PR #7763, opened 2020-09-28
Feature/lockfiles partial

### LINKED ISSUE #7739: [bug][question] Conan lockfile behavior change between conan 1.26 and conan 1.29.
We used conan.lock files from a CI build for a big set of comonents where all component versions were synchronized. Later developers used these lockfiles to build new versions of one of those components in the context of that lockfile locally, so they were not affected by new versions of upstreams that some other developer might have created in the meantime.

Simple example:
CA/1.0.0 <- CB/1.0.0  (read CB uses CA)

This combination was built in a CI job and the lockfile was generated. Then someone created a CA/2.0.0 package.
Now another developer works on a CB/2.0.0 package, but still wants to work in the context of the last successfull CI build, so he uses the lockfile and creates a CB/2.0.0 locally on his machine, but still wants to use CA/1.0.0 (which was locked in the CI lockfile).

This scenario used to work perfectly with conan 1.26, but fails with conan 1.29 with the error message: 
`ERROR: Couldn't find 'CB/2.0.0@local/testing' in lockfile`.

We need an alternative for the old behavior with the new conan version, but are at a loss how to do that, as lockfiles have become quite restrictive about building new versions of packages that were not previously locked.

I have attached a simple python script, that reproduces the case explained above. The script works with conan 1.26, but fails with the above error message with conan 1.29.

(Of course, our real word scenario involves a complete tree of sometimes dozens of upstream dependencies with potentially several version conflicts, not just one upstream package. That is why we need a way to fix the local environment to a known-good state until we push the new version of CB to the CI which then builds a new consistent state of all components.)

memsharded:
Changelog: Fix: Removed check in lockfiles computed from other lockfile that it should be part of it. Users can check the resulting lockfile themselves if they want to.
Docs: https://github.com/conan-io/docs/pull/1868

Fix https://github.com/conan-io/conan/issues/7739


conans/test/functional/graph_lock/dynamic_test.py
```diff
@@ -254,3 +256,35 @@ def augment_test_package_requires(self):
...
+        client.run("create . LibB/1.0@")
+        client.save({"conanfile.py": GenConanfile().with_require("LibB/[>=1.0]")})
+        client.run("create . LibC/1.0@")
+        client.run("lock create --reference=LibC/1.0 --lockfile-out=libc.lock")
+
+        # New version of LibA/1.0.1, that should never be used
+        client.save({"conanfile.py": GenConanfile()})
+        client.run("create . LibA/1.0.1@")
+
+        # Go back to B, we want to develop but keep depending on LibA/1.0.0
```
jgsogo:
```suggestion
        # Go back to B, we want to develop but keep depending on LibA/1.0
```

jgsogo (approved):
I agree it is ok to relax the condition here. As long as the nodes that are present in both lockfiles are the same (same package-id) I think it is ok to build the new lockfile.

I can think about other development scenarios where this can be useful: two teams that develop different libraries that will be consumed by the same application, both teams want to share a _master_ lockfile with all the dependencies to ensure they are using the same packages.

PR #7738, opened 2020-09-22
Feature/toolchain cmake simplify

memsharded:
Changelog: Omit
Docs: Omit

Improving the CMake toolchain, flattening template and integrating the code in the toolchain.

PR #7765, opened 2020-09-28
move values to the constructor of CMakeToolchain

memsharded:
Changelog: Omit
Docs: Omit

Follow up from https://github.com/conan-io/conan/pull/7738, moving things to constructor, so recipes could check and modify them as necessary.


solvingj (approved):
Seems straightforward.

PR #7776, opened 2020-09-30
Feature/detect libcxx

### LINKED ISSUE #7264: [bug] compiler.libcxx for dependencies with prebuilt packages
Source issue: https://github.com/cryfs/cryfs/issues/336

I'm using [conan_cmake_run](https://github.com/conan-io/cmake-conan) to call conan from cmake and pass in the argument `SETTINGS compiler.libcxx=libstdc++11`. This calls `conan install` with `-s compiler.libcxx=libstdc++11` before running the rest of the CMakeLists.txt because my application can only work with libstdc++11.

I noticed that the build works successfully when specifying `-DCMAKE_BUILD_TYPE=RelWithDebInfo` but runs into gcc ABI linking errors to one of the dependencies (spdlog) when specifying `-DCMAKE_BUILD_TYPE=Release`.

After some debugging, I noticed that `-DCMAKE_BUILD_TYPE=Release` uses a prebuilt package of spdlog from conan servers while `-DCMAKE_BUILD_TYPE=RelWithDebInfo` builds the dependency locally, possibly because there is no prebuilt package for `RelWithDebInfo`. Is it possible that it somehow downloads a package that was built with `libstdc++` instead of `libstdc++11` even though I set the global setting to `libstdc++11`?

If that analysis is correct and the `compiler.libcxx` setting is not considered when searching for prebuilt binaries, then that should probably be fixed by adding it to the metadata for prebuilt packages and only downloading and using a prebuilt package if the `compiler.libcxx` setting matches.

  * Operating System+version: CentOS 7
  * Compiler+version: devtoolset-7 (gcc 7)
  * Conan version: 1.26.1
  * Python version: 3.6

1. Start fresh CentOS 7 docker container and install requirements
```
$ sudo docker run -ti centos:centos7 /bin/bash
[root@a0098fc49d53]# yum install git wget make centos-release-scl python3 libcurl-devel fuse-devel
[root@a0098fc49d53]# yum install devtoolset-7
[root@a0098fc49d53]# scl enable devtoolset-7 bash
[root@a0098fc49d53]# pip3 install conan
[root@a0098fc49d53]# cd ~
[root@a0098fc49d53]# wget https://github.com/Kitware/CMake/releases/download/v3.14.6/cmake-3.14.6-Linux-x86_64.sh
[root@a0098fc49d53]# chmod +x cmake-3.14.6-Linux-x86_64.sh
[root@a0098fc49d53]# ./cmake-3.14.6-Linux-x86_64.sh
[root@a0098fc49d53]# export PATH=$PATH:/root/bin
```
2. Inside the docker container, build CryFS with `-DCMAKE_BUILD_TYPE=Release`
```
[root@a0098fc49d53]# git clone https://github.com/cryfs/cryfs
[root@a0098fc49d53]# git checkout f179b8c4a433a4ebf893fe43d82fd3bd2c4fe513
[root@a0098fc49d53]# cd cryfs
[root@a0098fc49d53]# mkdir build
[root@a0098fc49d53]# cd build
[root@a0098fc49d53]# cmake .. -DCMAKE_BUILD_TYPE=Release
[root@a0098fc49d53]# make
```
3. Notice that it fails.
4. Try the same thing with `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and notice that it succeeds.

Run with `-DCMAKE_BUILD_TYPE=Release` fails: https://pastebin.com/wMuQAEdL
Run with `-DCMAKE_BUILD_TYPE=RelWithDebInfo` succeeds: https://pastebin.com/979iSsxp

memsharded:
Changelog: Feature: Implement real detection of ``compiler.libcxx`` value for ``gcc`` compiler. Only enabled in `CONAN_V2_MODE`, otherwise it would be breaking.
Docs: Omit

Close https://github.com/conan-io/conan/issues/7264

### DIFF
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -228,7 +228,7 @@ def _detect_compiler_version(result, output, profile_path):
             result.append(("compiler.libcxx", "libc++"))
         elif compiler == "gcc":
             profile_name = os.path.basename(profile_path)
-            libcxx = _detect_gcc_libcxx("g++", version, profile_name, profile_path)
+            libcxx = _detect_gcc_libcxx("g++", version, output, profile_name, profile_path)
             result.append(("compiler.libcxx", libcxx))
         elif compiler == "cc":
             if platform.system() == "SunOS":
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -1,6 +1,8 @@
 import os
 import platform
 import re
+import tempfile
+import textwrap
 
 from conans.client.build.compiler_id import UNKNOWN_COMPILER, LLVM_GCC, detect_compiler_id
 from conans.client.output import Color
@@ -9,6 +11,7 @@ from conans.client.tools.win import latest_visual_studio_version_installed
 from conans.model.version import Version
 from conans.util.conan_v2_mode import CONAN_V2_MODE_ENVVAR
 from conans.util.env_reader import get_env
+from conans.util.files import save
 from conans.util.runners import detect_runner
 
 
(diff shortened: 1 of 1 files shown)


conans/client/conf/detect.py
```diff
@@ -161,6 +164,55 @@ def _get_profile_compiler_version(compiler, version, output):
...
+                       "***********************\n %s\n************************************"
+                       "************************************************\n\n\n" % msg,
+                       Color.BRIGHT_RED)
+        return "libstdc++"
+
+    main = textwrap.dedent("""
+        #include <string>
+
+        using namespace std;
+        static_assert(sizeof(std::string) != sizeof(void*), "using libstdc++");
```
SSE4, line 195: static_assert is C++11, can we use it?

SSE4: maybe we can just include <string> and check the value of `_GLIBCXX_USE_CXX11_ABI`?

memsharded: It is only applied for gcc>=5.1 which already support c++11 (passed in the command line)

> maybe we can just include and check the value of _GLIBCXX_USE_CXX11_ABI?

That is the original problem, as you described in https://github.com/conan-io/conan/issues/7264#issuecomment-651107547, this will be ignored. We need to check for the real thing.

SSE4: oh, I see, you're passing `-std=c++11` below

czoido (approved):

conans/client/conf/detect.py
```diff
@@ -161,6 +164,55 @@ def _get_profile_compiler_version(compiler, version, output):
...
+    os.chdir(t)
+    try:
+        error, output = detect_runner("%s main.cpp -std=c++11" % executable)
+        if error:
+            if "using libstdc++" in output:
+                return "libstdc++"
+            # Other error, but can't know, lets keep libstdc++11
+            output.warn("compiler.libcxx check error: %s" % output)
+            output.warn("Couldn't deduce compiler.libcxx for gcc>=5.1, assuming libstdc++11")
+        return "libstdc++11"
```
uilianries, line 211: I would add an info message about which libcxx will be used.

uilianries:
I can't see the warning message when running it on CentOS6 + Devtools7:

```
python conans/conan.py profile new default --detect
CC and CXX: /opt/rh/devtoolset-7/root/usr/bin/gcc, /opt/rh/devtoolset-7/root/usr/bin/g++ 
Found gcc 7.3
gcc>=5, using the major as version
Profile created with detected settings: /home/conan/.conan/profiles/default
[conan@eb84f8fab0db project]$ python conans/conan.py profile show default
Configuration for profile default:

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=7
compiler.libcxx=libstdc++
build_type=Release
[options]
[build_requires]
[env]
```

For regular GCC7 docker image, based on Ubuntu, it works fine:

```
$ python conans/conan.py profile new default --detect 
CC and CXX: /usr/bin/gcc, /usr/bin/g++ 
Found gcc 7.2
gcc>=5, using the major as version
Profile created with detected settings: /home/conan/.conan/profiles/default
conan@f5cb86db1e8f:~/project$ python conans/conan.py profile show default          
Configuration for profile default:

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=7
compiler.libcxx=libstdc++11
build_type=Release
[options]
[build_requires]
[env]
```

#### Steps to reproduce:

    $ cd <conan_project_folder>
    $ docker run --rm -ti -v ${PWD}:/home/conan/project conanio/gcc7-centos6:latest
    $ cd project
    $ export CONAN_V2_MODE=1
    $ export PYTHONPATH=${PWD}
    $ python conans/conan.py profile new default --detect
    $ python conans/conan.py profile show default

conans/client/conf/detect.py
memsharded: Agree, will add it to a refactor later

memsharded:
> I can't see the warning message when running it on CentOS6 + Devtools7:

This will not warn if it checks the libcxx correctly, it will just assign the correct value, and seems to work ok, doesn't it?

uilianries:
> This will not warn if it checks the libcxx correctly, it will just assign the correct value, and seems to work ok, doesn't it?

Hmm, correct!

PR #7777, opened 2020-09-30
refactor compiler detect

memsharded:
Changelog: Omit
Docs: Omit

Refactor following https://github.com/conan-io/conan/pull/7776
Adding some messages of the resulting compiler.libcxx


conans/client/conf/detect.py
```diff
@@ -204,10 +203,13 @@ def _detect_gcc_libcxx(executable, version, output, profile_name, profile_path):
         error, output = detect_runner("%s main.cpp -std=c++11" % executable)
         if error:
             if "using libstdc++" in output:
+                output.info("gcc C++ standard library: libstdc++")
```
uilianries, line 206: You have to use another variable here, output is overridden at 203, now it's a string, not an output object.

memsharded: Good catch, fixed.

PR #7779, opened 2020-09-30
Install a list of system packages 

### LINKED ISSUE #6960: [bug] SystemPackageTool fails to install package lists
  * Compiler+version: gcc 7.0
  * Conan version: 1.24.1
  * Python version: 3.6

I have this conan file
 ```
        self.global_system_requirements=True
        installer = tools.SystemPackageTool()
        libs = "libgtk2.0-dev libavcodec-dev libavresample-dev libv4l-dev libzbar-dev libavformat-dev libswscale-dev python3-dev python3-numpy libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev ffmpeg libvtk7-dev liblapack-dev openjdk-11-jdk"
        installer.install(libs)
```
When I run conan create . I get the following output:
```
dpkg-query: no packages found matching libgtk2.0-dev
dpkg-query: no packages found matching libavcodec-dev
dpkg-query: no packages found matching libavresample-dev
dpkg-query: no packages found matching libv4l-dev
dpkg-query: no packages found matching libzbar-dev
dpkg-query: no packages found matching libavformat-dev
dpkg-query: no packages found matching libswscale-dev
dpkg-query: no packages found matching libtbb2
dpkg-query: no packages found matching libtbb-dev
dpkg-query: no packages found matching libjpeg-dev
dpkg-query: no packages found matching libpng-dev
dpkg-query: no packages found matching libtiff-dev
dpkg-query: no packages found matching libdc1394-22-dev
dpkg-query: no packages found matching ffmpeg
dpkg-query: no packages found matching libvtk7-dev
dpkg-query: no packages found matching liblapack-dev
dpkg-query: no packages found matching openjdk-11-jdk
Package already installed: libgtk2.0-dev libavcodec-dev libavresample-dev libv4l-dev libzbar-dev libavformat-dev libswscale-dev python3-dev python3-numpy libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev ffmpeg libvtk7-dev liblapack-dev openjdk-11-jdk
```
The system correctly detects that the packages are not installed but then it actually does not install them, because one of them it is.

### LINKED ISSUE #6961: [feature] Accept list for SystemPackageTool
The [SystemPackageTool](https://docs.conan.io/en/latest/reference/conanfile/methods.html#systempackagetool) class offers the method `install` which only supports a string with all packages to be installed. However, even a list of packages must be passed as a string, it's a bit confuse, and get worst if you need to check if it's installed or not before: https://github.com/conan-io/conan/issues/6960#issuecomment-624620287

It would be great accepting both string (backward) and list:

```
installer = SystemPackageTool()
installer.install(["pkg1", "pkg2", "pkg3"])
```

### PR DESCRIPTION
**Pull request by uilianries**
Support a list of package to be installed by SystemPackageTools. The current behavior only supports a single package, or its variants by a list. However, most system package managers accept a list of different names e.g. `apt install vim nano emacs` 

Changelog: Feature: System package tools can install a list of different packages.
Docs: https://github.com/conan-io/docs/pull/1937
fixes #6961
fixes https://github.com/conan-io/conan/issues/6960

### CODE REVIEW

**Comment by uilianries** [comment]
Hey, why pr-merge didn't restart?

**Comment by danimtb** [comment]
Probably an issue with the Ci regarding the log. cc/ @czoido

**Review by memsharded — APPROVED** [review]
The problem with this is that it implements the functionality correctly, but it would probably not solve the underlying issue. I expect that many users will still do the ``install(["nano vim emacs"])``, and will not read the documentation deeply enough to realize there is an opt-in ``all=True`` argument they need to pass if they want all packages installed.

The only way to achieve this is breaking the contract. We could do it gradually, and propose something like a new ``install_packages`` method, that will try to install all of them, and if some want to propose variants, it could do in a nested tuple or list. Or any other new interface, while we deprecate the old one.

I am approving, but will ask for further reviews.

The ``SystemPackageTool`` IMHO is a monster which probably needs a whole refactor.

**Comment by jgsogo** [comment]
The extra line in the [docs](https://github.com/conan-io/conan/issues/5347) is quite unfortunate so we cannot change without breaking

image

besides that, the examples in the docs are not taking advantage of that feature, and the [PR in CCI is not using it either](https://github.com/conan-io/conan-center-index/pull/904/files):

image

😞 😞 😞 

I agree it is more expected to install all the packages from the list than to return as soon as one of them is installed. I'm trying to think about a different implementation, but I cannot figure out how not to break existing behavior (yes, I'm not a big fan of adding extra arguments).

If we were 100% sure we will have `SystemPackageTools` in Conan v2.0, I would propose a different approach:

```python

   spt = SystemPackageTools(v2_behavior=True)  # CONAN_V2_MODE activates it by default
   spt.install('pkg_a')  # <-- installs the only one package

   # Breaking change activated by the `v2_behavior=True`)
   spt.install(['pkg_a', 'pkg_b', 'pkg_c'])  # <-- installs three different packages (this would be the new CONAN_V2 behavior)

   # Using a list of lists: If a list of alternatives is provided, Conan installs only one from each group
   spt.install([('pkg_a-alt1', 'pkg_a-alt2'), 'pkg_b', 'pkg_c' ])  # <-- installs one package from each group
   spt.install([('pkg_a-alt1', 'pkg_a-alt2'), ])  # <-- installs one package from each group
```

IMHO it is more natural: **every entry in the list installs one package, if alternatives are given for entry only one of them is installed**.

--- 

I'm not asking for these changes, just proposing the alternative and, if we agree with it, then we can change the PR.

memsharded:
Yes, I also like the syntax:

```python
spt.install([('pkg_a-alt1', 'pkg_a-alt2'), 'pkg_b', 'pkg_c' ])
```

But I wouldn't like to have the flag explicit here: ``SystemPackageTools(v2_behavior=True)``, as this will be very ugly in Conan 2.0. In any case, I would say a syntax that could remain with a new method, something like:

```python
spt = SystemPackageTools()
spt.install_pkgs(("alt1", "alt2", "alt3"), "pkg2", "pkg3")  # or install_system...
```

There is a chance if we replace the whole ``SystemPackageTools`` with a new alternative when we do the "tools" general refactor, and we take advantage of the change, but that would be just another extra effort, not sure it is worth.

jgsogo:
Adding an explicit flag or adding a dedicated method is more or less the same (IMHO it's more explicit the flag than the method, unless we find **a name other than `install` that we want for that method in the future**). Both the flag and the new method require to modify recipes now (if needed) and both alternatives can be marked as deprecated in Conan v2 and move the functionality to the preferred implementation (remove the flag, or back to `install` method).

Whatever is used it is ok for me, I'll review again when the PR is updated.

memsharded:
> Adding an explicit flag or adding a dedicated method is more or less the same (IMHO it's more explicit the flag than the method, unless we find a name other than install that we want for that method in the future). Both the flag and the new method require to modify recipes now (if needed) and both alternatives can be marked as deprecated in Conan v2 and move the functionality to the preferred implementation (remove the flag, or back to install method).

 I was thinking of a new method name that can **remain forever**, so it is only necessary to change it once, before Conan 2.0, but no necessary to deprecate again and move back after 2.0.

jgsogo:
`SystemPackageTool` already has some kind of cache and now we need to name a new method... 😄 

image

Yes, totally, if that is the method we will use forever, much better.

uilianries:
I vote for the new method name.  Why? When reading the documentation, method are empathized, using bold color, but for parameters you have to read to find any detail, so anyone who reads the docs, will find the new method name easily. For Conan 2.0, we can break and use only one method, the tuple for variants seems be more correct.

Screenshot_2020-10-14 Methods — conan 1 30 1 documentation

conans/test/unittests/client/tools/system_pm_test.py
```diff
@@ -351,7 +351,7 @@ def system_package_tool_test(self, patched_with_apt):
                                  'choco search --local-only --exact a_package | '
                                  'findstr /c:"1 packages installed."')
 
-    def system_package_tool_try_multiple_test(self):
+    def test_system_package_tool_try_multiple_test(self):
```
jgsogo:
```suggestion
    def test_system_package_tool_try_multiple(self):
```

jgsogo (commented):
If we go for a new method, I prefer the suggested `install_pkgs` over `install_all`. Taking into account that the method will last forever and it will be the only method in Conan 2.0 I feel like the suffix `_all` is a bit unexpected. wdyt?

uilianries:
I agree, but I prefer the full name `install_packages`, no abbreviations

conans/client/tools/system_pm.py
```diff
@@ -91,10 +91,10 @@ def update(self):
         self._is_up_to_date = True
         self._tool.update()
 
-    def install(self, packages, update=True, force=False, arch_names=None):
-        """ Get the system package tool install command.
+    def install(self, packages, update=True, force=False, arch_names=None, all=False):
```
memsharded: The ``all`` argument is not used anymore?

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,41 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+    def install_packages(self, packages, update=True, force=False, arch_names=None):
+        """ Get the system package tool install command and install all packages.
+
+        :param packages: String with all packages to be installed e.g. "libusb-dev libfoobar-dev"
+        :param update: Run update command before to install
+        :param force: Force installing all packages
+        :param arch_names: Package suffix/prefix name used by installer tool e.g. {"x86_64": "amd64"}
+        :return: None
+        """
+        packages = [packages] if isinstance(packages, str) else list(packages)
```
memsharded: Wouldn't it make sense to rely on the existing implementation? (we could throw warnings controlled by an undocumented argument). This would also automatically allow the syntax that @jgsogo was proposing for
```python
spt.install([("pkg-variant1", "pkg-variant2"), "otherpkg", "thirdpkg"])
```

That would also probably help with error messages: if some package is not installed now, the ``verify`` mode will raise an error, but it will not tell you which package is missing.

jgsogo: I think it would be enough with:

```python
def install_packages(self,...):
   packages = [packages] if isinstance(packages, str) else list(packages)
   for pkg in packages:
      self.install(pkg, update=update, force=force, arch_names=arch_names)
``` 

and it will work with:
```python
spt.install_packages("pkg-variant1")  # (1) String input
spt.install_packages(["pkg-variant1", "otherpkg", "thirdpkg"])  # (2) List of strings: all must be installed
spt.install_packages([("pkg-variant1", "pkg-variant2"), "otherpkg", "thirdpkg"])  # (3) List with alternatives
```

The old method cannot implement (2) without an additional argument to disambiguate if it has to install _all_ packages or _any_ of them.

lasote: Agree.

lasote (changes requested):
Agree with the previous comments, add a new test for the (2).

conans/client/tools/system_pm.py
uilianries: > Wouldn't it make sense to rely on the existing implementation? (we could throw warnings controlled by an undocumented argument). This would also automatically allow the syntax that @jgsogo was proposing for

I see a breaking behavior, for now, if we pass a list of packages, only one is installed. By passing a new argument to execute a different behavior looks a prone error for me, that's why a preferred a new method.

> I think it would be enough with:
I want to avoid multiple executions of the same command, that's why I "duplicated" the implementation. Any system package manager that we support, can install more than one package by a single command line. I don't think is a good idea re-run for each package name.

uilianries:
This new commit is able to support (1) (2) (3) options:

> spt.install_packages("pkg-variant1")  # (1) String input
> spt.install_packages(["pkg-variant1", "otherpkg", "thirdpkg"])  # (2) List of strings: all must be installed
> spt.install_packages([("pkg-variant1", "pkg-variant2"), "otherpkg", "thirdpkg"])  # (3) List with alternatives

I still use a different when installing multiple packages, because it is faster install all packages at once.

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,49 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+        :param packages: String/List/Tuple with all packages to be installed e.g. "libusb-dev libfoobar-dev"
+        :param update: Run update command before to install
+        :param force: Force installing all packages
+        :param arch_names: Package suffix/prefix name used by installer tool e.g. {"x86_64": "amd64"}
+        :return: None
+        """
+        packages = [packages] if isinstance(packages, str) else list(packages)
+        variants = list(filter(lambda x: isinstance(x, tuple), packages))
+        for variant in variants:
+            self.install(variant)
```
jgsogo:
```suggestion
            self.install(variant, update=update, force=update, arch_names=arch_names)
```

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,49 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+            [("pkg-variant1", "pkg-variant2"), "otherpkg", "thirdpkg"] # (3) List with alternatives
+
+        :param packages: String/List/Tuple with all packages to be installed e.g. "libusb-dev libfoobar-dev"
+        :param update: Run update command before to install
+        :param force: Force installing all packages
+        :param arch_names: Package suffix/prefix name used by installer tool e.g. {"x86_64": "amd64"}
+        :return: None
+        """
+        packages = [packages] if isinstance(packages, str) else list(packages)
+        variants = list(filter(lambda x: isinstance(x, tuple), packages))
```
jgsogo: Do we want to consider here a `list` too?

```python
package_install([['pkg1-var1', 'pkg1-var2'], 'pkg2'])

File: conans/client/tools/system_pm.py [hunk 6]
```diff
@@ -126,6 +126,49 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
         self._install_any(packages)
 
+    def install_packages(self, packages, update=True, force=False, arch_names=None):
+        """ Get the system package tool install command and install all packages or variants.
+            Inputs:
+            "pkg-variant1"  # (1) String input: Install only one package
+            ["pkg-variant1", "otherpkg", "thirdpkg"] # (2) List of strings: all must be installed
+            [("pkg-variant1", "pkg-variant2"), "otherpkg", "thirdpkg"] # (3) List with alternatives
+
+        :param packages: String/List/Tuple with all packages to be installed e.g. "libusb-dev libfoobar-dev"
```
jgsogo: If we are accepting the syntax `install_packages("pkg1 pkg2")`, please add it to the docstring.

But, IMHO, we should avoid that syntax because it can be confusing: what does it mean?
 * will it install `pkg1` and `pkg2` or just one of them?
 * what if I use `install_packages(["pkg1 pkg2", "pkg3 pkg4"])`?

conans/client/tools/system_pm.py
uilianries: That's a good question. I think only tuple keep more explicit, but on the other hand, people need to read to documentation to understand that only tuple is allowed. I think accepting list should not be a problem, I'll update it.

conans/client/tools/system_pm.py
uilianries: > If we are accepting the syntax install_packages("pkg1 pkg2"), please add it to the docstring.

Sure, indeed the docstring is outdated since the last commit.

> what if I use install_packages(["pkg1 pkg2", "pkg3 pkg4"])?

I would like to avoid it, but is totally valid for the current behavior, it will install all packages.

uilianries:
New change: `install_packages` installs only what is not installed on the system, for example, if pass ["vim", "wget", "curl"], but "wget" is already installed, it will install only "vim" and "curl". However, if `force` is True, it will install all packages listed, including "wget" again.

Why? It's the natural behavior when installing packages when we don't force. Also, it's faster, as the system don't need to process and print "package xyz already installed".

conans/client/tools/system_pm.py
jgsogo: All packages? pkg1, pkg2, pkg3 and pkg4? It doesn't match the expectations when you use a _list of tuples_...

uilianries: `install_packages(["pkg1 pkg2", "pkg3 pkg4"])` will install all packages pkg1, pkg2, pkg3 and pkg4, yes. Variants should be a nested list or a tuple in a list, like:

`install_packages([["pkg1 pkg2"], "pkg3 pkg4"])`

jgsogo: IMHO, it is very confusing if `install_packages(["pkg1 pkg2", "pkg3 pkg4"])` installs all packages, but `install_packages([("pkg1", "pkg2"), ("pkg3", "pkg4")])` install only some of them. I'd totally forbid (raise) the syntax with blank-separated packages (`pkg1 pkg2`).

...but this is my point of view, what do other reviewers think about it? cc/ @memsharded

memsharded: I agree with @jgsogo 

Space separated list introduce another way to do the same thing, but the semantics are now not clear and difficult to dissambiguate. I think it is better to have the following logic:

- Each argument could be either a string, or a list/tuple.
- If a string, that is a package name to install
- If a list/tuple, that is a set of alternatives, install the first successful one.

uilianries: > If a string, that is a package name to install

Considering the previous case, it's a string and contains more than one package. It's totally valid for install(), it will install a variant, but only because install() only installs variants, not all packages. I'll update to set is as invalid, but it won't follow the same design as did for install().

uilianries: Done, now ["foo bar"] and "foo bar" are considered invalid. Actually doesn't matter if you pass a string or list, it's convert to a list at the first step.

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,64 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+        :param update: Run update command before to install
+        :param force: Force installing all packages, including all installed.
+        :param arch_names: Package suffix/prefix name used by installer tool e.g. {"x86_64": "amd64"}
+        :return: None
+        """
+        packages = [packages] if isinstance(packages, str) else list(packages)
+        # only one (first) variant will be installed
+        variants = list(filter(lambda x: isinstance(x, (tuple, list)), packages))
+        for variant in variants:
+            self.install(variant, update=update, force=force, arch_names=arch_names)
```
uilianries:
```suggestion
        list_variants = list(filter(lambda x: isinstance(x, (tuple, list)), packages))
        for variants in list_variants:
            self.install(variants, update=update, force=force, arch_names=arch_names)
```

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,64 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+
+        :param packages: Supports multiple formats (string,list,tuple). Lists and tuples into a list
+        are considered variants and is processed just like self.install(). A list of string is
+        considered a list of packages to be installed (only not installed yet).
+        :param update: Run update command before to install
+        :param force: Force installing all packages, including all installed.
+        :param arch_names: Package suffix/prefix name used by installer tool e.g. {"x86_64": "amd64"}
+        :return: None
+        """
+        packages = [packages] if isinstance(packages, str) else list(packages)
```
jgsogo: Unicode (py2) not expected here, but just in case.

```suggestion
        packages = [packages] if isinstance(packages, six.string_types) else list(packages)
```

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,64 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+        # only one (first) variant will be installed
+        list_variants = list(filter(lambda x: isinstance(x, (tuple, list)), packages))
+        for variants in list_variants:
+            self.install(variants, update=update, force=force, arch_names=arch_names)
+        # all packages will be installed
+        packages = list(filter(lambda x: not isinstance(x, (tuple, list)), packages))
+
+        if [pkg for pkg in packages if " " in pkg]:
+            raise ConanException("Each string must contain only one package to be installed. "
+                                 "Use a list instead e.g. ['foo', 'bar'].")
```
jgsogo, line 156: If we are going to raise, better to raise before running any install command. 

We can compute `list_variants` and `packages` at the very beginning. Then check for blanks in any of the strings (not only for the `packages` variable). We raise soon if needed, otherwise, we start installing packages.

conans/client/tools/system_pm.py
```diff
@@ -138,16 +196,31 @@ def _get_package_names(self, packages, arch_names):
...
         return packages
 
     def installed(self, package_name):
         return self._tool.installed(package_name)
 
+    def _to_be_installed(self, packages):
+        """ Returns a list with all not installed packages.
+        """
+        for pkg in packages:
+            if isinstance(pkg, (list, tuple)):
```
jgsogo: According to the filter in the function `install_packages`, every `pkg` is a string, never a list. No need for this check.

conans/client/tools/system_pm.py
```diff
@@ -138,16 +196,31 @@ def _get_package_names(self, packages, arch_names):
...
+                if not self._to_be_installed(pkg):
+                    return list(pkg)
+        installed = {pkg: self._installed([pkg]) for pkg in packages}
+        return [] if all(installed.values()) else [pkg for pkg, i in installed.items() if not i]
```
jgsogo:
```suggestion
        not_installed = [pkg for pkg in packages if not self._tool.installed(pkg)]
        return not_installed
```

conans/client/tools/system_pm.py
```diff
@@ -126,6 +126,64 @@ def install(self, packages, update=True, force=False, arch_names=None):
...
+                self._output.error("The following packages need to be installed:\n %s"
+                                   % "\n".join(packages))
+                raise ConanException("Aborted due to CONAN_SYSREQUIRES_MODE=%s. "
+                                     "Some system packages need to be installed" % mode)
+
+        to_be_installed = self._to_be_installed(packages)
+        if not force and not to_be_installed:
+            return
+        # install only what is not installed yet
+        packages = packages if force else to_be_installed
```
jgsogo: Do not run all the operations in `self._to_be_installed` if we are going to `force` the installation

```suggestion
        packages = packages if force else self._to_be_installed(packages)
        if not force and not packages:
            return
```

jgsogo (changes requested):

conans/client/tools/system_pm.py
uilianries: Done.

conans/client/tools/system_pm.py
uilianries: Done.

conans/client/tools/system_pm.py
uilianries: Done.

conans/client/tools/system_pm.py
uilianries: Done.

conans/client/tools/system_pm.py
uilianries: Done.

conans/test/unittests/client/tools/system_pm_test.py
```diff
@@ -526,3 +517,144 @@ def get_linux_error_message():
...
+            spt = SystemPackageTool(runner=runner, tool=AptTool(output=self.out),
+                                    output=self.out)
+            spt.install_packages(packages)
+            self.assertEqual(0, runner.calls)
+
+    def test_install_all_multiple_package_list(self):
+        """ Separated string list must be treated as full package list to be installed
+        """
+        packages = "varianta variantb", "variantc variantd"
+        with tools.environment_append({"CONAN_SYSREQUIRES_SUDO": "True"}):
```
jgsogo: Is this needed here?

jgsogo (commented):
Looking good. Let's finish this feature and then we can refactor this class a little bit

conans/test/unittests/client/tools/system_pm_test.py
uilianries: No, it's a copy-past error. gonna fix it.

jgsogo (approved):

uilianries:
One approval! Maybe it will be merged before US election ending!

memsharded (approved):

PR #7781, opened 2020-09-30
[bugfix] multiple remotes with single conan upload

### LINKED ISSUE #7780: [bug] Upload fails with multiple remotes in the same command
When uploading several packages at once (without specifying a remote) and some packages are uploaded to different remotes, the command will fail.

The bug appears to be due to reusing a closed thread pool in the loop in CmdUploader.upload() in uploader.py. Once thread pools are closed they can't be reused, so any iterations after the first will cause an error.

  * Operating System+version: RHEL 7.6
  * Conan version: Tested with 1.29.0 and 1.30.dev0 (develop)
  * Python version: 3.8.5

1. Set up two remotes
2. Create 2 recipes and add_ref one recipe to each remote
4. Upload using `conan upload "*" --all`

I can't get the logs off my machine, but the stack trace looks like this:
command.py:2103
command.py:1477
conan_api.py:94
conan_api.py:965
uploader.py:109
pool.py:364
pool.py:473
pool.py:350
ValueError: Pool not running

kralicky:
Changelog: Bugfix: Fixed bug where uploading to multiple remotes in a single conan upload command would fail.
Docs: omit

Fixes #7780 

This bug was due to closing and reusing a thread pool in a loop. Instead, a separate thread pool will be created at each loop iteration.

### DIFF
--- a/conans/client/cmd/uploader.py
+++ b/conans/client/cmd/uploader.py
@@ -201,12 +201,11 @@ class CmdUpload(object):
                                            all_packages, query)
 
         if parallel_upload:
-            self._upload_thread_pool = ThreadPool(8)
             self._user_io.disable_input()
-        else:
-            self._upload_thread_pool = ThreadPool(1)
 
         for remote, refs in refs_by_remote.items():
+            self._upload_thread_pool = ThreadPool(8 if parallel_upload else 1)
+
             self._output.info("Uploading to remote '{}':".format(remote.name))
 
             def upload_ref(ref_conanfile_prefs):
--- a/conans/client/cmd/uploader.py
+++ b/conans/client/cmd/uploader.py
@@ -23,6 +23,7 @@ from conans.util.log import logger
 from conans.util.tracer import log_recipe_upload, log_compressed_files, log_package_upload
 from conans.tools import cpu_count
 
+
 UPLOAD_POLICY_FORCE = "force-upload"
 UPLOAD_POLICY_NO_OVERWRITE = "no-overwrite"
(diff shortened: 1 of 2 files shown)


conans/client/cmd/uploader.py
```diff
@@ -201,12 +201,11 @@ def upload(self, reference_or_pattern, remotes, upload_recorder, package_id=None
                                            all_packages, query)
 
         if parallel_upload:
-            self._upload_thread_pool = ThreadPool(8)
             self._user_io.disable_input()
-        else:
-            self._upload_thread_pool = ThreadPool(1)
 
         for remote, refs in refs_by_remote.items():
+            self._upload_thread_pool = ThreadPool(8 if parallel_upload else 1)
```
kralicky: Maybe this should use tools.cpu_count()?

memsharded: This doesn't look right either, as the ``self._upload_thread_pool`` variable will be overwritten, losing access to the other pool from the previous remote.

memsharded (commented):
Thanks for submitting this PR.

I would like to understand better why it is failing in the first case, I cannot see an evident reason, so I'd need to try it a bit first.

conans/client/cmd/uploader.py
kralicky: Each iteration of this loop will complete before the next starts, since map() will block, then it will close the thread pool and start the next iteration.

kralicky: Wait, we can definitely reuse the thread pool here. Keeping it alive the whole time makes a lot more sense.

memsharded (approved):
Great! Now I understand the issue, it is totally clear.
Thanks very much for contributing this.

Issue #7800, opened 2020-10-02
[bug] Error during conan export-pkg when using md5 with FIPS


MichaelHCline:
I am getting the following error message when doing a conan export-pkg command (see traceback):

**ERROR: [digital envelope routines: EVP_DigestInit_ex] disabled for fips**

My company IT department did some updates on the system yesterday that brought about this problem.  (I am unsure which mod exactly broke it).  I can get conan to work again by making the following changes:

"/usr/local/lib/python3.6/site-packages/conans/util/files.py": line 126:
Changed:
md5alg = hashlibmd5()
To:
md5alg = hashlibmd5(usedforsecurity=False)

"/usr/lib64/python3.6/hashlib.py": line 176
usedforsecurity = kwargs.pop(‘usedforsecurity’, True)
To:
usedforsecurity = kwargs.pop(‘usedforsecurity’, False)

I would need to do this on every machine that we use conan, and over time I may have to keep redoing it, so it doesnt seem like the right answer.

### Environment Details (include every applicable attribute)
  * Operating System+version:  uname -a = Linux **** 3.10.0-1127.13.1.el7.x86_64 #1 SMP Fri Jun 12 14:34:17 EDT 2020 x86_64 x86_64 x86_64 GNU/Linux
  * Compiler+version:  (n/a)  gcc 4.8.5
  * Conan version:  1.24.0
  * Python version:  3.6

### Steps to reproduce (Include if Applicable)
conan export-pkg .  --profile=myprofile --force

### Logs (Executed commands with output) (Include/Attach if Applicable)

cd .. && conan export-pkg .  --profile Linux/conanprofile --force
Exporting package recipe
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/conans/client/command.py", line 2002, in run
    method(args[0][1:])
  File "/usr/local/lib/python3.6/site-packages/conans/client/command.py", line 970, in export_pkg
    ignore_dirty=args.ignore_dirty)
  File "/usr/local/lib/python3.6/site-packages/conans/client/conan_api.py", line 89, in wrapper
    return f(api, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/conans/client/conan_api.py", line 420, in export_pkg
    graph_lock=graph_info.graph_lock, ignore_dirty=ignore_dirty)
  File "/usr/local/lib/python3.6/site-packages/conans/client/cmd/export.py", line 155, in cmd_export
[...]

PR #7807, opened 2020-10-04
fix md5 error in fips systems

memsharded:
Changelog: Fix: Allow usage of MD5 checksums in FIPS systems that would raise error otherwise.
Docs: Omit

Close https://github.com/conan-io/conan/issues/7800

### DIFF
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -127,7 +127,10 @@ def normalize(text):
 
 
 def md5(content):
-    md5alg = hashlib.md5()
+    try:
+        md5alg = hashlib.md5()
+    except ValueError:  # FIPS error https://github.com/conan-io/conan/issues/7800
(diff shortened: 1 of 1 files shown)


jgsogo (changes requested):
Is it a `TypeError`? maybe both are needed depending on Python version

```
⇒  python
Python 3.8.2 (default, May  5 2020, 00:20:36) 
[Clang 11.0.3 (clang-1103.0.32.59)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> def function(argument):
...    print("function")
... 
>>> function(invalid_arg=23)
  File "<stdin>", line 1, in <module>
TypeError: function() got an unexpected keyword argument 'invalid_arg'
>>>
```

jgsogo (approved):

czoido (approved):

Issue #7826, opened 2020-10-06
[bug] Auto-detection of CMAKE_SYSTEM_NAME not working for all cross-compilation scenarios


jasal82:
Conan has an auto-detection feature for `CMAKE_SYSTEM_NAME` which is enabled by default if a value is not set explicitly. I found the method `_cmake_cross_build_defines()` in `CMakeDefinitionsBuilder` which seems to perform that auto-detection:

```python
        # System name and system version
        if cmake_system_name is not True:  # String not empty
            definitions["CMAKE_SYSTEM_NAME"] = cmake_system_name
        else:  # detect if we are cross building and the system name and version
            if cross_building(self._conanfile):  # We are cross building
                if os_ != os_build:
                    if os_:  # the_os is the host (regular setting)
                        definitions["CMAKE_SYSTEM_NAME"] = {"Macos": "Darwin",
                                                            "iOS": "Darwin",
                                                            "tvOS": "Darwin",
                                                            "watchOS": "Darwin",
                                                            "Neutrino": "QNX"}.get(os_, os_)
                    else:
                        definitions["CMAKE_SYSTEM_NAME"] = "Generic"
```

So if `os_ == os_build` the variable is never set which is clearly wrong. A typical cross-toolchain for arm would have `os = Linux` and `os_build = Linux` but `arch` and `arch_build` are different. My suggestion would be to add another `else:` block on the `if os_ != os_build` level that also sets `CMAKE_SYSTEM_NAME` to `Generic`.

memsharded:
Just to be sure are you suggesting?

```python
         if cross_building(self._conanfile):  # We are cross building
                if os_ != os_build:
                    if os_:  # the_os is the host (regular setting)
                        definitions["CMAKE_SYSTEM_NAME"] = {"Macos": "Darwin",
                                                            "iOS": "Darwin",
                                                            "tvOS": "Darwin",
                                                            "watchOS": "Darwin",
                                                            "Neutrino": "QNX"}.get(os_, os_)
                    else:
                        definitions["CMAKE_SYSTEM_NAME"] = "Generic"
          else:
                definitions["CMAKE_SYSTEM_NAME"] = "Generic"
```

Or why not something like:

```python
         if cross_building(self._conanfile):  # We are cross building
                if os_ != os_build and arch != arch_build:
                    if os_:  # the_os is the host (regular setting)
                        definitions["CMAKE_SYSTEM_NAME"] = {"Macos": "Darwin",
[...]

PR #7827, opened 2020-10-06
cross build arch check in cmake build helper

memsharded:
Changelog: Bugfix: Check comparing the host and the build architecture to decide if cross building and set `CMAKE_SYSTEM_NAME` in the ``CMake`` build helper.
Docs: Omit

#tags: slow

Close https://github.com/conan-io/conan/issues/7826

### DIFF
--- a/conans/client/build/cmake_flags.py
+++ b/conans/client/build/cmake_flags.py
@@ -182,7 +182,7 @@ class CMakeDefinitionsBuilder(object):
         env_sn = {"False": False, "True": True, "": None}.get(env_sn, env_sn)
         cmake_system_name = env_sn or self._forced_cmake_system_name
 
-        os_build, _, _, _ = get_cross_building_settings(self._conanfile)
+        os_build, arch_build, _, _ = get_cross_building_settings(self._conanfile)
         compiler = self._ss("compiler")
         libcxx = self._ss("compiler.libcxx")
 
@@ -203,7 +203,7 @@ class CMakeDefinitionsBuilder(object):
             definitions["CMAKE_SYSTEM_NAME"] = cmake_system_name
         else:  # detect if we are cross building and the system name and version
             if cross_building(self._conanfile):  # We are cross building
-                if os_ != os_build:
+                if os_ != os_build and arch != arch_build:
                     if os_:  # the_os is the host (regular setting)
                         definitions["CMAKE_SYSTEM_NAME"] = {"Macos": "Darwin",
                                                             "iOS": "Darwin",
--- a/conans/client/build/cmake_flags.py
+++ b/conans/client/build/cmake_flags.py
@@ -182,7 +182,7 @@ class CMakeDefinitionsBuilder(object):
         env_sn = {"False": False, "True": True, "": None}.get(env_sn, env_sn)
         cmake_system_name = env_sn or self._forced_cmake_system_name
 
-        os_build, arch_build, _, _ = get_cross_building_settings(self._conanfile)
+        os_build, _, _, _ = get_cross_building_settings(self._conanfile)
         compiler = self._ss("compiler")
         libcxx = self._ss("compiler.libcxx")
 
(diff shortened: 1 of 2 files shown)


memsharded:
Please @jasal82 have a look.

I have had to change the logic, and only change to CMAKE_SYSTEM_NAME=generic when the host ``os`` is not defined (embedded or the like), otherwise it will be breaking, almost for sure in some platforms as Windows that will get confused by the CMAKE_SYSTEM_NAME, and might potentially be changing users doing Linux 32bits from 64bits.

What is your host ``os``?

conans/client/build/cmake_flags.py
```diff
@@ -204,14 +204,19 @@ def _cmake_cross_build_defines(self):
...
-                    if os_:  # the_os is the host (regular setting)
+                    if os_:
                         definitions["CMAKE_SYSTEM_NAME"] = {"Macos": "Darwin",
                                                             "iOS": "Darwin",
                                                             "tvOS": "Darwin",
                                                             "watchOS": "Darwin",
                                                             "Neutrino": "QNX"}.get(os_, os_)
                     else:
                         definitions["CMAKE_SYSTEM_NAME"] = "Generic"
+                elif not os_:
```
jgsogo: I would expect something like this here:

```python
if cross_building(self._conanfile):  # We are cross building
    if os_:
        # If we are cross-building, it doesn't matter if it is because OS or ARCH
        definitions["CMAKE_SYSTEM_NAME"] = {...}.get(os_, os_)
    else:
        definitions["CMAKE_SYSTEM_NAME"] = "Generic"  # To avoid the CMake warning for missing *.cmake file
```

I'm sure there is a reason why we didn't have this more straightforward if/else conditions... maybe it is encoded in some test, but a comment here would be nice.

---

This implementation won't satisfy the ARM cross-compilation: `os:host=Linux`, `os:build=Linux`, `arch:host=x86_64`, `arch:build=arm`. According to [CMake docs](https://cmake.org/cmake/help/v3.6/manual/cmake-toolchains.7.html#cross-compiling-for-linux), here we should generate something with:

```
set(CMAKE_SYSTEM_NAME Linux)
```

jgsogo (commented):
I need more info, why not a more straightforward if/else as suggested?

conans/client/build/cmake_flags.py
memsharded: cross-building is returning ``True`` for Windows Visual building to x86 for x64, but seems that adding ``CMAKE_SYSTEM_NAME Windows`` for that case would be problematic, as cmake not really consider it cross-building, as far as I understood. Trying to do something like the proposed solution, I got some tests failing.

I am not saying that what we have is correct, certainly it is incomplete and fragile/buggy, but the risk of large breakings is important here.

jgsogo: Ok, probably it is worth to add a `CONAN_V2_MODE` block with a more straightforward implementation (if we think it is better):

```python
if cross_compiling:
    if CONAN_V2_MODE:
        if os_ == os_build and _os == 'Windows':
            # Nothing here?
        else:
            definitions["CMAKE_SYSTEM_NAME"] = {..., None, 'Generic'}.get(os_, os_)
    else:
        # Current logic plus patches.
        ...
```

I'll wait for answer [here](https://github.com/conan-io/conan/issues/7826#issuecomment-704861201), I really want to know why setting it to `Generic` instead of the actual value.

memsharded: This functionality is going to be moved anyway to toolchains, so CONAN_V2_MODE would probably not be necessary there.

jgsogo:
After the last answer in #7826 , I think we can add a test with ARM cross-compilation: `os:host=Linux`, `os:build=Linux`, `arch:host=x86_64`, `arch:build=arm` and check the resulting CMAKE_SYSTEM_NAME... and see if it doesn't break anything.

jgsogo (changes requested):
I think this is the way to go (and we should remove the Windows exception too), but I'm afraid this can break existing solutions when _cross compiling_ from Linux-x64 to Linux-x86.

Setting `CMAKE_SYSTEM_NAME` will make CMake set `CMAKE_CROSSCOMPILING` variable and it will have other implications: https://gitlab.kitware.com/search?utf8=%E2%9C%93&snippets=false&scope=&repository_ref=master&search=CMAKE_CROSSCOMPILING&group_id=415&project_id=541 

I think that we should keep working the same for that scenario: Linux x64 to x86. Something like this:

```python
skip_x64_x86 = conanfile.settings.get_safe('os') in ['Windows', 'Linux']
if cross_building(self._conanfile, skip_x64_x86=skip_x64_x86):
    ...
    definitions["CMAKE_SYSTEM_NAME"] = cmake_system_name_map.get(os_, os_)
```

I've realized about the implications after reading about `CMAKE_CROSSCOMPILING` 😞

conans/test/unittests/client/build/cmake_test.py
```diff
@@ -629,12 +629,17 @@ def loads_default_test(self):
...
             for cmake_system_name in (True, False):
                 cross_ver = ("-DCMAKE_SYSTEM_VERSION=\"%s\" " % os_ver) if os_ver else ""
-                cross = ("-DCMAKE_SYSTEM_NAME=\"%s\" %s-DCMAKE_SYSROOT=\"/path/to/sysroot\" "
-                         % ({"Macos": "Darwin"}.get(the_os, the_os), cross_ver)
-                         if (platform.system() != the_os and cmake_system_name) else "")
+                # FIXME: This test is complicated to maintain and see the logic, lets simplify it
+                cross = ""
+                if cmake_system_name and (the_os != platform.system() or
+                                          (arch != detected_architecture()
+                                           and the_os not in ("Windows", "Linux"))):
```
jgsogo: Is it necessary to add extra conditions here?

jgsogo (approved):
I'll help you to pay the 🍔  if needed, I really think this is the desired behavior (keeping the exceptional one about Win/Linux exception)

jgsogo (approved):

PR #7840, opened 2020-10-07
Feature/6207 config single file

### LINKED ISSUE #6207: [feature] conan config install a single file
When using `conan config install`, allow a single configuration file to be installed. 

Normally, the command requires a git repository, a local folder, or a zip file that contains the file(s) to be installed. If I have a single file I want installed, I would have to put it into a folder or zip it into a file.

The current result:
```
conan config install my_settings\settings.yml
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\command.py", line 1859, in run
    method(args[0][1:])
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\command.py", line 554, in config
    target_folder=args.target_folder)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\conan_api.py", line 78, in wrapper
    return f(*args, **kwargs)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\conan_api.py", line 611, in config_install
    source_folder=source_folder, target_folder=target_folder)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\conf\config_installer.py", line 235, in configuration_install
    _process_config(config, cache, output, requester)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\conf\config_installer.py", line 190, in _process_config
    _process_zip_file(config, config.uri, cache, output, tmp_folder)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\conf\config_installer.py", line 63, in _process_zip_file
    unzip(zippath, tmp_folder, output=output)
  File "C:\Users\Kyle.Kaja\AppData\Roaming\Python\Python37\site-packages\conans\client\tools\files.py", line 104, in unzip
    with zipfile.ZipFile(filename, "r") as z:
  File "C:\Program Files\Python37\lib\zipfile.py", line 1225, in __init__
    self._RealGetContents()
  File "C:\Program Files\Python37\lib\zipfile.py", line 1292, in _RealGetContents
    raise BadZipFile("File is not a zip file")
zipfile.BadZipFile: File is not a zip file

ERROR: File is not a zip file
```

This feature is useful in the case where you only want to install specific files from a single folder.
For example,
```
my_settings\
  settings.yml
  remotes.txt

conan config install my_settings  (Installs settings.yml and remotes.txt)
conan config install my_settings\settings.yml  (Installs settings.yml and ignores remotes.txt)
```

It is also useful when the current folder structure does not facilitate creating a folder specifically for configuration files (e.g. a git repository).
```
my_git_repo\
  settings.yml
  unrelated.txt
  ...

conan config install .\
Installing settings.yml
Copying file unrelated.txt to C:\Users\Kyle.Kaja\.conan\.  (I don't want this installed here!!)
...

OR

mkdir temp_settings
copy settings.yml temp_settings
conan config install temp_settings
Installing settings.yml
rmdir temp_settings /s  (Remove the directory so Git doesn't try to track that new file.)
(So much to do when all I want to do is install a single configuration file.)

PREFERRED

conan config install settings.yml  (Ahh! Simple, short, and intuitive!)
Installing settings.yml
```

sagarafr:
Changelog: Feature: Allow ``conan config install`` of a single file
Docs: https://github.com/conan-io/docs/pull/1908

Close https://github.com/conan-io/conan/issues/6207

* Add the possibility to specify a configuration for the command `conan config install`
* Copy only the configuration files in the directory, for the command `conan config install dirname`

Question :
* For what purpose, do we need the files copy, when the directory is pass by parameter ?

### DIFF
--- a/conans/client/conf/config_installer.py
+++ b/conans/client/conf/config_installer.py
@@ -183,8 +183,6 @@ class _ConfigOrigin(object):
                 config.type = "git"
             elif os.path.isdir(uri):
                 config.type = "dir"
-            # elif is_compressed_file(uri):
-            #     config.type = "compressed"
             elif os.path.isfile(uri):
                 config.type = "file"
             elif uri.startswith("http"):
@@ -207,7 +205,6 @@ def _process_config(config, cache, output, requester):
             _process_git_repo(config, cache, output)
         elif config.type == "dir":
             _process_folder(config, config.uri, cache, output)
-        # elif config.type == "compressed":
         elif config.type == "file":
             if is_compressed_file(config.uri):
                 with tmp_config_install_folder(cache) as tmp_folder:
--- a/conans/client/conf/config_installer.py
+++ b/conans/client/conf/config_installer.py
@@ -210,8 +210,8 @@ def _process_config(config, cache, output, requester):
         elif config.type == "file":
             from zipfile import is_zipfile
             if is_zipfile(config.uri):
-                    with tmp_config_install_folder(cache) as tmp_folder:
-                        _process_zip_file(config, config.uri, cache, output, tmp_folder)
+                with tmp_config_install_folder(cache) as tmp_folder:
+                    _process_zip_file(config, config.uri, cache, output, tmp_folder)
             else:
                 dirname, filename = os.path.dirname(config.uri), os.path.basename(config.uri)
                 _process_file(dirname, filename, config, cache, output, dirname)
--- a/conans/client/conf/config_installer.py
+++ b/conans/client/conf/config_installer.py
@@ -86,6 +86,45 @@ def _filecopy(src, filename, dst):
     shutil.copyfile(src, dst)
 
 
+def _process_file(directory, filename, config, cache, output, folder):
+    if filename == "settings.yml":
+        output.info("Installing settings.yml")
+        _filecopy(directory, filename, cache.cache_folder)
+    elif filename == "conan.conf":
+        output.info("Processing conan.conf")
+        _handle_conan_conf(cache.config, os.path.join(directory, filename))
+    elif filename == "remotes.txt":
+        output.info("Defining remotes from remotes.txt")
+        _handle_remotes(cache, os.path.join(directory, filename))
+    elif filename in ("registry.txt", "registry.json"):
+        try:
+            os.remove(cache.remotes_path)
+        except OSError:
+            pass
+        finally:
+            _filecopy(directory, filename, cache.cache_folder)
+            migrate_registry_file(cache, output)
+    elif filename == "remotes.json":
+        # Fix for Conan 2.0
+        raise ConanException("remotes.json install is not supported yet. Use 'remotes.txt'")
+    # TODO : do we need this ? in which case we need to copy all the folder source
+    # TODO : into the conan configuration file ?
+    # else:
+    #     # This is ugly, should be removed in Conan 2.0
+    #     if filename in ("README.md", "LICENSE.txt"):
+    #         output.info("Skip %s" % filename)
+    #     else:
+    #         relpath = os.path.relpath(directory, folder)
+    #         if config.target_folder:
+    #             target_folder = os.path.join(cache.cache_folder, config.target_folder,
+    #                                          relpath)
+    #         else:
+    #             target_folder = os.path.join(cache.cache_folder, relpath)
+    #         mkdir(target_folder)
+    #         output.info("Copying file %s to %s" % (filename, target_folder))
+    #         _filecopy(directory, filename, target_folder)
+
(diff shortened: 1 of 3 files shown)


conans/client/conf/config_installer.py
```diff
@@ -86,6 +86,45 @@ def _filecopy(src, filename, dst):
...
+            os.remove(cache.remotes_path)
+        except OSError:
+            pass
+        finally:
+            _filecopy(directory, filename, cache.cache_folder)
+            migrate_registry_file(cache, output)
+    elif filename == "remotes.json":
+        # Fix for Conan 2.0
+        raise ConanException("remotes.json install is not supported yet. Use 'remotes.txt'")
+    # TODO : do we need this ? in which case we need to copy all the folder source
```
memsharded: yes, please leave this as-is, it is a feature to copy any other file in the folder, not just the known ones. For example, it is possible to install python files and use them in recipes (not recommended, legacy, but possible)

conans/client/tools/files.py
```diff
@@ -101,41 +101,42 @@ def print_progress(the_size, uncomp_size):
...
-                    z.extract(file_, full_path)
-                    if keep_permissions:
-                        # Could be dangerous if the ZIP has been created in a non nix system
-                        # https://bugs.python.org/issue15795
-                        perm = file_.external_attr >> 16 & 0xFFF
-                        os.chmod(os.path.join(full_path, file_.filename), perm)
-                except Exception as e:
-                    output.error("Error extract %s\n%s" % (file_.filename, str(e)))
-        output.writeln("")
+    if zipfile.is_zipfile(filename):
```
memsharded: I don't understand this, this code is specifically for zip files, so no need to check it. 
At most it could be a "precondition" check that checks it and raise a meaningful error otherwise, but that would be a different feature (that will require its own unit test)

memsharded (changes requested):
Hi @sagarafr

Thanks very much for contributing this PR. Please have a look:

- the tests are broken. This shouldn' t be breaking existing tests (https://conan-ci.jfrog.info/blue/organizations/jenkins/ConanTestSuite/detail/PR-7840/1/pipeline)
- It would be necessary to add some test that covers the new functionality. Please ask for help if needed
- Please sign the CLA above (sometimes it is an issue of using the right user while creating commits, using the same user as github one)

conans/client/conf/config_installer.py
sagarafr: OK. I was not sure of the usage, and, in the issue, the user want no necessary file copy.

conans/client/tools/files.py
sagarafr: When I use to run the command `conan configure install` on a file, the file is categorized as a file, then, try to unzip this file and this raise an exception, because it's not a ZIP file.
My meaning, in this change, is to check if the file is a ZIP file, first, to avoid unnecessary exception and after that threat the file as a real file.
I can change that in order to first check if the input file is a configuration, and then, run the unzip function in order to keep the raising error. Or I can keep this change and add a meaning full error output. What is your preference ?

conans/client/conf/config_installer.py
```diff
@@ -201,9 +207,12 @@ def _process_config(config, cache, output, requester):
             _process_git_repo(config, cache, output)
         elif config.type == "dir":
             _process_folder(config, config.uri, cache, output)
-        elif config.type == "file":
+        elif config.type == "compressed":
             with tmp_config_install_folder(cache) as tmp_folder:
                 _process_zip_file(config, config.uri, cache, output, tmp_folder)
+        elif config.type == "file":
```
memsharded, line 208: Hi @sagarafr

I am having a look, and there is a problem: we cannot change existing behavior unless it is declared a bug. At the moment the type for a zipped file, the type is "file", and that cannot be changed. It is not good enough to introduce a ``compressed`` type for differentiating this. The current ``file`` behavior has to be respected and maintained.

So I think it would be better to try to process files capturing the exception when the decompression fails.

sagarafr: Indeed, I understand my mistake when I have push. I just try to find the right place to make the check in order to have a cleaner code and have all tests pass.

memsharded (approved):
Good, thanks very much for contributing this feature!

mgappa:
Thanks you for your patience

sagarafr:
Thanks your for your patience

PR #7841, opened 2020-10-07
[refact] What if... we use an inheritance based approach for CMake toolchains?

jgsogo:
Changelog: omit
Docs: omit

Main motivations here:

The generated toolchain file will be different for different cross-compiling scenarios. Probably some variables that make sense for some target architectures don't make sense for others (CMAKE_CXX_FLAGS). And for sure values will be different. 

I think that this _inheritance based_ approach will help us to keep source code cleaner:

The template itself:
 * the base template file can be simpler: probably a base template for variables and definitions that affect to every CMake based compilation: `set(...)`, `add_definitions(..)`, `BUILD_SHARED_LIBS`, `CMAKE_POSITION_INDEPENDENT_CODE`,... (or at least I'm pretty sure that, if they don't apply for a target platform, they will be ignored)
 * each inherited class can extend/append to that file the things that are specific to the target architecture. For example, the native one will add `-m64` to CONAN_CXX_FLAGS, while the Android one doesn't want to add anything there.

The logic inside the functions:
 * common logic can be implemented in the base file, but specific logic for a target platform can override the function and we will get rid of many nasty `if/else` to handle specific systems, cross-compiling logic, exceptional values,... 

At least this is the intention...

#tags: slow

### DIFF
--- a/conans/client/toolchain/cmake/__init__.py
+++ b/conans/client/toolchain/cmake/__init__.py
@@ -1,0 +1,16 @@
+from conans.client.tools import cross_building
+from .native import CMakeNativeToolchain
+from .android import CMakeAndroidToolchain
+
+
+def CMakeToolchain(conanfile, *args, **kwargs):
+    if not cross_building(conanfile):
+        return CMakeNativeToolchain(conanfile, *args, **kwargs)
+    else:
+        # Exceptions to cross-building scenarios
+        if conanfile.settings.os == 'Windows' and conanfile.settings_build.os == 'Windows':
+            return CMakeNativeToolchain(conanfile, *args, **kwargs)
+
+        # Actual cross-building
+        if conanfile.settings.os == 'Android':
+            return CMakeAndroidToolchain(conanfile, *args, **kwargs)
--- a/conans/client/toolchain/cmake/__init__.py
+++ b/conans/client/toolchain/cmake/__init__.py
@@ -1,16 +1,17 @@
 from conans.client.tools import cross_building
-from .native import CMakeNativeToolchain
+from conans.errors import ConanException
 from .android import CMakeAndroidToolchain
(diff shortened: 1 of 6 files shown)


memsharded (approved):
looks good to me

jgsogo:
I think that the test is failing because it is a cross-compiling scenario. I was not expecting to find it here. In purity it would require a `set(CMAKE_SYSTEM_NAME Linux)` but I need to have a closer look at it (that would mean that the toolchain as it is now is wrong).

If building from `x86_64` to `x86` doesn't require cross-building machinery neither for Windows nor for Linux, then maybe it is our `cross_compiling()` function the one we should fix.

Open questions: is it a cross-compiling scenario? Do we want to set `CMAKE_SYSTEM_NAME` and get `CMAKE_CROSSCOMPILING` set? What is the difference between the two approaches?

conans/client/toolchain/cmake/__init__.py
```diff
@@ -0,0 +1,17 @@
+from conans.client.tools import cross_building
+from conans.errors import ConanException
+from .android import CMakeAndroidToolchain
+from .native import CMakeNativeToolchain
+
+
+def CMakeToolchain(conanfile, *args, **kwargs):
+    skip_x64_x86 = conanfile.settings.get_safe('os') in ['Windows', 'Linux']
+    if not cross_building(conanfile, skip_x64_x86=skip_x64_x86):
```
memsharded: I would simplify this, and add new logic only when the derived classes are really created. So at the moment I would have just 2, the Android one which is very clear, and another CMakeGenericToolchain() that could be now both cross-building and native, until someone actually does some cross compilation.

memsharded (changes requested):

memsharded (approved):

PR #7843, opened 2020-10-07
[poc] CMake + Android

### LINKED ISSUE #7809: [feature] CMakeToolchain + Android POC
Implement a proof of concept of using the new CMakeToolchain feature for building:

- Android projects (from Linux/OSX)
- From Windows

Notes:
- Android NDK can be assumed installed
- Integration test.
- If enough time to implement reuse of a package, the ``cmake_find_package_multi`` generator must be used

jgsogo:
Changelog: Feature: Add POC on a toolchain for Android (using CMake provided modules).
Docs: https://github.com/conan-io/docs/pull/1902

Close #7809 

There are two main different approaches to compile to Android using CMake, both of them should be equivalent:
 * Using CMake provided toolchains: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-android-with-the-ndk
 * Using AndroidNDK provided toolchain: https://developer.android.com/ndk/guides/cmake#command-line_1

Not sure which one is better, here it is implemented the first one. Do we want to let the user choose?

---

- [ ] There are many TODOs to work on yet


conans/client/toolchain/cmake/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from .native import CMakeNativeToolchain
+from .android import CMakeAndroidToolchain
+
+
+def CMakeToolchain(conanfile, **kwargs):
+    if not cross_building(conanfile):
+        return CMakeNativeToolchain(conanfile=conanfile, **kwargs)
+    else:
+        # Exceptions to cross-building scenarios
+        if conanfile.settings.os == 'Windows' and conanfile.settings_build.os == 'Windows':
```
SSE4: why exceptions for Windows? anyway, this sounds like out of scope for Android, don't understand why it should be changed here

conans/test/integration/toolchains/android/test_using_cmake.py
```diff
@@ -0,0 +1,81 @@
...
+class SystemToolsTestCase(unittest.TestCase):
+    # This test assumes that 'CMake' and 'AndroidNDK' are available in the system
+    #
+    # Guidelines: https://developer.android.com/ndk/guides/cmake#command-line
+
+    @classmethod
+    def setUpClass(cls):
+        if not shutil.which('cmake'):
+            raise unittest.SkipTest("CMake expected in PATH")
+        if not shutil.which('cmake'):
```
SSE4: checking the same twice?

conans/client/toolchain/cmake/android.py
```diff
@@ -0,0 +1,129 @@
+import textwrap
+
+from .base import CMakeToolchainBase
+
+
+class CMakeAndroidToolchain(CMakeToolchainBase):
```
SSE4, line 9: so, do we need Toolchain class for every OS? this doesn't scale, and won't work with custom settings. is it our general strategy going forward?

conans/test/integration/toolchains/android/test_using_cmake.py
jgsogo: This is an ongoing effort. I'll remove that. My plan was to have something like `which('ndk-build')`, but it looks like it is not in the path by default.

conans/client/toolchain/cmake/__init__.py
jgsogo: More about it here (https://github.com/conan-io/conan/pull/7841), this PR is built on top of that one. But, specifically, comments related to the Windows exception here (reviews and comments: https://github.com/conan-io/conan/pull/7827), current Conan behavior doesn't consider a cross-building scenario when going from x64 to x86 in Windows or Linux.

conans/client/toolchain/cmake/android.py
jgsogo: IMO it can scale much better than adding `if/else` conditions inside the methods: we can compose templates and classes, and override methods. Otherwise, we probably end up with the template full of `{% if something %}{{ something }}{% endif %}` (like now) and methods full of `if/else` (like they are now in the build-helepers): harder to maintain, harder to reason about and really hard to modify them in parallel (many of us are working on toolchains right now).

At least, I think it would be convenient while we work in parallel on different toolchains, once we now the features around the toolchains we can think if this is the best implementation or not and refactor it.

SSE4: it seems like I don't understand the model fully. I was thinking like we are going to use `CMakeToolchain`, possibly specifying some additional things like base toolchain and additional definitions, if needed. for instance, for Android, it may look like:
```
tc = CMakeToolchain(self)
tc.base = os.path.join(android_ndk, "build", "cmake", "android.toolchain.cmake")
android_variables = {"ANDROID_PLATFORM": "android-21", "ANDROID_ABI": "arm64-v8a", "ANDROID_STL=c++_static"}
tc.variables += android_variables
```
similarly, we may use the same approach for other platforms. WDYT?

jgsogo: We really don't know, that's the main purpose of implementing different toolchains, to realize what they need and try to define an interface that satisfies all the different needs. But first, we need those POCs to see different scenarios.

IMHO, any recipe that builds with CMake should be exactly the same, nothing more than this (and everything could be simplified to a declaration):

image

and then we need to provide a way to build this package using other platforms and/or toolchains **without the need to modify the recipe itself**. Some recipes might require some tweaks, but those should be exceptions.

Having this in mind there are two approaches needed:
 * a way to inject a custom toolchain from outside (defined in a profile or coming from a build-requires): there is no POC for this, but IMO it should be built-in, it shouldn't require to modify the recipe at all.
 * some well-known scenarios can be managed by Conan (I would say that those that are listed in the default `settings.yml`). We can implement everything needed for Android behind the scenes as this PR does.

---

From this PR I learn two things:
 * the class-based + inheritance approach is cleaner, I don't need to modify the base template at all, no `if/else`, no increase in complexity, implementation is straightforward (right now it is intentionally duplicating code).
 * to work with Android we need the `ANDROID_NDK` path, we need to figure out how to pass that information to the toolchain implementation (right now it will be via environment variable).

conans/client/toolchain/cmake/android.py
```diff
@@ -1,5 +1,130 @@
...
+
 from .base import CMakeToolchainBase
 
 
 class CMakeAndroidToolchain(CMakeToolchainBase):
-    pass
+    _template_project_include = ''  # TODO: This file is not useful to Android, there is no MSVC runtime MD/MT
+
+    # TODO: Factorize with the native one
+    _template_toolchain = textwrap.dedent("""
```
memsharded: This is one of my concerns, that this diverge too early, and it is difficult to understand common parts and specialization parts, and fixes in other platforms fail to apply to others. 

Yes, please factorize this so we can better see isolated the differences of the Android toolchain.

jgsogo: That's another PR to `develop` and yes, it is totally necessary (WIP). From this PR we should take some learnings about the toolchain feature, the same for the rest of PRs related to CMake toolchain. In this one I'm _learning_.

You will love how it looks like after the refactor 😄

conans/client/toolchain/cmake/android.py
```diff
@@ -1,5 +1,92 @@
+import os
+import textwrap
+
 from .base import CMakeToolchainBase
+from ...tools import which, ConanException
```
memsharded: We are not using relative imports syntax.

conans/client/toolchain/cmake/android.py
```diff
@@ -1,5 +1,92 @@
...
+        return {"x86": "x86",
+                "x86_64": "x86_64",
+                "armv7": "armeabi-v7a",
+                "armv8": "arm64-v8a"}.get(str(self._conanfile.settings.arch))
+
+    def _get_android_stl(self):
+        libcxx_str = str(self._conanfile.settings.compiler.libcxx)
+        return libcxx_str  # TODO: only 'c++_shared' y 'c++_static' supported?
+
+    def _guess_android_ndk(self):
```
memsharded, line 74: This seems wrong, not returning anything?

conans/test/integration/toolchains/android/test_using_cmake.py
```diff
@@ -0,0 +1,70 @@
+import textwrap
+import unittest
+
+from conans.client.toolchain.cmake.base import CMakeToolchainBase
+from conans.client.tools import which
+from conans.test.utils.tools import TestClient
+from ._utils import create_library
+
+
+class SystemToolsTestCase(unittest.TestCase):
```
memsharded: Test name?

memsharded (approved):

Issue #7846, opened 2020-10-07
[bug] _/_ doesn't work the same as usual user/channel


alex-700:
- The quick reproducer in unit-test is [here](https://github.com/alex-700/conan/commit/0f9822c5891e1fad759bb74f02bbbedfe33de2b8).
- This bug was introduced in #7673. 
- The query `packagename/version@_/_` is not the same as `packagename/version`, because the latter can return also `packagename/version@someuser/somechannel` and client side filtration was deleted in the PR mentioned above.

memsharded:
Hi @alex-700 

Thanks for reporting this.

Just to clarify, the ``_/_`` user and channel should never appear in the UI, not as a ``requires`` definition, not in any Conan command. This doesn't appear at all in the docs. 

The correct code is ``requires = "pkg/version"``, or ``conan install pkg/version@``

Does the same unit-test fails if you change that?

This could be a regression, lets have a look, if such, it should be fixed asap in 1.30.1

alex-700:
@memsharded,

Thanks for the quick response.

Yeah, I know about `_/_` in the UI. It was just a fast way to show the problem in the conan client-server interaction. This bug affects the UI too (in the critical manner), e.g. if you have bincrafters remote before CCI in the remotes list and try to resolve some dependency (I know about `spdlog/[^1.2]`) you will get `spdlog/someversion/bincrafters/stable` instead of `spdlog/someversion`  (correct one from CCI).

No, the same unit-test doesn't fail. This test shows the situation when we need to resolve dependency from CCI, but conan-server get the query `somepackage/someversion` without `@_/_`.

memsharded:
This is weird. I am doing a variation of your test, removing the ``@_/_`` and it seems green here:
```diff
diff --git a/conans/test/unittests/client/graph/version_ranges_graph_test.py b/conans/test/unittests/client/graph/version_ranges_graph_test.py
index 210a5d8c0..653c86c2d 100644
--- a/conans/test/unittests/client/graph/version_ranges_graph_test.py
+++ b/conans/test/unittests/client/graph/version_ranges_graph_test.py
@@ -45,6 +45,8 @@ class VersionRangesTest(GraphTest):
             say_content = GenConanfile().with_name("Say").with_version(v)
             say_ref = ConanFileReference.loads("Say/%s@myuser/testing" % v)
             self.retriever.save_recipe(say_ref, say_content)
+            say_ref = ConanFileReference.loads("Say/%s" % v)
[...]

PR #7673, opened 2020-09-08
Use patterns in server query when resolving version ranges 

czoido:
Changelog: Fix: Use patterns in server query when resolving version ranges.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/7672

After a poor performance in resolution of version ranges in Artifactory was reported (https://github.com/conan-io/conan/issues/5432), Conan introduced a workaround for the issue in https://github.com/conan-io/conan/pull/5433. With these changes now the query to Artifactory server was made removing the patterns: 
`
http://localhost:8081/artifactory/api/conan/conan/v2/conans/search?q=cgal
`
instead of:
`
http://localhost:8081/artifactory/api/conan/conan/v2/conans/search?q=cgal%2F%2A%40user%2Fchannel&ignorecase=False
`
and then the ranges were resolved with all the packages on the client side.

This is causing a security gap as defining exclude patterns in Artifactory will not work as those patterns are not sent any more to the server as reported here: https://github.com/conan-io/conan/issues/7672. This branch reverts those changes. 

Please, @Daniel-Roberts-Bose could you try this branch and check if the performance issues are still happening?

- [X] Refer to the issue that supports this Pull Request.
- [X] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [X] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [X] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #7847, opened 2020-10-07
Fix/range resolver without userchannel

memsharded:
Changelog: Bugfix: Fix regression introduced in 1.30 (https://github.com/conan-io/conan/pull/7673), with incorrect matches of user/channel for version ranges.
Docs: Omit

Fix https://github.com/conan-io/conan/issues/7846


czoido (approved):
🤦‍♂️

PR #7850, opened 2020-10-08
[refact] Move to parent class those items that are common to all toolchains

jgsogo:
Changelog: omit
Docs: omit

Without being aggressive and moving more things, there are some common things that will be used by all the toolchains:
 * `variables`
 * `preprocessor_definitions`
 * `install_prefix`

### DIFF
--- a/conans/client/toolchain/cmake/base.py
+++ b/conans/client/toolchain/cmake/base.py
@@ -1,8 +1,32 @@
 import os
+from collections import OrderedDict, defaultdict
 
 from jinja2 import Template
 
 from conans.util.files import save
+
+
+class Variables(OrderedDict):
+    _configuration_types = None  # Needed for py27 to avoid infinite recursion
+
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


conans/client/toolchain/cmake/base.py
```diff
@@ -14,24 +38,38 @@ class CMakeToolchainBase(object):
...
         """ Returns two dictionaries, the context for the '_template_toolchain' and
             the context for the '_template_project_include' templates.
         """
-        return {}, {}
+        ctxt_toolchain = {
+            "variables": self.variables,
+            "variables_config": self.variables.configuration_types,
+            "preprocessor_definitions": self.preprocessor_definitions,
+            "preprocessor_definitions_config": self.preprocessor_definitions.configuration_types,
+            "install_prefix": self.install_prefix,
```
memsharded, line 59: We probably want to discuss a bit about the UI of this when this is local dev flow

memsharded (approved):

PR #7851, opened 2020-10-08
[refact] Refact/toolchain templates

jgsogo:
Changelog: omit
Docs: omit

~~After https://github.com/conan-io/conan/pull/7850~~

We can take some advantage of Jinja to organize the templates for the toolchains. More can be done, but I prefer to wait until we have other toolchains.

 * There is a `base_toolchain` in the parent class that _writes_ the variables that belong to the parent class (those that were moved in #7850).
 * The base `base_toolchain` template provides three blocks to add more data to the toolchain file (we should argue if these blocks are needed, but it is what we have right now):
   + `before_try_compile`: all configuration that is taken into account for every try-compile work
   + `main`: all configuration that is taken into account only for the project to build
   + `footer`: a placeholder after variables and preprocessor definitions (probably it can be removed in favor of `main`).
 * Child toolchains can extend that base template an populate the required blocks:

   + `CMakeGenericToolchain` is including some external templates for `before_try_compile` and `main` block while defining the block `footer` in the template (the difference is more for demo purpose than something needed):

       ```
       {% extends 'base_toolchain' %}

       {% block before_try_compile %}
       {% include 'before_try_compile' %}
       {% endblock %}

       {% block main %}
       {% include 'main' %}
       {% endblock %}

       {% block footer %}
       set(CMAKE_CXX_FLAGS_INIT "${CONAN_CXX_FLAGS}" CACHE STRING "" FORCE)
       set(CMAKE_C_FLAGS_INIT "${CONAN_C_FLAGS}" CACHE STRING "" FORCE)
       set(CMAKE_SHARED_LINKER_FLAGS_INIT "${CONAN_SHARED_LINKER_FLAGS}" CACHE STRING "" FORCE)
       set(CMAKE_EXE_LINKER_FLAGS_INIT "${CONAN_EXE_LINKER_FLAGS}" CACHE STRING "" FORCE)
       {% endblock %}
       ```

   + `CMakeAndroidToolchain` will [define everything in the template file](https://github.com/conan-io/conan/pull/7843/files#diff-20e8dfdd1e3b1978c7fcd373efc8f17aR7) (WIP) adding nothing to the `before_try_compile` block:

      ```
        {% extends 'base_toolchain' %}

        {% block main %}
        {{ super() }}

        # shared libs
        {% if shared_libs -%}
        message(STATUS "Conan toolchain: Setting BUILD_SHARED_LIBS= {{ shared_libs }}")
        set(BUILD_SHARED_LIBS {{ shared_libs }})
        {%- endif %}

        # Parallel builds
        {% if parallel -%}
        set(CONAN_CXX_FLAGS "${CONAN_CXX_FLAGS} {{ parallel }}")
        set(CONAN_C_FLAGS "${CONAN_C_FLAGS} {{ parallel }}")
        {%- endif %}

        # C++ Standard
        {% if cppstd -%}
        message(STATUS "Conan C++ Standard {{ cppstd }} with extensions {{ cppstd_extensions }}}")
        set(CMAKE_CXX_STANDARD {{ cppstd }})
        set(CMAKE_CXX_EXTENSIONS {{ cppstd_extensions }})
        {%- endif %}
        {% endblock %}

        {% block footer %}
        set(CMAKE_CXX_FLAGS_INIT "${CONAN_CXX_FLAGS}" CACHE STRING "" FORCE)
        set(CMAKE_C_FLAGS_INIT "${CONAN_C_FLAGS}" CACHE STRING "" FORCE)
        set(CMAKE_SHARED_LINKER_FLAGS_INIT "${CONAN_SHARED_LINKER_FLAGS}" CACHE STRING "" FORCE)
        set(CMAKE_EXE_LINKER_FLAGS_INIT "${CONAN_EXE_LINKER_FLAGS}" CACHE STRING "" FORCE)
        {% endblock %}
      ```

In the future, with a better understanding of the toolchains and the variables they need/use, we can iterate this templates and simplify them or move them to the specific mixin that will take care of the information.

### DIFF
--- a/conans/client/toolchain/cmake/base.py
+++ b/conans/client/toolchain/cmake/base.py
@@ -1,7 +1,9 @@
 import os
+import textwrap
 from collections import OrderedDict, defaultdict
 
-from jinja2 import Template
+from jinja2 import DictLoader
+from jinja2 import Environment
 
 from conans.util.files import save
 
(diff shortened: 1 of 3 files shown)


memsharded (commented):
I honestly don't know what to think about this. On one hand, it is very good factoring out and putting things in common and reuse them, but on the other hand, it is more difficult to read, understand the information, and debug things. All that is evaluated in the templates, at least IMHO is way more difficult to debug and track mentally. I guess it is just difficult to find the balance.

memsharded:
I have checked out this branch, and started navigating the code.
I have struggled to track the origin of things, to have a clear understanding of the resulting toolchain files.

I think I would prefer some more repetition but more evident files. Probably just one file, with ``if android`` and similar conditions. One single template containing everything. Let's get the feedback from the team.

jgsogo:
Yes, I'm waiting for that quick review by @danimtb  and @czoido .

I really find jinja2 templates intuitive and easy to follow, with the `extends` and `include` commands... the `block` and the `super`,... and the loaders. But I've been working with this kind of language for several years now.

czoido:
I would also prefer have more repetition in the sake of being able to follow the code easier but that's maybe because I'm not so familiar with jinja2 templates...

danimtb (approved):

lasote (approved):
I would say this is a bit nicer but I also worked with similar templates a time ago. 
A bit worried about the "hierarchy" being so much restrictive when a new derivated toolchain arrives in case it doesn't fit perfectly.
I understand the concerns of @memsharded and @czoido, this would require to get a bit confident with jinja2 templates to change something more than a detail.
So approved by `0.5` :rofl:

conans/client/toolchain/cmake/base.py
```diff
@@ -33,8 +34,94 @@ class CMakeToolchainBase(object):
...
+        # DO NOT EDIT MANUALLY, it will be overwritten
+
+        # Avoid including toolchain file several times (bad if appending to variables like
+        #   CMAKE_CXX_FLAGS. See https://github.com/android/ndk/issues/323
+        if(CONAN_TOOLCHAIN_INCLUDED)
+          return()
+        endif()
+        set(CONAN_TOOLCHAIN_INCLUDED TRUE)
+
+        {% block before_try_compile %}
```
memsharded, line 70: Maybe we could come up with a name that is more meaningful that ``before_try_compile``?

conans/client/toolchain/cmake/generic.py
```diff
@@ -123,55 +90,28 @@ class CMakeGenericToolchain(CMakeToolchainBase):
...
-        {%- endfor %}
+        {% block before_try_compile %}
+        {% include 'before_try_compile' %}
+        {% endblock %}
 
+        {% block main %}
+        {% include 'main' %}
+        {% endblock %}
 
+        {% block footer %}
```
memsharded: Maybe same with the name? Could this section be something like ``cmake_flags_init`` or something similar?

memsharded (approved):
Ok, so it seems there is a majority of Jinja template lovers, so I guess we should go with that, and I will have to get more used to it.

Probably I would try to improve the naming, making blocks a bit more meaningful.

memsharded:
@jgsogo if you please solve conflicts, it can be merged.

jgsogo:
Changes/comments after latest reviews:
 * Naming. I intentionally used `before_try_compile`, `main` and `footer` because they refer to the position more than to the functionality and I wanted to highlight it. I'm not sure if every variable is in the place it should be, maybe we should move some of them upper in the file so they are used by the try-compile calls?

   Said that, what about these names?
   * `before_try_compile` -> `toolchain_features_config`
   * `main` -> `main`? `project_config`?
   * `footer`: removed, everything in this block can be moved to the `main` one.

 * Changes in the `CMakeGenericToolchain`: I'm no longer _including_ other templates, it was only to show the possibility, now the template contains all the definitions in a single template ([here](https://github.com/conan-io/conan/pull/7851/files#diff-161115eaa7a0c8470add8d7723f1547229f2508568838ff63e0162f8d60598ddR17)). You can also have a look to the Android implementation here (https://github.com/jgsogo/conan/pull/45/files), it follows the same approach.

 * **Indentation**. This is a very controversial topic. When you use templates it is very hard to generate files with nice indentation. People usually prefer to indent the template and forget about generated files (you can have a look at any HTML generated by Django, there are tons of webs out there). Alternatives:
   1. worry about the template, do not worry about the file generated
   1. do not worry about the template and try to generate a nice file: template will get lots of `-` around the `{% %}` symbols trying to remove spaces, the template won't follow indentation rules (much harder to read),...
   1. worry about the template, fix the generated content before printing it in a dedicated step, basically running a beautifier.

memsharded (approved):
Looking good to me. Lets merge and we can iterate other things later.

I would like to have clean, readable generated toolchains. At the moment I would start with option 2, a beutifier for cmake will surely have its complexities.

czoido (approved):

PR #7855, opened 2020-10-09
[poc] CMake + iOS

### LINKED ISSUE #7810: [feature] CMakeToolchain + iOS projects PoC
Implement a proof of concept of using the new CMakeToolchain feature for building iOS projects

Notes:
- Toolchain can be assumed installed
- Integration test.
- If enough time to implement reuse of a packages, the ``cmake_find_package_multi`` generator must be used

czoido:
Changelog: Feature: Add POC on a toolchain for iOS (using CMake XCode generator).
Docs: https://github.com/conan-io/docs/pull/1906

Only working with XCode generator for the moment: https://cmake.org/cmake/help/v3.18/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos

Close https://github.com/conan-io/conan/issues/7810


conans/test/integration/toolchains/ios/test_using_cmake.py
```diff
@@ -0,0 +1,80 @@
...
+
+                def toolchain(self):
+                    tc = CMakeToolchain(self)
+                    tc.write_toolchain_files()
+
+                def build(self):
+                    cmake = self._configure_cmake()
+                    cmake.configure()
+                    cmake.build()
+                    self.run("lipo -info Release-iphoneos/libhello.a")
```
memsharded, line 43: Is this very needed? Should be part of the toolchain somehow? Can't be done by cmake? Is expected that the user runs this command manually in the terminal when building without conan?

conans/test/integration/toolchains/ios/_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    void HelloLib::hello(const std::string& name) {
+        #ifdef DEBUG
+        std::cout << "Hello " << name << " Debug!" <<std::endl;
+        #else
+        std::cout << "Hello " << name << " Release!" <<std::endl;
+        #endif
+    }
+""")
+
+cpp_wrapper_h = textwrap.dedent("""
```
memsharded: Isn't it possible to have  a pure C++ library compiled for iOS?

That would be much better, if the testing code would be generic?. If we could be building exactly the same "hello world" C++ library with all toolchains, that would be great.

conans/client/toolchain/cmake/ios.py
```diff
@@ -0,0 +1,133 @@
...
+    """)
+
+    def __init__(self, conanfile, build_type=None, **kwargs):
+        super(CMakeiOSToolchain, self).__init__(conanfile, build_type=build_type, **kwargs)
+        self.build_type = build_type or self._conanfile.settings.get_safe("build_type")
+        self.host_architecture = self._get_architecture()
+        self.host_os = self._conanfile.settings.get_safe("os")
+        self.host_os_version = self._conanfile.settings.get_safe("os.version")
+        self.host_sdk_name = self._get_sdk_name(self.host_architecture)
+        self.host_os_min_version = "9.0"
```
memsharded: This constant "9.0" is a bit concerning, please add at least an explanation, a warning or something.

conans/test/integration/toolchains/ios/test_using_cmake.py
```diff
@@ -0,0 +1,80 @@
...
+                compiler.libcxx=libc++
+                build_type=Release
+            """)
+        })
+
+    def test_xcode_generator(self):
+        """ Simplest approach:
+            https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos
+        """
+        self.t.save({'conanfile.py': self._conanfile.format(generator='"Xcode"')})
```
memsharded, line 67: To discuss what would be a way so the recipe doesn't need to hardcode ``generator=Xcode`` in it, so it can be generic to more platforms.

conans/test/integration/toolchains/ios/test_using_cmake.py
czoido: That's to check later in the test that the library we have compiled actually resulted with `arm64` architecture.

conans/test/integration/toolchains/ios/_utils.py
czoido: It is not strictly needed, I just liked the idea of having also the wrapper there, that you will need if you want to call the library from Swift. If if makes it cleaner I'll remove it 👍

conans/client/toolchain/cmake/ios.py
```diff
@@ -0,0 +1,97 @@
...
+            set(CMAKE_OSX_SYSROOT {{ CMAKE_OSX_SYSROOT }})
+            if(NOT DEFINED CMAKE_XCODE_ATTRIBUTE_DEVELOPMENT_TEAM)
+              set(CMAKE_XCODE_ATTRIBUTE_DEVELOPMENT_TEAM "123456789A" CACHE INTERNAL "")
+            endif()
+        {% endblock %}
+        {% block main %}
+            {{ super() }}
+            {% if shared_libs -%}
+            message(STATUS "Conan toolchain: Setting BUILD_SHARED_LIBS= {{ shared_libs }}")
+            set(BUILD_SHARED_LIBS {{ shared_libs }})
```
memsharded, line 28: This is already part of the ``CMakeGenericToolchain``, can't be reused? This is what I struggle to see with templates.

conans/client/toolchain/cmake/ios.py
```diff
@@ -0,0 +1,97 @@
...
+        # "-DCMAKE_OSX_ARCHITECTURES=armv7;armv7s;arm64;i386;x86_64"
+        arch = self._conanfile.settings.get_safe("arch")
+        return {"x86": "i386",
+                "x86_64": "x86_64",
+                "armv8": "arm64",
+                "armv8_32": "arm64_32"}.get(arch, arch)
+        return None
+
+    def _get_sdk_name(self, architecture):
+        os_name = self._conanfile.settings.get_safe("os")
```
memsharded: If this is a copy & paste, please add a comment with the original source: ``apple_sdk_name()``, for example, to track it when we refactor.

conans/test/integration/toolchains/ios/test_using_cmake.py
```diff
@@ -0,0 +1,79 @@
...
+
+        # Build in the cache
+        self.t.run('create . --profile:build=default --profile:host=ios_profile')
+        self.assertIn("Non-fat file: Release-iphoneos/libhello.a is architecture: arm64", self.t.out)
+
+        # Build locally
+        self.t.run('install . --profile:host=ios_profile --profile:build=default')
+        self.t.run_command('cmake . -G"Xcode" -DCMAKE_TOOLCHAIN_FILE={}'.format(CMakeToolchainBase.filename))
+        self.t.run_command('cmake --build . --config Release')
```
memsharded, line 76: Isn't it possible to test, with some binary inspection tool or something, that the binary is actually an iOS one? But directly against the binary?

conans/client/toolchain/cmake/ios.py
czoido: I was not aware of the other method in conan, I'll add a note

conans/test/integration/toolchains/ios/test_using_cmake.py
czoido: Yes, I'll add the same check I already had in the recipe to test that it is arm64 with lipo.

conans/client/toolchain/cmake/ios.py
jgsogo: It would be inherited... if this toolchain was inheriting from the CMakeGenericToolchain, but it isn't.

If we are sure that these values are going to be used by all the toolchains, then they can (and have to) be moved to the `CMakeToolchainBase`.

PR #7862, opened 2020-10-11
Autotools adds apple deployement target flag

MartinDelille:
Changelog: Fix: Automatically add OSX deployment flags in ``AutootoolsBuildEnvironment`` with the value of ``os_version``, unless the values are already defined in environment variables `CFLAGS` or `CXXFLAGS`.
Docs: Omit

After submitting [this pull request on CCI](https://github.com/conan-io/conan-center-index/pull/3170) I realized that maybe that was something that `AutoToolsBuildEnvironment` could do automatically since 4 recipes are using the same kind of code: [zlib](https://github.com/conan-io/conan-center-index/blob/f16b9ee7771b73d46662b38a50df00d147df937e/recipes/zlib/1.2.11/conanfile.py#L69), [icu](https://github.com/conan-io/conan-center-index/blob/f696331f762027cfb4c43ac0c01145d682e30163/recipes/icu/all/conanfile.py#L110), [libcurl](https://github.com/conan-io/conan-center-index/blob/27c119a06fc0303efa89c2d8cbbace35f32db1b6/recipes/libcurl/all/conanfile.py#L398), [openssl](https://github.com/conan-io/conan-center-index/blob/27c119a06fc0303efa89c2d8cbbace35f32db1b6/recipes/libcurl/all/conanfile.py#L398)

### DIFF
--- a/conans/client/build/autotools_environment.py
+++ b/conans/client/build/autotools_environment.py
@@ -11,6 +11,7 @@ from conans.client.build.compiler_flags import (architecture_flag, build_type_de
                                                 format_frameworks, format_framework_paths)
 from conans.client.build.cppstd_flags import cppstd_from_settings, \
     cppstd_flag_new as cppstd_flag
+from conans.client import tools
 from conans.client.tools.env import environment_append
 from conans.client.tools.oss import OSInfo, args_to_string, cpu_count, cross_building, \
     detected_architecture, detected_os, get_gnu_triplet, get_target_os_arch, get_build_os_arch
@@ -39,6 +40,7 @@ class AutoToolsBuildEnvironment(object):
         self.subsystem = OSInfo().detect_windows_subsystem() if self._win_bash else None
         self._deps_cpp_info = conanfile.deps_cpp_info
         self._os = conanfile.settings.get_safe("os")
+        self._os_version = conanfile.settings.get_safe("os.version")
         self._arch = conanfile.settings.get_safe("arch")
         self._os_target, self._arch_target = get_target_os_arch(conanfile)
 
@@ -339,6 +341,10 @@ class AutoToolsBuildEnvironment(object):
         tmp_compilation_flags = copy.copy(self.flags)
         if self.fpic:
             tmp_compilation_flags.append(pic_flag(self._conanfile.settings))
+        if tools.is_apple_os(self._os) and self._os_version:
+            flag = tools.apple_deployment_target_flag(self._os,
+                                                      self._os_version)
+            tmp_compilation_flags.append(flag)
 
         cxx_flags = append(tmp_compilation_flags, self.cxx_flags, self.cppstd_flag)
         c_flags = tmp_compilation_flags
--- a/conans/client/build/autotools_environment.py
+++ b/conans/client/build/autotools_environment.py
@@ -343,6 +343,8 @@ class AutoToolsBuildEnvironment(object):
             tmp_compilation_flags.append(pic_flag(self._conanfile.settings))
         if tools.is_apple_os(self._os):
             concat = " ".join(tmp_compilation_flags)
+            if os.environ.get("CFLAGS", None):
+                concat += " " + os.environ.get("CFLAGS", None)
             if self._os_version and "-version-min" not in concat and "-target" not in concat:
                 tmp_compilation_flags.append(tools.apple_deployment_target_flag(self._os,
                                                                                 self._os_version))
--- a/conans/client/build/autotools_environment.py
+++ b/conans/client/build/autotools_environment.py
@@ -345,6 +345,8 @@ class AutoToolsBuildEnvironment(object):
             concat = " ".join(tmp_compilation_flags)
             if os.environ.get("CFLAGS", None):
                 concat += " " + os.environ.get("CFLAGS", None)
+            if os.environ.get("CXXFLAGS", None):
+                concat += " " + os.environ.get("CXXFLAGS", None)
             if self._os_version and "-version-min" not in concat and "-target" not in concat:
                 tmp_compilation_flags.append(tools.apple_deployment_target_flag(self._os,
                                                                                 self._os_version))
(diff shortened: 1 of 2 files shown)


conans/client/build/autotools_environment.py
```diff
@@ -339,6 +341,10 @@ def append(*args):
         tmp_compilation_flags = copy.copy(self.flags)
         if self.fpic:
             tmp_compilation_flags.append(pic_flag(self._conanfile.settings))
+        if tools.is_apple_os(self._os) and self._os_version:
+            flag = tools.apple_deployment_target_flag(self._os,
+                                                      self._os_version)
+            tmp_compilation_flags.append(flag)
```
SSE4: this flag needs to be added carefully
we know from the https://github.com/conan-community/community/issues/304, some users may specify explicit deployment target, e.g. in CFLAGS/CXXFLAGS. we shouldn't overwrite that, otherwise it may break some use cases (e.g. Catalist)

MartinDelille: Ok thank you for your feedback! What do you think would be the best option?

1. Close this PR and apply the flag per recipe (like I did in #3170
2. Apply the same checks done here: https://github.com/conan-io/conan-center-index/pull/585/files
3. Move the flag addition to another part of the code more safe or logical
4. 2 and 3

SSE4: №2 looks as the best option for me. per recipe doesn't scale well.

MartinDelille: Ok I implemented just the part adding the deployement flags for now. I didn't wrote a corresponding test because I don't know how I can inject existing flags in the test, any idea?

conans/client/build/autotools_environment.py
```diff
@@ -339,6 +341,11 @@ def append(*args):
         tmp_compilation_flags = copy.copy(self.flags)
         if self.fpic:
             tmp_compilation_flags.append(pic_flag(self._conanfile.settings))
+        if tools.is_apple_os(self._os):
+            concat = " ".join(tmp_compilation_flags)
+            if self._os_version and "-version-min" not in concat and "-target" not in concat:
+                tmp_compilation_flags.append(tools.apple_deployment_target_flag(self._os,
```
memsharded, line 351: I would like to understand a bit better (not a OSX user) what is the effect of not adding this flag. Is it a compilation error? What have been the users doing so far for this?

Also, is there a chance that this could be breaking for other users that do not expect this flag to be added? Should this flag be added in other generators as well (cmake, xcode)?

conans/test/unittests/client/build/autotools_environment_test.py
```diff
@@ -671,3 +671,25 @@ def autotools_fpic_test(self):
...
+        settings = MockSettings({"os": "Macos",
+                                 "os.version": "10.13",
+                                 "compiler.version": "12.0"})
+        conanfile = MockConanfile(settings, options)
+        be = AutoToolsBuildEnvironment(conanfile)
+        expected = be.vars["CXXFLAGS"]
+        self.assertIn("10.13", expected)
+
+        with tools.environment_append({"CXXFLAGS": "-mmacosx-version-min=10.9"}):
+            be = AutoToolsBuildEnvironment(conanfile)
```
memsharded, line 692: This is failing because the addition to cxxflags is done after the processing of the flags, in:

```python@property
    def vars_dict(self):

        ld_flags, cpp_flags, libs, cxx_flags, c_flags = self._get_vars()  # HERE it gets the min-version flag


        if os.environ.get("CXXFLAGS", None):
            cxx_flags.append(os.environ.get("CXXFLAGS", None))
```

conans/client/build/autotools_environment.py
MartinDelille: If [this flag](https://clang.llvm.org/docs/CommandGuide/clang.html#target-selection-options) is not set and someone is trying to link a library with no minimal os version against an application explicitely specifying it then warning will span like mentioned here: https://github.com/bincrafters/community/issues/1264#issuecomment-705448208

Regarding the users that do not expect this flag to be added I wondered why they would have specify the `os.version` setting that (if I'm understanding well) states that the dependencies should support such OS.

Cmake generators allready handle the `os.version` flags by setting the CMAKE_OSX_DEPLOYMENT_TARGET definition: https://github.com/conan-io/conan/blob/develop/conans/client/build/cmake_flags.py#L216-L219

conans/test/unittests/client/build/autotools_environment_test.py
MartinDelille: Yes I pushed explicitely a failing test but I just fixed it in my last commit: 6e042ba3

conans/client/build/autotools_environment.py
MartinDelille: I found a way to test by simply setting an environment variable: https://github.com/conan-io/conan/pull/7862/files#diff-809c0e7ca852b9746fa8614ed3a57074892e114a74c0072cfc8d92fbad78e195R691-R695

conans/client/build/autotools_environment.py
```diff
@@ -339,6 +341,13 @@ def append(*args):
         tmp_compilation_flags = copy.copy(self.flags)
         if self.fpic:
             tmp_compilation_flags.append(pic_flag(self._conanfile.settings))
+        if tools.is_apple_os(self._os):
+            concat = " ".join(tmp_compilation_flags)
+            if os.environ.get("CFLAGS", None):
```
memsharded, line 346: Couldn't the ``version-min`` come defined also in CXXFLAGS? I think it would make sense check that var too, just in case, as C++ users are probably defining it in that var, what do you think?

memsharded (commented):
It is looking good, please have a look to the CXXFLAGS issue and tell me.

conans/client/build/autotools_environment.py
MartinDelille: Done

memsharded (approved):

SSE4 (approved):

jgsogo (approved):

memsharded:
Merged! Will be in Conan 1.31, thanks for your contribution @MartinDelille 

When it is released, maybe it is worth to revisit those recipes in conan-center-index to simplify them if the flags are no longer necessary in the recipes.

MartinDelille:
Ok I leave this in my inbox and will look at it when 1.31 will be out!

PR #7880, opened 2020-10-14
Test/build require override

### LINKED ISSUE #7864: [question] "To fix this conflict you need to override the package 'gtest' in your root package." even though we have specified the version in root package
Our root package was recently changed from having:

```
        "gtest/1.8.0@PORT/stable", \
```

to:

```
        "gtest/1.8.1@bloomberg/stable", \
```

(Note: both patch change and different `user`)

Test builds with `--build=missing` of that change worked.

But in a recent attempt to rebuild that root package with an updated version of another package 'pal' we are seeing an error like:

```
version[1.1.42] commitAtVersion[4f28b4b]
ERROR: Conflict in scubaclient/1.6@PORT/stable:
    'scubaclient/1.6@PORT/stable' requires 'gtest/1.8.0@PORT/stable' while 'pal/2.15.0@PORT/stable' requires 'gtest/1.8.1@bloomberg/stable'.
    To fix this conflict you need to override the package 'gtest' in your root package.
gtest/1.8.0@PORT/stable: Not found in local cache, looking in remotes...
```

I understand from statements in tickets like https://github.com/conan-io/conan/issues/6812#issuecomment-610451941 that:

> Yes, the explanation of the error message ... is that there are different packages using different "zlib" packages. On one branch of the dependency graph, someone is still using the old zlib/1.2.11@conan/stable, while some other more modern dependency is using the newer zlib/1.2.11, and that is a conflict. Specifying which of the two ones you want, the conflict is resolved. 

But in this case we *are* specifying the version of gtest we want into our root package.  So why do we still see this error?

Conan version 1.26.0
Windows 2012 R2 Server
vs2015_14_v140_Debug
 
memsharded:
Changelog: Omit
Docs: Omit

Close https://github.com/conan-io/conan/issues/7864

### DIFF
--- a/conans/test/functional/graph/version_range_override_test.py
+++ b/conans/test/functional/graph/version_range_override_test.py
@@ -53,3 +53,34 @@ class VersionRangeOverrideTestCase(unittest.TestCase):
                       " by your conanfile to libB/3.0@user/channel", self.t.out)
         self.assertIn("ERROR: Version range '<=2.0' required by 'libC/1.0@user/channel' not valid"
                       " for downstream requirement 'libB/3.0@user/channel'", self.t.out)
+
+
+class VersionRangeOverrideFailTestCase(unittest.TestCase):
+
+    def test(self):
+        # https://github.com/conan-io/conan/issues/7864
+        t = TestClient()
+        # t.run("config set general.revisions_enabled=True")
+        t.save({"conanfile.py": GenConanfile()})
+        t.run("create . gtest/1.8.0@PORT/stable")
+        print(t.out)
+        t.run("create . gtest/1.8.1@bloomberg/stable")
+        print(t.out)
+
+        t.save({"conanfile.py": GenConanfile().with_require("gtest/1.8.1@bloomberg/stable")})
+        t.run("create . bde/1.0@PORT/stable")
+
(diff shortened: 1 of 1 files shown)


conans/test/functional/graph/version_range_override_test.py
```diff
@@ -53,3 +53,31 @@ def test_override_version_range_outside(self):
...
+        t = TestClient()
+        t.save({"conanfile.py": GenConanfile()})
+        t.run("create . gtest/1.8.0@PORT/stable")
+        t.run("create . gtest/1.8.1@bloomberg/stable")
+
+        t.save({"conanfile.py": GenConanfile().with_require("gtest/1.8.0@PORT/stable")})
+        t.run("create . intermediate/1.0@PORT/stable")
+
+        t.save({"conanfile.py": GenConanfile().with_requires("intermediate/1.0@PORT/stable")
+                                              .with_build_requires("gtest/1.8.0@PORT/stable")})
```
jgsogo:
```suggestion
                                              .with_build_requirement("gtest/1.8.0@PORT/stable", force_host_context=True)})
```

This limitation/protection should apply when the `build_requires` belongs to the _build_ context, right?

memsharded: The user issue was with a regular declarative ``build_require`` in the build context. I can add another test with a build_require in the host context, it really doesn't matter much, the overriding is the same for both cases.

Issue #7863, opened 2020-10-12
[bug] Remote searches for channel-promoted packages are broken since 1.30.0


jasal82:
It seems that PR #7673 broke search queries which used to work in Conan < 1.30.0. We use the copy feature in Artifactory UI to promote packages from `master` to `release` channel. In previous versions Conan sent queries like

```
https://server/artifactory/api/conan/euler-conan-local/v2/conans/search?q=euler_threading&ignorecase=False
```

These returned all of the available packages with that name and filtered for channel on the client side. Now I see queries like

```
https://server/artifactory/api/conan/euler-conan-local/v2/conans/search?q=euler_threading%2F%2A%40sick%2Frelease&ignorecase=False
```

This works fine for packages which were uploaded with that exact user/channel combination. But when a package is promoted from channel `master` to `release` it is not part of the result for the query containing `release` in the expression. When I change the query to search for `master` (the original channel before promotion) the package is listed in the results but with the new (promoted) channel.

We're on Artifactory 6.21.0.

memsharded:
Hi @jasal82 

Just to clarify, are you using the private http api directly, or that is only the debugging information? Could you please clarify which Conan commands/recipes are you using that will trigger this issue? 

Also, please take into account that there was a bug related to this change, that has been fixed in Conan 1.30.1.

jasal82:
The queries are debugging traces from a simple `conan install .. --build missing --update`. It happens also with Conan 1.30.1.

jasal82:
It should be possible to reproduce this easily with any recipe by promoting its channel via Artifactory UI or JFrog CLI and then searching with a query similar to the above.

jasal82:
I think it's actually an Artifactory problem which was uncovered by the changes in the PR.

memsharded:
If the command is ``conan install .. --build missing --update``, then the search will be invoked by a version range. Such version range could be:

```python
requires = "pkg/[>1.0 <2.0]@user/channel"
# or
requires = "otherpkg/[>1.0 <2.0]"
```

In the first case, the version range should only resolve to packages that have the exact ``user/channel``, and in the second case it should only match packages without ``user/channel``. It if was matching other packages, it was a bug that was fixed.

I am trying with Artifactory 6.8, it seems to be fine:

```bash
$  conan install "hello/[>0.0]@james/testing" -r=local 
# and
$ conan install "hello/[>0.0]@" -r=local
```

Both manage to install the correct package with and without user/channel.

[...]

PR #7883, opened 2020-10-15
Let conan download preserve tarfiles in the cache

### LINKED ISSUE #7869: [feature] 'conan download': preserve tarfiles in the cache ('conan upload' to use it)
As of Conan v1.30, the command `conan download` does extract the packaged file (`conan_package.tgz`) into the Conan cache to make it ready to use. As a consequence, `conan upload` needs to package the files again before uploading to a remote repository.

This workflow is used as a workaround to _promote_ packages between different remote repositories, but it might modify the _package revision_ if these operations happen in different machines (see https://github.com/conan-io/conan/pull/7760#issuecomment-707041796 for information related to symlinks changes).

The proposal is to keep any file downloaded from a remote repository (whenever Conan downloads a package or install it from a remote) and use the same tarfile if available in the `conan upload` command. Some sentinel should be able to notice if the files in the cache have been modified and the tarfile needs to be invalidated/removed and generated again.

---

Details TBD (maybe 2.0).- Does `conan download` actually extract the files or only when they are going to be used (`conan install` using the package)? 

---

Side effect.- A workflow with `conan download + conan upload` will run much faster

czoido:
Changelog: Feature: Do not delete the downloaded `conan_package.tgz` when doing a conan download.
Docs: TODO?

The intention of this PR is to make possible a flow like doing conan download and conan upload without repackaging the conan_package.tgz each time Conan performs these operations. Besides the optimizations if will also enable the workaround commented here: https://github.com/conan-io/conan/issues/7863#issuecomment-707198363

Closes: https://github.com/conan-io/conan/issues/7869

#TAGS: slow
#REVISIONS: 1

### DIFF
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -279,9 +279,8 @@ def unzip_and_get_files(files, destination_dir, tgz_name, output):
     check_compressed_files(tgz_name, files)
     if tgz_file:
         uncompress_file(tgz_file, destination_dir, output=output)
-        file_name = os.path.basename(tgz_file)
         # https://github.com/conan-io/conan/issues/7869
-        if file_name != "conan_package.tgz":
+        if os.path.basename(tgz_file) != "conan_package.tgz":
             os.remove(tgz_file)
 
 
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -279,6 +279,10 @@ def unzip_and_get_files(files, destination_dir, tgz_name, output):
     check_compressed_files(tgz_name, files)
     if tgz_file:
         uncompress_file(tgz_file, destination_dir, output=output)
+        file_name = os.path.basename(tgz_file)
+        # https://github.com/conan-io/conan/issues/7869
+        if file_name != "conan_package.tgz":
+            os.remove(tgz_file)
 
 
 def uncompress_file(src_path, dest_folder, output):
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -279,7 +279,6 @@ def unzip_and_get_files(files, destination_dir, tgz_name, output):
     check_compressed_files(tgz_name, files)
     if tgz_file:
         uncompress_file(tgz_file, destination_dir, output=output)
-        os.remove(tgz_file)
 
 
 def uncompress_file(src_path, dest_folder, output):
--- a/conans/test/functional/cache/path_limit_test.py
+++ b/conans/test/functional/cache/path_limit_test.py
@@ -111,7 +111,7 @@ class PathLengthLimitTest(unittest.TestCase):
             files = os.listdir(package_folder)
             self.assertIn("myfile.txt", files)
             self.assertIn("myfile2.txt", files)
-            self.assertNotIn("conan_package.tgz", files)
+            self.assertIn("conan_package.tgz", files)
 
     def export_source_test(self):
         client = TestClient()
--- a/conans/client/generators/deploy.py
+++ b/conans/client/generators/deploy.py
@@ -9,7 +9,7 @@ from conans.paths import BUILD_INFO_DEPLOY
 from conans.util.files import mkdir, md5sum
 
 
-FILTERED_FILES = ["conaninfo.txt", "conanmanifest.txt", "conan_package.tgz"]
+FILTERED_FILES = ["conaninfo.txt", "conanmanifest.txt"]
 
 
 class DeployGenerator(Generator):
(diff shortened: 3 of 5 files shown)


conans/test/functional/cache/path_limit_test.py
```diff
@@ -111,7 +111,7 @@ def upload_test(self):
             files = os.listdir(package_folder)
             self.assertIn("myfile.txt", files)
             self.assertIn("myfile2.txt", files)
-            self.assertNotIn("conan_package.tgz", files)
+            self.assertIn("conan_package.tgz", files)
```
memsharded, line 114: I think this is concerning, to have the .tgz there, that might also be deployed with the ``-g deploy`` or ``deploy()`` methods.

conans/test/functional/command/info/info_test.py
```diff
@@ -354,8 +354,8 @@ def test_reuse(self):
                     Hello1/0.1@lasote/stable""")
 
         expected_output = expected_output % (
-                "\n    Revision: 4dfe7e755ac2ce2b39f1da54151c7636",
-                "\n    Revision: e003760cfa649c4ac4680fec3271b17a") \
+                "\n    Revision: 63865a1afa3a2666b2f75cbc7745e8a4",
```
memsharded: Why is this revision changing? This is also concerning

czoido: I think this revision change is related to this changes: https://github.com/conan-io/conan/pull/7783/files
The `conanfile.py` has changed from the previous time it was tested with revisions and is generating a new recipe revision from the manifest files.

conans/client/remote_manager.py
```diff
@@ -279,7 +279,9 @@ def unzip_and_get_files(files, destination_dir, tgz_name, output):
     check_compressed_files(tgz_name, files)
     if tgz_file:
         uncompress_file(tgz_file, destination_dir, output=output)
-        os.remove(tgz_file)
+        # https://github.com/conan-io/conan/issues/7869
+        if os.path.basename(tgz_file) != "conan_package.tgz":
```
memsharded, line 283: This is not enough. If I do a ``conan install -g deploy`` I get a copy in my user folder, which is most likely unwanted/unexpected:
```
1602790590
C:\Users\memsharded\conanws\kk\zlib\conan_package.tgz: c1853f575112001ebb494c9fc85ad3e5
C:\Users\memsharded\conanws\kk\zlib\include\zconf.h: e5b18c411a45727e543081145b983bbc
C:\Users\memsharded\conanws\kk\zlib\include\zlib.h: 0338828e9d00c94645648b1517108324
C:\Users\memsharded\conanws\kk\zlib\lib\zlib.lib: 94143d12f4e75d9aa11e725d7ff73c11
C:\Users\memsharded\conanws\kk\zlib\licenses\LICENSE: 1af2e5353c16e930435fff550f827c3e
```

czoido: My first thought was to filter `conan_package.tgz` in the deploy generator (like it's done with `conaninfo.txt` and `conanmanifest.txt`) but then I realized that, as things work now, if you do a `conan upload` and then `conan install` with the deploy generator you will also get a `conan_package.tgz` in the deployment so I allowed the deploy generator to copy the `conan_package.tgz` not to change how it works now.
What would be the correct behaviour?

conans/client/remote_manager.py
```diff
@@ -279,7 +279,10 @@ def unzip_and_get_files(files, destination_dir, tgz_name, output):
     check_compressed_files(tgz_name, files)
     if tgz_file:
         uncompress_file(tgz_file, destination_dir, output=output)
-        os.remove(tgz_file)
+        # move the .tgz to another place
+        tgzdir = destination_dir + "_tgz"
+        mkdir(tgzdir)
+        shutil.move(tgz_file, os.path.join(tgzdir, tgz_name))
```
czoido: The problem I see storing the file in another folder is some cases like this that you are using revisions and do this operations:

```
conan new hello/0.1
conan create .
conan upload hello/0.1 -r artifactory -c --all
conan download hello/0.1@ -r artifactory 
echo "# new revision" >> conanfile.py 
conan export .
```

You end up with a folder `hello/0.1/_/_/package/package_id_tgz`, but the original has been removed because Conan already took care of removing that folder in certain situations... so it's necessary now to also take care of handling this folder which maybe adds a little bit more complexity.

Before Conan was removing the `package_id` folder so the downloaded file would also be deleted.

WDYT?

memsharded:
@czoido Sorry,  my mistake, I didn't intend to push these commits here, but to a new branch to explore the idea.
I will revert the changes and push clean your previous state.

PR #7886, opened 2020-10-15
Feature/keep tgz download

memsharded:
Changelog: Feature: Download and keep the conan_package.tgz in the cache, so they are not affected by different Operating Systems compression and de-compression and uploading is way more efficient.
Docs: Omit

The same for other .tgz files will follow in separate PR in: https://github.com/conan-io/conan/pull/7938

#tags: slow
#revisions: 1


czoido (approved):

conans/client/cmd/copy.py
```diff
@@ -79,6 +79,13 @@ def package_copy(src_ref, user_channel, package_ids, cache, user_io, short_paths
         shutil.copytree(package_path_origin, package_path_dest, symlinks=True)
         user_io.out.info("Copied %s to %s" % (str(package_id), str(dest_ref)))
 
+    # The downloaded .tgz files can also be copied
+    src_download = src_layout.download_folder()
+    dst_download = dst_layout.download_folder()
+    rmdir(dst_download)
+    if os.path.isdir(src_download):
+        shutil.copytree(src_download, dst_download)
```
jgsogo: We should only remove those listed in the `package_ids` list (they are already removed in the line 75) and copy only the files corresponding to those.

conans/client/migrations.py
```diff
@@ -122,6 +125,27 @@ def _get_prefs(layout):
...
+    will be compressed again
+    """
+    out.info("Removing temporary .tgz files, they are stored in a different location now")
+    refs = _get_refs(cache)
+    for ref in refs:
+        try:
+            base_folder = os.path.normpath(os.path.join(cache.store, ref.dir_repr()))
+            for d, _, fs in os.walk(base_folder):
+                for f in fs:
+                    if f in (EXPORT_SOURCES_TGZ_NAME, EXPORT_TGZ_NAME, PACKAGE_TGZ_NAME):
```
jgsogo: Not yet, it will go in the other PR

```suggestion
                    if f in (PACKAGE_TGZ_NAME, ):
```

conans/client/remote_manager.py
```diff
@@ -133,37 +133,49 @@ def get_package(self, pref, layout, remote, output, recorder):
...
+
+            tgz_file = zipped_files.pop(PACKAGE_TGZ_NAME, None)
+            check_compressed_files(PACKAGE_TGZ_NAME, zipped_files)
+            package_folder = layout.package(pref)
+            if tgz_file:  # This must happen always, but just in case
+                # TODO: The output could be changed to the package one, but
+                uncompress_file(tgz_file, package_folder, output=self._output)
+            mkdir(package_folder)  # Just in case it doesn't exist, because uncompress did nothing
+            for file_name, file_path in zipped_files.items():  # copy CONANINFO and CONANMANIFEST
+                os.rename(file_path, os.path.join(package_folder, file_name))
```
jgsogo, line 168: Why are we moving these files now?

conans/client/remote_manager.py
```diff
@@ -133,37 +133,49 @@ def get_package(self, pref, layout, remote, output, recorder):
...
             pref = self._resolve_latest_pref(pref, remote)
             snapshot = self._call_remote(remote, "get_package_snapshot", pref)
             if not is_package_snapshot_complete(snapshot):
                 raise PackageNotFoundException(pref)
-            zipped_files = self._call_remote(remote, "get_package", pref, dest_folder)
 
-            package_checksums = calc_files_checksum(zipped_files)
+            download_pkg_folder = layout.download_package(pref)
+            # Download files to the pkg_tgz folder, not to the final one
+            zipped_files = self._call_remote(remote, "get_package", pref, download_pkg_folder)
```
jgsogo, line 147: Somewhere around these lines, we need to set/unset the dirty flag to the downloaded file, it can fail during the download. I think the `get_package` call is only assigning the dirty flag to the package folder.
I see we are no longer removing the folder in the `except BaseException`

jgsogo (changes requested):

conans/client/remote_manager.py
memsharded: Because they are being downloaded to the "download" folder, and these files need to be in the final package folder. There were 2 options:
- Download to the "package folder", then move the .tgz to other place (the download folder)
- Download to the "download folder", leave the .tgz there and move the other files to the package folder.

I have decided the second one, to don't move the "heavy" .tgz file.

conans/client/remote_manager.py
memsharded: That is the thing, as long as the whole package_folder is protected with a "dirty" flag, everything inside is dirty. There was not dirty flag for this download before, I would say it is not necessary. Probably the other dirty flags over individual .tgz files could also be removed, but I didn't want to change more things in this PR.

The removal of the folder also doesn't make sense, and apparently no test is covering it. If we have the "dirty" mechanism in place to protect interrupted processes, then it shouldn't be necessary to try to remove the folder (in some cases, it could even be annoying, as it won't let debug)

memsharded (commented):
Reviewed the PR:
- Removed the non related parts of the migration
- Fixed the copy (only affected package-ids)

Please check the comments

conans/client/remote_manager.py
jgsogo: Haven't tried it, but this is my concern:
 * Download starts
   + `package_folder` marked as dirty
 * Download fails
   + `package_folder` mark is still there
   + We have a broken file in `<layout>/download_folder/pid/xxxxxx.tgz` (not dirty)
 * Upload the package: lines in the uploader are checking the dirty for the tgz file, not the `package_folder`, will it upload a broken file?
   ```python
        download_pkg_folder = layout.download_package(pref)
        package_tgz = os.path.join(download_pkg_folder, PACKAGE_TGZ_NAME)
        if is_dirty(package_tgz):
            self._output.warn("%s: Removing %s, marked as dirty" % (str(pref), PACKAGE_TGZ_NAME))
            os.remove(package_tgz)
            clean_dirty(package_tgz)
       ... 
   ```

Without running the test I'm not sure if this is something that can happen or not.

memsharded: The code is doing both:

```python
    def _compress_package_files(self, layout, pref, integrity_check):
        t1 = time.time()
        if layout.package_is_dirty(pref):
            raise ConanException("Package %s is corrupted, aborting upload.\n"
                                 "Remove it with 'conan remove %s -p=%s'"
                                 % (pref, pref.ref, pref.id))

        download_pkg_folder = layout.download_package(pref)
        package_tgz = os.path.join(download_pkg_folder, PACKAGE_TGZ_NAME)
        if is_dirty(package_tgz):
            self._output.warn("%s: Removing %s, marked as dirty" % (str(pref), PACKAGE_TGZ_NAME))
            os.remove(package_tgz)
            clean_dirty(package_tgz)
```

First, it will raise if the package is dirty (as it doesn't know how to recover from that). If only the .tgz is dirty, it can try to remove it and re-compress.

memsharded: The test in ``test_upload_dirty()`` covers the case where the package is marked as dirty.

jgsogo: I see, we are covered with that `raise` when the package_folder is dirty. 👍

jgsogo (approved):

mietzen:
@memsharded Is it possible to turn this feature of? Some of our Devs don't want to store redundant data on their drives.

memsharded:
Hi @ngerke 

At the moment it is not possible. Partly, because this is a bit more than a feature, but it is also fixing some bugs. Read the rationale in the blog post: https://blog.conan.io/2020/11/05/New-conan-release-1-31.html. 

Could you please give an order of magnitude of your developer caches sizes? Trying to understand a bit the pain and  size of the issue.

In any case, the solution is likely to be based more on the ongoing work with the "LRU", and new features in order to keep the cache size limited. Have a look at: https://github.com/conan-io/conan/pull/7980

mietzen:
Our main product has a pkg size of ~ 2 GB (worst case) the SDK pkg ~ 300 MB. Also we have a other pkgs (e.g. Qt) with ~ 300MB if you store multiple versions of the pkgs, cache size of 20-30 GB are not unusual (Without tar.gz pkgs).
I will have a look at "LRU", is there any documentation about it?

Issue #7587, opened 2020-08-25
[feature] Overcome Windows system PATH length limitation


db4:
Our internally developed software consists of many dynamic libraries, each of them placed into a single Conan package.
So on Windows Conan modifies run environment adding to system PATH something like
```
c:\Users\ContainerAdministrator\.conan\data\package-name\1.0.0\company-name\testing\package\xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\bin (about 140 chars)
```
This is done for every package. But PATH on Windows has a very strict limitation: its length cannot exceed 8192 characters. Once it's reached, `cmd.exe` will fail to create a new process. About 2500 of them takes Visual Studio Developer Command Prompt; it leaves less than 6000 chars for a user (if he/she doesn't have any 3rd-party application installed), or about 40 Conan packages if `short_paths=True` is not used. I think it makes Windows development using Conan very problematic. What about another approach: place .dll/.exe symlinks into a single bin directory or even use [shim generator](https://github.com/chocolatey/shimgen) as Chocolatey does?

jgsogo:
Hi, @db4 

For other reason (but with the same effect) I'm proposing to add wrappers around the executables that are provided by packages (https://github.com/conan-io/conan/pull/7324). As a side effect, only the path to the wrappers would be added to PATH and probably they should be all together closer to the working directory.

We need to keep working on that approach.

db4:
Hi @jgsogo! 

> For other reason (but with the same effect) I'm proposing to add wrappers around the executables that are provided by packages (#7324)

That's great, but what about DLLs, provided by packages? They also should be in PATH. Why symlinking .exe/.dll into a single bin directory is not an option?

memsharded:
Closing this ticket as outdated, Conan 2.0 cache new layout will avoid too long paths

PR #7891, opened 2020-10-16
Remove duplicate entries while modifying PATH-like environment variables

db4:
Changelog: Fix: Remove duplicate entries while modifying PATH-like environment variables internally. Especially important for Windows where system PATH size is limited by 8192 charachers (when using cmd.exe).
Docs: omit

Partially solves https://github.com/conan-io/conan/issues/7587, especially when `conan` is launched from MSVC develper command prompt.

### DIFF
--- a/conans/client/tools/env.py
+++ b/conans/client/tools/env.py
@@ -1,6 +1,7 @@
 import os
 import platform
 import sys
+from collections import OrderedDict
 from contextlib import contextmanager
 
 from conans.client.run_environment import RunEnvironment
@@ -65,6 +66,9 @@ def _environment_add(env_vars, post=False):
                     apply_vars[name] = old + os.pathsep + apply_vars[name]
                 else:
                     apply_vars[name] += os.pathsep + old
+            # Remove possible duplicates, keeping the order of the remaining paths
+            items = apply_vars[name].split(os.pathsep)
+            apply_vars[name] = os.pathsep.join(OrderedDict.fromkeys(items))
         else:
             apply_vars[name] = value
 
--- a/conans/client/envvars/environment.py
+++ b/conans/client/envvars/environment.py
@@ -1,6 +1,7 @@
 import os
 import platform
 import textwrap
+from collections import OrderedDict
 
 from jinja2 import Template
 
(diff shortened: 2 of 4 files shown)


conans/client/tools/env.py
```diff
@@ -65,6 +69,7 @@ def _environment_add(env_vars, post=False):
                     apply_vars[name] = old + os.pathsep + apply_vars[name]
                 else:
                     apply_vars[name] += os.pathsep + old
+            apply_vars[name] = os.pathsep.join(_unique(apply_vars[name].split(os.pathsep)))
```
memsharded: It seems that it can be done in place with a more common python idiom like:

```python
items = apply_vars[name].split(os.pathsep)
apply_vars[name] = os.pathsep.join(OrderedDict.fromkeys(items))
```

memsharded (commented):
Could you please add a test that will cover this case? Ask for guidance if it is not evident how/where.

Thanks for the contribution!

conans/client/tools/env.py
db4: > 
> 
> It seems that it can be done in place with a more common python idiom like:
> 
> ```python
> items = apply_vars[name].split(os.pathsep)
> apply_vars[name] = os.pathsep.join(OrderedDict.fromkeys(items))
> ```

Yes, indeed. I was under the impression that `OrderedDict` was the Python3-only thing but actually it's included in Python 2.7 as well.

memsharded: Great. If you could manage to add some test (ask for help if necessary), this can be merged for next Conan 1.31.

db4: > Could you please add a test that will cover this case? Ask for guidance if it is not evident how/where.

Yes, please. I'm not familiar with Conan testing framework. Is  https://github.com/conan-io/conan/blob/4f9822deb2ef4fc7cd8b50a5091beff831dd546e/conans/test/unittests/client/tools/win/vcvars_test.py#L60 the correct place for these additional tests? And how to run Conan tests locally?

memsharded: It seems that the ``VirtualEnvGeneratorTest`` in conans/test/unittests/client/generators/virtualenv_test.py would be a good place, adding another variable with repeated paths.

Locally, you can run it with: ``nosetests conans/test/unittests/client/generators/virtualenv_test.py``. You might need to ``pip install -r requirements_dev.txt`` and ``pip install -r requirements_server.txt`` if missing deps. Please let me know if this helps, otherwise I can contribute the test myself. Thanks again for the contribution!

db4: > Please let me know if this helps, otherwise I can contribute the test myself.

Looks like sorting it out needs some time but right now I'm a little bit busy. If you can write tests yourself with little efforts please do!

jgsogo (approved):
Thanks a lot for the contribution, @db4 🎉

PR #7900, opened 2020-10-18
Refactor/rm package

memsharded:
Changelog: Omit
Docs: Omit

This is an internal refactor. It abstracts the operations:

- Remove package binary from the cache
- Set/check/unset dirty of a package binary in the cache
- Check if a package binary exists in the cache

These operations are at the moment distributed in the codebase based on folder operations. But as the cache changes, (for example if we try to create a different storage for .tgz files), everything breaks. It seems that we need to abstract the operations in the cache. This would be only the part of the package binaries, the problem remains for all other parts: exports, export_sources, sources...

#tags: slow
#revisions: 1


conans/client/installer.py
```diff
@@ -200,7 +199,6 @@ def build_package(self, node, keep_build, recorder, remotes):
 
         # BUILD & PACKAGE
         with package_layout.conanfile_read_lock(self._output):
-            _remove_folder_raising(package_folder)
```
memsharded, line 203: This is the only possible minor functional change. The ``package_folder`` was removed in this point when building from source. Now it is removed earlier in the process, as soon as the build will start.

conans/client/installer.py
```diff
@@ -488,14 +481,14 @@ def _handle_node_cache(self, node, keep_build, processed_package_references, rem
...
         layout = self._cache.package_layout(pref.ref, conanfile.short_paths)
-        package_folder = layout.package(pref)
 
         with layout.package_lock(pref):
             if pref not in processed_package_references:
                 processed_package_references.add(pref)
                 if node.binary == BINARY_BUILD:
                     assert node.prev is None, "PREV for %s to be built should be None" % str(pref)
-                    with set_dirty_context_manager(package_folder):
+                    layout.package_remove(pref)
```
memsharded, line 490: This is the place of removal of the package folder, instead of inside the ``_build_package()``. This has been changed because the ``set_dirty_context_manager()`` could leave it in an inconsistent state, when removing the package folder, could remove accidentally the .dirty flag.

conans/paths/package_layouts/package_cache_layout.py
```diff
@@ -110,6 +110,43 @@ def package(self, pref):
...
+
+    def package_set_dirty(self, pref):
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        return set_dirty(pkg_folder)
+
+    def package_clean_dirty(self, pref):
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        return clean_dirty(pkg_folder)
+
+    def package_id_exists(self, package_id):
```
memsharded, line 132: There is a similar ``package_exists()`` method, but I didn't want to change it. That one is doing checks of the revisions, for the ``conan remove`` functionality, better not to mess with it.

conans/client/remote_manager.py
```diff
@@ -147,6 +155,7 @@ def get_package(self, pref, dest_folder, remote, output, recorder):
                 metadata.packages[pref.id].revision = pref.revision
                 metadata.packages[pref.id].recipe_revision = pref.ref.revision
                 metadata.packages[pref.id].checksums = package_checksums
+                metadata.packages[pref.id].remote = remote.name
```
memsharded, line 158: This will assign the remote to ``conan download`` binaries. So far it seems it was not assigned, which seems a bit inconsistent, as the defined behavior is that packages keep the remote from which it was installed from.

conans/client/installer.py
```diff
@@ -397,15 +397,8 @@ def _download(n):
                 _download(node)
 
     def _download_pkg(self, layout, pref, node):
-        conanfile = node.conanfile
-        package_folder = layout.package(pref)
-        output = conanfile.output
-        with set_dirty_context_manager(package_folder):
```
memsharded, line 403: The ``dirty_context_manager`` was only applied for ``conan install`` command, but not for ``conan download`` commands, which could also leave the packages broken if interrupted. This has been moved to ``RemoteManager``, so no matter who calls the download of a package, the ``.dirty`` mechanism will be used.

conans/paths/package_layouts/package_cache_layout.py
jgsogo: I think it is the short_paths one, for example here:

image

It was `dst_layout.package(...)` and this function has the `@short_path` decorator applied.


Maybe, after changing the `package_remove` call, now it removes the one in the cache and we can perform the check on this folder instead of checking for the short one?

conans/client/remover.py
```diff
@@ -76,7 +76,7 @@ def remove_packages(self, package_layout, ids_filter=None):
                 if not package_layout.package_exists(pref):
                     raise PackageNotFoundException(pref)
                 pkg_folder = package_layout.package(pref)
-                self._remove(pkg_folder, package_layout.ref, "package:%s" % package_id)
+                package_layout.package_remove(pref)
                 self._remove_file(pkg_folder + ".dirty", package_layout.ref, "dirty flag")
                 self._remove_file(package_layout.system_reqs_package(pref), package_layout.ref,
                                   "%s/%s" % (package_id, SYSTEM_REQS))
```
jgsogo, line 82: Is this something we can move into the `package_remove` function? I would say that I can remove always the `.dirty` check and the system-requires associated with a package-ID if I remove the package-ID from the cache, right?

conans/client/remote_manager.py
jgsogo: 👍 

It was assigned for the `conan install`, but not for the `conan download`, bug or feature? 🤔  I prefer to be consistent.

conans/paths/package_layouts/package_cache_layout.py
```diff
@@ -110,6 +110,43 @@ def package(self, pref):
         assert pref.ref == self._ref, "{!r} != {!r}".format(pref.ref, self._ref)
         return os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
 
+    @contextmanager
+    def set_dirty_context_manager(self, pref):
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        set_dirty(pkg_folder)
```
jgsogo, line 116: What if the package is already _dirty_? Should we raise an exception? Same comment for the `package_set_dirty` method. Not for this PR because it wouldn't be a refactor, but wdyt?

conans/paths/package_layouts/package_cache_layout.py
```diff
@@ -110,6 +110,43 @@ def package(self, pref):
...
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        return is_dirty(pkg_folder)
+
+    def package_set_dirty(self, pref):
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        return set_dirty(pkg_folder)
+
+    def package_clean_dirty(self, pref):
+        pkg_folder = os.path.join(self._base_folder, PACKAGES_FOLDER, pref.id)
+        return clean_dirty(pkg_folder)
```
jgsogo, line 130: If `package_remove` removes also the `.dirty` sentinel for the package-ID, I think that methods `package_set_dirty` and `package_clean_dirty` won't be needed, only the context manager one, which is much safer for this kind of job.

conans/client/installer.py
```diff
@@ -509,6 +502,8 @@ def _handle_node_cache(self, node, keep_build, processed_package_references, rem
                     log_package_got_from_local_cache(pref)
                     self._recorder.package_fetched_from_cache(pref)
 
+            package_folder = layout.package(pref)
+            assert os.path.isdir(package_folder), "Package folder must exist %s" % package_folder
```
jgsogo: Add more info to the message: if it is using short-paths here we will get only an obfuscated path. Add the `pref` itself.

conans/paths/package_layouts/package_cache_layout.py
jgsogo: Nevertheless, what happens if I remove manually the short_paths folder? In `develop` this function would return False (it was checking the short-path), now it will return True (checking the one in the cache)

jgsogo (changes requested):
Some suggestions here and there

Doubt about the `package_id_exists` method, check the comment there: now it is checking the _standard cache_ path, it doesn't realize if accidentally the user has removed the short-paths folder.

Overall, I really like it,  I want to see this PR merged ASAP 🎉

conans/client/remote_manager.py
memsharded: To be honest, this was a surprise to me. That you did a ``conan download`` and the remote was not assigned, I observed it as a buggy behavior. Even more if there is a chance that such binary (maybe a previous revision) came from another remote, and maintains that remote. So I think we should be consistent and assign the remote: bug.

conans/paths/package_layouts/package_cache_layout.py
memsharded: yes, most likely, it would be great to add at least an assert, as in theory it shouldn't be possible. But an exception for readability from users would be good too, just in case.

conans/paths/package_layouts/package_cache_layout.py
memsharded: Yes, I have considered doing this. But this might slightly change the behavior, so I wanted to wait for feedback first.
Should I try in this PR or maybe in a new one after this?

memsharded:
> Doubt about the package_id_exists method, check the comment there: now it is checking the standard cache path, it doesn't realize if accidentally the user has removed the short-paths folder.

Yes, the existence of the package is based on the existence of a folder for regular packages. Lets make the same for short-paths packages. Yes, packages in the normal cache internally can be corrupted too, they could remove the manifest or conaninfo.txt or something, or someone could remove the short paths storage. But I think that should be another orthogonal mechanism (for both cases, regular and short paths cache).

Yes, the ``rm_package()`` clears both the short-paths if it exists, and also the folder in the cache, so ``package_id_exists()`` seems to be very consistent.

In the same way, the ".dirty" checks and storage of the ``conan_package.tgz`` has been moved to the regular cache. I think it is way more consistent and robust.

I am introducing a test removing the short-paths cache, to see the impact and error messages of this change.

lasote (approved):
I like it, go ahead with the @jgsogo comments and merge when green.

Just a thought for the future: Should we go even further with the file system access abstraction? With this PR we are abstracting part of it, so the operations that affect the package "as a whole" are encapsulated, but we still manage disk operations based on single folders. It might be difficult for the team to distinguish where the limit is. 
If you don't have in mind the possibility of moving the "tgz" to a different place is difficult to understand, for example, why I can get the package folder from the layout to unzip something but why we should use the encapsulated functions to check if the package exists.

memsharded:
> Just a thought for the future: Should we go even further with the file system access abstraction? With this PR we are abstracting part of it, so the operations that affect the package "as a whole" are encapsulated, but we still manage disk operations based on single folders. It might be difficult for the team to distinguish where the limit is.
If you don't have in mind the possibility of moving the "tgz" to a different place is difficult to understand, for example, why I can get the package folder from the layout to unzip something but why we should use the encapsulated functions to check if the package exists.

Yes, totally agree. The scope of this PR is limited, it is very likely that we want to keep moving in this direction. Especially because other challenges too, if we want to have a multi-revision cache for 2.0, we will need even better abstraction of the cache and package layouts.

jgsogo (approved):
👍  Those two comments can wait for another PR.

PR #7901, opened 2020-10-18
Add export-all command

sztomi:
Changelog: Feature: This PR adds an `export-all` command that allows exporting a monorepo of recipes quickly. For comparison, we are currently using an external script that launches multiple `conan export` commands in parallel. Because a no-op conan call is already about 500ms (command.py is HUGE), we repeatedly spend a lot of time redoing the same things in conan startup. The idea is to only do these things once and then repeatedly (in parallel) export the recipes.

This yields a 19x speedup for the same monorepo (containing 83 recipes):

```
❯ hyperfine -w 3 'rm -rf ./conan_home/.conan/data && ./dep_builder.py -o'
Benchmark #1: rm -rf ./conan_home/.conan/data && ./dep_builder.py -o
  Time (mean ± σ):     34.395 s ±  0.660 s    [User: 190.776 s, System: 11.921 s]
  Range (min … max):   33.169 s … 35.532 s    10 runs

❯ hyperfine -w 3 'rm -rf ./conan_home/.conan/data && python -m conans.conan export-all packages plex/stable'
Benchmark #1: rm -rf ./conan_home/.conan/data && python -m conans.conan export-all packages plex/stable
  Time (mean ± σ):      1.879 s ±  0.039 s    [User: 1.773 s, System: 0.957 s]
  Range (min … max):    1.826 s …  1.945 s    10 runs
```

This is a draft PR because this implementation is the bare minimum:

  * the output from each parallel run is interleaved
  * the threadpool is hardcoded to 8 threads

I'd like to get your opinion on this before I continue and solve the above. Would you be willing to add a command like this?

### DIFF
--- a/conans/client/command.py
+++ b/conans/client/command.py
@@ -1066,6 +1066,49 @@ class Command(object):
                                   lockfile_out=args.lockfile_out,
                                   ignore_dirty=args.ignore_dirty)
 
+    def export_all(self, *args):
+        """
+        Exports multiple recipes from a monorepo.
+
+        Each subdirectory of the given path will be considered a reciped folder.
+        """
+        parser = argparse.ArgumentParser(description=self.export.__doc__,
+                                         prog="conan export",
+                                         formatter_class=SmartFormatter)
+        parser.add_argument("path", help=_PATH_HELP)
+        parser.add_argument("reference", nargs='?', default=None,
+                            help="user/channel, Pkg/version@user/channel (if name "
+                                 "and version are not declared in the conanfile.py) "
+                                 "Pkg/version@ if user/channel is not relevant.")
+        parser.add_argument('-k', '-ks', '--keep-source', default=False, action='store_true',
+                            help=_KEEP_SOURCE_HELP)
+        parser.add_argument("-l", "--lockfile", action=OnceArgument,
+                            help="Path to a lockfile file.")
+        parser.add_argument("--lockfile-out", action=OnceArgument,
+                            help="Filename of the updated lockfile")
+        parser.add_argument("--ignore-dirty", default=False, action='store_true',
+                            help='When using the "scm" feature with "auto" values, capture the'
+                                 ' revision and url even if there are uncommitted changes')
+
+        args = parser.parse_args(*args)
+        self._warn_python_version()
+        if args.lockfile_out and not args.lockfile:
+            raise ConanException("lockfile_out cannot be specified if lockfile is not defined")
+
+        name, version, user, channel, _ = get_reference_fields(args.reference,
+                                                               user_channel_input=True)
+
+        if any([user, channel]) and not all([user, channel]):
+            # Or user/channel or nothing, but not partial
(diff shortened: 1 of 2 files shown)


conans/client/command.py
```diff
@@ -1066,6 +1066,49 @@ def export(self, *args):
...
+        """
+        parser = argparse.ArgumentParser(description=self.export.__doc__,
+                                         prog="conan export",
+                                         formatter_class=SmartFormatter)
+        parser.add_argument("path", help=_PATH_HELP)
+        parser.add_argument("reference", nargs='?', default=None,
+                            help="user/channel, Pkg/version@user/channel (if name "
+                                 "and version are not declared in the conanfile.py) "
+                                 "Pkg/version@ if user/channel is not relevant.")
+        parser.add_argument('-k', '-ks', '--keep-source', default=False, action='store_true',
```
memsharded, line 1083: Is the --keep-source very necessary for this use case? Maybe not? I would try to simplify and remove it.

conans/client/command.py
```diff
@@ -1066,6 +1066,49 @@ def export(self, *args):
...
+        """
+        Exports multiple recipes from a monorepo.
+
+        Each subdirectory of the given path will be considered a reciped folder.
+        """
+        parser = argparse.ArgumentParser(description=self.export.__doc__,
+                                         prog="conan export",
+                                         formatter_class=SmartFormatter)
+        parser.add_argument("path", help=_PATH_HELP)
+        parser.add_argument("reference", nargs='?', default=None,
```
memsharded, line 1079: Reference is unused

conans/client/conan_api.py
```diff
@@ -842,6 +842,37 @@ def export(self, path, name, version, user, channel, keep_source=False, cwd=None
...
+            graph_lock = graph_lock_file.graph_lock
+            self.app.out.info("Using lockfile: '{}'".format(lockfile))
+
+        self.app.load_remotes()
+
+        from concurrent.futures import as_completed, ThreadPoolExecutor
+        from pathlib import Path
+
+        futures = []
+        path = Path(path)
```
memsharded, line 861: We probably want to keep the code Py2 compatible, it won't cost much to use old style ``os.path`` syntax to work.

conans/client/conan_api.py
```diff
@@ -842,6 +842,37 @@ def export(self, path, name, version, user, channel, keep_source=False, cwd=None
...
+        conanfiles = [_get_conanfile_path(str(subdir.absolute()), cwd, py=True) for subdir in path.iterdir()]
+        with ThreadPoolExecutor(max_workers=8) as pool:
```
memsharded, line 863: Use configurable ``cpu_count()`` as parameter.

memsharded (commented):
The feature would need thorough, comprehensive testing:

- Making sure resulting lockfiles are consistent.
- ignore-dirty cases

sztomi:
After some discussion in slack DMs, I'll be closing this. We will instead use the (experimental) conan API to implement the same.

PR #7910, opened 2020-10-20
Set default storage folder to $CONAN_USER_HOME/.conan/data

fplk0:
Changelog: Bugfix: Set default storage_folder to .conan/data in case if storage_path entry fails to be defined by conan.conf.
Docs: Omit

Set default storage_folder to .conan/data in case if storage/path entry is missing from the conan.conf. Usage of .conan for storage root seems to be a mistake and leads to failures in case of various operations, like conan config install <some_relatively_deep_git_repo>

We've noticed that sometimes when doing conan install from a git repo, we're getting an error like:
```
conan config install <sample-repo>
Trying to clone repo: <sample-repo>
Repo cloned!
Defining remotes from remotes.txt
ERROR: Failed conan config install: Value provided for package version, '.git' (type Version), is an invalid name. Valid names MUST begin with a letter, number or underscore, have between 2-51 chars, including letters, numbers, underscore, dot and dash
```

It turned out that, if [storage] / path entry in conan.conf is missing, the default storage path would be set to $CONAN_USER_HOME/.conan ; which looks to be incorrect. Default per [here](https://github.com/conan-io/conan/blob/develop/conans/client/conf/__init__.py#L189) seems to be $CONAN_USER_HOME/.conan/data . 
Later during conan config install command execution, conan tries to traverse all packages to fix the remotes, but given that it clones the config repo to the temporary directory within .conan dir, it tries to read packages in the temporarily cloned repo and fails with the error above.
Adding "data" at the end of the path fixes the error.

### DIFF
--- a/conans/client/cache/cache.py
+++ b/conans/client/cache/cache.py
@@ -76,7 +76,7 @@ class ClientCache(object):
         self._config = None
         self.editable_packages = EditablePackages(self.cache_folder)
         # paths
-        self._store_folder = self.config.storage_path or self.cache_folder
+        self._store_folder = self.config.storage_path or os.path.join(self.cache_folder, "data")
         # Just call it to make it raise in case of short_paths misconfiguration
         _ = self.config.short_paths_home
 
--- a/conans/test/functional/command/config_install_test.py
+++ b/conans/test/functional/command/config_install_test.py
@@ -14,6 +14,7 @@ from conans.client.conf import ConanClientConfigParser
 from conans.client.conf.config_installer import _hide_password, _ConfigOrigin
 from conans.client.rest.file_downloader import FileDownloader
 from conans.errors import ConanException
+from conans.test.utils.genconanfile import GenConanfile
 from conans.test.utils.test_files import temp_folder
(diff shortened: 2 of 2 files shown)


memsharded:
Hi @Sfairat 

Thanks for contributing this. I think it would be necessary to have a red/green test for this fix, so we make sure that this is both fixed correctly and we don't break it in the future. Please have a look to the test suite, if the bug is fired by ``conan config install`` then ``config_install_test.py`` could be a good starting point. Don't hesitate to ask for guidance or help, we could contribute the tests ourselves too.

fplk0:
@memsharded Thanks for the feedback - I was convinced that tests are necessary here. Will add later this week.

memsharded:
> @memsharded Thanks for the feedback - I was convinced that tests are necessary here. Will add later this week.

Excellent. Conan 1.31 is planned for next week, if we could have the tests this week, this could be included. Adding it to the milestone. Thanks again!

memsharded:
Hi @Sfairat 

We are doing 1.31 this week, could you please add a test for this fix? If not possible, we might try to contribute the test, please tell. Thanks!

PR #7754, opened 2020-09-25
Feature/msbuild toolchain generator

memsharded:
Changelog: Feature: Allow ``MSBuildToolchain`` custom configurations in ``generate()`` method.
Docs: https://github.com/conan-io/docs/pull/1957

Note:  the docs of this slipped into 1.32 release, the Docs PR is already merged.

PR #7915, opened 2020-10-20
[feature][POC] toolchain cmake custom multi configurations

### LINKED ISSUE #7999: [question] Mapped configs in cmake_find_package_multi.
Hello,
I'm hitting a wall trying to consume the GTest conan package from a custom cmake config.

We are using a custom cmake config called Hybrid (Release with asserts). When consuming packages from conan, I try to map the imported configs to Release.

For ex.
```cmake
# Create custom config
set(CMAKE_CONFIGURATION_TYPES Debug Release Hybrid)

# Use release for our imported targets with find_package.
set(CMAKE_MAP_IMPORTED_CONFIG_HYBRID Release)

# Somewhere
set(CMAKE_MODULE_PATH ${CMAKE_BINARY_DIR} ${CMAKE_MODULE_PATH})
set(CMAKE_PREFIX_PATH ${CMAKE_BINARY_DIR} ${CMAKE_PREFIX_PATH})

# Later
find_package(GTest CONFIG REQUIRED)
target_link_libraries(my_tests PRIVATE GTest::GTest)
```

`find_package` does try to pickup the Release config for GTest::GTest. However, it fails with the error :
```
LINK : fatal error LNK1104: cannot open file 'CONAN_LIB::GTest_gtest_main_gtest_mainRELEASE-NOTFOUND.obj'
```

The build does work in Release and Debug, so I'm guessing something isn't getting picked up when building with our Hybrid config. Any help welcome, thx.

edit :

The conanfile.
```
[requires]
gtest/1.10.0@_/_

[generators]
cmake_find_package_multi

[options]
gtest:build_gmock=False

[imports]
bin, *.pdb -> ./bin
bin, *.pdb -> ./lib
bin, *.dll -> ./bin
```

memsharded:

Fix https://github.com/conan-io/conan/issues/7999

Similar approach than https://github.com/conan-io/conan/pull/7754, but for CMake custom configurations.
Biggest change in this PR is actually just templatizing the configurations in ``cmake_find_package_multi`` generator


jgsogo:
I really think changes in the generator side are nice and needed, I would split them to a different PR and merge it: I like to be able to list/add more configurations and the refactor to the template is reducing duplicated code.

I'm not sure how much useful it is to be able to assign a proper name to a set of settings/options.... but it is ok, it helps the toolchain to work together with the generator.


```
conan install . -s build_type=Debug
conan install . -s build_type=Release
conan install . -s build_type=Release -o hello:shared=True  # <-- generates 'releaseshared'

cmake --build . --config releaseshared
```

> Side note.- I wonder if there could be case-sensitive problems


But 😮 😮 😮 , there is something  I didn't realize at first in the previous commands (if used with the recipe provided in the tests within this PR):

```python
        class App(ConanFile):
            requires = "hello/0.1"

            def toolchain(self):
                generator = CMakeFindPackageMultiGenerator(self)
                generator.configurations.append("ReleaseShared")
                tc = CMakeToolchain(self)
                if self.options["hello"].shared:
                    generator.configuration = "ReleaseShared"
                    tc.preprocessor_definitions["MYDEFINE"] = "MYDEF_SHARED_VALUE"
                    tc.preprocessor_definitions.debug["MYDEFINE_CONFIG"] = "MYDEF_SHARED_DEBUG"
                    tc.preprocessor_definitions.release["MYDEFINE_CONFIG"] = "MYDEF_SHARED_RELEASE"
                    tc.preprocessor_definitions.releaseshared["MYDEFINE_CONFIG"] = "MYDEF_SHARED_RELEASE"
                else:
                    tc.preprocessor_definitions["MYDEFINE"] = "MYDEF_VALUE"
                    tc.preprocessor_definitions.release["MYDEFINE_CONFIG"] = "MYDEF_RELEASE"
                tc.write_toolchain_files()
                generator.write_generator_files()
```

 * The call to `conan install . -s build_type=Release -o hello:shared=True` is overriding the `conan_toolchain.cmake` file. When we started the development around the toolchain, one of the main objectives was to avoid `if` clauses inside the `toolchain` method, it was the reason why we decided to maintain the _configs_, we deal with generator expressions and the `CMAKE_PROJECT_INCLUDE`.


 * This two calls will generate the same `helloTarget-releaseshared.cmake` file and override it:
   
    ```
    conan install . -s build_type=Debug -o hello:shared=True
    conan install . -s build_type=Release -o hello:shared=True
    ``` 

    I'm not sure if this is a desired side effect or something that will be really confusing/problematic for users.


TBH, I feel like this is not a _pure_ feature and it introduces (and documents) an extra level of complexity. When I think about the original motivation, I wonder why we cannot recommend adding to each Configuration in the project the corresponding `.props` file, I'm afraid we are building a castle to avoid a very specific pain (and, if needed, **we can easily fix it in the Visual Studio plugin, not from the Conan client**).

memsharded:
> The call to conan install . -s build_type=Release -o hello:shared=True is overriding the conan_toolchain.cmake file. When we started the development around the toolchain, one of the main objectives was to avoid if clauses inside the toolchain method, it was the reason why we decided to maintain the configs, we deal with generator expressions and the CMAKE_PROJECT_INCLUDE.

Yes, this is happening, but because this PR is handling only the variability of the generators, not the toolchain. It could be extended to the toolchain, but how it fits in the generators was the first thing I wanted to check after the MSBuild discussion. Please consider only the generators part of this PR.

I have already extracted #7920 new PR with the template refactor only

> TBH, I feel like this is not a pure feature and it introduces (and documents) an extra level of complexity. When I think about the original motivation, I wonder why we cannot recommend adding to each Configuration in the project the corresponding .props file, I'm afraid we are building a castle to avoid a very specific pain (and, if needed, we can easily fix it in the Visual Studio plugin, not from the Conan client).

Yes, this is exactly what we want to discover and find a good proposal. If we decide to rely exclusively on the VS plugin, then it means that the generator and toolchain do not need to be multi anymore, and the plugin will manage to create a new toolchain+generator files for each configuration switch. Though I am afraid that such behavior might invalidate the build caches of every configuration. Something to check. And it introduce a hard dependency on the plugin, I am not fully convinced that we want to go that way.

On the other hand, I would also like to have some uniformity across generators/toolchains. Relying on VS plugin to implement the multi-config switch, but having cmake being multi-config toolchains/generators is a bit assymetric.

>  I wonder why we cannot recommend adding to each Configuration in the project the corresponding .props

Yes, this might be a completely valid solution. Conan becomes totally agnostic, it is not necessary to change the filenames, but several configurations can be ``conan install -if=conf1``, ``conan install -if=conf2``, and then VS will add the generated files of each configuration from the matching folder, per their configuration.

But the CMake one would still be an open question, as this has to be managed by Conan, it is not possible to be done in the user side (all they will see is a ``find_package()`` and a ``-DCMAKE_TOOLCHAIN_FILE=toolchain.cmake`` and that must implement all the variability for all their managed configurations.

memsharded:
Making explicit the problem with custom ``build_type`` in settings.yml for the ``cmake_find_package_multi`` generator in a test in: https://github.com/conan-io/conan/pull/7923

lasote (commented):
Code not reviewed, only the general idea/approach.

My conclusions so far:

- The `configuration` shouldn't be necessarily equal to the `build_type` and might even not belong to the `settings.build_type` list. `MyCoolConfig` is not a valid `settings.build_type`.
- So, the `cmake_find_package_multi` generator `needs` to know the list of configurations to support. And that configuration cannot come from the settings nor the build_type.
- But the same for the toolchain! As @memsharded said, it is missing in this PR a way to generate N toolchain files with a "main" one including using the "Config" (not with ifs but with the config expressions)
- That would be exactly the same idea as the Visual one.

> TBH, I feel like this is not a pure feature and it introduces (and documents) an extra level of complexity. When I think about the original motivation, I wonder why we cannot recommend adding to each Configuration in the project the corresponding .props file, I'm afraid we are building a castle to avoid a very specific pain (and, if needed, we can easily fix it in the Visual Studio plugin, not from the Conan client).

I might have agreed (not really :sweat_smile: ) about the plugin managing this situation, but with the `find_package` with the same situation, I think this is evidence.

p-groarke:
An easier path (or first step) may be supporting [CMAKE_MAP_IMPORTED_CONFIG_\<CONFIG\>](https://cmake.org/cmake/help/latest/variable/CMAKE_MAP_IMPORTED_CONFIG_CONFIG.html).

If I understand this PR correctly, you are passing on custom configs to consumed packages. This is desirable for packages you control.

However, you cannot guarantee external dependencies supports your custom config, so one must be able to consume a different config for external dependencies. That's where the `CMAKE_MAP_IMPORTED_CONFIG_<CONFIG>` comes in.

Unit test example
```cmake
# Create custom config
set(CMAKE_CONFIGURATION_TYPES Debug Release MyCoolConfig)

# Use Release for our imported targets with find_package, if packages don't support MyCoolConfig.
set(CMAKE_MAP_IMPORTED_CONFIG_MYCOOLCONFIG Release)

# This should find GTestTarget-release.cmake
find_package(GTest CONFIG REQUIRED)
```

memsharded:
Superseded by other previous PRs

PR #7916, opened 2020-10-20
Add 'provides' and 'deprecated' fields to 'conan info' output

jgsogo:
Changelog: Feature: Add `provides` and `deprecated` fields to `conan info` output
Docs: omit

Changes:
 * Add `provides` and `deprecated` fields to `conan info` output
 * Minor format changes:
   + test names to match _standard_ names expected by `pytest`
   + some `textwrap.dedent`

### DIFF
--- a/conans/client/conan_command_output.py
+++ b/conans/client/conan_command_output.py
@@ -14,6 +14,7 @@ from conans.util.dates import iso8601_to_str
 from conans.util.env_reader import get_env
 from conans.util.files import save
 from conans import __version__ as client_version
+from conans.util.misc import make_tuple
 
 
 class CommandOutputer(object):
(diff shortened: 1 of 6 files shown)


conans/client/conan_command_output.py
```diff
@@ -166,15 +167,16 @@ def _add_if_exists(attrib, as_list=False):
                     if not as_list:
                         item_data[attrib] = value
                     else:
-                        item_data[attrib] = list(value) if isinstance(value, (list, tuple, set)) \
-                            else [value, ]
+                        item_data[attrib] = make_tuple(value)
```
jgsogo, line 170: This is changing current output: from `list` to `tuple`, breaking? 

TBH, I don't think so, this information is consumed in two different ways: JSON and written on the screen. For JSON it is the same a list and a tuple, in the screen we are writing a `, `-separated list. Am I missing other scenarios?

memsharded (approved):

danimtb (commented):
✔️

PR #7919, opened 2020-10-21
[toolchain] CMake + Ninja

### LINKED ISSUE #7814: [feature] CMakeToolchain + Ninja PoC
Implement a proof of concept of using the new CMakeToolchain feature for building with Ninja:

- From Linux
- From Windows

Notes:
- Ninja can be assumed installed. 
- Opt-in to use Ninja instead of the default (MSBuild, Makefiles) build system, or hardcoded in the recipe, investigate alternatives
- Integration test.
- If enough time to implement reuse of a package, the ``cmake_find_package_multi`` generator must be used

uilianries:
Changelog: Omit
Docs: omit
closes #7814

Testing CMakeToolchain with Ninja generator

CMake + Ninja only uses a specific Ninja var: [CMAKE_NINJA_OUTPUT_PATH_PREFIX](https://cmake.org/cmake/help/v3.7/variable/CMAKE_NINJA_OUTPUT_PATH_PREFIX.html), but it doesn't affect our jobs. It is used for extra ninja files dir, by default is present in CMakeFiles/ folder.

More info about CMake and Ninja: https://cmake.org/cmake/help/latest/generator/Ninja.html

### DIFF
--- a/conans/test/integration/toolchains/cmake/test_ninja.py
+++ b/conans/test/integration/toolchains/cmake/test_ninja.py
@@ -6,7 +6,6 @@ import os
 from conans.test.utils.tools import TestClient
 from conans.test.utils.test_files import temp_folder
 from conans.client.tools import environment_append
-from conans.client.toolchain.cmake.base import CMakeToolchainBase
 
 
 class CppProject(object):
@@ -61,14 +60,18 @@ class CMakeNinjaTestCase(unittest.TestCase):
 
             def toolchain(self):
                 tc = CMakeToolchain(self)
+                # tc.preprocessor_definitions["CMAKE_NINJA_OUTPUT_PATH_PREFIX"] = "MyValue"
                 tc.write_toolchain_files()
 
             def build(self):
                 cmake = CMake(self)
                 cmake.configure()
+                cmake.build()
 
             def package(self):
                 cmake = CMake(self)
+
+                cmake.configure()
                 cmake.install()
     """)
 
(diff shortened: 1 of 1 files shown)


conans/test/integration/toolchains/cmake/test_ninja.py
```diff
@@ -0,0 +1,127 @@
...
+        build_folder = os.path.join(self.client.current_folder, "build")
+        package_folder = os.path.join(self.client.current_folder, "pkg")
+        with environment_append({"CONAN_PRINT_RUN_COMMANDS": "1"}):
+            self.client.run("export . foobar/0.1.0@")
+            self.client.run("install . --install-folder={}".format(build_folder))
+            self.client.run("build . --build-folder={}".format(build_folder))
+            self.assertIn('CMake command: cmake -G "Ninja" '
+                          '-DCMAKE_TOOLCHAIN_FILE="conan_toolchain.cmake"', self.client.out)
+            # FIXME: conan package tries to install on /usr/local. CMAKE_PREFIX_PATH is empty
+            self.client.run("package . --build-folder={} --package-folder={}"
```
memsharded, line 199: Don't worry about local packaging at the moment, this can wait, can be removed.

memsharded (changes requested):
Please have a look at other existing tests in the toolchains. It is very important to make sure that the settings are translated to the toolchain correctly, for different platforms. For example if using it for VS compiler, we want to check that the build type is the correct one, the architecture, the C++standard, the MT/MD runtime, etc. Same for gcc and libcxx too.

The toolchain feature is about making sure that things built in Conan cache, match the local build the developer does in their user folder and is aligned with the defined Conan settings.

uilianries:
I have to investigate better for Windows, but CMake + Ninja is limited there in terms of finding the compiler. I have to force the compiler otherwise it can't find visual studio. Maybe toolchain file could enforce it.

PR #7920, opened 2020-10-21
Refactor/generator cmake find multi jinja

memsharded:
Changelog: Omit
Docs: Omit

Use jinja to expand the CMake generator expressions in ``cmake_find_package_multi`` for configs ["Release", "Debug", "RelWithDebInfo", "MinSizeRel"]

Extracted from https://github.com/conan-io/conan/pull/7915

Make @jgsogo happy 😄  

#tags: slow


jgsogo (approved):
I'd have used a _templatetag_ instead of a _macro_ (that was my paln if you didn't include it), but this makes me happy too 😄

PR #7922, opened 2020-10-22
POC python profiles

### LINKED ISSUE #7659: [question] Extending environment variable values in profiles
I read there are two options to set env variables in profiles:
```
[env]
VAR=assign
PATH=[extend]
```

How can I extend non-path variables? This is a common scenario when including another profile and adding to flags.
```
[env]
CFLAGS=$CFLAGS -fsanitize=address   # $CFLAGS is not expanded - not a global variable
CFLAGS=[-fsanitize=address]         # Joined with ':', but this is not a path
```

danimtb:
Changelog: Feature: Allow usage of Python in profiles.
Docs: missing

- [x] Refer to the issue that supports this Pull Request: closes #7659

### DIFF
--- a/conans/model/env_info.py
+++ b/conans/model/env_info.py
@@ -42,6 +42,16 @@ class EnvValues(object):
         ret = EnvValues()
         if not text:
             return ret
+        if isinstance(text, dict):
+            for k, v in text.items():
+                package = None
+                if ":" in k:
+                    package, name = k.split(":", 1)
+                else:
+                    name = k
+                ret.add(name, v, package)
+            return ret
+
         for env_def in text.splitlines():
             try:
                 if env_def:
@@ -64,7 +74,6 @@ class EnvValues(object):
                 raise
             except Exception as exc:
                 raise ConanException("Error parsing the env values: %s" % str(exc))
-
         return ret
 
     def dumps(self):
--- a/conans/model/env_info.py
+++ b/conans/model/env_info.py
@@ -42,7 +42,6 @@ class EnvValues(object):
         ret = EnvValues()
         if not text:
             return ret
-
         for env_def in text.splitlines():
             try:
                 if env_def:
@@ -65,6 +64,7 @@ class EnvValues(object):
                 raise
             except Exception as exc:
                 raise ConanException("Error parsing the env values: %s" % str(exc))
+
         return ret
 
     def dumps(self):
--- a/conans/model/env_info.py
+++ b/conans/model/env_info.py
@@ -42,15 +42,6 @@ class EnvValues(object):
         ret = EnvValues()
         if not text:
             return ret
-        if isinstance(text, dict):
-            for k, v in text.items():
-                package = None
-                if ":" in k:
-                    package, name = k.split(":", 1)
-                else:
-                    name = k
-                ret.add(name, v, package)
-            return ret
 
         for env_def in text.splitlines():
             try:
--- a/conans/test/functional/profiles/test_python_profiles.py
+++ b/conans/test/functional/profiles/test_python_profiles.py
@@ -1,19 +1,11 @@
 import os
 import textwrap
-import traceback
 import unittest
-import uuid
-from _ast import Module
-from collections import OrderedDict
 
-import sys
 from importlib.util import spec_from_file_location, module_from_spec
-from os.path import splitext, basename
 
 from conans.client.tools import environment_append
-from conans.errors import ConanException
 from conans.test.utils.tools import TestClient
-from conans.tools import chdir
 
 
 class Profile(dict):
@@ -36,7 +28,7 @@ class Profile(dict):
 
 
 def _load_module_from_file(filename, parent_profile):
-    name = splitext(basename(filename))[0]
+    name = os.path.splitext(os.path.basename(filename))[0]
     spec = spec_from_file_location(name, filename)
     config = module_from_spec(spec)
     for key, value in parent_profile.items():
--- a/conans/test/functional/profiles/test_python_profiles.py
+++ b/conans/test/functional/profiles/test_python_profiles.py
@@ -1,58 +1,107 @@
+import os
 import textwrap
+import traceback
 import unittest
+import uuid
+from _ast import Module
+from collections import OrderedDict
 
+import sys
+from importlib.util import spec_from_file_location, module_from_spec
+from os.path import splitext, basename
+
+from conans.client.tools import environment_append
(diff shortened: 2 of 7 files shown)


conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+            settings = {
+                 "os": "Windows",
+                 "compiler": "Visual Studio",
+                 "compiler.version": "16",
+                }
+            options = {"shared": "2"}
+            env = {"pkg:env_var": "bar"}
+            cxxflags = os.environ.get('HOME', "")
```
memsharded: Use a custom defined env-var, not something that can be defined or not in the system, and then check it later in the test.

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,58 @@
...
+            if (cxxflags):
```
memsharded: Parenthesis not needed

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,58 @@
...
+                env['CXXFLAGS'] = cxxflags + "-stdlib=libc++11"
+        """)
+        profile_content2 = textwrap.dedent("""
+            from conans import tools
```
memsharded, line 71: Let's exclude importing conan things at the moment, wdyt?

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,58 @@
...
+            import os
+
+
+            settings = {
+                 "os": "Windows",
+                 "compiler": "Visual Studio",
+                 "compiler.version": "16",
+                }
+            options = {"shared": "2"}
+            env = {"pkg:env_var": "bar"}
```
memsharded, line 65: discuss the syntax, in case we want to do {"pkg": {"env_var": "bar"}} (probably not, but just in case)

conans/model/env_info.py
```diff
@@ -42,6 +42,16 @@ def loads(text):
         ret = EnvValues()
         if not text:
             return ret
+        if isinstance(text, dict):
```
memsharded: Better create a ``from_dict()`` static method or something.

memsharded (commented):
Looking good!

danimtb:
Forgot to comment, that one of the issues I found with this approach was a possible collision with the python module name when loading the profile. FOr example, when I wrote the test, I named the test profile file as `profile.py`, but when it was loaded python complained about a module that was named in the same way. Actually, there is a `profile.py` module distributed with python.

I think it is not a major problem as we can change the module name of each profile randomly, just wanted to comment the issue here as well.

memsharded:
> Forgot to comment, that one of the issues I found with this approach was a possible collision with the python module name when loading the profile. FOr example, when I wrote the test, I named the test profile file as profile.py, but when it was loaded python complained about a module that was named in the same way. Actually, there is a profile.py module distributed with python.

I think the name of the file should be irrelevant for loading it, as long as it is not being imported by other profile. Our python loader should be able to put them in their own namespace to not conflict with system ones.

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,99 @@
+import os
+import textwrap
+import unittest
+
+from importlib.util import spec_from_file_location, module_from_spec
```
danimtb, line 5: unfortunately, this is only python 3

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,99 @@
...
+            else:
+                self[k] = v
+
+
+def _load_module_from_file(filename, parent_profile):
+    name = os.path.splitext(os.path.basename(filename))[0]
+    spec = spec_from_file_location(name, filename)
+    config = module_from_spec(spec)
+    for key, value in parent_profile.items():
+        setattr(config, key, value)
```
danimtb, line 35: here attributes coming from parent profile are set so they are available in the module before loading it

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def _load_profile_python(profile_path, parent_profile=None):
+    # TODO: implement include() of other profiles? variable substitution?
+    parent_profile = parent_profile or Profile()
+    profile_content = _load_module_from_file(profile_path, parent_profile)
+    profile = Profile()
+    for key, value in profile_content.__dict__.items():
+        if not key.startswith("__"):
+            profile[key] = value
```
danimtb, line 47: here we set all the variables to the profile in order to achieve generic variable substitution

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,99 @@
...
+    parent_profile.update(profile)
+    return parent_profile
```
danimtb, line 49: update parent profile with the new values and return so inheritance is achieved

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,99 @@
...
+            cxxflags = os.environ.get('CXXFLAGS', "")
+            if cxxflags:
+                env['CXXFLAGS'] = cxxflags + " -stdlib=libc++11"
+        """)
+        pr2_content = textwrap.dedent("""
+            from conans import tools
+
+            settings = {"os": "Linux", "build_type": "Release"}
+            env_values = {"cpu_count": str(tools.cpu_count())}
+            if env:
```
memsharded, line 75: This reads a bit weird and non idiomatic. Doesn't this profile raise an error if not used together with a parent profile, because ``env`` is an undefined variable?

memsharded (changes requested):
I like this proposal, but I think we need to check for a better composition syntax, that is more idiomatic. Maybe something based on callbacks for custom behavior.

Lets continue to work on this a bit more in next 1.33

conans/test/functional/profiles/test_python_profiles.py
danimtb: yes, I think so

memsharded: Would you like to rethink the composition proposal yourself, or do you prefer to brainstorm or discuss first?

memsharded:

PR #7924, opened 2020-10-22
Set CMAKE_SYSTEM_NAME to iOS, tvOS, watchOS from CMake 3.14

### LINKED ISSUE #7875: [bug] CMAKE_SYSTEM_NAME is set to Darwin on iOS (watch and tv OS)
Coming from here, https://github.com/conan-io/conan/issues/4550 I can see that the change was not enough.

The `CMAKE_SYSTEM_NAME` is always set to Darwin, even for iOS (cross) builds, what is wrong

please check 
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-for-ios-tvos-or-watchos
that clearly specifies what should be used

and for example the help to the cmake IOS variable
https://cmake.org/cmake/help/latest/variable/IOS.html
> Set to 1 when the target system (CMAKE_SYSTEM_NAME) is iOS

clearly, setting this to Darwin for iOS is not good.

  * Operating System+version: OSX 10.15.7
  * Compiler+version: apple clang 12
  * Conan version: 1.30.1
  * Python version: 3.8.5

czoido:
Changelog: Fix: Set `CMAKE_SYSTEM_NAME` set to `iOS`, `tvOS` or `watchOS` or `Darwin` depending on the CMake version. 
Docs: Omit

Closes: https://github.com/conan-io/conan/issues/7875

Since CMake 3.14 setting `CMAKE_SYSTEM_NAME` to `iOS`, `tvOS` or `watchOS` is supported. This PR checks the version of CMake in the build helper and sets the definition accordingly.

TODO: Update this implementation in the toolchain files. I'll open an issue about this.


memsharded (approved):

a4z (approved):

conans/client/build/cmake_flags.py
```diff
@@ -172,7 +172,7 @@ def _get_cpp_standard_vars(self):
         definitions["CONAN_STD_CXX_FLAG"] = cppstd_flag(self._conanfile.settings)
         return definitions
 
-    def _cmake_cross_build_defines(self):
+    def _cmake_cross_build_defines(self, cmake_version=None):
```
jgsogo:
```suggestion
    def _cmake_cross_build_defines(self, cmake_version):
```

No need to add a default value. The call to `CMake::get_version()` returns the value or raises. It also simplifies the implementation.

jgsogo (changes requested):

conans/client/build/cmake_flags.py
```diff
@@ -204,10 +204,12 @@ def _cmake_cross_build_defines(self):
         else:  # detect if we are cross building and the system name and version
             skip_x64_x86 = os_ in ['Windows', 'Linux']
             if cross_building(self._conanfile, skip_x64_x86=skip_x64_x86):  # We are cross building
+                apple_system_name = "Darwin" if cmake_version and Version(cmake_version) < Version(
+                    "3.14") or not cmake_version else None
```
jgsogo:
```suggestion
                apple_system_name = "Darwin" if Version(cmake_version) < Version("3.14") else None
```

conans/client/build/cmake_flags.py
```diff
@@ -280,7 +282,7 @@ def _get_make_program_definition(self):
 
         return {}
 
-    def get_definitions(self):
+    def get_definitions(self, cmake_version=None):
```
jgsogo: Why a default value here? This function is called only from this line

```
self.definitions = builder.get_definitions(cmake_version=self.get_version())
```

and it will always pass a value or raise.

jgsogo (changes requested):

conans/client/build/cmake_flags.py
czoido: It was because a test was calling directly and also thinking about a tool or something that could be calling this... I'll update the test and remove that.

jgsogo (approved):

PR #7929, opened 2020-10-22
encapsulating package dirty

memsharded:
Changelog: omit
Docs: omit

Removing the package "dirty" flag automatically with the package. Adding a check that set_dirty is not applied to already dirty packages.

### DIFF
--- a/conans/client/remover.py
+++ b/conans/client/remover.py
@@ -75,9 +75,7 @@ class DiskRemover(object):
                 pref = PackageReference(package_layout.ref, package_id)
                 if not package_layout.package_exists(pref):
                     raise PackageNotFoundException(pref)
-                pkg_folder = package_layout.package(pref)
                 package_layout.package_remove(pref)
-                self._remove_file(pkg_folder + ".dirty", package_layout.ref, "dirty flag")
                 self._remove_file(package_layout.system_reqs_package(pref), package_layout.ref,
                                   "%s/%s" % (package_id, SYSTEM_REQS))
 
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -48,6 +48,7 @@ _DIRTY_FOLDER = ".dirty"
 
 def set_dirty(folder):
     dirty_file = os.path.normpath(folder) + _DIRTY_FOLDER
+    assert not os.path.exists(dirty_file)
     save(dirty_file, "")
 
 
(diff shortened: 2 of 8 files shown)


conans/client/remover.py
```diff
@@ -75,9 +75,7 @@ def remove_packages(self, package_layout, ids_filter=None):
                 pref = PackageReference(package_layout.ref, package_id)
                 if not package_layout.package_exists(pref):
                     raise PackageNotFoundException(pref)
-                pkg_folder = package_layout.package(pref)
                 package_layout.package_remove(pref)
-                self._remove_file(pkg_folder + ".dirty", package_layout.ref, "dirty flag")
                 self._remove_file(package_layout.system_reqs_package(pref), package_layout.ref,
                                   "%s/%s" % (package_id, SYSTEM_REQS))
```
jgsogo, line 80: Does it make sense to remove this file too when the package_id is removed?

conans/util/files.py
```diff
@@ -48,6 +48,7 @@ def make_file_read_only(file_path):
 
 def set_dirty(folder):
     dirty_file = os.path.normpath(folder) + _DIRTY_FOLDER
+    assert not os.path.exists(dirty_file)
```
jgsogo: Every assert with the corresponding message

```suggestion
    assert not os.path.exists(dirty_file), "Folder '{}' is already dirty".format(folder)
```

conans/client/remover.py
jgsogo: I mean, remove this file inside the `layout.package_remove` method

memsharded: Yeah, I didn't want to touch system-requirements in this PR, I prefer doing 1 thing at a time, that will follow in another PR

jgsogo (approved):

jgsogo:
Next steps:
 * try to handle removal of system-reqs file inside the `layout.package_remove` function
 * can we remove `set_dirty_context_manager` function (the one in `conans.util.files`)? Is it useful by itself?

memsharded:
> try to handle removal of system-reqs file inside the layout.package_remove function

Agree

> can we remove set_dirty_context_manager function (the one in conans.util.files)? Is it useful by itself?

Still used for other folders (source, export, etc). That might be removed when we abstract away the management of those folders in the same way we did with packages.

PR #7932, opened 2020-10-23
Fix cmake build helper get_definitions when failed CMake get version

czoido:
Changelog: omit
Docs: omit

Develop CI broken at: https://github.com/conan-io/conan/pull/7924

Just to test if it fixes develop after: https://conan-ci.jfrog.info/blue/organizations/jenkins/ConanTestSuite/detail/develop/695/pipeline/414

#TAGS: slow

### DIFF
--- a/conans/client/build/cmake_flags.py
+++ b/conans/client/build/cmake_flags.py
@@ -204,7 +204,8 @@ class CMakeDefinitionsBuilder(object):
         else:  # detect if we are cross building and the system name and version
             skip_x64_x86 = os_ in ['Windows', 'Linux']
             if cross_building(self._conanfile, skip_x64_x86=skip_x64_x86):  # We are cross building
-                apple_system_name = "Darwin" if Version(cmake_version) < Version("3.14") else None
+                apple_system_name = "Darwin" if cmake_version and Version(cmake_version) < Version(
+                    "3.14") or not cmake_version else None
                 cmake_system_name_map = {"Macos": "Darwin",
                                          "iOS": apple_system_name or "iOS",
                                          "tvOS": apple_system_name or "tvOS",
--- a/conans/client/build/cmake.py
+++ b/conans/client/build/cmake.py
@@ -98,7 +98,12 @@ class CMakeBuildHelper(object):
         # FIXME CONAN 2.0: CMake() interface should be always the constructor and self.definitions.
         # FIXME CONAN 2.0: Avoid properties and attributes to make the user interface more clear
 
-        self.definitions = builder.get_definitions(self.get_version())
+        try:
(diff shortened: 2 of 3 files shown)


memsharded (commented):
I am still a bit concerned about this. The change is basically a workaround for a failing test that is not very elegant, probably the test is not well defined either and needs a change too.

But it still reads very weird to me. I would try, very very strongly to isolate the "cmake version" checks in CMake code, but not in python.
This variability holds here, because this runs in the ``build()`` method, and cmake is supposed to be there, but any attempt to do something similar in the toolchain() will fail. It is good that we have decoupled and do not depend anymore on the ``builder.get_definitions()`` in the toolchain, but in the past we did.

jgsogo (changes requested):
Maybe I need more context.

The test fails because CMake is not in the `PATH` right? the test fails to run `cmake --version` and **it'd fail running any other `cmake ...` command**. This failure is related to the build-helper, it doesn't make sense a recipe that creates the build-helper and doesn't run it (although it is possible). 

If this PR is covering that use-case...  I'd need evidence that it is useful for someone and they cannot use existing `tools` to get whatever information they are consuming from the CMakeBuildHelper object without running it.

Otherwise, IMHO we shouldn't add anything to the codebase just because we need it to satisfy a test we have written ourselves.

conans/client/build/cmake_flags.py
```diff
@@ -204,7 +204,8 @@ def _cmake_cross_build_defines(self, cmake_version):
         else:  # detect if we are cross building and the system name and version
             skip_x64_x86 = os_ in ['Windows', 'Linux']
             if cross_building(self._conanfile, skip_x64_x86=skip_x64_x86):  # We are cross building
-                apple_system_name = "Darwin" if Version(cmake_version) < Version("3.14") else None
+                apple_system_name = "Darwin" if cmake_version and Version(cmake_version) < Version(
+                    "3.14") or not cmake_version else None
```
jgsogo, line 208:
```suggestion
                    "3.14") else None
```

jgsogo (approved):
Ok, let's fallback to the previous behavior

conans/client/build/cmake_flags.py
czoido: I think that If I change that it's going to resolve as if the cmake version is > 3.14 and I think we want to have the current behaviour when the cmake version call fails and we don't have it. I have added one more check to test this:

```python
definitions = definitions_builder.get_definitions(None)
self.assertEqual("Darwin", definitions["CMAKE_SYSTEM_NAME"])
```

jgsogo (approved):

memsharded (approved):

PR #7936, opened 2020-10-25
remove ancient .c_src 0.25 compat

memsharded:
Changelog: omit
Docs: omit

This code is more than 3 years old, from 0.25. I think it can be wiped from the codebase.


jgsogo (approved):
I'm totally ok with removing 0.x deprecated bits.

PR #7942, opened 2020-10-26
Remove arbitrary or unnecessary content from make toolchain

solvingj:
From conversation on slack with @memsharded :

> at the moment, most of the toolchain is unused, right?

> maybe we want to start super minimalistic instead?

> something like "things not necessary to have the tests passing".

This implementation removes everything not necessary to have tests passing on linux and windows (mingw32). I think it is minimalistic, but @memsharded please advise if anything else should be changed. 

Changelog: Fix: Simplified MakeToolchain to remove things that were not checked by tests or unused.
Docs: omit

#TAGS: slow


solvingj:
@memsharded please review (I don't have permission to add a reviewer.)

conans/client/toolchain/make.py
```diff
@@ -22,172 +16,26 @@ class MakeToolchain(object):
...
-        {%- endif %}
-
             # Recipe-Defined Variables
         {%- for it, value in variables.items() %}
             {{ it }} = {{ value }}
         {%- endfor %}
 
-            # Automatic Conan pre-processor definition: build_type_define
-        {%- if build_type_define %}
-            CONAN_TC_CPPFLAGS += -D{{build_type_define}}
```
memsharded: But this was being used, it was already part of the build

conans/client/toolchain/make.py
```diff
@@ -22,172 +16,26 @@ class MakeToolchain(object):
...
-        {%- endif %}
-
-            # Automatic Conan pre-processor definition: glibcxx_define
-        {%- if glibcxx_define %}
-            CONAN_TC_CPPFLAGS += -D{{glibcxx_define}}
```
memsharded: Same: this was part of the build

conans/client/toolchain/make.py
```diff
@@ -22,172 +16,26 @@ class MakeToolchain(object):
         ifndef CONAN_TOOLCHAIN_INCLUDED
             CONAN_TOOLCHAIN_INCLUDED = TRUE
 
-            # Automatic Conan Toolchain Variables
-        {%- if build_type %}
-            CONAN_TC_BUILD_TYPE = {{build_type}}
```
memsharded, line 28: This was not part of the build

memsharded (changes requested):
I am afraid there was some misunderstanding, it seems you went too far. 
I meant, all the vars that are not being used at all in the build, like CONAN_TC_ARCH_HOST.

Other vars, they were actually involved in the build, and they were being injected in CONAN_TC_SETUP, right? Then we have at least some evidence that they are not completely broken.

conans/client/toolchain/make.py
solvingj: restored

conans/client/toolchain/make.py
solvingj: no-op

conans/client/toolchain/make.py
solvingj: restored

solvingj:
I've restored the specific preprocessor definitions requested.  Please name or highlight any other specific flags or definitions which you would like restored.  I think it's best for you to choose the list explicitly in this case.

solvingj:
@memsharded Here are four items related to settings which were removed.  Please indicate any of these which should be restored.
```
libcxx_flag 
cppstd_flag
build_type_flags
arch_flag
```

memsharded:
> libcxx_flag 
cppstd_flag
build_type_flags
arch_flag

These should be restored, making sure they are actually being used in the compilation, checking the binary or introducing architecture or cppstd specific code, it depends.
Maybe it can be delayed until users feedback, to learn actual user needs and patterns. For example, is it expected that Conan defines "Debug" or "Release" CXXFLAGS? Maybe not, that would be part of users Makefiles, right? Maybe it is an opt-in for users that want us to give the utility?

memsharded:
I have merged it, lets build step by step on the minimalistic toolchain

PR #7945, opened 2020-10-26
POC env vars substitution in profiles

danimtb:
Changelog: Feature: Allow environment variable substitution in profiles
Docs: missing

After playing a little bit with ConfigParser, I found that the built-in variable declaration of profiles was a mess in the code in order to get the values right.

Also, you have to implement your own interpolator to get the values as well https://stackoverflow.com/questions/26586801/configparser-and-string-interpolation-with-env-variable/49529659#49529659

So finally, I found it easier to apply the `os.path.expandvars()` to the whole contents after processing the built-in variable declaration.

Another con is that the ConfigParser changed from python2 to python3 and the imports are a bit different (might not be a real issue).

However, if we take the decision to change our own profile parser and use ConfigParser utilities instead, I think the code might be less messy. Even though the `include()` functionality and the "inheritance" of vars declaration from one profile to another is probably not supported by ConfigParser.

- [x] Refer to the issue that supports this Pull Request: #7659


conans/test/functional/profiles/profiles_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+from conans.client.tools import environment_append
+from conans.test.utils.tools import TestClient
+
+
+class ProfileTest(unittest.TestCase):
+
+    def test_env_var_with_variable_substitution(self):
+        client = TestClient()
+        profile = """
```
memsharded: Please ``textwrap.dedent()`` and indent the contents with respect to the ``profile`` variable definition.

conans/client/profile_loader.py
```diff
@@ -147,6 +147,9 @@ def _load_profile(text, profile_path, default_folder):
         # Replace the variables from parents in the current profile
         profile_parser.apply_vars()
 
+        # Expand env vars after variable substitution
+        profile_parser.profile_text = os.path.expandvars(profile_parser.profile_text)
```
memsharded: This sounds a bit risky, to be honest. It is a ``os.path`` operation, so at some point it might do some assumptions about characters, whitespace including carriage return, etc.

conans/test/functional/profiles/profiles_test.py
```diff
@@ -39,3 +39,37 @@ def test_env_var_without_variable_substitution(self):
...
+        class CustomInterpolation(BasicInterpolation):
+            """Interpolation which expands environment variables in values."""
+
+            def before_get(self, parser, section, option, value, defaults):
+                rawval = parser.get(section, option, raw=True)
+                val = rawval % parser.defaults()
+                return val
+
+        parser1 = ConfigParser(interpolation=CustomInterpolation())
+        profile1 = """
```
memsharded: please ``textwrap.dedent`` the files contents (one indent more than the variable declaration)

conans/test/functional/profiles/profiles_test.py
```diff
@@ -39,3 +39,37 @@ def test_env_var_without_variable_substitution(self):
...
+        parser1.read_string(profile1)
+        p1_vars = dict(parser1.items("env"))
+        self.assertDictEqual({"bar": "bar1", "cxxflags": "-fpic", "zet": "zet"}, p1_vars)
+
+        parser2 = ConfigParser(interpolation=CustomInterpolation(), defaults=p1_vars)
+        profile2 = """
+        [env]
+        foo=foo1
+        bar=bar2
+        CXXFLAGS=%(cxxflags)s -fother
```
memsharded: Could we better use the flags with the same casing, ``%(CXXFLAGS)`` in this case?

memsharded (commented):
This is interesting. If you could also please add the "basic" case of settings and options updates (the higher precedence profile overwrites the existing values of the previous one, respecting other existing values not defined)

conans/test/functional/profiles/profiles_test.py
danimtb: done

conans/test/functional/profiles/profiles_test.py
```diff
@@ -0,0 +1,105 @@
...
+        if section:
+            if section in self.section_defaults:
+                result = self.section_defaults[section]
+        result.update(items)
+        return result
+
+
+class ProfileTest(unittest.TestCase):
+
+    def test_env(self):
```
memsharded, line 48: I am trying to have a test passing like:

```python
    def test_os_environ(self):
        with environment_append({"CXXFLAGS": "-fPIC"}):
            parser2 = ProfileConfigParser()
            profile2 = textwrap.dedent("""
                [env]
                foo=foo1
                bar=bar2
                CXXFLAGS=%(CXXFLAGS)s -fother
                OTHERFLAGS=%(missing_flag)s -fother
            """)
            parser2.read_string(profile2)
            self.assertDictEqual({"foo": "foo1", "CXXFLAGS": "-fPIC -fother", "bar": "bar2",
                                  "zet": "zet", "OTHERFLAGS": " -fother"},
                                 dict(parser2.items("env")))
```
But not evident. Env-vars should be readable from the environment too, not only from other profiles. Is this possible?

danimtb: Not in this POC, as the main issue was to create a substitution syntax from one profile to another. However, you could do the env var substitution by means of `os.path.expandvars` or similar the same way it is done the interpolation here with values from other profiles

memsharded:

PR #7946, opened 2020-10-26
Move 'CMAKE_INSTALL_PREFIX' to the command-line

jgsogo:
Changelog: omit
Docs: omit

If the value for `CMAKE_INSTALL_PREFIX` is not known in the local-workflow, it makes no sense to add that value to the toolchain template. And the cache-workflow can use the command line to set the value.

### DIFF
--- a/conans/client/build/cmake_toolchain_build_helper.py
+++ b/conans/client/build/cmake_toolchain_build_helper.py
@@ -77,7 +77,7 @@ class CMakeToolchainBuildHelper(object):
         mkdir(build_folder)
         arg_list = '-DCMAKE_TOOLCHAIN_FILE="{}" -DCMAKE_INSTALL_PREFIX="{}" "{}"'.format(
             CMakeToolchainBase.filename,
-            self._conanfile.package_folder.package_folder.replace("\\", "/"),
+            self._conanfile.package_folder.replace("\\", "/"),
             source)
 
         generator = '-G "{}" '.format(self._generator) if self._generator else ""
--- a/conans/client/build/cmake_toolchain_build_helper.py
+++ b/conans/client/build/cmake_toolchain_build_helper.py
@@ -76,7 +76,9 @@ class CMakeToolchainBuildHelper(object):
 
         mkdir(build_folder)
         arg_list = '-DCMAKE_TOOLCHAIN_FILE="{}" -DCMAKE_INSTALL_PREFIX="{}" "{}"'.format(
-            CMakeToolchainBase.filename, self._conanfile.package_folder, source)
+            CMakeToolchainBase.filename,
+            self._conanfile.package_folder.package_folder.replace("\\", "/"),
+            source)
 
         generator = '-G "{}" '.format(self._generator) if self._generator else ""
         command = "%s %s%s" % (self._cmake_program, generator, arg_list)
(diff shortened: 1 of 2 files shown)


conans/client/build/cmake_toolchain_build_helper.py
```diff
@@ -75,7 +75,8 @@ def configure(self, source_folder=None):
             build_folder = os.path.join(self._conanfile.build_folder, self._build_folder)
 
         mkdir(build_folder)
-        arg_list = '-DCMAKE_TOOLCHAIN_FILE="%s" "%s"' % (CMakeToolchainBase.filename, source)
+        arg_list = '-DCMAKE_TOOLCHAIN_FILE="{}" -DCMAKE_INSTALL_PREFIX="{}" "{}"'.format(
+            CMakeToolchainBase.filename, self._conanfile.package_folder, source)
```
memsharded: Probably add the ``.package_folder.replace("\\", "/")`` that was there? It might be necessary in Windows?

memsharded (approved):

conans/client/build/cmake_toolchain_build_helper.py
jgsogo: In the old build-helper we are not running the replacement (https://github.com/conan-io/conan/blob/e6385bbf639898400a10841cee37910f4fe10a38/conans/client/build/cmake_flags.py#L352)... but now I see we are setting many other variables. Should I set all of them? maybe use `GNUInstallDirs`?

memsharded: It seems ``GNUInstallDirs`` define the system ones, not the local ones, which sounds unwanted.
Passing all of them in the command line reads very ugly again. What is currently defined:

```python
definitions["CMAKE_INSTALL_BINDIR"] = DEFAULT_BIN
definitions["CMAKE_INSTALL_SBINDIR"] = DEFAULT_BIN
...
definitions["CMAKE_INSTALL_OLDINCLUDEDIR"] = DEFAULT_INCLUDE
definitions["CMAKE_INSTALL_DATAROOTDIR"] = DEFAULT_SHARE
```

Is also a bit wrong, as it is hardcoding some values, irrespective of what is defined in the ``package_info()`` and ``package()`` method. I intentionally left this out until we had some time to think about it. Shouldn't those values be defined by the library ``CMakeLists.txt``?

conans/client/build/cmake_toolchain_build_helper.py
```diff
@@ -75,7 +75,10 @@ def configure(self, source_folder=None):
             build_folder = os.path.join(self._conanfile.build_folder, self._build_folder)
 
         mkdir(build_folder)
-        arg_list = '-DCMAKE_TOOLCHAIN_FILE="%s" "%s"' % (CMakeToolchainBase.filename, source)
+        arg_list = '-DCMAKE_TOOLCHAIN_FILE="{}" -DCMAKE_INSTALL_PREFIX="{}" "{}"'.format(
+            CMakeToolchainBase.filename,
+            self._conanfile.package_folder.package_folder.replace("\\", "/"),
```
memsharded:
```suggestion
            self._conanfile.package_folder.replace("\\", "/"),
```

PR #7947, opened 2020-10-27
feat: purec test package

### LINKED ISSUE #7152: [question] Getting 'settings.compiler.libcxx' value not defined in a --pure-c project

Even though I've created a new package with the flag `--pure-c`, which conanfile.py template configures the deletion of the field `self.settings.compiler.libcxx`, the command conan create still fails with the error `ERROR: : 'settings.compiler.libcxx' value not defined`.

Am I doing something wrong? Is that a bug?

Here are the commands:

```
conan new myproj/1.2.3 --test --pure-c
```

```
# generated conanfile.py
    def configure(self):
            del self.settings.compiler.libcxx
```

```
conan create . user/testing -s compiler=clang -s compiler.version=10
Exporting package recipe
myproj/1.2.3@user/testing: A new conanfile.py version was exported
myproj/1.2.3@user/testing: Folder: /home/myuser/.conan/data/myproj/1.2.3/user/testing/export
myproj/1.2.3@user/testing: Exported revision: 71b84694e77a8c54b51a1f797a31116a
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=clang
compiler.version=10
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

ERROR: : 'settings.compiler.libcxx' value not defined
```

sdmg15:
Changelog: Feature: Provide correct `--pure_c` implementation to `conan new`.
Docs: omit 

Close #7152

### DIFF
--- a/conans/client/cmd/new.py
+++ b/conans/client/cmd/new.py
@@ -191,7 +191,31 @@ target_link_libraries(example ${CONAN_LIBS})
 #          COMMAND example)
 """
 
+test_cmake_pure_c = """cmake_minimum_required(VERSION 2.8.12)
+project(PackageTest C)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup()
+
+add_executable(example example.c)
+target_link_libraries(example ${CONAN_LIBS})
+
+# CTest is a testing tool that can be used to test your project.
+# enable_testing()
+# add_test(NAME example
+#          WORKING_DIRECTORY ${CMAKE_BINARY_DIR}/bin
+#          COMMAND example)
+"""
+
 test_main = """#include <iostream>
+#include "hello.h"
+
+int main() {
+    hello();
+}
+"""
+
+test_main_pure_c = """#include <stdio.h>
 #include "hello.h"
 
 int main() {
@@ -349,8 +373,12 @@ def cmd_new(ref, header=False, pure_c=False, test=False, exports_sources=False, 
         files["test_package/conanfile.py"] = test_conanfile.format(name=name, version=version,
                                                                    user=user, channel=channel,
                                                                    package_name=package_name)
-        files["test_package/CMakeLists.txt"] = test_cmake
-        files["test_package/example.cpp"] = test_main
+        if pure_c:
+            files["test_package/example.c"] = test_main_pure_c
+            files["test_package/CMakeLists.txt"] = test_cmake_pure_c
+        else:
+            files["test_package/example.cpp"] = test_main
+            files["test_package/CMakeLists.txt"] = test_cmake
 
     if gitignore:
         files[".gitignore"] = gitignore_template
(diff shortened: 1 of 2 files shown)


sdmg15:
@memsharded I'm unable to link with the `hello.h` how can I manage to?

conans/client/cmd/new.py
```diff
@@ -199,6 +215,14 @@ def test(self):
 }
 """
 
+test_main_pure_c = """#include <stdio.h>
```
memsharded: The include<stdio.h> is not necessary here.

memsharded (commented):
> @memsharded I'm unable to link with the hello.h how can I manage to?

Oh, you are right, there is a bug there. Passing the ``--pure-c`` flag is incompatible with the ``--source``, but the actual code that is in github, the "hello" project is only C++. I would recommend the following, to have a purely operation ``--pure-c`` option:

- We can declare it a bug, and then it can be fixed
- The ``--pure-c`` will automatically activate the ``--source``, instead of being incompatible.
- Let's generate a ``hello.c`` and ``hello.h`` pure C library "hello" package, with local sources.
- It would be great to update the test ``new_purec_test`` to reflect and capture the new behavior. I can help or contribute that if you need help.

Please let me know if this is clear, if you have any question, and if you think you will manage to do it. Thanks!

sdmg15:
So you mean while calling `--pure-c` we need to call the `--source` too ?

The generation of the `hello.c` and `hello.h` will be done from this source or we will update the github package?

memsharded:
> So you mean while calling --pure-c we need to call the --source too ?

Yes, exactly, make it obligatory, --pure-c only works with the local --source option, otherwise, it is an error.

> The generation of the hello.c and hello.h will be done from this source or we will update the github package?

Yes, include the generation of hello.c, hello.h, CMakeLists.txt of the package as well, not only the ``test_package``

sdmg15:
If we include the generation of the files hello.h and hello.c, would that still have the same effect as adding it as a dependency?

Would it be something like this?
```
hello_c = ""
#include <stdio.h>


```

...

memsharded:
It would be the same as it is already being done with the *hello_cpp* file, it is not a dependency, it is an example source file.

Yes, what you suggest seems correct, the *hello_h* can be common, and then it means adding the *hello_c* file you are suggesting in the ``conans/client/cmd/new.py`` file (and probably changing the CMakeLists.txt so it has ``project(MyHello C)`` instead of ``project(MyHello CXX)``

sdmg15:
Yeah will check that, the current changes are doing what you said: Update the cmakelists to 
`projecr(MyProject C)` will just go ahead and add the `hello` files

sdmg15:
@memsharded Your suggestions sound cool, I can now link :smile: please review

memsharded (approved):
Great! I think now it is good, lets merge it, and it will be available in next 1.32.
Many thanks!

Issue #7948, opened 2020-10-27
[question] Using CONAN_USER_HOME from scratch


ngrodzitski:
I use docker. And keep a volume for CONAN_USER_HOME (mounted to `/_conan_home` when running). On new machine machine it is empty. If I try to use `conan install` I got the following error:

```
Traceback (most recent call last):
  File "/usr/bin/conan", line 11, in <module>
    load_entry_point('conan==1.30.2', 'console_scripts', 'conan')()
  File "/usr/lib/python3.8/site-packages/conans/conan.py", line 11, in run
    main(sys.argv[1:])
  File "/usr/lib/python3.8/site-packages/conans/client/command.py", line 2248, in main
    conan_api, _, _ = Conan.factory()
  File "/usr/lib/python3.8/site-packages/conans/client/conan_api.py", line 228, in factory
    return cls(), None, None
  File "/usr/lib/python3.8/site-packages/conans/client/conan_api.py", line 240, in __init__
    migrator = ClientMigrator(self.cache_folder, Version(client_version), self.out)
  File "/usr/lib/python3.8/site-packages/conans/client/migrations.py", line 27, in __init__
    self.cache = ClientCache(cache_folder, out)
  File "/usr/lib/python3.8/site-packages/conans/client/cache/cache.py", line 79, in __init__
    self._store_folder = self.config.storage_path or self.cache_folder
  File "/usr/lib/python3.8/site-packages/conans/client/cache/cache.py", line 151, in config
    self.initialize_config()
  File "/usr/lib/python3.8/site-packages/conans/client/cache/cache.py", line 256, in initialize_config
    save(self.conan_conf_path, normalize(get_default_client_conf()))
  File "/usr/lib/python3.8/site-packages/conans/util/files.py", line 200, in save
    with open(path, "wb") as handle:
FileNotFoundError: [Errno 2] No such file or directory: '/_conan_home/.conan/conan.conf'
```

I tried to do `conan config init`
```
$ conan config init
Traceback (most recent call last):
  File "/usr/bin/conan", line 11, in <module>
    load_entry_point('conan==1.30.2', 'console_scripts', 'conan')()
  File "/usr/lib/python3.8/site-packages/conans/conan.py", line 11, in run
    main(sys.argv[1:])
  File "/usr/lib/python3.8/site-packages/conans/client/command.py", line 2248, in main
    conan_api, _, _ = Conan.factory()
[...]

PR #7949, opened 2020-10-27
avoid loading config when not running migrations

memsharded:
Changelog: Fix: Avoid unnecessary extra loading of *conan.conf* file in the version migrations check.
Docs: Omit

Discover this while investigating: https://github.com/conan-io/conan/issues/7948

We are loading the **conan.conf** too many times. One of them is for migrations, even if migration is not necessary, not executed, creating a ``ClientCache()`` object unnecessarily loads the config. Delayed it instantiation to avoid it.

### DIFF
--- a/conans/client/migrations.py
+++ b/conans/client/migrations.py
@@ -22,14 +22,13 @@ from conans.util.files import list_folder_subdirs, load, save
 class ClientMigrator(Migrator):
 
     def __init__(self, cache_folder, current_version, out):
-        self.cache = ClientCache(cache_folder, out)
-        super(ClientMigrator, self).__init__(self.cache.cache_folder, self.cache.store,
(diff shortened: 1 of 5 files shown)


conans/client/migrations.py
```diff
@@ -22,14 +22,13 @@
 class ClientMigrator(Migrator):
 
     def __init__(self, cache_folder, current_version, out):
-        self.cache = ClientCache(cache_folder, out)
-        super(ClientMigrator, self).__init__(self.cache.cache_folder, self.cache.store,
-                                             current_version, out)
+        self.cache_folder = cache_folder
```
uilianries, line 25: Okay, cache_folder can't be None, because ConanAPI forces an object.

uilianries (approved):

Issue #7818, opened 2020-10-05
Proof of Concept of Testing with multiple tooling


memsharded:
We want the unittest to pass, even if users don't have all tools installed. We use a bunch of them:

- Cmake, SVN, pkg-config, Visual...

We want control over which versions run by default for tools, and control of multiple versions for the same tool

jgsogo:
First approach to this issue. Alternatives:
 * `nose` is dead, it shouldn't be considered an alternative
 * `pytest`: (can run `unittest`)
 * `unittest`

---

## `unittest`

We can write our own tools on top of unittests. We would need something to add _tags_ or _marks_ to the tests and probably write our custom decorator. Not hard at first sight, but it might get complicated, custom and hard to maintain/understand.

```python
@conantesting_attribs(['cmake', 'gcc',...])
class MyTestCase(...): ...
```

```python
def conantesting_attribs(obj, attr):
    # Check list of attributes against some condition (JSON, PATH,...)
    condition_match = ...
    if condition_match:
        return lambda func: func
    return unittest.skip("{!r} doesn't satisfy attributes {!r}".format(obj, attr))
```

## `pytest`

pytest already provides all these functionality, with hooks and plugins. Much more powerful and better prepared to handle any situation we might imagine.

But, there are several drawbacks in order to adopt `pytest`. It uses a more modern approach and tests inheriting from `unittest.TestCase` have only very few features available (https://docs.pytest.org/en/stable/unittest.html), among those we cannot use they are listed custom hooks (probably not needed) and markers (POC needed here).

Nevertheless, a POC implementation of marks **works with classes inheriting from `unittest.TestCase`**. This works:

**conftest.py**
```python
import pytest

# More about hooks: https://doc.pytest.org/en/2.7.3/plugins.html?highlight=re#pytest-hook-reference


tools_available = {
    'cmake': '3.10',
    'gcc': '7'
}

tools_keys = tools_available.keys()

def pytest_runtest_setup(item):
    # Every mark is a required tool, some specify a version
    for mark in item.iter_markers():
        if mark.name not in tools_keys:
            pytest.skip("required {} not satisfied".format(mark.name))
        else:
            version = mark.kwargs.get('version', None)
            if version:  # Implement version_range, regex,...
                pytest.skip("required version {} {} not satisfied".format(mark.name, version))
```

**test_module.py**
```python
import unittest
import pytest

@pytest.mark.cmake
class TestModule(unittest.TestCase):

    @pytest.mark.gcc
[...]

Issue #6527, opened 2020-02-11
[feature] Move Conan tests to pytest


uilianries:
Hi!

As you know, I'm a pytest fan due its facility for testing. For example, when I write an unit test, I don't want to remember the method name for that specific test (e.g. assertNotIn), I just want to put `assert` and use a regular condition.

Also, sometimes I need to write a tool for my tests, like a small http server to mock my answer. Or let's say I'm running another python framework like Flask, and I need some tools for testing. Well, pytest offers +800 plugins, including Flask integration: https://plugincompat.herokuapp.com/

But you can say that there is big code base for testing in your project, and porting is not possible. No problem, pytest is able to run nose: https://docs.pytest.org/en/latest/existingtestsuite.html

In summary , pytest [helps you write better programs](https://docs.pytest.org/en/latest/#pytest-helps-you-write-better-programs)

This suggestion could be adopted for Conan 2.0

What I see to be done for moving Conan to pytest:

- Rename test files to `test_*.py`
- Rename test methods to `test_*()`
- Replace tools which are already covered by pytest plugins

- [X] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

jgsogo:
Count with my sword :wink: 

Right now we are starting to plan the transition to Conan 2.0, the compatibility mode and after that we should maintain two branches: something like an LTS for Conan 1.x and the `rc` for Conan 2.0. This is going to increase the complexity of our CI and test suite and moving to `pytest` cannot take precedence over the roadmap.

But... probably we can add a branch to `conan_ci_jenkins` using `pytest` (running this `develop` branch ) and start moving things and checking that everything works (and fails) the same way --> https://github.com/conan-io/conan_ci_jenkins/issues/15

memsharded:
> Rename test files to test_*.py

I would say this is no longer necessary, from the docs:

> File names should start or end with “test”, as in test_example.py or example_test.py

That doesn't mean we shouldn't have a convention. Our codebase use so far xxxx_test.py. The truth is that in that way it is a bit less evident when tests are skipped because they don't match, but I think renaming all the files in the test suite is too much, isn't it?

uilianries:
no problem with file names now, I think that was a limitation for older versions. The only limitation would be the method names, but now is fixed, so all good.

jgsogo:
[...]

PR #7952, opened 2020-10-27
1 more step to remove nose

lasote:
Changelog: Feature: Changed recommended way to launch test suite, with pytest over nosetests.
Docs: Omit

Doesn't need docs, just README instructions

- [x] Added pytest custom marks in addition to the current nose attrs.
- [x] Changed readme and tox to use pytest as the recommended way.

What's next?

- [ ] We could change the Jenkins to use pytest, but until this is not in master branch it will fail, but would be necessary to somehow try it sooner.
- [ ] Introduce new/better marks so https://github.com/conan-io/conan/issues/7818 can be implemented.

Part of #6527


conans/test/functional/build_helpers/cmake_flags_test.py
```diff
@@ -1,7 +1,9 @@
 import os
 import platform
+import pytest
```
memsharded: pytest is not a builtin python module, imports shouldn't be here (but not a big problem)

memsharded (approved):
Looks good!

jgsogo (commented):
In order to test it in the CI, I'd recommend to rename the Jenkinsfile and create a different multibranch project in Jenkins. It could also need a matching branch (using `pytest`) of the `conan_ci` library.

Existing Jenkins project will run the current Jenkinsfile with the current default `conan_ci` branch. New Jenkins project will be configured with the new Jenkinsfile and a different branch of the `conan_ci` library.

As this PR evolves or it is merged in `develop`, the old CI project will stop to run (no matching Jenkinsfile) and the new one will take care of pull-requests and changes. Eventually, when all PRs are rebased, we might consider using the previous default branch for `conan_ci`.

---

Following this approach, we can remove the `@attr(...)` in this PR too, remove `nose` from the requirements,... and check it works with pytest.

lasote:
I think it is easier to open first a PR in the Jenkins library to add a new function that will cause "pytest" to run instead of "nosetests". Then modify here the Jenkinsfile to call that. The "new pytest" library will keep evolving with new pipelines to run first only python tests, then tools, and so on (like the pictures at https://github.com/conan-io/conan_ci_jenkins/pull/10) so when merged, It will be automatically enabled (on each branch). Eventually, we could remove the old library that launches nosetest.

conans/test/functional/build_helpers/cmake_flags_test.py
lasote: right. I'll change them.

lasote:
Also, I think it is better to merge small chunks to avoid a gigantic PR. I'll fix the imports order and I vote to merge this to develop (because it doesn't affect anything but recommending to use pytests)

memsharded:
> Also, I think it is better to merge small chunks to avoid a gigantic PR. I'll fix the imports order and I vote to merge this to develop (because it doesn't affect anything but recommending to use pytests)

Yeah, I agree, we can go step by step, the CI job can an is better to be done in another PR.

czoido (approved):

PR #7957, opened 2020-10-28
[refact] Create LocalDB object in a single place

jgsogo:
Changelog: omit
Docs: omit

Motivation: Needed so we can configure it (assign some _encryption key_) in a single place

### DIFF
--- a/conans/client/cache/cache.py
+++ b/conans/client/cache/cache.py
@@ -2,7 +2,6 @@ import os
 import platform
 import shutil
 from collections import OrderedDict
-from os.path import join
 
 from jinja2 import Environment, select_autoescape, FileSystemLoader, ChoiceLoader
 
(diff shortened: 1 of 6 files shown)


danimtb (approved):

conans/client/cache/cache.py
```diff
@@ -154,37 +154,38 @@ def config(self):
 
     @property
     def localdb(self):
-        return join(self.cache_folder, LOCALDB)
+        localdb_filename = os.path.join(self.cache_folder, LOCALDB)
+        return LocalDB.create(localdb_filename)
```
memsharded, line 158: Do we want to cache the created instance, just in case someone abuse calling ``cache.localdb``? Or convert it to a method instead of a property, something like ``create_localdb()``?

memsharded (approved):
Looks good, just a minor question, but can be merged without changing anything too.

conans/client/cache/cache.py
jgsogo: I didn't want to make more changes. Before we were returning the path and creating the instance, now we are returning the instance just created. So, we are instantiating it the same number of times. Too many times.

I can create the localdb in the `__init__` method and then return the object that is already created `get_localdb()`. IMHO, if we want to cache at function level we should use a decorator to make it always in the same way.

memsharded: Agree

PR #7958, opened 2020-10-28
Use key/secret to obfuscate values in the local database

jgsogo:
Changelog: Feature: Add `CONAN_LOGIN_ENCRYPTION_KEY` environment variable to obfuscate stored auth token.
Docs: https://github.com/conan-io/docs/pull/1903

This PR implements a level of obfuscation over the data stored in the local database. It uses any value provided (if provided) in the `CONAN_LOGIN_ENCRYPTION_KEY` envvar to compose the data that will be stored in the database. **It cannot be considered an encryption mechanism, this is NOT OK FOR SECURITY**, but it can be useful for CI pipelines: the CI job assigns a random-unique value for each build, with this simple implementation the data stored in the Conan cache is no longer usable by other jobs/builds in the same machine.

#PYVERS: py27


danimtb:
@jgsogo please rebase from develop now that #7957 is merged

jgsogo:
> @jgsogo please rebase from develop now that #7957 is merged

It is already rebased 🤔

conans/util/encrypt.py
```diff
@@ -0,0 +1,20 @@
...
+    if six.PY3:
+        return urlsafe_b64encode(bytes(key + data, 'utf-8'))
+    else:
+        return urlsafe_b64encode((key.encode('utf-8') + data).encode('utf-8'))
+
+
+def decode(enc, key):
+    assert isinstance(enc, bytes), "Expected 'bytes', got '{}'".format(type(enc))
+    assert isinstance(key, str), "Expected 'str' type, got '{}'".format(type(key))
+    return urlsafe_b64decode(enc)[len(key):].decode('utf-8')
```
memsharded: This seems a bit weak, as it brings the length of the key only as the thing to "crack".  I would say it is worth to at least use something like the Vigenere-Cipher, that would require way more effort to crack, and seems quite straight forward to implement.

danimtb (approved):

czoido (approved):

memsharded (approved):

PR #7962, opened 2020-10-28
Bump cryptography to >= 3.2

### LINKED ISSUE #6290: [question] Cryptography version lock on OSX
Seems like Cryptography is locked to an old version specifically on OSX, I guess there was a reason for it at the time. We are using multiple packages that seems to be conflicting with that requirement and I'm wondering if anyone can take a look and check if it is still necessary or just left behind?
For sure it works on Linux and Windows with newer Cryptography version - only OSX is locked and conflicts.

- [X ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

czoido:
Changelog: Fix: Bump _cryptography_ dependency in MacOS to equal or later than 3.2.
Docs: omit

Close https://github.com/conan-io/conan/issues/6290

#TAGS: svn, slow
#REVISIONS: 1

### DIFF
--- a/conans/requirements.txt
+++ b/conans/requirements.txt
@@ -15,6 +15,6 @@ tqdm>=4.28.1, <5
 Jinja2>=2.9, <3
 python-dateutil>=2.7.0, <3
 idna==2.6 ; sys_platform == "darwin" # Solving conflict, somehow is installing 2.7 when requests require 2.6
-cryptography>=1.3.4, <2.4.0 ; sys_platform == "darwin"
+cryptography>=3.2, <4 ; sys_platform == "darwin"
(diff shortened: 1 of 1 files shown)


memsharded:
@lasote do you recall what issues did we have with OSX requests and SSL for adding this, and why the version limitation?

lasote:
I don't. We can only bump and see.

memsharded (approved):

jgsogo:
I've installed `cryptography 3.2.1` locally, I'll report back if I find any issue.

jeremad:
thanks <3

PR #7966, opened 2020-10-28
Better UX error for migrations

uilianries:
I didn't find specific test file only related to Migrator or ClientMigrator, so I created an unit test to validate the current error message.

This change should affect more tests, as now `save` raises an error. Let's see ...

Changelog: Fix: Improve permission error message when migrating cache folder.
Docs: Omit

Related to #7948

/cc @ngrodzitski @memsharded 

### DIFF
--- a/conans/client/migrations.py
+++ b/conans/client/migrations.py
@@ -95,7 +95,7 @@ class ClientMigrator(Migrator):
             migrate_editables_use_conanfile_name(cache)
 
         if old_version < Version("1.31.0"):
-            migrate_tgz_location(self.cache, self.out)
+            migrate_tgz_location(cache, self.out)
 
 
 def _get_refs(cache):
--- a/conans/migrations.py
+++ b/conans/migrations.py
@@ -32,6 +32,8 @@ class Migrator(object):
     def _update_version_file(self):
         try:
             save(self.file_version_path, str(self.current_version))
+        except OSError as error:
+            raise ConanException(str(error))
         except Exception:
             raise ConanException("Can't write version file in %s" % self.file_version_path)
 
--- a/conans/migrations.py
+++ b/conans/migrations.py
@@ -32,10 +32,9 @@ class Migrator(object):
     def _update_version_file(self):
         try:
             save(self.file_version_path, str(self.current_version))
-        except OSError as error:
-            raise ConanException(str(error))
-        except Exception:
-            raise ConanException("Can't write version file in %s" % self.file_version_path)
+        except Exception as error:
+            raise ConanException("Can't write version file in '{}': {}"
+                                 .format(self.file_version_path, str(error)))
 
     def _load_old_version(self):
         try:
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -193,7 +193,8 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
         try:
             os.makedirs(os.path.dirname(path))
         except OSError as error:
-            raise OSError(errmsg.format(dir_path, error.strerror))
+            if "permission denied" in error.strerror.lower():
+                raise OSError(errmsg.format(dir_path, error.strerror))
         except Exception as error:
             raise OSError(errmsg.format(dir_path, str(error)))
 
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -192,9 +192,6 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
         if error.errno == errno.EACCES:
             raise OSError("Could not create the directory '{}' due to lack of permission."
                           " Please, check your user write permission.".format(path))
-        elif error.errno == errno.EEXIST:
-            raise OSError("Could not create the directory '{}' because it already exists."
-                          .format(path))
     except Exception:
         pass
 
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -188,6 +188,13 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
     """
     try:
         os.makedirs(os.path.dirname(path))
+    except OSError as error:
+        if error.errno == errno.EACCES:
+            raise OSError("Could not create the directory '{}' due to lack of permission."
+                          " Please, check your user write permission.".format(path))
+        elif error.errno == errno.EEXIST:
+            raise OSError("Could not create the directory '{}' because it already exists."
+                          .format(path))
     except Exception:
         pass
 
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -189,14 +189,14 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
     """
     dir_path = os.path.dirname(path)
     if not os.path.isdir(dir_path):
-        errmsg = "The folder {} does not exist and could not be created ({})."
         try:
-            os.makedirs(os.path.dirname(path))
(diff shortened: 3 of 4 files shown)


conans/util/files.py
```diff
@@ -188,6 +188,13 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
     """
     try:
         os.makedirs(os.path.dirname(path))
+    except OSError as error:
+        if error.errno == errno.EACCES:
+            raise OSError("Could not create the directory '{}' due to lack of permission."
+                          " Please, check your user write permission.".format(path))
+        elif error.errno == errno.EEXIST:
+            raise OSError("Could not create the directory '{}' because it already exists."
```
memsharded: If it exists it is OK, no need to raise!

uilianries: Yes, after seeing tons of test errors on my local environment, I agree :joy:

conans/util/files.py
```diff
@@ -188,6 +188,10 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
     """
     try:
         os.makedirs(os.path.dirname(path))
+    except OSError as error:
+        if error.errno == errno.EACCES:
+            raise OSError("Could not create the directory '{}' due to lack of permission."
+                          " Please, check your user write permission.".format(path))
```
memsharded: Directory name is ``os.path.dirname(path)``, not ``path``

uilianries: Done

conans/client/migrations.py
```diff
@@ -95,7 +95,7 @@ def _make_migrations(self, old_version):
             migrate_editables_use_conanfile_name(cache)
 
         if old_version < Version("1.31.0"):
-            migrate_tgz_location(self.cache, self.out)
+            migrate_tgz_location(cache, self.out)
```
memsharded: This is weird, it is already fixed in develop

uilianries: lemme force a pull here

uilianries: Okay, fixed after doing a new pull.

memsharded (approved):

conans/util/files.py
```diff
@@ -189,6 +189,10 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
     """
     try:
         os.makedirs(os.path.dirname(path))
+    except OSError as error:
+        if error.errno == errno.EACCES:
```
memsharded: I think this has some potential to break something. The folder might exist, but created by another user, and instead of getting a EEXIST error, get an EACCES error? Who knows.

I think I would make sense to check the existence of the folder, something like:

```python
folder = os.path.dirname(path)
if not os.path.isdir(folder):
     try:
         os.makedirs(folder)
     except Exception as e:
         raise OSError("The folder {} does not exist and could not be created: {}".format(folder, str(e)))
```

conans/migrations.py
```diff
@@ -32,6 +32,8 @@ def _make_migrations(self, old_version):
     def _update_version_file(self):
         try:
             save(self.file_version_path, str(self.current_version))
+        except OSError as error:
```
memsharded: Better just capture ``except Exception as error`` and include the ``str(error)`` message in the "Can't write version file in...", as it would be more complete and informative

memsharded (changes requested):

conans/util/files.py
uilianries: On Linux I'm sure it will result in a Permission error (errno.EACCES) if I try to re-create a folder which already exists, but created by another user. Anyway, I think don't harm using your logic, because I can say nothing about OSX and Windows behavior for this case.

conans/migrations.py
uilianries: Yes, as utils.file.save raises an informative message, my concern was about duplicated message. But in terms of code, it's too generic here, we can't deduce the problem without checking save() first.

conans/util/files.py
uilianries: Done.

conans/migrations.py
uilianries: Done.

conans/util/files.py
```diff
@@ -187,10 +187,16 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
...
-        os.makedirs(os.path.dirname(path))
-    except Exception:
-        pass
+    dir_path = os.path.dirname(path)
+    if not os.path.isdir(dir_path):
+        errmsg = "The folder {} does not exist and could not be created ({})."
+        try:
+            os.makedirs(os.path.dirname(path))
+        except OSError as error:
+            if "permission denied" in error.strerror.lower():
```
uilianries: I don't like it, but it's required because of test_get_gunzip. Maybe we should use errno.EACCES instead

conans/util/files.py
```diff
@@ -187,10 +187,16 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
         only_if_modified: file won't be modified if the content hasn't changed
         encoding: target file text encoding
     """
-    try:
-        os.makedirs(os.path.dirname(path))
-    except Exception:
-        pass
+    dir_path = os.path.dirname(path)
+    if not os.path.isdir(dir_path):
+        errmsg = "The folder {} does not exist and could not be created ({})."
```
memsharded: Only create the error message when it is going to be launched

conans/util/files.py
```diff
@@ -187,10 +187,16 @@ def save(path, content, only_if_modified=False, encoding="utf-8"):
...
+        try:
+            os.makedirs(os.path.dirname(path))
```
memsharded: Use ``dir_path`` here.

conans/util/files.py
memsharded: This indeed reads ugly and shouldn't be necessary. Maybe the test should be reviewed too.
It should be simpler, we already checked that the folder didn't exist. So, except some weird concurrency issue, this makedirs should succeed and create the folder. Probably just checking errno.EEXIST, just to prevent any concurrency issue, and any other makedir exception should just be reraised (with an improved error message if necessary)

uilianries: Sounds a good solution

conans/util/files.py
uilianries: Done.

conans/util/files.py
uilianries: Done.

conans/util/files.py
uilianries: The test `test_get_gunzip` raises `errno.ENOENT`, or "No such file or directory"

memsharded (approved):

uilianries:
I skipped Windows test because os.chmod has no effect under Windows folders, I would need other libraries, like [pywin32](https://sourceforge.net/projects/pywin32) or [oschmod](https://github.com/YakDriver/oschmod/blob/main/oschmod/__init__.py). 

The documentation is clear about Windows support: https://docs.python.org/2.7/library/os.html#os.chmod

> Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

It works for file only ... 

For some reason, if I inject `os.stat()` before calling `os.makedirs()` it works, but I won't change `save()` only because a single test.

I also tried subprocess call with `attrib +R dir_path /D` but had no effect.

PR #7975, opened 2020-10-30
Feature/tag tools

lasote:
Changelog: Feature: Tagged tests and created a `conftest.py` to run the tests with `pytest` skipping the tests using not available tools (cmake, visual studio...).
Docs: omit

Some tests are tagged as `tool_compiler` because some way o another they suppose a `settings.compiler` will be detected. Most of them could be fixed but I didn't want to change tests in this PR.

- [ ] Windows/Mac (pending to run the pytest test-suite in these OSS)
- [ ] There is no tool version comparison yet, maybe we can implement it later
- [x] Tested with both `unittest` and `functional` docker images from https://github.com/conan-io/conan_ci_jenkins/pull/22
- [x] This PR shouldn't affect at all the current test-suite, but we could start to provide a new one more sophisticated based on tool availability.

NOTE: There is a `SharedChainTest` test failing (also in my Linux from several months ago) because something changed in the Linux oss regarding rpaths if I recall correctly. We should check.


memsharded (approved):
This is looking very good. I totally agree that it is better to mark them and from there try to improve the tests, make tool-independent when possible (most times it will probably be just defining  a base default profile not autodetected)

conans/test/functional/generators/virtualenv_test.py
```diff
@@ -183,9 +183,12 @@ def _run_virtualenv(self, generator):
 
         env_before = _load_env_file(os.path.join(self.test_folder, self.env_before))
         env_after = _load_env_file(os.path.join(self.test_folder, self.env_after))
-        if platform.system() == "Darwin":
-            env_after.pop(six.u("PS1"), None)  # TODO: FIXME: Needed for the test to pass
-            env_after.pop("PS1", None)  # TODO: FIXME: Needed for the test to pass
+        # FIXME: Not the best behavior
+        # The deactivate sets PS1 always, but sometimes it didn't exist previously
+        there_was_ps1 = os.getenv("PS1")
```
memsharded: For some reason this is broken in CI: https://conan-ci.jfrog.info/blue/organizations/jenkins/ConanTestSuite/detail/PR-7975/1/pipeline

conans/requirements_dev.txt
```diff
@@ -1,6 +1,7 @@
 nose>=1.3.7, <1.4.0
 pytest>=6.1.1, <7.0.0; python_version > '3.0'
 pytest>=4.6.11; python_version < '3.0'
+pytest-xdist  # To launch in N cores with pytest -n
```
jgsogo, line 4: I would wait before running tests in parallel in the CI, maybe it introduces some hard-to-find failure. Let's work first on the testing framework and then we can make this PR to introduce parallelism.

conans/test/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    tools_available.remove("git")
+if not which("svn"):
+    tools_available.remove("svn")
+
+if not which("autoconf") or not which("automake"):
+    tools_available.remove("autotools")
+if not which("pkg-config"):
+    tools_available.remove("pkg_config")
+if not which("premake"):
+    tools_available.remove("premake")
```
jgsogo, line 44: IMO, it is the CI orchestrator the one that should tell the test_suite which tools are available, we shouldn't detect them in runtime; or we can stop running tests without noticing.

conans/test/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def tool_check(mark):
+    tool_name = mark.name[5:]
+    if tool_name not in tools_available:
+        pytest.skip("required {} not satisfied".format(tool_name))
```
jgsogo, line 50: We cannot add this check in the test suite, a typo will result in a skipped test and a green CI. Every check should fail on the safe side.

jgsogo (commented):
I need to know where we want to go with these changes, how the final process to launch the tests will look like. If tests are going to be inside different folders (unittest, functional, integration) then some of these _marks_ are not needed, the test will fail if it isn't in the proper folder because tools won't be available.

conans/requirements_dev.txt
memsharded: Yes, I agree, this is not currently used in this PR, so can be omitted at the moment.

conans/test/conftest.py
memsharded: Yes, agree, but if I understood it correctly, this PR is the work oriented to classify tests based on their tooling, and make it available for devs. So the syntax to have the CI to force the installed tools should be done in a later PR, this is one with 97 files changed is good as-is in this regard, no need to do everything in the same PR.

jgsogo: Then the review is totally different having that in mind, I thought this was related to our CI.

memsharded: I'd say this is a necessary step towards running in our CI, but not yet. Better do it step by step, lets instrument first the test-suite without changing the CI, and we can move from there. At least that was my understanding, please @lasote tell.

jgsogo (commented):
Minor concern related to the `conftest.py` using the default naming. I prefer a different naming, or some explicit input to use the auto-detection and skip so many tests.

OTH, how will our CI force us to add those `marks`? Otherwise, it is very likely we will forget about them.

memsharded:
> OTH, how will our CI force us to add those marks? Otherwise, it is very likely we will forget about them.

But if the mark is not added, the test will be run in the machines without the necessary tools and they will fail, isn't it the goal of this?

I think we don't want to use the folders as a mechanism to control tools at all. It is likely that most of the tests in "integration" will require some tool, but as the definitions of which tools have to be done anyway, then I don't see the need of using folders.

One of the issues that could happen is that some test is never run in our CI, maybe because it has some extreme conditions, lets say that a test explicitly requires "cmake 2.8.12" to test our "legacy" integration, but we don't have that in our CI (but the test has value for local development anyway?). Or there could be a typo. I see 2 possible approaches:

- Have some tool in pytest that collects tests skipped, and report and the end of the whole build which tests haven't run
- Force that every "mark" (with every version, if possible) has an entry defined in the config, even if it needs to be explicitly None to disable that test.

lasote:
> I need to know where we want to go with these changes, how the final process to launch the tests will look like. If tests are going to be inside different folders (unittest, functional, integration) then some of these marks are not needed, the test will fail if it isn't in the proper folder because tools won't be available.

Well, I think this PR enables the possibility of:
- Launching "no tools" only tests (could be done first)
- Launching only a concrete tool in several versions (e.g only cmake 3.0 and only cmake 3.9)
- ...
- And most important feature, keep everything as it is right now, so you can develop the new pipeline step by step.
    
The pipeline has to be defined and developed, this is just "tagging" tests.

> Minor concern related to the conftest.py using the default naming. I prefer a different naming, or some explicit input to use the auto-detection and skip so many tests.

One of the requirements of the feature is to ease to run the tests for the community. I see reasonable that the default launcher run the tests for the available tool only. 

> OTH, how will our CI force us to add those marks? Otherwise, it is very likely we will forget about them.

It depends on the implementation of the pipeline, but if we control the test environment, and it is complete enough if a test is not correctly tagged, it will fail.

> Force that every "mark" (with every version, if possible) has an entry defined in the config, even if it needs to be explicitly None to disable that test.

I think that is the right approach.

jgsogo (approved):
I was missing the purpose of this PR.

Issue #4216, opened 2018-12-31
[suggestion] Having os=None would be awesome for bare metal cross compilation


pepe82sh:
To help us debug your issue please explain:

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [ ] I've specified the Conan version, operating system version and any tool that can be relevant.
- [ ] I've explained the steps to reproduce the error or the motivation/use case of the question/suggestion.

Currently there is no option for having no OS. I'm playing with cross compilation in a bare metal environment, and it'd be great to have "None" as an option. I am aware that I can provide my own `settings.yml`, yet I think it could be a nice option for the rest of the world.

pepe82sh:
I understand your point at this matter. I created/forked a arm-none-eabi-gcc package where I thought of limiting the os setting to bare metal, yet I don't want to introduce non standard settings at that point. Right now I just ignore that setting, which should work okay as well. Thanks!

DavidZemon:
Can you re-open this?

As a workaround for #4382 , I created an OS in my local `settings.yml` with an OS of type "None" (as in, the string value `"None"`). I then check if the OS is equal to `"None"` and act accordingly based on that. However, the recipe fails to build in Travis for _any_ OS simply because I'm comparing to a non-existent OS. See [my pull request here](https://github.com/argtable/argtable3/pull/27) and [one of the failing builds here](https://travis-ci.org/argtable/argtable3/jobs/485565370#L682).

lasote:
When you specify the `None` value in the `settings.yml` file it doesn't matter that you can specify the "None" value as a setting, it means that you can avoid specifying it (not using os=XXX in the profile or in the -s). Then you can check `if not self.settings.os:"

DavidZemon:
Ah, thank you! I did not realize Conan treated "None" specially. That makes a lot more sense and does, in fact, make my life much easier. I do wish you'd add `None` into the default `settings.yml` file though.

lasote:
As memsharded said, it would enable mistakes when the `os` is not specified. Only for embedded and some other special cases, the `None` makes sense.  We really encourage the users to use the `conan config install` and distribute the `settings.yml` and any other config that way.
Anyway, happy it works for you!

jhol:
I just encountered this issue myself - before encountering this Issue.

I made my own patch which adds `None` as an option for `os` and `os_target`:

```
[...]

PR #7977, opened 2020-10-31
Added xtensalx6 for ESP32/ESP8266 platforms

redradist:
Changelog: Feature: New setttings: _xtensalx6_ and _xtensalx106_ for ESP32/ESP8266 platforms.
Docs: omit

There is related pull-request in cmake-conan repository https://github.com/conan-io/cmake-conan/pull/293

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -20,7 +20,7 @@ _t_default_settings_yml = Template(textwrap.dedent("""
     # Only for building cross compilation tools, 'os_target/arch_target' is the system for
     # which the tools generate code
     os_target: [Windows, Linux, Macos, Android, iOS, watchOS, tvOS, FreeBSD, SunOS, AIX, Arduino, Neutrino]
-    arch_target: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7]
+    arch_target: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7, xtensalx6, xtensalx106]
 
     # Rest of the settings are "host" settings:
     # - For native building/cross building: Where the library/program will run.
@@ -57,7 +57,7 @@ _t_default_settings_yml = Template(textwrap.dedent("""
         Emscripten:
         Neutrino:
             version: ["6.4", "6.5", "6.6", "7.0", "7.1"]
-    arch: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv4, armv4i, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7]
+    arch: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv4, armv4i, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7, xtensalx6, xtensalx106]
     compiler:
         sun-cc:
             version: ["5.10", "5.11", "5.12", "5.13", "5.14", "5.15"]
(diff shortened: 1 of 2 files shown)


conans/client/conf/__init__.py
```diff
@@ -17,12 +17,12 @@
 _t_default_settings_yml = Template(textwrap.dedent("""
     # Only for cross building, 'os_build/arch_build' is the system that runs Conan
     os_build: [Windows, WindowsStore, Linux, Macos, FreeBSD, SunOS, AIX]
-    arch_build: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, sh4le]
+    arch_build: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, sh4le, xtensalx6]
```
memsharded: ``arch_build`` is the architecture of the build machine. Is this architecture one you would be using to actually compile packages, or is this typically a cross-compile one?

redradist: > `arch_build` is the architecture of the build machine. Is this architecture one you would be using to actually compile packages, or is this typically a cross-compile one?

@memsharded It is architecture for which I will build package (for ESP32, xtensalx6)

memsharded: Ok, then it can be removed from ``arch_build``, as you will never run Conan itself in this platform.

redradist: @memsharded Okay, one second ;)

redradist: @memsharded Done

redradist: @memsharded Do you have any other thoughts on this pull-request ?

redradist: @memsharded Could you check ? I have updated pull-request to exclude merge conflicts, but there is some migration test failed ...

redradist: @czoido Could you check this pull-request ?

czoido:
Hi @redradist,
I think adding `xtensalx6` as architecture may not be enough to model this.
For what I have seen, I think that it would be better to add a separate new os, that maybe could be `ESP` and under it a subsetting like `cpu` that depending on the chip could be `lx6` (ESP32), `l106`(ESP8266) and leave arch as just `xtensa`
I have also seen that are two different choices for the target that are `elf` or `linux` (https://gcc.gnu.org/install/specific.html#xtensa-x-elf). Maybe that could be another subsetting?

WDYT @memsharded, @SSE4 ?

redradist:
> ESP

@czoido But there is not such OS like ESP !!

ESP32 works on FreeRTOS and Zephyr

`xtensalx6` is exact name of architecture according the Wiki and it is similar to `armv7` and etc.

czoido:
> 
> @czoido But there is not such OS like ESP !!
>
@redradist, I know there is not such OS, I was just wondering how that can be modelled, and took a model like the one we have for `Arduino` or `Emscripten` that are also not os's but yes, if those chips work on `FreeRTOS` and `Zephyr` then those os's are the ones that should be in os.
One question, isn't `xtensalx6` just valid for `ESP32`? Shouldn't the one for `ESP8266` be something like `xtensalx106` ?
Could you please link that wiki?
Thanks

redradist:
> > @czoido But there is not such OS like ESP !!
> 
> @redradist, I know there is not such OS, I was just wondering how that can be modelled, and took a model like the one we have for `Arduino` or `Emscripten` that are also not os's but yes, if those chips work on `FreeRTOS` and `Zephyr` then those os's are the ones that should be in os.
> One question, isn't `xtensalx6` just valid for `ESP32`? Shouldn't the one for `ESP8266` be something like `xtensalx106` ?
> Could you please link that wiki?
> Thanks

@czoido ESP32 https://en.wikipedia.org/wiki/ESP32, for gcc seems like it is called just xtensa

For ESP8266 in wiki nothing mentioned regarding the subset of xtensa architechure (:

czoido:
I think then probably the best way of adding this is adding architectures: `xtensa_lx6` for _ESP32_ and `xtensa_lx106` for _ESP8266_ and also adding os `FreeRTOS` (is that the default os for these boards?)

redradist:
> `xtensa_lx6`

@memsharded @czoido Maybe better `xtensalx6` and `xtensalx106`, because it would be similar with `aarch64`, `arm64v8`, `aarch64`, `arm64`, `armv7` ...
What do you think ?

redradist:
@memsharded @czoido Have you seen my comment above ?

redradist:
@memsharded @czoido Have you seen my comment above ?

@madebr Could you also review this PR ?

It is only 4 times added `xtensalx6` target arch

madebr (approved):
LGTM.

ESP32 -> Xtensa LX6 (https://en.wikipedia.org/wiki/ESP32)
ESP8266 -> Xtensa L106 (https://en.wikipedia.org/wiki/ESP8266)

Is there a check that ensures all settings of the previous settings were copied correctly?

redradist:
@madebr 
> Is there a check that ensures all settings of the previous settings were copied correctly?

Hard to say ...

czoido:
Hi @redradist,
Sorry for the delay, I'm adding this PR to milestone 1.41 to be considered.
Mentioning https://github.com/conan-io/conan/issues/4216 that is slightly related

redradist:
@madebr @memsharded This PR passed CI

Could you please review it ?

czoido (approved):

czoido:
This will be released in Conan 1.41

redradist:
@czoido @memsharded Probably it would be nice to create the specific compiler for it `xtensa-esp32-elf-gcc` ??
Is there a guidance how to do it ?

PR #7978, opened 2020-11-01
Refactor time (current timestamp) function and remove dead FileTreeManifest.__str__

memsharded:
Changelog: Omit
Docs: Omit

#tags: slow, svn
#revisions: 1

### DIFF
--- a/conans/model/manifest.py
+++ b/conans/model/manifest.py
@@ -96,17 +96,6 @@ class FileTreeManifest(object):
         content = "\n".join(ret)
         return content
 
-    def __str__(self):
-        raise
-        # TODO: check if this is used in the UI
-        dt = datetime.datetime.utcfromtimestamp(self.time).strftime('%Y-%m-%d %H:%M:%S')
-        ret = ["Time: %s" % dt]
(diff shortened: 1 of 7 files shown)


conans/model/manifest.py
```diff
@@ -89,22 +88,14 @@ def load(folder):
         return FileTreeManifest.loads(text)
 
     def __repr__(self):
+        # Used for serialization and saving it to disk
         ret = ["%s" % self.time]
```
danimtb, line 91: Not sure if it is important, but I think this should be UTC format to be the same as in the old `__str__()` method

memsharded: Not sure what you mean, we cannot change this, this is the way the conanmanifest.txt files are stored on disk, it is an integer (seconds), always UTC, so this cannot be changed, it could break a lot of things.

memsharded: I have restored the ``__str__`` function, it was a very corner case for displaying a message when Uploading and the manifest in the server is newer.

danimtb (approved):

PR #7980, opened 2020-11-01
Feature/lru

### LINKED ISSUE #5453: Implement something like git gc to clean out older versions of dependencies.
On my windows development machine I just realized I had over 30GB of dependencies stored in conan/data directory. This is something our developers has commented on as well. One reason our data dirs baloon that much is because we have a revision system that changes the version of the packages per change - but the same will be true now that official binary revisions are implemented in conan - lot of stuff we be stored and will build up over time.

I suggest we start storing the last used date in the local package directory - so when a package is used (i.e. installed from a conanfile) a timestamp is written to the package dir somewhere. We can then scan through all versions of each package and remove the ones that haven't been used in a while.

Another option is to just remove everything but the latest version. But that might be hard to know what the latest version is if some packages are not using semver versioning for example.

This should mostly like be added as a option to the remove command and then maybe add it as a option to do this automatically on some kind of schedule.

memsharded:

Close https://github.com/conan-io/conan/issues/5453

#tags: slow
#revisions: 1


jgsogo (commented):
I like the feature, and much more implemented at package-id level. If I want to save local storage I want to get rid of packages I'm no longer using (for example, if I once tried to build emscripten, or old Android versions, or that project I deployed to RaspberryPI...) I keep using `boost` recipe, but I no longer want those old binaries. 🎉 

---
(not reviewing, just suggesting)

About the implementation, I would try to automate the assignment of the timestamp instead of calling explicitly to `lru_now()`, maybe add it to the exit of the context manager... but we would need another context manager for the package level:

* recipe level
```python
with layout.update_metadata() as metadata:
    cur_revision = metadata.recipe.revision
    cur_remote = metadata.recipe.remote
    # ``lru_now`` is assigned behind the scenes to recipe level
```

* package level
```python
with layout.update_metadata(package_id) as metadata:
    metadata.revision = prev
    metadata.recipe_revision = rrev
    # ``lru_now`` is assigned behind the scenes to package-id level
    # ``lru_now`` is assigned behind the scenes to recipe level
```

The problem with this approach is that sometimes we need to call the function just to update the value: it is not nice and we can easily forget about it.

image

---

I think we need to define **what it means to use a recipe**: 
 + only when the recipe is used inside a Conan graph? what about methods like ``inspect``, ``download`` or ``upload`` that doesn't build the graph?
 + can we associate this mark to some method in the ``PackageCacheLayout``? to the ``conanfile_read_lock`` and the ``package_lock`` ones?
 + what about packages used to set up a development environment? You create those packages once, generate the corresponding ``virtualenv.sh`` file, add that file to your IDE startup and you forget about these packages... is there a workaround for it?
 + if it is only related to the graph, maybe it is the graph itself, after building everything, the one that should iterate the packages/recipes and add the mark. I mean, a dedicated visitor that runs like a _post_action_. (<-- This concept of _plugable_ visitors IMO is something we need to investigate, it would clarify the graph source code a lot)


---

memsharded:
> can we associate this mark to some method in the PackageCacheLayout? to the conanfile_read_lock and the package_lock ones?

Updating the metadata is a relatively slow operation, it needs to load the metadata, parse it, modify, serialize it again and save it again. I tried not to add more calls if not necessary, so I search the existing places were this was already being done, and only adding it when there was no other option.

 > can we associate this mark to some method in the PackageCacheLayout? to the conanfile_read_lock and the package_lock ones?

I don't love this idea, it could be a weird coupling that could work in some cases and not others. I like more the idea of a separate post-action that visits the graph and annotate things. Problem with that is that we are missing that infrastructure, and it seems to much for this feature.

> what about packages used to set up a development environment? You create those packages once, generate the corresponding virtualenv.sh file, add that file to your IDE startup and you forget about these packages... is there a workaround for it?

Not only for virtualenvs or dev environments. Regular library packages could happen the same: you do ``conan install`` get your build files like *conanbuildinfo.cmake* generated in your project and forget about it and keep working. If that takes a long time and some ``conan remove --old=xxx`` is done, yes, it is going to remove those packages, but Conan cannot do anything at all here, because it shouldn't run at all, and we are not going to interfere with users build scripts to annotate the LRU of the Conan cache. I don't think this is a problem, this doesn't seem the use case of the user that will call this feature.

> The problem with this approach is that sometimes we need to call the function just to update the value: it is not nice and we can easily forget about it.

I think this problem is quite small, and this is what is addressed with testing, and it shouldn't evolve very much once it is slightly stabilized.


I see other open issues:
- The build-folder and source folders are not covered by this feature. They need to be removed globally with ``conan remove -s -b``, but not possible based on date/lru (without modifiying the package-metadata.json, but I didn't want to do that at this stage)
- Is the package-metadata the best place to put this information? I have played with file timestamps, but it is not nice to do that with files that can be later packaged. We could add yet another dedicated file with a value in it (like we do with .dirty or locks)
- Really defining the "usage" criteria: is download usage? is inspect usage?

So what would be the recommendation to proceed? Should we do some core changes, like the visitor thing? I don't see those changes and investment at this stage. What would be the next steps for a good MVP of this feature?

jgsogo:
* This will be an opt-in (you run or not the remove command), but the problem with the development environment is real. If I configure my tools with Conan to be used by my IDE, I will keep using Conan for libraries and dependencies, I would like to libraries I haven't used for long but preserve my tools. It is not incompatible, maybe it deserves a _mark_ to prevent some packages from being removed. IIRC @Manu343726 was preparing his development environment with Conan. **But these considerations can wait for an additional feature.**

This is perfect as an MVP for this feature, it is even good as the feature itself, but IMO it is very important to learn from it some of the things we will need in the future: this kind of features shouldn't interfere with the core Conan functionality, they should be easily pluggable (from the POV of the developer, not as a user).

For me, it is clear we will need a way to _visit_ the graph and probably we also need a place to annotate data associated with references and packages. I think `metadata.json` is a good place to store this data (it is the closest we have to a database), but I'd try to isolate it from other data that could compromise Conan core functionality. What if a custom command (CLI 2.0) want to store data associate with a reference? will it use the `update_metadata()` function? how will we provide a place to store arbitrary data? I think this LRU could include an MVP of such a place, which is probably a different entry inside the `metadata.json`.

memsharded:
> This will be an opt-in (you run or not the remove command), but the problem with the development environment is real. If I configure my tools with Conan to be used by my IDE, I will keep using Conan for libraries and dependencies, I would like to libraries I haven't used for long but preserve my tools. It is not incompatible, maybe it deserves a mark to prevent some packages from being removed. IIRC @Manu343726 was preparing his development environment with Conan. But these considerations can wait for an additional feature.

Ok, now I see. Yes, this is kind of related with the long term idea of having a secondary cache, that could be even shared between Conan concurrent jobs, that will be mostly read-only and could host very heavy packages for toolchains.

Yes, I agree, this can be a future feature, it will probably be requested by users, can wait a bit until the core feature stabilizes.

> This is perfect as an MVP for this feature, it is even good as the feature itself, but IMO it is very important to learn from it some of the things we will need in the future: this kind of features shouldn't interfere with the core Conan functionality, they should be easily pluggable (from the POV of the developer, not as a user).

Well, it still needs a bit of thinking around the build and source folders.

> For me, it is clear we will need a way to visit the graph and probably we also need a place to annotate data associated with references and packages. I think metadata.json is a good place to store this data (it is the closest we have to a database), but I'd try to isolate it from other data that could compromise Conan core functionality. What if a custom command (CLI 2.0) want to store data associate with a reference? will it use the update_metadata() function? how will we provide a place to store arbitrary data? I think this LRU could include an MVP of such a place, which is probably a different entry inside the metadata.json.
 
I am not sure what you mean, this is already a different entry in the metadata.json. Can you clarify?
I agree that in general we want to revisit how we manage metadata in the Conan cache.

memsharded:
Lets wait for this feature to be built-in in the new cache design for Conan 2.0 and multi-revision, it seems too much trouble for the value to be added in the existing cache.

memsharded:
@jgsogo has started working on a new cache for Conan 2.0 that will integrate locks, metadata, lru information, etc. This will not be integrated in Conan 1.X (risky), so lets wait for the new cache.

PR #7986, opened 2020-11-02
fix short_path folders removal

### LINKED ISSUE #7983: [bug] `short_paths` option leads to creation redundant subdirectories in C:/.conan after `conan remove`
I tested this with global override `CONAN_USE_ALWAYS_SHORT_PATHS=1`, but I trust it's just the same for every package with `short_paths=True` in conanfile.

  * Windows:
  * VS 2013 (should be independent from compiler):
  * Conan 1.27.1 / 1.31:
  * Python 3.7.4 / 3.8:

- Run `conan install` on conanfile.txt containing N packages with `short_paths` feature enabled
- Check `C:/.conan` (or your `CONAN_USER_HOME_SHORT` if set) for the number of folders (4×N)
- Run `conan remove *` and then `conan install` again
- Watch number of folders in shortener become 7×N
- Each repeated `remove+install` cycle will increase the number of folders

```
C:\conantest
λ set CONAN_USE_ALWAYS_SHORT_PATHS=1
λ set CONAN_USER_HOME=C:\conantest
λ conan install -if b conanfile-win32.txt -pr conanprofile.vc120 -s build_type=Release
...
λ ls .conan/data | wc -l
38

bash-3.1$ ls /c/.conan | wc -l
152
bash-3.1$ for x in *; do if [[ -f $x/real_path.txt ]]; then echo "$x: $(cat $x/real_path.txt)"; fi done | grep zlib | /usr/bin/sort -t':' -b +2
99a5b5: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\export_source
03f603: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\package\25ae31d511843d8b24249c76bfff96b70097a22f
505002: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\scm_source
140320: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\source

C:\conantest
λ conan remove * -f
λ ls .conan/data | wc -l
0

bash-3.1$ ls /c/.conan | wc -l
114
bash-3.1$ for x in *; do if [[ -f $x/real_path.txt ]]; then echo "$x: $(cat $x/real_path.txt)"; fi done | grep zlib | /usr/bin/sort -t':' -b +2
99a5b5: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\export_source
505002: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\scm_source
140320: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\source

C:\conantest
λ conan install -if b conanfile-win32.txt -pr conanprofile.vc120 -s build_type=Release
...
λ ls .conan/data | wc -l
38

bash-3.1$ ls /c/.conan | wc -l
266
bash-3.1$ for x in *; do if [[ -f $x/real_path.txt ]]; then echo "$x: $(cat $x/real_path.txt)"; fi done | grep zlib | /usr/bin/sort -t':' -b +2
99a5b5: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\export_source
99a5b5a: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\export_source
03f603: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\package\25ae31d511843d8b24249c76bfff96b70097a22f
5050029: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\scm_source
505002: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\scm_source
140320: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\source
140320c: C:\conantest\.conan\data\zlib\1.2.8\trassir\sdk120\source
```
The quirk, I believe, is `util/windows.py : hashed_redirect`:
```python
    for length in range(min_length, max_length):
        redirect = os.path.join(base, full_hash[:length])
        if not os.path.exists(redirect): # <<<
            return redirect
```
, so either the check should be smth like `if not exists(redirect) or content(join(redirect, CONAN_REAL_PATH)) == join(base, path)`, or at least `conan remove` should also remove those `\source` subdirs. However the same thing happens for `package` shorten subdir, if instead of `conan remove` you do simply `rm -rf /c/conantest/.conan/data`, so I believe the enhanced check is a way to go.
Otherwise after a dozen of such remove+install cycles the hasher will run into `max_length` and resolve back to nondeterministic paths (which contradicts #3971 where it first apperaed).

Note of interest: in Conan 1.27.1 (which we currently use in our CI) the `source` subdirs are not created immediately - some of them (like source) appear immediately after `conan install` command, but some (like export_source) - only after `conan remove`. Which is kinda weird, but probably isn't a problem anyway since it's 4 versions behind.

memsharded:
Changelog: Bugfix: Remove correctly short-paths folders in Windows.
Docs: Omit

Fix https://github.com/conan-io/conan/issues/7983

#tags: slow, svn
#revisions: 1


conans/paths/package_layouts/package_cache_layout.py
```diff
@@ -153,15 +153,31 @@ def package_remove(self, pref):
...
         rmdir(export_folder)
-        export_src_folder = self.export_sources()
+        export_src_folder = os.path.join(self._base_folder, EXPORT_SRC_FOLDER)
         rm_conandir(export_src_folder)
         download_export = self.download_export()
         rmdir(download_export)
-        scm_folder = self.scm_sources()
-        rmdir(scm_folder)
+        scm_folder = os.path.join(self._base_folder, SCM_SRC_FOLDER)
+        rm_conandir(scm_folder)
```
jgsogo, line 180: Do we need here the same copy/paste as above with try/catch?

```python
        try:
            rm_conandir(....)  # This will remove the shortened path too if exists
        except OSError as e:
            raise ConanException("%s\n\nFolder: %s\n"
                                 "Couldn't remove folder, might be busy or open\n"
                                 "Close any app using it, and retry" % (scm_folder, str(e)))
```

conans/paths/package_layouts/package_cache_layout.py
```diff
@@ -153,15 +153,31 @@ def package_remove(self, pref):
...
+            raise ConanException("%s\n\nFolder: %s\n"
+                                 "Couldn't remove folder, might be busy or open\n"
+                                 "Close any app using it, and retry" % (src_folder, str(e)))
+        scm_folder = os.path.join(self._base_folder, SCM_SRC_FOLDER)
+        try:
+            rm_conandir(scm_folder)  # This will remove the shortened path too if exists
+        except OSError as e:
+            raise ConanException("%s\n\nFolder: %s\n"
+                                 "Couldn't remove folder, might be busy or open\n"
+                                 "Close any app using it, and retry" % (scm_folder, str(e)))
```
jgsogo, line 170: In this message, I'm always missing information about the actual path being removed, the shortened one, which would be the one blocking.  I agree with the changes as they are now, but this is something to improve.

conans/test/assets/genconanfile.py
```diff
@@ -355,6 +366,11 @@ def __repr__(self):
             ret.append("    {}".format(self._provides_line))
         if self._deprecated_line:
             ret.append("    {}".format(self._deprecated_line))
+        if self._short_path is not None:
+            ret.append("    short_paths = {}".format(str(self._short_path)))
+        if self._exports_sources:
+            line = ", ".join('"%s"' % e for e in self._exports_sources)
```
jgsogo:
```suggestion
            line = ", ".join('"{}"'.format(e) for e in self._exports_sources)
```

conans/test/functional/cache/short_paths_test.py
```diff
@@ -132,3 +137,28 @@ class TestConan(ConanFile):
         client.run("install test/1.0@", assert_error=True)
         self.assertIn("ERROR: Missing binary: test/1.0:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9",
                       client.out)
+
+    @parameterized.expand([(True,), (False,)])
+    def test_leaking_folders(self, use_always_short_paths):
```
jgsogo, line 142: I think we are mixing two different tests here. One is to check that the shortened folders are removed too, and the other one is to check that ``general.use_always_short_paths`` and attribute ``short_paths`` have the same effects. In this PR I would expect only the first one, the other should be already tested.

jgsogo (changes requested):

conans/paths/package_layouts/package_cache_layout.py
memsharded: This is a good question. There wasn't any check before for these folders, and I am trying not to change those things unless necessary. I think we might want to do it more systematically in the future, but probably that is a different effort than this PR.

conans/test/functional/cache/short_paths_test.py
memsharded: The thing is that this bite me during the testing, so it is clearly not fully tested elsewhere. Actually ``use_always_short_paths`` is very undertested, only some test is checking the env-var, not the conan.conf. So I think it doesn't harm (it is actually good) to test that neither way of defining short-paths leak folders in the short-path storage

conans/paths/package_layouts/package_cache_layout.py
memsharded: But the OSError exception should contain the failing path, and the exception is printed too, right?

jgsogo: Uhm, that's probably right, but I cannot test it outside Windows...

jgsogo (approved):

PR #7988, opened 2020-11-02
Refactor/gen sources

memsharded:
Changelog: Omit
Docs: Omit

Moving some files from toolchains to the ``assets`` location, as templates.


conans/test/assets/sources.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+
+
+def gen_function_cpp(name, msg=None, includes=None, calls=None, preprocessor=None):
+    t = Template(_function_cpp)
+    msg = msg or name
+    includes = includes or []
+    calls = calls or []
+    preprocessor = preprocessor or []
+    return t.render(name=name, msg=msg, includes=includes, calls=calls, preprocessor=preprocessor)
```
jgsogo:
```suggestion
def gen_function_cpp(**context):
    t = Template(_function_cpp)
    return t.render(**context)
```

conans/test/assets/sources.py
```diff
@@ -0,0 +1,76 @@
...
+
+#include <iostream>
+
+{% for it in includes -%}
+#include "{{it}}.h"
+{%- endfor %}
+
+int {{name}}(){
+    #ifdef NDEBUG
+    std::cout << "{{msg}}: Release!\n";
```
jgsogo:
```suggestion
    std::cout << "{{ msg|default(name) }}: Release!\n";
```

conans/test/assets/sources.py
```diff
@@ -0,0 +1,76 @@
...
+_function_cpp = r"""
+{% if name != "main" %}
+#include "{{name}}.h"
+{% endif %}
+
+#include <iostream>
+
+{% for it in includes -%}
+#include "{{it}}.h"
+{%- endfor %}
```
jgsogo, line 12:
```suggestion
{% if includes %}{% for it in includes -%}
#include "{{it}}.h"
{%- endfor %}{% endif %}
```

or using a `default` filter

jgsogo (commented):
If we are using Jinja2, I think we need to start using the features it provides. These testing files can be a good place to start getting some practice with it.

I did a couple of suggestions with two basic things I would start to use.

wdyt? cc @danimtb

memsharded:
Yeah, I totally agree, I was trying to isolate the asset generation and show how they can be reused, saving test code, because that is exactly what later enables a cleaner code with these patterns. Thanks for the suggestions!

jgsogo (approved):

PR #7993, opened 2020-11-03
Relax lockfiles config_options and configure options evaluation

### LINKED ISSUE #7991: [bug] conan install followed by conan install with lockfile fails
This bug report started from this PR which has most relevant information. 

https://github.com/conan-io/conan-center-index/pull/3382

I believe I've identified the root cause of the problem, but could be wrong. 

The user is able to run `conan install` with the profile provided which creates a lockfile. If they run `conan install` again and pass that lockfile as a parameter, it fails with this error. 
```
ERROR: paho-mqtt-c/1.3.1: paho-mqtt-cpp/1.1 tried to change paho-mqtt-c/1.3.1 option shared to False
but it was already defined as True
```

The specific code from the Conanfile in the error is here in `config_options()`. 

https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-c/all/conanfile.py#L32-L43

```python
    def config_options(self):
            del self.options.fPIC
        # Static linking before 1.3.4 isn't supported
        if tools.Version(self.version) < "1.3.4":
            self.options.shared = True
```

So the most obvious explanation is that Conan initializes the whole graph and hard-sets all the values, and then forbids any value-changing operations.  But, I am not confident this is the problem because there would be an obvious problem, in that any recipe that sets any value in `config_options()` would simply be un-usable with lockfiles.  So, it seems like this problem have been encountered and resolved long ago unless it's a regression.  But then again, here's the code that raises the error and it's 2 years old. 

https://github.com/conan-io/conan/blob/f563439d18dacf4e7310c08c07386f95070eccfc/conans/model/options.py#L134-L137

  * Operating System+version: All 
  * Compiler+version: All
  * Conan version: 1.31.0
  * Python version: All

Of note, the original user report provided both a separate profile for host and build, but I'm not sure that's relevant.  

```
conan install . --install-folder build/android3/debug --profile:build native --profile:host nsdk-android-armv8 -g=virtualenv                                   
conan install . --install-folder build/android4/  -l build/android3/conan.lock

```

```
conan install . --install-folder build/android3/debug --profile:build native --profile:host nsdk-android-armv8 -g=virtualenv    

Configuration (profile_host):
[settings]
arch=armv8
build_type=Release
compiler=clang
compiler.libcxx=libc++
compiler.version=9
os=Android
os.api_level=21
[options]
libcurl:with_ssl=openssl
lua:compile_as_cpp=True
ms-gsl:on_contract_violation=throw
nlohmann_json:multiple_headers=True
paho-mqtt-c:asynchronous=True
paho-mqtt-c:shared=False
paho-mqtt-c:ssl=True
paho-mqtt-cpp:ssl=True
[build_requires]
*: android-ndk/r21d, cmake/3.18.4, ninja/1.10.1
[env]

Configuration (profile_build):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=12.0
os=Macos
os_build=Macos
[options]
[build_requires]
[env]


conan install . --install-folder build/android4/  -l build/android3/conan.lock
...
ERROR: paho-mqtt-c/1.3.1: paho-mqtt-cpp/1.1 tried to change paho-mqtt-c/1.3.1 option shared to False
but it was already defined as True

### PR DESCRIPTION
**Pull request by memsharded**
Changelog: Fix: When using lockfiles, allow ``config_options`` and ``configure`` to compute different options as long as the final evaluated values match the locked ones.
Docs: Omit

Fix https://github.com/conan-io/conan/issues/7991

#tags: slow, svn
#revisions: 1

### CODE REVIEW

File: conans/model/graph_lock.py [hunk 1]
```diff
@@ -494,7 +494,6 @@ def pre_lock_node(self, node):
             node.graph_lock_node = locked_node
             if locked_node.options is not None:  # This was a "partial" one, not a "base" one
                 node.conanfile.options.values = locked_node.options
-                node.conanfile.options.freeze()
```
memsharded, line 507: This is the key of this PR: the options are not freezed anymore, so they do not error while being evaluated and getting different values.

conans/client/graph/graph_binaries.py
```diff
@@ -343,6 +343,16 @@ def evaluate_graph(self, deps_graph, build_mode, update, remotes, nodes_subset=N
         for node in deps_graph.ordered_iterate(nodes_subset=nodes_subset):
             self._propagate_options(node)
 
+            # Make sure that locked options match
+            if (node.graph_lock_node is not None and
```
memsharded, line 349: This has replaced the previous ``options.freeze()`` instead of detecting when someone is trying to modify a frozen option value, check at the end of the process that the final options match the locked ones.

conans/test/functional/graph_lock/dynamic_test.py
```diff
@@ -403,18 +403,20 @@ def configure(self):
...
+        expected = textwrap.dedent("""\
+                       ERROR: LibA/1.0: Locked options do not match computed options
+                       Locked options:
+                       myoption=False
+                       Computed options:
+                       myoption=True""")
+        self.assertIn(expected, client.out)
+
+        # Order of LibC, LibB does matter, in this case it will not raise
+        client.save({"conanfile.py": GenConanfile().with_requires("LibC/1.0", "LibB/1.0")})
```
memsharded, line 455: This test has changed, it cannot raise an error anymore. It will be a new (missing) binary for the added ``LibC``, that will get the locked option for ``LibA:myoption``. Depending on the evaluation error it can be different, and this is annoying but at least it is correct in the sense that it respects the lockfile, but in some cases it will error early, and in other cases the locked value will prevail and if anything it can fail later (missing binary, or binary cannot be built).

This is not ideal, but I think it is complicated and it is not worth such a huge investment, and this can wait until the new graph in Conan 2.0

conans/client/conanfile/configure.py
```diff
@@ -32,6 +32,7 @@ def run_configure_method(conanfile, down_options, down_ref, ref):
         conanfile.settings.validate()  # All has to be ok!
         conanfile.options.validate()
         # Recipe provides its own name if nothing else is defined
+        # FIXME: This shouldn't be here
```
jgsogo: Where instead? `init()` function closer to `set_name`?

conans/client/graph/graph_binaries.py
```diff
@@ -343,6 +343,16 @@ def evaluate_graph(self, deps_graph, build_mode, update, remotes, nodes_subset=N
...
+                    node.graph_lock_node.options is not None and
+                    node.conanfile.options.values != node.graph_lock_node.options):
+                raise ConanException("{}: Locked options do not match computed options\n"
+                                     "Locked options:\n{}\n"
+                                     "Computed options:\n{}".format(node.ref,
+                                                                    node.conanfile.options.values,
+                                                                    node.graph_lock_node.options))
```
jgsogo: This should be the other way around, right?

```suggestion
                                                                    node.graph_lock_node.options,
                                                                    node.conanfile.options.values))
```

conans/test/functional/graph_lock/graph_lock_test.py
```diff
@@ -387,6 +387,77 @@ def test_base_options(self):
...
+            """)
+
+        client.save({"pahomqttc/conanfile.py": pahomqttc,
+                     "pahomqttcpp/conanfile.py": pahomqttcpp,
+                     "consumer/conanfile.txt": "[requires]\npahomqttcpp/1.0"})
+        client.run("export pahomqttc pahomqttc/1.0@")
+        client.run("export pahomqttcpp pahomqttcpp/1.0@")
+
+        client.save({"conanfile.txt": "[requires]\npahomqttcpp/1.0"}, clean_first=True)
+        client.run("install . --build -o paho-mqtt-c:shared=False")
```
jgsogo: You are creating `conanfile.txt` twice

```suggestion
        client.run("install consumer/conanfile.txt --build -o paho-mqtt-c:shared=False")
```

conans/test/functional/graph_lock/graph_lock_test.py
```diff
@@ -387,6 +387,77 @@ def test_base_options(self):
...
+        lockfile = client.load("conan.lock")
+        print(lockfile)
```
jgsogo:
```suggestion
```

conans/test/functional/graph_lock/graph_lock_test.py
```diff
@@ -387,6 +387,77 @@ def test_base_options(self):
...
+        self.assertIn('"options": "pahomqttc:shared=False"', lockfile)
+        self.assertNotIn('shared=True', lockfile)
+        client.run("install . --lockfile=conan.lock")
```
jgsogo:
```suggestion
        client.run("install consumer/conanfile.txt --lockfile=conan.lock")
```

conans/client/conanfile/configure.py
memsharded: Yes, I think that can be defined at ``Conanfile`` class ``initialize()`` time, together with the ``requires``, ``settings`` and other stuff that is initialized very early in the recipe. At that point in time, the ``name`` attribute should be already defined.

conans/test/functional/graph_lock/graph_lock_test.py
```diff
@@ -387,6 +387,77 @@ def test_base_options(self):
...
+            """)
+
+        client.save({"pahomqttc/conanfile.py": pahomqttc,
+                     "pahomqttcpp/conanfile.py": pahomqttcpp,
+                     "consumer/conanfile.txt": "[requires]\npahomqttcpp/1.0"})
+        client.run("export pahomqttc pahomqttc/1.0@")
+        client.run("export pahomqttcpp pahomqttcpp/1.0@")
+
+        client.save({"conanfile.txt": "[requires]\npahomqttcpp/1.0"}, clean_first=True)
+        client.run("install . --build -o paho-mqtt-c:shared=False")
```
jgsogo:
```suggestion
        client.run("install consumer/conanfile.txt --build -o paho-mqtt-c:shared=False")
```

conans/test/functional/graph_lock/graph_lock_test.py
```diff
@@ -387,6 +387,77 @@ def test_base_options(self):
...
+        lockfile = client.load("conan.lock")
+        self.assertIn('"options": "pahomqttc:shared=True"', lockfile)
+        # Check the trailing ", to not get the profile one
+        self.assertNotIn('shared=False"', lockfile)
+        client.run("install . --lockfile=conan.lock")
```
jgsogo:
```suggestion
        client.run("install consumer/conanfile.txt --lockfile=conan.lock")
```

jgsogo (commented):
I like the idea of validating the options after the graph has been built. I've added some suggestions, have a look at them, the most important one is the new exception message, I think that the order of parameters is wrong.

---

(Digression for Conan 2.0)

I've been playing with different scenarios and the resulting behavior makes sense to me. 
BUT I would like to consider a different approach to the `default_options` for requirements' options. I mean, we are used to recipes like:

```python
        from conans import ConanFile

        class LibC(ConanFile):
            requires = 'LibA/1.0'

            def configure(self):
                self.options['LibA'].myoption = False
```

What is the purpose of this statement (or the similar one `default_options = {'LibA:myoption': False}`)? To _force_ the value of the option of a requirement. What for? Many times it is because `LibC` doesn't compile unless `LibA` uses that value for that option.

What would be the consequences of deprecating the _set value of requirement_ in favor of _validate the final value_ (upcoming PR)?

```python
        from conans import ConanFile

        class LibC(ConanFile):
            requires = 'LibA/1.0'

            def validate(self):
                assert self.options['LibA'].myoption == False, "LibC requires LibA:myoption=False because ...."
```

The user will need to set the value of `libA:myoption` **explicitly** in the profile/cli or use a lockfile with the required value (like the test that is being modified here), I'm sure that the implementation and the option resolution would be much simpler,...

conans/client/graph/graph_binaries.py
memsharded: Yep, seems a mistake, fixing it.

conans/client/conanfile/configure.py
jgsogo: IIRC this is written here because the `provides` can be conditioned to options, e.g.: some recipe might provide `ssl` if an option like `vendorized_ssl`  is activated.

jgsogo: This is the test: https://github.com/conan-io/conan/blob/aeb969369707005488df6d40785acecbffb5887c/conans/test/functional/provides/test_conditional_provides.py#L18

```suggestion
```

jgsogo (commented):
IMO, the FIXME doesn't apply, the `provides` can be assigned in the `configure()` method (there is a test to check it).

jgsogo (approved):
I'll copy the comment about options to a new issue.

PR #7989, opened 2020-11-02
pure C fix plus parameterized project code

memsharded:
Changelog: Fix: Make the ``conan new --pure_c`` pure C template to remove both ``compiler.libcxx`` and ``compiler.cppstd`` settings, as described in the docs.
Changelog: Feature: Make the ``conan new bye/0.1 -s -t`` to provide variable filenames and messages that include the package name and version, instead of a hardcoded "hello" one.
Docs: omit

PR #7994, opened 2020-11-03
Fix develop

jgsogo:
Changelog: omit
Docs: omit

`develop` branch was broken after parameterizing `conan new` templates in https://github.com/conan-io/conan/pull/7989

#tags: slow


memsharded (approved):
Thanks! 

I should have added "slow" tags

PR #7996, opened 2020-11-03
Better message for case incompatible package found

jgsogo:
Changelog: omit
Docs: https://github.com/conan-io/docs/pull/1910

After comment here https://github.com/conan-io/conan-center-index/issues/3379#issuecomment-720401803 , trying to make the message more explicit (not sure if we should add the recommendation: remove existing package)

### DIFF
--- a/conans/test/functional/case_sensitive_test.py
+++ b/conans/test/functional/case_sensitive_test.py
@@ -18,9 +18,9 @@ conanfile = textwrap.dedent('''
 ''')
 
 
+@unittest.skipIf(platform.system() == 'Linux', "Only for case insensitive OS")
 class CaseSensitiveTest(unittest.TestCase):
 
-    @unittest.skipIf(platform.system() == 'Linux', "Only for case insensitive OS")
     def test_install(self):
         test_server = TestServer()
         servers = {"default": test_server}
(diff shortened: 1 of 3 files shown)


conans/test/functional/case_sensitive_test.py
```diff
@@ -1,24 +1,26 @@
...
-        self.output.info("Running source!")
-'''
+        def source(self):
+            self.output.info("Running source!")
+''')
 
 
 class CaseSensitiveTest(unittest.TestCase):
 
+    @unittest.skipIf(platform.system() == 'Linux', "Only for case sensitive OS")
```
memsharded: Skip message is wrong, it is the opposite

conans/test/functional/command/remove_test.py
```diff
@@ -14,12 +11,13 @@
...
 from conans.server.store.server_store import ServerStore
 from conans.test.assets.cpp_test_files import cpp_hello_conan_files
+from conans.test.utils.mocks import TestBufferConanOutput
 from conans.test.utils.test_files import temp_folder
 from conans.test.utils.tools import NO_SETTINGS_PACKAGE_ID, TestClient, \
     TestServer, GenConanfile
-from conans.test.utils.mocks import TestBufferConanOutput
 from conans.util.env_reader import get_env
 from conans.util.files import load
+from mock import Mock
```
memsharded: Incorrect position for import

conans/test/functional/case_sensitive_test.py
memsharded: Maybe move the skipIf to class level?

danimtb (approved):

memsharded (approved):

memsharded:
> (not sure if we should add the recommendation: remove existing package)

Probably yes, add this suggestion in the error message.

PR #8001, opened 2020-11-04
propagate options downstream in graph

memsharded:

#tags: slow
#revisions: 1


memsharded:
This is in a good direction, but it is too much risk, it is almost impossible to not break something.

PR #8007, opened 2020-11-04
Test/ignore case

### LINKED ISSUE #7990: [bug] Version 1.30.2 broken copy_deps at least on Windows when ignore_case is False
I had a set of working conan packages with dependencies (p1 depends on p2).
When deployed, p1 also deploys files from its dependency p2 and uses the copy_deps() method for that.
I had ignore_case = False set for all platforms and everything worked fine with conan 1.21.1 64bit but stopped working in conan 1.30.2 32 bit. The call succeeds but no files are copied.
Deploying p2 directly works, thus thre is not issue with the case of the files as the same params are used in both cases. I believe that there is a problem in computing the package path for dependencies. I believe that the path is correct except for the case.
The src parameter is empty. The pattern parameter has "include/*", which is in the correct case.
If I change "ignore_case" to True, things work.
I don't know how to debug this any further.

memsharded:

Could close https://github.com/conan-io/conan/issues/7990

### DIFF
--- a/conans/client/file_copier.py
+++ b/conans/client/file_copier.py
@@ -161,7 +161,10 @@ class FileCopier(object):
                                                                          pattern)]
 
         for exclude in excludes:
-            files_to_copy = [f for f in files_to_copy if not fnmatch.fnmatch(f, exclude)]
+            if ignore_case:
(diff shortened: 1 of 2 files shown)


memsharded:
Closed in favor of #8009

PR #8014, opened 2020-11-05
msbuild generator custom configs

memsharded:
Changelog: Feature: Allow (experimental) custom configuration of the ``msbuild`` generator.
Changelog: Feature: Rename ``msbuild`` generator to ``MSBuildDeps`` and use the new ``generate()`` method.
Docs: https://github.com/conan-io/docs/pull/1952

Simplified from https://github.com/conan-io/conan/pull/7754/files

This is ONLY the ``msbuild`` part. There are 2 external things to be addressed too, but lets discuss those in other issues:

- The import structure for tools, helpers, etc.
- The ``config`` mechanism to propagate things from config and command line to the recipes and codebase.

#tags: slow


conan/tools/cmake/base.py
```diff
@@ -162,6 +163,21 @@ def _get_template_context_data(self):
...
+               "******************************************************************\n"
+               "'write_toolchain_files()' has been deprecated and moved.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generate()' method instead.\n"
+               "********************************************************************\n"
+               "********************************************************************\n")
+        from conans.client.output import Color, ConanOutput
+        ConanOutput(self._conanfile.output._stream,
+                    color=self._conanfile.output._color).writeln(msg, front=Color.BRIGHT_RED)
+        warnings.warn(msg)
```
jgsogo: This can be a deprecation warning

```suggestion
        warnings.warn(msg, category=DeprecationWarning)
```

conan/tools/cmake/base.py
```diff
@@ -162,6 +163,21 @@ def _get_template_context_data(self):
...
+        msg = ("\n*****************************************************************\n"
+               "******************************************************************\n"
+               "'write_toolchain_files()' has been deprecated and moved.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generate()' method instead.\n"
+               "********************************************************************\n"
+               "********************************************************************\n")
+        from conans.client.output import Color, ConanOutput
+        ConanOutput(self._conanfile.output._stream,
+                    color=self._conanfile.output._color).writeln(msg, front=Color.BRIGHT_RED)
```
jgsogo: Why not...?

```python
self._conanfile.output.warn(msg)
```

conans/client/conanfile/configure.py
```diff
@@ -47,6 +47,8 @@ def run_configure_method(conanfile, down_options, down_ref, ref):
 
 def _validate_fpic(conanfile):
     v2_mode = get_env(CONAN_V2_MODE_ENVVAR, False)
+    # FIXME: The toolchain() has dissapeared, but now it is integrated with generators.
+    # FIXME: Not sure this check should be here, this could raise before the graph is fully evaluated
     toolchain = hasattr(conanfile, "toolchain")
```
jgsogo: ?
```suggestion
    toolchain = hasattr(conanfile, "toolchain") or hasattr(conanfile, "generate")
```

conans/client/generators/__init__.py
```diff
@@ -125,6 +153,15 @@ def write_generators(self, conanfile, path, output):
 
 def write_toolchain(conanfile, path, output):
     if hasattr(conanfile, "toolchain"):
+        msg = ("\n*****************************************************************\n"
+               "******************************************************************\n"
+               "The 'toolchain' attribute or method has been deprecated.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generators = ClassName' or 'generate()' method instead.\n"
```
jgsogo:
```suggestion
               "Use 'generators = \"ClassName\"' or 'generate()' method instead.\n"
```

conans/client/generators/__init__.py
```diff
@@ -138,6 +175,15 @@ def write_toolchain(conanfile, path, output):
                 raise ConanException("Unknown toolchain '%s'" % conanfile.toolchain)
             tc = toolchain(conanfile)
             with chdir(path):
-                tc.write_toolchain_files()
+                tc.generate()
+
+        # TODO: Lets discuss what to do with the environment
+
+    if hasattr(conanfile, "generate"):
+        assert callable(conanfile.generate), "generate should be a method, not an attribute"
```
jgsogo: We are not doing this for other methods. The Python error itself and the `conanfile_exception_formatter` should be enough.

```suggestion
```

conans/test/functional/toolchain/test_txt_cmdline.py
```diff
@@ -0,0 +1,50 @@
...
+        toolchain = client.load("conan_meson_native.ini")
+        self.assertIn("[project options]", toolchain)
+
+    def test_command_line(self):
+        conanfile = textwrap.dedent("""
+            [generators]
+            CMakeToolchain
+            MesonToolchain
+            MakeToolchain
+            MSBuildToolchain
```
jgsogo: If we are testing the command line `-g`, I think we should remove them from the `conanfile.txt`, right?

```suggestion
```

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -65,20 +65,54 @@ class MSBuildDeps(Generator):
...
+            return self._config_filename
+        # Default name
+        props = [("Configuration", self.configuration),
+                 ("Platform", self.platform)]
         name = "".join("_%s" % v for _, v in props if v is not None)
+        return name.lower()
+
+    @config_filename.setter
+    def config_filename(self, value):
+        self._config_filename = value
```
jgsogo: I'd remove this `config_filename` attribute, the user can already modify the name of the generated filename using `self.configuration` and `self.platform`, which are the arguments meaningful for VS IDE (`msbuild`) and are the ones considered in the __condition_.

conans/client/generators/__init__.py
```diff
@@ -83,10 +84,37 @@ def __contains__(self, name):
...
+            return CMakeToolchain
+        elif generator_name == "MakeToolchain":
+            from conan.tools.gnu import MakeToolchain
+            return MakeToolchain
+        elif generator_name == "MSBuildToolchain":
+            from conan.tools.microsoft import MSBuildToolchain
+            return MSBuildToolchain
+        elif generator_name == "MesonToolchain":
+            from conan.tools.meson import MesonToolchain
+            return MesonToolchain
```
jgsogo, line 97: See [line 121 in this file](https://github.com/conan-io/conan/pull/8014/files#diff-e235fa62562f3e2f47330538b120750466913d3d1733d75c6eee9a06906dc95fR121), we need to add these new generators to the list, otherwise the error message won't list all of them:

```python
                raise ConanException("Invalid generator '%s'. Available types: %s" %
                                     (generator_name, ", ".join(self.available)))
```

jgsogo (changes requested):
Some changes/suggestiongs, and main comment: IMO, we should remove the `config_filename`, it adds a degree of freedom that is not needed and it can be misleading because it is not considered in the _condition_:

```python
def toolchain(self):
   ms = MSBuildDeps(self)
   ms.configuration = "My"+str(self.settings.build_type)
   ms.platform = "My"+str(self.settings.arch)
   ms.config_filename = 'optTrue' if self.options.opt1 else 'optFalse'
   ms.generate()
```

We get two different files, but we cannot reach one of them, which one? It depends on the install order...

conan/tools/microsoft/msbuilddeps.py
memsharded: This was implemented because of your suggestions when thinking about this feature.
There is a use case, in which users can add explicitly specific generated .props files to their project files, without using the ``conan_depname.props`` that contains the condition.

I agree this seems a bit early, so I am going to remove it.

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -65,20 +64,40 @@ class MSBuildDeps(Generator):
...
-                 ("Platform", {'x86': 'Win32',
-                               'x86_64': 'x64'}.get(settings.get_safe("arch")))]
-
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self.configuration = conanfile.settings.build_type
+        self.platform = {'x86': 'Win32',
+                         'x86_64': 'x64'}.get(str(conanfile.settings.arch))
+        # TODO: this is ugly, improve this
+        self.output_path = os.getcwd()
```
jgsogo, line 73: Not needed here. We are already assigning a value before calling `generate()` **with a different value** (the class is instantiated before the `chdir` call): 

```python
                    generator = generator_class(conanfile)
                    output.highlight("Generator '{}' calling 'generate()'".format(generator_name))
                    generator.output_path = path
                    with chdir(path):
                        generator.generate()
```

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -65,20 +64,40 @@ class MSBuildDeps(Generator):
...
+
+    def generate(self):
+        # TODO: Apply config from command line, something like
+        # configuration = self.conanfile.config.generators["msbuild"].configuration
+        # if configuration is not None:
+        #     self.configuration = configuration
+        # platform
+        # config_filename
+        # TODO: This is duplicated in write_generators() function, would need to be moved
+        # to generators and called from there
```
jgsogo, line 83: We can check here that `self.configuration` and `self.platform` are different from `None`

jgsogo (changes requested):
Just a couple of comments

conan/tools/microsoft/msbuilddeps.py
memsharded: But this is needed if you are explicitly instantiating the ``MSBuildDeps()`` in the ``generate()`` method.
This is the reason it has a ``TODO``, I plan to put some time to put a nicer proposal regarding the "layout" or expected location of generated files, for implicit and explicit generators and both for cache and local flows.

conan/tools/microsoft/msbuilddeps.py
memsharded: I was adding the check:
```python
        if self.configuration is None:
            raise ConanException("MSBuildDeps.configuration is None, it should have a value")
        if self.platform is None:
            raise ConanException("MSBuildDeps.platform is None, it should have a value")
```
Then realized that both ``condition`` and ``configfilename`` already provide for a ``None`` value, filtering it. This might make sense in single-config environments (we have been reported that MSBuild can be used for Linux builds), so maybe it is not worth raising an exception?

conan/tools/microsoft/msbuilddeps.py
jgsogo: I never think about the generators like classes that the user will instantiate in the recipe 😓 

Ok, then it is needed, ugly, but needed 👍

conan/tools/microsoft/msbuilddeps.py
jgsogo: IIRC, VS IDE uses something like `AnyCPU` or `AnyPlatform` if the same configuration should apply for `x86` and `x64` (I cannot remember what was it like for the Configuration)... so it is assigning a value to the variable that could be evaluated in the _Condition_. I cannot check because I don't have a running Visual Studio.

I don't know which one is the right approach but if the IDE always assigns a value to those properties (Configuration and Platform) maybe it is better to take those values into account instead of ignoring them in the Condition... if that is the scenario, the user would need to assign `msbuild.platform = 'AnyCPU'`. 

No idea how it looks like for those Linux build, never seen them. Maybe there is strong evidence and all we need is a test.

conan/tools/microsoft/msbuilddeps.py
memsharded: > I never think about the generators like classes that the user will instantiate in the recipe

But this is exactly the use case one of the test is implementing:

```python
            from conans import ConanFile
            from conan.tools.microsoft import MSBuildDeps
            class Pkg(ConanFile):
                settings = "os", "compiler", "arch", "build_type"
                requires = "pkg/1.0"
                def generate(self):
                    ms = MSBuildDeps(self)
                    ms.configuration = "My"+str(self.settings.build_type)
                    ms.platform = "My"+str(self.settings.arch)
                    ms.generate()
```
The goal is to support user custom configurations in Visual Studio/MSBuild, which need to be parameterized by the users, and the best way is to be able to instantiate them directly in the ``generate()`` method. This is also how user provided custom generators (implemented in ``python_requires``) will need to be used too.

jgsogo: I know, but this is something new, I've never used it in my recipes...

conan/tools/microsoft/msbuilddeps.py
memsharded: You are right, it seems impossible to remove or leave empty this field, you can rename to a custom config, but you need to have one. Lets add the exceptionst then.

PR #8015, opened 2020-11-05
More pytest marks for windows

lasote:
Changelog: omit
Docs: omit

Used a Windows machine with nothing installed to verify the pytest marks are complete.
Added one more tool for meson and for "conan" itself (used in several tests). 
Fixed the `tool_check` function to check every tool mark applied to a test.

### DIFF
--- a/conans/test/functional/generators/virtualenv_test.py
+++ b/conans/test/functional/generators/virtualenv_test.py
@@ -5,6 +5,7 @@ import textwrap
 import unittest
 from collections import OrderedDict
 
+import pytest
 import six
 from parameterized.parameterized import parameterized_class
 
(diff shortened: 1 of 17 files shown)


conans/test/functional/generators/virtualenv_test.py
```diff
@@ -193,6 +194,7 @@ def _run_virtualenv(self, generator):
         self.assertDictEqual(env_before, env_after)  # Environment restored correctly
         return stdout, _load_env_file(os.path.join(self.test_folder, self.env_activated))
 
+    @pytest.mark.tool_conan
```
memsharded, line 197: This is bad, we should work to fix this

memsharded (approved):

PR #8016, opened 2020-11-05
Fix username & password in SCM for git clone with ssh

### LINKED ISSUE #8011: [bug] SCM username value is ignored
Conan ignores `username` SCM attribute.
For file:
```python
from conans import ConanFile

class TestRecipe(ConanFile):
    name = 'test-recipe'
    version = '0.1'
    scm = {
        'type': 'git',
        'url': 'ssh://github.com/conan-io/conan.git',
        'username': 'test-user',
        }
```
When cloning git repository, `test-user` is not used, but current logged-in user is used instead:
```bash
test-recipe/0.1: SCM: Getting sources from url: 'ssh://github.com/conan-io/conan.git'
ERROR: Command 'git -c http.sslVerify=true clone "ssh://github.com/conan-io/conan.git" .   ' returned non-zero exit status 128.
Cloning into '.'...
seva@github.com: Permission denied (publickey).
```
(conan.git repository is used just as example).

### Environment Details
Conan version: 1.31.0
Python version: 3.8.4, 3.7.3

### Steps to reproduce
- Create conanfile recipe (example is attached).
- run 'conan create .' in it's directory.

Conanfile and example output:
[username-bug.zip](https://github.com/conan-io/conan/files/5492773/username-bug.zip)

danimtb:
Changelog: Fix: Set username or password individually in git SCM with ssh.
Docs: omit

- [x] Refer to the issue that supports this Pull Request: closes #8011


jgsogo (approved):
There are a couple of tests

conans/test/unittests/client/tools/scm/test_scm_base.py
```diff
@@ -53,4 +56,27 @@ def test_local_windows(self):
...
+        self.assertEqual('ssh://dani:pass@github.com/conan-io/conan.git',
+                         scm.get_url_with_credentials("ssh://github.com/conan-io/conan.git"))
+
+    def test_ssh_username(self):
+        scm = SCMBase(username="dani")
+        self.assertEqual('ssh://dani@github.com/conan-io/conan.git',
+                         scm.get_url_with_credentials("ssh://github.com/conan-io/conan.git"))
+
+    def test_ssh_password(self):
+        scm = SCMBase(password="pass")
```
memsharded: I am a bit confused a bit about this test. So adding the password, but not the user, will not add the password, but will not error either?

danimtb: This was the behavior before but I am not sure how the user could set only the password. Maybe providing the user already in the URL?

memsharded: Well, that is the point, the previous behavior was requiring both user+password. Now we changed it, and adding only the user is a new valid behavior, with a new test checking it. 

I'd say that we need to define a new behavior for only the password: either an error for this case, or check that there is a valid scenario, like the user is already in the URL, but that should probably be a new test too.

danimtb:
I have added some warning messages for options that ignore parameters

conans/test/unittests/client/tools/scm/test_scm_base.py
```diff
@@ -53,4 +56,79 @@ def test_local_windows(self):
...
+    def test_ssh_password(self):
+        output = OutputMock()
+        scm = SCMBase(password="pass", output=output)
+        self.assertEqual('ssh://github.com/conan-io/conan.git',
+                         scm.get_url_with_credentials("ssh://github.com/conan-io/conan.git"))
+        self.assertIn("WARN: SCM username undefined, ignoring 'password' parameter", output.out)
+
+    def test_ssh_url_with_username_only_password(self):
+        scm = SCMBase(password="pass")
+        self.assertEqual('ssh://git:pass@github.com/conan-io/conan.git',
```
czoido: As we talked, I think you can't add the password in ssh url's

czoido (changes requested):

conans/test/unittests/client/tools/scm/test_scm_base.py
memsharded: yes, I have been investigating a bit, and I haven't found any evidence this is supported.

conans/test/functional/scm/issues/test_ssh_user.py
```diff
@@ -0,0 +1,35 @@
...
+    def test_ssh_username(self):
+        t = TestClient()
+        ref = ConanFileReference.loads("lib/version@issue/testing")
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile
+            
+            class Conan(ConanFile):
+                scm = {
+                        "type": "git",
+                        "url": "ssh://github.com/conan-io/conan.git",
```
czoido:
```suggestion
                        "url": "ssh://git@github.com/conan-io/conan.git",
```

conans/client/tools/scm.py
```diff
@@ -62,14 +62,38 @@ def run(self, command):
                         return self._runner(command)
 
     def get_url_with_credentials(self, url):
-        if not self._username or not self._password:
+        if not self._username and not self._password:
             return url
-        if urlparse(url).password:
+        url_parsed = urlparse(url)
+        if url_parsed.scheme == "ssh" and (url_parsed.password or self._password):
```
czoido: I think `url_parsed.scheme` does not work properly for scp-like syntax like: `git@github.com:memsharded/hello.git`

conans/test/functional/scm/issues/test_ssh_user.py
danimtb: with this test I wanted to reproduce the error in #8011. If I add the user to the URL, then the case is different

czoido: Oh!, I see 👍

danimtb:
I think the error in CI could be because we don't have ssh installed https://stackoverflow.com/questions/12569705/error-cannot-run-ssh-no-such-file-or-directory-when-trying-to-clone-on-windows ?

conans/client/tools/scm.py
danimtb: I don't know how to detect this if the scheme is not provided

czoido: Maybe we could detect the syntax with a regular expression, converting to a ssh:// url and then passing to urlparse? WDYT?

czoido: Maybe this could be enough if the `scheme==""` or we can normalize the url before urlparse

```python
if "@" in url and ":" in url:
    return "ssh://" + url.replace(":", "/")
```

danimtb: I did some tests and it is very fragile with that condition. It might work with a regex but manipulating the URL that much does not feel right to me. It will be weird for the user to change the provided URL. I will need a second opinion about it @memsharded

czoido: Yes, something like [this](https://github.com/golang/go/blob/979e1376096a4c1c7525f33dd4e76414f63c93fb/src/cmd/go/internal/vcs/vcs.go#L206) would be more robust in case we want to do this conversion.

memsharded: This whole feature feels a bit weird to me. Like, for a user defined the URL they want is a one liner, lets say:

```python
url = "https://{user}:{passwd}@someurl.com/repo".format(user=os.getenv("USER"), passwd=os.getenv("PASSWD")
```
Am I missing something?
While on our side is complicated and fragile, trying to cover all cases. Maybe we want to re-think the whole UX.

jgsogo: There are two sources for this URL: user-input or `auto` keyword. Parsing all formats can be very challenging ([😮  there are a lot](https://stackoverflow.com/questions/31801271/what-are-the-supported-git-url-formats)). Maybe we should:
 * check it matches some regex for well-known formats (those in Github)
 * if it doesn't match, the URL will be used without any transformation. If user/password are provided, print a warning telling the user that _"URL format unknown, user and password won't be used to clone"_ (and in this case, do not store user/pasword?).

czoido (approved):

conans/client/tools/scm.py
```diff
@@ -62,14 +62,67 @@ def run(self, command):
...
 
     def get_url_with_credentials(self, url):
-        if not self._username or not self._password:
+        if not self._username and not self._password:
             return url
-        if urlparse(url).password:
+        url_user = None
+        url_pass = None
+        scp_regex = re.compile("^([a-zA-Z0-9_]+)@([a-zA-Z0-9._-]+):(.*)$")
+        url_user_pass_regex = re.compile("^(file|http|git)s?:\/\/(\S+):(\S+)@("".*)$")
```
jgsogo: What are the quotes in this group `("".*)` intended for? I mean, they are concatenating the Python str... basically doing nothing here.

danimtb: looks like a leftover

jgsogo (commented):
My spider-sense says this is going to fail somewhere and somehow. I find too many if/else here. I'd suggest a dispatcher, something like:

```python
def handle_scp_pattern(user, domain, url):
    ... logic related to user, password and warnings.
    return "{usr}@{domain}:{url}".format(user=user, domain=domain, url=url)

def handle_url_pattern(protocol, url, user=None, password=Non):
    ... logic related to user, password and warnings
    auth = "{user}:{password}@" if xxxx else "{user}@" if yyyyy else ""
    return "{protocol}://{auth}{url}".format(protocol=protocol, auth=auth, url=url)

# Maybe other actions if the logic inside of them starts to be cumbersome

# Now a map/list we will iterate and use the matching dispatcher
url_patterns = {
   'scp': [
      re.compile("^(?P<user>[a-zA-Z0-9_]+)@(?P<domain>[a-zA-Z0-9._-]+):(?P<url>.*)$"),
      handle_scp_pattern
   ]
   'url_user_pass': [
      re.compile("^(?P<protocol>file|http|git)s?:\/\/(?P<user>\S+):(?P<password>\S+)@(?P<url>.*)$"),
      handle_url_pattern
   ]
   'url_user': [
      re.compile("^(?P<protocol>file|http|git)s?:\/\/(?P<user>\S+)@(?P<url>.*)$"),
      handle_url_pattern
   ]
}

for key, (pattern, action) in url_patterns:
    m = pattern.match(url)
    if m:
        return action(**m.groupdict())

self.output.warn("URL not recognized....")
return url
```

wdyt? This is just a suggestion. Let me know and I'll review carefully whatever you choose

danimtb:
@jgsogo I have followed your suggestion and I think the code is cleaner. Please check again as I have also added more tests. Thank you!

conans/client/tools/scm.py
```diff
@@ -61,15 +61,58 @@ def run(self, command):
                     else:
                         return self._runner(command)
 
+    def handle_scp_pattern(self, user, domain, url):
```
jgsogo: If not local to the function `get_url_with_credentials`, at least make them _private_

```suggestion
    def _handle_scp_pattern(self, user, domain, url):
```

conans/client/tools/scm.py
```diff
@@ -61,15 +61,58 @@ def run(self, command):
...
+        if self._password:
+            self._output.warn("SCM password cannot be set for scp url, ignoring parameter")
+        if self._username:
+            self._output.warn("SCM username got from URL, ignoring 'username' parameter")
+        return "{user}@{domain}:{url}".format(user=user, domain=domain, url=url)
+
+    def handle_url_pattern(self, scheme, url, user=None, password=None):
```
jgsogo: If not local to the function `get_url_with_credentials`, at least make them _private_

```suggestion
    def _handle_url_pattern(self, scheme, url, user=None, password=None):
```

conans/client/tools/scm.py
```diff
@@ -61,15 +61,58 @@ def run(self, command):
...
+        scp_regex = re.compile("^(?P<user>[a-zA-Z0-9_]+)@(?P<domain>[a-zA-Z0-9._-]+):(?P<url>.*)$")
+        url_user_pass_regex = re.compile("^(?P<scheme>file|http|https|git):\/\/(?P<user>\S+):(?P<password>\S+)@(?P<url>.*)$")
+        url_user_regex = re.compile("^(?P<scheme>file|http|https|git|ssh):\/\/(?P<user>\S+)@(?P<url>.*)$")
+        url_basic_regex = re.compile("^(?P<scheme>file|http|https|git|ssh):\/\/(?P<url>.*)$")
+
+        url_patterns = [
+            (scp_regex, self.handle_scp_pattern),
+            (url_user_pass_regex, self.handle_url_pattern),
+            (url_user_regex, self.handle_url_pattern),
+            (url_basic_regex, self.handle_url_pattern)
```
jgsogo:
```suggestion
            (scp_regex, self._handle_scp_pattern),
            (url_user_pass_regex, self._handle_url_pattern),
            (url_user_regex, self._handle_url_pattern),
            (url_basic_regex, self._handle_url_pattern)
```

conans/client/tools/scm.py
```diff
@@ -61,15 +61,58 @@ def run(self, command):
...
+                self._output.warn("SCM password cannot be set for file url, ignoring parameter")
+            return "{scheme}://{url}".format(scheme=scheme, url=url)
+        elif scheme == "ssh" and self._password:
+            self._output.warn("SCM password cannot be set for ssh url, ignoring parameter")
+
+        if user and self._username:
+            self._output.warn("SCM username got from URL, ignoring 'username' parameter")
+        if password and self._password:
+            self._output.warn("SCM password got from URL, ignoring 'password' parameter")
+        if the_password and the_user and scheme != "ssh":
```
jgsogo, line 88:
```suggestion

        the_user = user or self._username
        the_password = password or self._password
        if the_password and the_user and scheme != "ssh":
```

conans/client/tools/scm.py
```diff
@@ -61,15 +61,58 @@ def run(self, command):
...
+    def handle_scp_pattern(self, user, domain, url):
+        if self._password:
+            self._output.warn("SCM password cannot be set for scp url, ignoring parameter")
+        if self._username:
+            self._output.warn("SCM username got from URL, ignoring 'username' parameter")
+        return "{user}@{domain}:{url}".format(user=user, domain=domain, url=url)
+
+    def handle_url_pattern(self, scheme, url, user=None, password=None):
+        the_user = user or self._username
+        the_password = password or self._password
```
jgsogo: Let's move them closer to where they are used

```suggestion
```

jgsogo (commented):
Some suggestions. Now it's much easier to follow all the conditions 🎉

jgsogo (approved):

memsharded (approved):
This is much better structured and understandable, and likely to be more robust. Well done.

czoido (approved):

PR #8024, opened 2020-11-06
Feature/generator cmake custom configs

memsharded:
Changelog: Feature: New ``CMakeDeps`` generator (at the moment is the ``cmake_find_package_multi``, that allows custom configurations, like ``ReleaseShared``.
Docs: https://github.com/conan-io/docs/pull/1992

Similar to https://github.com/conan-io/conan/pull/8014/files, but for cmake

#tags: slow


conans/client/generators/cmake_find_package_multi.py
```diff
@@ -299,7 +313,7 @@ def content(self):
                 find_lib = target_template.format(name=pkg_findname, deps=deps,
                                                   build_type_suffix=build_type_suffix,
                                                   deps_names=deps_names)
-                ret["{}Target-{}.cmake".format(pkg_filename, build_type.lower())] = find_lib
+                ret["{}Target-{}.cmake".format(pkg_filename, self.configuration.lower())] = find_lib
```
SSE4, line 316: so, configuration no longer matches build type?

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+            self.client.run_command('cmake .. -G "Visual Studio 15 Win64"')
+            self.client.run_command('cmake --build . --config Release')
+            self.client.run_command(r"Release\\app.exe")
+            self.assertIn("hello/0.1: Hello World Release!", self.client.out)
+            self.assertIn("main: Release!", self.client.out)
+            self.client.run_command('cmake --build . --config ReleaseShared')
+            self.client.run_command(r"ReleaseShared\\app.exe")
+            self.assertIn("hello/0.1: Hello World Release!", self.client.out)
+            self.assertIn("main: Release!", self.client.out)
```
SSE4, line 96: why output is still `main: Release!` and not `main: ReleaseShared!`?

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
```diff
@@ -0,0 +1,93 @@
...
+        build_directory = os.path.join(self.client.current_folder, "build").replace("\\", "/")
+        with self.client.chdir(build_directory):
+            self.client.run("install .. %s -o hello:shared=True" % settings)
+            self.client.run("install .. %s -o hello:shared=False" % settings)
+
+            self.client.run_command('cmake .. -G "Visual Studio 15 Win64"')
+            self.client.run_command('cmake --build . --config Release')
+            self.client.run_command(r"Release\\app.exe")
+            self.assertIn("hello/0.1: Hello World Release!", self.client.out)
+            self.assertIn("main: Release!", self.client.out)
```
SSE4, line 92: do we check if ReleaseShared config cmake is generated?

conans/client/generators/cmake_find_package_multi.py
memsharded: It matches by default the ``build_type``. But the user can have custom configurations, like ``ReleaseShared``, so this allows the user to provide and manage those cases too.

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
memsharded: Because the source code is parameterized (#ifdef NDEBUG) with the Debug build_type, not the build-system configuration. It will output ``Release`` is both cases, but in one case it will be running with external DLLs dependencies, and in the other with static linked deps.

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
memsharded: It should be there, otherwise it would error. I have added checks to test to make sure:
```diff
+self.assertTrue(os.path.isfile(os.path.join(self.client.current_folder,
+                                                        "helloTarget-releaseshared.cmake")))
```

SSE4 (approved):

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
SSE4: yes, but I prefer explicit check, as cmake find logic is pretty complicated, the library might be found by other means

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -265,7 +279,7 @@ def filename(self):
     def content(self):
         ret = {}
         build_type = str(self.conanfile.settings.build_type).upper()
-        build_type_suffix = "_{}".format(build_type) if build_type else ""
+        build_type_suffix = "_{}".format(self.configuration.upper()) if self.configuration else ""
```
SSE4, line 282: I am not sure if it could cause problems if I have something like (corner case):
```
cmake.configurations = ["Debug", "debug", "DEBUG"]
```
do we already know If cmake configurations are case insensitive?

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
memsharded: Sure, check added explicit. In any case, it is a red-green test, if you remove the ``imports()`` method, the test will be red as it cannot execute due to the missing dll.

conans/client/generators/cmake_find_package_multi.py
memsharded: Yes, I guess it could cause problems. Up to my knowledge the CMake generator expressions for configuration are case insensitive, so that definition shouldn't be done.

conans/test/functional/generators/cmake_find_package_multi_configs_test.py
```diff
@@ -0,0 +1,97 @@
...
+                    "compiler.version": "15",
+                    "arch": "x86_64",
+                    "build_type": "Release",
+                    }
+
+        settings = " ".join('-s %s="%s"' % (k, v) for k, v in settings.items() if v)
+
+        # Run the configure corresponding to this test case
+        build_directory = os.path.join(self.client.current_folder, "build").replace("\\", "/")
+        with self.client.chdir(build_directory):
```
jgsogo: (Unrelated to this PR) Is this needed, is it not enough...?

```suggestion
        with self.client.chdir('build'):
```

jgsogo (approved):
👍 

Next PR we should read and _add_ the new configuration as the MSBuildDeps does.

memsharded:
>Next PR we should read and add the new configuration as the MSBuildDeps does.

Yep, I'd love to do that, but I am a bit concerned about breaking, I think I would prefer to do it on a copy ``CMakeDeps`` generator that can start diverging from the ``cmake_find_package_multi`` one.

PR #8026, opened 2020-11-07
Feature: set CMAKE_SYSTEM_PROCESSOR

### LINKED ISSUE #8025: [bug] CMAKE_SYSTEM_PROCESSOR isn't set
from https://github.com/conan-io/conan-center-index/pull/3392
/cc @KristianJerpetjon

[CMAKE_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/v3.7/variable/CMAKE_SYSTEM_PROCESSOR.html) is not set by conan in CMake build helper.
as result, cross-compiling for armv8 fails for certain recipes (at least **libpng** and **libjpeg-turbo**)

```
if the variable is wrong then it selects sse instead of neon.. no need to explain that bit

you get a linker error when trying to link sse assembly into arm compiled code which is obvius
```

  * Operating System+version: macOS Catalina
  * Compiler+version: Apple clang version 11.0.3 (clang-1103.0.32.62)
  * Conan version: Conan version 1.31.0-dev
  * Python version: Python 3.8.0

I reproduced it by running:
```
conan create . libjpeg-turbo/2.0.5@ -k -pr ios
```
out:
```
CMake Warning at source_subfolder/simd/CMakeLists.txt:5 (message):
  SIMD extensions not available for this CPU ().  Performance will suffer.
```
profile:
```
Configuration for profile ios:

[settings]
os=iOS
os.version=9.0
arch=armv8
compiler=apple-clang
compiler.version=11.0
compiler.libcxx=libc++
build_type=Release
os_build=Macos
arch_build=x86_64
[options]
[build_requires]
[env]
```

```
cmake_minimum_required(VERSION 2.8.12)
project(cmake_wrapper)
message("CONAN_CMAKE_SYSTEM_PROCESSOR  $ENV{CONAN_CMAKE_SYSTEM_PROCESSOR} ")
message("CMAKE_SYSTEM_PROCESSOR ${CMAKE_SYSTEM_PROCESSOR}")
include(conanbuildinfo.cmake)
conan_basic_setup(NO_OUTPUT_DIRS)
add_subdirectory("source_subfolder")
Result when running conan create is.. 
CONAN_CMAKE_SYSTEM_PROCESSOR   
```

SSE4:
@KristianJerpetjon could you test this POC for your case, especially for **libpng**, **libjpeg-turbo** and others?

closes: #8025

Changelog: Feature: set CMAKE_SYSTEM_PROCESSOR


madebr:
Some extra data points.

I did some tests on Visual Studio 2019 using a little cmake script:
```
project(dummy)
message("CMAKE_SYSTEM_PROCESSOR: ${CMAKE_SYSTEM_PROCESSOR}")
message("CMAKE_SIZEOF_VOID_P: ${CMAKE_SIZEOF_VOID_P}")
```

The following command was executed:
```
cmake -G "Visual Studio 16 2019" -A <PLATFORM>
```

PLATFORM | CMAKE_SYSTEM_PROCESSOR | CMAKE_SIZEOF_VOID_P
-- | -- | --
`Win32` | `AMD64` | 4
`x64` | `AMD64` | 8
`ARM` | `ARM64` | 4
`ARM64` | `ARM64` | 8

memsharded:
Thanks @madebr for the hints!

@SSE4 I think that for moving this forward we need to:

- First wait to see if this is very necessary to be fixed in the current CMake build helper. If 1.31.4 stabilizes, we will not touch there anything at all. It is too risky.
- We will move this to the new toolchains, in this case CMakeToolchain, which are experimental
- I think that at first, the CMAKE_HOST_XXX variables are not needed, we can rely on CMake for it largely. Lets try to define only CMAKE_SYSTEM_PROCESSOR, and have at least a good mapping of most Conan architectures

Windows
-  https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details
Windows mapping AMD64, IA64, ARM64, x86

Linux:
- https://stackoverflow.com/questions/45125516/possible-values-for-uname-m

madebr:
For arm support of my [mpg123 cci recipe](https://github.com/conan-io/conan-center-index/pull/3630),
CMAKE_SYSTEM_PROCESSOR will be necessary for ARM support on Windows.
See https://sourceforge.net/p/mpg123/bugs/298/ for upstream discussion.

As long as `CMAKE_SYSTEM_PROCESSOR` is somehow defined :smile:

SSE4:
@memsharded updated:
- set only `CMAKE_SYSTEM_PROCESSOR`
- added more architectures to mapping for Windows and *nix

testing on real recipes and cross-building  on real platforms is welcomed

conans/client/build/cmake_flags.py
```diff
@@ -172,6 +172,61 @@ def _get_cpp_standard_vars(self):
...
+                                 "tvOS": apple_system_name or "tvOS",
+                                 "watchOS": apple_system_name or "watchOS",
+                                 "Neutrino": "QNX",
+                                 "": "Generic",
+                                 None: "Generic"}
+        return cmake_system_name_map.get(os_, os_)
+
+    @staticmethod
+    def _cmake_system_processor(os_, arch):
+        if os_ in ["Windows", "WindowsStore", "WindowsCE"]:
```
madebr, line 192: I *think* these names are valid for Visual Studio, not for Windows systems.
mingw expects the gnu names instead of Microsoft tainted ones.

SSE4: do you have a reference? the [doc](https://cmake.org/cmake/help/v3.19/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR) says:
```
On Windows, this variable is set to the value of the environment variable PROCESSOR_ARCHITECTURE
```
it doesn't mention compilers

madebr: I have no other reference.
But the cmake documentation mentions GNU, and mingw is a GNU compiler.

Otoh, I searched for `CMAKE_SYSTEM_PROCESSOR arm` and `CMAKE_SYSTEM_PROCESSOR aarch` on github and it looks like there is no *universal standard*.
I think we have to face the reality that there is no universally-agreed-to-standard to use as `CMAKE_SYSTEM_PROCESSOR` variable.
Like everything in the c/c++ world (except for the stl standaard), there is no standard. CMake allows everything. CMAKE_SYSTEM_PROCESSOR is just a hint.

In reality, some project will either use one kind `CMAKE_SYSTEM_PROCESSOR` mapping from processor to cmake variable and not switch between a Microsoft convention or GNU convention. Or will not switch between Windows convention or Linux convention.

It's like [the `arch` argument of `MsBuild.build`](https://docs.conan.io/en/latest/reference/build_helpers/visual_studio.html#build): there will always be a project having a different mapping (uppercase/lowercase/64 suffix).

Making this variable too hidden will make it harder for people wanting to modify it.
I'm just saying it would be nice to have an api that allows this:

```python
self._cmake.cmake_host_processor = tools.cmake_host_processor(self.settings.arch, "gnu").tolower()
```
So people can hook it.
Cci recipe writers will thank you.

And conan providing a proper default: the Microsoft one if MSVC else GNU.

SSE4: yes, I think it's correct. CMake itself doesn't care much about `CMAKE_SYSTEM_PROCESSOR` and lists it as optional.
the main source of headache is libraries checking for `CMAKE_SYSTEM_PROCESSOR`, for instance, to include proper assembler sources, such as libjpeg-turbo case.
although CMake specifies `CMAKE_SYSTEM_PROCESSOR` to have values either of `uname` or `PROCESSOR_ARCHITECTURE`, I've seen in reality libraries checking for different values, e.g. for `armv8` they may check for `armv8`, `arm64` and `aarch64`.
examples:
- [aarch64](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/ccaba5d7894ecfb5a8f11e48d3f86e1f14d5a469/CMakeLists.txt#L57)
- [arm64](https://github.com/xianyi/OpenBLAS/blob/1c0b03efb42dfc829198c74391206ef4900441e6/cmake/system_check.cmake#L57)
- [armv8-a](https://github.com/hohocb/xmrig/blob/3296c0d7f3f78540cec7df98507330c8c9a0e953/cmake/cpu.cmake#L11)

KristianJerpetjon: On 1.33.0 we needed this patch to make the libjpeg-turbo cross compile for "ARM64/aarch64/armv8 etc." (not saying its not libjpeg turbos fault)

madebr:
Looking at this, I think it's fine to set these variables, when cross building.
As long as a recipe creator can override these.

sorny92:
> @memsharded updated:
> 
> * set only `CMAKE_SYSTEM_PROCESSOR`
> * added more architectures to mapping for Windows and *nix
> 
> testing on real recipes and cross-building on real platforms is welcomed

I tested in linux x86_64 targeting armv8 and the fix works for me.

memsharded:
Closing this PR as outdated, it was targeting the legacy integrations, not the new CMakeToolchain

PR #8040, opened 2020-11-10
appending existing LocalDebuggerEnvironment in msbuild generator

### LINKED ISSUE #8038: [bug] msbuild generator: setting of LocalDebuggerEnvironment does not work with multiple dependencies
  * Compiler+version: Visual Studio 2019
  * Conan version: 1.31.1
  * Python version: 3.8.3

1. Require at least two dependencies with shared libraries (e.g. boost and sqlite3)
2. Call `conan install . -g msbuild`
3. Include the generated `conan_deps.props`
4. Debug an executable in Visual Studio
5. The executable will fail to load the dlls

The problem is that the `PATH` environment variable set via the `<LocalDebuggerEnvironment>` property set in the props file does not include the directory of all dependencies.

E.g. with:
conan_deps.props:
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" ToolsVersion="4.0">
		<Import Project="conan_boost.props"/>
		<Import Project="conan_sqlite3.props"/>
	</ImportGroup>
</Project>
```
conan_boost.props:
```xml
[...]
	<PropertyGroup>
		<LocalDebuggerEnvironment>PATH=%PATH%;$(ConanboostBinaryDirectories)</LocalDebuggerEnvironment>
		<DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
	</PropertyGroup>
[...]
```
conan_sqlite3.props:
```xml
[...]
	<PropertyGroup>
		<LocalDebuggerEnvironment>PATH=%PATH%;$(Conansqlite3BinaryDirectories)</LocalDebuggerEnvironment>
		<DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
	</PropertyGroup>
[...]
```

The environment will only contain a modified `PATH` with the last included dependency, here sqlite3. It should hovered include the binary directories of all dependencies, like it is done by the `visual_studio` or `visual_studio_multi` generator.

Note that I myself am unsure how/if the chaining will work if one includes the `conan_boost.props` and `conan_sqlite3.props` separately in the same project (which from my understanding is the main advantage over using `visual_studio_multi`).

But for the use case of including `conan_deps.props` there would be a simple solution/workaround: overwrite the `<LocalDebuggerEnvironment>` property in the `conan_deps.props` file after the other props file are included.

memsharded:
Changelog: Bugfix: Append existing ``LocalDebuggerEnvironment`` in ``msbuild`` generator.
Docs: Omit

Close https://github.com/conan-io/conan/issues/8038

#tags: slow

### DIFF
--- a/conans/client/generators/msbuild.py
+++ b/conans/client/generators/msbuild.py
@@ -38,7 +38,7 @@ class MSBuildGenerator(Generator):
             <conan_{name}_props_imported>True</conan_{name}_props_imported>
           </PropertyGroup>
           <PropertyGroup>
-            <LocalDebuggerEnvironment>PATH=%PATH%;$(Conan{name}BinaryDirectories)</LocalDebuggerEnvironment>
+            <LocalDebuggerEnvironment>PATH=%PATH%;$(Conan{name}BinaryDirectories)$(LocalDebuggerEnvironment)</LocalDebuggerEnvironment>
             <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
           </PropertyGroup>
           <ItemDefinitionGroup>
--- a/conans/test/assets/sources.py
+++ b/conans/test/assets/sources.py
@@ -9,7 +9,7 @@ _function_cpp = r"""
 
 {% for it in includes -%}
 #include "{{it}}.h"
-{%- endfor %}
+{% endfor %}
 
 int {{name}}(){
     #ifdef NDEBUG
@@ -40,7 +40,7 @@ int {{name}}(){
 
     {% for it in calls -%}
     {{it}}();
-    {%- endfor %}
+    {% endfor %}
     return 0;
 }
 """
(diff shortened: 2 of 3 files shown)


conans/client/generators/msbuild.py
```diff
@@ -38,7 +38,7 @@ class MSBuildGenerator(Generator):
             <conan_{name}_props_imported>True</conan_{name}_props_imported>
           </PropertyGroup>
           <PropertyGroup>
-            <LocalDebuggerEnvironment>PATH=%PATH%;$(Conan{name}BinaryDirectories)</LocalDebuggerEnvironment>
+            <LocalDebuggerEnvironment>PATH=%PATH%;$(Conan{name}BinaryDirectories)$(LocalDebuggerEnvironment)</LocalDebuggerEnvironment>
```
SSE4, line 41: no separator (semicolon)?

memsharded: Nop, that works, and it is the same pattern used for include-dirs and other properties, check the current implementation of ``MSBuildGenerator`` (I was initially surprised too)

SSE4 (approved):

PR #4750, opened 2019-03-15
Fix/test utils used in external projects

lasote:
Changelog: omit
Docs: omit

Moved a solitary test.
Kept in the distribution package only test tools, that is not the same as tests, and used by CPT.
Closes #4746

PR #8057, opened 2020-11-13
Conan test assets is required for CPT

uilianries:
Exclude conan.tests from Conan python package.

Changelog: Omit
Docs: Omit

Basically, it reverts https://github.com/conan-io/conan/pull/4750

CPT should copy and use what it needs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ dev_requirements = get_requires("conans/requirements_dev.txt")
 # The tests utils are used by conan-package-tools
 exclude_test_packages = ["conans.test.{}*".format(d)
                          for d in os.listdir(os.path.join(here, "conans/test"))
-                         if os.path.isdir(os.path.join(here, "conans/test", d)) and d != "utils"]
+                         if os.path.isdir(os.path.join(here, "conans/test", d)) and d not in ("utils", "assets")]
 
 
 def load_version():
--- a/setup.py
+++ b/setup.py
@@ -29,10 +29,6 @@ def get_requires(filename):
 project_requirements = get_requires("conans/requirements.txt")
 project_requirements.extend(get_requires("conans/requirements_server.txt"))
 dev_requirements = get_requires("conans/requirements_dev.txt")
-# The tests utils are used by conan-package-tools
-exclude_test_packages = ["conans.test.{}*".format(d)
-                         for d in os.listdir(os.path.join(here, "conans/test"))
-                         if os.path.isdir(os.path.join(here, "conans/test", d)) and d not in ("utils", "assets")]
 
 
 def load_version():
@@ -91,7 +87,7 @@ setup(
 
     # You can just specify the packages manually here if your project is
     # simple. Or you can use find_packages().
-    packages=find_packages(exclude=exclude_test_packages),
+    packages=find_packages(exclude=("conans.test*")),
 
     # Alternatively, if you want to distribute just a my_module.py, uncomment
     # this:


memsharded (commented):
I don't like the idea of packaging a bunch of VS .sln files, CMakeLists templates, .cpp and .h files in the distributable package of Conan itself, because someone is reusing the test assets.

IMHO we should tend to the opposite direction, to have a CPT as decoupled as possible from the Conan, otherwise, it becomes broken again and again after each Conan release. Having CPT reusing TestClient and TestServer might be convenient, but is probably a smell.

uilianries:
Okay, so I'll copy Conan test utils to CPT and we can remove all test stuff from Conan python package.

uilianries:
@memsharded new change: Exclude Conan tests

memsharded:
> Okay, so I'll copy Conan test utils to CPT and we can remove all test stuff from Conan python package.

Uhm, I am not sure about this, it still sounds like a smell to me. CPT should probably use real calls to ``conan`` instead of using things like TestClient and TestServer. I know that testing might be more inconvenient or difficult, but still. Maintaining a copy of TestClient and TestServer inside CPT might be a new pain.

Doing a quick review of tests, I find some concerning things, like tests directly attacking https://api.bintray.com/conan/conan-community/conan, and expecting a failure because of no-credentials.

The thing is that as we are starting with Conan 2.0, probably it is not worth a big investment in this issue right now. Asking for @jgsogo feedback.

jgsogo:
* Conan shouldn't distribute/package anything from its test framework. I totally approve this PR, but I'd go a little bit further and move the `test` directory to a `tests` directory in the root of this repository (it's probably the standard for Python).

 * CPT probably deserves some attention, I'm really surprised it breaks with new Conan versions and requires new releases every time Conan is released. I would expect it to break only if Conan changes the CLI or modifies some input file (profiles, settings,...), I mean, it should be compatible with any Conan 1.x version, exactly the same as hooks or the ConanCenterIndex pipeline library.

   I think we can (and should) simplify that testing a lot, and then the `TestClient` or any other class/util won't be needed. It requires time, do we want to pay it now or later? If not now, maybe it is easier to copy anything needed by CPT to its repository (probably lots of functions can be removed) and forget about breaking.

uilianries:
> Conan shouldn't distribute/package anything from its test framework. I totally approve this PR, but I'd go a little bit further and move the test directory to a tests directory in the root of this repository (it's probably the standard for Python).

:+1:


> CPT probably deserves some attention, I'm really surprised it breaks with new Conan versions and requires new releases every time Conan is released. I would expect it to break only if Conan changes the CLI or modifies some input file (profiles, settings,...), I mean, it should be compatible with any Conan 1.x version, exactly the same as hooks or the ConanCenterIndex pipeline library.


CPT requires a new release only for checking if new Conan version is supported or not. At some point we move from conan commands to conan api and as the API is not stable and changed sometimes it broke CPT few times. However, now I consider the API stable enough to remove that rule. We can add a new cron job or even trigger a CPT test and new release branch is created for Conan. 

> I think we can (and should) simplify that testing a lot, and then the TestClient or any other class/util won't be needed. It requires time, do we want to pay it now or later? If not now, maybe it is easier to copy anything needed by CPT to its repository (probably lots of functions can be removed) and forget about breaking.

CPT requires TestClient because it tests upload steps. As it upload a matrix of packages, we check if the matrix is correct. Could it be changed? Of course, but we re-used a well known form from Conan which works pretty fine.

uilianries:
No longer relevant. All tests are internalized in CPT now.

PR #8070, opened 2020-11-17
keep python files in packaging .pyc, .pyo, etc

### LINKED ISSUE #4961: Add configurable pattern for files to be excluded from the package (repackaging '.pyc' files)
Conan version: 1.15.0-dev

The way `gather_files` works - it discards all .pyc files. Discarding files is not always desired if one wants to keep the original package content as-is.

In my case the legacy package doesn't have .py files. While it's possible to "uncompyle" .pyc files I want to have more control and have an option to include everything.

```python
conans/model/manifest.py
   11 def discarded_file(filename):
   12     """
   13     # The __conan pattern is to be prepared for the future, in case we want to manage our
   14     own files that shouldn't be uploaded
   15     """
   16     return filename == ".DS_Store" or filename.endswith(".pyc") or \
   17            filename.endswith(".pyo") or filename == "__pycache__" or \
   18            filename.startswith("__conan")

```

- [x] I've specified the Conan version, operating system version and any tool that can be relevant.

memsharded:
Changelog: Feature: Introduce configuration ``general.keep_python_files`` to allow packaging of Python .pyc files.
Docs: https://github.com/conan-io/docs/pull/1942

Close https://github.com/conan-io/conan/issues/4961

### DIFF
--- a/conans/model/manifest.py
+++ b/conans/model/manifest.py
@@ -7,19 +7,23 @@ from conans.util.env_reader import get_env
 from conans.util.files import load, md5, md5sum, save, walk
 
 
-def discarded_file(filename):
+def discarded_file(filename, keep_python):
     """
     # The __conan pattern is to be prepared for the future, in case we want to manage our
     own files that shouldn't be uploaded
     """
-    return (filename == ".DS_Store" or filename.endswith(".pyc") or
-            filename.endswith(".pyo") or filename == "__pycache__" or
-            filename.startswith("__conan"))
+    if not keep_python:
+        return (filename == ".DS_Store" or filename.endswith(".pyc") or
+                filename.endswith(".pyo") or filename == "__pycache__" or
+                filename.startswith("__conan"))
+    else:
+        return filename == ".DS_Store"
 
 
 def gather_files(folder):
     file_dict = {}
     symlinks = {}
+    keep_python = get_env("CONAN_KEEP_PYTHON_FILES", False)
     for root, dirs, files in walk(folder):
         dirs[:] = [d for d in dirs if d != "__pycache__"]  # Avoid recursing pycache
         for d in dirs:
@@ -28,7 +32,7 @@ def gather_files(folder):
                 rel_path = abs_path[len(folder) + 1:].replace("\\", "/")
                 symlinks[rel_path] = os.readlink(abs_path)
         for f in files:
-            if discarded_file(f):
+            if discarded_file(f, keep_python):
                 continue
             abs_path = os.path.join(root, f)
             rel_path = abs_path[len(folder) + 1:].replace("\\", "/")
@@ -42,7 +46,6 @@ def gather_files(folder):
                                          "'general.skip_broken_symlinks_check' at the conan.conf "
                                          "file."
                                          % abs_path)
-
     return file_dict, symlinks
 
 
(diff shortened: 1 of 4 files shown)


conans/model/manifest.py
```diff
@@ -7,19 +7,23 @@
...
     """
-    return (filename == ".DS_Store" or filename.endswith(".pyc") or
-            filename.endswith(".pyo") or filename == "__pycache__" or
-            filename.startswith("__conan"))
+    if not keep_python:
+        return (filename == ".DS_Store" or filename.endswith(".pyc") or
+                filename.endswith(".pyo") or filename == "__pycache__" or
+                filename.startswith("__conan"))
+    else:
+        return filename == ".DS_Store"
```
jgsogo, line 20: Add here the `filename.startswith("__conan")`, I think that the motivation was to reserve a filename just in case we want to add something in that folder in the future.

conans/test/test_package_python_files.py
```diff
@@ -0,0 +1,69 @@
...
+        manifest = load(os.path.join(pkg_folder, "conanmanifest.txt"))
+        self.assertIn("myfile.pyc", manifest)
+        self.assertIn("myfile.pyo", manifest)
+        self.assertNotIn(".DS_Store", manifest)
+
+
+
+
+
+
```
jgsogo:
```suggestion
```

jgsogo (commented):
This new config modifies the **recipe revision**... but I'd say it should be the default for Conan 2.0.

My doubts are related to mixing clients with different configurations (and revisions enabled), how does Conan deal with a recipe containing a mismatch in the recipe revision?
 * which value will be stored in `metadata.json`? 
 * which value will be used in lockfiles?
 * will Conan raise because of some mismatch in the `manifest.txt`?

If there are no uncontrolled errors or we can print a friendly message, I would say that using `keep_python=True` can be ok for a company where they should already be controlling the `conan.conf` and sharing it among their developers.

memsharded:
> This new config modifies the recipe revision... but I'd say it should be the default for Conan 2.0.

Yes, users should be aware of this. As this is mainly for users that want to package .pyc files (curiously this has been a recurrent issue), they will be aware, the revision will change because they will be able to package something they couldn't before.

And yes, as much as possible, Conan shouldn't be messing with which files are packaged and which are not. The only exception I have found so far is the changing .DS_Store, which we clearly want to ignore in packages. So I agree this should be the default in Conan 2.0

> My doubts are related to mixing clients with different configurations (and revisions enabled), how does Conan deal with a recipe containing a mismatch in the recipe revision?

This will be an issue only for creating the same package again under different configurations, but not for consuming. And if they are creating the package, and they don't have the config enabled, it will fail. It is the responsibility of the users to keep all their clients in sync, in exactly the same way that ``default_package_id_mode``. If different clients have different config, they will not work nicely together. This should be clearly documented, indeed.

In any case it is important to note that this will only happen if you are actually packaging .pyc files, which might not happen that often.

> Add here the filename.startswith("__conan"), I think that the motivation was to reserve a filename just in case we want to add something in that folder in the future.

Yep, I clearly recall the motivation. But I think we really want to avoid putting anything else inside the package folder. This has been always very problematic, including the .tgz themselves, that we have recently removed from the folders. Do we really want to put "__conan" files inside the ``export``, ``export_source`` or the ``package_folder``? Do we want to keep it anyway, just in case?

jgsogo:
👍  I don't like the idea of adding `__conan` files to those folders either. I can't imagine what to use it for that cannot be stored somewhere else.

I would say this is safe as a feature. Do you think it is safe to add it here as a default for Conan v2 mode: https://github.com/conan-io/conan/blob/58dd8268d94f27f49a50b4f155c45c1c05476d44/conans/client/conf/__init__.py#L177

memsharded:
> I would say this is safe as a feature. Do you think it is safe to add it here as a default for Conan v2 mode:

Yes, I think it would be good to add it, but as discussed, lets re-think the approach with V2_MODE in general, not for this feature only.

jgsogo (approved):

czoido (approved):

PR #8072, opened 2020-11-17
fix warning with multiple import of transitive deps

### LINKED ISSUE #8065: [bug] msbuild generator: warning MSB4011 - conan_<dep>.props cannot be imported again
  * Operating System+version: Windows
  * Compiler+version: VS 2019
  * Conan version: 1.31.1

1. Package A requires B
2. App requires A and B
3. In project file for App, include `conan_deps.props` 

`conan_deps.props` :
```xml
[...]
		<Import Project="conan_A.props"/>
		<Import Project="conan_B.props"/>
	</ImportGroup>
[...]
```

`conan_B.props` :
```xml
[...]
	<ImportGroup Label="TransitiveDependencies">
		<Import Condition="'$(conan_B_props_imported)' != 'True'" Project="conan_B.props"/>
	</ImportGroup>
[...]
```

The problem is that `conan_A.props` already includes `conan_B.props`, so including it again in `conan_deps.props` causes the warning.
It should probably only be included under the condition `Condition="'$(conan_B_props_imported)' != 'True'"`, like it is done in `conan_B.props`.

memsharded:
Changelog: Fix: Avoid warning in ``msbuild`` generator importing multiple times the same .props file due to transitive dependencies.
Docs: Omit

Close https://github.com/conan-io/conan/issues/8065

#tags: slow

### DIFF
--- a/conans/client/generators/msbuild.py
+++ b/conans/client/generators/msbuild.py
@@ -30,7 +30,7 @@ class MSBuildGenerator(Generator):
     _dep_props = textwrap.dedent("""\
         <?xml version="1.0" encoding="utf-8"?>
         <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-          <ImportGroup Label="TransitiveDependencies">
+          <ImportGroup Label="ConanDependencies">
           </ImportGroup>
           <ImportGroup Label="Configurations">
           </ImportGroup>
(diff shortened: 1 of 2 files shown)


lieser:
Tested your change branch, and can confirm the warning is gone. Thanks for the quick fix.

czoido (approved):

PR #8077, opened 2020-11-18
adding --lockfile-node-id argument to differentiate same ref in lockfile

### LINKED ISSUE #7985: [bug] Private packages: Cannot build [...] because it is already locked in the input lockfile
Dear reader,

In several cases we use private dependencies, which themselves have other dependencies. Sometimes, this causes dependencies to appear in a dependency graph more than once, with different options set.

Unfortunately, this seems to fail when using lockfiles. The dependency is added to the lockfile more than once, with different package IDs, and building it the first time is successful. However, the second time it fails with the output "Cannot build [...] because it is already locked in the input lockfile".

See below for a simple reproduction.

Any ideas for workarounds would be welcome!

Thanks in advance!

Best regards,
spoofedex

  * Compiler+version: N/A
  * Conan version: 1.31.0
  * Python version: 3.7.3

Consider the following three packages:

private/conanfile.py:
```
from conans import ConanFile

class PrivateConan(ConanFile):
    name = "private"
    version = "1.0"

    options = {"a": [True, False]}
    default_options = {"a": False}
```

dep/conanfile.py:
```
from conans import ConanFile

class DepConan(ConanFile):
    name = "dep"
    version = "1.0"

    requires = ("private/1.0@test/test", "private"),

    def configure(self):
        self.options["private"].a = True
```

user/conanfile.py:
```
from conans import ConanFile

class UserConan(ConanFile):
    name = "user"
    version = "1.0"

    requires = ("private/1.0@test/test", "private"), "dep/1.0@test/test"

    def configure(self):
        self.options["private"].a = False
```

(note: "private" is used twice, with different options)

Export the three packages:
```
conan export private test/test
conan export dep test/test
conan export user test/test
```

Create a lockfile for "user":
```
conan lock create --reference user/1.0@test/test --lockfile-out lock --build missing --build outdated --build cascade
```
(note: the lockfile contains "private" twice, with two different package IDs)

Get the build order:
```
conan lock build-order lock
```
This gives the following output:
```
[[('private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96', '5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9', 'host', '2'), ('private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96', '69bc772c8ddf1b2fc5cc8733ac2b9f26be365797', 'host', '4')], [('dep/1.0@test/test#8bf90605c1fa4abe800599ce620b32f4', 'a29584d1915d2d1e548f8941ece31b66859ee98f', 'host', '3')], [('user/1.0@test/test#b148ee0df9fbb3e7a665521078605c5f', '987bbf43b04222dc2db0b1cf0030bd8f66b70241', 'host', '1')]]
```

(note: "private" is listed twice, with different package IDs)

Build private and update the lock (this succeeds):
```
conan install private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96 --lockfile lock --build private --lockfile-out lock-updated
conan lock update lock lock-updated
```

Build-order is now:
```
[[('private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96', '69bc772c8ddf1b2fc5cc8733ac2b9f26be365797', 'host', '4')], [('dep/1.0@test/test#8bf90605c1fa4abe800599ce620b32f4', 'a29584d1915d2d1e548f8941ece31b66859ee98f', 'host', '3')], [('user/1.0@test/test#b148ee0df9fbb3e7a665521078605c5f', '987bbf43b04222dc2db0b1cf0030bd8f66b70241', 'host', '1')]]
```

Build private (the second time):
```
conan install private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96 --lockfile lock --build private --lockfile-out lock-updated2
```

This fails with error:
`ERROR: Cannot build 'private/1.0@test/test#f6f07d91208890a80c3d4fb7b7b40c96' because it is already locked in the input lockfile`

Note: if it were possible to pass the package ID to "conan install", then I think this would be easy to resolve for us.

memsharded:
Changelog: Feature: Add ``--lockfile-node-id`` argument to ``conan install --lockfile`` so it can target different packages with same reference (different binary, this can happen with private requirements).
Docs: https://github.com/conan-io/docs/pull/1938

This will be an experimental feature, and there are other cases that might be missing, like ``export-pkg``, ``create``, etc.

Close https://github.com/conan-io/conan/issues/7985

#revisions: 1


solvingj (approved):
Extremely surprising how straightforward this implementation appears to have been.

conans/test/functional/graph_lock/graph_lock_ci_test.py
```diff
@@ -704,3 +704,61 @@ def test_version_ranges(self):
...
+                             'd7d6ac48b43e368b0a5ff79015acea49b758ffdf', 'host', '1']]]
+        else:
+            build_order = [[['private/0.1@',
+                             'd2560ba1787c188a1d7fabeb5f8e012ac53301bb', 'host', '3'],
+                            ['private/0.1@',
+                             '5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9', 'host', '4']],
+                           [['pkga/0.1@',
+                             '5b0fc4382d9c849ae3ef02a57b62b26ad5137990', 'host', '2']],
+                           [['pkgb/0.1@',
+                             'd7d6ac48b43e368b0a5ff79015acea49b758ffdf', 'host', '1']]]
```
jgsogo, line 758: Is there any difference between the two branches?  🕵️

conans/test/functional/graph_lock/graph_lock_ci_test.py
```diff
@@ -704,3 +704,61 @@ def test_version_ranges(self):
...
+
+        self.assertEqual(to_build, build_order)
+
+        for ref, _, _, node_id in build_order[0]:
+            client.run("install %s --build=%s --lockfile=conan.lock --lockfile-out=conan.lock "
+                       "--lockfile-id=%s" % (ref, ref, node_id))
```
jgsogo: Maybe check it's been built?

```suggestion
        for ref, pid, _, node_id in build_order[0]:
            client.run("install %s --build=%s --lockfile=conan.lock --lockfile-out=conan.lock "
                       "--lockfile-id=%s" % (ref, ref, node_id))
            self.assertIn('private/0.1:{} - Build'.format(pid))
```

jgsogo (commented):
Shall we use `lockfile-id` as the argument name or is it more the **ID for a node**? `node-id`? `lockfile-node-id`?

conans/test/functional/graph_lock/graph_lock_ci_test.py
memsharded: Yes, adding revisions would make the revisions_enabled branch to fail, because it doesn't include the revisions. I need to fix that.

jgsogo:
> Shall we use `lockfile-id` as the argument name or is it more the **ID for a node**? `node-id`? `lockfile-node-id`?

I wanted to mention @czoido in this comment (nomenclator!), but I forgot it.

czoido:
> > Shall we use `lockfile-id` as the argument name or is it more the **ID for a node**? `node-id`? `lockfile-node-id`?
> 
> I wanted to mention @czoido in this comment (nomenclator!), but I forgot it.

Yes, I would say maybe `node-id` looks more appropriate.

solvingj (approved):

czoido (approved):

PR #8082, opened 2020-11-19
Call conan_find_libraries_abs_path for CONAN_LIBS

### LINKED ISSUE #3162: conan_find_libraries_abs_path is called for CONAN_LIBS_DEBUG and CONAN_LIBS_RELEASE but not CONAN_LIBS
If CMAKE_BUILD_TYPE is given, neither CONAN_LIBS_DEBUG nor CONAN_LIBS_RELEASE are set, only CONAN_LIBS. BUT(!) conan_find_libraries_abs_path is not called for CONAN_LIBS, leaving the user the task of determining the absolute path. CONAN_LIBS should in my opinion also be made absolute.

Tested with conan 1.4.5 on Ubuntu 16.04, with cmake-conan.

- [x] I've read the [CONTRIBUTING guide](https://raw.githubusercontent.com/conan-io/conan/develop/.github/CONTRIBUTING.md).
- [x] I've specified the Conan version, operating system version and any tool that can be relevant.

czoido:
Changelog: Fix: Set `CONAN_LIBS` with absolute paths found instead of library names.
Docs: omit

When using the `cmake` generator using the global config (without targets) `CONAN_LIBS` takes the value for the library names and not the absolute path to the libraries, that absolute path is set if using `cmake_multi` generator. The absolute paths are also searched and set to the target if using targets.

Closes: https://github.com/conan-io/conan/issues/3162

#TAGS: slow
#REVISIONS: 1

### DIFF
--- a/conans/client/generators/cmake_common.py
+++ b/conans/client/generators/cmake_common.py
@@ -642,6 +642,8 @@ class CMakeCommonMacros:
 
             link_directories(${CONAN_LIB_DIRS})
 
+            conan_find_libraries_abs_path("${CONAN_LIBS}" "${CONAN_LIB_DIRS}"
+                                          CONAN_LIBS)
             conan_find_libraries_abs_path("${CONAN_LIBS_DEBUG}" "${CONAN_LIB_DIRS_DEBUG}"
                                           CONAN_LIBS_DEBUG)
             conan_find_libraries_abs_path("${CONAN_LIBS_RELEASE}" "${CONAN_LIB_DIRS_RELEASE}"


conans/client/generators/cmake_common.py
```diff
@@ -642,6 +642,8 @@ class CMakeCommonMacros:
 
             link_directories(${CONAN_LIB_DIRS})
 
+            conan_find_libraries_abs_path("${CONAN_LIBS}" "${CONAN_LIB_DIRS}"
```
memsharded, line 645: I am very concerned about this change. From experience I'd say that this will probably break users that are checking, iterating or doing something with the CONAN_LIBS variable.

memsharded (commented):
We are already trying to deprecate the ``cmake`` generator, and move to toolchains + find_package generator, I am not sure if the change is worth the risk. Was there some other more recent feedback that this is very needed, besides https://github.com/conan-io/conan/issues/3162, that has not been updated in many months?

conans/client/generators/cmake_common.py
czoido: Yes, I think that's right, It's probably not worth the risk, I did not take into account that maybe users are checking the current CONAN_LIBS variable and we could break them.

PR #8101, opened 2020-11-23
Feature/generate

memsharded:
Changelog: Feature: Deprecate experimental ``toolchain()`` in favor of more generic ``generate()`` method. Deprecate toolchains ``write_toolchain_files()`` to new ``generate()`` method.
Docs: https://github.com/conan-io/docs/pull/1940

#tags: slow


conans/client/conanfile/configure.py
```diff
@@ -47,6 +47,8 @@ def run_configure_method(conanfile, down_options, down_ref, ref):
 
 def _validate_fpic(conanfile):
     v2_mode = get_env(CONAN_V2_MODE_ENVVAR, False)
+    # FIXME: The toolchain() has dissapeared, but now it is integrated with generators.
+    # FIXME: Not sure this check should be here, this could raise before the graph is fully evaluated
     toolchain = hasattr(conanfile, "toolchain")
```
jgsogo, line 52: ?
```suggestion
    toolchain = hasattr(conanfile, "toolchain") or hasattr(conanfile, "generate")
```

conan/tools/microsoft/toolchain.py
```diff
@@ -27,6 +28,21 @@ def _name_condition(settings):
...
+               "******************************************************************\n"
+               "'write_toolchain_files()' has been deprecated and moved.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generate()' method instead.\n"
+               "********************************************************************\n"
+               "********************************************************************\n")
+        from conans.client.output import Color, ConanOutput
+        ConanOutput(self._conanfile.output._stream,
+                    color=self._conanfile.output._color).writeln(msg, front=Color.BRIGHT_RED)
+        warnings.warn(msg)
```
jgsogo, line 42: We might add the `category=DeprecationWarning` here, but probably it doesn't worth it (it requires changes in the test to capture it)

conan/tools/microsoft/toolchain.py
```diff
@@ -27,6 +28,21 @@ def _name_condition(settings):
...
+        msg = ("\n*****************************************************************\n"
+               "******************************************************************\n"
+               "'write_toolchain_files()' has been deprecated and moved.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generate()' method instead.\n"
+               "********************************************************************\n"
+               "********************************************************************\n")
+        from conans.client.output import Color, ConanOutput
+        ConanOutput(self._conanfile.output._stream,
+                    color=self._conanfile.output._color).writeln(msg, front=Color.BRIGHT_RED)
```
jgsogo, line 41: Why not...?  (It is ok because we will remove all of this in the next release)

```python
self._conanfile.output.warn(msg)
```

conans/client/generators/__init__.py
```diff
@@ -83,10 +84,37 @@ def __contains__(self, name):
...
+            return CMakeToolchain
+        elif generator_name == "MakeToolchain":
+            from conan.tools.gnu import MakeToolchain
+            return MakeToolchain
+        elif generator_name == "MSBuildToolchain":
+            from conan.tools.microsoft import MSBuildToolchain
+            return MSBuildToolchain
+        elif generator_name == "MesonToolchain":
+            from conan.tools.meson import MesonToolchain
+            return MesonToolchain
```
jgsogo, line 99: See [line 121 in this file](https://github.com/conan-io/conan/pull/8014/files#diff-e235fa62562f3e2f47330538b120750466913d3d1733d75c6eee9a06906dc95fR121), we need to add these new generators to the list, otherwise the error message won't list all of them:

```python
                raise ConanException("Invalid generator '%s'. Available types: %s" %
                                     (generator_name, ", ".join(self.available)))
```

conans/client/generators/__init__.py
```diff
@@ -125,6 +153,15 @@ def write_generators(self, conanfile, path, output):
 
 def write_toolchain(conanfile, path, output):
     if hasattr(conanfile, "toolchain"):
+        msg = ("\n*****************************************************************\n"
+               "******************************************************************\n"
+               "The 'toolchain' attribute or method has been deprecated.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generators = ClassName' or 'generate()' method instead.\n"
```
jgsogo:
```suggestion
               "Use 'generators = \"ClassName\"' or 'generate()' method instead.\n"
```

conans/client/generators/__init__.py
```diff
@@ -138,6 +175,15 @@ def write_toolchain(conanfile, path, output):
                 raise ConanException("Unknown toolchain '%s'" % conanfile.toolchain)
             tc = toolchain(conanfile)
             with chdir(path):
-                tc.write_toolchain_files()
+                tc.generate()
+
+        # TODO: Lets discuss what to do with the environment
+
+    if hasattr(conanfile, "generate"):
+        assert callable(conanfile.generate), "generate should be a method, not an attribute"
```
jgsogo: We are not doing this for other methods. The Python error itself and the `conanfile_exception_formatter` should be enough.

```suggestion
```

conans/test/functional/toolchain/test_txt_cmdline.py
```diff
@@ -0,0 +1,50 @@
...
+        toolchain = client.load("conan_meson_native.ini")
+        self.assertIn("[project options]", toolchain)
+
+    def test_command_line(self):
+        conanfile = textwrap.dedent("""
+            [generators]
+            CMakeToolchain
+            MesonToolchain
+            MakeToolchain
+            MSBuildToolchain
```
jgsogo: If we are testing the command line `-g`, I think we should remove them from the `conanfile.txt`, right?

```suggestion
```

jgsogo (changes requested):

conan/tools/microsoft/toolchain.py
memsharded: Because it only prints in RED color the package name and version, so the message is not that visible, as it is printed in regular white, together with a wall of text. I wanted to make it super-visible and clear.

This is something that we might want to change for warnings and errors in our output, I'd say the UX is not the best (easy to miss important warnings or messages)

conans/test/functional/toolchain/test_txt_cmdline.py
memsharded: Yep, a mistake due to my laptop suspending, fixing it.

conan/tools/microsoft/toolchain.py
jgsogo: It prints everything in yellow/red, it is the same call (with a prefix and output to stderr):

```python
    def warn(self, data):
        self.writeln("WARN: {}".format(data), Color.BRIGHT_YELLOW, error=True)

    def error(self, data):
        self.writeln("ERROR: {}".format(data), Color.BRIGHT_RED, error=True)
```

conan/tools/microsoft/toolchain.py
memsharded: Yes, if it is going to be removed soon, probably not worth the effort of changing tests forth and back.

conans/client/conanfile/configure.py
memsharded: Problem here is that ``toolchain`` was both an attribute and a method. If it is defined, it was activating this behavior. Now, we could have the ``generate()`` method, that would fire it, but defining and using ``generators = "CMakeToolchain"`` will not. So it seemed a bit inconsistent, and I added the FIXMEs to reconsider how to approach this (it seems that this could be moved to ``validate()`` stage, when things are evaluated after the graph is computed.

conan/tools/microsoft/toolchain.py
memsharded: ``conanfile.output`` is of type ``ScopedOutput`` and it will only print the scope in color, not the whole text.

conan/tools/cmake/base.py
```diff
@@ -162,6 +163,21 @@ def _get_template_context_data(self):
...
+               "'write_toolchain_files()' has been deprecated and moved.\n"
+               "It will be removed in next Conan release.\n"
+               "Use 'generate()' method instead.\n"
+               "********************************************************************\n"
+               "********************************************************************\n")
+        from conans.client.output import Color, ConanOutput
+        ConanOutput(self._conanfile.output._stream,
+                    color=self._conanfile.output._color).writeln(msg, front=Color.BRIGHT_RED)
+        warnings.warn(msg)
+        return self.generate()
```
jgsogo:
```suggestion
       self.generate()
```

jgsogo (approved):
Good! Just a typo

PR #8106, opened 2020-11-25
Feature: update conan new to latest guidelines

SSE4:
update several minor things in `conan new` command:
- add `fPIC` option handling
- ~~add `homepage` attribute~~
- add `header_only` to `package_id`
- update to `cmake_minimum_required(VERSION 3.1)`

this makes it easier to create package skeletons - fewer things need to be edited

Changelog: Feature: Update `conan new` to latest guidelines.
Docs: omit

#tags: slow

### DIFF
--- a/conans/client/cmd/new.py
+++ b/conans/client/cmd/new.py
@@ -186,7 +186,7 @@ class {package_name}TestConan(ConanFile):
             self.run(".%sexample" % os.sep)
 """
 
-test_cmake = """cmake_minimum_required(VERSION 2.8.12)
+test_cmake = """cmake_minimum_required(VERSION 3.1)
 project(PackageTest CXX)
 
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
@@ -202,7 +202,7 @@ target_link_libraries(example ${CONAN_LIBS})
 #          COMMAND example)
 """
 
-test_cmake_pure_c = """cmake_minimum_required(VERSION 2.8.12)
+test_cmake_pure_c = """cmake_minimum_required(VERSION 3.1)
 project(PackageTest C)
 
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
@@ -261,7 +261,7 @@ void {name}(){{
 }}
 """
 
-cmake_pure_c = """cmake_minimum_required(VERSION 2.8)
+cmake_pure_c = """cmake_minimum_required(VERSION 3.1)
 project({name} C)
 
 include(${{CMAKE_BINARY_DIR}}/conanbuildinfo.cmake)
@@ -270,7 +270,7 @@ conan_basic_setup()
 add_library({name} {name}.c)
 """
 
-cmake = """cmake_minimum_required(VERSION 2.8)
+cmake = """cmake_minimum_required(VERSION 3.1)
 project({name} CXX)
 
 include(${{CMAKE_BINARY_DIR}}/conanbuildinfo.cmake)
--- a/conans/client/cmd/new.py
+++ b/conans/client/cmd/new.py
@@ -22,9 +22,13 @@ class {package_name}Conan(ConanFile):
     description = "<Description of {package_name} here>"
     topics = ("<Put some tag here>", "<here>", "<and here>")
     settings = "os", "compiler", "build_type", "arch"
-    options = {{"shared": [True, False]}}
-    default_options = {{"shared": False}}
+    options = {{"shared": [True, False], "fPIC": [True, False]}}
+    default_options = {{"shared": False, "fPIC": True}}
     generators = "cmake"
+
+    def config_options(self):
+        if self.settings.os == "Windows":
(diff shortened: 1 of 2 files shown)


memsharded (approved):

n-bes (changes requested):
Could you update ```test_cmake``` template? By default it is not allowed by hook ```KB-H048```:

```
[HOOK - conan-center.py] pre_export(): ERROR: [CMAKE VERSION REQUIRED (KB-H048)] The test_package/CMakeLists.txt requires CMake 3.1 at least. Update to 'cmake_minimum_required(VERSION 3.1)'. (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H048)
```

SSE4:
> Could you update `test_cmake` template? By default it is not allowed by hook `KB-H048`:
> 
> ```
> [HOOK - conan-center.py] pre_export(): ERROR: [CMAKE VERSION REQUIRED (KB-H048)] The test_package/CMakeLists.txt requires CMake 3.1 at least. Update to 'cmake_minimum_required(VERSION 3.1)'. (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H048)
> ```

np, updated

SSE4:
/cc @ericLemanissier @Croydon @madebr @uilianries @SpaceIm @prince-chrismc 
lemme know if there is something else useful to update in that shot

conans/client/cmd/new.py
```diff
@@ -91,11 +95,15 @@ class {package_name}Conan(ConanFile):
     description = "<Description of {package_name} here>"
     topics = ("<Put some tag here>", "<here>", "<and here>")
     settings = "os", "compiler", "build_type", "arch"
-    options = {{"shared": [True, False]}}
-    default_options = {{"shared": False}}
+    options = {{"shared": [True, False], "fPIC": [True, False]}}
+    default_options = {{"shared": False, "fPIC": True}}
```
prince-chrismc, line 99: Seems a little odd to have compilation settings if it's header only 🤔 

base on the example CCI points to https://github.com/conan-io/conan-center-index/blob/78363b7d99786ea0060bfe2a02ded90b43bf2d4a/recipes/rapidjson/all/conanfile.py

we should have

```suggestion
    no_copy_source = True
```

conans/client/cmd/new.py
```diff
@@ -22,10 +22,14 @@ class {package_name}Conan(ConanFile):
...
     generators = "cmake"
 
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
```
prince-chrismc, line 31: I am fairly certain this is also expected during review

```suggestion
    def config_options(self):
        if self.settings.os == "Windows":
            del self.options.fPIC

    def configure(self):
        if self.options.shared:
            del self.options.fPIC
```

prince-chrismc (commented):
A few suggestions because of comments that are common

conans/client/cmd/new.py
madebr: I think the `conanfile_header` temple below addresses header_only packages.
This one addresses building from current folder (=conanfile bundled with sources).

conans/client/cmd/new.py
memsharded: Not completely sure, many recipes do not implement it yet. I would say wait until this is discussed a bit more, seems that some Conan built-in helpers would be better.

SSE4: agree
```
    def config_options(self):
            del self.options.fPIC
```
this code has 361 occurrences on CCI
```
    def configure(self):
        if self.options.shared:
            del self.options.fPIC
```
this one is only 264.
so, probably, it might be too early to enforce that, as it's not an established practice on CCI.
earlier we also had proposal for so called `shared_library_package_id` to be used in recipes, I don't remember what was a conclusion.
we need to talk about these two IMO before we start to recommend them.

conans/client/cmd/new.py
```diff
@@ -145,6 +153,9 @@ def source(self):
 
     def package(self):
         self.copy("*.h", "include")
+
+    def package_id(self):
+        self.info.header_only()
```
danimtb, line 158: I understand this is more explicit than just a recipe without settings, but we are adding more code here that is unneeded. If this is a pattern followed in CCI for example, then I am ok with it. Maybe we should consider it for conan 2.0 (enforce settings always?)

danimtb (approved):

conans/client/cmd/new.py
SSE4: I guess it might be error prone without explicit, e.g. if you add some settings and options to the recipe after running `conan new`, it's easy to forget about this method.
same if you decide to add more requirements, as `header_only` now contains:
```
self.requires.clear()
```
I believe the same package id is expected to be generated regardless of requires, which is not the case in the current template.

jgsogo (approved):
LGTM

PR #8107, opened 2020-11-25
Feature: temporary acquire write permissions in replace_in_file

### LINKED ISSUE #8043: [bug] cmake.patch_config_paths() fails with PermissionError
Hi,

I'm building OpenCV 4.3.0 for aarch64.
It's cross build.
It's modified version of https://github.com/conan-community/conan-opencv
Everything compiles, then installs but this last patching step during packaging fails:
506     def package(self):
507         self.copy("LICENSE", dst="licenses", src=self._source_subfolder)
508         cmake = self._configure_cmake()
509         cmake.install()
510         cmake.patch_config_paths()

It fails with:
ERROR: opencv/4.3.0: Error in package() method, line 510
	cmake.patch_config_paths()
	PermissionError: [Errno 13] Permission denied: '/home/vagrant/.conan/data/opencv/4.3.0/_/_/build/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/build_subfolder/CPackSourceConfig.cmake'

I listed properties of that file and seems problem is that it is readonly:
vagrant@ubuntu-bionic:~/opencv_pkg$ ls -laF /home/vagrant/.conan/data/opencv/4.3.0/_/_/build/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/build_subfolder/CPackSourceConfig.cmake
-r--r--r-- 1 vagrant vagrant 10215 Nov 11 04:22 /home/vagrant/.conan/data/opencv/4.3.0/_/_/build/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/build_subfolder/CPackSourceConfig.cmake

I also built this thing for armv7hf target with different toolchain and it had no issues.
Seems it's caused by some difference in configuration, but I'm not sure which.
In any case, this seems like a bug. 
Seems CMake generates CPackSourceConfig.cmake as readonly and patching then fails.
So, either patching tools shall be able to deal with readonly case or there shall be a way to control how
permissions of CPackSourceConfig.cmake are set. However, I did not find how to change those.
If somebody can point it out I will gladly do that and dismiss this issue.
If there is no straight way to change permissions, then the fix would be to account permissions during patching.

  * Operating System+version: Ubuntu 18.04 (under vagrant)
  * Compiler+version: yokto sdk aarch64 gcc 9.2 
  * Conan version: 1.29.2
  * Python version: Python 3.6.9
cmake version 3.10.2
I also tried cmake 3.18.4 and got same output.

I'm not sure how to formulate those. It's just too big. 
See previous paragraphs for major parts description.

Tail of conan create output:

...
-- Installing: /home/vagrant/.conan/data/opencv/4.3.0/_/_/package/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/res/lbpcascades/lbpcascade_profileface.xml
-- Installing: /home/vagrant/.conan/data/opencv/4.3.0/_/_/package/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/res/lbpcascades/lbpcascade_silverware.xml
ERROR: opencv/4.3.0: Error in package() method, line 510
	cmake.patch_config_paths()
	PermissionError: [Errno 13] Permission denied: '/home/vagrant/.conan/data/opencv/4.3.0/_/_/build/8bf6f6dfbecc99ccf819a01821ecfa373bdda90d/build_subfolder/CPackSourceConfig.cmake'
+ return 1

Let me know if you see a possibility to fix. 
I guess I shall be able to workaround in some ugly way, but having a fix would be nicer.

Best regards,

Yuriy

SSE4:
closes: #8043

Changelog: Feature: Temporary acquire write permissions in `replace_in_file`.
Docs: omit

### DIFF
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -245,8 +245,7 @@ def replace_in_file(file_path, search, replace, strict=True, output=None, encodi
 
     encoding_in = encoding or "auto"
     encoding_out = encoding or "utf-8"
-    with add_permissions(file_path, stat.S_IREAD):
-        content = load(file_path, encoding=encoding_in)
+    content = load(file_path, encoding=encoding_in)
     if -1 == content.find(search):
         _manage_text_not_found(search, file_path, strict, "replace_in_file", output=output)
     content = content.replace(search, replace)
@@ -268,8 +267,7 @@ def replace_path_in_file(file_path, search, replace, strict=True, windows_paths=
 
     encoding_in = encoding or "auto"
     encoding_out = encoding or "utf-8"
-    with add_permissions(file_path, stat.S_IREAD):
-        content = load(file_path, encoding=encoding_in)
+    content = load(file_path, encoding=encoding_in)
     normalized_content = normalized_text(content)
     normalized_search = normalized_text(search)
     index = normalized_content.find(normalized_search)
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -226,7 +226,7 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
 
 
 @contextmanager
-def add_permissions(file_path, permissions):
+def _add_permissions(file_path, permissions):
     if os.path.isfile(file_path):
         saved_permissions = os.stat(file_path).st_mode
         if saved_permissions & permissions == permissions:
@@ -250,7 +250,7 @@ def replace_in_file(file_path, search, replace, strict=True, output=None, encodi
         _manage_text_not_found(search, file_path, strict, "replace_in_file", output=output)
     content = content.replace(search, replace)
     content = content.encode(encoding_out)
-    with add_permissions(file_path, stat.S_IWRITE):
+    with _add_permissions(file_path, stat.S_IWRITE):
         save(file_path, content, only_if_modified=False, encoding=encoding_out)
 
 
@@ -281,7 +281,7 @@ def replace_path_in_file(file_path, search, replace, strict=True, windows_paths=
         index = normalized_content.find(normalized_search)
 
     content = content.encode(encoding_out)
-    with add_permissions(file_path, stat.S_IWRITE):
+    with _add_permissions(file_path, stat.S_IWRITE):
         save(file_path, content, only_if_modified=False, encoding=encoding_out)
 
     return True
@@ -295,7 +295,7 @@ def replace_prefix_in_pc_file(pc_file, new_prefix):
             lines.append('prefix=%s' % new_prefix)
         else:
             lines.append(line)
-    with add_permissions(pc_file, stat.S_IWRITE):
+    with _add_permissions(pc_file, stat.S_IWRITE):
         save(pc_file, "\n".join(lines))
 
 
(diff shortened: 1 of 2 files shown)


conans/client/tools/files.py
```diff
@@ -224,17 +225,34 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
...
+    else:
+        yield
+
 def replace_in_file(file_path, search, replace, strict=True, output=None, encoding=None):
     output = default_output(output, 'conans.client.tools.files.replace_in_file')
 
     encoding_in = encoding or "auto"
     encoding_out = encoding or "utf-8"
-    content = load(file_path, encoding=encoding_in)
+    with add_permissions(file_path, stat.S_IREAD):
```
memsharded: Probably the READ case is not a real issue? If some package doesn't have read permission, something will be surely broken? What is a real use case of a package file without READ permission?

conans/client/tools/files.py
```diff
@@ -224,17 +225,34 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
...
+        content = load(file_path, encoding=encoding_in)
     if -1 == content.find(search):
         _manage_text_not_found(search, file_path, strict, "replace_in_file", output=output)
     content = content.replace(search, replace)
     content = content.encode(encoding_out)
-    save(file_path, content, only_if_modified=False, encoding=encoding_out)
+    with add_permissions(file_path, stat.S_IWRITE):
```
memsharded: I am not fully convinced about this approach, something seems not completely right, like this module shouldn't be responsible of doing this. If the user do a ``save()`` it will fail due to permissions, but ``replace_in_file()`` should succeed?

Asking @jgsogo for a second opinion.

conans/client/tools/files.py
SSE4: added mostly for symmetry, but I don't think it happens in reality, so maybe removed

conans/client/tools/files.py
SSE4: in the first attempt I put it to the `save` method, but it's used extensively in many other tools and internal functions, as result it caused many different tests to fail (e.g. `test_export_read_only`). so I decided to isolate it, and limit changes just to `replace_in_file`/`replace_path_in_file` family of functions.

conans/client/tools/files.py
memsharded: Yes, lets remove it, unless until this is reported as a real issue.

conans/client/tools/files.py
```diff
@@ -224,6 +225,21 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
         return False
 
 
+@contextmanager
+def add_permissions(file_path, permissions):
```
memsharded: Make it private, I don't want to expose more tools to users (even if not documented) that we don't have a strong evidence that they are very necessary an demanded by users

memsharded (approved):
Please do those minor changes, if done tomorrow early, we could try to merge this for 1.32.

conans/client/tools/files.py
SSE4: removed already :)

conans/client/tools/files.py
SSE4: changed

czoido (approved):

PR #8114, opened 2020-11-26
Fix linkage to a same global target of different package components in CMake generators

danimtb:
Changelog: BugFix: Fix linkage to a same global target of different package components in `cmake_find_package/_multi` generators.
Docs: https://github.com/conan-io/docs/pull/1946

- [x] Refer to the issue that supports this Pull Request: closes https://github.com/conan-io/conan/issues/7889


conans/test/functional/generators/components/cmake_find_package_multi_test.py
```diff
@@ -804,3 +804,146 @@ def build(self):
...
+            cmake_minimum_required(VERSION 3.1)
+
+            include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+            conan_basic_setup()
+
+            find_package(expected)
+            find_package(variant)
+
+            add_library(middle middle.cpp)
+            target_link_libraries(middle nonstd::expected nonstd::variant)
```
jgsogo: I think here we need to test how the behavior changes.  Before if using `nonstd::nonstd` it should fail to link, with the changes in this PR it will work (both mini-targets are available).

conans/test/functional/generators/components/cmake_find_package_multi_test.py
```diff
@@ -804,3 +804,146 @@ def build(self):
...
+            #include "middle.h"
+
+            int main() {
+                middle();
+            }
+            """)
+        client.save({"conanfile.py": conanfile,
+                     "src/CMakeLists.txt": cmakelists,
+                     "src/main.cpp": main_cpp}, clean_first=True)
+        client.run("create . consumer/1.0@")
```
jgsogo, line 939: Not sure if this last `consumer` adds anything to the test.

conans/test/functional/generators/components/cmake_find_package_multi_test.py
danimtb: Do you mean to split this into two PRs? one adding a test failing and another one with the fix and the change in the test?

conans/test/functional/generators/components/cmake_find_package_test.py
```diff
@@ -766,3 +766,146 @@ def build(self):
...
+            include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+            conan_basic_setup()
+
+            find_package(expected)
+            find_package(variant)
+
+            add_library(middle middle.cpp)
+            target_link_libraries(middle nonstd::expected nonstd::variant)
+            """)
+        middle_h = textwrap.dedent("""
```
memsharded: Use the ``gen_function_h()`` and ``gen_function_cpp()`` test helpers.

conans/test/functional/generators/components/cmake_find_package_test.py
```diff
@@ -766,3 +766,146 @@ def build(self):
...
+                settings = "os", "compiler", "build_type", "arch"
+                exports_sources = "src/*"
+                requires = "middle/1.0"
+
+                def build(self):
+                    cmake = CMake(self)
+                    cmake.configure(source_folder="src")
+                    cmake.build()
+                    bin_path = os.path.join("bin", "main")
+                    self.run(bin_path, run_environment=True)
```
memsharded: ``run_environment`` needed?

conans/test/functional/generators/components/cmake_find_package_multi_test.py
memsharded: It means that a test with ``nonstd::nonstd`` seems good, to prove it works now.
Probably no need another full test, just adding a new target, using the same file, would be enough.

memsharded (approved):
It is looking good, maybe improving the tests to make sure that the APPEND work, and ``nonstd::nonstd`` is now the accumulated of all.

conans/test/functional/generators/components/cmake_find_package_multi_test.py
danimtb: It is reproducing the case exposed in the issue more clearly, maybe a check to verify that the target link libraries are not set with multiple `nonstd::nonstd` targets is the part missing. But I agree that we can reduce the tests if we don't want to test this one

conans/test/functional/generators/components/cmake_find_package_test.py
danimtb: not needed, removed

conans/test/functional/generators/components/cmake_find_package_multi_test.py
jgsogo: That's it, we need to test that using `nonstd::nonstd`, it works (Conan aggregates targets coming from different packages). We already know it works if we use explicitly those two mini-targets.

conans/test/functional/generators/components/cmake_find_package_multi_test.py
jgsogo: Ok, this is useful to check we are not adding the same target several times.

conans/test/functional/generators/components/cmake_find_package_multi_test.py
jgsogo: ...but, it is true this is linking statically, so the consumer needs to link with both (expected and variant), so this test would fail without this fix.

conans/test/functional/generators/components/cmake_find_package_test.py
danimtb: used only `gen_function_h()` as the cpp file if different

conans/test/functional/generators/components/cmake_find_package_multi_test.py
danimtb: anyway, I have changed the targets for the ``nonstd::nonstd`` to make it more evident. thanks!

jgsogo (approved):

jgsogo (approved):
btw, I would modify the changelog, IMO this is more like a feature.

(I'm not merging because `develop` is 🔴  for several builds now and I don't want to add more noise if something is really broken)

memsharded (approved):

PR #8115, opened 2020-11-26
[refact] Use downloaders via a 'run_downloader' function

jgsogo:
Changelog: omit
Docs: omit

Other downloaders, like the ArtifactorySourcesBackup, will only need to be added in a single place.

A little bit more work can be done related to checksums (trying it in a following PR)


conans/client/downloaders/file_downloader.py
```diff
@@ -6,8 +6,7 @@
 import six
 
 from conans.client.rest import response_to_str
-from conans.errors import AuthenticationException, ConanConnectionError, ConanException, \
-    NotFoundException, ForbiddenException, RequestErrorException
+from conans.errors import ConanException, NotFoundException, AuthenticationException, ForbiddenException, ConanConnectionError, RequestErrorException
```
memsharded: Your editorconfig seems not working? This lenght is too much.

memsharded (approved):
Nice refactor, looking good.

conans/client/downloaders/file_downloader.py
jgsogo: It didn't apply the proper formatting when moving the class (right-click > refactor > move)... 🤷

czoido (approved):

PR #8117, opened 2020-11-26
possible fix to cmake.install() failing in test_package

### LINKED ISSUE #8080: [bug] Can't use cmake.install() in test_package/conanfile.py
When executing "conan create", and the conan package includes a test_package with a call to cmake.install() in its conanfile.py's build() method, the "conan create" step fails with following error message:

 ERROR: test/0.0.1@user/testing (test package): Error in build() method, line 18
        cmake.install()
        TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType

This issue makes it impossible to use cmake.install() in test packages.

### Environment Details
  * Compiler+version: MSVC 19.16.27038.0
  * Conan version: 1.28.2
  * Python version: 3.8.6

### Steps to reproduce
Run conan new test/0.0.1 -t
Add cmake.install() to test_package/conanfile.py -> build(self):
```
    def build(self):
        cmake = CMake(self)
        # Current dir is "test_package/build/<build_id>" and CMakeLists.txt is
        # in "test_package"
        cmake.configure()
        cmake.build()
        cmake.install()
```
Run conan create . test/0.0.1@user/testing 

**Note:** The example above was chosen for the sake of simplicity. It is generally expected that cmake.install() would fail in this example because the install target does not exist. However the above mentioned error is also reproducible in more complex examples that actually define an "install" target.

### Workaround
The issue can be worked around by manually setting the required vars:
```
    def build(self):
        cmake = CMake(self)
        # Hack: For some reason when running "conan test" the package_folder
        # and CMAKE_INSTALL_PREFIX are not defined, so we need to set them
        # Manually here. Conan bug?
        cmake._conanfile.package_folder = "package"
        cmake.definitions["CMAKE_INSTALL_PREFIX"] = cmake._conanfile.package_folder
        cmake.configure()
        cmake.build()
        cmake.install()
```

memsharded:
Changelog: Fix: Define ``package_folder`` in the ``test_package`` folder (defaulting to "package"), so the test recipe can execute ``cmake.install()`` in its ``build()`` method.
Docs: Omit

Close https://github.com/conan-io/conan/issues/8080

### DIFF
--- a/conans/client/cmd/test.py
+++ b/conans/client/cmd/test.py
@@ -37,9 +37,7 @@ def install_build_and_test(app, conanfile_abs_path, reference, graph_info,
                      keep_build=keep_build,
                      recorder=recorder)
         cmd_build(app, conanfile_abs_path, base_folder, test_build_folder,
-                  # FIXME: Probably we need a better (or user-defined) layout definition
-                  #package_folder=os.path.join(test_build_folder, "package"),
-                  package_folder=None,
+                  package_folder=os.path.join(test_build_folder, "package"),
                   install_folder=test_build_folder, test=reference)
     finally:
         if delete_after_build:
--- a/conans/client/cmd/test.py
+++ b/conans/client/cmd/test.py
@@ -37,7 +37,9 @@ def install_build_and_test(app, conanfile_abs_path, reference, graph_info,
                      keep_build=keep_build,
                      recorder=recorder)
         cmd_build(app, conanfile_abs_path, base_folder, test_build_folder,
-                  package_folder=test_build_folder,
+                  # FIXME: Probably we need a better (or user-defined) layout definition
+                  #package_folder=os.path.join(test_build_folder, "package"),
+                  package_folder=None,
                   install_folder=test_build_folder, test=reference)
     finally:
         if delete_after_build:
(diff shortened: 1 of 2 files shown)


conans/client/cmd/test.py
```diff
@@ -36,7 +36,8 @@ def install_build_and_test(app, conanfile_abs_path, reference, graph_info,
                      manifest_interactive=manifest_interactive,
                      keep_build=keep_build,
                      recorder=recorder)
-        cmd_build(app, conanfile_abs_path, base_folder, test_build_folder, package_folder=None,
+        cmd_build(app, conanfile_abs_path, base_folder, test_build_folder,
+                  package_folder=os.path.join(test_build_folder, "package"),
```
uilianries, line 40: I would use `conans.paths.PACKAGES_FOLDER` instead:

```python
from conans.paths import PACKAGES_FOLDER

...

cmd_build(app, conanfile_abs_path, base_folder, test_build_folder,
                  package_folder=os.path.join(test_build_folder, PACKAGES_FOLDER),
```

memsharded: I think it is better to leave it explicit and independent, that is an internal path definition for the Conan cache. We might want to change it anytime, to something like "pkg", but we don't want to break what happens in user space.

uilianries (approved):

PR #8124, opened 2020-11-29
make variables defined in cmake tolchain cache ones

### LINKED ISSUE #7832: [bug] CMakeToolchain now requires double configure to set cache variables
https://github.com/conan-io/conan/commit/a77b97817a73fbdd2fbeb5f7038ad400aafd1803 switched from using cached variables to normal cmake variables. This means that cmake configuration has to be run twice to override standard configurable settings. This is because on the first processing, the cache variable does not exist in the cache, so when the `set` function call that initializes the cache variable is reached, it overrides the variable that is set by toolchain. On a second configuration, the same `set` function call is a no-op, since the cache variable has already been initialized.

  * Conan version: 1.30.0
  * Python version: 3.6, 3.7, 3.8

CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.12)
project(my_project)
set(my_configurable_string "default value" CACHE STRING "my configurable string")
```

conanfile.py
```python
class MyClass(ConanFile):
    ...
    def toolchain(self):
        toolchain = CMakeToolchain(self)
        toolchain.variables["my_configurable_string"] = "my value"
        toolchain.write_toolchain_files()
```

memsharded:
Changelog: Fix: Make variables defined in ``CMakeToolchain`` cache variables, so they can define directly values defined in ``CMakeLists.txt``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/7832

### DIFF
--- a/conan/tools/cmake/base.py
+++ b/conan/tools/cmake/base.py
@@ -44,9 +44,11 @@ class CMakeToolchainBase(object):
                                           '$<IF:$<CONFIG:' + conf + '>,"' + value|string + '",' %}
                     {%- if loop.last %}{% set genexpr.str = genexpr.str + '""' -%}{%- endif -%}
                 {%- endfor -%}
-                {% for i in range(values|count) %}{%- set genexpr.str = genexpr.str + '>' %}{%- endfor -%}
+                {% for i in range(values|count) %}{%- set genexpr.str = genexpr.str + '>' %}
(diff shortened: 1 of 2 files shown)


jgsogo (approved):

jgsogo:
Waiting for a green build in `develop` before merging

PR #8125, opened 2020-11-30
Feature/qbs build helper

### LINKED ISSUE #8079: [feature] Qbs build helper
It would be nice to have a Qbs build helper. Since Qbs has its own profiles it is not possible to just run the qbs command to build an application.  
When building a conan package with qbs first a qbs profile has to be created with the conan profile settings. Adding this manually to every conanfile or build pipeline definition is tedious. A Qbs build helper could setup qbs profiles automatically.

Psy-Kai:
Changelog: Feature: Add `QbsToolchain` and a Qbs build helper class (currently working for Mcus, not for Android or iOS).
Docs: https://github.com/conan-io/docs/pull/1978

Closes #8079 


conans/client/build/qbs.py
```diff
@@ -0,0 +1,103 @@
+import os
+
+from conans import tools
+from conans.errors import ConanException
+
+
+class QbsException(ConanException):
```
memsharded: Do not redefine a class with same name again, this is problematic.
Actually, it is not necessary to define a QbsException, all the other build systems do not have a custom exception for this.

conans/client/build/qbs.py
```diff
@@ -0,0 +1,103 @@
...
+                b = 'true'
+            else:
+                b = 'false'
+            command_list.append('%s:%s' % (key, b))
+        else:
+            command_list.append('%s:%s' % (key, value))
+    return command_list
+
+
+class Qbs(object):
```
memsharded: Move it to ``conan.tools.qbs`` too

conan/tools/qbs/generic.py
```diff
@@ -0,0 +1,296 @@
...
+            raise QbsException('Could not parse LDFLAGS')
+        self._parse = self._add_linker_flag
+
+    def _add_linker_flag(self, token):
+        self.linker_flags.append(token)
+        self._parse = self._detect_wl_or_detect_comma_or_add_driver_linker_flag
+
+
+def _flags_from_env():
+    flags_from_env = {}
```
memsharded: The way we are going to use environment variables in toolchains is still not defined, so this might be broken in the future. I am ok with adding it now this way, as long as it is clear that this integration is very experimental and subject to future breaking change.

conan/tools/qbs/generic.py
```diff
@@ -0,0 +1,296 @@
...
+            }
+        }
+        ''')
+
+    def __init__(self, conanfile):
+        _check_for_compiler(conanfile)
+        self._conanfile = conanfile
+        _setup_toolchains(conanfile)
+        self._profile_values_from_setup = (
+            _read_qbs_toolchain_from_config(conanfile))
```
memsharded: I am a bit confused about this round-trip conversion, but probably it is because I don't know Qbs enough. If the config data is already present in a qbs format file, why is it necessary to parse it, and translate it to a conan-toolchain format one? Couldn't it be used directly from that original qbs origin?

conans/__init__.py
```diff
@@ -6,6 +6,7 @@
 from conans.client.build.cmake import CMake
 from conans.client.build.meson import Meson
 from conans.client.build.msbuild import MSBuild
+from conans.client.build.qbs import Qbs
```
memsharded: This is also changing, as users will import it also ``from conan.tools.qbs import Qbs``

memsharded (changes requested):
This is looking good, thanks very much.
Please have a look to the comments, and also to the broken CI, as it seems the import "toolchain.qbs.generic" is wrong

conan/tools/qbs/generic.py
Psy-Kai: I use the qbs command `qbs-setup-toolchains` here because it contains some logic on how to detect compilers and create a qbs-profile out of it. I did not want to mirror the logic here. The problem is that this tool creates qbs-settings which are by default user/system wide and are stored in a `.ini` file containing binary Qt data. So parsing this settings file cannot be easily be done by in python. This is why I use another qbs command `qbs-config` to read the values in a easy to parse format.

conan/tools/qbs/generic.py
Psy-Kai: I the environment should not be used, how do you plan to pass the compiler path or cross compiling flags? The current cross compiling goto way seems like environment variables: https://docs.conan.io/en/latest/systems_cross_building/cross_building.html#conan-v1-24-and-newer

conans/__init__.py
Psy-Kai: Will the build helper also be imported from `conan.tools`? I saw that the toolchain has moved there but not the build helper classes.

conans/client/build/qbs.py
Psy-Kai: I just defined a custom exception because I wanted to prepend the origin of the exception to the message (qbs build helper and qbs toolchain). Without the custom exception I would need to add this string to every message manually. But if this is not desired I can remove the exceptions.

Psy-Kai:
@memsharded Could it be that some tests run in an environment where `CC` is set?

Psy-Kai:
@memsharded Could you help me here? I don't know how to fix the error in the python2.7 test.

memsharded:
Hi @Psy-Kai 

Sorry for the delay. I am having a look, and doing some changes to align to latest changes in toolchains. I will push changes soon.

memsharded:
Hi @Psy-Kai 

I did a few days ago a PR with a review in https://github.com/Psy-Kai/conan/pull/1. Please have a look when possible.

Psy-Kai:
@memsharded There is still some problem on python2.7 regarding some *IO type class (like StringIO). I am not skilled enough to find the problem, sorry :/

memsharded:
> @memsharded There is still some problem on python2.7 regarding some *IO type class (like StringIO). I am not skilled enough to find the problem, sorry :/

Yeah, don't worry, it is not evident, I am having a look, hopefully I can fix it, I will let you know.

memsharded (approved):
I think this is a good start, as a first iteration, knowing that:

- The new toolchains and helpers are all experimental. They will change in things like folders layouts, or environment management. 
- ``conan.tools.xxx`` should only import thing from itself, things like ``tools.vcvars()`` usage will need to be refactored. Also new ``msvc`` compiler is coming soon, will need to be taken into account.
- It seems there are some implementation details/style that could be improved (more idiomatic python), but that can be done later.
- At some point it will be necessary to add some integration testing that actually calls qbs

I will request some other reviews too, but lets aim to merge this for Conan 1.33

conan/tools/qbs/qbstoolchain.py
```diff
@@ -0,0 +1,264 @@
...
+        if compiler == 'gcc':
+            return 'mingw'
+        if compiler == 'Visual Studio':
+            if tools.msvs_toolset(conanfile) == 'ClangCL':
+                return 'clang-cl'
+            return 'cl'
+        if compiler == 'clang':
+            return 'clang-cl'
+        raise ConanException('unknown windows compiler')
+
```
danimtb, line 100:
```suggestion
```

danimtb (approved):

conan/tools/qbs/qbs.py
```diff
@@ -0,0 +1,97 @@
...
+
+from conans import tools
+from conans.errors import ConanException
+
+
+def _configuration_dict_to_commandlist(name, config_dict):
+    command_list = ['config:%s' % name]
+    for key, value in config_dict.items():
+        if type(value) is bool:
+            if value:
```
SSE4, line 11: just FYI: shorter syntax:
```
b = 'true' if value else 'false'
```
(no need to change)

SSE4 (approved):

memsharded:
Merged, this will be in Conan 1.33.

Please, try to contribute the docs and update the first comment ``Docs: https://github.com/conan-io/docs/pull/XXXX`` section, the changelog is generated automatically from this. A few hints:

- Explicitly add that the toolchain, helper, etc, are experimental and subject to change
- The documentation can be concise in a first iteration, and completed later, maybe just the reference section in ``tools.qbs``.

PR #8130, opened 2020-11-30
Inject CMake build_modules after targets are created

### LINKED ISSUE #7261: [feature] CMake generators: Move includes of `build_modules` after targets are created in generated CMake find modules
The attribute `self.cpp_info.build_modules` can be used to set CMake files which get included in the generated find modules created by `cmake_find_package`.

Today, these `include` statements are put _before_ the targets which are defined by the generated find module files, rather than at the end of the file, which means the build modules script cannot interact with the exported targets directly.

@danimtb [has suggested](https://github.com/conan-io/conan/issues/7217#issuecomment-648862422) there may not have been a motivation in putting the build modules before the target, so if the is the case let's move them to the end of the file.

This would enable things like [setting additional properties on the targets](https://github.com/conan-io/conan/issues/7217#issuecomment-648843268) or [creating clones or aliases of them](https://github.com/TimSimpson/Lp3-Engine/blob/master/conan/CMakeLists.txt#L67).

### LINKED ISSUE #7615: [feature] cmake_find_package[_multi] generators: provide a way to create non namespaced imported targets
Currently, `cmake_find_package` and `cmake_find_package_multi` generators always create namespaced imported targets in generated module/config files.
While not considered as best practice, it's not uncommon that official CMake config file of a library create non-namespaced imported targets (and quite uncommon for module files because official ones are usually provided by Kitware).

This is a non-exhaustive list of CCI recipes whose official config file creates non-namespaced targets, to show you that it's not just a corner case:

```
ade
amqp-cpp
antlr4-cppruntime
arcus
arduinojson
argtable3
asyncplusplus
celero
cereal
charls
cjson
cppzmq
cryptopp
cwalk
czmq
dcmtk
debug_assert
docopt.cpp
easy_profiler
effolkronium-random
embree3
evix2
ezc3d
fast-cdr
fast-dds
fcl
foonathan-memory
freetype
g3log
gcem
gdcm
gflags
glfw
glslang
gtsam
hana
jinja2cpp
json-schema-validator
jsoncpp
kangaru
leptonica
libccd
libde265
libe57format
libgeotiff
libkml
libmediainfo
libpng
libsigcpp
libtins
libwebsockets
libyaml
libzen
litehtml
log4cxx
mbits-args
mfast
mpark-variant
msgpack
octomap
onnx
opencascade
opencolorio
opencv
opengv
openjpeg
openmesh
parallel-hashmap
pro-mdnsd
rapidcheck
rapidjson
raylib
read-excel
rxcpp
sfml
spirv-cross
spirv-tools
sqlitecpp
symengine
szip
tensorpipe
tesseract
tinyobjloader
tlx
tmx
type_safe
utfcpp
wslay
xproperty
xsimd
xtensor
xtl
yaml-cpp
zeromq
```

It's worth noting that a library may have an official module and config file, and one might create namespaced targets but not the other.
libpng and freetype are good examples.

| library | cmake target in module file | cmake target in config file |
|--------|--------| --------| 
| freetype | Freetype::Freetype | freetype |
| libpng | PNG::PNG | png |

danimtb:
Changelog: Fix: Inject build modules after CMake targets are created
Docs: https://github.com/conan-io/docs/pull/1944

This allows uisng build modules to generate non-namespaced targets. --

- [x] Refer to the issue that supports this Pull Request: closes #7261, closes #7615


conans/client/generators/cmake_find_package.py
```diff
@@ -46,6 +46,10 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
                 {find_dependencies_block}
             endif()
         endif()
+
+        foreach(_BUILD_MODULE_PATH ${{{name}_BUILD_MODULES_PATHS}})
```
danimtb, line 50: maybe we have to add another type of build modules, one to be loaded before targets are created (as it right now) and this one adding build modules after targets are created

conans/client/generators/cmake_find_package_common.py
```diff
@@ -72,10 +72,6 @@
 
 set(CMAKE_MODULE_PATH {deps.build_paths} ${{CMAKE_MODULE_PATH}})
 set(CMAKE_PREFIX_PATH {deps.build_paths} ${{CMAKE_PREFIX_PATH}})
-
-foreach(_BUILD_MODULE_PATH ${{{name}_BUILD_MODULES_PATHS{build_type_suffix}}})
```
memsharded, line 76: Wasn't this a feature already of both generators, the normal and the _multi one? removing this from here won't break the _multi one?

From the docs:
> The <PKG-NAME>Targets-.cmake files use <PKG-NAME>_BUILD_MODULES_<BUILD-TYPE> values to include the files using the include(…) CMake directive. This makes functions or utilities exported by the package available for consumers just by setting find_package(<PKG-NAME>) in the CMakeLists.txt.

danimtb: I conceived the PR as a POC as there are some points for discussion. For example, this is moving the injection of build modules AFTER the creation of the targets, so you are able to do things with them instead of having build modules declares BEFORE the targets (a point for discussion).

The other point is that you cannot create ALIAS targets, so instead, I used a normal target and linked it to the global one

conans/client/generators/cmake_find_package.py
danimtb: Directly related to https://github.com/conan-io/conan/issues/7261

conans/test/functional/generators/cmake_test.py
```diff
@@ -507,3 +506,69 @@ def build(self):
 
         # Test cmake_find_package_multi generator without components
         run_test("cmake_find_package_multi", cmakelists, with_components=False)
+
+    def test_build_modules_alias_target(self):
```
danimtb: Apparently, this was already working with the CMake generator

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -13,6 +13,7 @@
 class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
     name = "cmake_find_package_multi"
     _configurations = ["Release", "RelWithDebInfo", "MinSizeRel", "Debug"]
+    _configurations_upper = ["RELEASE", "RELWITHDEBINFO", "MINSIZEREL", "DEBUG"]
```
memsharded: Better remove this and do an ``.upper()`` call when necessary.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -68,6 +70,15 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
              {%- endfor %})
     """)
 
+    build_modules = Template("""
```
memsharded, line 73: This is a bit ugly, to have this template here, and then some line below, have an almost identical block that is not reusing this one. For just 1 reuse of this, I would much prefer to have it integrated in the template directly.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -255,6 +268,33 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
                          $<$<CONFIG:{{config}}>:{{ '${'+pkg_name+'_COMPONENTS_'+config.upper()+'}'}}>
                          {%- endfor %})
         endif()
+
+        ########## BUILD MODULES ####################################################################
+        #############################################################################################
+
+        {%- for comp_name, comp in components %}
```
memsharded, line 275: Does the component model makes sense for this, if the build_modules of each component is always unconditionally included? It doesn't depend on linking that component or not. 

If that is the case, I would restrict build-modules to global ones only.

conans/client/generators/cmake_find_package_multi.py
danimtb: I prefer to have this block here as this one is used to format the file generated when the package as no components, while the other is placed the one that is generated with components. I don't want to make the second one a "compound" of different blocks that will make it harder to read

conans/client/generators/cmake_find_package.py
```diff
@@ -156,10 +160,6 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
         set(CMAKE_MODULE_PATH {{ comp.build_paths }} ${CMAKE_MODULE_PATH})
         set(CMAKE_PREFIX_PATH {{ comp.build_paths }} ${CMAKE_PREFIX_PATH})
 
-        foreach(_BUILD_MODULE_PATH {{ '${'+pkg_name+'_'+comp_name+'_BUILD_MODULES_PATHS}' }})
-            include(${_BUILD_MODULE_PATH})
-        endforeach()
-
```
jgsogo, line 162: Maybe I'm wrong, but...

If components are iterated in order, then it is safe to add these build-modules here (just after the component). That way we ensure that a build-module declared at a component level is not using any target-component that is not available.

...and we include the `build_modules` declared at component level.

conans/test/functional/generators/cmake_find_package_test.py
```diff
@@ -13,6 +13,110 @@
 from conans.test.utils.tools import TestClient, NO_SETTINGS_PACKAGE_ID
 
 
+@pytest.mark.slow
+@pytest.mark.tool_cmake
+class TestCMakeFindPackageGenerator:
+
+    @pytest.mark.parametrize("use_components", [False, True])
+    def test_build_modules_alias_target(self, use_components):
```
jgsogo, line 20: In this test, when `use_components =True`, I can see the following:

**Findhello.cmake**
```cmake
...
set(namespace_BUILD_MODULES_PATHS "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmphcq2mp5yconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")
...
set(namespace_comp_BUILD_MODULES_PATHS "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmphcq2mp5yconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")
...
...
foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS})
    include(${_BUILD_MODULE_PATH})
endforeach()
```

I would expect the build-module to appear only at the component level... and, of course, a loop calling `include` for each one of them.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -69,6 +70,15 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
...
     """)
 
+    build_modules = Template("""
+# Build modules
+{%- for config in configs %}
+foreach(_BUILD_MODULE_PATH {{ '${'+name+'_BUILD_MODULES_PATHS_'+config.upper()+'}' }})
+    include(${_BUILD_MODULE_PATH})
+endforeach()
+{%- endfor %}
+    """)
```
jgsogo, line 80: Here we are adding unconditionally all the build-modules. The resulting cmake file contain these lines:

```cmake
# Build modules
foreach(_BUILD_MODULE_PATH ${hello_BUILD_MODULES_PATHS_DEBUG})
    include(${_BUILD_MODULE_PATH})
endforeach()
foreach(_BUILD_MODULE_PATH ${hello_BUILD_MODULES_PATHS_MINSIZEREL})
    include(${_BUILD_MODULE_PATH})
endforeach()
foreach(_BUILD_MODULE_PATH ${hello_BUILD_MODULES_PATHS_RELWITHDEBINFO})
    include(${_BUILD_MODULE_PATH})
endforeach()
foreach(_BUILD_MODULE_PATH ${hello_BUILD_MODULES_PATHS_RELEASE})
    include(${_BUILD_MODULE_PATH})
endforeach()
```

All of them are included regardless of the build_type used. We need a generator expression to expand and use only the build-modules for the matching build-type.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -256,6 +268,19 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
...
+
+        ########## GLOBAL BUILD MODULES #############################################################
+
+        {%- for config in configs %}
+
+        foreach(_BUILD_MODULE_PATH {{ '${'+pkg_name+'_BUILD_MODULES_PATHS_'+config.upper()+'}' }})
+            include(${_BUILD_MODULE_PATH})
+        endforeach()
+
+        {%- endfor %}
```
jgsogo, line 286: Same comment as above, here we add all the build-modules unconditionally, we need a generator expression to `include` only the ones for the matching `config`.

This is current output:

```cmake
########## BUILD MODULES ####################################################################
#############################################################################################

########## GLOBAL BUILD MODULES #############################################################

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_DEBUG})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_MINSIZEREL})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_RELWITHDEBINFO})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_RELEASE})
    include(${_BUILD_MODULE_PATH})
endforeach()
```

conans/test/functional/generators/cmake_find_package_test.py
memsharded, line 20: If this is for injecting a build_module for a dependency, could we better test it without actually needing to build any binaries? I think we need to reduce the tests a little bit, specially if we can avoid compiling that would be great.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -256,6 +268,19 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
...
         endif()
+
+        ########## BUILD MODULES ####################################################################
+        #############################################################################################
+
+        ########## GLOBAL BUILD MODULES #############################################################
+
+        {%- for config in configs %}
+
+        foreach(_BUILD_MODULE_PATH {{ '${'+pkg_name+'_BUILD_MODULES_PATHS_'+config.upper()+'}' }})
```
memsharded: I wouldn't mind if build_modules does not work/exist at all for the multi-config. In fact, this is something I think we should deprecate in Conan 2.0, the complexity is high, the use cases are probably not that legit anymore (the multi-config packages were created when multi-generators didn't exist)

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -139,6 +139,105 @@ def build(self):
         client.run("create .")
         self.assertIn("Printing using a external module!", client.out)
 
+    @parameterized.expand([(False,), (True,)])
+    def test_build_modules_alias_target(self, use_components):
```
jgsogo: Same issue as the one with the `cmake_find_package` generator: 

**helloTarget-release.cmake**
```cmake
...
set(namespace_BUILD_MODULES_PATHS_RELEASE "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmpf8fefhlrconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")
...
set(namespace_comp_BUILD_MODULES_PATHS_RELEASE "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmpf8fefhlrconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")
...
```

I expect the build-module to appear listed only under the `namespace_comp` ones...

**hello-config.cmake**
```cmake
...
########## BUILD MODULES ####################################################################
#############################################################################################

########## GLOBAL BUILD MODULES #############################################################

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_DEBUG})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_MINSIZEREL})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_RELWITHDEBINFO})
    include(${_BUILD_MODULE_PATH})
endforeach()

foreach(_BUILD_MODULE_PATH ${namespace_BUILD_MODULES_PATHS_RELEASE})
    include(${_BUILD_MODULE_PATH})
endforeach()
...
```

Build-modules per component are not included.

memsharded (approved):
Looking good, I think it is ready, somethings could be improved now, but I wouldn't oppose if they are fixed later.

jgsogo (changes requested):
This PR is a bit challenging. I'm assuming these expectations:

 * build modules are included after the element they refer to:
   + global build-modules at the very end
   + build-modules per component, after corresponding component is defined (nice to have, but they could go at the bottom too).
* per-component build-modules are not included as global ones as well.
* build-modules for a given configuration are added only when that configuration is used (requires some generator expression).


Please, tell me if I'm wrong with any of the previous points.

conans/client/generators/cmake_find_package.py
danimtb: The problem is that you don't have the component target declared at this point

conans/client/generators/cmake_find_package_multi.py
danimtb: I have tried modifying this to use generator expressions here, but those only work in target properties as generator expressions are specific for CMake generators. I have also tried using CMAKE_BUILD_TYPE to make the conditionals but it is not available. I am leaving this as it is right now

conans/test/functional/generators/cmake_find_package_test.py
danimtb: done

conans/client/generators/cmake_find_package_multi.py
jgsogo: That's right. Generator expressions only work with target-related properties... so for multi-generators this is unsolvable. 🤔 . Maybe something to document.

Then it is clearly a CMake limitation and I'd extend it to non-multi generators as well (for consistency). It means that the list of build-modules to include cannot be different based on the configuration name (typically `build_type`), recipe authors need to implement the build-module with this circumstance in mind. We'd go crazy trying to protect the user from it given our `cpp_info` object.

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -12,6 +12,80 @@
 from conans.util.files import load
 
 
+@pytest.mark.tool_cmake
+class TestCMakeFindPackageMultiGenerator:
+
+    @pytest.mark.parametrize("use_components", [False, True])
+    def test_build_modules_alias_target(self, use_components):
```
jgsogo, line 19: I see the build_modules at global level and at component level:

```cmake
set(namespace_comp_BUILD_MODULES_PATHS_RELEASE "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmpd7yqgq3vconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")

set(namespace_BUILD_MODULES_PATHS_RELEASE "/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmpd7yqgq3vconans/path with spaces/data/hello/1.0/_/_/package/f83037eff23ab3a94190d7f3f7b37a2d6d522241/share/cmake/target-alias.cmake")

```

conans/test/functional/generators/cmake_find_package_test.py
jgsogo: Still the same, the same build-module appear at component and global level.

jgsogo (changes requested):
Agree we cannot decide which build-modules to include based on the configuration for multi-config generator (the `include` function cannot use generator expressions). Agree we shouldn't add the feature to non-multi generator for consistency too.

But, I think we can include only build-modules at the level where they are declared.

I'll have a look to the implementation.

jgsogo (approved):
Wow, I'm no longer sure if it is worth fixing/adding that in this PR. After having a look at the sources I can see that **we are populating global variables even if we are using components**. This is intended and it is good according to [the docs](https://docs.conan.io/en/latest/reference/generators/cmake_find_package.html#variables-in-find-pkg-name-cmake) where we list the CMake variables that are available for the different generators.

And I see current implementations `include` only the proper build-modules: global ones or components, depending on the generator.

Good!

PR #8131, opened 2020-12-01
small refactor of add_permissions

memsharded:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -227,6 +227,7 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
 
 @contextmanager
 def _add_write_permissions(file_path):
+    # Assumes the file already exist in disk
     write = stat.S_IWRITE
     saved_permissions = os.stat(file_path).st_mode
     if saved_permissions & write == write:
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -226,19 +226,18 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
 
 
 @contextmanager
-def _add_permissions(file_path, permissions):
-    if os.path.isfile(file_path):
-        saved_permissions = os.stat(file_path).st_mode
-        if saved_permissions & permissions == permissions:
-            yield
-            return
-        try:
-            os.chmod(file_path, saved_permissions | permissions)
-            yield
-        finally:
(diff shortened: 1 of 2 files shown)


conans/client/tools/files.py
```diff
@@ -226,19 +226,18 @@ def _manage_text_not_found(search, file_path, strict, function_name, output):
...
-            return
-        try:
-            os.chmod(file_path, saved_permissions | permissions)
-            yield
-        finally:
-            os.chmod(file_path, saved_permissions)
-    else:
+def _add_write_permissions(file_path):
+    write = stat.S_IWRITE
+    saved_permissions = os.stat(file_path).st_mode
```
SSE4, line 232: now it may raise if file doesn't exist before save

memsharded: This is fine. The scope is changing permissions to files that already exists. Any usage of this function acts on an existing file that was ``load()`` before.
In other words, the ``else`` path of the first ``if`` is completely untested, non-covered by tests. It can be safely removed until it is really necessary. If the use case or usage changes later, this function should indeed break, this is exactly the behavior I want, because in that case, I want to understand what is the case of changing permissions of a file that doesn't exist.

Same happen with the embedding of S_IWRITE. In general, functions should receive the minimum necessary scope and functionality, there is no point of being able to change arbitrary permissions if there is no real use case so far and it is not really covered by tests. Premature generalization.

SSE4: okay, but I think there is potential risk for the file system race:
1. load file
2. another process removes/renames the file
3. add_permissions/save file

now it may raise here, which might be breaking

memsharded: There are tons of filesystem race conditions that can happen all around the Conan codebase if we think that any arbitrary operation can happen concurrently.

If a ``replace_in_file()`` is overwriting a file that has been concurrently modified/removed/renamed by other process, I prefer it to crash and know it, because there is something quite bad there. Besides that, ``replace_in_file()`` is a recipe operation, we can safely assume that on a given path/folder, cannot happen concurrently, because if that could happen, we would probably be in a way more problematic scenario, and this would be the least of our problems.

SSE4: I think it's fair, thanks for the clarification

SSE4 (approved):

PR #8141, opened 2020-12-02
Cli2.0 add file handler to log to files

czoido:
Changelog: Feature: Add logging to file capabilities to cli 2.0.
Docs: Omit


czoido:
Continuing the work in the Conan cli2.0 this PR is changing the logger color handling. Now a `ColorFormatter` is added so that the color is added there, that seems more correct than the previous approach.
Also, using the `CONAN_TRACE_FILE` env var (just fo the moment, we should decide if this stays or not for 2.0) adds a simple file logger that will not have color codes added to messages.
Looks like multiline log messages are not a problem and the output seems correct.

PR #8143, opened 2020-12-02
Fix pkg-config description parser

### LINKED ISSUE #8142: [bug] Recipe description breaks pkg-config generator
This bug comes from https://github.com/conan-io/conan-center-index/issues/3745

As you can see, Conan has a limitation to format strings when it comes with a tuple. Tuple is total valid on multi-line situation, and it's considered more pythonic.

 * Package Name/Version: gsl-lite/0.37.0
 * Operating System+version: Linux Ubuntu 18.04
 * Conan version: conan 1.31.3
 * Python version: Python 3.6.9

Described on https://github.com/conan-io/conan-center-index/issues/3745

Described on https://github.com/conan-io/conan-center-index/issues/3745

uilianries:
Changelog: Fix: Throw error when the recipe description is not a string.
Docs: Omit

fixes https://github.com/conan-io/conan/issues/8142

### DIFF
--- a/conans/test/functional/generators/pkg_config_test.py
+++ b/conans/test/functional/generators/pkg_config_test.py
@@ -203,24 +203,3 @@ class PkgConfigConan(ConanFile):
         pc = client.load("pkg.pc")
         self.assertNotIn("libdir=${prefix}/lib", pc)
         self.assertNotIn("includedir=${prefix}/include", pc)
-
-    def test_tuple_description(self):
-        conanfile = textwrap.dedent("""
-            from conans import ConanFile
-            class PkgConfigConan(ConanFile):
-                name = "foo"
-                version = "0.1"
-                description = ("Conan, ",
-                               "the package manager  ")
-
-                def build(self):
-                    pass
-            """)
-
-        client = TestClient()
-        client.save({"conanfile.py": conanfile})
-        client.run("create .")
-        client.run("install foo/0.1@ -g pkg_config")
-
-        pc_content = client.load("foo.pc")
-        self.assertIn("Description: Conan, the package manager", pc_content)
--- a/conans/test/functional/generators/pkg_config_test.py
+++ b/conans/test/functional/generators/pkg_config_test.py
@@ -205,19 +205,18 @@ class PkgConfigConan(ConanFile):
         self.assertNotIn("includedir=${prefix}/include", pc)
 
     def test_tuple_description(self):
-        conanfile = """
-from conans import ConanFile
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile
+            class PkgConfigConan(ConanFile):
+                name = "foo"
+                version = "0.1"
+                description = ("Conan, ",
+                               "the package manager  ")
 
+                def build(self):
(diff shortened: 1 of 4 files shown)


conans/test/functional/generators/pkg_config_test.py
```diff
@@ -203,3 +203,25 @@ def test_empty_include(self):
         pc = client.load("pkg.pc")
         self.assertNotIn("libdir=${prefix}/lib", pc)
         self.assertNotIn("includedir=${prefix}/include", pc)
+
+    def test_tuple_description(self):
+        conanfile = """
```
memsharded: USe textwrap.dedent() in tests

conans/test/functional/generators/pkg_config_test.py
```diff
@@ -203,3 +203,25 @@ def test_empty_include(self):
...
+from conans import ConanFile
+
+
+class PkgConfigConan(ConanFile):
+    name = "foo"
+    version = "0.1"
+    description = ("Conan, ",
```
memsharded: Is a tuple an allowed description? I don't recall that it is.

uilianries: Not a description, but a python feature. Months ago I watched a python meetup where that usage was more pythonic. And here is the reference: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses.

conans/test/functional/generators/pkg_config_test.py
uilianries: Done

conans/test/functional/generators/pkg_config_test.py
memsharded: That is without the comma @uilianries 

```python
description = ("Conan, "
               " the C++ package manager")
```
That will resolve to a single string, not a tuple

uilianries: I see, so it's more a prone error than a bug. Anyway, if we pass a tuple, we have an exception which doesn't help. So we could accept tuple, or raise a better error.

memsharded: Lets make a better error. I don't see here a use case in which we want to support tuples, lists, or other different ways of inputting the data, it should always be a string.

uilianries: Done.

memsharded (approved):

Croydon:
This change breaks some existing recipes, like https://github.com/conan-io/conan-center-index/issues/4726 

Shouldn't such changes only happen with Conan 2.0?

memsharded:
Hi @Croydon 

It is not a breaking change, it is a bug fix. As described in the documentation (https://docs.conan.io/en/latest/reference/conanfile/attributes.html#description), the description field is a string. Passing any other type, like a tuple, is an error and the recipe should fail (or could have undefined behavior if the error is not managed)

PR #8154, opened 2020-12-04
refactor to decouple sources fetching

memsharded:
Changelog: Omit
Docs: Omit

Very simple refactor to rename a function, and decoupling fetching the sources to copying the sources to the build folder, so the logic might be used in a potential "install sources" command.


solvingj (approved):
Appears very simple, no issue found.

PR #7416, opened 2020-07-24
Fix: Make env_diff be portable across different shells

ohanar:
The portable version of env_diff seems to have been lost in the rebasing of #6735.

Closes #8168.

Changelog: Fix: Make env_diff be portable across different shells
Docs: omit

- [x] Refer to the issue that supports this Pull Request.
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #8169, opened 2020-12-07
Fix env_diff() on Linux

### LINKED ISSUE #8168: [bug] tools.intel_compilervars_dict() not working
  * Operating System+version: CentOS 8
  * Compiler+version: Intel 16
  * Conan version: 1.32.0
  * Python version: 3.6.8

When using the tools.intel_compilervars_dict() function, the dict returned does not contain the expected values.
Expected is a dictionary with the environment variables set by the tools.intel_compilervars_command(). However all variable names have a prefix of "export " or "declare -x " (depending on the used shell).
Because the variable names are malformed, this dict cannot be used with tools.environment_append() as intended.

The problem also affects tools.intel_compilervars() which uses tools.intel_compilervars_dict() internally and also the CMake helper when used with a Intel compiler.

I beliefe the root cause is in env.env_diff() and the fix is quiet simple, I will provide a PR.

tools.vcvars_dict() is not affected as it does not use env.env_diff().

akleber:
Changelog: Bugfix: fixes #8168 

### DIFF
--- a/conans/client/tools/env.py
+++ b/conans/client/tools/env.py
@@ -124,7 +124,7 @@ def env_diff(cmd, only_diff):
     if platform.system() == "Windows":
         cmd += " && set"
     else:
-        cmd += " && export"
+        cmd += " && printenv"
     ret = check_output_runner(cmd)
     new_env = {}
     for line in ret.splitlines():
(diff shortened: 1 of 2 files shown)


ohanar:
Duplicate of #7416.

memsharded:
Environment management has changed with new ``VirtualBuildEnv`` generator, env-diff is not longer used, closing as outdated, thanks!

PR #8186, opened 2020-12-10
add build_requires to conandeps.props file in MSBuildDeps

### LINKED ISSUE #8170: [bug] MSBuildDeps generator: build_requires not added to conan_deps.props
Build requirements are not added to the generated `conan_foo.props` file.

  * Operating System+version:
  * Compiler+version:
  * Conan version: 1.32.0
  * Python version:

1. add a library package (e.g. `foo`) to `build_requires`
2. execute `conan install` using the `MSBuildDeps`/`msbuild` generator

- `conan_foo.props` is generated
- generated `conan_deps.props` does not contain an import of `conan_foo.props`

Background:
We have a test helper library as a dependency that is only used in the test. Previously we added it as a private requirement, which showed up in the `conan_foo.props`. From our current understanding of Conan, it however is better written as a build requirements.

memsharded:
Changelog: Fix: Include ``build_requires`` in the global ``conandeps.props`` file generated by MSBuildDeps.
Changelog: Fix: Change `MSBuildDeps` file ``conan_deps.props`` to ``conandeps.props`` to avoid collision with a package named "deps".
Docs: Omit

Close https://github.com/conan-io/conan/issues/8170

### DIFF
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -221,7 +221,10 @@ class MSBuildDeps(object):
         conf_name = self._config_filename()
         condition = self._condition()
         # Include all direct build_requires for both build & host context. This might change
-        build_requires_names = [name for name, _ in self._conanfile.build_requires.keys()]
+        if hasattr(self._conanfile, "build_requires"):
+            build_requires_names = [name for name, _ in self._conanfile.build_requires.keys()]
+        else:
+            build_requires_names = []
         public_deps = list(self._conanfile.requires.keys()) + build_requires_names
         result[general_name] = self._deps_props(general_name, public_deps)
         for dep_name, cpp_info in self._conanfile.deps_cpp_info.dependencies:
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -220,8 +220,8 @@ class MSBuildDeps(object):
         general_name = "conandeps.props"
         conf_name = self._config_filename()
         condition = self._condition()
-        # Include all direct build_requires for both build & host context. This might change
-        direct_deps = self._conanfile.deps_cpp_info.direct_deps
+        # Include all direct build_requires for host context. This might change
+        direct_deps = self._conanfile.deps_cpp_info.direct_host_deps
         result[general_name] = self._deps_props(general_name, direct_deps)
         for dep_name, cpp_info in self._conanfile.deps_cpp_info.dependencies:
             # One file per configuration, with just the variables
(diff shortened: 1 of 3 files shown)


jgsogo (changes requested):
As a generator, the information from the dependencies should be taken from the `deps_cpp_info` object, it already contains all the requirements (and build_requiremnts) given the logic implemented in the graph.

memsharded:
> As a generator, the information from the dependencies should be taken from the deps_cpp_info object, it already contains all the requirements (and build_requiremnts) given the logic implemented in the graph.

The problem is that ``deps_cpp_info`` object do not contain information about the immediate dependencies. Only the direct dependencies (requires and build-requires declared in the recipe) are being added to ``conandeps.props``.

lieser:
Tested your change branch, and can confirm the `build_requires` is now included. Thanks for the quick fix.

jgsogo:
But we agreed that the way to get information from the requirements was to use those classes (there was a refactor related to `cpp_info` and a bigger refactor pending for its implementation, that (probably) will be superseded by the new way to retrieve information from the graph nodes). We cannot access this information using a third new way that it is not the one every generator uses neither the new one.

If we need to differentiate between _direct/inmediate_ requirements and _transitive_ requirements, let's add a mark to them in the `deps_cpp_info` object and set its value in the graph, in the same place where we choose what to propagate (https://github.com/conan-io/conan/blob/develop/conans/client/installer.py#L579). It is legit to add the mark and not document it now as it is going to be superseded, but at least we keep all the logic for this graph propagation together and all the generators use the same interfaces. It will surely help us when rewriting everything related to this interface.

memsharded:
@lieser I did some changes to the way the direct dependencies are computed, tests are ok, but just in case you want to give the branch another run.

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -217,11 +217,12 @@ def _content(self):
         if not self._conanfile.settings.get_safe("build_type"):
             raise ConanException("The 'msbuild' generator requires a 'build_type' setting value")
         result = {}
-        general_name = "conan_deps.props"
+        general_name = "conandeps.props"
         conf_name = self._config_filename()
         condition = self._condition()
-        public_deps = self._conanfile.requires.keys()
-        result[general_name] = self._deps_props(general_name, public_deps)
+        # Include all direct build_requires for both build & host context. This might change
```
jgsogo:
```suggestion
        # Include all direct build_requires for host context
```

jgsogo (approved):
👍

lieser:
@memsharded tested the updated branch again successfully

conan/tools/microsoft/msbuilddeps.py
memsharded: Done, and fixed also, filtering CONTEXT_HOST in the code, not only the comment

jgsogo (approved):
_"Not only the comments should resemble the code, but the code should resemble the comments"_ (Conan developer, 2020)

 😄

PR #8192, opened 2020-12-10
duplicate and isolate cmake utilities

### LINKED ISSUE #7485: [feature] [toolchain.cmake] Set platform and toolset only if generator matches Visual Studio
Use case: Using MSVC with Ninja as CMake Generator. It doesn't support `PLATFORM` and `TOOLSET` variables and errors out on them. 

Consider generating the template like this:
```
if(CMAKE_GENERATOR MATCHES "Visual Studio")
        {% if generator_platform %}
        {% endif %}
        {% if toolset %}
        set(CMAKE_GENERATOR_TOOLSET "{{ toolset }}" CACHE STRING "" FORCE)
        {% endif%}
endif()
```

memsharded:
Changelog: Omit
Docs: Omit

Put a copy of some cmake helpers into ``conan.tools.cmake`` because changes will be needed to adapt to:

- build-profiles
- new settings.compiler=msvc
- using config instead of env-vars

Too risky to do it in the code used by old helpers.

Close https://github.com/conan-io/conan/issues/7485

#tags: slow

### DIFF
--- a/conan/tools/cmake/__init__.py
+++ b/conan/tools/cmake/__init__.py
@@ -1,4 +1,4 @@
 # noinspection PyUnresolvedReferences
 
-from .toolchain import CMakeToolchain
-from .cmake import CMake
+from conan.tools.cmake.toolchain import CMakeToolchain
+from conan.tools.cmake.cmake import CMake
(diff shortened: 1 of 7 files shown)


danimtb (approved):

danimtb:
there are some conflicts here

PR #8194, opened 2020-12-11
Feature/generator cmake custom configs settings

memsharded:
Changelog: Feature: ``CMakeDeps`` now takes values for configurations from ``settings.yml``.
Docs: https://github.com/conan-io/docs/pull/1992

I think that we need to make it dynamic, so this ``CMakeDeps`` generator doesn't rely on predefined settings.yml possibilities, and just on the actual ``conan install`` invocations. But until this is there, this test captures the desired behavior of not having to hardcode custom configurations in recipes.


conans/test/functional/generators/cmake_find_package_multi_configs_test.py
```diff
@@ -94,3 +96,93 @@ def test_generator_multi(self):
...
+            conan_basic_setup()
+            set(CMAKE_CXX_FLAGS_MYRELEASE ${CMAKE_CXX_FLAGS_RELEASE})
+            set(CMAKE_C_FLAGS_MYRELEASE ${CMAKE_C_FLAGS_RELEASE})
+            set(CMAKE_EXE_LINKER_FLAGS_MYRELEASE ${CMAKE_EXE_LINKER_FLAGS_RELEASE})
+            """)
+        self.client.save({"src/CMakeLists.txt": cmake})
+        self.client.run("create . hello/0.1@ -s compiler.version=15 "
+                        "-s build_type=MyRelease -o hello:shared=True")
+        self.client.run("create . hello/0.1@ -s compiler.version=15 "
+                        "-s build_type=MyRelease")
```
jgsogo: It is enough to create one of them, right?

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -260,8 +260,8 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
 
     def __init__(self, conanfile):
         super(CMakeFindPackageMultiGenerator, self).__init__(conanfile)
-        self.configurations = ["Release", "RelWithDebInfo", "MinSizeRel", "Debug"]
         self.configuration = str(self.conanfile.settings.build_type)
+        self.configurations = [v for v in conanfile.settings.build_type.values_range if v != "None"]
```
jgsogo, line 264: Maybe we can turn this into a property and prevent people from assigning/appending to it... or make it a private attribute.


```python
@property
def configurations(self):
    return [v for v in conanfile.settings.build_type.values_range if v != "None"]
```

jgsogo (commented):
A nice step before adding those configs dynamically.

conans/client/generators/cmake_find_package_multi.py
memsharded: It is still possible to add new configurations, that are not defined in ``settings.yml``, for example "ReleaseShared", that is defined from settings build_type + option shared. There is a test for it, and the only way is that recipes are able to explicitly add it. 

The moment the generator is fully dynamic, it can be removed, but atm, it will make that test fail.

PR #8201, opened 2020-12-12
Feature/msvc

### LINKED ISSUE #3573: Incorrect handling of MSVC versions
I am using Conan v1.7.4.

I've created a static library Conan package using VS 2017 15.7 and updated my Visual Studio to 15.8. After the update, Conan failed to detect a change in the compiler version and did not perform recompilation of the static library. Instead, it just downloaded the same package that was built with VS 2017 15.7. However, when linking to that package from VS 2017 15.8, I am getting following linker error:

```
LINK : fatal error C1047: The object or library file 'C:\Users\dodo\.conan\data\LogAndTimer\1.0.1\microblink\stable\package\8b2772668ccbf879d34a70a9d54c78ce70f38a1b\lib\LogAndTimer.lib' was created with an older compiler than other objects; rebuild old objects and libraries
2>LINK : fatal error LNK1257: code generation failed
```

However, the debug build does work (details how to reproduce are below).

I think the problem is in how Conan handles MSVC versions - instead of using the actual compiler version (CMake detects `MSVC 19.14.26431.0` in VS 2017 15.7 and `MSVC 19.15.26729.0` in VS 2017 15.8), it uses the version of Visual Studio (VS 2017 is _compiler version 15_, according to Conan).

I suggest that Conan should use the MSVC compiler version under `compiler.version` setting and additionally have `compiler.vs_version` for Visual Studio version. The reason for that is that sometimes libraries built with one version of MSVC are not compatible with later versions of MSVC, especially if you use link time code generation flag (`/LTCG` link flag and `/GL` compile flag).

## How to reproduce

- Install Visual Studio 2017, but not the latest update (at the time of writing, this is 15.8.4). For example, use VS 2017 15.7.x or 15.6.x.
- Create a conan package that contains a single static library. Make sure that `Release` build type of that package is compiled with `/GL` compile flag.
- Create an application that consumes that static library and make sure it works in `Release` build type (the app should be linked with `/LTCG` linker flag).
- Update Visual Studio 2017 to the latest version (15.8.4 at the time of writing).
- Build your application

### Expected behaviour
- `conan install` after updating the VS should calculate different hash for the library and should rebuild it, thus making sure application linking will succeed

### Actual behaviour
- Conan does not detect version change in the MSVC compiler and uses the incompatible binary, thus causing linker error described above.

Now, the main question is - could this be fixed without breaking existing conan VS packages, or we need to wait for Conan 2.0 (and disable LTCG until this issue gets fixed).

Also, as a side question, is there a possibility to configure Conan to trigger rebuild of libraries even when a minor/bugfix version of the compiler changes, regardless of the compiler being used (e.g. if we update GCC from 8.1.0 to 8.1.1, we would like to rebuild all our binaries, even though they are link-compatible). This would get us the benefits of always having binaries built with the latest compiler to get the latest bugfixes and performance improvements - this benefit is also mentioned in [conan documentation](https://docs.conan.io/en/latest/creating_packages/define_abi_compatibility.html#):

> What happens if we have a library that we can be built with GCC 4.8 and will preserve the ABI compatibility with GCC 4.9? (This kind of compatibility is easier to achieve for example for pure C libraries).
>
> **Although it could be argued that it is worth rebuilding with 4.9 too -to get fixes and performance improvements-.** Let’s suppose that we don’t want to create 2 different binaries, but just a single built with GCC 4.8 which also needs to be compatible for GCC 4.9 installations.

memsharded:
Changelog: Feature: Add new ``msvc`` compiler setting and preliminary support in ``conan.tools.cmake`` generator and toolchain.
Docs: https://github.com/conan-io/docs/pull/1991

Close https://github.com/conan-io/conan/issues/3573

It is providing a ``compatible_packages`` approach to compatibility with Visual Studio compiler.

- Support in ``conan.tools.microsoft`` following in a different PR
- Support for intel compiler will follow in a different PR, lets first do the very basics.
- Auto-detect for profile should probably be delayed until it is way more stable, not being tackled now.
- Definition of the VS IDE version to use for a given compiler version (when not the default) will require improving the "configuration" of the toolchains, that will be done in other PRs, at the moment only the default compiler-version <=> VS IDE version is supported


danimtb:
If it is possible I would add the compatibility by default to the Visual Studio compiler if we want to encourage users to move to this `msvc` compiler. In ConanCenter we can generate new configurations now, but generating all visual studio packages twice will be too much IMO. Probably we will need to do the compiler switch in Conan 2.0

SSE4:
is it possible to use a trick similar to Intel's `base_compatible`?

danimtb:
yes, something similar to that in package ID but it should be applied by default in conan 1.X IMO

SSE4:
to clarify, are we proposing one-way (one-directional) compatibility layer?
e.g. the following profile:
```
compiler = Visual Studio
compiler.version = 15
```
will be able to use packages from:
```
compiler = msvc
compiler.version = 1928
```
but not vice versa

solvingj:
> to clarify, are we proposing one-way (one-directional) compatibility layer?
> e.g. the following profile:
> 
> ```
> compiler = Visual Studio
> compiler.version = 15
> ```
> 
> will be able to use packages from:
> 
> ```
> compiler = msvc
> compiler.version = 1928
> ```
> 
> but not vice versa

what was the answer to this?

solvingj:
In any case, as a reviewer, i support the overall change of strategy.

memsharded:
> to clarify, are we proposing one-way (one-directional) compatibility layer?
> e.g. the following profile:
> 
> ```
> compiler = Visual Studio
> compiler.version = 15
> ```
> 
> will be able to use packages from:
> 
> ```
> compiler = msvc
> compiler.version = 1928
> ```
> 
> but not vice versa

No, the PR is proposing the opposite. The existing packages are the ``Visual Studio`` ones. To allow a smooth migration, when defining the ``msvc`` compiler, it should be able to fallback (as ``compatible_package``) to the existing ``Visual Studio`` binaries.

SSE4:
I am still not sure there is a huge benefit of going from `compiler.runtime` which has 4 possible values:
```
MD MDd MT MTd
```
into the two settings, `compiler.runtime` + `compiler.runtime_type` which still results into 4 possible variants:
```
static + Debug, static + Release, dynamic + Debug, dynamic + Release
```
as far as I understand from code, all 4 combinations are valid and not restricted.
IMO this will significantly complicate the logic of recipes.

SSE4:
> No, the PR is proposing the opposite. The existing packages are the `Visual Studio` ones. To allow a smooth migration, when defining the `msvc` compiler, it should be able to fallback (as `compatible_package`) to the existing `Visual Studio` binaries.

in this case, both `19.27` and `19.28` are compatible with Visual Studio 16 (`19.2x`) but they are incompatible with each other. what will happen in such a case (all 3 appear in the same graph)?
maybe we need some tests to ensure the result is correct in these situations.

memsharded:
> I am still not sure there is a huge benefit of going from compiler.runtime which has 4 possible values:

This has been driven by user feedback, and requested a few times. The main problem is that just switching the build type, like ``conan install ... -s build_type=Debug``, then suddenly requires to add the more obscure ``conan install ... -s build_type=Debug -s compiler.runtime=MDd`` to be consistent with most cases. Most users will forget, and get linking errors. Conan already has obscure logic, that will mess with the ``compiler.runtime`` to (partly) deduce it from the ``build_type``. UX don't want to define the flag value, they just want to specify the static or dynamic runtime, and get the flag value from the system.

Can you give an example why this would complicate the logic of recipes? The current usage I see are expressions like:

```python
and "MT" in str(self.settings.compiler.runtime) 
self._cmake.definitions["MSVC_USE_STATIC_CRT"] = str(self.settings.compiler.runtime) == "MT" or str(self.settings.compiler.runtime) == "MTd"
self._cmake.definitions["USE_MSVC_RUNTIME_LIBRARY_DLL"] = "MD" in self.settings.compiler.runtime
self._cmake.definitions["JANSSON_STATIC_CRT"] = str(self.settings.compiler.runtime).startswith("MT")

if self.settings.compiler == "Visual Studio":
      if "MT" in str(self.settings.compiler.runtime):
            self._autotools.defines.append("USE_STATIC_RTL")
      if "d" in str(self.settings.compiler.runtime):
           self._autotools.defines.append("USE_DEBUG_RTL")
```
I would say it is the opposite, in most cases, it will simplify the recipes, with ``if self.settings.compiler.runtime == "static"``. Of course there will be a few cases where recipes will want the flag value, but I think those cases should probably be a helper that maps from the settings to the compiler flag.

memsharded:
> in this case, both 19.27 and 19.28 are compatible with Visual Studio 16 (19.2x) but they are incompatible with each other. what will happen in such a case (all 3 appear in the same graph)?

``19.2x`` is not a version, only ``19.2`` exists in the settings. The compatibility logic is the same as existing gcc and clang ones (as requested by users feedback), if using ``19.2`` it is assumed compatibility between all the 19.2X versions, they can be mixed without problems. If using a specifc ``19.28``, then it defines compatibility only with itself.

That is only for the computation of ``package_id``, but it is totally independent of the fact that you can define different settings per package. If you define 3 different packages to have different ``msvc`` versions, then, the graph will do that. If they are incompatible, it will crash or fail to link. But in exactly the same way you can define completely different compilers for different packages in the graph, some with gcc 4.9 and others with "Visual Studio 2019". This is not something new, and it doesn't seem that we need to test those scenarios. We need to test the binary compatibility mechanisms, not that different settings are in fact compatible or incompatible.

SSE4:
the complication would be something like this:
```
if self.settings.compiler == "Visual Studio":
    runtime = str(self.settings.compiler.runtime)
elif self.settings.compiler == "msvc":
    if self.settings.compiler.runtime == "static" and self.settings.compiler.runtime_type == "debug":
        runtime = "MTd"
    < 3 more if branches >
```

memsharded:
> the complication would be something like this:
> 
> ```
> if self.settings.compiler == "Visual Studio":
>     runtime = str(self.settings.compiler.runtime)
> elif self.settings.compiler == "msvc":
>     if self.settings.compiler.runtime == "static" and self.settings.compiler.runtime_type == "debug":
>         runtime = "MTd"
>     < 3 more if branches >
> ```

Lets solve that with something like a ``from conan.tools.microsoft import msvc_runtime_flag``? That accept (self) as argument and returns MT/MD/MTd/MDd? While I see that this happens in recipes, it seems it is less frequent than the above ones, which are ugly too.

SSE4:
> Lets solve that with something like a `from conan.tools.microsoft import msvc_runtime_flag`? That accept (self) as argument and returns MT/MD/MTd/MDd? While I see that this happens in recipes, it seems it is less frequent than the above ones, which are ugly too.

build helper will be definitely helpful. grep reveals lots of cases like:
```
tools.replace_in_file(fn, "-MT", "-{}".format(str(self.settings.compiler.runtime)))
configure_args.append("--enable-msvc={}".format(self.settings.compiler.runtime))
build_flags.append('--msvc-runtime=%s' % str(self.settings.compiler.runtime))
static_lib = "amd_ags_{arch}_{vs_version}_{runtime}.lib".format(arch=win_arch, vs_version=vs_version, runtime=self.settings.compiler.runtime)
"cruntime=/%s" % self.settings.compiler.runtime,
```

still, for me two entities to describe a single thing are kinda excessive, but if users find it better I am okay. btw, we still have `build_type` in addition to `compiler.runtime_type` and they don't necessarily match...

conan/tools/cmake/utils.py
```diff
@@ -47,9 +47,18 @@ def get_generator(conanfile):
         return os.environ["CONAN_CMAKE_GENERATOR"]
 
     compiler = conanfile.settings.get_safe("compiler")
+    compiler_version = conanfile.settings.get_safe("compiler.version")
+    if compiler == "msvc":
+        version = compiler_version[:4]  # Remove the latest version number 19.1X if existing
+        _visuals = {'19.0': '14 2015',
+                    '19.1': '15 2017',
+                    '19.2': '16 2019'}[version]
```
jgsogo: Here it is where we need `vswhere`, right? We need it to tell us which is the VS version that provides that `msvc` version. Maybe we can move these lines to some `conan.tools.microsoft`, add the comments (FIXME/TODO) there and call it from here.

conans/client/conf/__init__.py
```diff
@@ -71,6 +71,13 @@
             threads: [None, posix, win32] #  Windows MinGW
             exception: [None, dwarf2, sjlj, seh] # Windows MinGW
             cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]
+        msvc:
+            version: ["19.0",
+                      "19.1", "19.10", "19.11", "19.12", "19.13", "19.14", "19.15", "19.16",
+                      "19.2", "19.20", "19.21", "19.22", "19.23", "19.24", "19.25", "19.26", "19.27", "19.28"]
+            runtime: [static, dynamic]
+            runtime_type: [Debug, Release]
```
jgsogo: Maybe this requires a `None` value, meaning that the value will follow the one defined by `build_type`? Otherwise it is explicit, and running `conan install ... -s build_type=Debug` will require explicit `conan install -s build_type=Debug -s runtime_type=Debug` which is what we are trying to avoid.

conans/client/graph/graph_binaries.py
```diff
@@ -339,6 +339,9 @@ def _compute_package_id(self, node, default_package_id_mode, default_python_requ
                                           python_requires=python_requires,
                                           default_python_requires_id_mode=
                                           default_python_requires_id_mode)
+        msvc_compatible = conanfile.info.msvc_compatible()
```
jgsogo: Make it opt-able?

jgsogo (commented):
IIRC, you can install different MSVC versions (different toolsets) in the same VS instance, this conditions some of my comments below:

We need this new compiler, making old packages (`Visual Studio`) consumable with new settings (`msvc`) is something needed for ConanCenter to reuse all the existing binaries, but only because we are using the default `msvc` for each VS version (this might not be true in other scenarios, people is crazy, this automatic compatibility assumes the defaults, but IMO it should be an opt-able, just in case).


 * Going from `msvc` version to VS IDE version requires `vswhere` (and msvc-toolset mapping), I think we should add the scaffolding in this PR, at this moment we can use a dictionary instead of running `vswhere`, but at least we have a FIXME/TODO and a place to write the implementation in the future. 
 * `intel` compiler can use `compiler.base=msvc`, I would add it to this PR too, it should be just a few more lines.
 * besides the tool requested by @SSE4 , I would find useful another one to compare msvc versions (our `tool.Version` is not valid). It needs some ad-hoc logic like `19.2 > 19.14` or `19.1 == 19.14` (under some loose consideration).

conan/tools/cmake/utils.py
memsharded: Yes, probably we need not only vswhere, but also some easy way to configure it, because if vswhere finds VS2017 and VS2019, and both have the v140 toolset for older versions, which one will it pick? Seems a task for the ``config`` feature, where this can be explicit. And if not explicit, then ``vswhere``, but as the explicit configuration is necessary anyway, I would start with it, then follow with vswhere one.

conans/client/conf/__init__.py
memsharded: This was automated. If ``runtime_type`` values is not defined, it will get the ``build_type`` one, which is the expected behavior in most cases.

conans/client/graph/graph_binaries.py
memsharded: Yes, agree.

conan/tools/cmake/utils.py
jgsogo: Does it make any difference? The tools (compiler, linker,...) will be exactly the same, won't they? Can you have more than one instance of the same MSVC installed and detected by `vswhere`? Maybe CMake generates something different (VS migrations), but under the hood everything should be exactly the same... I'd say that in this case the generated binary will be exactly the same. Maybe the config is not needed, we can use it if present, otherwise we can choose any at random (or the latest one). cc/ @SSE4

SSE4: that might be different instances of compilers for VS2019 and VS2017 with the same toolset. they probably might have the same version, but I think they installed and updated intependently.

SSE4:
> * besides the tool requested by @SSE4 , I would find useful another one to compare msvc versions (our `tool.Version` is not valid). It needs some ad-hoc logic like `19.2 > 19.14` or `19.1 == 19.14` (under some loose consideration).

I don't think assuming `19.1 == 19.14` is correct as it will break the transitivity of comparison.

jgsogo:
> > * besides the tool requested by @SSE4 , I would find useful another one to compare msvc versions (our `tool.Version` is not valid). It needs some ad-hoc logic like `19.2 > 19.14` or `19.1 == 19.14` (under some loose consideration).
> 
> I don't think assuming `19.1 == 19.14` is correct as it will break the transitivity of comparison.

I was thinking about the C++ standard support. I agree with you and we can remove this comparison from the API, but we will need to decide something in the context of `check_min_cppstd` tool (I'm thinking about this kind of checks: https://github.com/conan-io/conan-center-index/blob/master/recipes/entt/3.x.x/conanfile.py#L42)

SSE4:
> I was thinking about the C++ standard support. I agree with you and we can remove this comparison from the API, but we will need to decide something in the context of `check_min_cppstd` tool (I'm thinking about this kind of checks: https://github.com/conan-io/conan-center-index/blob/master/recipes/entt/3.x.x/conanfile.py#L42)

that might be not just the C++ standard that matters for users. there is a number of Microsoft-specific extensions that are widely used. there are multiple other standards (such as OpenMP) implemented by the compiler.

conan/tools/cmake/utils.py
memsharded: I don't think the approach of choosing one at random is a good one. Besides, asking information about the installed toolsets is not possible in vswhere: https://github.com/Microsoft/vswhere/issues/151

jgsogo: This is what vswhere does (from that same link): _"[vswhere] finds the installation path (and other basic information) of a version of Visual Studio that meets basic requirements"_ (https://github.com/conan-io/conan/issues/3573#issuecomment-522517346):

```
vswhere [-latest] -requires <component-toolset>
```

The toolset is a requirement, vswhere returns the VS version that satisfies that requirement (VS 2017, VS 2019,...) which is the input we need for the CMake generator.

memsharded: Yeah, I have tried that. The toolset component name that you pass to the ``--requires`` argument is the same (``Microsoft.VisualStudio.Component.VC.Tools.x86.x64``) , and it will map to the default toolset of the different VS versions, not the specific one that you want to discover. So far I have not been able to know which toolsets contain each VS versions, I don't know if it is possible at all.

jgsogo: Each toolset has a different component name (https://github.com/conan-io/conan/issues/3573#issuecomment-522517346): 

>   
>   * MSVC `v141`: `Microsoft.VisualStudio.Component.VC.v141.x86.x64`
>   * MSVC `v142` (14.20): `Microsoft.VisualStudio.Component.VC.14.20.x86.x64`
>   * MSVC `v142` (14.21): `Microsoft.VisualStudio.Component.VC.14.21.x86.x64`
>   * `v140`: `Microsoft.VisualStudio.Component.VC.140`
>   * ...
>   
>   It could be easier, but at least [it is documented](https://docs.microsoft.com/es-es/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019).

memsharded: I have tried them too, but don't seem to work as expected, or at least they are trickier than they look. I didn't succeed in identifying the toolsets I have installed in my different VS versions. I will try again.

memsharded: Results differ from documented (or at least documentation is not very clear).
As in any case, it is necessary to allow configuring the VS/MSBuild instance that is being used to build with a compiler version, let's do the configuration part first, and then we will try to automate to define it automatically.

```
INSTALLED:
    VS 2019: v14.28, v14.16
    VS 2017: v14.16, v14.00
DETECTED
    Microsoft.VisualStudio.Component.VC.Tools.x86.x64 => [VisualStudio.15, VisualStudio.16]
    
    Microsoft.VisualStudio.Component.VC.140 => VisualStudio.15
    Microsoft.VisualStudio.Component.VC.141 => []
    Microsoft.VisualStudio.Component.VC.142 => []

    Microsoft.VisualStudio.Component.VC.14.20 => []
    Microsoft.VisualStudio.Component.VC.14.28 => []
    Microsoft.VisualStudio.Component.VC.14.16 => []

    Microsoft.VisualStudio.Component.VC.14.20.x86.x64 => []
    Microsoft.VisualStudio.Component.VC.14.28.x86.x64 => []
    Microsoft.VisualStudio.Component.VC.14.16.x86.x64 => []

    Microsoft.VisualStudio.Component.VC.v140.x86.x64 => []
    Microsoft.VisualStudio.Component.VC.v141.x86.x64 => VisualStudio.16
    Microsoft.VisualStudio.Component.VC.v142.x86.x64 => []
```

conans/client/graph/graph_binaries.py
memsharded: Done

danimtb (approved):

conan/tools/cmake/generic.py
```diff
@@ -42,7 +42,7 @@ def get_generator_platform(settings, generator):
     if settings.get_safe("os") == "WindowsCE":
         return settings.get_safe("os.platform")
 
-    if (compiler == "Visual Studio" or compiler_base == "Visual Studio") and \
+    if (compiler in ("Visual Studio", "msvc") or compiler_base == "Visual Studio") and \
```
SSE4, line 45:
```suggestion
    if (compiler in ("Visual Studio", "msvc") or compiler_base in ("Visual Studio", "msvc")) and \
```

conan/tools/microsoft/__init__.py
```diff
@@ -2,3 +2,17 @@
 from .toolchain import MSBuildToolchain
 from .msbuild import MSBuild
 from .msbuilddeps import MSBuildDeps
+
+
+def msvc_runtime_flag(conanfile):
+    settings = conanfile.settings
+    compiler = settings.get_safe("compiler")
```
SSE4, line 8:
```suggestion
    compiler = settings.get_safe("compiler.base") or settings.get_safe("compiler")
```

conan/tools/microsoft/__init__.py
```diff
@@ -2,3 +2,17 @@
...
+    runtime = settings.get_safe("compiler.runtime")
```
SSE4, line 9:
```suggestion
    runtime = settings.get_safe("compiler.base.runtime") or settings.get_safe("compiler.runtime")
```

conan/tools/microsoft/__init__.py
memsharded: I'd prefer to condition it to ``intel`` compiler explicitly.

conan/tools/microsoft/__init__.py
```diff
@@ -2,3 +2,17 @@
...
+    if compiler == "Visual Studio":
+        return runtime
+    if compiler == "msvc":
+        runtime_type = settings.get_safe("compiler.runtime_type")
+        if runtime == "static":
```
SSE4: this is a better implementation (from below):
```
        runtime = "MT" if runtime == "static" else "MD"
        if  runtime_type == "Debug":
            runtime = "{}d".format(runtime)
```
also, why not reuse?

conans/client/conf/__init__.py
```diff
@@ -71,6 +71,13 @@
             threads: [None, posix, win32] #  Windows MinGW
             exception: [None, dwarf2, sjlj, seh] # Windows MinGW
             cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]
+        msvc:
+            version: ["19.0",
```
SSE4: are we going to define the previous versions as well (pre-19.0)?

conans/model/info.py
```diff
@@ -575,6 +575,27 @@ def header_only(self):
         self.options.clear()
         self.requires.clear()
 
+    def msvc_compatible(self):
+        if self.settings.compiler != "msvc":
+            return
+
+        compatible = self.clone()
+        version = compatible.settings.compiler.version
```
SSE4, line 585:
```suggestion
        version = compatible.settings.get_safe("compiler.version")
```

conans/model/info.py
```diff
@@ -575,6 +575,27 @@ def header_only(self):
...
+        runtime = compatible.settings.compiler.runtime
```
SSE4, line 586:
```suggestion
        runtime = compatible.settings.get_safe("compiler.runtime")
```

conans/model/info.py
```diff
@@ -575,6 +575,27 @@ def header_only(self):
...
+        runtime_type = compatible.settings.compiler.runtime_type
```
SSE4, line 587:
```suggestion
        runtime_type = compatible.settings.get_safe("compiler.runtime_type")
```

conans/model/info.py
```diff
@@ -575,6 +575,27 @@ def header_only(self):
...
+
+        compatible.settings.compiler = "Visual Studio"
+        version = str(version)[:4]
+        _visuals = {'19.0': '14',
+                    '19.1': '15',
+                    '19.2': '16'}
+        compatible.settings.compiler.version = _visuals[version]
```
SSE4, line 594: may raise, need a sanity check (maybe deserves a test)

conans/model/info.py
```diff
@@ -575,6 +575,27 @@ def header_only(self):
...
+        if self.settings.compiler != "msvc":
+            return
+
+        compatible = self.clone()
+        version = compatible.settings.compiler.version
+        runtime = compatible.settings.compiler.runtime
+        runtime_type = compatible.settings.compiler.runtime_type
+
+        compatible.settings.compiler = "Visual Studio"
+        version = str(version)[:4]
```
SSE4, line 590: may raise, need a sanity check (maybe deserves a test)

conan/tools/microsoft/__init__.py
SSE4: up to you :)

conan/tools/cmake/utils.py
```diff
@@ -48,6 +48,15 @@ def get_generator(conanfile):
 
     compiler = conanfile.settings.get_safe("compiler")
     compiler_version = conanfile.settings.get_safe("compiler.version")
+
+    if compiler == "msvc":
+        version = compiler_version[:4]  # Remove the latest version number 19.1X if existing
```
SSE4, line 56: this line (and the line below) may raise, need sanity checks (and maybe tests)

conans/client/conf/__init__.py
memsharded: Not at the moment, we are starting with VS 2015 for this proposal, and once the functionality is stabilized a bit, we will get the feedback from the community/tribe regarding the supported versions.

conans/model/info.py
memsharded: Nop, at this point, using the ``msvc`` compiler, the ``compiler.runtime`` must be defined, otherwise I want it to raise, not to continue silently. Why would you want/allow this to be undefined here? It is not possible in the settings (it doesn't have a ``None`` value)

conans/model/info.py
memsharded: Sure, this is a preliminary implementation, this needs to be improved.

conan/tools/microsoft/__init__.py
memsharded: Changed.

Premature reuse is like premature optimization. It leads to wrong and obscure abstractions. I prefer stronger encapsulation, and knowing ``msvc_runtime_flag()`` is fully self-contained. Anyway, which other piece of code is mapping the new msvc settings to the runtime flag? I see it nowhere else, it is indeed a new functionality necessary for the new model.

conan/tools/microsoft/__init__.py
memsharded: If this is for the intel compiler, I prefer to wait until the next iteration. For a start, the ``settings.yml`` needs to be changed too, this will not work without changing the settings. And we need to understand how the binary model for intel will work, now that we are defining a fallback from msvc => Visual Studio, because intel compiler was relying on that too, and it might have some collisions.

conan/tools/microsoft/__init__.py
SSE4: the code was from `def msvc_compatible(self):` (info.py).

conans/model/info.py
SSE4: that's very possible `settings.yml` might be modified by the user, either manually, or by running `conan config install`. user may add None value or remove `compiler.runtime` sub-setting completely. IMO code should be reliable for custom settings, as we're allowing them and promote it as a first-class feature (especially for Enterprise users).

conan/tools/microsoft/__init__.py
SSE4: again, that might not be not just Intel compiler, but any custom compiler with custom settings using `msvc` as a `compiler.base`.

conans/model/info.py
memsharded: So far this would be an unsupported case, the ``msvc`` compiler always has a ``runtime``, it is not possible to compile without it. What does a ``None`` value means? What is the use case? It is important to code new features initially more restrictive/limited scope, and learn from use cases, better than be relaxed, allow use cases that we completely ignore and that might have other implications. I strongly prefer to evolve from the basic "canonical" use case, and account for variants and special use cases later.

conans/client/conf/__init__.py
```diff
@@ -497,6 +504,14 @@ def full_transitive_package_id(self):
         except ConanException:
             return None
 
+    @property
+    def msvc_visual_incompatible(self):
+        try:
+            visual_comp = self.get_item("general.msvc_visual_incompatible")
+            return visual_comp.lower() in ("1", "true")
+        except ConanException:
+            return None
```
jgsogo: Here I prefer to return `False` as it is an explicit default value. Returning `None` is like letting the consumer choose any value. Does it make sense?

conans/client/settings_preprocessor.py
```diff
@@ -15,6 +15,14 @@ def _check_cppstd(settings):
     cppstd = settings.get_safe("cppstd")
     compiler_cppstd = settings.get_safe("compiler.cppstd")
 
+    if compiler == "msvc":
+        # This is hardcoded at the moment, because the minimum defined version is 19.0, with c++14
+        # TODO: When older versions of msvc are defined in settings, cppstd can be None
```
jgsogo:
```suggestion
        # TODO: Decide if we want to force the user to provide the 'cppstd' value always
```

jgsogo (approved):

conans/client/settings_preprocessor.py
memsharded: I have better removed the None value in settings leaving: ``cppstd: [14, 17, 20]``. Not defining it will raise an error. Current design was especulative, assuming older VS versions. For the current versions ``cppstd: [14, 17, 20]`` is the correct definition, lets wait to handed this in the settings_preprocessor until we really have the problem because defining older versions.

In that case we might want to add something like ``cppstd: [undefined, 14, 17, 20]`` instead of ``None``, in this way, it is still explicit and doesn't allow not defining it. The internal helpers like ``cppstd_flag`` should account for this, but this can wait.

PR #8208, opened 2020-12-14
Unzip flat directory

### LINKED ISSUE #6479: [feature] Automatic directory management in source unpack support
Debian's `dpkg-source` has a nice feature when unpacking upstream tarballs to ensure that the target directory always has a specific name. It will do the following:

 - Create the target directory that should be created
 - Unpack within that directory
 - If the top-level contents of the archive was just a single directory, move the contents of that directory one level up and remove that directory

The following situations exist then:

 - Archive contains a directory tree in the following form:
    `dir/`
    `dir/CMakeLists.txt`
    `dir/sourcefile.c`
  Calling `dpkg-source -x` would create a directory of the expected name (in Debian's case `package-version`) and that directory would contain the `CMakeLists.txt` file and the `sourcefile.c` file - regardless of the name `dir/` **within** the archive.

 - Archive contains a directory tree in the following form:
   `CMakeLists.txt`
   `sourcefile.c`
   Calling `dpkg-source -x` would create the same result as before, even though the archive doesn't contain a subdirectory.

I think something like that would be interesting for conan, because I see the following pattern very often in recipies:

 - unpack source archive
 - rename directory to `_source_subfolder` (or something otherwise deterministic)

Since the rename step requires the user to know the directory name within the source archive, and projects might change that name between releases, this can add quite a bit of complexity to the conan file.

Since this changes the behavior of unzip / get I would suggest adding a keyword argument `autodir` (default `False`) that introduces this new behavior.

Example usage could be:

    tools.get(url, sha256=hash, destination='_source_subfolder', autodir=True)

This would ensure that the contents of the downloaded archive is available in the directory `_source_subfolder` according to the described algorithm, regardless of how the archive was organized internally.

I could implement this and provide a pull request if this is something you'd be interested in adding to conan.

lasote:
Changelog: Feature: The `tools.get`, `tools.unzip` and `tools.untargz` now accept a new argument `strip_root=True` to unzip moving all the files to the parent folder when all of them belongs to a single folder.
Docs: https://github.com/conan-io/docs/pull/1967

Closes #6479 

When a zip file structure is the following:
```
- folder/file
- folder/file2
- folder/folder2/file3
```
If you use the `flat_folder=True` it will extract:
```
- file
- file2
- folder2/file3
```
- If there is more than one folder in the zip it will raise. 
- If there is a file in the root folder it will raise.

### DIFF
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -110,10 +110,10 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
             zip_info = [zi for zi in zip_info if fnmatch(zi.filename, pattern)]
         if flat_folder:
             names = [n.replace("\\", "/") for n in z.namelist()]
-            common_folder = os.path.commonprefix(names).split("/")[0]
+            common_folder = os.path.commonprefix(names).split("/", 1)[0]
             if not common_folder and len(names) > 1:
                 raise ConanException("The zip file contains more than 1 folder in the root")
-            if len(names) == 1 and len(names[0].split("/")) == 1:
+            if len(names) == 1 and len(names[0].split("/", 1)) == 1:
                 raise ConanException("The zip file contains a file in the root")
 
             for member in zip_info:
@@ -162,10 +162,10 @@ def untargz(filename, destination=".", pattern=None, flat_folder=False):
 
             if flat_folder:
                 names = [n.replace("\\", "/") for n in tarredgzippedFile.getnames()]
-                common_folder = os.path.commonprefix(names).split("/")[0]
+                common_folder = os.path.commonprefix(names).split("/", 1)[0]
                 if not common_folder and len(names) > 1:
                     raise ConanException("The tgz file contains more than 1 folder in the root")
-                if len(names) == 1 and len(names[0].split("/")) == 1:
+                if len(names) == 1 and len(names[0].split("/", 1)) == 1:
                     raise ConanException("The tgz file contains a file in the root")
 
                 for member in members:
--- a/conans/client/tools/files.py
+++ b/conans/client/tools/files.py
@@ -55,7 +55,8 @@ def human_size(size_bytes):
     return "%s%s" % (formatted_size, suffix)
 
 
-def unzip(filename, destination=".", keep_permissions=False, pattern=None, output=None):
+def unzip(filename, destination=".", keep_permissions=False, pattern=None, output=None,
+          flat_folder=False):
     """
     Unzip a zipped file
     :param filename: Path to the zip file
@@ -67,6 +68,7 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
     :param pattern: Extract only paths matching the pattern. This should be a
     Unix shell-style wildcard, see fnmatch documentation for more details.
     :param output: output
+    :param flat: If all the contents are in a single dir, flat that directory.
     :return:
     """
     output = default_output(output, 'conans.client.tools.files.unzip')
@@ -74,7 +76,7 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
     if (filename.endswith(".tar.gz") or filename.endswith(".tgz") or
             filename.endswith(".tbz2") or filename.endswith(".tar.bz2") or
             filename.endswith(".tar")):
-        return untargz(filename, destination, pattern)
+        return untargz(filename, destination, pattern, flat_folder)
     if filename.endswith(".gz"):
         with gzip.open(filename, 'rb') as f:
             file_content = f.read()
@@ -84,7 +86,7 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
     if filename.endswith(".tar.xz") or filename.endswith(".txz"):
         if six.PY2:
             raise ConanException("XZ format not supported in Python 2. Use Python 3 instead")
-        return untargz(filename, destination, pattern)
+        return untargz(filename, destination, pattern, flat_folder)
 
     import zipfile
     full_path = os.path.normpath(os.path.join(get_cwd(), destination))
@@ -103,10 +105,19 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
             pass
 
     with zipfile.ZipFile(filename, "r") as z:
-        if not pattern:
-            zip_info = z.infolist()
-        else:
-            zip_info = [zi for zi in z.infolist() if fnmatch(zi.filename, pattern)]
+        zip_info = z.infolist()
+        if pattern:
+            zip_info = [zi for zi in zip_info if fnmatch(zi.filename, pattern)]
+        if flat_folder:
+            common_folder = None
+            for info in zip_info:
+                _check_common = os.path.normpath(info.filename).split(os.sep)[0]
+                if common_folder is None:
+                    common_folder = _check_common
+                elif _check_common != common_folder:
+                    raise ConanException("The zip file contains more than 1 folder in the root")
+                info.filename = os.sep.join(os.path.normpath(info.filename).split(os.sep)[1:])
+
         uncompress_size = sum((file_.file_size for file_ in zip_info))
         if uncompress_size > 100000:
             output.info("Unzipping %s, this can take a while" % human_size(uncompress_size))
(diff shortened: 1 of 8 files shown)


conans/test/unittests/util/files/flat_extract_test.py
```diff
@@ -0,0 +1,166 @@
+import os
+import tarfile
+import unittest
+
+import six
+
+from client.tools import untargz, unzip
```
memsharded: Broken imports, should start with ``from conan.``

conans/client/tools/files.py
```diff
@@ -103,10 +105,22 @@ def print_progress(_, __):
             pass
 
     with zipfile.ZipFile(filename, "r") as z:
-        if not pattern:
-            zip_info = z.infolist()
-        else:
-            zip_info = [zi for zi in z.infolist() if fnmatch(zi.filename, pattern)]
+        zip_info = z.infolist()
```
SSE4, line 108: I was using `namelist` for `ZipFile` (and `getnames` for the `TarFile` similarly).
then, `os.path.commonprefix` on the returned names list.
if it is dot (`.`), then there are more than one file/folder on the root.

conans/client/tools/files.py
```diff
@@ -139,14 +153,29 @@ def print_progress(_, __):
...
-        if not pattern:
+        if not pattern and not flat_folder:
             tarredgzippedFile.extractall(destination)
         else:
-            members = list(filter(lambda m: fnmatch(m.name, pattern),
-                                  tarredgzippedFile.getmembers()))
+            members = tarredgzippedFile.getmembers()
+            if flat_folder:
+                common_folder = None
+                for member in tarredgzippedFile.getmembers():
```
memsharded: why not ``for member in members``

conans/client/tools/files.py
```diff
@@ -103,10 +105,22 @@ def print_progress(_, __):
...
+            for info in zip_info:
+                dir_split = os.path.normpath(info.filename).split(os.sep)
+                if len(dir_split) == 1:
+                    raise ConanException("The zip file contains a file in the root")
+                _check_common = dir_split[0]
+                if common_folder is None:
+                    common_folder = _check_common
+                elif _check_common != common_folder:
+                    raise ConanException("The zip file contains more than 1 folder in the root")
+                info.filename = os.sep.join(os.path.normpath(info.filename).split(os.sep)[1:])
```
memsharded: isnt ``os.path.normpath(info.filename).split(os.sep)`` = ``dir_split``?

SSE4:
what do you think on returning the flat directory, so it could be chained like:
```
with chdir(get(URL)):
    do_something()
```

lasote:
But this feature removes the directory... why do you want to return it? 
If you are thinking of another feature to return the directory but not "flat" it, I don't think that is the correct way to operate. We should avoid as much as possible to change directories in the recipes and **declare** everything as a recipe layout.

conans/test/functional/command/config_install_test.py
```diff
@@ -72,17 +71,6 @@
 """
 
 
-def zipdir(path, zipfilename):
```
lasote, line 75: Moved to test utils

conans/client/tools/files.py
```diff
@@ -103,10 +105,21 @@ def print_progress(_, __):
...
+            names = [n.replace("\\", "/") for n in z.namelist()]
+            common_folder = os.path.commonprefix(names).split("/")[0]
+            if not common_folder and len(names) > 1:
+                raise ConanException("The zip file contains more than 1 folder in the root")
+            if len(names) == 1 and len(names[0].split("/")) == 1:
+                raise ConanException("The zip file contains a file in the root")
+
+            for member in zip_info:
+                name = member.filename.replace("\\", "/")
+                member.filename = "/".join(name.split("/")[1:])
```
memsharded:
```suggestion
                member.filename = name.split("/", 1)[1]
```

conans/client/tools/files.py
```diff
@@ -103,10 +105,21 @@ def print_progress(_, __):
...
     with zipfile.ZipFile(filename, "r") as z:
-        if not pattern:
-            zip_info = z.infolist()
-        else:
-            zip_info = [zi for zi in z.infolist() if fnmatch(zi.filename, pattern)]
+        zip_info = z.infolist()
+        if pattern:
+            zip_info = [zi for zi in zip_info if fnmatch(zi.filename, pattern)]
+        if flat_folder:
+            names = [n.replace("\\", "/") for n in z.namelist()]
```
memsharded, line 112: Can't we assume that if ``flat_folder`` is given, there should be just 1 element in the root, and it be a directory? No need to check for common_prefix?

conans/client/tools/files.py
```diff
@@ -55,7 +55,8 @@ def human_size(size_bytes):
     return "%s%s" % (formatted_size, suffix)
 
 
-def unzip(filename, destination=".", keep_permissions=False, pattern=None, output=None):
+def unzip(filename, destination=".", keep_permissions=False, pattern=None, output=None,
+          flat_folder=False):
```
memsharded: flat_folder name could be improved?, I would read it as getting rid of all levels of subfolders, not just dropping the first level one

conans/client/tools/files.py
```diff
@@ -139,14 +152,29 @@ def print_progress(_, __):
...
+                names = [n.replace("\\", "/") for n in tarredgzippedFile.getnames()]
+                common_folder = os.path.commonprefix(names).split("/")[0]
+                if not common_folder and len(names) > 1:
+                    raise ConanException("The tgz file contains more than 1 folder in the root")
+                if len(names) == 1 and len(names[0].split("/")) == 1:
+                    raise ConanException("The tgz file contains a file in the root")
+
+                for member in members:
+                    name = member.name.replace("\\", "/")
+                    member.name = "/".join(name.split("/")[1:])
```
memsharded:
```suggestion
                    member.name = name.split("/", 1)[1]
```

conans/client/tools/files.py
lasote: sure, suggestions? `crop_first_folder`, `flat_first`....?

conans/client/tools/files.py
lasote: I think it could be error prone for the user. Because, if you don't check... what do you "flat" exactly? all the first levels in case of several dirs in the root?

conans/client/tools/files.py
```diff
@@ -103,10 +105,21 @@ def print_progress(_, __):
...
-        if not pattern:
-            zip_info = z.infolist()
-        else:
-            zip_info = [zi for zi in z.infolist() if fnmatch(zi.filename, pattern)]
+        zip_info = z.infolist()
+        if pattern:
+            zip_info = [zi for zi in zip_info if fnmatch(zi.filename, pattern)]
+        if flat_folder:
+            names = [n.replace("\\", "/") for n in z.namelist()]
+            common_folder = os.path.commonprefix(names).split("/")[0]
```
memsharded:
```suggestion
            common_folder = os.path.commonprefix(names).split("/", 1)[0]
```

conans/client/tools/files.py
```diff
@@ -139,14 +152,29 @@ def print_progress(_, __):
...
+        if not pattern and not flat_folder:
             tarredgzippedFile.extractall(destination)
         else:
-            members = list(filter(lambda m: fnmatch(m.name, pattern),
-                                  tarredgzippedFile.getmembers()))
+            members = tarredgzippedFile.getmembers()
+
+            if flat_folder:
+                names = [n.replace("\\", "/") for n in tarredgzippedFile.getnames()]
+                common_folder = os.path.commonprefix(names).split("/")[0]
```
memsharded:
```suggestion
                common_folder = os.path.commonprefix(names).split("/", 1)[0]
```

conans/client/tools/files.py
lasote: `ignore_root`?

lasote: `skip_root`?

lasote: `flat_root`?

conans/client/tools/files.py
```diff
@@ -103,10 +105,21 @@ def print_progress(_, __):
...
-            zip_info = [zi for zi in z.infolist() if fnmatch(zi.filename, pattern)]
+        zip_info = z.infolist()
+        if pattern:
+            zip_info = [zi for zi in zip_info if fnmatch(zi.filename, pattern)]
+        if flat_folder:
+            names = [n.replace("\\", "/") for n in z.namelist()]
+            common_folder = os.path.commonprefix(names).split("/")[0]
+            if not common_folder and len(names) > 1:
+                raise ConanException("The zip file contains more than 1 folder in the root")
+            if len(names) == 1 and len(names[0].split("/")) == 1:
```
memsharded:
```suggestion
            if len(names) == 1 and len(names[0].split("/", 1)) == 1:
```

conans/client/tools/files.py
```diff
@@ -139,14 +152,29 @@ def print_progress(_, __):
...
-            members = list(filter(lambda m: fnmatch(m.name, pattern),
-                                  tarredgzippedFile.getmembers()))
+            members = tarredgzippedFile.getmembers()
+
+            if flat_folder:
+                names = [n.replace("\\", "/") for n in tarredgzippedFile.getnames()]
+                common_folder = os.path.commonprefix(names).split("/")[0]
+                if not common_folder and len(names) > 1:
+                    raise ConanException("The tgz file contains more than 1 folder in the root")
+                if len(names) == 1 and len(names[0].split("/")) == 1:
```
memsharded:
```suggestion
                if len(names) == 1 and len(names[0].split("/", 1)) == 1:
```

conans/client/tools/files.py
memsharded: https://unix.stackexchange.com/questions/535772/what-does-strip-components-c-mean-in-tar

Probably ``strip_root``

memsharded (approved):

conans/client/tools/files.py
```diff
@@ -67,14 +68,15 @@ def unzip(filename, destination=".", keep_permissions=False, pattern=None, outpu
     :param pattern: Extract only paths matching the pattern. This should be a
     Unix shell-style wildcard, see fnmatch documentation for more details.
     :param output: output
+    :param flat: If all the contents are in a single dir, flat that directory.
```
SSE4, line 71: to clarify, it always strips only one level?

SSE4 (approved):

conans/client/tools/files.py
lasote: Yes

PR #8213, opened 2020-12-15
Pick the first free port to launch servers on tests

czoido:
Changelog: omit
Docs: omit

This is a little improvement over https://github.com/conan-io/conan/pull/8198
When tests launch servers sometimes the ports collide. It also could interfere with other project's tests in the same ci. This will reduce the probability of this to happen picking the first available port but will not completely solve the problem. It could happen that between the port is picked and it is passed to the server other process took this port but it should rarely happen.

#TAGS: slow


memsharded (approved):

jgsogo (approved):
👍

PR #8216, opened 2020-12-16
Feature/toolchain checkers

memsharded:
Changelog: Omit
Docs: Omit

Do some binary checks in CMake toolchain, extracted from MSBuild one.
Very preliminary, more refactors will follow.

### DIFF
--- a/conans/test/integration/toolchains/test_cmake.py
+++ b/conans/test/integration/toolchains/test_cmake.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import re
 import textwrap
 import time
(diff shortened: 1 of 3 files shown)


uilianries (approved):
I tested on my Windows 10 - With Visual Studio 16 (v142) and looks good.
I also tested with admin permission to validate any possible read/write permission, no errors.

As there are many points to be checked on generated cmake files, I uploaded test_toolchain_win_1_Release here: https://gist.github.com/uilianries/2ab755c6fb758b627262cdeb89308ed5

Issue #1839, opened 2017-10-02
Using conan with clang-cl and CMake on Windows


dutow:
Clang provides the `clang-cl` binary, which is compatible with MSVC's `cl`. It can also be used with the `Ninja` CMake generator, for example, the following way:

```
call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Auxiliary\Build\vcvarsall.bat" x64

set CC=clang-cl
set CXX=clang-cl
set CFLAGS=-m64 -fmsc-version=1910
set CXXFLAGS=-m64 -fmsc-version=1910
cmake ../ -GNinja
```

CMake's internal scripts correctly detect the compiler to be cl-compatible and use the cl style command line parameters in this case.

Conan itself has no knowledge about this compiler, but I assumed that if I compile the dependencies with MSVC (which is the default), I should be able to use them in my CMake project.

While this approach mostly works, the `conanbuildinfo.cmake` script generated by conan detects the compiler as Clang instead of as MSVC, and displays an error, because of the following line:

```
if( (CONAN_COMPILER STREQUAL "Visual Studio" AND NOT CMAKE_CXX_COMPILER_ID MATCHES MSVC) OR
```

While the Ninja generator sets the compiler id to Clang, it also sets the MSVC variable to 1 and sets the MSVC_VERSION correctly to 1910.

I also know that ideally compilers shouldn't be mixed, but I also assume that adding fully working clang/clang-cl support for conan on windows would be a significant work.

As a simpler solution, this check could be more permissive, for example:

1. By changing that line to just `if( (CONAN_COMPILER STREQUAL "Visual Studio" AND NOT MSVC)`
2. By changing the MSVC version checks to use `MSVC_VERSION` instead of the `CMAKE_CXX_COMPILER_VERSION` variable. (currently, this step isn't necessary, because as I described in #1838, the MSVC version checks are incomplete and pass even when they shouldn't)

This change would allow any compiler which claims to be compatible with the given MSVC version to be used. It would be also consistent with how most projects detect that they are compiled with MSVC.

memsharded:
Thanks for your detailed feedback and suggestions.

@SSE4 has been trying recently to use Clang in Windows, but not sure if clang-cl. Any update on this @SSE4 ?

So, @dutow the idea is that binaries compiled with ``clang-cl`` are fully interchangeable and binary-compatible with binaries built with ``cl``?

SSE4:
[...]

PR #8218, opened 2020-12-17
Test agains Clang for ClangCL toolset

### LINKED ISSUE #5896: Using clang-cl with libraries compiled with Visual Studio doesn't work
Conan version: 1.18.1
OS Name: Windows 10
Version: 1903
OS build: 18362.418
Visual Studio 2019 Version 16.3.3

When using clang-cl I want to use libraries that were compiled with Visual Studio since they are binary compatible and building most libraries with clang-cl is basically impossible right now(but thats another issue).

I made a little proof of concept of what I'm trying to do [here](https://github.com/munoah/clang-cl-poc).

Since it uses conan_cmake your default profile should look something like this:
```
[settings]
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.version=16
build_type=Release
```
In Visual Studio go to Open -> CMake and select the CMakeLists.txt
Everything works as expected in the x64-Debug Configuration, however when I switch to x64-Clang-Debug I get:
```
CMake Error at out/build/x64-Clang-Debug/conanbuildinfo.cmake:642 (message):
Incorrect 'Visual Studio'.  Toolset specifies compiler as 'MSVC' but CMake detected 'Clang'
```
I can `set(CONAN_DISABLE_CHECK_COMPILER TRUE)` to disable the whole function, but disabling sanity checks is not nice.
The check that the specified toolset uses the same compiler as the project seems a bit counterproductive, clang-cl and cl are binary compatible (and i guess the check further down is the same problem, maybe someone wants to compile my code with gcc, but the libraries with clang for performance reasons)

My fix for this would be reverting https://github.com/conan-io/conan/issues/2285, but that would presumably break toolchains, but these seem to work different in vs19 anyways

Relates Issues:
https://github.com/conan-io/conan/issues/1839
https://github.com/conan-io/conan/issues/2285

Psy-Kai:
Changelog: Fix: Let CMake generator generate code for checking against "ClangCL" msvc toolset.
Docs: Omit

Without that fix the generated `conanbuildinfo.cmake` would check for compiler and just fail because cmake detected clang but the code would compare with "MSVC".

Fixes #5896
May also fix the #1839 for Visual Studio 2019

### DIFF
--- a/conans/client/generators/cmake_common.py
+++ b/conans/client/generators/cmake_common.py
@@ -583,7 +583,8 @@ class CMakeCommonMacros:
             # If using VS, verify toolset
             if (CONAN_COMPILER STREQUAL "Visual Studio")
                 if (CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "LLVM" OR
-                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang")
+                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang" OR
+                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "ClangCL")
                     set(EXPECTED_CMAKE_CXX_COMPILER_ID "Clang")
                 elseif (CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "Intel")
                     set(EXPECTED_CMAKE_CXX_COMPILER_ID "Intel")
--- a/conans/client/generators/cmake_common.py
+++ b/conans/client/generators/cmake_common.py
@@ -583,8 +583,9 @@ class CMakeCommonMacros:
             # If using VS, verify toolset
             if (CONAN_COMPILER STREQUAL "Visual Studio")
                 if (CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "LLVM" OR
+                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "llvm" OR
                     CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang" OR
-                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "ClangCL")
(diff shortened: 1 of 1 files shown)


conans/client/generators/cmake_common.py
```diff
@@ -583,7 +583,8 @@ class CMakeCommonMacros:
             # If using VS, verify toolset
             if (CONAN_COMPILER STREQUAL "Visual Studio")
                 if (CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "LLVM" OR
-                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang")
+                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang" OR
+                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "ClangCL")
```
memsharded:
```suggestion
                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "llvm" OR
                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "clang" OR
                    CONAN_SETTINGS_COMPILER_TOOLSET MATCHES "Clang")
```

memsharded: The bug here is that MATCHES is case sensitive, so it would apply also to the "llvm" lowercase toolset, adding it too.

memsharded (approved):

memsharded:
Approved, but please take into account these hints for Conan direction:

- The ``cmake`` generator will be replaced by the ``CMakeToolchain`` and ``CMakeDeps`` (forked from ``cmake_find_package_multi`` generator), that are still experimental, but planned to replace it in Conan 2.0
- There is a mechanism to skip the compiler check, just as a tip to unblock: define the CMake variable ``CONAN_DISABLE_CHECK_COMPILER``

PR #8223, opened 2020-12-17
Rename tests build folder

czoido:
Changelog: omit
Docs: omit

After all, it seems that the main reason why pytest was not discovering part of the tests is because it will skip all the folders named "build" or "dist" by default...


memsharded (commented):
Why not changing the pytest ``norecursedirs`` var instead?

https://docs.pytest.org/en/stable/reference.html#confval-norecursedirs

czoido:
Yes, that would be much better, I'll add the `pytest.ini` file in other PR

PR #8230, opened 2020-12-18
temporary allow cmake_paths for CMakeToolchain

memsharded:
Changelog: Fix: Temporarily allow ``cmake_paths`` generator for ``CMakeToolchain``, to allow start using the toolchain for users that depend on that generator.
Docs: Omit

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -14,7 +14,7 @@ from conans.util.files import mkdir
 
 
 def _validate_recipe(conanfile):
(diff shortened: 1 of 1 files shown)


danimtb:
Are we testing the toolchain with this generator? at least to have it there so we are aware of this enabled behavior

memsharded:
> Are we testing the toolchain with this generator? at least to have it there so we are aware of this enabled behavior

The thing is that so far it is very disconnected from the generator functionality. Not to say they are completely independent. The check was only there to promote the usage of the new generators, but we didn't move it fast enough.

danimtb (approved):

Issue #6125, opened 2019-11-25
build_modules : refactor to support multiple build systems


solvingj:
This API...

`self.cpp_info.build_modules = []`

...should be re-implemented. Probably something like this: 

`self.cpp_info.build_modules['cmake'] = []`

This would then allow modules to be passed to any build system, not just cmake:

`self.cpp_info.build_modules['visual_studio'] = []`

The need for this seems obvious (every other member of `cpp_info` is build-system-agnostic, etc), and this re-implementation could be done without being a breaking change. 

Obviously, there are design questions to be answered, for example: 
- How to associate the members with the generators? 
    - Exact match?
    - Starts With?
    - Regex?
    - Something more robust

"exact match" seems intuitive, but is probably naive because we have several cases of competing generators for a given build system.  For cmake, it would be annoying to have to define the same `build_module` file three times (once for each cmake generator). 

"starts with" captures the current CMake implementation which could be described as "all three generators which start with the word `cmake` will include this module". It also captures the second prominent case of multiple generators for a single build system: `visual_studio` and `visual_studio_multi`.  

Of course, corner cases are likely, where someone wants a module to only be included in `visual_studio_multi` for example but not `visual_studio`.  This would be supported by a "starts with" strategy. Unfortunately, the converse would not (including a module in the `visual_studio` generator but not `visual_studio_multi`).

"regex" would support both cases... but regex just seems terrible for this API. 

Feel free to do something more robust, but it will certainly be more complex if you actually want to support highly flexible user-defined-mappings of `build_modules` to generators.  I would probably just do starts-with on the next iteration.

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
This could make sense in the same way that the ``cpp_info.name`` made sense. That one, we are fixing it with a ``cpp_info.names`` field, but this one, I am not sure what the approach would be without breaking. The docs doesn't say this field is experimental, but it should have been labeled as such.  What do you think @danimtb ?

But I still have a question. What is the use case for ``visual_studio`` generators? For ``cmake`` ones, it is more or less clear, the provided modules are injected and made available to the consumer CMakeLists.txt. Typical use case is a macro or function with some utility that helps consuming the package itself.

What would be an example use case for visual studio? How should them be injected to the consumers?

solvingj:
All build systems have the same need.  `.props` files can hold macros and functions exactly like `.cmake` files.

[...]

PR #8232, opened 2020-12-18
Specify build modules by the generator in cpp_info

danimtb:
Changelog: Feature: Specify build modules by the generator in `cpp_info`. Added backwards compatibility for `*.cmake` build modules added at global scope, but not for other file extensions.
Docs: https://github.com/conan-io/docs/pull/1986

- [x] Refer to the issue that supports this Pull Request: closes https://github.com/conan-io/conan/issues/6125


jgsogo (changes requested):
I agree this is more explicit behavior, but it will break existing behavior for recipes that are appending files to `cpp_info.build_modules.append(...)`, will they lose all the build_modules?

conans/client/generators/cmake_multi.py
```diff
@@ -14,6 +14,19 @@ def extend(cpp_info, config):
     if config_info:
         def add_lists(seq1, seq2):
             return seq1 + [s for s in seq2 if s not in seq1]
+
+        def add_dicts(dict1, dict2):
```
memsharded: Maybe this is a bit more idiomatic and efficient:

```python
def add_dicts(d1, d2):
    result = d1.copy()
    for k, v in d2.items():
        existing = result.get(k)
        if existing is None:
            result[k] = v
        else:
            existing.extend((item for item in v if item not in existing))
    return result
```

conans/model/build_info.py
```diff
@@ -35,6 +35,20 @@ def __copy__(self):
         return the_copy
 
 
+def merge_dicts(dict1, dict2):
```
memsharded: Seems the same as the above ``add_dicts``, and also could be shorter and a bit more idiomatic.

conans/model/build_info.py
```diff
@@ -333,7 +348,7 @@ def merge_lists(seq1, seq2):
         self.frameworkdirs = merge_lists(self.frameworkdirs, dep_cpp_info.framework_paths)
         self.libs = merge_lists(self.libs, dep_cpp_info.libs)
         self.frameworks = merge_lists(self.frameworks, dep_cpp_info.frameworks)
-        self.build_modules = merge_lists(self.build_modules, dep_cpp_info.build_modules_paths)
+        self.build_modules = merge_dicts(self.build_modules, dep_cpp_info.build_modules_paths)
```
memsharded, line 397: The thing is that the generators that provide 1 file per dependency ``cmake_find_package(_multi)``, ``MSBuildDeps``, which are the future, do not need an aggregated dictionary at all. Every dependency file will declare its own modules.

Maybe we could provide the new build-modules-per-generator for those generators only, and keep the old ones for the others (so we also don't break)? I can see this might be an ugly UI, but just in case, this could be an idea.

memsharded (commented):
As commented by @jgsogo if we would like to keep the previous "global" build_modules so we don't break to users (even if experimental), I would vote for adding the functionality for new generators only, that do not need an aggregated dict of build_modules, because every file will include its own ones.

conans/client/generators/cmake_multi.py
danimtb: ok, I have to check the order here as it is especially important to keep the behavior

conans/model/build_info.py
danimtb: actually not, this is appending the items in a different order

danimtb:
Added the possibility of reading build_modules as a list and adding the specific `.cmake` modules to all the current CMake generators. This is done for backward compatibility and should be removed in the following releases.

conans/model/build_info.py
```diff
@@ -35,6 +35,53 @@ def __copy__(self):
...
+    the_dict = {"cmake": [], "cmake_multi": [], "cmake_find_package": [],
+                "cmake_find_package_multi": []}
+    for item in build_modules:
+        if item.endswith(".cmake"):
+            for gen in the_dict.keys():
+                the_dict[gen].append(item)
+    return the_dict
+
+
+def merge_lists(seq1, seq2):
```
memsharded: Make this private, maybe even local to ``merge_dicts()`` to not confuse with the 2 other ``merge_lists()`` out there (even if the implementation matches, extracting to a different common module would probably not be necessary/desired.

conans/model/build_info.py
```diff
@@ -35,6 +35,53 @@ def __copy__(self):
...
+
+
+def build_modules_list_to_dict(build_modules):
+    the_dict = {"cmake": [], "cmake_multi": [], "cmake_find_package": [],
+                "cmake_find_package_multi": []}
+    for item in build_modules:
+        if item.endswith(".cmake"):
+            for gen in the_dict.keys():
+                the_dict[gen].append(item)
+    return the_dict
```
memsharded:
```suggestion
    for item in build_modules:
        if item.endswith(".cmake"):
            for val in the_dict.values():
                val.append(item)
    return the_dict
```

conans/model/build_info.py
```diff
@@ -89,8 +136,10 @@ def _filter_paths(self, paths):
     @property
     def build_modules_paths(self):
         if self._build_modules_paths is None:
-            self._build_modules_paths = [os.path.join(self.rootpath, p) if not os.path.isabs(p)
-                                         else p for p in self.build_modules]
+            if isinstance(self.build_modules, list):  # FIXME: This should be just a plain dict
+                self.build_modules = build_modules_list_to_dict(self.build_modules)
```
memsharded: This is confusing: in constructor ``self.build_modules = BuildModulesDict() ``,  but here ``self.build_modules`` is a plain dict? Shouldn't ``build_modules_list_to_dict`` return a ``BuildModulesDict()`` object? And in that case, better make it a class method?

conans/model/build_info.py
```diff
@@ -333,7 +383,7 @@ def merge_lists(seq1, seq2):
         self.frameworkdirs = merge_lists(self.frameworkdirs, dep_cpp_info.framework_paths)
         self.libs = merge_lists(self.libs, dep_cpp_info.libs)
         self.frameworks = merge_lists(self.frameworks, dep_cpp_info.frameworks)
-        self.build_modules = merge_lists(self.build_modules, dep_cpp_info.build_modules_paths)
+        self.build_modules = merge_dicts(self.build_modules, dep_cpp_info.build_modules_paths)
```
memsharded, line 397: Maybe also the ``merge_dicts`` is a ``BuildModulesDict`` method?

conans/model/build_info.py
```diff
@@ -89,8 +136,10 @@ def _filter_paths(self, paths):
     @property
     def build_modules_paths(self):
         if self._build_modules_paths is None:
-            self._build_modules_paths = [os.path.join(self.rootpath, p) if not os.path.isabs(p)
-                                         else p for p in self.build_modules]
+            if isinstance(self.build_modules, list):  # FIXME: This should be just a plain dict
+                self.build_modules = build_modules_list_to_dict(self.build_modules)
+            self._build_modules_paths = {g: [os.path.join(self.rootpath, p) if not os.path.isabs(p)
+                                         else p for p in l] for g, l in self.build_modules.items()}
```
memsharded: avoid the ``l`` variable name

conans/model/build_info.py
```diff
@@ -441,6 +491,17 @@ def _aggregated_paths(self, item):
         setattr(self, "_%s_paths" % item, paths)
         return paths
 
+    def _aggregated_dict_paths(self, item):
```
memsharded: Again, maybe a bit too early generic at this point. Maybe it only needs to be implemented as BuildModulesPath thing, not as a general any-dictionary of paths.

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -98,7 +98,7 @@ def package_info(self):
                     # (the other one should be found by CMAKE_MODULE_PATH in builddirs)
                     builddir = os.path.join("share", "cmake")
                     module = os.path.join(builddir, "my-module.cmake")
-                    self.cpp_info.build_modules.append(module)
+                    self.cpp_info.build_modules["cmake_find_package_multi"].append(module)
```
memsharded: In theory, this test would also pass if not modified, right?

conans/test/unittests/client/generators/cmake_test.py
```diff
@@ -557,37 +557,49 @@ def setUp(self):
         cpp_info = CppInfo(ref.name, "dummy_root_folder1")
         cpp_info.filter_empty = False  # For testing purposes only
         cpp_info.name = ref.name
-        cpp_info.build_modules = ["my-module.cmake"]
```
memsharded, line 560: Will the assignment fail? I see the existing implementation keeps ``append()`` and ``extend()``, but not sure about the ``build_modules = ["..."]`` one. Would this be breaking?

danimtb: No, you have a specific test for that case here https://github.com/conan-io/conan/pull/8232/files#diff-3779db7bbc92daafae06a9e61c096a7970fe9fb3e9345183947061183d7c5612R249

conans/test/functional/generators/cmake_find_package_test.py
```diff
@@ -390,7 +390,7 @@ def package_info(self):
             """)
         # This is a module that defines some functionality
         find_module = textwrap.dedent("""
-            function(conan_message MESSAGE_OUTPUT)
+            function(custom_message MESSAGE_OUTPUT)
```
danimtb, line 391: to avoid using `conan_message()` defined by the generator

memsharded (approved):

conans/model/build_info.py
```diff
@@ -35,6 +35,56 @@ def __copy__(self):
...
+            self.append(item)
+
+    @classmethod
+    def from_list(cls, build_modules):
+        the_dict = BuildModulesDict({"cmake": [], "cmake_multi": [], "cmake_find_package": [],
+                                     "cmake_find_package_multi": []})
+        for item in build_modules:
+            if item.endswith(".cmake"):
+                for val in the_dict.values():
+                    val.append(item)
```
jgsogo: We can rely on the logic written in the other methods, less duplication.

```suggestion
        the_dict = BuildModulesDict()
        the_dict.extend(build_modules)
```

conans/model/build_info.py
```diff
@@ -35,6 +35,56 @@ def __copy__(self):
...
+        return the_dict
+
+    def to_abs_paths(self, rootpath):
+        for generator, values in self.items():
+            self[generator] = [os.path.join(rootpath, p) if not os.path.isabs(p) else p
+                               for p in values]
```
jgsogo: If this `BuildModulesDict` should be a simple dictionary (your comment below), better not to add methods to this class that will be removed. We can just make it a free function.

conans/model/build_info.py
```diff
@@ -35,6 +35,56 @@ def __copy__(self):
...
+        if key not in self.keys():
+            super(BuildModulesDict, self).__setitem__(key, list())
+        return super(BuildModulesDict, self).__getitem__(key)
+
+    def append(self, item):
+        if item.endswith(".cmake"):
+            self["cmake"].append(item)
+            self["cmake_multi"].append(item)
+            self["cmake_find_package"].append(item)
+            self["cmake_find_package_multi"].append(item)
```
jgsogo, line 54: What if I'm already adding files with other extensions? Now they will be lost...

```python
self.cpp_info.build_modules.append("mything.pc")
```

Maybe here it is breaking and we can break (no idea if we already have any recipe with build-modules for other generators, at least in `conan-center-index`), we can raise and say that build-modules at global level are not allowed. wdyt?

conans/client/generators/cmake.py
```diff
@@ -63,11 +62,12 @@ def join_paths_single_var(values):
         self.exelinkflags_list = join_flags(";", cpp_info.exelinkflags)
 
         self.rootpath = join_paths([cpp_info.rootpath])
-        self.build_modules_paths = join_paths([path for path in cpp_info.build_modules_paths if
-                                               path.endswith(".cmake")])
+        self.build_modules_paths = join_paths(cpp_info.build_modules_paths.get(generator_name, ""))
```
jgsogo: empty list instead of empty string?

conans/model/build_info.py
jgsogo: Maybe better a warning (error using conan_v2_mode)

jgsogo (approved):
I would add a note to the changelog about "backward compatibility for `*.cmake` build-modules added at global scope, but not for file extensions".

danimtb:
added to the changelog!

PR #8237, opened 2020-12-19
Feature/new v2

memsharded:
Changelog: Omit
Docs: Omit

Let's have a ``conan new`` v2 mode that creates files using new features like CMakeToolchain, CMakeDeps, etc. Let's make it internal first, for us (for example, I need to use it for the new ``msvc`` compiler testing, that only works with new Toolchains), but it can be made public later


jgsogo (changes requested):
Why not making it a feature available to users? We can distribute this as template files and users will be able to instantiate them with `conan new ... --template conan_v2`. Then we can use it in tests as well.

We said we didn't want to provide a way to activate v2 behavior as it was not going to be maintainable (or even possible).

memsharded:
> Why not making it a feature available to users? We can distribute this as template files and users will be able to instantiate them with conan new ... --template conan_v2. Then we can use it in tests as well.

This is needed in testing of the ``msvc`` compiler, for example, so having it as template files is a bit ugly for testing, because they need to be installed from somewhere else, they will not be in the cache/templates/new by default. The goal of this is to replace the current ones, which use the old generators and build helpers, not an alternative that needs to be installed from an external origin. IMO, when the toolchains are more stabilized and we start to use them in ConanCenter, these new files should become the default ones, and we should start updating the documentation.

memsharded:
> We said we didn't want to provide a way to activate v2 behavior as it was not going to be maintainable (or even possible).

I didn't say that, I commented that maybe the current V2_MODE needed to be decoupled in 2:
   - A "V2_DEPRECATE", that just raises when using deprecated things. This can be defined in conan.conf and be used in production in Conan 1.X, to make sure you are as ready as possible for migration to 2.0
   - A "V2_BEHAVIOR", that might enable some new behavior more aligned with Conan 2.0. This would be only for experimentation, never for production.

I agree that we might drop, at least at the moment, the second one, it might be too much effort for the gain, we will already have Conan 2 in develop2 to experiment the new behavior.

jgsogo:
I can't remember.

I think v2-behavior is a promise we cannot fulfill. We tried to and we failed, there are some v2 features that are not activated with the current v2 variable. Sometimes it is not just an `if`, it adds complexity. Documenting it can create false expectations.

Without that variable, we can provide these new files as templates for everyone to use (testing too, they can be distributed together with Conan, in the same package, no need to install them from somewhere else or write to the user cache) or we add them to the `test/assets/files` folder. Any other alternative?

memsharded:
> Without that variable, we can provide these new files as templates for everyone to use (testing too, they can be distributed together with Conan, in the same package, no need to install them from somewhere else or write to the user cache) or we add them to the test/assets/files folder. Any other alternative?

Yes, this sounds a better approach, to add it as built-in templates, I think it is possible, just didn't think about it. I am going to try this instead.

conans/test/integration/toolchains/test_cmake.py
```diff
@@ -164,6 +164,20 @@ def _run_app(self, build_type, bin_folder=False, msg="App", dyld_path=None):
...
+        client = TestClient()
+        client.run("new pkg/0.1 --template=v2_cmake")
+        conanfile = client.load("conanfile.py")
+        self.assertIn("CMakeToolchain", conanfile)
+        conanfile = client.load("test_package/conanfile.py")
+        self.assertIn("CMakeToolchain", conanfile)
+        cmake = client.load("test_package/CMakeLists.txt")
+        self.assertIn("find_package", cmake)
+        client.run("create .")
+        self.assertIn("pkg/0.1: Hello World Release!", client.out)
```
jgsogo: Maybe we can move this test together with other tests of the `conan new` command (`test/functional/command/new_test.py`)

jgsogo (approved):
LGTM

conans/test/integration/toolchains/test_cmake.py
memsharded: If it doesn't build, yes, sure, I just wanted to check that the package is actually building, etc, including the ``test_package()``, this hasn't been test in the toolchain so far. But as it will be used by other tests soon, yep, lets move it (and remove the actual build)

memsharded: Done

jgsogo (approved):

PR #7716, opened 2020-09-17
Fix #7715, in rpath_flags use host OS instead of build OS to determine -rpath separator

Erlkoenig90:
Fix issue #7715.

Changelog: Bugfix: Change the `rpath_flags` flag to always use the comma separator instead of "=", because the current behaviour causes linker error messages when attempting to cross-compile to Mac OS, and the comma separator is accepted everywhere.
Docs: Omit

- [X] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [X] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [X] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

Issue #7715, opened 2020-09-17
[bug] -rpath flag separator depends on build OS instead of host OS


Erlkoenig90:
### Environment Details (include every applicable attribute)
  * Operating System+version: Linux Mint 20 Ulyana 
  * Compiler+version: clang version 10.0.0, Target: x86_64-apple-darwin19
  * Conan version: 1.29.0 with PR #7695 applied
  * Python version: 3.8.2

### Steps to reproduce (Include if Applicable)
Install [osxcross](https://github.com/tpoechtrager/osxcross) on Linux, create a profile with `os=Macos` and attempt to build `glib` with [this PR](https://github.com/conan-io/conan-center-index/pull/2925) applied, like this:

### Logs (Executed commands with output) (Include/Attach if Applicable)
```sh
conan create . -pr:h x86_64-apple-darwin19-macosx-clang -pr:b default -o glib:with_elf=False
```
results in
```
[111/398] Linking target glib/libglib-2.0.0.dylib
FAILED: glib/libglib-2.0.0.dylib 
x86_64-apple-darwin19-clang  -o glib/libglib-2.0.0.dylib ... lots of stuff ... -Wl,-rpath=/home/niklas.guertler/.conan/data/pcre/8.41/_/_/package/ddd42dd21f101343926e8ee9dff966bdba84b078/lib /home/niklas.guertler/.conan/data/bzip2/1.0.8/_/_/package/d81a2c40ef3ca9d7c9333cd75ac1bcdef9183a77/lib/libbz2.a -Wl,-rpath=/home/niklas.guertler/.conan/data/bzip2/1.0.8/_/_/package/d81a2c40ef3ca9d7c9333cd75ac1bcdef9183a77/lib /home/niklas.guertler/.conan/data/zlib/1.2.11/_/_/package/099680e166ebfb4e7a7b9e0126a86c3d8bf1e7a7/lib/libz.a -Wl,-rpath=/home/niklas.guertler/.conan/data/zlib/1.2.11/_/_/package/099680e166ebfb4e7a7b9e0126a86c3d8bf1e7a7/lib -lgnuintl -liconv -liconv
ld: unknown option: -rpath=/home/niklas.guertler/.conan/data/pcre/8.41/_/_/package/ddd42dd21f101343926e8ee9dff966bdba84b078/lib
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
```
The `libpcre.pc` file created by the `pkg_config` generator looks like this:
```
[...]

PR #8238, opened 2020-12-20
Feature/msvc msbuild

memsharded:
Changelog: Feature: Provide support for ``msvc`` compiler in ``MSBuild`` tools
Docs: https://github.com/conan-io/docs/pull/1991

Follow up of https://github.com/conan-io/conan/pull/8201, to provide support for ``msvc`` compiler in ``MSBuild`` tools

Same initial experimental support, just extending it for ``tools.microsoft`` (it was only for CMake):

- No support for intel yet
- No older visual studio versions. No validation of versions.
- Using the new conf ``core.package_id:msvc_visual_compatible`` in new *conan.cfg*
- Some refactors, decoupling logic of toolsets from the old build helpers, so we can change things without breaking.


danimtb (approved):

conan/tools/microsoft/toolchain.py
```diff
@@ -49,20 +51,59 @@ def generate(self):
...
+
+    @staticmethod
+    def _runtime_library(settings):
+        compiler = settings.compiler
+        runtime = settings.get_safe("compiler.runtime")
+        if compiler == "msvc":
+            build_type = settings.get_safe("build_type")
+            if build_type != "Debug":
+                runtime_library = {"static": "MultiThreaded",
+                                   "dyanmic": "MultiThreadedDLL"}.get(runtime, "")
```
czoido:
```suggestion
                                   "dynamic": "MultiThreadedDLL"}.get(runtime, "")
```

conan/tools/microsoft/toolchain.py
```diff
@@ -49,20 +51,59 @@ def generate(self):
...
+            else:
+                runtime_library = {"static": "MultiThreadedDebug",
+                                   "dyanmic": "MultiThreadedDebugDLL"}.get(runtime, "")
```
czoido:
```suggestion
                                   "dynamic": "MultiThreadedDebugDLL"}.get(runtime, "")
```

czoido (approved):

conan/tools/microsoft/toolchain.py
memsharded: Good catch! 

I have added an integration test, much faster than functional ones, that is capturing this error. So far the effort was using only functional tests for the big picture working, but it is needing also some integration ones.

conans/test/functional/generators/pkg_config_test.py
```diff
@@ -138,7 +136,7 @@ def package_info(self):
         pc_path = os.path.join(client.current_folder, "MyLib.pc")
         self.assertTrue(os.path.exists(pc_path))
         pc_content = load(pc_path)
-        self.assertIn("-Wl,-rpath=\"${libdir}\"", pc_content)
+        self.assertIn("-Wl,-rpath,\"${libdir}\"", pc_content)
```
jgsogo, line 139: These changes, was it a bug? Maybe this is worth an entry in the changelog.

memsharded: There was another PR for this issue, not this one, I am surprised this is here, need to check.

memsharded: https://github.com/conan-io/conan/issues/7715

memsharded: The PR https://github.com/conan-io/conan/pull/7716/files contains this change, and this was merged to develop, and this PR merged develop, so I think it might be a glitch of Github diff

PR #8243, opened 2020-12-21
Recipe Layouts: Step 0

lasote:
Changelog: omit
Docs: omit

- Pure refactor, no functional changes
- To check if the layout model is looking good?
- Pending: Agree on next steps (Pull requests):
  - Next step 2 (functional changes):
         - Local "install" method   
         - Local "source" method
         - Local "build" method
         - Local "package" method
         - Local "export-pkg" method
         - Build helpers and new toolchains
         - Editable packages
  - Next step 3:
          - Design the "package" & "deploy" "automatic" mechanism based on the layout.
          - ...
  - Next step 4 (adopt):
     - Deprecate access to folders directly:  Deprecate `self.source_folder` with `conan_v2_behavior` in favor of `self.layout.source_folder` 
     - Change all internal calls (including tests to access the folders with `self.layout.xxxx_folder`


conans/model/layout.py
```diff
@@ -0,0 +1,90 @@
...
+            return self._base_source_folder
+
+        return os.path.join(self._base_source_folder, self.source.folder)
+
+    def set_base_source_folder(self, folder):
+        self._base_source_folder = folder
+
+    @property
+    def build_folder(self):
+        print("Build folder: {}".format(self._base_build_folder))
```
memsharded: Remove prints

conans/model/layout.py
```diff
@@ -0,0 +1,90 @@
...
+        self.install.folder = ""
+
+        self.build = _LayoutEntry()
+        self.build.folder = ""  # Where the software is built (relative to _base_build_folder)
+        self.build.libdirs[""] = ["*.so", "*.so.*", "*.a", "*.lib", "*.dylib"]
+        self.build.bindirs[""] = ["*.exe", "*.dll"]
+
+        self.source = _LayoutEntry()
+        self.source.folder = ""
+        self.source.includedirs[""] = ["*.h", "*.hpp"]
```
memsharded: Maybe *.hxx too

conans/model/layout.py
```diff
@@ -0,0 +1,90 @@
+import os
+
+
+class _LayoutEntry(object):
+
+    def __init__(self):
+        self.folder = ""
+        self.includedirs = {}
```
memsharded: The model dict {"folder": "file patterns list"} seems a bit weird to me. I think it would be more explicit to have:

- Lists of folders
- Patterns ``lib_patterns``, ``bin_patterns`` with lists of files

conans/model/layout.py
```diff
@@ -0,0 +1,90 @@
...
+        self.source = _LayoutEntry()
+        self.source.folder = ""
+        self.source.includedirs[""] = ["*.h", "*.hpp"]
+
+        self.package = _LayoutEntry()  # Where the artifacts are installed
+        self.package.includedirs["include"] = ["*"]
+        self.package.bindirs["bin"] = ["*"]
+        self.package.libdirs["lib"] = ["*"]
+
+    def __str__(self):
```
memsharded: AS this seems apparently for debugging purposes, I would convert it to a ``__repr__`` and include the class name in it.

memsharded (commented):
Looking good, I like where this is going.

conans/model/layout.py
lasote: Check now

conans/model/layout.py
```diff
@@ -0,0 +1,103 @@
...
+class Layout(object):
+    def __init__(self):
+
+        self._base_source_folder = None
+        self._base_build_folder = None
+        self._base_install_folder = None
+        self._base_package_folder = None
+
+        self.install = _LayoutEntry()
+        self.install.folder = ""
```
memsharded: Already the default

conans/model/layout.py
```diff
@@ -0,0 +1,103 @@
...
+
+        self.source = _LayoutEntry()
+        self.source.folder = ""
+        self.source.includedirs = [""]
+        self.source.includepatterns = ["*.h", "*.hpp", "*.hxx"]  # To be packaged
+
+        self.build = _LayoutEntry()
+        self.build.folder = ""  # Where the software is built (relative to _base_build_folder)
```
memsharded: Already the default

memsharded (approved):
This is looking great. I still have some questions about ``self.package.includepatterns = ["*"]  # To be deployed``, and how that deploy will work using these layouts, but I think this should wait until we see how they are used in further PRs.

I am not sure what will be the release flow of this feature, if we should release this part first (should it be merged to develop already?) and see if by accident we are introducing some regressions, that would be easier to revert. Lets talk about this.

lasote:
I removed all unused code and will introduce it again in following PRs

PR #8244, opened 2020-12-21
Fix tests to run in pytest

czoido:
Changelog: omit
Docs: omit

Getting closer to have the same results in pytest and nose... some tests designed for windows where running and passing in other platforms, skipping them...

### DIFF
--- a/conans/test/unittests/client/generators/visual_studio_multi_test.py
+++ b/conans/test/unittests/client/generators/visual_studio_multi_test.py
@@ -18,7 +18,6 @@ from conans.test.utils.mocks import TestBufferConanOutput
 from conans.test.utils.test_files import temp_folder
 
 
-@attr('visual_studio')
 class VisualStudioMultiGeneratorTest(unittest.TestCase):
 
     @parameterized.expand([(False, ), (True, )])
--- a/conans/test/unittests/client/generators/visual_studio_multi_test.py
+++ b/conans/test/unittests/client/generators/visual_studio_multi_test.py
@@ -1,4 +1,5 @@
 import os
+import platform
 import unittest
 
 import pytest
@@ -19,6 +20,7 @@ from conans.test.utils.test_files import temp_folder
 
 @attr('visual_studio')
 @pytest.mark.tool_visual_studio
+@unittest.skipUnless(platform.system() == "Windows", "Requires Windows")
 class VisualStudioMultiGeneratorTest(unittest.TestCase):
 
     @parameterized.expand([(False, ), (True, )])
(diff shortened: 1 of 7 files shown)


conans/test/unittests/client/generators/visual_studio_multi_test.py
```diff
@@ -19,6 +20,7 @@
 
 @attr('visual_studio')
 @pytest.mark.tool_visual_studio
+@unittest.skipUnless(platform.system() == "Windows", "Requires Windows")
```
memsharded: The truth is that this test is a unittest, that shouldn't rely on the current platform. Just taking some hardcoded settings and outputting an xml file, should be completely platform independent (and it was passing in the past, right?) 

What is the difference now? What is failing?

conans/test/unittests/util/build_sln_command_test.py
```diff
@@ -22,6 +23,7 @@
 
 @attr('visual_studio')
 @pytest.mark.tool_visual_studio
+@unittest.skipUnless(platform.system() == "Windows", "Requires Windows")
```
memsharded: Same as above?

memsharded (approved):
I am fine with changing this, just I wouldn't mind to understand why pytest is failing for these cases.

conans/test/unittests/client/generators/visual_studio_multi_test.py
czoido: It's because we have pytest configured to skip tests that try to run when a marked tool is missing, as it was marked `@pytest.mark.tool_visual_studio` it was skipped for pytest but ran with nose:
https://github.com/conan-io/conan/blob/639450053a0359af5a757b3617a82a2ece2f7cf1/conans/test/conftest.py#L53-L56

memsharded: Oh, I see, then isn't it possible that the solution would be remove the ``@pytest.mark.tool_visual_studio``? As the test doesn't need Visual Studio installed at all. Do we really want that this unittest only run on Windows? 

Same might happen to every other generator and toolchain. Do we want to run ``pkg-config`` generator only on systems that have it installed?

czoido:
Following the conversation we had, I'm removing the skips and all pure unit tests will run in all platforms.
Also I have removed the attr visual_studio as it seems that it was not used.

memsharded (approved):

conans/test/unittests/client/generators/visual_studio_multi_test.py
jgsogo: This is exactly my main concern with adding marks to tests and use them to skip tests when the mark is not used... that a wrong mark will prevent the test from running. Every test should run unless something is added to the command line to skip it, not the other way around.

jgsogo (approved):
Removing marks is always ok, it can only lead to false-negatives (safe side), which is something safe in a test-suite.

PR #8246, opened 2020-12-22
Recipe Layouts: Step 1 (local methods + cache)

lasote:
Changelog: omit
Docs: omit

Based on #8243

- Local methods following the layouts and tests demonstrating the behavior
- Cache following layouts with a test


conan/tools/cmake/cmake.py
```diff
@@ -66,13 +66,13 @@ def configure(self, source_folder=None):
         if not self._conanfile.should_configure:
             return
 
-        source = self._conanfile.source_folder
+        source = self._conanfile.layout.source_folder
```
memsharded: Wasn't the initial idea of the PR0 that this is not necessary? Most Conan usage can still be done via ``self._conanfile.source_folder``?

memsharded (commented):
Not sure about the strategy, maybe instead this big refactor we want to focus on actually using it in some specific places? Maybe just in the new toolchains?

conan/tools/cmake/cmake.py
lasote: > Wasn't the initial idea of the PR0 that this is not necessary?

No, the idea was to not break, but I thought that we will promote the usage of the `layout` instead of the conanfile directly (Conan 2.0). That's why the "getters" in this PR have the deprecated function.

lasote:
> Not sure about the strategy, maybe instead this big refactor we want to focus on actually using it in some specific places? Maybe just in the new toolchains?

Ok! I'll base future PRs on the 0 instead of this one.

conans/client/conan_api.py
```diff
@@ -776,14 +776,16 @@ def package(self, path, build_folder, package_folder, source_folder=None, instal
...
-        package_folder = _make_abs_path(package_folder, cwd, default=default_pkg_folder)
 
-        if package_folder == build_folder:
-            raise ConanException("Cannot 'conan package' to the build folder. "
-                                 "--build-folder and package folder can't be the same")
         conanfile = self.app.graph_manager.load_consumer_conanfile(conanfile_path, install_folder,
                                                                    deps_info_required=True)
+
+        # If the layout declared a package folder follow it, do not invent a "package/"
+        # folder by default
```
memsharded: Or make the layout have a default always, never have a "package" harcoded here?

conans/test/functional/layout/in_cache_test.py
```diff
@@ -0,0 +1,12 @@
+import textwrap
+import os
+
+import unittest
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+class CacheWithLayoutTest(unittest.TestCase):
```
memsharded: You can start using pytest tests, they are simpler to write and more elegant

conans/test/functional/layout/local_commands_test.py
```diff
@@ -0,0 +1,224 @@
...
+    def configure(self):
+        self.layout.install.folder = "my_install"
+        self.layout.source.folder = "my_source"
+
+    def source(self):
+        tools.save("downloaded.h", "bar")
+    """
+        client.save({"conanfile.py": conan_file})
+        client.run("install . -if=common")
+        client.run("source . -if=common/my_install -sf=common")
```
memsharded: Don't we want to deprecate ``-sf``? Maybe it could raise if we are using layouts?

(same for ``-bf``, etc)

lasote: Yes, I would like to deprecate them because they won't be necessary but until then the "sense" is the "base" folder.

Probably not easy to raise because of two reasons:
- Not sure if when you have available the conanfile (and you can check if you have defined the layout) you lose the trace if the directory came from the argument `-sf` etc
- If, for example, we decide that the default "layout.package.folder" in the layout is "package", then is not that easy to check if the value has been modified or not, so it could be tricky.

I would prefer to just introduce the deprecation mechanism at the `conan_api` level when the arguments are declared but in the meantime that is the behavior, they should have.

conans/test/functional/layout/in_cache_test.py
lasote: I have to check how it looks

conans/client/conan_api.py
lasote: I might feel more comfortable with an empty default layout, easier to recognize if someone has started using the feature. Check my comment about the `-bf` below.

conans/client/installer.py
```diff
@@ -186,50 +189,50 @@ def build_package(self, node, keep_build, recorder, remotes):
...
-                    node.prev = prev
-                    log_file = os.path.join(build_folder, RUN_LOG_NAME)
-                    log_file = log_file if os.path.exists(log_file) else None
-                    log_package_built(pref, time.time() - t1, log_file)
-                    recorder.package_built(pref)
-                except ConanException as exc:
-                    recorder.package_install_error(pref, INSTALL_ERROR_BUILDING, str(exc),
-                                                   remote_name=None)
-                    raise exc
+            self._output.info('Building your package in %s' % base_build_folder)
```
lasote: No changes, just removed the `with tools.chdir(build_folder):` and untabbed

conans/client/conan_api.py
```diff
@@ -776,14 +776,16 @@ def package(self, path, build_folder, package_folder, source_folder=None, instal
...
+        default_pkg_folder = os.path.join(build_folder, "package") \
+            if not conanfile.layout.package.folder else None
+        package_folder = _make_abs_path(package_folder, cwd, default=default_pkg_folder)
```
memsharded, line 785: This is confusing. ``conanfile.layout.package.folder`` is not really used, only for the ``if``. Shouldn't be used? Maybe it is not covered by tests?

memsharded: Now I realize, the ``conanfile.layout.package.folder`` will be used later, internally. Maybe just a comment that ``package_folder`` will be ``None``

conans/test/functional/layout/in_cache_test.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_cache_in_layout():
+    """The layout in the cache is used too, always relative to the "base" folders that the cache
+    requires.
+    """
+    client = TestClient()
+    client.save({"conanfile.py": GenConanfile()})
+    client.run("create . base/1.0@")
+
+    conan_file = str(GenConanfile().with_import("from conans import tools").with_import("import os").
```
memsharded: ``import os `` seems unused

conans/test/functional/layout/in_cache_test.py
```diff
@@ -0,0 +1,63 @@
...
+    no_copy_sources = True
+
+    def configure(self):
+        self.layout.source.folder = "my_sources"
+        self.layout.build.folder = "my_build"
+        self.layout.package.folder = "my_package"
+
+    def source(self):
+        self.output.warn("Source folder: {}".format(self.source_folder))
+        tools.save("source.h", "foo")
```
memsharded: source.h unused?

conans/test/functional/layout/in_cache_test.py
```diff
@@ -0,0 +1,63 @@
...
+    pref = PackageReference(ref, "58083437fe22ef1faaa0ab4bb21d0a95bf28ae3d")
+    sf = client.cache.package_layout(ref).source()
+    bf = client.cache.package_layout(ref).build(pref)
+    pf = client.cache.package_layout(ref).package(pref)
+
+    csf = os.path.join(sf, "my_sources")
+    cbf = os.path.join(bf, "my_build")
+    cpf = os.path.join(pf, "my_package")
+    # Check folders match with the declared by the layout
+    assert "Source folder: {}".format(csf in client.out)
```
memsharded: This assert seems wrong? Check parenthesis
```
assert "Source folder: {}".format(csf) in client.out
```

memsharded (commented):
Looking good.
I think I would like one test that the same conanfile is good for both cache and local commands. Maybe share a fixture. I will play with the PR and maybe contribute something if I can find some time.

conans/test/functional/layout/in_cache_test.py
lasote: Yes it is, at: `os.path.join`

conans/client/conan_api.py
lasote: I have changed it to be more clear

lasote:
I've added a test with a fixture but not sure that what you meant because there are other "local flow" tests at `local_commands_test.py`.

conans/test/integration/layout/test_local_commands.py
```diff
@@ -0,0 +1,269 @@
...
+    #        ready
+    client = TestClient()
+    conan_file = str(GenConanfile().with_import("from conans import tools"))
+    conan_file += """
+    no_copy_source = True
+
+    def configure(self):
+        self.layout.source.folder = "my_source"
+        self.layout.build.folder = "my_build"
+        self.layout.package.folder = "my_package"
```
memsharded: This is confusing, or at least unexpected. When I create the package in the local cache, it puts the files under a "my_package" folder in the cache, so the files are not found (the set(CONAN_LIB_DIRS_MYPKG ) is empty in the ``conanbuildinfo.cmake``.

I am not sure how this folder would work in the cache, or what benefit/feature it will implement.

lasote: I pushed some changes fixing the location of the generated conaninfo and testing it.
About changing the package folder in the layout, a new the test, by modifying the `cpp_info`, is testing that the package is "usable". That part will be more automatic when the info belongs to the package.

The benefit is more clearly seen with the editable package I think. But nevertheless, you could use the `self.in_local_cache` to declare a different layout in the cache. I added a test.

lasote: Check last commit (https://github.com/conan-io/conan/pull/8246/commits/95ec8bce6cb2da0bdaffb211dd80bb9bafb7b4d5) where the cache, by default, is not following the layout for the package. I'm not convinced either about that. WDYT?

lasote:
Removed `package` layout after thinking about the local package method, which probably is not very useful anymore, if someone wants to try package something probably it is better to use `export-pkg`.

conans/test/integration/layout/test_in_cache.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def package(self):
+        self.output.warn("Package folder: {}".format(self.package_folder))
+        tools.save(os.path.join(self.package_folder, "LICENSE"), "bar")
+        self.copy("*.h", dst="include")
+        self.copy("*.lib", dst="lib")
+
+    def package_info(self):
+        # This will be easier when the layout declares also the includedirs etc
+        self.cpp_info.includedirs = [os.path.join(self.package_folder, "include")]
```
memsharded: Not sure about this. The ``includedirs`` was always relative to the package_folder, can't we keep it without this?

memsharded (approved):
Looking good, this could be merged, please check the comment about cpp_info.includedirs

conans/test/integration/layout/test_in_cache.py
lasote: Sure, now that there is no `package` layout, so that is not necessary.

memsharded (approved):

PR #8250, opened 2020-12-22
Moving unittests that used tools to functional

czoido:
Changelog: omit
Docs: omit

This is the first step in the tests reorganization, under unittests there should be no tests that run any external tools, just pure Python.


memsharded (approved):
Nice!

PR #8260, opened 2020-12-24
Fix skipif condition

czoido:
Changelog: omit
Docs: omit

fix develop

The [test fails in PY2](https://github.com/conan-io/conan/commit/e498cefbc8e89308d2f0570707d51a3405401368), when passing from skipunless to skipif the condition was not correctly ported.


jgsogo (approved):
👍 

I've seen before the changes it was `@unittest.skipUnless(platform.system() == "Windows" and six.PY3, "Requires MSBuild")`

Issue #8113, opened 2020-11-25
[question] Does conan support Macos on ARM (ARM on Apple desktop) with xcode >=12


madebr:
Today, we received [this bug report of an Apple developer](https://github.com/conan-io/conan-center-index/issues/3690) creating a x86_64 executable instead of an arm when wanting to natively building an arm application.

I think this is because `clang` builds x86_64 executabled by default and conan does not add the correct `CFLAGS/CXXFLAGS/LDFLAGS`.

I think conan needs to add the correct flags.
After googling a bit I found the following page: https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary

The example shows the following lines:
```
x86_app: main.c
    $(CC) main.c -o x86_app -target x86_64-apple-macos10.12
arm_app: main.c
    $(CC) main.c -o arm_app -target arm64-apple-macos11
```
So maybe `-target {ARCH}-apple-os{VERSION}` must be added (automatically)?

I don't own anything apple, so there's nothing more I can tell or debug.

SSE4:
I think we do automatically adjustments for CMake build helper, but not for the AutoToolsBuildEnvironment helper right now, that's the problem. there are two flags are usually needed for iOS/watchOS/tvsOS : `--isysroot` and `--arch`. I believe that's the same for ARM Macos.

theodelrieu:
I've built our entire project on macOS ARM, and here's the main problem I've encountered:

## No way to differentiate armv8 iPhone / Simulator

Up until now, Conan Apple helpers made the distinction based on `arch`. If there was `x86`, it was a simulator. Simple.

This is over now, because it can also be `armv8`. The workaround I've used is simply providing the `sdk` parameter in `tools.XCRun`.

But it did not help when I tried distributing simulator packages, because there is no setting that models the targeted XCode SDK, thus the package IDs will be the same.

I propose to add a `self.settings.os.sdk` subsetting, when `os == "iOS"`, with values `[None, iPhoneOS, iPhoneSimulator]`.

`None` will be valid only when `arch != "armv8"`.

SSE4:
@theodelrieu are arm64 simulator and arm64 iPhone binaries actually different?
I suspect arm64 simulator might be able to run the exactly same binaries as iPhone runs, so it's very possible they should render into the same package id.
maybe if you could try to build some dummy .a files, we can compare

theodelrieu:
@SSE4 No they are different, you cannot link both together, the error looks like:

[...]

PR #8263, opened 2020-12-25
Feature: add os.sdk sub-settings for Apple

SSE4:
related: #8113 

- add `os.sdk` sub-setting for Apple platforms (iOS/watchOS/tvOS/macOS)
- select correct SDK in `apple_sdk_name` based on `os.sdk`
- return correct `apple_deployment_target_flag` for simulators
- as now we have to accept `os.sdk` in tools, they were changed to accept `ConanFile` as a first argument
- `conan_v2_behavior` issued for old syntax (individual settings values instead of `ConanFile`)

it makes it possible to distinguish ARMv8 iPhone and ARMv8 iPhone Simulator (M1) binaries (they are really different, incompatible binaries).

/cc @theodelrieu 

Changelog: Feature: Add os.sdk sub-settings for Apple
Docs: https://github.com/conan-io/docs/pull/1981


conans/test/unittests/util/apple_test.py
```diff
@@ -44,22 +37,28 @@ def test_to_apple_arch(self):
         self.assertIsNone(tools.to_apple_arch('mips'))
 
     def test_apple_sdk_name(self):
-        self.assertEqual(tools.apple_sdk_name(FakeSettings('Macos', 'x86')), 'macosx')
```
memsharded: If the intent (it should be) is not to break existing calls, leave these tests unmodified, and exclusively add new tests.

memsharded (changes requested):
I think it is good, and could be merged once the tests are changed, to leave the old ones, making sure we are not breaking current behavior.

conans/test/unittests/util/apple_test.py
SSE4: okay, sure, I'll change

memsharded (commented):
Also, you could leave ``settings`` as argument for this functionality. You are doing 2 different things in this PR:
- The os.sdk functionality
- An unrelated change to allow passing the conanfile as argument

All tools new v2 mode will be moved to ``conan.tools`` new space, and imported from there explicitly, users will need to change their recipes to use the new tools, conanfile as argument can be enforced there.

SSE4:
@memsharded the reason for the change is that most apple tools accept individual settings values. right now, it turns out it's not enough, and complete settings context is needed.
the process might be done in 2 phases, tools might be changed to accept `settings`, then, at some later point, they might be changed to accept `conanfile` instance. IMO it would be a nightmare for recipe authors to change all recipe(s) twice, first time to accept `settings`, second time to accept `conanfile`. I personally would prefer if I only need to do such a transition once.

conans/client/tools/apple.py
```diff
@@ -50,12 +91,29 @@ def apple_deployment_target_env(os_, os_version):
...
+        os_sdk = conanfile.settings.get_safe("os.sdk")
+        arch = conanfile.settings.get_safe("arch")
+    else:
+        conan_v2_behavior('first argument of "apple_deployment_target_flag" has to be ConanFile,'
+                          'use "tools.apple_deployment_target_flag(self)" instead.')
+
+    if not os_sdk:
+        os_sdk = _guess_apple_sdk_name(os_, arch)
+    if os_sdk:
+        os_sdk = os_sdk.lower()
```
danimtb:
```suggestion
    os_sdk = _guess_apple_sdk_name(os_, arch) if not os_sdk else os_sdk.lower()
```

danimtb (changes requested):
I'd rather not change the apple tools that currently don't need to use the conanfile. In this case, none of them need to change to accept a ConanFile object in order to better support the new `os.sdk` subsetting. IMO, let's just add the new parameters to `apple_deployment_target_flag` and leave the other tools for a future discussion

SSE4:
@memsharded @danimtb I have updated according to your suggestions

danimtb (approved):

SSE4:
~~@danimtb any idea why `virtualenv` might be MIA?~~

conans/client/tools/apple.py
```diff
@@ -39,6 +35,17 @@ def apple_sdk_name(settings):
...
 
 
+def apple_sdk_name(settings):
+    """returns proper SDK name suitable for OS and architecture
+    we're building for (considering simulators)"""
+    arch = settings.get_safe('arch')
+    os_ = settings.get_safe('os')
+    os_sdk = settings.get_safe('os.sdk')
+    if os_sdk:
+        return os_sdk.lower()
```
memsharded: If in every single place the sdk is used is lowercase, wouldn't it be better to define them as lowercase directly in the settings?

memsharded (approved):
LGTM, please fix conflict and comment and it can be merged for 1.33.

conans/client/tools/apple.py
SSE4: I'd personally prefer always lower-case as it's less error-prone. @theodelrieu  WDYT?

theodelrieu: @SSE4 Agreed, lowercase looks good.

SSE4: changed to lowercase

PR #8281, opened 2021-01-04
fix per-package settings exact match for packages without user/channel

memsharded:
Changelog: BugFix: Fix per-package settings exact match for packages without user/channel.
Docs: Omit

Could this be breaking in some way? Please review.


jgsogo (approved):
Ok, given how it works now, we need this extra _exact_ match.

--- 

Thinking about Conan v2.0, IMO, we should use here only the name of the package. Is there a use-case where you want to modify something (in the profile) only for a certain version?

czoido (approved):

PR #8298, opened 2021-01-07
Move more tests that use tools from integration to functional

czoido:
Changelog: omit
Docs: omit

Continue moving tests and adding tool marks.


memsharded (approved):

conans/test/functional/build_helpers/meson_test.py
```diff
@@ -10,6 +10,7 @@ class MesonTest(unittest.TestCase):
 
     @pytest.mark.skipif(platform.system() != "Windows", reason="Needs windows for vcvars")
     @pytest.mark.visual_studio
+    @pytest.mark.tool_meson
```
memsharded:
```suggestion
```

memsharded: It is good to move it to ``functional``, as it requires Visual Studio installed, for the vcvars, but it doesn't need Meson installed, as the calls to meson are being captured by the ``def run()`` override.

memsharded (approved):

PR #8299, opened 2021-01-07
Fix some tests names and autotools v2 tests

czoido:
Changelog: omit
Docs: omit

Conan v2 _build_helpers_ tests never run because of the naming. Also the AutotoolsBuildHelperTestCase was failing because configure.ac and Makefile.am were missing.


conans/test/functional/conan_v2/build_helpers/test_autotools.py
```diff
@@ -0,0 +1,58 @@
...
+import platform
+
+import pytest
+
+from conans.test.utils.conan_v2_tests import ConanV2ModeTestCase
+
+
+@pytest.mark.skipif(platform.system() != "Linux", reason="Requires make")
+@pytest.mark.tool_autotools
+class AutotoolsBuildHelperTestCase(ConanV2ModeTestCase):
```
memsharded: I think it is much better to simplify, something like:

```diff
-    @pytest.mark.skipif(platform.system() != "Linux", reason="Requires make")
+    #@pytest.mark.skipif(platform.system() != "Linux", reason="Requires make")
     def test_no_build_type(self):
         t = self.get_client()
         conanfile = textwrap.dedent("""
@@ -17,15 +16,16 @@ class AutotoolsBuildHelperTestCase(ConanV2ModeTestCase):
             class Pkg(ConanFile):
                 settings = "os", "arch", "compiler"
                 def build(self):
-                  autotools = AutoToolsBuildEnvironment(self)
-                  autotools.configure()
-                  autotools.make()
+                    autotools = AutoToolsBuildEnvironment(self)
+                    autotools.make()
+                def run(cmd, *args, **kwargs):
+                    self.output.info("RUNNING CMD!: {}!!".format(cmd))
```

- The ``skipif`` can be removed, this would be more basic test, it doesn't require any tool installed at all.
- The trick of overriding the ``def run()`` method can be very useful for testing these things
- The test can be moved to ``integration`` folder then. 

We need to try to simplify tests, and avoid **functional** tests as much as possible.

memsharded (changes requested):

PR #8301, opened 2021-01-07
Bugfix in MakeGenerator

memsharded:
Changelog: Bugfix: Fix bug in definition of ``ROOT`` variables in ``MakeGenerator``.
Docs: Omit

MakeGenerator was using ``rootpath`` instead of ``root``.

I also did some refactors to simplify it a bit, reducing the number of methods, making them private,


solvingj:
Yes, this looks like a bug, good find.

solvingj (approved):

PR #8306, opened 2021-01-08
Feature: add venv command

SSE4:
Changelog: Feature: Add venv command.
Docs: TODO

### DIFF
--- a/conans/client/cmd/venv.py
+++ b/conans/client/cmd/venv.py
@@ -1,0 +1,2 @@
+def cmd_venv(conanfile, cmd):
+    conanfile.run(" ".join(cmd), run_environment=True)
--- a/conans/client/command.py
+++ b/conans/client/command.py
@@ -11,6 +11,7 @@ from six.moves import input as user_input
 
 from conans import __version__ as client_version
 from conans.client.cmd.frogarian import cmd_frogarian
+from conans.client.cmd.venv import cmd_venv
 from conans.client.cmd.uploader import UPLOAD_POLICY_FORCE, \
     UPLOAD_POLICY_NO_OVERWRITE, UPLOAD_POLICY_NO_OVERWRITE_RECIPE, UPLOAD_POLICY_SKIP
 from conans.client.conan_api import Conan, default_manifest_folder, _make_abs_path, ProfileData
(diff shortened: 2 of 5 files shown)


uilianries:
Good idea!

memsharded (commented):
I would like to learn a bit better the use case for this PR:

- What would be some real life use cases for this command
- User requests that would be satisfied with this feature
- Why not using virtualenv generators

SSE4:
> I would like to learn a bit better the use case for this PR:
> 
> * What would be some real life use cases for this command
> * User requests that would be satisfied with this feature
> * Why not using virtualenv generators

the use case is the same as `VirtuanEnv` itself, basically. actually, it's more likely a syntax sugar on top of `VirtualEnv`.
right now, to get cmake and execute command with it I need 4 conan commands:
```
conan install cmake/3.19@ -g virtualenv
./activate.sh (or ps1, or bat)
cmake --version
./deactivate.sh (or ps1, or bat)
```
with new feature, it would be just:
```
conan venv cmake/3.19@ -- cmake --version
```
it's easier to remember just one command than 4 commands, right?
also, it is shell-agnostic, so you don't need to always check which script to execute.
further, I can use this command as an alias, or environment variable, like:
```
alias cmake=conan venv cmake/3.19@ --
```
```
export CONAN_CMAKE_PROGRAM=conan venv cmake/3.19@ --
```
so the similar one-liner could be easily used everywhere I need to pass for instance CC/CXX.

memsharded:
Thanks very much for describing this in detail, now I understand better.

I think you described this accurately: ``it's more likely a syntax sugar on top of``. As such, I think this is not the right time to introduce it. We are just starting the process of the new toolchains, and moving there Autotools and also managing build-requires will probably require some big efforts in managing virtualenvs. We will put the foundations first, which, at the moment they will be based on files (Virtualenv generator like), and we will consider higher level commands later.

Despite its apparent simplicity, some of the usages that you are already commenting reads to me like a Pandora box waiting to be opened, like ``export CONAN_CMAKE_PROGRAM=conan venv cmake/3.19@ --`` means that now when you launch Conan to install something, it might launch Conan itself as a subprocess. What if ``cmake/3.19`` doesn't have a binary, will it build it from source? Some of the locking mechanisms of the cache might fail, as they are not prepared to be recursive. I am not sure we want to open this right now, I recommend reviewing this later down the road, at least after maturing the whole build system new toolchain integrations and having a clear view of how the environments look like in general.

memsharded:
Created issue in https://github.com/conan-io/conan/issues/8393, and assigned it for checking later, lets close this PR at this moment, we will revisit later following the issue.

PR #8321, opened 2021-01-11
Allow global build modules with components

danimtb:
Changelog: Feature: Allow the usage of global scope build modules with components in `cpp_info`
Docs: omit

Note that build modules can still be used in components although they are always injected in the global scope.

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -261,7 +261,6 @@ class CppInfo(_CppInfo):
              self.cxxflags or
              self.sharedlinkflags or
              self.exelinkflags or
-             self.build_modules or
              self.requires):
             raise ConanException("self.cpp_info.components cannot be used with self.cpp_info "
                                  "global values at the same time")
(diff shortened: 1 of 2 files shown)


jgsogo (commented):
Not really sure. Do we want `build_modules` at the global level when there are components? This _global_ `build_modules` will be included always or only when the consumer requires the global target (`requires = [pkg::pkg]`)?

memsharded:
> Not really sure. Do we want build_modules at the global level when there are components? This global build_modules will be included always or only when the consumer requires the global target (requires = [pkg::pkg])?

They are global anyway, as the include mechanism is defined at the xxxx-config.cmake level, isn't it? So what doesn't seem feasible is to have build_modules per component? Can you please clarify @danimtb ?

danimtb:
Following the discussion, we need to evaluate the approach suggested by @jgsogo regarding including**only** build modules from the components required. I am closing this PR for the moment

PR #8333, opened 2021-01-13
add CMAKE_GENERATOR_PLATFORM and CMAKE_GENERATOR_TOOLSET conditionall…

### LINKED ISSUE #7485: [feature] [toolchain.cmake] Set platform and toolset only if generator matches Visual Studio

memsharded:
…y to Visual Studio only in CMakeToolchain

Changelog: Fix: Do not define CMAKE_GENERATOR_PLATFORM and CMAKE_GENERATOR_TOOLSET in the ``CMakeToolchain`` file unless the CMake generator is "Visual Studio". Fix https://github.com/conan-io/conan/issues/7485
Docs: Omit

This might be a temporary measure, as the generator needs to become more explicit at some point (via configuration or some other way), so the ``CMakeToolchain`` might decide to completely skip those definitions if they don't apply.

### DIFF
--- a/conan/tools/cmake/generic.py
+++ b/conan/tools/cmake/generic.py
@@ -57,8 +57,12 @@ class CMakeGenericToolchain(CMakeToolchainBase):
 
         {% block before_try_compile %}
             {{ super() }}
-            {% if generator_platform %}set(CMAKE_GENERATOR_PLATFORM "{{ generator_platform }}" CACHE STRING "" FORCE){% endif %}
-            {% if toolset %}set(CMAKE_GENERATOR_TOOLSET "{{ toolset }}" CACHE STRING "" FORCE){% endif %}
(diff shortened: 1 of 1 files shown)


memsharded:
It seems this is not necessary. The toolchain already incorporated checks, if the generator is not "Visual" one, it will not add any longer those variables. This can be closed.

PR #8337, opened 2021-01-14
Documenting some recomendations to add testing with Pytest

czoido:
Changelog: Omit
Docs: Omit

Add documentation on some general guidelines when adding tests to Conan: naming of files and tests, marks that should be added, tests folder organization...

### DIFF
--- a/conans/test/README.md
+++ b/conans/test/README.md
@@ -54,9 +54,10 @@ class TestSomeFunctionality:
 
 ### Marking tests
 
-Please mark your tests if they need to. Besides the builtin Pytest markers we interpret some markers
-related to external tools: `cmake`, `gcc`, `clang`, `visual_studio`, `mingw`, `autotools`,
-`pkg_config`, `premake`, `meson`, `file`, `git`, `svn`, `compiler`, `conan`. For example:
+Please mark your tests if they need to. Besides the [builtin Pytest
+markers](https://docs.pytest.org/en/stable/mark.html#mark) we interpret some markers related to
+external tools: `cmake`, `gcc`, `clang`, `visual_studio`, `mingw`, `autotools`, `pkg_config`,
+`premake`, `meson`, `file`, `git`, `svn`, `compiler`, `conan`. For example:
 
 ```python
 @pytest.mark.skipif(platform.system() != "Windows", reason="Needs windows for vcvars")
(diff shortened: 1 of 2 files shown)


conans/test/README.md
```diff
@@ -0,0 +1,88 @@
...
+discovery:
+- Name your Python test files ending in `_test.py`.
+
+```
+test
+├── README.md
+├── conftest.py
+├── unittests
+│   ├── __init__.py
+│   ├── basic_build_test.py
```
memsharded: Better start names with ``test_``. It makes evident when something is not named (and not running)

memsharded (approved):
LGTM with that minor change about filenames

conans/test/README.md
```diff
@@ -0,0 +1,88 @@
...
+```
+
+If the test needs any of those tools to run it should be marked as using that tool and moved to the `conans/test/functional` folder.
+
+### Parametrizing tests
+
+Please, if you need to run several combinations of the same testing code use parameterization. You can use the builtin `pytest.mark.parametrize` decorator to enable parametrization of arguments for a test function:
+
+```python
+@pytest.mark.parametrize("use_components", [False, True])
```
jgsogo:
```suggestion
    @pytest.mark.parametrize("use_components", [False, True])
```

conans/test/README.md
```diff
@@ -0,0 +1,88 @@
...
+
+Pytest follows this [convention](https://docs.pytest.org/en/stable/goodpractices.html) for test
+discovery:
+- Name your Python test files starting in `test_`.
+
+```
+test
+├── README.md
+├── conftest.py
+├── unittests
```
jgsogo, line 37:
```suggestion
├── integration
```

Related to `unittests`, which is the example here, but it was something I wanted to comment in this PR.

This is me **completely opinionated about this**: I REALLY like the tests under `unittests` to follow EXACTLY the same directory/files structure as the codebase, why?
 * no doubts where to add a test
 * no doubts where to check if a function is tested before writing one more test (was it already tested?).
 * when you are refactoring something, you only need to run one folder while changing the implementation.

We can talk about this later, but then I'd change the example here and show a file from `functional` or `integration`

memsharded: I agree, unittests should follow the same layout as the prod codebase.

memsharded (approved):

PR #8338, opened 2021-01-14
[test] Add test creating a target without namespaces (CMake)

### LINKED ISSUE #7615: [feature] cmake_find_package[_multi] generators: provide a way to create non namespaced imported targets

jgsogo:
Changelog: omit
Docs: omit

Adds a test using `build_modules` to create a target without namespace. This target consumes the targets provided by Conan.

closes https://github.com/conan-io/conan/issues/7615 - At this moment we are not implementing anything else in the model to create targets without namespaces (it is very CMake specific and `cpp_info` should have some level of abstraction over different build-systems). ping @madebr @SpaceIm

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -2,8 +2,6 @@ import os
 from collections import OrderedDict
 from copy import copy
 
-import six
-
 from conans.errors import ConanException
 from conans.util.conan_v2_mode import conan_v2_behavior
 
@@ -84,7 +82,6 @@ def dict_to_abs_paths(the_dict, rootpath):
 def merge_dicts(d1, d2):
     def merge_lists(seq1, seq2):
         return [s for s in seq1 if s not in seq2] + seq2
-
     result = d1.copy()
     for k, v in d2.items():
         if k not in d1.keys():
@@ -148,12 +145,7 @@ class _CppInfo(object):
     @property
     def build_modules_paths(self):
         if self._build_modules_paths is None:
-            # FIXME: This should be just a plain dict
-            if isinstance(self.build_modules, six.string_types):
-                conan_v2_behavior("Use 'self.cpp_info.build_modules[\"<generator>\"] = "
-                                  "{the_list}' instead".format(the_list=[self.build_modules, ]))
-                self.build_modules = BuildModulesDict.from_list([self.build_modules, ])
-            if isinstance(self.build_modules, list):
+            if isinstance(self.build_modules, list):  # FIXME: This should be just a plain dict
                 conan_v2_behavior("Use 'self.cpp_info.build_modules[\"<generator>\"] = "
                                   "{the_list}' instead".format(the_list=self.build_modules))
                 self.build_modules = BuildModulesDict.from_list(self.build_modules)
@@ -354,8 +346,7 @@ class CppInfo(_CppInfo):
                 reason = None
                 if comp_require not in pkg_requires:
                     reason = "not defined as a recipe requirement"
-                elif package_requires[comp_require].private and package_requires[
-                    comp_require].override:
+                elif package_requires[comp_require].private and package_requires[comp_require].override:
                     reason = "it was defined as an overridden private recipe requirement"
                 elif package_requires[comp_require].private:
                     reason = "it was defined as a private recipe requirement"
@@ -389,6 +380,7 @@ class _BaseDepsCppInfo(_CppInfo):
         super(_BaseDepsCppInfo, self).__init__()
 
     def update(self, dep_cpp_info):
+
         def merge_lists(seq1, seq2):
             return [s for s in seq1 if s not in seq2] + seq2
 
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -2,6 +2,8 @@ import os
 from collections import OrderedDict
 from copy import copy
 
+import six
+
 from conans.errors import ConanException
 from conans.util.conan_v2_mode import conan_v2_behavior
 
@@ -82,6 +84,7 @@ def dict_to_abs_paths(the_dict, rootpath):
 def merge_dicts(d1, d2):
     def merge_lists(seq1, seq2):
         return [s for s in seq1 if s not in seq2] + seq2
+
     result = d1.copy()
     for k, v in d2.items():
         if k not in d1.keys():
@@ -145,7 +148,12 @@ class _CppInfo(object):
     @property
     def build_modules_paths(self):
         if self._build_modules_paths is None:
-            if isinstance(self.build_modules, list):  # FIXME: This should be just a plain dict
+            # FIXME: This should be just a plain dict
+            if isinstance(self.build_modules, six.string_types):
+                conan_v2_behavior("Use 'self.cpp_info.build_modules[\"<generator>\"] = "
+                                  "{the_list}' instead".format(the_list=[self.build_modules, ]))
+                self.build_modules = BuildModulesDict.from_list([self.build_modules, ])
+            if isinstance(self.build_modules, list):
                 conan_v2_behavior("Use 'self.cpp_info.build_modules[\"<generator>\"] = "
                                   "{the_list}' instead".format(the_list=self.build_modules))
                 self.build_modules = BuildModulesDict.from_list(self.build_modules)
@@ -346,7 +354,8 @@ class CppInfo(_CppInfo):
                 reason = None
                 if comp_require not in pkg_requires:
                     reason = "not defined as a recipe requirement"
-                elif package_requires[comp_require].private and package_requires[comp_require].override:
+                elif package_requires[comp_require].private and package_requires[
+                    comp_require].override:
                     reason = "it was defined as an overridden private recipe requirement"
                 elif package_requires[comp_require].private:
                     reason = "it was defined as a private recipe requirement"
@@ -380,7 +389,6 @@ class _BaseDepsCppInfo(_CppInfo):
         super(_BaseDepsCppInfo, self).__init__()
 
     def update(self, dep_cpp_info):
-
         def merge_lists(seq1, seq2):
             return [s for s in seq1 if s not in seq2] + seq2
 
--- a/conans/test/functional/generators/cmake_find_package_multi_test.py
+++ b/conans/test/functional/generators/cmake_find_package_multi_test.py
@@ -539,7 +539,7 @@ class TestNoNamespaceTarget:
         from conans import ConanFile, CMake
 
         class Recipe(ConanFile):
-            settings = "os", "compiler", "arch"
+            settings = "os", "compiler", "arch", "build_type"
             exports_sources = ["src/*", "build-module.cmake"]
             generators = "cmake"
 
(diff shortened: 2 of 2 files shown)


conans/model/build_info.py
```diff
@@ -145,7 +148,12 @@ def _filter_paths(self, paths):
     @property
     def build_modules_paths(self):
         if self._build_modules_paths is None:
-            if isinstance(self.build_modules, list):  # FIXME: This should be just a plain dict
+            # FIXME: This should be just a plain dict
+            if isinstance(self.build_modules, six.string_types):
+                conan_v2_behavior("Use 'self.cpp_info.build_modules[\"<generator>\"] = "
+                                  "{the_list}' instead".format(the_list=[self.build_modules, ]))
+                self.build_modules = BuildModulesDict.from_list([self.build_modules, ])
```
jgsogo: Shall I open a different PR with this change? I'm not sure if we want to cover this scenario (maybe not), but the current error is very ugly... I preferred to implement it than to raise. wdyt?

SpaceIm:
What happen if a consumer doesn't use `cmake_find_package(_multi)` generators. `library::library` doesn't exist anymore, so I guess `target_link_libraries(nonamespace INTERFACE library::library)` will fail?

conans/model/build_info.py
memsharded: I would say raise always with a clear error message?

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -526,3 +526,121 @@ def build(self):
...
+
+        int main() {
+            library();
+        }
+    """)
+
+    @classmethod
+    def setup_class(cls):
+        cls.t = t = TestClient()
+        t.current_folder = '/private/var/folders/fc/6mvcrc952dqcjfhl4c7c11ph0000gn/T/tmpiq_fqfmfconans/path with spaces'
```
memsharded: Why this folder?

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -526,3 +526,121 @@ def build(self):
...
+                self.cpp_info.build_modules = module
+    """)
+
+    build_module = textwrap.dedent("""
+        message(">> Build-module is included")
+
+        if(NOT TARGET nonamespace)
+            add_library(nonamespace INTERFACE IMPORTED)
+            target_link_libraries(nonamespace INTERFACE library::library)
+        endif()
```
SpaceIm, line 573: What happen if a consumer relies on another generator than `cmake_find_package(_multi)` (`cmake` for example)? `library::library` doesn't exist anymore, so I guess `target_link_libraries(nonamespace INTERFACE library::library)` will fail?

Maybe this?

```suggestion
        if(NOT TARGET nonamespace AND TARGET library::library)
            add_library(nonamespace INTERFACE IMPORTED)
            target_link_libraries(nonamespace INTERFACE library::library)
        endif()
```

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -526,3 +526,121 @@ def build(self):
...
+
+    main = textwrap.dedent("""
+        #include "library.h"
+
+        int main() {
+            library();
+        }
+    """)
+
+    @classmethod
```
memsharded, line 602: why not using pytest fixtures? I have started to used them and so far look great, if we are going pytest, probably we want to start adopting its practices?

memsharded (approved):

conans/test/functional/generators/cmake_find_package_multi_test.py
jgsogo: ops, clearly need to remove this line. 🙈

conans/test/functional/generators/cmake_find_package_multi_test.py
jgsogo: because the purpose here is to execute these lines once for all the tests in this class (I could use class-scoped fixture).

A fixture typically provides a way to execute (and inject) something before/after tests, but IMHO **something intended to be reusable by many tests in the test suite**. This one probably applies only to this test-class, I think these class methods are better than fixtures when they apply only to one specific test.

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -526,3 +526,121 @@ def build(self):
...
+                self.copy("*.dll", dst="bin", keep_path=False)
+                self.copy("*.dylib*", dst="lib", keep_path=False)
+                self.copy("*.so", dst="lib", keep_path=False)
+                self.copy("*.a", dst="lib", keep_path=False)
+                self.copy("build-module.cmake", dst="share/cmake")
+
+            def package_info(self):
+                self.cpp_info.libs = ["library"]
+                module = os.path.join("share", "cmake", "build-module.cmake")
+                self.cpp_info.build_modules = module
```
SpaceIm:
```suggestion
                self.cpp_info.build_modules = [module]
```

conans/model/build_info.py
jgsogo: Moving to a different PR

jgsogo: No, I'm not suggesting this change anymore. It should be something to check and validate for _all_ `cpp_info` fields, it doesn't make sense to check just this one.

conans/test/functional/generators/cmake_find_package_multi_test.py
jgsogo: I wrote a comment, but it went away with some of the commits 🤷 .

As you can see here, I can specify a different build-module per generator (https://github.com/conan-io/conan/pull/8338/commits/f2163c6c30037c0476aa48733a58e060bda65b82), otherwise it is the recipe author the one that needs to implement the required logic inside the build-module, of course, as you have suggested.

SpaceIm: ok, so I understand that we could already (conan 1.32.0) provide non namespaced CMake imported targets in CCI recipe? conan 1.32.0 doesn't allow build_modules per generator but with some tricks it should work.

jgsogo: Limitation in 1.32 is that `build_modules` are included before the Conan targets. It was changed here: https://github.com/conan-io/conan/pull/8130

SpaceIm: Surprinsingly, it works in 1.32: https://github.com/conan-io/conan-center-index/pull/4258
If I add `message()` in the build_modules file, I see that it's called twice: one before targets found (in `cmake_find_package_multi`) and another one after.

jgsogo: Dammit! 😅   It should be some kind of (evil) CMake laziness resolution...

madebr: So the idea is to create cmake build modules that are aliases to the conan generated ones?
This is supposed to happen in `build`, `package` or copied through from `exports_sources`?

Sounds good to me! 💯 

My next question is to add a build helper, as this is going to be often needed in cci.

Something like:

```python
from conan.tools.cmake import cmake_tools

def package(self):
    cmake_tools.write_alias_module
        {
            "library": "library::library",
            "library2": "library::library2",
        },
        filename=None,  # default is os.path.join("lib", "cmake", "build-modules.cmake"),
        generators=None,  # (generators' default is `cmake_find_package` and `cmake_find_package_multi`)
    )

def package_info(self):
    # ...
    cmake_tools.append_alias_module_filename(
        self.cpp_info,
        filename=None,  # default is os.path.join("lib", "cmake", "build-modules.cmake"),
        generators=None,  # default is os.path.join("lib", "cmake", "build-modules.cmake"),
    )
```

The idea is to need no extra arguments by default.

prince-chrismc: Please add a helper :heart:

SpaceIm: > Dammit! 😅 It should be some kind of (evil) CMake laziness resolution...

Actually, it seems to work in conan 1.32 if there are no components in the recipe. It fails in opencv recipe (lot of components): https://github.com/conan-io/conan-center-index/pull/4267

danimtb (approved):

conans/test/functional/generators/cmake_find_package_multi_test.py
jgsogo: About the helper, something for a feature request, not for this PR. Here we only wanted to check that adding targets without namespaces was possible.

SpaceIm:
Is there a test with CMake imported executables targets?

I've tried to add this in `geographiclib` recipe:
https://github.com/SpaceIm/conan-geographiclib/blob/a143f702c9fb369651356e7bd28b590032044e1e/conanfile.py#L142
https://github.com/SpaceIm/conan-geographiclib/blob/a143f702c9fb369651356e7bd28b590032044e1e/test_package/CMakeLists.txt#L16

And it works... https://github.com/SpaceIm/conan-geographiclib/actions/runs/494023796

jgsogo:
Hi! I would say it should be more a built-in feature than something for a _custom_ build-module. I'd vote for this issue (https://github.com/conan-io/conan/issues/7240) and then cmake generators can create those targets as well.

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -526,3 +526,125 @@ def build(self):
...
+                self.cpp_info.build_modules['cmake_find_package'] = [module, ]
+                self.cpp_info.build_modules['cmake_find_package_multi'] = [module, ]
+    """)
+
+    build_module = textwrap.dedent("""
+        message(">> Build-module is included")
+
+        if(NOT TARGET nonamespace)
+            add_library(nonamespace INTERFACE IMPORTED)
+            target_link_libraries(nonamespace INTERFACE library::library)
```
SpaceIm, line 572: It seems that `set_property(TARGET nonamespace PROPERTY INTERFACE_LINK_LIBRARIES library::library)` has wider support across CMake versions (because nonamespace is an IMPORTED target I guess):
https://github.com/conan-io/conan-center-index/issues/4369

PR #8005, opened 2020-11-04
Validate CMake Toolchain + Ninja 

uilianries:
Changelog: Feature: ``CMakeToolchain`` will generate ``conanvcvars.bat`` for Ninja builds for ``msvc``.
Docs: Omit

- [ ] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [ ] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [ ] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #8343, opened 2021-01-15
improving toolchain checkers

memsharded:
Changelog: Omit
Docs: Omit


memsharded:
The failure in https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-8343/1/pipeline is good:

```
  main __cplusplus201103
  main __GNUC__8
  main __GNUC_MINOR__1
```

This is because Mingw installed there is 8.1. There is actually nothing in the toolchain that checks, validates or enforces the compiler version. The ``cmake`` generator has logic to check this.

SSE4:
@memsharded I am not sure, what do you mean? if the merge that "good failure", CI will be red forever, am I missing the point?

memsharded:
> @memsharded I am not sure, what do you mean? if the merge that "good failure", CI will be red forever, am I missing the point?

No, no, do not merge it! :) Such a test should be marked with pytest.xfail. 
What I am saying is that it is good that this extra checker is proving something is not great with the compiler versions. The test is defining gcc 4.9 and is green. The new checker is proving that the used gcc is 8.1, but no one knows. It means that:

- We could change the test to provide the correct gcc version, or not use a gcc version at all.
- We could implement checkers in the toolchain, as ``conanbuildinfo.cmake`` does. (This would be needed in the CMAKE_PROJECT_INCLUDE, and validate that the current compiler version is the expected one.

memsharded:
``check_exe_run()`` will not validate the version and cppstd if they are None. Good to merge, as the checkers is fine.
Now we need to improve the testing and toolchain functionality to check this better.

uilianries (approved):
Tested on Windows. LGTM. Didn't affect #8005 (conflicts)

PR #8344, opened 2021-01-16
Proposal for lockfiles bundle management (multiple products and configurations)

### LINKED ISSUE #7180: [feature] update multiple lockfiles if they share modified/built package and package_id
Summary:

Give users a way to update multiple lockfiles when they share a common package/packageID which has been `"modified/built"`.  Possible ways to approach it include:
* make `conan create` accept multiple lockfiles (probably not the right approach)
* make `conan install` able to detect that the package in the local cache has changed and just update without rebuild (probably not the right approach)
* add `conan graph copy-modified` to copy a status update from one lockfile to N lockfiles, when each of those lockfiles features the same identical packageID for the "Modified/built" node..  (probably the most straightforward). 

Background

During experimentation with CI workflows and lockfiles, there's a major scalability issue with an extremely common scenario.  The premise of the scenario is that many organizations have long lists of products with complicated options, but for which they share binaries for large numbers of low level components built with a common set of conan settings and options.

Example Scenario: 
zlib
 -> openssl
    ->  50 products

* For simplicity, lets imagine our CI just builds a single profile/variant for each product
* Each product has slightly different package options
* All the options used for zlib and openssl are identical between all the profiles
* Thus we have 50 distinct lockfiles

Despite having 50 distinct lockfiles, they all resolve to the same package IDs for zlib and OpenSSL.  If we make a patch to zlib and use revisions and lockfiles to rebuild everything, the current guidance for Conan + CI is: "build zlib once for every product lockfile".  That's 50 builds of zlib and openssl, each of them all having the same package ID as the one before.  This is an unacceptable approach.  Whether run in parallel or sequentially, it's wildly inefficient.  This was just a modest example, if you imagine building 4 or 16 lockfiles for each product, it's many many hundreds of redundant builds of zlib when 4 or 16 would suffice. 

The reason the guidance says "build once with each unique lockfile"  is that the only way to mark a lockfile with "modified/built" under zlib is to run  `conan create --lockfile`.  We need each lockfile to get this update, because they'll all be used when we pass them to their appropriate downstreams for the 50 unique product builds. 

In my local environment, i've used the following trick to workaround this redundant rebuilds by reading the lockfiles and grouping them by the packageID they contain for `zlib`.  Then, I build zlib only once for each unique packageID of zlib found by doing: 

`conan create --lockfile <random_lockfile_from_group>` 

Then, I update the rest of the lockfiles with this trick:

`conan create --lockfile <random_lockfile_from_group> --keep-build --test-folder=None` 

This re-runs the conan create, but skips the actual build and test package.  So, it basically just updates the remaining lockfiles.  It actually does re-run the `package()` method which is ok-ish for now.  IF we had a flag called `--keep-package` , that would be even faster, but, obviously, that's all just workaround optimization. 

The simplest and most desirable solution to this specific problem (given the current implementation of lockfiles), is the one suggested at the start.  Give users a way to update many lockfiles with a single build when all lockfiles share the same package id for the "modified/built" package. 

memsharded:
Changelog: Feature: New lockfile "bundle" feature that can integrate different lockfiles for different configurations and different graphs into a single lockfile bundle that can be used to vastly optimize CI (specially for multiple products), implementing bundle build-order and bundle update operations.
Docs: https://github.com/conan-io/docs/pull/2030

Close https://github.com/conan-io/conan/issues/7180

Note the division of the nodes in the "multi-lock" graph is per recipe reference, not per package. Because there could be different versions and revisions of the same package, I am not sure it really makes sense to build ``pkga/1.0#rev1``, ``pkga/2.0#rev2``, ``pkga/2.0#rev3``, etc in the same task, but every parallel task should be based also in a commit, so every recipe reference sounds the reasonable division so far.


conans/model/lock_multi.py
```diff
@@ -0,0 +1,102 @@
...
+import os
+
+from conans.model.graph_lock import GraphLockFile
+from conans.util.files import load, save
+
+
+class LockMulti(object):
+
+    def __init__(self):
+        self.nodes = {}
```
solvingj: Consider a descriptive top level member above `nodes` similar to existing lockfiles such as `graph-lock` to make it apparent what we're looking at when we open a `multi-lock` file. Could also just implement it during `dumps`.

conans/test/integration/graph_lock/test_multi_lock.py
```diff
@@ -0,0 +1,55 @@
...
+            for pkg_id, lockfile_info in pkg_ids.items():
+                print("    Processing ", pkg_id)
+                lockfiles = lockfile_info["lockfiles"]
+                lockfile = next(iter(lockfiles))
+                print("    install ", ref, lockfile)
+                client.run("install {ref} --build={ref} --lockfile={lockfile} "
+                           "--lockfile-out={lockfile}".format(ref=ref, lockfile=lockfile))
+                client.run("lock update-multi multi.lock")
+
+    print(client.load("multi.lock"))
```
solvingj: I see the output has these which I believe would be cleaned in final draft: 
```
                "prev": null,
                "modified": null
```

solvingj (approved):
Ok, I've spent some time analyzing the PR and now I think I now understand the scope of the features here.  There have been several discussions covering a broad range of possible scopes and implementation strategies, so I will try to summarize what I see and add the context of the challenges we're addressing. 

First, this provides a dedicated command for creating a single `multi-build-order.json` file from a list of arbitrary lockfiles.  This is nice, because previously we had to use a custom external script to did this in CI flows.  It wasn't particularly complicated script because the math is quite simple, but it was unfortunate that every conan user would have to implement this script for such a common case.  

Second, it tries to provide a solution for the problem listed here: 
https://github.com/conan-io/conan/issues/7180

In summary, when creating a dynamic CI flow driven by Conan lockfiles, it's currently extremely common to need to update multiple lockfiles in the scope of a pipeline, and for those lockfiles to reference a common dependency with the same `package_id`.  With a naïve approach of building affected nodes "once-per-lockfile", this will lead to multiple re-builds of a common `package_id`, once for each lockfile in the job, which has multiple unfortunate implications.

The implementation provides a dedicated command to take a list of arbitrary lockfiles and produce the `multi-lock` file, which is a dictionary of:
`package_ref#rrev` -> `list(package_id)` -> `lockfile_info` (`lockfile_path`, `prev`, `modified`)

Once the user has generated this multi-lock file, they can do the following in the context of a CI job for `pkgc/1.0` for example: 

1. load multi-lock file and parse as json
2. find the reference for the package currently building
3. enumerate the list of package_id
4. spawn a `build step` for each package id
5. for each build step 
     - choose a lockfile from the list of lockfiles. *
     - build the package with that lockfile with `conan install --build pkgc --lockfile chosen.lock`
     - run `conan lock update-multi multi-lock`

The final command will then iterate through all the lockfiles, and update all those which have empty `prev` and marking them as `modified : true`. 

I consider this a a big improvement.  It makes updating the "subsequent" lockfiles a first class feature. Previously, to achieve this I had resorted to iterating over each of the remaining lockfiles and manually update the two fields (`prev` and `modified` themselves).  While actually fairly simple to implement, this again was unfortunate logic for every user to need to figure out and implement themselves.  Also, it was error prone and potentially unsafe, and manual modifications to lockfiles are effectively unsupportable. Now there is a dedicated command and guidance for doing this operation safely, and that's a good thing.  

I consider this ready to move forward and try in next release.

*Note: The need for the user to choose an arbitrary lockfile from the list to run the build is slightly more inconvenient than it probably seems.  If this could somehow be avoided, the user could achieve this entire part of the flow directly with conan commands at their shell.  As it stands, the user needs to actually load and parse json just to choose an entry.  I understand why this design choice was made at this time.  Perhaps later, if we choose to move forward with this `multi.lock` strategy, we can provide some convenience commands for working with them, such as `conan lock search-multi package_id --first` (obviously not that specifically, but something where users can get what they need for common case from conan's stdout).

conans/test/integration/graph_lock/test_multi_lock.py
memsharded: This is fine, the intent is to include them. A bundle can track of the package revision it implements, and if it has been modified during this build (so this can be used to upload or something). Any problem?

conans/model/lock_multi.py
memsharded: What do you mean? The output is json, cannot contain comments.

solvingj (approved):

conans/model/lock_multi.py
solvingj: I meant that in lockfiles, we have a top-level node called `graph-lock`. This makes it clear what the purpose of these nodes are. 

    {
     "graph_lock": {
      "nodes": {
       "4": {

So perhaps:  

    { 
      "lock_bundle": {

solvingj:
The code looks very elegant and simple, and does solve the first use case of providing a consolidated build order across multiple lockfiles as needed.  This will eliminate a custom script in the CI training course. To me, it is ready to go.

solvingj (approved):

jgsogo:
I wonder if this is very CI specific and not a package-manager thing. Maybe these are scripts for a CI dedicated library (or at least extensions via CLI 2.0) and not core Conan features, wdyt?

memsharded:
> I wonder if this is very CI specific and not a package-manager thing. Maybe these are scripts for a CI dedicated library (or at least extensions via CLI 2.0) and not core Conan features, wdyt?

It is very CI specific, but also some of the features we already have. We are also bundling conan_server and conan_build_info in the same release:

- Waiting until CLI 2.0 is not an option, too far away
- Separating in another library is very inconvenient, for our users and also ourselves. 

So our best alternative is released it under the ``conan lock`` umbrella. The risk is very low, maintenance is easy, and it cannot be more convenient for our users.

czoido (approved):

PR #8347, opened 2021-01-18
Fix: Allow modification of system/cpu_family/cpu/endian in generated MesonToolchain toolchain file

### LINKED ISSUE #8312: [feature] Allow modification of system/cpu_family/cpu/endian in generated MesonToolchain toolchain file
Hello,

I'm currently creating recipes+packages aimed for the Nintendo Switch.
I added `Nintendo Switch` to `os` in my `settings.yml`.

While creating a recipe for the mesa port, I'm using the MesonToolchain class.
This creates the following section:
```
[build_machine]

system = 'linux'
cpu_family = 'x86_64'
cpu = 'x86_64'
endian = 'little'

[host_machine]

system = 'nintendo switch'
cpu_family = 'aarch64'
cpu = 'armv8'
endian = 'little'

[target_machine]

system = 'nintendo switch'
cpu_family = 'aarch64'
cpu = 'armv8'
endian = 'little'
```

The correct data for nintendo should be:
```
[host_machine]
system = 'horizon'
cpu_family = 'aarch64'
cpu = 'armv8'
endian = 'little'
```

What I'm asking for is a feature to override these default/calculated values.

Right now, the `toolchain` method is the following:
```python
    def toolchain(self):
        tc = MesonToolchain(self)
        tc.generate()
```
It would be useful to make this possible:
```
    def toolchain(self):
        tc = MesonToolchain(self)
        tc.host_machine.system = "horizon"
        tc.target_machine.system = "horizon"
        tc.generate()
```
Or something similar...

CC'ing @SSE4 for being the strong man behind `MesonToolchain`.

SSE4:
closes: #8312

based on #8266

Changelog: Fix: Allow modification of system/cpu_family/cpu/endian in generated MesonToolchain toolchain file
Docs: omit

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -190,7 +190,7 @@ class MesonToolchain(object):
     def _native_content(self):
         return self._render(self._native_file_template, self._context)
 
-    def _to_meson_machine(self, machine_os, machine_arch):
+    def _to_meson_machine(self, machine_os, machine_arch, conf):
         # https://mesonbuild.com/Reference-tables.html#operating-system-names
         system_map = {'Android': 'android',
                       'Macos': 'darwin',
@@ -232,10 +232,12 @@ class MesonToolchain(object):
                           'wasm': ('wasm32', 'wasm32', 'little'),
                           'x86': ('x86', 'x86', 'little'),
                           'x86_64': ('x86_64', 'x86_64', 'little')}
-        system = system_map.get(machine_os, machine_os.lower())
+        system = conf.system or system_map.get(machine_os, machine_os.lower())
         default_cpu_tuple = (machine_arch.lower(), machine_arch.lower(), 'little')
         cpu_tuple = cpu_family_map.get(machine_arch, default_cpu_tuple)
-        cpu_family, cpu, endian = cpu_tuple[0], cpu_tuple[1], cpu_tuple[2]
+        cpu_family = conf.cpu_family or cpu_tuple[0]
+        cpu = conf.cpu or cpu_tuple[1]
+        endian = conf.endian or cpu_tuple[2]
         context = {
             'system': self._to_meson_value(system),
             'cpu_family': self._to_meson_value(cpu_family),
@@ -249,9 +251,13 @@ class MesonToolchain(object):
         os_build, arch_build, os_host, arch_host = get_cross_building_settings(self._conanfile)
         os_target, arch_target = os_host, arch_host  # TODO: assume target the same as a host for now?
 
-        build_machine = self._to_meson_machine(os_build, arch_build)
-        host_machine = self._to_meson_machine(os_host, arch_host)
-        target_machine = self._to_meson_machine(os_target, arch_target)
+        build_machine_conf = self._conanfile.conf["tools.meson.mesontoolchain.build_machine"]
+        host_machine_conf = self._conanfile.conf["tools.meson.mesontoolchain.host_machine"]
+        target_machine_conf = self._conanfile.conf["tools.meson.mesontoolchain.target_machine"]
+
+        build_machine = self._to_meson_machine(os_build, arch_build, build_machine_conf)
+        host_machine = self._to_meson_machine(os_host, arch_host, host_machine_conf)
+        target_machine = self._to_meson_machine(os_target, arch_target, target_machine_conf)
 
         context = self._context
         context['build_machine'] = build_machine
--- a/conans/test/functional/toolchains/meson/test_conf.py
+++ b/conans/test/functional/toolchains/meson/test_conf.py
@@ -1,0 +1,93 @@
+import os
+import textwrap
+
+from conans.test.assets.sources import gen_function_cpp, gen_function_h
+from conans.test.functional.toolchains.meson._base import TestMesonBase
+
+
+class MesonConfTest(TestMesonBase):
+    _conanfile_py = textwrap.dedent("""
+    from conans import ConanFile, tools
+    from conan.tools.meson import Meson, MesonToolchain
+
+
+    class App(ConanFile):
+        settings = "os", "arch", "compiler", "build_type"
+        options = {"shared": [True, False], "fPIC": [True, False]}
+        default_options = {"shared": False, "fPIC": True}
+
+        def config_options(self):
+            if self.settings.os == "Windows":
+                del self.options.fPIC
+
+        def generate(self):
+            tc = MesonToolchain(self)
+            tc.generate()
+
+        def build(self):
+            meson = Meson(self)
+            meson.configure()
+            meson.build()
+    """)
+
+    _meson_build = textwrap.dedent("""
+    project('tutorial', 'cpp')
+    hello = library('hello', 'hello.cpp')
+    executable('demo', 'main.cpp', link_with: hello)
+    """)
+
+    def test_build(self):
+        hello_h = gen_function_h(name="hello")
+        hello_cpp = gen_function_cpp(name="hello")
+        app = gen_function_cpp(name="main", includes=["hello"], calls=["hello"])
+
+        profile_host = textwrap.dedent("""
+    include(default)
+    [settings]
+    arch=armv8
+    os=FreeBSD
+    [conf]
+    tools.meson.mesontoolchain.host_machine:system=nes
+    tools.meson.mesontoolchain.host_machine:cpu_family=MOS
+    tools.meson.mesontoolchain.host_machine:cpu=6502
+    tools.meson.mesontoolchain.host_machine:endian=big
+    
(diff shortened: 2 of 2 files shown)


conans/test/functional/toolchains/meson/test_conf.py
```diff
@@ -0,0 +1,93 @@
...
+    executable('demo', 'main.cpp', link_with: hello)
+    """)
+
+    def test_build(self):
+        hello_h = gen_function_h(name="hello")
+        hello_cpp = gen_function_cpp(name="hello")
+        app = gen_function_cpp(name="main", includes=["hello"], calls=["hello"])
+
+        profile_host = textwrap.dedent("""
+    include(default)
```
memsharded: Plese indent this so it is indented 1 more than ``profile_host`` variable.

conan/tools/meson/toolchain.py
```diff
@@ -249,9 +251,13 @@ def _cross_content(self):
         os_build, arch_build, os_host, arch_host = get_cross_building_settings(self._conanfile)
         os_target, arch_target = os_host, arch_host  # TODO: assume target the same as a host for now?
 
-        build_machine = self._to_meson_machine(os_build, arch_build)
-        host_machine = self._to_meson_machine(os_host, arch_host)
-        target_machine = self._to_meson_machine(os_target, arch_target)
+        build_machine_conf = self._conanfile.conf["tools.meson.mesontoolchain.build_machine"]
```
memsharded, line 254: I am concerned that this is an alternate way to have settings defined, totally bypassing the recipe and profile settings.
So you can have something in your profile settings, totally disconnected from the ``[conf]`` and it will still work.

I think the feature should be more oriented to complement definitions in ``settings`` than a full replacement. In that sense, what is desired is a mapping between custom user architectures and the toolchain values, but the input data should come from the settings, not directly from the [conf]. Please do a proof of concept of this approach.

memsharded (changes requested):

conan/tools/meson/toolchain.py
SSE4: I am not sure I understand. there are no settings defined at all right now, if we are talking about conan settings (`settings.yml`). there are no settings like `os`, `compiler`, etc. defined in `conf`.
I mean this code doesn't modify settings at all, and doesn't touch them - `self.settings.get_safe("whatever")` will still return that it returned before...

memsharded: There should be some custom user settings to model different binaries. There are settings like architecture, that are being completely overriden by [conf]. You can put arch=x86 in your profile, and it will be ignored and build with armv8, because you said it in [conf]. This is incorrect.

The definitions for configurations should be settings as much as possible. Only very exceptional cases that wouldn't fit in settings could be possible. If a specific build system needs to map/convert those settings to its own representation, we can provide the mapping between settings and that representation in [conf] if necessary.

SSE4: if I understand correctly, the mapping should happen outside of conan then.
otherwise it's impossible to have custom settings supported that way.

memsharded: Yes, it could serve the purpose of augmenting built-in mappings, like ``cpu_family_map``, to make it extensible for user custom settings. Where this mapping could live? There could be several options:
- Make it available in the toolchain, but it would require to modify recipes to account for it, not the best.
- Is it possible to pass mappings in ``[conf]``? Maybe it is possible to pass the values, or an external mapping file? (whatever we implement, lets discuss to make it a general approach for other build systems).

SSE4: let's be more specific :)
the feature request has started for Nintendo Switch (`arch=armv8`), so what should be in conf for that particular case?
I believe something like:
```
[conf]
tools.meson.mesontoolchain.armv8.build_machine = nintendo switch
```
or if we do a mapping in conf:
```
[conf]
tools.meson.mesontoolchain.build_machine_mapping = armv8: nintendo switch
```
do I understand your idea correctly?

memsharded: I am not sure. I need to know the full input settings for ``os``, ``compiler``, etc, to know a bit better. If ``arch=armv8``, why not have it in the settings and just use it in Meson? I would say that ``Nintendo Switch`` should be a ``settings.os``, isn't it? If that requires to add an entry to ``system_map = {'Android': 'android',``, because it is not built-in and otherwise it will not map correctly, that is the behavior we are looking to inject into the ``MesonToolchain`` right? (and as many other build systems as possible whenever it makes sense)

SSE4: I am also not sure, it could be a weak defined area.
e.g. if PS4 runs FreeBSD (modified), should it be `os=FreeBSD` + `arch=x86_64`?
should it be different from regular (desktop) FreeBSD x86_64?
if I install vanilla FreeBSD on PS4, what should it be then?
and if PS5 also runs FreeBSD, should it also be the same settings combo?
or should we define `os.subsystem` for all these cases?
anyway, mapping could be very arbitrary (e.g. some devices don't even have OS at all, so it could be bare `arch` mapping for them).

solvingj: Indeed, as @memsharded said, this is a clear example of an alternate way to define variables which will be sent to the compiler and will deeply impact the generated binary.  I agree that it undermines the basic automatic behavior that Conan settings are intended to provide, which is that changing any one of these compiler-related values will result in a changed `package_id`, and the build output will be treated as a separate binary.  The same thing could have been done with environment variables before, and we would have recommended against it in that case.

However, this seems so obvious, it leads me to the question: Why did @SSE4 not see it this way, and choose to try this implementation?

My guess is this:  
The default settings included with Conan don't contain any of the primitive abstractions (variables) which Meson uses for it's cross-building model.  Meanwhile, @SSE4 is tasked with trying to create a "generic" toolchain for Meson which is built-in with Conan works out of the box for most Meson users.  It seems completely impossible to deduce values for these Meson variables from Conan's existing settings. So, there are really only two choices:

1. Add these settings to Conan
2. Use `conf` in this way
3. Maintain a MesonToolchain which really isn't that useful out-of-the-box
 
@SSE4 please confirm or correct me if I am wrong. 

This does seem like a difficult choice to be honest.

SSE4: @solvingj yes, that's exactly it. I've been thinking about it, and I guess:
>  There are settings like architecture, that are being completely overriden by [conf].

this is not accurate. settings are not overridden, e.g. `package_id` will still have that is defined in settings. we only override particular build system specific variables, which might (or might not) match settings values.

a) so, may we have `settings.arch=x86` and have `arm` binaries built actually in package?
yes, we definitely can. it's already possible (e.g. via environment variables like `CFLAGS` or just by using another `gcc` executable)

b) should conan control/prevent that?
I don't think so, it's beyond of conan responsibilities

c) what do I need from conan?
I just want conan to allow me to specify arbitrary build system variables (e.g. `--host` in case of configure, env vars like `CFLAGS` or `machine` in case of meson). that things might be valid, or invalid, might or might not match the settings. it's basically garbage in - garbage out in this case, e.g. if I am specifying `settings.arch=x86` and then set `CFLAGS=-march=arm`, that's my problem, not conan's.

so there are two major approaches:
1. allow `conf` to directly specify low-level build system variables.

2. let user model his specific platform in `settings.yml` and then specify mapping for build system variables for:
2.1. cmake
2.2. autotools
2.3. meson
2.4. whatever else he might need

first off, approach 2. will not work if we declare mapping from settings to build system variables just in conan - it will require conan to model everything out of the box (all the exotic platforms), and if we missed something, it will require users either to wait for conan release, or to fork.

therefore, we need to define this new extensibility feature, and it seems like it should be turing complete (e.g. it cannot be modeled just by plain `conf` mechanism) - we don't know what could be all the custom settings and sub-settings, and how complicated could be that mapping in general way.

so in practice, approach 2. will require user to put python file implementing function like:
```
def get_build_system_variables(settings):
    vars = dict()
    if settings.os == "FreeBSD" and settings.os.subsystem = "Horizon" and self.settings.os.version = "2.0":
        vars["tools.meson.mesontoolchain.target_machine"] = "horizon"
    return vars
```
that seems to be only realistic and flexible enough approach.

still, I don't see huge benefit over 1. here: e.g. nothing prevents user from producing `arm` binaries for `settings.arch=x86` in that model, right?

but it seems like it was the problem we were trying to solve with approach 2., but it clearly failed, only introducing additional complexity.

memsharded:
Closing this as outdated and not planned.

Conan 2 already supports the ``tools.meson.mesontoolchain:extra_machine_files`` that allows passing Meson full machine files, in the same spirit that CMake toolchains can be passed to ``CMakeToolchain`` for any arbitrary configuration.

PR #8348, opened 2021-01-18
Move some conan_v2 tests from integration to functional

czoido:
Changelog: omit
Docs: omit

They use the compiler to calculate the profile.


memsharded (commented):
Most of these tests can pass by defining a fixed profile/compiler in test command line arguments. The compiler is only needed to auto-detect one, but integration tests can work with a fixed one, no issue. Unless the test is really compiling something, the goal is that they won't need the ``pytest.mark.tool_compiler``

PR #8352, opened 2021-01-18
Feature: retry http requests with 5xx responses

### LINKED ISSUE #8330: [feature] Retry most http requests
Currently our CI builds sometimes fail after conan failed on simple http requests to our artifactory instance despite having reasonable general.retry and general.retry_wait config values.

After inspecting the code, I have learned that retries are implemented for file upload and download queries and
when creating requests.HTTPAdapter:
`            adapter = HTTPAdapter(max_retries=config.retry)`
https://github.com/conan-io/conan/blob/develop/conans/client/rest/conan_requester.py#L28

But requests documentation states:
> Note, this applies only to failed DNS lookups, socket connections and connection timeouts, never to requests where data has made it to the server. By default, Requests does not retry failed connections. If you need granular control over the conditions under which we retry a request, import urllib3’s Retry class and pass that instead.
https://2.python-requests.org/en/master/api/#requests.adapters.HTTPAdapter

So I suggest the following:
1. Configure requests.HTTPAdapter to retry failed http requests with 5xx status codes by passing urllib3.Retry instance instead of int.
2. Maybe, if this is implemented, retry code in file_uploader.py and file_downloader.py is no longer needed because retries will be do by requests lib?

I am ready to contribute this feature myself.

Fefer-Ivan:
Changelog: Feature: Use built-in retries in requests lib to retry http requests with _5xx_ response code.
Closes: https://github.com/conan-io/conan/issues/8330
Docs: Omit

#tags: slow

### DIFF
--- a/conans/client/rest/conan_requester.py
+++ b/conans/client/rest/conan_requester.py
@@ -5,6 +5,8 @@ import platform
 import time
 import warnings
 
+import http
+import urllib3
 import requests
 from requests.adapters import HTTPAdapter
 
@@ -25,7 +27,8 @@ class ConanRequester(object):
             self._http_requester = http_requester
         else:
             self._http_requester = requests.Session()
-            adapter = HTTPAdapter(max_retries=config.retry)
+            adapter = HTTPAdapter(max_retries=self._get_retries(config.retry, config.retry_wait))
+
             self._http_requester.mount("http://", adapter)
             self._http_requester.mount("https://", adapter)
 
@@ -60,6 +63,20 @@ class ConanRequester(object):
                                              self._client_cert_key_path)
             else:
                 self._client_certificates = self._client_cert_path
+
+    def _get_retries(self, retry, retry_wait):
+        retry = retry if retry is not None else 2
+        retry_wait = retry_wait if retry_wait is not None else 5
+        retry_status_code_set = set()
+        for status in http.HTTPStatus:
+            code = int(status)
+            if code // 100 == 5:
+                retry_status_code_set.add(code)
+        return urllib3.Retry(
+            total=retry,
+            backoff_factor=(retry * retry_wait) / (2**(retry-1)),
+            status_forcelist=retry_status_code_set
+        )
 
     def _should_skip_proxy(self, url):
         for entry in self._no_proxy_match:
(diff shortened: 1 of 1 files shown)


conans/client/rest/conan_requester.py
```diff
@@ -61,6 +64,22 @@ def __init__(self, config, http_requester=None):
...
+        if retry == 0:
+            return 0
+        retry_status_code_set = set()
+        for status in http.HTTPStatus:
+            code = int(status)
+            if code // 100 == 5:
+                retry_status_code_set.add(code)
+        return urllib3.Retry(
+            total=retry,
+            backoff_factor=(retry * retry_wait) / (2**(retry-1)),
```
Fefer-Ivan: This line doesn't look pretty because urllib3 Retry doesn't support constant waits between retries.

Instead, it has backoff_factor argument and in docs it states that it will sleep for total of 
{backoff factor} * (2 ** ({number of total retries} - 1)) seconds
https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html

conan is expected to sleep for retry * retry_wait seconds, so to achieve it I set backoff factor to
retry * retry_wait / (2 ** (retry - 1) seconds.

memsharded: I would say that this is not necessary. So far, if the number of retries was >0, HttpAdapter was already applying its internal policy of retries with backoff_factor, not following the Conan ``retry_wait``. So we can simplify and leave it.

The main change is that it will retry for some more conditions than before. I am fine with this, but if I understood correctly the docs, this will not implement the current Conan "retry" functionality, which will retry a transfer that is half-way interrupted, and is only valid for initiating calls, but once some data has hit the server, it will not be retried.

I have also checked that our requests>=2.8.1 seems good and supports passing the Retry object instead of int.

memsharded (changes requested):

memsharded (approved):

conans/client/rest/conan_requester.py
Fefer-Ivan: I removed backoff factor, but it looks like default value for it is zero, so there will be no backoff whatsoever. Maybe we can add some default small value like 0.1 seconds?

I thought about it when I was first writing this code and decided that conan already have a retry_wait config parameter so I choose backoff_factor with respect to it

memsharded: Yeah, maybe it is good to give the server at least some time, if it did a 500 or some resource limit hit. Lets give it a small default.

jgsogo (approved):

PR #8353, opened 2021-01-18
Fix: add preprocessor_definitions to Meson + CC/CXX from build requirements

### LINKED ISSUE #8311: [bug] MesonToolchain does not pick up CC/CXX from build requirements (in build profile)
The compiler paths of a build requirement with a cross compiler that sets the CC/CXX environment variables
are not available in `conan_meson_cross.ini`.

I have a build requirement that sets some compiler variables:
```
def package_info(self):
    self.env_info.CC = os.path.join(self.package_folder, "bin", "aarch64-none-elf-gcc")
    self.env_info.CC_LD = os.path.join(self.package_folder, "bin", "aarch64-none-elf-gcc")
    self.env_info.CXX = os.path.join(self.package_folder, "bin", "aarch64-none-elf-g++")
    self.env_info.CXX_LD = os.path.join(self.package_folder, "bin", "aarch64-none-elf-g++")
    self.env_info.AR = os.path.join(self.package_folder, "bin", "aarch64-none-elf-ar")
```

But the `[binaries]` section of the cross file remains empty.

When I run conan with `CC=abc conan create .... -pr:h ... -pr:b ...`, then `c = 'abc' is present in the `binaries` section.

`CFLAGS/CXXFLAGS` from a build requirement are also not propagated.

  * Operating System+version: Fedora Linux 30
  * Compiler+version: native: gcc 9.3, cross: gcc 10.2
  * Conan version: 1.32.1
  * Python version: 3.7.7

1. Create a conan recipe that packages a compiler (sets `C`/`CXX`/...). Let's name it `crosscompiler/0.1`.
2. Create a host profile that adds the package created in step 1 as a build requirement, let's name the profile: `cross`.
3. Create a build profile for your native compiler. Let's name it `native`.
4. Cross build a meson project (using `MesonToolchain` like: `conan create . meson-project/0.1@ -pr:h cross -pr:b native`

The generated `conan_meson_cross.ini` won't contain the `properties.c` values, set by `crosscompiler/0.1`.
If you'd run `CC=abc conan create . meson-project/0.1@ -pr:h cross -pr:b native`, then it would.

CC'ing @SSE4 for being the author of `MesonToolchain`.

SSE4:
Changelog: Fix: The new `MesonToolchain` now takes the declared environment variables (`CC`, `CXX`...) from build-requires and profiles to set the variables `c`, `cpp`, `c_ld`, `cpp_ld` etc, into the `conan_meson_native.ini`
Changelog: Fix: Added new `preprocessor_definitions` to new Meson build helper. 
Changelog: Fix: The new `MesonToolchain` now allows adjusting any variable before generating the `conan_meson_native.ini` file.
Docs: https://github.com/conan-io/docs/pull/2139

Closes https://github.com/conan-io/conan/issues/8311

this is mostly to align interface with `CMake`, `Make` and `MSBuild` (feature parity)
important side change: `c_args`, `c_link_args`, `cpp_args` and `cpp_link_args` are now arrays as documented at https://mesonbuild.com/Builtin-options.html#compiler-options:
```
c_args |   | free-form comma-separated list | C compile arguments to use
-- | -- | -- | --
c_link_args |   | free-form comma-separated list | C link arguments to use

cpp_args |   | free-form comma-separated list | C++ compile arguments to use
-- | -- | -- | --
cpp_link_args |   | free-form comma-separated list | C++ link arguments to use
```


SSE4:
blocked by #8311
actually, we don't capture ENV vars from the profile (and it causes build to fail), it's meson who captures them in the `build` method.
previously I didn't set `c_args`, so meson used `CFLAGS`
no I try to set `c_args` to `CFLAGS` + preprocessor_definitions, and `CFLAGS` are empty within `generate`.

conans/test/functional/toolchains/meson/test_ios.py
```diff
@@ -71,11 +71,13 @@ def env(self):
         cflags += " -isysroot " + self.xcrun.sdk_path
         cflags += " -arch " + to_apple_arch(self.arch)
         cxxflags = cflags
+        ldflags = cflags
```
solvingj: This looks wrong on the surface.

solvingj: according to @SSE4 apparently this is correct

SSE4: anyway, changed the code to be more accurate

solvingj (approved):

conan/tools/meson/toolchain.py
```diff
@@ -71,7 +73,9 @@ def __init__(self, conanfile, env=os.environ):
         self._shared = self._conanfile.options.get_safe("shared")
         self._fpic = self._conanfile.options.get_safe("fPIC")
         self.definitions = dict()
-        self._env = env
+        self.preprocessor_definitions = dict()
+        self._env = env.copy()
```
memsharded: Leave the ``env`` outside of this PR. This PR can implement the ``preprocessor_definitions`` without env

conan/tools/meson/toolchain.py
```diff
@@ -171,13 +179,12 @@ def _context(self):
...
             # https://mesonbuild.com/Builtin-options.html#compiler-options
             "cpp_std": self._to_meson_cppstd(self._cppstd) if self._cppstd else None,
-            "c_args": self._none_if_empty(self._env.get("CPPFLAGS", '') +
-                                          self._env.get("CFLAGS", '')),
-            "c_link_args": self._env.get("LDFLAGS", None),
-            "cpp_args": self._none_if_empty(self._env.get("CPPFLAGS", '') +
-                                            self._env.get("CXXFLAGS", '')),
-            "cpp_link_args": self._env.get("LDFLAGS", None),
-            "pkg_config_path": "'%s'" % os.getcwd()
+            "c_args": self._to_meson_value(self._env_array('CPPFLAGS') + self._env_array('CFLAGS')),
```
memsharded: New toolchains will provide a different way to manage the environment, please don't use it yet.

memsharded (changes requested):

conan/tools/meson/toolchain.py
SSE4: no, unfortunately we cannot.
if I override `cflags` with preprocessor definitions, it no longer uses default and no longer fetches `CFLAGS` from the environment.

memsharded: Isn't a way in Meson in which you can just define the preprocessor definitions, and append them to existing in the Meson toolchain (not in Conan)? BTW, just t make sure, these are not CFLAGS at all, these would be CPPFLAGS, right?

SSE4: I didn't find any way to just add preprocessor definitions on their own. the only way seems to be apply them to `c_args`.

memsharded: It seems we will need to wait until we correctly propagate the environment, unless a different way that doesn't require reading the CFLAGS env-var is necessary.

SSE4: okay, we can wait. no-one actually requested it (yet). I've added just for the feature parity between toolchains.

lasote:
@memsharded @SSE4 I updated the PR to use the new environment model. Check if the changes are correct and respond to the requested in the #8311

memsharded (approved):

memsharded (approved):

conan/tools/meson/toolchain.py
```diff
@@ -71,8 +74,44 @@ def __init__(self, conanfile, env=os.environ):
         self._cppstd = cppstd_from_settings(self._conanfile.settings)
         self._shared = self._conanfile.options.get_safe("shared")
         self._fpic = self._conanfile.options.get_safe("fPIC")
+        self._build_env = VirtualEnv(self._conanfile).build_environment()
```
memsharded, line 77: I have merged this PR, because it is already an improvement.
But this is a bit "hardcoded". The ``VirtualEnv`` might be used in ``generate()`` in a different way, to customize what gets pulled into the environment, and users might prioritize dependencies, [conf], environment, or other stuff, but that will not be taken into account here.

It seems the ideal is that the MesonToolchain could use the ``conanbuildenv.xx`` script, or some other way to reuse the build environment. But I ignore how Meson works for this, I guess the activation of ``conanbuildenv.xx``  is still necessary?

PR #8361, opened 2021-01-19
removing warnings for toolchain imports

memsharded:
Changelog: Fix: Remove warnings for old toolchains imports and ``generate_toolchain_files()`` calls (use new imports and ``generate()`` calls.
Docs: Omit

### DIFF
--- a/conan/tools/cmake/base.py
+++ b/conan/tools/cmake/base.py
@@ -1,5 +1,4 @@
 import textwrap
-import warnings
 from collections import OrderedDict, defaultdict
 
 from jinja2 import DictLoader, Environment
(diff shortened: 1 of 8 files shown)


uilianries (approved):
LGTM
- `toolchain` attribute generates an error: 
```
ERROR:                                                              
*****************************************************************   
******************************************************************  
The 'toolchain' attribute or method has been deprecated and removed 
Use 'generators = "ClassName"' or 'generate()' method instead.      
********************************************************************
********************************************************************`
```

- It's impossible to import `CMakeToolchain`, thus `write_toolchain_files` no longer exists:
```
 File "C:\Users\uilia\Development\conan\conan-io\build\conanfile.py", line 1, in <module>
    from conans import ConanFile, CMake, CMakeToolchain
ImportError: cannot import name 'CMakeToolchain' from 'conans' (C:\Users\uilia\Development\conan\conan-io\conans\__init__.py)
```

memsharded:
> It's impossible to import CMakeToolchain, thus write_toolchain_files no longer exists:

It is possible via ``from conan.tools.cmake import CMakeToolchain``, so it would still be pssible to try to call ``write_toolchain_files``, but it will still fail.

PR #8362, opened 2021-01-19
BugFix: MesonToolchain does not pick up CC/CXX from build requirements (in build profile)

### LINKED ISSUE #8311: [bug] MesonToolchain does not pick up CC/CXX from build requirements (in build profile)

SSE4:
closes: #8311
on top of #8353

Changelog: BugFix: MesonToolchain does not pick up CC/CXX from build requirements (in build profile)
Docs: omit

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -74,7 +74,8 @@ class MesonToolchain(object):
         self._fpic = self._conanfile.options.get_safe("fPIC")
         self.definitions = dict()
         self.preprocessor_definitions = dict()
-        self._env = env
+        self._env = env.copy()
+        self._env.update(self._conanfile.env)
 
     @staticmethod
     def _to_meson_value(value):
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -28,6 +28,8 @@ class MesonToolchain(object):
     {% if pkgconfig %}pkgconfig = {{pkgconfig}}{% endif %}
 
     [built-in options]
+    preprocessor_definitions = [{% for it, value in preprocessor_definitions.items() -%}
+    '-D{{ it }}="{{ value}}"'{%- if not loop.last %}, {% endif %}{% endfor %}]
     {% if buildtype %}buildtype = {{buildtype}}{% endif %}
     {% if debug %}debug = {{debug}}{% endif %}
     {% if default_library %}default_library = {{default_library}}{% endif %}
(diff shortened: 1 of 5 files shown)


conan/tools/meson/toolchain.py
```diff
@@ -71,7 +73,9 @@ def __init__(self, conanfile, env=os.environ):
         self._shared = self._conanfile.options.get_safe("shared")
         self._fpic = self._conanfile.options.get_safe("fPIC")
         self.definitions = dict()
-        self._env = env
+        self.preprocessor_definitions = dict()
+        self._env = env.copy()
```
memsharded, line 77: I'd prefer to wait until we have the new env management there: https://github.com/conan-io/conan/pull/8534, because partially providing it, then breaking could be probably more annoying that having to wait.

lasote:
I've updated https://github.com/conan-io/conan/pull/8353/files fixing this with the new environment model.

PR #8371, opened 2021-01-20
Feature/cmakedeps

memsharded:
Changelog: Feature: Forked ``cmake_find_package_multi`` into ``CMakeDeps``, to allow evolution without breaking.
Docs: Omit

Made a complete copy, flattening and adaptation of everything: the templates, the mixin, the inheritance. Should be functionally equivalent, but this is the start point for doing changes without touching anything else.


memsharded:
Some challenges identified:

- The transitive find_dependency() are not conditional to the build_type, and not possible to do it. Need to check if it can fail for conditional requirements conditioned on build_type

czoido (approved):

Issue #2675, opened 2018-03-24
exports_sources loses part of relative path when ../ is used


solvingj:
When `../` is used in the path for a path in the `export_sources` attribute of a conanfile, the relative directory structure in the generated package under the `exports_sources` directory behaves "unexpectedly".  It's actually unclear what the "expected" behavior would be when the `../` is used, as it's not intuitive or obvious when you consider that we're passing a pattern and not `source` or `destination` parameters. 

_Note that this is a corner case, as the vast majority of uses of `exports_sources` involve the `conanfile.py` existing in the root of the sources, not some external subdirectory of a parent directory as in this example   It's possible that the use of `../` in `exports_sources` should simply be unsupported if a predictable/sensible behavior cannot be agreed upon ._

Expected/desired directory structure in `exports_sources` directory: 
`hello/src/file.c`

Actual directory structure:
`file.c`

```
from conans import ConanFile, CMake
class HelloConan(ConanFile):
    name = "hello"
    version = "0.1"
    exports_sources = "../../hello/src/file.c",

```

If I were to define the behavior that makes the most sense for my use case, it would be to retain the directory structure with all `../`'s stripped from it, but I'm not even sure that makes sense for all cases. 

This is not a blocking issue for me, as I can define where i want the files to go with the `package()` step in my case. 

[exports_sources_ssue.zip](https://github.com/conan-io/conan/files/1844723/exports_sources_ssue.zip)

- [x] I've read the [CONTRIBUTING guide](https://raw.githubusercontent.com/conan-io/conan/develop/.github/CONTRIBUTING.md).
- [x] I've specified the Conan version, operating system version and any tool that can be relevant.
- [x] I've explained the steps to reproduce the error or the motivation/use case of the question/suggestion.

memsharded:
Submitted https://github.com/conan-io/conan/pull/2676, but not sure if breaking.

In the meantime you might try to workaround it with:

```python
class HelloConan(ConanFile):
    name = "hello"
    version = "0.1"
    exports_sources = "../../hello*src*file.c",
```

sourcedelica:
Related to this, we need a way to `exports` a Python script from a common directory which could be a parent or sibling to the `conanfile` directory (also discussed in #1271).   The Python script could contain a `Conanfile` base class and/or functions.

lasote:
I think we should postpone to 1.5 at least because the fix is not very clear for me and the exports method, that could be a fix for this issue, it is neither clear.

[...]

PR #2676, opened 2018-03-25
fix exports relative paths

memsharded:
- [x] Refer to the issue that supports this Pull Request.
Close #2675

Not sure if this could be considered breaking or fixing a bug.

PR #8374, opened 2021-01-21
Bugfix: Fix file copying with .. patterns and keep_path=True

JoachimKuebart-TomTom:
Changelog: Bugfix: Fix file copying with .. patterns and keep_path=True
Docs: N/A

When copying files using a pattern such as `"../libfoo/*.c"`, the expected outcome in the destination directory is:

libfoo/
└── file.c

Instead, all files matched by the pattern are currently flattened into the destination directory. This PR fixes that.

### DIFF
--- a/conans/client/file_copier.py
+++ b/conans/client/file_copier.py
@@ -90,10 +90,9 @@ class FileCopier(object):
     def _copy(self, base_src, pattern, src, dst, symlinks, ignore_case, excludes, keep_path,
               excluded_folders):
         # Check for ../ patterns and allow them
-        if pattern.startswith(".."):
-            rel_dir = os.path.abspath(os.path.join(base_src, pattern))
-            base_src = os.path.dirname(rel_dir)
-            pattern = os.path.basename(rel_dir)
+        while pattern.startswith(".."):
+            base_src = os.path.abspath(os.path.join(base_src, ".."))
+            pattern = pattern[3:]
 
         src = os.path.join(base_src, src)
         dst = os.path.join(self._dst_folder, dst)
--- a/conans/test/functional/command/export/export_test.py
+++ b/conans/test/functional/command/export/export_test.py
@@ -142,7 +142,7 @@ class TestConan(ConanFile):
         client.run("export . lasote/stable")
         ref = ConanFileReference("Hello", "1.2", "lasote", "stable")
         export_path = client.cache.package_layout(ref).export()
-        content = load(os.path.join(export_path, "file.txt"))
+        content = load(os.path.join(export_path, "sibling/file.txt"))
         self.assertEqual("Hello World!", content)
 
     def test_code_several_sibling(self):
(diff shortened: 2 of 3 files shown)


JoachimKuebart-TomTom:
I realise now that it's a breaking change as evidenced by the `functional/command/export` tests I had to fix. So I guess it's not really feasible, at least without deprecation.

I still find the current behaviour surprising, i.e. the fact that `exports` is interpreted as `keep_path=True` *unless* it starts with `../`.

conans/client/file_copier.py
```diff
@@ -90,10 +90,9 @@ def __call__(self, pattern, dst="", src="", keep_path=True, links=False, symlink
     def _copy(self, base_src, pattern, src, dst, symlinks, ignore_case, excludes, keep_path,
               excluded_folders):
         # Check for ../ patterns and allow them
-        if pattern.startswith(".."):
-            rel_dir = os.path.abspath(os.path.join(base_src, pattern))
-            base_src = os.path.dirname(rel_dir)
-            pattern = os.path.basename(rel_dir)
+        while pattern.startswith(".."):
```
SSE4, line 93: this check is not correct as `..` might be a part of the file name (e.g. run `touch ...1`)

memsharded:
I think I tried something similar some time ago: https://github.com/conan-io/conan/pull/2676, to resolve issue https://github.com/conan-io/conan/issues/2675, but it was considered breaking and was closed. We might try to fix it in Conan 2.0.
At the moment the best way to approach this could be to use the ``export()`` and ``export_sources()`` method instead of the attribute.

conans/client/file_copier.py
JoachimKuebart-TomTom: No it's not, and I believe it wasn't before, or was it? Only if the code was a no-op in such cases, which it isn't in general because it reduces the pattern to just its `basename()`.

It could be made a little more correct by saying
```
        while pattern.startswith(".." + os.sep) or \
            (os.altsep is not None and pattern.startswith(".." + os.altsep)):
```
ugly, but perhaps safer?

JoachimKuebart-TomTom:
@memsharded I can confirm this is a duplicate of #2675.

conans/test/functional/command/export/export_test.py
```diff
@@ -142,7 +142,7 @@ class TestConan(ConanFile):
         client.run("export . lasote/stable")
         ref = ConanFileReference("Hello", "1.2", "lasote", "stable")
         export_path = client.cache.package_layout(ref).export()
-        content = load(os.path.join(export_path, "file.txt"))
+        content = load(os.path.join(export_path, "sibling/file.txt"))
```
memsharded, line 145: Yes, this is what would be breaking to other users relying on this.

memsharded (commented):
This will need to wait until Conan 2.0. Please use the ``export()`` and ``exports_sources()`` methods in the meantime. Thanks again for contributing this!

memsharded:
FileCopier is going to be replaced by a explicit ``copy()`` free function tool. Things will be more explicit, including the source and target folders, so this will no longer be an issue. This was risky for Conan 1.X, potentially breaking, and the new tools in 2.0 will address this more natively. (Also, using ``export()`` method is preferred over the class attribute for any special case), so I think this can be closed. Thanks!

Issue #8375, opened 2021-01-21
[question] Does diamond graphs ``configure()`` methods that remove options shared/fpic could be broken?


memsharded:
It is possible if evaluated more than once, with different branches values.

SSE4:
I spotted a serious issue in `validate` implementation: #8383

memsharded:
This is not an issue at all in the ``validate()``, it is working as expected. It is an issue in the options processing: https://github.com/conan-io/conan/pull/8383#pullrequestreview-574079761

memsharded:
The new graph resolution model in Conan 2.0 will try not to re-evaluate same recipe when diamonds, just check consistency. So that might solve this issue, which is otherwise simply too risky to solve for Conan 1.X

memsharded:
Solved in 2.0, closing.

PR #8383, opened 2021-01-22
- add new validate test

SSE4:
https://github.com/conan-io/docs/issues/2003
#8375
https://github.com/conan-io/conan-center-index/issues/4261

I think I have found a serious bug in the newly added `validate` function:
1. library dep has two options, `fPIC` and `shared`
2. `shared=False` is default, `fPIC=True` is default
3. `fPIC` is removed for `shared=False` in `config_options` (according to the https://docs.conan.io/en/latest/mastering/conditional.html?highlight=conditional)
4. consumer overrides `dep:shared=True`
5. consumer checks in validate: `dep:shared == True` - OK
6. consumer checks in validate: `dep:fPIC == True` - raises!

now we see that validates incorrectly computes value of `fPIC` option and shows a following error message:
```
ERROR: pkg/0.1: Error in validate() method, line 13
	assert self.options["dep"].fPIC == True
	ConanException: option 'fPIC' doesn't exist
Possible options are ['shared']
```

this is incorrect, since `dep:shared=True`, `fPIC` option exists

UPD1: validates works correctly if the option is removed in `configure` instead of `config_options`.
UPD2: validates also works incorrectly if the option is overridden from the command line (`-o dep:shared=True`)

### DIFF
--- a/conans/test/integration/package_id/test_validate.py
+++ b/conans/test/integration/package_id/test_validate.py
@@ -247,3 +247,43 @@ class TestValidate(unittest.TestCase):
                       "exist for this configuration):", client.out)
         self.assertIn("dep/0.1: Invalid ID: Windows not supported", client.out)
         self.assertIn("pkg/0.1: Invalid ID: Invalid transitive dependencies", client.out)
+
+    def test_override(self):
+        client = TestClient()
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile, tools
+
+
+            class Lib(ConanFile):
+                settings = "os", "arch", "compiler", "build_type"
+                options = {"shared": [True, False], "fPIC": [True, False]}
+                default_options = {"shared": False, "fPIC": True}
+
+                def config_options(self):
+                    if not self.options.shared:
+                        del self.options.fPIC
+
+                def build(self):
+                    pass
+        """)
+
+        client.save({"conanfile.py": conanfile})
+        client.run("export . dep/1.0@")
+
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile, tools
+
+
+            class App(ConanFile):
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Good catch! 

This is not a ``validate()`` error, it is a ``config_options`` or ``options`` bug. The ``del self.options.fPIC`` is being executed and the option is being deleted, it is expected that the ``validate()`` will fail.

memsharded:
Actually the documentation says it explicitly:

> config_options() is used to configure or constraint the available options in a package, before they are given a value. So when a value is tried to be assigned it will raise an error. For example, let’s suppose that a certain package library cannot be built as shared library in Windows, it can be done:

>def config_options(self):
        del self.options.shared

>This will be executed before the actual assignment of options (then, such options values cannot be used inside this function), 

**such options values cannot be used inside this function**

It is documented behavior. The following test using ``configure()`` seems to work fine:

```python
    def test_override(self):
        client = TestClient()
           from conans import ConanFile
           class Lib(ConanFile):
               options = {"shared": [True, False], "fPIC": [True, False]}
               default_options = {"shared": False, "fPIC": True}
               def configure(self):
                   if not self.options.shared:
                       self.output.warn("REMOVING fPIC!!!!")
                       del self.options.fPIC
           """)

        client.run("export . dep/1.0@")

           from conans import ConanFile
           class App(ConanFile):
               requires = "dep/1.0"

               def configure(self):
                   self.options["dep"].shared = True

               def validate(self):
                   assert self.options["dep"].shared == True
                   assert self.options["dep"].fPIC == True
           """)

        client.save({"conanfile.py": conanfile}, clean_first=True)
        client.run("create . pkg/0.1@ --build missing")
        print(client.out)
```

PR #8394, opened 2021-01-25
untangle a bit graph_info to prepare for removal

memsharded:
Changelog: Omit
Docs: Omit

#revisions: 1
#tags: slow


memsharded:
This is just a refactor, to depend less on the "graph_info" object, making explicit the profiles and lockfile arguments.

czoido (approved):

PR #8428, opened 2021-02-02
Remove disabled_revisions mode from 2.0: Clean enabled_revisions

czoido:
Changelog: Remove: Some cleaning after removing conan api v1.
Docs: omit

### DIFF
--- a/conans/client/cmd/search.py
+++ b/conans/client/cmd/search.py
@@ -49,10 +49,6 @@ class Search(object):
         if not remote_name:
             return self._search_packages_in_local(ref, query, outdated)
 
-        if ref.revision and not self._cache.config.revisions_enabled:
-            raise ConanException("Revisions not enabled in the client, specify a "
-                                 "reference without revision")
-
         if remote_name == 'all':
             return self._search_packages_in_all(ref, query, outdated)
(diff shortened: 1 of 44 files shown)


conans/client/cmd/search.py
```diff
@@ -70,7 +66,7 @@ def _search_packages_in_local(self, ref=None, query=None, outdated=False):
 
         if outdated:
             ordered_packages = filter_outdated(ordered_packages, recipe_hash)
```
memsharded, line 68: I want to understand what "outdated" means now with revisions. Apparently nothing, but at least some tests should break. I think we want to remove this.

conans/model/package_metadata.py
```diff
@@ -45,15 +43,15 @@ def revision(self):
 
     @revision.setter
     def revision(self, r):
-        self._revision = DEFAULT_REVISION_V1 if r is None else r
+        self._revision = r
```
memsharded, line 47: A ``assert r is not None`` here is the typical thing that can be useful to spot errors quickly

memsharded (approved):

memsharded (approved):

PR #8439, opened 2021-02-04
Extract the defaults of cpp_info so they can be different

lasote:
Changelog: omit
Docs: omit

Pure refactor. It helps me to move forward the layouts feature because:
  - The checks about using components and root at the same time bite otherwise.
  - The defaults make only sense for a cppinfo of a package.

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -232,17 +232,34 @@ class _CppInfo(object):
 
 class Component(_CppInfo):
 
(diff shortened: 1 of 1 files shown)


memsharded (approved):

danimtb (approved):
I hate to have another level of indirection here, but it is what it is... Hopefully we will improve the code around cpp_info in conan 2.0

PR #8445, opened 2021-02-04
Tests/modernize

### LINKED ISSUE #8434: [bug] CMakeDeps not working correctly for custom filenames and target names.
Reported by @mpusz

For some reason the forked CMakeDeps is behaving differently to ``cmake_find_package_multi`` with regards to generated target names.

memsharded:
Changelog: Bugfix: Fixing a `CMakeDeps` bug with components, not finding the _conan_macros.cmake_ file.
Docs: Omit

Fix: https://github.com/conan-io/conan/issues/8434

Also a refactor of test that highly reduce time, reusing same built package.


conan/tools/cmake/cmakedeps.py
```diff
@@ -358,7 +358,7 @@ class CMakeDeps(object):
     components_target_build_type_tpl = Template(textwrap.dedent("""\
         ########## MACROS ###########################################################################
         #############################################################################################
-        include(${{CMAKE_CURRENT_LIST_DIR}}/cmakedeps_macros.cmake)
+        include(${CMAKE_CURRENT_LIST_DIR}/cmakedeps_macros.cmake)
```
lasote, line 361: there is no test for this, right? components and CMakeDeps is barely tested

memsharded: Not really, I think it was partially discovered by some changes I did in the test, but not a explicit one. The thing is that we will move (copying first) all existing tests in cmake_find_package_multi to CMakeDeps, so all this will be more than covered, so it is probably not deserving its own new test for this.

czoido (approved):

PR #8456, opened 2021-02-06
Remove/outdated

memsharded:
Changelog: Remove: ``--build=outdated`` no longer make sense with revisions, can be removed
Docs: Omit


memsharded (commented):
LGTM

PR #8457, opened 2021-02-08
New AutotoolsDeps, AutotoolsToolchain helpers in conan.tools.gnu

### LINKED ISSUE #7070: [bug] Invert the logic of _GLIBCXX_USE_CXX11_ABI ?
  * Compiler+version: gcc9
  * Conan version: 1.25.1
  * Python version: 3.6.9

The default for _GLIBCXX_USE_CXX11_ABI is 1 for several years now. Can we invert the logic of compiler.libcxx to use libstdc++ancient for the older version and use libstdc++ by default, or have libstdc++ _NOT_ emit _GLIBCXX_USE_CXX11_ABI ?

The reason I'm asking this is that we have a rule for formatting global macros in clang-tidy which is tripped by this macro, and there is no way to suppress it in tidy, since it does not come from a source file, but it is injected on the command line.

I suppose this is a bug in tidy, but there is some redundancy in Conan / CMake injecting this variable explicitly when it is not needed for quite some time now.

memsharded:
Changelog: Feature: New AutotoolsDeps, AutotoolsToolchain helpers in conan.tools.gnu
Docs: https://github.com/conan-io/docs/pull/2057

Close https://github.com/conan-io/conan/issues/7070 (will invert the logic in _GLIBCXX_USE_CXX11_ABI)


conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+            except ConanException as exc:
+                self._conanfile.output.warn(str(exc))
+                target = None
+        else:
+            target = None
+
+        if hasattr(self._conanfile, 'settings_build'):
+            os_build, arch_build = get_build_os_arch(self._conanfile)
+        else:
+            # FIXME: Why not use 'os_build' and 'arch_build' from conanfile.settings?
```
SSE4: if it's new helper, can we avoid all these os_build/arch_build mess and just use modern two-profile approach?

conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+            os_build = detected_os() or platform.system()
```
SSE4: I'd say, that may be going forward, avoid auto-detecting altogether and specify everything explicitly in profile/settings, to ensure it's reproducible?

conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+                                                  " ".join(triplet_args))
+                self._conanfile.output.info("Calling:\n > %s" % command)
+                self._conanfile.run(command, win_bash=self._win_bash, subsystem=self.subsystem)"""
+
+        cmd = "bash -c 'source autotoolsdeps.sh && source autotools.sh && %s/configure'" % configure_dir
+        self._conanfile.output.info("Calling:\n > %s" % cmd)
+        self._conanfile.run(cmd, win_bash=self._win_bash, subsystem=self.subsystem)
+
+    def _configure_help_output(self, configure_path):
+        from six import StringIO  # Python 2 and 3 compatible
```
SSE4: for the new helper, I think we can avoid python 2 and six

conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+                                                       cpu_count_option]),
+                                win_bash=self._win_bash, subsystem=self.subsystem)"""
+
+        make_program = self._conanfile.conf["tools.gnu"].make_program
+        if make_program is None:
+            make_program = "mingw32-make" if platform.system() == "Windows" else "make"
+        if platform.system() == "Windows":
+            cmd = "autotoolsdeps.bat && autotools.bat && {}".format(make_program)
+        else:
+            cmd = "bash -c 'source autotoolsdeps.sh "\
```
SSE4: I think we shouldn't hard-code bash here

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -0,0 +1,62 @@
...
+        cpp_flags = []
+        include_paths = ['-I"%s"' % p for p in self.include_paths]
+        cpp_flags.extend(include_paths)
+        cpp_flags.extend(["-D%s" % define for define in self.defines])
+
+        # Libs
+        libs = ["-l%s" % library for library in self.libs]
+
+        # Ldflags
+        # TODO: Discuss, should the helper filter frameworks based on compiler?
```
SSE4, line 75: I vote to keep the least possible logic in these helpers. if some filtering has to be done, it doesn't belong here. and IMO it's the responsibility of the recipe author to ensure the correct CppInfo is built for the given configuration.

conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+            cpu_count_option = (("-j%s" % cpu_count(output=self._conanfile.output))
+                                if ("-j" not in str_args and "nmake" not in make_program.lower())
+                                else None)
+            self._conanfile.run("%s" % join_arguments([make_program, target, str_args,
+                                                       cpu_count_option]),
+                                win_bash=self._win_bash, subsystem=self.subsystem)"""
+
+        make_program = self._conanfile.conf["tools.gnu"].make_program
+        if make_program is None:
+            make_program = "mingw32-make" if platform.system() == "Windows" else "make"
```
SSE4: I also don't think it's always desired to enforce `mingw32-make` on Windows. it should go from the configuration.

conan/tools/gnu/autotools.py
```diff
@@ -0,0 +1,239 @@
...
+                if arg:
+                    if isinstance(arg, list):
+                        ret.extend(arg)
+                    else:
+                        ret.append(arg)
+            return ret
+
+        tmp_compilation_flags = copy.copy(self.flags)
+
+        if tools.is_apple_os(self._os):
```
SSE4: I don't think it belongs here as well. let's keep all platform-specific code as minimal as possible, and isolated. the way we did it previously was proven to be fragile and doesn't scale well.

SSE4 (changes requested):
may we try to avoid complicated logic (e.g. platform-specific), at least in the first iteration?

conan/tools/env/environment.py
```diff
@@ -0,0 +1,138 @@
+import textwrap
```
SSE4: maybe this one deserves its own first class generator? and new Virtual*Env generators might be based on that one?

conan/tools/gnu/autotools.py
memsharded: Yes, definitely, it is very ongoing work, I plan to refactor everything and use the 2 profiles.

conan/tools/gnu/autotools.py
memsharded: 100% agree, there will be no auto-detection here, only deterministic from profiles.

conan/tools/gnu/autotools.py
memsharded: Agree, ongoing work.

conan/tools/gnu/autotools.py
memsharded: Unless it is the only way to make it work reproducibly. The autotoolsdeps.sh, etc are using bash. If we are supporting other shells, we need tests, I don't want to add things into the new codebase that cannot be tested.

conan/tools/gnu/autotools.py
memsharded: It is probably  a sane default in windows, the same as ``make`` is in other platforms. But we will see as we keep adding support for other subsystems, I am fine with removing it.

conan/tools/gnu/autotools.py
memsharded: Agree, this code is not used at the moment, needs to be worked on.

conan/tools/env/environment.py
memsharded: Not yet. This could be a first class model, like ``self.env_info = Environment()`` in the recipes. This is separated in another branch and will have a separate PR once it is stabilized.

conan/tools/gnu/autotools.py
SSE4: the make command is not hard-coded and got from the config.
the same, more or less, should be done for the bash I believe.
also, maybe it's enough to just use plain `sh` or `$SHELL`?
I don't think we need advanced bash-specific features, and it would be nice to be portable and require only posix sh.
bash is not always available and not a default shell everywhere.

memsharded: The ``make`` command is hardcoded, most of the times it doesn't come from config: https://github.com/conan-io/conan/blob/2ccc79c48ce42972a19c7d3587e4ab8b585f220a/conans/client/build/autotools_environment.py#L237

I guess you are aware that many recipes in ConanCenter are doing ``make_program=which("make") or which("mingw32-make")`` to be able to work on Windows. No config here, but recipes hardcoding it.

Same as ``bash``, hardcoded here: https://github.com/conan-io/conan/blob/2ccc79c48ce42972a19c7d3587e4ab8b585f220a/conans/client/tools/oss.py#L389.

The fact that they have configs, doesn't mean that they are not hardcoded in the codebase. Same I am doing here, we can discuss some defaults as we gain knowledge, but I am not introducing new hardcoded things.

The new ``Environment`` files at the moment only work on ``bash``. If we can generalize them to work on different shells, or to provide different generated files, for different shells, good for me.

SSE4: I know there are recipes doing that, as I wrote some of them.
that might be a sign we're not managing things the best way.
in some cases, `mingw32-make` is valid defaults, in some cases valid default could be `nmake` or just plain `make`.
that might be the reason we need to provide helper to deduce sane default based on settings, e.g. it may return MinGW make, nmake, GNU make or BSD make depending on system and compiler.
but that we know already the logic doesn't believe  to the AutoToolsBuildEnvironment or recipes.

memsharded: Yes, agree, a helper would be the best. This helper can hardcode some defaults for some given configurations (settings, etc), and always have a ``[conf]`` to be able to override it. 

But lets wait a bit and have a minimum set of build helpers, just the most common "AutotoolsXXX" ones, that work well and make recipes simpler for a majority of cases, before extracting new helpers like ``get_make_program(conanfile)``, when we gain more insights about these defaults, our new helpers will be closer to something stable.

memsharded:
I would like some help from @SSE4 and @czoido for CI. The tests from ``pytest conans\test\functional\toolchains\gnu\test_autotools.py -s`` pass in my computer (all 4 for Windows execute, I have installed Msys, Mingw and Cygwin), but they fail in CI. The failure is at an early stage (CMake build of dependency). Some things that could help:

- (@czoido ) Have all different subsytems: Msys, Cygwin, Mingw (not a subsystem, but yeah), have different gcc compiler versions. That could help to identify issues and make sure it is building with the right subsystem.
- (@czoido) running the tests manually in the CI server, just login in the machine, 3 tests fail and 1 is skipped. I am not sure why, I guess there is something configured in CI, that is not happening while running manually.
- Check what is happening with CMake builds. Make sure the tests are running in CI. Maybe we can have tests similar to what @SSE4 wrote directly with gcc, but with CMake (no conan, no recipes, pure CMake to build under the different subsystems).

SSE4:
I'd say, let's go step by step, if possible.
- we have working tests which are using plain makefiles, that's nice
- let's add some (hello-world) tests using plain `./configure` + `make` (no build helper), for all subsystems
- let's add some (hello-world) tests using plain `cmake` + `make` (no build helper), for all subsystems

also, that might be very possible, some tools could be picked from the wrong sub-system (e.g. if tool such as sed or whatever is missing in MSYS/Cygwin installation, tool with the same name could be used from Git bash, or WSL, if it's in PATH)

conans/test/functional/toolchains/gnu/test_autotools.py
```diff
@@ -0,0 +1,206 @@
...
+        arch=x86_64
+        build_type=Release
+        """)
+    out = build_windows_subsystem(gcc, make_program="mingw32-make")
+    print(out)
+    assert "__MSYS__" not in out
+    assert "main2 __MINGW64__1" in out
+
+
+@pytest.mark.tool_mingw64
```
SSE4, line 133:
```suggestion
@pytest.mark.tool_msys2
@pytest.mark.tool_mingw64
```
MinGW64 works on top of MSYS2

conans/test/functional/toolchains/gnu/test_autotools.py
```diff
@@ -0,0 +1,206 @@
...
+        build_type=Release
+        """)
+    out = build_windows_subsystem(gcc, make_program="make")
+    print(out)
+    assert "__MSYS__" not in out
+    assert "MINGW" not in out
+    assert "main2 __CYGWIN__1" in out
+
+
+@pytest.mark.tool_mingw
```
SSE4:
```suggestion
@pytest.mark.tool_msys2
@pytest.mark.tool_mingw32
```
not sure about an old `tool_mingw`, let's use the one from the MSYS2

memsharded: It is another mingw, quite popular that doesn't work under the umbrella of Msys2. We need to make sure it is working too, the commands, paths, syntax, might be different of what Msys2 is doing, so lets keep this variant too.

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _rpaths_link(self):
+        # TODO: Not implemented yet
+        pass
+
+    # TODO: Apple: tools.apple_deployment_target_flag,
+    # TODO:  tools.XCRun(self._conanfile.settings).sdk_path
+    # TODO: "-arch", tools.to_apple_arch(self._arch)
+
+    def _cxx11_abi_define(self):
```
SSE4, line 41: belongs to `conan/tools/_compilers.py`?

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -0,0 +1,111 @@
...
+        libcxx = settings.get_safe("compiler.libcxx")
+        if not libcxx:
+            return
+
+        compiler = settings.get_safe("compiler.base") or settings.get_safe("compiler")
+        if compiler == "gcc":
+            if libcxx == 'libstdc++':
+                return '_GLIBCXX_USE_CXX11_ABI=0'
+
+    def _libcxx(self):
```
SSE4, line 54: belongs to `conan/tools/_compilers.py`?

conan/tools/gnu/autotoolsgen.py
```diff
@@ -0,0 +1,35 @@
...
+        run_env = self.env.run_environment()
+        return run_env
+
+    def generate(self):
+        build_env = self.build_environment()
+        run_env = self.run_environment()
+        # FIXME: Use settings, not platform Not always defined :(
+        # os_ = self._conanfile.settings_build.get_safe("os")
+        if platform.system() == "Windows":
+            build_env.save_bat("conanbuildenv.bat")
```
SSE4, line 31: name is too generic? may overwrite other generators, if multiple are used together...

conan/tools/gnu/autotoolstoolchain.py
memsharded: The idea is to factor out to a common place like ``_compilers.py`` only when there is a repetition/duplicated code needed somewhere else.

conan/tools/gnu/autotoolsgen.py
memsharded: ``conanxxxx`` filenames are generated by Conan (probably we should document this more explicitly too).
This filename is automatically used by ``self.run()`` and is the same filename generated by ``VirtualEnv`` (to be automatically injected too). The idea is that this file is already the aggregation of the necessary build environments, in a single file, oriented also to make user life easier at the consumer flow. 

Ideas welcome, any suggestion of other filename?

SSE4: I am not sure if I got the idea. what if we add similar generator e.g. for premake, which will also need `VirtualEnv`.
and what if project uses both `autotoolsgen` and `premake` (e.g. it has to sub-directories with different build systems)
now both will try to create `conanbuildenv.bat` to override each other?

memsharded: ``Autotools`` generators do not create ``conanbuildenv.bat``, ``AutotoolsGen`` that wraps both Autotools and VirtualEnv does. In the same spirit, if you want to combine Autotools and Premake in an ordered, structured way, you can do it directly in the recipe, aggregating environments before saving ``conanbuildenv.bat`` or creating a generator that aggregates them and using it.

PR #8462, opened 2021-02-08
CMakeDeps with components broken inclusion of "CMAKE_CURRENT_LIST_DIR"

lasote:
Changelog: omit
Docs: omit

Bug only present in develop.


conans/test/functional/generators/cmake_deps_test.py
```diff
@@ -0,0 +1,39 @@
...
+
+               def generate(self):
+                   cmake = CMakeDeps(self)
+                   cmake.generate()
+
+           """)
+
+    t.save({"conanfile.py": conanfile})
+    t.run("install .")
+    contents = load(os.path.join(t.current_folder, "mylibTarget-release.cmake"))
```
memsharded, line 38: ``contents = t.load("mylibTarget-release.cmake")``

memsharded (commented):
This is already covered by https://github.com/conan-io/conan/pull/8445, but it has not been merged.

The test shouldn't go to ``functional`` but to ``integration`` as it doesn't really require CMake to run.

Issue #8205, opened 2020-12-14
[question]  ERROR: Attempt to modify locked lib when trying to create recipe revision lockfile


michaelmaguire:
We are following through the examples in https://docs.conan.io/en/latest/versioning/lockfiles/ci.html#package-pipeline and we get to a point where we have already run the "package" lockfile creation steps:

```
$ cd libb

# Derive one lockfile per profile/configuration
$ conan lock create conanfile.py --name=libb --version=0.2
  --user=user --channel=testing --lockfile=../locks/libb_deps_base.lock
  --lockfile-out=../locks/libb_deps_debug.lock -s build_type=Debug
$ conan lock create conanfile.py --name=libb --version=0.2
  --user=user --channel=testing --lockfile=../locks/libb_deps_base.lock
  --lockfile-out=../locks/libb_deps_release.lock

# Create the package binaries, one with each lockfile
$ conan create . libb/0.2@user/testing --lockfile=../locks/libb_deps_release.lock
$ conan create . libb/0.2@user/testing --lockfile=../locks/libb_deps_debug.lock
```

and now we were looking at this optional step:

> When using revisions (not this example), it is important to capture the recipe revision, and lock it too. We can capture the recipe revision doing an export, creating a new libb_base.lock lockfile:

So we ran in our CI after the steps above:

```
 conan export . libb/0.2@user/testing --lockfile=../locks/libb_deps_base.lock
  --lockfile-out=../locks/libb_base.lock
```

But we get this output  (note: we used slightly different lockfile output names to be clearer about package versus recipe lockfiles):

```
2020-12-14 05:31:31 |  [Conan_pipeline_master] $ cmd.exe /C "conan export . libb/0.2@conan/stable --lockfile=package_deps_base.lock --lockfile-out=recipe_base.lock && exit %%ERRORLEVEL%%"
2020-12-14 05:31:33 |  [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'url'. It is recommended to add it as attribute
2020-12-14 05:31:33 |  [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'license'. It is recommended to add it as attribute
2020-12-14 05:31:33 |  [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'description'. It is recommended to add it as attribute
[...]

PR #8463, opened 2021-02-08
Fix profile repeated build requires

memsharded:
Changelog: Bugfix: Fix repeated ``build_requires``, including conflicting versions in profile composition or inclusion that repeats ``[build_requires]`` values.
Docs: Omit

Comes from https://github.com/conan-io/conan/issues/8205#issuecomment-775032229

### DIFF
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -137,7 +137,7 @@ def _load_profile(text, profile_path, default_folder):
         for include in profile_parser.get_includes():
             # Recursion !!
             profile, included_vars = read_profile(include, cwd, default_folder)
-            inherited_profile.update(profile)
+            inherited_profile.update_profile(profile)
             profile_parser.update_vars(included_vars)
 
         # Apply the automatic PROFILE_DIR variable
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -137,7 +137,7 @@ def _load_profile(text, profile_path, default_folder):
         for include in profile_parser.get_includes():
             # Recursion !!
             profile, included_vars = read_profile(include, cwd, default_folder)
-            inherited_profile.update_profile(profile)
+            inherited_profile.compose(profile)
             profile_parser.update_vars(included_vars)
 
         # Apply the automatic PROFILE_DIR variable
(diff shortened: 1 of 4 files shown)


conans/model/profile.py
```diff
@@ -89,15 +90,28 @@ def dumps(self):
...
         self.options.update(other.options)
+        # It is possible that build_requires are repeated, or same package but different versions
         for pattern, req_list in other.build_requires.items():
-            self.build_requires.setdefault(pattern, []).extend(req_list)
+            existing_build_requires = self.build_requires.get(pattern)
+            existing = OrderedDict()
+            if existing_build_requires is not None:
+                for br in existing_build_requires:
+                    r = (ConanFileReference.loads(br)
+                         if not isinstance(br, ConanFileReference) else br)
```
jgsogo:
```suggestion
                    r = ConanFileReference.loads(br) \
                         if not isinstance(br, ConanFileReference) else br
```

better if it doesn't look like a generator.

conans/model/profile.py
```diff
@@ -89,15 +90,28 @@ def dumps(self):
...
+                    existing[r.name] = br
+            for req in req_list:
+                r = (ConanFileReference.loads(req)
+                     if not isinstance(req, ConanFileReference) else req)
```
jgsogo:
```suggestion
                r = ConanFileReference.loads(req) \
                     if not isinstance(req, ConanFileReference) else req
```

jgsogo (approved):

jgsogo:
I like the changes, I think they make sense, IMO existing test was testing the bug.

conans/client/profile_loader.py
```diff
@@ -137,7 +137,7 @@ def _load_profile(text, profile_path, default_folder):
         for include in profile_parser.get_includes():
             # Recursion !!
             profile, included_vars = read_profile(include, cwd, default_folder)
-            inherited_profile.update(profile)
+            inherited_profile.update_profile(profile)
```
SSE4: IMO `profile.update_profile()` looks too redundand. `profile.update()` is nice and self-explaining.

conans/model/profile.py
memsharded: It is a line break, otherwise it needs a backslash

conans/client/profile_loader.py
memsharded: Until you try to search for occurrences, or code calling this method and it becomes almost impossible to know, and you need to rely on CI running to find all occurrences.

conans/model/profile.py
jgsogo: Added to the suggestion

conans/client/profile_loader.py
memsharded: Agree that the ``update_profile()`` is not great and redundant. I have replaced it for ``compose()`` because it doesn't conflict with common dicts update() and also because in our docs we refer to this operation as "Profile composition".

SSE4 (approved):

PR #8470, opened 2021-02-09
Add windows subsystems testing

czoido:
Changelog: omit
Docs: omit

Add windows subsystems to test with pytest.

Just testing a way of adding tools to paths...

### DIFF
--- a/conans/test/functional/subsystems_build_test.py
+++ b/conans/test/functional/subsystems_build_test.py
@@ -23,4 +23,4 @@ class TestSubsystemsBuild:
     def test_tool_not_available(self):
         client = TestClient()
         client.run_command('uname', assert_error=True)
-        assert "command not found" in client.out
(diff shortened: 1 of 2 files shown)


conans/test/functional/subsystems_build_test.py
```diff
@@ -0,0 +1,28 @@
+import platform
+
+import pytest
+
+from conans.test.utils.tools import TestClient
+
+
+@pytest.fixture(scope="session")
+def client():
```
memsharded: Unnecessary, and ``session`` is very risky. Remove.

memsharded (approved):
Looking very good, thanks.

I think next step is to start generalizing a bit the ``tool_locations``, with some other mechanism than env-vars (could it be user files?)

SSE4 (approved):

PR #8490, opened 2021-02-12
Remove old migrations for Conan 2.0

### LINKED ISSUE #8441: [Conan 2.0] Remove old migrations
I think it is safe to assume that Conan 2.0 doesn't need to run migrations from older (rather than latest 1.X at the moment of 2.0 GA) Conan version.

So all migrations can be removed (and their tests)

uilianries:
Changelog: Feature: Remove old migration settings
Docs: Omit
closes #8441


memsharded (changes requested):
The migrations that can be removed are those like:

- test_migrate_config_install
- test_migration_to_default_profile
- etc

Those migrations can be completely removed.

Regarding the settings migration we want to take a different approach, at least one of:
- The settings to be migrated don't need to be stored in Conan, that is a nightmare. At the moment of creating them in the cache, a copy is made. That local copy is the one that can be used to check if the file has been modified or not.
- Move to round-trip pyyaml, that can actually add and migrate things for real in the ``settings.yml``.

I recommend splitting this in 2 PR:
- One PR to remove the old migrations (not the settings.yml one)
- Another PR to implement the update-if-not-modified, in which the comparison is done against a local copy of the settings.yml.

And then we will discuss about possibilities for round-trip, if it makes sense and the possible problems.

uilianries:
It's not entirely clear to me, because we have a lot of code involving Migration. I could remove all migration stuff, since tests and classes like Migrator. As we will adopt a new way for migration, I don't think they will be necessary. However, I'm stepping on eggs, not sure if we will re-use some parts of current migration.

memsharded:
You can leave the infrastructure for migrations, that is ok.
But the specific migrations for 1.X -> 1.Y can be removed, together with the tests.

```python
 if old_version < Version("1.0"):
            _migrate_lock_files(cache, self.out)

        if old_version < Version("1.12.0"):
            migrate_plugins_to_hooks(cache)

        if old_version < Version("1.13.0"):
            # MIGRATE LOCAL CACHE TO GENERATE MISSING METADATA.json
            _migrate_create_metadata(cache, self.out)

        if old_version < Version("1.14.0"):
            migrate_config_install(cache)

        if old_version < Version("1.14.2"):
            _migrate_full_metadata(cache, self.out)

        if old_version < Version("1.15.0"):
            migrate_registry_file(cache, self.out)

        if old_version < Version("1.19.0"):
            migrate_localdb_refresh_token(cache, self.out)

        if old_version < Version("1.26.0"):
            migrate_editables_use_conanfile_name(cache)

        if old_version < Version("1.31.0"):
            migrate_tgz_location(cache, self.out)
```
can be safely removed, as we will not attempt to upgrade from those to Conan 2.0.

uilianries:
Understood.

conans/client/migrations.py
```diff
@@ -69,34 +69,6 @@ def _make_migrations(self, old_version):
         cache = ClientCache(self.cache_folder, self.out)
         self._update_settings_yml(cache, old_version)
 
-        if old_version < Version("1.0"):
-            _migrate_lock_files(cache, self.out)
```
memsharded, line 73: You are leaving all the implementations of these calls as dead code, remove it too.

conans/client/migrations_settings.py
```diff
@@ -1,1944 +1,4 @@
...
-                exceptions: [None]
-
-build_type: [None, Debug, Release, RelWithDebInfo, MinSizeRel]
-
-
-cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]  # Deprecated, use compiler.cppstd
-"""
-
-settings_1_33_1 = """
+settings_2_0_0 = """
```
memsharded: This will also be removed in the next PR.

memsharded (changes requested):

conans/client/migrations_settings.py
uilianries: Yes. I just kept to be a good way when comparing the old migrations settings. If we put all on same PR it will be easier for review.

memsharded (changes requested):
Still not removed:
- ``def migrate_tgz_location(cache, out):``
- ``def migrate_localdb_refresh_token(cache, out):``
- etc

Those should be dead code, can be removed.

uilianries:
The class ClientMigrator is completely dead. Conan API consumes it but only runs migrate() which is implemented by Migrator class. Conan is 2kg lighter!

conans/client/conan_api.py
```diff
@@ -233,7 +233,7 @@ def __init__(self, cache_folder=None, output=None, user_io=None, http_requester=
         self.runner = runner
         self.app = None  # Api calls will create a new one every call
         # Migration system
-        migrator = ClientMigrator(self.cache_folder, Version(client_version), self.out)
+        migrator = Migrator(self.cache_folder, Version(client_version), self.out)
```
memsharded: But better leave the ClientMigrator. It is going to be needed, migrations will be needed in future Conan 2.X versions, and the Migrator infra is also used in the server, so it would be dirty to put migrations in the base ``Migrator`` class. The class can remain, even if empty.

uilianries: Yes, it makes sense. I gonna revert it.

uilianries:
Weird, virtualenv errors saying Conan is not installed.

```
pytest.fail("Required tool: '{}' is not available".format(tool_name))
E      Failed: Required tool: 'conan' is not available

[...]
=========================== short test summary info ============================
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_0::test_basic_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_0::test_find_program
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_0::test_list_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_0::test_list_with_spaces
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_1::test_basic_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_1::test_find_program
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_1::test_list_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_1::test_list_with_spaces
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_2::test_basic_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_2::test_find_program
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_2::test_list_variable
ERROR conans/test/functional/generators/virtualenv_test.py::VirtualEnvIntegrationTestCase_2::test_list_with_spaces
ERROR conans/test/functional/generators/cmake_multi_test.py::CMakeMultiTest::test_cmake_multi_find
= 713 passed, 142 skipped, 1 xfailed, 80 warnings, 13 errors in 156.92s (0:02:36) =
script returned exit code 1
```

memsharded:
Rebase from develop2 branch, it was broken.

uilianries:
Great!! Next step: Think how to solve migration for Conan 2.0.

PR #8492, opened 2021-02-12
- check cmake

SSE4:
see #8457 

Changelog: Omit
Docs: Omit

### DIFF
--- a/conans/test/functional/subsystems_build_test.py
+++ b/conans/test/functional/subsystems_build_test.py
@@ -220,3 +220,88 @@ class TestSubsystemsAutotoolsBuild:
         assert "__MINGW32__" not in client.out
         assert "__MINGW64__" not in client.out
         assert "__MSYS__" not in client.out
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Tests Windows Subsystems")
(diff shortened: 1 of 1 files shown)


memsharded (commented):
This is very good, great approach. Lets make sure that minimal CMake examples can be built in tests and in our infra, before suffering trying to build our Conan packages. Keep this going.

SSE4:
the error in CI is:
```
C:\msys64\mingw64\bin\gcc.exe      -Wl,--whole-archive CMakeFiles\cmTC_d4b34.dir/objects.a -Wl,--no-whole-archive  -o cmTC_d4b34.exe -Wl,--out-implib,libcmTC_d4b34.dll.a -Wl,--major-image-version,0,--minor-image-version,0 @CMakeFiles\cmTC_d4b34.dir\linklibs.rsp

E               collect2.exe: error: ld returned 5 exit status
```
locally, it doesn't reproduce, unfortunately.
status 5 sounds like an access violation, which might be a crash.
there are some conflicting tools already in the PATH on the CI, we need to keep digging with @czoido

SSE4:
passed! thanks to @czoido !

PR #8496, opened 2021-02-12
Fix extra spaces and colons in flags and environment variables

### LINKED ISSUE #8495: [bug] Path expansion in virtualrunenv generator adds extra colon for empty path
### Environment Details
  * Operating System+version: Debian GNU/Linux 10
  * Conan version: 1.34.0-dev
  * Python version: 3.7.0
  * Shell: GNU bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)

~~~bash
export LD_LIBRARY_PATH=
# install a recipe with "-g virtualrunenv"
source activate.sh
echo $LD_LIBRARY_PATH
~~~
The last echo prints path with extra ":" at the end.

### Explanation

According to [Bash expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):
> **${parameter:+word}**
> If parameter is null or unset, nothing is substituted, otherwise the expansion of word is substituted.

Conan uses:
~~~bash
PATH="..."${PATH+:$PATH}
~~~

Using `${var+:$var}`:
~~~bash
export DEFINED_EMPTY_VAR=
echo "xxx"${DEFINED_EMPTY_VAR+:$DEFINED_EMPTY_VAR}
echo "xxx"${UNDEFINED_VAR+:$UNDEFINED_VAR}
~~~
Results
~~~
xxx:
xxx
~~~

Correct form `${var:+:$var}`:
~~~bash
export DEFINED_EMPTY_VAR=
echo "xxx"${DEFINED_EMPTY_VAR:+:$DEFINED_EMPTY_VAR}
echo "xxx"${UNDEFINED_VAR:+:$UNDEFINED_VAR}
~~~
Results
~~~
xxx
xxx
~~~

klimkin:
According to [Bash expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html):

    ${parameter:+word}

Correct form for PATH:

    ${PATH:+:$PATH}

Correct form for CFLAGS:

    ${CFLAGS:+ $CFLAGS}

Fixes #8495

Changelog: Fix: Remove extra spaces in flags and colons in path variables.
Docs: omit

### DIFF
--- a/conans/test/functional/generators/virtualenv_test.py
+++ b/conans/test/functional/generators/virtualenv_test.py
@@ -265,6 +265,10 @@ class VirtualEnvIntegrationTestCase(unittest.TestCase):
 
     @pytest.mark.tool_conan
     def test_empty_defined_list_variable(self):
+        if os_info.is_windows:
+            # Windows doesn't make distinction between empty and undefined
+            # environment variables.
+            return
         self.assertNotIn("WHATEVER", os.environ)
         try:
             os.environ["WHATEVER"] = ""
--- a/conans/test/functional/generators/virtualenv_test.py
+++ b/conans/test/functional/generators/virtualenv_test.py
@@ -278,10 +278,7 @@ class VirtualEnvIntegrationTestCase(unittest.TestCase):
 
             _, environment = self._run_virtualenv(generator)
 
-            # FIXME: extra separator in Windows
-            extra_separator = os.pathsep if platform.system() == "Windows" else ""
-            self.assertEqual(environment["WHATEVER"],
-                             "{}{}{}{}".format("list", os.pathsep, "other", extra_separator))
+            self.assertEqual(environment["WHATEVER"], "{}{}{}".format("list", os.pathsep, "other"))
         finally:
             del os.environ["WHATEVER"]
 
--- a/conans/test/functional/generators/virtualenv_test.py
+++ b/conans/test/functional/generators/virtualenv_test.py
@@ -221,14 +221,14 @@ class VirtualEnvIntegrationTestCase(unittest.TestCase):
 
         _, environment = self._run_virtualenv(generator)
 
-        self.assertEqual(environment["CFLAGS"], "-O2 ")  # FIXME: Trailing blank
-        self.assertEqual(environment["CL"], "-MD -DNDEBUG -O2 -Ob2 ")  # FIXME: Trailing blank
+        extra_blank = " " if platform.system() == "Windows" else ""  # FIXME: Extra blank under Windows
+        self.assertEqual(environment["CFLAGS"], "-O2" + extra_blank)
+        self.assertEqual(environment["CL"], "-MD -DNDEBUG -O2 -Ob2" + extra_blank)
 
         with environment_append({"CFLAGS": "cflags", "CL": "cl"}):
             _, environment = self._run_virtualenv(generator)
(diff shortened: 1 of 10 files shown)


klimkin:
@jgsogo, @memsharded FYI

memsharded:
Thanks for contributing this @klimkin 

I am a bit surprised that our tests didn't catch such bug. I am experimenting with a new way to handle the environment in https://github.com/conan-io/conan/pull/8426, and I haven't been able to reproduce, it seems that modern bash can process the expansion correctly without the ``:``. 

I think this is most likely good to be merged, if it is the "canonical" way, but it would be great to have also first a failing test that proves that the current expansion without ":" doesn't work as intended. Could you please try to contribute such a test, or give some hints about the failures that could happen if the ":" is missing?

klimkin:
Added unit-test: https://github.com/conan-io/conan/pull/8496/commits/e64fee7bcbef7cdf007f6fd88d58767fca1e6c16

@memsharded Here is a nice write-up: https://tldp.org/LDP/abs/html/parameter-substitution.html
> **{parameter+alt_value}, ${parameter:+alt_value}**
> If parameter set, use alt_value, else use null string.
>
> Both forms nearly equivalent. The : makes a difference only when parameter has been declared and is null, see below.

~~~
param=

a=${param+xyz}
echo "a = $a"      # a = xyz

a=${param:+xyz}
echo "a = $a"      # a =
# Different result from   a=${param5+xyz}
~~~

conans/test/functional/generators/virtualenv_test.py
```diff
@@ -250,11 +248,43 @@ def test_list_variable(self):
...
         extra_separator = os.pathsep if platform.system() == "Windows" else ""
         self.assertEqual(environment["WHATEVER"],
                          "{}{}{}{}".format("list", os.pathsep, "other", extra_separator))
 
+    @pytest.mark.tool_conan
+    def test_empty_defined_list_variable(self):
+        if os_info.is_windows:
+            # Windows doesn't make distinction between empty and undefined
+            # environment variables.
+            return
```
jgsogo: You can use `@pytest.mark.skipif(platform.system() == "Windows", reason="...")`

conans/test/functional/generators/virtualenv_test.py
```diff
@@ -250,11 +248,43 @@ def test_list_variable(self):
...
+
+            generator = VirtualEnvGenerator(ConanFileMock())
+            generator.env = {"WHATEVER": ["list", "other"]}
+
+            _, environment = self._run_virtualenv(generator)
+
+            # FIXME: extra separator in Windows
+            extra_separator = os.pathsep if platform.system() == "Windows" else ""
+            self.assertEqual(environment["WHATEVER"],
+                             "{}{}{}{}".format("list", os.pathsep, "other", extra_separator))
```
jgsogo: If this test is not running for Windows, we can simplify

```suggestion
            self.assertEqual(environment["WHATEVER"], "{}{}{}".format("list", os.pathsep, "other"))
```

jgsogo (approved):
Love to see all those `FIXME` being actually fixed! 💯

conans/test/functional/generators/virtualenv_test.py
klimkin: That's what I was looking for! Thanks!

jgsogo (approved):

Issue #8476, opened 2021-02-10
[bug] Autotools in Conan with MSYS2


Facon:
### Environment Details (include every applicable attribute)
  * Operating System+version: Windows 10
  * Compiler+version: GCC 10
  * Conan version: 1.33.1
  * Python version: 3.8.7

### Steps to reproduce (Include if Applicable)
1. Install MSYS2 environment.
2. Open MSYS2 MINGW32 console and install Python, Autotools, CMake, GCC, etc.
3. Install conan with "pip install conan".
4. Create a profile that uses GCC from MSYS2 MINGW32.
5. Follow the example tutorial from Conan.io at [https://docs.conan.io/en/latest/getting_started.html]().

### Logs
**Output from conan:**
```
User@DESKTOP-LC9EG87 MINGW32 /c/Users/User/Documents/Sources/testing/conan_examples
$ conan install . -pr=mingw32debug -b=missing
Configuration:
[settings]
arch=x86
arch_build=x86
build_type=Debug
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=10
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
CC=C:\msys64\mingw32\bin\gcc.exe
CXX=C:\msys64\mingw32\bin\g++.exe
WARN: apr-util/1.6.1: requirement expat/2.2.9 overridden by poco/1.10.1 to expat/2.2.10
WARN: libmysqlclient/8.0.17: requirement openssl/1.1.1i overridden by poco/1.10.1 to openssl/1.1.1h
apr/1.7.0: WARN: Package binary is corrupted, removing: 945174b4e713b95463410bd3cc7cfb90a533f4d8
conanfile.txt: Installing package
Requirements
    apr/1.7.0 from 'conan-center' - Cache
    apr-util/1.6.1 from 'conan-center' - Cache
    bzip2/1.0.8 from 'conan-center' - Cache
    expat/2.2.10 from 'conan-center' - Cache
    libmysqlclient/8.0.17 from 'conan-center' - Cache
    libpq/11.5 from 'conan-center' - Cache
    openssl/1.1.1h from 'conan-center' - Cache
    pcre/8.41 from 'conan-center' - Cache
    poco/1.10.1 from 'conan-center' - Cache
    sqlite3/3.33.0 from 'conan-center' - Cache
    zlib/1.2.11 from 'conan-center' - Cache
Packages
[...]

PR #8506, opened 2021-02-15
BugFix: allow `run_in_windows_bash` in MSYS/Cygwin

SSE4:
related to #8476

running `AutoToolsBuildEnvironment(self, win_bash=True)` in MSYS2 results in an error:
```
ERROR: apr/1.7.0: Error in build() method, line 92
        autotools = self._configure_autotools()
while calling '_configure_autotools', line 76
        self._autotools.configure(args=conf_args, configure_dir=self._source_subfolder)
        ConanException: Command only for Windows operating system
```
the fix allows `run_in_windows_bash` to work in sub-systems as well.

Changelog: BugFix: Allow `run_in_windows_bash` in MSYS/Cygwin.
Docs: omit

### DIFF
--- a/conans/client/tools/win.py
+++ b/conans/client/tools/win.py
@@ -635,7 +635,7 @@ def run_in_windows_bash(conanfile, bashcmd, cwd=None, subsystem=None, msys_mingw
         It requires to have MSYS2, CYGWIN, or WSL
     """
     env = env or {}
-    if platform.system() != "Windows":
+    if not OSInfo().is_windows:
         raise ConanException("Command only for Windows operating system")
     subsystem = subsystem or OSInfo.detect_windows_subsystem()
 
@@ -693,7 +693,11 @@ def run_in_windows_bash(conanfile, bashcmd, cwd=None, subsystem=None, msys_mingw
         bash_path = OSInfo.bash_path()
         bash_path = '"%s"' % bash_path if " " in bash_path else bash_path
         login = "--login" if with_login else ""
-        wincmd = '%s %s -c %s' % (bash_path, login, escape_windows_cmd(to_run))
+        if platform.system() == "Windows":
+            # cmd.exe shell
+            wincmd = '%s %s -c %s' % (bash_path, login, escape_windows_cmd(to_run))
+        else:
+            wincmd = '%s %s -c %s' % (bash_path, login, to_run)
         conanfile.output.info('run_in_windows_bash: %s' % wincmd)
 
         # If is there any other env var that we know it contains paths, convert it to unix_path


conans/client/tools/win.py
```diff
@@ -635,7 +635,7 @@ def run_in_windows_bash(conanfile, bashcmd, cwd=None, subsystem=None, msys_mingw
         It requires to have MSYS2, CYGWIN, or WSL
     """
     env = env or {}
-    if platform.system() != "Windows":
```
memsharded, line 638: Cygwin also returns "Windows" here, could be only Msys will be different

memsharded (approved):

conans/client/tools/win.py
SSE4: didn't check Cygwin (may check tomorrow), msys returns a different value for sure

memsharded: The fix seems fine anyway, it was just an observation. Would you like to check something more, or can it be merged?

memsharded:
I think this can be merged, but I would like to note that ``run_in_windows_bash()`` is very complicated and obscure thing inside Conan codebase. I really want to improve this while we define the new helpers, mostly for Autotools, simplifying when possible, make it a bit more explicit and configurable, and being covered by real tests better.

conans/client/tools/win.py
SSE4: for the completeness.
Cygwin:
```
sse4@DESKTOP-LAD63V0 ~
$ python -c "import platform; print(platform.system())"
CYGWIN_NT-10.0-18363
```
MSYS2:
```
sse4@DESKTOP-LAD63V0 MSYS ~
$ python -c "import platform; print(platform.system())"
MSYS_NT-10.0-18363
```
MinGW32 (MSYS2):
```
sse4@DESKTOP-LAD63V0 MINGW32 ~
$ python -c "import platform; print(platform.system())"
MINGW32_NT-10.0-18363
```
MinGW64 (MSYS2):
```
sse4@DESKTOP-LAD63V0 MINGW64 ~
$ python -c "import platform; print(platform.system())"
MINGW64_NT-10.0-18363
```

SSE4:
checking cygwin, it seems to be not enough:
```
apr/1.7.0: run_in_windows_bash: /usr/bin/bash --login -c cd "/cygdrive/home/sse4/.conan/data/apr/1.7.0/_/_/build/520c5865652102ac1331f89be33c82aad2520eb9" && /cygdrivesource_subfolder/configure '--with-installbuilddir=${prefix}/bin/build-1' '--enable-shared=no' '--enable-static=yes' '--prefix=/home/sse4/.conan/data/apr/1.7.0/_/_/package/520c5865652102ac1331f89be33c82aad2520eb9'
/bin/sh: /cygdrivesource_subfolder/configure: No such file or directory
apr/1.7.0:
apr/1.7.0: ERROR: Package '520c5865652102ac1331f89be33c82aad2520eb9' build failed
apr/1.7.0: WARN: Build folder /home/sse4/.conan/data/apr/1.7.0/_/_/build/520c5865652102ac1331f89be33c82aad2520eb9
ERROR: apr/1.7.0: Error in build() method, line 94
        autotools = self._configure_autotools()
while calling '_configure_autotools', line 78
        self._autotools.configure(args=conf_args, configure_dir=self._source_subfolder)
        ConanException: Error 127 while executing /cygdrivesource_subfolder/configure '--with-installbuilddir=${prefix}/bin/build-1' '--enable-shared=no' '--enable-static=yes' '--prefix=/home/sse4/.conan/data/apr/1.7.0/_/_/package/520c5865652102ac1331f89be33c82aad2520eb9'
```

I guess the error is in `unix_path` function which needs a simple fix:
```
diff --git a/conans/client/tools/win.py b/conans/client/tools/win.py
index 43db28b5a..bd69a3e47 100644
--- a/conans/client/tools/win.py
+++ b/conans/client/tools/win.py
@@ -620,7 +620,10 @@ def unix_path(path, path_flavor=None):
     if path_flavor in (MSYS, MSYS2):
         return path.lower()
     elif path_flavor == CYGWIN:
-        return '/cygdrive' + path.lower()
+        if os.path.isabs(path):
+            return '/cygdrive' + path.lower()
+        else:
+            return path.lower()
     elif path_flavor == WSL:
         return '/mnt' + path[0:2].lower() + path[2:]
     elif path_flavor == SFU:
```

which allows the build to start.
actually, it needs a more complete fix (e.g. for other sub-systems, plus it should translate only paths starting with drives like C:, D:, because paths like `/home` don't need to be adjusted).

conans/client/tools/win.py
memsharded: Curiously, it seems it can change, I am not sure why:

Cygwin:
``` ~
$ python -c "import platform; print(platform.system())"
Windows
```

In any case, doesn't seem relevant, if it can change in other computers, need to take it into account.

memsharded:
@SSE4 How would you like to proceed? Fixing those things in this PR?

conans/client/tools/win.py
SSE4: this is probably not a python from a Cygwin itself. run `which python` to check.
if it's `/usr/bin/python`, then it's cygwin's one.
if it's something like `/cygdrive/c/users...`, then it's installed in the main Windows system.

SSE4:
> @SSE4 How would you like to proceed? Fixing those things in this PR?

nope, this deserves its own PR, and I want to write some tests for `unix_path`.

Issue #5513, opened 2019-07-18
[feature request] add support for simultaneous multiple conan caches


DoDoENT:
This idea was first discussed in issue #5498 in [this comment](https://github.com/conan-io/conan/issues/5498#issuecomment-512525224).

The idea is to enable conan to support multiple conan data folders simultaneously. The packages that are dependencies of the package being installed would be installed into the default cache folder and the current package could be exported into the ephemeral local cache required just for testing.

The motivation for this comes from the following example: let's say that my repository contains two packages `A` and `B`. `B` depends on `A` and they always need to be developed together (think of an example with `OcrEngine` and `OcrEngineModelBuilder` from the description of issue #4753). So, in order to ensure that package `B` can be tested with `conan test`, it also needs the same version of package `A`, which is still not exported. If multiple conan cache folders were supported, the Jenkins executor could export both `A` and `B` into its temporary folder to perform the `conan test`, while still using the default conan cache for other dependencies of `A` and `B`.

If the same package is found in multiple cache folders, an error could be raised which could be corrected by specifying from which cache the package should be consumed.

jgsogo:
This feature request is very interesting and it is the path we are following when we refactor the codebase trying to make more use of the `PackageLayouts` https://github.com/conan-io/conan/pull/5105 (currently we have the `PackageCacheLayout` and `PackageEditableLayout`). Once we manage to detach the packages from the cache it would be easy to have packages stored or coming from several caches 😉 

This will require a huge effort, at some Conan stages it is not easy to decouple a package from the cache, but we are walking in that direction step by step.

KerstinKeller:
While setting up our CI Workflow we stumbled upon the same issue.
It would be very useful to have one "global" cache where a package build could consume all its dependencies from, and a local cache where all the building takes place.

A huge benefit would be that each build job can clean up after itself, if the build cache is located in the current CI build folder, while the "official" cache for consuming packages is shared between all build jobs.

It might not sound like a big deal, but if you use CI and Conan to create packages as part of quality checks on every branch / commit for large projects, you do produce a lot of artifacts, which are not even supposed to be released / promoted.

jgsogo:
[...]

PR #8510, opened 2021-02-16
[WIP] Locks + cache for multiple revisions

jgsogo:

Requirements

- [ ] multiple revisions
- [ ] locking mechanism (threads and processes)
- [ ] LRU implementation
- [ ] use deterministic folders
- [ ] shortest paths as possible
- [ ] sqlite3 rollback mechanism and _blocks_
- [ ] ...

Keep in mind:

https://github.com/conan-io/conan/issues/5513


conan/cache/exceptions.py
```diff
@@ -0,0 +1,14 @@
+from conans.errors import ConanException
+from conans.model.ref import ConanFileReference, PackageReference
+
+
+class DuplicateReferenceException(ConanException):
```
memsharded, line 11: Please don't create additional exceptions until there is a except-capture use case for them.

conan/cache/cache_database.py
```diff
@@ -0,0 +1,180 @@
...
+        where_clause, where_values = self._where_clause(item, filter_packages=True)
+        query = f"UPDATE {self._table_name} " \
+                f"SET    {self._column_path} = '{new_path}' " \
+                f"WHERE {where_clause}"
+        with self.connect() as conn:
+            r = conn.execute(query, where_values)
+            assert r.rowcount > 0
+
+
+class CacheDatabaseSqlite3Memory(CacheDatabase, Sqlite3MemoryMixin):
```
memsharded, line 145: Prefer composition over inheritance. Try to avoid multiple inheritance too.
Specially, the "Memory" domain should exist in test scope only, not in production code. 

Please, use composition for these, and have the Sqlite3Memory functionality defined in testing only, injected as a dependency, or patched.

conan/locks/backend_sqlite3.py
```diff
@@ -0,0 +1,69 @@
...
+        CREATE TABLE {guard} {self._table_name} (
+            {self._column_resource} text NOT NULL,
+            {self._column_pid} integer NOT NULL,
+            {self._column_writer} BOOLEAN NOT NULL CHECK ({self._column_writer} IN (0,1))
+        );
+        """
+        with self.connect() as conn:
+            conn.execute(query)
+
+    def try_acquire(self, resource: str, blocking: bool) -> LockId:
```
memsharded, line 38: This ``try_acquire`` is not atomic. More than one connection can enter at a time, can ``select`` and can do checks, that will pass (no one has written yet), so writes will be incorrect. Please use an atomic ``update_if_not_exist`` idiom.

conan/locks/backend_sqlite3.py
```diff
@@ -0,0 +1,69 @@
...
+            result = conn.execute(f'INSERT INTO {self._table_name} '
+                                  f'VALUES (?, ?, ?)', (resource, os.getpid(), blocking_value,))
+            return result.lastrowid
+
+    def release(self, backend_id: LockId):
+        with self.connect() as conn:
+            conn.execute(f'DELETE FROM {self._table_name} WHERE rowid=?', (backend_id,))
+
+
+class LockBackendSqlite3Memory(Sqlite3MemoryMixin, LockBackendSqlite3):
```
memsharded, line 73: Same as above (no "memory" classes in production code, no multiple inheritance). Simplify, avoid premature generalization, we can work now assuming sqlite only.

memsharded (changes requested):
I am concerned about a system that by design leaves the things locked when the process dies or is killed, because in practice this happens frequently. We should aim to implement locks that unlock automatically after Conan process dying or being killed.

conan/locks/backend_sqlite3.py
jgsogo: According to docs, it's atomic, the `with self.connect() as conn:` should take care of it.

jgsogo:
> I am concerned about a system that by design leaves the things locked when the process dies or is killed, because in practice this happens frequently. We should aim to implement locks that unlock automatically after Conan process dying or being killed.

Me too. Any suggestion? My proposal was to register an `at_exit` function to clean the database from locks matching the `pid` and provide a command-line command to `purge` it.... but suggestions are welcome.

conan/locks/backend_sqlite3.py
memsharded: I am not sure about this. The following code works, adding 2 lines to the table, both with a 1=blocking state over the same resource:

```python
 with connect() as conn1:
        with connect() as conn2:
            # Check if any is using the resource
            print("SELECT from conn1")
            result = conn1.execute(f'SELECT Pid, Writer '
                                    f'FROM MyTable '
                                    f'WHERE Resource = ?;', (resource,))

            print("SELECT from conn2")
            result2 = conn2.execute(f'SELECT Pid, Writer '
                                    f'FROM MyTable '
                                    f'WHERE Resource = ?;', (resource,))

            if result.fetchone():
                raise Exception("Resource {} locked".format(resource))

            if result2.fetchone():
                raise Exception("Resource {} locked".format(resource))

            # Check if a writer (exclusive) is blocking
            blocked = any([it[1] for it in result.fetchall()])
            if blocked:
                raise Exception("Resource {} locked".format(resource))

             # Check if a writer (exclusive) is blocking
            blocked = any([it[1] for it in result2.fetchall()])
            if blocked:
                raise Exception("Resource {} locked".format(resource))

            # Add me as a blocker, reader or writer
            blocking_value = 1 if blocking else 0

            result2 = conn2.execute(f'INSERT INTO MyTable '
                                    f'VALUES (?, ?, ?)', (resource, id(conn2), blocking_value,))

        result = conn1.execute(f'INSERT INTO MyTable '
                               f'VALUES (?, ?, ?)', (resource, id(conn1), blocking_value,))
        print("RESULT", result.lastrowid)
        print("RESULT2", result2.lastrowid)
```

jgsogo: Yes, but this is not what is written in the PR... I'll force that connection concurrency, and verify what is written in the docs.

jgsogo: Ok, it doesn't block. The transaction is rolled back in case of an exception. It is in the TODO list.

jgsogo: I keep having a loot to it, but looks like, something like this will block the database during all the queries (it blocks and raises for the second one):

```python
    @contextmanager
    def connect(self):
        conn = sqlite3.connect(....)
        conn.isolation_level = None
        try:
            conn.execute('begin')
            yield conn.cursor()
            conn.execute("commit")
        except Exception as e:
            conn.execute("rollback")
            raise e
        finally:
            conn.close()
```

memsharded:
> Me too. Any suggestion? My proposal was to register an at_exit function to clean the database from locks matching the pid and provide a command-line command to purge it.... but suggestions are welcome.

We need a system level mechanism, atexit is still a python mechanism, if sqlite is not providing any system level utility for this we need to resort back to file locking, and fasteners seems our best fit.

jgsogo:
Fasteners relies on the filesystem as well, I don't see how it is different.

If the Python process is terminated, `fasteners` will leave files behind (https://github.com/harlowja/fasteners/issues/26) and `sqlite3` will leave content inside a file. In both cases we need _something else_ to take care of them. And, if we need to remove manually what is left behind, I prefer to query a database and remove entries than to inspect files and modify their content.

The only mechanism I can think about is something that ensures that all the locks are removed properly if the process crashes... something like a sidecar process that will clean entries in the database or remove files created by `fasteners` if the process being monitored disappears.

memsharded:
> Fasteners relies on the filesystem as well, I don't see how it is different.

It is implemented at the system level apis. If the process dies or is killed, the resource is released automatically. The OS manages it. It is way different than an application level mechanism. The file is only the reflect in disk of the lock, it doesn't matter if it is there after it is released, what is important is the OS lock on that file in the filesystem.

jgsogo:
Then, that is what we should use without any doubt. It makes absolutely no sense to try to implement it using something different. 🎉

conan/locks/backend_fasteners.py
```diff
@@ -0,0 +1,109 @@
...
+
+from conan.locks.backend import LockBackend
+from conan.locks.exceptions import AlreadyLockedException
+
+log = logging.getLogger(__name__)
+
+
+class RWLock(object):
+    def __init__(self, resource: str, interprocess_lock: str):
+        self.w_lock = threading.Lock()
```
memsharded, line 17: I am checking possible multi-threading race conditions, and given the structure of the program, typical operation (parallel uploads, parallel downloads), etc., we can guarantee by the business logic that multithreaded operations within the same process will not collide or produce race conditions with itself. Thus, lets simplify this, leaving only the interprocess lock, the simplicity will pay off

memsharded:
The work here is being continued in https://github.com/conan-io/conan/pull/8796

PR #8514, opened 2021-02-17
Make tests fail if tools are not available

czoido:
Changelog: omit
Docs: omit

Default testing behaviour with tools marks is that if the tools are not found the test is skipped, but this can make that for any reason like for example a ci node was not configured properly we skip tests that we would like to run. This PR alters that behaviour to fail if tools are not found.


memsharded (approved):
LGTM

jgsogo (approved):

PR #8517, opened 2021-02-17
Avoid concurrency issues with load_metadata function

jgsogo:
Changelog: omit
Docs: omit

Use the locking mechanism to read the metadata file, not only with the contextmanager when you are going to modify it.

---

I'm trying to figure out which could be the worst scenario that leads to the `assert` failure:

```
assert node.prev, "PREV for %s is None: %s" % (str(pref), metadata.dumps())
```

### DIFF
--- a/conans/paths/package_layouts/package_cache_layout.py
+++ b/conans/paths/package_layouts/package_cache_layout.py
@@ -238,6 +238,10 @@ class PackageCacheLayout(object):
 
     # Metadata
     def load_metadata(self):
+        with self.update_metadata() as metadata:
+            return metadata
+
+    def _load_metadata(self):
         try:
             text = load(self.package_metadata())
         except IOError:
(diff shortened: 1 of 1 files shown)


jgsogo:
Bitten by https://github.com/harlowja/fasteners/issues/26 and the layout we are using to store the lockfiles and determine if a package exists. Not sure if it will require too many changes 😞

jgsogo:
Alternatives:

* tell `fasteners` to store the files somewhere else
* force removal of the `metadata.json.lock` file if it is the only one in the folder here (we will face a concurrency challenge): https://github.com/conan-io/conan/blob/8e8ebcea93a44cb1617cfb5f5ef60754258eeeaa/conans/util/locks.py#L109

memsharded:
Obsolete, closing

PR #8519, opened 2021-02-17
Fail if we are about to save a package without revision to 'metadata.json'

jgsogo:
Changelog: omit
Docs: omit

Fail if we are saving a package without a revision. I think this can be the easiest way to figure out what is happening here https://github.com/conan-io/conan-center-index/issues/4599. 

There are places in the codebase where we are using `metadata.packages['pkgID'].member` which will create an empty `_BinaryPackageMetadata` and return `None`. No exception, no error, just an empty item is added to the dictionary. We should use `.get('pkgID')` instead... but ATM, this `assert` will be enough to check if/where we are saving a package without `revision` to `metadata.json` file.

Blocked by https://github.com/conan-io/conan/pull/8520

### DIFF
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -1068,6 +1068,7 @@ class ConanAPIV1(object):
             m = metadata.packages.get(pref.id)
             if m and m.remote:
                 raise ConanException("%s already exists. Use update" % str(pref))
+            # Create the package entry although we know nothing else about it
             metadata.packages[pref.id].remote = remote.name
 
     @api_method
--- a/conans/model/package_metadata.py
+++ b/conans/model/package_metadata.py
@@ -56,9 +56,12 @@ class _BinaryPackageMetadata(object):
         self._recipe_revision = DEFAULT_REVISION_V1 if r is None else r
 
     def to_dict(self):
-        # It is totally unexpected to save 'metadata.json' file without package-revsion, when
-        # even with revisions deactivated, it takes the value DEFAULT_REVISION_V1.
-        assert self._revision is not None, 'Trying to save to metadata packages without revision!'
+        # It is totally unexpected to save 'metadata.json' file without package-revision,
+        # even if revisions are deactivated, it takes the value DEFAULT_REVISION_V1.
+        # EXCEPT, for a requested feature to be able to asign the remote for a package-id before
+        # installing it to the cache.
+        if self.remote is None:
+            assert self._revision is not None, 'Trying to save to metadata without package revision!'
 
         ret = {"revision": self.revision,
                "recipe_revision": self.recipe_revision,
--- a/conans/model/package_metadata.py
+++ b/conans/model/package_metadata.py
@@ -56,6 +56,11 @@ class _BinaryPackageMetadata(object):
         self._recipe_revision = DEFAULT_REVISION_V1 if r is None else r
 
     def to_dict(self):
+        if not self.revision:
+            # It is totally unexpected to save 'metadata.json' file without package-revsion, when
+            # even with revisions deactivated, it takes the value DEFAULT_REVISION_V1.
+            assert False, 'Trying to save to metadata packages without revision!'
+
         ret = {"revision": self.revision,
                "recipe_revision": self.recipe_revision,
                "remote": self.remote,
--- a/conans/model/package_metadata.py
+++ b/conans/model/package_metadata.py
@@ -56,10 +56,9 @@ class _BinaryPackageMetadata(object):
         self._recipe_revision = DEFAULT_REVISION_V1 if r is None else r
 
     def to_dict(self):
-        if not self.revision:
-            # It is totally unexpected to save 'metadata.json' file without package-revsion, when
-            # even with revisions deactivated, it takes the value DEFAULT_REVISION_V1.
-            assert False, 'Trying to save to metadata packages without revision!'
+        # It is totally unexpected to save 'metadata.json' file without package-revsion, when
+        # even with revisions deactivated, it takes the value DEFAULT_REVISION_V1.
+        assert self._revision is not None, 'Trying to save to metadata packages without revision!'
 
         ret = {"revision": self.revision,
                "recipe_revision": self.recipe_revision,
--- a/conans/test/functional/command/remote_test.py
+++ b/conans/test/functional/command/remote_test.py
@@ -415,11 +415,10 @@ class RemoteTest(unittest.TestCase):
         self.assertIn("Hello1/0.1@user/testing: remote2", self.client.out)
 
     def test_package_refs(self):
-
         self.client.run("remote add_pref Hello/0.1@user/testing:555 remote0")
         self.client.run("remote list_pref Hello/0.1@user/testing")
         self.assertIn("Hello/0.1@user/testing:555: remote0", self.client.out)
-
+        
         self.client.run("remote add_pref Hello1/0.1@user/testing:555 remote1")
         self.client.run("remote list_pref Hello1/0.1@user/testing")
         self.assertIn("Hello1/0.1@user/testing:555: remote1", self.client.out)


jgsogo:
If everything goes smoothly in C3i (https://github.com/conan-io/c3i_jenkins/pull/641), we probably need these changes in 1.33.x as well (we can even wait if we only want to release one version).

memsharded (commented):
I would probably introduce at least a log in 

```python
@revision.setter
def revision(self, r):
    self._revision = DEFAULT_REVISION_V1 if r is None else r
```

Maybe someone is assigning a ``None`` revision to the package? Is this possible?

jgsogo:
We (@SSE4 ) are thinking that we can be executing these lines:

```python
        if not node.binary:
            node.binary = BINARY_CACHE
            metadata = metadata or package_layout.load_metadata()
            node.prev = metadata.packages[pref.id].revision
            assert node.prev, "PREV for %s is None: %s" % (str(pref), metadata.dumps())
```

in a cache with a `metadata.json` file without packages, but with the pkg-id folder there (that was not removed because of something).

SSE4:
what we already know so far?
- seems like it happens only on Windows slaves
- assertion happens for the dependency packages, not for the package being built (e.g. for `jinja2cpp` it asserts for the `expected-lite`, for `openssl` it asserts for `nasm` or `zlib`

my theory so far is not a concurrency and not a bad write to the `metadata.json`.
I suppose `metadata.json` has empty `packages` array so far.
our code uses [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict), which never raises `KeyError`, but instead constructs an empty (invalid) `_BinaryPackageMetadata` instances, right above the assert.

the one possible scenario for that is an incorrect cache cleanup (e.g. having a messed up cache). this could happen, for instance, if CI job didn't clean up cache fully, which may happen if:
- `rmdir` failed for some reason (e.g. permission error because of locked file or running process) - as I understand we never actually check that errors
- `rmdir` wasn't called at all (e.g. CI process was cancelled or crashed long before execution point reached the cleanup, or in the middle of it)

I was able to find out how to reproduce it very easy and consistently:

```
# ensure bzip2 doesn't exist
conan remove bzip2/* -f

# create an empty package folder to simulate incomplete CI cleanup - your package id will be different
mkdir -p $HOME/.conan/data/bzip2/1.0.8/_/_/package/75fe66fbfe61230e9fcbb361f13b6b3baa483e0a

# run an install
conan install jinja2cpp/1.1.0@ --profile=default --build=jinja2cpp

# BOOM!
  File "/Users/sse4/bincrafters/conan-io/conans/client/command.py", line 2115, in run
    method(args[0][1:])
  File "/Users/sse4/bincrafters/conan-io/conans/client/command.py", line 516, in install
    info = self._conan.install_reference(ref,
  File "/Users/sse4/bincrafters/conan-io/conans/client/conan_api.py", line 94, in wrapper
    return f(api, *args, **kwargs)
  File "/Users/sse4/bincrafters/conan-io/conans/client/conan_api.py", line 554, in install_reference
    deps_install(self.app, ref_or_path=reference, install_folder=install_folder,
  File "/Users/sse4/bincrafters/conan-io/conans/client/manager.py", line 55, in deps_install
    deps_graph = graph_manager.load_graph(ref_or_path, create_reference, graph_info, build_modes,
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_manager.py", line 114, in load_graph
    deps_graph = self._resolve_graph(root_node, graph_info, build_mode, check_updates, update,
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_manager.py", line 240, in _resolve_graph
    deps_graph = self._load_graph(root_node, check_updates, update,
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_manager.py", line 362, in _load_graph
    self._recurse_build_requires(graph, builder, check_updates, update, build_mode,
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_manager.py", line 286, in _recurse_build_requires
    self._binary_analyzer.evaluate_graph(graph, build_mode, update, remotes, nodes_subset, root)
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_binaries.py", line 389, in evaluate_graph
    self._evaluate_node(node, build_mode, update, remotes)
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_binaries.py", line 176, in _evaluate_node
    self._process_node(node, pref, build_mode, update, remotes)
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_binaries.py", line 242, in _process_node
    self._evaluate_cache_pkg(node, package_layout, pref, metadata, remote, remotes, update)
  File "/Users/sse4/bincrafters/conan-io/conans/client/graph/graph_binaries.py", line 95, in _evaluate_cache_pkg
    assert node.prev, "PREV for %s is None: %s" % (str(pref), metadata.dumps())
AssertionError: PREV for bzip2/1.0.8:75fe66fbfe61230e9fcbb361f13b6b3baa483e0a is None: {"recipe": {"revision": "0", "remote": "conan-center", "properties": {}, "checksums": {"conanmanifest.txt": {"md5": "608446fc59b46ac42597f4c6ed845190", "sha1": "2f0f0b6d28e4e7f0660f26269d995b016c0e7155"}, "conanfile.py": {"md5": "e9d524f692b231e1aa141a1995dac7aa", "sha1": "4deb01e7b5764b8e6694b64ee3a8cc039726a116"}, "conan_export.tgz": {"md5": "44e99ec27f446c8c2903d7ead4a7e8e4", "sha1": "5a2eb16ade9bced471bfe964bcc6a726c864fad6"}}}, "packages": {"75fe66fbfe61230e9fcbb361f13b6b3baa483e0a": {"revision": null, "recipe_revision": null, "remote": null, "properties": {}, "checksums": {}}}}

```

SSE4:
the evil insertion happens on this line:
```
remote_name = metadata.packages[pref.id].remote or metadata.recipe.remote
```
this could be fixed by a simple patch (I am not sure if it's correct, please double check):
```
diff --git a/conans/client/graph/graph_binaries.py b/conans/client/graph/graph_binaries.py
index e2bcb8602..d13f25baf 100644
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -91,6 +91,9 @@ class GraphBinariesAnalyzer(object):
         if not node.binary:
             node.binary = BINARY_CACHE
             metadata = metadata or package_layout.load_metadata()
+            if pref.id not in metadata.packages:
+                self._evaluate_remote_pkg(node, pref, remote, remotes)
+                return
             node.prev = metadata.packages[pref.id].revision
             assert node.prev, "PREV for %s is None: %s" % (str(pref), metadata.dumps())

@@ -235,7 +238,10 @@ class GraphBinariesAnalyzer(object):
             # If the remote_name is not given, follow the binary remote, or the recipe remote
             # If it is defined it won't iterate (might change in conan2.0)
             metadata = metadata or package_layout.load_metadata()
-            remote_name = metadata.packages[pref.id].remote or metadata.recipe.remote
+            if pref.id in metadata.packages:
+                remote_name = metadata.packages[pref.id].remote or metadata.recipe.remote
+            else:
+                remote_name = metadata.recipe.remote
             remote = remotes.get(remote_name)

         if package_layout.package_id_exists(pref.id):  # Binary already in cache, check for updates
```
this at least fixes an install command for me which runs gracefully without an assertion.

in general, there are many places where an evil insertion could happen - almost every index access to the `metadata.packages` should be guarded or replaced with `get`, otherwise it's dangerous.

so there are several things to be addressed IMO:
- evil insertions of an empty `_BinaryPackageMetadata` via `defaultdict`
- conan relies too much upon cache integrity, which in reality could be easily in an inconsistent state
- conan-center-index CI which is not doing/checking clean-ups properly

SSE4 (approved):

jgsogo:
@SSE4 , can you create a PR to `release/1.32` (we are at the point of upgrading CCI to `1.33`, so maybe we need to switch branches) with your patch so we can see how it runs with all the test-suite?

I'm going to open a PR to `c3i_jenkins` dumping all the files in the workspace before using it, we should be able to check if there are any files there before running any Conan command. Zero risk.

jgsogo:
Another reason why we might get a _corrupted cache_ when reusing a workspace: the job was downloading the files for a package when it was cancelled. Is the file `metadata.json` written at the end of this process? If so, we would get the directory for the package-id, but no `metadata.json` file.

jgsogo:
We've seen this error again in ConanCenter, thanks @SSE4 from noticing so quickly.

This is `metadata.json`:
```
{
   "recipe":{
      "revision":"c9b5dccbc620ae581bb1c60547ced620",
      "remote":"conan-center",
      "properties":{
         
      },
      "checksums":{
         "conan_export.tgz":{
            "md5":"01ec85a77c93003990b0a7123f5886f1",
            "sha1":"6f71b0b82e093b7fab90e70fc19ef3eafb7ba1aa"
         },
         "conanmanifest.txt":{
            "md5":"212e2e8cebbe3886f094e2834d40cf14",
            "sha1":"0bb80bef838223fac4864c42dc9b728270da1c61"
         },
         "conanfile.py":{
            "md5":"add4439326e7c09165b9242748bb4175",
            "sha1":"c246ba1c15aa93325ec57e29bdf46492611f4f18"
         }
      }
   },
   "packages":{
      
   }
}
```

and the directory for a package exists:

image

but there are no files inside the package, only the directories

memsharded:
Obsolete, closing.

PR #8521, opened 2021-02-17
Fix mingw marks

czoido:
Changelog: Fix: Change mingw mark to mingw32 or mingw64
Docs: omit

Using the new mingw mechanism to add the tool to path, some tests were skipped because the tool was not found. Removing mingw mark, now the test must be marked as mingw64 or mingw32.

#TAGS: slow

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -19,7 +19,7 @@ tool_environments = {
 
 tools_available = [
     'cmake',
-    'gcc', 'clang', 'visual_studio', 'mingw', 'xcode',
+    'gcc', 'clang', 'visual_studio', 'xcode',
     'msys2', 'cygwin', 'mingw32', 'mingw64',
     'autotools', 'pkg_config', 'premake', 'meson',
     'file',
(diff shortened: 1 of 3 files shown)


jgsogo (approved):
How do we know if we need to add `mingw32` or `mingw64`?

Suggestion: we should register the markers so the CI will fail (in the PRs) if we are adding a mark that contains a typo or is not considered in the pipelines.

czoido:
> How do we know if we need to add `mingw32` or `mingw64`?
> 

Maybe @SSE4 can answer that question

> Suggestion: we should register the markers so the CI will fail (in the PRs) if we are adding a mark that contains a typo or is not considered in the pipelines.

Good idea, I'll work on that

SSE4:
> > How do we know if we need to add `mingw32` or `mingw64`?

MinGW32 is 32-bit GCC (produces `x86` binaries), MinGW64 is 64-bit GCC (produces `x64` binaries), that's it.
depends on what you need to test. I'd say `MinGW64` is sane default, as it's not a cross-building scenario (if you just want to test generic MinGW and don't care much about bitness).

SSE4 (approved):

PR #8532, opened 2021-02-19
BugFix: prevent evil insertions into metadata.json resulted in corrupted package and inability to install

SSE4:
#TAGS: slow

see #8519

Changelog: BugFix: Prevent evil insertions into metadata.json resulted in corrupted package and inability to install.
Docs: omit

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -231,14 +231,19 @@ class GraphBinariesAnalyzer(object):
         metadata = self._evaluate_clean_pkg_folder_dirty(node, package_layout, pref)
 
         remote = remotes.selected
+
+        metadata = metadata or package_layout.load_metadata()
         if not remote:
             # If the remote_name is not given, follow the binary remote, or the recipe remote
             # If it is defined it won't iterate (might change in conan2.0)
-            metadata = metadata or package_layout.load_metadata()
-            remote_name = metadata.packages[pref.id].remote or metadata.recipe.remote
+            if pref.id in metadata.packages:
+                remote_name = metadata.packages[pref.id].remote or metadata.recipe.remote
+            else:
+                remote_name = metadata.recipe.remote
             remote = remotes.get(remote_name)
 
-        if package_layout.package_id_exists(pref.id):  # Binary already in cache, check for updates
+        if package_layout.package_id_exists(pref.id) and pref.id in metadata.packages:
+            # Binary already in cache, check for updates
             self._evaluate_cache_pkg(node, package_layout, pref, metadata, remote, remotes, update)
             recipe_hash = None
         else:  # Binary does NOT exist locally
--- a/conans/test/integration/command/install/install_update_test.py
+++ b/conans/test/integration/command/install/install_update_test.py
@@ -289,3 +289,23 @@ def test_fail_usefully_when_failing_retrieving_package():
     # Try to install ref2, it will try to download the binary for ref1
     client.run("install {}".format(ref2), assert_error=True)
     assert "ERROR: Error downloading binary package: '{}'".format(pref1) in client.out
+
+def test_evil_insertions():
+    ref = ConanFileReference.loads("lib1/1.0@conan/stable")
+    ref2 = ConanFileReference.loads("lib2/1.0@conan/stable")
+
+    client = TurboTestClient(servers={"default": TestServer()})
+    pref1 = client.create(ref)
+    client.upload_all(ref)
+
+    client.create(ref2, conanfile=GenConanfile().with_requirement(ref))
+    client.upload_all(ref2)
+
+    client.run("remove {} -p {} -f".format(pref1.ref, pref1.id))
+
+    # Even if we create the package folder artificially, the folder will be discarded and installed again.
+    os.makedirs(os.path.join(client.cache_folder, "data", ref.dir_repr(), "package", pref1.id))
+
+    client.run("install {}".format(ref2))
+
+    assert "AssertionError: PREV" not in client.out


conans/client/graph/graph_binaries.py
```diff
@@ -91,6 +91,9 @@ def _evaluate_cache_pkg(self, node, package_layout, pref, metadata, remote, remo
         if not node.binary:
             node.binary = BINARY_CACHE
             metadata = metadata or package_layout.load_metadata()
+            if pref.id not in metadata.packages:
+                self._evaluate_remote_pkg(node, pref, remote, remotes)
```
memsharded: Calling ``_evaluate_remote_pkg`` is a bit confusing. I see the idea of protecting against wrong insertions, but this ``_evaluate_cache_pkg`` should probably not be called if there is not really a package, so maybe try to evaluate his possible condition before calling ``_evaluate_cache_pkg``:
```python
        # Maybe here, to complement the condition of ``package_id_exists``?
        if package_layout.package_id_exists(pref.id):  # Binary already in cache, check for updates
            self._evaluate_cache_pkg(node, package_layout, pref, metadata, remote, remotes, update)
            recipe_hash = None
```

memsharded (changes requested):
I like the idea to protect against this, but it seems that this should be captured earlier to redirect the flow earlier.

conans/client/graph/graph_binaries.py
SSE4: yes, I've checked, and it works as well. the function `_evaluate_cache_pkg` does some things before assertion happens (e.g. it tries to update revision from remote), but I guess it's okay to skip this logic if `pref.id` doesn't exist in `metadata`.

SSE4: ~~nope, unfortunately, an early check breaks several tests...~~
ah, I see:
```
AttributeError: 'NoneType' object has no attribute 'packages'
```
`metadata` could be `None`

conans/test/functional/command/install/install_update_test.py
```diff
@@ -316,3 +316,21 @@ def test_fail_usefully_when_failing_retrieving_package(self):
...
+        pref1 = client.create(ref)
+        client.upload_all(ref)
+
+        client.create(ref2, conanfile=GenConanfile().with_requirement(ref))
+        client.upload_all(ref2)
+
+        client.run("remove {} -p {} -f".format(pref1.ref, pref1.id))
+
+        os.makedirs(os.path.join(client.cache_folder, "data", "lib1", "1.0", "conan", "stable", "package",
+                                 pref1.id))
```
jgsogo: You can use `ref.dir_repr()`, it will return the `lib1/1.0/conan/stable` part

jgsogo (approved):
LGTM. Safer than it was.

memsharded (approved):
I think this is good, and should be low risk.

Recall this is targeted at release/1.32, we want it in 1.33 (and of course develop)

jgsogo:
We are waiting for the error to happen again in ConanCenter. We added a command just at the moment of entering a workspace to list all the files in it. So far, the workspace is always empty, and AFAIK the error hasn't happened again.

Heisenbug hunting  🐛 

image

jgsogo:
We show the error again, it happened as expected: https://github.com/conan-io/conan/pull/8519#issuecomment-839945387

I will add this to a release so it doesn't get lost. This patch is a step forward, for sure.

conans/test/functional/command/install/install_update_test.py
```diff
@@ -316,3 +316,20 @@ def test_fail_usefully_when_failing_retrieving_package(self):
...
+        client.upload_all(ref)
+
+        client.create(ref2, conanfile=GenConanfile().with_requirement(ref))
+        client.upload_all(ref2)
+
+        client.run("remove {} -p {} -f".format(pref1.ref, pref1.id))
+
+        os.makedirs(os.path.join(client.cache_folder, "data", ref.dir_repr(), "package", pref1.id))
+
+        client.run("install {}".format(ref2))
```
lasote: What is or what is not happening? In this cases I would like at least to assert that something has downloaded (for example) and adding a comment about what we are testing.

lasote (changes requested):
Please change the changelog line to something more useful

conans/test/functional/command/install/install_update_test.py
SSE4: it was raising here before, but now command is successful
`client.run` already checks for conan install status code, but maybe I need to think on how to make it more explicit

SSE4:
```suggestion
        client.run("install {}".format(ref2))
        self.assertNotIn("AssertionError: PREV", client.out)
```

SSE4: I added a check for an output (similar for other tests)

conans/test/functional/command/install/install_update_test.py
```diff
@@ -316,3 +316,20 @@ def test_fail_usefully_when_failing_retrieving_package(self):
...
+        client = TurboTestClient(servers={"default": TestServer()})
+        pref1 = client.create(ref)
+        client.upload_all(ref)
+
+        client.create(ref2, conanfile=GenConanfile().with_requirement(ref))
+        client.upload_all(ref2)
+
+        client.run("remove {} -p {} -f".format(pref1.ref, pref1.id))
+
+        os.makedirs(os.path.join(client.cache_folder, "data", ref.dir_repr(), "package", pref1.id))
```
lasote:
```suggestion
        # Even if we create the package folder artificially, the folder will be discarded and installed again.
        os.makedirs(os.path.join(client.cache_folder, "data", ref.dir_repr(), "package", pref1.id))
```

lasote (approved):

PR #8533, opened 2021-02-19
Feature: add property "cmake_target_aliases" to create some CMake alias targets (CMakeDeps)

### LINKED ISSUE #8360: [feature] Tool: create module official cmake targets
A build helper often used in CCI which could be part of Conan Tools.

First idea [here](https://github.com/conan-io/conan/pull/8338#discussion_r557653422).

Also present in CCI now:
https://github.com/conan-io/conan-center-index/pull/4257
https://github.com/conan-io/conan-center-index/pull/4258

```python
from conans.tools import create_module_official_cmake_targets

...

def package(self):
    create_module_official_cmake_targets(
            os.path.join(self.package_folder, "lib", "cmake", "conan-official-foo-targets.cmake"), 
            [{"official": "foo", "namespaced": "foo::foo"}]
     )
```

/cc @SpaceIm @madebr 

SSE4:
Changelog: Feature: Add property "cmake_target_aliases" to create some CMake alias targets using CMakeDeps.
Docs: omit

closes: #8360

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -28,20 +28,18 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         cmake_target_aliases = self.conanfile.cpp_info.\
             get_property("cmake_target_aliases", "CMakeDeps") or dict()
 
-        if isinstance(cmake_target_aliases, (list, tuple)):
-            target = "%s::%s" % (self.target_namespace, self.global_target_name)
-            cmake_target_aliases = {alias: target for alias in cmake_target_aliases}
+        target = "%s::%s" % (self.target_namespace, self.global_target_name)
+        cmake_target_aliases = {alias: target for alias in cmake_target_aliases}
 
         cmake_component_target_aliases = dict()
         for comp_name in self.conanfile.cpp_info.components:
             aliases = \
                 self.conanfile.cpp_info.components[comp_name].\
                 get_property("cmake_target_aliases", "CMakeDeps") or dict()
-            if isinstance(aliases, (list, tuple)):
-                target = "%s::%s" % (self.target_namespace,
-                                     self.get_component_alias(self.conanfile, comp_name))
-                aliases = {alias: target for alias in aliases}
-            cmake_component_target_aliases[comp_name] = aliases
+
+            target = "%s::%s" % (self.target_namespace,
+                                 self.get_component_alias(self.conanfile, comp_name))
+            cmake_component_target_aliases[comp_name] = {alias: target for alias in aliases}
 
         ret = {"pkg_name": self.pkg_name,
                "target_namespace": self.target_namespace,
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -81,6 +81,8 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         if(NOT TARGET {{alias}})
             add_library({{alias}} INTERFACE IMPORTED)
             set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
+        else()
+            message(WARNING "target '{{alias}}' already exists, alias for target '{{target}}' won't be created!")
         endif()
 
         {%- endfor %}
@@ -92,6 +94,8 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         if(NOT TARGET {{alias}})
             add_library({{alias}} INTERFACE IMPORTED)
             set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
+        else()
+            message(WARNING "target '{{alias}}' already exists, alias for target '{{target}}' won't be created!")
         endif()
 
             {%- endfor %}
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -28,11 +28,20 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         cmake_target_aliases = self.conanfile.new_cpp_info.\
             get_property("cmake_target_aliases", "CMakeDeps") or dict()
 
+        if isinstance(cmake_target_aliases, (list, tuple)):
+            target = "%s::%s" % (self.target_namespace, self.global_target_name)
+            cmake_target_aliases = {alias: target for alias in cmake_target_aliases}
+
         cmake_component_target_aliases = dict()
         for comp_name in self.conanfile.new_cpp_info.components:
-            cmake_component_target_aliases[comp_name] = \
+            aliases = \
                 self.conanfile.new_cpp_info.components[comp_name].\
                 get_property("cmake_target_aliases", "CMakeDeps") or dict()
+            if isinstance(aliases, (list, tuple)):
+                target = "%s::%s" % (self.target_namespace,
+                                     self.get_component_alias(self.conanfile, comp_name))
+                aliases = {alias: target for alias in aliases}
+            cmake_component_target_aliases[comp_name] = aliases
 
         ret = {"pkg_name": self.pkg_name,
                "target_namespace": self.target_namespace,
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -25,12 +25,23 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         target_pattern = "" if not self.find_module_mode else "module-"
         target_pattern += "{}-Target-*.cmake".format(self.file_name)
 
+        cmake_target_aliases = self.conanfile.new_cpp_info.\
+            get_property("cmake_target_aliases", "CMakeDeps") or dict()
+
+        cmake_component_target_aliases = dict()
+        for comp_name in self.conanfile.new_cpp_info.components:
+            cmake_component_target_aliases[comp_name] = \
+                self.conanfile.new_cpp_info.components[comp_name].\
+                get_property("cmake_target_aliases", "CMakeDeps") or dict()
+
         ret = {"pkg_name": self.pkg_name,
                "target_namespace": self.target_namespace,
                "global_target_name": self.global_target_name,
                "file_name": self.file_name,
                "data_pattern": data_pattern,
-               "target_pattern": target_pattern}
+               "target_pattern": target_pattern,
+               "cmake_target_aliases": cmake_target_aliases,
+               "cmake_component_target_aliases": cmake_component_target_aliases}
 
         return ret
 
@@ -57,6 +68,28 @@ class TargetsTemplate(CMakeDepsFileTemplate):
             add_library({{ target_namespace }}::{{ global_target_name }} INTERFACE IMPORTED)
             conan_message(STATUS "Conan: Target declared '{{ target_namespace }}::{{ global_target_name }}'")
         endif()
+
+        {%- for alias, target in cmake_target_aliases.items() %}
+
+        if(NOT TARGET {{alias}})
+            add_library({{alias}} INTERFACE IMPORTED)
+            set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
+        endif()
+
+        {%- endfor %}
+
+        {%- for comp_name, component_aliases in cmake_component_target_aliases.items() %}
+
+            {%- for alias, target in component_aliases.items() %}
+
+        if(NOT TARGET {{alias}})
+            add_library({{alias}} INTERFACE IMPORTED)
+            set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
+        endif()
+
+            {%- endfor %}
+
+        {%- endfor %}
 
         # Load the debug and release library finders
         get_filename_component(_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -25,7 +25,7 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         target_pattern = "" if not self.find_module_mode else "module-"
         target_pattern += "{}-Target-*.cmake".format(self.file_name)
 
-        cmake_target_aliases = self.conanfile.new_cpp_info.\
+        cmake_target_aliases = self.conanfile.cpp_info.\
             get_property("cmake_target_aliases", "CMakeDeps") or dict()
 
         if isinstance(cmake_target_aliases, (list, tuple)):
@@ -33,9 +33,9 @@ class TargetsTemplate(CMakeDepsFileTemplate):
             cmake_target_aliases = {alias: target for alias in cmake_target_aliases}
 
         cmake_component_target_aliases = dict()
-        for comp_name in self.conanfile.new_cpp_info.components:
+        for comp_name in self.conanfile.cpp_info.components:
             aliases = \
-                self.conanfile.new_cpp_info.components[comp_name].\
+                self.conanfile.cpp_info.components[comp_name].\
                 get_property("cmake_target_aliases", "CMakeDeps") or dict()
             if isinstance(aliases, (list, tuple)):
                 target = "%s::%s" % (self.target_namespace,
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
@@ -33,7 +33,7 @@ def test_global_alias():
         settings = "os", "compiler", "build_type", "arch"
 
         def package_info(self):
-            self.cpp_info.set_property("cmake_target_aliases", {"hello": "hello::hello"}, "CMakeDeps")
+            self.cpp_info.set_property("cmake_target_aliases", ["hello"], "CMakeDeps")
     """)
 
     cmakelists = textwrap.dedent("""
@@ -66,7 +66,7 @@ def test_component_alias():
 
         def package_info(self):
             self.cpp_info.components["buy"].set_property("cmake_target_aliases",
-                {"hola::adios": "hello::buy"}, "CMakeDeps")
+                ["hola::adios"], "CMakeDeps")
     """)
 
     cmakelists = textwrap.dedent("""
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
@@ -66,7 +66,7 @@ def test_component_alias():
 
         def package_info(self):
             self.cpp_info.components["buy"].set_property("cmake_target_aliases",
-                ["hola::adios"], "CMakeDeps")
+                ["hola::adios"])
     """)
 
     cmakelists = textwrap.dedent("""
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
@@ -1,0 +1,87 @@
+import textwrap
+import pytest
+
+from conans.test.utils.tools import TestClient
+
+consumer = textwrap.dedent("""
+from conans import ConanFile
+from conan.tools.cmake import CMake
+
+class Consumer(ConanFile):
+    name = "consumer"
+    version = "1.0"
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps", "CMakeToolchain"
+    exports_sources = ["CMakeLists.txt"]
+    requires = "hello/1.0"
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+""")
+
+
+@pytest.mark.tool_cmake
+def test_global_alias():
+    conanfile = textwrap.dedent("""
+    from conans import ConanFile
+
+    class Hello(ConanFile):
+        name = "hello"
+        version = "1.0"
+        settings = "os", "compiler", "build_type", "arch"
+
+        def package_info(self):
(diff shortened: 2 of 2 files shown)


prince-chrismc (approved):

conans/client/tools/cmake.py
```diff
@@ -0,0 +1,11 @@
+from conans.client.tools.files import save
+
+
+def create_cmake_module_alias_targets(_, module, targets):
```
memsharded: Better to put it in the ``conan.tools.cmake`` space?

SSE4: it's already in `conan.client.tools.cmake` namespace

uilianries (approved):

SSE4:
if someone has better ideas for UX, please share. right now it requires modifications in two methods: `package` and `package_info`:
```
                def package(self):
                    tools.create_cmake_module_alias_targets(self, self._module,
                        {"hello": "hello::hello"})

                def package_info(self):
                    self.cpp_info.builddirs = [os.path.dirname(self._module)]
                    self.cpp_info.build_modules = [self._module]
```
it would be nice, if one tool invocation could set everything needed (e.g. `builddirs` and `build_modules`), but:
- `CppInfo` cannot be set from the `package`
- `build_modules` requires existing file that has to be packaged on the package creation

I am not sure if there is a nice way to break a vicious circle, and reduce the chance of mistake and amount of code needed.

uilianries:
Up!

danimtb:
Would like to know the feedback from @memsharded and @lasote about this issue.

I am not sure, but I believe this use case is already considered in the new cpp_info model and doing this will not be necessary.
If that is correct, introducing this tool in current recipes will be another extra thing to migrate when we introduce conan 2.0 in ConanCenter. I think we should think about that as well as we want the recipes to be as compatible as possible.

memsharded (changes requested):
Please try a different approach:

- Define a ``cpp_info.set_property()`` that defines this data
- Have the ``CMakeDeps`` generator to honor this property and generate the necessary alias or cmake code to implement this
- Recipes code should be just 1 line with the set_property definition

SSE4:
@memsharded I've spent some type to prototype something with better UX via `set_property` as you have suggested.
now it requires just a single line from the recipe author.
for global alias:
```
def package_info(self):
    self.cpp_info.set_property("aliases", ["hello"], "cmake_find_package_multi")
```
or for components alias:
```
def package_info(self):
    self.cpp_info.components["adios"].set_property("aliases", {"hola::adios": "hello::adios"}, "cmake_find_package_multi")
```

pros:
- simple UX: requires just 1 intuitive line
- no need to package a build module
- the author may define as many aliases as he wants
- the author may use arbitrary namespaces (or no namespace at all)
- and even different namespaces per target

we may add custom CMake variables a similar way

prince-chrismc:
I am sold on the idea!

prince-chrismc:
👋 @SpaceIm @madebr (great feature to review)

lasote (commented):
Looks nice! but we cannot implement it in the `cmake_find_package_multi` and not in the `CMakeDeps`. It should be considered "the king" and "the king" goes first.

SSE4:
> Looks nice! but we cannot implement it in the `cmake_find_package_multi` and not in the `CMakeDeps`. It should be considered "the king" and "the king" goes first.

sure, I can implement it in `CMakeDeps` as well if overall UX looks good. pinging @memsharded for the feedback :)

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -11,6 +11,91 @@
...
+            import os
+            from conans import ConanFile, CMake, tools
+
+            class Conan(ConanFile):
+                name = "hello"
+                version = "1.0"
+                settings = "os", "arch", "compiler", "build_type"
+
+                def package_info(self):
+                    self.cpp_info.set_property("aliases", ["hello"], "cmake_find_package_multi")
```
memsharded: The property name is probably not enough:

- Properties are designed to be generic for multiple generators. In most cases, ``package_info()`` will not specify the generator name.
- If this is intended for the CMake build system, something like ``cmake_target_alias`` should do the work.

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -44,6 +46,15 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
         endforeach()
         """)
 
+    alias_template = Template(textwrap.dedent("""\
+        {%- for alias, target in aliases.items() %}
+        if(NOT TARGET {alias})
+            add_library({{ alias }} INTERFACE IMPORTED)
```
memsharded: Why not ``add_library(<name> ALIAS <target>)``?

memsharded (commented):
I would like this feature only in CMakeDeps.

It doesn't make much sense to provide it in the legacy, have some recipes update to support this in ``cmake_find_package_xxx``, then remove that generator for ``CMakeDeps``.

conans/client/generators/cmake_find_package_multi.py
SSE4: I think we're using it due to some compatibility issues we had with `ALIAS`?
```
New in version 3.11: An ALIAS can target a GLOBAL Imported Target

New in version 3.18: An ALIAS can target a non-GLOBAL Imported Target. Such alias is scoped to the directory in which it is created and below. The ALIAS_GLOBAL target property can be used to check if the alias is global or not.
```
so it doesn't require very latest cmake to work properly for all the cases

conans/test/functional/generators/cmake_find_package_multi_test.py
SSE4: I don't see why alias properly couldn't be generic for several generators.
so far we only needed it in cmake, but theoretically, any build system providing aliases and targets could benefit from it (premake, b2, qbs, etc)

memsharded: You are specializing the call for ``self.cpp_info.set_property("aliases", ["hello"], "cmake_find_package_multi")`` for the ``cmake_find_package_multi`` generator. This shouldn't be the case, the line should be something like:

```
self.cpp_info.set_property("aliases", ["hello"])
```

Then, IMO, it reads too generic, given that there are not known build systems that will use this information, and the information there, by its format pkg::component might not be used by other build systems. So something like ``cmake_target_alias`` makes more sense.

SSE4: I don't know, the property might be generic, but the value of the property might be different for the different generators.
we already know e.g. `name` and `filename` are generic properties, but different generators use different values, right?
for instance, name in cmake might be `Boost`, but in pkg-config it's `boost`.

memsharded: Exactly: ``name`` and ``filename`` are legacy, to be deprecated by the new properties: ``cmake_target_name``, ``cmake_file_name``. Same pattern to be used.

SSE4: okay, I'll change then

SSE4:
@memsharded I finally had some time to revisit it, now I've changed code to your suggestions:
- only `CMakeDeps` generator is supported
- property name is `cmake_target_aliases` (so it indicates it is specific for the CMake only)

and a new UX, for global aliases:
```
self.cpp_info.set_property("cmake_target_aliases", {"hello": "hello::hello"}, "CMakeDeps")
```
for component aliases:
```
self.cpp_info.components["buy"].set_property("cmake_target_aliases",  {"hola::adios": "hello::buy"}, "CMakeDeps")
```
and I think only dict values make sense, as they are explicit and straightforward - alias target is always specified.
so now you can create as many arbitrary aliases as you want, for any targets, with or without namespaces.
I think it should cover 99% of use-cases in conan-center.

madebr:
If this property should supposedly be used for each component to create an alias, why does it need to be a dict?
If I want to add an alias to a component, then Conan already knows the name of the original target?

```
self.cpp_info.set_property("cmake_target_aliases", ["alias1", "alias2::alias2"])
```

I also find having 2 properties for global aliases and component aliases potentially confusing.
Is it possible to have one property `cmake_target_aliases`?
From a recipe writer perspective, they shouldn't care about this. A cmake alias is a cmake alias.

SSE4:
> If this property should supposedly be used for each component to create an alias, why does it need to be a dict?

if you ever want to create an alias to the target other than the current component? maybe it's useless, in that case, array would be enough.

> If I want to add an alias to a component, then Conan already knows the name of the original target?

yes, it knows the name of the original target. I am not sure if this extra flexibility is needed.

> I also find having 2 properties for global aliases and component aliases potentially confusing. Is it possible to have one property `cmake_target_aliases`? 

it is possible, but I think there are recipes that don't define any components, and there are recipes using components.

madebr:
> > I also find having 2 properties for global aliases and component aliases potentially confusing. Is it possible to have one property `cmake_target_aliases`?
> 
> it is possible, but I think there are recipes that don't define any components, and there are recipes using components.

I don't know the Conan implementation, but can't it detect it from the context?

```
self.cpp_info.set_property("cmake_target_aliases", ["alias1", "alias2::alias2"])
```
versus
```
self.cpp_info.modules["comp"].set_property("cmake_target_aliases", ["alias1", "alias2::alias2"])
```

madebr:
> >    If this property should supposedly be used for each component to create an alias, why does it need to be a dict?

> if you ever want to create an alias to the target other than the current component? maybe it's useless, in that case, array would be enough.

The only use I can think of would be to create an alias of a target of a dependency. But I haven't yet seen a need for this.

SSE4:
@madebr I've added an automatic deduction of an alias target, now the UX is:
```
self.cpp_info.set_property("cmake_target_aliases", ["hello"], "CMakeDeps")
self.cpp_info.components["buy"].set_property("cmake_target_aliases",  ["hola::adios"] "CMakeDeps")
```

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
```diff
@@ -0,0 +1,87 @@
...
+    from conans import ConanFile
+
+    class Hello(ConanFile):
+        name = "hello"
+        version = "1.0"
+        settings = "os", "compiler", "build_type", "arch"
+
+        def package_info(self):
+            self.cpp_info.components["buy"].set_property("cmake_target_aliases",
+                ["hola::adios"], "CMakeDeps")
```
lasote:
```suggestion
            self.cpp_info.components["buy"].set_property("cmake_target_aliases",
                ["hola::adios"])
```

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
```diff
@@ -0,0 +1,87 @@
...
+    conanfile = textwrap.dedent("""
+    from conans import ConanFile
+
+    class Hello(ConanFile):
+        name = "hello"
+        version = "1.0"
+        settings = "os", "compiler", "build_type", "arch"
+
+        def package_info(self):
+            self.cpp_info.set_property("cmake_target_aliases", ["hello"], "CMakeDeps")
```
lasote:
```suggestion
            self.cpp_info.set_property("cmake_target_aliases", ["hello"])
```

conan/tools/cmake/cmakedeps/templates/targets.py
```diff
@@ -58,6 +78,28 @@ def template(self):
             conan_message(STATUS "Conan: Target declared '{{ target_namespace }}::{{ global_target_name }}'")
         endif()
 
+        {%- for alias, target in cmake_target_aliases.items() %}
+
+        if(NOT TARGET {{alias}})
+            add_library({{alias}} INTERFACE IMPORTED)
+            set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
+        endif()
```
lasote, line 86: Should we warn if the target already exists?

conan/tools/cmake/cmakedeps/templates/targets.py
```diff
@@ -25,12 +25,32 @@ def context(self):
         target_pattern = "" if not self.find_module_mode else "module-"
         target_pattern += "{}-Target-*.cmake".format(self.file_name)
 
+        cmake_target_aliases = self.conanfile.new_cpp_info.\
```
lasote: I think you have to update develop and merge, the `new_cpp_info` access is done now using just `cpp_info`, you should see that in other places at the `cmakedeps` after rebase.

conan/tools/cmake/cmakedeps/templates/targets.py
```diff
@@ -25,12 +25,32 @@ def context(self):
...
+            get_property("cmake_target_aliases", "CMakeDeps") or dict()
+
+        if isinstance(cmake_target_aliases, (list, tuple)):
```
lasote: Why this check? What if not? Should we error, right?

conan/tools/cmake/cmakedeps/templates/targets.py
```diff
@@ -25,12 +25,32 @@ def context(self):
...
+            target = "%s::%s" % (self.target_namespace, self.global_target_name)
+            cmake_target_aliases = {alias: target for alias in cmake_target_aliases}
+
+        cmake_component_target_aliases = dict()
+        for comp_name in self.conanfile.new_cpp_info.components:
+            aliases = \
+                self.conanfile.new_cpp_info.components[comp_name].\
+                get_property("cmake_target_aliases", "CMakeDeps") or dict()
+            if isinstance(aliases, (list, tuple)):
```
lasote: same

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
lasote: I would put a comment there saying: `The default global target is "hello::hello"`

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
lasote: Spanish 101 🤓  👍

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
```diff
@@ -0,0 +1,87 @@
...
+    """)
+
+    client = TestClient()
+    client.save({"conanfile.py": conanfile})
+    client.run("create .")
+
+    client.save({"conanfile.py": consumer, "CMakeLists.txt": cmakelists})
+    client.run("create .")
+
+    assert "hola::adios link libraries: hello::buy" in client.out
```
lasote, line 88: Maybe we can add a new test assigning different names (cmake_target_name, cmake_target_namespace) to make sure nothing is messed up.

lasote (changes requested):
I think this is good. I was not convinced about the alias, not distingüising the namespace, but now I see is simpler this way.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
SSE4: added one more test with `cmake_target_name`/`cmake_target_namespace`

conan/tools/cmake/cmakedeps/templates/targets.py
SSE4: removed the check

conan/tools/cmake/cmakedeps/templates/targets.py
SSE4: removed the check

conan/tools/cmake/cmakedeps/templates/targets.py
SSE4: rebased

conan/tools/cmake/cmakedeps/templates/targets.py
SSE4: added a warning

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
SSE4: added a comment

lasote (approved):

czoido (approved):

franramirez688:
@SSE4 is there any kind of documentation for this feature? I think it could be interesting to have a little explanation there, couldn't it?

SpaceIm:
> I would like this feature only in CMakeDeps.
> 
> It doesn't make much sense to provide it in the legacy, have some recipes update to support this in `cmake_find_package_xxx`, then remove that generator for `CMakeDeps`.

@memsharded Why? It prevents adoption of this new property in conan-center until conan v2, because it would break many downstream recipes or users (we can't know which generator is used by downstream). With this strategy, we have to keep ugly workarounds in CCI.

SpaceIm:
I have also another comment, slightly off-topic:

If `cmake_target_name` property was designed to generate exactly the target name (and not automatically adding a namespace), `cmake_target_aliases` and `cmake_target_namespace` properties wouldn't have been required. It would also have solved some issues of conflicting global targets between recipes (Vulkan & mongo recipes for example).

SSE4:
> If `cmake_target_name` property was designed to generate exactly the target name (and not automatically adding a namespace), `cmake_target_aliases` and `cmake_target_namespace` properties wouldn't have been required. It would also have solved some issues of conflicting global targets between recipes (Vulkan & mongo recipes for example).

my two cents, just a personal opinion
there are different use-cases for that properties:
1. `cmake_target_name` is for declaring a canonical target name, if it's different from the default (which is `package name::package name`). for instance, if I have a package named `sqlite3` but I want a canonical target named `sqlite::sqlite` (without version `3`), then I would use `cmake_target_name`.
2. `cmake_target_namespace` is just to control over namespace (in reality, there are no namespaces in CMake, it's just an integral part of the name). we already faced libraries that use different namespaces for different targets, for instance `OpenEXR` may use something like `Imath::Imath` and `OpenEXR::Iex` together.
3.  `cmake_target_alias` is if you want to declare more than one name for the same target. e.g. for `sqlite3`, you may want to define both `sqlite::sqlite` and `sqlite` (without namespace), because both are provided by the upstream. in that case, there are multiple canonical names for the same target.

so `cmake_target_name` is mostly for single target configuration, `cmake_target_namespace` is useful when you have multiple targets with different namespaces, and `cmake_target_alias` is useful then you have multiple canonical names for the same target. of course, there is an overlap between them, but ideologically they all serve different purposes.

SpaceIm:
>  `cmake_target_namespace` is just to control over namespace (in reality, there are no namespaces in CMake, it's just an integral part of the name). we already faced libraries that use different namespaces for different targets, for instance `OpenEXR` may use something like `Imath::Imath` and `OpenEXR::Iex` together.

From https://github.com/conan-io/conan/pull/10005#discussion_r748063200, I understand that `cmake_target_namespace` doesn't work at component level. So you can't define both `Imath::Imath` and `OpenEXR::Iex` in the same recipe with this property. `cmake_target_alias` is needed here.
But I say that this complexity could have been avoided, to allow to define exactly the relevant targets, without intermediate useless targets which could conflict with targets of others recipes.

SSE4:
> > `cmake_target_namespace` is just to control over namespace (in reality, there are no namespaces in CMake, it's just an integral part of the name). we already faced libraries that use different namespaces for different targets, for instance `OpenEXR` may use something like `Imath::Imath` and `OpenEXR::Iex` together.
> 
> From [#10005 (comment)](https://github.com/conan-io/conan/pull/10005#discussion_r748063200), I understand that `cmake_target_namespace` doesn't work at component level. So you can't define both `Imath::Imath` and `OpenEXR::Iex` in the same recipe with this property. `cmake_target_alias` is needed here. But I say that this complexity could have been avoided, to allow to define exactly the relevant targets, without intermediate useless targets which could conflict with targets of others recipes.

I totally agree, we shouldn't pollute the CMake scope with targets no one needs.
I personally would love to see `cmake_target_namespace` implemented on a component level, I don't get the reason why it can't be done that way.
IMO you should be able to declare both `Imath::Imath` and `OpenEXR::Iex` without any intermediate targets.
as it's experimental, I think it's not late to change and avoid unnecessary complexity, if any.

SSE4:
@SpaceIm we have discussed an issue internally, and have the following proposal:

- we're going to reduce complexity and overlap, starting with 1.43+
- changes might be breaking (expected, as `set_property` is experimental)
- `cmake_target_name` will accept a list of targets to declare as many aliases as needed
- `cmake_target_name` will always specify the full name as is (means, if you want a target named `zlib::zlib`, you specify `zlib::zlib` into the property)
- `cmake_target_namespace` and `cmake_target_aliases` properties will be removed, as redundant (as same effect could be achieved with just `cmake_target_name`)
- there will be no intermediate targets defined, you will get only targets you asked for, and cmake scope won't be polluted
- `cmake_tatget_name` will work on both component and global level
- if you don't specify property explicitly, there will be `recipe name::recipe name` and `recipe name::component name` targets generated by default

there are still details to be figured out, we're going to open new issue and invite you to the discussion

hopefully, another wave of `set_property` changes won't be so bad for ConanCenter, as there are not so many recipes already using `set_property` right now (maybe around 20 recipes).

SpaceIm:
I like this new logic.

> cmake_target_name will always specify the full name as is (means, if you want a target named zlib::zlib, you specify zlib::zlib into the property)

Ok, so it's urgent to do nothing in CCI, except go back to `.names` to avoid breaking change, until `set_property()` is stable enough. One drawback for `CMakeDeps`: https://github.com/conan-io/conan/issues/9943

Question: What will be the name of the component when you call `find_package(Package COMPONENT component1 component2)`? Will you split `::` and take the second part by default? Few libs have a different name for the component and the imported target (SFML for example), will there be a way to define this value without creating a fake target?

SSE4:
> Question: What will be the name of the component when you call `find_package(Package COMPONENT component1 component2)`? Will you split `::` and take the second part by default? Few libs have a different name for the component and the imported target (SFML for example), will there be a way to define this value without creating a fake target?

that's an interesting question, and probably will require some investigation.
CMake [documentation](https://cmake.org/cmake/help/latest/command/find_package.html) is a bit vague:

```
The set of available components and their meaning are defined by the target package. 
Formally, it is up to the target package how to interpret the component information given to it, 
but it should follow the expectations stated above. 
For calls where no components are specified,
 there is no single expected behavior and target packages should clearly define what occurs in such cases. 
Common arrangements include assuming it should find all components, 
no components or some well-defined subset of the available components.
```

is it possible that I do `find_package(Package COMPONENT component_name)` and get a target with name different from `Package::component_name`? IDK right now, maybe need to do some research with existing libraries

Issue #8335, opened 2021-01-14
[question] how to lock build-requires in a derived 'partial' lockfile


DukeOfKirkcaldy:
When integrating changes into a product in CI, I rebuild any dependent packages by calculating the build-order.  For each package that is rebuilt I want to have  a 'partial' lockfile for a package which contains all of the settings, options, dependencies, etc. needed to completely reproduce the package again in the future.  I am currently making this file like this (3rd command):

```
// rebuild downstream consumer
conan install ${ref} --build=${ref} --build=missing --lockfile=${product_lockfile} --lockfile-out=${product_lockfile_updated}

// update product lockfile with information about newly built package
conan lock update ${product_lockfile} ${product_lockfile_updated}

// derive a partial lockfile which can be used to reproduce this package later
conan lock create --reference=${ref} --lockfile=${product_lockfile_updated} --lockfile-out=${package_lockfile}
```

But it seems build requirements are missing from this lock causing rebuilding from lockfile to fail:

```conan install ${ref} --build=${ref} --lockfile=${package_lockfile}```

> ERROR: Build-require '<my_build_require>' cannot be found in lockfile

So how can I lock the build-requires of the package?  I tried to force them with a build-policy flag but it raised an error because the reference is already locked:

```
conan lock create --reference=${ref} --build=${ref} --lockfile=${product_lockfile_updated} --lockfile-out=${package_lockfile}
```

> ERROR: Cannot build '<ref>' because it is already locked in the input lockfile

**Comment by memsharded** [comment]
Hi @GordonJess 

I think this was fixed in 1.32, in the changelog:

> BugFix: Allow lockfiles to be relaxed with the –build argument. https://github.com/conan-io/conan/pull/8054 . Docs https://github.com/conan-io/docs/pull/1939/files


Can you please confirm that you are using 1.32? thanks!

**Comment by ytimenkov** [comment]
I get the message `Build-require cannot be found in lockfile` if the same requirement is both in `requires` and `build_requires`.

conanfile.py:
```python
from conans import ConanFile


class ConanTestConan(ConanFile):
    name = "conan-test"
    version = "1.0"

    requires = ["trompeloeil/39"]
    build_requires = ["trompeloeil/39"]
```

```sh
$ conan lock create --lockfile-out=deps.lock conanfile.py
...
Generated lockfile: /home/default/test/conan-repro/deps.lock


$ conan create --lockfile=deps.lock --lockfile-out=conan.lock conanfile.py 
[...]

PR #5592, opened 2019-08-07
[1/n] [xbuild] Add support for crossbuilding into the graph

jgsogo:
Changelog: Feature: Add support for cross-building into the graph
Docs: omit

- [ ] Generators: does it make sense to generate a `FindXXXX.cmake` for a `build_require` for the _build_machine_?

Thi PR adds the needed changes to the graph to support two contexts for nodes that build allow the implementation of a new cross-building model into Conan.  `CONTEXT_BUILD` is for nodes referring to the _build machine_ and `CONTEXT_HOST` is for nodes referring to the _host machine_.

It is **very important to note that the default behavior for `build_requires` is different depending on the profiles provided**:
 * if a `profile_build` is provided, then `build_requires` will be associated by default with the _build machine_, while
 * if the `profile_build` is not provided then the `build_requires` will be associated ~~by default~~ with the _host machine_  (current behavior).

~~This PR adds the posibility to specify the context for a `build_requires` using the verbose declaration: `self.build_requires("....", context="build")` to force the association of the required reference to the given context (it will raise an exception if require is forced to `build` but no `profile_build` is provided).~~

The recipe can force a build-require to reside in the _host context_ using `self.build_requires("...", force_host_context=True)`, but it cannot force it to be in the _build context_ when no profile for the _build_machine_ is provided.

CLI is proposed here: https://github.com/conan-io/conan/pull/5594

Superseedes: https://github.com/conan-io/conan/pull/5184

PR #5184, opened 2019-05-20
[xbuild] Manage host/build contexts inside the graph

jgsogo:
Changelog: (Feature | Fix | Bugfix): Describe here your pull request
Docs: https://github.com/conan-io/docs/pull/XXXX

This PR is here to serve as an overview of the current implementation of the cross building model. Feel free to comment about it, but do not review in detail as many things will change (and probably this one won't be the final PR).

This is implementing **only the _backend_** part of the feature, for the command line interface, visit https://github.com/jgsogo/conan/pull/13.

There are some changes from this PR that are on its own PR to review them better or to be merged in advance:
 * Helper class to store `Node` objects: https://github.com/conan-io/conan/pull/5179

PR #8534, opened 2021-02-19
Feature/environment propagate

memsharded:
Changelog: Feature: New environment management, with correct value capturing and restoring on deactivate
Docs: https://github.com/conan-io/docs/pull/2060

New approach to environment (env-vars) management:

**VirtualEnv**
- New generator that will generate ``buildenv.xxx`` and ``runenv.xxx``
- Collects info from dependencies visiting the graph, *not* by the accumulated deps_env_info
- It also collects information from profiles
- It automatically collects runtime information based on ``cpp_info`` definition: exes, bin-paths, lib_paths to define PATH, LD_LIBRARY_PATH, DYLD_LIBRARY_PATH envvars

**ConanfileDependencies**
- Entry point in each recipe to visit its dependencies
- Access to dependencies conanfiles is restricted by a "read-only" conanfilewrapper

**CMakeGen**
- Proof of concept of an aggregated meta-generator that includes CMakeToolchain, CMakeDeps and VirtualEnv
- CMake build helper will integrate calls to VirtualEnv generated files


conans/test/unittests/tools/env/test_env.py
```diff
@@ -0,0 +1,212 @@
...
+import pytest
+
+from conan.tools.env import Environment
+from conan.tools.env.environment import ProfileEnvironment
+from conans.client.tools import chdir
+from conans.test.utils.test_files import temp_folder
+from conans.util.files import save
+
+
+def test_compose():
```
lasote: More tests:
 - composing prepend/append, append/prepend and so on
 - Using separators

conan/tools/env/environment.py
```diff
@@ -0,0 +1,230 @@
+import fnmatch
+import os
+import textwrap
+from collections import OrderedDict
+
+from conans.errors import ConanException
+from conans.util.files import save
+
+_ENV_VAR_PLACEHOLDER = "$PREVIOUS_ENV_VAR_VALUE%"
```
lasote: should be named "CONAN_SOMETHING" like the `_PATHSEP`?

conans/test/unittests/tools/env/test_env.py
```diff
@@ -0,0 +1,212 @@
...
+    assert env3.value("MyVar") == "MyNewValue"
+    assert env3.value("MyVar2") == 'MyValue2 MyNewValue2'
+    assert env3.value("MyVar3") == 'MyNewValue3 MyValue3'
+    assert env3.value("MyVar4") == ""
+    assert env3.value("MyVar5") == 'MyNewValue5'
+
+
+def test_profile():
+    myprofile = textwrap.dedent("""
+        # define
```
lasote: Is it possible to specify a separator in the profile?

lasote (commented):
The overall idea looks pretty good. I miss some more unit tests mostly and we need to try this with more pieces (toolchains etc) and real recipes to confirm that it fits.

memsharded:
Please @lasote check the changes in https://github.com/conan-io/conan/pull/8534/commits/5f323b1ac58ced09997629f5d882e03da95177d8:

- Proposes ``CMakeGen`` integration that aggregates everything: CMakeToolchain, CMakeDeps, VirtualEnv
- CMake will use the ``buildenv`` environment if existing, to inject env-vars

lasote:
@memsharded sublime. Love it.

conans/test/unittests/tools/env/test_env.py
memsharded: No, it seems the syntax will be weird, especially to designate whitespaces

conan/tools/env/environment.py
memsharded: Fixed: I have replaced those strings for real types, to be able to filter the list of values more reliably

conans/test/unittests/tools/env/test_env.py
memsharded: Adding a few more tests, please check.

solvingj:
Excellent work!  I really now appreciate the cleaner separation and management between the environment variables for build and run. I think the `CMakeGen` meta-wrapper makes sense now that we've had some time for the other parts to evolve. 

I will use this opportunity to point out that the `CMake` helper would have been a good candidate to use the alternative generic/python `.env` strategy I proposed offline (provide a third option, de-coupled from user shells):
- generate generic `run.env` and `build.env` files with pure key/value pairs
- provide a conan command which can read and use these env files: 
    - `conan exec --withenv=run.env cmake build ...`
    - this would just be a super-simple wrapper for  `tools.with_env(env):` and `os.subprocess('whatever')`
- use this mechanism in `CMake` helper instead of calling the `.bat`/`.sh` scripts

I understand it's a bit too ambitious and un-proven as a concept internally yet, so it's just a suggestion for future.  But, I'm confident it's a great idea to provide this option. I definitely think it would be a better internal implementation for our own python code such as this (and every other future build system wrapper).  But also, consider the vast number of other users trying to orchestrate Conan and these environments with their own python code, or cmake scripts, etc.  It's really unnecessary and has significant disadvantages to have to go through the shell scripts with to get to the environment variables like this: 

https://github.com/conan-io/conan/pull/8534/files#diff-56b36c23589a66b0bd803c545703684c74400cbb7bd8538b8949ed84613e3fc7R22-R27

```python
def environment_wrap_command(filename, cmd):
    if filename.endswith(".bat"):
        return "{} && {}".format(filename, cmd)
    elif filename.endswith(".sh"):
        return 'bash -c ". {} && {}"'.format(filename, cmd.replace('"', r'\"'))
    raise Exception("Unsupported environment file type {}".format(filename))
```

The fact that we're not even providing any option to produce generic `.env` files that can be automated with custom scripts/languages feels like an obvious thing we could change and improve.  Whether or not we provide the wrapper command could be a separate discussion.

That was the only thing I would suggest for fundamental change, perhaps in the next iteration.  Even without that, this is a huge improvement and feels like we're closer to having this part really mastered.

memsharded:
Thanks @solvingj for the review.

There are a few things that prevent to have a nice solution based on generic ``.env`` files. Basically the assumption ``generate generic run.env and build.env files with pure key/value pairs`` doesn't hold:

- The composition logic of environments results in complicated patterns to resolve. One environment should be able to do something like ``NEW_VAR = Val1 Val2 <Previous NEW_VAR> Val3 Val4``, as a result of appends and prepends of information. Different shell scripts should now be able to parse those lines, interpret and correctly evaluate them.
- There is no unique placeholder for representing the current/previous value of the environment variable. While it is very easy to encode such as $VAR, %VAR%, achieving that when the variable name is a variable itself is at least challenging, because the shell and batch evaluation is often based on expansion and not pure variable evaluation.
- Paths need to be handled, because the path separator is different between OSs, but also different for different subsystems. So it cannot be replaced in the ``.env`` file, and some mechanism to identify "this is a path", and concatenate them is necessary. Now the parsing becomes more complex because paths with path spaces, and need to quote/unquote things, which makes things more complicated and fragile.

Note that the shell/batch script is still mandatory, we cannot go without it. After a ``conan install`` it must be possible to use native tools, and be able to build without calling any other Conan command again. If we don't provide the shell scripts, then a ``conan build`` is mandatory again, and forbids many other user flows and tools.

Said that, I think that generating  serialized xxxx.env files from the ``Environment`` class is an easy thing, assuming they will be deserialized and used by Conan only, and not by the shell scripts. But having our code rely on that functionality might end in non-working user shell scripts, (because Conan internal .env management will eventually diverge from the scripts, which is exactly what we are trying to achieve with the new build system integration paradigm). So at the moment I prefer to focus on one way to do this thing, and make sure that it works reliably, robustly, and with low cognitive overhead, and there will be time to consider UX enhancements like this one.

conan/tools/env/environment.py
```diff
@@ -0,0 +1,265 @@
...
+
+class _Sep(str):
+    pass
+
+
+class _PathSep:
+    pass
+
+
+def environment_wrap_command(filename, cmd):
```
ytimenkov: Can this be made somehow extensible (even maybe a jinja template)? bash may not be present everywhere (well, maybe `sh` is required for POSIX).
What about powershell on Windows? 
Also for linux `sh` will do for running from toolchain, but support for other shells should be possible for user.

conan/tools/env/environment.py
```diff
@@ -0,0 +1,265 @@
...
+    if filename.endswith(".bat"):
+        return "{} && {}".format(filename, cmd)
+    elif filename.endswith(".sh"):
+        return 'bash -c ". {} && {}"'.format(filename, cmd.replace('"', r'\"'))
```
ytimenkov: This looks like an escaping hell (backslashes need to be escaped as well for example, maybe spaces...)

conan/tools/env/environment.py
```diff
@@ -0,0 +1,265 @@
...
+    def prepend(self, name, value, separator=" "):
+        value = self._list_value(value, _Sep(separator))
+        self._values[name] = value + [_Sep(separator)] + [_EnvVarPlaceHolder]
+
+    def prepend_path(self, name, value):
+        value = self._list_value(value, _PathSep)
+        self._values[name] = value + [_PathSep] + [_EnvVarPlaceHolder]
+
+    def save_bat(self, filename, generate_deactivate=True, pathsep=os.pathsep):
+        deactivate = textwrap.dedent("""\
```
ytimenkov: Another point about Jinja: if you have those templates then you can just go through the list of templates and generate each one. This will be customizeable and easily extensible. For example add setting:
`virtualenv.templates=buildenv.sh.j2,runenv.sh.j2,buildenv.fish.j2`
(Also modules / blocks can be loaded from config file location).

I rewrote my generator in Jinja. The template is arguably easier to read but python code is significantly simpler without all if'ed logic.

Another benefit of such user-extensibility is that it would be easier to write plain `.env` files as @solvingj  asked and which could be used in IDE (at least many VS Code plugins start supporting `.env` files: python, C++ debugger).

Also doing this as a template doesn't mean immediate decision to publish as a user-facing contract, but still (I believe) will reduce complexity of python code.

The only trick is to carefully handle this for pyinstaller / zip-friendlyness.

ytimenkov (commented):
Just to understand better (I always confuse build vs host, to me it's host vs target 😊):
1. build requirement has `bin_dirs` / `lib_dirs` (e.g. a tool cmake, flatc) -> ends up in `buildenv` (`PATH` / `LD_LIBRARY_PATH`)
2. package has `bin_dirs` / `lib_dirs` and it's application -> ends up in `runenv` when I do `conan install <ref> -f VirtualEnv`
3. Package is a (host) requirement and has `bin_dirs` / `lib_dirs` -> ends up in `runenv`.

correct? So for development I activate only `buildenv` and get my tools. 

But then `buildenv_info` / `runenv_info` are less clear:
1. a package is build requirement: where do its `build` / `run` infos go?
2. a package is a (host) requirement: ditto, where do explicit `build`/`run` envs end up?

Now the most unclear thing is `env_info.PYTHONPATH`. We use python packages and they can be both build requirement (as a plug in for build tool) and ordinary (a test runner when package is consumed).
Naturally I would expect to use `buildenv` in first case and `runenv` in the second. What are the needed changes for my setup? Should those recipes be changed to set both `build_env_info` and `run_env`_info properties?

Another thing to confirm: requirements from profile are always build ones, right? Asking because we have Conan packages with gcc and python and they're part of profile. However when running tests I would like to have them: python for test runner and gcov to process coverage (thus python should be in `PATH`) so I have to activate both build and run environments, right?

The last part is a bit tricky with lockfiles: I can't specify profile when using lockfile, so I have to create a simple consumer file with single dependency. Then (as in item 3) my `bin_dirs` / `lib_dirs` should end up in `runenv`, right?

memsharded:
The mapping of explicitly define env-vars in ``buildenv_info`` and ``runenv_info`` should be straightforward, each one goes to its respective file

> a package is build requirement: where do its build / run infos go?

A build requirement should only define ``buildenv_info``, as by definition the build-requirement should not exist in runtime.

> a package is a (host) requirement: ditto, where do explicit build/run envs end up?

``buildenv_info`` will end in ``buildenv.xxx`` and ``runenv_info`` ends in ``runenv.xxx``

memsharded:
> Now the most unclear thing is env_info.PYTHONPATH. We use python packages and they can be both build requirement (as a plug in for build tool) and ordinary (a test runner when package is consumed).
Naturally I would expect to use buildenv in first case and runenv in the second. What are the needed changes for my setup? Should those recipes be changed to set both build_env_info and run_env_info properties?

Yes, I guess so. Maybe if the tests are only intended to run in the ``build()`` method, it could be considered a "buildenv" only. But it might be perfectly possible that different python utilities are used for the build and other for the tests, so having them in each build/run info, would make sense, wouldn't it? I am not sure if clear from the code, but it would be enough to define one ``Environment`` and assign it to both.

> Another thing to confirm: requirements from profile are always build ones, right? Asking because we have Conan packages with gcc and python and they're part of profile. However when running tests I would like to have them: python for test runner and gcov to process coverage (thus python should be in PATH) so I have to activate both build and run environments, right?

The profiles will also contain ``[buildenv]`` and ``[runenv]`` both for build and host context while building. So far the ``[env]`` was only build one, but with this new proposal is possible to also define vars for the runenv, which might be useful for many users.

> The last part is a bit tricky with lockfiles: I can't specify profile when using lockfile, so I have to create a simple consumer file with single dependency. Then (as in item 3) my bin_dirs / lib_dirs should end up in runenv, right?

This is not very clear, lockfiles contain a copy of the profiles, shouldn't require any specific setup?

conan/tools/env/environment.py
memsharded: The problem is that the current scripts are using bash syntax, it is not just calling it, the scripts themselves should be adapted for other shells. If it is possible to make the xxxenv.sh scripts shell generic, that would be great.

The important bit I needed specific bash features if for capturing the environment for the ``deactivate_xxx.sh`` script.

conan/tools/env/environment.py
memsharded: Our tests run by default with spaces in paths, but indeed we should check it. We are adding more real tests that will hopefully cover these cases.

conan/tools/env/environment.py
memsharded: Yes, I agree it would be great to provide templates for this, and provide user extensibility. At this point my priority is to define the new paradigm:

- How generators access dependencies information via a visitor (ConanFileDependencies)
- The new separation of build and run envs
- The explicit generation of environment scripts for handling env-vars instead of doing everything dynamically in Conan (and thus not allowing many user flows)
- A clear and explicit mechanism to define env-var operations: append, prepend, clear, define, and how these compose for multiple levels (profile, build-requires, toolchains...)

Still the problem of .env files is not fully clear. In Conan we clearly have the user demand for those different operations: append, prepend, define, unset, and those .env files cannot do that. Also, in Conan users might need to concatenate more than one environment definition (for example the definition coming from profiles and build-requires, and another definition coming from a toolchain), that both will need to perform different operations like appending. Those .env files cannot do this either, up to my knowledge.

memsharded:
@ytimenkov 

I am experimenting with removing the environment capture, and it works without bash, which I agree would be better. I'll check if there is a possibility to capture the "deactivate_xxx.sh" script in pure shell without using bash, and at the moment, I am making the generation of "deactivate_xxxx.xxx" scripts opt-in, instead of opt-out.

conan/tools/env/environment.py
ytimenkov: > Still the problem of .env files is not fully clear.

At least in VS Code many plugins can use `.env` files instead of having settings. For example, python extension appends `PYTHONPATH` from it to its internal search path, C++ extension has similar option in debug configuration (for example to run program with modified `LD_LIBRARY_PATH`).

It may not even be possible to run IDE in such an activated environment, because, for example for workspace each folder (repo / project) has own settings / environment.

I'm not suggesting adding this by default, but some users may find it useful to customize and produce own files, even if they for example capture complete variables at the install time.

ytimenkov (commented):
> Maybe if the tests are only intended to run in the build() method, it could be considered a "buildenv" only.

We run tests separately. So job1 builds conan package, uploads it and publishes a lockfile as an artifact. Next jobs take this artifact, do `echo -e "[requires]\n$REF" > conanfile.txt && conan install . -g virtualenv -g virtualrunenv -o pkgname:testing=True`  and then like `python3 test.py` which is in the path and package-specific so it's the purpose of Conan package to provide right info to run particular tests.

However this is slightly irrelevant: `PYTHONPATH` is defined explicitly and I can put it to both run_env and build_env. This part is straight-forward.

I think my primary question was about `autorun_environment` (new function in this PR) and where do `PATH` and library paths end up for requirements and build requirements (including ones coming from profiles).

conan/tools/cmake/__init__.py
```diff
@@ -1,3 +1,4 @@
 from conan.tools.cmake.toolchain import CMakeToolchain
 from conan.tools.cmake.cmake import CMake
 from conan.tools.cmake.cmakedeps import CMakeDeps
+from conan.tools.cmake.cmakegen import CMakeGen
```
memsharded: The name ``CMakeGen`` is not great, ideas welcome (it is a wrapper around CMakeToolchain, CMakeDeps and VirtualEnv)

conan/tools/env/environment.py
memsharded: The bash requirement has been removed. Powershell support will also be added.

conan/tools/env/environment.py
```diff
@@ -0,0 +1,269 @@
...
+                        new_value = new_value[1:]
+                    val = new_value[-1]
+                    if isinstance(val, _Sep) or val is _PathSep:
+                        new_value = new_value[:-1]
+                    self._values[k] = new_value
+        return self
+
+
+class ProfileEnvironment:
+    def __init__(self):
```
SSE4: may ctor receive arguments as loads? I see usage pattern is `x = ProfileEnvironment(); x.load(something)`

conan/tools/cmake/cmakegen.py
```diff
@@ -0,0 +1,14 @@
+from conan.tools.cmake import CMakeDeps, CMakeToolchain
+from conan.tools.env import VirtualEnv
+
+
+class CMakeGen:
```
SSE4: I am not sure, why is it needed? it seems to be only combination of 3 other generators

memsharded: It is a high level wrapper of those three. Otherwise, all recipes will have to declare ``CMakeToolchain``, ``CMakeDeps`` and ``VirtualEnv`` (or other toolchains). Its intention was to simplify the general UI for many cases.

In some other toolchains it will be more evident, like ``Autootools``. If a recipe uses ``AutooToolsToolchain``, ``AutoToolsDeps``, ``VirtualEnv``, then it will get 3 environment files, like ``buildenv.sh``, ``toolchain.sh`` and ``conandeps.sh``, and will have to activate one after the other for normal user usage. If using a higher ``AutootoolsGen`` it is possible to compose the output of the three very easily in a single ``buildenv.sh`` which will be simpler for the user.

conan/tools/env/environment.py
memsharded: Yes, lets change it to ``ProfileEnvirionment.loads(something)`` staticmethod, as it will be more aligned with the rest of the codebase.

conan/tools/cmake/__init__.py
ytimenkov: Build system? Workflow? Toolset?

Maybe toolchain was a bad name in first place. In cmake it's a kind of bootstrap parameters /initial variables' values to override search rules in enable_language / project.

Maybe InitVars is a better name in Conan instead of toolchain.

memsharded: I agree I kind of like ``Toolchain`` name for the wrapper, but on the other hand ``Toolchain`` name was chosen based on what seems to be the most common usage (CMake like): the toolchain file of a build system is typically a file that defines the compiler, architecture, cross-build parameters, etc. This is basically what the Conan ``settings`` and ``options`` is, so the main responsibility of ``XXXXToolchain`` classes is to map from settings and options to build system specific mechanism. And also, because the file is the one that is passed as ``-DCMAKE_TOOLCHAIN_FILE=conantoolchain.cmake``. In that sense the name seems good for the current usage.

solvingj:
I have had several discussions with @SSE4 and @memsharded about the relationship between the `virtualenv` generator features, and the `generate()`/`build()` methods for Conan 2.0.  I have thought about it extensively over the past few weeks.  As I understand it, there is a fundamental problem that is not currently addressed.  I don't think we should move any of this forward until we solve it, because it influences a lot of the design shown here. 

The Conan 1.0 application makes a implicit guarantee to it's **callers**, and it holds true for 100% of recipes.  That guarantee is relied upon by the vast majority of Conan users today.  The new strategy for `virtualenv`+`generate`+`build` breaks this guarantee.  

That guarantee is that `build_requires` will ALWAYS be prepended to the `PATH` environment variable during the `build()` method, and therefor they will always be used in every recipe whenever `self.run` is called. The recipe authors do not have to do anything to make their recipes exhibit this behavior, and they can't reasonably do anything to escape this behavior.  It's implicit, and therefor it's guaranteed. It is a cornerstone of the `build_requires` feature.

The new strategy recognizes that many implicit things in Conan have some drawbacks, such as confusion for users and lack of reproducibility (or difficulty to reproduce).  Thus, it represents a radical shift to be 100% explicit to avoid these drawbacks.  I was in favor and supportive of this strategy, up until the point I realized that it breaks the guarantee of `build_requires` being used in recipes 100% of the time.  The new strategy (as discussed so far), relies on the recipe author to do `something` specifically and correctly in the recipe, otherwise `build_requires` will not be used in that recipe. 

I don't have an answer to the problem. I've thought about several ways to address it, and each of them is pretty un-elegant. 
Nonetheless, in my opinion, we must find some way to uphold the existing guarantee for `build_requires` in the `build()` (and `package()`) methods before we move forward: full-stop.

solvingj:
Ok, I see, it's new as of Cmake 3.15

solvingj (changes requested):
In summary, I feel that we need to have a working concept which guarantees all build_requires work for all recipes automatically, and does not rely on the author to know or do anything.

memsharded:
@solvingj it was left out at this stage, I thought that it was adding yet another point of UI, but I understand the concerns about not seeing it here. I have added it in https://github.com/conan-io/conan/pull/8534/commits/1f66a6c964a62cca71ebd650ce8674035598ffe5, please check.

solvingj:
Thanks for giving it some attention @memsharded .  From what I see, each recipe still needs to do `something` correctly, otherwise their recipe won't incorporate `build_requirements`, is that right?

memsharded:
> Thanks for giving it some attention @memsharded . From what I see, each recipe still needs to do something correctly, otherwise their recipe won't incorporate build_requirements, is that right?

Nop, only to use ``VirtualEnv`` generator, directly or implicitly (if used by any other high level generator). 
This is an actual design goal of Conan 2.0. It is not good to have to inject something, always, in all recipes, in all methods, about the environment, which typically you have no control of what is being injected. Recall that via ``VirtualEnv`` generator, you can get the flexibility and customization you want, for example, both "build" and "run" environments are generated, and used in different situations, which only a recipe knows and can decide.

solvingj:
> Nop, only to use VirtualEnv generator, directly or implicitly (if used by any other high level generator).

I'm translating this to "Yes, the user must do `something`, or `build_requires` will not work in the recipe.". 

> This is an actual design goal of Conan 2.0. 

This is the fundamental problem that I do not think is acceptable. This breaks the long-standing and widely-depended on feature/promise/guarantee of `build_requirements`. 

> It is not good to have to inject something, always, in all recipes, in all methods, about the environment, which typically you have no control of what is being injected. 

I believe you've conflated two (actually three) separate things and that is the fundamental problem with this new design.  I agree with part, but disagree with the other part.  Lets pull them apart:

1.  >  you have no control of what is being injected. 
    - I agree, this is the big problem, and I think we need to change it.  Specifically, I think the problem is that the CALLER has no control over what is being injected (not so much the recipe AUTHOR).  We can approach this problem in a number of different ways. Most importantly, I do NOT think it requires breaking `build_requires` in the default case, or requiring authors to do something in the default case.  We can, for example, approach it is as a configuration item which defaults to "on" but can be opt-out.  We have always had `conan.conf`, but now we can look to the more suitable `[conf]` mechanism to control this behavior. This now also lets us think about more granular and interesting controls within the generated environment files, for dealing with the differences between build and run envs, etc.  But, definitely, I agree that the CALLER should be in full control of what is being injected. 

2.  > It is not good to have to inject something, always, in all recipes, in all methods, about the environment
    - This is a different thing altogether.  There is tremendous value in the predictability and reliability of things that an application provides/injects always, automatically.  IMO, the overwhelming success and adoption of `build_requires` owes a lot to this guarantee.  If it only worked for "some recipes if the author did it right", I feel strongly that it would not be used, because it can't be counted on.  The problem of not being able to opt-out or control anything is a separate problem.  But, there is tremendous value in the Conan giving every single recipe something completely free and automatically. 

3. You didn't call it out in this discussion, but we're both aware of a third topic driving this new vision which is the demand for local reproducibility of the build environment. Implicit/automatic stuff available only in the recipe has historically been a challenge for local reprodicibility.
    - We now know that there's a very important and new high-level goal: 
        - "make it trivial for me to run `conan install` and then reproduce Conan's `build()` method locally outside the recipe". 
    - Again, we have a variety of ways to meet this goal for our users, but they do NOT require the elimination of automatic behavior in the conan recipe.  They only really require the ability to reproduce that automatic behavior outside the recipe.  Again, a very different thing.  

Finally, to re-iterate.  IMO, it's just not acceptable to take away the promise that all recipes will respect `build_requires` automatically, by default.  I now hope that it's a bit more clear that it's also not strictly necessary to do so.  We can achieve the other goals of local reproducibility and giving user complete control over the generation process, without a philosophical goal of "no automatic recipe behavior ever".  I believe most other users will agree with these sentiments. Also, I feel strongly, but I do acknowledge that I could still be wrong.

memsharded:
@solvingj yes, I think you are right, lets make this opt-out instead of opt-in. I'll change the PR.

jgsogo (commented):
What is the difference between `buildenv_info` and `runenv_info` in the `package_info()`?

How it manages a conflicting env-var between different packages providing the same env-var name?

solvingj:
@memsharded 
>  lets make this opt-out instead of opt-in

Great!  I reviewed the changes. I'm not sure if it gets overridden somewhere, but I do see that the `self.run()` function still has `run_environment=False`, as a default.  Will `self.run()` pick up the build_environment by default?

memsharded:
@solvingj 

>Great! I reviewed the changes. I'm not sure if it gets overridden somewhere, but I do see that the self.run() function still has run_environment=False, as a default. Will self.run() pick up the build_environment by default?

That is a different argument. Check the ``self.run(..., env="buildenv")`` one.

solvingj:
> Check the self.run(..., env="buildenv") one.

What does it mean?  Do recipe authors have to do `something` to make build requires work?  Can they do something in the recipe like do `self.run(..., env=None)` which will cause `build_requires` not to work in that command invocation?  If so, then it's no good.

The recipe author must not have the ability to outright deny the caller from using `build_requires`.

memsharded:
They have to do nothing, that is the default. Yes, they can change it, for example if for some reason they customize the ``VirtualEnv`` generator and generate a different environment script

> The recipe author must not have the ability to outright deny the caller from using build_requires.

This is not something new. They already have it, it is called ``apply_env``, and opts-out for activating the environment. It was requested by users to be able to opt-out specifically in some recipes: https://docs.conan.io/en/latest/reference/conanfile/attributes.html#apply-env

solvingj:
> They have to do nothing, that is the default. 
Awesome, thank you!

> This is not something new. They already have it, it is called apply_env, and opts-out for activating the environment. It was requested by users to be able to opt-out specifically in some recipes:

I see, I was not aware of that.  In that case, I can agree that there was a reasonable argument for this feature prior to `[conf]` .  With the caller now having the ability to control this value on a per-recipe basis with `[conf]`, this feature seems unnecessary and problematic.  Maybe it will be used as precedent to grandfather this kind of behavior in, but I would consider that a mistake.

ytimenkov:
I'm not sure I understand "caller" term but Jerry's comments made me thinking:

The problem with allowing recipe authors to opt-in may lead to troubles because you don't really have control over final environment or build requirements: it's enough that they come from profile and it can be tuned to meet particular environment or constraints.

What works initially may break in some circumstances.

For example not having `run_environment=True` in the build methods forced me to patch every recipe because we have compiler as a build requirement (which I'm really grateful Conan allows) but it requires proper `PATH`, `LD_LIBRARY_PATH`, `CXX` and other variables properly set. So default `self.run()` didn't work in most cases.

Another example is the need to set `INCLUDES` / `LIBRARY_PATH` in the "local" profile to make gcc accept for Debian's layout.

memsharded:
> What is the difference between buildenv_info and runenv_info in the package_info()?

``buildenv_info`` is mapped to a ``buildenv.sh|bat`` script in the consumer, which is now used by default. This is approximately the current ``env`` model, to apply for building consumers, which works in the build context. This would typically contain things like ``CC``, ``CXX``, ``CXXFLAGS`` and many other toolchain necessary environment to build.

``runenv_info`` is information you need strictly for running that thing in the host context and mapped to ``runenv.sh|bat``. This can be opted-in in recipes, for example for running tests in the host context without pollution from the build context, can be possible now with it. This is the one that can be automatically appended with information from ``cpp_info.libdirs``, etc. to define ``PATH``, ``LD_LIBRARY_PATH``, etc. This is very likely less common, but I have seen libraries requiring some env-vars to be defined in runtime to locate resources or other stuff.

>How it manages a conflicting env-var between different packages providing the same env-var name?

The ``xxxenv_info`` and profile ``xxxenv`` info are ``Environment`` objects, that contain a well defined logic for all possible operations: define, append, prepend, unset. These operations also compose in order. If two packages contain the same env-var, they will be composed in the right order (direct dependencies have higher priority, are composed last), and that will handle in the defined way the env-vars.

memsharded:
> The problem with allowing recipe authors to opt-in may lead to troubles because you don't really have control over final environment or build requirements: it's enough that they come from profile and it can be tuned to meet particular environment or constraints.

@ytimenkov I have already changed the PR to be opt-out, that means, recipes by default will use the ``buildenv`` of their build-requires

> For example not having run_environment=True in the build methods forced me to patch every recipe because we have compiler as a build requirement (which I'm really grateful Conan allows) but it requires proper PATH, LD_LIBRARY_PATH, CXX and other variables properly set. So default self.run() didn't work in most cases.

I think that might be still the case with the approach, because it requires to explicitly switch to the ``runenv`` environment. I would say that a recipe that wants to be used as a build-require should explicitly define PATH, LD_LIBRARY_PATH, etc, while the automatic ``runenv`` for runtime enviroment (not buildtime environment), should probably affect only to regular libraries, not build_requires

ytimenkov:
@memsharded Your last 2 comments made things much clearer to me about the idea of build / run env separation.

1. I think the case of build requirement being a tool is fairly common (what else do you usually put there? compiler, code generator, cmake?)? 
2. I really liked that Conan took care of handling of `PATH` and other variables in a nice and cross-platform way (and that macos `DYLD_LIBRARY_PATH` prepending...). Made a lot of things much simpler.

Given the above, would it be possible to provide some way to tell that particular package is normally a build requirement but its runtime information should be put into build_env?

(I didn't read the latest version, so sorry if this has been already addressed).

memsharded:
> Given the above, would it be possible to provide some way to tell that particular package is normally a build requirement but its runtime information should be put into build_env?

You mean, lets say we have the ``cmake`` recipe, and its conanfile.py:
```python
def package_info(self):
     self.buildenv_info.append("PATH", self.package_folder..)
     self.buildenv_info.append("LD_LIBRARY_PATH", self.package_folder..)

     # self.runenv_info not defined, as this is a build_require shouldn't be in runtime.
```

Isn't that what you want? It is important to see that such LD_LIBRARY_PATH will be used only when building, but not when running. What would be the UI that you would like to see in the recipes?

ytimenkov:
I would like to have effect of this, but with less verbosity. The example above is only linux. Then on winodws it's library dirs need to be added to `PATH` and on MacOSX it should be `DYLD_LIBRARY_PATH`. This every recipes author needs to write.

Then things get more complicated. Say, CMake requires OpenSSL which is also a Conan package and built as a shared library. How will it's paths end up in the build env? For OpenSSL `LD_LIBRARY_PATH` is clearly a run env. And again, whether OpenSSL in this case is a shared or a static library - out of control of CMake's recipe author.

That's why I was thinking about simply marking such CMake package as a kind of "build tool" and thus having it's complete run env being in build env when it's a build requirement. 🤷‍♂️.

memsharded:
> Then things get more complicated. Say, CMake requires OpenSSL which is also a Conan package and built as a shared library. How will it's paths end up in the build env? For OpenSSL LD_LIBRARY_PATH is clearly a run env. And again, whether OpenSSL in this case is a shared or a static library - out of control of CMake's recipe author.

That is already there! :)

```python
def build_environment(self):
        """ collects the buildtime information from dependencies. This is the typical use case
        of build_requires defining information for consumers
        """
        # Visit all dependencies
        deps_env = Environment()
        # TODO: The visitor of dependencies needs to be implemented correctly
        # TODO: The environment should probably be composed with direct dependencies first
        # TODO: in paths, but this is the opposite
        for dep in self._conanfile.dependencies.all:
            # environ_info is always "build"
            dep_env = dep.buildenv_info
            if dep_env is not None:
                deps_env.compose(dep_env)
            if dep.context == CONTEXT_BUILD:
                runenv = self._runenv_from_cpp_info(dep)
                deps_env.compose(runenv)

        # The profile environment has precedence, applied last
        profile_env = self._conanfile.buildenv
        deps_env.compose(profile_env)
        return deps_env
```
Transitive dependencies for cmake, like openssl will automatically add their ``runenv`` be added to the ``buildenv``.

ytimenkov:
> That is already there! :)

You should have started with this! (I had suspicions when was reading the code, but wanted to double-check)😊

In other words there is no need to do any of bin/library path handling in a tool recipe ("cmake")  at all, just use conventional paths and bin_dir -> `PATH` propagation.

jgsogo:
I still don't understand the difference between `buildenv_info`  and `runenv_info` (1).

IMHO, every build-require can be used used as a tool for its own _host_ context: CMake, compilers, code generators,... everything that can be a build-requires will be a tool in its own host too. The environment I need while using `protoc` or `cmake` as a build-requires is the same environment I need when I want to run it from the command-line.

As a build-requires:

```python
class MyPackage:
     build_requires = "cmake/3.16"
     
     def build(self):
         self.run('cmake ....', ...)
```

but it is also a package I would like to work with using some `runenv.sh|bat` generated by Conan:

```
conan install cmake/3.16 --profile:host=this-machine --generator <some-virtual-runevn>
./runevn.sh && cmake ...
```

In my mind, it translates to recipes having all the _build/run_env_ information duplicated like:

```python
class Recipe:
    name = 'cmake'

    def package_info(self):
         self.buildenv_info.append("PATH", self.package_folder..)
         self.buildenv_info.append("LD_LIBRARY_PATH", self.package_folder..)

         # Exactly the same for runenv_info
         self.runenv_info.append("PATH", self.package_folder..)
         self.runenv_info.append("LD_LIBRARY_PATH", self.package_folder..)
``` 

I'm sure I'm missing something 🤷  , maybe a typical example with mocked recipes for AndroidNDK, protobuf, CMake,... could help me to understand how the information is divided and used.

---
(1) I've only heard about one **exception** (that I cannot remember right now) where it would be useful to have `buildenv_info` to propagate (as a built-in feature) some information from a recipe in the _host_ context... but `user_info` was enough to overcome that situation (and I'm not sure if the recipe was _correct_ or it was just legacy).

memsharded:
> I'm sure I'm missing something 🤷 , maybe a typical example with mocked recipes for AndroidNDK, protobuf, CMake,... could help me to understand how the information is divided and used.

There are tests in the PR like this https://github.com/conan-io/conan/pull/8534/files#diff-3a4e0a05c2bdf172f5ec0214e40a6b9a00dff75838ed1111a1a5be6f3cc70fac, that are using mock recipes of CMake with transitive requirement to OpenSSL. There is also one real test that is using actual shared libraries in https://github.com/conan-io/conan/pull/8534/files#diff-7c521be4b7aaef4d9b562408819d40e913482764e81ba13efbdf4939792424cf.

The solution is as follows:

- In many cases, it is not necessary to implement the PATH and LD_LIBRARY_PATH vars. The definition from ``cpp_info`` is enough to deduce it and apply it correctly to the right script either ``conanbuildenv`` or ``conanrunenv``
- For the cases where the user wants to add things to the ``self.buildenv_info`` and ``self.runenv_info``, the logic is very straightforward: what is defined in ``self.buildenv_info`` goes to ``conanbuildenv`` and what is defined in ``self.runenv_info`` goes to ``conanrunenv``.
- If you want for some reason to have the same values for both, you can declare an ``e = Environment()``, fill the values and do ``self.buildenv_info = self.runenv_info = e``. No need to repeat things.
- BUT, the most important part here is that in most cases this is not what is expected. ``cmake/3.14`` Conan package binary is not intended to be deployed in production, in the servers, or to be deployed and run in the released application. It is a development tool, intended to serve as a ``build_require`` always.
- The intended usage of ``cmake``, even as developer will be using the build-environment:
  ```
  conan install cmake/3.16 --build-require --profile:build=this-machine --generator <some-virtual-runevn>
  ./buildenv.sh && cmake ...
  ```
- ``runenv`` is always intended to run the final app or tests in the host context.
- This is equivalent to the need of having the ``test_package`` test its package as ``build_require`` and not as a ``require``. Both cases are implemented in this POC: https://github.com/conan-io/conan/pull/8627

With this way, the usage of ``cmake/3.14`` is completely homegeneous for all use cases: build-required from recipe, build-require from test_package, build-required from the command line.

SSE4:
@memsharded this PR is getting large, it changes lots of things altogether, may we split it into several iterations for easier review?
e.g. I don't think we should add `CMakeGen` right away (at least in the current state), but `Environment` class is pretty solid already

memsharded:
> e.g. I don't think we should add CMakeGen right away 

I am fine of letting the ``CMakeGen`` class out, but it was added to this PR, it was not originally there, as a result of reviews and feedback from @lasote, but it is true that at that moment the ``VirtualEnv`` was opt-in and not opt-out, so lets see what @lasote suggests.

lasote:
If now it is not that clear, better to remove it. It can (and I bet it will) be added anytime.

SSE4:
> I am fine of letting the `CMakeGen` class out, but it was added to this PR, it was not originally there, as a result of reviews and feedback from @lasote, but it is true that at that moment the `VirtualEnv` was opt-in and not opt-out, so lets see what @lasote suggests.

seems like he agreed, but I actually meant to split it e.g. into 5 independent PRs (`CMakeGen` I've used as just an example which is clearly thing on its own):
- Environment
- ProfileEnvironment
- VirtualEnv
- ConanfileDependencies
- CMakeGen

(maybe some of them don't make sense on their own, please comment). right now it's too hard to agree on everything at one shot.

memsharded:
As a reference, potential recipes in conan-center that needs a ``conanrunenv`` because they contain resources that are necessary at consumer application run time:

- ``self.env_info.DCMDICTPATH = dcmdictpath`` in ``dcmtk``
- ``self.env_info.FONTCONFIG_FILE = fontconfig_file`` from ``fontconfig`` (also ``FONTCONFIG_PATH``)
- ``self.env_info.FT2_CONFIG = freetype_config`` from ``freetype``
- ``self.env_info.GDAL_DATA = gdal_data_path`` from ``gdal``
- ``self.env_info.GDK_PIXBUF_PIXDATA = os.path.join(self.package_folder, 'bin', 'gdk-pixbuf-pixdata')`` from ``gdk-pixbuf``
- ``self.env_info.GLIB_COMPILE_SCHEMAS = os.path.join`` from ``glib``
- ``self.env_info.GST_PLUGIN_SCANNER`` (and many others) from ``gstreamer``
- ``self.env_info.ICU_DATA.append(data_path)`` from ``icu``
- ``self.env_info.ALSA_CONFIG_DIR = os.path.join(self.package_folder, "res", "alsa")`` from ``libalsa``
- ``self.env_info.MARIADB_PLUGIN_DIR = plugin_dir`` from ``mariadb-connector-c``
- ``self.env_info.OATPP_SWAGGER_RES_PATH = res_path`` from ``oatp-swagger``
- ``self.env_info.POPPLER_DATADIR = datadir`` from ``poppler``
- ``self.env_info.PROJ_LIB.append(res_path)`` from ``proj``
- ``self.env_info.VK_LAYER_PATH.append(vk_layer_path)`` from ``vulkan-validationlayers``

This seems a compelling use case for ``self.runenv_info`` in recipes, so I am keeping it.

jgsogo:
From what I understand (the part related to build/runenv declared in the recipes), this is the proposal: recipes can declare `runenv_info` and `buildenv_info` in the `package_info()` function:

```python

class Recipe(ConanFile):
    name = 'gcc'

    requires = 'zlib', 'openssl'

    def package_info(self):
        self.cpp_info.bindirs = ['path/to/gcc']
        # self.env_info... would be superseded by the following

        self.runenv_info.RUNINFO = ['something']

        self.buildenv_info.LD_FLAGS = ['flags-from-gcc']

```

Let's define _build info_ and _run info_ as follows (I will use them below):
 * we call _**build info**_ of a given recipe to the PATHS inferred from `cpp_info` and the information coming from `buildenv_info`.
 * we call _**run info**_ of a given recipe to the PATHS inferred from `cpp_info` plus all the information coming from `runenv_info`. 

Now here it comes the (my) example: take it like pseudo-code
 * when building a package:

   ```python
   class Recipe:
      name = 'app'

      requires = 'library'

      def build_requirements(self):
         self.build_requires('gtest', force_host_context=True)
         self.build_requires('gcc')

      def build(self):
         with self.activate_build_environment(...):
            self.run('gcc ....')
         
          with self.activate_run_environment(...):
            self.run('gtest ...')
   ```
   
   The call to `activate_build_environment` will populate the environment with:
      * the **build info** from `gcc`  plus the **run info** from `zlib` and `openssl`.

   The call to `activate_run_environment` will populate the environment with:
      * the **run info** from `library` and the **run info** from `gtest` (and also all the **run info** from all the requirements in the _host_ context).

   So far this is what we have right now.

 * BUT, there is a different way to consume `gcc`. We may want to use it like _standalone_ application and activate its environment, something like:
 
    ```
    conan gimme gcc/10@ ... --generator <run-environment>
    ```
    
    And it will provide us with a `runenv.sh/bat` file containing the **run info** from `gcc` plus the **run info** from `zlib` and `openssl`.

    Alternatively, I can ask Conan to generate the build environment like:
    
    ```
    conan gimme gcc/10@ ... --generator <build-environment>
    ```
    
    And it will provide us with a `buildenv.sh/bat` file containing the **build info** from `gcc`  plus the **run info** from `zlib` and `openssl`.
    
Of course, I've invented names and commands, but so far I think we can understand the intention of all of it. Is this correct? Does it reflect the purpose of this PR? If not, stop reading here.

---

Now, these are my comments: with the proposed model we are able to provide two slightly different context to run `gcc`. One of them contains the `LD_FLAGS` (we need them to build our packages) but not `RUNINFO` while the other contains `RUNINFO` but not `LD_FLAGS`. This adds some functionality but I'm questioning if any recipe will ever use this functionality (while being something we can recommend as the way to write a good Conan recipe).

My point is that there is no such a thing as **build info**, but many different **build info**s depending on the settings targets, so the recipes will _always_ be as follows:

```python
class Gcc(ConanFile):
    name = 'gcc'

    def package_info(self):
        self.cpp_info.bindirs = ['path/to/gcc']
        # self.env_info... would be superseded by the following
        
        self.runenv_info.RUNINFO = ['something']
        
        if self.settings_target.xxxx = 'yyyyy'
            self.buildenv_info.LD_FLAGS = ['flags-from-gcc-for-yyy']
        elif self.settings_target.xxxx = 'zzzzz'
            self.buildenv_info.LD_FLAGS = ['flags-from-gcc-for-zzz']
        elif self.settings_target.xxxx = 'ssss'
            self.buildenv_info.LD_FLAGS = ['flags-from-gcc-for-ssss']
        else:
            pass
```

This affects our scenario where we wanted to use `gcc` as a standalone application and we wanted Conan to generate the build environment. Now we can optionally provide a `profile:target`:


```
conan gimme gcc/10@ ... --generator <build-environment> --profile:target=some-target-profile
```
    
And it will provide us with a `buildenv.sh/bat` file containing the **build info** from `gcc`  plus the **run info** from `zlib` and `openssl`. And **this is great, this is something we've been asking for long**. And also, typically, **if no `profile:target` is provided it won't add anything to `buildenv_info` (like AndroidNDK)** and others cross-compilers that are being used right now.

Up to now, we have just renamed `env_info` to `buildenv_info`. Nothing new.

---

What about **self.runenv_info**?

This is what I'm missing here: We need to figure out and find a recipe where we want some environment variable to be present while we run the package as standalone, **but we don't want that variable while running that package to build (inside an `activate_build_environment` or like a standalone application)**. Some recipe where we want:

```
conan gimme gcc/10@ ... --generator <build-environment>  # without a profile:target
conan gimme gcc/10@ ... --generator <run-environment>
```

to generate different environments. 

---

For those that think that this is something we need, please, can you provide some sound examples? What could I write into `runenv_info` that cannot go into `buildenv_info` (when no `profile:target` is provided)?   @memsharded @ytimenkov @lasote @solvingj 



---

##### ⚠️ ⚠️ ⚠️  Of course, my suggestion would be to keep using the name `env_info` to avoid the confusion that renaming it to `[build|run]env_info` introduces.

solvingj:
Maybe think about the question in the other direction.  When we produce a `run_env` file, we don't want to leak all the build environment variables into it. It should be strictly the information needed at runtime, as specified by the recipes.  

To clarify, I think there is no question that we had the original data model and abstraction slightly wrong before ( @jgsogo if you disagree, please say so).  But, it seems clear that there are definitely two completely separate categories of information that need to be declared.  Conan currently only has one place to declare them.  If Conan can't model them as two separate things with separate storage, users can never distinguish them in generators and scripts when consuming them. 

I think your key point is that if we don't have bug reports where mixing them together has caused actual problems, then maybe there's not a strong enough argument for a breaking change (or for making this a priority). That is certainly a reasonable argument. If that's the case, it's just a subjective matter of choosing whether or not "correctness" in this area is important enough to warrant prioritization above other outstanding items, and making a breaking change. 

In terms of necessity/priority, I would label this very high.  That's even with the understanding that we already have an overwhelming number of things on the roadmap.  But, as a build engineer I have personally been impacted by the mixing of these two things into one category.  When managing hundreds of packages, you have to configure a LOT of environment variables over a lot of profiles and recipes, coming from all sorts of hard-to-find documentation and code.  I was super happy Conan gave me first class features such as `[env]` in profiles and `env_info` in recipes.  But, after you set things up, you quickly forget the purpose of each variable, and it's usually super difficult to figure out where it comes from and where it's consumed (build or run). This is especially true with cross-compilation tools, makefiles, and lots of wrapper scripts in multiple script languages.  When debugging a build issue or runtime/test issue that comes out of Conan, I was double and triple checking environment variables a lot, and frequently wondering if something is even needed for the build or not.   The separation would have done wonders during debugging, and also for training new people, who often don't even understand how the environment variables work in the first place.  

I think it's pretty hard to over-value correctness for the fundamental data model.

ytimenkov:
I really like the summarization of how build_env and run_env are composed, @jgsogo 👍

I don't have good answers, but I was thinking in the same direction as @solvingj: build requirements are fine, but maybe it's ordinary requirements which may need different handling. But it's tricky: if such dependency happens to be a transitive dependency of some build requirement - you'll get its variables back. Maybe they won't even be consistent (can build context has different version?).

Another thought is that probably 90% of cases are nicely handled by Conans abstractions: bin/lib dirs, cppflags, etc. so the real need for custom environment variabels is quite minimal (so I guess your example is not very correct - `LD_FLAGS` should be a result of link_flags passed through AutotoolsToolchain or similar).

One example I can give of gcc being consumed both by build and run - I need gcov to process coverage data when I run my package, and gcov is a part of gcc package (maybe it shouldn't be 🤷‍♂️).

Written the second paragraph I begin to doubt myself whether a separate build_env (in recipe) is needed... There are also build modules (now per generator)....

solvingj:
@ytimenkov Just remember, if you have both, you can easily combine, can even create a "all_env" wrapper or whatever if we wanted.  However, if they're mixed together, separating them is impossible.  Since we can't predict every use case and interaction between the two environments, it seems strictly safer to separate them at the declaration status.

ytimenkov:
I don't disagree. I doubt 😊

I think the most important part of this pr is fusion of virtualrunenv (autorunenv) with information provided by recipe (env or runenv), because now I have to combine them by specifying virtualenv and virtualrunenv all the time. Having one file to activate world be much simpler.

memsharded:
There is one premise that is not the current design: to run ``gcc`` as a standalone application, defined as normal "host" package.
The ``gcc`` package is always a package used for building, not for deploying in your server to run your application. Unless you plan to do that, it is a ``build-require`` and will generate ``conanbuildenv`` scripts, not ``runenv`` scripts.

The existence and necessity of 2 different environments, one for building and one for running is evident from transitive requires: 
- If you have cmake as build-require that depends on both openssl/1.0 and zlib/1.0 as shared libraries
- And the package application executable depends on openssl/2.0 and zlib/2.0 shared libraries.

Then, when building, you don't want openssl/2.0 and zlib/2.0 in LD_LIBRARY_PATH, but you want openssl/1.0 and zlib/1.0, and viceversa, when you are running your application tests, you certainly want openssl/2.0 and zlib/2.0 in LD_LIBRARY_PATH.

So we need 2 different environment, one for build time and another for run/execution time.

The case for ``runenv`` scripts and ``runenv_info`` are the ones exposed above: https://github.com/conan-io/conan/pull/8534#issuecomment-796637499. When you need some environment to run your final application. The ``myapp`` executable that you are deploying to your servers. 

The separation of the different environments is very clear, there are env-vars that need to be defined only at build time (LDFLAGS), and other env-vars that need to be defined only at run time (https://github.com/conan-io/conan/pull/8534#issuecomment-796637499), if there are 2 different output environment, there is simply no point in having just one single definition of an ``env_info`` that will be applied both at build time and runtime.

memsharded:
@ytimenkov 

> I think the most important part of this pr is fusion of virtualrunenv (autorunenv) with information provided by recipe (env or runenv), because now I have to combine them by specifying virtualenv and virtualrunenv all the time. Having one file to activate world be much simpler.

The new proposed ``VirtualEnv`` is doing the necessary aggregation of env definitions of the build requires, so everything needed to use your build-requires will be contained in ``conanbuildenv`` script. Including all the ``virtualrunenv`` of your build-requires, like PATH, LD_LIBRARY_PATH, etc. The new ``conanrunenv`` is a new script that should be used for running executables, but it should not be aggregated to the ``conanbuildenv``, it is intended to be different and independent.

SSE4:
the same library (and therefore recipe) could be consumed as regular requires (let's just call it requires) and build-requires.
if I work on some library (let it be `protobuf`) I may have:
- build tools, such as `protoc.exe`
- libraries, such as `libprotobuf.dll`
- unit tests, such as `protobuf_unit_tests.exe`

in general way, I could be cross compiling, and running tests in some sort of simulator (not necessarily, e.g. for x86_64 -> x86 I can just run), but let's assume for now I am using x86 build machine and arm as a host device to deploy on.

this could be described, more or less, with `--profile:build x86 --profile:host arm`, at least in two profiles model, which is de-facto standard right now, and I guess everyone understands what are they.
in example x86 profile simply defines `settings.arch= x86` and arm profile defines `settings.arch= arm`.

my recipe for library I develop could look like:
```
def build_requirements(self):
    self.build_requires("protobuf/1.0")  # get protobuf to run protoc.exe

def requirements(self):
    self.requires("protobuf/1.0") # get protobuf to link with

def build(self):
    with activate_build_environment(self, "protobuf"):
        # I want to use build tools here (from the build profile), so I activate build environment for the build profile
        self.run("protoc.exe somefile.proto")

def test(self):
    with activate_run_environment(self, "protobuf"):
        # I want to use compiled executables for host device, so I activate run environment for the host profile
        self.run("myarmdevicesimulator.exe protobuf_unit_tests.exe")  # I want to run tests to ensure it still works on my platform
        # myarmdevicesimulator.exe something I have installed in the system (e.g. from Android Studio), not managed by conan
```

so, right now we have at least:
- `activate_build_environment`, to set up build environment for the build profile
- `activate_run_environment`, to set up run environment for the host profile

I believe, in order to make them actually work, we should have in model:
- `conanfile.build_env_info` - build environment for the build profile
- `conanfile.run_env_info` - run environment for the host profile

the same could be achieved with command line:
```
# I want to use build tools here (from the build profile), so I activate build environment for the build profile
# it launches x86 executable (build profile)
$ conan install . --profile:build x86 --profile:host arm -g virtualbuildenv && ./activate_build.sh protoc.exe somefile.proto

# I want to use compiled executables for host device, so I activate run environment for the host profile
# it launches arm executable (host profile)
$ conan install . --profile:build x86 --profile:host arm -g virtualrunenv && ./activate_run.sh myarmdevicesimulator.exe protobuf_unit_tests.exe
```

so we should have the following:
- `virtualbuildenv` generator, to generate build environment for the build profile from `conanfile.build_env_info`
- `virtualrunenv` generator, to generate run environment for the host profile from `conanfile.run_env_info`
- `activate_build.sh` as an output of `virtualbuildenv` generator
- `activate_run.sh` as an output of `virtualrunenv` generator

I think the example above clearly illustrates the difference and necessity of these two things in the model.

the only thing I would consider is probably renaming `run_env` into `host_env`, to make the mapping between run/host clear and avoid the confusion.

jgsogo:
We already have in Conan many different contexts, it is not that we are mixin all the environment variables together:
 * we have the _host_ context (1)
 * and we have one _build_ context **per package** (N). 

@solvingj we are not leaking build variables to the run environment. If that is your case, please, let me have a look at those conanfiles because they are probably wrong.

I'm not saying that we don't need run_environment and build_environment, what I'm saying is that **we already have build_environment and run_environment**. In this PR we are introducing a third category, that multiplies the number of build_environments by 2, it introduces build_environment1 (with information from `self.buildenv_info`) and build_environment2 (with information from `self.runenv_info`), and then it _artificially_ decides which one is activated in which scenario. If this goes forward all recipes will look like this:

```python

def package_info(self):
    self.runenv_info = ....

    self.buildenv_info = self.runenv_info  # This line will be forced/checked by a HOOK
```

We don't need to differentiate that, it is already working, please read the +30 issues gathered [here](https://github.com/conan-io/conan/projects/4#card-56760401) and their answers: the problem is not related to the number of environments, **we only need to transfer more control to the user and the recipe author**.

Please, I started to work on this model on [May 2019](https://github.com/conan-io/conan/pull/5184), the first PR was merged on [Aug 2019](https://github.com/conan-io/conan/pull/5592), since then I've been answering most of the questions related to this model, in this repo and in `conan-center-index`. We are successfully providing recipes for cross-compilers, for tools, for libraries that also package code generators... and they are working great, they are being used in many different scenarios and they work. I've not started to think about this issue after reading this PR, it is something I'm being working on for almost two years now.

All of you are suggesting hypothetical scenarios, in those scenarios I could ask for a `self.myenv_info` because I want to. **I'm only asking for a real scenario where this is needed, ONLY ONE**. It doesn't need to be a recipe from ConanCenter, just something sound.

Please, focus on the environments, don't get lost with the commands and the use-cases, those are only interfaces for the model.**Given the contexts we already have** we can decide to merge generators, change commands, add arguments,... we can rename files to `runenv.sh`, `buildenv.sh` or `buildenv-for-xxxx-package.sh`. That is something completely different that is not related to the core problem. We **already** can generate all those files with the information we **already** have with the (N+1) contexts we are **already** managing.

ytimenkov:
Q: Can other toolchains (or recipe) contribute to its own build_env?

Asking because we use waf and it has strange handing of source / output dirs and the easiest way to solve this is to set `WAFLOCKFILE` variable which I currently do in my generator by piggybacking on virtualenv generator which I'd like to avoid.
Would be much better if I could specify something like `self.build_env.WAFLOCKFILE=f".lock-{md5(self.install_folder}"` somewhere in toolchain and this could be used by stock VirtualEnv generator.

(Actually this is a shortcut for `with tools.append_environment()` in the recipe, but in the same spirit: remove the extra typing for developer when working on the code)

Maybe other tools could also benefit from such feature.

ytimenkov:
@SSE4 Unfortunately right now you can't have same package in build and host requirement. See https://github.com/conan-io/conan-center-index/issues/4013 And also https://github.com/conan-io/conan/issues/8335

jgsogo:
> @SSE4 Unfortunately right now you can't have same package in build and host requirement. See [conan-io/conan-center-index#4013](https://github.com/conan-io/conan-center-index/issues/4013) And also #8335

Yes, you can, and you can also have different versions of the same package and they won't raise any conflict. You only need to use the approach with two-profiles (`--profile:host` and `--profile:build` in the command line)... which is not the default in Conan 1.x because it would be breaking.

(I'm moving to that linked issue)

Psy-Kai:
As @jgsogo said having `self.runenv_info`and `self.buildenv_info` seems a little bit too much. Today settings `self.env_info` in a package consumed as build requirement puts the env vars in a build environment. Consuming that package as "normal" requirement puts the vars in the runtime environment. The only problem is that today runtime environment variables get lost in cross compile build.  
What I welcome is an easy way to activate build and runtime environment. In Conan 1.x I am not able to use (build) env vars set in a profile in the `generate` function (I wanted to pass some compiler flags that way). I needed to create a package just propagating the flags. And since the env vars were not merged properly I needed to do it manually. So having that feature build in is a huge improvement! But I don't see why I would need to pack different `self.buildenv_info` and `self.runenv_info`.

lasote:
For me is very important that everyone understand that we are building now the foundations of the Conan 2.0 and only a complete model will allow us to build on top of it the N layers to complete a successful Conan 2.0. So my vote is a big YES to merge this. 
If later we realize that we want to revert it, we can perfectly do it, it is just an addition that it is not breaking anything in the `Conan 1.X`. I think it doesn't make any sense to keep this blocked and discussing if we can live without it when the real question is: Is this a better model? and the answer is, yes, it represents better the reality than the previous one. The rest of the considerations are just UX.

jgsogo:
No, it doesn't represent the reality better. It is not about UX. It is about the model. It is adding an artificial environment that is not needed at all (we are waiting for an example, we have +750 recipes in ConanCenter and no one of them requires it). Without a minimal example, it is as arbitrary as adding `self.sandboxenv_info` or `self.testenv_info`.

lasote:
> Without a minimal example, it is as arbitrary as adding self.sandboxenv_info or self.testenv_info.

Multiple examples of variables that should exist at `run` but not on `build` already provided [here](https://github.com/conan-io/conan/pull/8534#issuecomment-796637499)

>No, it doesn't represent the reality better. It is not about UX. It is about the model. It is adding an artificial environment that is not needed at all (we are waiting for an example, we have +750 recipes in ConanCenter and no one of them requires it). 

"Not needed" is different from "useless". Can the recipes work without this feature, sure! they work! (even with vcpkg :P ) Can these variables exist at build time? sure, they can! but _they shouldn't_. That's what I mean by a better model.

jgsogo:
> > Without a minimal example, it is as arbitrary as adding self.sandboxenv_info or self.testenv_info.
> 
> Multiple examples of variables that should exist at `run` but not on `build` already provided [here](https://github.com/conan-io/conan/pull/8534#issuecomment-796637499)
> 


All those variables are needed when running as a build-requires and when running as _standalone_ application. Why do you think you don't need those variables when using those packages as build-requires?

lasote:
> Why do you think you don't need those variables when using those packages as build-requires?

I don't think that, a better model that doesn't mix things up allows to manage it nicely:

https://github.com/conan-io/conan/pull/8534#issuecomment-797103642

https://github.com/conan-io/conan/pull/8534/files#diff-5a2b3f42b8880cba4179072199deb461111c266409a3ce051e22e3bb1816084fR30-R32

Psy-Kai:
@lasote do you talk about the need of something like
```python
def package_info(self):
    self.buildenv_info.FOO = "Bar"
    self.runenv_info.HELLO = "World"
```
or something other? Because setting a `runenv_info` for a build-requires seems very odd (and vice versa). How would you explain the user that a build-requires package leaks information into the run-context? And how does a "normal" requires package leaks info into the build-context? That is very hard to explain. Mixing this things will lead to more complex package definitions.

Setting a single `self.env_info` and let the (build-)requires context decide how to propagate makes it simpler to understand and to use. And I cannot imagine an example where you would like to leak a build-context env variable into the run-context and vice versa. If the user want to mix these variables he sure can do manually in his conanfile (or by extracting them from conanbuildinfo.json). But I think that this will be a very very rare case.

SSE4:
@jgsogo to be honest, I don't think `ICU_DATA` and friends are good examples.
first off, `ICU_DATA` is used as by executables, as by libraries, so it should be always set, basically.
second, it's needed in both contexts, as I may run x86 (build) ICU binaries:
```
$ conan install . --profile:build x86 --profile:host arm -g virtualbuildenv && ./activate_build.sh icurun <args>
```
but at the same time I may run ARM (host) binaries:
```
$ conan install . --profile:build x86 --profile:host arm -g virtualrunenv && ./activate_run.sh <myemu> icurun <args>
```
I need `ICU_DATA` to be set in both cases, for certain.
meanwhile, if I have in ICU conanfile something like:
```
    self.buildenv_info.ICU_DATA= "Foo"
    self.runenv_info.ICU_DATA= "Bar"
```
I don't see how it helps, it only complicates the things, introducing additional challenges in UX.
now I not only have build and host (run) environments, but I need to distinguish if I want to use something as build tool, or just regular run (there might be some conceptual or ideological differences, but still, both are just launching executables in the end).
so I technically can have 4 situations now:

- I want to **BUILD** with icurun in **BUILD** context  (launch x86 icurun binaries, but as a part of build process or as a build tool)
- I want to **RUN** icurun in **BUILD**context (launch x86 icurun binaries, for other purposes)
- I want to **BUILD** with icurun in **HOST** context  (launch ARM icurun binaries, but as a part of build process or as a build tool)
- I want to **RUN** icurun in **HOST** context  (launch ARM icurun binaries, for other purposes)

IMO this is too much and super confusing.

memsharded:
@Psy-Kai 

There are recipes that declare 2 different types of environment variables, and there are examples of both usages in ConanCenter. I am not even talking about build-requires, just a regular recipe containing a library as regular requires, in the “host” context:

Environment variables for build-time, most common example is XXX_ROOT

```python
self.env_info.MYLIB_ROOT = self.package_folder.replace("\\", "/")
```
This one is an environment variable for building against it, more specifically for CMake find_package functionality to find it: https://cmake.org/cmake/help/latest/envvar/PackageName_ROOT.html. There are a few packages in ConanCenter using this approach.

This variable must exist in the ``conanbuildenv`` environment script, so CMake finds it when building.

On the other hand, there are env-vars for runtime, typically something like:

```python
self.env_info.MYLIB_DATADIR = os.path.join(bindir, "datadir")
```
This is also something that can be found in ConanCenter recipes. This is the location needed at application runtime, to locate the necessary resources to run applications linking with this library. This information must exist in the ``conanrunenv`` environment script, so when the application is run, it will find that data directory at runtime.

The necessary representation for this is something like:

```python
self.buildenv_info.MYLIB_ROOT = self.package_folder.replace("\\", "/")
self.runenv_info.MYLIB_DATADIR = os.path.join(bindir, "datadir")
```

Otherwise it will be impossible to distinguish the right target environment script for each one.

memsharded:
Seeing all the info, I have decided to move the PR forward as-is, there are a few pieces here that we need to keep moving forward. Even if it adds extra-model, it seems to exist real use cases that requires this flexibility, so I am keeping it. This part will not be documented at this moment, and we will keep evolving and investigating what is necessary and what not as we evolve the graph model. If we realize later that a single definition of environment is good enough, we will revert it.

Thanks all for the feedback!

ytimenkov:
Is it possible to keep all 3 instead of manual combination? `env_info`, `runenv_info` and `buildenv_info` In the recipe's `package_info()` and then use `env_info + runenv_info` or `env_info + buildenv_info` when generating? This will allow recipes to opt-in into differentiated behavior where needed so we can see later the demand for this feature.
If this is also used only with new generators there shouldn't be compatibility issues.

jgsogo:
Thanks for the example. Now I understand the motivation after that proposal. This is something people have been doing with `self.user_info`: the consumer knows the libraries it requires and decides to use (or not) the variables provided by its requirements in _host_ via `self.user_info_host` (legacy `self.deps_user_info`).

**We can change our minds and start recommending this approach with environment variables, no problem.** 👍 

---

Questions in order to help users in future issues:
 * When the same variable comes from _host_ context and _build_ context, which one takes precedence? How it affects any future approach based on wrappers/shims to execute build-requires?

 * What is the plan for these _new_ environment variables coming from _host_? Are they available before or after evaluating the method `package_info()` of the build-requires?

 * From my comment [here](https://github.com/conan-io/conan/pull/8534#issuecomment-796978654), if I understand it correctly, the environment gathered from a build-requires is the `self.buildenv_info` from the build-requires plus the `self.runenv_info` from the requirements of the build-requires. Can a consumer decide that it wants the `self.buildenv_info` from the requirements of the build-requires?
 
   Example: a user that implements an SDK with a package that requires other tools like `cmake`, `pkg-config`, `AndroidNDK`,... those would be requires of the `sdk-package`. For sure, when that user declares in a consumer that he build-requires that `sdk-package`, he will need the `self.buildenv_info` from all of them. Is this by default and my comment was wrong? Is there a way to opt-in for this behavior?

   Note.- This is the scenario for the `emsdk` recipe (with `python` and `nodejs` packages).

And then, there are lots of details, but for me at the moment, these are some important points that will help me to make up my mind. Thanks

PR #8537, opened 2021-02-21
Rename QbsToolchain to QbsProfile

Psy-Kai:
This generator generates a qbs profile. Renaming the class from Toolchain
to Profile makes this more obvious.

Changelog: Fix: Renamed generator `QbsToolchain` to `QbsProfile`.
Changelog: Fix: Renamed default filename of _QbsProfile_ generated file to _conan_toolchain_profile_.qbs.
Changelog: Fix: Renamed Qbs attribute `use_toolchain_profile` to `profile`.
Docs: https://github.com/conan-io/docs/pull/2027

### DIFF
--- a/conan/tools/qbs/__init__.py
+++ b/conan/tools/qbs/__init__.py
@@ -1,2 +1,3 @@
-from conan.tools.qbs.qbstoolchain import QbsToolchain
+from conan.tools.qbs.qbsprofile import QbsProfile
+from conan.tools.qbs.qbsprofile import QbsProfile as QbsToolchain
 from conan.tools.qbs.qbs import Qbs
--- a/conan/tools/qbs/qbs.py
+++ b/conan/tools/qbs/qbs.py
@@ -21,7 +21,7 @@ def _configuration_dict_to_commandlist(name, config_dict):
 class Qbs(object):
     def __init__(self, conanfile, project_file=None):
         # hardcoded name, see qbs toolchain
-        self.use_profile = 'conan_toolchain_profile'
+        self.profile = 'conan_toolchain_profile'
         self._conanfile = conanfile
         self._set_project_file(project_file)
         self.jobs = tools.cpu_count()
@@ -52,8 +52,8 @@ class Qbs(object):
 
         args.extend(['--jobs', '%s' % self.jobs])
 
-        if self.use_profile:
-            args.append('profile:%s' % self.use_profile)
+        if self.profile:
+            args.append('profile:%s' % self.profile)
 
         for name in self._configuration:
             config = self._configuration[name]
@@ -72,8 +72,8 @@ class Qbs(object):
 
         args.extend(['--jobs', '%s' % self.jobs])
 
-        if self.use_profile:
-            args.append('profile:%s' % self.use_profile)
+        if self.profile:
+            args.append('profile:%s' % self.profile)
 
         for name in self._configuration:
             config = self._configuration[name]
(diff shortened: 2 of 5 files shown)


memsharded (approved):
I am fine with this change, mainly because you are Qbs maintainers. It is very likely that I would reject this PR otherwise. Even if the feature is experimental, we try to minimize breaking changes. Renaming or other "cosmetic" changes are typically not enough to break.

In this case I trust you that Qbs users will not be very annoyed for this break, but please consider it for future changes.

Many thanks for keeping maintaining it!

Psy-Kai:
Thanks for making that exception :)
I tried to not break the things for now (thats why I let the `from conan.tools.qbs.qbsprofile import QbsProfile as QbsToolchain` in there) so other users have time to migrate.

memsharded:
> I tried to not break the things for now (thats why I let the from conan.tools.qbs.qbsprofile import QbsProfile as QbsToolchain in there) so other users have time to migrate.

Oh, that is true, good point, I skipped that part, well done. It is also possible to create a temporary alias ``QbsToolchain`` that print warnings, but not necessary, as it was released recently and the documentation change, it is fine. Thanks!

memsharded (commented):
It seems this change didn't include the renaming of the text generator:

```python
generators = "QbsToolchain"
```
Would you like the generator for the Qbs toolchain to be called QbsProfile?

Psy-Kai:
Oh yes, it seems like I forgot something. Do you mean https://github.com/conan-io/conan/blob/052a68f891702bcb616724983ca3f61593b4b29b/conans/client/generators/__init__.py#L110? Or is there another place I forgot to change?

memsharded:
Hi @Psy-Kai 

Yes, that is the place.

In any case, I am wondering what is the impact on Conan users. All other build sytems will have a XXXXToolchain which is the responsible of creating any files that the build system needs. In CMake it will be find_package scripts, in MSBuild .props properties files... Having this class named differently for Qbs might probably confuse initially Conan users using other build systems too. 

Please tell me, because we can still add it to Conan 1.34 before it is released (is already branched)

Psy-Kai:
I just thought that the "Toolchain" naming was a little off. For Qbs a [Profile](https://doc.qt.io/qbs/qml-qbslanguageitems-profile.html) is generated. So it would be more self explaining if I write something like `generator = QbsProfile`.  The "Toolchain" may be a little bit confusing for Qbs users because in the Qbs Documentation there is no real mention of such thing.

So my intention was self explaining code. And I thought that this "Toolchain"-Generators will be renamed some time because they are just generators (I found comments in the code hinting such a thing). The BuildInfo generators are not called "cmakebuildinfo" too (even if that would be some step towards more self explaining code).

memsharded:
Ok, I understand. Our current naming is:

- XXXDeps: Generator like CMakeDeps that handle dependencies build information (find_packages)
- XXXToolchain: Generator like CMakeToolchain or AutotoolsToolchain that handle the mapping of Conan settings and options to build system specific , like ``conantoolchain.cmake`` or ``autotoolstoolchain.sh`` shell script with env-vars. Toolchain is not necessarily a "toolchain" file, is any file that the build system could use to define its build configuration
- XXXX: The build helper command line wrapper, like CMake
- XXXXGen (currently under consideration): A high level generator that bundle at least XXXDeps, XXXToolchain and VirtualEnv

I am fine if you want to use the ``QbsProfile`` and it will be easier for your users. Please do a PR asap to the **release/1.34** branch, fixing the above issue (I can manage too if you can't do it now). Thanks!

PR #8547, opened 2021-02-24
POC Python profiles with django settings inspired syntax

danimtb:
Coming from https://github.com/conan-io/conan/pull/7922

See django settings docs: https://docs.djangoproject.com/en/3.1/topics/settings/


conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,164 @@
...
+        if parent_profile_path:
+            ProfileLoader._load_profile(parent_profile_path)
+        profile = ProfileLoader._load_profile(profile_path)
+        ProfileLoader._init_profile()
+        return profile.copy()
+
+
+class PythonProfilesTest(unittest.TestCase):
+
+    def setUp(self):
```
memsharded: Please start using modern pytest syntax with fixtures.

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,164 @@
...
+        self.pr1_content = textwrap.dedent("""
+            import os
+            from conans import settings, options, env
+
+            settings.os = "Linux"
+            settings.build_type = "Release"
+
+            options.shared = True
+
+            env.env_var = "bar"
```
memsharded: Lets aim to use the Environment composition defined in https://github.com/conan-io/conan/pull/8534

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,164 @@
...
+        return profile.copy()
+
+
+class PythonProfilesTest(unittest.TestCase):
+
+    def setUp(self):
+        self.client = TestClient()
+        self.pr1_content = textwrap.dedent("""
+            import os
+            from conans import settings, options, env
```
memsharded: This import of a global object, and the way the object is re-initialized for each profile doesn't look good.

It would be cleaner to do something like:

```python
from conan.profiles import Settings, Options  # To be define where it is, definitely in the new ``conan.`` scope, but not in the root
from conan.tools.env import Environment # the new one

settings = Settings()
settings.os = "Windows"

env = Environment()
env.append_path("PATH", os.getenv("MyLocalPathVar"))

```

conans/test/functional/profiles/test_python_profiles.py
danimtb: moved tests to pytest

conans/__init__.py
```diff
@@ -10,6 +12,64 @@
 from conans.util.files import load
 
 
+class ProfileOptions(defaultdict):
```
memsharded, line 15: These classes cannot be here, this is too global of scope. Better move them to ``conan.profile``

conans/__init__.py
```diff
@@ -10,6 +12,64 @@
...
+            super(ProfileSettings, self).__setattr__(key, value)
+
+    def __getattr__(self, item):
+        return self._data[item]
+
+    def to_dict(self):
+        return {key: item.value() for key, item in self._data.items()}
+
+
+settings = ProfileSettings()
```
memsharded, line 68: Remove these global objects, this is an anti-pattern.

conans/test/functional/profiles/test_python_profiles.py
danimtb: I agree, the re-initilialization is not very elegant that way but wanted to try the clean syntax of not having to initialize a new object in the profile file. 

I think it is easier to just have `seettings` initialized and you can apply the values without having to "compose" the "Settings()" object. You only load the files in the given order and just apply pure python rules.

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,217 @@
...
+from conans.client.tools import environment_append
+from conans.test.utils.tools import TestClient
+from conans import ProfileSettings, ProfileOptions, ProfileEnv
+
+
+class ProfileLoader(object):
+
+    @staticmethod
+    def _init_profile():
+        import conans
```
memsharded, line 17: Continuation of the anti-pattern of the global objects, please avoid it.

conans/test/functional/profiles/test_python_profiles.py
danimtb: I have added a similar test with append, prepend, delete and so syntax to the `env` object

conans/test/functional/profiles/test_python_profiles.py
```diff
@@ -0,0 +1,217 @@
...
+        if parent_profile_path:
+            ProfileLoader._load_profile(parent_profile_path)
+        profile = ProfileLoader._load_profile(profile_path)
+        ProfileLoader._init_profile()
+        return profile.copy()
+
+
+@pytest.fixture
+def client():
+    return TestClient()
```
memsharded, line 51: A fixture of 1 line provides no value, they are for initialization reuse, but 1 liner is not worth (actually counter productive, losing locality is bad for reading tests). You can put all 3 fixtures in just 1, if you are always using them together: ``(client, pr1_content, pr2_content):``. Also include the ``client.save()`` in the fixture.

memsharded (changes requested):

conans/test/functional/profiles/test_python_profiles.py
memsharded: Not worth if we incur in such code anti-pattern, this access to global objects is bad, and we have also had very bad experience with it.

conans/test/functional/profiles/test_python_profiles.py
memsharded: This can wait until #8534 moves forward. You can leave it as is by now, but it would be good to use the same interface, the same that will be used in recipes defined in #8534 in ``Environment``.

memsharded:

PR #8554, opened 2021-02-25
More layouts: Generators folder, CMake Toolchain, Editable packages and more

lasote:
Changelog: Feature: Introduced the `layout()` method to the recipe to be able to declare the folder structure both for the local development methods (conan source, conan build...) and in the cache. Also, associated to the folders, cppinfo objects to be used in editable packages and file pattern descriptions to enable "auto packaging".
Docs: https://github.com/conan-io/docs/pull/2092

- The layout declares a generator folder (defaulted to install_folder if not declared in layout) where the generators and toolchain files are written. The conaninfo and the other weird stuff is kept in the install_folder (probably to be removed in Conan 2.0)
- Also new `cppinfo` objects in the layout entries.
- The CMake toolchain is injected with an absolute path to the generator folder.
- A little helper for `clion` layout as an example (not convinced about interface/UX)
- Auto package() based on layout.
- Auto package_info() based on layout.
- Editables following the build and source layouts.


conan/tools/cmake/cmake.py
```diff
@@ -46,7 +46,8 @@ class CMake(object):
     are passed to the command line, plus the ``--config Release`` for builds in multi-config
     """
 
-    def __init__(self, conanfile, generator=None, build_folder=None, parallel=True):
+    def __init__(self, conanfile, generator=None, parallel=True):
```
lasote: The new cmake build helper doesn't need the build_folder anymore

conan/tools/cmake/base.py
```diff
@@ -87,10 +87,10 @@ class CMakeToolchainBase(object):
             set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY ON)
             # To support the cmake_find_package generators
             {% if cmake_module_path -%}
-            set(CMAKE_MODULE_PATH {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
+            set(CMAKE_MODULE_PATH "{{ cmake_module_path }}" ${CMAKE_MODULE_PATH})
```
lasote: The path is now absolute so we quote it.

conans/client/cmd/build.py
```diff
@@ -44,6 +44,8 @@ def cmd_build(app, conanfile_path, source_folder, build_folder, package_folder,
         conan_file.layout.set_base_build_folder(build_folder)
         conan_file.layout.set_base_source_folder(source_folder)
         conan_file.layout.set_base_package_folder(package_folder)
+        conan_file.layout.set_base_generators_folder(base_path)
```
lasote: if `layout.generator.folder` is adjusted, that is relative to the base. Otherwise it uses the `install_folder`.
Note: All these folder parameters (source_folder, build_folder, package_folder...) should disappear in Conan 2.0

conans/client/conan_api.py
```diff
@@ -827,7 +828,6 @@ def imports(self, path, dest=None, info_folder=None, cwd=None):
         dest = _make_abs_path(dest, cwd)
 
         self.app.load_remotes()
-        mkdir(dest)
```
lasote, line 829: Now the mkdir is done inside the `run_imports(conanfile, dest)` always

conans/client/recorder/action_recorder.py
```diff
@@ -32,6 +32,10 @@ def _cpp_info_to_dict(cpp_info):
             doc["configs"] = configs_data
             continue
 
+        if it == "_rootpath":
```
lasote: This hack is because the json of the action recorder now receives the "rootpath" and I made it private to control the propagation of the rootpath to the components when it is changed (in the setter), So I needed to "hack" it to not break.

conans/model/build_info.py
```diff
@@ -136,11 +136,17 @@ def __init__(self):
         self._framework_paths = None
         self.version = None  # Version of the conan package
         self.description = None  # Description of the conan package
-        # When package is editable, filter_empty=False, so empty dirs are maintained
-        self.filter_empty = True
+        # By default we don't filter empty, because we want to filter ONLY when the final cppinfo
+        # from package is full of files but before that we will play with the build and source layout
+        # and cppinfo and it mess a lot
+        self.filter_empty = False
```
lasote: Important: These forced me to change some tests where it assumed the filter empty is always active. 
NOTE: The filter_empty should not belong to the cpp_info, when the final generator get the information of the dep tree it should clear the empty if needed, but not sooner.

conans/model/build_info.py
```diff
@@ -290,12 +335,35 @@ def __init__(self, ref_name, root_folder, default_values=None):
             self.builddirs.append(self._default_values.builddir)
         if self._default_values.frameworkdir is not None:
             self.frameworkdirs.append(self._default_values.frameworkdir)
-        self.components = DefaultOrderedDict(lambda: Component(self.rootpath,
-                                                               self.version, self._default_values))
+
+        self.components = DefaultOrderedDict(lambda: Component(self._rootpath,
```
lasote: The constructor now receives the filter_empty, so any changes in the variable are propagated when we create a new component in the `package_info`

conans/client/installer.py
```diff
@@ -190,50 +191,50 @@ def build_package(self, node, keep_build, recorder, remotes):
...
-                set_dirty(build_folder)
+                set_dirty(base_build_folder)
                 self._prepare_sources(conanfile, pref, package_layout, remotes)
-                self._copy_sources(conanfile, source_folder, build_folder)
+                self._copy_sources(conanfile, base_source_folder, base_build_folder)
 
         # BUILD & PACKAGE
         with package_layout.conanfile_read_lock(self._output):
-            mkdir(build_folder)
-            with tools.chdir(build_folder):
```
memsharded, line 209: Removing this seems a bit risky, I am not sure this chdir is done elsewhere, but removing might break users

conans/model/build_info.py
memsharded: This sounds it could be breaking for any user with a custom generator that relies on this behavior.

conans/client/installer.py
```diff
@@ -624,17 +652,31 @@ def _call_package_info(self, conanfile, package_folder, ref):
         with pythonpath(conanfile):
             with tools.chdir(package_folder):
                 with conanfile_exception_formatter(str(conanfile), "package_info"):
-                    conanfile.layout.set_base_package_folder(package_folder)
-                    conanfile.layout.set_base_source_folder(None)
-                    conanfile.layout.set_base_build_folder(None)
-                    conanfile.layout.set_base_install_folder(None)
                     self._hook_manager.execute("pre_package_info", conanfile=conanfile,
                                                reference=ref)
+                    conanfile.cpp_info.filter_empty = not is_editable
```
memsharded: Ok, I see this is reverting the filter_empty to True, good

conans/client/installer.py
lasote: Now It is done in the `_build`, do you see risk still? I could keep it here but seems unnecessary

conans/client/build/cmake.py
```diff
@@ -259,14 +259,14 @@ def configure(self, args=None, defs=None, source_dir=None, build_dir=None,
...
-                       os.pathsep.join(get_abs_path(f, self._conanfile.install_folder)
+                       os.pathsep.join(get_abs_path(f, self._conanfile.layout.generators_folder)
                                        for f in pkg_config_paths)}
         else:
             # If we are using pkg_config generator automate the pcs location, otherwise it could
             # read wrong files
             set_env = "pkg_config" in self._conanfile.generators \
                       and "PKG_CONFIG_PATH" not in os.environ
-            pkg_env = {"PKG_CONFIG_PATH": self._conanfile.install_folder} if set_env else None
+            pkg_env = {"PKG_CONFIG_PATH": self._conanfile.generators_folder} if set_env else None
```
memsharded:
```suggestion
            pkg_env = {"PKG_CONFIG_PATH": self._conanfile.layout.generators_folder} if set_env else None
```

conans/client/build/cmake.py
```diff
@@ -259,14 +259,14 @@ def configure(self, args=None, defs=None, source_dir=None, build_dir=None,
 
         if pkg_config_paths:
             pkg_env = {"PKG_CONFIG_PATH":
-                       os.pathsep.join(get_abs_path(f, self._conanfile.install_folder)
+                       os.pathsep.join(get_abs_path(f, self._conanfile.layout.generators_folder)
```
memsharded: maybe it makes sense to leave the old ``install_folder`` for the old build helpers, and only adopt it in the new generators in ``conan.tools.xxx``? Could this hide a regression that users with custom generators using ``.install_folder`` might suffer?

conans/client/build/cmake_flags.py
```diff
@@ -375,16 +375,16 @@ def get_definitions(self, cmake_version):
 
         # Adjust automatically the module path in case the conanfile is using the
         # cmake_find_package or cmake_find_package_multi
-        install_folder = self._conanfile.install_folder.replace("\\", "/")
+        generators_folder = self._conanfile.generators_folder.replace("\\", "/")
```
memsharded:
```suggestion
        generators_folder = self._conanfile.layout.generators_folder.replace("\\", "/")
```

Confused about when to use it from ``layout`` or from ``conanfile`` directly.

conans/client/graph/graph_manager.py
```diff
@@ -97,8 +97,9 @@ def load_consumer_conanfile(self, conanfile_path, info_folder,
             if test:
                 conanfile.display_name = "%s (test package)" % str(test)
                 conanfile.output.scope = conanfile.display_name
-
             run_configure_method(conanfile, down_options=None, down_ref=None, ref=None)
+            if hasattr(conanfile, "shape"):
```
memsharded: We still need a method name :)

conans/client/installer.py
```diff
@@ -452,29 +453,48 @@ def _build(self, nodes_by_level, keep_build, root_node, profile_host, profile_bu
...
+            write_toolchain(conanfile, conanfile.generators_folder, output)
+            output.info("Generated toolchain")
+            graph_info_node = GraphInfo(profile_host, root_ref=node.ref)
+            graph_info_node.options = node.conanfile.options.values
+            graph_info_node.graph_lock = graph_lock
+            graph_info_node.save(base_path)
+            output.info("Generated conan.lock")
+            # FIXME: Imports probably should go to a new imports_folder
+            copied_files = run_imports(conanfile, conanfile.build_folder)
+            report_copied_files(copied_files, output)
```
memsharded, line 484: Shouldn't this return here? It seems it is doing both the new ``shape`` logic and the previous old one.

conans/test/functional/layout/test_clion_layout.py
```diff
@@ -0,0 +1,87 @@
...
+
+    conanfile += """
+    def source(self):
+        tools.save("include/myheader.h", "")
+
+    def build(self):
+        tools.save("mylib.lib", "")
+
+    def shape(self):
+        clion_layout(self)
```
memsharded: need a way to override/select/conf by users

lasote: You can access/change the `self.layout` after that, or what do you mean?

conans/client/installer.py
lasote: Yeah, ok.

conans/client/graph/graph_manager.py
lasote: yup. I hate it.

conans/client/build/cmake_flags.py
lasote: Always with `conanfile` is the rule. Only when declaring layout except for: Special cases (have to be checked one by one), (some bugs LoL), and internal logic should access to the `conanfile.layout.xxx_folder`.

conans/client/build/cmake.py
lasote: Well, that is safer, but is an "all or nothing" situation. The layout mechanism is designed (except for bugs) to be compatible, but I know, I know... the regressions will ocurr.

conans/client/build/cmake.py
lasote: I think it is ok. At line `262` I think there is a bug because it is using the `.layout.generators_folder`. I'm checking.

conans/client/build/cmake_flags.py
lasote: And that's because the `conanfile.xxx_folder` prepends the correct root folder, the other one `conanfile.layout.xxx_folder` is just relative, so usually is not something to be used.

conans/client/build/cmake.py
lasote: Changed the other one to `self._conanfile.generators_folder`

conans/test/functional/layout/test_local_commands.py
```diff
@@ -0,0 +1,240 @@
...
+    client = TestClient()
+    conan_file = str(GenConanfile().with_import("from conans import tools"))
+    conan_file += """
+
+    def shape(self):
+        self.layout.generators.folder = "my_generators"
+        self.layout.build.folder = "my_build"
+
+    def build(self):
+        self.output.warn("Generators folder: {}".format(self.layout.generators_folder))
```
memsharded:
```suggestion
        self.output.warn("Generators folder: {}".format(self.generators_folder))
```

conans/client/build/autotools_environment.py
```diff
@@ -164,12 +164,12 @@ def configure(self, configure_dir=None, args=None, build=None, host=None, target
 
         if pkg_config_paths:
             pkg_env = {"PKG_CONFIG_PATH":
-                       [os.pathsep.join(get_abs_path(f, self._conanfile.install_folder)
+                       [os.pathsep.join(get_abs_path(f, self._conanfile.generators_folder)
```
memsharded: I cannot find the place where this is defined, and that by default should be the same folder it was before. I'd like to be sure of this.

conans/client/cmd/build.py
```diff
@@ -8,9 +8,9 @@
 from conans.util.log import logger
 
 
-def cmd_build(app, conanfile_path, source_folder, build_folder, package_folder, install_folder,
-              test=False, should_configure=True, should_build=True, should_install=True,
-              should_test=True):
+def cmd_build(app, conanfile_path, base_path, source_folder, build_folder, package_folder,
```
memsharded, line 11: why not call it "generators_folder"?

conan/tools/cmake/toolchain.py
```diff
@@ -373,8 +373,8 @@ class FindConfigFiles(Block):
     def context(self):
         # To find the generated cmake_find_package finders
         # TODO: Change this for parameterized output location of CMakeDeps
-        cmake_prefix_path = "${CMAKE_CURRENT_LIST_DIR}"
-        cmake_module_path = "${CMAKE_CURRENT_LIST_DIR}"
+        cmake_prefix_path = self._conanfile.generators_folder.replace("\\", "/")
```
memsharded: This one is tricky. Apparently it can stay, as the toolchain file will always be generated in the same folder (generator_folder) as the xxxxconfig.cmake scripts, apparently ``${CMAKE_CURRENT_LIST_DIR}`` will always be correct.

conan/tools/layout/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    conanfile.build_cpp_info.libdirs = ["."]
+
+    conanfile.folders.generators.folder = os.path.join(base, "generators")
+
+    conanfile.folders.source.folder = "."
+    conanfile.source_cpp_info.includedirs = ["include"]
+
+
+# FIXME: Not sure about the location, interface, name etc.
+#        Do we want a public explicit FileCopier maybe? Review and suggest!
```
memsharded: Yes, I have been wanting to make ``FileCopier`` explicit in recipes for a while. This seems related.

conans/test/functional/layout/test_in_cache.py
```diff
@@ -43,6 +43,75 @@ def package_info(self):
...
+
+        class HelloTestConan(ConanFile):
+            settings = "os", "compiler", "build_type", "arch"
+
+            def generate(self):
+                deps = CMakeDeps(self)
+                deps.generate()
+                tc = CMakeToolchain(self)
+                tc.generate()
+
```
memsharded:
```suggestion

        class HelloTestConan(ConanFile):
            settings = "os", "compiler", "build_type", "arch"
            generators = "CMakeDeps", "CMakeToolchain"
```

Just trying to make tests more compact

memsharded (commented):
I like this. I am going to play a little bit at the user level.

conans/model/conan_file.py
```diff
@@ -150,14 +151,38 @@ def __init__(self, output, runner, display_name="", user=None, channel=None):
...
         self._conan_node = None  # access to container Node object, to access info, context, deps...
         self.virtualenv = True  # Set to false to opt-out automatic usage of VirtualEnv
 
-        self._conan_new_cpp_info = None  # Will be calculated lazy in the getter
+        self._conan_new_cpp_info = None   # Will be calculated lazy in the getter
+
+        # layout() method related variables:
+        self.folders = Folders()
+        self.patterns = Patterns()
+        self.infos = Infos()
```
lasote: Not a good name. Please suggest.

conans/client/conanfile/package.py
```diff
@@ -14,12 +14,15 @@
...
-def run_package_method(conanfile, package_id, hook_manager, conanfile_path, ref, copy_info=False):
+def run_package_method(conanfile, package_id, hook_manager, conanfile_path, ref, copy_info=False,
+                       use_pkg_layout=False):
     """ calls the recipe "package()" method
     - Assigns folders to conanfile.package_folder, source_folder, install_folder, build_folder
     - Calls pre-post package hook
     - Prepares FileCopier helper for self.copy
     """
+    if use_pkg_layout:
+        conanfile.folders.set_base_package(conanfile.folders.local_package_folder)
```
memsharded: This is the ugly bit. why passing itself (folders) as argument?

memsharded: Probably can be called even before ``run_package_method()``, and avoid the argument completely.

conans/client/graph/graph_builder.py
```diff
@@ -121,6 +121,10 @@ def _expand_node(self, node, graph, down_reqs, down_ref, down_options, check_upd
         new_options, new_reqs = self._get_node_requirements(node, graph, down_ref, down_options,
                                                             down_reqs, graph_lock, update, remotes)
 
+        # Once the node is configured call the layout()
+        if hasattr(node.conanfile, "layout"):
```
memsharded: Can this be put inside ``run_configure_method()``?

conans/client/manager.py
```diff
@@ -92,16 +91,27 @@ def deps_install(app, ref_or_path, install_folder, graph_info, remotes=None, bui
...
-            write_toolchain(conanfile, install_folder, output)
+        tmp = list(conanfile.generators)  # Add the command line specified generators
+        generators = set(generators) if generators else set()
+        tmp.extend([g for g in generators if g not in tmp])
+        conanfile.generators = tmp
+        app.generator_manager.write_generators(conanfile, install_folder,
+                                               conanfile.generators_folder,
+                                               output)
+        write_toolchain(conanfile, conanfile.generators_folder, output)
+        if add_txt_generator:
```
memsharded: Why not add ``txt`` before and let the first call to ``write_generators()`` hadle it?

conans/model/conan_file.py
```diff
@@ -150,14 +151,38 @@ def __init__(self, output, runner, display_name="", user=None, channel=None):
...
+        self.infos.package.includedirs = ["include"]
+        self.infos.package.libdirs = ["lib"]
+        self.infos.package.bindirs = ["bin"]
+        self.infos.package.resdirs = ["res"]
+        self.infos.package.builddirs = [""]
+        self.infos.package.frameworkdirs = ["Frameworks"]
+
+        # Init all fields with None so we can certainly know when a field has been declared
+        # to be able to merge them to the package one
+        self.infos.source = NewCppInfo()
```
memsharded: Isn't this the ``Infos()`` default already?

memsharded (commented):
Getting there, good job!

conans/model/conan_file.py
lasote: yes, this is no longer necessary because I simplified the None stuff.

conans/model/conan_file.py
lasote: Changed to `self.cpp`

conan/tools/cmake/cmakedeps.py
```diff
@@ -33,18 +33,18 @@
...
-   function(conan_package_library_targets libraries package_libdir deps out_libraries out_libraries_target build_type package_name)
+   function(conan_package_library_targets libraries package_libdir deps out_libraries out_libraries_target config_suffix package_name)
        unset(_CONAN_ACTUAL_TARGETS CACHE)
 
        foreach(_LIBRARY_NAME ${libraries})
            find_library(CONAN_FOUND_LIBRARY NAME ${_LIBRARY_NAME} PATHS ${package_libdir}
                         NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
            if(CONAN_FOUND_LIBRARY)
-               conan_message(STATUS "Library ${_LIBRARY_NAME} found ${CONAN_FOUND_LIBRARY}")
+               conan_message(DEBUG "Library ${_LIBRARY_NAME} found ${CONAN_FOUND_LIBRARY}")
```
lasote, line 43: Might be a good idea to keep the ".a" found as DEBUG because we have the new traces with the targets that probably is more useful.

conan/tools/cmake/cmakedeps.py
```diff
@@ -54,17 +54,17 @@
                    conan_message(STATUS "Skipping already existing target: ${_LIB_NAME}")
                endif()
                list(APPEND _out_libraries_target ${_LIB_NAME})
-               conan_message(STATUS "Found: ${CONAN_FOUND_LIBRARY}")
+               conan_message(DEBUG "Found: ${CONAN_FOUND_LIBRARY}")
            else()
-               conan_message(ERROR "Library ${_LIBRARY_NAME} not found in package")
+               conan_message(FATAL_ERROR "Library '${_LIBRARY_NAME}' not found in package. If '${_LIBRARY_NAME}' is a system library, declare it with 'cpp_info.system_libs' property")
```
lasote, line 59: Please confirm behavior, "ERROR" is not a valid tag, Do we want to fail, right? Old generators didn't fail but printed a message.

memsharded: Yes, we want to fail.

conans/test/functional/conan_build_info/new_build_info_test.py
```diff
@@ -1,8 +1,8 @@
 import pytest
 
-from conans.errors import ConanException
 from conans.model.build_info import CppInfo
-from conans.model.new_build_info import NewCppInfo, _DIRS_VAR_NAMES, _FIELD_VAR_NAMES
+from conans.model.new_build_info import NewCppInfo, _DIRS_VAR_NAMES, _FIELD_VAR_NAMES, \
```
memsharded, line 4: Pending a refactor to move this file to unittests, it doesn't seem functional at all:
- functional: actually calling cmake and building things
- integration: can be very complete, but without building nor any external tool needed: cmake, ninja (those would be functional).

Not for this PR, of course.

memsharded (commented):
I have pushed a couple of minor things:
- CMake() was broken for some local flows, because not finding the ``conanbuild.json`` file (sorry, no tests, I was checking manually)
- Redundant ``layout()`` call, it was already run inside ``run_configure_method()``

Please check https://github.com/conan-io/conan/pull/8554/commits/ffaa4cb59d81f320bf6c25cc0a216530f9593f01

conans/test/functional/conan_build_info/new_build_info_test.py
lasote: Sure, that is unittest I will move it no problem.

PR #8599, opened 2021-03-04
Feature: set CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON

### LINKED ISSUE #8582: [feature] Set `CMAKE_FIND_PACKAGE_PREFER_CONFIG` in `conan_toolchain.cmake`
Please consider setting `CMAKE_FIND_PACKAGE_PREFER_CONFIG` in the `conan_toolchain.cmake`. This will allow CMake to prefer configuration files generated by `CMakeDeps` over the ones provided with CMake release (i.e. `FindGTest.cmake`).

SSE4:
closes: #8582 
/cc @mpusz 
Changelog: Feature: Set `CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`.
Docs: omit

### DIFF
--- a/conan/tools/cmake/base.py
+++ b/conan/tools/cmake/base.py
@@ -152,6 +152,8 @@ class CMakeToolchainBase(object):
             conanfile.conf["tools.cmake.cmaketoolchain"].find_package_prefer_config
         if self.find_package_prefer_config is not None:
             self.find_package_prefer_config = "OFF" if self.find_package_prefer_config.lower() in ("false", "0", "off") else "ON"
+        else:
+            self.find_package_prefer_config = "ON"  # assume ON by default if not specified in conf
     def _get_templates(self):
         return {
             'toolchain_macros': self._toolchain_macros_tpl,
--- a/conan/tools/cmake/base.py
+++ b/conan/tools/cmake/base.py
@@ -85,7 +85,10 @@ class CMakeToolchainBase(object):
             # We are going to adjust automagically many things as requested by Conan
             #   these are the things done by 'conan_basic_setup()'
             set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY ON)
-            set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
+            {%- if find_package_prefer_config %}
+            set(CMAKE_FIND_PACKAGE_PREFER_CONFIG {{ find_package_prefer_config }})
+            {%- endif %}
+            #set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
             # To support the cmake_find_package generators
             {% if cmake_module_path -%}
             set(CMAKE_MODULE_PATH {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
@@ -134,6 +137,8 @@ class CMakeToolchainBase(object):
         """)
 
     def __init__(self, conanfile, **kwargs):
+        import six
+
         self._conanfile = conanfile
         self.variables = Variables()
         self.preprocessor_definitions = Variables()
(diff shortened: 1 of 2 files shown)


solvingj:
It seems to me that this should be controlled via `[conf]`. As @SSE4 has said in the past, if many people using CMake try to move over to the toolchain generator and it turns out this behavior is breaking for them, it would be great if they had a way to opt-out right away.

solvingj (changes requested):
Consider adding `[conf]` item for this.

memsharded:
Sounds good, lets add a [conf] to opt-out this behavior, I'd suggest something like:
```
[conf]
tools.cmake.cmaketoolchain:find_package_prefer_config=False
```

SSE4:
now `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is controlled by `tools.cmake.cmaketoolchain:find_package_prefer_config`

conan/tools/cmake/base.py
```diff
@@ -143,6 +149,15 @@ def __init__(self, conanfile, **kwargs):
...
 
+        self.find_package_prefer_config = \
+            conanfile.conf["tools.cmake.cmaketoolchain"].find_package_prefer_config
+        if self.find_package_prefer_config is not None:
+            if isinstance(self.find_package_prefer_config, six.string_types):
+                self.find_package_prefer_config = self.find_package_prefer_config == "True"
+            elif isinstance(self.find_package_prefer_config, six.integer_types):
+                self.find_package_prefer_config = bool(self.find_package_prefer_config)
+            self.find_package_prefer_config = "ON" if self.find_package_prefer_config else "OFF"
+
```
SSE4: two notes:
- right now, `conanfile.conf` always returns string, so it returns string value `"True"` instead of boolean `True`. I am not sure if it will be always the same, so I check type for the safety
- I believe it should be a function (just like [to_meson_value](https://github.com/conan-io/conan/blob/639450053a0359af5a757b3617a82a2ece2f7cf1/conan/tools/meson/toolchain.py#L77)), but I guess we will do it to address #8638  anyway

solvingj (approved):

conan/tools/cmake/base.py
```diff
@@ -143,6 +148,15 @@ def __init__(self, conanfile, **kwargs):
 
         self.build_type = None
 
+        self.find_package_prefer_config = \
+            conanfile.conf["tools.cmake.cmaketoolchain"].find_package_prefer_config
+        if self.find_package_prefer_config is not None:
```
memsharded, line 153: This is not opting-out, it is opting-in please recall my comment:

> Sounds good, lets add a [conf] to opt-out this behavior, I'd suggest something like


Then simplify code, as you said they are always strings, so go for a oneliner:

```python
if self.find_package_prefer_config is not None:
    self.find_package_prefer_config = "OFF" if self.find_package_prefer_config.lower() in ("false", "0", "off") else "ON"
```

memsharded (changes requested):

conan/tools/cmake/base.py
SSE4: I am not sure I understand. this is opt-out, the behavior matches CMake exactly:
- conf not defined -> CMAKE_FIND_PACKAGE_PREFER_CONFIG not defined
- conf=True -> CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON
- conf=False -> CMAKE_FIND_PACKAGE_PREFER_CONFIG=OFF

mpusz: The whole idea was to have it ON by default so inexperienced users using `conanfile.txt` without any `conf` sections have a friendly dev env. Advanced users may want to disable it if they so desire.

As a side note, still there are many inexperienced users out there both for Conan and CMake. Doing things that we consider trivial is not easy for them at all. For the last two days, I am remotely helping one of them to enable `mp-units` support via `conanfile.txt` even though I thought that my docs are clear: https://mpusz.github.io/units/usage.html#conan-cmake-release. Today I finally saw that person's CMake file and it turned out to be:

image

SSE4: okay, I added the ON as default again. there was a confusion because initially I implemented ON as default, then others said it should be controlled by conf instead. hopefully the current behavior makes sense:
- conf not defined -> CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON
- conf=True -> CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON
- conf=False -> CMAKE_FIND_PACKAGE_PREFER_CONFIG=OFF

(although it's not exactly an opt out for the default CMake behavior, it should help for your case)

mpusz: Actually, in this case, a user correctly applied `CONFIG REQUIRED` as suggested in my docs. The problem was with the last two lines 😉

memsharded: @SSE4 my request was:

> Sounds good, lets add a [conf] to opt-out this behavior, I'd suggest something like:

A config required to **opt-out**. That means that by default it is ON, and you need to do something proactively to opt-out of that feature. If you need to define a conf for activating the feature, it is not an opt-out feature, it is an opt-in feature.

PR #8601, opened 2021-03-04
Support PKG_CONFIG env var to define the pkg-config executbale used by tools.PkgConfig

### LINKED ISSUE #8591: [feature] tools.PkgConfig should honor $PKG_CONFIG.
(My Version Info: Conan 1.33, pip-installed on Debian unstable)

Many thanks for providing conan! I might have a suggestion how to improve the tools.PkgConfig integration, as it is currently not behaving as expected-by-convention: It is ignoring the PKG_CONFIG env…

The autotools documentation mentions this in [1] that one can specify the pkg-config tool to be used by setting env PKG_CONFIG. It's also mentioned in the manpage of pkg-config (autotools macro section). Also, cmake mentions it in
[2] and my experiments shows that indeed autotools and cmake honors it. Calling ./configure, on projects needing it, without pkg-config being present tells you to set PKG_CONFIG to point to the executable [3].

Unfortunately conan's tools.PkgConfig does not honor it; so the "wrong" pkg-config will be called. Of course, tools.PkgConfig can be hinted explicitly, but this is cumbersome to do so in every recipe if not using the documented PKG_CONFIG way.

My use case: I'm running into a pkg-config bug [4]  Until that one is fixed, I need a local wrapper to work around the problem and the script is provided by an conan package and overrides the default pkg-config (but still needs to call it.)

This would make it more easy to check other pkg-config implementations, e.g pkgconf (which did not help with my issue, as it has the same issue [5]…)

[1] https://autotools.io/pkgconfig/cross-compiling.html  -- Section 5.2. Tool Calling Conventions
Quote:  "This is supported by the autoconf macros provided by the package, as they all respect $PKG_CONFIG if set in the environment,"
[2] https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#variables-affecting-behavior
Quote: The PKG_CONFIG environment variable can be used as a hint.
[3] https://autotools.io/pkgconfig/pkg_check_modules.html -- see the example output in 3.4. Optional Modules section or
     another IRL example at https://bugs.debian.org/528169
[4] https://gitlab.freedesktop.org/pkg-config/pkg-config/-/issues/52
[5] https://github.com/pkgconf/pkgconf/issues/213

danimtb:
Changelog: Feature: Support `PKG_CONFIG` env var to define the pkg-config executbale used by `tools.PkgConfig`

- [x] Refer to the issue that supports this Pull Request: closes #8591

### DIFF
--- a/conans/client/tools/pkg_config.py
+++ b/conans/client/tools/pkg_config.py
@@ -6,7 +6,6 @@ import subprocess
 
 from conans.errors import ConanException
 from conans.util.runners import check_output_runner
-from util.env_reader import get_env
 
 
 class PkgConfig(object):
--- a/conans/client/tools/pkg_config.py
+++ b/conans/client/tools/pkg_config.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import os
 import subprocess
 
 from conans.errors import ConanException
 from conans.util.runners import check_output_runner
+from util.env_reader import get_env
 
 
 class PkgConfig(object):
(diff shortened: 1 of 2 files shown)


memsharded:
Outdated, please create a new PR for the new tools if it still makes sense (I think that just adding the pkg-config to the PATH ``[buildenv]`` should be good for this, or is it that the executable itself has different names?)

PR #8604, opened 2021-03-04
Fix transitive python requires

### LINKED ISSUE #8546: [question] Unclear how to use transitive python_requires
The documentation at https://docs.conan.io/en/latest/extending/python_requires.html#resolution-of-python-requires states that:

> python-requires can python-require other recipes too, but this should probably be limited to very few cases, we recommend to use the simplest possible structure.

But it is unclear how that should work.

I made an example where I define a base class in a python_requires module. This base class uses a helper function from another python_requires module. How can I transitively define that the base class has another python_requires? At the moment I can only get it to work by specifying all python_requires explicitly in the consumer.

Example: [transitive-python_requires.tar.gz](https://github.com/conan-io/conan/files/6035128/transitive-python_requires.tar.gz)

To run the example:
- `conan export shared-function`
- `conan export base-class`
- `conan create consumer`

It will fail with:
```
ERROR: consumer/1.0.0: Error in requirements() method, line 8
	assert self.python_requires["shared-function"].module.helper(1, 2) == 3
	ConanException: 'shared-function' is not a python_require
```

Alternatively, using the old python_requires method results in a crash.

danimtb:
Changelog: Bugfix: Fix: Transitive python requires not working with the new syntax.
Docs: omit

- [x] Refer to the issue that supports this Pull Request: closes #8546


conans/client/graph/python_requires.py
```diff
@@ -54,7 +54,12 @@ def __getitem__(self, item):
         try:
             return self._pyrequires[item]
         except KeyError:
-            raise ConanException("'%s' is not a python_require" % item)
+            # https://github.com/conan-io/conan/issues/8546
```
memsharded, line 57: Sounds like this class could get a refactor, but lets wait until moving this forward first, if we think it is correct.

memsharded (approved):

czoido (approved):

PR #8605, opened 2021-03-04
[wip] Fix config install not working when .git* folder is in the path

### LINKED ISSUE #8594: [bug] conan config install skips files if ".git" is in the path
I had set my `CONAN_USER_HOME` env-var to "$HOME/.gitlab-conan" (doesn't matter why).  
Later, I ran the `conan config install` command, providing a git repo that stores shared-company config, but that command has appeared to silently fail, or do nothing at all, as the config hasn't been applied, at all.  
The configuration repo is properly configured as I've tested it on other environments where conan's home is set to the default path, `~/.conan`, and it works as expected.  

When I run the command, I always get the output `Repo cloned!` and that's it, it quits right after. So I've decided to dig in and find out why, and I found that inside the [conans/client/conf/config_installer.py file](https://github.com/conan-io/conan/blob/03b3d71890c537df1154b3d04cc35ed2654a5894/conans/client/conf/config_installer.py), on [this line](https://github.com/conan-io/conan/blob/03b3d71890c537df1154b3d04cc35ed2654a5894/conans/client/conf/config_installer.py#L133), there's a check for the `.git` word being in the `root` string returned from `walk()`. I get the idea here, but it might be required to re-implement this to look for `.git` only in the last component of the `root`, as this is what actually matters, allowing other components of the path to anything the user desires.  
It might not be that trivial to solve this nicely, so an alternative could be to warn users in the docs about setting conan home to paths that include `.git` in it.

  * Operating System+version: Ubuntu 16.04 (Docker container)
  * Conan version: 1.34.0
  * Python version: 3.8.8

1. Set the conan home to a path that has `.git` in it, no matter where exactly, e.g. `.gitlab`
2. Try to install some config using `conan config install`

danimtb:
Changelog: Bugfix: Fix config install not working when .git* folder is in the path.
Docs: omit

closes #8594

### DIFF
--- a/conans/client/conf/config_installer.py
+++ b/conans/client/conf/config_installer.py
@@ -130,7 +130,7 @@ def _process_folder(config, folder, cache, output):
         folder = os.path.join(folder, config.source_folder)
     for root, dirs, files in walk(folder):
         dirs[:] = [d for d in dirs if d != ".git"]
-        if ".git" in root:
+        if os.path.join(".git", "") in root:
             continue
         for f in files:
             _process_file(root, f, config, cache, output, folder)
--- a/conans/client/conf/config_installer.py
+++ b/conans/client/conf/config_installer.py
@@ -130,8 +130,6 @@ def _process_folder(config, folder, cache, output):
         folder = os.path.join(folder, config.source_folder)
     for root, dirs, files in walk(folder):
         dirs[:] = [d for d in dirs if d != ".git"]
-        if os.path.join(".git", "") in root:
-            continue
         for f in files:
             _process_file(root, f, config, cache, output, folder)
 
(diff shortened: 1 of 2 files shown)


conans/client/conf/config_installer.py
```diff
@@ -130,7 +130,7 @@ def _process_folder(config, folder, cache, output):
         folder = os.path.join(folder, config.source_folder)
     for root, dirs, files in walk(folder):
         dirs[:] = [d for d in dirs if d != ".git"]
-        if ".git" in root:
+        if os.path.join(".git", "") in root:
```
czoido: I'm not sure if this will make a difference in the end, the line above: `dirs[:] = [d for d in dirs if d != ".git"]`
Is already filtering the _.git_ folder.
Also, the new comparison will compare if `.git/` is in root that will have the value `my/root/path/git` so it will always be false because of the forward slash join is going to add in the end.
I don't know if I'm missing something, but is this comparison needed in the end?

danimtb: you are right, I have removed that line as it looks useless and added a check in the test

czoido (approved):

PR #8618, opened 2021-03-09
Replace custom TestBufferConanOutput with Python Mock

memsharded:
Changelog: Omit
Docs: Omit


lasote (approved):
Agree. If we are not going to check anything in the output it is not needed to pass a custom mock.

PR #8625, opened 2021-03-10
Do not remove sh from path in the new CMake helper

### LINKED ISSUE #8597: Remove removal of "sh" from path for MinGW in CMake
New conan.tools.cmake CMake helper contains:

```python
 if is_windows_mingw:
                with tools.remove_from_path("sh"):
                    self._conanfile.run(command)
```

This is no longer necessary, can be replaced by ``set(DCMAKE_SH="CMAKE_SH-NOTFOUND")``. Please remove it and add a red/green test for it.

danimtb:
Changelog: Feature: Do not remove sh from the path in the new CMake helper.
Docs: https://github.com/conan-io/docs/pull/2055

#TAGS: slow

- [x] Refer to the issue that supports this Pull Request: closes #8597

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -563,6 +563,6 @@ class TestMinGW:
         client.run_command("where cmake")
         print("CMAKE PATH: %s" % client.out)
         client.run_command("cmake --version")
-        assert "cmake version 3.16.9" in client.out
+        assert "cmake version 3.16" in client.out
         client.run("create . test/1.0@ --profile profile", assert_error=True)
         print(client.out)
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -546,6 +546,7 @@ class TestMinGW:
     main_cpp = gen_function_cpp(name="main")
 
     @pytest.mark.tool_mingw64
+    @pytest.mark.tool_cmake(version="3.16")
     def test_mingw64(self):
         profile = textwrap.dedent("""
             [settings]
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -272,6 +272,7 @@ class WinTest(Base):
     @parameterized.expand([("Debug", "libstdc++", "4.9", "98", "x86_64", True),
                            ("Release", "libstdc++", "4.9", "11", "x86_64", False)])
     @pytest.mark.tool_mingw64
+    @pytest.mark.tool_cmake(version="3.15")
     def test_toolchain_mingw_win(self, build_type, libcxx, version, cppstd, arch, shared):
         # FIXME: The version and cppstd are wrong, toolchain doesn't enforce it
         settings = {"compiler": "gcc",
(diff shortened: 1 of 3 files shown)


SSE4 (approved):

memsharded (changes requested):
~~This goes to ``develop2`` branch!~~ Update: My bad, it is good in ``develop``, it only affects new CMake tools

It will need the CMAKE-SH_NOTFOUND, maybe this is not sufficiently covered by tests, I would expect something to fail. Maybe it is a "slow" test?

danimtb:
I was expecting the test added here or other ones using mingw to fail with this PR but it is not happening. The Ci is using CMake 3.17.3 (https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-8625/4/pipeline/307/#step-355-log-1441) that do not raise error if ssh is in the path (it was fixed in CMake 3.17.0 https://gitlab.kitware.com/cmake/cmake/-/merge_requests/3978).

We need to run a lower version in the CI in order to test this properly.

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -517,3 +517,52 @@ def build(self):
...
+            compiler.libcxx=libstdc++
+            compiler.cppstd=98
+            """)
+        client = TestClient()
+        client.save({"conanfile.py": self.conanfile, "CMakeLists.txt": self.cmakelists,
+                     "main.cpp": self.main_cpp, "profile": profile})
+        client.run_command("where cmake")
+        print("CMAKE PATH: %s" % client.out)
+        client.run_command("cmake --version")
+        assert "cmake version 3.16.9" in client.out
```
danimtb:
```suggestion
        assert "cmake version 3.16" in client.out
```
```suggestion
        assert "cmake version 3.16.9" in client.out
```

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -528,3 +528,53 @@ def build(self):
         client.run("install .")
         client.run("build .")
         self.assertIn("VALUE OF CONFIG STRING: my new value", client.out)
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Tests Windows MinGW")
```
memsharded: ``TestSubsystemsCMakeBuild`` is already building very similar tests, isn't there any overlap?

danimtb: Still a draft. I made another test for better isolation of the test but I will unify them to the existing MinGW ones once I figure out the cmake behavior

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -528,3 +528,51 @@ def build(self):
...
+            compiler.version=4.9
+            compiler.libcxx=libstdc++
+            compiler.cppstd=98
+            """)
+        client = TestClient()
+        client.save({"conanfile.py": self.conanfile, "CMakeLists.txt": self.cmakelists,
+                     "main.cpp": self.main_cpp, "profile": profile})
+        client.run_command("cmake --version")
+        assert "cmake version 3.15" in client.out
+        client.run_command('cmake . -G "MinGW Makefiles" -DCMAKE_SH=CMAKE_SH-NOTFOUND')
```
memsharded: I have investigated, and yes, it is hardcoded in the CMake codebase:

```cpp
if (this->AlreadyInCache) {
    // If the user specifies the entry on the command line without a
    // type we should add the type and docstring but keep the original
    // value.
    if (this->AlreadyInCacheWithoutMetaInfo) {
      this->Makefile->AddCacheDefinition(this->VariableName, "",
                                         this->VariableDocumentation.c_str(),
                                         cmStateEnums::FILEPATH);
    }
    return true;
  }
```
This is the only way it will work, in the command line argument. Please add it to the command line conditionally to MinGW and Windows

memsharded (changes requested):

PR #8627, opened 2021-03-10
poc of testing and installing build_requires as build_requires

memsharded:
Changelog: Feature: New ``conan install <ref> --build-require`` and ``conan create <path> --build-require``  (when not using ``test_package``) arguments to explicitly define that the installed or created package has to be a ``build-require``, receiving the build profile instead of the host one.
Docs: omit

This follows #8787 that added support for the ``test_package``. 
To discuss: Should we use ``--build-require`` always for ``conan create``, even if it is defined in ``test_package/conanfile.py`` with ``test_type = "build_requires"``?

#tags: slow


conans/client/graph/graph_manager.py
```diff
@@ -282,6 +296,7 @@ def _recurse_build_requires(self, graph, builder, check_updates,
         """
         :param graph: This is the full dependency graph with all nodes from all recursions
         """
+        print("RECUSRS BUILD REQUIRES!!!!!!!!!!!!!!!!!!!!!!!!!", apply_build_requires, nodes_subset)
```
lasote: rm

conans/client/installer.py
```diff
@@ -576,7 +577,10 @@ def _propagate_info(self, node, using_build_profile):
 
         for n in node_order:
             if n not in transitive:
-                conan_file.output.info("Applying build-requirement: %s" % str(n.ref))
+                try:
+                    conan_file.output.info("Applying build-requirement: %s" % str(n.ref))
```
lasote: ?

conans/test/integration/build_requires/test_install_test_build_require.py
```diff
@@ -0,0 +1,123 @@
...
+    cmd = ". ./activate.sh && mycmake.sh" if platform.system() != "Windows" else \
+        "activate.bat && mycmake.bat"
+    client.run_command(cmd)
+    system = {"Darwin": "Macos"}.get(platform.system(), platform.system())
+    assert "MYCMAKE={}!!".format(system) in client.out
+    assert "MYOPENSSL={}!!".format(system) in client.out
+
+
+def test_build_require_command_line_build_context(client):
+    client.run("install mycmake/1.0@ --build-require -g virtualenv -pr:b=default")
```
lasote, line 194: Shouldn't we change the build profile to see if the output changes?

conans/client/graph/graph_manager.py
```diff
@@ -290,11 +304,11 @@ def _recurse_build_requires(self, graph, builder, check_updates,
         for node in graph.ordered_iterate(nodes_subset):
             # Virtual conanfiles doesn't have output, but conanfile.py and conanfile.txt do
             # FIXME: To be improved and build a explicit model for this
-            if node.recipe == RECIPE_VIRTUAL:
-                continue
+            #if node.recipe == RECIPE_VIRTUAL:
```
lasote: ?

conans/client/installer.py
memsharded: The printing of virtual consumers is disabled, because it was very confusing to users, seeing something like ``(virtual): Applying build-requirement...``. So ``conanfile.output.info`` or any other printing command will raise for ``RECIPE_VIRTUAL`` recipe.

conans/test/integration/build_requires/test_install_test_build_require.py
memsharded: Not sure what you mean, the build-profile is different in each tested platform, and the output changes accordingly, it is parameterized to the system

conans/client/graph/graph_manager.py
memsharded: ``Virtual`` recipes couldn't have ``build_requires`` so far, only regular ``requires``. This is opening the door for this possibility.

conans/client/loader.py
```diff
@@ -306,8 +306,11 @@ def load_virtual(self, references, profile_host, scope_options=True,
         conanfile.conf = profile_host.conf.get_conanfile_conf(None)
         conanfile.settings = profile_host.processed_settings.copy_values()
 
-        for reference in references:
-            conanfile.requires.add_ref(reference)
+        if is_build_require:
```
SSE4, line 310: may they have the same interface? e.g.
```
        if is_build_require:
            for reference in references:
                conanfile.build_requires.add_ref(reference)
        else:
            for reference in references:
                conanfile.requires.add_ref(reference)
```

conans/test/integration/build_requires/test_install_test_build_require.py
```diff
@@ -0,0 +1,123 @@
...
+        assert "MYOPENSSL={}!!".format(system) in client.out
+
+
+def test_build_require_conanfile_text(client):
+    client.save({"conanfile.txt": "[build_requires]\nmycmake/1.0"}, clean_first=True)
+    client.run("install . -g virtualenv")
+    cmd = ". ./activate.sh && mycmake.sh" if platform.system() != "Windows" else \
+        "activate.bat && mycmake.bat"
+    client.run_command(cmd)
+    system = {"Darwin": "Macos"}.get(platform.system(), platform.system())
```
SSE4, line 178: why this pattern is repeating and needed?

lasote (approved):

czoido (approved):

PR #8635, opened 2021-03-11
decouple in new class the preparation logic of packages before the actual upload

memsharded:
Changelog: Omit
Docs: Omit

A first step towards having an independent definition of package to upload.

This is a pure refactor, no functionality should be changed at all. Some minor cleanings to the interfaces could be done, but those wouldn't be pure refactors, I prefer to leave them for subsequent PRs, to make sure to make them more visible

### DIFF
--- a/conans/client/cmd/uploader.py
+++ b/conans/client/cmd/uploader.py
@@ -150,6 +150,259 @@ class _UploadCollecter(object):
         return refs_by_remote
 
 
+class _PackagePreparator(object):
+    def __init__(self, cache, remote_manager, hook_manager, output):
+        self._cache = cache
+        self._remote_manager = remote_manager
+        self._output = output
+        self._hook_manager = hook_manager
+
+    def prepare_recipe(self, ref, conanfile, remote, remotes, policy):
+        """ do a bunch of things that are necessary before actually executing the upload:
+        - retrieve exports_sources to complete the recipe if necessary
+        - compress the artifacts in conan_export.tgz and conan_export_sources.tgz
+        - check if package is ok to be uploaded, if scm info missing, will raise
+        - check if the remote recipe is newer, raise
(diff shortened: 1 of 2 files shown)


conans/client/cmd/uploader.py
```diff
@@ -150,6 +150,259 @@ def _collect_packages_to_upload(self, refs, confirm, remotes, all_packages, quer
...
+class _PackagePreparator(object):
+    def __init__(self, cache, remote_manager, hook_manager, output):
+        self._cache = cache
+        self._remote_manager = remote_manager
+        self._output = output
+        self._hook_manager = hook_manager
+
+    def prepare_recipe(self, ref, conanfile, remote, remotes, policy):
+        """ do a bunch of things that are necessary before actually executing the upload:
+        - retrieve exports_sources to complete the recipe if necessary
```
solvingj, line 162: Is it too ambitious to separate this step out even further? Recall that we have multiple people who would like to see a new command or flag added which can force conan to retrieve the `exports_sources` for debug purposes.  If we pull it out to a separate function here, makes it that much easier to add that feature in near future (potentially pre-2.x)

solvingj (approved):
Seems like a really great first step while it's fresh in mind.

conans/client/cmd/uploader.py
memsharded: Please note that this is only calling the ``retrieve_exports_sources(self._remote_manager, self._cache, conanfile, ref, remotes)``, which is only getting the ``export_source`` files, but not the actual complete source files necessary to build and debug, that will be the result of merging that with exported files, scm and any other action that the ``source()`` method is doing. So it seems the entry point for that functionality would be closer to the ``source.py`` files, and shouldn't be extracted from here.

PR #8645, opened 2021-03-14
allow integers as CMakeToolchain.preprocessor_definitions

### LINKED ISSUE #8632: [feature] allow for non-string configuration-specific preprocesor definitions on CMakeToolchain
Currently CMakeToolchain only allows for string literal definitions due to quotes included on [this line of jinja template](https://github.com/conan-io/conan/blob/develop/conan/tools/cmake/base.py#L128).
`add_definitions(-D{{ it }}="{{ value }}")`

This limits the hability to declare integer or other definitions on generate() method. I hope this could be easily workaround by removing the quotes from the jinja template and, instead, adding them on python string variable

memsharded:
Changelog: Feature: Allow integers as ``preprocessor_definitions`` in ``CMakeToolchain``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/8632

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -346,7 +346,10 @@ class LinuxTest(Base):
         pic_str = "" if shared else "ON"
         arch_str = "-m32" if arch == "x86" else "-m64"
         cxx11_abi_str = "1" if libcxx == "libstdc++11" else "0"
-        defines = '_GLIBCXX_USE_CXX11_ABI=%s;MYDEFINE="MYDEF_VALUE";MYDEFINEINT=42;' % cxx11_abi_str
+        defines = '_GLIBCXX_USE_CXX11_ABI=%s;MYDEFINE="MYDEF_VALUE";MYDEFINEINT=42;'\
+                  'MYDEFINEINT_CONFIG=$<IF:$<CONFIG:debug>,421,$<IF:$<CONFIG:release>,422,"">>;'\
+                  'MYDEFINE_CONFIG=$<IF:$<CONFIG:debug>,"MYDEF_DEBUG",'\
+                  '$<IF:$<CONFIG:release>,"MYDEF_RELEASE","">>' % cxx11_abi_str
         vals = {"CMAKE_CXX_STANDARD": "14",
                 "CMAKE_CXX_EXTENSIONS": extensions_str,
                 "CMAKE_BUILD_TYPE": build_type,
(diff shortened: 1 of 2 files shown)


conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -36,8 +36,11 @@ def generate(self):
                 tc.variables.debug["MYVAR2_CONFIG"] = "MYVAR2_DEBUG"
                 tc.variables.release["MYVAR2_CONFIG"] = "MYVAR2_RELEASE"
                 tc.preprocessor_definitions["MYDEFINE"] = "MYDEF_VALUE"
+                tc.preprocessor_definitions["MYDEFINEINT"] = 42
                 tc.preprocessor_definitions.debug["MYDEFINE_CONFIG"] = "MYDEF_DEBUG"
                 tc.preprocessor_definitions.release["MYDEFINE_CONFIG"] = "MYDEF_RELEASE"
+                tc.preprocessor_definitions.debug["MYDEFINEINT_CONFIG"] = 421
```
SSE4, line 42: do we need this? I though you planned to remove that

memsharded: I have not yet fully decided, on one hand it is clear there is demand, on the other, it could be a pure build-system thing, and Conan agnostic about it. In the meantime, it totally makes sense that users want to provide integer preprocessor definitions, as they are very common. Maybe @guiserle could elaborate a bit further about their use case, and why such preprocessor definitions are not integrated in the build system scripts directly.

guiserle: First of all, thanks a lot for considering my request.

My use case is to define configuration values for libraries such as FreeRTOS or MCU SDK packages. Currently we define conan options for such values and pass those values to cmake on conan `build()` method. However that means for every configuration value (and these libraries have lots of them) we need to add logic on both conanfile.py and cmake files.

On top of that we are using Clion as IDE and we need to add these cmake variables (and others such as version numbers) on the IDE project settings manually. I was thinking about setting these values directly on `generate()` method so they would all be included in toolchain file and it would only be necessary to add the toolchain as value for CMAKE_TOOLCHAIN_FILE.

If that is not the intended use case for the generate method and this should be implemented on another way, I will understand it. Also I have considered the implementation of a custom generator for my particular use case if necessary

memsharded: I am going to merge this, seems a valid use case, thanks for the feedback.

Just beware that ``self.options`` are also always strings, even if you defined them as ``option: [1, 2, 3]``, internally, they are always strings. So you might need to convert it to ints, if you use it from there like ``cmake.preprocessor_defintions["MYDEF"] = int(self.options.myoption)``

guiserle: That's great. Thanks for taking my use case into account!

PR #8651, opened 2021-03-15
Continue working in conan local flows [3]

czoido:
Changelog: omit
Docs: omit


conans/test/integration/graph/half_diamond_test.py
```diff
@@ -30,24 +30,20 @@ def config_options(self):
         if export:
             self.client.run("export . lasote/stable")
 
-    def test_reuse(self):
```
memsharded: This probably should remove the check of conaninfot.txt, but keep the test, the installation and the graph seems a valid use case

memsharded (approved):
Great job, can be merged!

PR #8656, opened 2021-03-16
Test different versions of tools

czoido:
Changelog: Feature: Add support for testing with different tools versions.
Changelog: Feature: Add different CMake versions for testing.
Docs: omit

This is just to test if we can add some information to tests to select the version of the tools if it's installed in the CI.

#TAGS: slow

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -1,4 +1,5 @@
 import os
+import platform
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


conans/test/conftest.py
```diff
@@ -73,11 +81,20 @@ def add_tool(request):
     env_tools = dict()
     for mark in request.node.iter_markers():
         if mark.name.startswith("tool_"):
+            version = mark.kwargs.get('version', 'default')
             tool_name = mark.name[5:]
-            if tool_name in tool_locations:
-                add_tools.append(tool_locations[tool_name])
-            if tool_name in tool_environments:
-                env_tools.update(tool_environments[tool_name])
+            if tool_name in tool_locations and tool_locations[tool_name].get(platform.system()):
```
SSE4: the code starts to become complicated and hard to read...
maybe if we extract `tool_locations[tool_name].get(platform.system())` (seems to be a repeating pattern) it will look better? or maybe split it.

memsharded (approved):
Looking good to me, can be moved forward.

SSE4 (approved):

PR #8665, opened 2021-03-17
More fine-grained control (using [conf]) for build parallelization

### LINKED ISSUE #8598: Propose ``cpu_count`` alternative in [conf] for build systems paralelism
For build helpers in ``conan.tools.xxxx``, the parallelism should be configurable via [conf] (and not ``cpu_count``).

Please propose a solution, some guidelines:

- No need that recipes specify it, not explicitly, not as arguments, and they can default back to whatever the build system likes.
- All build systems can share a "parallel" value, no need to do one per build-system/helper
- Might be different to concurrent upload/download

jgsogo:
Changelog: Feature: More fine-grained control (using [conf]) for build parallelization.
Docs: https://github.com/conan-io/docs/pull/2061

This PR introduces the following entries in `conf`:
 * `tools.build:processes`: number of processes to use for every build-helper
 * `tools.gnu.make:jobs`: argument for the `--jobs` parameter when running `make` (overrides the general `tools.build:processes`).
 * `tools.microsoft.msbuild:max_cpu_count`: argument for the `/m` (`/maxCpuCount`) when running `MSBuild` standalone or via CMake (overrides the general `tools.build:processes`).
 * `tools.ninja:jobs`: argument for the `--jobs` parameter when running Ninja generator via CMake or Meson. (overrides the general `tools.build:processes`). 

If none of these configuration options is found, Conan won't add anything to the command line running these tools, it will use whatever it is the default.

close https://github.com/conan-io/conan/issues/8598


conan/tools/microsoft/msbuild.py
```diff
@@ -1,17 +1,26 @@
+from conans.client.tools.oss import cpu_count
 from conans.errors import ConanException
 
 
 def msbuild_verbosity_cmd_line_arg(conanfile):
-    verbosity = conanfile.conf["tools.microsoft"].msbuild_verbosity
+    verbosity = conanfile.conf["tools.microsoft.msbuild"].verbosity
```
jgsogo: @memsharded ,  Is `conf["tools.microsoft"].msbuild_verbosity` intended? I see others like `conf["tools.microsoft.msbuild"].vs_version` and `conf["tools.microsoft.msbuildtoolchain"].compile_options` and I feel like `conf["tools.microsoft.msbuild"].verbosity` is more aligned with them.

SSE4:
IMO, it will be hard to remember the name for the given build helper, as it's different (`jobs` vs `processes` vs `maxCpuCount`)
I guess it should be unified, and all new custom build helper should be following the same practice.

jgsogo:
> IMO, it will be hard to remember the name for the given build helper, as it's different (`jobs` vs `processes` vs `maxCpuCount`)
> I guess it should be unified, and all new custom build helper should be following the same practice.

Sure, this is something to decide. This PR tries to cover both approaches: from the Conan perspective it is unified in the `core.build:processes` variable, but if you want to provide different values for MSBuild, make, Meson,... you can do it using more explicit variables like  `tools.microsoft.msbuild:maxCpuCount`, `tools.gnu.make:jobs`,... where the name of the variable matches the one of the tool.

conan/tools/microsoft/msbuild.py
```diff
@@ -1,17 +1,26 @@
...
+def msbuild_max_cpu_count_cmd_line_arg(conanfile):
+    max_cpu_count = conanfile.conf["tools.microsoft.msbuild"].maxCpuCount or \
+                    conanfile.conf["core.build"].processes or \
+                    cpu_count(output=conanfile.output)
+    return "/m:{}".format(max_cpu_count)
+
+
 class MSBuild(object):
-    def __init__(self, conanfile):
+    def __init__(self, conanfile, parallel=True):
```
memsharded: No arguments for this in recipes, please remove.

conan/tools/gnu/make.py
```diff
@@ -3,11 +3,20 @@
...
+
 from conans.client.build.compiler_flags import build_type_define, libcxx_define
+from conans.client.tools.oss import cpu_count
 from conans.client.tools.oss import detected_architecture, detected_os, get_build_os_arch
 from conans.util.files import save
 
 
+def make_jobs_cmd_line_arg(conanfile):
+    njobs = conanfile.conf["tools.gnu.make"].jobs or \
+            conanfile.conf["core.build"].processes or \
```
memsharded: core.xxxx is internal, it is not exposed to recipes, and cannot be defined in profiles.

Use ``tools.build`` or something like that.

memsharded: This requires some tests.

conan/tools/meson/meson.py
```diff
@@ -5,9 +5,20 @@
 from conans.client.tools.oss import cross_building, cpu_count
 
 
+def ninja_jobs_cmd_line_arg(conanfile, parallel):
+    if parallel:
+        njobs = conanfile.conf["tools.ninja"].jobs or \
+                conanfile.conf["core.build"].processes or \
```
memsharded: Avoid the ``parallel`` argument, it should be implicit from the `conf`. Do not fall back to `cpu_count()`

conan/tools/microsoft/msbuild.py
memsharded: Yes, agree, lets keep ``conf["tools.microsoft.msbuild"].verbosity``. The reason was it didn't exclusively belong to the ``MSBuild`` helper, but I am fine with that.

conan/tools/microsoft/msbuild.py
```diff
@@ -1,17 +1,26 @@
...
-    verbosity = conanfile.conf["tools.microsoft"].msbuild_verbosity
+    verbosity = conanfile.conf["tools.microsoft.msbuild"].verbosity
     if verbosity:
         if verbosity not in ("Quiet", "Minimal", "Normal", "Detailed", "Diagnostic"):
             raise ConanException("Unknown msbuild verbosity: {}".format(verbosity))
         return '/verbosity:{}'.format(verbosity)
 
 
+def msbuild_max_cpu_count_cmd_line_arg(conanfile):
+    max_cpu_count = conanfile.conf["tools.microsoft.msbuild"].maxCpuCount or \
```
memsharded: Better avoid ``maxCpuCount`` CamelCase, lets aim for [conf] variables all lowercase.

conan/tools/microsoft/msbuild.py
jgsogo: 👍   Moved to its own PR: https://github.com/conan-io/conan/pull/8692

memsharded (approved):

jgsogo:
Please, review again. Besides the tests, there are two minor changes to logic: https://github.com/conan-io/conan/pull/8665/commits/0092d4103db8702dd4c034004359fd8b46ebcd92#diff-d72013a45b00a0adf06f4536d6a8c8844461e51b72911937d63e6dda9a3d440a

conan/tools/cmake/cmake.py
```diff
@@ -23,15 +25,27 @@ def _validate_recipe(conanfile):
...
-        elif "Visual Studio" in generator and compiler_version and Version(compiler_version) >= "10":
-            # Parallel for building projects in the solution
-            args.append("/m:%i" % cpu_count(output=conanfile.output))
-
-    if generator and "Visual Studio" in generator:
+    if not generator:
+        return args
+
+    # Arguments related to parallel
+    if "Makefiles" in generator and "NMake" not in generator and parallel:
```
SSE4: maybe check for parallel just once and use nested cond? we're checking 3 times right now

conans/test/unittests/tools/microsoft/test_msbuild.py
```diff
@@ -0,0 +1,27 @@
+import textwrap
+
+from conan.tools.microsoft import MSBuild
+from conans.model.conf import ConfDefinition
+from conans.test.utils.mocks import ConanFileMock, MockSettings
+
+
+def test_meson_build():
```
memsharded:
```suggestion
def test_msbuild_cpu_count():
```

memsharded (approved):

PR #8666, opened 2021-03-17
further minor improvements to CMakeDeps

memsharded:
Changelog: Fix: Simplifications to ``CMakeDeps`` generator to remove legacy code.
Docs: omit


lasote (approved):
We should copy/transform tests from the find package multi to CMakeDeps, is low tested specially on non-windows machines.

danimtb (approved):

PR #8670, opened 2021-03-18
Continue working in conan local flows [4]

czoido:
Changelog: Remove: Remove _txt_ generator and all references and uses of _conanbuildinfo.txt_.
Docs: omit


memsharded (approved):
Great job

PR #8682, opened 2021-03-22
Feature: MSBuildDeps generator: CAExcludePath

### LINKED ISSUE #8633: [feature] MSBuildDeps generator: CAExcludePath
In most cases if one wants to run the code analysis included in Visual Studio, one is not interested in findings in external code.

This is possible via the (poorly documented) `CAExcludePath` property, see e.g. https://stackoverflow.com/questions/43755499/suppress-warnings-for-external-headers-in-vs2017-code-analysis.

It would be nice if the `MSBuildDeps` generator could populate the `CAExcludePath` with the include paths of the dependencies.

Depending on how well the `visual_studio`/`visual_studio_multi` generators are still maintained, consider adding it there too.

Note that the few examples usages I found on the internet with a quick search seem to have the property under a `<PropertyGroup Condition="'$(Language)'=='C++'">` group.
A quick test suggest this is not needed, e.g. if I manually modify `conan_boost.props` to
```xml
...
<PropertyGroup>
	<LocalDebuggerEnvironment>PATH=%PATH%;$(ConanboostBinaryDirectories)$(LocalDebuggerEnvironment)</LocalDebuggerEnvironment>
	<DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
	<CAExcludePath>$(ConanboostIncludeDirectories);$(CAExcludePath)</CAExcludePath>
</PropertyGroup>
...
```
the exclude seems to works fine.

SSE4:
closes: #8633 

Changelog: Feature: MSBuildDeps generator: CAExcludePath
Docs: TODO


conan/tools/microsoft/msbuilddeps.py
```diff
@@ -84,6 +88,14 @@ def generate(self):
...
         # TODO: This is duplicated in write_generators() function, would need to be moved
         # to generators and called from there
+        exclude_code_analysis = self._conanfile.conf["tools.microsoft.msbuilddeps"].exclude_code_analysis
+        if exclude_code_analysis:
+            if isinstance(exclude_code_analysis, str):
+                self.exclude_code_analysis = exclude_code_analysis.split(";")
+        if self.exclude_code_analysis in [["True"], True]:
+            self.exclude_code_analysis = ["*"]
+        elif self.exclude_code_analysis in [["False"], False]:
+            self.exclude_code_analysis = None
```
memsharded: By definition exclude_code_analysis is always ``None`` or a string, this seems could be greatly simplified, maybe something like this:

```suggestion
        exclude_code_analysis = self._conanfile.conf["tools.microsoft.msbuilddeps"].exclude_code_analysis
        if exclude_code_analysis is not None:
                self.exclude_code_analysis = eval(exclude_code_analysis)
                assert isinstance(self.exclude_code_analysis, (bool, list))
```

The idea is not to reinvent a syntax to create a list of things, just use python syntax and eval()

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -164,7 +183,8 @@ def add_valid_ext(libname):
             'definitions': "".join("%s;" % d for d in cpp_info.defines),
             'compiler_flags': " ".join(cpp_info.cxxflags + cpp_info.cflags),
             'linker_flags': " ".join(cpp_info.sharedlinkflags),
-            'exe_flags': " ".join(cpp_info.exelinkflags)
+            'exe_flags': " ".join(cpp_info.exelinkflags),
+            'ca_exclude_dirs': '$(Conan%sIncludeDirectories)' % name if exclude_code_analisys else ''
```
memsharded:
```suggestion
            'ca_exclude_dirs': '$(Conan{}IncludeDirectories)'.format(name) if exclude_code_analysis else ''
```

Typo in ``exclude_code_analysis``. Also the ``.format()`` helps dissambiguate precedence of % name vs if

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -25,6 +26,7 @@ class MSBuildDeps(object):
             <Conan{name}BinaryDirectories>{bin_dirs}</Conan{name}BinaryDirectories>
             <Conan{name}Libraries>{libs}</Conan{name}Libraries>
             <Conan{name}SystemDeps>{system_libs}</Conan{name}SystemDeps>
+            <Conan{name}CAExcludeDirectories>{ca_exclude_dirs}</Conan{name}CAExcludeDirectories>
```
memsharded: This is not new "data" definition, shouldn't be defined here, but just in the group below, as an ``if ca_exclude_path``.

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -42,6 +44,7 @@ class MSBuildDeps(object):
           <PropertyGroup>
             <LocalDebuggerEnvironment>PATH=%PATH%;$(Conan{name}BinaryDirectories)$(LocalDebuggerEnvironment)</LocalDebuggerEnvironment>
             <DebuggerFlavor>WindowsLocalDebugger</DebuggerFlavor>
+            <CAExcludePath>$(Conan{name}CAExcludeDirectories);$(CAExcludePath)</CAExcludePath>
```
memsharded: Put the conditional here, to define it or not (by default, this line shouldn't appear)

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -152,6 +164,13 @@ def add_valid_ext(libname):
             ext = os.path.splitext(libname)[1]
             return '%s;' % libname if ext in VALID_LIB_EXTENSIONS else '%s.lib;' % libname
 
+        exclude_code_analisys = False
+        if self.exclude_code_analysis:
+            for pattern in self.exclude_code_analysis:
+                if fnmatch.fnmatch(name, pattern):
+                    exclude_code_analisys = True
+                    break
+
```
memsharded:
```suggestion
        exclude_code_analysis = False
        if isinstance(self.exclude_code_analysis, list):
              for pattern in self.exclude_code_analysis :
                  if fnmatch.fnmatch(name, pattern):
                      exclude_code_analysis = True
                      break
        else:
             exclude_code_analysis = self.exclude_code_analysis
```

Lets leverage the ``bool`` type, not convert to ``*`` pattern analyze the pattern to retrieve a new ``True``.

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -84,6 +88,10 @@ def generate(self):
         # config_filename
         # TODO: This is duplicated in write_generators() function, would need to be moved
         # to generators and called from there
+        exclude_code_analysis = self._conanfile.conf["tools.microsoft.msbuilddeps"].exclude_code_analysis
+        if exclude_code_analysis is not None:
+                self.exclude_code_analysis = eval(exclude_code_analysis)
+                assert isinstance(self.exclude_code_analysis, (bool, list))
```
SSE4: @memsharded I've applied your suggestions, but now, string pattern doesn't work any longer, as you accept only bool and list here.

conan/tools/microsoft/msbuilddeps.py
SSE4: why do I need condition here? `$(Conan{name}CAExcludeDirectories)` will just resolve to an empty string, so line will have no effect.

conan/tools/microsoft/msbuilddeps.py
SSE4: not sure I follow - it uses the same structure as another definitions...

memsharded: This section contains the pure translation of ``cpp_info`` fields, nothing more. 
The custom logic of this generator is not defined in this section, but somewhere else.

memsharded (commented):
I have pushed some changes

conan/tools/microsoft/msbuilddeps.py
SSE4: got it, then I agree it shouldn't be there

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -75,6 +80,13 @@ def __init__(self, conanfile):
                          'x86_64': 'x64'}.get(str(conanfile.settings.arch))
         # TODO: this is ugly, improve this
         self.output_path = os.getcwd()
+        # ca_exclude section
+        self.exclude_code_analysis = None
+        ca_exclude = self._conanfile.conf["tools.microsoft.msbuilddeps"].exclude_code_analysis
+        if ca_exclude is not None:
+            # TODO: Accept single strings, not lists
+            self.exclude_code_analysis = eval(ca_exclude)
```
SSE4, line 88: are we really okay if `eval` usage? it seems to be pretty dangerous, as it may allow arbitrary code execution:
https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html

memsharded: Yes, as well as every Conan recipe, how this would be more dangerous than Conan python evaluation?

SSE4: maybe we can avoid eval then? if use just a regexp, if regexp matches nothing, it acts effectively as disabled matching

memsharded: Problem is that all the matches we have in Conan for matching packages references are fnmatch based. Introducing now regex for this would be very asymmetric. And for fnmatch, we should probably allow multiple patterns, as it is not as expressive.

memsharded: Lets merge it, but keep it not documented while we finalize the UX, but there are some template changes that are useful for other PRs.

PR #8685, opened 2021-03-22
Relative profile path should be valid for creation commands

### LINKED ISSUE #8678: [question] profile search order
From some testing, it seems conan tries to resolve relative paths to profiles in the following order:

1. resolve from `.conan/profiles`
2. resolve from current directory

I'm a bit confused why this order was chosen, as it doesn't seem to be intuitive: Why try the more general solution before the more specific one?

---

I stumbled upon this when developing some profiles to later be shared and used with `conan config install`. One referenced profile from the local directory structure (path scheme: `../profiles/xyz`) caused some confusion, as there was another profile installed at `.conan/profiles/xyz` and the latter was chosen instead.

uilianries:
Conan accepts absolute and relative paths for profiles, that's well [documented](https://docs.conan.io/en/latest/reference/profiles.html).

We know the main recommendation is absolute, but relative path is not working well when is associated to same tree level, and profile names is same from default profile folder. If a default profile doesn't exist, it works well, but when duplicated, Conan prefers default profiles folder instead.

fixes #8678

Changelog: Fix: Accept relative profile path when folder is on same tree level.
Docs: https://github.com/conan-io/docs/pull/2049


conans/test/functional/configuration/profile_test.py
```diff
@@ -501,6 +501,77 @@ def config(self):
         self.assertIn('''Requires:
         WinRequire/0.1@lasote/stable''', self.client.out)
 
+    def test_install_profile_absolute_path(self):
```
memsharded: This feature should not require a ``functional`` test. Those are the heaviest and slowest, and using external tools like cmake or actually compiling.

Do at most a ``integration`` test, and it sounds it should be much sorter

conans/client/profile_loader.py
```diff
@@ -90,7 +90,9 @@ def valid_path(_profile_path):
     if os.path.isabs(profile_name):
         return valid_path(profile_name)
 
-    if profile_name[:2] in ("./", ".\\"):  # local
+    # relative local profile path
+    if profile_name[:2] in ("./", ".\\") or \
```
SSE4: why not use `os.path.isabs` instead of our own bicycles?
that will end up in many misdetected cases, e.g. in posix `..\\myfile` is valid file name

uilianries: Yes, makes total sense, I gonna improve it. Thanks.

conans/test/functional/configuration/profile_test.py
uilianries: Yes, it's possible running a simple unit test, because is very related to an internal method. Command arguments are safe for this case.

conans/client/profile_loader.py
uilianries: Done.

conans/test/functional/configuration/profile_test.py
uilianries: Done.

conans/test/unittests/client/profile_loader/profile_loader_test.py
```diff
@@ -22,336 +20,416 @@ def create_profile(folder, name, settings=None, package_settings=None, env=None,
...
+        """)
+    save(os.path.join(tmp, "profile2.txt"), profile2)
+    profile, variables = read_profile("./profile2.txt", tmp, None)
+
+    assert variables == {"MYVAR": "fromProfile2",
+                         "PROFILE_DIR": tmp.replace('\\', '/')}
+    assert profile.settings["os"] == "fromProfile2"
+
+
+def test_profile_load_absolute_path():
```
uilianries, line 360: New tests start here. All other changes are related to pytest. You can check each commit separately, I didn't mix pytest with new test, each commit follows a single change.

memsharded (commented):
It is great the modernizing of tests to pytest, thanks for taking the time!,

But on the other hand it is practically impossible to see if something might be breaking. Could you please split this into 2 PRs, one pure-refactor of tests, nothing else changing, and then the new fix with the new tests covering it? 

Thanks!

uilianries:
Yes, totally. Indeed is terrify for review.

uilianries:
@memsharded Done

uilianries:
Second PR: https://github.com/conan-io/conan/pull/8729

memsharded (changes requested):
This is quite risky:

- I have reverted the changes to ``profile_loader.py``, and the new tests are green! This means the tests are not testing this change, but just the previous existing functionality.
- It is likely that we are breaking something, the condition might change for paths like ``relative/path/to/profile.txt`` that do not start with ``.``
- It is clear that the different paths functionality is not sufficiently covered by tests, we would need more thorough testing to make sure we are not breaking.

So I think we cannot merge these changes as-is.

uilianries:
> I have reverted the changes to profile_loader.py, and the new tests are green! This means the tests are not testing this change, but just the previous existing functionality.

Indeed is tricky, I needed to add `profiles` folder to reproduce same behavior. The commit [6b8de0d35d2](https://github.com/conan-io/conan/pull/8685/commits/6b8de0d35d279e5345a87ca94b4606d2aca97a64) should reproduce that error.

> It is likely that we are breaking something, the condition might change for paths like relative/path/to/profile.txt that do not start with .

The fragility is because we need to follow not equal profile names, but also profiles folder.

uilianries:
Okay, unit test working as spectated: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-8685/8/pipeline/#step-278-log-1537

conans/test/functional/configuration/profile_test.py
```diff
@@ -635,3 +635,13 @@ def test_profile_crazy_inheritance(self):
                              compiler.runtime=MD
                              compiler.version=15
                              os=Windows"""), self.client.out)
+
+
+def test_profile_from_cache_path():
```
memsharded, line 640: @uilianries please check this new test that I have provided. This would be proving that the previous fix was breaking, very likely massively breaking many users (all of them that manages profiles in subfolders in the cache).

The new proposed fix is more conservative and passes your new tests, please check if there is some other case to take into account.

uilianries: I would love adding a functional test to validate relative paths too. I know those unit tests cover very well `get_profile_path()`, but I would like having and end-to-end, because we have a considerable stack before calling it.

stacktrace


For regression it can help us. WDYT?

memsharded: Actually this should go to ``integration``, not functional (as they don't depend on external tools), but until this is fixed, yes, it can co here.

Yes, it makes sense to have a few more tests here to cover better this functionality, please add some. Try to keep them simple, as the proposed one. Thanks!

conans/test/functional/configuration/profile_test.py
```diff
@@ -635,3 +635,60 @@ def test_profile_crazy_inheritance(self):
...
+    """ When passing relative ../path as profile file, it MUST be used
+        conan install . -pr=../profiles/default
+        /tmp/profiles/default MUST be consumed as target profile
+    """
+    folder = temp_folder(False)
+    client = TestClient()
+    client.save({os.path.join(folder, "profiles", "default"): "[settings]\nos=AIX",
+                 os.path.join(folder, "current", "conanfile.txt"): ""})
+    with client.chdir(os.path.join(folder, "current")):
+        client.run("install . -pr=../profiles/default")
```
memsharded:
```suggestion
    client = TestClient()
    client.save({"profiles/default": "[settings]\nos=AIX",
                      "current/conanfile.txt": ""})
    with client.chdir("current"):
        client.run("install . -pr=../profiles/default")
```

You can simplify the tests by using the TestClient current folder which is already a temporary one, intended for these cases.

memsharded (commented):
Please try the proposed simplification to tests.

conans/test/functional/configuration/profile_test.py
uilianries: True, I forgot client.current_folder != client.cache_folder. Thanks

memsharded (approved):

memsharded:
I have fixed the broken test in https://github.com/conan-io/conan/pull/8685/commits/8295aefbbb762f6d59b3fa229155f43f980732d9. It was incorrectly managing paths, and it was creating folders in the code folder repo, making it dirty (and with risk to commit them accidentally)

PR #8697, opened 2021-03-24
Added existing tests to CMakeDeps and fixed some things about it

lasote:
Changelog: omit
Docs: omit

### DIFF
--- a/conan/tools/cmake/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps.py
@@ -627,10 +627,10 @@ endforeach()
             save(generator_file, content)
 
     def _data_filename(self, pkg_filename):
-        data_fname = "{}-".format(pkg_filename)
+        data_fname = "{}-{}".format(pkg_filename, self.configuration.lower())
         if self.arch:
-            data_fname += "{}-".format(self.arch)
-        data_fname += "{}-data.cmake".format(self.configuration)
+            data_fname += "-{}".format(self.arch)
+        data_fname += "-data.cmake"
         return data_fname
 
     @property
--- a/conan/tools/cmake/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps.py
@@ -267,7 +267,7 @@ class CMakeDeps(object):
 
         # Load the debug and release library finders
         get_filename_component(_DIR "${{CMAKE_CURRENT_LIST_FILE}}" PATH)
-        file(GLOB DATA_FILES "${{_DIR}}/{filename}-*-*-data.cmake")
+        file(GLOB DATA_FILES "${{_DIR}}/{filename}-*-data.cmake")
 
         foreach(f ${{DATA_FILES}})
             include(${{f}})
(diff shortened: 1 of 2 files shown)


conans/test/functional/generators/components/test_components_cmake_generators.py
```diff
@@ -50,12 +50,16 @@ def package_info(self):
                 elif self.options.components == "custom":
                     self.cpp_info.filenames["cmake_find_package_multi"] = "MYG"
                     self.cpp_info.filenames["cmake_find_package"] = "MYG"
+                    self.cpp_info.filenames["CMakeDeps"] = "MYG"
```
memsharded: This might not be necessary, there is an adaptor to use ``cmake_find_package_multi`` definition in ``CMakeDeps`` too (necessary for painless migration)

memsharded (approved):

conan/tools/cmake/cmakedeps.py
```diff
@@ -626,6 +626,13 @@ def generate(self):
             generator_file = os.path.join(self.output_path, generator_file)
             save(generator_file, content)
 
+    def _data_filename(self, pkg_filename):
+        data_fname = "{}-{}".format(pkg_filename, self.configuration.lower())
+        if self.arch:
```
danimtb, line 631: This is just a detail: If we are already using build_type `none` when the build_type does not exist, why not doing the same with the architecture and follow the same filename pattern?

conans/test/functional/generators/components/test_components_cmake_generators.py
```diff
@@ -286,17 +289,22 @@ def test_standard_names(setup_client_with_greetings, generator):
 
 @pytest.mark.parametrize("generator", ["cmake_find_package_multi", "cmake_find_package"])
 def test_custom_names(setup_client_with_greetings, generator):
+    # NOTE: cpp_info.names is deprecated for CMakeDeps and doesn't work anymore so this test doesn't
```
danimtb: What does this mean? Are things like `self.cpp_info.names["CMakeDeps"] = "MyChat"` not allowed?

conan/tools/cmake/cmakedeps.py
lasote: That's a good observation. I didn't realize that `none` was a valid pattern. @memsharded what do you prefer?
I see a bit confusing to convert a None to str. If we want that behavior I would do it more explicit.

conans/test/functional/generators/components/test_components_cmake_generators.py
lasote: I wrote that and later at night I thought... it doesn't make sense. It should work, probably some bug. I'm checking

conan/tools/cmake/cmakedeps.py
lasote: I just realized: If the build type is not declared and you do: `str(self._conanfile.settings.build_type)` it will crash, right? I think the `build_type` is right now mandatory in this generator as it was in the `cmake_find_package_multi`

memsharded: build_type ``none`` might be undefined behavior, I don't think the generator expressions in the targets etc will handle this correctly. We need to figure out the multiplicity (build_type, arch) of the generator, and how they connect the producer build_type and the consumer build_type.


So I suggest, leave it as-is, whatever it passes current tests is already a big improvement, and we will address this later.

danimtb (approved):

Issue #5415, opened 2019-06-27
conanfile.py is silently deleted by "conan create" in local "source" cache


climblinne:
To help us debug your issue please explain:

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've specified the Conan version, operating system version and any tool that can be relevant.
- [x] I've explained the steps to reproduce the error or the motivation/use case of the question/suggestion.

I have Conan 1.16.1 installed on Windows 10.

I used the same example as in https://github.com/conan-io/conan/issues/4013.

Clone the test repo:
```
git clone https://github.com/db4/conan-scm_test.git
```
Execute

```
conan create . demo/testing
```

When looking into the source folder in the local cache, I get the following output of `git status`:

```bash
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        deleted:    conanfile.py

no changes added to commit (use "git add" and/or "git commit -a")
```

So the removing of the `conanfile.py` leads for our automatic version numbering detection (based on `git describe --dirty`) to an dirty working copy.
What is the reason to remove the `conanfile.py` from the git source folder? All other files in our packages are always still there.

jgsogo:
As you can see [here](https://github.com/conan-io/conan/blob/b6840ccfb53c1d0a38eec9f75df5caa8fd8aad5d/conans/client/source.py#L244) we are not sure about it, we have marked it to review for Conan 2.0, but as this issue has arised, we should investigate if it is possible to keep the `conanfile.py` in the source directory without breaking any workflow.

Nevertheless, any dynamic value in your recipe should be resolved during the `export` stage, can I see that `conanfile.py` to understand why are you calling `git` in the sources folder and check if it is possible to resolve that value during the export.

Thanks

climblinne:
For automatic builds it would be fine to just generate the version number by Conan. The problem is, that when somebody build it locally, the version number would stay at the "timeslot", when we call `conan install`. So no detection for a dirty version locally. That's the reason, why we always have a prebuild script in our main projects, which check the current status of the git repo and generate a version header.

jgsogo:
> That's the reason, why we always have a prebuild script in our main projects, which check the current status of the git repo and generate a version header.

[...]

PR #8700, opened 2021-03-24
#5415: removed a seemingly spurious deletion of conanfile.py during t…

lexknuther:
#5415: removed a seemingly spurious deletion of conanfile.py during the source method.  This prevented projects using cmake-conan and conan_cmake_install(PATH_OR_REFERENCE conanfile.py) as it deleted the project's conanfile.py before cmake-conan could used it.

Changelog: (Feature | Fix | Bugfix): Bugfix
Docs: https://github.com/conan-io/docs/pull/5415

### DIFF
--- a/conans/client/source.py
+++ b/conans/client/source.py
@@ -159,7 +159,7 @@ def _run_source(conanfile, conanfile_path, hook_manager, reference, cache,
 
 def _clean_source_folder(folder):
     for f in (EXPORT_TGZ_NAME, EXPORT_SOURCES_TGZ_NAME, CONANFILE+"c",
-              CONANFILE+"o", CONANFILE, CONAN_MANIFEST):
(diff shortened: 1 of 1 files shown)


lexknuther:
Looking at the failing unit tests, they seem to all be just that some list also now includes conanfile.py.

matthew-reid:
@lexknuther Any idea why this was closed or whether there is a work around? I have the same problem, I need to CMake install my conanfile.py (so that I have a locally installed package I can use with `conan editable`), but the conanfile.py is silently deleted.

memsharded:
@matthew-reid if you can specify with more detail what are you trying, that could help (better as a new issue). Are you using the ``layout()`` method? Are you using the modern integrations? With those it is not necessary at all to run a cmake install to have the package locally, and it can be used from the build folders.

Issue #8145, opened 2020-12-02
[question] find_package with components behavior


ericriff:
I've recently started using the (still under development) recipe of boost which introduces components support and I've noticed a strange behavior when we use `findBoost.cmake` when we specify the components we want instead of looking for the whole boost. Maybe it is the normal behavior, I'm not sure.
The "issue" that I'm seeing is that if you ask `find_package()` to provide you with some of the boost components, the log makes it look like conan still looks for all of them. This a portion of the `test_package` of said recipe:

```
    if(WITH_RANDOM)
        find_package(Boost COMPONENTS random REQUIRED)
        add_executable(random_exe random.cpp)
        target_link_libraries(random_exe PRIVATE Boost::random)
    endif()
```

So I would expect `FindBoost.cmake` to give me just `boost.random`. The log shows:

```
-- Conan: Using autogenerated FindBoost.cmake
-- Found Boost: 1.74.0 (found version "1.74.0") 
-- Conan: Component 'random' found in package 'Boost'
-- Library boost_program_options found /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_program_options.a
-- Found: /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_program_options.a
-- Library boost_stacktrace_noop found /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_stacktrace_noop.a
-- Found: /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_stacktrace_noop.a
-- Library boost_wserialization found /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_wserialization.a
-- Found: /home/eric/brewst/.conan/data/boost/1.74.0/eriff-dev/stable/package/0fdc43fcd9ad9cbd2530f4b832dbf5f2bb7ec3d1/lib/libboost_wserialization.a
[...]

PR #8715, opened 2021-03-27
Reduce CMake spam

klimkin:
Changelog: Feature: Use CMAKE_MESSAGE_LOG_LEVEL to control the amount of output of from cmake files

Note, `CMAKE_MESSAGE_LOG_LEVEL` requires CMake 3.17. For older versions the messages will be dropped. It's possible to have  make a workaround in `conan_message()` to convert the type to `STATUS` for any older version.
https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_LOG_LEVEL.html

Solves https://github.com/conan-io/conan/issues/8145

From this:
~~~
-- Conan: Using autogenerated FindSystemCLanguage.cmake
-- Found SystemCLanguage: 2.3.3 (found version "2.3.3")
-- Library systemc found .conan/data/systemc/2.3.3/syssim/stable/package/ecec4cd1dc0d521a95bc6a23f7972092d3c15d7f/lib/libsystemc.so
-- Found: .conan/data/systemc/2.3.3/syssim/stable/package/ecec4cd1dc0d521a95bc6a23f7972092d3c15d7f/lib/libsystemc.so
-- Library systemc found .conan/data/systemc/2.3.3/syssim/stable/package/ecec4cd1dc0d521a95bc6a23f7972092d3c15d7f/lib/libsystemc.so
-- Found: .conan/data/systemc/2.3.3/syssim/stable/package/ecec4cd1dc0d521a95bc6a23f7972092d3c15d7f/lib/libsystemc.so
-- Conan: Using autogenerated FindGTest.cmake
-- Found GTest: 1.10.0 (found version "1.10.0")
-- Library gtest_main found .conan/data/gtest/1.10.0/_/_/package/00673baad6d72163dd851981a9d73649926a72ff/lib/libgtest_main.a
-- Found: .conan/data/gtest/1.10.0/_/_/package/00673baad6d72163dd851981a9d73649926a72ff/lib/libgtest_main.a
-- Library gmock_main found .conan/data/gtest/1.10.0/_/_/package/00673baad6d72163dd851981a9d73649926a72ff/lib/libgmock_main.a
-- Found: .conan/data/gtest/1.10.0/_/_/package/00673baad6d72163dd851981a9d73649926a72ff/lib/libgmock_main.a
...
~~~

To this:
~~~
-- Found SystemCLanguage: 2.3.3 (found version "2.3.3")
-- Found GTest: 1.10.0 (found version "1.10.0")
~~~


solvingj:
It's definitely a nice thing to control.  It should probably be configurable via `[conf]`.

memsharded (commented):
Conan 2.0 will provide support for CMake 3.15, and current Conan 1.X supports in some cases down to CMake 2.8, and in the find_package generator, something like 3.3 I think is the supported case.

Changes done need to take this into account, please don't change if it will not work properly in older CMake versions. Everything in ``conan.tools.xxxx`` can be considered 2.0-oriented, and thus can contain CMake 3.15 only features, but other generators like ``cmake_find_package`` must guarantee backwards compatibility with older CMake versions.

solvingj:
@memsharded comments on using `[conf]`?

memsharded (changes requested):
Hi @klimkin 

If you would like to move this forward:

- This cannot rely on https://cmake.org/cmake/help/latest/variable/CMAKE_MESSAGE_LOG_LEVEL.html, because that is from cmake 3.17.
- New integrations in ``conan.tools.cmake`` will need to support CMake 3.15
- Do not touch the previous integrations, as they need to support even older CMake versions, and will be removed in Conan 2.0, not worth the risk.

I understand that probably not worth to do a custom solution for this, or too much effort, please tell.

memsharded:
Ok, as this is a bit risky (depends on CMake>3.17, which is breaking for many users), cannot be merged. So I propose the following plan:

- Lets close this PR, it is a bit outdated and conflicted.
- Conan 2.0 is now refactoring the api and the output. The output will include logging and verbosity levels.
- Try to align the default verbosity levels of build systems with the Conan new levels
- Possibly propose a possibility of setting build-system logging levels (general ones, not only CMake), with different values than the Conan logging levels
- Do this only for the new generators, better not touch the legacy ones, not worth the risk

In short, we like the feature and we will implement i , (without modifying the legacy ones, only for the new). Thanks very much for your contribution

klimkin:
That's fine, thank you for reviewing this. Currently, there is a middle-ground solution to suppress spam using `set(CONAN_CMAKE_SILENT_OUTPUT TRUE)`. So you can keep focus to get better logging in 2.0.

PR #8716, opened 2021-03-27
Use find_package_handle_standard_args() to handle CMake components

klimkin:
- Use standard CMake macro to handle and report components
- Only "REQUIRED COMPONENTS" should result in CMake error

For `cmake_find_package` generator this changes output from:
~~~
-- Conan: Component 'gtest_main' found in package 'GTest'
~~~
To
~~~
-- Found GTest: 1.10.0 (found version "1.10.0") found components: gtest_main
~~~

Related to https://github.com/conan-io/conan/issues/8145

Changelog: Bugfix: Use find_package_handle_standard_args() to handle CMake components


memsharded (commented):
Though this is looking good, I think I'd prefer to do changes to ``CMakeDeps`` only and leave without modification ``cmake_find_package[_multi]`` generators, if there is any risk of breaking users. If there is a small chance of breaking other users (older CMake versions, down to 3.1, non fatal errors at the moment that would be fatal errors now, etc), then better do changes that are only strictly necessary, like bug fixes.

memsharded:
Hi @klimkin 

It seems that this PR got stale and outdated, sorry for not being able to push it back then.
I am closing this PR, if you still consider that it makes sense to make these changes, please let us know (they should be targeted to the ``release/2.0`` branch, and to ``CMakeDeps`` exclusively)

PR #8718, opened 2021-03-28
Feature: allow arbitrary defines in conan new templates

SSE4:
allow arbitrary defines in `conan new` templates

e.g. instead of running:
```
$ conan new zip/0.1.30 -m cci.cmake
```
I may run:
```
conan new zip/0.1.30 -m cci.cmake -d license=Unlicense
```
which allows further automation on top of `conan new`, for instance, some script may fetch license, description, topics, homepage, tarball URL and sha256 in order to pass em to the template.

Changelog: Feature: Allow arbitrary defines in `conan new` templates.
Docs: https://github.com/conan-io/docs/pull/2051

### DIFF
--- a/conans/client/cmd/new.py
+++ b/conans/client/cmd/new.py
@@ -286,16 +286,17 @@ test_package/build
 """
 
 
-def _render_template(text, name, version, package_name):
+def _render_template(text, name, version, package_name, defines):
     context = {'name': name,
                'version': version,
                'package_name': package_name,
                'conan_version': client_version}
+    context.update(defines)
     t = Template(text, keep_trailing_newline=True)
     return t.render(**context)
 
 
-def _get_files_from_template_dir(template_dir, name, version, package_name):
+def _get_files_from_template_dir(template_dir, name, version, package_name, defines):
     files = []
     for d, _, fs in os.walk(template_dir):
         for f in fs:
@@ -306,9 +307,10 @@ def _get_files_from_template_dir(template_dir, name, version, package_name):
     out_files = dict()
     for f in files:
         f_path = os.path.join(template_dir, f)
-        rendered_path = _render_template(f, name=name, version=version, package_name=package_name)
+        rendered_path = _render_template(f, name=name, version=version, package_name=package_name,
+                                         defines=defines)
         rendered_file = _render_template(load(f_path), name=name, version=version,
-                                         package_name=package_name)
+                                         package_name=package_name, defines=defines)
         out_files[rendered_path] = rendered_file
 
     return out_files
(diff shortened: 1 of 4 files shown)


conans/client/cmd/new.py
```diff
@@ -347,6 +349,8 @@ def cmd_new(ref, header=False, pure_c=False, test=False, exports_sources=False,
         raise ConanException("'template' is incompatible with 'header', "
                              "'sources', 'pure-c' and 'bare'")
 
+    defines = defines or []
+    defines = dict((n, v) for n, v in (d.split('=') for d in defines))
```
memsharded: Better pass a dict, don't need to parse the inputs so inside, better do it in command.py and pass a dict to the api.

conans/client/command.py
```diff
@@ -199,6 +199,7 @@ def new(self, *args):
                             help='Generate a .gitignore with the known patterns to excluded')
         parser.add_argument("-ciu", "--ci-upload-url",
                             help='Define URL of the repository to upload')
+        parser.add_argument('-d', '--define', action='append')
```
memsharded, line 202: Sure this doesn't need the ``Extender`` custom action?

memsharded (changes requested):
Needs tests, please add.

conans/client/cmd/new.py
SSE4: agree, I suppose API method should already accept the dict, so conversion should happen earlier, just after the command line parsing

conans/client/command.py
SSE4: sorry, I am not sure I understand, what do you mean... could you possibly describe your question in more details, e.g. with some examples?

conans/client/cmd/new.py
SSE4: moved

SSE4:
@memsharded added test

conans/client/command.py
memsharded: We have Extender class in our command.py file: https://github.com/conan-io/conan/blob/beaca4195d700c944f0f7b098e267a509f78acb4/conans/client/command.py#L34

It is basically there to implement the ``append`` action it seems. Is the Extender class legacy code not needed? If that is the case maybe we can get rid of it? (Not in this PR, of course, in another one, but just wondering if the ``action='append'`` could have some limitation or problem)

memsharded (approved):

conans/client/command.py
SSE4: seems so, at least `append` exists in 2.7, so `Extender` seems to be a redundant bicycle for me, could be removed in conan 2.0

memsharded: If it is really redundant, then let's drop it today, for next 1.35? Of course if we are relatively sure this will not be breaking users, but if we are, no need to wait until Conan 2.0.

This looks good, then, merging.

PR #8719, opened 2021-03-28
creating conanvcvars file for toolchains

memsharded:
Changelog: Feature: Use ``conancvvars.bat`` file for Meson toolchain
Docs: Omit

### DIFF
--- a/conan/tools/meson/meson.py
+++ b/conan/tools/meson/meson.py
@@ -1,7 +1,6 @@
 import os
 
 from conan.tools.meson import MesonToolchain
-from conan.tools.microsoft.visual import vcvars_command, vcvars_arch
 from conans.client.tools.oss import cross_building
 
 
@@ -16,13 +15,6 @@ class Meson(object):
     def __init__(self, conanfile, build_folder='build'):
         self._conanfile = conanfile
         self._build_folder = build_folder
-
-    def _run(self, cmd):
-        if self._conanfile.settings.get_safe("compiler") == "Visual Studio":
-            vcvars = vcvars_command(self._conanfile.settings.get_safe("compiler.version"),
-                                    vcvars_arch(self._conanfile))
-            cmd = '%s && %s' % (vcvars, cmd)
-        self._conanfile.run(cmd)
 
     @property
     def _build_dir(self):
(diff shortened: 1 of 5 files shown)


conan/tools/meson/meson.py
```diff
@@ -44,7 +36,8 @@ def configure(self, source_folder=None):
         cmd += ' "{}" "{}"'.format(self._build_dir, source)
         if self._conanfile.package_folder:
             cmd += ' -Dprefix="{}"'.format(self._conanfile.package_folder)
-        self._run(cmd)
+        vcvars = os.path.join(self._conanfile.install_folder, "conanvcvars")
```
SSE4, line 39: code is repeating 3 times, maybe move it inside `self._run` instead?

SSE4:
are we sure `install_folder` is always available? I remember it was missing for some cases, e.g. `test_package` and magic `root` package

conan/tools/meson/meson.py
memsharded: ``self._run`` might use different locations of environment scripts, depending if they are from Conan or from user ones.

SSE4: and how do we know this locations?

memsharded: If it is a generated file, it will be by definition in the ``install_folder``. 
If it is a user file, the user will specify its location.

SSE4: but right now it uses `self._conanfile.install_folder` always, so these 2 lines are repeated.

memsharded: but if we put it inside ``self._run``, it will try to apply it to all the environment scripts, which is incorrect

SSE4: I mean doing:
```
    def _run(self, cmd):
       vcvars = os.path.join(self._conanfile.install_folder, "conanvcvars")
        self._conanfile.run(cmd, env=["conanbuildenv", vcvars])
```
are there other environment scripts in meson toolchain? I can't spot any right now

memsharded: Oh, I see what you mean, I was confusing it with ``self._conanfile.run()``. I put some TODOs and some comments there regarding the ``install`` or ``test`` that I didn't know if they really need the vcvars environment at all. If they need it, then I guess it is ok, if not, each method should define its own environment.

SSE4: not sure, probably best add vcvars everywhere, just in case install needs tools from vcvars (e.g. dumpbin or whatever)

memsharded: In any case, it is a small detail, can be refactored later, I am merging this, I prefer to have this in the next release.

PR #8724, opened 2021-03-29
Feature: premake_multi generator

### LINKED ISSUE #6634: [feature] Provide a 'premake_multi' generator
I'm trying to use the new experimental premake generator with premake5 based on these guides:
https://docs.conan.io/en/latest/reference/generators/premake.html
https://docs.conan.io/en/latest/integrations/build_system/premake.html#premake

But I can't figure out how to hook up multiple configurations in premake.

If I use `conan install . -s built_type=Debug -if ./debug` to install Debug/Release builds into separate directories, then I end up with two `conanbuildinfo.premake.lua` files in two separate directories (debug and release).

If I then have this in premake5.lua:
```
include "debug/conanbuildinfo.premake.lua"
```

That'll include the debug configuration. But then what about the release configuration?

If I do this:

```
include "debug/conanbuildinfo.premake.lua"
include "release/conanbuildinfo.premake.lua"
```

Then I'm just loading two different tables on top of each other. Is this just not supported on premake, or am I doing something wrong?
Any help on this would be appreciated.

fschoenm:
Changelog: (Feature | Fix | Bugfix): experimental premake_multi generator

Closes #6634. I changed the existing premake generator to generate real Lua modules instead of include files. It's just some experimental code. I'm not a Lua/premake expert so this may be solved more elegantly. And I'm not sure if I have enough time to add tests/documentation. But maybe the PR can help somebody else.

It can be used like this:

```
$ conan install . -s build_type=RelWithDebInfo
$ conan install . -s build_type=Debug
```
This creates two files: conanbuildinfo_debug.premake.lua and connabuildinfo_relwithdebinfo.premake.lua.

The modules can then be used in the premake script like this:

```
conan_dbg = require("conanbuildinfo_debug.premake")
conan_rel = require("conanbuildinfo_relwithdebinfo.premake")

solution("Program")
  -- ...
  filter "Debug"
    conan_dbg.conan_basic_setup()
  filter "Release"
    conan_rel.conan_basic_setup()
```
or directly:
```
solution("Program")
  -- ...
  filter "Debug"
    require("conanbuildinfo_debug.premake").conan_basic_setup()
  filter "Release"
    require("conanbuildinfo_relwithdebinfo.premake").conan_basic_setup()
```
I.e. the ``conan_basic_setup()`` functions are now clearly separated into different modules.

### DIFF
--- a/conan/tools/premake/__init__.py
+++ b/conan/tools/premake/__init__.py
@@ -1,0 +1,2 @@
+
+from conan.tools.premake.premakedeps import PremakeDeps
--- a/conan/tools/premake/__init__.py
+++ b/conan/tools/premake/__init__.py
@@ -1,2 +1,1 @@
-
 from conan.tools.premake.premakedeps import PremakeDeps
--- a/conan/tools/premake/premakedeps.py
+++ b/conan/tools/premake/premakedeps.py
@@ -1,0 +1,109 @@
+import os
+import textwrap
+
+from conans.errors import ConanException
+from conans.model.build_info import DepCppInfo
+from conans.util.files import load, save
+
+VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
+
+
+class PremakeDeps(object):
+    
+    _vars_conf_props = textwrap.dedent("""\
+       {lua_module}.conan_includedirs{name} = {{{include_dirs}}}
+       {lua_module}.conan_libdirs{name} = {{{lib_dirs}}}
+       {lua_module}.conan_bindirs{name} = {{{bin_dirs}}}
+       {lua_module}.conan_libs{name} = {{{libs}}}
+       {lua_module}.conan_system_libs{name} = {{{system_libs}}}
+       {lua_module}.conan_defines{name} = {{{definitions}}}
+       {lua_module}.conan_cxxflags{name} = {{{compiler_flags}}}
+       {lua_module}.conan_cflags{name} = {{{compiler_flags}}}
+       {lua_module}.conan_sharedlinkflags{name} = {{{linker_flags}}}
+       {lua_module}.conan_exelinkflags{name} = {{{linker_flags}}}
+       """)
+
+    _vars_luafile = textwrap.dedent("""\
+        local {lua_module} = {{}}
+
+        {lua_module}.conan_build_type = "{build_type}"
+        {lua_module}.conan_arch = "{arch}"
+
+        {deps}
+
+        function {lua_module}.conan_basic_setup()
+            configurations{{{lua_module}.conan_build_type}}
+            architecture({lua_module}.conan_arch)
+            includedirs{{{lua_module}.conan_includedirs}}
+            libdirs{{{lua_module}.conan_libdirs}}
+            links{{{lua_module}.conan_libs}}
+            links{{{lua_module}.conan_system_libs}}
+            defines{{{lua_module}.conan_defines}}
+            bindirs{{{lua_module}.conan_bindirs}}
+        end
+        
+        return {lua_module}
+        """)
+
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self.configuration = conanfile.settings.build_type
+        self.arch = conanfile.settings.arch
+        self.output_path = os.getcwd()
+
+    def generate(self):
+        if self.configuration is None:
+            raise ConanException("PremakeDeps.configuration is None, it should have a value")
+
+        self._lua_module = "conan_{config}".format(config=str(self.configuration).lower())
+
+        generator_files = self._content()
+        for generator_file, content in generator_files.items():
+            generator_file = os.path.join(self.output_path, generator_file)
+            save(generator_file, content)
+
+    def _pkg_props(self, name, cpp_info):
+        def add_valid_ext(libname):
+            (base, ext) = os.path.splitext(libname)
+            return libname if ext == 'framework' else base
+
+        fields = {
+            'name': name,
+            'lua_module': self._lua_module,
+            'root_folder': cpp_info.rootpath,
+            'bin_dirs': ";".join('"%s"' % p for p in cpp_info.bin_paths),
+            #'res_dirs': ";".join('"%s"' % p for p in cpp_info.res_paths),
+            'include_dirs': ";".join('"%s"' % p for p in cpp_info.include_paths),
+            'lib_dirs': ";".join('"%s"' % p for p in cpp_info.lib_paths),
+            'libs': ";".join('"%s"' % add_valid_ext(lib) for lib in cpp_info.libs),
+            'system_libs': ";".join('"%s"' % add_valid_ext(sys_dep) for sys_dep in cpp_info.system_libs),
+            'definitions': ";".join('"%s"' % d for d in cpp_info.defines),
+            'compiler_flags': " ".join(cpp_info.cxxflags + cpp_info.cflags),
+            'linker_flags': " ".join(cpp_info.sharedlinkflags),
+            'exe_flags': " ".join(cpp_info.exelinkflags),
+        }
+        formatted_template = self._vars_conf_props.format(**fields)
+        return formatted_template
+
+    def _content(self):
+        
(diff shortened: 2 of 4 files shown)


conans/client/generators/premake_multi.py
```diff
@@ -0,0 +1,86 @@
...
+            deps = PremakeDeps(dep_cpp_info)
+            dep_name = dep_name.replace("-", "_")
+            dep_flags = template_deps.format(lua_module=lua_module, dep="_" + dep_name, deps=deps)
+            sections.append(dep_flags)
+
+        sections.append(
+            "function {lua_module}.conan_basic_setup()".format(lua_module=lua_module)
+        )
+        sections.append(
+            f"    configurations{{{lua_module}.conan_build_type}}\n"
```
memsharded: Do not use f-strings yet, we need to pass CI

memsharded (changes requested):
We are already in the way to Conan 2.0
Current efforts in toolchains are done in ``conan.tools.xxxx`` space, and following new paradigms, so I think this premake efforts should be directly targeted there.

- The visiting of dependencies should be done following ``conanfile.dependencies`` access (check ``MSBuildDeps`` for example)
- No need to extend Generator
- Only 1 ``premake`` generator should be implemented, valid both for single and multi-config cases

conans/client/generators/premake_multi.py
fschoenm: Hm, the CI run passed with it. Is there another, more thorough CI? Anyway, I replaced them.

fschoenm:
> We are already in the way to Conan 2.0
> Current efforts in toolchains are done in `conan.tools.xxxx` space, and following new paradigms, so I think this premake efforts should be directly targeted there.

My problem is that I needed the functionality right now for a project. If you can't use it right now maybe I can leave the PR for inspiration and try again later after we switched to Conan 2.0.

However, the idea is simple: Instead of an include file I created real Lua modules that don't have colliding function names so multiple configurations can be used in parallel. Might be relatively easy to port to the new toolchain.

memsharded:
> My problem is that I needed the functionality right now for a project. If you can't use it right now maybe I can leave the PR for inspiration and try again later after we switched to Conan 2.0.

It can be done now, but inside ``conan.tools.xxxx``, which is following the design and guidelines that will survive in Conan 2.0. It is also experimental there by default. The ``CMakeDeps``, ``MSBuildDeps`` and others (coming this 1.35 release) are already there in 1.34. No need to wait for Conan 2.0 to get it merged.

fschoenm:
> > My problem is that I needed the functionality right now for a project. If you can't use it right now maybe I can leave the PR for inspiration and try again later after we switched to Conan 2.0.
> 
> It can be done now, but inside `conan.tools.xxxx`, which is following the design and guidelines that will survive in Conan 2.0. It is also experimental there by default. The `CMakeDeps`, `MSBuildDeps` and others (coming this 1.35 release) are already there in 1.34. No need to wait for Conan 2.0 to get it merged.

Ok, I tried again and integrated a PremakeDeps in conan.tools. Not sure how I did but in my simple example it worked similar to my first approach.

Still not a premake/conan expert so I hope someone could iron out the kinks.

memsharded (commented):
This is indeed looking much better now, and well aligned with the new tools.

It would need at least some unittests, that will use it and checks its output. It might be difficult to have actual functional tests building with premake in our CI, but it can be added with ``@pytest.mark.tools_premake`` and they can at least be ran locally when premake is installed.

Do you think you could try to contribute some of these tests? Would you need help here?

Thanks for contributing and iterating this!

fschoenm:
I added a functional test that does a release build. I took inspiration from one of the MSBuildDeps tests.

I'd like to do a release and debug build in the same test, however I wasn't sure how to do the conan run for the debug dependencies in this case.

memsharded:
I have been trying to execute the test in my machine, but so far without success:

- Installed premake5 from the official site
- It seems that it has a Remote Code Execution vulnerability they are aware, but it fails
- They haven't release a fixed release, so I'd need to build premake from sources
- Even in that case, it seems there was an error locating the conanbuildinfo.release.lua file, that is the reason the RCE was being called.

Also, the new generators are generating, when possible, a different file per dependency, and handling transitive dependencies in the build system, so it makes possible to have dependencies of different applications in the same project to different dependencies. I am not sure if this is possible with premake, but if it is, we should try to follow this approach too.

I have pushed some minor pep8 changes to code. Once the test is passing locally successfully, it can be marked with ``@pytest.mark.xfail``, because we don't have premake in CI machines yet.

fschoenm:
> It seems that it has a Remote Code Execution vulnerability they are aware, but it fails

I think this happens if premake cannot find the file that is ```require```d. It's a weird Lua/premake behavior that can only be fixed on their side.

Not really sure why in your test run it couldn't find the _conanbuildinfo_release.premake.lua_. Do you have more information about what happened? Did the PremakeDeps generator not run or why did the file not exist/could not be found?

> Also, the new generators are generating, when possible, a different file per dependency, and handling transitive dependencies in the build system, so it makes possible to have dependencies of different applications in the same project to different dependencies. I am not sure if this is possible with premake, but if it is, we should try to follow this approach too.

Yeah, I've seen that in some of the other generators. That's really where I could use some help from somebody who knows premake much better than me. So far, I just adapted the existing premake generator. Also, I haven't really found anything regarding transitive dependencies in premake like they could be done in CMake.

redorav:
As far as I know that vulnerability thing is not present in the latest code (but the latest version of the tool doesn't yet have it)

I am quite keen in getting multi config in premake, I only just started evaluating conan and it is a bit of a roadblock for the way I have things set up.

**EDIT: By the way one reason why it might not work is that premake doesn't accept require("something.lua"), you need to remove the .lua extension**

fschoenm:
> I am quite keen in getting multi config in premake, I only just started evaluating conan and it is a bit of a roadblock for the way I have things set up.

I'm working on a more advanced approach (haven't pushed the changes yet) that works more like the CMake and MSBuild dependency generators, using one file per package and configuration, with inspiration from this existing premake module: https://github.com/mversluys/premake-ghp. I hope I have time in the next days to push this approach. However, it's still going to handle the settings (like include directories) for a project manually.

What I'm really missing in premake is the ability to define external projects for already prebuilt libs. Also, CMake's approach with targets that can act as transitive dependencies is not (or not obviously?) modeled in premake, I think.

@redorav, if you have any input on that feel free to comment or give ideas.

redorav:
@fschoenm that premake module you link to is very interesting! I'm going to take a look.

wrt prebuilt projects, all I can point to is this https://github.com/premake/premake-core/wiki/externalproject with the caveat that I haven't used it. I don't define external dependencies as projects in premake, I have a wrapper class for them and use as appropriate

The other thing is that I think premake does not bring in dependencies transitively as far as I know, it is a bit more manual. I have rarely found this to be a problem to be fair but I'm sure other people's experiences vary. https://github.com/premake/premake-core/issues/627. 

VS does do the transitive dependency thing in a very odd fashion, where e.g. if library B links to A, it will amalgamate both and when you link to B somewhere else, both are brought together. I find this approach to be horrible, as you'll get duplicate symbol clashes and bloated libraries slowing down your link times. That's just an implementation detail though, I don't have much experience with cmake other than as a consumer of projects, so I cannot really compare what it does and whether it doesn't do this.

fschoenm:
> wrt prebuilt projects, all I can point to is this https://github.com/premake/premake-core/wiki/externalproject with the caveat that I haven't used it.

Yeah, I came across that function but it didn't work. Even the documentation states that it is "Visual Studio only" so I didn't spend much time investigating.

memsharded:
Sorry that this PR couldn't be moved forward back then. I hope that with the possibility of user custom generators that could have been workarounded.

This PR has become a bit too outdated, building on ``deps_cpp_info`` and similar things that do not longer exist in Conan 2.0, so I am closing this. Recall that Conan 2.0 keeps supporting custom generators both inside ``python-requires`` (recommended for build-systems like this) and installable in cache with ``conan config install`` since Conan 2.0.5. Thanks very much for your feedback and contribution!

Issue #8600, opened 2021-03-04
Discuss future approach for cpp_info.names and cpp_info.filenames


memsharded:
Coming from https://github.com/conan-io/conan/pull/8568:

- The ``cpp_info.names["cmake_find_package_multi"]`` strategy for custom generators behaviors seem too specific
- Consider using something more generic like ``cmake`` (when this doesn't break for current cmake generator)
- Or a conceptual name: ``cmake_targets_names``, that could be exploited by different generators

memsharded:
Related https://github.com/conan-io/conan/issues/7661

lasote:
Note: It is not strictly related to this issue but as an observation, for the `CMakeDeps` only the `cpp_info.filenames` field affects to the name of the `xxx-config.cmake` file, it doesn't default to the `self.cpp_info.names` like the older generators.

memsharded:
Closed by https://github.com/conan-io/conan/pull/8727

PR #8727, opened 2021-03-29
Alternative way of setting properties that depend on generator specific features.

### LINKED ISSUE #7661: [feature] Allow arbitrary sections in pkg_config .pc files
pkg_config files can contain extra information, next to include paths/library paths/libs/cflags/requires/...
This is used by some programs to pass around configuration paths.

Because the generated .pc files by conan, are missing this information, configuration fails.

e.g. glib uses this. When using the files generated by conan, the following error appears:
```
Run-time dependency gio-2.0 found: YES 2.65.1
Dependency gio-2.0 found: YES 2.65.1 (cached)
Using 'PKG_CONFIG_PATH' from environment with value: '/home/maarten/.conan/data/glib-networking/2.65.1/_/_/build/eac566419e1470979e589f72ef984a2778d4785a'
Run-time dependency gmodule-2.0 found: YES 2.65.1
WARNING: pkgconfig variable 'giomoduledir' not defined for dependency gio-2.0.

source_subfolder/meson.build:68:0: ERROR: Assert failed: GIO_MODULE_DIR is missing from gio-2.0.pc
```

The `gio-2.0.pc` file, installed in my system contains:
```
prefix=/usr
libdir=${prefix}/lib64
includedir=${prefix}/include

datadir=${prefix}/share
schemasdir=${datadir}/glib-2.0/schemas
bindir=${prefix}/bin
giomoduledir=${libdir}/gio/modules
glib_compile_schemas=${bindir}/glib-compile-schemas
glib_compile_resources=${bindir}/glib-compile-resources
gdbus_codegen=${bindir}/gdbus-codegen

Name: GIO
Description: glib I/O library
Version: 2.60.7
Requires: glib-2.0, gobject-2.0
Requires.private: gmodule-no-export-2.0, zlib, mount >= 2.23, libselinux
Libs: -L${libdir} -lgio-2.0
Libs.private: -ldl -pthread -lresolv
Cflags: -I${includedir}
[maarten@localhost eac5664
```

The generated `gio-2.0.pc` contains:
```
prefix=/home/maarten/.conan/data/glib/2.65.1/_/_/package/f9b484c68439b8cf20fdfbb3f876d6cf74a4f970
libdir=${prefix}/lib
includedir=${prefix}/include

Name: glib-gio-2.0
Description: Conan package: glib-gio-2.0
Version: 2.65.1
Libs: -L${libdir} -lgio-2.0  -lresolv  -ldl  -Wl,-rpath="${libdir}"
Cflags: -I${includedir}
Requires: glib-2.0 gobject-2.0 gmodule-2.0 zlib mount libselinux
```

The generated pc file is missing a lot of keys, of which some are custom.
So it would be useful to add custom key/values to pkg_config files.

This feature is used by glib/glib-networking.

czoido:
Changelog: Feature: Add `set_property` and `get_property` to set properties and access them in generators. Can be set only for a specific generator or as a default value for all of them.
Changelog: Feature: Use `set_property` and `get_property` to support custom defined content in `pkg_config` generator.
Changelog: Feature: Add new property names: `cmake_target_name`, `cmake_file_name`, `pkg_config_name` and `cmake_build_modules` that can be used for multiple generators of the same type allowing also an easier migration of `names`, `filenames` and `build_modules` properties to this model.
Docs: https://github.com/conan-io/docs/pull/2082

Closes: https://github.com/conan-io/conan/issues/7661 , https://github.com/conan-io/conan/issues/8600

Set properties of the _cpp_info_ that are related to generators (things like names, filenames, build_modules...) with `set_property` and `get_property`. Tests with `filenames`, `names` and `build_modules` properties.
Using new properties names `cmake_target_name`, `cmake_file_name`, `pkg_config_name` and `cmake_build_modules` will make the migration a little bit less risky. The new property name will be given preferences over the old one.

This:

``` python
self.cpp_info.components["mycomponent"].names["cmake_find_package"] = "mycomponent-name"
self.cpp_info.components["mycomponent"].names["cmake_find_package_multi"] = "mycomponent-name"
```
would be expressed as this:

``` python
self.cpp_info.components["mycomponent"].set_property("cmake_target_name", "mycomponent-name")
```

It's also possible to make it specific for generator but that most of cases it won't be necessary:

``` python
self.cpp_info.components["mycomponent"].set_property("cmake_target_name", "mypkg-name", "cmake_find_package")
```

Also, not specifying the generator will set this value as a default for every generator. So if you set:

``` python
self.cpp_info.components["mycomponent"].set_property("my-property", "mycomponent-name")
```

A new custom generator can access that information with:

``` python
for pkg_name, cpp_info in self.deps_build_info.dependencies:
    names = cpp_info.get_property("my-property", generator=self.name))
```

Related to: https://github.com/conan-io/conan/issues/8600, https://github.com/conan-io/conan/pull/8568 and https://github.com/conan-io/conan/issues/7661

#tags: slow


memsharded (commented):
I like the idea of generalizing the information for generators, keep going

conans/model/build_info.py
```diff
@@ -209,14 +210,44 @@ def name(self):
...
-        return self.get_name(generator)
+    def get_generator_property(self, property_name, generator=None):
+        generator = generator or "conan_default_generators_value"
+        try:
+            gen_dict = self._generator_properties.get(generator)
+            if gen_dict:
+                return gen_dict.get(property_name)
+            elif self._generator_properties.get("conan_default_generators_value"):
+                return self._generator_properties.get("conan_default_generators_value").get(property_name)
+        except KeyError:
```
memsharded, line 265: This probably doesn't raise if all are ``.get()``?

Also the double ``self._generator_properties.get("conan_default_generators_value")`` would be nice to avoid.

conans/model/build_info.py
```diff
@@ -209,14 +210,44 @@ def name(self):
...
+            return None
+
+    def update_generator_properties(self):
```
memsharded: Name is a bit misleading, if it is some migration helper to move ``names`` and ``filenames`` don't be afraid of name it after that like ``translate_names_filenames()``

conans/model/build_info.py
```diff
@@ -209,14 +210,44 @@ def name(self):
     def name(self, value):
         self._name = value
 
-    def get_name(self, generator):
-        return self.names.get(generator, self._name)
+    def set_generator_property(self, property_name, value, generator=None):
```
memsharded: I'd probably remove the "generator" thing, because it is possible that other consumers, like custom commands or someone else might be interested in the property, so ``set_property`` looks good?

memsharded (commented):
This is looking very good, I like it.

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+def setup_client():
+    client = TestClient()
+    custom_generator = textwrap.dedent("""
+        from conans.model import Generator
+        from conans import ConanFile
+        from conans.model.conan_generator import GeneratorComponentsMixin
+        import textwrap
+        import os
+
+        class custom_generator(GeneratorComponentsMixin, Generator):
```
memsharded, line 20: We need to get rid of this inheritance too. New generators do not need to extend anything, just implement ``generate()`` method. Components helpers should go elsewhere. 

Not now, not in this PR, but in the future.

conans/model/build_info.py
```diff
@@ -209,15 +210,49 @@ def name(self):
...
+                ret_dict[generator] = values.get("cmake_build_modules")
+        return ret_dict if ret_dict else self.build_modules
+
+    def set_property(self, property_name, value, generator=None):
+        generator = generator or "conan_default_generators_value"
+        gen_dict = self._generator_properties.get(generator)
+        if gen_dict:
+            gen_dict.update({property_name: value})
+        else:
+            self._generator_properties.update({generator: {property_name: value}})
```
memsharded:
```suggestion
        self._generator_properties.setdefault(generator, {})[property_name] = value
```

- Use ``None`` as the value that means "no generator". No need to use a custom string
- Avoid ``mydict.update({k: v})`` why not ``mydict[k] = v``?
- ``setdefault`` is more idiomatic

conans/model/build_info.py
```diff
@@ -209,15 +210,49 @@ def name(self):
...
+
+    def get_property(self, property_name, generator=None):
+        generator = generator or "conan_default_generators_value"
+        gen_dict = self._generator_properties.get(generator)
+        if gen_dict and gen_dict.get(property_name):
+            return gen_dict.get(property_name)
+        else:
+            gen_dict = self._generator_properties.get("conan_default_generators_value")
+            return gen_dict.get(property_name) if gen_dict else None
+
```
memsharded:
```suggestion
        if generator:
            try:
                return self._generator_properties[generator][property_name]
            except KeyError:
                pass
        try:
            return self._generator_properties[None][property_name]
        except KeyError:
            pass
```

- avoid getting twice ``gen_dict.get(property_name)``
- There is a redundant double processing when ``generator=None``, same thing is being evaluated twice.

conans/model/build_info.py
```diff
@@ -209,15 +210,49 @@ def name(self):
     def name(self, value):
         self._name = value
 
+    # TODO: Deprecate for 2.0. Only cmake and pkg_config generators should access this.
+    #  Use get_property for 2.0
     def get_name(self, generator):
-        return self.names.get(generator, self._name)
+        property_name = "cmake_target_name" if "cmake" in generator else "pkg_config_name"
```
memsharded: It is a bit confusing this "pkg_config_name" for every other thing not "cmake"

memsharded (commented):
Some points to consider and to discuss, I still need to understand a bit more the new migration proposal, now that it seems there is no longer an internal translation to properties.

conans/model/build_info.py
```diff
@@ -209,15 +210,55 @@ def name(self):
...
+    # TODO: Deprecate for 2.0. Only cmake generators should access this. Use get_property for 2.0
     def get_filename(self, generator):
-        result = self.filenames.get(generator)
+        result = self.get_property("cmake_file_name", generator) or self.filenames.get(generator)
         if result:
             return result
         return self.get_name(generator)
 
+    # TODO: Deprecate for 2.0. Use get_property for 2.0
+    def get_build_modules(self):
```
memsharded, line 234: It is a bit undesired that this is called so many times, it could be cached, but the above is redefining ``self.build_modules = BuildModulesDict.from_list(self.get_build_modules())``, which is used in the method, so that might break such caching.

Recall that things in CppInfo might be called mamy many times while propagating graph, so this is one of the bottlenecks in graph resolution, we should be careful with adding here inefficient things.

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+@pytest.fixture(scope="module")
+def setup_client():
+    client = TestClient()
+    custom_generator = textwrap.dedent("""
+        from conans.model import Generator
+        from conans import ConanFile
+        from conans.model.conan_generator import GeneratorComponentsMixin
+        import textwrap
```
memsharded: Possibly unused imports

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+                self.cpp_info.components["mycomponent"].set_property("cmake_build_modules", [os.path.join("lib", "mypkg_bm.cmake")])
+                self.cpp_info.components["mycomponent"].set_property("custom_name", "mycomponent-name", "custom_generator")
+        """)
+
+    client.save({"mypkg.py": mypkg})
+    client.run("export mypkg.py")
+    client.run("install consumer.py --build missing -s build_type=Release")
+
+    with open(os.path.join(client.current_folder, "my-generator.txt")) as custom_gen_file:
+        assert "mycomponent:mycomponent-name" in custom_gen_file.read()
```
memsharded:
```suggestion
    my_generator= client.load("my-generator.txt")
    assert "mycomponent:mycomponent-name" in my_generator
```

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+
+    files_to_compare = ["FindMyFileName.cmake", "MyFileNameConfig.cmake", "MyFileNameTargets.cmake",
+                        "MyFileNameTarget-release.cmake", "MyFileNameConfigVersion.cmake", "mypkg.pc",
+                        "mycomponent.pc"]
+    new_approach_contents = get_files_contents(client.current_folder, files_to_compare)
+
+    mypkg = textwrap.dedent("""
+        import os
+        from conans import ConanFile, CMake, tools
```
memsharded: Unused imports

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+                "mypkg_bm.cmake": build_module, "mypkg_anootherbm.cmake": another_build_module})
+    return client
+
+
+def get_files_contents(folder, filenames):
+    ret = []
+    for filename in filenames:
+        with open(os.path.join(folder, filename)) as properties_package_file:
+            ret.append(properties_package_file.read())
+    return ret
```
memsharded:
```suggestion
def get_files_contents(folder, filenames):
    return [load(os.path.join(folder, f)) for f in filenames]
```

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+def test_same_results_components(setup_client):
+    client = setup_client
+    mypkg = textwrap.dedent("""
+        import os
+        from conans import ConanFile, CMake, tools
```
memsharded, line 68: Unused imports

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -0,0 +1,266 @@
...
+                self.cpp_info.names["cmake_find_package"] = "mypkg-name"
+                self.cpp_info.names["cmake_find_package_multi"] = "mypkg-name"
+                self.cpp_info.names["custom_generator"] = "mypkg-name"
+                self.cpp_info.build_modules.append(os.path.join("lib", "mypkg_bm.cmake"))
+        """)
+    client.save({"mypkg.py": mypkg})
+    client.run("create mypkg.py")
+    client.run("install consumer.py -s build_type=Release")
+
+    old_approach_contents = get_files_contents(client.current_folder, files_to_compare)
```
memsharded: I like very much these tests comparing the previous and new results, good job

memsharded (approved):
A couple of minor stylistic things, but overall good job, I think this is ready to be merged

PR #8739, opened 2021-03-31
client cache: Use /etc/conan global config if available

### LINKED ISSUE #8738: [feature] Global config /etc/conan/
Conan shall check if /etc/conan/* configs exist before reading configs from user home. This is similar behaviour to many tools, for example Git.

MaciejKucia:
Changelog: (Feature): Use /etc/conan global config if available

fixes #8738

### DIFF
--- a/conans/client/cache/cache.py
+++ b/conans/client/cache/cache.py
@@ -25,6 +25,7 @@ from conans.paths.package_layouts.package_editable_layout import PackageEditable
 from conans.util.files import list_folder_subdirs, load, normalize, save, remove
 from conans.util.locks import Lock
 
+CONAN_GLOBAL_CONF = '/etc/conan/'
 CONAN_CONF = 'conan.conf'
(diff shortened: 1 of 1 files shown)


memsharded (commented):
This cannot be merged at this moment:

- We introduce mechanisms that work the same in all platforms. If a global location is being introduced, it should also be defined for Windows and OSX
- But most importantly, Conan configuration is not managed globally, but per instance/cache/user, as it can change from application to application. There is a mechanism, ``conan config install`` to populate the cache with the necessary configuration. Having the config global will collide for different users or even different concurrent jobs in the same CI agent.
- If global configuration were to be introduced, it should be cumulative, it should read the global one, but also the cache one, and accumulate it, with the relevant logic to prioritize them (cache config has higher priority). This would add some added complexity that we think it wouldn't be worth at this moment.

Thanks for trying to contribute to Conan.

Issue #8703, opened 2021-03-25
[bug] cmake link options treated as files with cmake_find_package generator


srnwk:
### Environment Details (include every applicable attribute)
  * Operating System+version: Windows 10 20H2
  * Compiler+version: MSVC 19.28
  * Conan version: 1.34.1
  * Python version: 3.9.1

### Steps to reproduce
Install siglost/1.2.0:
```sh
conan install sigslot/1.2.0 -g cmake_find_package_multi
```
Consume Pal::Sigslot via cmake:
```cmake
find_package(PalSigslot REQUIRED)
target_link_libraries(example PRIVATE Pal::Sigslot)
```
Configure & build:
```sh
cmake -G Ninja -B build . && ninja -C build
```
### Logs
```
ninja: error: '/OPT:NOICF', needed by 'bin/example.exe', missing and no known rule to make it
```

sigslot/1.2.0 declares a link option `/OPT:NOICF`.
When using the cmake_find_package_multi generator or cmake_find_package,
that gets added verbatim to the INTERFACE_LINK_LIBRARIES property of the declared target.

Since it doesn't start with `-`, it's treated as a library name by cmake instead of a linker flag,
and with the Ninja generator, cmake ends up generating something like: `build bin\example.exe: ... \OPT$:NOICF ...`
leading to the error shown above.

With cmake 3.13+ INTERFACE_LINK_OPTIONS could be used to avoid this, but for earlier versions I'm not sure.
cl.exe, link.exe etc. will accept flags with `-` instead of `/` as well, so this could be used as a workaround in the recipe,
but IIRC this is entirely undocumented...

SpaceIm:
It could also be protected with double quotes for CMake: `self.cpp_info.components["_sigslot"].exelinkflags.append("\"/OPT:NOICF\"")`
But I don't know if it would be a good solution for non-CMake generators.

memsharded:
I have been able to reproduce:

- Using quotes doesn't help, (at least CMake 3.15)
- Using -/OPT:NOICF with the ``-`` works, and without quotes.

memsharded:
This is surprising it has not been raised before. Apparently the ``-`` is necessary for the CMake generators, and I have submitted a potential fix in https://github.com/conan-io/conan/pull/8740, but we need to carefully analyze if it could be breaking somehow.

memsharded:
Closed by https://github.com/conan-io/conan/pull/8740 ,will be in 1.35.1

[...]

PR #8740, opened 2021-03-31
fix linker flags in cmake (find_package based) generators

memsharded:
Changelog: Bugfix: Fix linker flags in cmake (find_package based) generators.
Docs: Omit

Fix: https://github.com/conan-io/conan/issues/8703

#tags: slow

### DIFF
--- a/conan/tools/cmake/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps.py
@@ -217,7 +217,15 @@ class DepsCppCmake(object):
             return '"%s"' % ";".join(p.replace('\\', '/').replace('$', '\\$') for p in values)
 
         def format_link_flags(link_flags):
-            return [(f if f.startswith("-") else"-{}".format(f)) for f in link_flags]
+            result = []
+            for f in link_flags:
+                if f.startswith("-"):
+                    result.append(f)
+                else:
+                    if f.startswith("/"):  # msvc link flag
+                        f = f[1:]  # Remove the initial / and use only "-"
+                    result.append("-{}".format(f))
+            return result
 
         self.include_paths = join_paths(cpp_info.include_paths)
         self.include_path = join_paths_single_var(cpp_info.include_paths)
--- a/conan/tools/cmake/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps.py
@@ -235,8 +235,9 @@ class DepsCppCmake(object):
         # Issue: #1251
         self.cxxflags_list = join_flags(";", cpp_info.cxxflags)
         self.cflags_list = join_flags(";", cpp_info.cflags)
-        self.sharedlinkflags_list = join_flags(";", cpp_info.sharedlinkflags)
-        self.exelinkflags_list = join_flags(";", cpp_info.exelinkflags)
+        self.sharedlinkflags_list = join_flags(";", ["-{}".format(f)
+                                                     for f in cpp_info.sharedlinkflags])
+        self.exelinkflags_list = join_flags(";", ["-{}".format(f) for f in cpp_info.exelinkflags])
 
         self.rootpath = join_paths([cpp_info.rootpath])
         self.build_modules_paths = join_paths(cpp_info.build_modules_paths.get(generator_name, []))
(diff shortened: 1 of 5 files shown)


conan/tools/cmake/cmakedeps.py
```diff
@@ -235,8 +235,9 @@ def join_paths_single_var(values):
         # Issue: #1251
         self.cxxflags_list = join_flags(";", cpp_info.cxxflags)
         self.cflags_list = join_flags(";", cpp_info.cflags)
-        self.sharedlinkflags_list = join_flags(";", cpp_info.sharedlinkflags)
-        self.exelinkflags_list = join_flags(";", cpp_info.exelinkflags)
+        self.sharedlinkflags_list = join_flags(";", ["-{}".format(f)
+                                                     for f in cpp_info.sharedlinkflags])
+        self.exelinkflags_list = join_flags(";", ["-{}".format(f) for f in cpp_info.exelinkflags])
```
srnwk: Could this break flags that already start with a `-` or `--`, for non-MSVC linkers?

memsharded: Yes, good catch, I will verify that too.

memsharded: Keeping flags that already start with "-" in https://github.com/conan-io/conan/pull/8740/commits/4ebe404bbd85c43c71a44626816dd419dc37c6b1

conans/test/unittests/client/generators/cmake_find_package_multi_test.py
```diff
@@ -27,3 +28,27 @@ def test_cmake_find_package_multi_version(self):
...
+    cpp_info = CppInfo("mypkg", "dummy_root_folder1")
+    cpp_info.sharedlinkflags = ["/NODEFAULTLIB", "/OTHERFLAG"]
+    cpp_info.exelinkflags = ["/OPT:NOICF"]
+    conanfile.deps_cpp_info.add("mypkg", cpp_info)
+
+    gen = CMakeFindPackageMultiGenerator(conanfile)
+    files = gen.content
+    d = files["mypkgTarget-release.cmake"]
+    assert "$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,EXECUTABLE>:-/OPT:NOICF>" in d
+    assert "$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,MODULE_LIBRARY>:-/NODEFAULTLIB;-/OTHERFLAG>" in d
```
SSE4: so, is it really expected to pass `-/OPT` with both dash and slash?

memsharded: Apparently it worked, but you are right, better double check, I'll check it.

conan/tools/cmake/cmakedeps.py
```diff
@@ -216,6 +216,9 @@ def join_paths_single_var(values):
             """
             return '"%s"' % ";".join(p.replace('\\', '/').replace('$', '\\$') for p in values)
 
+        def format_link_flags(link_flags):
+            return [(f if f.startswith("-") else"-{}".format(f)) for f in link_flags]
```
danimtb:
```suggestion
            return [(f if f.startswith("-") else "-{}".format(f)) for f in link_flags]
```

conans/client/generators/cmake.py
```diff
@@ -34,6 +34,9 @@ def join_paths_single_var(values):
             """
             return '"%s"' % ";".join(p.replace('\\', '/').replace('$', '\\$') for p in values)
 
+        def format_link_flags(link_flags):
+            return [(f if f.startswith("-") else"-{}".format(f)) for f in link_flags]
```
danimtb:
```suggestion
            return [(f if f.startswith("-") else "-{}".format(f)) for f in link_flags]
```

conans/test/unittests/client/generators/cmake_find_package_multi_test.py
danimtb: yeah, that syntax looks a bit weird to me

memsharded: Ok, it was working, it didn't raise an error, but just a warning. Dropped the extra /, please check.

danimtb (approved):

SSE4 (approved):

Issue #8741, opened 2021-03-31
[bug] Conan install fails after updating to 1.35.0


SpiderCh:
### Environment Details (include every applicable attribute)
  * OS: Windows 7 SP1, MacOS 11.2.3
  * Compiler: MSVC 19.28.29337.0, clang 5.0, clang 11.0
  * Conan: 1.35.0
  * Python version: 3.8.2, 3.9

### Steps to reproduce (Include if Applicable)
Install conan 1.35.0. Try to call conan install.

### Logs (Executed commands with output) (Include/Attach if Applicable)

$ conan install path_to_conanfile -if install_path -pr your_profile
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Debug
compiler=clang
compiler.libcxx=libc++
compiler.version=5.0
os=Macos
os_build=Macos
[options]
[build_requires]
[env]

Traceback (most recent call last):
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/command.py", line 2164, in run
    method(args[0][1:])
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/command.py", line 500, in install
    info = self._conan.install(path=args.path_or_reference,
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/conan_api.py", line 93, in wrapper
    return f(api, *args, **kwargs)
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/conan_api.py", line 598, in install
    deps_install(app=self.app,
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/manager.py", line 57, in deps_install
    deps_graph = graph_manager.load_graph(ref_or_path, create_reference, graph_info, build_modes,
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/graph/graph_manager.py", line 116, in load_graph
    root_node = self._load_root_node(reference, create_reference, profile_host, graph_lock,
  File "/usr/local/Cellar/conan/1.35.0/libexec/lib/python3.9/site-packages/conans/client/graph/graph_manager.py", line 158, in _load_root_node
    root_node = self._load_root_consumer(path, graph_lock, profile_host, root_ref)
[...]

PR #8746, opened 2021-03-31
avoid breaking users calling forbidden private api conanfile.__init__

memsharded:
Changelog: Fix: Avoid breaking users calling forbidden private api ``conanfile.__init__``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/8741

#tags: slow

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -47,8 +47,8 @@ class ConanFileLoader(object):
         """
         cached = self._cached_conanfile_classes.get(conanfile_path)
         if cached and cached[1] == lock_python_requires:
-            conanfile = cached[0](self._output, self._runner, display, user, channel,
-                                  self._requester)
+            conanfile = cached[0](self._output, self._runner, display, user, channel)
+            conanfile._conan_requires = self._requester
             if hasattr(conanfile, "init") and callable(conanfile.init):
                 with conanfile_exception_formatter(str(conanfile), "init"):
                     conanfile.init()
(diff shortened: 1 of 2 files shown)


conans/client/loader.py
```diff
@@ -47,8 +47,8 @@ def load_basic_module(self, conanfile_path, lock_python_requires=None, user=None
         """
         cached = self._cached_conanfile_classes.get(conanfile_path)
         if cached and cached[1] == lock_python_requires:
-            conanfile = cached[0](self._output, self._runner, display, user, channel,
-                                  self._requester)
+            conanfile = cached[0](self._output, self._runner, display, user, channel)
+            conanfile._conan_requires = self._requester
```
czoido:
```suggestion
            conanfile._conan_requester = self._requester
```

SSE4:
@memsharded is it really needed to be fixed? that appears to be private API, thus breakage is totally expected. (and if breakage of private API is not expected, then this particular fix breaks behavior of 1.35.0).
I suggest to close #8741 instead of adding workarounds to the codebase.

czoido (approved):
I think we all agree this is not the ideal solution as commented in the PR: https://github.com/conan-io/conan/issues/8741#issuecomment-811158727 but I also would prefer to give some time for this users to migrate and revert this changes in the near future. 
Although it's already stated, maybe we could emphasise a bit more in the documentation that making use of the private implementation could lead to breakage.

SSE4:
if we want to avoid breakage, this cannot be merged as is, because right now we have two categories of users:
1. pre-1.35 users who used private ctor with 4 arguments
2. 1.35 users who used private ctor with 5 arguments

the changes breaks second category now, as there is no 5 arguments ctor any more. IMO it would be better to provide two overloads of ctor then.

czoido:
> 1. pre-1.35 users who used private ctor with 4 arguments
> 2. 1.35 users who used private ctor with 5 arguments
> 
> the changes breaks second category now, as there is no 5 arguments ctor any more. IMO it would be better to provide two overloads of ctor then.

Technically, yes, we will break people who used that 5 arguments constructor between the release of 1.35.0 and the moment we release 1.35.1, but I hope no one decided to do this in this short period of time... so IMHO I would try to help those users that have hundreds of packages affected by this now and fix this in the near future...

PR #8749, opened 2021-03-31
CMakeToolchain customization and extensibility mechanism

### LINKED ISSUE #7484: [question] How to inject ANDROID_ABI, and use the NDK and toolchain file provided by Android?
I see that in [!4538](https://github.com/conan-io/conan/pull/4538), [cmake_flags.py](https://github.com/bincrafters/conan-android_ndk_installer/blob/4c98a5c87f76feaa1be9693064d72a61cc562278/conanfile.py#L183) was augmented to set the `ANDROID_ABI` definition in CMake.  But code only applies this if a toolchain file is not provided.

My problem is that I'm using the CMake toolchain file provided in the NDK (and using a per-existing NDK), however it requires `ANDROID_ABI` to be set externally (before it).  Thus, I have no way to supply `ANDROID_ABI` to the toolchain file, and then it just assumed armv7.

Is there a different way to inject `ANDROID_ABI`?  Or can`cmake_flags.py` be modified to set the `ANDROID_ABI` definition before checking if there's a toolchain file?  Because the one from the NDK - if given `ANDROID_ABI` does a lot of the heavy lifting.  I could drop using the NDK supplied toolchain file in favour of relying on conan to get the values right, but (no criticism of conan here) it feels like depending on Android for the NDK toolchain file is more sustainable.

### LINKED ISSUE #7908: [feature] cmake toolchains should be generator agnostic
We regularly use a few different cmake generators for development, even when targeting a single platform. Because `conan_toolchain.cmake` has different content depending on the generator conan is targeting, right now our developers have to regularly change their `CONAN_CMAKE_GENERATOR` environment variable when switching generators, which ends up being a fiddly process that is frequently forgotten, and leads to cryptic errors.

Additionally, for recipes that hard-code a specific generator, a developer of a conan package has to both pass the generator to the `CMakeToolchain` as well as the `CMake` build helper. This is again easy to forget, especially when the cmake toolchain doesn't need any other special configuration (i.e. `toolchain = "cmake"` fails to work as a developer might expect).

Instead, the `conan_toolchain.cmake` file's contents should not depend upon the generator. Where generator specific content is required, that can be checked at cmake runtime by examining the `CMAKE_GENERATOR` variable, so for instance the Visual Studio toolset could be set after checking that `CMAKE_GENERATOR` is a Visual Studio project, or similar for passing the `/MP` flag.

memsharded:
Changelog: Feature: New ``CMakeToolchain`` customization and extensibility mechanism with blocks of components instead of inheritance.
Docs: https://github.com/conan-io/docs/pull/2085

#tags: slow

Close https://github.com/conan-io/conan/issues/7484
Close https://github.com/conan-io/conan/issues/7908


czoido (approved):
I like this approach much more than the inheritance previous one, for me this is more clear, organized and scalable.

SSE4 (approved):
much better

lasote (approved):
Neat

PR #8765, opened 2021-04-06
Feature/visitor more data

memsharded:
Changelog: Feature: Make available the full recipe and package reference to consumers via ``.dependencies``.
Docs: Omit

Note: I am not documenting this yet, I'd prefer until the Conan 2.0 ongoing work with the graph settles a bit, because we will learn a lot of things there, better wait a bit.

#tags: slow


czoido (approved):

lasote (approved):
:+1:

PR #8766, opened 2021-04-06
New cppinfo and CMakeDeps as example of usage

lasote:
Changelog: omit
Docs: omit

Summary:

- Not using `deps_cpp_info` in the generator anymore.
- Simplified cppinfo class that now doesn't use root folders internally until the usage of `to_absolute_paths(self, root_folder)`. 

NOTE: Removed multi config packages


conans/model/new_build_info.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def to_absolute_paths(self, root_folder):
+        for name in _DIRS_VAR_NAMES:
+            setattr(self, name, [os.path.join(root_folder, c) for c in getattr(self, name)])
+
+        self.build_modules = {k: [os.path.join(root_folder, e) for e in v]
+                                  for k, v in self.build_modules.items()}
+
+
+class NewCppInfo(_BaseNewCppInfo):
```
memsharded: Maybe it would simplify the design a bit to assume that there is always at least 1 component, without a name, or with a canonical name like the package name, and forward all the base definitions to that component, and start managing everything as a component?

```python
# assuming we use ``None`` as internal identifier of the package component
self.cpp_info.includedirs = ["myinc"]  # equal self.cpp_info.components[None].includedirs = ["myinc"]
```
Just an idea to consider.

conan/tools/cmake/cmakedeps.py
```diff
@@ -764,3 +669,81 @@ def _config(self, filename, name, version, public_deps_names):
...
+            for comp_name, comp in req.cpp_info.components.items():
+                components[comp_name] = _get_real_name(comp, comp_name)
+            if req.name not in components:
+                # Default component (no component) => package name
+                components[req.name] = name
+            ret[req.name] = _PackageNames(name, filename, components)
+        return ret
+
+    @staticmethod
+    def get_target_names_and_filenames(req, cpp_info, computed_names):
```
memsharded: The management of names and filenames is being improved by @czoido in https://github.com/conan-io/conan/pull/8727, probably this will be simplified by that

conans/model/new_build_info.py
```diff
@@ -0,0 +1,172 @@
...
+        self.build_patterns = []
+        self.res_patterns = []
+        self.framework_patterns = []
+
+    def get_name(self, generator):
+        return self.names.get(generator)
+
+    def to_absolute_paths(self, root_folder):
+        for name in _DIRS_VAR_NAMES:
+            setattr(self, name, [os.path.join(root_folder, c) for c in getattr(self, name)])
```
memsharded: Is it fully necessary that paths are converted to absolute paths? I can totally see a generator like CMakeDeps generating:

```cmake
set(MYPKG_ROOT_FOLDER /some/path/to/package/13234)
set(MYPKG_INCLUDEDIRS ${MYPKG_ROOT_FOLDER}/include ${MYPKG_ROOT_FOLDER}/myinc)
set(MYPKG_LIBDIRS ${MYPKG_ROOT_FOLDER)/lib)
```

memsharded (commented):
I like this, this is a good step forward in the right direction.

conans/model/new_build_info.py
lasote: Interesting. I will check it.

conans/model/new_build_info.py
lasote: Totally right

conans/model/new_build_info.py
```diff
@@ -0,0 +1,172 @@
...
+        self.defines = []  # preprocessor definitions
+        self.cflags = []  # pure C flags
+        self.cxxflags = []  # C++ compilation flags
+        self.sharedlinkflags = []  # linker flags
+        self.exelinkflags = []  # linker flags
+
+        self.sysroot = ""
+        self.requires = []
+
+        # ##### PATTERNS
```
lasote: Note: The patterns are not necessary for this PR so I will remove it and when we move this forward will add it to the layouts PR.

conans/model/new_build_info.py
lasote: Done, review

conans/model/new_build_info.py
lasote: Done. check

conans/client/graph/conanfile_dependencies.py
```diff
@@ -57,7 +58,7 @@ def host_requires(self):
                     if not edge.build_require and not edge.private]
         requires.extend([ConanFileInterface(edge.dst.conanfile)
                          for edge in self._node.dependencies if edge.build_require])
-        return DependencyOrderedSet(requires)
+        return DependencyOrderedSet([r for r in requires if r.context == CONTEXT_HOST])
```
lasote: @memsharded BIG check on this. I'm not sure what I'm doing :sweat_smile:

memsharded: Yes, this seems correct, and a bug, I will do a separate PR for this.

memsharded: https://github.com/conan-io/conan/pull/8823

conan/tools/cmake/cmakedeps.py
```diff
@@ -615,6 +555,20 @@ def _data_filename(self, pkg_filename):
         data_fname += "-data.cmake"
         return data_fname
 
+    def get_name(self, req):
+        return req.cpp_info.get_property("cmake_target_name", self.name) or req.ref.name
```
memsharded: This is good, but we might have a transition problem, the CMakeDeps will not work with ConanCenter packages in a very long time. Maybe we should consider postponing this.

conans/model/new_build_info.py
```diff
@@ -0,0 +1,169 @@
+import copy
```
memsharded, line 1: I LOVE this new file and implementation.

memsharded (commented):
This is a huge step forward, way more clear, readable and structured, I love it.

My only concern would be the migration path for this, lets talk and lets have this merged for next 1.36!

conan/tools/cmake/cmakedeps.py
lasote: done

conans/model/new_build_info.py
lasote: nice!

conan/tools/cmake/cmakedeps.py
```diff
@@ -619,83 +583,88 @@ def content(self):
...
-            pkg_public_deps_filenames = [self._get_filename(self._conanfile.deps_cpp_info[it[0]],
-                                                            pkg_name)
-                                         for it in public_deps]
-            config_version = self.config_version_template.format(version=pkg_version)
+        host_requires = {r.ref.name: r for r in
+                         self._conanfile.dependencies.transitive_host_requires}
+        # FIXME: convert to NewCppInfo into the conanfile_interface.py
+        #  "cpp_info" getter for all the new generators
+        for req in host_requires.values():
+            req._cpp_info = NewCppInfo.from_old_cppinfo(req._conanfile.cpp_info)
```
memsharded: It shouldn't be the ``CMakeDeps`` the one creating and assigning a private ``req._cpp_info``. I think maybe this responsibility can be moved to ``conanfile.py``?

PR #8769, opened 2021-04-06
Remove invalid config_install_interval

### LINKED ISSUE #8713: [bug] ERROR: Incorrect definition of general.config_install_interval: 1s
executing the following command:
```
$ conan config set general.config_install_interval=1s
```
makes conan completely unusable, any conan command will fail:
```
$ conan config set general.config_install_interval=1h
ERROR: Incorrect definition of general.config_install_interval: 1s
$ conan create . stc/1.0.0-rc1@
ERROR: Incorrect definition of general.config_install_interval: 1s
```
and it seems like completely impossible to restore, at least from the command line.

looking at the code: https://github.com/conan-io/conan/blob/8e964922074538899de71bff068bb959fcfae6ac/conans/util/dates.py#L33
it doesn't allow seconds there, but seconds are allowed on `conan config set`, which is inconsistent...

I also don't like usage of in-house bicycle for parsing dates. I suggest to use [isodate](https://pypi.org/project/isodate/) module and its function `parse_duration`. it will be more flexible, as it will allow to use an arbitrary time intervals - e.g. not just an integral values of hours/minutes/seconds, but e.g. `1H30M` (1 hour and 30 minutes).

the new implementation may look like:
```
def timedelta_from_text(interval):
    import isodate
    interval = interval.upper()
    if not interval.startswith("PT"):
        interval = "PT" + interval
    return isodate.parse_duration(interval)
```

I also think `conan config set` command should validate value the same way (via trying `timedelta_from_text`), right now it allows arbitrary strings, e.g. the following is always successful command:
```
$ conan config set general.config_install_interval=xxx
```
(alternatively, read of config could default to 0 or infinity on parse error)
as every command reads the config, it's pretty annoying

/cc @uilianries 

  * Operating System+version:
  * Compiler+version:
  * Conan version:
  * Python version:

uilianries:
Hello there.

I've added new units for `seconds` and `weeks`. Now when loading `conan.conf`, if `config_install_interval` is invalid, it will be removed.

I did another change that may break users, but I think necessary. Before this PR, if `config_install_interval` value had no value (e.g. config_install_interval=1), it would be considered **days**, implicitly. Now, an unit is mandatory (explicit is better than implicit).

I need to update docs, because those units are not clear there, but first I would like some review.

fixes #8713

Changelog: Fix: `config_install_interval` no longer enter in loop when invalid.
Docs: https://github.com/conan-io/docs/pull/2067

### DIFF
--- a/conans/test/functional/command/config_install_test.py
+++ b/conans/test/functional/command/config_install_test.py
@@ -8,6 +8,7 @@ import unittest
 import pytest
 import six
 from mock import patch
+from parameterized import parameterized
 
 from conans.client.cache.remote_registry import Remote
 from conans.client.conf import ConanClientConfigParser
@@ -637,6 +638,7 @@ class ConfigInstallSchedTest(unittest.TestCase):
         self.folder = temp_folder(path_with_spaces=False)
         save_files(self.folder, {"conan.conf": conanconf_interval})
         self.client = TestClient()
+        self.client.save({"conanfile.txt": ""})
 
     def test_config_install_sched_file(self):
         """ Config install can be executed without restriction
@@ -685,14 +687,17 @@ class ConfigInstallSchedTest(unittest.TestCase):
         self.client.run('config get general.config_install_interval', assert_error=True)
         self.assertIn("config_install_interval defined, but no config_install file", self.client.out)
 
-    def test_invalid_time_interval(self):
-        """ config_install_interval only accepts minutes, hours or days
+    @parameterized.expand([("1y",), ("2015t",), ("42",)])
+    def test_invalid_time_interval(self, internal):
+        """ config_install_interval only accepts seconds, minutes, hours, days and weeks.
         """
-        self.client.run('config set general.config_install_interval=1s')
+        self.client.run('config set general.config_install_interval={}'.format(internal))
         # Any conan invocation will fire the configuration error
         self.client.run('install .', assert_error=True)
-        self.assertIn("ERROR: Incorrect definition of general.config_install_interval: 1s",
+        self.assertIn("ERROR: Incorrect definition of general.config_install_interval: {}. "
+                      "Removing it from conan.conf to avoid possible loop error.".format(internal),
                       self.client.out)
+        self.client.run('install .')
 
     @pytest.mark.tool_git
     def test_config_install_remove_git_repo(self):
--- a/conans/util/dates.py
+++ b/conans/util/dates.py
@@ -34,15 +34,11 @@ def timedelta_from_text(interval):
     match = re.search(r"(\d+)([smhdw])", interval)
     try:
         value, unit = match.group(1), match.group(2)
-        if unit == 's':
-            return datetime.timedelta(seconds=float(value))
-        elif unit == 'm':
-            return datetime.timedelta(minutes=float(value))
-        elif unit == 'h':
-            return datetime.timedelta(hours=float(value))
-        elif unit == 'd':
-            return datetime.timedelta(days=float(value))
-        elif unit == 'w':
-            return datetime.timedelta(weeks=float(value))
+        name = {'s': 'seconds',
+                'm': 'minutes',
+                'h': 'hours',
+                'd': 'days',
+                'w': 'weeks'}[unit]
+        return datetime.timedelta(**{name: float(value)})
     except Exception:
         raise ConanException("Incorrect time interval definition: %s" % interval)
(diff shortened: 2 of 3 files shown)


conans/test/functional/command/config_install_test.py
```diff
@@ -685,14 +687,17 @@ def test_invalid_scheduler(self):
...
         """
-        self.client.run('config set general.config_install_interval=1s')
+        self.client.run('config set general.config_install_interval={}'.format(internal))
         # Any conan invocation will fire the configuration error
         self.client.run('install .', assert_error=True)
-        self.assertIn("ERROR: Incorrect definition of general.config_install_interval: 1s",
+        self.assertIn("ERROR: Incorrect definition of general.config_install_interval: {}. "
+                      "Removing it from conan.conf to avoid possible loop error.".format(internal),
                       self.client.out)
+        self.client.run('install .')
```
uilianries, line 700: Once removed, no error should occur, as we have default conanfile.txt to be installed.

conans/util/dates.py
```diff
@@ -31,14 +31,18 @@ def timestamp_to_str(timestamp):
...
         value, unit = match.group(1), match.group(2)
-        if unit == 'm':
+        if unit == 's':
+            return datetime.timedelta(seconds=float(value))
+        elif unit == 'm':
             return datetime.timedelta(minutes=float(value))
         elif unit == 'h':
             return datetime.timedelta(hours=float(value))
-        else:
+        elif unit == 'd':
```
uilianries: It may be a breaking change. Default was `days`, now we no longer accept implicit values. I don't think it's a good idea keeping implicit support, we would need read docs to understand what's the behavior.

memsharded: In the docs: 

> The config_install_interval variable starts a time scheduler which runs conan config install according the time interval configured. It only accepts the follow time intervals: minutes, hours and days.

It is lacking some details. It should also have been labeled as experimental, this could be a problem, I'd probably keep the default with a big warning.

memsharded (commented):
Looking good, just some concern about breaking, it is indeed likely that this will break, so probably keep old default "days".

uilianries:
> Looking good, just some concern about breaking, it is indeed likely that this will break, so probably keep old default "days".

Checking again, it was an impossible condition. At line 34 has a regex which only accepts values with units. I just reverted `conans/util/dates.py` and executed new tests, which tries to validate with no unit and it failed as expected.

conans/util/dates.py
```diff
@@ -31,14 +31,18 @@ def timestamp_to_str(timestamp):
 
 
 def timedelta_from_text(interval):
-    match = re.search(r"(\d+)([mhd])", interval)
+    match = re.search(r"(\d+)([smhdw])", interval)
     try:
         value, unit = match.group(1), match.group(2)
```
SSE4, line 36: the overall code could be significantly simplified in order to avoid if spagetti:
```
name = {'s': 'seconds',
        'm': 'minutes',
        'h': 'hours',
        'd': 'days',
        'w': 'weeks'}[unit]

print(datetime.timedelta(**{name: value}))
```

conans/util/dates.py
```diff
@@ -31,14 +31,18 @@ def timestamp_to_str(timestamp):
...
+            return datetime.timedelta(seconds=float(value))
+        elif unit == 'm':
             return datetime.timedelta(minutes=float(value))
         elif unit == 'h':
             return datetime.timedelta(hours=float(value))
-        else:
+        elif unit == 'd':
             return datetime.timedelta(days=float(value))
+        elif unit == 'w':
+            return datetime.timedelta(weeks=float(value))
```
SSE4: there is no else branch in order unit isn't any of s/m/h/d/w

memsharded:
> Checking again, it was an impossible condition. At line 34 has a regex which only accepts values with units. I just reverted conans/util/dates.py and executed new tests, which tries to validate with no unit and it failed as expected.

You are right, good!

conans/util/dates.py
memsharded: Not really necessary if the previous regex guarantees unit is always defined. But yes, probably the dictionary is a bit neater.

conans/util/dates.py
uilianries: Thank you, I was looking for it!

uilianries: Okay, added to current code.

memsharded (approved):

PR #8778, opened 2021-04-09
fix python 3.9 test, different checksum of files created with tar.

### LINKED ISSUE #8020: [bug] Python 3.9 makes different tar files
The test `test_conan_data_as_source` fails in Python 3.9 because the checksums of the files are not the same in Python 3.9.
The change is introduced in Python because there was a "bug" generating tar files reported [here](https://bugs.python.org/issue18819) and fixed [here](https://github.com/python/cpython/commit/674935b8caf33e47c78f1b8e197b1b77a04992d2)
The thing is, even changing the "type" to either `CHRTYPE`,  `BLKTYPE` won't solve the issue because it changes also the file.

So this cannot be solved without patching the python `tarfile.py` module `def _create_header(info, format, encoding, errors):`.

BUT maybe this is not really an issue in Conan and we have just to patch the test.

memsharded:
Changelog: Omit
Docs: Omit

Fix a test broken in Python3.9, different checksum of files created with python tar implementation. 

Close https://github.com/conan-io/conan/issues/8020

### DIFF
--- a/conans/test/integration/conanfile/conan_data_test.py
+++ b/conans/test/integration/conanfile/conan_data_test.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import textwrap
 import unittest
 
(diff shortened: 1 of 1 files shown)


memsharded:
Not enough, followed by https://github.com/conan-io/conan/pull/8779

PR #8787, opened 2021-04-12
Feature/test package br

memsharded:
Changelog: Feature: Implement ``test_build_require`` in ``test_package/conanfile.py`` recipes, so build_requires can be tested as such.
Docs: https://github.com/conan-io/docs/pull/2081

Comes from https://github.com/conan-io/conan/pull/8627, but isolating only the ``test_package`` functionality, which is way more necessary, for example for ConanCenter.


jgsogo (commented):
Looking good, just some comment with ConanCenter in mind:
 * Right now we are only using one profile. I see the test works the same, but the `cmake_find_package` file is generated... is it a problem in this context?
 * The test suite needs a test case using `conan test` command to execute this functionality with a package that already exists (in the cache and/or in the remote), it should be exactly the same.
 * We need a test (additional feature?) where the package being tested is used as a `requirement` **and also** as a `build_requirement`. Recipes for packages like `protobuf` are already doing it.

memsharded:
> We need a test (additional feature?) where the package being tested is used as a requirement and also as a build_requirement. Recipes for packages like protobuf are already doing it.

I see. Then, this might need some changes, I'll try to come up with a proposal for this.

jgsogo:
If we go with the _explicit attribute approach_, maybe it is enough to write explicitly:

```python
        class Pkg(ConanFile):

            test_build_require = True
            test_require = True
```

memsharded:
Proposed: 
```
test_type = "build_requires", "requires"
```

To have something that scales a bit better if we have other kind of requires in the future, and also changed the name, to avoid confusion with so many "xxx-requires" things.

I have also added ``conan test`` command checks.

jgsogo (commented):
I'm testing it locally and indeed the `conan test` command is not adding the `build_requires` when using only one profile... 😞  is this something that can be fixed?

In different scenarios in ConanCenter the packages are already built and we want to run only the tests: a commit in a PR that modifies only the `test_package` files, a PR that is modifying only `test_package/***`,...

If this is not something we can fix, we need to be really confident that the same packageID will be computed if using one profile or two (which, btw, is something we need to start checking sooner than later)...

memsharded:
> I'm testing it locally and indeed the conan test command is not adding the build_requires when using only one profile... 😞 is this something that can be fixed?

This is not related to this feature. When using only 1 profile, there is only 1 context, and both requirements are the same node in host context, because they are compatible and do not conflict. Is kind of diamond with only 1 vertex. This cannot be fixed when using only 1 profile.

PR #8806, opened 2021-04-15
remove corrupt cache files and retry download 3 times

blackliner:
Changelog: Feature: Validate checksum and retry download for corrupted downloaded cache files.
Docs: omit

For testing, it would be cool if this feature could be switched on&off (default on), so the old test could remain with the feature switched off. Or mocking the Server to actually deliver corrupt files could also be an option to improve test coverage of the functionality.

I am also unsure about the logging, I have seen quite a multitude of ways to inform the user, from using the python `logging`, to `_out.warn` up to custom actions in a queue 🤔 

Issue: https://github.com/conan-io/conan/issues/8807

### DIFF
--- a/conans/client/downloaders/cached_file_downloader.py
+++ b/conans/client/downloaders/cached_file_downloader.py
@@ -55,6 +55,7 @@ class CachedFileDownloader(object):
                     self._file_downloader.download(url=url, file_path=cached_path, md5=md5,
                                                 sha1=sha1, sha256=sha256, **kwargs)
                     clean_dirty(cached_path)
+                    break
                 else:
                     # specific check for corrupted cached files, will log an error
                     # and raise after the third failed download attempt
--- a/conans/client/downloaders/cached_file_downloader.py
+++ b/conans/client/downloaders/cached_file_downloader.py
@@ -48,28 +48,21 @@ class CachedFileDownloader(object):
                 if os.path.exists(cached_path):
                     os.remove(cached_path)
                 clean_dirty(cached_path)
-            
-            for attempt in range(3):
-                if not os.path.exists(cached_path):
-                    set_dirty(cached_path)
-                    self._file_downloader.download(url=url, file_path=cached_path, md5=md5,
-                                                sha1=sha1, sha256=sha256, **kwargs)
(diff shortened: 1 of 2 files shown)


conans/client/downloaders/cached_file_downloader.py
```diff
@@ -47,19 +48,27 @@ def download(self, url, file_path=None, md5=None, sha1=None, sha256=None, **kwar
...
-                except ConanException as e:
-                    raise ConanException("%s\nCached downloaded file corrupted: %s"
-                                         % (str(e), cached_path))
+            
+            for attempt in range(3):
+                if not os.path.exists(cached_path):
+                    set_dirty(cached_path)
+                    self._file_downloader.download(url=url, file_path=cached_path, md5=md5,
+                                                sha1=sha1, sha256=sha256, **kwargs)
+                    clean_dirty(cached_path)
```
memsharded: if successful, it shouldn't iterate, because ``file_downloader.download()`` already implements a checksum inside. Most likely a break is necessary here too.

blackliner: did not know that detail, done
```suggestion
                    clean_dirty(cached_path)
                    break
```

memsharded (commented):
The ``file_downloader.download()`` already implements a ``retry`` following user configuration, and if it finally doesn't succeed, it will raise anyway, so the for loop will not iterate again. So the proposed iteration was not exactly achieving a retry. I have realized that it is enough to check for corruption first, remove if corrupted and then call download. Such download will try to download and validate checksum up to X times (typically 3)

blackliner:
@memsharded thats amazing, thanks for the commit! Anything else I need to do to get this merged? Documentation?

memsharded:
> @memsharded thats amazing, thanks for the commit! Anything else I need to do to get this merged? Documentation?

Nop, it is fine, as this could be considered a good/better natural behavior of the download cache, I think it is not necessary to document it, I have read https://docs.conan.io/en/latest/configuration/download_cache.html, and seems we are good.

Assigned it to next 1.36.

memsharded (approved):

czoido (approved):

Issue #8811, opened 2021-04-15
[bug] Regression: PR #8740 breaks adding objects to link through sharedlinkflags


Vhab:
`mimalloc/1.6.7` uses `self.cpp_info.sharedlinkflags` and `self.cpp_info.exelinkflags` to add objects to link.
(Do note, I am not the maintainer of this package, just a user)

https://github.com/conan-io/conan/pull/8740 makes a change that assumes any 'sharedlinkflags' starting with `/` is a msvc flag changes it to `-`.
This breaks absolute unix urls, like a full path to an object, by changing `/some/path.o` to `-some/path.o`.

I'm not sure if this is a regression due to the assumption made in #8740 or due to miss-use of sharedlinkflags by mimalloc. But it's regressed usage of mimalloc which previously worked.

memsharded:
Hi @Vhab 

I am not sure either it is a regression or it is an abuse of sharedlinkflags to provide something that is not a flag, but a filename. I'll check the recipe, thanks for reporting.

memsharded:
A couple of quick questions:

- This configuration of the recipe (option) is not built by default by ConanCenter, so it is difficult to know if this feature breaks or not in the CI. So far this seems a bit of abuse of the ``xxxlinkflags`` feature to provide a linkage that is not implemented in Conan. Is it possible to fallback to normal static library while we figure out? Because the right approach seems to explicitly model ``self.cpp_info.objects`` or something like that. 
- From https://github.com/microsoft/mimalloc#static-override, to link with that library object, it should be linked as the first object, I am not sure if this really works on consumers side. How are you dealing with it? How are you guaranteeing that it is the first object?

memsharded:
Furthermore, it seems the recipe ``test_package`` is not covering this case at all, so it is impossible to reproduce from the repo.

memsharded:
Lets revert that PR first: https://github.com/conan-io/conan/pull/8812, and we will take care of the issue that caused that PR later.

elqueffo:
> A couple of quick questions:
> 
> * This configuration of the recipe (option) is not built by default by ConanCenter, so it is difficult to know if this feature breaks or not in the CI. So far this seems a bit of abuse of the `xxxlinkflags` feature to provide a linkage that is not implemented in Conan. Is it possible to fallback to normal static library while we figure out? Because the right approach seems to explicitly model `self.cpp_info.objects` or something like that.

[...]

PR #8812, opened 2021-04-15
revert cmake linkflags handling

memsharded:
Changelog: Bugfix: Revert regression that replaces first ``/`` by ``-`` in ``cpp_info.xxxxlinkflags`` in _CMake_ generators because it can break passing objects and other paths that start with ``/``.
Docs: Omit

Fix https://github.com/conan-io/conan/issues/8811


SSE4 (approved):
okay for the revert
hope for the better solution in the next version

solvingj (approved):

Issue #8813, opened 2021-04-15
[feature] Recognize CMake's "Ninja Multi-Config" generator


mrjoel:
Add support for the "Ninja Multi-Config" generator of CMake, which was added in CMake 3.17.

- [X] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Closed by https://github.com/conan-io/conan/pull/8814

PR #8814, opened 2021-04-15
Add support for "Ninja Multi-Config" CMake generator (#8813)

mrjoel:
The "Ninja Multi-Config" generator was added in CMake 3.17. Add
support for recognizing it as multi-config from
is_multi_configuration(), as well as inserting needed envvars as with
regular (single configuration) Ninja usage.

Signed-off-by: Joel Johnson <mrjoel@lixil.net>

Changelog: Fix: Recognize ``Ninja Multi-Config`` as a  CMake multi-configuration generator.
Docs: Omit

### DIFF
--- a/conan/tools/cmake/utils.py
+++ b/conan/tools/cmake/utils.py
@@ -7,7 +7,7 @@ from conans.errors import ConanException
 def is_multi_configuration(generator):
     if not generator:
         return False
-    return "Visual" in generator or "Xcode" in generator
+    return "Visual" in generator or "Xcode" in generator or "Multi-Config" in generator
 
 
 def get_generator(conanfile):
--- a/conans/client/build/cmake.py
+++ b/conans/client/build/cmake.py
@@ -226,13 +226,14 @@ class CMake(object):
         context = tools.no_op()
 
         if (is_msvc or is_clangcl) and platform.system() == "Windows":
-            if self.generator in ["Ninja", "NMake Makefiles", "NMake Makefiles JOM"]:
+            if self.generator in ["Ninja", "Ninja Multi-Config",
+                                  "NMake Makefiles", "NMake Makefiles JOM"]:
                 vcvars_dict = tools.vcvars_dict(self._settings, force=True, filter_known_paths=False,
                                                 output=self._conanfile.output)
                 context = _environment_add(vcvars_dict, post=self._append_vcvars)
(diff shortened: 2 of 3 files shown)


memsharded (commented):
Thanks for contributing this. If we want to add support for Ninja Multi-Config, we need to do a few more things:

- It is very important that changes to build systems are at least (or only) targeted to the new tools in ``conan.tools.cmake``, not only the old ``conans.xxxx`` space.
- We are adding functional tests for almost all things. This was merged just today: https://github.com/conan-io/conan/pull/8005. We really need to make sure things really work, otherwise there is too much untested code that easily breaks. 

We really want to add support for Ninja Multi-Config, but it will require a bit more effort. Adding tests is not that straightforward, we can help with this at some point if necessary.

mrjoel:
Yep, I just needed a minimal `is_multi_configuration` test to work for some internal items. I started to take a look at adding test cases but also saw it as not straightforward. I didn't see any prior activity on the Ninja Multi-Config generator though so wanted to get something out as a starting point.

mrjoel:
>     * It is very important that changes to build systems are at least (or only) targeted to the new tools in `conan.tools.cmake`, not only the old `conans.xxxx` space.

What additional changes would you expect here? I did look at a brief look in `conan/tools/cmake` and the only Ninja usage are currently `"Ninja" in generator` which will match the multi-config one as well (and the behavior should also be desired), or in `toolchain.py` uses the `is_multi_configuration(generator)` path.

memsharded:
> What additional changes would you expect here? I did look at a brief look in conan/tools/cmake and the only Ninja usage are currently "Ninja" in generator which will match the multi-config one as well (and the behavior should also be desired), or in toolchain.py uses the is_multi_configuration(generator) path.

Yes, this was more a general consideration of how to think regarding changes in build system integrations, and being aware that the future integrations for Conan 2.0 are those in ``conan.tools.xxx``.

Sure, thanks, this is good as a starting point and will help to push things. We will try to help with the tests, it is important to add some, specially in the new part that is currently under heavy development.

memsharded:
We will be adding tests for Ninja multi-config in CI when possible, but that might take some time. As this change seems no risk at all, I think it is ok to merge it now for 1.36, but I have changed the changelog, until we have full proper testing, better not to say that we added support for it.

Issue #7105, opened 2020-05-27
Reorganize tools


memsharded:
The current tools approach has some issues:

- Internal code ugliness, due to global objects (output, requester, being globally managed: https://github.com/conan-io/conan/issues/4933)
- No clear import and usage pattern. Some are ``from conans import CMake, load``, but ``save`` is ``from conans.tools import save``.

There are few alternatives:

- Use a conanfile attribute object:
```python

def build(self):
     self.tools.files.save("myfile.txt", "mycontent")
     cmake = self.tools.CMake(...) 
```

- Import an object (hacking a bit the import system)

```python
from conans import tools  # tools is an object we control

def build(self):
     tools.files.save("myfile.txt", "mycontent")
     cmake = tools.CMake(...) 
```

- All tools use a ``conafile/ self`` argument as the first argument. Access/inject what is necessary via a ``conanfile`` private reference to collaborators (output, requester, config

```python
from conans.tools.files import load, save
from conans.tools.build import CMake

def build(self):
     save(self, "myfile.txt", "mycontent")  # A bit ugly, but uniform.
     cmake = CMake(...) 
```

Related:
- https://github.com/conan-io/conan/issues/4933

TheQwertiest:
My (zero-weighted) vote: solution 2.
Solution 1 and Solution 3 are not good for the same reason: some tools do not require `ConanFile` at all, so it doesn't make much sense to drag `ConanFile` object or definition in places where it's not needed at all (e.g. `tools.files.save()`).

jgsogo:
I'm not sure if adding scope to the tools is something we should do, it would make it difficult to find the tools. I mean, from the user POV, I prefer `tools.save` over `tools.files.save`.... so I don't need to check the documentation every time I want to use a tool.

About the main issue here, I like option (2), but I'm not sure if it will require too much monkey-patching (keep in mind that it cannot block eventual parallelism in Conan)

memsharded:
> so it doesn't make much sense to drag ConanFile object or definition in places where it's not needed at all (e.g. tools.files.save()).

Until you want ``save()`` to be able to emit a warning because it is overwriting an existing file, for example, that looks like:

```bash
pkg/0.1@user/testing: warn: save() method is overwriting "myfile.txt"
```
Then you really need the ``save()`` method to have access to the ``conanfile`` to get the ``pkg/0.1@user/testing``. 

[...]

PR #8819, opened 2021-04-16
Feature : cross-building tests for new AutoTools build helper

### LINKED ISSUE #8762: [feature] Implementation of cross-build triplets in new conan.tools.gnu AutotoolsToolchain
- Triplets must be defined at toolchain time (``generate()``), and stored in a file or any mechanism that makes as easy as possible to be used by users. In the worst case, yes, copy&paste from file to avoid retyping.
- The ``Autootools()`` helper in the ``build()`` method will just use this from the result of ``generate()``
- Use exclusively the build/host contexts and profiles. No usage of any detection or current platform, everything must come from profiles.
- If necessary use the new [conf] to define extra complements, but try to avoid as much as possible and draft a minimalistic approach
- It is fine if it only starts with build/host but not the target one
- Try to include a functional test that really cross-build, even if it doesn't pass CI now (we can talk to @czoido to install something, should be the easiest to install cross-compiler), you can use an @pytest.mark.xfail at the moment

SSE4:
closes: #8762
Changelog: Feature: Add cross-building tests for new AutoTools build helper.
Docs: omit

### DIFF
--- a/conan/tools/gnu/autotoolsgen.py
+++ b/conan/tools/gnu/autotoolsgen.py
@@ -34,4 +34,4 @@ class AutotoolsGen:
             build_env.save_sh("conanbuildenv.sh")
             run_env.save_sh("conanrunenv.sh")
 
-        self.toolchain.generate()
+        self.toolchain.generate_args()
--- a/conan/tools/gnu/autotoolsgen.py
+++ b/conan/tools/gnu/autotoolsgen.py
@@ -33,3 +33,5 @@ class AutotoolsGen:
         else:
             build_env.save_sh("conanbuildenv.sh")
             run_env.save_sh("conanrunenv.sh")
+
+        self.toolchain.generate()
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -1,7 +1,12 @@
+import json
+
+from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
 from conan.tools._compilers import architecture_flag, build_type_flags
 from conan.tools.env import Environment
 # FIXME: need to refactor this import and bring to conan.tools
 from conans.client.build.cppstd_flags import cppstd_flag_new
+from conans.client.tools.oss import cross_building, get_cross_building_settings, get_gnu_triplet
+from conans.util.files import save
 
 
 class AutotoolsToolchain:
@@ -29,6 +34,15 @@ class AutotoolsToolchain:
         self.arch_flag = architecture_flag(self._conanfile.settings)
         # TODO: This is also covering compilers like Visual Studio, necessary to test it (&remove?)
         self.build_type_flags = build_type_flags(self._conanfile.settings)
+
+        self._host = None
+        self._build = None
+        self._target = None
+
+        if cross_building(self._conanfile):
+            os_build, arch_build, os_host, arch_host = get_cross_building_settings(self._conanfile)
+            self._host = get_gnu_triplet(os_host, arch_host)
+            self._build = get_gnu_triplet(os_build, arch_build)
 
     def _rpaths_link(self):
         # TODO: Not implemented yet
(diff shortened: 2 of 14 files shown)


conan/tools/gnu/autotoolsgen.py
```diff
@@ -33,3 +33,5 @@ def generate(self):
         else:
             build_env.save_sh("conanbuildenv.sh")
             run_env.save_sh("conanrunenv.sh")
+
+        self.toolchain.generate()
```
memsharded: This is confusing, as it will generate an extra, unused file. 
Move the AutotoolsToolchain CONAN_TOOLCHAIN_ARGS_FILE functionality to its own method, and call just that method.

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -1,7 +1,12 @@
+import json
+
+from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
 from conan.tools._compilers import architecture_flag, build_type_flags
 from conan.tools.env import Environment
 # FIXME: need to refactor this import and bring to conan.tools
 from conans.client.build.cppstd_flags import cppstd_flag_new
+from conans.client.tools.oss import cross_building, get_cross_building_settings, get_gnu_triplet
```
memsharded: In general try to avoid things from old ``conans.client....`` scope. The ``cross_building, get_cross_building_settings, get_gnu_triplet`` should be copied here, because they are going to be modernized. This one already has a FIXME above, so it can be done in a later PR, but at least the idea for other PRs.

memsharded (changes requested):

conan/tools/gnu/autotoolsgen.py
SSE4: I am not sure I got the idea behind the new method, but I'll do

conan/tools/gnu/autotoolstoolchain.py
SSE4: okay, let's copy them

conan/tools/gnu/autotoolsgen.py
memsharded: To avoid the creation of an extra ``conanautotooltoolchain.sh`` script that will not be used.

conan/tools/gnu/autotoolstoolchain.py
SSE4: made a copies of `cross_building`, `get_cross_building_settings`, `get_gnu_triplet` into `conan.tools.oss`. it's like a spider web, one method pulling another one from the `conans.client.tools`, so I had to cut `OSInfo` to only used methods to avoid copying almost everything now (e.g. it requires `which`, `environment_append` and so on for full functionality, it actually does to much - classic god object).

conan/tools/gnu/autotoolsgen.py
SSE4: okay, I've extracted the method `generate_args`.
still would be nice if you could explain your idea a bit more, since for CMake toolchain we don't do a file, it just writes `CONAN_TOOLCHAIN_ARGS_FILE` inside the `generate`.
I am not sure what is an extra and unused file, as `CONAN_TOOLCHAIN_ARGS_FILE` is definitely used.
and as I understand the point of `generate` is for developer to have exactly all the files needed, so I am not sure why some files belong to `generate`, but some don't.

conan/tools/oss/__init__.py
```diff
@@ -0,0 +1,3 @@
+from conan.tools.oss.cross_building import cross_building
```
memsharded: This is not a new namespace at the moment (https://github.com/conan-io/conan/issues/7105)

Put them atm as private inside ``tools.gnu``

memsharded (approved):

conan/tools/oss/__init__.py
SSE4: okay, moved one more time, to `tools.gnu`

conan/tools/gnu/autotoolstoolchain.py
memsharded: yes, this is one of the purposes of doing this copy. All this should be distilled and reduced much, specially if using both profiles, and not needing to detect anything at all. This will follow in future PRs.

memsharded (approved):

PR #8821, opened 2021-04-16
Listing of tools and core configurations

### LINKED ISSUE #8696: [feature] Implement discoverability for [conf] items
Some usages of the new `[conf]` construct are starting to be instrumented throughout the codebase. Currently, it's primarily in generators and other experimental places. While it's still early for `conf`, I think we should consider providing a good experience for users to print out all available `conf` items , current values, etc.  

Also, I think we should do something different from `conan config get`, which prints the config file and doesn't reflect environment variables, and doesn't show a comprehensive list of settings which can be set. Instead, I suggest doing something similar to `argparse`, where every possible `conf` item within the codebase is declared very early on in the main setup.  Such a strategy presents some significant challenges. For example, the actual values of the variables are reconciled as "instances" per-package during various different times.  Still, perhaps it's possible.  Another strategy might be annotations. 

While I don't know all the options for implementing, I still want to suggest the feature be considered and looked into.  We keep adding new `[conf]` items in each new release of Conan.  I feel that as users get used to adding `conf` items to profiles, it will become increasingly annoying for users to figure out what all the possible `conf` items are.  They'll have to check their Conan version, then go to the specific version of the docs, and then search the docs to find out what `conf` items are available.  Exposing them at the CLI for discovery similar to `conan --help` and `conan install --help` seems a much better pattern if it can be implemented.

uilianries:
Some points need to be defined yet:

For global printing (conan config get), we can group conan.conf, global.conf and profile[conf] at same output.

However, get single item, update and adding a new one is more complicated. So far `conan config` only manages `conan.conf` file. We will need to support a new argument, like `--global`.

Also, we can't print all possible configurations to be used by [conf], because `ConfDefition` is empty, and all items are read when a specific condition occurs. We would need to centralize all configuration again, just like we did for conan.conf, but I'm sure if we want it.

Another point, we have 3 configurations, conan.conf, global.conf and profile[conf]. Printing all this is a mess, we could create a new layer to abstract them, otherwise we would need called separately. See [line 629](https://github.com/conan-io/conan/compare/develop...uilianries:feature/manage-conf?expand=1#diff-c15f4aacf65aa58b31a041bac63e4dd7337355d9ec44a8f32b2a5815d5eb5ea9R630)

Changelog: Feature: List tools and core from profile and _global.conf_.
Docs: https://github.com/conan-io/docs/pull/2077

closes #8696


conans/client/conan_api.py
```diff
@@ -626,6 +626,11 @@ def config_get(self, item):
             result = self.app.config.storage_path
         else:
             result = self.app.config.get_item(item)
+        if item is None:
+            # TODO: Provide a helper
+            result += "\n[conf]\n"
+            result += self.app.cache.new_config.dumps() + "\n"
```
memsharded: No, this is not the feature. It is not about listing current configuration.

It is about listing all possible configurations that are in our codebase, together with an explanation/documentation of what it does. Kind of ``conan --help`` for commands, but for configurations. Could be in ``conan config --help``, for example, or ``conan config list``.

At some point we might use it to validate user inputs, to avoid typos, for example.

memsharded (changes requested):

conans/client/conan_api.py
uilianries: We don't populate new_config, we just read if a key is present or not. For example: https://github.com/conan-io/conan/blob/12600f3975704031782bfe0dd3c571a2ac36a05d/conans/client/conf/required_version.py#L33

We will need to centralize them and populate with empty values.

memsharded: Centralize their definition:
```python
confs = {
    "core.required_conan_version": "help for this config",
    "tools.microsoft.....": "help for this one too",
}
```

uilianries:
@memsharded I've added a new subcommand: `conan config list`, which should list all possible configurations.

I also collected all core and tools key for new_config. However, it's a bit fragile, because all developers when adding a new key, need to remember about DEFAULT_CONFIGURATION too. Another option is populating ConfDefinition with default values + help, but still we can't avoid people adding new keys, but not updating ConfDefinition.

The current output is:

```
$ conan config list
Supported Conan conan.conf properties:
core.required_conan_version: Will raise if the current Conan version does not match the defined version range.
core.package_id.msvc_visual_incompatible: Allows opting-out the fallback from the new msvc compiler to the Visual Studio compiler existing binaries
tools.microsoft.msbuild:verbosity: Verbosity level for MSBuild: 'Quiet', 'Minimal', 'Normal', 'Detailed', 'Diagnostic'
tools.microsoft.msbuild:max_cpu_count: Argument for the /m (/maxCpuCount) when running MSBuild
tools.microsoft.msbuild:vs_version: Defines the compiler version when using using the new msvc compiler
tools.microsoft.msbuilddeps:exclude_code_analysis: Suppress MSBuild code analysis for patterns
tools.microsoft.msbuildtoolchain:compile_options: Dictionary with MSBuild compiler options
tools.build:processes: Default jobs number
tools.ninja:jobs: Argument for the --jobs parameter when running Ninja generator
tools.gnu.make:jobs: Argument for the -j parameter when running Make generator
tools.gnu:make_program: Indicate path to make program
tools.env.virtualenv:auto_use: Automatically activate virtualenvs when changing into a directory
tools.cmake.cmaketoolchain:msvc_parallel_compile: Argument for the /MP when running msvc
tools.cmake.cmaketoolchain:find_package_prefer_config: Argument for the CMAKE_FIND_PACKAGE_PREFER_CONFIG
tools.android:ndk_path: Argument for the CMAKE_ANDROID_NDK
tools.files.download:retry: Number of retries in case of failure when downloading
tools.files.download:retry_wait: Seconds to wait between download attempts
```

memsharded:
> I also collected all core and tools key for new_config. However, it's a bit fragile, because all developers when adding a new key, need to remember about DEFAULT_CONFIGURATION too. Another option is populating ConfDefinition with default values + help, but still we can't avoid people adding new keys, but not updating ConfDefinition.

Yes, I know. We should don't forget now in code reviews. But we should remind to document them anyway, so at least let's start to build the DB now, we could add later checks/validation against this DB (so people adding new confs without adding them here will raise)

conans/test/integration/command/config_test.py
```diff
@@ -164,3 +166,16 @@ def test_init_overwrite(self):
     def _assert_dict_subset(self, expected, actual):
         actual = {k: v for k, v in actual.items() if k in expected}
         self.assertDictEqual(expected, actual)
+
+    @parameterized.expand([("", True), ("--raw", False)])
```
memsharded: Please recall to write new tests in pytest framework better.

conans/client/command.py
```diff
@@ -590,6 +592,8 @@ def config(self, *args):
         set_subparser.add_argument("item", help="'item=value' to set")
         init_subparser.add_argument('-f', '--force', default=False, action='store_true',
                                     help='Overwrite existing Conan configuration files')
+        list_subparser.add_argument('-r', '--raw', default=False, action='store_true',
```
memsharded: Avoid the --raw for now, lets keep it simple. Users shouldn't parse this output for anything.

conans/client/command.py
```diff
@@ -627,6 +631,11 @@ def config(self, *args):
                                               target_folder=args.target_folder)
         elif args.subcommand == 'init':
             return self._conan.config_init(force=args.force)
+        elif args.subcommand == "list":
+            if not args.raw:
+                self._out.info("Supported Conan conan.conf properties:")
```
memsharded:
```suggestion
                self._out.info("Supported Conan *experimental* conan.conf properties:")
```

memsharded (commented):
Looks good, minor fixes, lets try to merge it for Conan 1.36

conans/test/integration/command/config_test.py
uilianries: Sure, actually I'll do better, I can update this file to pytest on a next PR, so I don't mess this one.

conans/client/command.py
uilianries: Done

conans/client/command.py
uilianries: Done

uilianries:
@memsharded All done, please, review again.

memsharded (approved):

uilianries:
Docs PR: https://github.com/conan-io/docs/pull/2077

PR #8835, opened 2021-04-20
Remove not-export argument for conan crete

czoido:
Changelog: Remove: Remove `--not-export` argument for conan create.
Docs: omit


memsharded (approved):
Wow. It is not even tested at all 😱 Great that we are removing it.

PR #8843, opened 2021-04-21
CMakeDeps reading name from cmake_find_package_multi too 

lasote:
Changelog: Feature: The `CMakeDeps` generator will print CMake traces with the declared targets. e.g: `Target declared: 'OpenSSL::Crypto'`.
Docs: omit

ALSO A MINI FEATURE FOR PRINTING THE FOUND (GENERATED) TARGETS FOR THE CONSUMER.

### DIFF
--- a/conan/tools/cmake/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps.py
@@ -552,17 +552,11 @@ endforeach()
     def get_name(self, req):
         ret = req.new_cpp_info.get_property("cmake_target_name", self.name)
         if not ret:
-            # The old cpp info
-            ret = req.cpp_info.get_name(self.name, default_name=False)
-        if not ret:
             ret = req.cpp_info.get_name("cmake_find_package_multi", default_name=False)
         return ret or req.ref.name
 
     def get_filename(self, req):
         ret = req.new_cpp_info.get_property("cmake_file_name", self.name)
-        if not ret:
-            # The old cpp info
-            ret = req.cpp_info.get_filename(self.name, default_name=False)
         if not ret:
             ret = req.cpp_info.get_filename("cmake_find_package_multi", default_name=False)
         return ret or req.ref.name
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/cmakedeps.py
```diff
@@ -552,26 +553,32 @@ def get_name(self, req):
         ret = req.new_cpp_info.get_property("cmake_target_name", self.name)
         if not ret:
             # The old cpp info
-            ret = req.cpp_info.get_name(self.name)
+            ret = req.cpp_info.get_name(self.name, default_name=False)
```
memsharded: This is for packages that have declared ``self.cpp_info.names["CMakeDeps"]``, isn't it? If this is the case, maybe we want to already deprecate it? ConanCenter is not using it at all, I don't think users have started doing this much either. So we can go with:
- New ``cmake_target_name`` property
- Or fallback to ``get_name("cmake_find_package_multi", default_name=False)`` for compatibility with ConanCenter and other existing recipes.

conan/tools/cmake/cmakedeps.py
```diff
@@ -552,26 +553,32 @@ def get_name(self, req):
...
+        if not ret:
+            ret = req.cpp_info.get_name("cmake_find_package_multi", default_name=False)
         return ret or req.ref.name
 
     def get_filename(self, req):
         ret = req.new_cpp_info.get_property("cmake_file_name", self.name)
         if not ret:
             # The old cpp info
-            ret = req.cpp_info.get_filename(self.name)
+            ret = req.cpp_info.get_filename(self.name, default_name=False)
```
memsharded: Same as above

conan/tools/cmake/cmakedeps.py
```diff
@@ -552,26 +553,32 @@ def get_name(self, req):
...
             if req.ref.name == comp_name:  # foo::foo might be referencing the root cppinfo
                 return self.get_name(req)
             raise KeyError(comp_name)
-        ret = req.new_cpp_info.components[comp_name].get_property("cmake_target_name",
-                                                                  self.name)
+        ret = req.new_cpp_info.components[comp_name].get_property("cmake_target_name", self.name)
         if not ret:
             # The old cpp info
-            ret = req.cpp_info.components[comp_name].get_name(self.name)
+            ret = req.cpp_info.components[comp_name].get_name(self.name, default_name=False)
```
memsharded: Same?

memsharded (approved):

PR #8849, opened 2021-04-22
Use cmake 3.19 in ci for tests that use Xcode cmake generator

czoido:
Changelog: omit
Docs: omit

CMake XCode generator does not work for latest XCode versions for [versions of CMake <3.19](https://cmake.org/cmake/help/latest/release/3.19.html).

### DIFF
--- a/conans/test/functional/generators/cmake_find_package_multi_test.py
+++ b/conans/test/functional/generators/cmake_find_package_multi_test.py
@@ -549,6 +549,7 @@ class TestNoNamespaceTarget:
 
     @pytest.mark.skipif(platform.system() != "Darwin", reason="Requires Macos")
     @pytest.mark.tool_xcode
+    @pytest.mark.tool_cmake(version="3.19")
     def test_multi_generator_macos(self):
         t = self.t
         with t.chdir('multi_macos'):
--- a/conans/test/functional/generators/cmake_apple_frameworks_test.py
+++ b/conans/test/functional/generators/cmake_apple_frameworks_test.py
@@ -80,6 +80,7 @@ class CMakeAppleFrameworksTestCase(unittest.TestCase):
         self.t.run("build .")
         self._check_frameworks_found(str(self.t.out))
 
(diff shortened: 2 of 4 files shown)


SSE4 (approved):

conans/test/functional/generators/cmake_find_package_multi_test.py
```diff
@@ -549,6 +549,7 @@ def test_multi_generator_windows(self):
 
     @pytest.mark.skipif(platform.system() != "Darwin", reason="Requires Macos")
     @pytest.mark.tool_xcode
+    @pytest.mark.tool_cmake(version="3.19")
```
memsharded, line 552: I am a bit concerned that we are introducing a hard requirement to cmake 3.19, but users will not know that easily, if something fail, specifically because we declare that we support 3.15. 

Is there a way we could introduce the 3.19 requirements so users get the right informative error? Maybe in ``CMakeToolchain`` when ``generator==xcode`` we can introduce a cmake_minimum_required(3.19)?

czoido: I agree we should introduce some kind of warning. Also to consider, from CMake 3.19 using generator Xcode is only compatible with Xcode > 12.0 but there's an environment variable [CMAKE_XCODE_BUILD_SYSTEM](https://cmake.org/cmake/help/latest/variable/CMAKE_XCODE_BUILD_SYSTEM.html#variable:CMAKE_XCODE_BUILD_SYSTEM) to switch to the XCode 11.X build system, also can be set by [the command line](https://cmake.org/cmake/help/latest/release/3.19.html#generators) (`-T buildsystem=1`).

SSE4: yeah, that seems to require an additional effort from `CMakeToolchain` side to properly support `CMAKE_XCODE_BUILD_SYSTEM`, but that's probably out of the scope of this PR

PR #8861, opened 2021-04-24
AptTool: add repo key before running apt-add-repository (#8857)

### LINKED ISSUE #8857: [bug] AptTool's use of apt-add-repository may update the apt index before the repository's key is added
  * Compiler+version: gcc 9.3.0
  * Conan version: 1.35.1
  * Python version: Python 3.8.5

In a `system_requirements` method, run

```python
if os_info.linux_distro == "ubuntu":
    apt = SystemPackageTool(tool=AptTool())
    apt_key = "https://apt.llvm.org/llvm-snapshot.gpg.key"

    if os_info.os_version == "20.04":
	apt.add_repository("'deb http://apt.llvm.org/focal/ llvm-toolchain-focal main'", apt_key)
``` 
On Ubuntu 20.04, `apt-add-repository` (which is invoked by `apt.add_repository`) defaults to running `apt update` after adding the repository line to apt's sources.  However, for a repository like `apt.llvm.org` whose signing key must usually be installed on the system, this will fail.

This is because `AptTool` installs the repository key _after_ running `apt-add-repository`.

<details>
<summary>Log of a `conan install` of a Conanfile containing the above `system_requirements` code</summary>

```
recipes/toolchain-clang# conan create . toolchain-clang/11.0@
Exporting package recipe
toolchain-clang/11.0: The stored package has not changed
toolchain-clang/11.0: Exported revision: 29c135f3c45a6d6410c47bd6ecd52928
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++
compiler.version=9
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

toolchain-clang/11.0: Forced build from source
Installing package: toolchain-clang/11.0
Requirements
    toolchain-clang/11.0 from local cache - Cache
Packages
    toolchain-clang/11.0:4db1be536558d833e52e862fd84d64d75c2b3656 - Build

Running: sudo apt-add-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal main'
Hit:1 http://security.ubuntu.com/ubuntu focal-security InRelease
Hit:2 http://archive.ubuntu.com/ubuntu focal InRelease                            
Hit:4 http://archive.ubuntu.com/ubuntu focal-updates InRelease                    
Hit:5 http://archive.ubuntu.com/ubuntu focal-backports InRelease
Get:3 https://apt.llvm.org/focal llvm-toolchain-focal InRelease [5526 B]
Err:3 https://apt.llvm.org/focal llvm-toolchain-focal InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 15CF4D18AF4F7421
Reading package lists... Done
W: GPG error: https://apt.llvm.org/focal llvm-toolchain-focal InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 15CF4D18AF4F7421
E: The repository 'http://apt.llvm.org/focal llvm-toolchain-focal InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
toolchain-clang/11.0: ERROR: while executing system_requirements(): Command 'sudo apt-add-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal main'' failed
```

</details>

hannahwhy:
Changelog: Bugfix: AptTool: add repo key before running apt-add-repository.
Docs: omit

Closes #8857

In Ubuntu 20.04 (and maybe earlier), the behavior of

    apt-add-repository 'deb https://example.com/repo example main'

is to add the named repository to apt's sources *and* to update the
package index.  If the new repository's package key is not already in
apt's keyring, this will fail.

It is possible to change this behavior by running apt-add-repository -n,
but adding the key before running apt-add-repository also works, doesn't
require new command-line switches, and should be backwards-compatible
with apt-add-repository versions that might not behave as described
above.

### DIFF
--- a/conans/client/tools/system_pm.py
+++ b/conans/client/tools/system_pm.py
@@ -272,11 +272,11 @@ class NullTool(BaseTool):
 
 class AptTool(BaseTool):
     def add_repository(self, repository, repo_key=None):
-        _run(self._runner, "%sapt-add-repository %s" % (self._sudo_str, repository),
-             output=self._output)
         if repo_key:
             _run(self._runner, "wget -qO - %s | %sapt-key add -" % (repo_key, self._sudo_str),
                  output=self._output)
+        _run(self._runner, "%sapt-add-repository %s" % (self._sudo_str, repository),
+             output=self._output)
 
     def update(self):
         _run(self._runner, "%sapt-get update" % self._sudo_str, output=self._output)
--- a/conans/test/unittests/client/tools/system_pm_test.py
+++ b/conans/test/unittests/client/tools/system_pm_test.py
@@ -100,10 +100,10 @@ class SystemPackageToolTest(unittest.TestCase):
                 sudo_cmd = "sudo " if isatty else "sudo -A "
 
             runner = RunnerOrderedMock()
-            runner.commands.append(("{}apt-add-repository {}".format(sudo_cmd, repository), 0))
             if gpg_key:
                 runner.commands.append(
                     ("wget -qO - {} | {}apt-key add -".format(gpg_key, sudo_cmd), 0))
+            runner.commands.append(("{}apt-add-repository {}".format(sudo_cmd, repository), 0))
             if update:
                 runner.commands.append(("{}apt-get update".format(sudo_cmd), 0))
 


hannahwhy:
As described in https://github.com/conan-io/conan/issues/8857, this proposed change will have AptTool running an extra index update, which does waste some time.  However:

1. The approach here doesn't change the interface expected from `apt-add-repository`.
2. Two index updates already occur under Ubuntu 20.04 without this change.

The version of `apt-add-repository` in Ubuntu 20.04 has a `-n, --no-update` option that skips the index update.  If we use that, then the order of operations in AptTool can remain the way they are.  However, I have yet not confirmed that `-n` and/or `--no-update` exist in all versions of `apt-add-repository`, not only in Debian and Ubuntu but also other Debian derivatives.

hannahwhy:
Looks like I'm running afoul of some expected command sequences in the test suite.  I'll check those in a little while.

memsharded:
Note for the reviewers:

- The disadvantage of this solution is that it will run an "update" twice, second might be redundant, and it might be less efficient. This doesn't happen for all cases (only when adding repository and providing key), but better to be sure impact is reasonable.

lasote (approved):
I think the added time is reasonably low and probably we don't have a "safer" way to do it without breaking. For me, it could be merged.

memsharded (approved):

PR #8878, opened 2021-04-27
autotools: apply unix_path to CONAN_MAKE_PROGRAM

ericLemanissier:
Changelog: Bugfix: enable use of `make/*` recipe in msys environment

this intends to fix the use of `make/*` recipe in msys environment. [Currently, the error is](https://ci.appveyor.com/project/SpaceIm/conan-libiconv/builds/38884285/job/lrjjsifc5m39chyo#L772):
```
libiconv/1.16@SpaceIm/testing: run_in_windows_bash: "C:\Program Files\Git\usr\bin\bash.exe" --login -c ^"cd \^"/c/users/appveyor/.conan/data/libiconv/1.16/spaceim/testing/build/dbc97e02589c20e3ca37b15d05c8e5853fae244f/source_subfolder\^" ^&^& PATH=\^"/c/users/appveyor/.conan/data/mingw_installer/1.0/conan/stable/package/4b8f5a096ecfd6aabff608521169f167b01837ec/bin:$PATH\^" ^&^& C:\Users\appveyor\.conan\data\make\4.2.1\_\_\package\0a420ff5c47119e668867cdb51baff0eca1fdb68\bin\gnumake.exe -j2 ^"
/usr/bin/bash: C:Usersappveyor.conandatamake4.2.1__package0a420ff5c47119e668867cdb51baff0eca1fdb68bingnumake.exe: command not found
```

CC @SpaceIm

### DIFF
--- a/conans/client/build/autotools_environment.py
+++ b/conans/client/build/autotools_environment.py
@@ -234,7 +234,10 @@ class AutoToolsBuildEnvironment(object):
         if not self._conanfile.should_build:
             return
         conan_v2_error("build_type setting should be defined.", not self._build_type)
-        make_program = os.getenv("CONAN_MAKE_PROGRAM") or make_program or "make"
+        conan_make_program = os.getenv("CONAN_MAKE_PROGRAM")
+        if conan_make_program and self._win_bash:
+            conan_make_program = tools.unix_path(conan_make_program, subsystem)
+        make_program = conan_make_program or make_program or "make"
         with environment_append(vars or self.vars):
             str_args = args_to_string(args)
             cpu_count_option = (("-j%s" % cpu_count(output=self._conanfile.output))
--- a/conans/client/build/autotools_environment.py
+++ b/conans/client/build/autotools_environment.py
@@ -236,7 +236,7 @@ class AutoToolsBuildEnvironment(object):
         conan_v2_error("build_type setting should be defined.", not self._build_type)
         conan_make_program = os.getenv("CONAN_MAKE_PROGRAM")
         if conan_make_program and self._win_bash:
-            conan_make_program = tools.unix_path(conan_make_program, subsystem)
+            conan_make_program = tools.unix_path(conan_make_program, path_flavor=self.subsystem)
         make_program = conan_make_program or make_program or "make"
         with environment_append(vars or self.vars):
             str_args = args_to_string(args)


ericLemanissier:
@madebr what do you think of this PR ?

madebr:
The change looks legit to me.
`tools.unix` is only applied when in winbash.

conans/client/build/autotools_environment.py
```diff
@@ -234,7 +234,10 @@ def make(self, args="", make_program=None, target=None, vars=None):
         if not self._conanfile.should_build:
             return
         conan_v2_error("build_type setting should be defined.", not self._build_type)
-        make_program = os.getenv("CONAN_MAKE_PROGRAM") or make_program or "make"
+        conan_make_program = os.getenv("CONAN_MAKE_PROGRAM")
+        if conan_make_program and self._win_bash:
+            conan_make_program = tools.unix_path(conan_make_program, path_flavor=self.subsystem)
```
memsharded, line 239: The problem with this is that this is based on auto-detection, not the real subsystem:
```python
self.subsystem = OSInfo().detect_windows_subsystem() if self._win_bash else None
```
There are reports already, that this is buggy, and might take the wrong subsystem. So this PR might break existing users, I'd prefer not to take the risk, and the users will probably have the workaround of specifying the ``CONAN_MAKE_PROGRAM`` directly in the subsystem path style? As it is very likely that make itself is in such subsytem, that might be good enough?

As we are moving fast towards the new AutotoolsDeps, AutotoolsToolchain, also re-doing the ``win_bash`` code, and basing it on user configuration instead of auto-detection, I'd prefer to wait and submit this issue there (if relevant, the ``CONAN_MAKE_PROGRAM`` will not be an env-var anymore, but ``[conf]`` instead).

ericLemanissier: so you think correcting the path directly in https://github.com/conan-io/conan-center-index/blob/master/recipes/make/all/conanfile.py#L55 is better right ?

EDIT: It seems to me that if the subsystem is not properly detected, users are already broken, because it is used few lines below in `self._conanfile.run` invocation

memsharded: Well, changing it in the recipe it won't be very pretty either, but maybe it has some smaller chance to break users.

Yes, I agree that it might be broken, but at least it is the "same broken". If it was broken for your case in the past it will continue to be broken, if it happen to work, even by luck, it will keep working. We have had too many experiences lately, breaking users with apparently insignificant changes (we have passed the limit of Hyrums Law clearly), so my current preference is try to avoid this kind of change unless very necessary, and move new changes, experimentation, etc to the new domains in preparation for Conan 2.0 (``conan.tools`` new stuff)

ericLemanissier: actually it would not be correct to transform the path in [here](https://github.com/conan-io/conan-center-index/blob/master/recipes/make/all/conanfile.py#L55), because there is no way we can know at that point if make will be invoked in a bash environment or not.

ericLemanissier:
I have no more time for this, and there does not seem to be a possible solution yet

PR #8895, opened 2021-04-28
trying config from build_require

memsharded:
Changelog: Feature: Build-requires can define [conf] for its consumers.
Docs: https://github.com/conan-io/docs/pull/2107

- It is injected directly in the conanfile, just before calling ``generate()``, so toolchains will have the [conf] objects available.
- Generating file like ``conandeps.conf`` seems useless, as the classes like ``CMakeToolchain`` inside ``generate()`` need to have the ``conanfile.conf`` completely defined.
- The new conanfile.conf will remain updated in the ``build()`` and ``package()`` methods.
- Local flow is a problem in Conan 1.X, but as the dependency graph will be computed in Conan 2.0, this will not be a problem.
- Native build by users ``cmake ....`` instead of ``conan build`` does not require and cannot use a conandeps.conf file.


lasote (commented):
Looking good! I think we could merge it and give it a try

czoido (approved):

PR #8901, opened 2021-04-29
simplifying [conf] access

memsharded:
Changelog: Fix: Simplify code access to [conf] variables removing attribute based access.
Docs: Omit

#tags: slow


lasote (approved):
It is a pleasure to see less "magic"

czoido (approved):

PR #8912, opened 2021-05-03
Fixed #8884 bug

### LINKED ISSUE #8884: [bug] Conan client fails when server charset is specified.
I'm experimenting with implementing a conan server. While testing said server I ran into problems because my server fully qualifies the "Content-Type" HTTP header to "application/json; charset=utf-8". The conan client though errors when talking to that server with this error: `Response from remote is not json, but 'application/json; charset=utf-8'.`. JSON content requires utf-8 encoding and charset, which makes the addition in the content-type of the charset optional by the server. But regardless of what the charset is, utf-8 or ascii, clients need to handle there being a charset being specified. Which the conan client doesn't appear to be doing.

  * Operating System+version: 5.11.14-1-MANJARO #1 SMP PREEMPT Wed Apr 14 08:25:29 UTC 2021 x86_64 GNU/Linux
  * Compiler+version: NA
  * Conan version: 1.35.1
  * Python version: 3.9.3

ghost:
Changelog: Bugfix: Raise an exception when response header `Content-type` is different than `application/json` or `application/json; charset=utf-8`.
Docs: omit

Fixes #8884 

Conan client now raise an exception when response header `Content-type` is different than `application/json` or `application/json; charset=utf-8`. Only utf-8 charset is available, because json must be encoded in Unicode, and its default encoding is utf-8.

### DIFF
--- a/conans/client/rest/rest_client_common.py
+++ b/conans/client/rest/rest_client_common.py
@@ -196,7 +196,7 @@ class RestCommonMethods(object):
 
         content = decode_text(response.content)
         content_type = response.headers.get("Content-Type")
-        if content_type != 'application/json':
+        if content_type != 'application/json' and content_type != 'application/json; charset=utf-8':
             raise ConanException("%s\n\nResponse from remote is not json, but '%s'"
                                  % (content, content_type))
 


memsharded (approved):

conans/client/rest/rest_client_common.py
```diff
@@ -196,7 +196,7 @@ def get_json(self, url, data=None, headers=None):
 
         content = decode_text(response.content)
         content_type = response.headers.get("Content-Type")
-        if content_type != 'application/json':
+        if content_type != 'application/json' and content_type != 'application/json; charset=utf-8':
```
SSE4, line 199: looks good.
in future, we may consider accepting more encodings there, according to RFC:
https://tools.ietf.org/html/rfc4627
https://tools.ietf.org/html/rfc7158#section-8.1

SSE4 (approved):

PR #8914, opened 2021-05-04
Update urlib3 to >=1.25.8, <1.26

czoido:
Changelog: Feature: Update urlib3 Conan dependency setting version `>=1.25.8` to avoid CVE-2020-7212.
Docs: omit

#TAGS: slow


lasote (approved):
Looks good.

PR #8926, opened 2021-05-06
using always 2 profiles

memsharded:
Changelog: Feature: Use always 2 profiles (build/host)
Docs: Omit


conans/test/functional/generators/cmake_apple_frameworks_test.py
```diff
@@ -258,6 +258,7 @@ def test_apple_own_framework_cmake(self, settings):
             target_link_libraries(timer ${CONAN_LIBS})
         """)
 
+        # FIXME: test() The normal requires do not propagate environment in new build/host contexts
```
lasote, line 261: So what is the plan?

conans/test/integration/environment/compile_helpers_test.py
```diff
@@ -44,6 +44,7 @@ def test_build_with_profile(self):
 
         self.client.save({CONANFILE: conanfile_scope_env}, clean_first=True)
         self.client.run("build . --build=missing -pr scopes_env")
+        print(self.client.out)
```
lasote: print

conans/test/functional/generators/virtualenv_windows_bash_test.py
```diff
@@ -1,119 +0,0 @@
-import os
```
lasote, line 1: Is this something related to the two profiles mechanism?

conans/test/functional/graph/private_deps_test.py
```diff
@@ -309,62 +309,3 @@ def test_consumer_private(self):
                          self.client.out)
         self.assertNotIn("Hello1/0.1@lasote/stable: Generating the package",
                          self.client.out)
-
```
lasote, line 312: Is this something related to the two profiles mechanism?

lasote (approved):
Looks good but I don't really understand the removal of some tests

conans/test/integration/graph_lock/graph_lock_ci_test.py
```diff
@@ -634,9 +637,11 @@ def test_version_ranges(self):
         self.assertIn("PkgD/0.1@user/channel: DEP FILE PkgC: HelloC", client.out)
 
         # Go back to main orchestrator
-        client.run("lock create --reference=PkgD/0.1@user/channel --build -pr=myprofile "
+        client.run("lock create --reference=PkgD/0.1@user/channel --build -pr:h=myprofile -pr:b=myprofile "
                    " --lockfile-out=conan.lock")
 
+        print(client.load("conan.lock"))
```
czoido:
```suggestion
```

czoido (approved):

conans/test/functional/graph/private_deps_test.py
memsharded: Yes, this is a bug of the "private" requires, with the 2 profiles they are treated like a "build_requires", and it will not add itself to the ``self.deps_cpp_info``. And all of this will change in Conan 2.0 graph model, so probably I thought it was not worth trying to fix it, but maybe I can try to fix it for Conan 1.X.

memsharded (commented):
Tests removed are mostly because of this: https://github.com/conan-io/conan/blob/ee107d5a77d47743f30eca15cab83a37135941d6/conans/client/installer.py#L588

With 2 profiles, build_requires and private requires do not add any more to ``self.deps_cpp_info`` and viceversa, normal requires do not add to ``self.deps_env_info``.

This is challenging to fix with the current generators, and until the new generators with the new graph visitor are there, these tests would be failing. Is it better to ``pytest.mark.xfail()`` them?

lasote (approved):
Ok, so if the removed tests don't have any sense anymore, better to remove them

PR #8929, opened 2021-05-06
removing workspace

memsharded:
Changelog: Remove: Remove old ``workspace`` functionality. Will probably be restored from scratch after layouts and editables are working smoothly.
Docs: Omit


memsharded:
I am merging this @czoido, my other PR is breaking because of workspaces, I'd like to avoid unproductive conflicts.

PR #8932, opened 2021-05-06
removing cmake_find_package generators

memsharded:
Changelog: Remove ``cmake_find_package`` and ``cmake_find_package_multi`` generators, superseded by ``CMakeDeps``
Docs: Omit


SpaceIm:
I don't know if it's the good place to discuss about that, but if `cmake_find_package` is removed in conan v2 and `CMakeDeps` doesn't model Find module files, but config files only, how consumers and recipes will be able to generate files that can take precedence over built-in Find module files like FindZLIB.cmake, FindOpenSSL.cmake?

memsharded:
Hi @SpaceIm 

> I don't know if it's the good place to discuss about that, but if cmake_find_package is removed in conan v2 and CMakeDeps doesn't model Find module files, but config files only, how consumers and recipes will be able to generate files that can take precedence over built-in Find module files like FindZLIB.cmake, FindOpenSSL.cmake?

Up to my knowledge, this is not an issue, this is how ``cmake_find_package_multi`` works, and it has been working fine for a long time. ``find_package(OpenSSL)`` can resolve to the generated ``openssl-config.cmake`` too, it doesn't need to be a ``FindOpenSSL.cmake`` one. The ``CMakeToolchain`` now defines the priority of config files over find files too with ``CMAKE_FIND_PACKAGE_PREFER_CONFIG``, to help with that. Am I missing something?

SpaceIm:
> Hi @SpaceIm
> 
> > I don't know if it's the good place to discuss about that, but if cmake_find_package is removed in conan v2 and CMakeDeps doesn't model Find module files, but config files only, how consumers and recipes will be able to generate files that can take precedence over built-in Find module files like FindZLIB.cmake, FindOpenSSL.cmake?
> 
> Up to my knowledge, this is not an issue, this is how `cmake_find_package_multi` works, and it has been working fine for a long time. `find_package(OpenSSL)` can resolve to the generated `openssl-config.cmake` too, it doesn't need to be a `FindOpenSSL.cmake` one. The `CMakeToolchain` now defines the priority of config files over find files too with `CMAKE_FIND_PACKAGE_PREFER_CONFIG`, to help with that. Am I missing something?

It was not working if `CMAKE_FIND_PACKAGE_PREFER_CONFIG` was not set, and if a CMakeLists was relying on find_package without CONFIG keyword for libraries with built-in Find modules files. There are several recipes in CCI which need `cmake_find_package` AND `cmake_find_package_muti`, because `find_package(OpenSSL REQUIRED)` (or ZLIB, or CURL etc) is the way to find OpenSSL, `find_package(OpenSSL REQUIRED CONFIG)` wouldn't make sense in an agnostic CMakeLists.

`CMAKE_FIND_PACKAGE_PREFER_CONFIG` doesn't work for CMake < 3.15: https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html, it's quite a high requirement.

memsharded:
But wont ``find_package(OpenSSL REQUIRED)`` find the openssl-config.cmake with ``CMAKE_FIND_PACKAGE_PREFER_CONFIG`` ? I think it does, that is the intention of that variable, right?


>CMAKE_FIND_PACKAGE_PREFER_CONFIG doesn't work for CMake < 3.15:

Not an issue, CMakeDeps is for Conan 2.0, and the tribe already approved CMake>= 3.15. It already contains ``if(${{CMAKE_VERSION}} VERSION_LESS "3.15")`` in its code.

SpaceIm:
> But wont find_package(OpenSSL REQUIRED) find the openssl-config.cmake with CMAKE_FIND_PACKAGE_PREFER_CONFIG ? I think it does, that is the intention of that variable, right?

Sure, but we don't do that in CCI because the current policy is to stick to upstream min CMake version, which is almost never 3.15 or higher (more close to 3.1, sometimes 2.8).

> Not an issue, CMakeDeps is for Conan 2.0, and the tribe already approved CMake>= 3.15. It already contains `if(${{CMAKE_VERSION}} VERSION_LESS "3.15")` in its code.

I see, then it's ok. Now my only concern is that injecting `CMAKE_FIND_PACKAGE_PREFER_CONFIG` changes the default state of CMake, so it's against the principle of least surprise. Though, I don't know what could be the side effects in agnostic CMakeLists.

memsharded:
> I see, then it's ok. Now my only concern is that injecting CMAKE_FIND_PACKAGE_PREFER_CONFIG changes the default state of CMake, so it's against the principle of least surprise. Though, I don't know what could be the side effects in agnostic CMakeLists.

Yes, all of this is intended for Conan 2.0, and the whole point of having it in 1.X is to mature it, stabilize, get feedback... There are some guidelines we are trying to follow:

- We want to avoid to have different generators for the same build system. It is super confusing to users, and very error prone. 
- We want to provide one integration, as solid, standard, tested and flexible as possible. CMakeDeps and ``generate()`` provide a whole new level of configuration. If something is necessary it will be a parameterization of CMakeDeps.
- We are dropping support for older CMake < 3.15. The pressure for moving forward is high, and the Tribe already voted for that.
- The ``CMAKE_FIND_PACKAGE_PREFER_CONFIG`` is injected by the ``CMakeToolchain``. It is also highly configurable, and could be skipped. 
- Conan will be trying to provide a good default behavior, in this case, as ``CMakeDeps`` create config files, it will default to find those config files. The opposite will be the "least expected" behavior in terms of someone using Conan right? If you are using the ``conan.tools.cmake`` CMakeToolchain and CMakeDeps is generating config files for you, why shouldn't the toolchain prioritize those config files instead of possible find.cmake in the system? 

Please let me know if this make sense. Lets keep working on CMakeDeps (and others), please keep the feedback coming. Thanks!

czoido (approved):

ruilvo:
>     * The `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is injected by the `CMakeToolchain`. It is also highly configurable, and could be skipped.

What does this mean? How will this variable be injected with the `CMakeDeps` generator?

PR #8945, opened 2021-05-11
Feature/toolchain cmake userfiles

### LINKED ISSUE #8860: [question] What is the recommended way to get Conan to use an existing CMake toolchain file?

I recently wrote a CMake toolchain file for the Texas Instrument C2000 Code Generation Tools toolchain. After verifying that it works as intended, I sought to find a way to get this toolchain file to work with Conan.

The first thing I found in my web searches  that conan has tools for generating a CMake toolchain file on the fly and including it in the call to CMake. However, after going over the source code of the CMakeToolchain class it seems that I am unable to specify the path of the compiler executable. Unless there is some way for me to set the path of the compiler executable in the `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` variables, I cannot use the CMakeToolchain toolchain generator.

Further research led me to the `CONAN_CMAKE_TOOLCHAIN_FILE` variable and CMake build helper. Defining the `CONAN_CMAKE_TOOLCHAIN_FILE` supposedly inserts a `-DCMAKE_TOOLCHAIN_FILE` option into the CMake command line. However, this can be rather inflexible if one is working with multiple toolchains. In the CMake build helper class, there is the `configure` method which has the defs parameter where I can pass a dictionary with a key `CMAKE_TOOLCHAIN_FILE` and the corresponding value being that of the path to the toolchain file. This looks to be the most promising as I can pick out which toolchain to use based on the Conan settings.

If there are any other ways to use a handwritten CMake toolchain file in Conan, I would like to have them described in a reply to this post.

memsharded:
Changelog: Feature: Define ``[conf]`` for defining the user toolchain for ``CMakeToolchain``, both for injecting a user toolchain in the ``CMakeToolchain`` generated ``conan_toolchain.cmake`` and for completely replacing ``conan_toolchain.cmake``.

Close https://github.com/conan-io/conan/issues/8860


jsanchez19:
hi @memsharded - could you show what this looks like in an example? Is `[conf]` a section in a profile or elsewhere?

memsharded:
Yes, it would follow the syntax in https://docs.conan.io/en/latest/reference/config_files/global_conf.html#tools-configurations

So something in your profile like:
```
[settings]
....
[conf]
tools.cmake.cmaketoolchain:toolchain_file=mytoolchain.cmake
```

conan/tools/cmake/toolchain.py
```diff
@@ -610,12 +627,24 @@ def content(self):
...
 
     def generate(self):
-        save(self.filename, self.content)
+        toolchain_file = self._conanfile.conf["tools.cmake.cmaketoolchain:toolchain_file"]
+        if toolchain_file is None:  # The main toolchain file generated only if user dont define
+            save(self.filename, self.content)
         # Generators like Ninja or NMake requires an active vcvars
         if self.generator is not None and "Visual" not in self.generator:
             write_conanvcvars(self._conanfile)
+        self._writebuild(toolchain_file)
```
lasote, line 636: `_write_toolchain_args_file()` would be a better name?

lasote (approved):

PR #8947, opened 2021-05-12
Expand user-agent string to include OS info.

grafikrobot:
It's useful to have the operating system information in the user-agent of the client to provide better analytics. Such analytics can help in determining if some particular set of clients are having problems. This adds the operating system, version, and machine type. In a format similar to what regular browsers provide.

Changelog: Feature: Expand user-agent string to include OS info.
Docs: omit

### DIFF
--- a/conans/client/rest/conan_requester.py
+++ b/conans/client/rest/conan_requester.py
@@ -105,8 +105,12 @@ class ConanRequester(object):
 
         # Only set User-Agent if none was provided
         if not kwargs["headers"].get("User-Agent"):
-            user_agent = "Conan/%s (Python %s) %s" % (client_version, platform.python_version(),
-                                                      requests.utils.default_user_agent())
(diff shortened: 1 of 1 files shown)


danimtb (approved):
Agree this is pretty common information and can be useful to determine the usage in conan center

jgsogo (approved):
I've been having a look at sources on some backend services and I cannot find any usage of it. Maybe it is hidden in some script or even in some private script someone executed when we were asking for statistics about Python versions. We could fix that script if needed.

BTW, do we want to show `platform.processor()` for completeness?

jgsogo:
I think it is good as it is now in the PR. IMHO we could change it in the future without considering it breaking (like we are changing it now).

memsharded:
This is breaking something in https://github.com/conan-io/conan/issues/9036, investigating

PR #8951, opened 2021-05-12
Add test cases for build require invalid configuration

danimtb:
Changelog: omit
Docs: omit

- [x] Refer to the issue that supports this Pull Request: Related to https://github.com/conan-io/conan-center-index/pull/5511

cc/ @jgsogo @SSE4


memsharded (changes requested):
The tests are not running because you need to name the file ``test_xxx.py``

Then, if you could elaborate/explain a bit better what is the issue, what tests are expected to do or fail, that would help.

jgsogo:
There are two different scenarios that start in these tests:
 * if the `br` doesn't remove the `build_type` setting: it should fail (missing package and then invalid config) unless it declares Release as a valid fallback for Debug (compatible_packages).
 * if the `br` does remove `build_type` setting: it shouldn't fail because the resulting packageID is already available in the cache.

This can be enough for a first iteration, it gets harder when the build-requires has other dependencies.

danimtb:
@jgsogo I will add the cases with the `package_id()`

conans/test/functional/build_requires/test_build_require_invalid.py
```diff
@@ -0,0 +1,317 @@
...
+    client = setup_client_with_build_requires
+    client.run("create br_build.py")
+    client.run("create consumer.py -s build_type=Debug", assert_error=True)
+
+
+def test_create_validate_remove_build_type(setup_client_with_build_requires):
+    client = setup_client_with_build_requires
+    client.run("create br_validate_remove_build_type.py")
+    # FIXME: This should NOT fail with -> br/1.0.0: Invalid ID: br cannot be built in debug mode
+    client.run("create consumer.py -s build_type=Debug", assert_error=True)
```
jgsogo, line 214: This is the use-case we are exercising currently in ConanCenter and **we need to fix in the Conan client**. Because of this error we are not generating some configurations of libraries.

conans/test/functional/build_requires/test_build_require_invalid.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def test_create_configure_remove_build_type(setup_client_with_build_requires):
+    client = setup_client_with_build_requires
+    client.run("create br_configure_remove_build_type.py")
+    # FIXME: This should NOT fail with -> ERROR: br/1.0.0: Invalid configuration: br cannot be built in debug mode
+    client.run("create consumer.py -s build_type=Debug", assert_error=True)
```
jgsogo, line 221: When using one profile, this scenario is really important too. Many checks are being raised from `configure` method, like the ones related to `cppstd` ([CMake](https://github.com/conan-io/conan-center-index/blob/master/recipes/cmake/3.x.x/conanfile.py)), which are then removed together with the `settings.compiler`.

With this error, it is not possible to use `CMake` as a build-requires and use explicit `-s compiler.cppstd=98`

conans/test/functional/build_requires/test_build_require_invalid.py
```diff
@@ -0,0 +1,317 @@
...
+    client.run("create br_build_remove_build_type.py")
+    client.run("create consumer.py -s build_type=Debug")
+    assert "br/1.0.0 from local cache - Cache" in client.out
+
+
+def test_create_validate_compatible(setup_client_with_build_requires):
+    client = setup_client_with_build_requires
+    client.run("create br_validate_compatible.py")
+    client.run("create consumer.py -s build_type=Debug")
+    assert "Using compatible package" in client.out
```
jgsogo, line 235: 👌

conans/test/functional/build_requires/test_build_require_invalid.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+def test_create_configure_compatible(setup_client_with_build_requires):
+    client = setup_client_with_build_requires
+    client.run("create br_configure_compatible.py")
+    # FIXME: This should NOT fail with -> ERROR: br/1.0.0: Invalid configuration: br cannot be built in debug mode
+    client.run("create consumer.py -s build_type=Debug", assert_error=True)
```
jgsogo, line 242: When using one profile, this scenario is really important too. Many checks are being raised from `configure` method, like the ones related to `cppstd` ([CMake](https://github.com/conan-io/conan-center-index/blob/master/recipes/cmake/3.x.x/conanfile.py)), which are then removed together with the `settings.compiler`.

With this error, it is not possible to use `CMake` as a build-requires and use explicit `-s compiler.cppstd=98`

conans/test/functional/build_requires/test_build_require_invalid.py
memsharded: Yes, agree, this is the main case I am trying to address in the client.

conans/test/functional/build_requires/test_build_require_invalid.py
memsharded: If I understood it correctly, the idea here is to move invalid checks from ``configure()`` to ``validate()``. I am not considering adding extra logic to ``configure()`` step exceptions.

conans/test/functional/build_requires/test_build_require_invalid.py
memsharded: Same as above.

conans/test/functional/build_requires/test_build_require_invalid.py
jgsogo: Yes, probably **all** the checks can be moved to `validate()` and focus the issue/fix on that function. Maybe this is something we can add explicitly somewhere in ConanCenter docs. Right now, I cannot think about any reason why someone could need to raise from `configure()` instead of `validate()`.

memsharded:
Superseded by https://github.com/conan-io/conan/pull/9268 (for Conan 2.0)

PR #8952, opened 2021-05-12
Add support for PyJWT 2.0

### LINKED ISSUE #8876: [feature] Lift version limitation for the PyJWT
Currently Conan has version limitation for PyJWT like this:
PyJWT>=1.4.0, <2.0.0

Popular frameworks, like PyGithub, are gradually lifting minimal PyJWT version to the 2.0+, and it becomes problematic to use conan together with the rest of the scripting. Situation is temporary resolvable using virtuenv, but python version switching is yet another dimension for human mistakes, especially for non-experts in python. 

It would be way easier if conan can be installed together with mainstream packages in a typical environment. Can PyJWT version limit be lifted to the 2.0+?

uilianries:
PyJWT 2.0 made some breaking changes:

1) profile.exp no longer is converted to integer, we need to do it manually

The difference from [1.4.0](https://github.com/jpadilla/pyjwt/blob/1.4.0/jwt/api_jwt.py#L46) to [2.1.0](
https://github.com/jpadilla/pyjwt/blob/2.1.0/jwt/api_jwt.py#L53) is because now they do a copy of profile. I thought it was a bug, but actually their [tests](https://github.com/jpadilla/pyjwt/blob/2.1.0/tests/test_api_jwt.py#L218) proves that is correct.

2) The `ExpiredSignature` exception is now [deprecated](https://pyjwt.readthedocs.io/en/stable/changelog.html#dropped-deprecated-errors)

No problem, we still can use [ExpiredSignatureError](https://github.com/jpadilla/pyjwt/blob/1.4.0/jwt/exceptions.py#L46), which is compatible since 1.4.0 and it's recommended way for 2.0: 

3) PyJWT+2 [dropped](https://pyjwt.readthedocs.io/en/stable/changelog.html#drop-support-for-python-2-and-python-3-0-3-5) Python 2 support

We still can adopt [PEP-508](https://www.python.org/dev/peps/pep-0508/), which allows conditions in requirements.txt.

4) The method `jwt.encode` returns [encoded bytes](https://github.com/jpadilla/pyjwt/blob/1.4.0/jwt/api_jwt.py#L53) for 1.4.0. However, 2.1.0 returns [string text](https://github.com/jpadilla/pyjwt/blob/2.1.0/jwt/api_jwt.py#L63)

No problem, we can use six to detect string or bytes and encode when necessary.

Changelog: Feature: Add support for PyJWT 2.0
Docs: Omit
closes #8876

/cc @ezaiakinsc @memsharded 


uilianries:
Another breaking change: https://github.com/jpadilla/pyjwt/issues/602

Need to find where secret is breaking.

memsharded (changes requested):
Hi @uilianries 

It seems this is massively breaking. It might be high risk to do this upgrade, could you please have another look to see what it could take to upgrade?

conans/server/crypto/jwt/jwt_manager.py
```diff
@@ -22,7 +24,12 @@ def get_token_for(self, profile_fields=None):
         if self.expire_time:
             profile_fields["exp"] = datetime.utcnow() + self.expire_time
 
-        return jwt.encode(profile_fields, self.secret, algorithm="HS256")
+        encoded = jwt.encode(profile_fields, self.secret, algorithm="HS256")
+        if self.expire_time and isinstance(profile_fields.get("exp"), datetime):
```
memsharded: Passing this after the ``jwt.encode()`` reads confusing or even dangerous? What if ``jwt.encode`` does an internal copy? This field would be ignored, right? Is this extracted from jwt documentation?

memsharded: Is this in addition to the above ``profile_fields["exp"] = datetime.utcnow() + self.expire_time``? 

I might be worth commenting a bit these changes, what the code does is not obvious.

uilianries: >  Is this extracted from jwt documentation?

No, I copied directly from their source code, to keep the compatibility: https://github.com/jpadilla/pyjwt/compare/1.4.0...2.1.0#diff-6893ad4a1c5a36b8af3028db8c8bc3b62418149843fc382faf901eaab008e380R53

>  What if jwt.encode does an internal copy?

Indeed there is a copy for payload. Let me invert it.

uilianries: Done, I've added some comments. Indeed won't remember about these changes few weeks later, so comments a really helpful here.

I also inverted the encoder order, as profile_fields is copied internally for 2.x, we don't want to risk any possible abnormal behavior for future minor versions.

conans/requirements.txt
```diff
@@ -1,4 +1,5 @@
-PyJWT>=1.4.0, <2.0.0
+PyJWT>=1.4.0, <2.0.0; python_version < '3.0'
+PyJWT>=2.0.0, <3.0.0; python_version >= '3.0'
```
memsharded: Maybe better maintain PyJWT>=1.4.0

memsharded (commented):
I am concerned about this.

This has been reported inconvenient, and needing to use virtualenvs by 1 user, but it is likely that out of the thousands of users using Conan, there will be some that are also using PyJWT and haven't still ugpraded to PyJWT 2.0 yet. If we introduce this constraint, now we will be breaking them.

This dependency is used only for the ``conan_server``, not the client, so if we decouple it to its own PyPI package this problem will be over (this is plan, at least for Conan 2.0)

At the moment, it might be enough to workaround to force the version, if not using the ``conan_server`` in the same virtualenv (this is strongly discouraged, the server should run in its own env), it will not be an issue and nothing will break.

So maybe better to wait a bit before changing this?

conans/requirements.txt
uilianries: have to check if prefers the first, but yes.

uilianries:
> So maybe better to wait a bit before changing this?

I agree. It should work, but some features of PyJWT 2.x are still under discussion, it adds some risk of breaking in future, due a new minor version. Users still can force the installed version if they need.

memsharded (changes requested):
Please target this to ``develop2`` branch.
Py2 support is removed too
Put it only in the ``requirements_server.txt`` file

uilianries:
> Please target this to `develop2` branch. Py2 support is removed too Put it only in the `requirements_server.txt` file

Done. I moved pyjwt requirement.

uilianries:
Need to convert datetime object to int, otherwise, the token would contain datetime in its dictionary, but when decoding, it will became an integer. It's a regular behavior from PyJWT 2.0, not present on 1.x.

memsharded (approved):

PR #8956, opened 2021-05-14
Improve error message when a folder is not an SCM repo

DavidSpickett:
Changelog: Fix: Improve error message when a directory doesn't contain a valid repository.
Docs: omit

Before you would get this:
ConanException: Not a valid 'svn' repository

Which doesn't help much, especially if you're using
"conan create" which may be off in a folder you didn't
create.

Add the path of the folder to the error message to instead
get:
ConanException: '/a/b/not_an_svn_repo' is not a valid 'svn' repository

### DIFF
--- a/conans/client/tools/scm.py
+++ b/conans/client/tools/scm.py
@@ -16,7 +16,7 @@ from conans.util.runners import check_output_runner, version_runner, muted_runne
 
 
 def _check_repo(cmd, folder):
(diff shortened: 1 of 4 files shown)


lasote (approved):

memsharded:
Thanks for contributing this! Merging this to be included in next 1.37

PR #8958, opened 2021-05-14
First implementation of the Bazel toolchain

tapia:


conan/tools/google/bazel.py
```diff
@@ -0,0 +1,49 @@
...
+
+from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
+from conans.util.files import load
+
+class Bazel(object):
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._get_bazel_project_configuration()
+
+    def configure(self, args=None):
```
memsharded, line 12: If no config step in bazel, you can safely remove this method.

conan/tools/google/bazeldeps.py
```diff
@@ -0,0 +1,95 @@
...
+            result += '    static_library = "{0}/lib{1}.a"\n'.format(
+                dependency.new_cpp_info.libdirs[0],
+                lib
+            )
+            result += ')\n\n'
+
+        result += 'cc_library(\n'
+        result += '    name = "{}",\n'.format(dependency.ref.name)
+
+        if dependency.new_cpp_info.includedirs:
```
memsharded: This might read a bit easier with a jinja template

memsharded (commented):
Looking good!

Please start adding some "integration" tests that check things without building anything or calling bazel at all. We will discuss what to do with CI.

conan/tools/google/bazeldeps.py
tapia: Done

conan/tools/google/bazel.py
tapia: As far as I've seen, the convention when writing the `build()` method for a project's `conanfile.py` is:

```
def build(self):
    bazel = Bazel(self)
    bazel.configure()
    bazel.build(whatever)
```

Wouldn't deleting the configure method break this pattern?

PR #8972, opened 2021-05-18
Feature/test build require invalid sol

memsharded:
Changelog: Feature: Adds a new ``ConanUnbuildableConfiguration`` exception that can be raised in ``validate()`` to indicate that a binary cannot be built, but a compatible package_id might exist.

The proposal is:

```python

        from conans.errors import ConanUnbuildableConfiguration

        class Conan(ConanFile):
     
            def validate(self):
                if self.settings.build_type == "Debug":
                    raise ConanUnbuildableConfiguration("br cannot be built in debug mode")

            def package_id(self):
                del self.info.settings.build_type
```

Test case from: https://github.com/conan-io/conan/pull/8951


danimtb:
Having another exception seems like the short and easy approach to fix the issue but on the other hand, it gives the responsibility to the users to specify this correctly in the `validate()` being (at first) a vague difference between an invalid configuration and an unbuildable one.

In other words, I feel that, with this approach, we are duplicating the information of the recipe in `validate()` and `package_id()`. With two different exceptions that can be raised from `validate()`, users now have to think not only about configurations that do not build but also define if any other configurations may or may not result in a compatible package.
IMO, with this `UnbuildableConfiguration`, we are representing exactly the same thing `InvalidConfiguration` was conceived for, and from the point of view of the recipe creator, I'd (almost? 🤔) always use `UnbuildableConfiguration` from now on because this is what invalid means to me.

I think it should be better to use only one exception type (InvalidCofiguration) and leave the "cognitive effort" of defining the binary compatibility to the `package_id()`. Probably this would require that the `package_id()` is evaluated before the `validate()` and maybe changing this behavior is breaking or it has other implications (like breaking other use cases that current `validate()` behavior might be covering).

memsharded:
>  Probably this would require that the package_id() is evaluated before the validate()

package_id() is already evaluated before the validate()


> I think it should be better to use only one exception type (InvalidCofiguration) and leave the "cognitive effort" of defining the binary compatibility to the package_id()

If defining it in the ``package_id()`` is preferred, something like this might be totally possible, same meaning, different UX:

```python
        class Conan(ConanFile):
     
            def package_id(self):
                del self.info.settings.build_type
                if self.settings.build_type == "Debug":
                        self.info.invalid = BINARY_UNBUILDABLE
```

I implemented it in the validate, because the ``package_id()`` is oriented to compute the binary package_id, irrespective wheter such package_id is realizable or not, while the ``validate()`` seemed the place to implement: "this configuration can not be used at all (ConanInvalidConfiguration)" or "this configuration cannot be built (ConanUnbuildableConfiguration), but this configuration might be valid input".

jgsogo:
I need to say that I agree with @danimtb 100%. The purpose of the ConanInvalidConfiguration was very likely this one, I don't see why another exception (or explicit value in `package_id`) is needed. Information is already there, Conan only needs to work with it properly.

IMHO, I prefer to wait and start using two profiles in ConanCenter over adding extra complexity to recipes.

danimtb:
> package_id() is already evaluated before the validate()

In that case, I'd rather use the `validate()`. I don't like the approach of package ID.

I guess the other use case that `InvalidConfiguration` covers and `UnbuildableConfiguration` does not is this one?

```
class Pkg(ConanFile):
    requires = "dep/0.1"

    def validate(self):
        if self.options["dep"].myoption == 2:
            raise ConanInvalidConfiguration("Option 2 of 'dep' not supported")
```

🤔 I am still not convinced of having two different exceptions...

memsharded:
> IMHO, I prefer to wait and start using two profiles in ConanCenter over adding extra complexity to recipes.

The 2 profiles doesn't solve this issue. Even when the build profile is used, if something is deleted in ``package_id``, it can have exactly the same issue, as the tests already prove.

> The purpose of the ConanInvalidConfiguration was very likely this one, I don't see why another exception (or explicit value in package_id) is needed. Information is already there, Conan only needs to work with it properly.

I have tried to approach the problem with 1 single configuration, in https://github.com/conan-io/conan/pull/8976. It is possible to go far, but there is still one case that is impossible to cover, this one:

```python
           from conans.errors import ConanInvalidConfiguration
           class Pkg(ConanFile):
               requires = "dep/0.1"

               def validate(self):
                   if self.options["dep"].myoption == 2:
                       raise ConanInvalidConfiguration("Option 2 of 'dep' not supported")
```

The problem is when a ``validate()`` raises a ``ConanInvalidConfiguration`` for something that doesn't directly produce a variation in the ``package_id``. This is a likely scenario, users might raise invalid configurations for example for some specific versions of its dependencies, that will still produce the same package id, because they fall in the same semver bucket.

If we want to support this case, and it seems we need to, then, it is fundamentally not possible to use 1 single indicator of the situation, like ``ConanInvalidConfiguration``, because there are 2 different states to represent:

- We want to raise an error exception, even if I have a matching package_id for the current ones, because there is something in the graph, that doesn't affect my package_id, that I know will make my current, built binary, invalid
- We want to raise an error for the case that the current package_id cannot be built with the current settings (and there was no compatible packages to fallback)


So we can discuss the UI, but it is necessary to specify in recipes these 2 different user intentions.

jgsogo:
I understand that situation, but maybe it is a problem more related to the packageID than to the build process. With the default mode (semver-major), you've chosen to ignore everything but the major version of your dependencies, assuming that it will be ABI compatible. Here, you are adding an extra check to take into account the value of an option of your dependencies, it can be equivalent to check a minor version:

```python

class Recipe(ConanFile):
   def requirements(self):
      self.requires('dep/1.0.0')

   def validate(self):
      if self.deps_cpp_info['dep'].version == '1.0.2':
         raise ConanInvalidConfiguration("Patch version not supported")
```

If you are afraid of ABI issues because of patch releases or options, then you can choose to use a package-id-mode that will take those changes into account (adding more components of your requirements: full version, package-id, revision,...).

I would say that these `ConanInvalidConfiguration` exceptions belong to the `build()` stage: while building, they prevent users from starting a build that won't succeed (save build-time). OTH, while consuming binaries, the package-id is king, if package-id is available, then use it... 

---

Summing up, in my opinion, if it is not enforced in the recipe, it is up to the consumer. If we want to invalidate that combination for consumers, then we should add that information to the package-id, something like:

```python
class Recipe(ConanFile):
   def requirements(self):
      self.requires('dep/1.0.0')

   def validate(self):
      if self.deps_cpp_info['dep'].version == '1.0.2':
         raise ConanInvalidConfiguration("Patch version 'dep/1.0.2' is breaking and won't compile")

   def package_id(self):
      self.info.requires["dep"].full_package_mode()  # FIXME: semver-major + package-id (at least)
```

I prefer these modifications to `package_id()` (and some improvements) over adding one new exception. And, I would say, that they are easier to understand and more aligned with the `compatible_package` feature.

memsharded:
I don't understand what you mean, this is not related to the package_id, but to valid configurations.

Lets say that we have 1 package, without dependencies, the AndroidNDK package. It is some java code, which is totally portable and runs in all platforms, so we want it to be exactly 1 package_id, the same to be used in all cases.

Now there are 2 different situations:
- The package itself has 2 options that can be set, and they define some behavior of the package, lets say, the ``api_level`` and the ``android_sdk``. There are some configurations that will fail, for example, you cannot use an ``api_level=13``, with an ``android_sdk=True``
- On the other hand, this package cannot be created successfully on OSX, for some problem while unzipping some things from the Android source site

We still need recipes to signal 2 different conditions:

- I cannot build this package in this platform, which is OSX, when someone tries to build it in that platform.
- This package, which exists and is there, cannot be used with the user selected options ``api_level=13`` and ``android_sdk=True``

jgsogo:
But those options and/or settings are related to `android_ndk` recipe itself... The problem here is about options/configuration/versions **from requirements** that are incompatible with the consumer. Not about the consumer itself.

In your example, it won't exist an Android package with `api_level=13` and `android_sdk=True` because those options belong to the `android_ndk` recipe and they are _always_ included in its package-id. It will fail to build and that package-id will never exist.

memsharded:
> In your example, it won't exist an Android package with api_level=13 and android_sdk=True because those options belong to the android_ndk recipe and they are always included in its package-id. It will fail to build and that package-id will never exist.

Yes, in my example, there is 1 unique package_id for the AndroidNDK, because there is only 1 AndroidNDK which is, lets say a  portable android.jar. This android package will have 1 single package_id that can be used by all projects, similar to what an AndroidNDK installation in the system does. There will not be multiple different package_ids for different configurations of options ``api_level`` and ``android_sdk``, because these options do not affect the binary, they only affect the information this binary propagates to consumer. This is a common usage by Conan users, which remove ``self.info.options`` in their ``package_id()``.

There are no requirements involved at all, just 1 package, that is being used by:

conanfile.txt
```
[build_requires]
androidndk/1.0

[options]
androidndk:api_level=13
androidndk:android_sdk=True
```

Now we still need those 2 different outputs:
- If you are using ``api_level=13 and android_sdk=True``, your configuration is wrong, I am a valid package, sitting here in your cache, and I am being used by other projects successfully, but I won't work with those options
- If you are using perfectly compatible options, but there is no AndroidNDK binary, and you try to build it with ``--build`` in OSX, I won't be able to build in this OSX platform

jgsogo:
I think they would validate the combination in the `package_info()` method, which is the place where they have to do something based on that option (this is not how the Android recipe works, but I'm trying to follow your hypothesis):

```python
def package_info(self):
   if self.options.api_level=54:
      raise ConanException('API level 54 is not available in this package')
   else:
      self.env_info.ANDROID_NATIVE_API_LEVEL = str(self.options.api_level)
```

Maybe there is some real example we can use and it would be easier to reason about it.

memsharded:
The ``package_info()`` will be evaluated later in the flow, so it will be useless to raise there exceptions, it will be too late to get the useful information we want to get early in the process to see if some dependency graph can be built/installed or not.

jgsogo:
Yes, this is why my suggestion is to look for a real example: a package that contains options that are removed from its package-id but, if used (the right way) by consumers, can lead to this situation. Maybe it's just a theoretical scenario... 

But if you know the scenario is valid and the error has to be raised at that time and from `validate()`... then, of course, it can only be achieved by adding an additional argument to `ConanInvalidConfiguration("...", XXXX)` or adding another exception. I'm just trying to build with the pieces we already have.

memsharded:
Superseded by #9268

PR #8985, opened 2021-05-21
CMakeDeps build_requires full support

### LINKED ISSUE #7719: [question] 'build_modules' of `build_requires` are not included when building consumers with build/host profiles
I'm experimenting with the build and host context profiles and many of my builds are breaking due to the changes in how `build_requires` are handled.  So I'm just trying to understand if this is intentional and whether I need to update my recipes.

For example, I have a tool which is used to generate source code from .xml files in the consumer's repository.  The sources are then compiled into a static library and packaged.  Since consumers only need this tool when building from sources, they have it defined as a `build_requires`.  (Though since the code generator version can have an affect on the code, I'm wondering if it should actually be a normal `requires`? )

To simplify it's usage for consumers, the tool package provides functions to consumers through a `build_module`.  But now it seems that `build_module`s of `build_requires` dependencies are not included by _conan_basic_setup()_.

My current workaround is to `force_host_context=true` for these dependencies, but I'm not sure if this is the correct way to go about it.

### LINKED ISSUE #8303: [bug] cmake generator does not write build requirements info when using dual profiles
  * Operating System+version: macOS Big Sur
  * Conan version: develop (1.33.0-dev)
  * Python version: 3.8.5

With this `conanfile.py`:

```python
from conans import ConanFile, CMake, tools


class BugConan(ConanFile):
    name = "bug"
    version = "0.1"
    license = "<Put the package license here>"
    author = "<Put your name here> <And your email here>"
    url = "<Package recipe repository url here, for issues about the package>"
    description = "<Description of Bug here>"
    topics = ("<Put some tag here>", "<here>", "<and here>")
    options = {"shared": [True, False], "fPIC": [True, False]}
    default_options = {"shared": False, "fPIC": True}
    generators = "cmake"

    def config_options(self):
            del self.options.fPIC

    def build_requirements(self):
        self.build_requires("doctest/2.3.8")

    def source(self):
        pass

    def build(self):
        pass

    def package(self):
        pass

    def package_info(self):
        pass
```

Run `conan install . --profile:host=default --profile:build=default`

The generated `conanbuildinfo.cmake` does not contain `doctest` targets, but it does when `conan install . --profile=default` is used.

lasote:
Changelog: Feature: Introduced new options for the `CMakeDeps` generator allowing to manage `build_requires` even declaring the same package as a `require` and `build_require` avoiding the collision of the `config` cmake files and enabling to specify which `build_modules` should be included (e.g protobuf issue)
Docs: https://github.com/conan-io/docs/pull/2104

Closes #8303
Closes #7719

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/__init__.py
+++ b/conan/tools/cmake/cmakedeps/templates/__init__.py
@@ -15,11 +15,6 @@ class CMakeDepsFileTemplate(object):
         return self.conanfile.ref.name + self.suffix
 
     @property
-    def package_folder(self):
-        return self.conanfile.package_folder.\
-            replace('\\', '/').replace('$', '\\$').replace('"', '\\"')
-
-    @property
     def target_namespace(self):
         return get_target_namespace(self.conanfile) + self.suffix
 
--- a/conan/tools/cmake/cmakedeps/templates/__init__.py
+++ b/conan/tools/cmake/cmakedeps/templates/__init__.py
@@ -8,13 +8,37 @@ class CMakeDepsFileTemplate(object):
 
     def __init__(self, cmakedeps, req):
         self.cmakedeps = cmakedeps
-        if req is not None:
-            self.conanfile = req
-            self.pkg_name = req.ref.name
-            self.package_folder = req.package_folder.\
-                replace('\\', '/').replace('$', '\\$').replace('"', '\\"')
-            self.target_namespace = get_target_namespace(self.conanfile)
-            self.file_name = get_file_name(self.conanfile)
+        self.conanfile = req
+
+    @property
+    def pkg_name(self):
+        return self.conanfile.ref.name + self.suffix
+
+    @property
+    def package_folder(self):
+        return self.conanfile.package_folder.\
+            replace('\\', '/').replace('$', '\\$').replace('"', '\\"')
+
+    @property
+    def target_namespace(self):
+        return get_target_namespace(self.conanfile) + self.suffix
+
+    @property
+    def file_name(self):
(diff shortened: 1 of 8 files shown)


conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -8,13 +8,37 @@ class CMakeDepsFileTemplate(object):
...
-            self.file_name = get_file_name(self.conanfile)
+        self.conanfile = req
+
+    @property
+    def pkg_name(self):
+        return self.conanfile.ref.name + self.suffix
+
+    @property
+    def package_folder(self):
+        return self.conanfile.package_folder.\
```
memsharded: This name is a bit confusing, it will return a escaped package_folder

conans/test/functional/toolchains/cmake/cmakedeps/test_build_context_protobuf.py
```diff
@@ -0,0 +1,233 @@
...
+        find_package(protobuff)
+        find_package(protobuff_KK)
+        add_executable(app main.cpp)
+        foo_generate()
+        target_link_libraries(app protobuff::protobuff)
+        """)
+
+    cmake_deps_conf = """
+        deps.build_context_build_modules = []
+        deps.build_context_suffix = {"protobuff": "_KK"}
```
memsharded: more elegant suffix :)

memsharded (approved):
Looking good progress!

SeanSnyders:
Hi, I filed a bug relating to these updates for the `CMakeDeps` generator:
https://github.com/conan-io/conan/issues/9047

PR #8991, opened 2021-05-24
First implementation of the Bazel toolchain

### LINKED ISSUE #6235: [feature] Generator for Bazel
Dependencies with Bazel are a pain sometimes. This has the potential to solve it. [google/cargo-raze](/google/cargo-raze) is a tool that interprets Rust's `Cargo.toml` and generates build files from which simplifies the import story for Rust packages. There is the possibility of creating something similar for your `conanfile.txt` in the form of a Bazel generator. This would generate build files for each of the libraries and make them available as dependencies to other actions.

The main question that needs to be answered is how to reason with the various combinations of local cache vs remote and build-from-source vs prebuilt. Some ideas:

- Cargo raze is the simplest approach in that it always downloads the source code and builds locally
- We have the potential to, where they exist, download precompiled binaries and depend on them with `cc_import` (fast)

tapia:
Changelog: Feature: Implements a new experimental ``conan.tools.google`` Bazel integration with ``BazelDeps``, ``BazelToolchain`` and ``Bazel``.
Docs: https://github.com/conan-io/docs/pull/2109

Close https://github.com/conan-io/conan/issues/6235

Shout-out to @technoir42, who worked on a first implementation of the generator (https://github.com/technoir42/conan-bazel)

### DIFF
--- a/conan/tools/google/bazel.py
+++ b/conan/tools/google/bazel.py
@@ -1,0 +1,49 @@
+import os
+import json
+
+from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
+from conans.util.files import load
+
+class Bazel(object):
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._get_bazel_project_configuration()
+
+    def configure(self, args=None):
+        pass
+
+    def build(self, args=None, label=None):
+        # TODO: Change the directory where bazel builds the project (by default, /var/tmp/_bazel_<username> )
+
+        bazelrc_path = '--bazelrc={}'.format(self._bazelrc_path) if self._bazelrc_path else ''
+        bazel_config = '--config={}'.format(self._bazel_config) if self._bazel_config else ''
+
+        # arch = self._conanfile.settings.get_safe("arch")
+        # cpu = {
+        #     "armv8": "arm64",
+        #     "x86_64": ""
+        # }.get(arch, arch)
+        #
+        # command = 'bazel {} build --sandbox_debug --subcommands=pretty_print --cpu={} {} {}'.format(
+        #     bazelrc_path,
+        #     cpu,
+        #     bazel_config,
+        #     label
+        # )
+
+        command = 'bazel {} build {} {}'.format(
+            bazelrc_path,
+            bazel_config,
+            label
+        )
+
+        self._conanfile.run(command)
+
+    def _get_bazel_project_configuration(self):
+        self._bazel_config = None
+        self._bazelrc_path = None
+
+        if os.path.exists(CONAN_TOOLCHAIN_ARGS_FILE):
+            conan_toolchain_args = json.loads(load(CONAN_TOOLCHAIN_ARGS_FILE))
+            self._bazel_config = conan_toolchain_args.get("bazel_config", None)
+            self._bazelrc_path = conan_toolchain_args.get("bazelrc_path", None)
--- a/conans/test/functional/toolchains/google/test_bazel.py
+++ b/conans/test/functional/toolchains/google/test_bazel.py
@@ -1,0 +1,193 @@
+import os
+import platform
+import textwrap
+import time
+import unittest
+
+import pytest
+from parameterized.parameterized import parameterized
+
+from conans.model.ref import ConanFileReference, PackageReference
+from conans.test.assets.cmake import gen_cmakelists
+from conans.test.assets.sources import gen_function_cpp, gen_function_h
+from conans.test.functional.utils import check_vs_runtime, check_exe_run
+from conans.test.utils.tools import TestClient
+from conans.util.files import save
+
+
+@pytest.mark.toolchain
+@pytest.mark.tool_bazel
+class Base(unittest.TestCase):
+
+    conanfile = textwrap.dedent("""
+        from conan.tools.google import Bazel
+        from conans import ConanFile
+
+        class App(ConanFile):
+            name="test_bazel_app"
+            version="0.0"
+            settings = "os", "compiler", "build_type", "arch"
+            generators = "BazelDeps", "BazelToolchain"
+            exports_sources = "WORKSPACE", "app/*"
+
+            def build(self):
+                bazel = Bazel(self)
+                bazel.configure()
+                bazel.build(label="//app:main")
+
+            def package(self):
+                self.copy('*', src='bazel-bin')
+        """)
+
+    buildfile = textwrap.dedent("""
+    load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")
+
+    cc_library(
+        name = "hello",
+        srcs = ["hello.cpp"],
+        hdrs = ["hello.h",],
+    )
+
+    cc_binary(
+        name = "main",
+        srcs = ["main.cpp"],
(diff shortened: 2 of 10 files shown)


conan/tools/google/bazel.py
```diff
@@ -0,0 +1,49 @@
...
+
+from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
+from conans.util.files import load
+
+class Bazel(object):
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._get_bazel_project_configuration()
+
+    def configure(self, args=None):
```
lasote, line 12: If there is no configure step for bazel you can nicely remove it

conan/tools/google/bazel.py
```diff
@@ -0,0 +1,49 @@
...
+        pass
+
+    def build(self, args=None, label=None):
+        # TODO: Change the directory where bazel builds the project (by default, /var/tmp/_bazel_<username> )
```
lasote, line 16: About the TODO. Why?

conans/test/functional/toolchains/google/test_bazel.py
```diff
@@ -0,0 +1,193 @@
...
+        command_str = "bazel-bin/app/main.exe" if bin_folder else "bazel-bin/app/main"
+        if platform.system() == "Windows":
+            command_str = command_str.replace("/", "\\")
+
+        self.client.run_command(command_str)
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Only for windows")
+class WinTest(Base):
+    @parameterized.expand(["Debug"])
```
lasote, line 119: Not very "expandable" right? LoL

lasote (commented):
Looking good and clean!

conan/tools/google/bazel.py
tapia: As I said to @memsharded in the other PR (I had to close it due to merge complications):

As far as I've seen, the convention when writing the build() method on a project's conanfile.py is:

```
def build(self):
    bazel = Bazel(self)
    bazel.configure()
    bazel.build(whatever)
```

Wouldn't deleting the configure method break this pattern?

conans/test/functional/toolchains/google/test_bazel.py
tapia: That's obviously temporal. The bazel generator can't handle build types yet, but it will 😄

conan/tools/google/bazel.py
tapia: That's @memsharded's suggestion. He said that the project should be built in the conan's cache. Shouldn't it?

conan/tools/google/bazel.py
memsharded: Not necessarily, the ``MSBuild`` helper does not have ``configure()`` step. It can be removed.

memsharded (approved):
Looks good to me for a start and getting users feedback. Lets add it to 1.37?

conan/tools/google/bazel.py
aiuto: I'm not sure that the concept "the directory where bazel builds the project" makes sense, as much as 'the directory of the Bazel project'

bazel must be run from somewhere in the source tree under a WORKSPACE file. It executes each build action in a symlink tree created for the action and places results in bazel-out in trees under bazel-out and bazel-bin.

After building a particular label or labels, they will appear somewhere under bazel-bin. From there, you might want to copy them out to conan's cache. It's hard to make a particular suggestion without seeing an example of how all the intermediate files should fit together.  For example, what would you expect for a configure based project that depends on libfoo where libfoo is built by Bazel.

You obviously need to have the glue that says
- for target X
- which produces an output file that you can't easily name (libfoo.a, libfoo.so, libfoo.lib, ...)
- tell configure how to use that library

Somehow we need to be able to tell configure not to used the installed libfoo. It needs to use the file under bazel-bin.  At least, that is what I presume. I don't know the conan workflow enough to know.

memsharded: Hi @aiuto, thanks for the feedback!

Let me briefly clarify the different flows:

For the consumer point of view: a project using bazel to build that wants to use Conan existing packages as dependencies:

- This seems to be the most requested use case
- The BazelDeps Conan generator can output files that Bazel could understand with cc_import, etc instructions to let Bazel locate those Conan packages in the Conan cache. Those packages could have been built with other build system, not a problem, Conan abstracts the package information and can translate it to different build systems.

For the producer point of view, a Conan package which build system is Bazel. 
- This is the case this ``Bazel`` Conan helper tries to implement. 
-  Conan packages when building from source, are built in the Conan cache, typically in ``.conan`` user folder, to not pollute the user projects, and to be able to reuse existing packages among projects.
- When building from source, Conan also contains a copy of the sources in such ".conan" cache, which might have been retrieved from SCM, for example.
- For all the other build systems, the build is done in a path that looks like ``.conan/data/mypkg/1.0/user/channel/build/binary_id/``. All the binaries and temporary files are there, and when Conan wants to do the final package, will copy selected artifacts to something like ``.conan/data/mypkg/1.0/user/channel/package/binary_id/``


It would be good if Bazel could be instructed to put the temporary build files in that location, instead of somewhere in ``bazel-out and bazel-bin``, that if I understood correctly, might live in ``/var/tmp/_bazel`` by default?

aiuto: Let's look at the consumer view.
I think a nicely transparant integration is one that points to the conan packages and then creates a local bazel repository that containds a BUILD file with the cc_import rules directly, so that the we can refer to them directly.
So the user experience is this:

WORKSPACE
```
workspace(name = 'myproject')
conan_repository(
  name = "foo", 
   artifacts=['a, 'b', .. list of top build targets from the conan package...]
)
```

BUILD
```
cc_library(
  name = 'myapp',
  deps = ['@foo//:a']
  ...
)
```

and the glue generated by the `conan_repository` rule is 

WORKSPACE
```
workspace(name=foo)
```
BUILD
```
cc_import(
  name = 'a',
  hdrs = [exported headers of the artifact a],
  static_library = "libmylib.a",
)
```
libmylib.a: symlink to libmylib.a built as a result of doing the conan build.

memsharded:
Let's merge this and released it in 1.37 (tomorrow), even if there is feedback to check, and potential follow-up improvements, this has 0 risk, and better have it merged for further feedback and testing in 1.37 release.

conan/tools/google/bazel.py
tapia: The merged generator is pretty similar to your idea. The dependencies are declared in the `conanfile.py`, though, not in the `WORKSPACE` file. For example:

```
class BazelExampleConan(ConanFile):
    name = "bazel-example"
    ....
    build_requires = "boost/1.76.0"
```

Then, the only thing you have to add to the `WORKSPACE` file is this:

```
load("@//conandeps:dependencies.bzl", "load_conan_dependencies")
load_conan_dependencies()
```

After that, just update the `BUILD` files where you need to use the new dependency:

```
cc_binary(
    name = "hello-world",
    srcs = ["hello-world.cc"],
    deps = [
        "@boost//:boost",
    ],
)
```

Give it a try and let us know what you think :-)

PR #9005, opened 2021-05-26
Feature: use CMake File API

### LINKED ISSUE #7764: [feature] Extract CMake project information with code model
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md), sir.

When a library maintainer package his library with conan, he needs to duplicate some information like C++ flags and definitions.
It is possible to use the [cmake file api](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html) to get the information about the project and targets.

## How it works

```
<build folder>/.cmake/api/v1/query/client-conan/query.json
```
```json
{
    "requests": [
        {
            "kind": "codemodel",
            "version": 2
        }
    ]
}
```

During the `configure` phase, CMake will generate files inside `<build folder>/.cmake/api/v1/reply/` folder. This folder contains an index file named `<build folder>/.cmake/api/v1/reply/index-<unspecified>.json` (Ex: `index-2020-09-24T15-20-32-0066.json`):
```json
{
	...
	"reply" : 
	{
		"client-conan" : 
		{
			"query.json" : 
			{
				"requests" : 
				[
					{
						"kind" : "codemodel",
						"version" : 2
					}
				],
				"responses" : 
				[
					{
						"jsonFile" : "codemodel-v2-a5a1537322975cd7e68c.json",
						"kind" : "codemodel",
						"version" : 
						{
							"major" : 2,
							"minor" : 0
						}
					}
				]
			}
		},
        ...
	}
}
```
The `codemodel-v2-<undefined>.json` contains all the information about projects, targets and configurations (It's a huge file so i dont paste an example here). [Its reference documentation is here](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2).

## How to integrate it with Conan

In my opinion, the best way is to integrate it into the `configure` method of the CMake build helper. The helper can create the request file and extract information on demand.

## Known limitations

- The codemodel isn't accessible inside `package_info` method because the codemodel is produced at build time. It may however be possible to export the codemodel in a file inside the package folder and parse it in the `package_info` method.
- The codemodel currently contains both public and private definitions and we can't determine their scope.

SSE4:
closes: #7764
(mainly based on my earlier POC: https://github.com/SSE4/cci.cmake-file-api)

implementation is based on several important pieces:

1. `CMakeFileAPI` object (conan.tools.cmake.file_api).
the object is responsible to communicate with [CMake File API](https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html), primarily, composing requests and receiving (and parsing responses). the object itself doesn't invoke CMake configuration, so it relies on the existing `CMake` build helper (conan.tools.cmake). right now, only `codemodel-v2` requests/responses are implemented, but in future, we could add others on demand. right now it's very limited, and extracts the following information only:
- target names
- library name to link with
- requirements

2. `CppPackage` object (conans.model.cpp_package).
~~NOTE: I personally don't like the name at the moment, as it introduces a lot of confusing, as in internal code, as for recipe authors, but we agreed before for file named `conan_package.json`, although "conan package" represents a different entity.~~
represents a `cpp_package.json` file, containing serialized, build-system agnostic package information (e.g. similar to CppInfo). it's also not a conan-specific, so theoretically, this model could be used wider in C++ community. the purpose of object is to serialize and de-serialize data, as well as provide `package_info` helper method to automatically populate `CppInfo` object.
(`package` method is executed by the package creator, while `package_info` is executed by the consumer. in order to preserve the information, it has to be somehow packaged together with other artifacts, like conanfile itself, exported sources, headers and libraries)
3. ~~`.autopackage`~~
~~it might be boring to always manually copy the `conan_package.json`, so I have a speculation, most of consumers and creators might want this to be performed automatically and transparently, behind the scenes. therefore, small code block always copies all the contents of the `.autopackage` subdirectory (which is expected to contain `conan_package.json`).~~
~~on the other hand, this could be considered a "magic", so I am okay to remove it, as it's not strictly required.~~
4. ~~`package_info`~~
~~now, `package_info` always looks for the `conan_package.json`, and if it's already there, it is de-serialized, and `ConanPackage` object is constructed, in order to populate `CppInfo`. this happens before original conanfile's `package_info` invocation, so recipe creator always has a power to correct, or append the information.~~
5. there are some small modifications to the `gen_cmakelists`, mostly to support several targets and dependencies between them.

/cc @memsharded 

Changelog: Feature: Use _CMake_ File API.
Docs: omit

### DIFF
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -12,6 +12,8 @@ class CMakeFileAPI(object):
     """
     CODEMODELV2 = "codemodel-v2"
     SKIP_TARGETS = ["ZERO_CHECK", "ALL_BUILD"]
+    # https://cmake.org/cmake/help/v3.21/prop_tgt/TYPE.html
+    SUPPORTED_TARGET_TYPES = ['STATIC_LIBRARY', 'SHARED_LIBRARY', 'MODULE_LIBRARY']
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
@@ -129,6 +131,8 @@ class CMakeFileAPI(object):
             conan_package = CppPackage()
 
             for name, target in self._components.items():
+                if target['type'] not in CMakeFileAPI.SUPPORTED_TARGET_TYPES:
+                    continue
                 component = conan_package.add_component(name)
                 # TODO: CMakeDeps has nothing to do here
                 component.names["CMakeDeps"] = name
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -1,7 +1,9 @@
+import json
 import fnmatch
 import os
 
-from conans.util.files import save
+from conans.model.cpp_package import CppPackage
+from conans.util.files import save, load
 
 class CMakeFileAPI(object):
     """
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -2,8 +2,9 @@ import json
 import fnmatch
 import os
 
-from conans.model.cpp_package import CppPackage
+from conan.tools.files import CppPackage
 from conans.util.files import save, load
+
 
 class CMakeFileAPI(object):
     """
@@ -16,18 +17,11 @@ class CMakeFileAPI(object):
         self._conanfile = conanfile
 
     @property
-    def build_dir(self):
-        """
-        :return: CMake build directory (the one containing CMakeCache.txt)
-        """
-        return self._conanfile.build_folder
-
-    @property
     def api_dir(self):
         """
         :return: api directory <build>/.cmake/api/v1/
         """
-        return os.path.join(self.build_dir, ".cmake", "api", "v1")
+        return os.path.join(self._conanfile.build_folder, ".cmake", "api", "v1")
 
     @property
     def query_dir(self):
@@ -50,7 +44,11 @@ class CMakeFileAPI(object):
         :return: new query object
         """
         if query == self.CODEMODELV2:
-            return self.CodeModelQueryV2(self)
+            # implements codemodel-v2 query
+            # https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+            os.makedirs(self.query_dir)
+            save(os.path.join(self.query_dir, "codemodel-v2"), "")
+            return
         raise NotImplementedError()
 
     def reply(self, reply):
@@ -69,18 +67,6 @@ class CMakeFileAPI(object):
         :return: active build type (configuration)
         """
         return self._conanfile.settings.get_safe("build_type")
-
-    class CodeModelQueryV2(object):
-        """
-        implements codemodel-v2 query
-        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
-        """
-        def __init__(self, api):
-            """
-            :param api:
-            """
-            os.makedirs(api.query_dir)
-            save(os.path.join(api.query_dir, "codemodel-v2"), "")
 
     class CodeModelReplyV2(object):
         """
@@ -110,20 +96,21 @@ class CMakeFileAPI(object):
                         loadjs(os.path.join(api.reply_dir, target['jsonFile']))
 
         @classmethod
-        def _name_on_disk2lib(cls, nameOnDisk):
+        def _name_on_disk2lib(cls, name_on_disk):
             """
             convert raw library file name into conan-friendly one
-            :param nameOnDisk: raw library file name read from target.json
+            :param name_on_disk: raw library file name read from target.json
             :return: conan-friendly library name, without suffix and prefix
             """
-            if nameOnDisk.endswith('.lib'):
-                return nameOnDisk[:-4]
-            elif nameOnDisk.endswith('.dll'):
-                return nameOnDisk[:-4]
-            elif nameOnDisk.startswith('lib') and nameOnDisk.endswith('.a'):
-                return nameOnDisk[3:-2]
+            if name_on_disk.endswith('.lib'):
+                return name_on_disk[:-4]
+            elif name_on_disk.endswith('.dll'):
+                return name_on_disk[:-4]
+            elif name_on_disk.startswith('lib') and name_on_disk.endswith('.a'):
+                return name_on_disk[3:-2]
             else:
-                raise Exception("don't know how to convert %s" % nameOnDisk)
+                # FIXME: This fails to parse executables
+                raise Exception("don't know how to convert %s" % name_on_disk)
 
         @classmethod
         def _parse_dep_name(cls, name):
@@ -143,6 +130,7 @@ class CMakeFileAPI(object):
 
             for name, target in self._components.items():
                 component = conan_package.add_component(name)
+                # TODO: CMakeDeps has nothing to do here
                 component.names["CMakeDeps"] = name
                 component.libs = [self._name_on_disk2lib(target['nameOnDisk'])]
                 deps = target["dependencies"] if 'dependencies' in target else []
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -138,7 +138,6 @@ class CMakeFileAPI(object):
             converts codemodel-v2 into conan_package.json object
             :return: ConanPackage instance
             """
-            from conans.model.cpp_package import CppPackage
 
             conan_package = CppPackage()
 
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -1,0 +1,153 @@
+import fnmatch
+import os
+
+from conans.util.files import save
+
+class CMakeFileAPI(object):
+    """
+    implements https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html
+    """
+    CODEMODELV2 = "codemodel-v2"
+    SKIP_TARGETS = ["ZERO_CHECK", "ALL_BUILD"]
+
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+
+    @property
+    def build_dir(self):
+        """
+        :return: CMake build directory (the one containing CMakeCache.txt)
+        """
+        return self._conanfile.build_folder
+
+    @property
+    def api_dir(self):
+        """
+        :return: api directory <build>/.cmake/api/v1/
+        """
+        return os.path.join(self.build_dir, ".cmake", "api", "v1")
+
+    @property
+    def query_dir(self):
+        """
+        :return: api query sub-directory <build>/.cmake/api/v1/query
+        """
+        return os.path.join(self.api_dir, "query")
+
+    @property
+    def reply_dir(self):
+        """
+        :return: api reply sub-directory <build>/.cmake/api/v1/reply
+        """
+        return os.path.join(self.api_dir, "reply")
+
+    def query(self, query):
+        """
+        prepare the CMake File API query (the actual query will be done during the configure step)
+        :param query: type of the CMake File API query (e.g. CODEMODELV2)
+        :return: new query object
+        """
+        if query == self.CODEMODELV2:
+            return self.CodeModelQueryV2(self)
+        raise NotImplementedError()
+
+    def reply(self, reply):
+        """
+        obtain the CMake File API reply (which should have been made during the configure step)
+        :param reply: type of the CMake File API reply (e.g. CODEMODELV2)
+        :return: new reply object
+        """
+        if reply == self.CODEMODELV2:
+            return self.CodeModelReplyV2(self)
+        raise NotImplementedError()
+
+    @property
+    def build_type(self):
+        """
+        :return: active build type (configuration)
+        """
+        return self._conanfile.settings.get_safe("build_type")
+
+    class CodeModelQueryV2(object):
+        """
+        implements codemodel-v2 query
+        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+        """
+        def __init__(self, api):
+            """
+            :param api:
+            """
+            os.makedirs(api.query_dir)
+            save(os.path.join(api.query_dir, "codemodel-v2"), "")
+
+    class CodeModelReplyV2(object):
+        """
+        implements codemodel-v2 reply
+        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+        """
+        def __init__(self, api):
+
+            def loadjs(filename):
+                from conans.util.files import load
+                import json
+                return json.loads(load(filename))
+
+            codemodels = os.listdir(api.reply_dir)
+            codemodels = [c for c in codemodels if fnmatch.fnmatch(c, "codemodel-v2-*.json")]
+            assert len(codemodels) == 1
+            self._codemodel = loadjs(os.path.join(api.reply_dir, codemodels[0]))
+
+            self._configurations = dict()
+            for configuration in self._codemodel['configurations']:
+                if configuration["name"] != api.build_type:
+                    continue
+
+                self._components = dict()
+                for target in configuration["targets"]:
+                    if target['name'] in CMakeFileAPI.SKIP_TARGETS:
+                        continue
+                    self._components[target["name"]] = \
+                        loadjs(os.path.join(api.reply_dir, target['jsonFile']))
+
+        @classmethod
+        def _name_on_disk2lib(cls, nameOnDisk):
+            """
+            convert raw library file name into conan-friendly one
+            :param nameOnDisk: raw library file name read from target.json
+            :return: conan-friendly library name, without suffix and prefix
+            """
+            if nameOnDisk.endswith('.lib'):
+                return nameOnDisk[:-4]
+            elif nameOnDisk.endswith('.dll'):
+                return nameOnDisk[:-4]
+            elif nameOnDisk.startswith('lib') and nameOnDisk.endswith('.a'):
+                return nameOnDisk[3:-2]
+            else:
+                raise Exception("don't know how to convert %s" % nameOnDisk)
+
+        @classmethod
+        def _parse_dep_name(cls, name):
+            """
+            :param name: extract dependency name from the id like 'decoder::@5310dfab9e417c587352'
+            :return: dependency name (part before ::@ token)
+            """
+            return name.split("::@")[0]
+
+        def to_conan_package(self):
+            """
+            converts codemodel-v2 into conan_package.json object
+            :return: ConanPackage instance
+            """
+            from conans.model.cpp_package import CppPackage
+
+            conan_package = CppPackage()
+
+            for name, target in self._components.items():
+                component = conan_package.add_component(name)
+                component.names["CMakeDeps"] = name
+                component.libs = [self._name_on_disk2lib(target['nameOnDisk'])]
+                deps = target["dependencies"] if 'dependencies' in target else []
+                deps = [self._parse_dep_name(d['id']) for d in deps]
+                component.requires = [d for d in deps if d not in CMakeFileAPI.SKIP_TARGETS]
+
+            return conan_package
--- a/conan/tools/cmake/file_api.py
+++ b/conan/tools/cmake/file_api.py
@@ -90,8 +90,6 @@ class CMakeFileAPI(object):
         def __init__(self, api):
 
             def loadjs(filename):
-                from conans.util.files import load
-                import json
                 return json.loads(load(filename))
 
             codemodels = os.listdir(api.reply_dir)
--- a/conan/tools/files/cpp_package.py
+++ b/conan/tools/files/cpp_package.py
@@ -1,6 +1,7 @@
 import json
 from conans.util.files import load, save
 
+
 class CppPackage(object):
     """
     models conan_package.json, serializable object
@@ -33,7 +34,9 @@ class CppPackage(object):
         :param filename: path to the cpp_package.json
         :return: None
         """
-        save(filename, json.dumps(self, default=lambda o: o.__dict__))
+        text = json.dumps(self, default=lambda o: o.__dict__, indent=4)
+        print(text)
+        save(filename, text)
 
     def package_info(self, conanfile):
         """
--- a/conan/tools/files/cpp_package.py
+++ b/conan/tools/files/cpp_package.py
@@ -35,7 +35,6 @@ class CppPackage(object):
         :return: None
         """
         text = json.dumps(self, default=lambda o: o.__dict__, indent=4)
-        print(text)
         save(filename, text)
 
     def package_info(self, conanfile):
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -38,7 +38,7 @@ class CppPackage(object):
     def package_info(self, conanfile):
         """
         performs an automatically generated package_info method on conanfile, populating
-        conanfile.package_info with the information available inside conan_package.json
+        conanfile.package_info with the information available inside cpp_package.json
         :return: None
         """
         for cname, component in self.components.items():
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -18,7 +18,6 @@ class CppPackage(object):
         :param filename: path to the conan_package.json
         :return: a new instance of ConanPackage
         """
-        import json
         from conans.util.files import load
         def from_json(o):
             if "components" in o:
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -33,8 +33,6 @@ class CppPackage(object):
         :param filename: path to the conan_package.json
         :return: None
         """
-        import json
-        from conans.util.files import save
         save(filename, json.dumps(self, default=lambda o: o.__dict__))
 
     def package_info(self, conanfile):
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -29,8 +29,8 @@ class CppPackage(object):
 
     def save(self, filename=DEFAULT_FILENAME):
         """
-        saves package model from memory into the conan_package.json file
-        :param filename: path to the conan_package.json
+        saves package model from memory into the cpp_package.json file
+        :param filename: path to the cpp_package.json
         :return: None
         """
         save(filename, json.dumps(self, default=lambda o: o.__dict__))
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -15,8 +15,8 @@ class CppPackage(object):
     def load(cls, filename=DEFAULT_FILENAME):
         """
         loads package model into memory from the conan_package.json file
-        :param filename: path to the conan_package.json
-        :return: a new instance of ConanPackage
+        :param filename: path to the cpp_package.json
+        :return: a new instance of CppPackage
         """
         def from_json(o):
             if "components" in o:
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -49,9 +49,9 @@ class CppPackage(object):
 
     def add_component(self, name):
         """
-        appens a new ConanPackage.Configuration into the internal dictionary
+        appens a new CppPackage.Configuration into the internal dictionary
         :param name: name of the given configuration (e.g. Debug)
-        :return: a new ConanPackage.Configuration instance (empty)
+        :return: a new CppPackage.Configuration instance (empty)
         """
         self.components[name] = CppPackage.Component()
         return self.components[name]
--- a/conans/model/cpp_package.py
+++ b/conans/model/cpp_package.py
@@ -45,7 +45,7 @@ class CppPackage(object):
             conanfile.cpp_info.components[cname].libs = component.libs
             conanfile.cpp_info.components[cname].requires = component.requires
             for generator, gname in component.names.items():
-                conanfile.cpp_info.components[cname].names[generator] = gname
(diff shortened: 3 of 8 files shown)


conans/client/conanfile/package.py
```diff
@@ -51,6 +51,14 @@ def _call_package(conanfile, package_id, hook_manager, conanfile_path, ref, copy
                                    "'self.info' access in package() method is deprecated"):
                 conanfile.package()
 
+    # auto-package: all the files within .autopackage sub-directory are automatically copied
+    # so package author doesn't need to explicitly copy them
+    auto_package_src = os.path.join(conanfile.build_folder, ".autopackage")
+    auto_package_dst = os.path.join(conanfile.package_folder, ".autopackage")
+    if os.path.isdir(auto_package_src):
+        copier = FileCopier([auto_package_src], auto_package_dst)
```
memsharded: Avoid ``FileCopier`` for something that should be a ``shutil.copy``

conans/model/conan_package.py
```diff
@@ -0,0 +1,94 @@
...
+    def add_configuration(self, name):
+        """
+        appens a new ConanPackage.Configuration into the internal dictionary
+        :param name: name of the given configuration (e.g. Debug)
+        :return: a new ConanPackage.Configuration instance (empty)
+        """
+        self.configurations[name] = ConanPackage.Configuration()
+        return self.configurations[name]
+
+    class Configuration(object):
```
memsharded: Package descriptions should be configuration agnostic, please do not include specific Debug/Release things.

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,142 @@
...
+
+        class Triunfo(ConanFile):
+            name = "triunfo"
+            version = "1.0"
+            settings = "os", "compiler", "arch", "build_type"
+            exports_sources = "*CMakeLists.txt", "*.cpp", "*.h"
+            generators = "CMakeToolchain"
+            _cmake = None
+
+            def _configure_cmake(self):
```
memsharded: This idiom should not be longer necessary with CMakeDeps and CMakeToolchain, please avoid using this ``_configure_cmake()`` method.

memsharded (commented):
Looks good, I would like to try it as a user, please try to make the tests green, thanks!

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: how come it is no longer necessary? it will invoke `configure` twice, once for `build`, and once `package` otherwise

memsharded: package() is called after build(), no need to call configure() again.
That was due to an old misbehavior in the helper that required it to be called twice to do the right thing. No longer necessary with the new toolchain.

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,133 @@
...
+                cmake = CMake(self)
+                cmake.configure()
+                reply = file_api.reply(CMakeFileAPI.CODEMODELV2)
+                package = reply.to_conan_package()
+                package.save()
+                cmake.build()
+
+            def package(self):
+                cmake = CMake(self)
+                cmake.configure()
```
SSE4:
```suggestion
```

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,133 @@
...
+    use_transcoder_cpp = gen_function_cpp(name="main", includes=["transcoder"], calls=["transcoder"])
+    use_decoder_cmake = gen_cmakelists(appname="use_decoder", appsources=["use_decoder.cpp"],
+                                       find_package={"triunfo": "decoder"})
+    use_encoder_cmake = gen_cmakelists(appname="use_encoder", appsources=["use_encoder.cpp"],
+                                       find_package={"triunfo": "encoder"})
+    use_transcoder_cmake = gen_cmakelists(appname="use_transcoder", appsources=["use_transcoder.cpp"],
+                                          find_package={"triunfo": "transcoder"})
+    common_cmake = textwrap.dedent("""
+        cmake_minimum_required(VERSION 2.8)
+        project(triunfo)
```
SSE4:
```suggestion
        project(elogio)
```

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,133 @@
...
+from conans.test.utils.tools import TestClient
+
+
+@pytest.mark.tool_cmake
+def test_file_api():
+    """
+    simple library providing 3 targets:
+    - decoder
+    - encoder
+    - transcoder (requires decoder and encoder)
```
SSE4, line 17:
```suggestion
    - transcoder (requires decoder and encoder)
    generates the following targets:
    - triunfo::decoder
    - triunfo::encoder
    - triunfo::transcoder (depends on triunfo::decoder and triunfo::encoder)
    consumer uses find_package(triunfo COMPONENTS <component>)
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+    class CodeModelQueryV2(object):
+        """
+        implements codemodel-v2 query
+        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+        """
+        def __init__(self, api):
+            """
+            :param api:
+            """
+            from conans.util.files import save
```
memsharded: No need to do local imports, can be moved to file top

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+            os.makedirs(api.query_dir)
+            save(os.path.join(api.query_dir, "codemodel-v2"), "")
+
+    class CodeModelReplyV2(object):
+        """
+        implements codemodel-v2 reply
+        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+        """
+        def __init__(self, api):
+            import fnmatch
```
memsharded: Same with imports.

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+
+                self._components = dict()
+                for target in configuration["targets"]:
+                    if target['name'] in CMakeFileAPI.SKIP_TARGETS:
+                        continue
+                    self._components[target["name"]] = \
+                        loadjs(os.path.join(api.reply_dir, target['jsonFile']))
+
+        @classmethod
+        def _nameOnDisk2lib(cls, nameOnDisk):
```
memsharded: Please pep-8 naming (lowercase for methods and variables)

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+            """
+            convert raw library file name into conan-friendly one
+            :param nameOnDisk: raw library file name read from target.json
+            :return: conan-friendly library name, without suffix and prefix
+            """
+            import fnmatch
+            if fnmatch.fnmatch(nameOnDisk, '*.lib'):
```
memsharded: why fnmatch and not ``endswith``?

conans/client/conanfile/package.py
```diff
@@ -51,6 +51,13 @@ def _call_package(conanfile, package_id, hook_manager, conanfile_path, ref, copy
                                    "'self.info' access in package() method is deprecated"):
                 conanfile.package()
 
+    # auto-package: all the files within .autopackage sub-directory are automatically copied
+    # so package author doesn't need to explicitly copy them
+    auto_package_src = os.path.join(conanfile.build_folder, ".autopackage")
```
memsharded: Yes, better not hard-code this in the core. This can be made part later of the ``FileCopier``, or any other helper that users will use in their ``package()`` method. Lets make the copy explicit in recipes atm, no need of ``.autopackage`` now.

conans/client/installer.py
```diff
@@ -671,6 +672,10 @@ def _call_package_info(self, conanfile, package_folder, ref, is_editable):
                         else:
                             conanfile.cpp_info.filter_empty = False
 
+                    if os.path.isfile(ConanPackage.DEFAULT_FILENAME):
```
memsharded: Same, do not harcode this in the core, recipes that want to use this feature can read it explicitly in recipes, lets have something in the line:

```python
def package_info(self):
     self.cpp = ConanPackage.load("cpp_package.json")
```

conans/model/conan_package.py
```diff
@@ -0,0 +1,74 @@
+import os
+
+class ConanPackage(object):
+    """
+    models conan_package.json, serializable object
+    """
+
+    DEFAULT_FILENAME = os.path.join(".autopackage", "conan_info.json")
```
memsharded: Yes, lets call this ``cpp_package.json``, lets make it generic.

conans/model/conan_package.py
```diff
@@ -0,0 +1,74 @@
+import os
+
+class ConanPackage(object):
```
memsharded: Agree with your comments, lets call it ``CppPackage``

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,137 @@
...
+        class Triunfo(ConanFile):
+            name = "triunfo"
+            version = "1.0"
+            settings = "os", "compiler", "arch", "build_type"
+            exports_sources = "*CMakeLists.txt", "*.cpp", "*.h"
+            generators = "CMakeToolchain"
+
+            def build(self):
+                file_api = CMakeFileAPI(self)
+                file_api.query(CMakeFileAPI.CODEMODELV2)
```
memsharded, line 40: This (the query) needs to happen before calling cmake? or can be done later?

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,137 @@
...
+                cmake = CMake(self)
+                cmake.configure()
+                reply = file_api.reply(CMakeFileAPI.CODEMODELV2)
```
memsharded, line 43: I think I would probably allow some defaults, like not needing CODEMODELV2. Most likely later, no need to do it now, but UX seems a bit too detailed here.

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,137 @@
...
+                package = reply.to_conan_package()
+                package.save()
+                cmake.build()
+
+            def package(self):
+                cmake = CMake(self)
+                cmake.install()
+
+            def package_info(self):
+                # auto-generated from CMakeFileAPI
```
memsharded: Make it explicit

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,137 @@
...
+                cmake = CMake(self)
+                cmake.configure()
+                reply = file_api.reply(CMakeFileAPI.CODEMODELV2)
+                package = reply.to_conan_package()
+                package.save()
+                cmake.build()
+
+            def package(self):
+                cmake = CMake(self)
+                cmake.install()
```
memsharded, line 50: Make it explicit

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,137 @@
...
+    use_encoder_cpp = gen_function_cpp(name="main", includes=["encoder"], calls=["encoder"])
+    use_transcoder_cpp = gen_function_cpp(name="main", includes=["transcoder"], calls=["transcoder"])
+    use_decoder_cmake = gen_cmakelists(appname="use_decoder", appsources=["use_decoder.cpp"],
+                                       find_package={"triunfo": "decoder"})
+    use_encoder_cmake = gen_cmakelists(appname="use_encoder", appsources=["use_encoder.cpp"],
+                                       find_package={"triunfo": "encoder"})
+    use_transcoder_cmake = gen_cmakelists(appname="use_transcoder", appsources=["use_transcoder.cpp"],
+                                          find_package={"triunfo": "transcoder"})
+    common_cmake = textwrap.dedent("""
+        cmake_minimum_required(VERSION 2.8)
```
memsharded, line 127: Use 3.15 as a base

conans/test/functional/toolchains/cmake/test_ninja.py
```diff
@@ -78,7 +78,7 @@ def test_locally_build_linux(build_type, shared, client):
     assert 'cmake -G "Ninja"' in client.out
     assert "main: {}!".format(build_type) in client.out
     client.run("install hello/1.0@ -g=deploy -if=mydeploy {}".format(settings))
-    ldpath = os.path.join(client.current_folder, "mydeploy", "hello")
+    ldpath = os.path.join(client.current_folder, "mydeploy", "hello", "lib")
```
memsharded, line 81: Why? This seems to be fine without this change, and also with this change?

memsharded (commented):
If we remove the parts from the core, and make recipes use it explicitly, I think this would be low risk and we could release it soon, to start getting feedback from users.

conan/tools/cmake/file_api.py
SSE4:
```suggestion
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
+import os
```
SSE4, line 3:
```suggestion
import fnmatch
import os

from conans.util.files import save
```

conan/tools/cmake/file_api.py
SSE4:
```suggestion
```

conan/tools/cmake/file_api.py
SSE4:
```suggestion
            if nameOnDisk.endswith('.lib'):
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+                        loadjs(os.path.join(api.reply_dir, target['jsonFile']))
+
+        @classmethod
+        def _nameOnDisk2lib(cls, nameOnDisk):
+            """
+            convert raw library file name into conan-friendly one
+            :param nameOnDisk: raw library file name read from target.json
+            :return: conan-friendly library name, without suffix and prefix
+            """
+            import fnmatch
```
SSE4:
```suggestion
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+            if fnmatch.fnmatch(nameOnDisk, '*.lib'):
+                return nameOnDisk[:-4]
+            elif fnmatch.fnmatch(nameOnDisk, '*.dll'):
```
SSE4:
```suggestion
            elif nameOnDisk.endswith('.dll'):
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+                return nameOnDisk[:-4]
+            elif fnmatch.fnmatch(nameOnDisk, 'lib*.a'):
```
SSE4:
```suggestion
            elif nameOnDisk.startswith('lib') and nameOnDisk.endswith('.a'):
```

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: why? I prefer not to set unnecessarily high requirements, but instead ensure maximum compatibility

conans/test/functional/toolchains/cmake/test_ninja.py
SSE4: I've added `LIBRARY DESTINATION lib`, so libraries are deployed to the `lib` sub-directory with `gen_cmakelists`.
otherwise, all recipes will have to add `self.package_info.libdirs.append(self.package_folder)` (while `lib` is automatically added by conan)

conan/tools/cmake/file_api.py
SSE4: both will work and will have the same effect, so I's just style preference. it's again python providing too many ways to do the same thing. changed to `endswith`.

conan/tools/cmake/file_api.py
SSE4:
```suggestion
        def _name_on_disk2lib(cls, nameOnDisk):
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+            :return: ConanPackage instance
+            """
+            from conans.model.conan_package import ConanPackage
+
+            conan_package = ConanPackage()
+
+            for name, target in self._components.items():
+                component = conan_package.add_component(name)
+                component.names["CMakeDeps"] = name
+                component.libs = [self._nameOnDisk2lib(target['nameOnDisk'])]
```
SSE4:
```suggestion
                component.libs = [self._name_on_disk2lib(target['nameOnDisk'])]
```

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: I think we already agreed to prefer explicit things in conan codebase, so I don't have to think which API call is magically invoked under the hood.

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: yes, before

conans/model/conan_package.py
SSE4:
```suggestion
    DEFAULT_FILENAME = "cpp_package.json"
```

conan/tools/cmake/file_api.py
SSE4: moved to the top

conan/tools/cmake/file_api.py
SSE4: moved to the top

conan/tools/cmake/file_api.py
SSE4: renamed

conans/client/conanfile/package.py
SSE4: removed that part for now

conans/client/installer.py
SSE4: removed that part for now

conans/model/conan_package.py
SSE4: renamed

conans/model/conan_package.py
SSE4: renamed

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: made it explicit

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: made it explicit

conans/test/functional/toolchains/cmake/test_file_api.py
memsharded: Because policies change based on this version, and something might not work the same with an older version, it is not the first time something in Conan didn't work as expected because of this version.

conans/model/cpp_package.py
```diff
@@ -0,0 +1,72 @@
...
+        self.components = dict()
+
+    @classmethod
+    def load(cls, filename=DEFAULT_FILENAME):
+        """
+        loads package model into memory from the conan_package.json file
+        :param filename: path to the conan_package.json
+        :return: a new instance of ConanPackage
+        """
+        import json
```
memsharded: More local imports, I didn't mark them all, please do not use local imports unless necessary to avoid cycles.

conans/test/functional/toolchains/cmake/test_file_api.py
SSE4: they do change, but most of conan tests use `2.8` or `2.8.12` right now, so I am not sure, what exactly are we trying to catch here.
cmake files here are pretty simple and don't use any sophisticated features that would require newer release.

conans/test/functional/toolchains/cmake/test_file_api.py
memsharded: One thing is to have things explicitly, no magic, like having to instantiate explicitly ``CMakeFileApi`` in the ``build()`` method, and a different one is that helpers like this one cannot have some defaults that simplify the UI for most of the cases, or at least for some reference default.

memsharded (commented):
Looking better, thanks!

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+    class CodeModelReplyV2(object):
+        """
+        implements codemodel-v2 reply
+        https://cmake.org/cmake/help/latest/manual/cmake-file-api.7.html#codemodel-version-2
+        """
+        def __init__(self, api):
+
+            def loadjs(filename):
+                from conans.util.files import load
+                import json
```
SSE4:
```suggestion
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
+import fnmatch
+import os
+
+from conans.util.files import save
```
SSE4:
```suggestion
import json
import fnmatch
import os

from conans.model.cpp_package import CppPackage
from conans.util.files import save, load
```

conan/tools/cmake/file_api.py
```diff
@@ -0,0 +1,153 @@
...
+            :return: dependency name (part before ::@ token)
+            """
+            return name.split("::@")[0]
+
+        def to_conan_package(self):
+            """
+            converts codemodel-v2 into conan_package.json object
+            :return: ConanPackage instance
+            """
+            from conans.model.cpp_package import CppPackage
```
SSE4:
```suggestion
```

conans/test/functional/toolchains/cmake/test_file_api.py
memsharded: Many of the new tests of the new integration are already using ``cmake_minimum_required(VERSION 3.15)``, because the behavior when using the generated xxx-config.cmake or conan_toolchain.cmake, was not correct. Maybe in this test not an issue yet, I am fine to leave it, but it might fail later down the road when something change.

conans/model/cpp_package.py
```diff
@@ -0,0 +1,72 @@
+class CppPackage(object):
```
SSE4:
```suggestion
import json
from conans.util.files import load, save

class CppPackage(object):
```

conans/model/cpp_package.py
SSE4:
```suggestion
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @classmethod
+    def load(cls, filename=DEFAULT_FILENAME):
+        """
+        loads package model into memory from the conan_package.json file
+        :param filename: path to the conan_package.json
+        :return: a new instance of ConanPackage
+        """
+        import json
+        from conans.util.files import load
```
SSE4:
```suggestion
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,72 @@
...
+        return conan_package
+
+    def save(self, filename=DEFAULT_FILENAME):
+        """
+        saves package model from memory into the conan_package.json file
+        :param filename: path to the conan_package.json
+        :return: None
+        """
+        import json
+        from conans.util.files import save
```
SSE4:
```suggestion
```

conans/model/cpp_package.py
SSE4: removed them all

conans/model/cpp_package.py
```diff
@@ -0,0 +1,71 @@
...
+            else:
+                return o
+        conan_package = CppPackage()
+        conan_package.components = json.loads(load(filename), object_hook=from_json)
+        return conan_package
+
+    def save(self, filename=DEFAULT_FILENAME):
+        """
+        saves package model from memory into the conan_package.json file
+        :param filename: path to the conan_package.json
```
SSE4:
```suggestion
        saves package model from memory into the cpp_package.json file
        :param filename: path to the cpp_package.json
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def __init__(self):
+        self.components = dict()
+
+    @classmethod
+    def load(cls, filename=DEFAULT_FILENAME):
+        """
+        loads package model into memory from the conan_package.json file
+        :param filename: path to the conan_package.json
+        :return: a new instance of ConanPackage
```
SSE4:
```suggestion
        :param filename: path to the cpp_package.json
        :return: a new instance of CppPackage
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,71 @@
...
+        saves package model from memory into the conan_package.json file
+        :param filename: path to the conan_package.json
+        :return: None
+        """
+        save(filename, json.dumps(self, default=lambda o: o.__dict__))
+
+    def package_info(self, conanfile):
+        """
+        performs an automatically generated package_info method on conanfile, populating
+        conanfile.package_info with the information available inside conan_package.json
```
SSE4:
```suggestion
        conanfile.package_info with the information available inside cpp_package.json
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,71 @@
...
+            conanfile.cpp_info.components[cname].libs = component.libs
+            conanfile.cpp_info.components[cname].requires = component.requires
+            for generator, gname in component.names.items():
+                conanfile.cpp_info.components[cname].names[generator] = gname
+
+    def add_component(self, name):
+        """
+        appens a new ConanPackage.Configuration into the internal dictionary
+        :param name: name of the given configuration (e.g. Debug)
+        :return: a new ConanPackage.Configuration instance (empty)
```
SSE4:
```suggestion
        appens a new CppPackage.Configuration into the internal dictionary
        :param name: name of the given configuration (e.g. Debug)
        :return: a new CppPackage.Configuration instance (empty)
```

conans/model/cpp_package.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        performs an automatically generated package_info method on conanfile, populating
+        conanfile.package_info with the information available inside cpp_package.json
+        :return: None
+        """
+        for cname, component in self.components.items():
+            conanfile.cpp_info.components[cname].libs = component.libs
+            conanfile.cpp_info.components[cname].requires = component.requires
+            for generator, gname in component.names.items():
+                conanfile.cpp_info.components[cname].names[generator] = gname
```
memsharded: ``names`` no longer the way to define these things, better user ``set_property()``.

The UI might need to evolve, like it would be better to return a ``NewCppInfo`` information instead, maybe? We can leave it from now, and see how it evolves.

Isn't it possible to get the ``include`` directories from the API? I would expect to obtain:

- Includedirs
- Libdirs
- libraries
- Component names

SSE4: > names no longer the way to define these things, better user set_property().

not sure I understand, `set_property`, do you mean CMake [thing](https://cmake.org/cmake/help/latest/command/set_property.html)? I don't see how can I set it from within `CppInfo`. do you have an example?

> The UI might need to evolve, like it would be better to return a NewCppInfo information instead, maybe? We can leave it from now, and see how it evolves.

from the functional perspective yes, this makes `CppInfo` an immutable object, and `package_info` will be pure function, which should result in several benefits

> Isn't it possible to get the include directories from the API?

unfortunately, I don't see anything related to the `includedirs` in the current response.
still, I suppose, having some information auto-generated is much better than having no information at all. author always can manually fill missing `CppInfo` fields, as you have stated previously.
for the most of cases, it won't be ever needed, as for standard layout `include`/`lib` directories are added automatically.

SSE4:
```suggestion
                conanfile.cpp_info.components[cname].set_property("cmake_target_name", gname, generator)
```

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -0,0 +1,146 @@
...
+                 os.path.join("decoder", "decoder.h"): decoder_h,
+                 os.path.join("encoder", "encoder.h"): encoder_h,
+                 os.path.join("transcoder", "transcoder.h"): transcoder_h,
+                 os.path.join("decoder", "CMakeLists.txt"): decoder_cmake,
+                 os.path.join("encoder", "CMakeLists.txt"): encoder_cmake,
+                 os.path.join("transcoder", "CMakeLists.txt"): transcoder_cmake,
+                 "CMakeLists.txt": common_cmake,
+                 })
+    client.run("create .")
+    print(client.out)
```
SSE4:
```suggestion
```

conan/tools/files/cpp_package.py
```diff
@@ -0,0 +1,74 @@
...
+        return conan_package
+
+    def save(self, filename=DEFAULT_FILENAME):
+        """
+        saves package model from memory into the cpp_package.json file
+        :param filename: path to the cpp_package.json
+        :return: None
+        """
+        text = json.dumps(self, default=lambda o: o.__dict__, indent=4)
+        print(text)
```
SSE4:
```suggestion
```

memsharded (approved):
Ok, lets merge it.
There are some minor changes to be done, but that can be done in another PR. I will do it, while checking the CppPackage model, which is the core part that would be good to evolve first.

Do not document it yet, some interfaces might have small changes, better wait.

PR #9009, opened 2021-05-26
CMakeToolchain: more blocks before try_compile

memsharded:
Changelog: Feature: simplified ``CMakeToolchain`` with only 1 category of blocks, made ``try-compile`` template code as another block, and reordered blocks so relevant flags for try-compile are taken into account.

#tags: slow


lasote (approved):
Solve conflict

PR #9012, opened 2021-05-27
trying to fix MSBuildDeps for packages with dot names

### LINKED ISSUE #9010: [question] MSBuildDeps - Cannot add property sheet due to circular dependency with property sheets
Trying to ascertain the suitability of the the MSBuildDeps generator over using CMake for legacy projects.  I tried the following steps - 

-Created recipe file with one build_requires and using the MSBuildDeps generator.  Build step is using the MSBuild helper passing an MSBuild .proj file.
-Run Conan install 
-Opened the VS solution and and tried to add the conandeps.props prop sheet to the project.  I get the following error - 

```
---------------------------
Microsoft Visual Studio
---------------------------
Cannot add property sheet 'C:\core\abc\conandeps.props' as it is either an invalid property sheet, or adding it would cause a circular inheritance with one of the selected property sheets.
---------------------------
OK   
---------------------------
```

Also tried adding the actual dep conan_xyz.props file, still get the same error.  This is with VS 2017.  Has this error been seen before?

I did get it working with the older visualstudio and visualstudio_mult generators (well sort of, the property sheets at least loaded, but I did have to tweak the includes outside of the property sheet...didn't look further as I believe MSBuildDeps is the way to go)

Thanks
Scott

memsharded:
Changelog: Bugfix: Allow ``MSBuildDeps`` to correctly process packages with dots in the package name.
Docs: Omit

Fix https://github.com/conan-io/conan/issues/9010

### DIFF
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -36,12 +36,12 @@ class MSBuildDeps(object):
     _conf_props = textwrap.dedent("""\
         <?xml version="1.0" encoding="utf-8"?>
         <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-          <ImportGroup Label="ConanDependencies">
+          <ImportGroup Label="PropertySheets">
             {% for dep in deps %}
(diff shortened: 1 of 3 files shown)


czoido (approved):

memsharded:
Last change in ec29acf allows property sheets to be correctly displayed in IDE, thanks @sd185233 !! image

PR #9017, opened 2021-05-27
Test default virtualenv Apply

lasote:
Changelog: omit
Docs: omit

### DIFF
--- a/conans/client/generators/__init__.py
+++ b/conans/client/generators/__init__.py
@@ -223,8 +223,7 @@ def write_toolchain(conanfile, path, output):
                 conanfile.generate()
 
     # tools.env.virtualenv:auto_use will be always True in Conan 2.0
-    auto_use = conanfile.conf["tools.env.virtualenv:auto_use"]
-    if (auto_use is None or auto_use) and conanfile.virtualenv:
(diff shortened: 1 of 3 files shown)


conans/client/generators/__init__.py
```diff
@@ -223,7 +223,8 @@ def write_toolchain(conanfile, path, output):
                 conanfile.generate()
 
     # tools.env.virtualenv:auto_use will be always True in Conan 2.0
-    if conanfile.conf["tools.env.virtualenv:auto_use"] and conanfile.virtualenv:
+    auto_use = conanfile.conf["tools.env.virtualenv:auto_use"]
```
memsharded: This will activate the new VirtualEnv, which might be breaking, when the user doesn't define anything at all at [conf].

lasote: 🤔 then who is activating it when I use the CMakeDeps + tool chain? Not very intuitive

memsharded: Current tests, either use explicitly ``generators = "VirtualEnv"``, or set up the ``auto_use`` conf. But it should be either opt-in, to not break.

memsharded (approved):

PR #9024, opened 2021-05-30
Feature/default skip rpaths

### LINKED ISSUE #9007: [feature] Change default CMakeToolchain to not remove Rpaths
Because that should be an opt-in and tools like cmake install already strip the rpaths while installaing.

Behavior in Cmake was likely implemented in Cmake 3.9, so we didn't take that into account while developing generators for older Cmake versions.

memsharded:
Changelog: Feature: Remove ``CMAKE_SKIP_RPATHS`` by default to True in ``CMakeToolchain``, it is not necessary by default, users can opt-in, and new test validates shared libs will work with ``VirtualEnv`` generator ``conanrunenv``.

#tags: slow

Close https://github.com/conan-io/conan/issues/9007

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -175,16 +175,18 @@ class GLibCXXBlock(Block):
 
 class SkipRPath(Block):
     template = textwrap.dedent("""
+        {% if skip_rpath %}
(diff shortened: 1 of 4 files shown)


memsharded:
@lasote, I have merged this to prepare 1.37 release for tomorrow, but please review anyway, just in case.

lasote (commented):
Looking good

PR #9030, opened 2021-05-31
User access to the new environment

lasote:
Changelog: Feature: Provide access in the recipes to the environment declared with the [new environment system](https://docs.conan.io/en/latest/reference/conanfile/tools/env.html). 
Docs: https://github.com/conan-io/docs/pull/2136

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -225,16 +225,33 @@ class Environment:
                     self._values[k] = new_value
         return self
 
-    def to_dict(self):
-        ret = OrderedDict()
-        for varname, varvalues in self._values.items():
-            value = self._format_value(varname, varvalues, "", os.pathsep)
-            ret[varname] = value
-        return ret
+    # Methods to user access to the environment object as a dict, replacing the placeholder with
+    # the current environment value
+    def _get_final_value(self, name):
+        if name not in self._values:
+            raise KeyError("No environment variable: " + name)
+        previous_value = os.getenv(name) or ""
+        return self._format_value(name, self._values[name], previous_value, os.pathsep)
+
+    def __getitem__(self, name):
+        return self._get_final_value(name)
+
+    def keys(self):
+        return self._values.keys()
+
+    def items(self):
+        for k in self._values.keys():
+            yield k, self._get_final_value(k)
+
+    def __eq__(self, other):
+        return other._values == self._values
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
 
     @contextmanager
     def apply(self):
-        apply_vars = self.to_dict()
+        apply_vars = self.items()
         old_env = dict(os.environ)
         os.environ.update(apply_vars)
         try:
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -3,6 +3,7 @@ import os
 import textwrap
 from collections import OrderedDict
 
+from conans.client.tools.env import _environment_add
 from conans.errors import ConanException
 from conans.util.files import save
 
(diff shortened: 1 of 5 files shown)


conan/tools/env/environment.py
```diff
@@ -224,6 +225,16 @@ def compose(self, other):
                     self._values[k] = new_value
         return self
 
+    def to_dict(self):
+        ret = OrderedDict()
+        for varname, varvalues in self._values.items():
+            value = self._format_value(varname, varvalues, "", os.pathsep)
```
memsharded: Maybe for the variable access we should provide access to the "structure"? Like if it is a list of items, it should be a list of items, not always a string that they will need to parse? 
I also understand the reason for always a string, to resemble the native env-vars, just a question.

conan/tools/env/environment.py
```diff
@@ -3,6 +3,7 @@
 import textwrap
 from collections import OrderedDict
 
+from conans.client.tools.env import _environment_add
```
memsharded: The goal is not to depend on anything of the previous ``conans.client.tools``, but to create new helpers here. The thing is that ``_environment_add()`` has all the legacy, like checking if something is a list and doing ``os.pathsep.join()`` logic directly there. 

Maybe not necessary to do as part of this PR if it works so far, but certainly a TODO

conan/tools/env/environment.py
lasote: I don't know if that would be common, probably you will just consume the variable with the computed value.

conans/test/integration/toolchains/env/test_virtualenv_object_access.py
```diff
@@ -0,0 +1,68 @@
...
+    import os
+    from conans import ConanFile
+    from conan.tools.env import VirtualEnv
+
+    class ConanFileToolsTest(ConanFile):
+        requires = "foo/1.0"
+
+        def build(self):
+          env = VirtualEnv(self)
+          build_env = env.build_environment().to_dict()
```
memsharded: Everything looks good, just a small thing, this ``to_dict()`` doesn't read great to me. Maybe is just the name? Or should we create a dict-like access to ``Environment`` class?

In any case, probably just personal opinion, so I am good with this.

conan/tools/env/environment.py
```diff
@@ -224,6 +225,24 @@ def compose(self, other):
                     self._values[k] = new_value
         return self
 
+    def to_dict(self):
+        ret = OrderedDict()
+        for varname, varvalues in self._values.items():
+            value = self._format_value(varname, varvalues, "", os.pathsep)
```
memsharded: Beware, this might have issues, you are removing placeholder for previous values of the env-var, like ``"{name}"`` default argument.  Every platform will need to define its one placeholder like ``%{name}%`` in Windows.

This might also do badly for subsystems in which the ``os.pathsep`` is not the system one (msys, cygwin, etc)

lasote: 🤔 I would need to think about it but at first sight, I'm not sure that a value with a placeholder in a python variable makes any sense for a user willing to have a value. That only makes sense in a shell script. right? should we emulate that replacement of the placeholder reading the actual `os.environ`?

memsharded: Yep, exactly, I guess this should be handled in-memory, resolving against the current environment.

conans/test/integration/toolchains/env/test_virtualenv_object_access.py
lasote: updated

conan/tools/env/environment.py
lasote: updated

memsharded (commented):
Looking much better

memsharded (approved):

PR #9032, opened 2021-05-31
Toolchain declare DIR find package variables

lasote:
Changelog: Feature: The `conan_toolchain.cmake` now includes `xxx_DIR` variables for the dependencies to ease the `find_package` mechanism to locate them. The declaration of these directories is a must when cross-building in OSX where CMake ignores `CMAKE_PREFIX_PATH` and `CMAKE_MODULE_PATH` to look only at the system framework directories.
Docs: https://github.com/conan-io/docs/pull/2108

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -9,7 +9,7 @@ from jinja2 import Template
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE
 from conan.tools._compilers import architecture_flag, use_win_mingw
-from conan.tools.cmake.utils import is_multi_configuration
+from conan.tools.cmake.utils import is_multi_configuration, get_file_name
 from conan.tools.microsoft.toolchain import write_conanvcvars, vs_ide_version
 from conans.errors import ConanException
 from conans.util.files import load, save
@@ -223,6 +223,20 @@ class CppStdBlock(Block):
             cppstd = compiler_cppstd
             cppstd_extensions = "OFF"
         return {"cppstd": cppstd, "cppstd_extensions": cppstd_extensions}
+
+
+class FindPackageDirVarsBlock(Block):
+    template = textwrap.dedent("""
+        # Help locating the find_package config files
+        {% for name in find_names %}
+        set({{ name }}_DIR "${CMAKE_BINARY_DIR}")
+        {% endfor %}
+        """)
+
+    def context(self):
+        host_req = self._conanfile.dependencies.transitive_host_requires
+        names = [get_file_name(req) for req in host_req]
+        return {"find_names": names}
 
 
 class SharedLibBock(Block):
@@ -618,7 +632,8 @@ class CMakeToolchain(object):
                                        ("try_compile", TryCompileBlock),
                                        ("find_paths", FindConfigFiles),
                                        ("rpath", SkipRPath),
-                                       ("shared", SharedLibBock)])
+                                       ("shared", SharedLibBock),
+                                       ("find_package_dir_vars", FindPackageDirVarsBlock)])
 
     def _context(self):
         """ Returns dict, the context for the template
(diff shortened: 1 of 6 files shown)


conan/tools/cmake/toolchain.py
```diff
@@ -225,6 +225,20 @@ def context(self):
         return {"cppstd": cppstd, "cppstd_extensions": cppstd_extensions}
 
 
+class FindPackageDirVarsBlock(Block):
+    template = textwrap.dedent("""
+        # Help locating the find_package config files
+        {% for name in find_names %}
+        set({{ name }}_DIR "${CMAKE_BINARY_DIR}")
```
memsharded: Why ``CMAKE_BINARY_DIR``? Shouldn't be the ``generators_folder``?

conan/tools/cmake/toolchain.py
```diff
@@ -618,7 +632,8 @@ def __init__(self, conanfile, generator=None):
                                        ("try_compile", TryCompileBlock),
                                        ("find_paths", FindConfigFiles),
                                        ("rpath", SkipRPath),
-                                       ("shared", SharedLibBock)])
+                                       ("shared", SharedLibBock),
+                                       ("find_package_dir_vars", FindPackageDirVarsBlock)])
```
memsharded: Maybe put it together with the ``find_paths`` one, as they are related.

conan/tools/cmake/utils.py
```diff
@@ -2,3 +2,12 @@ def is_multi_configuration(generator):
     if not generator:
         return False
     return "Visual" in generator or "Xcode" in generator or "Multi-Config" in generator
+
+
+def get_file_name(conanfile):
+    """Get the name of the file for the find_package(XXX)"""
+    ret = conanfile.new_cpp_info.get_property("cmake_file_name", "CMakeDeps")
```
memsharded, line 11: This is not very great now, that it depends explicitly on ``CMakeDeps``, while it has been made common and is being used not in CMakeDeps, but in CMakeToolchain. I understand why, but reads a bit confusing if you don't know it.

conan/tools/cmake/toolchain.py
```diff
@@ -225,6 +225,20 @@ def context(self):
...
+class FindPackageDirVarsBlock(Block):
+    template = textwrap.dedent("""
+        # Help locating the find_package config files
+        {% for name in find_names %}
+        set({{ name }}_DIR "${CMAKE_BINARY_DIR}")
+        {% endfor %}
+        """)
+
+    def context(self):
+        host_req = self._conanfile.dependencies.transitive_host_requires
```
memsharded: If only necessary for OSX, maybe it should be added by default only for OSX (or Apple) systems?

conan/tools/cmake/toolchain.py
lasote: Probably `${CMAKE_CURRENT_LIST_DIR}` to be consistent with `cmake_module_path`

conan/tools/cmake/utils.py
lasote: It is used in both of them. Maybe that property shouldn't be attached to the `CMakeDeps` generator by the recipe creator.

lasote: I added a comment. Suggestion welcome.

memsharded: I am not fully convinced on the proposition of both having the property name and the specific generator as a specialization. We might want to revisit that, but yes, I guess at the moment this is fine here.

conan/tools/cmake/toolchain.py
```diff
@@ -409,10 +401,17 @@ def context(self):
 
         os_ = self._conanfile.settings.get_safe("os")
         android_prefix = "${CMAKE_CURRENT_LIST_DIR}" if os_ == "Android" else None
+
+        host_req = self._conanfile.dependencies.transitive_host_requires
+        find_names_needed = os_ in ('iOS', "watchOS", "tvOS")
```
memsharded, line 406: This doesn't happen in OSX, right, only for these systems?

memsharded (approved):

Issue #8966, opened 2021-05-17
[bug] Alias resolution bug when mixing fix-version and alias


Doome161:
[conan_mix_fix_versions_with_alias.zip](https://github.com/conan-io/conan/files/6492511/conan_mix_fix_versions_with_alias.zip)

The bug occurs when using fixed versions and aliases in a dependency tree. The order of the requirements in the conanfiles plays a decisive role. If, for example, the requires are swapped in component CD, everything runs through successfully. It seems that this issue is related to #2576 which is already marked as fixed.

I have tested various conan versions and was able to reproduce the bug: 1.19, 1.22, 1.26, 1.35 and 1.36.

### Environment Details (include every applicable attribute)
  * Operating System+version: Linux
  * Compiler+version: gcc+7
  * Conan version: 1.35.0
  * Python version: 3.6.5

### Steps to reproduce (Include if Applicable)
To reproduce the bug I have attached a zipfile. Just run the build.sh and the bug will appear.
The script executes a `conan create . user/test` for each component and then creates an alias via `conan alias ....`

### Logs (Executed commands with output) (Include/Attach if Applicable)

```
WARN: CB/1.0.0@user/test: requirement CA/1.0.0@user/test overridden by CC/1.0.0@user/test to CA/RELEASE@user/test 
ERROR: CB/1.0.0@user/test: Incompatible requirements obtained in different evaluations of 'requirements'
    Previous requirements: [CA/1.0.0@user/test]
    New requirements: [CA/RELEASE@user/test]
```

monsdar:
We also get many of these warnings: `WARN: geotiff/1.6.0@thirdparty/release: requirement libtiff/4.1.0@thirdparty/release overridden by gdal/3.1.2 to libtiff/LATEST@thirdparty/release` where the `LATEST`-alias points to the exact same version that is required (`LATEST == 4.1.0`).

Seems that the timepoint where alias-resolution happens is not correct.

memsharded:
https://github.com/conan-io/conan/pull/9217 implemented new alias syntax that would resolve this issue, in 1.39.
Thanks for reporting this, let us know how it works!

PR #9039, opened 2021-06-01
Added integration test for mixing aliases and fix versions

Doome161:
Changelog: Omit
Docs: Omit

Added integration test for the bug described in #8966 

### DIFF
--- a/conans/test/integration/command/create_test.py
+++ b/conans/test/integration/command/create_test.py
@@ -540,3 +540,26 @@ class MyPkg(ConanFile):
         self.assertNotIn("requires", cpp_info_data["components"]["pkg1"])
         self.assertIn("libpkg2", cpp_info_data["components"]["pkg2"]["libs"])
         self.assertListEqual(["pkg1"], cpp_info_data["components"]["pkg2"]["requires"])
+
+    def test_mixing_aliases_and_fix_versions(self):
+        client = TestClient()
+
+        client.save({"conanfile.py": GenConanfile("ca", "1.0")})
+        client.run("create . user/testing")
+        client.run("alias ca/latest@user/testing ca/1.0@user/testing")
+
+        client.save({"conanfile.py": GenConanfile("cb", "1.0")
+                    .with_requirement("ca/1.0@user/testing")})
+        client.run("create . cb/1.0@user/testing")
+        client.run("alias cb/latest@user/testing cb/1.0@user/testing")
+
+        client.save({"conanfile.py": GenConanfile("cc", "1.0")
+                    .with_requirement("cb/latest@user/testing")
+                    .with_requirement("ca/latest@user/testing")})
+        client.run("create . user/testing")
+        client.run("alias cc/latest@user/testing cc/1.0@user/testing")
+
+        client.save({"conanfile.py": GenConanfile("cd", "1.0")
+                    .with_requirement("cb/latest@user/testing")
+                    .with_requirement("cc/latest@user/testing")})
+        client.run("create . user/testing", assert_error=False)


memsharded:
Hi @Doome161,

Thanks very much for doing this PR and reproduce this issue, really helpful.

I am investigating this, seems complex to fix, I am on it. But what I can suggest meanwhile is try to minimize the usage of the ``requirements()`` method and use plain ``requires`` class attribute whenever possible, this helps to minimize conflicts. Have you tried that?

In your example just changing ``client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requires("ca/1.0@user/testing")})`` will make the test pass.

memsharded:
I have been able to understand a bit what is happening. Indeed very complex to fix, at least in Conan 1.X, as the risk of breaking something will be probably too high.

The thing is that alias are way more complicated from the implementation point of view than it seems, and makes things challenging compared with other requirements. Even more than version_ranges, the main challenge is that when you do see a require, you don't know it is an alias at all, until you resolve the recipe, but that comes later. So you can well have 2 different requires, like ``pkga/somealias`` and ``pgka/otheralias``, that might conflict or not, but until both of them are resolved, it is impossible to know.

As I would like to improve the processing of alias for Conan 2.0, I am going to propose to make them explicit, same as version ranges. Lets say an alias could always be like ``pkga/(somealias)``, and in this way, it will be very evident that it is, understanding what is happening, also from the consumer point of view will be nicer, and it could greatly improve the internal processing that Conan needs to do, maybe even resulting in faster dependencies resolution. What do you think?

monsdar:
> As I would like to improve the processing of alias for Conan 2.0

I'd like to hear the opinions of the swarm in this regards. Having aliases explicit helps in a lot of ways, but there will be downfalls as well I guess.

> Indeed very complex to fix, at least in Conan 1.X, as the risk of breaking something will be probably too high.

Can we support this somehow? Perhaps by providing additional testcases? We use Aliases extensively and lately found that this issue in particular is breaking our local developer workflow. In most cases it's just warnings, but in others developer can't compile because Conan is not able to resolve the dependencies in the right order.

memsharded:
Hi @monsdar 

I have prepared a proposal for the Tribe, to gather the feedback.

> Can we support this somehow? Perhaps by providing additional testcases? We use Aliases extensively and lately found that this issue in particular is breaking our local developer workflow. In most cases it's just warnings, but in others developer can't compile because Conan is not able to resolve the dependencies in the right order.

I am not sure, I have tried a couple of things and failed to come up with something that wouldn't have risk. I also tried in ``develop2`` branch, where I have more freedom to not break things, and even there I struggled to come up with something that made sense. Then I experimented a bit with the explicit approach, and at the moment it seems the only viable thing.

I will give it another try and let you know. In the worst case, I think we can look for other workarounds, many times these alias problems are solved just by changing order of declarations of ``requires``, or putting things in ``requires`` attribute instead of a method.

monsdar:
It looks like there is nobody arguing against having explicit aliases in the Tribe request. Given that there is nothing coming up in the next days could we implement explicit aliases in Conan 1.x and solve the above issue with these? Then we could already migrate to explicit aliases in our codebase and have our problems resolved.

In Conan 1.x this could be an extension to the existing alias implementation. With Conan 2.x the non-explicit alias implementation would then be removed.

memsharded:
> Given that there is nothing coming up in the next days could we implement explicit aliases in Conan 1.x and solve the above issue with these? Then we could already migrate to explicit aliases in our codebase and have our problems resolved.

As much as I would like that, it would be breaking behavior, and I don't think it is possible in Conan 1.X because it will break other users. We will add the syntax to help the transition to Conan 2.0, but we cannot force it and make compulsory or fail because it will be breaking.

I have still not given up, and I plan to try a couple of things more. But in any case, I would like to suggest to explore some other possible workarounds, there might be something doable that won't require a lot of effort on your side. 
This is the original failing test:

```python
def test_mixing_aliases_and_fix_versions():
    # cd ----------------------------------> cb/latest -(alias) -> cb/1.0 -> ca/1.0
    #  \--> cc/latest -(alias) -> cc/1.0  ----/                               /
    #                               \   ---> ca/latest -(alias) -------------/
    client = TestClient()

    client.save({"conanfile.py": GenConanfile("ca", "1.0")})
    client.run("create . ")
    client.run("alias ca/latest@ ca/1.0@")

    client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requirement("ca/1.0@")})
    client.run("create .")
    client.run("alias cb/latest@ cb/1.0@")

    client.save({"conanfile.py": GenConanfile("cc", "1.0")
                .with_requirement("cb/latest@")
                .with_requirement("ca/latest@")})
    client.run("create . ")
    client.run("alias cc/latest@ cc/1.0@")

    client.save({"conanfile.py": GenConanfile("cd", "1.0")
                .with_requirement("cb/latest@")
                .with_requirement("cc/latest@")})
    client.run("create . ")
```

Changing the ``requirement()`` method for a plain ``requires = "xxxx"`` definition will make the test pass:

```diff
@@ -502,7 +502,7 @@ def test_mixing_aliases_and_fix_versions():
     client.run("create . ")
     client.run("alias ca/latest@ ca/1.0@")

-    client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requirement("ca/1.0@")})
+    client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requires("ca/1.0@")})
     client.run("create .")
     client.run("alias cb/latest@ cb/1.0@")
```

Making the dependency from cb to ca use the alias too, will make the test pass (what is problematic is to merge alias with explicit requires in the same diamond of the graph):
```diff
@@ -502,7 +502,7 @@ def test_mixing_aliases_and_fix_versions():
     client.run("create . ")
     client.run("alias ca/latest@ ca/1.0@")

-    client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requirement("ca/1.0@")})
+    client.save({"conanfile.py": GenConanfile("cb", "1.0").with_requirement("ca/latest@")})
     client.run("create .")
     client.run("alias cb/latest@ cb/1.0@")
```

Changing the order of the downstream declaration, will also make the test pass (the order of the declaration matters, if the alias is resolved first or not):
```diff
@@ -513,6 +513,6 @@ def test_mixing_aliases_and_fix_versions():
     client.run("alias cc/latest@ cc/1.0@")

     client.save({"conanfile.py": GenConanfile("cd", "1.0")
-                .with_requirement("cb/latest@")
-                .with_requirement("cc/latest@")})
+                .with_requirement("cc/latest@")
+                .with_requirement("cb/latest@")})
     client.run("create . ")
```

Maybe some of these alternatives is a long hanging fruit that would solve your issue?

monsdar:
My idea would be to add the explicit format in addition to the current non-explicit way. That should be non-breaking, but for the explicit aliases one could already implement better resolution mechanisms. The following could be legal in Conan 1.3x:

```
def requirements( self):
        self.requires("MyCompA/1.0.0@core/release")
        self.requires("MyCompB/[~=1.0]@core/release")
        self.requires("MyCompC/MyAlias@core/release")
        self.requires("MyCompD/(MyAlias)@core/release")
```

This could work in Conan 1.x, where in Conan 2.0 then the declaration for MyCompC would be invalid.

monsdar:
If that is not an option we'll investigate how to automatically find a valid order for the requirements. Our dependency tree is too complex to find this manually, but perhaps (with some magic) we can provide a tool that outputs the optimal order for any developer encountering this issue.

memsharded:
Hi @Doome161 @monsdar 

I am trying in https://github.com/conan-io/conan/pull/9217 to implement this feature. It might be possible to consider it for Conan 1.X as the new syntax ``pkg/(alias)`` is an opt-in apparently at all levels, so it wouldn't be breaking. 

I have added already a few tests, so far it seems fine, but if you want to experiment with this feature, that would be extremely helpful, you might run Conan from source from my branch. Thanks!

Doome161:
Hi @memsharded ,

sounds great! I will experiment around with the feature in the next few days and will share my experience here

Doome161:
Hi @memsharded ,

I have been experimenting with the explicit aliases now and it works great! I was able to build huge trees with mixed aliases and fixed versions without problems.

I do have one small comment though:

We also use aliases for python_requires. But there you get an error if you use the explicit aliase syntax; 
```shell
Unable to find python_requires("comp/(LATEST)@user/channel") in remotes
```

Beside that, I would be very happy if this feature would be included in one of the next conan releases. It would help our developers a lot. Our current workaround for this bug is very dirty and often leads to confusion and a lot more work.

memsharded:
@Doome161 

Added in the PR support for python-requires alias, if you want to try that too. Thanks!

memsharded:
@Doome161 did you manage to try the ``python_requires`` support in my PR?
If we aim to release this in Conan 1.39, ti would be great to have feedback by this week, as this will require some internal discussion about the architecture and risks.

Doome161:
Hi @memsharded ,

sorry for the late feedback. I tried it on my system and I still get the error shown above

memsharded:
Which ``python_requires`` are you using? The class attribute or the global function?

Doome161:
The global function

memsharded:
> The global function

That explains it. It has only be implemented in the class attribute, it superseded the global function long time ago. Only the class attribute will be maintained in Conan 2.0: https://docs.conan.io/en/latest/extending/python_requires_legacy.html

It is unlikely that we can port this thing to the global function, it is being removed because it always made things extra-complicated, because of the inversion of control (the python requires are resolved directly by the python interpreter, not Conan).

Doome161:
Ok, I think we can live with that 👍

memsharded:
Hi all!

We have merged https://github.com/conan-io/conan/pull/9217, to be released in Conan 1.39.
This should be enough for this case, so we can close it, but please create a new issue or PR to report any problem with the new syntax. We hope this enable a smooth transition to Conan 2.0, and also allows to solve all the alias related issues earlier, from 1.39.

Many thanks!

PR #9043, opened 2021-06-02
Fix self.folders.source

lasote:
Changelog:  Fix: When using the new `self.folders.source` (at `layout(self)` method) the sources (from `export`, `export_sources` and `scm`) are copied to the base source folder and not to the `self.folders.source` that is intended to describe where the sources are after fetching them.
Docs: https://github.com/conan-io/docs/pull/2117

### DIFF
--- a/conans/test/functional/layout/test_source_folder.py
+++ b/conans/test/functional/layout/test_source_folder.py
@@ -1,3 +1,6 @@
+import os
+import platform
+
 import pytest
 
 from conans.test.assets.cmake import gen_cmakelists
@@ -5,6 +8,8 @@ from conans.test.assets.genconanfile import GenConanfile
 from conans.test.assets.sources import gen_function_cpp
 from conans.test.utils.scm import create_local_git_repo
 from conans.test.utils.tools import TestClient
+
+app_name = "my_app.exe" if platform.system() == "Windows" else "my_app"
 
 
 @pytest.mark.parametrize("no_copy_source", ["False", "True"])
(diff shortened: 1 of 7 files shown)


conans/test/functional/layout/test_source_folder.py
```diff
@@ -0,0 +1,131 @@
...
+
+from conans.test.assets.cmake import gen_cmakelists
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.utils.scm import create_local_git_repo
+from conans.test.utils.tools import TestClient
+
+
+@pytest.mark.parametrize("no_copy_source", ["False", "True"])
+def test_exports_source_with_src_subfolder(no_copy_source):
```
memsharded, line 16: It would be very relevant to learn how this behaves with ``editable`` mode, maybe another test?. I would say there could be issues, as the ``export`` process do change layouts?

memsharded (commented):
Looks good, please check the failing test, and it would be nice to see how this behaves for editable mode.

conans/test/functional/layout/test_source_folder.py
lasote: I would say at `test_editables_layout.py` more than one test is using `self.folders.source` and editable packages. 
The idea still works because the "root" of the editable package is the current project directory, so the `self.folders.source="my_src"` is correct as it is with conan local commands.

memsharded (approved):

PR #9045, opened 2021-06-02
Skip when no arch is present

lasote:
Changelog: Bugfix: The `CMakeToolchain` generator now manages correctly a recipe without `arch` declared in an Apple system.
Docs: omit

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -344,6 +344,8 @@ class AppleSystemBlock(Block):
     def context(self):
         os_ = self._conanfile.settings.get_safe("os")
         host_architecture = self._get_architecture()
+        if host_architecture is None:
+            return
         # We could be cross building from Macos x64 to Macos armv8 (m1)
         build_architecture = self._get_architecture(host_context=False)
         if os_ not in ('iOS', "watchOS", "tvOS") and \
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/toolchain.py
```diff
@@ -344,6 +344,8 @@ def _apple_sdk_name(self):
     def context(self):
         os_ = self._conanfile.settings.get_safe("os")
         host_architecture = self._get_architecture()
+        if host_architecture is None:
```
memsharded: Shouldn't at least the CMAKE_SYSTEM_NAME, CMAKE_SYSTEM_VERSION for iOS, tvOS, watchOS be defined? Is it possible to have those target systems without arch?

lasote: we can try...

lasote: Yes, it is possible. It allowed to simplify the code, please check again.

memsharded (approved):

conans/client/migrations_settings.py
```diff
@@ -2302,3 +2302,5 @@
 
 cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20, 23, gnu23]  # Deprecated, use compiler.cppstd
 """
+settings_1_37_1 = settings_1_37_0
+settings_1_38_0 = settings_1_37_1
```
czoido:
```suggestion
```
1.38 should not appear in this branch

PR #9046, opened 2021-06-02
if the build_require not activated shouldn't fail

memsharded:
Changelog: Fix: Do not fail in ``CMakeDeps`` when there are ``build_requires`` with same name as host requires, unless ``build_context_activated`` is enabled for those and a different suffix has not been defined.
Docs: Omit

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -41,13 +41,16 @@ class CMakeDeps(object):
         build_req = self._conanfile.dependencies.build_requires_build_context
 
         # Check if the same package is at host and build and the same time
-        common = {r.ref.name for r in host_req}.intersection({r.ref.name for r in build_req})
-        for name in common:
-            if name not in self.build_context_suffix:
-                raise ConanException("The package '{}' exists both as 'require' and as "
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -41,13 +41,16 @@ def content(self):
         build_req = self._conanfile.dependencies.build_requires_build_context
 
         # Check if the same package is at host and build and the same time
-        common = {r.ref.name for r in host_req}.intersection({r.ref.name for r in build_req})
```
lasote, line 44: I think it would be more readable and short just filtering the activated, something like:

```
activated_br = {r.ref.name for r in build_req if r.ref.name in self.build_context_activated}
common_names = {r.ref.name for r in host_req}.intersection(activated_br)
```

memsharded: Changed, please review.

In any case, the important feedback here is if this makes sense as default behavior, or I am missing some other thing.

lasote (approved):

conan/tools/cmake/cmakedeps/cmakedeps.py
lasote: This check? Yes, it was missing and makes perfect sense.

PR #9056, opened 2021-06-03
fixing preprocessor_definitions in MSBuildToolchain

memsharded:
Changelog: BugFix: Do not quote all values and allow integer and macro referencing in ``MSBuildToolchain.preprocessor_definitions``
Docs: Omit

This might be breaking for other users, but better fix it soon or it will be impossible to fix later.

### DIFF
--- a/conan/tools/microsoft/toolchain.py
+++ b/conan/tools/microsoft/toolchain.py
@@ -135,7 +135,7 @@ class MSBuildToolchain(object):
     def _write_config_toolchain(self, config_filename):
 
         def format_macro(key, value):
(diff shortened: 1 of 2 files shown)


lasote (commented):
I cannot find the `. preprocessor_definitions` at the docs,  it is not documented right?. Probably we should do it and put some examples of quoted values clarifying the behavior.

memsharded:
Yes, it was not documented on purpose, this user reporting it see it seeing the source code, so not a big issue that it is "breaking".

Maybe it is time to document it, but I would like to have a discussion of what can be parameterized via the toolchains, besides the preprocessor_definitions.

PR #9057, opened 2021-06-04
Test to show issue with editable and cmake layout()

memsharded:
Changelog: Feature: New ``cmake_layout()`` layout helper to define a multi-platform CMake layout that will work for different generators (ninja, xcode, visual, unix), and is multi-config.
Docs: https://github.com/conan-io/docs/pull/2141

This PR is highlighting one issue regarding the layout() feature: We need information about the CMake generator to know about the layout, as it changes depending on it. It is not good in Visual Studio to go with build/release or build-release and build/debug or build-debug folders, because you get 2 different independent VS projects there, which you cannot switch from the IDE (they are 2 single config projects), and I think this defeats the purpose of editable packages.

I think the problem is solvable, by extracting the information of the generator from the toolchain to a common, reusable place. Maybe it is possible to delay the evaluation of ``layout()`` too, that might help in the extraction and reusage of CMake generator info.
 
#tags: slow


lasote:
Ok, I'm working on it, I'll open a PR for your branch.

lasote (commented):
Looking good, I submited a PR here in case you want to support more generators, https://github.com/memsharded/conan/pull/39

memsharded (commented):
The XCode generator is failing, CMAKE_CXX_COMPILER  not defined. It might be necessary to add them to the toolchain or something (an unrelated bug to this PR, maybe).

I removed Ninja multi-config in OSX, it is not installed in CI (or cmake support missing), we can wait to add it too.

The MinGW generator is working for me locally, but fails in CI, not sure why

lasote:
> The XCode generator is failing, CMAKE_CXX_COMPILER not defined. 

Oh, that is because the `3.19` CMake version is required to work with the newest XCodes. 
Other tests are marked as `@pytest.mark.tool_cmake(version="3.19")` but it will affect all the parametrized.

> I removed Ninja multi-config in OSX, it is not installed in CI (or cmake support missing), we can wait to add it too.

Good 

> The MinGW generator is working for me locally, but fails in CI, not sure why

Check with @czoido , other mingw test are decorated with:

```
@pytest.mark.tool_mingw64
@pytest.mark.tool_cmake(version="3.16")
@pytest.mark.skipif(platform.system() != "Windows", reason="Mingw test")`
```

lasote (approved):

PR #9062, opened 2021-06-06
Feature/new conanfile dependencies

memsharded:
Changelog: Feature: New ``conanfile.dependencies`` model, using a dict {requirement: ConanFileInterface} to prepare for Conan 2.0.
Docs: https://github.com/conan-io/docs/pull/2134

#tags: slow


conans/test/integration/configuration/conf/test_conf_from_br.py
```diff
@@ -34,8 +34,8 @@ def generate(self):
         """)
     client.save({"conanfile.py": consumer,
                  "android": android_profile}, clean_first=True)
-    client.run("install .")
+    client.run("install . -pr:b=default")
```
lasote, line 37: Do we need this change? Might be breaking?

conans/client/generators/__init__.py
```diff
@@ -157,6 +156,7 @@ def write_generators(self, conanfile, old_gen_folder, new_gen_folder, output):
                         generator.generate()
                     continue
                 except Exception as e:
+                    print(traceback.format_exc())
```
lasote: traceback

lasote (commented):
Looking good! I don't see any reason to stop this moving forward.

conans/test/integration/configuration/conf/test_conf_from_br.py
memsharded: Actually, this was kind of a bug or error. It allowed to bring [conf] to consumers from build requires in the host context (i.e. test requires like gtest), and at least the initial intention was to limit this to build requires in the build context, only for tools like AndroidNDK and similar build tools.

conans/client/generators/__init__.py
memsharded: Removed. This is something to fix elsewhere, when a generator crash, it won't show a useful traceback to diagnose the crash, I think we should improve this in another PR.

lasote: Totally agree

PR #9073, opened 2021-06-08
Fix 9071

### LINKED ISSUE #9071: [bug] CMakeDeps generator: component and package with the same name
Hi!

During refactoring #8971 behavior of `tvalue` [macro](https://github.com/conan-io/conan/blob/80e99ca9a7d730e95adbc8bde5aa33da4f4fbb44/conan/tools/cmake/cmakedeps/templates/target_configuration.py#L35) was changed by adding some special case when `pkg_name` and `comp_name` are the same. 

It accidentally break this case, because `tvalue` macro is inlined in declaration of several variables (see [this](https://github.com/conan-io/conan/blob/80e99ca9a7d730e95adbc8bde5aa33da4f4fbb44/conan/tools/cmake/cmakedeps/templates/target_configuration.py#L98) for example). So these variables have one name (`{pkg_name}_{comp_name}_...`) during setting and the other name (`{comp_name}_...`) during using it in `set_property`.

cc @lasote

lasote:
Changelog: Bugfix: The `CMakeDeps` generator variables were named wrongly when a component had the same name as the package.
Docs: omit

Closes #9071

### DIFF
--- a/conans/test/unittests/tools/cmake/test_cmakedeps.py
+++ b/conans/test/unittests/tools/cmake/test_cmakedeps.py
@@ -143,3 +143,52 @@ def test_cmake_deps_links_flags():
             data_cmake = files["mypkg-release-x86-data.cmake"]
             assert "set(mypkg_SHARED_LINK_FLAGS_RELEASE -NODEFAULTLIB;-OTHERFLAG)" in data_cmake
             assert "set(mypkg_EXE_LINK_FLAGS_RELEASE -OPT:NOICF)" in data_cmake
+
+
+def test_component_name_same_package():
+    """
+    When the package and the component are the same the variables declared in data and linked
+    to the target have to be the same.
+    https://github.com/conan-io/conan/issues/9071"""
+    conanfile = ConanFile(Mock(), None)
+    conanfile._conan_node = Mock()
+    conanfile._conan_node.context = "host"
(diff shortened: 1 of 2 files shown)


conans/test/unittests/tools/cmake/test_cmakedeps.py
```diff
@@ -143,3 +143,52 @@ def test_cmake_deps_links_flags():
...
+
+    # We adjust the component with the same name as the package on purpose
+    cpp_info.components["mypkg"].includedirs = ["includedirs1"]
+
+    conanfile_dep = ConanFile(Mock(), None)
+    conanfile_dep.cpp_info = cpp_info
+    conanfile_dep._conan_node = Mock()
+    conanfile_dep._conan_node.context = "host"
+
+    with mock.patch('conans.ConanFile.ref', new_callable=mock.PropertyMock) as mock_ref:
```
memsharded, line 174: This patch can be avoided, assigning the ref above directly (same as done with "host" context, like in https://github.com/conan-io/conan/pull/9062/files, this allows to simplify the test a bit.

Not a big deal, can be merged, in any case it will conflict with https://github.com/conan-io/conan/pull/9062, so I can update that PR later.

memsharded (approved):

memsharded (approved):

PR #9087, opened 2021-06-10
"external_package" property for CMakeDeps

### LINKED ISSUE #8919: [question] using `boost/system` recipe with system-wide config

Hi! My question is about `conan` recipes for system packages (like [gtk/system](https://conan.io/center/gtk) or [xorg/system](https://conan.io/center/xorg) from CCI). In our ecosystem we want to use `boost/system` (custom one) together with `boost/1.7?` from CCI (we need a transparent switching between them in the root of dependency tree). We use `CMakeDeps` generator. So before build it generates (among other things) `Boost-config.cmake`, which is very hard to reproduce in `boost/system` . 

- Before `1.36` we use the following smart crutch which exploits a bug (?) in generator. Our recipe declare package `boost/system` with filename (in latest terms `cmake_file_name`) `SystemBoost` and with name (in latest terms `cmake_target_name`) `Boost`. When somebody (e.g. `A`) wants to depend on boost, `CMakeDeps` generates **unused** `SystemBoost-config.cmake` and call `find_package(Boost ... ) ` in `A-config.cmake`. This `find_package` uses system-wide `FindBoost.cmake` or config from `boost-cmake` project and works perfectly. 
- In `1.36` this bug was fixed. ([the method](https://github.com/conan-io/conan/blob/develop/conan/tools/cmake/cmakedeps.py#L698) was introduced). So the crutch is not working now.
- Could you elaborate this situation?

There are my thoughts about it:
- I like the idea to use system-wide config in the specific case of system boost.
- Generating conan's `Boost-config.cmake` to emulate a system one seems very hard. In CCI the only way to do this is using `PkgConfig`, but Boost has no official `.pc` file AFAIK.
- Maybe `CMakeDeps` can take a system-wide boost config somehow (copying or symlinking it instead of generate its own) in this specific case?
- Maybe conan can legitimate the crutch described above? 

Thanks in advance!

lasote:
Changelog: Feature: Support for `CMakeDeps` generator of a new property `"skip_deps_file"` to be declared in the `cpp_info` of a package to skip creating `xxx-config.cmake` files for it, allowing to create "system wrapper" recipes easily. 
Docs: https://github.com/conan-io/docs/pull/2121

Closes #8919

### DIFF
--- a/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
+++ b/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
@@ -15,7 +15,7 @@ def test_package_from_system():
                .with_settings("os", "arch", "build_type", "compiler"))
     dep2 += """
     def package_info(self):
-        self.cpp_info.set_property("external_package", True)
+        self.cpp_info.set_property("skip_deps_file", True)
         self.cpp_info.set_property("cmake_file_name", "custom_dep2")
 
     """
--- a/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
+++ b/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
@@ -1,0 +1,39 @@
+import os
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+def test_package_from_system():
+    """
+    If a node declares "system_package" property, the CMakeDeps generator will skip generating
+    the -config.cmake and the other files for that node but will keep the "find_dependency" for
+    the nodes depending on it. That will cause that cmake looks for the config files elsewhere
+    https://github.com/conan-io/conan/issues/8919"""
+    client = TestClient()
+    dep2 = str(GenConanfile().with_name("dep2").with_version("1.0")
+               .with_settings("os", "arch", "build_type", "compiler"))
+    dep2 += """
+    def package_info(self):
+        self.cpp_info.set_property("external_package", True)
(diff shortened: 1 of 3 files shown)


danimtb (approved):
Simple and easy! Looking good to me

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -0,0 +1,39 @@
...
+    If a node declares "system_package" property, the CMakeDeps generator will skip generating
+    the -config.cmake and the other files for that node but will keep the "find_dependency" for
+    the nodes depending on it. That will cause that cmake looks for the config files elsewhere
+    https://github.com/conan-io/conan/issues/8919"""
+    client = TestClient()
+    dep2 = str(GenConanfile().with_name("dep2").with_version("1.0")
+               .with_settings("os", "arch", "build_type", "compiler"))
+    dep2 += """
+    def package_info(self):
+        self.cpp_info.set_property("external_package", True)
```
memsharded: I'd probably try to make it more explicit like ``skip_deps_file``, or ``external_build_script``, or  something like that, because "external package" could be a system package, but you still want the xxx-config.cmake or equivalent created to work.

But I guess mostly a personal taste and naming opinion, so I am fine with it, you can merge it as-is.

memsharded (approved):

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
lasote: I was waiting for your name suggestion. `skip_deps_file` will be

PR #9092, opened 2021-06-11
minor refactor of names

memsharded:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -37,8 +37,8 @@ class CMakeDeps(object):
         macros = MacrosTemplate()
         ret = {macros.filename: macros.render()}
 
-        host_req = self._conanfile.dependencies.host_requires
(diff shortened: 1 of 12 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -37,8 +37,8 @@ def content(self):
         macros = MacrosTemplate()
         ret = {macros.filename: macros.render()}
 
-        host_req = self._conanfile.dependencies.host_requires
-        build_req = self._conanfile.dependencies.direct_build_requires
+        host_req = self._conanfile.dependencies.host
```
memsharded, line 40: I thought while testing something as user, that both dependencies + requires in the user code was reading weird better:

```python
def xxx():
      self.dependencies.host["zlib"]
      # than
      self.dependencies.host_requires["zlib"]
```

As it seemed that the expression would be returning a ``require`` (Requirement), instead of a ``dependency`` (ConanFileInterface)

PR #9095, opened 2021-06-13
simple poc of template profiles

memsharded:

Very simple POC of using templates for profiles for advanced functionality. Still to test how composition could be improved, maybe also variable management.

### DIFF
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -1,5 +1,8 @@
 import os
+import platform
 from collections import OrderedDict, defaultdict
+
+from jinja2 import Template
 
 from conan.tools.env.environment import ProfileEnvironment
 from conans.errors import ConanException, ConanV2Exception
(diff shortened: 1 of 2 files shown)


czoido (approved):
I think this is a great idea and could be a very powerful tool to use in profiles.

conans/client/profile_loader.py
```diff
@@ -117,6 +120,11 @@ def read_profile(profile_name, cwd, default_folder):
     logger.debug("PROFILE LOAD: %s" % profile_path)
     text = load(profile_path)
 
+    if profile_name.endswith(".tpl"):
```
lasote, line 123: Why that extension? Might be `.jinja`?
https://jinja.palletsprojects.com/en/3.0.x/templates/#template-file-extension

memsharded: Yes, probably better to use ``jinja``.
But this was only an implementation trick, maybe it is not necessary to use the extension, other alternatives:

- Render always as a template. Jinja syntax is not valid at the moment, so sounds it won't be breaking.
- Use other argument like ``--profile-template=myprofile``
- Use a header first line in profile files like ``# jinja2 template`` (this could be good to later model other templating mechanisms or parameters of the rendering)

lasote: > Render always as a template. Jinja syntax is not valid at the moment, so sounds it won't be breaking.

It shouldn't but...might it be some profile with `{%` or `{{` ?

Probably the simpler to implement and maintain and more user-friendly was the extension.

memsharded:

PR #9096, opened 2021-06-13
refactor of profile loading

memsharded:
Changelog: Omit
Docs: Omit

Avoid ugly import of ``_load_profile()`` that does recursive evaluation of profiles, when it is not necessary. Minor refactors to group the parsing functionality.

The new class might be joined with ``ProfileParser`` but I didn't want to change too much atm.


lasote (approved):
Difficult to read if there could be breaking changes but go ahead.

memsharded:
> Difficult to read if there could be breaking changes but go ahead.

It can wait, lets wait until we see about the template jinja2 profiles, and I might re-target this to Conan 2.0 if not necessary to process the templates.

PR #9099, opened 2021-06-14
[feature] Add support for custom server path via env or cli

### LINKED ISSUE #9098: [feature] Support for custom directory for conan_server
Currently conan_server only supports being deployed to a default location - `~/.conan_server/`. My suggestion is to let the user set a custom directory by adding command-line argument that will let the user set where the server config and data will be stored. As far as I see from the code it should be something that is rather easy to do, and I would like to open a PR to add the functionality.

jenia90:
Changelog: Feature: Added support for using server config from a custom location, setting `CONAN_SERVER_HOME` env variable or using `-d` or `--server_dir` flag when launching the server with `conan_server` command. 
Docs: https://github.com/conan-io/docs/pull/2125

When launching from the command line the new CLI flag will be prioritized over the environment variable,  this is because we want to let the user test the server with custom dirs even if an env var has been set.
I've also added some tests to test the new feature (even though currently the tests are pretty basic).

Close  #9098. 

Please let me know if there are any improvements to be made :)

### DIFF
--- a/conans/conan_server.py
+++ b/conans/conan_server.py
@@ -1,14 +1,19 @@
 import argparse
+from conans import server
 
 from conans.server.launcher import ServerLauncher
+from conans.util.env_reader import get_env
 
 
 def run():
     parser = argparse.ArgumentParser(description='Launch the server')
     parser.add_argument('--migrate', default=False, action='store_true',
                         help='Run the pending migrations')
+    parser.add_argument('--server_dir', '-d', default=None,
+                        help='Specify where to store server config and data.')
     args = parser.parse_args()
-    launcher = ServerLauncher(force_migration=args.migrate)
+    launcher = ServerLauncher(force_migration=args.migrate,
+                              server_dir=args.server_dir or get_env("CONAN_SERVER_HOME"))
     launcher.launch()
 
 
--- a/conans/conan_server.py
+++ b/conans/conan_server.py
@@ -1,5 +1,4 @@
 import argparse
-from conans import server
 
 from conans.server.launcher import ServerLauncher
 from conans.util.env_reader import get_env
--- a/conans/server/launcher.py
+++ b/conans/server/launcher.py
@@ -18,14 +18,13 @@ class ServerLauncher(object):
     def __init__(self, force_migration=False, server_dir=None):
         self.force_migration = force_migration
         if server_dir:
-            server_folder = server_dir
+            user_folder = server_folder = server_dir
         else:
             user_folder = conan_expand_user("~")
             server_folder = os.path.join(user_folder, '.conan_server')
 
-        is_custom_path = server_dir is not None
         server_config = migrate_and_get_server_config(
-            user_folder, is_custom_path, self.force_migration
+            user_folder, self.force_migration, server_dir is not None
         )
         custom_auth = server_config.custom_authenticator
         if custom_auth:
--- a/conans/server/launcher.py
+++ b/conans/server/launcher.py
@@ -15,12 +15,18 @@ from conans.server.service.authorize import BasicAuthorizer, BasicAuthenticator
 
 
 class ServerLauncher(object):
-    def __init__(self, force_migration=False):
+    def __init__(self, force_migration=False, server_dir=None):
         self.force_migration = force_migration
-        user_folder = conan_expand_user("~")
-        server_folder = os.path.join(user_folder, '.conan_server')
+        if server_dir:
+            server_folder = server_dir
+        else:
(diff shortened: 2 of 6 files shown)


conans/conan_server.py
```diff
@@ -1,14 +1,19 @@
 import argparse
+from conans import server
```
memsharded: Unused import?

conans/server/server_launcher.py
```diff
@@ -1,6 +1,8 @@
 from conans.server.launcher import ServerLauncher
 
-launcher = ServerLauncher()
+from conans.util.env_reader import get_env
```
memsharded, line 3: Is this ``server_launcher.py`` file needed at all? @lasote?

memsharded (approved):
Looks good!

conans/conan_server.py
jenia90: Yup, creeped in by accident...

conans/server/server_launcher.py
lasote: As far as I remember is a convenience module to launch a WSGI server like gunicorn:

`gunicorn -b 0.0.0.0:9300 -w 4 -t 300 conans.server.server_launcher:app`

conans/server/launcher.py
```diff
@@ -15,12 +15,18 @@
...
+        if server_dir:
+            server_folder = server_dir
+        else:
+            user_folder = conan_expand_user("~")
+            server_folder = os.path.join(user_folder, '.conan_server')
 
-        server_config = migrate_and_get_server_config(user_folder, self.force_migration)
+        is_custom_path = server_dir is not None
+        server_config = migrate_and_get_server_config(
+            user_folder, is_custom_path, self.force_migration
```
czoido: This is calling the arguments in the wrong order

czoido: `user_folder` will be referenced before assignment here and fail

czoido (changes requested):

conans/server/launcher.py
jenia90: Yikes, sorry, fixing it

conans/server/migrate.py
```diff
@@ -5,8 +5,8 @@
 from conans.util.log import logger
 
 
-def migrate_and_get_server_config(base_folder, force_migration=False):
-    server_config = ConanServerConfigParser(base_folder)
+def migrate_and_get_server_config(base_folder, force_migration=False, is_custom_path=False):
+    server_config = ConanServerConfigParser(base_folder, is_custom_path)
```
czoido: This is going to fail because of the positional argument not being the path

czoido:
I think I would prefer to have the server folder stored under the `serve_folder/.conan_server` insted of directly in `server_dir` to be more symmetric with the conan client, but that's just a personal opinion... so I would be ok with that.

jenia90:
Hi @czoido 
Thanks for the feedback, I'm resolving the issues that came up, so hopefully I will have a fix in a bit. 
Regarding the folder location, initially I considered placing `.conan_server` folder inside the custom path, but then I thought that it would give the user more flexibility when choosing where to store their configs and data. If you or your team think it's better to place it under `.conan_server` inside the custom path, I'll happily make the relevant changes :)

memsharded:
> I think I would prefer to have the server folder stored under the serve_folder/.conan_server insted of directly in server_dir to be more symmetric with the conan client, but that's just a personal opinion... so I would be ok with that.

Actually, this is something that will change in Conan 2.0, and the client side will be able to define directly the cache folder, without the extra ``.conan`` folder. So I am fine with it the way it is.

memsharded (approved):

czoido (approved):

memsharded:
Thanks for contributing this. If possible, it would be nice to do a PR to the docs repo to update the docs with this new feature. I think adding a couple of lines will be good enough. Then, update the ``TODO`` in the first  comment, we automatically generate the changelog from there.

jenia90:
> 
> 
> Thanks for contributing this. If possible, it would be nice to do a PR to the docs repo to update the docs with this new feature. I think adding a couple of lines will be good enough. Then, update the `TODO` in the first comment, we automatically generate the changelog from there.

@memsharded, thank you for accepting my PR! To update the docs I need to create a new issue on the docs repo and then submit a PR, correct?

memsharded:
> @memsharded, thank you for accepting my PR! To update the docs I need to create a new issue on the docs repo and then submit a PR, correct?

No need for the issue, this issue or PR is good enough for it, go directly for the PR.

PR #9115, opened 2021-06-16
Feature/toolchain cmake cross

### LINKED ISSUE #9072: [feature] Cross-compilation model in CMakeToolchain
The new ``CMakeToolchain`` is not managing cross-building yet, variables like CMAKE_SYSTEM_PROCESSOR not set.

The mapping between Conan settings and this variable might not be straightforward, but the old integration does something. It is necessary to check this, and start adding support and testing this case in ``CMakeToolchain``

- https://docs.microsoft.com/en-us/windows/win32/winprog64/wow64-implementation-details
- Windows mapping? AMD64, IA64, ARM64, x86
- Linux: https://stackoverflow.com/questions/45125516/possible-values-for-uname-m

memsharded:
Changelog: Feature: Initial cross-build support in ``CMakeToolchain`` with definition of ``CMAKE_SYSTEM_NAME``, ``CMAKE_SYSTEM_PROCESSOR`` and ``CMAKE_SYSTEM_VERSION``, deduced from ``self.settings_build`` (only using the 2 profiles) and from new ``[conf]`` items.
Docs: https://github.com/conan-io/docs/pull/2140

Close https://github.com/conan-io/conan/issues/9072


conan/tools/cmake/toolchain.py
```diff
@@ -513,6 +526,38 @@ def _get_generator_platform(self, generator):
...
+        if hasattr(self._conanfile, "settings_build"):
+            os_ = settings.get_safe("os")
+            arch = settings.get_safe("arch")
+            settings_build = self._conanfile.settings_build
+            os_build = settings_build.get_safe("os")
+            arch_build = settings_build.get_safe("arch")
+
+            if system_name is None:  # Try to deduce
+                if os_ not in ('Macos', 'iOS', 'watchOS', 'tvOS'):  # Handled by AppleBlock
+                    if os != os_build:
```
czoido:
```suggestion
                    if os_ != os_build:
```

memsharded: Good catch, I have fixed it, added an extra check and a new test to cover it.

czoido (approved):

conan/tools/cmake/toolchain.py
```diff
@@ -513,6 +526,38 @@ def _get_generator_platform(self, generator):
...
+                    if os_ != os_build:
+                        cmake_system_name_map = {"Neutrino": "QNX",
+                                                 "": "Generic",
+                                                 None: "Generic"}
+                        system_name = cmake_system_name_map.get(os_, os_)
+
+            if system_name is not None and system_version is None:
+                os_ver_str = "os.api_level" if os_ == "Android" else "os.version"
+                op_system_version = settings.get_safe(os_ver_str)
+                system_version = op_system_version
```
lasote:
```suggestion
                system_version = settings.get_safe(os_ver_str)
```

lasote (approved):
Looking good!

- Missing docs
- I'm wondering if we should add  more`conf` entries for the rest of the typical stuff that CMake uses in a toolchain file. How would be the alternative for the user without modifying the recipe ( toolchain blocks ) ?
 
```
set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR arm)

set(CMAKE_SYSROOT /home/devel/rasp-pi-rootfs)
set(CMAKE_STAGING_PREFIX /home/devel/stage)

set(tools /home/devel/gcc-4.7-linaro-rpi-gnueabihf)
set(CMAKE_C_COMPILER ${tools}/bin/arm-linux-gnueabihf-gcc)
set(CMAKE_CXX_COMPILER ${tools}/bin/arm-linux-gnueabihf-g++)

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```

memsharded:
> I'm wondering if we should add moreconf entries for the rest of the typical stuff that CMake uses in a toolchain file. How would be the alternative for the user without modifying the recipe ( toolchain blocks ) ?

It is possible to directly inject a complete toolchain, with ``"tools.cmake.cmaketoolchain:user_toolchain"`` conf, in the ``UserToolchain`` block. Is this what you want?


Regarding the ``CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`` variables, yes, maybe we should add some of those lines directly in the toolchain?

KerstinKeller:
I just saw this PR, and I wanted to give some feedback. We're currently using some dirty hacks to induce the correct Toolchain files for cross builds.
Basically every profile which is a cross compilation profile, sets an environment variable 
```
[env]
TOOLCHAIN_FILE=/common/toolchain-files/arm64_toolchainfile.cmake
```
But this toolchain file is usually present in the docker images which run the build, back then we didn't find a better solution.

Also it's nice that Conan will now offer to generate these toolchains (or induce variables in the generic toolchain), *however* from my experience, (real life) cmake toolchain files for cross compilation go beyond just setting sysroot and the compilers, so it might be hard for Conan to generate them correctly.

If profiles could reference existing CMake toolchain files (e.g. don't know where they could be stored, ideally *with*/*next to* the profile), and if these existing toolchains would be induced to the one that Conan generates, or (if the user doesn't use the toolchain generator, if Conan could provide a path to the specified toolchain in the profile), that would be extremly useful and beneficial.

memsharded:
Hi @KerstinKeller 

If I understood correctly, I think that this is already available in 1.37 in the ``[conf]`` of your profile:

- ``tools.cmake.cmaketoolchain:user_toolchain`` in your profile will ``include(xxx)`` that information from the ``conan_toolchain.cmake``
- ``tools.cmake.cmaketoolchain:toolchain_file`` will completely replace the ``conan_toolchain.cmake`` by the user provided one.

Does this help? Is this the feature that you want?
Maybe one UX issue is the definition of paths, that would be absolute or something, something we need to check.

KerstinKeller:
@memsharded 
Ah perfect 👍  , I seem never to be able to keep up with all the new features available 🙈 

about the paths of the toolchain files, they are usually tightly coupled to proflex, e.g. one toolchain file per profile, so it would be cool to have it there. Alternative could be a `build_requires` package, that contains the toolchain, but then I have no clue how to reference that from the profile...

memsharded:
> Alternative could be a build_requires package, that contains the toolchain, but then I have no clue how to reference that from the profile...

We also introduced the possibility for ``build_requires`` to pass [conf] items to their consumers, so if the toolchain is inside a package that is ``build_required`` it could define that config and point it to its own file: https://docs.conan.io/en/latest/reference/config_files/global_conf.html#configuration-from-build-requires

KerstinKeller:
Oh that's very good to know and very handy! Thanks for the info!

memsharded:
> Oh that's very good to know and very handy! Thanks for the info!

:)  

Keep in mind that it is experimental, we are doing a ton of work in this, and this is expected to evolve. Feedback very welcome of course!

lasote (approved):
Don't forget the docs and explain that `tools.cmake.cmaketoolchain:user_toolchain` invalidates the cross build block

PR #9119, opened 2021-06-17
Implementation of the "conan search" command with the new CLI system

### LINKED ISSUE #9105: [feature] Conan 2.0 CLI: Search command
Develop the new Search command for the Conan 2.0 CLI

tapia:
Closes #9105


conans/client/recorder/search_recorder.py
```diff
@@ -5,16 +5,27 @@ class _SearchRecipe(namedtuple("SearchRecipe", "ref")):
...
 
 
 class _SearchPackage(namedtuple("SearchPackage",
                                 "package_id, options, settings, requires")):
-
     def to_dict(self):
-        return {"id": self.package_id, "options": self.options, "settings": self.settings,
-                "requires": self.requires}
+        return {
+            "id": self.package_id,
```
memsharded, line 24: Things like ``package_id`` will no longer make sense in the new search. In any case the ``SearchRecorder`` will probably dissapear or get a big refactor.

conans/client/api/conan_api.py
```diff
@@ -90,19 +167,37 @@ def user_update(self, user_name, user_pasword):
         return {}
 
     @api_method
-    def search_recipes(self, query, remote_patterns=None, local_cache=False):
-        remote = None
-        if remote_patterns is not None and len(remote_patterns) > 0:
-            remote = remote_patterns[0].replace("*", "remote")
+    def search_recipes(self, remotes, query):
+        search_recorder = SearchRecorder()
```
memsharded, line 171: The intermediate conversion to ``SearchRecorder`` doesn't make sense anymore, we need to improve this:

- Define the result data structure. It can directly be generic json if we want to, no problem.
- ``search.search_recipes`` can directly return the result data structure
- Maybe consider the error handling mechanism, what is an error and what not. I don't love attaching the .info field to the exception.

conans/cli/commands/search.py
```diff
@@ -5,38 +5,72 @@
 from conans.cli.command import conan_command, Extender
 
 
+def print_output(info, show_colors):
```
memsharded: If this is not used outside the ``search.py`` module, make it ``_print_output()``. We tend to do this by default, to identify at first glance what is the public interface of a module.

memsharded (commented):
It is looking good, I think there is a lot of refactoring and changes we need to improve inside the ConanApi, but this can be done later, this PR is about the CLI and the output.

conans/cli/commands/search.py
```diff
@@ -5,38 +5,71 @@
...
+            reference = recipe["recipe"]["id"]
+            cli_out_write("{}{}".format(" " * (indentation + 2), reference), reference_color)
+
+
+def _output_search_cli(info):
+    _print_output(info, show_colors=True)
+
+
+def _output_search_raw(info):
+    _print_output(info, show_colors=False)
```
memsharded: Not sure about the ``show_colors`` parametrization. Colorama approach is to add always colors, and if colors are not enabled, then it will filter out the escape caracters. So color is something that is configurable at the app level, maybe not necessary to implement 2 outputs for each command, one with color and one without.

conans/test/integration/command_v2/search_test.py
```diff
@@ -0,0 +1,216 @@
...
+        self.client.run("search whatever", assert_error=True)
+        self.assertIn("ERROR: The remotes registry is empty", self.client.out)
+
+    def test_search_disabled_remote(self):
+        self.client.run("remote disable remote1")
+        self.client.run("search whatever -r remote1", assert_error=True)
+        self.assertIn("ERROR: Remote 'remote1' is disabled", self.client.out)
+
+
+class SearchRemotesTest(unittest.TestCase):
```
memsharded: For new tests, please use ``pytest`` style.

conans/test/integration/command_v2/search_test.py
```diff
@@ -0,0 +1,216 @@
+import textwrap
+import unittest
+from collections import OrderedDict
+
+from conans.test.utils.tools import TestClient, TestServer
+
+
+class SearchTest(unittest.TestCase):
```
memsharded: pytest

conans/test/integration/command_v2/search_test.py
tapia: Done

conans/test/integration/command_v2/search_test.py
tapia: Done

conans/cli/commands/search.py
tapia: Done

conans/cli/commands/search.py
tapia: I removed the whole `raw` formatter. It doesn't add any value to this command.

PR #9124, opened 2021-06-18
Change sources management

lasote:
Changelog: Fix: Fixed behavior in the `self.folders` feature whereby the sources saved or downloaded inside the `source(self)` were saved at the `self.folders.source` folder. But `self.folders.source` is intended to describe where the sources are instead of forcing where the sources are saved.
Docs: https://github.com/conan-io/docs/pull/2127


memsharded (approved):
So far looks very reasonable, I think it makes sense specifier-layout!=force-to-save-sources in that layout

PR #9129, opened 2021-06-21
testing cmake imported config

### LINKED ISSUE #7999: [question] Mapped configs in cmake_find_package_multi.

memsharded:
Changelog: Omit
Docs: Omit

This test proves that CMAKE_MAP_IMPORTED_CONFIG mappings are not needed with CMakeDeps, as it is able to mix build_types from upstream dependencies.

What is necessary is to improve the UX definition, and also to allow defining the build_type just for the consumer, even if it is a conanfile.txt, there is no syntax for that. I'll create a new ticket for it.

Close https://github.com/conan-io/conan/issues/7999

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmakedeps_custom_configs.py
+++ b/conans/test/functional/toolchains/cmake/test_cmakedeps_custom_configs.py
@@ -212,31 +212,44 @@ def test_changing_build_type():
         cmake_minimum_required(VERSION 3.15)
         project(App CXX)
 
-        set(CMAKE_MAP_IMPORTED_CONFIG_DEBUG Release)
-
         find_package(dep REQUIRED)
         add_executable(app app.cpp)
         target_link_libraries(app PRIVATE dep::dep)
         """)
     app = gen_function_cpp(name="main", includes=["hello"], calls=["hello"])
-    pkg_conanfile = GenConanfile().with_requires("dep/0.1").\
+    pkg_conanfile = GenConanfile("pkg", "0.1").with_requires("dep/0.1").\
         with_generator("CMakeDeps").with_generator("CMakeToolchain").\
         with_settings("os", "compiler", "arch", "build_type")
     client.save({"conanfile.py": pkg_conanfile,
                  "CMakeLists.txt": cmakelists,
                  "app.cpp": app}, clean_first=True)
 
-    client.run("install . -s build_type=Debug -s dep:build_type=Release")
+    # in MSVC multi-config -s pkg:build_type=Debug is not really necesary, toolchain do nothing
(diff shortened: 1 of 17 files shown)


conans/test/functional/toolchains/cmake/test_cmakedeps_custom_configs.py
```diff
@@ -183,3 +184,54 @@ def test_generator_multi(self):
...
+        with_settings("os", "compiler", "arch", "build_type")
+    client.save({"conanfile.py": pkg_conanfile,
+                 "CMakeLists.txt": cmakelists,
+                 "app.cpp": app}, clean_first=True)
+
+    # in MSVC multi-config -s pkg:build_type=Debug is not really necesary, toolchain do nothing
+    # TODO: Challenge how to define consumer build_type for conanfile.txt
+    client.run("install . -s pkg:build_type=Debug -s build_type=Release")
+    client.run_command("cmake . -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake")
+    client.run_command("cmake --build . --config Debug")
```
lasote, line 232: This is only for multi-config windows, right? why is it working in Linux/Osx?

lasote (approved):

PR #9131, opened 2021-06-21
Feature/autotools new improvements

### LINKED ISSUE #9069: [feature] Keep modernizing and testing the new ``conan.tools.gnu`` Autotools integration
It is necessary to keep pushing the new Autotools integration:

- Migrate helpers like ``cppstd_flag_new`` that still live in the old context. 
- Modernize them: using always 2 profiles, no auto-detection at all (for example the one in ``_detected_os()`` in ``get_cross_building()``
- Keep implementing TODOS like:
```python
    def _rpaths_link(self):
        # TODO: Not implemented yet
        pass

    # TODO: Apple: tools.apple_deployment_target_flag,
    # TODO:  tools.XCRun(self._conanfile.settings).sdk_path
    # TODO: "-arch", tools.to_apple_arch(self._arch)
```
- More comprehensive testing coverage of features.
- Check configurability and extensitibility of provided tools is good, or check alternatives.
- Usage of new cpp info

lasote:
Changelog: Feature: Improved the new `AutotoolsToolchain`, `AutotoolsDeps` and `Autotools` build helper.
Docs: https://github.com/conan-io/docs/pull/2135

Closes https://github.com/conan-io/conan/issues/9069


conan/tools/files/files.py
```diff
@@ -181,3 +183,15 @@ def rename(conanfile, src, dst):
             os.rename(src, dst)
         except Exception as err:
             raise ConanException("rename {} to {} failed: {}".format(src, dst, err))
+
+
+def load_build_json(conanfile):
```
lasote, line 188: Probably better wait until we decide the new format.

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -1,91 +1,63 @@
...
         env.append("LDFLAGS", ldflags)
         env.append("CXXFLAGS", cxxflags)
         env.append("CFLAGS", cflags)
         return env
 
-    def generate(self):
-        env = self.environment()
-        env.save_sh("conanautotoolsdeps.sh")
-        env.save_bat("conanautotoolsdeps.bat")
+    def generate(self, env=None):
```
lasote, line 61: To discuss: What if we want to modify/add some variable in the environment before generate? what would be the best interface?

conan/tools/gnu/autotools.py
```diff
@@ -3,69 +3,50 @@
...
-        self._compiler_version = conanfile.settings.get_safe("compiler.version")
-        self._build = None
-        self._host = None
-        self._target = None
-        if os.path.isfile(CONAN_TOOLCHAIN_ARGS_FILE):
-            args = json.loads(load(CONAN_TOOLCHAIN_ARGS_FILE))
-            self._build = args["build"] if "build" in args else None
-            self._host = args["host"] if "host" in args else None
-            self._target = args["target"] if "target" in args else None
+        self.environment_files = ["conanbuildenv", "conanautotoolstoolchain", "conanautotoolsdeps",
```
lasote, line 15: To discuss: I removed the `AutoolsGen` because I think it is easier to be explicit but we should discuss if this is the best idea of "applying" the environments or we can think of a better one. I have seen something similar in the Bazel one.

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -1,91 +1,63 @@
...
-                self.sysroot = dep_cpp_info.sysroot
-
-        for d in self._conanfile.dependencies.host.values():
-            merge(d)
+        self._cpp_info = None
+
+    @property
+    def cpp_info(self):
+        if self._cpp_info is None:
+            self._cpp_info = NewCppInfo()
```
lasote, line 16: Only CMake and now Autotools use the new cpp info. We should plan move the rest of them.

conans/model/conan_file.py
```diff
@@ -387,7 +387,7 @@ def package_info(self):
     def run(self, command, output=True, cwd=None, win_bash=False, subsystem=None, msys_mingw=True,
             ignore_errors=False, run_environment=False, with_login=True, env="conanbuildenv"):
 
-        command = environment_wrap_command(env, command)
+        command = environment_wrap_command(env, command, cwd=self.generators_folder)
```
lasote, line 390: Now the wraps will have absolute paths to the generators folder because sometimes you have to run the command inside (cwd) a different folder.

conan/tools/env/environment.py
```diff
@@ -197,6 +199,13 @@ def save_sh(self, filename, generate_deactivate=False, pathsep=os.pathsep):
         content = "\n".join(result)
         save(filename, content)
 
+    def save_script(self, name):
```
lasote, line 202: Improved UX, the same code was starting to be repeated in several places

conan/tools/gnu/autotoolsgen.py
```diff
@@ -1,37 +0,0 @@
-import platform
```
lasote, line 1: I removed this but not because I don't like the aggregator, but the way it was creating a single `conanbuildenv.bat` made it difficultly to "split" the generators and keep the build helpers working. See the other comment in the build helper where I'm not convinced either about the best way to wrap the command with all the available environment scripts.

conan/tools/cross_building.py
```diff
@@ -0,0 +1,29 @@
+
+
+def cross_building(conanfile):
```
memsharded: This is no longer true for 64->32 bits for example, and there is code in CMake that filters it. Not to change now, I understand this is an initial refactor moving things into scope, but we certainly need to consolidate all of this.

conan/tools/files/files.py
memsharded: Yes, do not make this public in ``conan/tools/files/__init__.py``, keep it internal at the moment.

conan/tools/gnu/get_cppstd.py
```diff
@@ -0,0 +1,270 @@
+from conans.model.version import Version
+
+
+def cppstd_flag(settings):
```
memsharded: This file doesn't look like a ``gnu`` thing only, it might be common to cmake, meson and others

conan/tools/gnu/autotools.py
memsharded: I agree is nicer to be explicit for Conan, so good for me to remove it. The only thing we should consider is if there is other way to improve the UX of the developer flow that does ``conan install`` + activate several env scripts (order matter? should I activate all of them?)

memsharded (commented):
Overall is a good improvement, great job.

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -115,21 +124,46 @@ def environment(self):
...
         env.append("CXXFLAGS", self.cxxflags)
         env.append("CFLAGS", self.cflags)
         env.append("LDFLAGS", self.ldflags)
         return env
 
-    def generate(self):
-        env = self.environment()
-        env.save_sh("conanautotoolstoolchain.sh")
-        env.save_bat("conanautotoolstoolchain.bat")
+    def generate(self, env=None):
```
memsharded, line 141: What is the purpose of this ``env=None`` environment?

Note that it will fail to override an empty environment, some users might want "do nothing", and pass an empty dict, but it will still define the default env?

memsharded (approved):
Good to go, I suggest merging and keep working and testing it.

conan/tools/gnu/autotoolstoolchain.py
lasote: It was related to my previous comment: `"To discuss: What if we want to modify/add some variable in the environment before generate? what would be the best interface?"` 
Users might call first to `environment()` then modify something and finally generate, but I'm not convinced, that's why I was asking to discuss it.

PR #9137, opened 2021-06-22
adding context information to conan-info command

### LINKED ISSUE #9121: [feature] Return node context in `conan info` command
It would be very useful to get **information about the node context (_host_ or _build_) in the `conan info` output**.... of course, it only makes sense when using the `--dry-build` argument and build-requires are expanded, but it probably should be added to the output always (console and JSON).

memsharded:
Changelog: Feature: Add ``context`` information to ``conan info`` output both to stdout and json outputs.
Docs: https://github.com/conan-io/docs/pull/2142

Close https://github.com/conan-io/conan/issues/9121


czoido (approved):

jgsogo (commented):
No time to try it, but it should work 🤷

PR #9148, opened 2021-06-23
fix ConanFileDependencies.build[dep] access

memsharded:
Changelog: BugFix: ``ConanFileDependencies.build["dep"]`` was retrieving the ``host`` dependency if existing, or failing otherwise, because the default requires was hardcoded to fetch the host (build=False) dependency.

Docs: https://github.com/conan-io/docs/pull/2134


lasote (approved):
Looks much better

PR #9161, opened 2021-06-25
Environment Automanage

lasote:
Changelog: Fix: The new `VirtualEnv` generator has been split into `VirtualRunEnv` and `VirtualBuildEnv`. Both are automatically generated as before but only `VirtualBuildEnv` will be activated by default.
Changelog: Feature: A unique environment launcher (`conanenv.bat/sh`) is generated to aggregate all the environment generators (`VirtualRunEnv`, `VirtualBuildEnv`, `AutotoolsToolchain` and `AutotoolsDeps`) that had been generated so the user can easily activate all of them with one command.

Docs: https://github.com/conan-io/docs/pull/2151

The conanfile keeps a list of environment scripts generated at `self.environment_scripts`, then after `generate()`, conan generates a wrapper (one for `.bat` and one for `.sh`) called always `conanenv` so the "runners" can default to that environment always.

The user could adjust the `self.environment_scripts` in the generate method to, for example, delete some element.

IMPORTANT: 
1) This changes the previous behavior as by default activates also the `conanrunenv`
2) The automatic virtualenv prevent the user to remove the `conanrunenv` or `conanbuildenv` from `self.environment_scripts` because they are not added yet to the list.


conan/tools/microsoft/msbuild.py
```diff
@@ -53,8 +53,7 @@ def command(self, sln):
 
     def build(self, sln):
         cmd = self.command(sln)
-        vcvars = os.path.join(self._conanfile.generators_folder, "conanvcvars")
-        self._conanfile.run(cmd, env=["conanbuildenv", vcvars])
+        self._conanfile.run(cmd)
```
memsharded, line 56: I don't see who is adding ``conanvcvars`` now

memsharded (commented):
I like the idea a lot

conan/tools/microsoft/msbuild.py
lasote: The idea is to have a explicit `Vcvars(self).generate()` or `generators="Vcvars"` and it will be automatically added to the list of environments.

lasote: I have a draft in a branch but with more things and very dirty yet.

memsharded: The problem with the vcvars generator is that it might be too coupled to other generator, like ``CMakeToolchain`` is the one that has the information to decide whether to generate ``vcvars`` or not. I don't see it, but if you have a branch we can check it later.

memsharded:
> This changes the previous behavior as by default activates also the conanrunenv

Not clear, by default this probably shouldn't be activated, it often can conflict with the buildenv

conan/tools/microsoft/msbuild.py
lasote: Now the `write_conanvcvars()` function is adding the script to the list. We can add later (in other ticket/release) the generator for the user.

lasote:
The `VirtualEnv` has been split into `VirtualRunEnv` and `VirtualBuildEnv`. It was getting too much complicated to get them in only one generator but managing automatically the `VirtualBuildEnv` only. I think now is more explicit and clear and with a more direct relation with `self.runenv_info` and `self.buildenv_info`

conan/tools/env/environment.py
```diff
@@ -351,3 +344,21 @@ def loads(text):
...
+        env.save_bat(path)
+    else:
+        path = os.path.join(conanfile.generators_folder, "{}.sh".format(name))
+        env.save_sh(path)
+
+    register_environment_script(conanfile, path)
+
+
+def register_environment_script(conanfile, path):
+    element = os.path.basename(path)
```
memsharded: Are we registering only the names, not the paths to the scripts? Then assume it will always be relative to generators_folder? What if some tool reference some script that is in the profiles folder?

conans/client/generators/__init__.py
```diff
@@ -226,8 +231,34 @@ def write_toolchain(conanfile, path, output):
                 conanfile.generate()
 
     # tools.env.virtualenv:auto_use will be always True in Conan 2.0
-    if conanfile.conf["tools.env.virtualenv:auto_use"] and conanfile.virtualenv:
+    if conanfile.conf["tools.env.virtualenv:auto_use"] and conanfile.virtualbuildenv:
         with chdir(path):
-            from conan.tools.env.virtualenv import VirtualEnv
-            env = VirtualEnv(conanfile)
+            from conan.tools.env.virtualbuildenv import VirtualBuildEnv
```
memsharded, line 237: We need also the ``VirtualRunEnv`` to be able to run things with shared dependencies, as a relevant and common use case.

memsharded (commented):
I would say not to kill the ``VirtualRunEnv`` as built-in, too, it is a very common use case: wanting to run with shared deps. Recipes should work when some deps change ``option:shared=True``, without much effort or needing to explicitly change every recipe to account for this use case.

conan/tools/env/environment.py
lasote: I didn't think about this being a public function and not sure about accepting scripts in the profile folder, but ok, we can add the full path. Just a bit more ugly if someone wants to "alter" the list by hand.

conans/client/generators/__init__.py
lasote: Done

memsharded (approved):
I think this is ready, good job.

PR #9168, opened 2021-06-25
making Paths explicit in new Environment

memsharded:
Changelog: Omit
Docs: Omit

Making the different between a Path and other env-vars in the new ``Environment``.

The final extraction of values for the ``win_bash`` and others, would need some minor changes, like changing ``items()`` to return the new ``_EnvValue`` so the ``is_path`` property can be checked, but I haven't done it because better do it depending on the consumer use case.

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -52,6 +52,9 @@ class _EnvValue:
         self._path = path
         self._sep = separator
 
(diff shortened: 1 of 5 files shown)


lasote (approved):
Like it better. The only change of not allowing different separators is totally fine.

lasote (approved):

PR #9174, opened 2021-06-29
minor changes while inspecting cli

memsharded:
Changelog: Omit
Docs: Omit

Just some minor changes while reading trying to refresh my understanding of the new CLI code.

### DIFF
--- a/conans/cli/output.py
+++ b/conans/cli/output.py
@@ -153,3 +153,8 @@ class ConanOutput(object):
                 # TODO: check if colorama checks for stripping colors are enough
                 colorama.init()
             return True
+
+
+def cli_out_write(data, fg=None, bg=None, endline="\n"):
+    data = "{}{}{}{}{}".format(fg or '', bg or '', data, Style.RESET_ALL, endline)
+    sys.stderr.write(data)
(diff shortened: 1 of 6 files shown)


conans/cli/output.py
```diff
@@ -153,3 +153,8 @@ def _init_colors():
                 # TODO: check if colorama checks for stripping colors are enough
                 colorama.init()
             return True
+
+
+def cli_out_write(data, fg=None, bg=None, endline="\n"):
+    data = "{}{}{}{}{}".format(fg or '', bg or '', data, Style.RESET_ALL, endline)
+    sys.stderr.write(data)
```
memsharded: Not sure about the intention of this, if ``conan help`` output for example should go to stderr or to stdout? Is in this case the output considered the cli printing?

tapia: I think this should be `sys.stdout`, not `stderr`

czoido: Yes, this one should go to `sys.stdout ` this is the one that is used in the formatters of the commands and all the output should go to `stdout`.

memsharded: Fixed

czoido (approved):

Issue #9170, opened 2021-06-28
Remove PackageCacheLayout from Conan 2.0


czoido:
Now that first implementation of the multi-revision Conan Cache for 2.0 is already merged: #8796 
We should remove all the code from the old Conan Cache for 1.X

- The class PackageCacheLayout is still used in through the Conan Test Suite calling underneath to the new 2.0 Cache. We should remove that.
- There are some recurrent structures through the test suite for Conan 2.0 to get latest revisions and new layouts so we could introduce new helpers for the TestClient class in 2.0.

czoido:
Closed by: https://github.com/conan-io/conan/pull/9179

PR #9179, opened 2021-06-29
Remove PackageCacheLayout from Conan 2.0 #9170

franramirez688:
Changelog: Feature: Removing any call to old `PackageCacheLayout` from `conans.*`. Adapting it to new cache layout usage.
- Deleted module `conans/paths/package_layouts/package_cache_layout.py`
- PackageCacheLayout does not exist anymore.
- Using newest `ClientCache` methods which are calling to new cache.

Docs: omit
Closes: #9170


conans/test/integration/command/download/download_test.py
```diff
@@ -1,212 +1,224 @@
...
-        self.assertIn("pkg/1.0: Downloading pkg/1.0:%s" % NO_SETTINGS_PACKAGE_ID, client.out)
-        self.assertIn("pkg/1.0: Package installed %s" % NO_SETTINGS_PACKAGE_ID, client.out)
-
-        # All
-        client.run("remove * -f")
-        client.run("download pkg/1.0@")
-        self.assertIn("pkg/1.0: Downloading pkg/1.0:%s" % NO_SETTINGS_PACKAGE_ID, client.out)
-        self.assertIn("pkg/1.0: Package installed %s" % NO_SETTINGS_PACKAGE_ID, client.out)
+@pytest.fixture()
+def ref():
```
memsharded: One line fixtures introduce more complexity than they provide. Just create the ``ConanFileReference`` in every test.

conans/client/cache/cache.py
```diff
@@ -153,6 +169,7 @@ def config_install_file(self):
     # TODO: cache2.0 this will be removed in the future is just to adapt to some tests
     #  that call this directly
     def package_layout(self, ref, short_paths=None):
+        raise Exception("buuuuuuuhhhhh")
```
memsharded: Buuuuuhhhhhh!!! 😄 


If this is no longer used, just remove the method.

conans/client/conan_command_output.py
```diff
@@ -71,7 +71,7 @@ def _read_dates(self, deps_graph):
         for node in sorted(deps_graph.nodes):
             ref = node.ref
             if node.recipe not in (RECIPE_CONSUMER, RECIPE_VIRTUAL, RECIPE_EDITABLE):
-                manifest = self._cache.package_layout(ref).recipe_manifest()
+                manifest = self._cache.get_ref_layout(ref).recipe_manifest()
```
franramirez688, line 74: @czoido @memsharded should I have used `get_latest_` instead?

conans/client/installer.py
```diff
@@ -462,7 +462,7 @@ def _handle_node_editable(self, node, profile_host, profile_build, graph_lock):
         # Get source of information
         conanfile = node.conanfile
         ref = node.ref
-        package_layout = self._cache.package_layout(ref)
+        package_layout = self._cache.get_ref_layout(ref)
```
franramirez688: @czoido @memsharded should I have used `get_latest_` instead?

conans/client/conan_command_output.py
```diff
@@ -125,15 +125,14 @@ def _grab_info_data(self, deps_graph, grab_paths):
 
             # Paths
             if isinstance(ref, ConanFileReference) and grab_paths:
-                package_layout = self._cache.package_layout(ref, conanfile.short_paths)
-                item_data["export_folder"] = package_layout.export()
-                item_data["source_folder"] = package_layout.source()
+                ref_layout = self._cache.get_ref_layout(ref)
```
franramirez688, line 129: @czoido @memsharded should I have used `get_latest_` instead?

conans/client/conan_command_output.py
memsharded: When the graph is resolved, the ``ref`` should already have the ``revision``. Internally, it has an assert for it:

```python
 def get_reference_layout(self, ref: ConanReference):
        assert ref.rrev, "Recipe revision must be known to get the reference layout"
```

conans/client/cache/cache.py
franramirez688: I just pushed the new code 😉

conans/client/cache/cache.py
```diff
@@ -134,6 +133,22 @@ def get_latest_prev(self, ref):
         prevs = self.get_package_revisions(ref, True)
         return prevs[0] if prevs else None
 
+    def get_latest_pkg_layout(self, ref: ConanReference) -> PackageLayout:
```
memsharded: This looks mostly like test code in production scope:

- Better move them to ``TestClient``
- When necessary, in production code, call explicitly ``cache.get_latest_rrev()`` to get the latest revision (in the ``conan get``)

This way we will have more control over who is resolving to latest revision in the cache, at the moment it is only used in a couple of places (uploader, proxy).

conans/client/cache/cache.py
```diff
@@ -134,6 +133,22 @@ def get_latest_prev(self, ref):
...
+        """Get the latest PackageLayout given a file reference"""
+        latest_rrev = self.get_latest_rrev(ref)
+        # Given the latest recipe revision we can get the latest package ID and
+        # the corresponding PackageLayout
+        all_package_ids = self.get_package_ids(latest_rrev)
```
memsharded: This doesn't look correct for production code, seems pretty arbitrary getting just the first package id in ``all_package_ids[0]``. Could be good as a simplification for testing, but looks dangerous here.

conans/client/conan_command_output.py
memsharded: Same as above, ref already has a revision, get_ref_layout will check and assert.

memsharded (changes requested):

conans/client/cache/cache.py
franramirez688: Yeah, you're completely right. I should have realized about having something like this one:

```python
    def get_latest_pkg_layout(self, ref: PackageReference) -> PackageLayout:
        """Get the latest PackageLayout given a package reference"""
        latest_prev = self.get_latest_prev(ref)
        pkg_layout = self.get_pkg_layout(latest_prev)
        return pkg_layout
```

But yes, agree with having it explicitly in production scope 👍

conans/client/conan_command_output.py
```diff
@@ -125,15 +125,14 @@ def _grab_info_data(self, deps_graph, grab_paths):
 
             # Paths
             if isinstance(ref, ConanFileReference) and grab_paths:
-                package_layout = self._cache.package_layout(ref, conanfile.short_paths)
-                item_data["export_folder"] = package_layout.export()
-                item_data["source_folder"] = package_layout.source()
+                ref_layout = self._cache.get_ref_layout(self._cache.get_latest_rrev(ref))
```
memsharded: Are you sure the ``get_latest_rrev()`` is needed here? If it is from ``info``, it should have resolved the revision already.

conans/client/installer.py
```diff
@@ -462,7 +462,7 @@ def _handle_node_editable(self, node, profile_host, profile_build, graph_lock):
         # Get source of information
         conanfile = node.conanfile
         ref = node.ref
-        package_layout = self._cache.package_layout(ref)
+        package_layout = self._cache.get_ref_layout(self._cache.get_latest_rrev(ref))
```
memsharded: This is probably wrong, editables are not in the cache, asking for the layout of the latest revision in teh cache doesn't make sense.
In the past, ``cache.package_layout()`` could return a specific EditableLayout, but that doesn't seem the case anymore. Please check this

conans/test/functional/command/devflow_test.py
```diff
@@ -62,7 +62,7 @@ def test_parallel_folders(self):
         client.run("export-pkg . Pkg/0.1@lasote/testing -bf=../build")
 
         ref = ConanFileReference.loads("Pkg/0.1@lasote/testing")
-        cache_package_folder = client.cache.package_layout(ref).packages()[0]
+        cache_package_folder = client.get_latest_pkg_layout(ref).package()
```
memsharded: This is weird, how can you get the pkg_layout if you don't provide the ``package_id``. It is a bit risky and confusing, maybe the argument should always be a ``pref``, not a ``ref``

conans/test/functional/layout/test_in_cache.py
```diff
@@ -122,9 +122,9 @@ def test_cache_in_layout(conanfile):
     package_id = re.search(r"lib/1.0:(\S+)", str(client.out)).group(1)
     ref = ConanFileReference.loads("lib/1.0@")
     pref = PackageReference(ref, package_id)
-    sf = client.cache.package_layout(ref).source()
-    bf = client.cache.package_layout(ref).build(pref)
-    pf = client.cache.package_layout(ref).package(pref)
+    sf = client.get_latest_ref_layout(ref).source()
+    bf = client.get_latest_pkg_layout(ref).build()
```
memsharded: Same here ,shouldn't this receive ``pref`` instead of ``ref``

conans/test/functional/command/devflow_test.py
franramirez688: Yes, that was I was thinking... Ok, let me check all the cases because it's not the only one in the tests

conans/client/conan_command_output.py
```diff
@@ -125,12 +125,12 @@ def _grab_info_data(self, deps_graph, grab_paths):
...
             if isinstance(ref, ConanFileReference) and grab_paths:
-                ref_layout = self._cache.get_ref_layout(self._cache.get_latest_rrev(ref))
+                ref_layout = self._cache.get_ref_layout(ref)
                 item_data["export_folder"] = ref_layout.export()
                 item_data["source_folder"] = ref_layout.source()
                 pref_build_id = build_id(conanfile) or package_id
-                pref_build = self._cache.get_latest_prev(PackageReference(ref, pref_build_id))
-                pref_package = self._cache.get_latest_prev(PackageReference(ref, package_id))
+                pref_build = PackageReference(ref, pref_build_id)
+                pref_package = PackageReference(ref, package_id)
```
memsharded: ``ref`` above had recipe revision ``rrev`` but this ``pref`` do NOT have package revision, still the command is returning a "build_folder". This is quite confusing @czoido and probably a cause of the refactor, I think it might be even creating a new temporary folder for a ``prev``?

czoido: Yes, this fails but the tests for this (`InfoFoldersTest`) have the xfail mark and that's why this is passing green.
This will not create a temporal folder but raise an error because Package revision must be known to get the package layout.

conans/client/conan_api.py
```diff
@@ -1100,13 +1098,42 @@ def read_profile(self, profile=None):
...
+                :param path: path relative to the cache reference or package folder
+            """
+            assert not os.path.isabs(path)
+
+            if package_id is None:  # Get the file in the exported files
+                latest_rrev = cache.get_latest_rrev(ref)
+                folder = cache.get_ref_layout(latest_rrev).export()
+            else:
+                latest_pref = cache.get_latest_prev(PackageReference(ref, package_id))
+                folder = cache.get_pkg_layout(latest_pref).package()
```
czoido: This will fail because you don't have the recipe revision for the reference. It's passing in the PR because `ConanGetTest` are marked as xfail

conans/test/functional/command/devflow_test.py
czoido: The thing I see here is that we force to give a package id for this, then we will need to add more code just to get the list of package_ids and after that the list of package revisions. Old tests don't fit so well with the new cache, they try to get packages folders from just the `ConanFileReference` in many places... but yes, probably it's better to fix that

conans/test/functional/only_source_test.py
```diff
@@ -1,163 +1,164 @@
...
-        # self.assertFalse(os.path.exists(other_client.cache.package_layout(ref).builds()))
-        # self.assertFalse(os.path.exists(other_client.cache.package_layout(ref).packages()))
-
-        # Now even if the package is in local store, check that's rebuilded
-        other_client.run("install %s -b Hello*" % str(ref))
-        self.assertIn("Copying sources to build folder", other_client.out)
-
-        other_client.run("install %s" % str(ref))
-        self.assertNotIn("Copying sources to build folder", other_client.out)
+def test_conan_test():
```
czoido: Lots of changes here are not related to the intention of the PR, I would prefer to have these separated in another PR

conans/test/unittests/client/cmd/export/test_update_revision_in_metadata.py
```diff
@@ -18,8 +17,9 @@ class UpdateRevisionInMetadataTests(unittest.TestCase):
 
     def setUp(self):
         ref = ConanFileReference.loads("lib/version@user/channel")
-        self.package_layout = PackageCacheLayout(base_folder=temp_folder(), ref=ref,
-                                                 short_paths=False, no_lock=True)
+        # FIXME: 2.0: PackageCacheLayout does not exist anymore
```
czoido, line 21: I think we can remove all these tests, maybe adding a TODO to add some unittests for the cache on getting the fields that replace the metadata for 2.0.

conans/test/functional/only_source_test.py
memsharded: Agree, in general, tests:

- New tests should be written in pytest
- Old tests, with very minor changes, do only those minor things in the PR
- Submit a separate PR with the pytest-modernizing of the tests, as a pure refactor (that can be done before or after)

conans/test/utils/tools.py
```diff
@@ -659,14 +661,37 @@ def scm_info_cache(self, reference):
...
+        all_package_ids = self.cache.get_package_ids(latest_rrev)
+        return self.cache.get_latest_prev(all_package_ids[0])
+
+    def get_latest_pkg_layout(self, pref: PackageReference) -> PackageLayout:
+        """Get the latest PackageLayout given a file reference"""
+        # Let's make it easier for all the test clients
+        if pref.ref.revision is None:
+            pref_ = PackageReference(self.cache.get_latest_rrev(pref.ref), pref.id)
+        else:
+            pref_ = pref
```
franramirez688: @memsharded @czoido here, I had doubts about assuming that we're always passing the pref with the recipe revision or helping test clients to pass it or not, what do you think?

czoido: I think that I would opt for assuming that the PackageReference comes with the rrev and fail if it doesn't. Then, if you have to get the latest rrev it could be done in the test before calling this method.

franramirez688: Agree but the thing is that if I request a pref with recipe revision included, almost all the occurrences will look like:

```python
pref = client.get_latest_pref(ANY_REF)
pkg_layout = client.get_pkg_layout(pref)
```
So, totally agree with these two lines will appear a lot of times, right?

memsharded: you say ``get_latest_pref()`` Is that a typo, you mean ``get_latest_prev()``?

The main problem in the above is not the latest revision, but the ``all_package_ids[0]``. It doesn't make sense to pick the first (and assuming only) package-id in the cache, so the package_id needs to be explicit (it is relatively easy to capture it in the output while creating the package). The latest revisions, both recipe and package could be automated in the ``TestClient``, IMO.

franramirez688: Ok, I'm passing the tests locally and it's not happening for all the cases so it should not be a big deal. Applying the changes...

conans/test/unittests/client/cmd/export/test_update_revision_in_metadata.py
franramirez688: Yeah, but I think it's better to keep them for now and, open another PR with those new tests and get rid of these ones. Anyway, I'm gonna add that TODO

conans/client/conan_api.py
```diff
@@ -3,11 +3,13 @@
 import sys
 from collections import OrderedDict
 from collections import namedtuple
-
 from io import StringIO
 
+from conans.model.manifest import discarded_file
```
memsharded: Use IDE optimize imports or any other tools, that will arrange the imports in order.

conans/client/installer.py
memsharded: I think this is wrong, this must be a ``package_folder``, but it seems you are using a ``recipe folder``

memsharded: I am surprised that this is not breaking tests, but ok, I guess too many commented tests now.

conans/test/functional/build_helpers/autotools_environment_test.py
```diff
@@ -55,15 +53,18 @@ def package_id(self):
                      "hello.h": header,
                      "hello.cpp": body})
         client.run("create . danimtb/testing")
-        pref = PackageReference.loads("test/1.0@danimtb/testing:%s" % NO_SETTINGS_PACKAGE_ID)
-        pkg_path = client.cache.package_layout(pref.ref).package(pref)
+        pref = client.get_latest_prev(ConanFileReference.loads("test/1.0@danimtb/testing"),
```
memsharded, line 56: Tip: allow ``TestClient.get_latest_prev()`` to accept a string too, and convert internally to ``ConanFileFerence``. (keep the possibility of passing a ConanFileReference)
TestClient helpers are there to make test code simpler.

conans/test/utils/tools.py
```diff
@@ -659,14 +661,36 @@ def scm_info_cache(self, reference):
...
         else:
             return self._create_scm_info(dict())
 
+    def get_latest_prev(self, ref: ConanReference, package_id=None) -> PackageReference:
+        """Get the latest PackageReference given a ConanReference"""
+        latest_rrev = self.cache.get_latest_rrev(ref)
+        if package_id:
+            pref = PackageReference(latest_rrev, package_id)
+        else:
+            pref = self.cache.get_package_ids(latest_rrev)[0]
```
memsharded: This is a bit dangerous. I understand the utility, but please, at least raise an error if there is more than 1 package_id in the package.

memsharded (changes requested):

conans/client/installer.py
franramirez688: @memsharded  Would it be fine if add

```python
        package_layout = self._cache.get_pkg_layout(node.pref)
```
?

franramirez688: I guess the same, the test is not raising anything because of all the commented tests...

czoido: Yes, all tests related to editables are xfailed so changes here will probably not affect.

memsharded: Ok, yes, that is true, that is editable... I guess we can leave it at the moment, until editables are fixed.

franramirez688: Ok then, keeping it for now

memsharded (approved):

czoido (approved):

Issue #9183, opened 2021-06-30
[bug] Disabling remote prevents build from proceeding


cassava:
Disabling a broken remote prevents me from building a Conan package, even though I have all packages I need in my local cache.

### Environment Details (include every applicable attribute)
  * Operating System: Ubuntu 18.04
  * Compiler: GCC=8
  * Conan version: 1.37.2
  * Python version: 3.6.9

### Steps to reproduce

1. `conan remote disable artifactory`
2. `conan install --build=missing --install-folder=build -g cmake`

### Logs

1. Timeout
```console
$ conan install --build=missing --install-folder build -g cmake .
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=RelWithDebInfo
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

ERROR: HTTPSConnectionPool(host='artifactory.example.com', port=443): Max retries exceeded with url: /artifactory/api/conan/conan-local/v1/conans/search?q=fmt%2F%2A&ignorecase=False (Caused by ReadTimeoutError("HTTPSConnectionPool(host='artifactory.example.com', port=443): Read timed out. (read timeout=60.0)",))

Unable to connect to artifactory=https://artifactory.example.com/artifactory/api/conan/conan-local
1. Make sure the remote is reachable or,
2. Disable it by using conan remote disable,
Then try again.
```

2. Disable artifactory remote:
```console
$ conan remote disable artifactory
```

3. Try building again:
```console
$ conan install --build=missing --install-folder build -g cmake .
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=RelWithDebInfo
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

ERROR: Remote 'artifactory' is disabled
```

memsharded:
Hi @cassava 

Thanks for the detailed report, already reproduced, indeed seems a bug, and I am proposing some potential fix in #9184

memsharded:
#9184 has been last minuted merged into 1.38, so it will be released today

PR #9184, opened 2021-06-30
Fix/disabled remotes

memsharded:
Changelog: Bugfix: Disabled remotes shouldn't fail if not used at all
Docs: Omit

Fix https://github.com/conan-io/conan/issues/9183

### DIFF
--- a/conans/test/integration/command/install/install_test.py
+++ b/conans/test/integration/command/install/install_test.py
@@ -363,8 +363,8 @@ def test_install_disabled_remote(client):
     client.run("create . Pkg/0.1@lasote/testing")
     client.run("upload * --confirm --all -r default")
     client.run("remote disable default")
-    client.run("install Pkg/0.1@lasote/testing -r default", assert_error=True)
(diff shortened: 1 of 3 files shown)


conans/test/integration/command/install/install_test.py
```diff
@@ -363,16 +363,16 @@ def test_install_disabled_remote(client):
     client.run("create . Pkg/0.1@lasote/testing")
     client.run("upload * --confirm --all -r default")
     client.run("remote disable default")
-    client.run("install Pkg/0.1@lasote/testing -r default", assert_error=True)
-    assert "ERROR: Remote 'default' is disabled" in client.out
+    client.run("install Pkg/0.1@lasote/testing -r default")
+    assert "Pkg/0.1@lasote/testing: Already installed!" in client.out
```
memsharded, line 367: Note this behavior change, could be considered a bug, why would fail if not really hitting the server at all?

lasote (approved):

Issue #9164, opened 2021-06-25
[feature] Continue working in the implementation of the Conan Cache for 2.0: --update flows


czoido:
The first implementation of the multi-revision Conan Cache for 2.0 is already merged: https://github.com/conan-io/conan/pull/8796

Following: https://github.com/conan-io/conan/issues/9076

Next step is implementing --update flows with the new cache to answer some questions like:
- What does it mean to --update in a multi-revision cache
- To what date the revisions are set when taken from the servers
- ...

Some pending features of the cache that should be implemented:

- Concurrent cache. Use of locks: multiprocess required, also could be multithread
- Multiple levels cache: Read-Only and Write cache
- LRU implementation
- Shortest paths as possible
- Deterministic folders

PR #9191, opened 2021-07-01
Implement --update flows for revisions in conan cache 2.0

czoido:
Changelog: Feature: Implement --update flows for conan cache 2.0.
Docs: missing (2.0)

This PR is focused on the logic of resolutions when we have multiple revisions in remotes and also multiple revisions in the cache when the version is already decided the proxy will take care of deciding which revision of that version we should take. The part of the logic related to the version ranges is here: https://github.com/conan-io/conan/pull/9350

- When we install a new revision we don't set the time of the installed revision to the current time which the installation takes place but the date the package has in the remote (this may change in the future, we can test if this model fits).

- Because we set the date in the local cache to the one from the remote it was considered adding a new `--make-latest` argument  to force that we want to update the date of that revision in the cache to make it the latest one. This may be part of commands like create or install like is implemented in this PR or may be moved to other command to manipulate revions?. **Finally this functionality will not be added, is a little bit speculative, let's wait to see if there's a real need for that.**

Closes: https://github.com/conan-io/conan/issues/9164

The proposed update flows are:

image


czoido:
image

conans/client/cmd/download.py
```diff
@@ -14,7 +14,7 @@ def download(app, ref, package_ids, remote, recipe, recorder, remotes):
     hook_manager.execute("pre_download", reference=ref, remote=remote)
 
     try:
-        ref = remote_manager.get_recipe(ref, remote)
+        ref, ref_time = remote_manager.get_recipe(ref, remote)
```
memsharded: Unused ref_time?

conans/client/conan_api.py
```diff
@@ -275,7 +275,7 @@ def inspect(self, path, attributes, remote_name=None):
                 except NotFoundException:
                     raise RecipeNotFoundException(ref)
                 else:
-                    ref = self.app.remote_manager.get_recipe(ref, remote)
+                    ref, ref_time = self.app.remote_manager.get_recipe(ref, remote)
```
memsharded: Unused ref_time?

conans/client/graph/graph.py
```diff
@@ -5,6 +5,7 @@
 RECIPE_DOWNLOADED = "Downloaded"
 RECIPE_INCACHE = "Cache"  # The previously installed recipe in cache is being used
 RECIPE_UPDATED = "Updated"
+RECIPE_INCACHE_DATE_UPDATED = "Cache (Updated date)"
```
memsharded, line 8: This doesn't sound it should be a recipe state, but I still need to understand all the details.

conans/client/graph/graph_binaries.py
```diff
@@ -60,33 +63,37 @@ def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
...
+    def _evaluate_cache_pkg(self, node, pref, remote, remotes, update):
         if update:
             output = node.conanfile.output
             if remote:
                 try:
-                    tmp = self._remote_manager.get_package_manifest(pref, remote)
-                    upstream_manifest, pref = tmp
+                    # if there's a later package revision in the remote we will take that one
+                    pkg_id = PackageReference(pref.ref, pref.id)
+                    remote_prevs = self._remote_manager.get_package_revisions(pkg_id, remote)
```
memsharded, line 58: We need to benchmark this a bit, this could be relatively slow and scale poorly on the server side.

conans/client/graph/graph_binaries.py
```diff
@@ -60,33 +63,37 @@ def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
...
+                    remote_latest_prev = PackageReference(pref.ref, pref.id,
+                                                          revision=remote_prevs[0].get("revision"))
+                    remote_latest_prev_time = from_iso8601_to_datetime(remote_prevs[0].get("time"))
+                    cache_time = datetime.fromtimestamp(self._cache.get_timestamp(pref), gettz())
```
memsharded: I think it would be better to compare directly timestamps, why the conversion?

conans/client/graph/graph_binaries.py
```diff
@@ -60,33 +63,37 @@ def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
...
                 except NoRemoteAvailable:
                     output.warn("Can't update, no remote defined")
                 else:
-                    package_folder = package_layout.package()
-                    if self._check_update(upstream_manifest, package_folder, output):
+                    if cache_time < remote_latest_prev_time and remote_latest_prev != pref:
                         node.binary = BINARY_UPDATE
-                        node.prev = pref.revision  # With revision
+                        node.prev = remote_latest_prev.revision
+                        output.warn("Current package is older than remote upstream one")
```
memsharded: Why warning? It got an update, seems the intended behavior.

conans/client/graph/graph_binaries.py
```diff
@@ -60,33 +63,37 @@ def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
...
+                    else:
+                        output.warn("Current package is newer than remote upstream one")
```
memsharded: Maybe better explicitly say ``Current package revision is newer than ...``

conans/client/graph/graph_binaries.py
```diff
@@ -238,13 +245,14 @@ def _process_node(self, node, pref, build_mode, update, remotes):
         if not remote:
             # If the remote_name is not given, follow the binary remote, or the recipe remote
             # If it is defined it won't iterate (might change in conan2.0)
-            remote_name = self._cache.get_remote(latest_prev_for_pkg_id.ref) if latest_prev_for_pkg_id else None
+            remote_name = self._cache.get_remote(
```
memsharded: Maybe?
```
remote_name = self._cache.get_remote(latest_prev_for_pkg_id or pref.ref)
```

conans/client/graph/proxy.py
```diff
@@ -59,67 +63,99 @@ def _get_recipe(self, reference, check_updates, update, remotes):
...
             status = RECIPE_INCACHE
+            return conanfile_path, status, cur_remote, ref
+
+    def _get_rrev_from_remotes(self, reference, remotes, check_all_servers):
+        results = []
+        output = ScopedOutput(str(reference), self._out)
+
+        # TODO: cache2.0 --update strategies: when we have specified the revision we don't want to
+        #  check all the remotes, just return the first match
+        output.info(f"Checking all remotes: ({', '.join([remote.name for remote in remotes])})")
```
memsharded: I agree that this sounds as a good functional behavior, but I am also concerned about the performance penalty of resolving graphs when there are several (lets say 5, 10) remotes defined, as this multiplies xN the slowest part of the resolution, which is doing API calls to remote servers.

Not something to change here, but something to talk with @lasote and discuss.

conans/client/graph/proxy.py
```diff
@@ -59,67 +63,99 @@ def _get_recipe(self, reference, check_updates, update, remotes):
...
+    def _download_recipe(self, ref, output, remotes, remote, check_all_servers=True,
+                         make_latest=False):
 
-        def _retrieve_from_remote(the_remote):
+        def _retrieve_from_remote(the_remote, reference):
             output.info("Trying with '%s'..." % the_remote.name)
             # If incomplete, resolve the latest in server
-            _ref = self._remote_manager.get_recipe(ref, the_remote)
+            _ref, _ref_time = self._remote_manager.get_recipe(reference, the_remote)
+            new_timestamp = datetime.timestamp(_ref_time) if not make_latest else time.time()
```
memsharded: This time conversion from the remote seems confusing, wouldn't it be possible to communicate with servers with timestamp and not datetimes?

memsharded (commented):
Good job!, lots of things that I still need to process, some other things to discuss.

memsharded:
Checking last scenario in the flow graphs: I don't understand the case for ``install ref#rrev --update`` when it is in the cache and returning only the first match of a remote, and not the latest one from all remotes, which is the behavior for ``install ref#rrev`` without update and remote args (all remotes). Seems asymmetrical:

When a ref#rrev is not in the cache:
- ``conan install ref#rrev --update`` will not obtain the latest from all remotes, only first match
- ``conan install ref#rrev`` will obtain the latest from all remotes

czoido:
> Checking last scenario in the flow graphs: I don't understand the case for `install ref#rrev --update` when it is in the cache and returning only the first match of a remote, and not the latest one from all remotes, which is the behavior for `install ref#rrev` without update and remote args (all remotes). Seems asymmetrical:
> 
> When a ref#rrev is not in the cache:
> 
> * `conan install ref#rrev --update` will not obtain the latest from all remotes, only first match
> * `conan install ref#rrev` will obtain the latest from all remotes

Following the flow of the [diagram](https://github.com/conan-io/conan/pull/9191#issuecomment-882256402), the result is the same when a `ref#rrev` is not in the cache:
* `conan install ref#rrev --update` will not obtain the latest from all remotes, only first match
* `conan install ref#rrev` will not obtain the latest from all remotes, only first match

and when `ref#rrev` is in the cache:
* `conan install ref#rrev --update` check all remotes and update date of revision in the cache if there's a newer date for the revision (not downloading anything as we consider revisions inmutable) otherwise do nothing.
* `conan install ref#rrev` do nothing

conans/client/graph/graph.py
memsharded: Ok, it is good atm, lets leave it.

czoido:
> > Checking last scenario in the flow graphs: I don't understand the case for `install ref#rrev --update` when it is in the cache and returning only the first match of a remote, and not the latest one from all remotes, which is the behavior for `install ref#rrev` without update and remote args (all remotes). Seems asymmetrical:
> > When a ref#rrev is not in the cache:
> > 
> > * `conan install ref#rrev --update` will not obtain the latest from all remotes, only first match
> > * `conan install ref#rrev` will obtain the latest from all remotes
> 
> Following the flow of the [diagram](https://github.com/conan-io/conan/pull/9191#issuecomment-882256402), the result is the same when a `ref#rrev` is not in the cache:
> 
> * `conan install ref#rrev --update` will not obtain the latest from all remotes, only first match
> * `conan install ref#rrev` will not obtain the latest from all remotes, only first match
> 
> and when `ref#rrev` is in the cache:
> 
> * `conan install ref#rrev --update` check all remotes and update date of revision in the cache if there's a newer date for the revision (not downloading anything as we consider revisions inmutable) otherwise do nothing.
> * `conan install ref#rrev` do nothing

Finally it's clear we have to change the flow, if revision is not in cache and you add the --update argument we have to check all remotes.

conans/client/remote_manager.py
```diff
@@ -99,12 +101,18 @@ def get_recipe(self, ref, remote):
...
+        except ConanReferenceDoesNotExist:
+            layout = self._cache.create_ref_layout(ref)
+
         layout.export_remove()
 
         t1 = time.time()
         download_export = layout.download_export()
         zipped_files = self._call_remote(remote, "get_recipe", ref, download_export)
+        remote_revisions = self._call_remote(remote, "get_recipe_revisions", ref)
+        ref_time = from_iso8601_to_timestamp(remote_revisions[0].get("time"))
```
memsharded: It seems the server is returning an iso date instead of the raw timestamp? Looks like something to change in the server API?

conans/client/graph/graph_binaries.py
memsharded: Please check ⬆️

conans/client/graph/proxy.py
```diff
@@ -62,67 +71,95 @@ def _get_recipe(self, reference, check_updates, update, remotes):
...
+        output = ScopedOutput(str(reference), self._out)
+
+        # TODO: cache2.0 --update strategies: when we have specified the revision we don't want to
+        #  check all the remotes, just return the first match
+        output.info(f"Checking all remotes: ({', '.join([remote.name for remote in remotes])})")
+
+        for remote in remotes:
+            try:
+                output.info(f"Checking remote: {remote.name}")
+                remote_rrevs = self._remote_manager.get_recipe_revisions(reference, remote)
```
memsharded: If this is a "getting latest" use case that could be implemented by the remote API, lets hide it at least behind the ``remote_manager``, instead of computing here explicitly.

conans/client/remote_manager.py
memsharded: Maybe at least also hide it behind the remote call (implement it in the rest api)?

memsharded (approved):
Seems this could be merged and continue work from there, I think it is needed a lot of architectural work to make things more maintainable and specially understandable, I am struggling myself to follow the flows in code.

conans/client/graph/proxy.py
```diff
@@ -62,67 +68,95 @@ def _get_recipe(self, reference, check_updates, update, remotes):
...
-                    return conanfile_path, status, selected_remote, ref
+            remote, latest_rrev, remote_time = self._get_rrev_from_remotes(reference,
+                                                                           remotes.values(),
+                                                                           check_all_servers=True)
+            if latest_rrev:
+                # check if we already have the latest in local cache
+                # TODO: cache2.0 here if we already have a revision in the cache but we add the
+                #  --update argument and we find that same revision in server, we will not
+                #  download anything but we will UPDATE the date of that revision in the
+                #  local cache and WE ARE ALSO UPDATING THE REMOTE
```
lasote, line 86: I don't remember how the timestamp in the sever was assigned but... wasn't it automatically assigned with the server time at the time you upload a revision? How could we force a timestamp in the server based on a local one? I'm clearly missing something.

conans/client/graph/proxy.py
lasote: Yep, we might opt-in/out that behavior, you could always use Virtual repositories.

lasote: note: we should benchmark virtual repos

conans/client/graph/proxy.py
czoido: Sorry, the sentence is a little bit confusing, I'm referring to update the remote associated to the reference in the database (what previously was stored in the metadata). In this PR only local timestamps are changes because as you say we can't modify the remote ones.

memsharded (commented):
Please rebase from develop2, the range_resolver got merged, this PR should probably be reduced.

conan/cache/db/references.py
```diff
@@ -62,106 +57,68 @@ def _set_clause(self, ref: ConanReference, path=None, timestamp=None, remote=Non
...
         query = f"DELETE FROM {self.table_name} " \
                 f"WHERE {where_clause};"
-        r = conn.execute(query)
+        r = self._conn.execute(query)
         return r.lastrowid
 
-    def all(self, conn, only_latest_rrev=False):
+    def all(self, only_latest_rrev=False):
         if only_latest_rrev:
             query = f'SELECT DISTINCT {self.columns.reference}, ' \
```
lasote, line 124: Why using `DISTINCT` if you are already doing `GROUP BY` the reference?

lasote: Why selecting `MAX({self.columns.timestamp})` if you are ordering the timestamp DESC? Shouldn't it return only the latest already?

lasote: What is the meaning of `{self.columns.prev} IS NULL` ? binary packages?

conans/test/functional/revisions_test.py
```diff
@@ -88,21 +72,25 @@ def test_update_recipe_iterating_remotes(self):
             other_v2.create(self.ref, conanfile=conanfile)
         other_v2.upload_all(self.ref, remote="remote2")
 
+        remote2_latest_prev = other_v2.get_latest_prev(self.ref)
+
         # Install, it wont resolve the remote2 because it is in the registry, it will use the cache
+        # doing --update it will find a newer same revision in remote2 and update the date and server
+        # of the revision in the local cache
+        # TODO: cache2.0 we get the recipe from remote 2 but package rev still is the one from
+        #  is this what we want?
```
lasote: Check the TODO comment, is it still valid?

conans/client/graph/proxy.py
```diff
@@ -62,67 +68,95 @@ def _get_recipe(self, reference, check_updates, update, remotes):
...
+                        status = RECIPE_NEWER
                 else:
-                    status = RECIPE_UPDATEABLE
+                    # TODO: cache2.0 we are returning RECIPE_UPDATED just because we are updating
+                    #  the date
+                    if cache_time >= remote_time:
+                        status = RECIPE_INCACHE
+                    else:
+                        selected_remote = remote
+                        remotes.select(remote.name)
```
memsharded: This ``remotes.select()`` here seems very weird. I think the ``select()`` with only 1 remote as possible selection needs to dissapear. But also, this wouldn't be a place to define a "selection" of remotes, that is intended for user input, isn't it?

conans/client/graph/proxy.py
memsharded: This has been finally changed, right @czoido? now it only do this for ``--update``?

conans/client/remote_manager.py
```diff
@@ -99,12 +100,18 @@ def get_recipe(self, ref, remote):
         self._hook_manager.execute("pre_download_recipe", reference=ref, remote=remote)
 
         ref = self._resolve_latest_ref(ref, remote)
-        layout = self._cache.create_ref_layout(ref)
+        try:
```
memsharded: I think this might be worth moving this in a single call in a ``ClientCache`` class method, (so it is cleaner to optimize it later if necessary, moving down the ``DataCache`` as a single DB query). The ``RemoteManager`` shouldn't be dealing with this Exception, this responsibility doesn't belong here.

conans/client/remote_manager.py
```diff
@@ -157,7 +164,12 @@ def get_package(self, conanfile, pref, remote, output, recorder):
 
         output.info("Retrieving package %s from remote '%s' " % (pref.id, remote.name))
         latest_prev = self.get_latest_package_revision(pref, remote)
-        pkg_layout = self._cache.create_pkg_layout(latest_prev)
+
+        try:
```
memsharded: Same as above.

conans/client/rest/rest_client_v2.py
```diff
@@ -293,6 +294,9 @@ def remove_conanfile(self, ref):
     def get_recipe_revisions(self, ref):
         url = self.router.recipe_revisions(ref)
         tmp = self.get_json(url)["revisions"]
+        # FIXME: the server API is returning an iso date, we have to convert to timestamp
```
memsharded, line 297: Please remember to annotate this in the doc for server potential changes.

conans/errors.py
```diff
@@ -102,6 +102,16 @@ def __str__(self):
         return exception_message_safe(msg)
 
 
+class ConanReferenceDoesNotExist(ConanException):
```
memsharded: If these are DB/Cache errors, maybe add DB/Cache to the name itself? It could read as a 404, server side missing package error.

conans/client/rest/rest_client_v2.py
czoido: It's already there 👍

conans/client/graph/proxy.py
czoido: Yes, checking all the remotes will only happen with `--update`, probably this message that will only make sense with that argument is redundant or should be changed, I'll remove it for the moment.

conan/cache/db/references.py
czoido: Yes, totally, `GROUP BY` and `MAX({self.columns.timestamp})` are redundant. I'll remove them.
This method lists only the different Conan references (name/version@user/channel) I'm going to change the name because it's confusing. That's why the `{self.columns.prev} IS NULL`, because it should not list all the binaries.

conans/test/functional/revisions_test.py
czoido: I don't think it's valid any more, I'll remove that.

conans/client/remote_manager.py
czoido: Moved to `get_or_create_ref_layout` and `get_or_create_pkg_layout` methods in ClientCache.

lasote (approved):

PR #9194, opened 2021-07-02
new win_bash mechanism

### LINKED ISSUE #8883: [bug] unix_path() uses WSL path flavor on Windows
When `conans.client.tools.win.unix_path()` is called and the path_flavor is left to be determined automatically,
it will call OSInfo.detect_windows_subsystem().
When this happens on a Windows 10 system with WSL installed, this will find C:\Windows\System32\bash.exe,
run `uname -a` inside WSL and identify the subsystem as WSL, and choose the WSL path flavor.
Example of the generated conanbuildinfo.txt:

```
[ENV_xapian-core]
AUTOMAKE_CONAN_INCLUDES=["/mnt/d/conan_data/xapian-core/1.4.16/_/_/package/d140711d95cc16a85766a8fc3a551dfafe84cf63/bin/share/aclocal"]
PATH=["D:\conan_data\xapian-core\1.4.16\_\_\package\d140711d95cc16a85766a8fc3a551dfafe84cf63\bin"]
```

This also makes `conan install` run slower than it needs to when it has to wait for WSL startup before misidentifying the subsystem.

  * Operating System+version: Windows 10 19042.928
  * Compiler+version: Visual Studio 2019 16
  * Conan version: 1.35.2 - 1.36 develop
  * Python version: 3.9.1 x64

`conan install xapian-core/1.4.16`

### LINKED ISSUE #9068: [feature] win_bash new approach
The current win_bash and run_in_windows_bash needs some improvements. This ticket would be the first iteration for improving that, to better serve the new build system integrations and for Conan 2.0.
Important points:

- It should have strictly 0 auto-detection. All information that it would need should be extracted from settings or new [conf]
- It should have a simple ``win_bash=True`` syntax, not necessary to do ugly checks in recipes about the platform. Internally, it should do nothing if not in Windows.
- It should use strictly the 2 profiles approach.
- Consider skipping using the WSL as a win_bash, as that would be a Linux environment, not a Windows one?
- ``tools.run_in_windows_bash()`` can probably be dropped if nicely supported in ``self.run(..., win_bash=True)``
- No need to integrate it with the existing build integrations, or to be backwards compatible, it can be treated as a completely new feature to replace the previous one.

Start simple, probably just the ``bash`` wrapper itself, but not necessarily handling all the possible environment and path (cygwin, and the like) variants.

Related issue: https://github.com/conan-io/conan/issues/8883, difficult to solve without breaking...

lasote:
Changelog: Feature: New `self.win_bash` mechanism to enable running commands in a bash shell in Windows. It works only with the new environment definition from the dependencies (`env_buildinfo` and `run_buildinfo`) as long as the new `AutotoolsToolchain`, `AutotoolsDeps` and `Autotools` build helper. It supports automatic conversion of the environment variables values declared as "path" according to the declared subsystem in the conf `tools.win.bash:subsystem`, that is not being auto-detected anymore.
Docs: https://github.com/conan-io/docs/pull/2152

Closes #8883
Closes #9068


conan/tools/env/__init__.py
```diff
@@ -1,3 +1,3 @@
-from conan.tools.env.environment import Environment
+from conan.tools.env.environment import Environment, save_script, register_environment_script
```
memsharded: Is the register_environment_script intended for users? I see ``save_script`` used by some test, but not this one.

conan/tools/microsoft/__init__.py
```diff
@@ -2,3 +2,4 @@
 from conan.tools.microsoft.msbuild import MSBuild
 from conan.tools.microsoft.msbuilddeps import MSBuildDeps
 from conan.tools.microsoft.visual import msvc_runtime_flag
+from conan.tools.microsoft.win import unix_path
```
memsharded: Is this exposed to users?

conan/tools/microsoft/win.py
```diff
@@ -0,0 +1,150 @@
...
+    """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
+    if env:
+        # Passing env invalidates the conanfile.environment_scripts
+        env_win = [env] if not isinstance(env, list) else env
+        env_shell = []
+    else:
+        env_shell = [f for f in conanfile.environment_scripts if f.lower().endswith(".sh")]
+        env_win = env or [f for f in conanfile.environment_scripts if f.lower().endswith(".bat")]
+
+    subsystem = conanfile.conf["tools.win.bash:subsystem"]
```
memsharded: Better follow import path ``tools.microsoft.bash:subsystem``?

conan/tools/microsoft/win.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    quoted_arg = subprocess.list2cmdline([command])
+    return "".join(["^%s" % arg if arg in r'()%!^"<>&|' else arg for arg in quoted_arg])
+
+
+def unix_path(conanfile, path):
+    """"Used to translate windows paths to MSYS unix paths like
+    c/users/path/to/file. Not working in a regular console or MinGW!"""
+    if not path:
+        return None
```
memsharded: Someone not passing a path or path None? Shouldn`t it raise or let it error later, or assert?

conan/tools/microsoft/win.py
```diff
@@ -0,0 +1,150 @@
+import os
```
memsharded: Maybe rename the file to ``subsystems.py``?

conans/model/conan_file.py
```diff
@@ -387,16 +391,21 @@ def package_info(self):
...
-        def _run():
-            if not win_bash:
-                return self._conan_runner(command, output, os.path.abspath(RUN_LOG_NAME), cwd)
+        def _run(cmd, _env):
             # FIXME: run in windows bash is not using output
-            return tools.run_in_windows_bash(self, bashcmd=command, cwd=cwd, subsystem=subsystem,
-                                             msys_mingw=msys_mingw, with_login=with_login)
+            if platform.system() == "Windows":
+                if win_bash:
+                    return tools.run_in_windows_bash(self, bashcmd=cmd, cwd=cwd)
```
memsharded: You are removing the ``with_login``, etc, this could be breaking. Shouldn't the call be identical to what it was?

memsharded (commented):
Doesn't look that bad on the surface, indeed a couple of ugly definitions, like propagating the conanfile to _EnvItem, but maybe worth the value of the proposal.

conan/tools/microsoft/__init__.py
lasote: Yes. I think this is still useful, The old one one public too: https://docs.conan.io/en/latest/reference/tools.html?highlight=unix_path#tools-unix-path

conan/tools/microsoft/win.py
lasote: Yup. copy paste from old one. I'll fix it

conans/model/conan_file.py
lasote: My mistake, I also removed by mistake `subsystem=subsystem, msys_mingw=msys_mingw, with_login`. I'll fix it

conans/test/functional/toolchains/gnu/autotools/test_win_bash.py
```diff
@@ -0,0 +1,78 @@
...
+        from conan.tools.env import save_script, Environment
+
+        class TestConan(ConanFile):
+            settings = "os", "compiler", "arch", "build_type"
+            exports_sources = "configure.ac", "Makefile.am", "main.cpp"
+            generators = "AutotoolsToolchain"
+            win_bash = True
+
+            def generate(self):
+                # Add vcvars launcher
```
lasote, line 43: @memsharded Please take a look at these two FIXME in case we want to add more functionality now.

lasote:
Now it is not passing conanfile to `_EnvValue` only to needed methods.

conan/tools/env/environment.py
```diff
@@ -285,15 +306,22 @@ def __init__(self):
...
         """
-        result = Environment()
+        result = Environment(conanfile)
         for pattern, env in self._environments.items():
             if pattern is None or fnmatch.fnmatch(str(ref), pattern):
                 result = env.compose(result)
+
+        # FIXME: Ugly
+        # Fill the conanfile
+        result._conanfile = conanfile
```
memsharded, line 334: No longer necessary?

conan/tools/microsoft/__init__.py
```diff
@@ -2,3 +2,4 @@
 from conan.tools.microsoft.msbuild import MSBuild
 from conan.tools.microsoft.msbuilddeps import MSBuildDeps
 from conan.tools.microsoft.visual import msvc_runtime_flag
+from conan.tools.microsoft.subsystems import unix_path
```
memsharded, line 5: Quick question, are the subsystems a ``microsoft`` thing, like ``msvc``, or a ``gnu`` thing?

conan/tools/microsoft/subsystems.py
```diff
@@ -0,0 +1,149 @@
...
+MSYS = 'msys'
+CYGWIN = 'cygwin'
+WSL = 'wsl'  # Windows Subsystem for Linux
+SFU = 'sfu'  # Windows Services for UNIX
+
+
+def run_in_windows_shell(conanfile, command, cwd=None, env=None):
+    """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
+    if env:
+        # Passing env invalidates the conanfile.environment_scripts
```
memsharded, line 20: Maybe call it ``env_scripts``, at first sight I thought it was requiring an environment or an env-var dict

conan/tools/microsoft/subsystems.py
```diff
@@ -0,0 +1,149 @@
...
+    wrapped_user_cmd = command
+    if env_shell:
+        # Wrapping the inside_command enable to prioritize our environment, otherwise /usr/bin go
+        # first and there could be commands that we want to skip
+        wrapped_user_cmd = environment_wrap_command(conanfile, env_shell, command,
+                                                    cwd=conanfile.generators_folder)
+    inside_command = 'cd "{cwd_inside}" && ' \
+                     '{wrapped_user_cmd}'.format(cwd_inside=cwd_inside,
+                                                 wrapped_user_cmd=wrapped_user_cmd)
+    if platform.system() == "Windows":
```
memsharded: This check for Windows, wasn't the above one enough?

conans/test/functional/toolchains/gnu/autotools/test_win_bash.py
memsharded: If it is going to be exposed to public already, then maybe yes, add a generator like class. But can be done in a later PR.

conans/test/integration/environment/test_env.py
```diff
@@ -171,12 +173,11 @@ def test_profile_buildenv():
...
         import os, platform
         from conans import ConanFile
+        from conan.tools.env import save_script
         class Pkg(ConanFile):
             def generate(self):
-                if platform.system() == "Windows":
-                    self.buildenv.save_bat("pkgenv.bat")
-                else:
-                    self.buildenv.save_sh("pkgenv.sh")
+                save_script(self, self.buildenv, "pkgenv")
```
memsharded: as the ``Environment`` is always an arg, maybe it is a bit more idiomatic (and require 1 import less):

```
self.buildenv.save_script(self, "pkgenv")
```

memsharded (approved):
There are some minor comments, but this is looking great, lets discuss those minor things and lets merge it.

conan/tools/microsoft/__init__.py
lasote: Hard to know, I would see it more "Microsoft" than "gnu". Take into account that not every "linux/unix" thing is "gnu", not even package config is really gnu.

conans/test/functional/toolchains/gnu/autotools/test_win_bash.py
lasote: Ok. created https://github.com/conan-io/conan/issues/9213

conans/test/integration/environment/test_env.py
lasote: The conanfile is not needed either, because it is always there.

memsharded (approved):
Great, lets merge it!

PR #9195, opened 2021-07-02
--require-override in cli

### LINKED ISSUE #2140: override requirements in conan install
In the docs you explain requirement overriding. I guess that can be used to handle diamond dependency conflicts. You force existing packages to use different versions of their requirements. 

But what if a library needs to be rebuilt to correctly use different versions of its requirements? We would need to create different packages of that library. We can handle that by adding an option and using conditional requirements in the conanfile. Wouldn't it be nice to be able to override the requirements the same way you override options and settings in the conan install command? Someting like:

`conan install . --requires="LibA/0.1@demo/testing"`

memsharded:
Changelog: Feature: Introduce the ``-require-override`` argument to define dependency overrides directly on command line.
Docs: https://github.com/conan-io/docs/pull/2170

Close https://github.com/conan-io/conan/issues/2140

Implementation note: It is a bit dirty that for implementing this kind of feature, 95% of the code is just passing around arguments that add to function calls that already contains +10 arguments, we need to revisit this architecture.

### DIFF
--- a/conans/client/cmd/create.py
+++ b/conans/client/cmd/create.py
@@ -28,7 +28,7 @@ def _get_test_conanfile_path(tf, conanfile_path):
 
 def create(app, ref, graph_info, remotes, update, build_modes,
            manifest_folder, manifest_verify, manifest_interactive, keep_build, test_build_folder,
-           test_folder, conanfile_path, recorder, is_build_require=False):
+           test_folder, conanfile_path, recorder, is_build_require=False, require_overrides=None):
     assert isinstance(ref, ConanFileReference), "ref needed"
     test_conanfile_path = _get_test_conanfile_path(test_folder, conanfile_path)
 
@@ -76,7 +76,9 @@ def create(app, ref, graph_info, remotes, update, build_modes,
                                    manifest_interactive=manifest_interactive,
                                    keep_build=keep_build,
                                    test_build_folder=test_build_folder,
-                                   recorder=recorder)
+                                   recorder=recorder,
+                                   require_overrides=require_overrides
+                                   )
     else:
         deps_install(app=app,
                      ref_or_path=ref,
@@ -92,4 +94,5 @@ def create(app, ref, graph_info, remotes, update, build_modes,
                      update=update,
                      keep_build=keep_build,
                      recorder=recorder,
-                     is_build_require=is_build_require)
+                     is_build_require=is_build_require,
+                     require_overrides=require_overrides)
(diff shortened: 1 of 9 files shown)


conans/client/cmd/create.py
```diff
@@ -28,7 +28,7 @@ def _get_test_conanfile_path(tf, conanfile_path):
 
 def create(app, ref, graph_info, remotes, update, build_modes,
            manifest_folder, manifest_verify, manifest_interactive, keep_build, test_build_folder,
-           test_folder, conanfile_path, recorder, is_build_require=False):
```
gsemet, line 31: Does it also overrides build_requirements ?

memsharded: Only normal requires in this proposal. ``build_requires`` are not overriden in the same way, they don't have the same ``override`` concept. When you add a ``[build_requires]`` in your profile, it will inject that build_requires to dependencies, always, not just overriding the existing versions if they exist. It might be possible to overwrite the ``profile`` build_requires on the fly as well, but it should be clear that the behavior is quite different to this feature.

lasote (approved):
Apart from your comment about the architecture that it is totally true, I see it fine.
It could be merged.

ytimenkov:
Nice feature! 👍
I wonder why there is no similar functionality for `lock create` command where it could make sense with version ranges?

memsharded:
Hi @ytimenkov 

Yes, we wanted to roll it out for the basics (install/create), if everything is working fine, yes, it could make sense for lockfiles too. I will open a ticket from your comment.

puetzk:
any chance we could also see a `[require-override]` section in (host) profiles? Most of the other things one can set on the CLI (e.g. `--settings`, `--options`, `--env`, `--conf`) have a counterpart in profiles, and what I would mainly like to use `--require-override` for is forcing things to use "system" library versions whose runtime is already baked into a target's sysroot. So setting such overrides along with the toolchain, build_requires, and other such things to cross-compile to a particular sysroot would be very natural and convenient...

memsharded:
Hi @puetzk 

> any chance we could also see a [require-override] section in (host) profiles?

I think it seems a bit not aligned the initial proposal of this feature. This was born as a feature to not have to edit files to change/override requirements, and being able to do that in the command line. If we are allowing to change requirements, isn't it better to do it in the conanfile? that is basically the main purpose of the conanfile, a place to define the dependencies. Wdyt @lasote?

lasote:
I think in that case (with the sysroot), might be more reproducible use the `requirements(self)` method in the recipe conditioning the requirements based on settings or options. Not exactly the same "use case" for what the command-line argument has been introduced.

PR #9197, opened 2021-07-02
trying to understand layouts 2

memsharded:
Changelog: Omit
Docs: Omit

Experimenting with the cache 2.0 and layouts:

It is necessary to clear express intention ``cache.ref_layout`` and ``cache.get_ref_layout`` are confusing. Lets go with something a bit more explicit, maybe something like: 
- ``cache.temp_ref_layout()`` For rrev=None, exporting with a temporary rrev
- ``cache.create_ref_layout()`` For rrev!=None, downloading or installing something to the cache, that is not there yet, but we should already know the rrev from the server.
- ``cache.ref_layout()``: It must exist already in the cache and rrev!=None

Let me know if this makes sense, or what problems could it have.

### DIFF
--- a/conans/client/graph/proxy.py
+++ b/conans/client/graph/proxy.py
@@ -51,7 +51,7 @@ class ConanProxy(object):
         # NOT in disk, must be retrieved from remotes
         if not ref:
             remote, new_ref = self._download_recipe(reference, output, remotes, remotes.selected)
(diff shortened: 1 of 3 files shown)


czoido:
I agree with this, I also realised it's a bit confusing, better make it more explicit.

czoido (approved):

PR #9211, opened 2021-07-06
More explicit layout handling for cache 2.0.

czoido:
Changelog: Feature: More explicit layout handling for cache 2.0.
Docs: omit

Using:
- ``cache.create_temp_ref_layout()`` For rrev=None, exporting with a temporary rrev
- ``cache.create_ref_layout()`` For rrev!=None, downloading or installing something to the cache, that is not there yet, but we should already know the rrev from the server.
- ``cache.ref_layout()``: It must exist already in the cache and rrev!=None

Adding some tests to xfail related with the fact that conan download did not check if the downloaded package where already present in the cache. This will change for 2.X because we consider revisions to be inmutable so it will be fixed in an upcoming PR.

Related to: https://github.com/conan-io/conan/pull/9197


conan/cache/cache.py
```diff
@@ -51,60 +51,70 @@ def base_folder(self):
...
+        return PackageLayout(pref, os.path.join(self.base_folder, package_path))
 
-        package_path, created = self._db.get_or_create_reference(package_path, pref)
-        self._create_path(package_path, remove_contents=created)
+    def create_reference_layout(self, ref: ConanReference):
+        assert ref.rrev, "Recipe revision must be known to create the package layout"
+        ref = ConanReference(ref.name, ref.version, ref.user, ref.channel, ref.rrev,
+                             ref.pkgid, ref.prev)
+        reference_path = self._get_path(ref)
+        self._db.get_or_create_reference(reference_path, ref)
```
memsharded: Maybe this should also be a ``db.create_reference``, to ensure nothing was there in the DB for this reference.

memsharded (commented):
Looking much better!

memsharded (approved):

PR #9217, opened 2021-07-06
poc of new alias

### LINKED ISSUE #8966: [bug] Alias resolution bug when mixing fix-version and alias

memsharded:
Changelog: Feature: Implement a new ``requires = "pkg/(alias)"`` syntax to be able to dissambiguate alias requirements and resolve them earlier in the flow, solving some limitations of the previous alias definition. This approach is intended to be the one in Conan 2.0 (issue backported from https://github.com/conan-io/tribe/pull/25).
Docs: https://github.com/conan-io/docs/pull/2169

Close https://github.com/conan-io/conan/issues/8966

From https://github.com/conan-io/conan/pull/9039, trying a proof of concept, if the new ``pkg/(alias)`` syntax could help addressing problems and help to transition smoothly to Conan 2.0

### DIFF
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -249,6 +249,7 @@ class DepsGraph(object):
         self.nodes = set()
         self.root = None
         self.aliased = {}
+        self.new_aliased = {}
         self._node_counter = initial_node_id if initial_node_id is not None else -1
 
     def add_node(self, node):
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -94,6 +94,8 @@ class DepsGraphBuilder(object):
             # TODO: Add info about context?
             graph_lock.lock_node(node, build_requires, build_requires=True)
 
+        for require in build_requires:
+            self._resolve_alias(node, require, graph, update, update, remotes)
         self._resolve_ranges(graph, build_requires, scope, update, remotes)
 
         for br in build_requires:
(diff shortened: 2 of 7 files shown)


lasote (approved):
Looking good! it is clearly opt-in so I don't think we can break something.

franramirez688 (approved):
I think it's good enough for a POC! 😁  It's done by well-separated functions so it could be easily removed in case of any problem.

czoido (approved):

PR #9218, opened 2021-07-06
Skip cmake.test and meson.test using [conf] profile

### LINKED ISSUE #8454: [feature] Using [conf] block in profiles to skip running tests as part of build()
The new [[conf] block](https://blog.conan.io/2021/01/27/New-conan-release-1-33.html) added to profiles in Conan 1.33 seems like a great addition that has been missing so far, so thanks a lot for adding this!

I think it might be useful to allow users to skip running tests as part of the build. The official recommendation is to run tests as part of `build()` like this :
```python
def build(self):
    cmake = CMake(self)
    cmake.configure()
    cmake.build()
    cmake.test()
```

Unfortunately, with larger test suites or tests that take long to run, it can get impractical when you just want to quickly perform a rebuild. Since we wanted to have an easy way of building without running tests, we are currently having our unit tests in test_package but this only works when your tests do not depend on any internal headers/APIs (and it also comes with other issues).

Now with [conf] available, I'd like to give this a try:
```python
def build(self):
    cmake = CMake(self)
    cmake.configure()
    cmake.build()
    if not self.conf["user:skip_tests"]:
        cmake.test()
```

What might be a problem though: currently these conf values can only be set as part of a profile and not on the CLI. To be really useful for enabling/disabling tests, there would need to be a way to specify these conf values as parameters to `conan build`.

uilianries:
Not only CMake offers `test`, but Meson also has the same method.

The original issue uses `skip_tests` (plural), but I think `test` is better, because is same method name.

Changelog: Feature: Skip build helper test and using [conf].
Docs: https://github.com/conan-io/docs/pull/2154

close #8454

/cc @chausner


memsharded (commented):
I think this is interesting, but probably we should further standardize the "run-test" concepts. There is demand for something more built-in and more generic, also to create conditional requirements, etc. So maybe the ``cmake:skip_test`` and ``meson:skip_test`` are not the best, and we should define something general, in the line of ``tools.build:skip_tests``. Or maybe the default should be to skip tests and tests are only enabled on demand? ``tools.build:tests``?

Lets see what others think.

uilianries:
> tools.build:tests

I like it. More generic and can be re-used for other helpers.

memsharded (commented):
I like this.

My main question is what is the right default for the future:

- Recipes, including if we want ConanCenter can add building and running tests support.
- By default tests do not build and run (good default for ConanCenter). It is possible to ``[conf]`` sometihng like ``tools.build:tests=True``, and tests will build and run
- By default tests build and run (bad default for ConanCenter, probably the ``tools.build:skip_tests=True`` should be enforced by CI, and hooks).

Wdyt @lasote?

czoido (approved):

memsharded (approved):

PR #9221, opened 2021-07-07
Check for revisions already in cache for conan download in Conan 2.0

czoido:
Changelog: Feature: Conan download checks for already existing recipe or package revisions in cache when downloading from a remote instead of trying to install the recipe without checking.
Docs: missing (2.0)


conans/client/cmd/download.py
```diff
@@ -14,9 +14,12 @@ def download(app, ref, package_ids, remote, recipe, recorder, remotes):
     hook_manager.execute("pre_download", reference=ref, remote=remote)
 
     try:
-        ref = remote_manager.get_recipe(ref, remote)
+        _, ref = remote_manager.get_recipe_manifest(ref, remote)
```
memsharded: We should probably avoid file transfers just to get the revision. Isn't there any other endpoint better than the manifest. I think in Conan 2.0 the manifest retrieval can actually be dropped in whole codebase.

conans/client/cmd/download.py
```diff
@@ -42,13 +45,21 @@ def download(app, ref, package_ids, remote, recipe, recorder, remotes):
 
 def _download_binaries(conanfile, ref, package_ids, cache, remote_manager, remote, output,
                        recorder, parallel):
-    short_paths = conanfile.short_paths
 
     def _download(package_id):
         pref = PackageReference(ref, package_id)
+        # TODO: cache2.0 here when specifying a package_id with conan download
```
memsharded: But this doesn't look right, when we query to the server for package_ids, don't we get the revisions too?

conans/client/cmd/download.py
```diff
@@ -48,11 +49,15 @@ def _download_binaries(conanfile, ref, package_ids, cache, remote_manager, remot
 
     def _download(package_id):
         pref = PackageReference(ref, package_id)
-        # TODO: cache2.0 here when specifying a package_id with conan download
-        #  we are not checking the remote for the latest package revision
-        #  maybe we want to add an argument to conan download like --update
-        check_pref_cache = cache.get_package_revisions(pref)
-        skip_download = (len(check_pref_cache) > 0)
+        try:
+            if not pref.revision:
```
memsharded, line 53: It is impossible that pref has revision here, you just constructed it above, empty revision?

conans/client/cmd/download.py
```diff
@@ -14,9 +14,13 @@ def download(app, ref, package_ids, remote, recipe, recorder, remotes):
     hook_manager.execute("pre_download", reference=ref, remote=remote)
 
     try:
-        ref = remote_manager.get_recipe(ref, remote)
+        if not ref.revision:
+            ref = remote_manager.get_latest_recipe_revision(ref, remote)
     except NotFoundException:
         raise RecipeNotFoundException(ref)
+    else:
+        ref_in_cache = cache.get_recipe_revisions(ref)
```
memsharded: This ``cache.get_recipe_revision(ref)`` when ref contains revision, and that is supposed to return just 1 revision, the passed one, is a bit confusing to read. Maybe is worth to add an explicit ``bool recipe_revision_exist(ref)`?

memsharded (approved):

memsharded (approved):

PR #9223, opened 2021-07-07
AutotoolsDeps multi level customization

### LINKED ISSUE #9176: [feature] New AutotoolsDeps customization
The `AutotoolsDeps` generator should precalculate things in the constructor to allow the user to modify them before calling the `generate()` method.

lasote:

### DIFF
--- a/conan/tools/gnu/autotoolsdeps.py
+++ b/conan/tools/gnu/autotoolsdeps.py
@@ -1,5 +1,6 @@
 from conan.tools.env import Environment
 from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
+from conans.errors import ConanException
 from conans.model.new_build_info import NewCppInfo
 
 
@@ -9,9 +10,18 @@ class AutotoolsDeps:
         # alter some value
         self._conanfile = conanfile
         self._cpp_info = None
+        self._cached_flags = False
+        self._libs_flags = None
+        self._cpp_flags = None
+        self._ld_flags = None
+        self._cxx_flags = None
+        self._c_flags = None
+
+        self._env = None
 
     @property
     def cpp_info(self):
+        # The user can modify the cpp_info object before accessing the flags or the environment
         if self._cpp_info is None:
             self._cpp_info = NewCppInfo()
             for dep in self._conanfile.dependencies.host.values():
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/autotoolsdeps.py
```diff
@@ -21,43 +31,84 @@ def cpp_info(self):
...
-        env.append("CXXFLAGS", cxxflags)
-        env.append("CFLAGS", cflags)
-        return env
-
-    def generate(self, env=None, auto_activate=True):
-        env = env or self.environment()
-        env.save_script("conanautotoolsdeps", auto_activate=auto_activate)
+        if not self._env:
+            env = Environment(self._conanfile)
+            env.append("CPPFLAGS", self._cpp_flags or self.cpp_flags)
```
memsharded, line 105: This idiom here ``self._cpp_flags or self.cpp_flags`` with the second also computing the former is very confusing.

conans/test/unittests/client/toolchain/autotools/autotools_deps_test.py
```diff
@@ -85,5 +94,54 @@ def test_foo():
...
+             "compiler.libcxx": "libstdc++11",
+             "compiler.version": "7.1",
+             "cppstd": "17"})
+        deps = AutotoolsDeps(consumer)
+
+        # We can access the cpp_info and modify it
+        deps.cpp_info.includedirs = ["/other/includedir"]
+
+        # Later we can access to the flags and modify them
+        deps.cpp_flags.append("NewCppFlag")
```
memsharded, line 126: I am not sure that adding flags is a use case for ``AutotoolsDeps``, it should to go ``AutotoolsToolchain``.
I think that some control over the dependencies flags, like enabling/disabling/filter some or all of the dependencies flags might be useful or necessary, but adding a new flag? Probably not here?

memsharded (commented):
Yes, this code is a bit difficult to know what is doing, with the ``_check_env_already_calculated`` and the laziness. A

conans/test/unittests/client/toolchain/autotools/autotools_deps_test.py
lasote: Well, that is a test, the cpp_flags are the ones filled from the deps information.

lasote:
So? I totally agree the laziness is confusing, so any conclusion about what should we allow to customize?
Maybe just the `cpp_info` is enough and the more natural.

conan/tools/gnu/autotoolsdeps.py
lasote: Agree. Needed tricks to allow that level of customization. I went to the maximum level of customization so we can decide the level we want.

lasote:
After discussing I'm closing this PR to open a new one only allowing to change the environment.

PR #9225, opened 2021-07-07
[fix] CONAN_TOOLCHAIN_ARGS_FILE as configuration file

### LINKED ISSUE #9130: [fix] `CONAN_TOOLCHAIN_ARGS_FILE` is not very practical being a json file
The CONAN_TOOLCHAIN_ARGS_FILE will be frequently used to "store" arguments to pass to the build system in the "generate()" method. These arguments can contain paths and other things quoted. One of the principal use cases for the CONAN_TOOLCHAIN_ARGS_FILE is to be able to "cat" the file and paste the contents to call directly the build system in the command line, emulating exactly what Conan does. But the json format escapes the quotes, and you cannot directly copy the contents without editing it and that is a big pain.

Think about another format... ¿yml? ¿txt?...

franramirez688:
Changelog: Fix: Saving all the toolchain args information into `conanbuild.conf` instead of json file.
Docs: https://github.com/conan-io/docs/pull/2156

* Saving all the toolchain args information into `conanbuild.conf`.
* Using a default configuration section called `toolchain`.
* Using ConfigParser as internal mechanism to manage the load and saving of the toolchain-file content.

Closes https://github.com/conan-io/conan/issues/9130

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -195,13 +195,16 @@ def load_toolchain_args(generators_folder=None):
     args_file = os.path.join(generators_folder, CONAN_TOOLCHAIN_ARGS_FILE) if generators_folder \
         else CONAN_TOOLCHAIN_ARGS_FILE
     toolchain_config = configparser.ConfigParser()
-    toolchain_config.read(args_file)
+    toolchain_file = toolchain_config.read(args_file)
+    if not toolchain_file:
+        raise ConanException("The file %s does not exist. Please, make sure that it was not"
+                             " generated in another folder." % args_file)
     try:
         return toolchain_config[CONAN_TOOLCHAIN_ARGS_SECTION]
     except KeyError:
-        raise ConanException("Something was wrong reading the %s file. Review if the path is "
-                             "correct or the section [%s] exist" % (args_file,
-                                                                    CONAN_TOOLCHAIN_ARGS_SECTION))
+        raise ConanException("The primary section [%s] does not exist in the file %s. Please, add it"
+                             " as the default one of all your configuration variables." %
+                             (CONAN_TOOLCHAIN_ARGS_SECTION, args_file))
 
 
 def save_toolchain_args(content, generators_folder=None):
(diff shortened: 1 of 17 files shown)


conan/tools/files/files.py
```diff
@@ -185,13 +185,34 @@ def rename(conanfile, src, dst):
...
+def load_toolchain_args(generators_folder=None):
+    """
+    Helper function to load the content of any CONAN_TOOLCHAIN_ARGS_FILE
+
+    :param generators_folder: `str` folder where is located the CONAN_TOOLCHAIN_ARGS_FILE.
+    :return: <class 'configparser.SectionProxy'>
+    """
+    args_file = os.path.join(generators_folder, CONAN_TOOLCHAIN_ARGS_FILE) if generators_folder \
+        else CONAN_TOOLCHAIN_ARGS_FILE
+    if os.path.exists(args_file):
```
memsharded: Shouldn't it fail if it doesn't exist? I think it would be safer.

conan/tools/files/files.py
```diff
@@ -185,13 +185,34 @@ def rename(conanfile, src, dst):
...
+def save_toolchain_args(content, generators_folder=None):
+    """
+    Helper function to save the content into the CONAN_TOOLCHAIN_ARGS_FILE
+
+    :param content: `dict` all the information to be saved into the toolchain file.
+    :param generators_folder: `str` folder where is located the CONAN_TOOLCHAIN_ARGS_FILE
+    """
+    # Let's prune None values
+    content_ = {k: v for k, v in content.items() if v is not None}
+    if content_:
```
memsharded: Same, probably it is safer to create the file, even if empty.

memsharded (commented):
Looking good so far, please check comments.

conan/tools/files/files.py
franramirez688: I'm rethinking it and I'm not completely sure about it. Now, some tests are failing because it's not supposed to be any `conanbuild.conf` created... Are we sure that we want that behavior then? @lasote let us know your opinion, please

lasote: Yes, I would say it has to exist. Some build helpers need more that file than others but in general, if they read the file, the file should be there.

memsharded (approved):

lasote (approved):

PR #9228, opened 2021-07-07
Filename PKG-NAME_COMP-NAME.pc for PkgConfigDeps

### LINKED ISSUE #9185: Filename ``PKG-NAME_COMP-NAME.pc`` for PkgConfigDeps?
Isn't the name ``PKG-NAME_COMP-NAME.pc``? I think it could be better this way, to avoid potential conflicts with components of different packages. Or is not possible?

_Originally posted by @memsharded in https://github.com/conan-io/docs/pull/2133#discussion_r660957122_

franramirez688:
Changelog: Fix: Use filename `[PKG-NAME]-[COMP-NAME]` for `PkgConfigDeps`.
Docs: https://github.com/conan-io/docs/pull/2148
Closes https://github.com/conan-io/conan/issues/9185

Now, we don't have any conflict for components of different packages.
* `<COMP-NAME>.pc` -> `<PKG-NAME_COMP-NAME>.pc`

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -87,9 +87,11 @@ class PkgConfigDeps(object):
             if dep.new_cpp_info.has_components:
                 components = self._get_components(dep.ref.name, dep)
                 for comp_genname, comp_cpp_info, comp_requires_gennames in components:
-                    ret["%s.pc" % comp_genname] = self._pc_file_content(
-                        "%s-%s" % (pkg_genname, comp_genname),
-                        comp_cpp_info, comp_requires_gennames, dep.package_folder, dep.ref.version)
+                    pkg_comp_genname = "%s-%s" % (pkg_genname, comp_genname)
+                    ret["%s.pc" % pkg_comp_genname] = self._pc_file_content(
+                        pkg_comp_genname, comp_cpp_info,
+                        comp_requires_gennames, dep.package_folder,
+                        dep.ref.version)
                 comp_gennames = [comp_genname for comp_genname, _, _ in components]
                 if pkg_genname not in comp_gennames:
                     ret["%s.pc" % pkg_genname] = self._global_pc_file_contents(pkg_genname,
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -87,7 +87,7 @@ class PkgConfigDeps(object):
             if dep.new_cpp_info.has_components:
                 components = self._get_components(dep.ref.name, dep)
                 for comp_genname, comp_cpp_info, comp_requires_gennames in components:
-                    pkg_comp_genname = "%s_%s" % (pkg_genname, comp_genname)
+                    pkg_comp_genname = "%s-%s" % (pkg_genname, comp_genname)
                     ret["%s.pc" % pkg_comp_genname] = self._pc_file_content(
                         pkg_comp_genname, comp_cpp_info,
                         comp_requires_gennames, dep.package_folder,
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -87,7 +87,7 @@ class PkgConfigDeps(object):
             if dep.new_cpp_info.has_components:
                 components = self._get_components(dep.ref.name, dep)
                 for comp_genname, comp_cpp_info, comp_requires_gennames in components:
-                    pkg_comp_genname = "%s-%s" % (pkg_genname, comp_genname)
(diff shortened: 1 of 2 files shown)


memsharded (approved):

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -87,9 +87,11 @@ def content(self):
             if dep.new_cpp_info.has_components:
                 components = self._get_components(dep.ref.name, dep)
                 for comp_genname, comp_cpp_info, comp_requires_gennames in components:
-                    ret["%s.pc" % comp_genname] = self._pc_file_content(
-                        "%s-%s" % (pkg_genname, comp_genname),
-                        comp_cpp_info, comp_requires_gennames, dep.package_folder, dep.ref.version)
+                    pkg_comp_genname = "%s_%s" % (pkg_genname, comp_genname)
```
memsharded: While this seems correct, and much better, I am re-thinking that I don't know what pkg_config is expecting in terms of filenames, if the standard, expected .pc names is only the target, then this might not work? 

Asking for @SSE4 feedback.

SSE4: yes, I think it would be breaking if you change pkg-config names.
e.g. for vorbis, there are:
```
vorbis.pc
vorbisenc.pc
vorbisfile.pc
```
they don't follow the convention of `PKG-NAME_COMP_NAME.pc`.
for pkg-config, the name is important, as libraries use `pkg-config <pc_file_name>`
OTH, you may try to create additional pkg-config files, it should be harmless, as long as you don't overwrite others.

memsharded: Are ``enc`` and ``file`` components of vorbis? Is this the common default?

The thing is that this was previously generating ``enc.pc`` and ``files.pc`` for vorbis::enc and vorbis::files components, so we need to map these components names to something that makes sense and is understood by pkg_config.

What would be the general approach?

SSE4: there is no common pattern, some may use dashes like:
```
cairo-egl.pc
cairo-fc.pc
cairo-ft.pc
cairo-gl.pc
cairo-glx.pc
cairo-pdf.pc
cairo-png.pc
cairo-ps.pc
cairo-script.pc
cairo-svg.pc
cairo-tee.pc
cairo-xcb-shm.pc
cairo-xcb.pc
cairo-xlib-xrender.pc
cairo-xlib.pc
cairo.pc
```
some may use underscores, or other characters (maybe dot, colon, or whatever) to separate library and component names.
so I believe it's not a part of the standard, but a per-library choice.
I also suppose this information could be encoded in recipes, not in generator.

SSE4 (approved):

conan/tools/gnu/pkgconfigdeps.py
SSE4: some update (might be irrelevant):
usually, pkg-config files from different libraries are installed into the same location in the system (e.g. `/usr/lib/pkg-config`).
therefore, they cannot use `component.pc` pattern, is it's easy to collide between libraries (e.g. it's very high probability two libraries may have 'gl' component). that's why library name is also usually part of the .pc file.

memsharded: Ok, so lets change it to ``pkg-component.pc`` which seems slightly more common?
Please, @franramirez688 have a look to a populated ``/usr/lib/pkg-config`` folder in a linux distro, ad choose the most common pattern.

memsharded (changes requested):

conan/tools/gnu/pkgconfigdeps.py
SSE4: another (counter)example is glib:
```
/usr/lib64/pkgconfig/glib-2.0.pc
/usr/lib64/pkgconfig/gmodule-2.0.pc
/usr/lib64/pkgconfig/gmodule-export-2.0.pc
/usr/lib64/pkgconfig/gmodule-no-export-2.0.pc
/usr/lib64/pkgconfig/gobject-2.0.pc
/usr/lib64/pkgconfig/gthread-2.0.pc
```
they just use the `component-version.pc` model (without library name)...
so there is no common pattern, as I have stated previously
the only valid way to implement this as I can see Is to encode information in recipe (not generator)

memsharded: Yes, ok, but that is another problem, we would need to introduce a new property for this.
Lets get this merge with ``-``, it seems it is the most common thing, as this is already an improvement to the current approach, that is likely to conflict.

memsharded (approved):

franramirez688:
I just figured it out a little bit more and it seems to be more popular to use `pkg-component.pc` rather than `pkg_component.pc`

PR #9230, opened 2021-07-07
VCVars generator

### LINKED ISSUE #9213: [feature] Add a generator like class wrapping write_conanvcvars
Create a public generator `VCVars()` class to wrap the `write_conanvcvars` command.
After https://github.com/conan-io/conan/pull/9194
Change the conans/test/functional/toolchains/gnu/autotools/test_win_bash.py test

lasote:
Changelog: Feature: New `VCVars` generator that generates a `conanvcvars.bat` that activates the Visual Studio Developer Command Prompt.
Docs: https://github.com/conan-io/docs/pull/2153

Close #9213

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -19,7 +19,6 @@ class VCVars:
 
 
 def write_conanvcvars(conanfile, auto_activate=True):
-    # FIXME: Write a VCVars generator for the final user
     """
     write a conanvcvars.bat file with the good args from settings
     """
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -15,10 +15,10 @@ class VCVars:
         self._conanfile = conanfile
 
     def generate(self, auto_activate=True):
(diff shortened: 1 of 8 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -1,7 +1,65 @@
...
+
+class VCVars:
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+
+    def generate(self, auto_activate=True):
+        write_conanvcvars(self._conanfile, auto_activate=auto_activate)
+
+
+def write_conanvcvars(conanfile, auto_activate=True):
```
memsharded: I think if we are going to give the ``VCVars`` interface to users, lets use this as interface always, and hide the function ``write_conanvccars()``. Seems more consistent with other generators, and there is no issue in other generator just callling ``VCVars(conanfile).generate()`` and the ``auto_activate`` can be parameters in the constructor or attribute, to have the ``generate()`` interface match (no parameters anywhere).

memsharded (commented):
Looks good.
Just a minor detail about the UI, wdyt about it?

memsharded (approved):

Issue #8580, opened 2021-03-01
[feature] get profile name from lockfile


DukeOfKirkcaldy:
Very excited about the new bundles feature and have been trying to take your examples one step further and test it out in a CI flow where packages are rebuilt by triggering their respective CI jobs.

When triggering the build I should only need to share the bundle and lockfiles with the downstream build.  But one issue with this is the loss of the profile name, and so far the name has been quite useful for organisation purposes.  Naming the lockfile upload paths for example:

`${env.JOB_NAME}/${env.BUILD_NUMBER}/${name}/${version}/${user}/${channel}/${profile}/conan.lock`

I can understand why you might not want to store the name in the lockfile itself, but for me it would be quite useful to trace a lockfile back to a profile name to quickly know which configuration it was created for.  Perhaps you can detect if any existing profile matches the values stored in the lockfile and return the name?  Just an idea!

solvingj:
Hi @GordonJess , so, there's a related feature request here, and your input would be appreciated:

https://github.com/conan-io/conan/issues/7207

I will defer to @memsharded to decide whether or not to consider this feature for the future.  However, I will share my opinion, as someone who is working with lockfiles a lot. 

I think this feature would be exceptionally challenging and awkward to try to implement, and what could be delivered in the end would be extremely unreliable and unpredictable.  Here are a few complications:
- Users can pass multiple profiles
- Users can mix profiles and CLI args
- What the lockfile actually records is the "reconciled profile" (also labeled "Configuration" in `conan install` log). 

This will actually almost never match line-for-line or even string-for-string with a profile provided by the user. 

So, for these reasons, it seems un-reasonable to implement. 

With that said, the feature request I linked tries to provide exactly what you're describing but in a different way.  It could allow the user to actually include the `profile` and potentially even the `settings.yml` that were used to create the package.  Basically, anything a packager wishes. 

However, that is an extensive feature request, which requires server side changes. So rather than putting all my eggs in that basket, i'm also requesting an alternative feature which is more approachable. 

`conan lockfile extract profile`

The idea here is simply a first-class way to re-construct the profile from the lockfile so that it could be used in any number of workflows that people might come up with. I don't know if it addresses your actual situation at all (I suspect not actually), but please let me know either way.

memsharded:
Closing as outdated, as 2.0 lockfiles do not contain profiles anymore

PR #9235, opened 2021-07-07
Extract profiles from lockfiles

### LINKED ISSUE #8684: [feature] conan lock extract-profiles
In experimenting with workflows surrounding lockfiles, and in slack conversations several users have expressed interest or need in extracting the profile string from a lockfile for various purposes.  For example, creating new lockfiles from scratch for related packages.  So, the request is for something like:

`conan lock extract-profiles <host_profile_name> <build_profile_name>`

Since this feature seem relatively trivial and low risk to implement, it seems reasonable to add support for it now.  It does seem like the kind of feature which provides a good building block for adding flexibility to the existing lockfile featureset. 

Alternative to: https://github.com/conan-io/conan/issues/8580

uilianries:
Hello!

Add the new subcommand for `lock`: `extract-profile`

    conan lock extract-profiles <host_profile> <build_profile> --lockfile=<conan.lock>

This new command read a lockfile, by default it reads `./conan.lock`, but also accepts any file path by `--lockfile` argument;

The profiles extracted are written to both file paths passed as argument. When build profile is not listed in conan.lock, the build profile file is not generated.

Changelog: Feature: Extract host and build profiles from conan.lock
Docs: TODO

closes #8684

### DIFF
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -1495,6 +1495,22 @@ class ConanAPIV1(object):
         graph_lock_file.save(lockfile_out)
         self.app.out.info("Generated lockfile: %s" % lockfile_out)
 
+    @api_method
+    def lock_extract_profiles(self, host_profile_name, build_profile_name, lockfile=None, cwd=None):
+        cwd = cwd or os.getcwd()
+        lockfile = lockfile or LOCKFILE
+        lockfile_path = _make_abs_path(lockfile, cwd)
+        host_profile_path = _make_abs_path(host_profile_name, cwd)
+        build_profile_path = _make_abs_path(build_profile_name, cwd)
+        revisions_enabled = self.app.config.revisions_enabled
+        lockfile_handle = GraphLockFile.load(lockfile_path, revisions_enabled)
+        if lockfile_handle.profile_host:
+            save(host_profile_path, lockfile_handle.profile_host.dumps())
+            self.app.out.info("Generated host profile: {}".format(host_profile_path))
+        if lockfile_handle.profile_build:
+            save(build_profile_path, lockfile_handle.profile_build.dumps())
+            self.app.out.info("Generated build profile: {}".format(build_profile_path))
+
 
 Conan = ConanAPIV1
 
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -1497,6 +1497,13 @@ class ConanAPIV1(object):
 
     @api_method
     def lock_extract_profiles(self, host_profile_name, build_profile_name, lockfile=None, cwd=None):
+        """ Extract host and build profile from a conan lockfile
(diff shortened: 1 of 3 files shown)


conans/client/conan_api.py
```diff
@@ -1495,6 +1495,22 @@ def lock_create(self, path, lockfile_out,
...
+        lockfile = lockfile or LOCKFILE
+        lockfile_path = _make_abs_path(lockfile, cwd)
+        host_profile_path = _make_abs_path(host_profile_name, cwd)
+        build_profile_path = _make_abs_path(build_profile_name, cwd)
+        revisions_enabled = self.app.config.revisions_enabled
+        lockfile_handle = GraphLockFile.load(lockfile_path, revisions_enabled)
+        if lockfile_handle.profile_host:
+            save(host_profile_path, lockfile_handle.profile_host.dumps())
+            self.app.out.info("Generated host profile: {}".format(host_profile_path))
+        if lockfile_handle.profile_build:
```
uilianries, line 1517: We could add a warning message log if build profile is not listed, but by default it's not listed, so I preferred don't alarming.

memsharded:
Conan 2.0 lockfiles no longer contain profiles, so this probably don't make sense anymore, so closing it. Thanks!

mkmkme:
@memsharded 
> Conan 2.0 lockfiles no longer contain profiles, so this probably don't make sense anymore, so closing it. Thanks!

Can I ask why and what's the alternative?

Also, in that case #8580 should be closed

memsharded:
The idea is that configuration and lockfiles are orthogonal. When you install things you keep doing:
```
conan install . -pr=mywindowsprofile
# or
conan install . -pr=mylinuxprofile
```
And if you want to have the locked set of versions and revisions you provide it:
```
conan install -pr=mywindowsprofile --lockfile=mylockfile.lock
# or
conan install . -pr=mylinuxprofile --lockfile=mylockfile.lock
```
Because one of the major changes in 2.0 is that lockfiles are multi-configuration, the same lockfile file can be used to lock all possible configurations, and that simplifies things.
So all Conan commands keep the exact same syntax with and without lockfiles, and just add the ``--lockfile`` arguments when desired. This makes the CLI behavior more consistent.

mkmkme:
Does that mean that if one wants to have, say, a release branch with lock files, they will now have to store the profiles next to the lock files for reproducibility?
Because in continuous development, the profile with the same name (say, `acme-corp-default`) might evolve and change the compiler version.

memsharded:
Yes, the configuration (as everything that is source and input files) is something to be versioned.

The recommendation is (and it already was in 1.X), to put the configuration, profiles, etc in a git repo, and install it with ``conan config install``. So all modifications to configuration can be perfectly tracked, versioned, recovered in time, etc. ``conan config install`` allows to fetch github releases tarballs for example, or it also has extra args that could be used to checkout a given tag.

PR #9237, opened 2021-07-07
removing CONAN_USERNAME, CONAN_CHANNEL env-vars

memsharded:
Changelog: Remove: Env-vars CONAN_USERNAME, CONAN_CHANNEL are no longer used, they are defined in recipes or in command line.
Docs: TODO


lasote:
For me this is ok but I'm a bit worried about the "Docs: TODO" because I don't know how and when we are going to document the 2.0 changes. We might end with terrible docs with outdated stuff.

czoido:
> For me this is ok but I'm a bit worried about the "Docs: TODO" because I don't know how and when we are going to document the 2.0 changes. We might end with terrible docs with outdated stuff.

I agree with that, we should start thinking about the documentation for 2.0, maybe create a develop2 branch in the docs too or rethinking the whole thing?

czoido (approved):

memsharded:
Conan 2.0 docs will need to be re-thought completely.

PR #9240, opened 2021-07-08
New "list recipes" command

### LINKED ISSUE #9189: [feature] Conan 2.0 CLI: List command
Develop the new `list` command for the Conan 2.0 CLI

tapia:
Fixes #9189


conans/test/integration/command_v2/list_recipes_test.py
```diff
@@ -0,0 +1,267 @@
...
+            r"    test_recipe/2.1.0@user/channel#.*\n"
+            r"remote1:\n"
+            r"  test_recipe\n"
+            r"    test_recipe/1.0.0@user/channel\n"
+            r"    test_recipe/1.1.0@user/channel\n"
+            r"remote2:\n"
+            r"  test_recipe\n"
+            r"    test_recipe/2.0.0@user/channel\n"
+            r"    test_recipe/2.1.0@user/channel\n"
+        )
```
franramirez688, line 130: Shouldn't it be cleaner to use `textwrap` for all the `expected_output`s?
For instance:

```python3
expected_output = textwrap.dedent("""
    Local Cache:
      test_recipe
        test_recipe/1.0.0@user/channel#.*
        test_recipe/1.1.0@user/channel#.*
        test_recipe/2.0.0@user/channel#.*
        test_recipe/2.1.0@user/channel#.*
    remote1:
      test_recipe
        test_recipe/1.0.0@user/channel
        test_recipe/1.1.0@user/channel
    remote2:
      test_recipe
        test_recipe/2.0.0@user/channel
        test_recipe/2.1.0@user/channel
""")
```

tapia: I tried, but the string is a regular expression (note the `.*` in the Local Cache recipes), and it doesn't work with textwrap (or at least I didn't manage to make it work)

memsharded: `` textwrap.dedent(r"""...)`` didn't work?

conans/cli/commands/help.py
```diff
@@ -3,21 +3,21 @@
 from conans.cli.output import cli_out_write
 from conans.client.output import Color
 from conans.errors import ConanException
-from conans.cli.command import conan_command
+from conans.cli.command import conan_command, COMMAND_GROUPS
```
memsharded:
```suggestion
from conans.cli.command import conan_command
```

Unused import?

conans/cli/commands/list.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+@conan_subcommand(formatters=search_formatters)
+def list_recipes(conan_api, parser, subparser, *args):
+    """
+    Search available recipes in the local cache or in the remotes
+    """
+    subparser.add_argument(
+        "query",
+        nargs='?',
```
memsharded: Under the new CLI design, we wanted to be always explicit, it seems this will allow an empty argument. 

Maybe force always one argument, and it should be at least ``conan list recipes *`` to list everything? cc/ @czoido

conans/cli/commands/remote.py
```diff
@@ -1,7 +1,7 @@
 import json
 
 from conans.cli.output import cli_out_write
-from conans.cli.command import conan_command, conan_subcommand, OnceArgument
+from conans.cli.command import conan_command, conan_subcommand, OnceArgument, COMMAND_GROUPS
```
memsharded:
```suggestion
from conans.cli.command import conan_command, conan_subcommand, OnceArgument
```

conans/cli/commands/search.py
```diff
@@ -1,42 +1,27 @@
...
     args = parser.parse_args(*args)
 
-    remotes = args.remote or []
-    info = conan_api.search_recipes(args.query, remote_patterns=remotes, local_cache=args.cache)
-    return info
+    remotes = conan_api.get_active_remotes(args.remote)
+
+    results = []
+    for remote in remotes:
+        results.append(conan_api.search_remote_recipes(remote, args.query))
```
memsharded, line 26: How would incremental output (print in CLI for every remote searched, not need to wait until the end, this can be slow) look like? No need for this PR, but maybe something to consider in the design of the formatters

conans/client/api/conan_api.py
```diff
@@ -90,19 +166,81 @@ def user_update(self, user_name, user_pasword):
...
+            references = search.search_local_recipes(query)
+        except ConanException as exc:
+            data["error"] = True
+            raise
+
+        results = []
+        for reference in references:
+            reference = repr(reference)
+            result = {
+                "name": get_recipe_name(reference),
```
memsharded: Maybe this is not necessary, if you can do ``name = reference.name`` before converting to string. ``get_recipe_name()`` should be unnecessary function.

conans/client/conan_api.py
```diff
@@ -871,49 +870,31 @@ def users_list(self, remote_name=None):
...
-            if remote_ref.ordered_packages:
-                for package_id, properties in remote_ref.ordered_packages.items():
-                    # Artifactory uses field 'requires', conan_center 'full_requires'
-                    requires = properties.get("requires", []) or properties.get("full_requires", [])
-                    search_recorder.add_package(remote_name, ref,
-                                                package_id, properties.get("options", []),
-                                                properties.get("settings", []),
-                                                requires)
-        return search_recorder.get_info()
+    # @api_method
```
memsharded, line 873: Why not removing it?

conans/test/integration/command_v2/list_recipes_test.py
```diff
@@ -0,0 +1,267 @@
...
+        self.client.run("user username -p passwd -r {}".format(remote_name))
+
+    def _create_recipe(self, reference):
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile
+            class MyLib(ConanFile):
+                pass
+            """)
+
+        self.client.save({'conanfile.py': conanfile})
```
memsharded:
```suggestion
        self.client.save({'conanfile.py': GenConanfile()})
```

conans/test/integration/command_v2/list_recipes_test.py
```diff
@@ -0,0 +1,267 @@
...
+        self.client.run("create . {}".format(reference))
+
+    def _upload_recipe(self, remote, reference):
+        conanfile = textwrap.dedent("""
+            from conans import ConanFile
+            class MyLib(ConanFile):
+                pass
+            """)
+
+        self.client.save({'conanfile.py': conanfile})
```
memsharded:
```suggestion
        self.client.save({'conanfile.py': GenConanfile()})
```

conans/test/integration/command_v2/search_test.py
```diff
@@ -0,0 +1,221 @@
...
+from collections import OrderedDict
+
+import pytest
+
+from conans.test.utils.tools import TestClient, TestServer
+
+
+class TestSearch:
+
+    @pytest.fixture(autouse=True)
```
memsharded: If the fixture is not used in 50% of the cases, better inject it explicitly

memsharded (commented):
Looking good, clean, some improvements besides the feature.

conans/cli/commands/list.py
```diff
@@ -0,0 +1,90 @@
...
+        args.cache = True
+        remotes = conan_api.get_active_remotes(args.remote)
+    elif args.cache and not args.remote:
+        # If cache is defined but not remote, show results only from the cache
+        remotes = None
+    else:
+        remotes = conan_api.get_active_remotes(args.remote)
+
+    if remotes and not args.query:
+        parser.error("If searching in a remote, a query string must be provided.")
```
lasote: For me, this is a bit confusing. This is my experience as a user:

1) I type `conan list recipes`, and I receive the error `conan list: error: If searching in a **remote**, a query string must be provided.`... then I think.... uh? I use `conan search` for searching in the remote.
2) I type a pattern: `conan list recipes json*` and the first thing I see is:

```
Local Cache:
  There are no matching recipes
conan-center:
  json-c
  ...
```

So it is indeed searching in the local cache as I wanted...but the pattern is mandatory because it also searches on the remote, so basically, I cannot search the cache without a pattern? 

3) Then I see that the `-c` is for searching locally, oh, finally it made sense.

So, I know we decided that `conan list` will search both locally and in a remote but for me something is not comfortable enough.

lasote: Note: I don't know how to improve it except doing it by default (without args) only local.

conans/test/integration/command_v2/list_recipes_test.py
tapia: Nope. My guess is that `dedent()` always returns a plain string, not a raw string.

conans/cli/commands/search.py
tapia: The current design of the command output wouldn't allow it. It needs the command function to return, and then the the full return value is passed to the formatter function, via the `conan_command` decorator.

Maybe we can use `yield` instead of `return` and refactor the `conan_command` decorator. I agree this is something worth looking, as it would allow us to eventually start using asynchronous calls.

conans/client/api/conan_api.py
tapia: You're absolutely right. Fixed.

conans/client/conan_api.py
tapia: This function must be rewritten and moved to `client/api/conan_api.py`, and I want to temporarely keep it to use it as reference. I had to comment it, though, because it uses `SearchRecorder`, that doesn't exist anymore.

When the new function is done, I'll delete this.

conans/cli/commands/help.py
tapia: Ouch, I missed this from a merge conflict 😬  Fixed.

conans/test/integration/command_v2/list_recipes_test.py
tapia: Cool, I didn't know this 🙂

conans/test/integration/command_v2/search_test.py
tapia: Fixed

conans/cli/commands/list.py
tapia: As we agreed, now it works like this:

* The query string is now __always__ required. Explicit is better than implicit.
* `conan list recipes whatever` will only search in the cache.
* `conan list recipes -r remote1 whatever` will only search in the `remote1` remote
* `conan list recipes -a whatever` will search in all the remotes (but not the cache)
* `conan list recipes -a -c whatever` will search in all the remotes and in the cache.

conans/client/api/conan_api.py
```diff
@@ -23,10 +96,8 @@ def wrapper(api, *args, **kwargs):
         except EnvironmentError:
             old_curdir = None
         try:
-            # TODO: Removing ConanApp creation, should we make it different for 2.0?
-            # Also removing the logger call, maybe conan_command can handle it
-            cache = ClientCache(api.cache_folder, api.out)
-            with environment_append(cache.config.env_vars):
+            api.create_app()
```
memsharded, line 99: This was using a shortcut on purpose to avoid the creation of tons of things that won't be used.

Anyway, this is ok, I think we will eventually remove this.

conans/client/api/conan_api.py
```diff
@@ -90,19 +166,79 @@ def user_update(self, user_name, user_pasword):
...
+            results.append(result)
+
+        data["results"] = results
+        return data
+
+    @api_method
+    def search_remote_recipes(self, remote, query):
+        data = {
+            "remote": remote.name,
+            "error": False,
```
memsharded, line 200: This ``error`` is never being set to True

conans/client/api/conan_api.py
```diff
@@ -90,19 +166,79 @@ def user_update(self, user_name, user_pasword):
...
+                result = {
+                    "name": reference.name,
+                    "id": repr(reference)
+                }
+                results.append(result)
+
+        data["results"] = results
+        return data
+
+    @api_method
```
memsharded, line 220: Probably this is not an ``api_method``, but an auxiliary one. ``api_methods`` are those that match the Conan command, and need to have a "fresh state". Something to be discussed for the api, not this PR.

conans/client/api/helpers/search.py
```diff
@@ -0,0 +1,20 @@
+from collections import OrderedDict
+
+from conans.search.search import search_recipes
+
+
+class Search(object):
```
memsharded, line 6: This class is maybe a bit boilerplate at this stage, 2 bare functions would make it better.

memsharded (approved):
I think this is good, and it can be merged soon, then keep iterating some of the minor details, some of my comments are indirect observations of things that need to be revisited, like the api.

conans/client/api/conan_api.py
```diff
@@ -90,19 +166,79 @@ def user_update(self, user_name, user_pasword):
...
+        }
+
+        remotes = self.app.cache.registry.load_remotes()
+        search = Search(self.app.cache, self.app.remote_manager, remotes)
+
+        try:
+            references = search.search_local_recipes(query)
+        except ConanException as exc:
+            data["error"] = True
+            raise
```
franramirez688, line 183: I guess it does not make sense to have the `data['error']` field because we're raising the exception anyway so we'll always have the `'error' = False` at the end of the process, won't it?

memsharded: Yes, that is true, that is probably legacy for when using multiple remotes and 1 could fail, but not the others.

lasote (approved):

lasote:
Let's merge this so we can evolve it later.

PR #9267, opened 2021-07-13
Allow defining settings for a downstream consumer

### LINKED ISSUE #9143: [feature] Allow defining settings for a downstream consumer (different than upstream dependencies)
Continuation from https://github.com/conan-io/conan/pull/9129

The command ``conan install . -s pkg:build_type=Debug -s build_type=Release``, to have all dependencies in Release, but build your current project as Debug has some limitations:

- It requires a conanfile.py, that defines the ``name = "pkg"``
- conanfile.txt will not allow this
- conanfile.py without a name will not allow this

The solution could be some kind of placeholder that means "current consumer", something like ``conan install . -s $:build_type=Debug -s build_type=Release``

lasote:
Changelog: Feature: Now it is possible to define settings for a downstream consumer using `-s &:setting=value` or the same syntax in the profile even if the consumer is a `conanfile.txt` or a `conanfile.py` not declaring a name. e.g: Building a Debug application with Release dependencies: `-s build_type=Release -s &:build_type=Debug`
Docs: https://github.com/conan-io/docs/pull/2160

Close #9143

### DIFF
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -368,7 +368,6 @@ class DepsGraphBuilder(object):
 
         locked_id = requirement.locked_id
         lock_py_requires = graph_lock.python_requires(locked_id) if locked_id is not None else None
-
         dep_conanfile = self._loader.load_conanfile(conanfile_path, profile, ref=requirement.ref,
                                                     lock_python_requires=lock_py_requires)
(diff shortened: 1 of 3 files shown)


memsharded (approved):
It seems this is very low risk, and the feature is quite useful, solves a problem from long time.

Lets merge it, please document it as experimental, just in case.

PR #8976, opened 2021-05-18
Feature/test build require invalid sol2

memsharded:
Changelog: (Feature | Fix | Bugfix): Describe here your pull request
Docs: https://github.com/conan-io/docs/pull/XXXX

Alternate approach to https://github.com/conan-io/conan/pull/8972.

It might not viable, follow https://github.com/conan-io/conan/pull/8972 for discussion.

PR #9268, opened 2021-07-13
Introducing ConanErrorConfiguration exception

memsharded:
Changelog: Feature: New ``ConanErrorConfiguration``, to define a hard invalid configuration error, as opposed to the current ``ConanInvalidConfiguration`` exception that can still succeed using a compatible binary defined with  ``package_id()`` or ``compatible_packages``, even if the current configuration cannot be used to build a binary from source.

Close https://github.com/conan-io/conan/pull/8976
Close https://github.com/conan-io/conan/pull/8972
Close https://github.com/conan-io/conan/pull/8951


SSE4:
@memsharded there is no issue linked, can you explain more the background of the PR?
e.g. is there a user story, scenario? do you find `ConanInvalidBuildConfiguration` to be useful? why do you think `ConanInvalidConfiguration` is not enough?

conans/client/conanfile/configure.py
```diff
@@ -26,8 +26,6 @@ def run_configure_method(conanfile, down_options, down_ref, ref):
         with conanfile_exception_formatter(str(conanfile), "configure"):
             conanfile.configure()
 
-        conanfile.settings.validate()  # All has to be ok!
```
memsharded, line 29: This is an important change!

It is delaying the failure, or avoiding it, when some setting is missing. If someone does not provide a compiler for a package ``cmake/3.14``, but the package does ``del self.info.settings.compiler`` in its package_id, this will avoid failure, and the package will be able to resolve, even if compiler no defined. It will only fail if trying to build from source and setting.compiler not specified.

conans/client/graph/graph_binaries.py
```diff
@@ -209,11 +206,20 @@ def _evaluate_node(self, node, build_mode, update, remotes):
                 # package_id was not locked, this means a base lockfile that is being completed
                 locked.complete_base_node(node.package_id, node.prev)
 
+        if (node.binary in (BINARY_BUILD, BINARY_MISSING) and node.conanfile.info.invalid and
```
memsharded, line 209: This is the other important change: The first PACKAGE_ID_INVALID is not set directly, only after the package has been processed and it was detected that it is necessary to build it or that it is missing (no fallback to compatible packages, no removal of info in package_id()). As this is delayed, it allows to work without all settings defined if they are not part of the package_id, and the binary already exists.

conans/test/integration/graph/core/test_build_require_invalid.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def test_valid_build_require_two_profiles(self, client):
+        conanfile_consumer = GenConanfile().with_build_requires("pkg/0.1").with_settings("os")
+        client.save({"consumer/conanfile.py": conanfile_consumer})
+        client.run("install consumer -s:b os=Linux -s:h os=Windows")
+        assert "pkg/0.1:{} - Cache".format(self.linux_package_id) in client.out
+        assert "pkg/0.1: Already installed!" in client.out
+
+
+class TestErrorConfiguration(TestInvalidConfiguration):
```
lasote, line 70: Unused?

conans/test/integration/graph/core/test_build_require_invalid.py
```diff
@@ -0,0 +1,207 @@
...
+                    raise ConanErrorConfiguration("Package does not work in Windows!")
+
+            def package_id(self):
+                del self.info.settings.os
+        """)
+    linux_package_id = NO_SETTINGS_PACKAGE_ID
+    invalid = "ConfigurationError"
+
+
+class TestErrorConfigurationCompatible(TestInvalidConfiguration):
```
lasote, line 92: Unused?

conans/test/integration/graph/core/test_build_require_invalid.py
```diff
@@ -0,0 +1,207 @@
...
+        client.run("install consumer -s:b os=Linux -s:h os=Windows")
+        assert "pkg/0.1:{} - Cache".format(self.linux_package_id) in client.out
+        assert "pkg/0.1: Already installed!" in client.out
+
+        client.run("install consumer -s:b os=Windows -s:h os=Windows")
+        assert "pkg/0.1:{} - Cache".format(self.linux_package_id) in client.out
+        assert "pkg/0.1: Already installed!" in client.out
+
+
+class TestInvalidBuildCompatible(TestInvalidBuildPackageID):
```
lasote, line 186: Unused?

conans/test/integration/package_id/test_validate.py
```diff
@@ -64,17 +64,43 @@ def package_id(self):
         self.assertIn("pkg/0.1: Package 'cb054d0b3e1ca595dc66bc2339d40f1f8f04ab31' created",
                       client.out)
 
-        client.run("create . pkg/0.1@ -s os=Windows")
-        self.assertIn("pkg/0.1: Main binary package 'INVALID' missing. "
-                      "Using compatible package 'cb054d0b3e1ca595dc66bc2339d40f1f8f04ab31'",
-                      client.out)
-        self.assertIn("pkg/0.1:cb054d0b3e1ca595dc66bc2339d40f1f8f04ab31 - Cache", client.out)
+        client.run("create . pkg/0.1@ -s os=Windows", assert_error=True)
```
lasote, line 67: So now `ConanInvalidConfiguration` doesn't fallback to compatible packages?

conans/test/integration/package_id/test_validate.py
```diff
@@ -64,17 +64,43 @@ def package_id(self):
...
+        self.assertIn("pkg/0.1: Invalid: Windows not supported", client.out)
+        self.assertIn("pkg/0.1:INVALID - Invalid", client.out)
         client.run("info pkg/0.1@ -s os=Windows")
-        self.assertIn("pkg/0.1: Main binary package 'INVALID' missing. "
-                      "Using compatible package 'cb054d0b3e1ca595dc66bc2339d40f1f8f04ab31'",
+        self.assertIn("pkg/0.1: Main binary package '3475bd55b91ae904ac96fde0f106a136ab951a5e' "
+                      "missing. Using compatible package 'cb054d0b3e1ca595dc66bc2339d40f1f8f04ab31'",
```
lasote, line 72: Why the info says it use the compatible but the create failed?

lasote (commented):
Sorry but I don't get it, lets jump in a call

conans/client/graph/graph_binaries.py
```diff
@@ -343,7 +349,15 @@ def _compute_package_id(self, node, default_package_id_mode, default_python_requ
                 try:
                     conanfile.validate()
                 except ConanInvalidConfiguration as e:
-                    conanfile.info.invalid = str(e)
+                    conanfile.info.invalid = BINARY_INVALID, str(e)
```
lasote, line 352: we should improve this before merge if possible, the tuple is very tricky

lasote (approved):

memsharded:
@SSE4 @uilianries 

The ``ConanInvalidConfiguration`` fails today when settings are removed in ``package_id`` like ``del self.info.settings.compiler``, while it happens that there could exist a binary that would work. It works today when ``compatible_packages`` are explicitly defined, which is clearly an inconsistent behavior.

On the other hand, there is no mechanism to say: "this configuration is invalid", even if the binary is valid, when some options or target settings are wrong: example an ``android_ndk`` package that you have a binary the binary is valid, but the provided host settings is "iOS". It is not that you want to fallback and look for a binary compatible package for ``android_ndk``, the configuration is directly wrong, not the binary. So:

- ConanInvalidConfiguration: This binary cannot be built with these settings, but a compatible binary might exist (via either self.info.settings removal or compatible_packages). This scenario might not fail in many situations where the binary is not necessary, or it is necessary but a compatible one can be obtained.
- ConanErrorConfiguration: The binary exists and it is ok, but the requested configuration is wrong, and it will always directly fail.

uilianries:
> It works today when compatible_packages are explicitly defined, which is clearly an inconsistent behavior.

That's the point which makes the difference. On CCI we don't see it because we don't use `compatible_packages` there, but it makes sense for enterprises and others. Thanks for explaining bettter.

czoido (approved):

conans/test/integration/package_id/test_validate.py
danimtb: I need some clarification about this one as well

conans/test/integration/package_id/test_validate.py
danimtb: I need some clarification about this one as well

danimtb (commented):
✅

conans/test/integration/package_id/test_validate.py
memsharded: Yes, it is not possible to build a compatible configuration. Compatible configurations, both with ``compatible_packages`` and with deletion of ``self.info.settings`` in ``package_id()`` should exist in order to use them, but they are not built as a result of a fallback. It would be usually challenging, or directly impossible to define the full configuration for that binary to be built, as the fallback defines only minimum information, but not the whole profile that might be necessary to built that configuration from source.

Note that the previous behavior was really confusing, we did a ``conan create`` for one configuration and the package was not built at all. Note that if you did this command as a result of changing some source code, it could be ignored. A ``conan create`` without changing the build policy should always do a package build.

conans/test/integration/package_id/test_validate.py
memsharded: The ``conan info`` command doesn't build, it is a proxy for ``install`` not for create. If we did ``info --dry-build=pkg``, it would not fall back and report ``pkg/0.1 ID: INVALID``

PR #9272, opened 2021-07-15
New conan2 command: list recipe-revisions

tapia:
Changelog: Feature: New conan2 command: list recipe-revisions
Docs: omit, TODO (2.0 docs)


conans/cli/commands/list.py
```diff
@@ -0,0 +1,165 @@
...
+from conans.client.output import Color
+from conans.util.dates import iso8601_to_str
+
+indentation = 2
+remote_color = Color.BRIGHT_BLUE
+recipe_color = Color.BRIGHT_WHITE
+reference_color = Color.WHITE
+
+
+def list_recipes_cli_formatter(results):
```
franramirez688, line 14: I start thinking of we should keep all the `xxxx_formatter` functions in another specific module, e.g., `cli/formatters.py`

tapia: Let's wait and see if we need it. Except the `json_formatter` function, the command and formatter functions are completely coupled, so it makes sense to me to keep them in the same module.

memsharded: I prefer the formatters to be close to their specific commands, there is little value in putting them together, as they can be very different, decoupled outputs, while for every command, the data structure is 90% of the implementation, so IMHO it makes sense to have formatters grouped by command.

They could be in commands/list_formatters.py, if too much code for the ``list.py`` file.

conans/client/api/conan_api.py
```diff
@@ -165,6 +166,28 @@ def user_remove(self, remote_name):
...
+
+        if not all_remotes:
+            raise ConanException("The remotes registry is empty. "
+                                 "Please add at least one valid remote.")
+
+        # If no remote is specified, search in all of them
+        if not remote_names:
+            # Exclude disabled remotes
+            all_remotes = [remote for remote in all_remotes if not remote.disabled]
+            return all_remotes
```
memsharded, line 182: This seems to be leaking undocumented classes via the API. So far, the goal was that the API both received and returned plain/primitive python objects. Here it is returning a "Remote" object, that we should make part of the public API too, documented, and commit to stability.

Again, API design, to be discussed separately, not this PR.

conans/test/integration/command_v2/list_recipe_revisions_test.py
```diff
@@ -0,0 +1,181 @@
...
+import pytest
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient, TestServer
+
+
+class TestListRecipeRevisionsBase:
+    @pytest.fixture(autouse=True)
+    def _setup(self):
+        self.users = {}
```
memsharded: Unused self.users member?

memsharded (approved):
Looks good, I think this can be merged.

conans/test/integration/command_v2/list_recipe_revisions_test.py
tapia: Yes, it's a legacy variable. Good catch!

conans/client/api/conan_api.py
tapia: Yes, this Remote class should be documented, because it's extensively used.

franramirez688 (approved):

conans/client/api/conan_api.py
memsharded: But one thing is the programming interface and a much different one is the API one.
If we propagate this object through the API, then we will be exposing all its public interface, and even if we restrict it in documentation, we could have a stability problem. It could make sense to return through the api simplified objects, proxy wrappers, or something that could protect the internal programming interface.

PR #9273, opened 2021-07-15
POC of an "install" in a local folder +  "post_install" method

lasote:

To discuss what we want

### DIFF
--- a/conans/client/manager.py
+++ b/conans/client/manager.py
@@ -106,8 +106,8 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
             _f = os.path.join(conanfile.generators_folder, _name)
             shutil.copytree(cf._conanfile.package_folder, _f, symlinks=True)
             cf._conanfile.folders.set_base_package(_f)
-            if hasattr(cf._conanfile, "post_install"):  # TODO we need a name
-                cf._conanfile.post_install()
+            if hasattr(cf._conanfile, "install"):  # TODO we need a name
+                cf._conanfile.install()
 
     output = conanfile.output if root_node.recipe != RECIPE_VIRTUAL else out
 
--- a/conans/client/manager.py
+++ b/conans/client/manager.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 
 from conans.client.graph.build_mode import BuildMode
 from conans.client.graph.graph import RECIPE_CONSUMER, RECIPE_VIRTUAL
@@ -22,7 +23,7 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
                  build_modes=None, update=False, manifest_folder=None, manifest_verify=False,
                  manifest_interactive=False, generators=None, no_imports=False,
                  create_reference=None, keep_build=False, recorder=None, lockfile_node_id=None,
-                 is_build_require=False, add_txt_generator=True):
+                 is_build_require=False, add_txt_generator=True, local_folder=False):
 
     """ Fetch and build all dependencies for the given reference
     @param app: The ConanApp instance with all collaborators
@@ -96,6 +97,17 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
     conanfile.folders.set_base_install(install_folder)
     conanfile.folders.set_base_imports(install_folder)
     conanfile.folders.set_base_generators(base_folder)
+
+    if local_folder:
+        for cf in conanfile.dependencies.host.values():  # TODO: build requires? collisions suffix?
+            _name = cf.ref.name
+            if conanfile.settings.get_safe("build_type"):
+                _name += "-{}".format(str(conanfile.settings.build_type))
+            _f = os.path.join(conanfile.generators_folder, _name)
+            shutil.copytree(cf._conanfile.package_folder, _f, symlinks=True)
+            cf._conanfile.folders.set_base_package(_f)
+            if hasattr(cf._conanfile, "post_install"):  # TODO we need a name
+                cf._conanfile.post_install()
 
     output = conanfile.output if root_node.recipe != RECIPE_VIRTUAL else out
 
--- a/conans/client/manager.py
+++ b/conans/client/manager.py
@@ -99,14 +99,16 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
     conanfile.folders.set_base_generators(base_folder)
 
     if local_folder:
-        for cf in conanfile.dependencies.host.values():  # TODO: build requires? collisions suffix?
+        for cf in conanfile.dependencies.values():
             _name = cf.ref.name
             if conanfile.settings.get_safe("build_type"):
                 _name += "-{}".format(str(conanfile.settings.build_type))
-            _f = os.path.join(conanfile.generators_folder, _name)
+            _pre_folder = "build" if cf.is_build_context else "host"
+            _f = os.path.join(conanfile.generators_folder, _pre_folder, _name)
+            # The folder _f is .../host/hello-Release/ inside the generators folder
             shutil.copytree(cf._conanfile.package_folder, _f, symlinks=True)
             cf._conanfile.folders.set_base_package(_f)
-            if hasattr(cf._conanfile, "install"):  # TODO we need a name
+            if hasattr(cf._conanfile, "install"):
                 cf._conanfile.install()
 
     output = conanfile.output if root_node.recipe != RECIPE_VIRTUAL else out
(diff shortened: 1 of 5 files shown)


conans/test/integration/command/install/install_local_package_test.py
```diff
@@ -0,0 +1,65 @@
...
+               .with_import("from conan.tools.files import save").with_import("import os")
+               .with_requires("hello/1.0"))
+    chat += """
+    def package(self):
+        save(self, os.path.join(self.package_folder, "my_chat.lib"), "contents")
+    """
+    client.save({"conanfile.py": chat})
+    client.run("create .")
+
+    consumer = str(GenConanfile()
```
memsharded: Maybe for this just write the conanfile.py directly? :)

conans/test/integration/command/install/install_local_package_test.py
```diff
@@ -0,0 +1,65 @@
...
+                   .with_settings("os", "arch", "compiler", "build_type")
+                   .with_requirement("chat/1.0")
+                   .with_generator("CMakeDeps")
+               )
+    consumer += """
+    def layout(self):
+        cmake_layout(self)
+    """
+    client.save({"conanfile.py": consumer})
+    client.run("install . --local-folder -if=my_install")
```
memsharded: ``my_install`` folder actually used for something?

conans/client/manager.py
```diff
@@ -97,6 +98,17 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
     conanfile.folders.set_base_imports(install_folder)
     conanfile.folders.set_base_generators(base_folder)
 
+    if local_folder:
+        for cf in conanfile.dependencies.host.values():  # TODO: build requires? collisions suffix?
+            _name = cf.ref.name
+            if conanfile.settings.get_safe("build_type"):
```
memsharded, line 104: Probably the ``build_type`` not necessary here if we actually leverage the ``layout()``, as it can be seen in the test below with ``path = os.path.join(client.current_folder, "cmake-build-release", "conan")`` path.

 I didn't think of that, but maybe it fits nicely, just using the layout definition, taking advantage that is already paremeterized for different build_type (and possibly architectures, or even other custom configs)

On the other hand, having it separately from the build folders could be nice, easier to .gitignore the build, or clean the build more easily, without actually removing the deps.

memsharded (commented):
Probably the next thing to figure out is your ``# TODO: build requires? collisions suffix?``. What happens with different versions or configurations that would otherwise conflict? A first level folder "host" & "build" would highly reduce this I guess.

conans/client/manager.py
lasote: It is necessary for multi config, where the dependency binary packages must be different (that's why the suffix) but the generators folder according to the layout won't. (build/conan)

conans/test/integration/command/install/install_local_package_test.py
lasote: not really. Just hate to pollute the current dir even in tests (toc)

lasote: I will remove for clarity

lasote: Note: In case the layout doesn't declare generators folder it will use it.

conans/test/integration/command/install/install_local_package_test.py
lasote: Probably

lasote:
If this whole idea doesn't sound too crazy I can add functional test actually building something and multiconfig.

conans/test/integration/command/install/install_local_package_test.py
```diff
@@ -0,0 +1,70 @@
...
+        from conan.tools.layout import cmake_layout
+        class HelloConan(ConanFile):
+            name = 'consumer'
+            version = '1.0'
+            settings = "os", "arch", "compiler", "build_type"
+            generators = "CMakeDeps"
+            requires = ("chat/1.0", )
+
+            def requirements(self):
+                self.requires("chat/1.0")
```
memsharded:
```suggestion
            settings = "os", "arch", "compiler", "build_type"
            generators = "CMakeDeps"
            requires = ("chat/1.0", )
            
```

conans/test/integration/command/install/install_local_package_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+    client = TestClient()
+    hello = str(GenConanfile().with_name("hello").with_version("1.0")
+                .with_import("from conan.tools.files import save").with_import("import os"))
+    hello += """
+
+    def package(self):
+        save(self, os.path.join(self.package_folder, "my_hello.lib"), "contents")
+
+    def post_install(self):
```
memsharded: I'd probably call it ``install()``

memsharded (commented):
Yes, I would say the idea is worth pursuing a bit more.

conans/client/manager.py
memsharded: Yes, it would be something equivalent to ``libdirs`` that it will have either the base folder conditioned (``cmake-build-release``), or the lib folder conditioned ``build/conan/Release``. Sounds something that could be parameterized and modeled in ``layout``

memsharded (commented):
Looking very good, it is nice to see what can be achieved with so little lines of code.

blackliner:
I would like to play around with this branch, any advise how to get a working copy for local testing? Like 1) git checkout branch 2) `pip3 install` .... or so ?

From the current changes I can't really imagine how the folder structure would look like.

My personal solution at the moment:
- use `deploy` generator to get a flat hierarchy
- use `cmake_find_package` only to generate find modules, to avoid build type conflicts
- search&replace paths in generated find modules so they work from the deployed folder structure
- automatically delete build_requirements, and static libararies
- figure out that some of the static libraries headerfiles are required, so add them back

🙈

memsharded:
Hi @blackliner 

Yes, this is the same flow as described in https://github.com/conan-io/conan README file:

- git clone <url> conan_io && cd conan_io
- git checkout/remote add/depending on what have you cloned
- (recommended) create a python venv
- pip install -e .

The key is ``pip install -e .`` install the current project in "editable" mode. Every change you do to the source code is immediately available in the command line ``conan`` command.

blackliner:
I added `export CONAN_V2_MODE=1` to the env, but it seems that still the old conan is called, as it cannot find the `--local-folder` argument
```
(env) fberchtold@W10-RIG:~/workspace/conan/tmp$ conan install lum_common_log/1.0.0@ --local-folder
usage: conan install [-h] [-g GENERATOR] [-if INSTALL_FOLDER] [-m [MANIFESTS]]
                     [-mi [MANIFESTS_INTERACTIVE]] [-v [VERIFY]]
                     [--no-imports] [--build-require] [-j JSON] [-b [BUILD]]
                     [-r REMOTE] [-u] [-l LOCKFILE]
                     [--lockfile-out LOCKFILE_OUT] [-e ENV_HOST]
                     [-e:b ENV_BUILD] [-e:h ENV_HOST] [-o OPTIONS_HOST]
                     [-o:b OPTIONS_BUILD] [-o:h OPTIONS_HOST]
                     [-pr PROFILE_HOST] [-pr:b PROFILE_BUILD]
                     [-pr:h PROFILE_HOST] [-s SETTINGS_HOST]
                     [-s:b SETTINGS_BUILD] [-s:h SETTINGS_HOST] [-c CONF_HOST]
                     [-c:b CONF_BUILD] [-c:h CONF_HOST]
                     [--lockfile-node-id LOCKFILE_NODE_ID]
                     [--require-override REQUIRE_OVERRIDE]
                     path_or_reference [reference]
conan install: error: unrecognized arguments: --local-folder
ERROR: Exiting with code: 2
```

lasote:
You don't need to use the `CONAN_V2_MODE`. Have you checkout my fork and this branch (feature/new2_install_folder) before doing the `pip install -e .` ?

blackliner:
Alright, got another chance to look at it, this time it worked 🤷 

I see a deploy-like `host` folder that contains the packages content, and my assumption is that the `Find*.cmake` modules are not modified at this stage of the PR yet. I want to use the time though to bring in my point of view to the two statements from my initial email:

- The deploy initially looks like that could be a "flat" layout, maybe just 1 folder per package. But then, private requirements, build requirements and others that could have different versions, different user/channel, and even different package-id for the same package and version can co-exist in the graph. What to do then? If we store them in a folder with the necessary structure, that is mostly the Conan cache!

IMHO, `private`/`build_` requirements should not be brought into the filesystem anayways. They must not, by definition, show up in the final binary as public dependencies, being headerfiles or shared libraries. So that problem should be solved by definition.

- Is the deploy single-config or multi-config? It is likely that many users will want to have multi-config, and then the problem with the layout also becomes an issue.

I am actually interested in single-config for now (`cmake_find_package`), but from the direction conanV2 is going, I am aware that the future is the multi `CMakeDeps` generator. This does not look like a big issue though, other than figuring out a flat vs tree-like folder structure.

memsharded:
This use case has been implemented for 2.0 with the new "deployers" in https://github.com/conan-io/conan/pull/10507. Please track that PR and its docs, and the 2.0 next alpha.4 release to start trying it.

PR #9286, opened 2021-07-18
Feature/new cpp paths

memsharded:
Changelog: Omit
Docs: Omit

After some work, just one new test, a new ``pkg_cmake_test`` helper and some minor refactors.

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -44,10 +44,10 @@ def environment_wrap_command(conanfile, env_filenames, cmd, cwd=None):
 
     if bats:
         launchers = " && ".join('"{}"'.format(b) for b in bats)
-        return '{} && "{}"'.format(launchers, cmd)
+        return '{} && {}'.format(launchers, cmd)
     elif shs:
         launchers = " && ".join('. "{}"'.format(f) for f in shs)
-        return '{} && "{}"'.format(launchers, cmd)
+        return '{} && {}'.format(launchers, cmd)
     else:
         return cmd
 
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -44,10 +44,10 @@ def environment_wrap_command(conanfile, env_filenames, cmd, cwd=None):
 
     if bats:
         launchers = " && ".join('"{}"'.format(b) for b in bats)
-        return '{} && {}'.format(launchers, cmd)
+        return '{} && "{}"'.format(launchers, cmd)
     elif shs:
         launchers = " && ".join('. "{}"'.format(f) for f in shs)
-        return '{} && {}'.format(launchers, cmd)
+        return '{} && "{}"'.format(launchers, cmd)
     else:
         return cmd
 
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -26,17 +26,10 @@ class Patterns(object):
 
 class Infos(object):
 
-    def __init__(self, conanfile):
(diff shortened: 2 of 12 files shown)


conan/tools/env/environment.py
```diff
@@ -44,10 +44,10 @@ def environment_wrap_command(conanfile, env_filenames, cmd, cwd=None):
 
     if bats:
         launchers = " && ".join('"{}"'.format(b) for b in bats)
-        return '{} && {}'.format(launchers, cmd)
+        return '{} && "{}"'.format(launchers, cmd)
```
memsharded: This was failing for paths with spaces, new test failed because of this.

conans/model/new_build_info.py
```diff
@@ -42,33 +41,34 @@ def __init__(self):
     @property
     def required_component_names(self):
         """ Names of the required components of the same package (not scoped with ::)"""
-        if not self.requires:
+        if self.requires is None:
```
memsharded, line 44: Did some minor review of this file while understanding it:

- In general, it is better and faster to check for ``None`` explicitly.

conans/model/new_build_info.py
```diff
@@ -42,33 +41,34 @@ def __init__(self):
...
-        if not self._generator_properties:
+        if self._generator_properties is None:
             self._generator_properties = {}
         self._generator_properties.setdefault(generator, {})[property_name] = value
 
     def get_property(self, property_name, generator=None):
-        if not self._generator_properties:
+        if self._generator_properties is None:
             return None
-        if generator:
```
memsharded, line 57: This seemed redundant.

conans/model/layout.py
```diff
@@ -27,16 +26,22 @@ def __init__(self):
...
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
         self.source = NewCppInfo()
         self.build = NewCppInfo()
         self.package = NewCppInfo()
 
+    @property
+    def build_info(self):
+        build_cppinfo = self._conanfile.cpp.build.copy()
+        build_cppinfo.set_relative_base_folder(self._conanfile.folders.build_folder)
```
memsharded: I didn't reuse this in the ``installer.py``, because there it calls:

```python
build_cppinfo.set_relative_base_folder(self._conanfile.folders.build)
```
Instead of ``conanfile.folders.build_folder``, so there it is still a relative path, while here we need an absolute path.

lasote (commented):
I don't like it much. It is a magic object (with a very common name: build_info) that provides the same but absolute paths. I understand the pain but between a magic object and joining paths I prefer explicitly joining paths.

conans/test/assets/pkg_cmake.py
```diff
@@ -68,6 +68,40 @@ def package_info(self):
...
+                cmake_layout(self)
+
+            def build(self):
+                cmake = CMake(self)
+                cmake.configure()
+                cmake.build()
+
+            def test(self):
+                cmd = os.path.join(self.cpp.build_info.bindirs[0], "test")
+                cmd = '"{}"'.format(cmd) # to deal with spaces in path
```
memsharded: This is ugly, need to know in advance that the path might have spaces.

memsharded:
> I don't like it much. It is a magic object (with a very common name: build_info) that provides the same but absolute paths. I understand the pain but between a magic object and joining paths I prefer explicitly joining paths.

Agree, removed it, and simplified it down to just a test, no new features, and some minor refactors.

PR #9288, opened 2021-07-19
Feature/list package revisions

franramirez688:
Changelog: (Feature 2.0): Added new `list` subcommand:

```
conan list package-revisions -h
usage: conan list package-revisions [-h] [-f {cli,json}] [-r REMOTE | -a] [-c] package_reference

positional arguments:
  package_reference

optional arguments:
  -f {cli,json}, --format {cli,json}
                        Select the output format: cli, json. 'cli' is the default output.
  -r REMOTE, --remote REMOTE
                        Name of the remote to add
  -a, --all-remotes     Search in all the remotes
  -c, --cache           Search in the local cache
```

Depends on: https://github.com/conan-io/conan/pull/9272

### DIFF
--- a/conans/cli/commands/list.py
+++ b/conans/cli/commands/list.py
@@ -1,7 +1,7 @@
 import json
 
-from conans.cli.output import cli_out_write
 from conans.cli.command import conan_command, conan_subcommand, Extender, COMMAND_GROUPS
+from conans.cli.output import cli_out_write
 from conans.client.output import Color
 from conans.util.dates import iso8601_to_str
 
--- a/conans/cli/commands/list.py
+++ b/conans/cli/commands/list.py
@@ -31,7 +31,7 @@ def list_recipes_cli_formatter(results):
                 cli_out_write(current_recipe, fg=recipe_color, indentation=2)
 
             reference = recipe["id"]
-            cli_out_write(reference, fg=reference_color, indentation=2)
+            cli_out_write(reference, fg=reference_color, indentation=4)
 
 
 def _list_revisions_cli_formatter(results, is_package=False):
--- a/conans/cli/commands/list.py
+++ b/conans/cli/commands/list.py
@@ -5,7 +5,6 @@ from conans.cli.output import cli_out_write
 from conans.client.output import Color
 from conans.util.dates import iso8601_to_str
 
-indentation = 2
 remote_color = Color.BRIGHT_BLUE
 recipe_color = Color.BRIGHT_WHITE
 reference_color = Color.WHITE
@@ -20,19 +19,19 @@ def list_recipes_cli_formatter(results):
         if not remote_results.get("remote"):
             cli_out_write("Local Cache:", remote_color)
         else:
-            cli_out_write("{}:".format(remote_results["remote"]), remote_color)
+            cli_out_write(f"{remote_results['remote']}:", fg=remote_color)
 
         if not remote_results.get("results"):
-            cli_out_write("{}There are no matching recipes".format(" " * indentation))
+            cli_out_write("There are no matching recipes", indentation=2)
 
         current_recipe = None
         for recipe in remote_results["results"]:
             if recipe["name"] != current_recipe:
                 current_recipe = recipe["name"]
-                cli_out_write("{}{}".format(" " * indentation, current_recipe), recipe_color)
+                cli_out_write(current_recipe, fg=recipe_color, indentation=2)
 
             reference = recipe["id"]
-            cli_out_write("{}{}".format(" " * (indentation * 2), reference), reference_color)
+            cli_out_write(reference, fg=reference_color, indentation=2)
 
 
 def _list_revisions_cli_formatter(results, is_package=False):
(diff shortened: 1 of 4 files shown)


conans/client/api/conan_api.py
```diff
@@ -218,27 +241,41 @@ def search_remote_recipes(self, remote, query):
...
-        if not remote_names:
-            # Exclude disabled remotes
-            all_remotes = [remote for remote in all_remotes if not remote.disabled]
-            return all_remotes
-
-        active_remotes = []
-        for remote_name in remote_names:
-            remote = remotes[remote_name]
-            active_remotes.append(remote)
+    def get_revisions(self, reference, is_package=False, remote=None):
```
franramirez688: I've merged both logic for packages and recipes. I think it's not getting convoluted but let me know your opinion @tapia In my opinion, it would be a pain to have all those lines duplicated.

conans/cli/commands/list.py
```diff
@@ -34,14 +35,54 @@ def list_recipes_cli_formatter(info):
...
+
+        if not remote_results.get("remote"):
+            cli_out_write("Local Cache:", remote_color)
+        else:
+            cli_out_write("{}:".format(remote_results["remote"]), remote_color)
+
+        tab_space = " " * indentation
+        if not remote_results.get("results"):
+            ref_type = "packages" if is_package else "recipes"
+            cli_out_write(f"{tab_space}There are no matching {ref_type}")
```
tapia: I'd rather avoid mixing f-strings and `format()` in the same file

conans/client/api/conan_api.py
tapia: I don't like boolean flags (`is_package`). To me it's an antipattern, because it's not always obvious what they mean (you have to read the implementation to guess what is it if not a package).

I'd make two explicit API methods, calling an auxiliary function:

```python
def _get_revisions(self, reference, type_of_reference, remote):
    if type_of_reference == 'package':
        blahblah()
    elif type_of_reference == 'recipe':
        more_blah()

@api_method
def get_package_revisions(self, reference, remote):
    return self._get_revisions(reference, 'package', remote)
    
@api_method
def get_recipe_revisions(self, reference, remote):
    return self._get_revisions(reference, 'recipe', remote)
```

Much easier to read.

memsharded: Yes, agree, specially for API methods.
It could also implement the different type checking of ``reference`` outside of the ``_get_revisions()`` method.

conans/cli/commands/list.py
franramirez688: Completely right, I did not see the other `format`, my bad 😅

conans/client/api/conan_api.py
```diff
@@ -239,39 +239,47 @@ def search_remote_recipes(self, query, remote):
         data["results"] = results
         return data
 
-    @api_method
-    def get_remote_recipe_revisions(self, reference, remote):
-        ref = ConanFileReference.loads(reference)
+    def _get_revisions(self, reference, ref_type, remote=None):
+        if ref_type == "package":
```
memsharded: This type conversion is better done outside the ``_get_revisions``, then you avoid the ``ref_type`` argument, which would be redundant (you use the ``reference`` argument type, more idiomatic)

memsharded (approved):
Good enough to keep moving forward, lets merge it when possible.

conans/client/api/conan_api.py
```diff
@@ -239,39 +239,47 @@ def search_remote_recipes(self, query, remote):
...
+                results.append(result)
+            return results
 
     @api_method
-    def get_local_recipe_revisions(self, reference):
-        ref = ConanFileReference.loads(reference)
-        if ref.revision:
-            raise ConanException("Cannot list the revisions of a specific recipe revision")
+    def get_package_revisions(self, reference, remote=None):
+        return self._get_revisions(reference, 'package', remote=remote)
```
tapia: Following @memsharded [proposal](https://github.com/conan-io/conan/pull/9288/files#r675644552), this should become:

```python
    @api_method
    def get_package_revisions(self, reference, remote=None):
        reference = PackageReference.loads(reference)
        return self._get_revisions(reference, remote=remote)
```

conans/client/api/conan_api.py
tapia: This is correct, but you also need `ref_type` to know which remote method you are calling (line 249).

You could do that by checking the type of `reference`, though. The method becomes a little messier, but you're right that from a UX perspective, `ref_type` should be implicit.

```python
if isinstance(ref, PackageReference):
    method_name = 'get_package_revisions'
elif isinstance(ref, ConanFileReference):
    method_name = 'get_recipe_revisions'
```

memsharded: ``ref_type`` should be automatically deduced from the ``reference`` type, I agree is not perfect, but still better than having to pass ``"package"`` string explicitly as an argument

czoido (approved):

conans/client/api/conan_api.py
franramirez688: Much better 👍

PR #9310, opened 2021-07-22
Feature/New CLI command "list package-ids"

franramirez688:
Changelog: omit
Docs: omit 

```
$ conan list package-ids -h
usage: conan list package-ids [-h] [-f {cli,json}] [-r REMOTE | -a] [-c] reference

positional arguments:
  reference

optional arguments:
  -f {cli,json}, --format {cli,json}
                        Select the output format: cli, json. 'cli' is the default output.
  -r REMOTE, --remote REMOTE
                        Name of the remote to add
  -a, --all-remotes     Search in all the remotes
  -c, --cache           Search in the local cache
```

Depends on: https://github.com/conan-io/conan/pull/9288

### DIFF
--- a/conans/cli/commands/list.py
+++ b/conans/cli/commands/list.py
@@ -3,6 +3,7 @@ import json
 from conans.cli.command import conan_command, conan_subcommand, Extender, COMMAND_GROUPS
 from conans.cli.output import cli_out_write
 from conans.client.output import Color
+from conans.model.ref import PackageReference
 from conans.util.dates import iso8601_to_str
 
 remote_color = Color.BRIGHT_BLUE
@@ -40,7 +41,7 @@ def list_recipes_cli_formatter(results):
 def _list_revisions_cli_formatter(results, ref_type):
     for remote_results in results:
         _print_common_list_cli_output(remote_results, ref_type)
-        reference = remote_results["package_reference" if ref_type == "packages" else "reference"]
+        reference = remote_results["reference"]
         for revisions in remote_results["results"]:
             rev = revisions["revision"]
             date = iso8601_to_str(revisions["time"])
@@ -56,14 +57,24 @@ def list_package_revisions_cli_formatter(results):
 
 
 def list_package_ids_cli_formatter(results):
+    # Artifactory uses field 'requires', conan_center 'full_requires'
+    requires_fields = ("requires", "full_requires")
+    general_fields = ("options", "settings")
+
     for remote_results in results:
         _print_common_list_cli_output(remote_results, "references")
-
-        reference = remote_results["package_reference" if ref_type == "packages" else "reference"]
-        for revisions in remote_results["results"]:
-            rev = revisions["revision"]
-            date = iso8601_to_str(revisions["time"])
-            cli_out_write(f"{reference}#{rev} ({date})", fg=recipe_color, indentation=2)
+        reference = remote_results["reference"]
+        for pkg_id, props in remote_results["results"].items():
+            cli_out_write(repr(PackageReference(reference, pkg_id)), fg=reference_color, indentation=2)
+            for prop_name, values in props.items():
+                if prop_name in requires_fields:
+                    cli_out_write("requires:", fg=Color.BRIGHT_YELLOW, indentation=4)
+                    for req in values:
+                        cli_out_write(req, fg=Color.CYAN, indentation=6)
+                elif prop_name in general_fields:
+                    cli_out_write(f"{prop_name}:", fg=Color.BRIGHT_YELLOW, indentation=4)
+                    for name, val in values.items():
+                        cli_out_write(f"{name}={val}", fg=Color.CYAN, indentation=6)
 
 
 # FIXME: it's a general formatter, perhaps we should look for another module
@@ -149,33 +160,24 @@ def _get_remotes(conan_api, args):
 @conan_subcommand(formatters=list_recipe_revisions_formatters)
 def list_recipe_revisions(conan_api, parser, subparser, *args):
     """
-    List all the revisions of a recipe
+    List all the revisions of a recipe reference.
     """
     _add_common_list_subcommands(subparser, "reference")
     args = parser.parse_args(*args)
 
-    if not args.cache and not args.remote and not args.all_remotes:
-        # If neither remote nor cache are defined, show results only from cache
-        args.cache = True
-
-    remotes = _get_remotes(conan_api, args)
-
+    use_remotes = any([args.remote, args.all_remotes])
     results = []
-    if args.cache:
-        result = {
-            'remote': None,
-            'reference': args.reference,
-            'results': conan_api.get_recipe_revisions(args.reference)
-        }
+
+    # If neither remote nor cache are defined, show results only from cache
+    if args.cache or not use_remotes:
+        result = conan_api.get_recipe_revisions(args.reference)
         results.append(result)
 
-    for remote in remotes:
-        result = {
-            'remote': remote.name,
-            'reference': args.reference,
-            'results': conan_api.get_recipe_revisions(args.reference, remote=remote)
-        }
-        results.append(result)
+    if use_remotes:
+        remotes = _get_remotes(conan_api, args)
+        for remote in remotes:
+            result = conan_api.get_recipe_revisions(args.reference, remote=remote)
+            results.append(result)
 
     return results
 
@@ -183,33 +185,24 @@ def list_recipe_revisions(conan_api, parser, subparser, *args):
 @conan_subcommand(formatters=list_package_revisions_formatters)
 def list_package_revisions(conan_api, parser, subparser, *args):
     """
-    List all the revisions of a package ID
+    List all the revisions of a package ID reference.
     """
     _add_common_list_subcommands(subparser, "package_reference")
     args = parser.parse_args(*args)
 
-    if not args.cache and not args.remote and not args.all_remotes:
-        # If neither remote nor cache are defined, show results only from cache
-        args.cache = True
-
-    remotes = _get_remotes(conan_api, args)
-
+    use_remotes = any([args.remote, args.all_remotes])
     results = []
-    if args.cache:
-        result = {
-            'remote': None,
-            'package_reference': args.package_reference,
-            'results': conan_api.get_package_revisions(args.package_reference)
-        }
+
+    # If neither remote nor cache are defined, show results only from cache
+    if args.cache or not use_remotes:
+        result = conan_api.get_package_revisions(args.package_reference)
         results.append(result)
 
-    for remote in remotes:
-        result = {
-            'remote': remote.name,
-            'package_reference': args.package_reference,
-            'results': conan_api.get_package_revisions(args.package_reference, remote=remote)
-        }
-        results.append(result)
+    if use_remotes:
+        remotes = _get_remotes(conan_api, args)
+        for remote in remotes:
+            result = conan_api.get_package_revisions(args.package_reference, remote=remote)
+            results.append(result)
 
     return results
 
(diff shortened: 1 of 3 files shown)


conans/cli/commands/list.py
```diff
@@ -1,67 +1,83 @@
...
-                recipe_color
-            )
+
+def list_package_revisions_cli_formatter(results):
+    _list_revisions_cli_formatter(results, "packages")
 
 
+def list_package_ids_cli_formatter(results):
+    # Artifactory uses field 'requires', conan_center 'full_requires'
+    requires_fields = ("requires", "full_requires")
```
memsharded, line 61: This will be changed too withe new package_id, I know the ``full_requires`` thing is weird.

conans/client/api/conan_api.py
```diff
@@ -239,39 +241,120 @@ def search_remote_recipes(self, query, remote):
...
+                  "reference": "libcurl/7.77.0#2a9c4fcc8d76d891e4db529efbe24242",
+                  "results": {"d5f16437dd4989cc688211b95c24525589acaafd": {
+                                    "settings": {"compiler": "apple-clang",...},
+                                    "options": {'options': {'shared': 'False',...}},
+                                    "requires": [
+                                        'mylib/1.0.8:3df6ebb8a308d309e882b21988fd9ea103560e16',...]}
+                            }
+                 }
+        """
+        ref = ConanFileReference.loads(reference)
```
memsharded, line 356: When this gets too heavy on business logic, we should delegate the code to helper files.
No need to change this here, we will do it when we check the whole api.

conans/test/integration/command_v2/list_package_ids_test.py
```diff
@@ -0,0 +1,180 @@
...
+    def _upload_recipe(self, remote, reference):
+        self.client.save({'conanfile.py': GenConanfile()})
+        self.client.run("export . {}".format(reference))
+        self.client.run("upload --force -r {} {}".format(remote, reference))
+
+
+class TestParams(TestListPackageIdsBase):
+    def test_fail_if_reference_is_not_correct(self):
+        self.client.run("list package-ids whatever", assert_error=True)
+        assert "ERROR: Specify the 'name' and the 'version'" in self.client.out
```
memsharded, line 51: We need to improve the UX (not necessarily here). Error message should probably be on the form:

``"whatever" is not a valid reference, provide a reference in the form "name/version"``

conans/test/integration/command_v2/list_package_ids_test.py
```diff
@@ -0,0 +1,180 @@
...
+        remote_name = "remote1"
+        recipe_name = "test_recipe/1.0.0@user/channel"
+        self._add_remote(remote_name)
+        self._upload_recipe(remote_name, recipe_name)
+
+        self.client.run("list package-ids -r remote1 test_recipe/1.0.0")
+
+        expected_output = (
+            "remote1:\n"
+            "  There are no matching recipes\n"
```
memsharded: I cannot find a test with recipe but without binaries: What is the output in that case?

memsharded (approved):
This is good enough too, can be merged.

There is some engineering work to do, but that can wait until functionality stabilizes a bit more.

conans/test/integration/command_v2/list_package_ids_test.py
```diff
@@ -0,0 +1,180 @@
...
+        self.client.run("create . {}".format(reference))
+
+    def _upload_recipe(self, remote, reference):
+        self.client.save({'conanfile.py': GenConanfile()})
+        self.client.run("export . {}".format(reference))
+        self.client.run("upload --force -r {} {}".format(remote, reference))
+
+
+class TestParams(TestListPackageIdsBase):
+    def test_fail_if_reference_is_not_correct(self):
```
czoido, line 49: We are not testing a reference that does not exist in cache but has a valid format, like:
`conan list package-ids "notexisting/1.0@user/channel#whatever_not_existing"`
This is not returning an error.

conans/cli/commands/list.py
```diff
@@ -1,67 +1,83 @@
...
+    general_fields = ("options", "settings")
+
+    for remote_results in results:
+        _print_common_list_cli_output(remote_results, "references")
+        reference = remote_results["reference"]
+        for pkg_id, props in remote_results["results"].items():
+            cli_out_write(repr(PackageReference(reference, pkg_id)), fg=reference_color, indentation=2)
+            for prop_name, values in props.items():
+                if prop_name in requires_fields:
```
czoido: I think we should check here if `values` is empty or not so that it does not print the title like `requires:` if there are no requires.

conans/test/integration/command_v2/list_package_ids_test.py
franramirez688: Agree. This is coming from `ref.py` so it should go in another PR 😉

conans/test/integration/command_v2/list_package_ids_test.py
franramirez688: @czoido we're doing that plenty of times along the test suite. For example, `test_by_default_search_only_in_cache` is checking:

```
        Local Cache:
          There are no matching references
```
So it should be a valid test 😉  This test is done for cache and remote as well.

czoido: That test is failing in the PR: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuitev2/detail/PR-9310/2/pipeline/34
Anyway, I think if you want to check the package-id's of a specific recipe revision that does not exist in the cache, we should raise an error not returning that there are not matching recipes but saying that that recipe revision is not in the cache.

conans/client/api/conan_api.py
```diff
@@ -268,23 +269,100 @@ def _get_revisions(self, ref, remote=None):
...
+            try:
+                packages_props = self.app.remote_manager.search_packages(remote, rrev, None)
+            except NotFoundException:
+                # This exception must be catched manually due to a server inconsistency:
+                # Artifactory API returns an empty result if the recipe doesn't exist, but
+                # Conan Server returns a 404. This probably should be fixed server side,
+                # but in the meantime we must handle it here
+                packages_props = {}
+        else:
+            rrev = ref if ref.revision else self.app.cache.get_latest_rrev(remote, ref)
```
czoido: This method does not have the remote argument.

czoido (changes requested):
I think we are missing some tests, check how this one fails:
```python
def test_existing_reference(self):
    self.client.save({"conanfile.py": GenConanfile().with_name("pkg").with_version("1.0")})
    self.client.run(f"create .")
    self.client.run(f"list package-ids pkg/1.0")
```

conans/client/api/conan_api.py
franramirez688: Completely right. The thing is that I have a test for cache and remote cases but as you said I'm missing something... Thanks for the catch @czoido

conans/test/integration/command_v2/list_package_ids_test.py
franramirez688: I tend to say that this check could go in another PR because it should be something that affects to all the list context, not only the `package-ids` subcommand

memsharded (approved):

conans/cli/commands/list.py
```diff
@@ -40,7 +41,7 @@ def list_recipes_cli_formatter(results):
 def _list_revisions_cli_formatter(results, ref_type):
     for remote_results in results:
         _print_common_list_cli_output(remote_results, ref_type)
```
tapia: This function's name (`_print_common_list_cli_output`) is not very descriptive. How about `_print_remote_header()` or something like that?

franramirez688: That function is printing the `Local Cache` line as well so I did not know which could be the best name... 🤔  Anyway, let me change the var name `remote_results` to `result` because it does not make really sense

conans/cli/commands/list.py
tapia: Sorry for reopening this, but I think it would be cleaner if you check `values` only once:

```python
for prop_name, values in props.items():
    if not values:
        continue

    if prop_name in requires_fields:
        blablabla
    elif prop_name in general_fields:
        blablabla
```

conans/cli/commands/list.py
```diff
@@ -134,68 +163,76 @@ def _get_remotes(conan_api, args):
...
+
+    return results
+
 
-    remotes = _get_remotes(conan_api, args)
+@conan_subcommand(formatters=list_package_ids_formatters)
+def list_package_ids(conan_api, parser, subparser, *args):
+    """
+    List all the package IDs for a given recipe reference. If it doesn't include the revision,
+    the command will retrieve all the package IDs for the latest one by default.
```
tapia: A little confusing. I'd rephrase it slightly:

List all the package IDs for a given recipe reference. If the reference doesn't include the recipe revision, the command will retrieve all the package IDs for the most recent revision.

conans/test/integration/command_v2/list_package_ids_test.py
tapia: As @franramirez688 says, this is not related to this PR. I just created [an issue](https://github.com/conan-io/conan/issues/9332) to fix this.

conans/test/integration/command_v2/list_package_ids_test.py
```diff
@@ -0,0 +1,277 @@
...
+          %s:.{40}
+            settings:
+              arch=.*
+              build_type=.*
+              os=.*
+            options:
+              shared=False
+            requires:
+              pkg/0.1@user/channel:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
+        """ % repr(rrev))
```
tapia, line 157: Why are you building all the expected outputs with `%` formatting?

franramirez688: If you realize I have one line like this `%s:.{40}` so I don't want to have this `40` as a `f-string` value but a regex one

conans/client/api/conan_api.py
```diff
@@ -268,23 +269,100 @@ def _get_revisions(self, ref, remote=None):
...
+        Get all the recipe revisions given a reference from cache or remote.
+
+        :param reference: `PackageReference` without the revision
+        :param remote: `Remote` object
+        :return: `dict` with all the results, e.g.,
+                 {"remote": "my_remote_name",  # or None
+                  "reference": "libyaml/0.2.5#80b7cbe095ac7f38844b6511e69e453a:ef93ea55bee154729e264db35ca6a16ecab77eb7",
+                  "results": [{"revision": "80b7cbe095ac7f38844b6511e69e453a",
+                               "time": "2021-07-20 00:56:25 UTC"}, ...]}
+        """
```
tapia, line 294: please reformat the example to be more easily read (Also in `get_recipe_revisions` and `get_package_ids`:

```
:return: `dict` with all the results, e.g.,
  {
    "remote": "my_remote_name",  # or None
    "reference": "libyaml/0.2.5#80b7cbe095ac7f38844b6511e69e453a:ef93ea55bee154729e264db35ca6a16ecab77eb7",
    "results": [
      {
        "revision": "80b7cbe095ac7f38844b6511e69e453a",
        "time": "2021-07-20 00:56:25 UTC"
      },
      ...
    ]
  }
```

conans/cli/commands/list.py
tapia: I'm not sure this function makes a lot of sense. Besides refactoring repeated code, what does it do? Print the header? Handle errors? Show a message when the results are empty?

I mean, if you had to write the docstring of the function, what would it say?

franramirez688: It's only supposed to avoid repeating code so I think it makes sense to have it. Perhaps the best name is `_print_common_headers()`, what do you think @tapia ?

franramirez688: I've renamed it as `_print_common_headers`

PR #9348, opened 2021-07-29
Feature/develop2 remove old env

memsharded:
Changelog: Remove: Remove old environment management
Docs: Omit


memsharded:
Can @czoido, @franramirez688, @tapia  or @lasote help with the OSX tests that fail?

conans/client/profile_loader.py
```diff
@@ -225,13 +225,6 @@ def get_package_name_value(item):
     if doc.options:
         base_profile.options.update(OptionsValues.loads(doc.options))
 
-    # The env vars from the current profile (read in doc)
-    # are updated with the included profiles (base_profile)
-    # the current env values has priority
-    current_env_values = EnvValues.loads(doc.env)
-    current_env_values.update(base_profile.env_values)
-    base_profile.env_values = current_env_values
-
```
lasote, line 234: https://github.com/conan-io/conan/pull/9348/files#diff-711453b1a75934b6fa7f423dc82a3363a385a278e45ab7516105c67c60a82edfR162-R165

Should we remove the section [env] from there?

conans/model/profile.py
```diff
@@ -84,7 +82,7 @@ def dumps(self):
             result.append("%s: %s" % (pattern, ", ".join(str(r) for r in req_list)))
 
         result.append("[env]")
```
lasote, line 84: Why are you leaving the empty section?

lasote (approved):
It is great to get rid of it.

czoido (approved):

conans/model/profile.py
memsharded: it required to change even more tests, I preferred to do empty output [env], this can be removed in a later PR. Same as the ``no_op()`` everywhere, to reduce the diff.

conans/client/profile_loader.py
memsharded: The link doesn't point to a line in the diff, could you clarify it?

PR #9350, opened 2021-07-30
New version ranges resolution logic for conan 2.0 with --update

czoido:
Changelog: Feature: New version ranges resolution logic for conan 2.0.
Docs: TODO (2.0 docs)

Extracted version ranges part from: https://github.com/conan-io/conan/pull/9191

This PR fixes some not clear behaviours when using version ranges and also using the --update argument. The logic should be exactly the same than [the one used with revisions](https://github.com/conan-io/conan/pull/9191#issuecomment-882256402) but translated to version ranges.

Version ranges are always resolved before revisions. So for example requiring a range [>1.0.0] with --update will check all the remotes and take the latest version that satisfies the range, for example 1.3, then if that version is in several remotes it will take the latest from all of them in the proxy. TODO: optimize all this flow to make as few API calls as we can.

### DIFF
--- a/conans/client/graph/range_resolver.py
+++ b/conans/client/graph/range_resolver.py
@@ -1,5 +1,4 @@
 import re
-from collections import defaultdict
 
 from conans.errors import ConanException
 from conans.model.ref import ConanFileReference
--- a/conans/client/graph/range_resolver.py
+++ b/conans/client/graph/range_resolver.py
@@ -119,12 +119,15 @@ class RangeResolver(object):
         search_ref = ConanFileReference(ref.name, "*", ref.user, ref.channel)
 
         remote_name = None
+        remote_resolved_ref = None
         resolved_ref = self._resolve_local(search_ref, version_range)
         if not resolved_ref or update:
-            resolved_ref, remote_name = self._resolve_remote(search_ref, version_range, remotes,
-                                                             check_all_remotes=update)
+            remote_resolved_ref, remote_name = self._resolve_remote(search_ref, version_range, remotes,
+                                                                    check_all_remotes=update)
+            resolved_ref = self._resolve_version(version_range, [resolved_ref, remote_resolved_ref])
 
-        origin = ("remote '%s'" % remote_name) if remote_name else "local cache"
+        origin = f"remote '{remote_name}'" if resolved_ref == remote_resolved_ref and remote_name else "local cache"
+
         if resolved_ref:
             self._result.append("Version range '%s' required by '%s' resolved to '%s' in %s"
                                 % (version_range, base_conanref, str(resolved_ref), origin))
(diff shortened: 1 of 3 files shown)


conans/client/graph/range_resolver.py
```diff
@@ -147,37 +145,50 @@ def _resolve_local(self, search_ref, version_range):
...
             # Empty list, just in case it returns None
             self._cached_remote_found[search_ref] = found_refs, remote_name
-        if found_refs:
-            return self._resolve_version(version_range, found_refs), remote_name
-        return None, None
+        return (found_refs, remote_name) if found_refs else (None, None)
 
     def _resolve_version(self, version_range, refs_found):
-        versions = {ref.version: ref for ref in refs_found}
+        versions = {ref.version: ref for ref in refs_found if ref}
```
memsharded, line 203: This ref is a bit concerning, can it be None? From where?

memsharded (commented):
This is looking good, but I'd say it is necessary to add another change: do cache the remote refs **per remote**. This seems to be caching only the things from the successful remote.

conans/client/graph/range_resolver.py
czoido: I added that because of this addition: https://github.com/conan-io/conan/blob/0c0c2e28972ab9acd7f2add0ab21012a27c2c638/conans/client/graph/range_resolver.py#L125-L127
I thought that maybe if that code does not resolve a reference in the remotes this could fail

conans/client/graph/range_resolver.py
```diff
@@ -118,18 +118,16 @@ def resolve(self, require, base_conanref, update, remotes):
...
-                resolved_ref, remote_name = self._resolve_remote(search_ref, version_range, remotes)
+        remote_name = None
+        remote_resolved_ref = None
+        resolved_ref = self._resolve_local(search_ref, version_range)
+        if not resolved_ref or update:
+            remote_resolved_ref, remote_name = self._resolve_remote(search_ref, version_range, remotes,
+                                                                    check_all_remotes=update)
+            resolved_ref = self._resolve_version(version_range, [resolved_ref, remote_resolved_ref])
+
+        origin = f"remote '{remote_name}'" if resolved_ref == remote_resolved_ref and remote_name else "local cache"
```
memsharded, line 147: Next step, in another PR is to remove this ``self._result`` custom output.

conans/client/graph/range_resolver.py
```diff
@@ -147,37 +145,65 @@ def _resolve_local(self, search_ref, version_range):
...
-                    return result, remote.name
-        return None, None
-
-    def _resolve_remote(self, search_ref, version_range, remotes):
-        # We should use ignorecase=False, we want the exact case!
-        found_refs, remote_name = self._cached_remote_found.get(search_ref, (None, None))
-        if found_refs is None:
+                cached_search = self._cached_remote_found.get(search_ref)
+                if cached_search:
+                    cached_refs = cached_search.get(remote.name) or []
```
memsharded: You can do ``.get(remote.name, [])``

conans/client/graph/range_resolver.py
```diff
@@ -147,37 +145,65 @@ def _resolve_local(self, search_ref, version_range):
...
+                    if resolved_ref in cached_refs:
+                        resolved_remote = remote.name
+                    if not check_all_remotes and resolved_ref:
+                        return resolved_ref, resolved_remote
+        return resolved_ref, resolved_remote
+
+    def _resolve_remote(self, search_ref, version_range, remotes, check_all_remotes):
+        resolved_ref, remote_name = self._resolve_cached_reference(search_ref, version_range,
+                                                                   remotes, check_all_remotes)
+        if resolved_ref is None:
```
memsharded: I am not sure the caching is correct. The ``resolved_ref`` might still be ``None``, because it was not matching the previous version range query, but the ``pkg/*@user/channel`` versions from remote were indeed retrieved and cached, and we don't want to query them again to the server.

PR #9355, opened 2021-07-30
trying to avoid touching remote when binary not found in -r=remote

### LINKED ISSUE #9333: [question] What is the expected behavior of `conan install --remote REMOTE`
I'm trying to understand the expected behavior of `conan install --remote REMOTE`. We have a local Artifactory server for our own packages, but still have `conan-center` as a remote, even though it is not really used.

Output of `conan remote list`:

```
virtual-remote: https://[...] [Verify SSL: True]
conan-center: https://conan.bintray.com [Verify SSL: True]
```

Whether or not I call `conan install --remote virtual-remote` or  just `conan install`, I always get these "extra" calls in the trace log.

```
{"_action": "REST_API_CALL", "duration": 0.21838855743408203, "headers": {"User-Agent": "Conan/1.37.1 (Windows 10; Python 3.9.5; AMD64) python-requests/2.25.1", "X-Client-Anonymous-Id": "**********", "X-Client-Id": ""}, "method": "GET", "time": 1627460264.1806884, "url": "https://conan.bintray.com/v1/ping"}
{"_action": "REST_API_CALL", "duration": 0.016086578369140625, "headers": {"User-Agent": "Conan/1.37.1 (Windows 10; Python 3.9.5; AMD64) python-requests/2.25.1", "X-Client-Anonymous-Id": "**********", "X-Client-Id": ""}, "method": "GET", "time": 1627456231.6621335, "url": "https://conan.bintray.com/v2/conans/[...]/revisions/35c21a513276d9605b75da32df276b87/packages/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/latest"}
```

Is this to be expected and is there any way to avoid it (besides deleting the remote?)

Versions:
* `Python 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)] on win32`
* `Conan version 1.37.1`

- [y] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Edit: added Python and Conan version

memsharded:
Changelog: Fix: Avoid checking other remotes when ``-r=remote`` is defined and revisions are activated and binary is not found in the defined remote.
Docs: Omit

Attempt to fix https://github.com/conan-io/conan/issues/9333, not sure if this is a bug or expected behavior, lets see if this change breaks something.

Close https://github.com/conan-io/conan/issues/9333

#REVISIONS: 1

### DIFF
--- a/conans/test/integration/remote/test_request_headers.py
+++ b/conans/test/integration/remote/test_request_headers.py
@@ -50,7 +50,7 @@ class RequestHeadersTestCase(unittest.TestCase):
 
     def _get_header(self, requester, header_name):
         hits = sum([header_name in headers for _, headers in requester.requests])
-        assert hits == 2 if self.revs_enabled else 1
+        assert hits <= 2 if self.revs_enabled else 1
         for url, headers in requester.requests:
             if header_name in headers:
                 if self.revs_enabled:
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -101,11 +101,10 @@ class GraphBinariesAnalyzer(object):
         return self._remote_manager.get_package_info(pref, remote, info=node.conanfile.info)
 
     def _evaluate_remote_pkg(self, node, pref, remote, remotes, remote_selected):
-        """
-        :param remote_selected: bool, The user specified a remote with -r
-        """
         remote_info = None
-        if remote_selected:
+        # If the remote is pinned (remote_selected) we won't iterate the remotes.
+        # The "remote" can come from -r or from the registry (associated ref)
+        if remote_selected or remote:
             try:
                 remote_info, pref = self._get_package_info(node, pref, remote)
             except NotFoundException:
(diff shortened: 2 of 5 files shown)


lasote (commented):
I don't see today why I introduced that "or" and doesn't look good to search other remotes if you specified "-r". Said that we are highly probably breaking something, but we have to try.
I'm adding the `#REVISIONS: 1` to see if some test give us feedback.

lasote:
As I expected there are some tests failing apparently not using `-r`. I'm going to check why.

conans/test/integration/remote/test_request_headers.py
```diff
@@ -50,7 +50,7 @@ def setUp(self):
 
     def _get_header(self, requester, header_name):
         hits = sum([header_name in headers for _, headers in requester.requests])
-        assert hits == 2 if self.revs_enabled else 1
+        assert hits <= 2 if self.revs_enabled else 1
```
lasote, line 53: Because of the removed: `# FIXME: Here we hit the same remote we did before`

memsharded:
> I don't see today why I introduced that "or" and doesn't look good to search other remotes if you specified "-r". Said that we are highly probably breaking something, but we have to try.

Seems your fixes pass the new test and not break other existing tests, so it looks good. I understand this might be breaking some users abusing this behavior, but this totally reads like a bug, so I agree we need to try to fix it.

lasote:
Yes, I iterated a bit pushing commits until I finally find something that doesn't look very breaking.

memsharded (commented):
This will conflict with @czoido work on updates for Conan 2.0, asking for a review, just in case.

czoido (approved):
This makes sense, I'll update my develop2 PR according to this when merged

PR #9356, opened 2021-07-30
caching also cache status in range resolver

memsharded:
Changelog: Omit
Docs: Omit


czoido (approved):
I think this is more expected behaviour than having different results depending on the order of declaration of the requirements.

lasote (approved):

PR #9361, opened 2021-08-02
Feature/patches following layout

lasote:
Changelog: Fix: The tool `conan.tools.files.apply_conandata_patches` will use the root source folder to find the patch file and the tool `conan.tools.files.patch` will take the current source folder declared in the `layout()` method to know where is the source to apply the patches.
Docs: https://github.com/conan-io/docs/pull/2207

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -226,7 +226,7 @@ def save_toolchain_args(content, generators_folder=None):
 
 
 @contextmanager
-def chdir(newdir):
+def chdir(conanfile, newdir):
     old_path = os.getcwd()
     os.chdir(newdir)
     try:
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -3,6 +3,7 @@ import errno
 import os
 import platform
 import subprocess
+from contextlib import contextmanager
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
 from conans.client.downloaders.download import run_downloader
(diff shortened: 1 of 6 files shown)


conan/tools/files/patches.py
```diff
@@ -37,8 +38,10 @@ def patch(conanfile, base_path=None, patch_file=None, patch_string=None,
     :param kwargs: Extra parameters that can be added and will contribute to output information
     """
 
+    base_path = base_path or conanfile.source_folder
```
lasote: This might have some risk.

madebr: This makes patching depend on `no_copy_source`.
I think the behavior of `tools.patch` we want to see is to apply the patches relative to the working directory.

`apply_conandata_patches` can contain some logic to patch sources relative to the base source folder.

lasote: `source_folder` always points to the right location, if `no_copy_source` if points to source, otherwise it points to `build_folder`

madebr: So this will break:
```python
class CC(ConanFile):
    no_copy_source = True
    def build(self):
        with tools.chdir(self.build_folder):
            tools.patch("my-patch.patch")
```

lasote: yes, because the patches are not in the build folder.

lasote:
TODO: Check with local methods too, I think there is still something weird.

conan/tools/files/files.py
```diff
@@ -222,3 +223,13 @@ def save_toolchain_args(content, generators_folder=None):
     toolchain_config[CONAN_TOOLCHAIN_ARGS_SECTION] = content_
     with open(args_file, "w") as f:
         toolchain_config.write(f)
+
+
+@contextmanager
+def chdir(newdir):
```
memsharded: ``mkdir()`` uses ``conanfile`` as first argument, shouldn't we follow a similar pattern for ``chdir()``? Or remove the ``mkdir`` first argument.

conan/tools/files/patches.py
memsharded: What is the root of the ``base_path``? Shouldn't it be relative to some folder of the layout?
Recall this ``conan.tools.files patches`` is experimental, not used yet, we can change interface and behavior here, not a huge risk.

memsharded (approved):

PR #9371, opened 2021-08-03
Use `frameworks` in Premake generator

### LINKED ISSUE #9367: [bug] Premake generator does not use frameworks
  * Operating System+version: MacOS
  * Compiler+version:
  * Conan version: 1.39.0
  * Python version:

The `premake` generator does not take `frameworks` into account for MacOS. For Premake5, these need to be added as `links` with a `.framework` suffix: https://premake.github.io/docs/links#examples

sheepsteak:
Fixes #9367 

Changelog: Fix: Use `frameworks` in Premake generator.
Docs: https://github.com/conan-io/docs/pull/2210

This PR allows the system frameworks in a package like SDL2 to now be associated with Premake. There is still the issue with how to use `frameworkdirs`. It seems like Premake5 does support a `framworkdirs` property itself but it's only used for the Xcode generator. There is a PR (https://github.com/premake/premake-core/pull/1661) to add this to the `gmake` generator, however.

Although after looking at the `make` generator I'm a bit confused about `frameworkdirs`. The `conanbuildinfo.txt` has `framework_dirs` but the `make` generator (and unit tests) seem to use `framework_paths`. Are they the same thing? 

### DIFF
--- a/conans/client/generators/premake.py
+++ b/conans/client/generators/premake.py
@@ -18,6 +18,7 @@ class PremakeDeps(object):
         self.cflags = ", ".join('"%s"' % p for p in deps_cpp_info.cflags)
         self.sharedlinkflags = ", ".join('"%s"' % p.replace('"', '\\"') for p in deps_cpp_info.sharedlinkflags)
         self.exelinkflags = ", ".join('"%s"' % p.replace('"', '\\"') for p in deps_cpp_info.exelinkflags)
+        self.frameworks = ", ".join('"%s.framework"' % p.replace('"', '\\"') for p in deps_cpp_info.frameworks)
 
         self.rootpath = "%s" % deps_cpp_info.rootpath.replace("\\", "/")
 
@@ -40,7 +41,8 @@ class PremakeGenerator(Generator):
                     'conan_cxxflags{dep} = {{{deps.cxxflags}}}\n'
                     'conan_cflags{dep} = {{{deps.cflags}}}\n'
                     'conan_sharedlinkflags{dep} = {{{deps.sharedlinkflags}}}\n'
-                    'conan_exelinkflags{dep} = {{{deps.exelinkflags}}}\n')
+                    'conan_exelinkflags{dep} = {{{deps.exelinkflags}}}\n'
+                    'conan_frameworks{dep} = {{{deps.frameworks}}}\n')
 
         sections = ["#!lua"]
 
@@ -68,6 +70,7 @@ class PremakeGenerator(Generator):
             "    libdirs{conan_libdirs}\n"
             "    links{conan_libs}\n"
             "    links{conan_system_libs}\n"
+            "    links{conan_frameworks}\n"
             "    defines{conan_defines}\n"
             "    bindirs{conan_bindirs}\n"
             "end\n")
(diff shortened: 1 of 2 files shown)


lasote:
Thank you so much for helping with this!! Looks good!
Please, to accept the PR we need you to sign the CLA.

SSE4 (approved):
sounds reasonable
this is unfortunate premake doesn't support framework dirs.
we may still populate `conan_framework_dirs{dep}` vars just in case (someone manually specify `-F` flags)

SSE4:
```
Although after looking at the make generator I'm a bit confused about frameworkdirs. The conanbuildinfo.txt has framework_dirs but the make generator (and unit tests) seem to use framework_paths.
```
`frameworkdirs` is a value you populate within `package_info` method, e.g.:
```
self.cpp_info.frameworkdirs = ['Frameworks']
```
it's usually a relative paths to your the root directory of your package (theoretically, might be an absolute path as well, but I don't see good use-case for that at the moment)
`framework_paths` is the final value computed, which is a full, an absolute path to the directory in your local cache. this one is intended to be passed into the compiler/linker, and conan supplies that into generator (thus, files like `conanbuildinfo.cmake` contain such final paths).

that's more or less the same for other directories:
- `includedirs` -> `include_paths`
- `libdirs` -> `lib_paths`
- `bindirs` -> `bin_paths`
- `resdirs` -> `res_paths`

I think you see the logic, it's just a convention in our `CppInfo` model.

uilianries:
@sheepsteak Thank you for your contribution. Please, don't forget to sign Conan's CLA as described [here](https://github.com/conan-io/conan/pull/9371#issuecomment-892205138). Otherwise, this PR could not be merged.

uilianries (approved):
Agreed with @SSE4 

The Premake's documentation shows the supported dirs [here](https://premake.github.io/docs/Build-Settings)

Premake doesn't list a `frameworkdir`, but it also doesn't separate Mac frameworks, instead, it uses [links](https://premake.github.io/docs/links) and it's indicated to use absolute path. However, we can consider `libdirs` in this case.

The [framework](https://premake.github.io/docs/framework) feature supported by Premake is related to .NET. I hope people don't get confused about it.

memsharded (commented):
If I understood this correctly, this will not work for packaged frameworks, only for system frameworks?

Because the frameworks_dir variable is simply not managed, not considered at all here.

SSE4:
> If I understood this correctly, this will not work for packaged frameworks, only for system frameworks?
> 
> Because the frameworks_dir variable is simply not managed, not considered at all here.

yes. we can do nothing for packaged frameworks, as it's not supported by the build system.

memsharded:
> yes. we can do nothing for packaged frameworks, as it's not supported by the build system.

> Premake doesn't list a frameworkdir, but it also doesn't separate Mac frameworks, instead, it uses links and it's indicated to use absolute path. However, we can consider libdirs in this case.

@uilianries  you saying that could in-package frameworks work if the generator provided full paths to the framework somehow?

uilianries:
@memsharded if a framework is a single file, yes. But considering [link example](https://premake.github.io/docs/links/#examples) using Cocoa.Framework and based on old Conan behavior (before 1.22), only the framework name should work, since the `DYLD_LIBRARY_PATH` is configured correctly.

Well, actually `DYLD_FRAMEWORK_PATH` should be used, but again Premake doesn't have support for it, instead, it configures the regular `DYLD_LIBRARY_PATH` only.

PR #9374, opened 2021-08-04
post package-install in cache

lasote:
Changelog: Feature: New conanfile `post_package` method to patch packages after it is installed in the cache.

PENDING NAMING

### DIFF
--- a/conan/cache/conan_reference_layout.py
+++ b/conan/cache/conan_reference_layout.py
@@ -12,6 +12,7 @@ from conans.util.files import set_dirty, clean_dirty, is_dirty, rmdir
 SRC_FOLDER = "source"
 BUILD_FOLDER = "build"
 PACKAGES_FOLDER = "package"
+PACKAGES_INSTALL_FOLDER = "install"
 EXPORT_SRC_FOLDER = "export_source"
 SYSTEM_REQS_FOLDER = "system_reqs"
 SCM_SRC_FOLDER = "scm_source"
@@ -121,6 +122,9 @@ class PackageLayout(LayoutBase):
     def package(self):
         return os.path.join(self.base_folder, PACKAGES_FOLDER)
 
+    def post_package(self):
+        return os.path.join(self.base_folder, PACKAGES_INSTALL_FOLDER)
+
     def download_package(self):
         return os.path.join(self.base_folder, "dl")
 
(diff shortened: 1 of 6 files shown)


conan/cache/conan_reference_layout.py
```diff
@@ -121,6 +122,9 @@ def build(self):
     def package(self):
         return os.path.join(self.base_folder, PACKAGES_FOLDER)
 
+    def post_package(self):
```
memsharded, line 125: Lets call it "install" at the moment, not perfect, but ``post_package`` is confusing, seems like the hook and something only after ``package()`` not installation.

conans/client/installer.py
```diff
@@ -536,13 +539,28 @@ def _handle_node_cache(self, node, processed_package_references, remotes, pkg_la
...
+                package_folder = pkg_layout.package()
             assert os.path.isdir(package_folder), ("Package '%s' folder must exist: %s\n"
                                                    % (str(pref), package_folder))
             # Call the info method
             self._call_package_info(conanfile, package_folder, ref=pref.ref, is_editable=False)
             self._recorder.package_cpp_info(pref, conanfile.cpp_info)
 
+    def _call_post_package(self, conanfile, output, pkg_layout):
+        # FIXME: better name? "install" might be confusing with the old install_folder mechanism
+        # FIXME: Maybe it is better to remove first the install folder here in develop2
```
memsharded, line 555: Yes, probably removing the "install_folder" clears the way.

conans/client/installer.py
```diff
@@ -536,13 +539,28 @@ def _handle_node_cache(self, node, processed_package_references, remotes, pkg_la
...
+        # TODO: dirty?
```
memsharded, line 556: Yes, seems "post_package" needs its own dirty control check, as it can be interrupted, fail, etc.

conans/model/conan_file.py
```diff
@@ -234,6 +234,10 @@ def package_folder(self):
     def package_folder(self, folder):
         self.folders.set_base_package(folder)
 
+    @property
+    def post_package_folder(self):
```
memsharded, line 238: Maybe the layout could be obviated? Users will be defined the current "post_package" folder as the package one in the "post_package()" method?

memsharded (commented):
Just annotating some things to discuss, but overall looks good, simple enough, and that could provide value. My concern now is more about potential abuses of the feature than by the implementation itself.

puetzk:
This is a feature I've wanted a couple of times (mostly just to patch in the correct paths once I learn the location where a downloaded set of binaries was extracted into the package cache). It's obviously preferable to figure out how to make the package fully position-independent, but it can be quite hard :-)

memsharded:
Closing this, as in Conan 2.0 we will recommend:

- Packages that need to be localized to a given abs path can do whatever they need in their ``package()`` method, marking the package ``upload_policy = "skip"`` and ``build_policy = "missing"``, so the package is always built locally
- If the cost of building from source is too much, such packages can just ``require`` or ``tool_require`` other package that would be relocatable, and then copy its artefacts and localize them to the current computer. That required package can be fully uploaded and downloaded, so it doesn't need to be built from source over and over, only the "relocating" part is executed each time

This is not as integrated as the current proposal, but this use case is also very infrequent, just happening in some special tools like sdks and the like, so it sounds that adding the extra complexity of a general ``post-install`` approach would be too much at this moment. So closing this PR.

PR #9387, opened 2021-08-05
Feature/list improvements

franramirez688:
Changelog: Feature: Adding several improvements to the new `list` command:
* Handle remote errors.
* Better error messages.
* Fixing minor problems.

### DIFF
--- a/conans/cli/commands/list.py
+++ b/conans/cli/commands/list.py
@@ -3,8 +3,9 @@ import json
 from conans.cli.command import conan_command, conan_subcommand, Extender, COMMAND_GROUPS
 from conans.cli.output import cli_out_write
 from conans.client.output import Color
+from conans.errors import ConanException, InvalidNameException
+from conans.model.ref import PackageReference, ConanFileReference
 from conans.util.dates import from_timestamp_to_iso8601
-from conans.model.ref import PackageReference
 
 remote_color = Color.BRIGHT_BLUE
 recipe_color = Color.BRIGHT_WHITE
@@ -105,43 +106,6 @@ list_package_ids_formatters = {
 }
 
 
-@conan_subcommand(formatters=list_recipes_formatters)
-def list_recipes(conan_api, parser, subparser, *args):
-    """
-    Search available recipes in the local cache or in the remotes
-    """
-    subparser.add_argument(
-        "query",
-        help="Search query to find package recipe reference, e.g., 'boost', 'lib*'"
-    )
-    remotes_group = subparser.add_mutually_exclusive_group()
-    remotes_group.add_argument("-r", "--remote", default=None, action=Extender,
-                               help="Name of the remote to add")
-    remotes_group.add_argument("-a", "--all-remotes", action='store_true',
-                               help="Search in all the remotes")
-    subparser.add_argument("-c", "--cache", action='store_true', help="Search in the local cache")
-    args = parser.parse_args(*args)
-
-    if not args.cache and not args.remote and not args.all_remotes:
-        # If neither remote nor cache are defined, show results only from cache
-        args.cache = True
-
-    remotes = None
-    if args.all_remotes:
-        remotes = conan_api.get_active_remotes(None)
-    elif args.remote:
-        remotes = conan_api.get_active_remotes(args.remote)
-
-    results = []
-    if args.cache:
-        results.append(conan_api.search_local_recipes(args.query))
-    if remotes:
-        for remote in remotes:
-            results.append(conan_api.search_remote_recipes(args.query, remote))
-
-    return results
-
-
 def _add_remotes_and_cache_options(subparser):
     remotes_group = subparser.add_mutually_exclusive_group()
     remotes_group.add_argument("-r", "--remote", default=None, action=Extender,
@@ -153,11 +117,48 @@ def _add_remotes_and_cache_options(subparser):
 
 def _get_remotes(conan_api, args):
     remotes = []
+    error = None
     if args.all_remotes:
-        remotes = conan_api.get_active_remotes(None)
+        remotes, error = conan_api.get_active_remotes(None)
     elif args.remote:
-        remotes = conan_api.get_active_remotes(args.remote)
+        remotes, error = conan_api.get_active_remotes(args.remote)
+    if error:
+        raise ConanException(error)
     return remotes
+
+
+@conan_subcommand(formatters=list_recipes_formatters)
+def list_recipes(conan_api, parser, subparser, *args):
+    """
+    Search available recipes in the local cache or in the remotes
+    """
+    subparser.add_argument(
+        "query",
+        help="Search query to find package recipe reference, e.g., 'boost', 'lib*'"
+    )
+    _add_remotes_and_cache_options(subparser)
+    args = parser.parse_args(*args)
+
+    use_remotes = any([args.remote, args.all_remotes])
+    results = []
+
+    # If neither remote nor cache are defined, show results only from cache
+    if args.cache or not use_remotes:
+        result, error = conan_api.search_local_recipes(args.query)
+        results.append({
+            "error": error,
+            "results": result or []
+        })
+    if use_remotes:
+        remotes = _get_remotes(conan_api, args)
+        for remote in remotes:
+            result, error = conan_api.search_remote_recipes(args.query, remote)
+            results.append({
+                "remote": remote.name,
+                "error": error,
+                "results": result or []
+            })
+    return results
 
 
 @conan_subcommand(formatters=list_recipe_revisions_formatters)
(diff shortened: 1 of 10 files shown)


conans/client/api/conan_api.py
```diff
@@ -348,23 +360,34 @@ def get_package_ids(self, reference, remote=None):
                         "d5f16437dd4989cc688211b95c24525589acaafd": {
                             "settings": {"compiler": "apple-clang",...},
                             "options": {'options': {'shared': 'False',...}},
-                            "requires": ['mylib/1.0.8:3df6ebb8a308d309e882b21988fd9ea103560e16',...]
+                            "requires": ['mylib/1.0.8#3df6ebb8a308d309e882b21988fd9ea103560e16',...]
```
memsharded: This seems too long to be a revision, looks a package_id.

conans/client/api/conan_api.py
```diff
@@ -348,23 +360,34 @@ def get_package_ids(self, reference, remote=None):
...
-                # This exception must be catched manually due to a server inconsistency:
+            except (NotFoundException, PackageNotFoundException):
+                # This exception must be caught manually due to a server inconsistency:
                 # Artifactory API returns an empty result if the recipe doesn't exist, but
                 # Conan Server returns a 404. This probably should be fixed server side,
                 # but in the meantime we must handle it here
-                packages_props = {}
+                pass
+            except ConanConnectionError:
+                error = "There was a connection problem."
```
memsharded: This is discarding possibly interesting information about the connection error (timeout, auth, connection reset, network down...)

conans/client/api/conan_api.py
franramirez688: My bad, I did not want to change this one 🤦

conans/client/api/conan_api.py
franramirez688: You're completely right. I have to perform some new tests so that's the reason I did not realize it.

conans/client/api/conan_api.py
```diff
@@ -92,26 +93,36 @@ def load_remotes(self, remote_name=None, update=False, check_updates=False):
...
+        except ConanConnectionError as e:
+            error = f"There was a connection problem: {str(e)}"
+        except BaseException as e:
+            error = str(e)
         finally:
             if old_curdir:
                 os.chdir(old_curdir)
-
+            # Let's return the results and any possible error
+            return results, error
```
franramirez688: @memsharded @lasote let me know your opinion about this `return` change in this decorator. In my opinion, it was the most general and simplest way to manage all the possible errors. It lets all the API consumers manage any error as they really want but I still have doubts about any other better possibility to do it.

conans/cli/commands/list.py
```diff
@@ -104,6 +119,27 @@ def json_formatter(info):
...
+
+def _get_remotes(conan_api, args):
+    remotes = []
+    error = None
+    if args.all_remotes:
+        remotes, error = conan_api.get_active_remotes(None)
+    elif args.remote:
+        remotes, error = conan_api.get_active_remotes(args.remote)
+    if error:
+        raise ConanException(error)
```
tapia: If the API method returns an error, why delegate the exception to the caller? Let the API method raise the exception instead of returning a tuple

conans/cli/commands/search.py
```diff
@@ -19,9 +19,16 @@ def search(conan_api, parser, *args, **kwargs):
                         help="Remote to search. Accepts wildcards. To search in all remotes use *")
     args = parser.parse_args(*args)
 
-    remotes = conan_api.get_active_remotes(args.remote)
+    remotes, error = conan_api.get_active_remotes(args.remote)
+    if error:
+        raise ConanException(error)
```
tapia: If the API method returns an error, why delegate the exception to the caller? Let the API method raise the exception instead of returning a tuple

conans/cli/commands/user.py
```diff
@@ -28,7 +28,8 @@ def user_list(conan_api, parser, subparser, *args):
                                 "If no remote is specified it will show the users for all "
                                 "the remotes")
     args = parser.parse_args(*args)
-    info = conan_api.user_list(args.remote)
+    # FIXME: Handle errors
+    info, error = conan_api.user_list(args.remote)
```
tapia: If the API method returns an error, why delegate the exception to the caller? Let the API method raise the exception instead of returning a tuple

conans/client/api/conan_api.py
tapia: If the API method returns an error, why delegate the exception to the caller? Let the API method raise the exception. Returning a tuple with the error is not very pythonic.

conans/cli/commands/list.py
```diff
@@ -40,7 +43,13 @@ def list_recipes_cli_formatter(results):
...
 def _list_revisions_cli_formatter(results, ref_type):
     for result in results:
-        _print_common_headers(result, ref_type)
+        _print_common_headers(result)
+        if result.get("error"):
+            cli_out_write(f"ERROR: {result['error']}", fg=error_color, indentation=2)
+            continue
+        elif not result.get("results"):
+            cli_out_write(f"There are no matching {ref_type}", indentation=2)
+            continue
```
tapia: This two `continue` look weird. How about this:

```python
for result in results:
    _print_common_headers(result)

    if result.get("error"):
        cli_out_write(f"ERROR: {result['error']}", fg=error_color, indentation=2)
    elif not result.get("results"):
        cli_out_write(f"There are no matching {ref_type}", indentation=2)
    else:
        reference = result["reference"]
        for revisions in result["results"]:
            rev = revisions["revision"]
```

conans/cli/commands/list.py
```diff
@@ -62,22 +71,28 @@ def list_package_ids_cli_formatter(results):
...
 
     for result in results:
-        _print_common_headers(result, "references")
+        _print_common_headers(result)
+        if result.get("error"):
+            cli_out_write(f"ERROR: {result['error']}", fg=error_color, indentation=2)
+            continue
+        elif not result.get("results"):
+            cli_out_write("There are no matching recipe references", indentation=2)
+            continue
```
tapia: Same here. This two `continue` should be just an `else`

conans/cli/commands/search.py
```diff
@@ -20,8 +20,19 @@ def search(conan_api, parser, *args, **kwargs):
     args = parser.parse_args(*args)
 
     remotes = conan_api.get_active_remotes(args.remote)
-
     results = []
+
     for remote in remotes:
-        results.append(conan_api.search_remote_recipes(args.query, remote))
+        error = None
```
memsharded, line 27: This implementation is not fully clear to me. If this command is going to be an alias of ``conan list recipes --all``, then maybe it should be implemented as such? I would expect to see that equivalence directly in code.

conans/client/api/conan_api.py
```diff
@@ -92,26 +92,30 @@ def load_remotes(self, remote_name=None, update=False, check_updates=False):
...
-            except BaseException:
-                pass
-            raise
+        except (NotFoundException, PackageNotFoundException) as e:
+            # This exception must be caught manually due to a server inconsistency:
+            # Artifactory API returns an empty result if the recipe doesn't exist, but
+            # Conan Server returns a 404. This probably should be fixed server side,
+            # but in the meantime we must handle it here
+            raise ConanException(f"NotFoundError: {str(e)}")
+        except ConanConnectionError as e:
```
memsharded: This re-raise as ConanExceptions of all exceptions reads a bit weird, specially as it is not adding more context data, just the name of the exception, and that could be encoded directly in the exception?

conans/test/integration/command_v2/list_recipe_revisions_test.py
```diff
@@ -27,15 +31,15 @@ def _upload_recipe(self, remote, reference):
...
-    def test_fail_if_reference_is_not_correct(self):
-        self.client.run("list recipe-revisions whatever", assert_error=True)
-        assert "ERROR: Specify the 'name' and the 'version'" in self.client.out
-
-        self.client.run("list recipe-revisions whatever/", assert_error=True)
-        assert "ERROR: Specify the 'name' and the 'version'" in self.client.out
-
-        self.client.run("list recipe-revisions whatever/1", assert_error=True)
-        assert "ERROR: Value provided for package version, '1' (type Version), is too short" in self.client.out
+    @pytest.mark.parametrize("ref", [
```
memsharded, line 34: This test is repeated several times, maybe extract to a ``list_ux_test.py`` file?

conans/cli/commands/list.py
```diff
@@ -3,31 +3,40 @@
...
-
 
 def list_recipes_cli_formatter(results):
+    no_results_msg = "There are no matching recipe references"
     for result in results:
-        _print_common_headers(result, "recipes")
+        _print_common_headers(result)
+        if result.get("error"):
+            error = result['error']
+            if error.startswith("NotFoundError:"):
```
memsharded: This error message pattern seems fragile, seems connected to the below comments about exceptions.

memsharded (commented):
Looking good, lets get this reviewed by others too.

conans/client/api/conan_api.py
tapia: I agree. These except/raise should be deleted. If the decorated function raises an exception, it should just bubble up.

Note: the `except (NotFoundException, PackageNotFoundException) as e:` should NOT be deleted, though, because it's handling a special user case (as described in the comment).

conans/client/api/conan_api.py
```diff
@@ -236,31 +219,15 @@ def search_remote_recipes(self, query, remote):
...
-
-    def _get_revisions(self, ref, remote=None):
-        if isinstance(ref, PackageReference):
-            method_name = 'get_package_revisions'
-        elif isinstance(ref, ConanFileReference):
-            method_name = 'get_recipe_revisions'
-        else:
-            raise ConanException(f"Unknown reference type: {ref}")
-
+    def _get_revisions(self, ref, getter_name, remote=None):
```
tapia, line 213: Please add a brief comment explaining what the `getter_name` parameter does, for future reference

conans/cli/commands/search.py
tapia: They are aliases in the sense that they're both just calling `conan_api.search_remote_recipes()`, but they're not literal aliases, as they accept different arguments.

franramirez688: Completely agree but I think we should open another PR to change it.

conans/test/integration/command_v2/list_recipe_revisions_test.py
franramirez688: This is not the only one. I was thinking of having only one module parameterized with all the list-subcommands.

franramirez688: I tried it but it looked quite convoluted because of all the different parameterizations to define all the expected messages, references, queries, etc.

memsharded (approved):

tapia (approved):
Looks good to me. Let's merge, and see if some refactoring is necessary when we develop the next commands

PR #9389, opened 2021-08-05
Feature add an option to only copy files if they are different

boussaffawalid:
Changelog: added new param to self.copy function to only copy files if they are different. Similar to the cmake function `copy_if_different`
Mainly for performance purpose to speedup imports when reusing the same build directory.

Docs: https://github.com/conan-io/docs/pull/2186

### DIFF
--- a/conans/client/file_copier.py
+++ b/conans/client/file_copier.py
@@ -1,6 +1,7 @@
 import fnmatch
 import os
 import shutil
+import filecmp
 from collections import defaultdict
 
 from conans.errors import ConanException
@@ -51,7 +52,7 @@ class FileCopier(object):
         return report_copied_files(self._copied, output)
 
     def __call__(self, pattern, dst="", src="", keep_path=True, links=False, symlinks=None,
-                 excludes=None, ignore_case=True):
+                 excludes=None, ignore_case=True, if_different=True):
         """
         param pattern: an fnmatch file pattern of the files that should be copied. Eg. *.dll
         param dst: the destination local folder, wrt to current conanfile dir, to which
@@ -65,6 +66,7 @@ class FileCopier(object):
         param links: True to activate symlink copying
         param excludes: Single pattern or a tuple of patterns to be excluded from the copy
         param ignore_case: will do a case-insensitive pattern matching when True
+        param if_different: only copy file if destination is different from the source
         return: list of copied files
         """
         # TODO: Remove the old "links" arg for Conan 2.0
@@ -75,7 +77,7 @@ class FileCopier(object):
             # Avoid repeatedly copying absolute paths
             return self._copy(os.curdir, pattern, src, dst, links,
                               ignore_case, excludes, keep_path,
-                              excluded_folders=[self._dst_folder])
+                              excluded_folders=[self._dst_folder], if_different=if_different)
 
         files = []
         for src_folder in self._src_folders:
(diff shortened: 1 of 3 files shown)


conans/client/file_copier.py
```diff
@@ -51,7 +52,7 @@ def report(self, output):
         return report_copied_files(self._copied, output)
 
     def __call__(self, pattern, dst="", src="", keep_path=True, links=False, symlinks=None,
-                 excludes=None, ignore_case=True):
+                 excludes=None, ignore_case=True, if_different=True):
```
memsharded: It is not clear in which scenarios it will be copying a file and the target already exists. The idea is that the ``self.copy()`` copies to an empty package folder, so this shouldn't be an issue.

Why is this being added? Performance?

In any case, it is extremely recommended to not change the current default, as the likelihood of breaking is much much higher. In this case the default should be ``if_different=False``

conans/client/file_copier.py
```diff
@@ -237,6 +239,7 @@ def _copy_files(files, src, dst, keep_path, symlinks):
                     pass
                 os.symlink(linkto, abs_dst_name)  # @UndefinedVariable
             else:
-                shutil.copy2(abs_src_name, abs_dst_name)
+                if not if_different or not os.path.exists(abs_dst_name) or not filecmp.cmp(abs_src_name, abs_dst_name):
```
memsharded, line 242: If the reason is performance, some data/benchmark about the savings would be great.

memsharded (commented):
The feature is not completely clear.

In the case this will progress, it would need some tests, to be added in the PR before merging.

conans/client/file_copier.py
boussaffawalid: I made this draft PR to start a discussion, that is why there are no tests so far.
My understanding is that this function is being used in the conan imports method as well.

Use case: 
When doing local development, it is very common to reuse the same build directory. Adding this new param `if_different` will optimize copying files to local build directory, specially when:
* large files needs to be copied to the build directory
* Some files in the build directory are already used by another process (currently the import fail, even there is no need to re import this file as it already exist)

memsharded: Understood, thanks, yes, it makes sense.

Some hints about the future though:

- We will make explicit the ``self.copy()``, so maybe this could be added to the imports case only
- In Conan 2.0 we will probably reduce or integrate the ``imports`` case with the ``deploy`` case. The general idea is that copying files over and over is indeed inefficient, and with virtualenvs it is not needed in most cases, specially development cases. It will only be needed for final extraction of files for final deployment, but not for regular development.

boussaffawalid:
I've use this benchmarking class 
```
class DemoConan(ConanFile):
    name = 'demo'
    version = '1.0'
    license = ''
    settings = 'os', 'compiler', 'build_type', 'arch'
    #generators = 'cmake_find_package_multi'
    keep_imports = True
    no_copy_source = True

    def requirements(self):
        self.requires.add("cmake/3.20.2")

    def imports(self):
        start = time.time()
        for i in range(100):
            self.copy('*', root_package='cmake', src='bin', dst='bin', if_different=False)
        end = time.time()
        runtime = end - start
        print("100 imports took {} seconds to complete its execution for if_different=False".format(runtime))

        start = time.time()
        for i in range(100):
            self.copy('*', root_package='cmake', src='bin', dst='bin1', if_different=True)
        end = time.time()
        runtime = end - start
        print("100 imports took {} seconds to complete its execution for if_different=True".format(runtime))

    def build(self):
        pass

    def package(self):
        pass
```
and I got these results
```
100 imports took 10.610847234725952 seconds to complete its execution for if_different=False
100 imports took 0.14913082122802734 seconds to complete its execution for if_different=True
```

boussaffawalid:
@memsharded could you re-review this PR

memsharded:
I am afraid that this PR being red was buried in our notifications and never got reviewed properly.
Now the ``FileCopier`` has been deprecated and removed in 2.0, in favor of a ``copy()`` method, so I am closing this as outdated. If this functionality is desired there, it would be necessary to report against 2.0. Thanks for your contribution!

PR #9393, opened 2021-08-06
Subfolder for the configure script

### LINKED ISSUE #9375: [feature+bug] conan.tools.gnu.autotools.Autotools can only configure in root source folder
Some times, configure scripts can be found in subdirectories.
Right now, `ConanFile.source_folder` is hard coded at
https://github.com/conan-io/conan/blob/916cb10b99ed92d319cd0719462708ee0501ecd4/conan/tools/gnu/autotools.py#L25

It would be VERY useful to be able to change this.

  * Operating System+version: all
  * Compiler+version: all
  * Conan version: 1.39.0
  * Python version: 3.9.6

lasote:
Changelog: Fix: The new `Autotools` build helper accepts a `build_script_folder ` argument in the `configure()` method to specify are subfolder where the configure script is.
Docs: https://github.com/conan-io/docs/pull/2208

Closes #9375

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -73,14 +73,14 @@ class CMake(object):
 
         self._cmake_program = "cmake"  # Path to CMake should be handled by environment
 
-    def configure(self, source_folder=None):
+    def configure(self, build_script_folder=None):
         # TODO: environment?
         if not self._conanfile.should_configure:
             return
 
         source = self._conanfile.source_folder
-        if source_folder:
-            source = os.path.join(self._conanfile.source_folder, source_folder)
+        if build_script_folder:
+            source = os.path.join(self._conanfile.source_folder, build_script_folder)
 
         build_folder = self._conanfile.build_folder
         generator_folder = self._conanfile.generators_folder
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -78,9 +78,9 @@ class CMake(object):
         if not self._conanfile.should_configure:
             return
 
-        source = self._conanfile.source_folder
+        cmakelist_folder = self._conanfile.source_folder
         if build_script_folder:
-            source = os.path.join(self._conanfile.source_folder, build_script_folder)
+            cmakelist_folder = os.path.join(self._conanfile.source_folder, build_script_folder)
 
         build_folder = self._conanfile.build_folder
         generator_folder = self._conanfile.generators_folder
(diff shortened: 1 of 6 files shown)


conan/tools/gnu/autotools.py
```diff
@@ -13,7 +14,7 @@ def __init__(self, conanfile):
         self._configure_args = toolchain_file_content.get("configure_args")
         self._make_args = toolchain_file_content.get("make_args")
 
-    def configure(self):
+    def configure(self, source_folder=None):
```
memsharded: The current ``tools.cmake.CMake`` implementation is probably a legacy one, with the ``source_folder``.
I understand the use case, maybe it is just the name that is confusing. The ``source_folder`` is already defined by the layout. Maybe we should rename it to something like ``configure(build_script_folder=xxxx)``? (Just an idea, could be ``configure_folder``, ``cmakelists_folder``, or any other thing).

memsharded (approved):
Besides the naming thing, this looks necessary, so looks good to me.

conan/tools/gnu/autotools.py
lasote: totally agree

memsharded (approved):

conan/tools/cmake/cmake.py
```diff
@@ -73,14 +73,14 @@ def __init__(self, conanfile, parallel=True):
...
+    def configure(self, build_script_folder=None):
         # TODO: environment?
         if not self._conanfile.should_configure:
             return
 
         source = self._conanfile.source_folder
-        if source_folder:
-            source = os.path.join(self._conanfile.source_folder, source_folder)
+        if build_script_folder:
+            source = os.path.join(self._conanfile.source_folder, build_script_folder)
```
memsharded: Probably I'd refactor this to name it ``cmakelist_folder``, to clearly differentiate it might not be the root src folder, and just in case somewhere else in the code used the ``source`` folder (but it is not an issue here, as it is not used), but it is ok, just an internal detail.

conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py
```diff
@@ -74,7 +74,7 @@ class AppleframeworkConan(ConanFile):
 
                 def build(self):
                     cmake = CMake(self)
-                    cmake.configure(source_folder="src")
+                    cmake.configure(build_script_folder="src")
```
memsharded: this test is confusing, abusing the ``build_script_folder``, instead of defining the correct ``layout`` with the correct ``source``, which is the case. I think that it is ok, but it would be worth adding a comment to say it is known, and this is not the intended usage.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_build_modules.py
```diff
@@ -175,7 +175,7 @@ class Recipe(ConanFile):
 
             def build(self):
                 cmake = CMake(self)
-                cmake.configure(source_folder="src")
+                cmake.configure(build_script_folder="src")
```
memsharded: Same as above.

memsharded (approved):

memsharded (approved):

PR #9405, opened 2021-08-10
Bump urllib3 version to 1.26.6

tapia:
Changelog: Fix: Bump urllib3 version to 1.26.6
Docs: omit

urllib3 version 1.25 has severe security issues, so we must update it


memsharded (approved):
This will probably break some user in some not mainstream python setup, but security updates are necessary, lets merge and see what happens.

PR #9423, opened 2021-08-13
operator "in" in self.dependencies

lasote:
Changelog: omit
Docs: omit

Operator `in` for `self.dependencies`


memsharded (approved):
I thought the idiom:
```python
dep = self.dependencies.get("name") # with or without args
if dep is not None:
      # use dep
```
was good enough, and the proposal:
```python
if "name" in self.dependencies:  # always without args
      dep = self.dependencies.get("name"): # do not specify here args, or it could even fail
      # use dep
```

doesn't add much, besides a minor potential inconsistency between the check and the actual get.

But if you think this helps, I am fine with it, can be merged (I have fixed py27 broken test)

PR #9443, opened 2021-08-18
support custom name in virtualbuildenv and virtualbuildenv

boussaffawalid:
Changelog: Feature : support using a custom name in virtualbuildenv and virtualbuildenv.
This is useful for example when using multi config ide like visual studio. In this this case: we want to have two scripts:
* conanrunenv_release.bat
* conanrunenv_debug.bat 

Currently `conanrunenv.bat` for release overrides `conanrunenv.bat` for debug.

Docs: https://github.com/conan-io/docs/pull/2199

### DIFF
--- a/conan/tools/env/virtualbuildenv.py
+++ b/conan/tools/env/virtualbuildenv.py
@@ -40,7 +40,7 @@ class VirtualBuildEnv:
 
         return build_env
 
-    def generate(self, auto_activate=True):
+    def generate(self, auto_activate=True, name="conanbuildenv"):
         build_env = self.environment()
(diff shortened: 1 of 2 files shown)


memsharded (changes requested):
I think this seems a valid feature request, and a possible necessary use case for the consumer/developer flow, that we have not approached yet with these new generators.

The custom name doesn't seem enough, as it would require all recipes to implement such naming, and it seems something more automatic would be needed. Probably adding the desired user stories as tests (these would belong to "integration" tests) would throw some extra light on this.

PR #9455, opened 2021-08-20
CMakeDeps/CMakeToolchain: Several improvements for open issues

lasote:
Changelog: Feature: **CMakeToolchain** new member `find_builddirs` defaulted to `True` to add the `cpp_info.builddirs` from the requirements to the `CMAKE_PREFIX_PATH/CMAKE_MODULE_PATH`. That would allow finding the config files packaged and to be able to `include()` them from the consumer `CMakeLists.txt`.
Changelog: Bugfix: **CMakeToolchain**. Fixed a bugfix whereby a variable declared at the `.variables` containing a boolean ended at CMake with a quoted `"True"` or `"False"` values, instead of `ON` / `OFF`
Changelog: Feature: **CMakeDeps**. Added a new property `cmake_find_mode` with possible values to `config`(default), `module`, `both` or `none` to control the files to be generated from a package itself. The `none` replaces the current `skip_deps_file` property.
Changelog: Feature: **CMakeDeps**: Added two new properties `cmake_module_file_name` and `cmake_module_target_name`, analog to `cmake_file_name` and `cmake_target_name`, but to configure the name of `FindXXX.cmake` file and the target declared inside. 
    
Docs: https://github.com/conan-io/docs/pull/2209

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -1,4 +1,5 @@
 import os
+from fnmatch import fnmatch
 
 from conan.tools._check_build_profile import check_using_build_profile
 from conan.tools.cmake.cmakedeps.templates.config import ConfigTemplate
@@ -26,6 +27,9 @@ class CMakeDeps(object):
         # If specified, the files/targets/variables for the build context will be renamed appending
         # a suffix. It is necessary in case of same require and build_require and will cause an error
         self.build_context_suffix = {}
+
+        # Patterns to be excluded
+        self.excluded_requirement_patterns = []
 
         check_using_build_profile(self._conanfile)
 
@@ -75,26 +79,39 @@ class CMakeDeps(object):
             if dep.is_build_context and dep.ref.name not in self.build_context_activated:
                 continue
 
+            # Skip from the consumer
+            if any(fnmatch(str(dep.ref), pattern) for pattern in self.excluded_requirement_patterns):
+                self._conanfile.output.info("CMakeDeps: Skipped '{}'".format(dep.ref))
+                continue
+
+            # Skip from the requirement
             if dep.new_cpp_info.get_property("skip_deps_file", "CMakeDeps"):
                 # Skip the generation of config files for this node, it will be located externally
                 continue
 
-            config_version = ConfigVersionTemplate(self, require, dep)
-            ret[config_version.filename] = config_version.render()
+            generate_find_module = dep.new_cpp_info.get_property("cmake_module_file_name",
+                                                                 "CMakeDeps") is not None
 
-            data_target = ConfigDataTemplate(self, require, dep)
-            ret[data_target.filename] = data_target.render()
+            for find_modules_mode in ([False, True] if generate_find_module else [False]):
+                if not find_modules_mode:
+                    config_version = ConfigVersionTemplate(self, require, dep)
+                    ret[config_version.filename] = config_version.render()
 
-            target_configuration = TargetConfigurationTemplate(self, require, dep)
-            ret[target_configuration.filename] = target_configuration.render()
+                data_target = ConfigDataTemplate(self, require, dep, find_modules_mode)
+                ret[data_target.filename] = data_target.render()
 
-            targets = TargetsTemplate(self, require, dep)
-            ret[targets.filename] = targets.render()
+                target_configuration = TargetConfigurationTemplate(self, require, dep,
+                                                                   find_modules_mode)
+                ret[target_configuration.filename] = target_configuration.render()
 
-            config = ConfigTemplate(self, require, dep)
-            # Check if the XXConfig.cmake exists to keep the first generated configuration
-            # to only include the build_modules from the first conan install. The rest of the
-            # file is common for the different configurations.
-            if not os.path.exists(config.filename):
-                ret[config.filename] = config.render()
+                targets = TargetsTemplate(self, require, dep, find_modules_mode)
+                ret[targets.filename] = targets.render()
+
+                config = ConfigTemplate(self, require, dep, find_modules_mode)
+                # Check if the XXConfig.cmake exists to keep the first generated configuration
+                # to only include the build_modules from the first conan install. The rest of the
+                # file is common for the different configurations.
+                if not os.path.exists(config.filename):
+                    ret[config.filename] = config.render()
+
         return ret
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -92,22 +92,22 @@ class CMakeDeps(object):
             generate_find_module = dep.new_cpp_info.get_property("cmake_module_file_name",
                                                                  "CMakeDeps") is not None
 
-            for find_modules_mode in ([False, True] if generate_find_module else [False]):
-                if not find_modules_mode:
+            for find_module_mode in ([False, True] if generate_find_module else [False]):
+                if not find_module_mode:
                     config_version = ConfigVersionTemplate(self, require, dep)
                     ret[config_version.filename] = config_version.render()
 
-                data_target = ConfigDataTemplate(self, require, dep, find_modules_mode)
+                data_target = ConfigDataTemplate(self, require, dep, find_module_mode)
                 ret[data_target.filename] = data_target.render()
 
                 target_configuration = TargetConfigurationTemplate(self, require, dep,
-                                                                   find_modules_mode)
+                                                                   find_module_mode)
                 ret[target_configuration.filename] = target_configuration.render()
 
-                targets = TargetsTemplate(self, require, dep, find_modules_mode)
+                targets = TargetsTemplate(self, require, dep, find_module_mode)
                 ret[targets.filename] = targets.render()
 
-                config = ConfigTemplate(self, require, dep, find_modules_mode)
+                config = ConfigTemplate(self, require, dep, find_module_mode)
                 # Check if the XXConfig.cmake exists to keep the first generated configuration
                 # to only include the build_modules from the first conan install. The rest of the
                 # file is common for the different configurations.
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -12,6 +12,11 @@ from conans.errors import ConanException
 from conans.util.files import save
 
 
+FIND_MODE_MODULE = "module"
+FIND_MODE_CONFIG = "config"
+FIND_MODE_NONE = "none"
+FIND_MODE_BOTH = "both"
+
 class CMakeDeps(object):
 
     def __init__(self, conanfile):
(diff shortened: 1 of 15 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -75,26 +79,39 @@ def content(self):
...
             if dep.new_cpp_info.get_property("skip_deps_file", "CMakeDeps"):
                 # Skip the generation of config files for this node, it will be located externally
                 continue
 
-            config_version = ConfigVersionTemplate(self, require, dep)
-            ret[config_version.filename] = config_version.render()
+            generate_find_module = dep.new_cpp_info.get_property("cmake_module_file_name",
+                                                                 "CMakeDeps") is not None
+
+            for find_module_mode in ([False, True] if generate_find_module else [False]):
```
memsharded: This iteration reads a bit weird, maybe put the contents inside a local function and call it twice.

conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -14,18 +14,24 @@ class ConfigTemplate(CMakeDepsFileTemplate):
 
     @property
     def filename(self):
-        if self.file_name == self.file_name.lower():
-            return "{}-config.cmake".format(self.file_name)
+        if self.find_module_mode:
```
memsharded, line 17: I thought that better a new class, ``FindTemplate`` instead of ``ConfigTemplate``, but then I saw the large amount of repeated code, and agree this is better implementation.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -14,8 +13,9 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
 
     @property
     def filename(self):
-        return "{}Target-{}.cmake".format(self.file_name,
-                                          self.cmakedeps.configuration.lower())
+        name = "" if not self.find_module_mode else "module-"
```
memsharded, line 16: I was not aware about this "module" prefix, I thought the .cmake scripts had to be called ``find_zlib.cmake`` or something similar.

conan/tools/cmake/cmakedeps/cmakedeps.py
memsharded: Are you sure about generating both the Config and the Find? What would be the use case? Shouldn't be a selection? Maybe for each pattern/dependency to be able to select: None (not generate anything), Find (find_package) or Config (default)?

memsharded (commented):
I'd say that overall, it looks good, lets talk about the comments.

Breaking them into smaller pieces would be nice, yes, some things like the ``cmake_value`` template function is fine and can be merged already, but not necessary, the PR is not that long, so as you prefer.

jgsogo:
So far I think this kind of flexibility will be needed if we want to support all the scenarios out there and adapt to the existing [nightmare] ecosystem. Probably we need a decision about the use-cases that we won't cover and how to workaround them: build-modules, patches to consumer sources, _duplicated_ files for Meson,... that can help us to understand the full picture.


> * **CMakeToolchain**: Added the build modules paths from the requirements to the `CMAKE_PREFIX_PATH/CMAKE_MODULE_PATH`. That would allow finding the config files packaged. THAT COULD BE OPT-IN MAYBE!

There can be different files that one consumer will always think it is the best one:
 1. Conan generated files
 2. Official files provided by the dependency (packaged in the Conan package)
 3. Files provided by the consumer itself: some libraries provide their own scripts to find the dependencies and populate custom variables
 4. Official CMake files (related: https://github.com/conan-io/conan-center-index/issues/2311)

Depending on how intrusive Conan wants to be, we will need more or fewer config parameters in the recipe to fit each scenario.

> * **CMakeDeps**: Exclude generate the configs for a given dependency FROM THE CONSUMER, actually only available from the recipe of the dependency as a property. That would allow skipping the Conan-generated config file to use another one, like the packaged or whatever.

I don't like to have too many config parameters... but sometimes the same library mixes several of the previous approaches. You need some files provided by the library and for some others Conan generated ones are better. In this scenario you can [remove selected ones in the recipe](https://github.com/conan-io/conan-center-index/blob/16864a28736a442e5a6c17870b6884bfd676876d/recipes/openscenegraph/all/conanfile.py#L154):

```python
        for package in ("Fontconfig", "Freetype", "GDAL", "GIFLIB", "GTA", "Jasper", "OpenEXR"):
            # Prefer conan's find package scripts over osg's
            os.unlink(os.path.join(self._source_subfolder, "CMakeModules", "Find{}.cmake".format(package)))
```

> * **CMakeDeps**: Added two new properties `cmake_module_file_name` and `cmake_module_target_name` (analog to `cmake_file_name` and `cmake_target_name`, but these new ones will activate the generation of `FindXXX.cmake` cmake scripts in order to use modules. This issue respond to this situation like this:
>   
>   * CMake installation has FindZLIB.cmake but the author (irrespective if we package or not the config files) have a `zlib-config.cmake`.
>   * So Conan recipe needs to decide the name of the target...to make the consumers happy. `ZLIB`or `zlib`?
>   * But not only that, a consumer might be doing `find_package(zlib MODULE)` so he won't be able to use config files.
>   * That is happening is Conan center, and causing the recipe authors to use `cmake`, `cmake_find_package` or `cmake_find_package_multi` depending on how the CMakeLists consume the dependency, because `cmake_find_package`  generate modules, `cmake_find_package_multi` generate config...

Life is hard 😢 

---

This weekend I've been fighting with [DART](https://github.com/dartsim/dart), I think it is a library with good build-system files (CMake) but with subtle details that make it a bit complicated: library-provided module files that use pkg_config to find dependencies, problems with multi-config generators, casing and names conflicts,... It would be very challenging to provide something that just is a drop-in replacement.

SSE4:
@SpaceIm @prince-chrismc @madebr WDYT? will it make your life easier/harder/same? any comments are welcomed.

conan/tools/cmake/cmakedeps/cmakedeps.py
lasote: Yes. You have to generate both because that property would be declared in recipes of libraries that have a FindXXX in the CMake installation so there might be consumers using `find_package(XXX MODULE`, of course, there might be consumers calling `find_package(xxx)` expecting to find the author's config file (that will find the Conan's xxx-config.cmake). 
So, it is not a consumer choice but a recipe choice.

lasote: But I agree we might use a single property with different values set to "none" (confusing as str)?  (current skip_deps_file), `config`, `module`, `both`.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
lasote: The `FindZLib.cmake` file is not prefixed with `module`, only the other files that are included from it.

conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -74,29 +75,34 @@ def arch(self):
...
+
+    def get_target_namespace(self, req):
+        if self.find_module_mode:
+            ret = req.new_cpp_info.get_property("cmake_module_target_name", "CMakeDeps")
+            if ret:
+                return ret
+
+        ret = req.new_cpp_info.get_property("cmake_target_name", "CMakeDeps")
+        if not ret:
+            ret = req.cpp_info.get_name("cmake_find_package_multi", default_name=False)
```
prince-chrismc, line 89: No fallback to `"cmake_find_package"` for module?

conan/tools/cmake/cmakedeps/templates/target_configuration.py
prince-chrismc: Is there a specific example where this is being done? look at https://gitlab.kitware.com/cmake/cmake/-/tree/master/Modules I did not find any and I have yet to compress this before

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -11,6 +12,11 @@
 from conans.util.files import save
 
 
+FIND_MODE_MODULE = "module"
+FIND_MODE_CONFIG = "config"
+FIND_MODE_NONE = "none"
+FIND_MODE_BOTH = "both"
```
prince-chrismc, line 18: I am curious why a `"both"` option? 
AFAIK official CMake are modules (And there are a few projects the provide their own) and most project install targets + follow https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html which are your config packages

but I've not come across a project that does both 🤔 with the old generators it was the costume who decided what to install but now it's defined by the recipe... we'll need to be more careful to define it correctly

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
```diff
@@ -0,0 +1,94 @@
...
+
+            def package_info(self):
+
+                self.cpp_info.set_property("cmake_find_mode", "both")
+
+                self.cpp_info.set_property("cmake_file_name", "MyDep")
+                self.cpp_info.set_property("cmake_target_name", "MyDepTarget")
+
+                self.cpp_info.set_property("cmake_module_file_name", "mi_dependencia")
+                self.cpp_info.set_property("cmake_module_target_name", "mi_dependencia_target")
```
prince-chrismc, line 50: What about modules without targets? some of the older only use variables... how can we provide those?

conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -74,29 +75,34 @@ def arch(self):
...
-    return ret or comp_name
+        return get_file_name(self.conanfile, find_module_mode=self.find_module_mode)
+
+    def get_target_namespace(self, req):
+        if self.find_module_mode:
+            ret = req.new_cpp_info.get_property("cmake_module_target_name", "CMakeDeps")
+            if ret:
+                return ret
+
+        ret = req.new_cpp_info.get_property("cmake_target_name", "CMakeDeps")
```
prince-chrismc, line 87: In the language I am familiar with, target names are not the same as the namespace. This is a little confusing for me

prince-chrismc (commented):
The template looks really really complete! 🚀 Great work

> WDYT? will it make your life easier/harder/same? any comments are welcomed.

Gut reaction, the same, but it's hard to say only reviewing the code.... Some of the property names are a little ambiguous for me.


Some questions that might be outside of the diff.

- Could config be the default? that's what the majority of projects install (or they are supposed to)
- Maybe an example of setting the package name, namespace and a target?
- How about namespaces per target?

I also have a few questions about the code

conan/tools/cmake/cmakedeps/cmakedeps.py
lasote: `both` cover the case where the library has a module in cmake, typical `FindXXX.cmake` but also the author's provided config (irrespective if we package it or not). In that case, you don't know how the consumer will expect to consume the package, maybe `find_package(XXX MODULE)` maybe `find_package(xxx CONFIG)` but we want Conan to be as "ready" as possible to consume any conan center package. So `CMakeDeps` will generate both. Note that the `config` namespace and target name should be declared typically following the author's one.

conan/tools/cmake/cmakedeps/templates/__init__.py
lasote: I've been greping `conan-center-index` and the name is always set for both of them:

```
        self.cpp_info.names["cmake_find_package"] = "Crc32c"
        self.cpp_info.names["cmake_find_package_multi"] = "Crc32c"
```

So the fallback looks enough. We introduce this fallback mostly for Conan center.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
lasote: The issue with these modules without targets is, Which variables should we declare? probably nothing to be automated. This might be another reason to enable a property to introduce custom cmake code that would be declared when consuming the module/config but we didn't agree yet about that.

conan/tools/cmake/cmakedeps/templates/__init__.py
lasote: You also have the possibility to declare components:
https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html#properties

lasote: Thanks for the review by the way! :)

memsharded: This is followed up by this issue https://github.com/conan-io/conan/issues/9511, right?

memsharded (approved):
Good to me, can be merged and continue working and improving over it (like https://github.com/conan-io/conan/issues/9511)

conan/tools/cmake/cmakedeps/templates/__init__.py
lasote: Yes, exactly

conan/tools/cmake/cmakedeps/cmakedeps.py
prince-chrismc: That makes sense, I understand the motivation but I am not aware of any examples.

There are a few project out there that provide their own module files even though the upstream project provides config files. I wonder if this bonus feature will help there 🤔

PR #9468, opened 2021-08-23
conf to default the usage of host and build profile

### LINKED ISSUE #9336: [feature] Create new [conf] to activate using always the build profile (defaulting to "default" profile)
It might be useful to switch to the new paradigm, without needing to modify every single CLI invocation.

memsharded:
Changelog: Feature: Provide new ``[conf]`` ``core:default_build_profile`` to enable the usage of the build profile as default, and to allow definition of the host profile default in new ``[conf]`` ``core:default_profile``.
Docs: https://github.com/conan-io/docs/pull/2213

Close https://github.com/conan-io/conan/issues/9336

This was way more difficult than expected, the implementation is ugly and confusing, but it is due to the current state of legacy 1.X, all this needs to be fixed for 2.0

### DIFF
--- a/conans/test/integration/configuration/default_profile_test.py
+++ b/conans/test/integration/configuration/default_profile_test.py
@@ -1,4 +1,5 @@
 import os
+import textwrap
 import unittest
 
 from conans.client import tools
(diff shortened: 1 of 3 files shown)


lasote (approved):

franramirez688 (approved):

conans/test/integration/configuration/default_profile_test.py
```diff
@@ -180,3 +181,19 @@ def build(self):
                           "an existing profile file.", client.out)
 
 
+def test_conf_default_two_profiles():
+    client = TestClient()
+    save(os.path.join(client.cache.profiles_path, "mydefault"), "[settings]\nos=FreeBSD")
+    save(os.path.join(client.cache.profiles_path, "mydefault_build"), "[settings]\nos=Android")
+    global_conf = textwrap.dedent("""
+        core:default_profile=mydefault
```
czoido, line 189: Just one question, why is the name `default_profile` instead of `default_host_profile`?

memsharded: To keep it a bit more aligned with the current naming, documentation, etc.

czoido (approved):

Issue #9295, opened 2021-07-20
[feature] add clang support on windows


AndreyMlashkin:
Relates to https://github.com/conan-io/conan-center-index/issues/6414#issuecomment-883193586

If I try to compile with clang, that is not coming from MSVC, conan still try to use it as a MSVC toolchain

MinGW also does not work for windows.

- [ ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Hi @AndreyMlashkin 

We would need a bit more details, please.

We have MinGW tests in our test suite that pass for every commit, on Windows. 

Please extract from the conan-center thread the relevant details, and try to provide some minimum reproducible example (can we just try it with a ``conan new`` "hello world" recipe? and vision for the user interface of the feature.

Also, these might be 2 different features, don't hesitate to submit 2 separate issues, if they are actually different.

AndreyMlashkin:
Ok, totally make sense, I will change caption from 'add clang and mingw support on windows' to 'add clang support on windows'

A reproducible example you can find here:
https://github.com/AndreyMlashkin/windows-clang-conan/blob/main/Dockerfile

AndreyMlashkin:
Maybe I have just filled a conan profile wrong?
https://github.com/AndreyMlashkin/windows-clang-conan/blob/main/profile

But I have tried different configurations...

memsharded:
Well, this could also be that the ConanCenter ``zlib/1.2.11`` recipe could not be prepared to build for clang in Windows. There are many libraries in ConanCenter which build scripts are simply not prepared, or sometimes the recipe itself might hardcode some assumptions about the build systems. This is why I would suggest to use a hello world example, which can be easily generated with ``conan new hello/0.1 -s``.

The "legacy" build system integrations, which are the stable ones, and generated with ``conan new hello/0.1 -s`` and the used in ConanCenter are now in maintenance mode only, and we are preparing Conan 2.0. Most of the work and new features is done in the new build-system integrations in https://docs.conan.io/en/latest/reference/conanfile/tools.html. 

Maybe a better ``clang`` support is not easy to be introduced in the stable build system integrations without breaking, so if possible, it would be great to target directly the new ones, which can be also tested with ``conan new hello/0.1 -m v2_cmake``

AndreyMlashkin:
I have re-written test code to use a 'hello' test. It still demonstrate the problem. Please, investigate.

[...]

PR #9477, opened 2021-08-25
Feature/clang

memsharded:
Changelog: Feature: Testing support for Windows CMake + Clang (independent LLVM, not VS) + Ninja/MinGW builds, and CMake + Clang (Visual Studio 16 internal LLVM 11 via ClangCL toolset).
Docs: Omit

Address: https://github.com/conan-io/conan/issues/9295

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -515,6 +515,11 @@ class GenericSystemBlock(Block):
                     return "version=14.{}".format(minor)
                 else:
                     return "v14{}".format(minor)
+        elif compiler == "clang":
+            if generator and "Visual" in generator:
+                if "Visual Studio 16" in generator:
+                    return "ClangCL"
+                # TODO: Handle other cases
         return None
 
     def _get_generator_platform(self, generator):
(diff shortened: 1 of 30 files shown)


franramirez688 (approved):

memsharded:
@czoido clang in Windows in CI is not installed yet, but this can be a separate feature.

czoido (approved):

conan/tools/cmake/toolchain.py
```diff
@@ -515,6 +515,12 @@ def _get_toolset(self, generator):
                     return "version=14.{}".format(minor)
                 else:
                     return "v14{}".format(minor)
+        elif compiler == "clang":
+            if generator and "Visual" in generator:
```
arttet, line 519: Why cannot we use other generators such as `cmake` ?

memsharded: This is the CMake generator. You need to select one of the CMake supported ones: ``MinGW Makefiles``, ``Visual Studio XX YYYY``, etc.

Also, ``CMakeToolchain`` will only support the ``CMakeDeps`` generator (based on find_package()), but not the legacy ``cmake``, ``cmake_find_package[_multi]``, becase these later ones will be removed in 2.0.

arttet: Ok, I got it, thank you.

PR #9513, opened 2021-09-01
CMakeDeps with custom namespace

### LINKED ISSUE #9511: [feature] CMakeDeps different namespace and name for the global target
According to the documentation, the property:

```
cmake_target_name: Name of the target to be consumed. 
When set on the root cpp_info, it changes the namespace of the target. 
When set to a component, it changes the name of the target (see the example below).
``` 

1) It is not the best UI, not easy to understand without reading the docs.
2) There is no way to create a "global" target with different `namespace::name`
3) There are packages like `pybind11` containing modules that declare the global target itself, linking with targets from the components, but not with all of them. In these cases, it would be useful to declare a global target `pybind11::global` or similar and let the module declare the `pybind11::pybind11` with the target expected by the consumers.

cc @danimtb @ericLemanissier @SpaceIm @ericriff @prince-chrismc

Related https://github.com/conan-io/conan-center-index/issues/6605

lasote:
Changelog: Feature: Introduced a new cpp_info property `cmake_target_namespace` to declare the target namespace for the `CMakeDeps` generator. This feature allows declaring a global target with a different namespace like `Foo::Bar`.
Docs: https://github.com/conan-io/docs/pull/2209

Close https://github.com/conan-io/conan/issues/9511

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/__init__.py
+++ b/conan/tools/cmake/cmakedeps/templates/__init__.py
@@ -21,6 +21,10 @@ class CMakeDepsFileTemplate(object):
         return get_target_namespace(self.conanfile) + self.suffix
 
     @property
+    def global_target_name(self):
+        return get_global_target_name(self.conanfile) + self.suffix
+
+    @property
     def file_name(self):
         return get_file_name(self.conanfile) + self.suffix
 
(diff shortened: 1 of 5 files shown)


memsharded (approved):

danimtb (approved):

danimtb:
I'd mention `global_target_name` in the changelog as well

PR #9516, opened 2021-09-01
update conan new modern templates

memsharded:
Changelog: Feature: Update ``conan new`` modern templates ``--template=cmake_lib`` and ``--template=cmake_exe``.
Docs: https://github.com/conan-io/docs/pull/2211

### DIFF
--- a/conans/assets/templates/new_v2_cmake.py
+++ b/conans/assets/templates/new_v2_cmake.py
@@ -39,12 +39,8 @@ class {package_name}Conan(ConanFile):
         cmake.build()
 
     def package(self):
-        self.copy("*.h", dst="include")
-        self.copy("*.lib", dst="lib", keep_path=False)
-        self.copy("*.dll", dst="bin", keep_path=False)
-        self.copy("*.dylib*", dst="lib", keep_path=False)
-        self.copy("*.so", dst="lib", keep_path=False)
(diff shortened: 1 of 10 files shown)


conans/assets/templates/new_v2_cmake.py
```diff
@@ -140,6 +220,88 @@ def get_files(name, version, package_name="Pkg"):
...
                                                                    version=version,
                                                                    package_name=package_name),
-             "test_package/example.cpp": test_main.format(name=name),
-             "test_package/CMakeLists.txt": test_cmake_v2.format(name=name)}
+             "test_package/src/example.cpp": test_main.format(name=name),
+             "test_package/src/CMakeLists.txt": test_cmake_v2.format(name=name)}
+    return files
+
+
+conanfile_exe = """from conans import ConanFile
```
lasote, line 228: Nonrelated question, are we moving/copying the ConanFile to `conan` namespace at some point?

lasote (approved):
🥇

conans/assets/templates/new_v2_cmake.py
memsharded: Yes, we are.
I didn't want to do it yet, because I am thinking of using the move to clean some potentially breaking things, and the opt-in would be the new import.

Issue #9438, opened 2021-08-17
[feature] Support CMake OBJECT libraries


bemehiser:
Add `cmake_find_package` and `cmake_find_package_multi` support for CMake `OBJECT` libraries.

CMake allows the creation of [`OBJECT` libraries](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) (as contrasting with `STATIC` and `SHARED`). These `OBJECT` libraries consist of a lot of compiled object files which haven't been linked into a library or binary.

```
add_library(example_obj_lib OBJECT
        include/alib.h
        src/alib.c)

target_include_directories(example_obj_lib
        PUBLIC
        include)
```

These libraries can then be linked into other libraries directly without needing to recompile the objects. This is particularly useful when creating SDK libraries, where multiple libraries' objects' symbols can be combined into a single static library, so a single library binary can be passed to other developers with all relevant symbols.

```
add_library(example_sdk STATIC
        $<TARGET_OBJECTS:example_obj_lib>
        include/example.h
        src/example.c)

target_link_libraries(example_sdk
        PUBLIC
        include
       $<TARGET_PROPERTY:example_obj_lib,INTERFACE_INCLUDE_DIRECTORIES>)
```

As a developer of both `example_obj_lib` and `example_sdk`, I would like to be able to package `example_obj_lib` `OBJECT` libraries via Conan, and generate CMake Find scripts with those `OBJECT` libraries in my `example_sdk project`.

- [*] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
I don't think this is a specific CMake object libraries, but it seems this would be a duplicate of this one: https://github.com/conan-io/conan/issues/5159 (and related to https://github.com/conan-io/conan/issues/8703), and what needs to be done is:
- To explicitly model "objects" in packages, besides "libs" and "exes"
- To make build helpers and generators to handle this information for consumers.

Can we close this one as duplicate and track it in https://github.com/conan-io/conan/issues/5159?

memsharded:
``CMakeDeps`` added support for this long time ago, closing.

PR #9520, opened 2021-09-02
Adding cpp_info objects model for linking object files

czoido:
Changelog: Feature: Add objects attribute to the cpp_info so that we can add object files to the linker without having to add those mixed with linker flags.
Docs: https://github.com/conan-io/docs/pull/2243

For the moment it just works with CMakeDeps generator.

It's used like:

```
def package_info(self):
    self.cpp_info.components.objects = ["myobject.o"]
```
```
def package_info(self):
    self.cpp_info.components["say"].objects = ["mycomponent.o"]
```

Related with: 
- https://github.com/conan-io/conan/issues/8703
- https://github.com/conan-io/conan/issues/9438
- https://github.com/conan-io/conan/issues/5159
- https://github.com/conan-io/conan/pull/8740
- https://github.com/conan-io/conan/issues/8811

Related info:
- https://cmake.org/cmake/help/latest/prop_tgt/LINK_OPTIONS.html#prop_tgt:LINK_OPTIONS
- https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_OPTIONS.html
- https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries
- https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES.html
- https://cmake.org/cmake/help/latest/prop_tgt/LINK_LIBRARIES.html
- https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#object-libraries


memsharded (commented):
One functional test that really links and execute an app with a couple of objects, for this feature please

czoido:
> One functional test that really links and execute an app with a couple of objects, for this feature please

I'm on it, I already tested the feature with mimalloc: https://github.com/conan-io/conan-center-index/tree/master/recipes/mimalloc

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -278,6 +278,60 @@ def package_info(self):
...
+    object_cpp = gen_function_cpp(name="myobject")
+    object_h = gen_function_h(name="myobject")
+    lib_cpp = gen_function_cpp(name="hello")
+    lib_h = gen_function_h(name="hello")
+    cmakelists = textwrap.dedent("""
+        cmake_minimum_required(VERSION 3.15)
+        project(HelloLib)
+        file(GLOB HEADERS *.h)
+        add_library(myobject OBJECT myobject.cpp)
+        add_library(hello hello.cpp hello.h)
```
memsharded: Why the "hello" library? Is not possible to have a package with just the objects? (I would say it is also an expected use case). I guess everything should work, but with CMake, who knows...

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -278,6 +278,62 @@ def package_info(self):
     assert "MYVAR=>Like a Rolling Stone" in c.out
 
 
+@pytest.mark.tool_cmake
+def test_cpp_info_link_objects():
+    client = TestClient()
+
+    obj_ext = "obj" if platform.system() == "Windows" else "o"
+    cpp_info = {"objects": ["lib/myobject.{}".format(obj_ext)], "libs": ["hello"]}
```
lasote: I didn't know you can do this in a `GenConanfile` 😶‍🌫️.

lasote (commented):
Pretty cool!

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -66,6 +66,7 @@ def template(self):
               set({{ pkg_name }}_DEFINITIONS{{ config_suffix }} {{ global_cpp.defines }})
               set({{ pkg_name }}_SHARED_LINK_FLAGS{{ config_suffix }} {{ global_cpp.sharedlinkflags_list }})
               set({{ pkg_name }}_EXE_LINK_FLAGS{{ config_suffix }} {{ global_cpp.exelinkflags_list }})
+              set({{ pkg_name }}_OBJECTS{{ config_suffix }} {{ global_cpp.objects_list }})
```
memsharded, line 69: Only global? The cmake target_configuration seems to try to apply per-component ones, but seems per-component not defined here?

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -85,3 +85,24 @@ def package_info(self):
...
+
+
+def test_cpp_info_component_objects():
+    client = TestClient()
+    conan_hello = textwrap.dedent("""
+        from conans import ConanFile
+        class Pkg(ConanFile):
+            settings = "os", "arch", "build_type"
+            def package_info(self):
+                self.cpp_info.components["say"].objects = ["/path/to/mycomponent.o"]
```
memsharded: It is not clear what the object should be, always an absolute path? How does it work with relative paths?

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -278,6 +278,66 @@ def package_info(self):
     assert "MYVAR=>Like a Rolling Stone" in c.out
 
 
+@pytest.mark.tool_cmake
+def test_cpp_info_link_objects():
+    client = TestClient()
+    obj_ext = "obj" if platform.system() == "Windows" else "o"
+    cpp_info = {"objects": [os.path.join("lib", "myobject.{}".format(obj_ext))], "libs": ["hello"]}
```
memsharded: This path is relative, but seems to work (see note below)

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
czoido: Sorry, this is a mistake, it should be a relative path

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps.py
czoido: Yes, it should always be a relative path, I have changed the others that were mistaken

memsharded (approved):

lasote (approved):

Issue #9026, opened 2021-05-31
[feature] Create minor update versions for the Intel compiler in the default settings


memsharded:
The intel compiler has some behavior similar to gcc, there are minor versions, that are relevant to compatibility.

```
version: ["11", "12", "13", "14", "15", "16", "17", "18", "19", "19.1"]
```

It might be necessary or useful to represent the update too:

```
version: ["11", "12", "13", "14", "15", "16", "17", "18", "19", "19.0.3", "19.0.5"]
```

The update can be 19.0.3.203, so maybe use a ``update = [None, "ANY"]`` subsetting

@ansutremel

memsharded:
It seems that https://github.com/conan-io/conan/issues/9304 IntelOneApi will require some changes to the Intel compiler definition. Lets handle these 2 issues together.

memsharded:
https://github.com/conan-io/conan/pull/9522 implemented this, to be released in 1.41.

PR #9522, opened 2021-09-02
Adding Intel oneAPI support

### LINKED ISSUE #9304: [feature] Support for intel OneApi classic compilers
Recently, intel has made their c/c++/Fortran compilers publicly available by means of their [OneApi toolkits](https://software.intel.com/content/www/us/en/develop/tools/oneapi/all-toolkits.html#gs.70oc5a) that include their _classic_ compilers (based on VS for windows and gcc for linux) as well as their new compilers based on LLVM technologies. However, this _classic_ compilers break the compatibility with conan, as their installation path is different and their way of setting the variables is also different.

The changes needed to support this new versions of the compiler would consist on modifying `conans/client/tools/intel.py`, so that function `intel_installation_path` points to the installation path `/opt/intel/oneapi/compiler/latest/linux/` on the OneApi compilers and function `intel_compilervars_command` executes the OneApi compiler variables script in `/opt/intel/oneapi/setvars.sh`. It appears also that the test present in `conan/conans/test/unittests/client/tools/intel/compilervars_test.py` should also be modified.

franramirez688:
Changelog: Feature: New "compiler" `intel-cc` with different modes (icx, dpcpp, classic) for Intel oneAPI.
Changelog: Feature: Add Intel oneAPI support for `CMakeToolChain`, `MSBuildToolchain` and `VCVars`.
Docs: https://github.com/conan-io/docs/pull/2233

Closes: https://github.com/conan-io/conan/issues/9304 https://github.com/conan-io/conan/issues/9026

* Added new "compiler" `intel-cc` with different modes (icx, dpcpp, classic):
       * icx == Intel oneAPI C++ Compiler (icx/icpx)
       * dpcpp == Intel oneAPI DPC++ Compiler (dpcpp)
       * classic == Intel C++ Compiler Classic (icc/icpc for Linux/macOS and icl for Windows)
* Added as part of the new `conan.tools` flow:
       * Using: `CMakeToolChain`, `MSBuildToolchain` or `VCVars`
      

conan/tools/intel/oneapi.py
```diff
@@ -0,0 +1,91 @@
...
+    return command
+
+
+def get_intel_setvars_dict(conanfile, arch=None, force=False, only_diff=True):
+    cmd = get_intel_setvars_command(conanfile, arch, force)
+    return env_diff(cmd, only_diff)
+
+
+@contextmanager
+def intel_setvars(*args, **kwargs):
```
memsharded: We wont be using this way to set environment anymore. Tools need to create a file (recommended for generators in ``generate()`` method), or return an ``Environment`` object that can be applied.

conan/tools/intel/oneapi.py
```diff
@@ -0,0 +1,91 @@
...
+
+def get_intel_installation_path(out):
+    system = platform.system()
+
+    if system == "Darwin":
+        # Let's show WARNING for macOS
+        out.warn("The macOS* operating system is not supported by the "
+                 "Intel oneAPI DPC++/C++ Compiler")
+
+    installation_path = get_env("CONAN_INTEL_INSTALLATION_PATH")
```
memsharded: Do not use env-vars for configuration, only new ``[conf]`` items.

conan/tools/intel/oneapi.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    return int(compiler_version.split(".")[0]) > 19
+
+
+def get_intel_installation_path(out):
+    system = platform.system()
+
+    if system == "Darwin":
+        # Let's show WARNING for macOS
+        out.warn("The macOS* operating system is not supported by the "
```
memsharded: Why not exception?

conan/tools/microsoft/visual.py
```diff
@@ -49,13 +56,14 @@ def _write_conanvcvars(conanfile, auto_activate=True):
             minor = version_components[1]
             # The equivalent of compiler 19.26 is toolset 14.26
             vcvars_ver = "14.{}".format(minor)
-        cvars = vcvars_command(vs_version, architecture=vcvarsarch, platform_type=None,
-                               winsdk_version=None, vcvars_ver=vcvars_ver)
-    if cvars:
+        intel_vars_setup_command = vcvars_command(vs_version, architecture=vcvarsarch,
```
memsharded: Using the ``intel`` name here is very confusing, this one is not related to intel, right?

conans/client/build/cmake.py
```diff
@@ -241,8 +243,12 @@ def _run(self, command):
         elif is_intel:
             if self.generator in ["Ninja", "Ninja Multi-Config",
                                   "NMake Makefiles", "NMake Makefiles JOM", "Unix Makefiles"]:
-                intel_compilervars_dict = tools.intel_compilervars_dict(self._conanfile, force=True)
-                context = _environment_add(intel_compilervars_dict, post=self._append_vcvars)
+                compiler_version = self._settings.get_safe("compiler.version")
```
memsharded: No need to support legacy integrations.

conans/client/migrations_settings.py
```diff
@@ -2403,7 +2403,9 @@
         libcxx: [libstdc++, libc++]
         cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20]
     intel:
-        version: ["11", "12", "13", "14", "15", "16", "17", "18", "19", "19.1"]
+        version: ["11", "12", "13", "14", "15", "16", "17", "18", "19", "19.1",
+                  "2021.1", "2021.2", "2021.3"]
+            update = [None, ANY]
```
memsharded: spacing

memsharded (changes requested):

conan/tools/intel/oneapi.py
franramirez688: What if you are using the classic Compiler instead of the newest one? https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-setup/using-the-command-line/specifying-the-location-of-compiler-components.html

conans/client/build/cmake.py
franramirez688: Then, if we don't need to support legacy ones, shall I keep only the newest Intel versions? -> `2021.x.x`

franramirez688: I just realized about this is our old-fashioned cmake file, so, perhaps, here we can keep the legacy one and adapt the new `conan.tools.cmake` to the new oneAPI module.

conan/tools/intel/oneapi.py
```diff
@@ -0,0 +1,91 @@
...
+                 "Intel oneAPI DPC++/C++ Compiler")
+
+    installation_path = get_env("CONAN_INTEL_INSTALLATION_PATH")
+    if not installation_path:
+        # Let's try the default dirs
+        if system == ["Linux", "Darwin"]:
+            installation_path = os.path.join(os.sep, "opt", "intel", "oneapi")
+        elif system == "Windows":
+            installation_path = "C:\\Program Files (x86)\\Intel\\oneAPI"
```
SSE4: why hard-coded? shouldn't it be read from the registry?

conans/client/migrations_settings.py
SSE4: why switch to year instead of compiler version?

franramirez688: Apparently, it's the version https://software.intel.com/content/www/us/en/develop/articles/oneapi-toolkit-version-to-compiler-version-mapping.html

SSE4: really? can you double-check what is returned for `__INTEL_COMPILER` and `__INTEL_COMPILER_UPDATE`?
if you have an access to the compiler, ofcourse.
curious, cause it may break our detection.

conans/client/migrations_settings.py
```diff
@@ -2413,6 +2414,12 @@
                 <<: *visual_studio
             apple-clang:
                 <<: *apple_clang
+    intel-cc:
+        version: ["2021.1", "2021.2", "2021.3"]
+            update: [None, ANY]
+        mode: [None, "classic", "dpcpp"]
```
SSE4: what is mode `None` in this case? is it ICX?
we have 3 compiler:
```
        - ICX - Intel oneAPI DPC++/C++ Compiler
        - ICC Classic - Intel C++ Compiler Classic
        - DPCPP - Intel oneAPI DPC++/C++ Compiler
 ```
if so, how about explicit ["icx", "classic", "dpcpp"]?

**Comment by SSE4** [comment]
what about compatibility with GCC/VS (compiler.base)? it was an important feature before to allow mix of existing GCC binaries with Intel.

File: conans/client/migrations_settings.py [hunk 9]
```diff
@@ -2413,6 +2414,12 @@
...
+        libcxx: [None, libstdc++, libstdc++11, libc++, c++_shared, c++_static]
```
SSE4: does Intel support c++_shared/c++_static? these two were Android specific.

conans/client/migrations_settings.py
```diff
@@ -2413,6 +2414,12 @@
                 <<: *visual_studio
             apple-clang:
                 <<: *apple_clang
+    intel-cc:
```
SSE4: in general, I'd love to see an explanation/motivation on these scheme and naming
e.g., why `classic` flavor is not just a new version of the existing `intel` compiler?
why sub-setting fits better than several new compilers?
also, what about `intel-clang` as new compilers seems to be clang based? do they share many things in common?

conan/tools/intel/intel.py
```diff
@@ -0,0 +1,288 @@
...
+        # TODO: hardcoded value but will it depend on the version?
+        return "Intel C++ Compiler 19.2"
+
+
+class _InteloneAPIClang(_InteloneAPIBase):
+    """Intel oneAPI DPC++/C++ Compiler"""
+
+    @property
+    def ms_toolset(self):
+        # By default, we'll assume you are using C/C++ mode else DPC++
```
SSE4: there are 3 different toolsets
```
Intel(R) oneAPI DPC++ Compiler
Intel C++ Compiler 2021
Intel C++ Compiler 19.2
```

conan/tools/intel/intel.py
```diff
@@ -0,0 +1,288 @@
...
+        return command
+
+
+class _InteloneAPIClassic(_InteloneAPIBase):
+    """Intel oneAPI C++ Compiler Classic """
+
+    @property
+    def ms_toolset(self):
+        # TODO: hardcoded value but will it depend on the version?
+        return "Intel C++ Compiler 19.2"
```
SSE4: for sure, it will. needs to be computed from the settings.

conans/client/migrations_settings.py
franramirez688: Let's go step by step and discuss the different points:
* The naming, for sure, we can think of another one that can fit better. This name is easy to read/write so, why not?
* Why different `modes` rather than different compilers? According to [this Intel reference](https://software.intel.com/content/www/us/en/develop/articles/porting-guide-for-icc-users-to-dpcpp-or-icx.html):

```
The following are the guiding principles for ICX:

* ICX and ICC Classic use different compiler drivers. The ICC Classic drivers are icc, icpc, and icl. The ICX drivers are icx and icpx. Use icx to compile and link C programs, and icpx for C++ programs. 
Unlike the icc driver, icx does not use the file extension to determine whether to compile as C or C+. Users must invoke icpx to compile C+ files. In addition to providing a core C++ Compiler, ICX is the base compiler for the Intel® oneAPI Data Parallel C++ Compiler and its new driver, dpcpp.
* For Data Parallel C++ applications: The underlying compiler is ICX and the driver is dpcpp. 
* For Standard C++ applications: Intel recommends using the ICC Classic production compiler and the icc/icpc/icl driver.
```
All the compilers are more or less related between them and they are different for sure, but we can keep it simpler by changing only the mode instead of declaring 3 different ones with the same flags, installation paths, etc.
* They are indeed LLVM-based and, they admit the Clang flags but, in my opinion, all the code related to compilers based on another one is a bit convoluted so I think it's better to avoid it.

I hope it's clearer now. By the way, @SSE4 @memsharded please, share your thoughts here and let me know what you think about this new approach, it would be great 😁

conan/tools/_compilers.py
```diff
@@ -396,3 +403,36 @@ def _cppstd_mcst_lcc(mcst_lcc_version, cppstd):
...
+    vgnu23 = "gnu++2b"
+
+    flag = {"98": v98, "gnu98": vgnu98,
+            "03": v03, "gnu03": vgnu03,
+            "11": v11, "gnu11": vgnu11,
+            "14": v14, "gnu14": vgnu14,
+            "17": v17, "gnu17": vgnu17,
+            "20": v20, "gnu20": vgnu20,
+            "23": v23, "gnu23": vgnu23}.get(cppstd, None)
+    return "-std=%s" % flag if flag else None
```
franramirez688, line 438: For now and as far as I know, there are no limitations to use any of these flags for the new Intel oneAPI C++/DPC++/Classis compilers

SSE4:
I think there are some points still missing:
- what about compiler detection?
- what about compatibility with GCC/Visual Studio (package_id)?

franramirez688:
> I think there are some points still missing:
 > * what about compiler detection?

Yes, this is another step that I have to add for sure.

> * what about compatibility with GCC/Visual Studio (package_id)?

Do we need it? I mean, we're creating this new one to simplify the complexity added by the `compiler.base` implementation, aren't we? I guess we're losing this compatibility but the thing is, are there important reasons to maintain that compatibility if the Intel oneAPI has its compilers regardless of what they are based on (Clang or whatever)?

franramirez688:
More clarifications about this PR. I'm adding this `intel-cc` new compiler to the new `conan.tools` mechanism so I'm not changing anything regarding the legacy part (cmake, compiler flags, etc).

franramirez688:
> I think there are some points still missing:
> 
> * what about compiler detection?

@SSE4 @memsharded
I've been looking into it and, in my humble opinion, I think it's not worth implementing it for now.
Auto-detection is based on checking the output and having that compiler already added into the PATH to execute some kind of command to get the compiler version from there, so, for this case, you'll even need to have run the `setvars.bat|sh`.
I think it's better to let the users define the `CC` and `CXX` instead for every case regardless of the `mode` selected, e.g., if `compiler.mode=icx` you should define into your `profile` something like:

```
...
compiler=intel-cc
compiler.version=2021.3
compiler.mode=icx
...
[env]
CC=icx
CXX=icx
```
By the way, would it be possible to define a mechanism to write a default `CC|CXX` depending on the `mode` instead? what do you think?

conan/tools/intel/inteloneapi.py
```diff
@@ -0,0 +1,178 @@
...
+def get_inteloneapi_installation_path():
+    system = platform.system()
+    # Let's try the default dirs
+    if system == "Windows":
+        if is_win64():
+            base = r"SOFTWARE\WOW6432Node"
+        else:
+            base = r"SOFTWARE"
+        path = r"{base}\Intel\Products\IntelOneAPI".format(base=base)
+        from six.moves import winreg  # @UnresolvedImport
```
memsharded: You might avoid ``six`` usage now, and fail without managing error if Py2, as this will directly fail when merging this branch to ``develop2`` otherwise.

conan/tools/intel/inteloneapi.py
```diff
@@ -0,0 +1,178 @@
...
+
+def is_using_intel_oneapi(compiler_version):
+    """Check if the Intel compiler to be used belongs to Intel oneAPI
+
+    Note: Intel oneAPI Toolkit first version is 2021.1
+    """
+    return int(compiler_version.split(".")[0]) >= 2021
+
+
+def get_inteloneapi_installation_path():
```
memsharded: make it module private

conan/tools/intel/inteloneapi.py
```diff
@@ -0,0 +1,178 @@
...
+import os
+import platform
+import textwrap
+
+from conan.tools.env.environment import create_env_script
+from conans.client.tools.win import is_win64, _system_registry_key
+from conans.errors import ConanException
+
+
+def is_using_intel_oneapi(compiler_version):
```
memsharded: Make it module private

conan/tools/intel/inteloneapi.py
```diff
@@ -0,0 +1,178 @@
...
+                vs2019: Microsoft Visual Studio* 2019
+                vs2017: Microsoft Visual Studio 2017
+
+        :return: `str` setvars.sh|bat command to be run
+        """
+        # Let's check if user wants to use some custom arguments to run the setvars script
+        command_args = self._conanfile.conf["tools.intel:setvars_args"] or ""
+        # The setvars script is going to be loaded/cleared up every conanfile.run() execution
+        # but we will check this env variable just in case
+        if str(os.getenv("SETVARS_COMPLETED", "")) == "1" and "force" not in command_args:
```
memsharded: Maybe this is not the right place, because this command is not executed right now, but can be executed later.
I would probably add this check inside the generated script file, so it is checked right before doing the actual call to activate the intel environment (e.g. someone calling the activate script twice)

conans/client/conf/__init__.py
```diff
@@ -115,6 +116,12 @@
                     <<: *visual_studio
                 apple-clang:
                     <<: *apple_clang
+        intel-cc:
+            version: ["2021.1", "2021.2", "2021.3"]
+            mode: ["icx", "classic", "dpcpp"]
+            libcxx: [None, libstdc++, libstdc++11, libc++]
+            cppstd: [None, 98, gnu98, 03, gnu03, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20, 23, gnu23]
+            runtime: [None, MD, MT, MTd, MDd]
```
memsharded: Do we want to model the runtime as the ``msvc`` more modern one? (dynamic/static)? Cause it is less confusing and it needs to keep in sync with build_type.

memsharded (commented):
This is looking very good.

I think I would add at least 1 full functional test, with the ``pytest.mark.tool_icc`` or equivalent, so it doesn't run in CI yet, but at least we have a full test that compiles something with icc.

conan/tools/intel/inteloneapi.py
franramirez688: Indeed, there is no need to have it there either. The `setvars` script has already a mechanism to handle it. Removing that check

conan/tools/microsoft/visual.py
```diff
@@ -27,9 +27,11 @@ def _write_conanvcvars(conanfile, group):
         return
 
     compiler = conanfile.settings.get_safe("compiler")
-    cvars = None
+    vcvars = None
     if compiler == "intel":
-        cvars = intel_compilervars_command(conanfile)
+        vcvars = intel_compilervars_command(conanfile)
+    elif compiler == "intel-cc":
```
memsharded: I am not convinced this belongs here. The ``CMakeToolchain`` is already generating the intel vars script. Isn't a bit weird that you call ``VCVars().generate()`` and it will create a intel file instead of a vcvars one?

conans/test/conftest.py
```diff
@@ -102,6 +102,11 @@
         "default": "system",
         "system": {"path": {'Windows': 'C:/bazel/bin',
                             "Darwin": '/Users/jenkins/bin'}},
+    },
+    "intel_oneapi": {
```
memsharded: If this is NOT installed in CI, you need to put it in you ``conftest_user.py`` to try things (it is good if you put it in this file as a comment, to explain it, but not activate it and then ``xfail`` the test)

conans/test/functional/toolchains/intel/test_inteloneapi.py
```diff
@@ -0,0 +1,99 @@
...
+
+@pytest.mark.tool_cmake
+@pytest.mark.tool_intel_oneapi
+@pytest.mark.xfail(reason="Intel oneAPI Toolkit is not installed on CI yet")
+@pytest.mark.skipif(platform.system() != "Linux", reason="Only for Linux")
+class TestInteloneAPI:
+
+    @pytest.fixture(autouse=True)
+    def _setUp(self):
+        conanfile = textwrap.dedent("""
```
memsharded: You can get this setup, recommended with ``conan new hello/0.1 -m=cmake_lib`` (or using ``pkg_cmake()`` helper internally should be very close)

memsharded (commented):
Looking great, almost there!

conans/test/functional/toolchains/intel/test_inteloneapi.py
franramirez688: Yes, I know but I thought it was a better idea to create it from scratch instead of using another internal command. Anyway, it's ok for me 👍

memsharded: ``pkg_cmake()`` is actually an internal testing utility designed for this.

conan/tools/microsoft/visual.py
franramirez688: Yes, you're right. Removing it from here. Indeed, it's already integrated into `MSBuildToolchain` as a possible new toolset.

memsharded (approved):

conan/tools/intel/intel_cc.py
```diff
@@ -0,0 +1,177 @@
...
+
+def _is_using_intel_oneapi(compiler_version):
+    """Check if the Intel compiler to be used belongs to Intel oneAPI
+
+    Note: Intel oneAPI Toolkit first version is 2021.1
+    """
+    return int(compiler_version.split(".")[0]) >= 2021
+
+
+def _get_intel_cc_installation_path():
```
lasote: I would remove this, to be honest. I'm sure it works and it is a great job but looks like something hard to maintain and probably can differ in different Linux distributions/package managers. I think the users using that new Intel compiler will be glad to introduce the config. We can mention in the documentation the typical routes.
WDYT @franramirez688 ?

conans/test/functional/toolchains/intel/test_intel_cc.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.mark.tool_cmake
+@pytest.mark.tool_intel_oneapi
+@pytest.mark.xfail(reason="Intel oneAPI Toolkit is not installed on CI yet")
+@pytest.mark.skipif(platform.system() != "Linux", reason="Only for Linux")
+class TestIntelCC:
+    """Tests for Intel oneAPI C++/DPC++ compilers"""
+
+    @pytest.fixture(autouse=True)
+    def _setUp(self):
```
lasote: Why do you need a fixture for a single test? (just curious)

conan/tools/intel/intel_cc.py
franramirez688: Yeah, you're absolutely right but I don't know what to answer... For us, it will be the simplest solution for sure but I think it does not matter if we "suggest" two different default paths if the user did not, does it? In terms of maintenance, I think we can skip it saying that you can put any path through the `[conf]` so we won't have to be aware of the Intel default path changes, right?

conans/test/functional/toolchains/intel/test_intel_cc.py
franramirez688: We don't need it for sure. It was supposed to be for several functional tests but, like the CI is not prepared yet, I only made a single one and forgot to remove it.

conan/tools/intel/intel_cc.py
franramirez688: Well, let's do that. Removing it to simplify the mechanism to find the correct path 👍  Simpler and easier

conan/tools/intel/intel_cc.py
```diff
@@ -118,8 +87,11 @@ def generate(self, group="build"):
     @property
     def installation_path(self):
         """Get the Intel oneAPI installation root path"""
-        installation_path = self._conanfile.conf["tools.intel:installation_path"] or \
-                            _get_intel_cc_installation_path()
+        installation_path = self._conanfile.conf["tools.intel:installation_path"]
+        if not installation_path:
+            raise ConanException("Don't know how to find the Intel oneAPI folder in your system. "
+                                 "Please, provide a path through the [conf] section and the "
+                                 "tools.intel:installation_path variable")
```
lasote:
```suggestion
            raise ConanException("To use Intel oneAPI, specify a [conf] entry " 
                                 "'tools.intel:installation_path' containing the path to the "
                                 "installation folder.")
```

lasote (approved):

PR #9536, opened 2021-09-06
Export to base src and test with git clone

### LINKED ISSUE #9532: Layout source folder when using ``source()`` to ``git clone`` a third-party fails
- Layout defines source folder as ``src``
- Conan exports at least ``conanfile.py`` and ``conanmanifests.txt`` into the ``export`` folder in the cache
- The ``export`` folder in the cache is automatically copied to ``source/src`` in the cache.
- If the package tries to do a ``git clone <url> src`` to put the sources inside the ``src`` folder as the layout specified, it fails with "folder already exists".

Should we?
- Do not copy the ``export`` into the ``source/src`` folder, only the ``exports_sources``?
- Document some git gymnastics to do it, like cloning in a different folder?

lasote:
Changelog: BugFix: Fixed a bug where using the new `layout()` the `exports` went to the `conanfile.source_folder` instead of the base source folder in the cache and only the `exports_sources` should go there.
Docs: omit

Closes #9532

### DIFF
--- a/conans/client/source.py
+++ b/conans/client/source.py
@@ -105,7 +105,7 @@ def config_source(export_folder, export_source_folder, scm_sources_folder, conan
                 # First of all get the exported scm sources (if auto) or clone (if fixed)
                 _run_cache_scm(conanfile, scm_sources_folder, output)
                 # so self exported files have precedence over python_requires ones
-                merge_directories(export_folder, conanfile.source_folder)
+                merge_directories(export_folder, conanfile.folders.base_source)
                 # Now move the export-sources to the right location
                 merge_directories(export_source_folder, conanfile.folders.base_source)
 
@@ -139,7 +139,8 @@ def _run_source(conanfile, conanfile_path, hook_manager, reference, cache,
                 get_sources_from_exports()
 
                 if cache:
-                    _clean_source_folder(src_folder)  # TODO: Why is it needed in cache?
+                    # Clear the conanfile.py to avoid errors cloning git repositories.
+                    _clean_source_folder(src_folder)
                 with conanfile_exception_formatter(conanfile.display_name, "source"):
 
                     with conan_v2_property(conanfile, 'settings',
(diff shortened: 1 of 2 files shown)


conans/test/functional/layout/test_in_cache.py
```diff
@@ -270,3 +271,40 @@ def generate(self):
...
+               def layout(self):
+                   self.folders.source = "src"
+
+               def source(self):
+                   self.run('git clone "{}" src')
+       """).format(repo.replace("\\", "/"))
+
+    client.save({"conanfile.py": conanfile,
+                 "myfile.txt": "My file is copied"})
+    with client.chdir(repo):
```
memsharded, line 294: We have the ``client.init_git_repo()`` method, maybe that can help.

conans/client/source.py
```diff
@@ -105,7 +105,7 @@ def get_sources_from_exports():
                 # First of all get the exported scm sources (if auto) or clone (if fixed)
                 _run_cache_scm(conanfile, scm_sources_folder, output)
                 # so self exported files have precedence over python_requires ones
-                merge_directories(export_folder, conanfile.source_folder)
+                merge_directories(export_folder, conanfile.folders.base_source)
```
memsharded, line 108: I am confused that the ``export_folder`` went to source_folder, and instead below ``export_source_folder`` was already targeted to ``folders.base_source``? I expected the opposite. I guess that ``export_source_folder`` is already containing the "source" layout, is it that?

memsharded (commented):
I'd say this behavior is better than the previous one, I was very confused to find conanfile.py and conanmanifest.txt inside cache ``source/src`` folder.

conans/client/source.py
lasote: Yes, the  `export_source_folder` is containing the `src/` so it goes to the base.

memsharded (approved):

PR #9538, opened 2021-09-06
testing that cmake_paths functionality is covered by CMakeToolchain

memsharded:
Changelog: Omit
Docs: Omit

I was going to remove ``cmake_paths`` from 2.0, but couldn't find explicit tests about its functionality covered by new integrations, adding this test.


conans/test/functional/toolchains/cmake/test_cmaketoolchain_paths.py
```diff
@@ -0,0 +1,57 @@
...
+    client = TestClient()
+    conanfile = textwrap.dedent("""
+        from conans import ConanFile
+        class TestConan(ConanFile):
+            exports = "*"
+            def layout(self):
+                pass
+            def package(self):
+                self.copy(pattern="*", keep_path=False)
+        """)
```
lasote, line 26: Missing `cpp_info.buildirs = ["."]` ? I would say it shouldn't work without it.

memsharded: Test passes green, I thought that too initially... 🤷 😄

lasote: We have to check why

memsharded: I'll check, yes

memsharded: It seems it makes sense:

- ``CppInfo`` declares a ``builddirs = ""``
- ``NewCppInfo`` inherits that value
- ``CMakeToolchain`` generates ``CMAKE_MODULE_PATH``, ``CMAKE_PREFIX_PATH`` from ``for p in cppinfo.builddirs``, doing a os.path.join(), pointing then to the root of the package
- The test is putting the find and config files in the root of the package (many packages will do that)

lasote: that makes sense. I thought the default was `[]`. Actually, I don't think the default should be the root... why the root? why by default? Do we want that for Conan 2.0?

memsharded: Because it is common to create xxx-config.cmake files or find modules in the root of the package, I guess.

What I am concerned here is the transition, does it mean that recipes should add in Conan 1.X all the ``includedirs``, ``libdirs`` directories to their ``package_info()`` now, because there are no defaults for any of them? Lets discuss it.

memsharded: Lets check what is the default of cmake.install() generating a config.cmake file.

memsharded: I have added a real helloConfig.cmake. There is no real default in cmake, and it will look in many different locations: https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure. Libraries CMakeLists.txt will define different locations, but none of them seems to make sense as a default for Conan. In the test I have used ``hello/cmake`` as copied from the internet. The only possibility I see is the root of the package (``CMAKE_INSTALL_PREFIX``)

conans/test/functional/toolchains/cmake/test_cmaketoolchain_paths.py
```diff
@@ -0,0 +1,112 @@
...
+        set(CMAKE_CXX_ABI_COMPILED 1)
+        project(MyHello CXX)
+        cmake_minimum_required(VERSION 3.15)
+
+        find_package({package} REQUIRED)
+        include(myowncmake)
+        """).format(package=package)
+
+    client.save({"CMakeLists.txt": consumer}, clean_first=True)
+    client.run("install hello/0.1@ -g CMakeToolchain")
```
lasote: I think we should ensure that specifying `CMakeDeps` it takes the generated config and not the package one.

memsharded: Yes, I have added new checks to the test, generating CMakeDeps to verify it has priority over the packaged one.

lasote (approved):

PR #9543, opened 2021-09-06
making new env multi

memsharded:
Changelog: Feature: Make new environment generators multi-config (Release/Debug and arch).
Changelog: Fix: Make automatic call of ``VirtualBuildEnv`` and ``VirtualRunEnv`` independent.
Docs: https://github.com/conan-io/docs/pull/2247

Close https://github.com/conan-io/conan/pull/9443/

#tags: slow


conans/test/functional/toolchains/env/test_complete.py
```diff
@@ -122,18 +122,22 @@ class Pkg(ConanFile):
...
             generators = "CMakeDeps", "CMakeToolchain", "VirtualBuildEnv", "VirtualRunEnv"
+            apply_env = False
 
             def build(self):
                 cmake = CMake(self)
                 cmake.configure()
                 cmake.build()
                 self.run("mycmake")
                 self.output.info("RUNNING MYAPP")
+                filename = "conanrunenv-{}-{}".format(self.settings.build_type,
```
lasote: Not a good UI,  Should we enable all the `env` collaborators a common interface to get the name? something like `VirtualRunEnv(self).filename`,  `VCVars(self).filename`, ... 
The bad part is that you need to import the class.

lasote (approved):
I never liked much to have to recall the filename to activate the environment, but now compose it with the settings I think is too much. But overall is good.

conans/test/functional/toolchains/env/test_complete.py
memsharded: But I have made it configurable, in the ``generate()`` method you can change that (same pattern as CMakeDeps)
So you cannot create a new instance, we would need to serialize somewhere the name of the file, and that would be ugly too...

memsharded: I have rewritten the whole ``env_scripts`` approach, now can be used for build, run (and potentially other environments at the user control), so this solves this UX issue. It was necessary some breaking changes, please review.

conan/tools/env/environment.py
```diff
@@ -427,6 +427,7 @@ def loads(text):
         return result
 
 
-def register_environment_script(conanfile, path):
-    if path not in conanfile.environment_scripts:
-        conanfile.environment_scripts.append(path)
+def register_env_script(conanfile, path, group):
```
franramirez688: Please, add a docstring explaining a little bit the function and its parameters.

lasote (approved):
Good idea! now is much more comfortable.

franramirez688 (approved):

conans/test/functional/toolchains/gnu/autotools/test_win_bash.py
```diff
@@ -67,11 +67,13 @@ def build(self):
                  "main.cpp": main})
     client.run("install .")
     client.run("build .")
+    print(client.out)
+    print(client.current_folder)
```
czoido:
```suggestion
```

czoido (approved):

PR #9553, opened 2021-09-07
[develop2] New flat, simple, dynamic and multi-config lockfiles

memsharded:
Changelog: Feature: New flat, simple, dynamic and multi-config lockfiles
Docs: Omit

- New "flat" lockfile format, containing only lists of references (requires, build_requires and py_requires lists)
- Lists are ordered, more modern versions first, so they are resolved first in version ranges
- The approach only works with versions and version ranges, not with revisions yet (we would need to be able to sort by time)
- A lockfile can mix packages and versions for different configurations, the same list can contain a ``winonly/1.0`` package and ``linuxonly/1.0`` package, that are conditionally included based on OS. It requires 2 consecutive evaluations of ``conan lock create``, but they will be added.
- The CI flow of modifying a package in the middle of the graph adds the new version to the lockfile, making it available, so the flow is orders of magnitude simpler.
- I am removing commands like ``conan lock build-order`` and ``conan lock install``, they become regular ``info`` and ``install`` operations, just with a lockfile as argument
- "bundle" operations have been moved to the ``conan graph build-order`` that can be aggregated now, and becomes independent of lockfiles


conans/test/integration/lockfile/test_ci.py
```diff
@@ -0,0 +1,363 @@
...
+               "--lockfile=pkgd.lock --lockfile-out=b_linux.lock")
+    assert "SELF OS: Linux!!" in client.out
+    assert "pkgb/0.2: DEP FILE pkga: HelloA" in client.out
+    b_linux = client.load("b_linux.lock")
+    print(b_linux)
+
+    # To bundle, lockfiles should be complete?
+    client.run("lock bundle create b_win.lock b_linux.lock")
+    print(client.load("lock.bundle"))
+    return
```
lasote: What is the current status of the bundles with this implementation?

lasote: Oh, I just read your comment `"bundle" operations are not implemented yet, the "multi-product" pipeline might still need it, but is not completely clear.`

lasote: I would say the bundles are still needed right? otherwise, the packages not contained in 1 application will be clean from the second and the other way around right?
Should the bundle be just a merge of the lists? could/shold the isolated projects resolve to different versions of the dependencies? 
 I would talk about this to really understand.

memsharded: Yes, I think the bundles would still be needed for multi-product pipelines that needs to run in parallel.
As far as I know, the bundle cannot be a merge of the lists, because different product lockfiles can really belong to different points in time. It stills needs to be a managed set of lockfiles, and computing the aggregated build order is the major challenge, I think it is doable, but still need work. I wanted to make sure that at least the basic approach is good enough and makes sense.

lasote:
How does this manage different "private" requirements to the same library with different versions? (or revisions)

memsharded:
> How does this manage different "private" requirements to the same library with different versions? (or revisions)

Lets say 2 different packages have 2 different, incompatible requirements, private, build-requires, or whatever.
```
pkga -> zlib/1.2.11
pkgb -> zlib/1.2.8
```
The lockfile will store something like:
```
requires=[
zlib/1.2.11,
zlib/1.2.8
]
```
When the dependency resolution algorithm searches for the respective match in the lockfile, it will find it, not an issue (the definition is kind of already locked, lets forget about revisions by now). 

What happens if we have version ranges:
```
pkga -> zlib/[>1.2.8]  # resolves to  zlib/1.2.11
pkgb -> zlib/[<1.2.11] # resolves to  zlib/1.2.8
```
This will resolve to the above lockfile too. When the above lockfile is used for these version ranges, each version range will find the matching locked value in the lockfile, avoiding the resolution of the version range against the cache or remotes.

If we have compatible ranges, that resolve to the same version, the lockfile will only contain ``zlib/1.2.11`` and won't be an issue.
```
pkga -> zlib/[>1.2.8]  # resolves to  zlib/1.2.11
pkgb -> zlib/[>1.0.0] # resolves to  zlib/1.2.11
```
If we have partially compatible ranges, that resolve to different versions
```
pkga -> zlib/[>1.2.8 <1.3]  # resolves to  zlib/1.2.11
pkgb -> zlib/[>1.0.0] # resolves to  zlib/1.3
```
Lockfile will be (always ordered):
```
requires=[
zlib/1.3,
zlib/1.2.11
]
```
Which both of the above ranges will resolve correctly to the locked one (always taking the first valid match).

The reasoning extend for compatible version ranges, and mixing version ranges and fixed versions. As long as:

- The dependency resolution algorithm is deterministic, and do not resolve the same range in different packages to different versions
- The lockfile can be ordered from newer to older (need to do it in revisions too to work)
- The policy in version ranges is always using the latest valid in the range.

conans/test/integration/lockfile/test_ci.py
memsharded: Instead of using bundles, which added an extra layer of complexity on lockfiles, I have proposed an aggregation of "build-order" files, which is independent of lockfiles (build-order files can be computed while applying a lockfiles, and the distributed build from an aggregated build-order can and should use lockfiles too).

Please re-review, thanks!

conans/test/integration/lockfile/test_lock_pyrequires.py
```diff
@@ -0,0 +1,123 @@
...
+
+    client.run("export dep dep/0.1@")
+    client.run("export dep dep/0.2@")
+    client.run("export toola toola/0.1@")
+    client.run("export toolb toolb/0.1@")
+    client.run("create pkga pkga/0.1@")
+    client.run("create pkgb pkgb/0.1@")
+    client.run("lock create app/conanfile.py --lockfile-out=conan.lock")
+
+    client.run("export dep dep/0.2@")
```
lasote: Exporting again `dep/0.2` without changes?

conans/test/integration/lockfile/test_lock_pyrequires.py
```diff
@@ -0,0 +1,123 @@
...
+
+    client.run("install app/conanfile.py --lockfile=conan.lock")
+    assert "pkga/0.1: toola: toola/0.1!!" in client.out
+    assert "pkgb/0.1: toolb: toolb/0.1!!" in client.out
+    assert "pkga/0.1: dep: dep/0.1!!" in client.out
+    assert "pkgb/0.1: dep: dep/0.2!!" in client.out
+
+    client.run("install app/conanfile.py")
```
lasote, line 81: Add a comment: Without version range declared, the resolution is deterministic and the lockfile doesn't mess up at all.
Is that the idea?

conans/test/integration/lockfile/test_lock_requires.py
```diff
@@ -0,0 +1,253 @@
...
+    client.run("create pkg pkg/0.1@ -s os=Linux")
+
+    client.run("lock create consumer/conanfile.txt  --lockfile-out=conan.lock -s os=Windows"
+               " -s:b os=Windows")
+    print(client.load("conan.lock"))
+    assert "win/0.1 from local cache - Cache" in client.out
+    assert "pkg/0.1 from local cache - Cache" in client.out
+    client.run("lock create consumer/conanfile.txt  --lockfile=conan.lock "
+               "--lockfile-out=conan.lock -s os=Linux -s:b os=Linux")
+    print(client.load("conan.lock"))
```
lasote: caught!

conans/test/integration/lockfile/test_lock_requires.py
```diff
@@ -0,0 +1,253 @@
...
+                 "consumer/conanfile.txt": f"[{requires}]\npkg/0.1"})
+    client.run("create dep win/0.1@")
+    client.run("create dep nix/0.1@")
+
+    client.run("create pkg pkg/0.1@ -s os=Windows")
+    client.run("create pkg pkg/0.1@ -s os=Linux")
+
+    client.run("lock create consumer/conanfile.txt  --lockfile-out=conan.lock -s os=Windows"
+               " -s:b os=Windows")
+    print(client.load("conan.lock"))
```
lasote: caught!

conans/test/integration/lockfile/test_lock_requires.py
```diff
@@ -0,0 +1,253 @@
...
+    client.run("install consumer --lockfile=conan.lock -s os=Windows -s:b os=Windows")
+    assert "win/0.1 from local cache - Cache" in client.out
+    assert "win/0.2" not in client.out
+    client.run("install consumer -s os=Windows -s:b os=Windows")
+    assert "win/0.2 from local cache - Cache" in client.out
+    assert "win/0.1" not in client.out
+    assert "nix/0.1" not in client.out
+
+    client.run("install consumer --lockfile=conan.lock -s os=Linux -s:b os=Linux")
+    print(client.out)
```
lasote: caught!

conans/test/integration/lockfile/test_lock_requires.py
```diff
@@ -0,0 +1,253 @@
...
+
+    client.run("lock create consumer/conanfile.txt  --lockfile-out=conan.lock -s os=Linux"
+               " -s:b os=Linux")
+    assert "dep/0.2 from local cache - Cache" in client.out
+    assert "dep/0.1" not in client.out
+    client.run("lock create consumer/conanfile.txt  --lockfile=conan.lock "
+               "--lockfile-out=conan.lock -s os=Windows -s:b os=Windows")
+    assert "dep/0.1 from local cache - Cache" in client.out
+    assert "dep/0.2" not in client.out
+    print(client.load("conan.lock"))
```
lasote: caught!

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
...
+        """)
+
+    client.save({"ffmepg/conanfile.py": ffmpeg,
+                 "variant/conanfile.py": variant})
+    client.run("export ffmepg ffmpeg/1.0@")
+    client.run("export variant nano/1.0@")
+
+    client.run("lock create --reference=nano/1.0@ --build --lockfile-out=conan.lock")
+    lockfile = client.load("conan.lock")
+    print(lockfile)
```
lasote: caught!

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
...
+    # assert '"options": "variation=nano"' in lockfile
+
+    client.run("lock build-order conan.lock --lockfile-out=conan.lock --build=missing "
+               "--json=build_order.json")
+
+    json_file = client.load("build_order.json")
+    print("JSON: ", json_file)
```
lasote: caught!

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
...
+    to_build = json.loads(json_file)
+    f = to_build[0]
+    print(f)
```
lasote: caught!

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
...
+
+    cmd = "install {} --build={} {} --lockfile=conan.lock".format(f[0], f[0], f[3])
+    print(cmd)
```
lasote: caught!

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
...
+    client.run(cmd)
+    print(client.out)
```
lasote: caught!

conans/test/integration/lockfile/test_lock_requires.py
```diff
@@ -0,0 +1,253 @@
...
+    client.run("create dep dep/1.1@")
+
+    client.run("create pkg pkg/0.1@ -s os=Windows")
+    client.run("create pkg pkg/0.1@ -s os=Linux")
+
+    client.run("lock create consumer/conanfile.txt  --lockfile-out=conan.lock -s os=Windows"
+               " -s:b os=Windows")
+    assert "dep/0.1 from local cache - Cache" in client.out
+    assert "dep/1.1" not in client.out
+    client.run("lock create consumer/conanfile.txt  --lockfile=conan.lock "
```
lasote, line 212: Add a comment: `The dep/0.1 is ignored from the lock file but nothing fail`
I get it but this point is a bit controversial. We have a locked version that is not compatible so it is not used.
Have you thought about a strict mode where the references should be taken from the lock file or fail? Do we want to enter that "garden"?

conans/test/integration/lockfile/test_options.py
```diff
@@ -0,0 +1,52 @@
+import json
+import textwrap
+
+import pytest
+
+from conans.test.utils.tools import TestClient
+
+
+@pytest.mark.xfail(reason="lockfiles wip")
```
lasote, line 9: Real still wip, right?

lasote (approved):
🥳

conans/test/integration/lockfile/test_lock_pyrequires.py
memsharded: Yes, it was copied from a previous test, removing.

conans/test/integration/lockfile/test_lock_pyrequires.py
memsharded: Actually this test tried to do something with revisions, but then I was blocked by the lack of revisions time, so I didn't try to solve it now, and the test is a bit confusing. I am adding comments with TODO to test the correct revisions behavior.

conans/test/integration/lockfile/test_lock_requires.py
memsharded: Added comments about this behavior, and added extra checks in the lockfile and creating new versions of dep that will be safely ignored by using the lockfile.

conans/test/integration/lockfile/test_options.py
memsharded: Yes, the options need to be fixed first

conans/test/integration/lockfile/test_lock_requires.py
memsharded: Added the strict mode for consuming with ``conan install``

conans/client/conan_api.py
```diff
@@ -233,7 +232,7 @@ def create(self, conanfile_path, name=None, version=None, user=None, channel=Non
 
             if lockfile_out:
                 lockfile_out = _make_abs_path(lockfile_out, cwd)
-                graph_lock_file = GraphLockFile(profile_host, profile_build, graph_lock)
+                graph_lock_file = GraphLockFile(graph_lock)
```
czoido: If we do a `conan create . --lockfile-out=result.lock` without input lockfile this would crash when saving after this line?

memsharded: Yep, at the moment it crashes. All the UX related to who can create and when lockfiles is to be defined. We could completely avoid the ``conan lock create`` and replace it with the new ``conan graph info --lockfile-out=conan.lock``, would be basically equivalent.

czoido (approved):
This is a more intuitive way of locking and using lockfiles that what we had in 1.X, you can even use it right away with just checking the command help.

Issue #7456, opened 2020-07-30
[feature] Update package remote when uploading


puetzk:
To control our dependencies, we keep the conan-center usually disabled in our conan config, and copy recipes which we intend to use (and binaries built with the desired profiles) onto our own enterprise artifactory instance. However, after doing this, the instance that did the builds will still have the recipe in its cache with the remote set to conan-center.

This can be changed with `conan remote update_ref`, but that makes it into a two step process, and it is somewhat easy to forget.

I would like to ask for the `conan upload -r <remote>` to get an `--update_ref` option, which would update the ref of the just-uploaded binaries to now consider the location they were just uploaded to as the upstream remote. `conan upload` already sets set the remote if there isn't one: https://github.com/conan-io/conan/blob/d67545c1cff4853c44f8c979eeb6bf4819614f34/conans/client/cmd/uploader.py#L327-L329 
so this is not entirely novel behavior. I would just like an (optional, off-by-default) way to tell it to overwrite it even if already set, for this use case of internalizing recipes we have rebuilt. I look at this as being somewhat analogous to `git push --set-upstream`. It's something you can already do (in both cases), but doing it in two separate commands means you have to specify the reference(s) twice, and offers the chance to mismatch them (or other mistakes, e.g. do update_ref even if the upload failed) that leave the cache out of sync with reality. Combining it would be convenient and slightly less mistake-prone.

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Hi @puetzk 

Let me understand a couple of things:

- Why are you concerned about defining this value? Are you using it for later --updates?
- Wouldn't it be good to do the updates explicitly, specifying ``--remote``?
- Are you using revisions already?
- Are you reusing the cache of the job that uploads the package to your Artifactory? If you used clean caches in those jobs, then it wouldn't be an issue, would it?

We are probably going to remove the usage of remote information in Conan 2.0, and updates will be remote independent. I would like to understand better the pain to start thinking in the right direction for Conan 2.0.

Regarding the feature request, it doesn't seem difficult but it seems a bit too ad-hoc, if possible I would like to do a step towards something as close as the Conan 2.0 behavior as possible.

puetzk:
> Why are you concerned about defining this value?

[...]

PR #9574, opened 2021-09-09
Remove reference with remote association

czoido:
Changelog: Remove: Remove references with remotes association in database.
Docs: omit

Now references are not associated with remotes in any way. Also, for example the conan upload command will fail if you don't add a remote, before this it took the associated remote or the default remote.
Some logic has also changed in the retrieve_exports_sources. Before, if there was a remote associated with a package in the cache that did not have the sources downloaded, Conan tried to download the sources from that remote. Now that we have revisions and consider the those inmutable we can iterate the remotes and try to download the sources from any remote in the remotes because we have the revision of the package.


puetzk:
Looks like this approach would close #7456

It doesn't implement the original feature request, but if there's no association being recorded it can't be out-of-date and doesn't need to be updated.

conan/cache/db/cache_database.py
```diff
@@ -23,8 +23,8 @@ def dump(self, output: StringIO):
         self._references.dump(output)
 
     def update_reference(self, old_ref: ConanReference, new_ref: ConanReference = None,
-                         new_path=None, new_remote=None, new_timestamp=None, new_build_id=None):
-        self._references.update(old_ref, new_ref, new_path, new_remote, new_timestamp, new_build_id)
+                         new_path=None, new_timestamp=None, new_build_id=None):
+        self._references.update(old_ref, new_ref, new_path, new_timestamp, new_build_id)
```
memsharded, line 27: You are completely removing the remote information. I thought that we would still store (and only store, do nothing with it, just information) that remote. But if it is not being used, then it might be better to just simplify it, remove from the codebase, and if you need it, with revisions it is a ``conan list`` command and you will get where it is from the remotes. Wdyt @lasote?

conans/client/cache/remote_registry.py
```diff
@@ -273,77 +273,28 @@ def load_remotes(self):
...
-                            if self._cache.get_remote(prev) == renamed:
-                                self._cache.set_remote(prev, remote_name)
 
     def update(self, remote_name, url, verify_ssl=True, insert=None):
         self._validate_url(url)
         remotes = self.load_remotes()
         remotes.update(remote_name, url, verify_ssl, insert)
         remotes.save(self._filename)
 
-    def clear(self):
```
memsharded, line 294: Is ``conan remote clean`` command removed?

conans/client/printer.py
```diff
@@ -113,12 +113,6 @@ def _print(it_field, show_field=None, name=None, color=Color.BRIGHT_GREEN):
             _print("package_revision", name="Package revision", color=None)
             _print("binary", name="Binary", color=None)
 
-            if show("binary_remote") and is_ref:
```
memsharded, line 102: Why this removed but not the above ``if show("remote") and is_ref:``?

conans/client/source.py
```diff
@@ -23,23 +23,16 @@ def retrieve_exports_sources(remote_manager, cache, recipe_layout, conanfile, re
...
         return None
 
-    # TODO: cache2.0 check what behaviour for this in 2.0
-    # If not path to sources exists, we have a problem, at least an empty folder
-    # should be there
-    current_remote = cache.get_remote(recipe_layout.reference)
-    if current_remote:
-        current_remote = remotes[current_remote]
-    if not current_remote:
+    if not remote:
```
memsharded: Probably the ``retrieve_export_sources`` shouldn't be tied to the current remote. If this happens in a ``conan upload`` command, it will use the remote you are uploading to, but instead the sources must be downloaded from other remote that contains this revision.

memsharded (commented):
This is looking good, I am looking forward this simplification.
There are a couple of hypothesis/behaviors that need to be checked.

conans/client/printer.py
czoido: Sorry, I missed this, removing

conan/cache/db/cache_database.py
lasote: I thought about removing it too in the branch I'm working on. Let's simplify to the max first, we will have years to complicate it.

conans/client/source.py
memsharded: I have checked, this is being called in ``uploader`` with the same remote that you are going to upload it, so this would fail there (not sure why it is not failing in the test suite now). We need to check this.

memsharded (changes requested):
I would like to move this forward, but there are still some pending things, for example, the ``retrieve_exports_sources`` seems to be wrong, but the tests are passing.

Otherwise, I think this is looking good, great job, looking forward to be able to merge it.

conans/client/source.py
czoido: I removed the remote argument. Now if there's some remote selected it's taken and otherwise it searches for the first remote with sources to download them.

conans/client/cache/cache.py
```diff
@@ -105,14 +105,14 @@ def remove_layout(self, layout):
         layout.remove()
         self._data_cache.remove(ConanReference(layout.reference))
 
+    def get_timestamp(self, ref):
+        return self._data_cache.get_timestamp(ConanReference(ref))
+
     def get_remote(self, ref):
-        return self._data_cache.get_remote(ConanReference(ref))
+        return None
```
memsharded: Mark this as FIXME to be removed? It seems they are here to avoid test breaking?

conans/client/graph/graph_binaries.py
```diff
@@ -233,29 +196,50 @@ def _process_node(self, node, pref, build_mode, update, remotes):
...
-
-        if latest_prev_for_pkg_id:  # Binary already exists in local, check if we want to update
-            node.prev = latest_prev_for_pkg_id.revision
-            self._evaluate_cache_pkg(node, latest_prev_for_pkg_id, remote, remotes, update)
-        else:  # Binary does NOT exist locally
-            # Returned remote might be different than the passed one if iterating remotes
-            remote = self._evaluate_remote_pkg(node, pref, remote, remotes, remote_selected)
-
-        node.binary_remote = remote
+        if not node.binary:
```
memsharded: This should belong to the ``else`` clause in line 213 (the ``if`` clearly always define ``node.binary``)

conans/client/rest/rest_client_v2.py
```diff
@@ -320,13 +320,10 @@ def get_package_revisions(self, pref):
     def get_latest_recipe_revision(self, ref):
         url = self.router.recipe_latest(ref)
         data = self.get_json(url)
-        rev = data["revision"]
-        # Ignored data["time"]
-        return ref.copy_with_rev(rev)
+        return ref.copy_with_rev(data.get("revision")), from_iso8601_to_timestamp(data.get("time"))
```
memsharded, line 327: Anotate this with a FIXME; if the server should return timestamp instead of ISO

conans/client/source.py
```diff
@@ -5,15 +5,31 @@
...
 from conans.paths import CONANFILE, CONAN_MANIFEST, EXPORT_SOURCES_TGZ_NAME, EXPORT_TGZ_NAME
 from conans.util.conan_v2_mode import conan_v2_property
 from conans.util.files import (is_dirty, mkdir, rmdir, set_dirty_context_manager,
                                merge_directories, clean_dirty)
 
 
-def retrieve_exports_sources(remote_manager, cache, recipe_layout, conanfile, ref, remotes):
+def _try_get_sources(ref, remote_manager, recipe_layout, remote):
+    try:
+        remote_manager.get_recipe_revisions(ref, remote)
```
memsharded: why the double remote_manager call? Can't we call just ``get_recipe_sources`` and filter on return exception?

memsharded (approved):
Still some open issues, but can be merged and iterated later.

conans/client/cache/cache.py
czoido: I think it's ok to remove this, I used this in the beginning to develop the feature and forgot to remove that, also the set_remote. I'm removing both.

conans/client/graph/graph_binaries.py
```diff
@@ -232,29 +195,50 @@ def _process_node(self, node, pref, build_mode, update, remotes):
...
-
-        node.binary_remote = remote
+            if update:
+                try:
+                    remote, remote_prev, prev_time = self._get_package_from_remotes(node, pref,
+                                                                                    remotes, update)
+                except NotFoundException:
+                    output.warning("Can't update, no package in remote")
+                except NoRemoteAvailable:
+                    output.warning("Can't update, no remote defined")
```
lasote: Should this happen? Does it mean that there are no remotes configured or...? <= in that case we should improve the message?

lasote (approved):
Great job. 👏  Very clear code and consistent with the "update flows".
Just a comment of something I don't fully understand.

PR #9580, opened 2021-09-09
remove legacy os_build arch_build

memsharded:
Changelog: Remove: remove legacy os_build arch_build settings
Docs: Omit


memsharded (commented):
Note: many of the legacy CMake and VS helpers tests that rely on these ``xxx_build`` settings have been directly removed, as removing the legacy CMake helper will come in the next PR

czoido (approved):

lasote (approved):

PR #9583, opened 2021-09-10
develop2 new package_id computation

memsharded:
Changelog: Improvements in package_id computation
Docs: Omit


memsharded:
Too conflicted, broken and outdated. Seems nothing to rescue here

PR #9584, opened 2021-09-10
Fix symlinked conan cache dir bug

dscole:
Changelog: Bugfix: Fix symlinked conan cache directory bug
Docs: https://github.com/conan-io/docs/pull/9579

This PR fixes the situation when the conan cache dir (`~/.cache`) is a symlink and you attempt to create a package that also has internal symlinks in it. Previously, the symlinks were lost during the package creation process.

### DIFF
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -409,7 +409,7 @@ def merge_directories(src, dst, excluded=None):
             linkto = os.path.join(os.path.dirname(pointer_src), linkto)
 
         # Check if it is outside the sources
-        out_of_source = os.path.relpath(linkto, os.path.realpath(src)).startswith(".")
+        out_of_source = os.path.relpath(linkto, src).startswith(".")
         if out_of_source:
             # May warn about out of sources symlink
             return


memsharded (commented):
While apparently this doesn't break the CI test suite, I am slightly concern that this could break anything, symlinks are complicated and it is possible that not enough testing is done to cover this, and users have wildly different usages, with often intensive use of symlinks.

Below the:
```python
for src_dir, dirs, files in walk(src, followlinks=True):
```
Is already following the symlinks, in that case a symlink inside the folder would require the `` os.path.realpath(src)`` in `` out_of_source = os.path.relpath(linkto, os.path.realpath(src)).startswith(".")`` to be in the same base? I don't see this case tested in the suite.

Asking for @lasote feedback who implemented this, but it was time ago, so we would need to wait until some more testing is added.

lasote:
Honestly, I don't recall either understand most of the code. Not sure why but it will break something almost for sure.

memsharded:
It seems that this PR became outdated, without enough evidence of the problem it is solving (no tests provided), and also with a potential high risk of breaking other users.
It seems also that symlinking the cache wouldn't be the most recommended approach, it is already possible and recommended to define the CONAN_HOME (Conan 2.0), CONAN_USER_HOME (Conan 1.X) to define the cache location in a Conan-aware fashion.

Closing this PR, thanks very much for the contribution, don't hesitate to create new tickets to discuss the topic if needed.

PR #9595, opened 2021-09-13
Feature/cache short second attempt

lasote:
Changelog:  omit
Docs: omit

- To discuss the length we want. 
- This alleviates the PATH env variables limit and windows users are forced to use the user home folder instead of `C:/`
- Introduced assert and simplified apparently dead code

### DIFF
--- a/conan/cache/cache.py
+++ b/conan/cache/cache.py
@@ -1,4 +1,4 @@
-import base64
+import hashlib
 import hashlib
 import os
 import shutil
@@ -6,15 +6,16 @@ import time
 import uuid
 from io import StringIO
 
+from conan.cache.conan_reference import ConanReference
+from conan.cache.conan_reference_layout import RecipeLayout, PackageLayout
 # TODO: Random folders are no longer accessible, how to get rid of them asap?
 # TODO: Add timestamp for LRU
 # TODO: We need the workflow to remove existing references.
 from conan.cache.db.cache_database import CacheDatabase
-from conan.cache.conan_reference import ConanReference
-from conan.cache.conan_reference_layout import RecipeLayout, PackageLayout
-from conans.errors import ConanException, ConanReferenceAlreadyExistsInDB, ConanReferenceDoesNotExistInDB
+from conans.errors import ConanException, ConanReferenceAlreadyExistsInDB, \
+    ConanReferenceDoesNotExistInDB
 from conans.model.info import RREV_UNKNOWN, PREV_UNKNOWN
-from conans.util.files import md5, rmdir
+from conans.util.files import rmdir
 
 
 class DataCache:
@@ -55,11 +56,8 @@ class DataCache:
         hash = hash.encode("utf-8")
         md = hashlib.sha256()
         md.update(hash)
-        sha_bytes = md.digest()
-        tmp = base64.b64encode(sha_bytes)  # Trick to reduce the len from 64 to 44
-        # 9 is the default len of the shorted git commit, as this is base64 => 6
-        # FIXME: This "/" replace is a bit dirty
-        return tmp.decode("ascii")[0:6].replace("/", "_")  # '/' is a base64 valid char, replace it
+        sha_bytes = md.hexdigest()
+        return sha_bytes[0:9]
 
     @staticmethod
     def _get_tmp_path():
--- a/conan/cache/cache.py
+++ b/conan/cache/cache.py
@@ -58,7 +58,8 @@ class DataCache:
         sha_bytes = md.digest()
         tmp = base64.b64encode(sha_bytes)  # Trick to reduce the len from 64 to 44
         # 9 is the default len of the shorted git commit, as this is base64 => 6
-        return tmp.decode("ascii")[0:6]
+        # FIXME: This "/" replace is a bit dirty
+        return tmp.decode("ascii")[0:6].replace("/", "_")  # '/' is a base64 valid char, replace it
 
     @staticmethod
     def _get_tmp_path():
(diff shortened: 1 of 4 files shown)


conan/cache/cache.py
```diff
@@ -53,7 +54,9 @@ def _get_tmp_path():
 
     @staticmethod
     def _get_path(ref: ConanReference):
-        return md5(ref.full_reference)
+        value = ref.full_reference.encode("utf-8")
```
memsharded: can't we assume at this point that all Conan strings are encoded to utf-8?

memsharded (approved):
There is pending an improvement in the graph management in the "install binaries" step, that would mitigate this ugly pattern, but that should be done later.

conan/cache/cache.py
```diff
@@ -53,7 +54,9 @@ def _get_tmp_path():
...
+        sha = sha256(value)
```
memsharded: BTW, a common trick to reduce from 2->1,3 ratio characters is to encode the hash from hexadecimal to base64

conan/cache/cache.py
lasote: at that point, according to test failures, I would say it is an unicode object

conan/cache/cache.py
lasote: Nice idea

conan/cache/cache.py
memsharded: Oh, I see, the thing is that ``sha256`` requires bytes, didn't realize it.

czoido:
What about making this configurable (per package or global) and by default using the whole path? Maybe for most of the cases we don't need the shortening and it reduces the probability of collision?

Also, and out of the scope of this PR but related, regarding the length of the environment variables, maybe we could also shorten a little bit more when generating the environment scripts. I mean that for example we generate a _conanrunenv.sh_ like this:

```
echo Configuring environment variables

export PATH="/path/to/conan/home/that/may/be/long/d8eea279520277c9a1cd8a3d91e45b79/package/bin:$PATH"
```
 but could that also be shortened using this?

```
echo Configuring environment variables
export CONAN_PATH = "/path/to/conan/home/that/may/be/long"
export PATH="$CONAN_PATH/d8eea279520277c9a1cd8a3d91e45b79/package/bin:$PATH"
```

In case you have many transitive dependencies you could also save lots of characters from the max of the PATH environment variable. Don't know if this have been discussed before or not...

memsharded:
> What about making this configurable (per package or global) and by default using the whole path? Maybe for most of the cases we don't need the shortening and it reduces the probability of collision?

I don't think so at this stage, let's keep it as simple as possible.

I agree with the environment improvements, but that should be a different effort.

czoido:
I'm not convinced about shortening the paths by default. We have two risks here, in one hand we have the probability of someone having problems in Windows with the paths being too long and in the other hand someone having lots of packages that will have collisions in the cache. With the time we know that both of them will happen. I don't know which one of those is more likely...

Adding some numbers here so we can evaluate the risk of collision vs the risk of having problems with path length in Windows:

| chars for hash (not base64) | number recipes + binaries | % chance of collision |
|-----------------------------|---------------------------|-------------|
| 8                           | 1000                      | 0.011%      |
| 8                           | 10.000                    | 1.15%       |
| 8                           | 100.000                   | 68.78%      |
| 10                          | 1000                      | 4.56*10^-5  |
| 10                          | 10.000                    | 0.004%      |
| 10                          | 100.000                   | 0.45%       |
| 12                          | 1000                      | 1.7*10^-7   |
| 12                          | 10.000                    | 1.7*10^-5   |
| 12                          | 100.000                   | 0.0017      |
| 16                          | 1000                      | 2.7*10^-12  |
| 16                          | 10.000                    | 2.7*10^-10  |
| 16                          | 100.000                   | 2.7*10^-8   |
| 16                          | 1.000.0000                | 2.7*10^-6   |

Used
```
chance_of_collision = 1 - e^(-n^2 / (2 * d))
 d: 2^(bits_used_for_hash)
 bits_used_for_hash: 4*truncated_chars_length
 n: number_packages, number of hashes to check the probability for
```
to calculate. More info: https://stackoverflow.com/questions/51622061/how-can-i-calculate-the-impact-on-collision-probability-when-truncating-a-hash and https://preshing.com/20110504/hash-collision-probabilities/

Maybe at least we should start with a higher number of bytes and see if someone has problems with collisions or path length. With 16 characters (can be reduced with base64) it should be pretty safe.

lasote:
I've removed the base64 trick because:
a) "/" is a valid character and it mess with the paths 
b) The casing "a" vs "A" can cause additional collisions in Windows/Macos

lasote:
@memsharded check if we should increment to 16

memsharded:
> I've removed the base64 trick because:

You applied it in the wrong place. You can take the base64 of the first N chars of the hexadecimal sha256. That will not have any problem, and will be faster to compute.

memsharded:
> @memsharded check if we should increment to 16

Yes, I'd say we increment to 12 base64 characters => first 16 hexadecimal chars of the sha256

lasote:
> You applied it in the wrong place. You can take the base64 of the first N chars of the hexadecimal sha256. That will not have any problem, and will be faster to compute.

I don't get why it won't have any problem. The output of a base64 contains "/" and letters with casing "a", "A".

memsharded:
You are right, I didn't realize that base64 includes the backslash
character, and indeed the casing could cause issues.

Let's keep the hexadecimal then

czoido (approved):

PR #9596, opened 2021-09-13
Change default cmake_layout() source folder

### LINKED ISSUE #9549: Default ``cmake_layout()`` source folder
This layout is assuming that the main library ``CMakeLists.txt`` is inside a ``src`` subfolder in the repo.

This is probably not the most common user layout, in which they will have a CMakeLists.txt in the root of the folder:

- How to model this?
- Should we change the ``cmake_layout()`` default to ``folders.source = "."``?
- Check implications of defining the root as source folder.

tapia:
Changelog: Feature: Change default `cmake_layout()` source folder from 'src' to '.'
Docs: https://github.com/conan-io/docs/pull/2225

Closes #9549

### DIFF
--- a/conans/assets/templates/new_v2_cmake.py
+++ b/conans/assets/templates/new_v2_cmake.py
@@ -20,7 +20,7 @@ class {package_name}Conan(ConanFile):
     default_options = {{"shared": False, "fPIC": True}}
 
     # Sources are located in the same place as this recipe, copy them to the recipe
-    exports_sources = "src/*"
+    exports_sources = "CMakeLists.txt", "src/*"
 
     def config_options(self):
         if self.settings.os == "Windows":
@@ -81,7 +81,7 @@ project(PackageTest CXX)
 
 find_package({name} CONFIG REQUIRED)
 
-add_executable(example example.cpp)
+add_executable(example src/example.cpp)
 target_link_libraries(example {name}::{name})
 """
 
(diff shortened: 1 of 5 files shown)


conans/assets/templates/new_v2_cmake.py
```diff
@@ -216,12 +216,12 @@ def get_cmake_lib_files(name, version, package_name="Pkg"):
                                                          package_name=package_name),
              "src/{}.cpp".format(name): source_cpp.format(name=name, version=version),
              "src/{}.h".format(name): source_h.format(name=name, version=version),
-             "src/CMakeLists.txt": cmake_v2.format(name=name, version=version),
+             "CMakeLists.txt": cmake_v2.format(name=name, version=version),
```
memsharded, line 219: Maybe we want to follow CMake common pattern that is the top root uses a ``add_subdirectory(src)`` and keep a CMakeLists.txt inside "src" with the heavy logic.

memsharded:
Please, target the PR to ``release/1.40`` branch @tapia

conans/assets/templates/new_v2_cmake.py
tapia: Wouldn't having two `CMakeLists.txt` be too verbose? I mean, what would go into the root file, besides the `add_subdirectory()` call?

memsharded: It is verbose because it is a hello-world simple example. But it would probably be the most common layout out there in CMake projects. It will contain the project(), cmake_minimum_required() and the ``add_subdirectory()``

tapia: Looks completely unnecessary to me, but if it's the consensus, who am I to disagree 😄 What do you think, @lasote ?

lasote: I dont want to see two CMakeLists.txt in a hello world, please.

lasote (approved):

PR #9597, opened 2021-09-13
add failing test CMakeToolchain find_package() fails

memsharded:
Changelog: Bugfix: CMakeToolchain makes CMakeDeps ``find_package()`` without ``REQUIRED`` to fail if version doesn't match.
Docs: Omit


memsharded:
Superseded by https://github.com/conan-io/conan/pull/9598, it was me using locally cmake 3.15.1 instead of 3.15.7

PR #8399, opened 2021-01-27
Feature/conan v2 error

memsharded:
Changelog: Fix: `CONAN_V2_MODE` env-var does not longer alter behavior, only raises errors for Conan 2.0 incompatibilities
Docs: https://github.com/conan-io/docs/pull/2031

#tags: slow
#revisions: 1

PR #9615, opened 2021-09-16
Fix migration settings comparison

### LINKED ISSUE #9614: [bug] 'settings.yml is locally modified, can't be updated' every time I upgrade
  * Operating System+version: macOS 11.5.2, Windows 10 21H1
  * Compiler+version: AppleClang 12, MSVC 19.29.30133.0 (VS2019)
  * Conan version: 1.40.1
  * Python version: Python 3.9.7

Run:
```
pip3 install conan --upgrade
conan --version
```

I get this warning:
```
WARN: Migration: Updating settings.yml
WARN: ****************************************
WARN: settings.yml is locally modified, can't be updated
WARN: The new settings.yml has been stored in: /Users/steve/.conan/settings.yml.new
WARN: ****************************************
```

I'm not making any changes to this file so I would not expect Conan to consider it locally modified. It used to never happen, but has happened on every release recently. I haven't experienced any downstream impacts, though I usually move settings.yml.new to settings.yml manually just to make sure I have the latest.

czoido:
Changelog: Fix: Fix migrations settings comparison.
Docs: omit

There was a newline difference between default settings in __init__.py and the ones from migrations_settings.py, changing the literal string comparison to the same comparison that's tested in `test_migrations_matches_config`. 

Closes: https://github.com/conan-io/conan/issues/9614

### DIFF
--- a/conans/client/migrations.py
+++ b/conans/client/migrations.py
@@ -47,7 +47,7 @@ class ClientMigrator(Migrator):
         self.out.warn("Migration: Updating settings.yml")
         if hasattr(migrations_settings, var_name):
             version_default_contents = getattr(migrations_settings, var_name)
-            if version_default_contents != get_default_settings_yml():
+            if version_default_contents.splitlines() != get_default_settings_yml().splitlines():
                 current_settings = load(cache.settings_path)
                 if current_settings != version_default_contents:
                     save_new()
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -131,6 +131,7 @@ _t_default_settings_yml = Template(textwrap.dedent("""
 
 
     cppstd: [None, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20, 23, gnu23]  # Deprecated, use compiler.cppstd
+
     """))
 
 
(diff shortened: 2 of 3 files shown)


conans/client/migrations.py
```diff
@@ -47,7 +47,7 @@ def save_new():
         self.out.warn("Migration: Updating settings.yml")
         if hasattr(migrations_settings, var_name):
             version_default_contents = getattr(migrations_settings, var_name)
-            if version_default_contents != get_default_settings_yml():
+            if version_default_contents.splitlines() != get_default_settings_yml().splitlines():
```
memsharded, line 50: How did the line endings became different now?
Shouldn't the lineendings be the same, and not requiring the splitlines?

czoido: Looks like the ___init__.py_ _yaml_ does not include a new line in the end of the file [after removing some jinja templates](https://github.com/conan-io/conan/pull/8399) for conan 1.34. So now if someone wants to update from any version > 1.33 they will have a _settings.yml_ that is going to be considered different unless we modify the migrations_settings for all versions from 1.33
I thought this may be more reliable for things like this happening in the future and may allow users from 1.33 upgrade, but the other option is also possible.

czoido: I have added a literal comparison in the test after adding a newline to the generated settings.yml but left the splitlines for checking in the migration if it can be upgraded, wdyt?

memsharded: Good for me 👍

memsharded (approved):

lasote (approved):

PR #9624, opened 2021-09-17
Improve error message

### LINKED ISSUE #9626: [bug] Unhelpful error message about conflicting package versions between requires and build_requires
Having conflicting version overrides for the same package in build_requires and requires results in the following output: `ERROR: True`. This makes the error hard to find. The error message should be more helpful on what is going wrong.

#4753 might be related.

  * Operating System+version: Arch Linux Kernel 5.12.12
  * Conan version: 1.40.0
  * Python version: 3.9.5

Try to `conan install` this example conanfile.txt:

```
[requires]
libdeflate/1.7
[build_requires]
libdeflate/1.8
```

```$ conan install conanfile.txt
Configuration:
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
[options]
[build_requires]
[env]

ERROR: True
```

martin-schulze-vireso:
Fixes #9626, which contains a reproducer. While working with a project that had two conflicting version overrides for a dependency in build_requires and requires I got the unhelpful message:

```
ERROR: True
```

This tries to improve upon that message.
For closer inspection: I found `node.ref` to be `None`, which skips the outputs inside `self._conflicting_references`.

Changelog: Fix:  Don't just print  "ERROR: True" on unresolvable conflict.
Docs: Omit

Close #9626

### DIFF
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -288,7 +288,7 @@ class DepsGraphBuilder(object):
                 # Maybe it was an ALIAS, so we can check conflict again
                 conflict = self._conflicting_references(previous, require.ref, node.ref)
                 if conflict:
-                    raise ConanException(conflict)
+                    raise ConanException("Unresolvable conflict between {} and {}".format(previous, require.ref))
 
             # Add current ancestors to the previous node and upstream deps
             for n in previous.public_closure:
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -288,7 +288,7 @@ class DepsGraphBuilder(object):
                 # Maybe it was an ALIAS, so we can check conflict again
                 conflict = self._conflicting_references(previous, require.ref, node.ref)
                 if conflict:
-                    raise ConanException("Unresolvable conflict between {} and {}".format(previous, require.ref))
+                    raise ConanException(conflict)
 
             # Add current ancestors to the previous node and upstream deps
             for n in previous.public_closure:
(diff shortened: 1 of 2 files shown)


memsharded:
Thanks for contributing this @martin-schulze-vireso 

It seems this behavior was not properly covered by tests, and that is the reason it was that unhelpful. Do you think you could manage to contribute a test for this fix? If not, please tell and we will try to contribute it.

Update: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-9624/1/pipeline it seems the change broke some tests. Maybe the type of the conflict needs to be check, or other alternative solution.

martin-schulze-vireso:
@memsharded I already saw the test failures and must admit I am a bit intimidated. ;) However, I created and linked an issue (#9626) with a reproducer that might help you to come up with a test quite fast.

memsharded:
Contributed a test, and moved the message to another location to avoid breaking existing tests.

Note, however, that we are already in the path of moving to 2.0, the whole conflict reporting will change in 2.0, most likely this has already been addressed there (in ``develop2`` branch).

conans/client/graph/graph_builder.py
```diff
@@ -328,7 +328,8 @@ def _conflicting_references(previous, new_ref, consumer_ref=None):
                         "in your root package."
                         % (consumer_ref, consumer_ref, new_ref, next(iter(previous.dependants)).src,
                            previous.ref, new_ref.name))
-            return True
+            return "Unresolvable conflict between {} and {}".format(previous.ref, new_ref)
```
martin-schulze-vireso, line 331: There are two more returns in this function. Should they be changed as well? Also, I believe that the one on line 341 was the actual return path in my case.

memsharded: The use case you provided is actually satisfied with this change, other cases might need other tests then.

martin-schulze-vireso: I guess the other paths are those parts that resulted in the test errors from my fix above. It still feels a bit icky to wrap a True into an exception, but since we're on the way to 2.0 anyways...

martin-schulze-vireso:
Thanks for the fast response.

PR #9634, opened 2021-09-20
[develop2] Improving installation (build, binary download) process

memsharded:
Changelog: Omit
Docs: Omit

Use an explicit model of processing the same ``pref`` while installing (downloading, building), instead of an implicit ``processed_references`` cache.

This could be used as a base for the lockfile bundles build-order, I am investigating this.


czoido (approved):

lasote (approved):
I think i get the general idea and looks great, but I would need further explanations to really understand it.

PR #9645, opened 2021-09-22
[develop2] Remove the automatic creation of default profile

memsharded:
Changelog: Remove: According to the Tribe 2.0, remove the automatic creation of default profile
Docs: Omit

This has been accidentally partially mixed with https://github.com/conan-io/conan/pull/9648, needs to process it first.


conans/test/functional/layout/test_in_cache.py
```diff
@@ -147,6 +147,7 @@ def test_cache_in_layout(conanfile):
     assert "Package_ID: {}".format(package_id) in client.out
 
 
+@pytest.mark.xfail(reason="The conan-source command do not handle yet the layout, to do in other PR")
```
memsharded, line 150: @lasote the ``conan source`` case has not been addressed in this PR, it will be later in a separate one, better to keep this one focused on the profile issue.

conans/client/conan_api.py
```diff
@@ -484,7 +484,7 @@ def config_init(self, force=False):
         else:
             self.app.cache.initialize_config()
             self.app.cache.registry.initialize_remotes()
-            self.app.cache.initialize_default_profile()
+            self.app.cache.reset_default_profile()
```
lasote: The `self.app.cache.reset_default_profile()` is done when `force` and `else`. So it could be extracted from both.

conans/client/cache/cache.py
```diff
@@ -262,10 +257,8 @@ def generators_path(self):
 
     @property
     def default_profile_path(self):
-        if os.path.isabs(self.config.default_profile):
-            return self.config.default_profile
-        else:
-            return os.path.join(self.cache_folder, PROFILES_FOLDER, self.config.default_profile)
+        # Used only in testing
```
lasote: Also used at `reset_default_profile`

lasote (approved):

czoido (approved):

PR #9647, opened 2021-09-22
Remove JSON, MD and YCM generators for Conan 2.0

czoido:
Changelog: Remove: Remove markdown, youcompleteme and json generators.
Docs: omit

- JSON generator will probably be substituted for a json output in `conan info` command.
- Markdown and YouCompleteMe generators will probably be provided by an external repository and be installed via `conan config install`


memsharded (approved):

lo1ol:
Hi there!

I continue migration to conan 2.0 and noticed that json generator was removed. Qbs's [ConanfileProbe](https://doc.qt.io/qbs/qml-qbsprobes-conanfileprobe.html) is based on the output of this generator. So it's doesn't work anymore(

In this MR was the suggestion, that json generator will be replaced with output  `conan info`.  Will be returned `info` command in future?

memsharded:
Yes, there are a few toolsl that don't fit that clearly into the concept of "generators", which we use mostly for toolchains.

The ``conan graph info --format=json`` will probably be the place were all the different graph json outputs will be centralized (there is some overlap and divergence between them). 

Also the concept that will be more powerful in 2.0 is the custom commands. As the python API that returns a dependency graph will be public and documented, and it will be possible to create your own custom commands over that API (and distribute them with ``conan config install``), then it is possible that some of these tools will be implemented via a command (but this is still an ongoing discussion, nothing decided yet)

lo1ol:
I got your idea about custom commands. It's will be useful and modular. Our team can use this feature for safe uploading packages:)

Also, I explored `graph info` command and found that it can't show `package_folder`. As I understand, path to this folder ts discover at install  stage (after graph_compute). So, after this changes, `package_folder` become available:
```python
diff --git a/conans/cli/commands/install.py b/conans/cli/commands/install.py
index 7a5e0d60..aa212e74 100644
--- a/conans/cli/commands/install.py
+++ b/conans/cli/commands/install.py
@@ -89,6 +89,10 @@ def graph_compute(args, conan_api, strict_lockfile=True):
     conan_api.graph.analyze_binaries(deps_graph, args.build, remotes=remotes, update=args.update)
 
+    remote = get_multiple_remotes(conan_api, args.remote)
+    conan_api.install.install_binaries(deps_graph=deps_graph, build_modes=args.build,
+                                       remotes=remote, update=args.update)
+
     return deps_graph, lockfile
```

I know it's bad patch, because installing some packages on  "get info stage" is not obviously. Nevertheless, I believe that it's possible to get` package_folder` without binary packages installation.

Also, I think that I can call conan directly (with deploy generator), as temporary solution for my qbs projects

PR #9651, opened 2021-09-22
Update deploy generator for 2.0 and remove legacy classes

czoido:
Changelog: Feature: Update `deploy` generator. Remove old `Generator` inheritance. Also using (for the moment) `generators_folder` as output folder.
Docs: (pending 2.0)

Also removing all legacy Generator classes and therefore the custom generators.


memsharded (approved):

lo1ol:
Hi!

I tries to migrate my projects to conan 2.0. I notice that I can't use custom generators after this changes. My generator setups qbs settings directory via specific commands (QbsToolchains doesn't call their).

There will be no opportunity to define custom generators in conan 2.0 anymore? Is there some workarounds?

memsharded:
HI @lo1ol 

Custom generators in Conan 2.0 are indeed possible, the difference is that they are explicit:

- Define whatever logic you want in a "python_requires" recipe. Make sure to use the new ``self.dependencies`` access
- In the consumer that you want to use that, ``python_requires= "your package"``
- Then use the functionality you want, directly in the ``generate()`` method, in the same way the built-in generators are often used.

lo1ol:
Thank you for so fast answer! I try it, now)

lo1ol:
Your suggestion is works)

memsharded:
> Your suggestion is works)

Happy to hear that :)

One  added benefit of the new generators inside ``python_requires`` is that you can use the versioning to induce new builds:

- python_requires affect the ``package_id`` of its users, by default in "minor_mode". If you don't bump the version or bump the patch, it will not force a rebuild from source from its users
- But if you bump the minor version of the python_requires containing your generator, the packages using it will require a re-build from source.

This is a better default than the previous one for generators inside regular packages.

PR #9652, opened 2021-09-22
remove generator manager

memsharded:
Changelog: Omit
Docs: Omit

Remove:
- Old GeneratorManager
- Legacy ``write_toolchain()``

Continuation of https://github.com/conan-io/conan/pull/9651


ShuangLiu1992:
seems like when calling requires() on editable packages, conf defiend by tool_requires are not propagated and the generate() function that rely on conf information fails for those packages

PR #9654, opened 2021-09-22
remove legacy should_xxx

memsharded:
Changelog: Remove: Remove legacy ``--should-configure``, ``--should-build``, etc. will be replaced with ``conf`` items if makes sense (``tools.build:skip_test``).
Docs: Omit


lasote (approved):
Merge when green

PR #9657, opened 2021-09-23
Remove CppInfo and update to NewCppInfo

czoido:
Changelog: Remove: Remove `CppInfo` and update to `NewCppInfo`.
Docs: Pending (2.0)

Now, components in the `CppInfo` are not initialised with default paths like in 1.X (this can be discussed), only the global component is initialised with this information:

```
includedirs = ["include"]
libdirs = ["lib"]
resdirs = ["res"]
bindirs = ["bin"]
builddirs = [""]
frameworkdirs = ["Frameworks"] 
```


conans/client/installer.py
```diff
@@ -541,19 +536,21 @@ def _call_package_info(self, conanfile, package_folder, ref, is_editable):
...
                     if not is_editable:
                         package_cppinfo = conanfile.cpp.package.copy()
                         package_cppinfo.set_relative_base_folder(conanfile.folders.package)
-                        # Copy the infos.package into the old cppinfo
-                        fill_old_cppinfo(conanfile.cpp.package, conanfile.cpp_info)
+                        conanfile.cpp_info = package_cppinfo
                     else:
-                        conanfile.cpp_info.filter_empty = False
+                        conanfile.cpp_info = conanfile.cpp.package.copy()
+                        conanfile.cpp_info.clear()
```
czoido: This is because you could previously declare a component attribute in the `layout` method in the conanfile and in package_info you are appending a path to it, as it was previously initialised. But in editable mode that information is discarded and you don't want to use that path but still you have an append to the attribute, so here we get the information that the component attribute was declared in the `cpp.package` but discard the path it was set to with `clear()`

conan/tools/google/bazeldeps.py
```diff
@@ -65,34 +65,34 @@ def _get_dependency_buildfile_content(self, dependency):
...
+        for linkopt in dependency.cpp_info.system_libs:
             linkopts.append('"-l{}"'.format(linkopt))
         linkopts = ', '.join(linkopts)
 
         context = {
             "name": dependency.ref.name,
-            "libs": dependency.new_cpp_info.libs,
-            "libdir": dependency.new_cpp_info.libdirs[0],
+            "libs": dependency.cpp_info.libs,
+            "libdir": dependency.cpp_info.libdirs[0],
```
lasote: Is this `dependency.cpp_info.libdirs[0]` still a relative path right? Is this working?

conans/client/installer.py
```diff
@@ -541,19 +536,21 @@ def _call_package_info(self, conanfile, package_folder, ref, is_editable):
                 self._hook_manager.execute("pre_package_info", conanfile=conanfile,
                                            reference=ref)
                 if hasattr(conanfile, "layout"):
-                    # Old cpp info without defaults (the defaults are in the new one)
-                    conanfile.cpp_info = CppInfo(conanfile.name, package_folder,
-                                                 default_values=CppInfoDefaultValues())
                     if not is_editable:
                         package_cppinfo = conanfile.cpp.package.copy()
                         package_cppinfo.set_relative_base_folder(conanfile.folders.package)
```
lasote: I don't think we want to have absolute paths here, it was done before to maintain the compatibility, as the old `self.cpp_info` contained absolute paths.

conans/client/installer.py
lasote: I don't get it :( 
I would need an example, maybe we can simplify something.

conans/model/new_build_info.py
```diff
@@ -2,15 +2,32 @@
...
-from conans.model.build_info import DefaultOrderedDict
-
 _DIRS_VAR_NAMES = ["includedirs", "srcdirs", "libdirs", "resdirs", "bindirs", "builddirs",
                    "frameworkdirs"]
 _FIELD_VAR_NAMES = ["system_libs", "frameworks", "libs", "defines", "cflags", "cxxflags",
                     "sharedlinkflags", "exelinkflags"]
 
 
-class _NewComponent(object):
+class DefaultOrderedDict(OrderedDict):
```
lasote: We probably should rename the module `new_build_info.py` too.

lasote (commented):
I'll send the partial review as I'm talking to you right now to discuss it.

conans/client/installer.py
czoido: As discussed, all that part has been removed now. Now the cpp_info from the conanfile is the same as cpp.info and the old logic will probably be changed.

conan/tools/google/bazeldeps.py
czoido: Yes, that's a relative path, looks like it works fine...

conans/test/functional/layout/test_editables_layout.py
```diff
@@ -17,7 +17,10 @@ def layout(self):
         self.cpp.build.includedirs = ["my_include"]
         self.cpp.build.libdirs = ["my_libdir"]
         self.cpp.build.libs = ["hello"]
-        self.cpp.build.frameworkdirs = []  # Empty list is also explicit priority declaration
+
+        # TODO: check which is the desired behaviour here... an empty list will not overwrite the
+        # conanfile cpp_info, so this will not override the default value already set.
+        self.cpp.build.frameworkdirs = []
```
lasote, line 22: I think we need a way to say, in editables I don't have/want whatever.
I agree the "clear_none" and all that stuff is too much complicated but there might be another way.

conans/model/build_info.py
```diff
@@ -2,10 +2,10 @@
 import os
 from collections import OrderedDict
 
-_DIRS_VAR_NAMES = ["includedirs", "srcdirs", "libdirs", "resdirs", "bindirs", "builddirs",
+_DIRS_VAR_NAMES = ["_includedirs", "_srcdirs", "_libdirs", "_resdirs", "_bindirs", "_builddirs",
                    "frameworkdirs"]
```
lasote: `_ frameworkdirs`?

conans/model/build_info.py
```diff
@@ -2,10 +2,10 @@
...
-_FIELD_VAR_NAMES = ["system_libs", "frameworks", "libs", "defines", "cflags", "cxxflags",
-                    "sharedlinkflags", "exelinkflags"]
+_FIELD_VAR_NAMES = ["_system_libs", "_frameworks", "_libs", "_defines", "_cflags", "_cxxflags",
+                    "sharedlinkflags", "_exelinkflags"]
```
lasote: `_ sharedlinkflags`?

lasote (commented):
Not strictly related to this PR: To discuss later with the team, should we keep a `conanfile.new_cpp_info` as an alias of `conanfile.cpp_info` in case someone already migrated? Should we change now in "develop" branch the `new_cpp_info` to a better name? @memsharded

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -100,7 +100,7 @@ def setUp(self):
             from conans.tools import save
             import os
             class Pkg(ConanFile):
-                settings = "build_type"
+                settings = "build_type", "compiler"
```
lasote: why needed?

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -127,7 +127,7 @@ def _run_build(self, settings=None, options=None):
         # Run the configure corresponding to this test case
         build_directory = os.path.join(self.client.current_folder, "build").replace("\\", "/")
         with self.client.chdir(build_directory):
-            self.client.run("build .. %s %s" % (settings, options))
+            self.client.run("build .. %s %s --build=missing" % (settings, options))
```
lasote: ?

conans/test/functional/toolchains/cmake/test_file_api.py
```diff
@@ -8,6 +8,7 @@
 from conans.test.utils.tools import TestClient
 
 
+@pytest.mark.xfail(reason="NewCppInfo: xfail until it adds default folders for the components")
```
lasote, line 11: I don't see components here. but I trust you, just wanted to double check

conans/test/integration/command/test_package_test.py
```diff
@@ -128,46 +130,6 @@ def test_build_folder_handling(self):
                                                     "build_folder")))
         self.assertFalse(os.path.exists(default_build_dir))
 
-    def test_check_version(self):
```
lasote: how is this related?

conans/test/integration/conanfile/conanfile_errors_test.py
```diff
@@ -1,6 +1,8 @@
 import textwrap
 import unittest
 
+import pytest
```
lasote: 🤔

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -1,6 +1,8 @@
 import os
 import textwrap
 
+import pytest
```
lasote: 🤔

conans/test/unittests/model/conanfile_test.py
```diff
@@ -1,5 +1,6 @@
 import unittest
 
+import pytest
```
lasote: 🤔

lasote (approved):
Great! some comments about stuff in the tests files that I don't really understand.

conans/test/functional/toolchains/cmake/test_file_api.py
czoido: Yes, the CMakeFileAPI implementation fails because we removed the defaults in folder for components, this is something I imagine will be easy to fix later as @SSE4 knows that code I did not wanted to get into it because I have not so much experience with that part.

conans/test/integration/command/test_package_test.py
czoido: I removed it because it uses the old cpp_info attributes in the test: 
```
            class Dep(ConanFile):
                def package_info(self):
                    self.cpp_info.name = "MyDep"
```

conans/test/functional/toolchains/cmake/test_cmake.py
czoido: Sorry, that slipped there after some changes as well as the other, reverting.

lasote (approved):

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -81,7 +81,7 @@ def content(self):
             if dep.is_build_context and dep.ref.name not in self.build_context_activated:
                 continue
 
-            cmake_find_mode = dep.new_cpp_info.get_property("cmake_find_mode", "CMakeDeps")
+            cmake_find_mode = dep.cpp_info.get_property("cmake_find_mode", "CMakeDeps")
```
memsharded: This rename is my major concern. I think we need to think the path of migrating from 1.X to 2.0 regarding this, and it probably can't be the legacy ``cpp_info``

memsharded (commented):
I think this is fantastic, great job.

My only pending concern is the name for the migration. Users will need to implement their own integrations, generations, etc, and they must use both ``self.dependencies`` and the ``new_cpp_info``. But then, this code will live in ``python_requires`` recipes, and moving to Conan 2.0 will break it, no way we can make it work, right?

conan/tools/cmake/cmakedeps/cmakedeps.py
memsharded: I have a proposal: As the change is done in the ``ConanFileInterface`` new interface, accessed by ``self.dependencies``, which is completely new model, experimental we can do:

- Rename ``ConanFileInterface.new_cpp_info`` to ``ConanFileInterface.cpp_info``: is the ``new_cpp_info``, already converted, just a rename of the property, but it really returns a NewCppInfo object.
- Rename the ````ConanFileInterface.cpp_info`` to ````ConanFileInterface.legacy_cpp_info`` in case it is necessary somewhere. Maybe we could even remove it, and force ``self.dependencies`` adopters to use exclusively the new model.

Ping @lasote @czoido

PR #9662, opened 2021-09-24
Minor output improvements

lasote:
Minor improvements to the message color printing and removed dead code from the command launcher.

### DIFF
--- a/conans/cli/output.py
+++ b/conans/cli/output.py
@@ -130,10 +130,10 @@ class ConanOutput(object):
         self._write_message(msg, fg=fg, bg=bg)
 
     def highlight(self, data):
-        self.info(data, Color.BRIGHT_MAGENTA)
+        self.info(data, fg=Color.BRIGHT_MAGENTA)
 
     def success(self, data):
-        self.info(data, Color.BRIGHT_GREEN)
+        self.info(data, fg=Color.BRIGHT_GREEN)
 
     def warning(self, msg):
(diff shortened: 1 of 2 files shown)


conans/client/command.py
```diff
@@ -1964,25 +1964,9 @@ def run(self, *args):
             except IndexError:  # No parameters
                 self._show_help()
                 return False
-            try:
-                commands = self._commands()
-                method = commands[command]
-            except KeyError as exc:
```
lasote, line 1970: All this is managed in the v2 cli launcher

conans/cli/output.py
```diff
@@ -109,12 +109,19 @@ def rewrite_line(self, line):
         self._color = tmp_color
 
     def _write_message(self, msg, fg=None, bg=None):
+        tmp = ""
         if self._scope:
-            msg = "{}: {}".format(self.scope, msg)
-        if self._color:
-            msg = "{}{}{}{}".format(fg or '', bg or '', msg, Style.RESET_ALL)
+            if self._color:
+                tmp = "{}{}{}:{} ".format(fg or '', bg or '', self.scope, Style.RESET_ALL)
```
memsharded, line 115: If we want to simplify output coloring, we can do it too (now the key is that only the ``pkg/version:`` part of the message is colored, while the rest is plain white. But we can change that if we want.

memsharded (approved):

PR #9665, opened 2021-09-24
upload takes into account revisions

memsharded:
Changelog: Feature: Smart ``upload`` that uses ``revisions`` to optimize when to download things
Changelog: Remove: removed ``upload --no-overwrite`` argument, with revisions it doesn't make sense.
Docs: Omit

Changes:
- Removed ``get_recipe_manifest`` and ``get_package_manifest`` from the whole stack, it was used only for upload, doesn't make sense with revisions
- Removed ``get_package_snapshot``, not necessary with revisions. ``get_recipe_snapshot`` had to stay for the case of ``-force`` an upload of recipe that needed to remove some files in the server (like removed conan_sources.tgz).
- Removed some checks like ``is_package_snapshot_complete``, that could be remotely useful, but were done in the wrong place, in the client, handling corrupted packages in the server side
- If revision exist in server=>do nothing, do not compress, do not upload. 
- Removed ``--no-overwrite`` arguments, that is the default behavior now.(users might still require a new mode, of not being able to upload a NEW revision of the same version, that might be not possible.
- Removed grouping by remotes in the ``CmdUpload``, because now by design, it only uploads to 1 remote.
- Removed dead ``get_package_info`` api


lasote (approved):
👍 👍

czoido (approved):

PR #9668, opened 2021-09-24
fix generators_folder relative path to conanfile

memsharded:
Changelog: Bugfix: ``generator_folder`` was using the ``cwd`` instead of a relative folder to the conanfile, producing layout errors.
Docs: Omit

#tags: slow


lasote (changes requested):
So, is this superseded by https://github.com/conan-io/conan/pull/9674/files ?
I would say this is breaking if not conditioned to the usage of the layout.

conans/test/functional/toolchains/ios/test_using_cmake.py
```diff
@@ -55,6 +55,7 @@ def package(self):
             """)
         })
 
+    @pytest.mark.tool_cmake(version="3.20")
```
franramirez688: We will need to add it in `conftest.py`, e.g.:

```python
        "3.20": {
            "path": {'Windows': 'C:/cmake/cmake-3.20.0-win64-x64/bin',
                     'Darwin': '/Users/jenkins/cmake/cmake-3.20.0/bin',
                     'Linux': '/usr/share/cmake-3.20.0/bin'}
        },
```

Anyway, let's wait for CI members to check out the final installation folders.

conans/test/functional/toolchains/cmake/test_v2_cmake_template.py
```diff
@@ -7,6 +11,15 @@ def test_cmake_lib_template():
     # Local flow works
     client.run("install . -if=install")
     client.run("build . -if=install")
+    client.run("package . -if=install")
+    print(os.path.join(client.current_folder))
```
lasote: print

conans/client/conan_api.py
```diff
@@ -796,10 +797,19 @@ def package(self, path, build_folder, package_folder, source_folder=None, instal
...
-        conanfile.folders.set_base_build(build_folder)
-        conanfile.folders.set_base_source(source_folder)
-        conanfile.folders.set_base_package(package_folder)
-        conanfile.folders.set_base_install(install_folder)
+        if hasattr(conanfile, "layout"):
+            dir_path = os.path.dirname(conanfile_path)
+            conanfile.folders.set_base_generators(dir_path)
+            conanfile.folders.set_base_build(dir_path)
+            conanfile.folders.set_base_source(dir_path)
+            # FIXME: this is bad, but conanfile.folders.package=xxx is broken atm, cant be used
```
lasote, line 805: Please, explain.

lasote: I mean, I don't understand why/what is broken.

lasote (approved):
I think we should open an issue to fix `folders.package` or remove it if we really don't want it. But now, being hardcoded is useless.

franramirez688 (approved):

conan/tools/cmake/cmake.py
```diff
@@ -141,7 +141,20 @@ def install(self, build_type=None):
...
+
+        bt = build_type or self._conanfile.settings.get_safe("build_type")
+        if not bt:
+            raise ConanException("build_type setting should be defined.")
+        is_multi = is_multi_configuration(self._generator)
+        build_config = "--config {}".format(bt) if bt and is_multi else ""
+
+        pkg_folder = args_to_string([self._conanfile.package_folder.replace("\\", "/")])
+        build_folder = args_to_string([self._conanfile.build_folder])
+        arg_list = ["--install", build_folder, build_config, "--prefix", pkg_folder]
```
yowidin, line 153: The `--install` command line argument isn't available in earlier versions of the CMake, it was probably introduced only in CMake 3.15. Please consider changing the call back to a build call with the `install` target.

PR #9675, opened 2021-09-27
Conan 2.0: Input testing new mock

lasote:
Changelog: Fix: Changed `TestClient` the user/password mocking to a generic `inputs` array that will mock directly the `stdin` and return the `inputs` in order when a string is requested. The mocking this way is much more easy and robust.

The "real" changes (not only tests) are here: `conans/test/utils/tools.py`


memsharded (approved):
This is great.

czoido (approved):

PR #9678, opened 2021-09-27
VirtualRunEnv: avoid generating linux/mac env variables on Windows

tapia:
… env variables on Windows

Changelog: Fix: Avoid "overcrowding" Windows environment variables with `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH`, not used in Windows.
Docs: Omit


conan/tools/env/virtualrunenv.py
```diff
@@ -10,11 +10,14 @@ def runenv_from_cpp_info(conanfile, cpp_info):
     if cpp_info.bin_paths:  # cpp_info.exes is not defined yet
         dyn_runenv.prepend_path("PATH", cpp_info.bin_paths)
     # If it is a build_require this will be the build-os, otherwise it will be the host-os
-    if cpp_info.lib_paths:
-        dyn_runenv.prepend_path("LD_LIBRARY_PATH", cpp_info.lib_paths)
-        dyn_runenv.prepend_path("DYLD_LIBRARY_PATH", cpp_info.lib_paths)
-    if cpp_info.framework_paths:
-        dyn_runenv.prepend_path("DYLD_FRAMEWORK_PATH", cpp_info.framework_paths)
+
+    os = str(conanfile.settings.get_safe("os"))
```
memsharded: Actually, it is a bit more subtle than this:

- Build context should get the ``settings_build.os``
- Host context should get the ``settings.os``

memsharded (commented):
Tests with dual contexts would be necessary.

conans/test/integration/environment/test_env.py
```diff
@@ -105,14 +105,14 @@ def build(self):
     client.run("install . -s:b os=Windows -s:h os=Linux --build=missing")
     # Run the BUILD environment
     ext = "bat" if platform.system() == "Windows" else "sh"  # TODO: Decide on logic .bat vs .sh
-    cmd = environment_wrap_command(ConanFileMock(), "conanbuildenv", "mycmake.{}".format(ext),
+    cmd = environment_wrap_command(ConanFileMock(), "conanbuildenv.{}".format(ext), "mycmake.{}".format(ext),
```
memsharded: This shouldn't be necessary, why did you add it?

conans/test/integration/environment/test_env.py
```diff
@@ -95,35 +95,33 @@ def build_requirements(self):
                 self.build_requires("mygtest/1.0", force_host_context=True)
 
             def build(self):
-                mybuild_cmd = "mycmake.bat" if platform.system() == "Windows" else "mycmake.sh"
-                self.run(mybuild_cmd, env="conanbuildenv")
-                mytest_cmd = "mygtest.bat" if platform.system() == "Windows" else "mygtest.sh"
-                self.run(mytest_cmd, env="conanrunenv")
+                self.run("mycmake.bat", env="conanbuildenv")
```
memsharded, line 98: But this test did work for ``platform.system() != Windows``, with an ``.sh`` script. Probably the ``build()`` can use ``self.settings.os`` to decide which one, but seems this test is testing less than it did before.

conans/test/integration/environment/test_env.py
```diff
@@ -389,3 +387,66 @@ def generate(self):
...
+        assert "LD_LIBRARY_PATH" not in conanbuildenv
+        assert "LD_LIBRARY_PATH" not in conanrunenv
+    else:
+        conanbuildenv = client.load("conanbuildenv.sh")
+        conanrunenv = client.load("conanrunenv.sh")
+        assert "LD_LIBRARY_PATH" in conanbuildenv
+        assert "LD_LIBRARY_PATH" in conanrunenv
+
+    # Build context LINUX - Host context LINUX
+    client.run("install consumer_pkg -s:b os=Linux -s:h os=Linux --build")
```
memsharded, line 427: Better be sure to remove the previous generated script files, or to split the test or something. If something was not working, it would read the script file from the previous calls.

memsharded (approved):

conans/test/integration/environment/test_env.py
lasote: Well, this test is/was a mess. Previously the choice of `.bat` `.sh` on the conan client code was based on `platform.system()` so according to that, the test, independently of the applied profiles, could work in any operating system, but that was wrong! As this PR now solves that issue and the generated "launchers" follow the build profile, not every check can be done in all systems.

memsharded: yes, agree, probably this test should be splitted in Windows and Linux specific ones. Not necessarily in this PR, good for me atm.

lasote (approved):

PR #9682, opened 2021-09-28
Feature/2.0 clean http requester

lasote:
Changelog: Fix: Stop passing the http requester to the Conan apis and apps, it was used only in testing.

### DIFF
--- a/conans/test/utils/tools.py
+++ b/conans/test/utils/tools.py
@@ -16,6 +16,7 @@ from contextlib import contextmanager
 from urllib.parse import urlsplit, urlunsplit
 
 import bottle
+import mock
 import requests
 from mock import Mock
 from requests.exceptions import HTTPError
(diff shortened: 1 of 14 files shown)


conans/test/utils/tools.py
```diff
@@ -542,7 +527,12 @@ def run(self, command_line, assert_error=False):
             self.out = RedirectedTestOutput()  # Initialize each command
             with redirect_output(self.out):
                 with redirect_input(self.user_inputs):
-                    return self.run_cli(command_line, assert_error=assert_error)
+                    if self.requester_class:
```
memsharded: I think you have removed the possibility to run the ``artifactory_ready`` tests against Artifactory, is this intended?

memsharded (commented):
Changes looks great, but check comment about Artifactory tests.

conans/test/utils/tools.py
lasote: That was kind of working because the `TestRequester` was doing a fallback to `requests` when the URL didn't match `fake***` but I changed the code to be explicit and not mock the http_requester.

memsharded (approved):

PR #9686, opened 2021-09-28
MSBuildDeps handle build_requires executables

### LINKED ISSUE #9545: [bug] cross-compile visual_studio generator does not use build requirements bin folders
  * Compiler+version: visual studio (tested on 15 and 16)
  * Conan version: 1.39 (tested pre-release 1.40)
  * Python version: 3.8.10

os = Windows
Create a host profile with arch  = x86
Create a build profile with arch = x86_64
Both profiles specify the visual studio comiler, other profile settings can be the same.

Create a recipe using visual_studio generator.

In the recipe, list a package (such as protobuf or flatbuffers) as both a requirement and a build requirement.  In this case, the  requirement is needed for package libraries/headers but the build requirement is needed for the package executable (protoc or flatc). The libraries/headers should be the x86 variant and the executables should be the x86_64 variant

Generate visual studio properties using:

conan install . -g visual_studio -pr:h host_profile -pr:b build_profile

On examination of the "props" file created, it will be seen that the property file has a Conan-"pkg"-Root macro pointing to the requirements (x86) build of the two packages but nothing pointing to the build version (x68_64) of the package.

You might say this is no problem because x86 executables can run on an x86_64 build machine but this logic fails to consider the overall cross compile concept and would not work if I were trying to cross build ARM visual studio projects.  While I have not tested yet, I'm also concerned about cmake generators but perhaps that one is correct.

Expected results:

Conan-"pkg"-Root should still exist as it could be used for accessing the dependency package headers/libraries although these are generally already handled automatically.

NEW:
Conan-"pkg"-Build-Root must exist to get access to the bin folder of the build requirement to access protoc and flatc.  Note that in the case of flatbuffers, the package is actually smart enough not to build flatc if it is being used as a compile dependency so attempting to reference the x86 flatc fails as it is not built.

EXTRA CREDIT:
In researching this, I came across feature request #7162 asking to add ExecutablePath to the visual_studio Props file:

<ExecutablePath>D:\.conan\bfef1e\1\bin;$(ExecutablePath)</ExecutablePath>

In the above case, the paths added to the executable path obviously must be the build requires and not the requires package bin paths.  In fact, "requires" paths should not be added as this might cause attempts to execute bad executables in an ARM cross-compile.

memsharded:
Changelog: Feature: ``MSBuildDeps`` generator learned how to handle ``build_requires`` to use executables from them.
Docs: https://github.com/conan-io/docs/pull/2248

Close https://github.com/conan-io/conan/issues/9545

### DIFF
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -7,7 +7,6 @@ from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
 from conans.errors import ConanException
-from conans.model.build_info import DepCppInfo
 from conans.util.files import load, save
 
 VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
@@ -150,7 +149,7 @@ class MSBuildDeps(object):
         condition = " And ".join("'$(%s)' == '%s'" % (k, v) for k, v in props)
         return condition
 
-    def _vars_props_file(self, name, cpp_info, deps, build=False):
+    def _vars_props_file(self, dep, name, cpp_info, deps, build=False):
         """
         content for conan_vars_poco_x86_release.props, containing the variables
         """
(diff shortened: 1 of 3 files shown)


conan/tools/microsoft/msbuilddeps.py
```diff
@@ -284,4 +295,28 @@ def _content(self):
             dep_content = self._dep_props_file(dep_name, file_dep_name, props_name, condition)
             result[file_dep_name] = dep_content
 
+        build_req = list(self._conanfile.dependencies.build.values())
+        for dep in build_req:
+            dep_name = dep.ref.name
+            dep_name = dep_name.replace(".", "_") + "_build"
+            cpp_info = DepCppInfo(dep.cpp_info)  # To account for automatic component aggregation
```
lasote: We could start using the `new_cpp_info` here and calling `aggregate_components` for that, the same for the `host/test`

lasote (approved):
cool!

conan/tools/microsoft/msbuilddeps.py
memsharded: Agree, lets try it. 
However, we should probably refactor that, so we can pick a name for ``new_cpp_info`` that will survive in the transition to 2.0

lasote (approved):

PR #9705, opened 2021-10-01
[2.0] Search dead code and moved code

lasote:
- Moved `conans.client.api` to `conans.cli.api`
- Removed search functionality from the old command (already moved to cli)


czoido (commented):
I just have some doubts about moving the api to `conans.cli.api` maybe we can move it to `conan.api`?

memsharded (approved):

memsharded:
> I just have some doubts about moving the api to conans.cli.api maybe we can move it to conan.api?

I think that a re-organization of the whole folders structure is necessary, in the ``conan`` domain, not in the ``conans`` one.

czoido (approved):

PR #9706, opened 2021-10-01
[2.0] Remove commands 'config set/rm/get'

lasote:
We are going to avoid the commands that only edit/read files.


memsharded:
> What is going to be the equivalent? Is there going to be an equivalent? Asking because I use config set.

@grafikrobot the idea is that editing text files shouldn't require learning a specific command. The alternatives for this could be:

- Manage your own config file, with ``conan config install``, instead of relying on local editions of a config file. More robust and escalable.
- Many of the config, specifically the tools config, can be moved to profiles now with the new ``[conf]`` section. 
- New config might be provided also on the command line with ``-c key=value``
- The new ``global.conf`` syntax is key=value store, the idea is that Conan will still help giving the path to the file. Users can still do stuff like concat something to the file, but probably using the above strategies would be better.

conans/client/conan_api.py
```diff
@@ -401,21 +401,6 @@ def install(self, path="", name=None, version=None, user=None, channel=None,
...
-    def config_get(self, item):
-        app = ConanApp(self.cache_folder)
-        if item == "storage.path":
-            result = app.config.storage_path
-        else:
-            result = app.config.get_item(item)
-        ConanOutput().info(result)
-        return result
-
-    @api_method
```
memsharded, line 414: Seems like below ``config_rm()`` should be removed too?

memsharded (approved):

memsharded:
>Since I'm using the API directly on CI and wrappers programs to set single options non of those seem like usable alternatives to me.

Why not? The python API can also call ``conan config install`` via API. Or pass a ``--conf`` item as an argument to the API


> Just a minor FYI.. The more editing of config files of various kinds there are the higher the entry bar is for new users.

Sure, but new users shouldn't be changing the default config in most cases. When it becomes necessary (and they are no longer new users, but more experienced), it is much better to manage the configuration in code, in a repo that can be used for ``conan config install``, because that would also provide for a ton of other configuration items that are extremely recommended: your remotes, your profiles, your custom settings, your hooks...

memsharded (approved):

PR #9711, opened 2021-10-02
Feature: MesonToolchain: improve cross-compilation workflow on Apple systems

### LINKED ISSUE #9710: [feature] MesonToolchain: improve cross-compilation workflow on Apple systems
unlike CMake, meson is dumb and doesn't attempt to guess a path to the compiler in case of cross-compiling mode.
it has to be explicitly specified within cross-file

we probably can deduce required things in the conan client, but it's kinda auto-magic...
OTH it makes recipes easier

we may try to populate the contents of the cross-file similar to the examples meson provides:
https://github.com/mesonbuild/meson/blob/master/cross/iphone.txt
https://github.com/mesonbuild/meson/blob/master/cross/tvos.txt

see also:
https://github.com/conan-io/conan-center-index/pull/7360
https://github.com/conan-io/conan-center-index/pull/6678

/cc @ericLemanissier @madebr @SpaceIm

SSE4:
closes: #9710

Changelog: Feature: MesonToolchain: improve cross-compilation workflow on Apple systems
Docs: omit

- adds a mixin `MesonAppleMixin` to add some important flags specific to the Apple systems 
- adds missing `obj[cpp]_[link]_args` to the cross-file (compiler/linker flags for Objective-C and Objective-C++ languages)
- adds missing `root` to the cross-file (sysroot)
- new test for M1 cross-building

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -3,6 +3,7 @@ import textwrap
 from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
+from conan.tools.apple.apple import is_apple_os, to_apple_arch, apple_min_version_flag, XCRun
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.microsoft import VCVars
 from conans.client.build.cppstd_flags import cppstd_from_settings
@@ -10,7 +11,7 @@ from conan.tools.cross_building import cross_building, get_cross_building_settin
 from conans.util.files import save
 
 
-class MesonToolchain(object):
+class MesonBaseToolchain(object):
     native_filename = "conan_meson_native.ini"
     cross_filename = "conan_meson_cross.ini"
 
@@ -46,7 +47,14 @@ class MesonToolchain(object):
     c_link_args = {{c_link_args}}
     cpp_args = {{cpp_args}} + preprocessor_definitions
     cpp_link_args = {{cpp_link_args}}
+    objc_args = {{objc_args}} + preprocessor_definitions
+    objc_link_args = {{objc_link_args}}
+    objcpp_args = {{objcpp_args}} + preprocessor_definitions
+    objcpp_link_args = {{objcpp_link_args}}
     {% if pkg_config_path %}pkg_config_path = {{pkg_config_path}}{% endif %}
+
+    [properties]
+    {% if root %}root = {{root}}{% endif %}
     """)
 
     _cross_file_template = _native_file_template + textwrap.dedent("""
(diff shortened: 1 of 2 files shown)


conan/tools/meson/toolchain.py
```diff
@@ -314,3 +342,45 @@ def generate(self):
...
+        self.c_link_args.extend(args)
+        self.cpp_args.extend(args)
+        self.cpp_link_args.extend(args)
+        self.objc_args.extend(args)
+        self.objc_link_args.extend(args)
+        self.objcpp_args.extend(args)
+        self.objcpp_link_args.extend(args)
+
+
+class MesonToolchain(MesonBaseToolchain, MesonAppleMixin):
```
memsharded, line 403: Please, no multiple inheritance. And avoid inheritance as much as possible. Go with composition unless there is an extremely compelling case.

conan/tools/meson/toolchain.py
```diff
@@ -314,3 +342,45 @@ def generate(self):
...
+class MesonAppleMixin(object):
+    def __init__(self, conanfile):
+        if not cross_building(conanfile):
+            return
+        if self._base_compiler != 'apple-clang':
+            return
+        if not is_apple_os(conanfile.settings.get_safe("os")):
+            return
+
+        xcrun = XCRun(conanfile.settings)
```
memsharded, line 375: Better move XCRun first to modern approach. It should be less detection based and more configuration based.

SSE4: the XCRun is already imported from a new namespace `conan.tools.apple` or what exactly do you mean?

conan/tools/meson/toolchain.py
SSE4: I think it's a well-known mixin pattern?
/cc @jgsogo 
suggestion on how to to it better are welcomed

SSE4: @memsharded do you want me to implement `AppleSystemBlock` like it was done in `CMakeToolchain`?
https://github.com/conan-io/conan/blob/develop/conan/tools/cmake/toolchain.py#L280

SSE4:
I would love to see this moving forward
the bad things are:
- old meson build helper doesn't support cross-building at all
- new meson build helper doesn't work for M1 out of the box

we have a plenty of foundational recipes using `meson` as a build system, e.g.:
- GLib
- Cairo
- Pango
- GStreamer and its plug-ins

wbehrens-on-gh:
@SSE4 Does the new meson build helper only fail to cross compile to M1 or does it fail to cross build in all cases?

SSE4:
> @SSE4 Does the new meson build helper only fail to cross compile to M1 or does it fail to cross build in all cases?

no, not just Apple. see issue #9922 for the similar incident with Visual Studio cross-compilation (x86_64 -> x86).
in general, Meson, unlike CMake, cannot deduce a compiler executable in case of cross-compiling.

conan/tools/meson/toolchain.py
SSE4: anyway, I think improvements of XCRun, if any, could be done in another, their own PR. this is outside of the scope of this PR IMO.

conan/tools/meson/toolchain.py
memsharded: Yes, please, avoid inheritance and mixins unless it is strictly necessary. Composition scales better and is more flexible for user customization. Going with a "blocks" approach like the ``CMakeToolchain`` will be better, please go that way.

conan/tools/meson/toolchain.py
memsharded: The idea is that generators like this one do not auto-detect anything. XCRun is deducing things by executing some commands right? That should be moved to input configuration, like the new ``[conf]``. 

This is a general effort we are doing in 2.0, we are removing all the platform, architecture, compiler, env-vars detection in generators and helpers, and that should happen in the input configuration. Please try that approach better.

SSE4: I don't know right now how it should look like and how it should be implemented.
if you have some example, I can take a look, but my point it belongs to XCRun, not to the meson toolchain.
in other words, as I understand, XCRun will have some code to read `conf` file instead of executing commands.
I don't see any reason why it should be done inside this particular PR - it is just using XCRun, and for meson toolchain it's an implementation detail how exactly it is implemented under the hood.
I also don't understand how is it going to work in CI, e.g. in case of ConanCenter, e.g. who, when and how exactly should populate conf with the input configutation?
really, I am just relying on the existing implementations of `conan.tools` namespace here, if they are somehow bbad/buggy or don't meet the conan 2.0 requirements, let's create an another issue and improve in another PR.

conan/tools/meson/toolchain.py
SSE4: I am looking at how to implement blocks similarly to how it was done in CMake toolchain.
there are several ideas, and I think I'll need some advice.
first off, there are two main things Apple block should define:
- compiler executables, `c` (known as `CC` in autotools world) and `cpp` (known as `CXX`)
- flags for compilers and linkers, like `c_args` (`CFLAGS`), `cpp_args` (`CXXFLAGS`), `link_args` (`LDFLAGS`), etc.

for executables, it should take the very least priority (only use when not explicitly defined), for variables, it should effectively append them to any existing.

then, unlike `CMake`, meson machine files are simple .ini files with very limited syntax. they are not turing-complete, and cannot contain conditions, loop, and other flow control instructions. there are very few operators supported, like `+` and `/` (https://mesonbuild.com/Machine-files.html#constants), but that's it.

on the other hand, there is an interesting feature that meson supports multiple cross-files overriding each other: https://mesonbuild.com/Machine-files.html#loading-multiple-machine-files

so, we may use something like `--cross-file conan_meson_cross_apple.ini --cross-file conan_meson_cross_main.ini`.

therefore, there are likely 3 major approaches on how to implement that:
1. use multiple cross-files overriding each other
2. use a single cross-file with several blocks
3. use a single cross-file with one block, but compose several contexts within the toolchain

1) is easy to implement, but it might be kinda misleading: too many files to deal with for the consumer, e.g. too long command lines, and easy to mess up with an order
2) might be tricky with a limited .ini syntax. e.g. I can't define the same value twice within the same machine file, I will get an error:
```
configparser.DuplicateOptionError
option 'buildtype' in section 'built-in options' already exists
```
and I can't check if it's already defined above, without using an external check in the code.

3) is also easy to implement, it will be just a matter of creating several context dictionaries and merging them according to our specific rules (executables override, flags append). but it's not really a blocks approach.

looking for the advice, thanks!

franramirez688:
In the end, it was superseded by https://github.com/conan-io/conan/pull/10174

Issue #9731, opened 2021-10-05
[bug] conan update removed custom cacert.pem


fschoenm:
### Environment Details (include every applicable attribute)
  * Operating System+version: Windows 10
  * Compiler+version:
  * Conan version: 1.40.3
  * Python version:

### Steps to reproduce (Include if Applicable)

I updated conan with `pip install -U conan`. In the next run, conan removed the cacert.pem (and probably replaced it by some kind of default file even though that wasn't what the output said):

	WARN: Migration: Updating settings.yml
	WARN: ****************************************
	WARN: settings.yml is locally modified, can't be updated
	WARN: The new settings.yml has been stored in: C:\Users\fschoenm\.conan\settings.yml.new
	WARN: ****************************************
	Removing the 'cacert.pem' file...
	Conan version 1.40.3

We're using a cacert.pem file that is customized for our self-signed server certificate and get's installed by ``conan config install``.

The update broke the access to our remote.

Probably conan should handle cacert.pem like the settings.yml and not overwrite it (see above) or add a way to add additional certificates in ``conan config install``.

czoido:
Hi @fschoenm,
Sorry for that, you are right, this should be handled with some similar mechanism from that of the `settings.yml`
We are fixing this in a patch version: _1.40.4_ that we are releasing very soon.
Thanks a lot for reporting.

czoido:
Closed by https://github.com/conan-io/conan/pull/9734 merge in Conan 1.40.4 just released

PR #9734, opened 2021-10-05
Check old cacert.pem before deleting when updating Conan

czoido:
Changelog: Fix: Check current _cacert.pem_ file when updating Conan and only migrate if the user has not modified the file. If  the local file is modified then create a new cacert file but don't overwrite current.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/9731

### DIFF
--- a/conans/client/migrations.py
+++ b/conans/client/migrations.py
@@ -1,6 +1,8 @@
 import os
 import shutil
 
+import six
+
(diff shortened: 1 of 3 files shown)


memsharded (approved):
Tests are broken in 2.7 because of unicode conversion and comparison apparently. But otherwise looks good.

PR #9744, opened 2021-10-06
Added IntelCC as generator

franramirez688:
Changelog: IntelCC put as new possible generator
Docs: omit

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -203,7 +203,7 @@ class Environment:
             for %%v in ({vars}) do (
                 set foundenvvar=
                 for /f "delims== tokens=1,2" %%a in ('set') do (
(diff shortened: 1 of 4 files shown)


memsharded (commented):
I am not sure this is a use case.
The general idea is that compiling with intel-cc could be done transparently, as with other compilers, without needing to add extra stuff to the recipe beyond the build system (like CMakeToolchain) stuff, that is general to all compilers.

lasote:
Then why document it? At first sight, I see it as very similar to VCVars. If you are not using `CMakeToolchain` or `MSBuildToolchain` you might need it to set up the compiler. Anyway, in my opinion, we can document it and add it as a regular generator or not document it at all, but if it is "public" it should be complete.

memsharded (approved):

Issue #5295, opened 2019-06-05
Add an environment variable pointing to the .conan folder


MathPlayer:
I know about the `CONAN_USER_HOME` variable pointing to a folder in which conan will create the `.conan` folder. This allows for multiple `.conan` folders, especially useful for CI systems where one might want to build everything from scratch, maybe even based on a different settings repository.

Is there a reason against having a `CONAN_HOME` variable to point to the exact `.conan` folder? This seems pretty similar with the idea behind `CONAN_USER_HOME`, but with a better directory structure on disk.

I am using multiple conan config settings which are installed in different folders; I am switching between them by adjusting the `CONAN_USER_HOME` variable. The directory structure I currently have looks like this:
```
~
├── .conan_user_home_1
│   └── .conan
│       ├── remotes.json
│       └── settings.yml
│       ...
└── .conan_user_home_2
    └── .conan
        ├── remotes.json
        └── settings.yml
        ...
```
After having `CONAN_HOME` I can switch to something like this:
```
~
├── .conan_1
│   ├── remotes.json
│   └── settings.yml
│   ...
└── .conan_2
    ├── remotes.json
    └── settings.yml
    ...
```

danimtb:
Hi @MathPlayer,

yes, I think this totally makes sense and in some cases the behavior of CONAN_USER_HOME is a bit strange. At least it was confusing for me at first.

I don't know if there is a reason behind this CONAN_USER_HOME instead of just a CONAN_HOME, so I will ask the team for feedback.

Thanks!

danimtb:
Hi @mathieu,

I checked this with the team and we think the current CONAN_USER_HOME covers the use case you propose and the CONAN_HOME could be confusing, as usually, this points to the folder that contains the executable of the program (for example JAVA_HOME, https://confluence.atlassian.com/doc/setting-the-java_home-variable-in-windows-8895.html).

So finally, I agree, a direct variable to the CONAN_CACHE_FOLDER, or something like this may be useful, but it is already covered by CONAN_USER_HOME. We could consider this change for Conan 2.0 maybe

memsharded:
Actually I did an attempt to remove the extra ``.conan`` folder, but it was a bit breaking. The internal code is prepared, it might be possible to make it the default in Conan 2.0.

unzap:
@memsharded is there any rough idea when the 2.0 should be released?

danimtb:
[...]

PR #9750, opened 2021-10-06
fix CONAN_USER_HOME/.conan env-var

memsharded:
Changelog: Fix: Avoid creating always an extra ``.conan`` subfolder in the cache, if the usre defines CONAN_USER_HOME, that can be directly be the cache folder. Use it only for the default ``<userhome>/.conan``
Docs: Omit

Close https://github.com/conan-io/conan/issues/5295

### DIFF
--- a/conans/client/cache/cache.py
+++ b/conans/client/cache/cache.py
@@ -48,7 +48,7 @@ class ClientCache(object):
         self._new_config = None
         self.editable_packages = EditablePackages(self.cache_folder)
         # paths
-        self._store_folder = self.config.storage_path or os.path.join(self.cache_folder, "data")
+        self._store_folder = os.path.join(self.cache_folder, "p")
 
         mkdir(self._store_folder)
         db_filename = os.path.join(self._store_folder, 'cache.sqlite3')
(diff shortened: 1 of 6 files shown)


conans/client/cache/cache.py
```diff
@@ -48,7 +48,7 @@ def __init__(self, cache_folder):
         self._new_config = None
         self.editable_packages = EditablePackages(self.cache_folder)
         # paths
-        self._store_folder = self.config.storage_path or os.path.join(self.cache_folder, "data")
+        self._store_folder = os.path.join(self.cache_folder, "p")
```
madebr, line 51: Is this intentional?
```suggestion
        self._store_folder = os.path.join(self.cache_folder, "data")
```

memsharded: Yes. New cache will not have any ``short_paths``, so better keep all folders to the minimum. The inner folder are already shortened to "e", "s", and the like.

madebr: Thanks!

lasote (approved):

PR #9773, opened 2021-10-11
[2.0] [Refactor] Remove check_updates

lasote:
According to some TODO comments, it is possible that the argument `check_update` is not used anymore and the `update` is enough. Opened draft to see the CI opinion.

Changelog: omit
Docs: omit


memsharded (commented):
``check_update`` is actually used in ``conan info`` command, to know where there will be updates for recipes and packages, without actually retrieving them. A different thing is if this is a valuable use case that we want to maintain or not.

lasote:
we could say then that the tests are not complete or maybe not checking that the recipe hasn't been downloaded?

memsharded:
> we could say then that the tests are not complete or maybe not checking that the recipe hasn't been downloaded?

Yes, it is likely that some tests have been disabled or removed during all the previous changes in develop2.

lasote:
Tests and check_updates functionality for conan info recovered at: #9838

PR #9776, opened 2021-10-12
Feature/get safe options test package

### LINKED ISSUE #7548: [bug] self.options object in test_package recipe behaves different then object in main recipe
The `self.options` object inside a main recipe has `.get_safe` attribute to get the value of a option or a default value if it does not exist.
But in a test package conanfile.py, this does not exist.

I believe options object in the main and test recipe should have the same options.

  * Operating System+version: Linux+Windows
  * Conan version: 1.28.1
  * Python version: python 3.7.7

Create `rec1` and `rec2` recipe + test of rec2 using:

```sh
cat >rec1.py <<EOF
from conans import ConanFile
class Rec1Conan(ConanFile):
    name = "rec1"
    version = "1.0"
    options = {
        "opt": [True, False],
    }
    default_options = {
        "opt": True,
    }
EOF

cat >rec2.py <<EOF
from conans import ConanFile
class Rec2Conan(ConanFile):
    name = "rec2"
    version = "1.0"
    requires = "rec1/1.0"
EOF

mkdir tp_rec2

cat >tp_rec2/conanfile.py <<EOF
from conans import ConanFile
class TestRec2Conan(ConanFile):
    def build(self):
        self.output.info("rec1.opt={}".format(self.options["rec1"].get_safe("opt")))
EOF
```

Run the following commands:
```sh
conan remove -f rec1
conan remove -f rec2
conan create rec1.py rec1/1.0@ -tf None
conan create rec2.py rec2/1.0@ -tf tp_rec2
```

```sh
rec2/1.0 (test package): Calling build()
  File "/home/maarten/.local/lib/python3.7/site-packages/conans/errors.py", line 34, in conanfile_exception_formatter
    yield
  File "/home/maarten/.local/lib/python3.7/site-packages/conans/client/conanfile/build.py", line 16, in run_build_method
    conanfile.build()
  File "/tmp/poc2/tp_rec2/conanfile.py", line 6, in build
    self.output.info("rec1.opt={}".format(self.options["rec1"].get_safe("opt")))
  File "/home/maarten/.local/lib/python3.7/site-packages/conans/model/options.py", line 75, in __getattr__
    raise ConanException(option_not_exist_msg(attr, list(self._dict.keys())))
conans.errors.ConanException: option 'get_safe' doesn't exist
Possible options are ['opt']
```

gfidder:
Changelog: Bugfix: Adding `get_safe` to `self.options` object
Docs: skip

closes #7548

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,10 @@ conan.conf
 #VScode folder
 .vscode
 
+#Python env
+env/
+test/
+
 #Generated certificate file
 cacert.pem
 
--- a/conans/model/options.py
+++ b/conans/model/options.py
@@ -156,6 +156,9 @@ class PackageOptionValues(object):
     def serialize(self):
         return self.items()
 
+    def get_safe(self, field, default=None):
+        return self._dict.get(field, default)
+
     @property
     def sha(self):
         result = []
--- a/conans/test/unittests/model/options_test.py
+++ b/conans/test/unittests/model/options_test.py
@@ -240,6 +240,24 @@ Poco:deps_bundled=True""")
         self.assertEqual("None", self.sut.get_safe("path", True))
         self.assertEqual(True, self.sut.get_safe("unknown", True))
 
+    def test_get_safe_package_options(self):
+        boost_values = PackageOptionValues()
+        boost_values.add_option("static", True)
+        boost_values.add_option("optimized", 3)
+        boost_values.add_option("path", "NOTDEF")
+
+        self.assertEqual(True, boost_values.get_safe("static"))
+        self.assertEqual(3, boost_values.get_safe("optimized"))
+        self.assertEqual("NOTDEF", boost_values.get_safe("path"))
+        self.assertEqual(None, boost_values.get_safe("unknown"))
+        boost_values.static = False
+        boost_values.path = "None"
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -87,6 +87,10 @@ conan.conf
 #VScode folder
 .vscode
 
+#Python env
+env/
+test/
```
memsharded, line 92: gitignoring the env/ test/ folder is a bit dangerous, the project could have a "test/" or an env/ folder in the root, right? These are very normal and regular folder names in a project.

memsharded (commented):
There is an intrinsic problem with access of dependencies information with:

```
self.options["rec1"].get_safe("opt")
```
These options are **not** necessarily the actual values of the dependencies options. Because that only get a value when it is defined from the downstream (profile, or other recipe), of the current recipe. But it will not give you the real value of the dependencies options in every case.

We have changed the interface and model how to access dependencies information. It will always be done with ``self.dependencies["mydep"].options``, and it will give you the real upstream value, always. This ``options`` object, already has the ``get_safe()`` method.

Then, I think we could close this PR, as this is no longer recommended practice and there is a new, recomended way to do it from now, and will be the only one in Conan 2.0.

Thanks very much for submitting this contribution to Conan anyway!

PR #9800, opened 2021-10-14
rename ConanFileInterface.new_cpp_info

memsharded:
Changelog: Fix: Rename ``ConanFileInterface.new_cpp_info`` to ``ConanFileInterface.cpp_info`` (being it a ``NewCppInfo`` object). The `ConanFileInterface` object is only used when accessing `self.dependencies` in a conanfile, so we can make sure that the new develop generators use the final `cpp_info` name.
Docs: https://github.com/conan-io/docs/pull/2280

#tags: slow

### DIFF
--- a/conan/tools/cmake/utils.py
+++ b/conan/tools/cmake/utils.py
@@ -13,7 +13,5 @@ def get_file_name(conanfile, find_module_mode=False):
         if ret:
             return ret
     ret = conanfile.cpp_info.get_property("cmake_file_name", "CMakeDeps")
-    if not ret:
-        ret = conanfile.cpp_info.get_filename("cmake_find_package_multi", default_name=False)
     return ret or conanfile.ref.name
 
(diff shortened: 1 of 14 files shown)


conan/tools/cmake/utils.py
```diff
@@ -9,11 +9,9 @@ def get_file_name(conanfile, find_module_mode=False):
     # This is used by the CMakeToolchain to adjust the XXX_DIR variables and the CMakeDeps. Both
     # to know the file name that will have the XXX-config.cmake files.
     if find_module_mode:
-        ret = conanfile.new_cpp_info.get_property("cmake_module_file_name", "CMakeDeps")
+        ret = conanfile.cpp_info.get_property("cmake_module_file_name", "CMakeDeps")
         if ret:
             return ret
-    ret = conanfile.new_cpp_info.get_property("cmake_file_name", "CMakeDeps")
-    if not ret:
```
memsharded, line 16: This is important, if we want to follow this approach, we cannot keep the compatibility with the legacy ``filename`` and ``names`` approaches, must use properties. I am fine with this.

lasote (approved):

czoido (approved):

PR #9801, opened 2021-10-14
Fix unsafe check of the compiler in the CMakeToolchain generator

jwillikers:
Changelog: Bugfix: Avoid crash in ``CMakeToolchain`` for custom generator when compiler is not defined.
Docs: Omit

The compiler variable here must not be a NoneType object to use `in`.
This will result in an error when attempting to package a header-only library.
This commit fixes this check.
It first verifies the compiler is not None before checking it.


memsharded (approved):

memsharded (commented):
Thanks for fixing this!
In general it would require a test, unit, integration, something to cover the bug and fix. But in this case as this crash is pretty obvious, we can move it foward.

PR #9806, opened 2021-10-15
Feature/refactor pkgconfigdeps

### LINKED ISSUE #9748: [feature] Refactor new `conan.tools.gnu.pkgconfigdeps.py` module
Modernize the `PkgConfigDeps` class as it's done in other `conan.tools` modules:

* Use templates to generate the content.
* Simplify the content generation.

franramirez688:
Changelog: Feature: Refactor `conan.tools.gnu.pkgconfigdeps` module.
Docs: omit

Closes:  https://github.com/conan-io/conan/issues/9748

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -26,16 +26,16 @@ from conans.errors import ConanException
 from conans.util.files import save
 
 
-def get_package_namespace(req):
+def get_package_name(req):
     ret = req.cpp_info.get_property("pkg_config_name", "PkgConfigDeps")
     return ret or req.ref.name
 
 
-def get_component_alias(req, comp_name):
+def get_component_name(req, comp_name):
     if comp_name not in req.cpp_info.components:
         # foo::foo might be referencing the root cppinfo
         if req.ref.name == comp_name:
-            return get_package_namespace(req)
+            return get_package_name(req)
         raise ConanException("Component '{name}::{cname}' not found in '{name}' "
                              "package requirement".format(name=req.ref.name, cname=comp_name))
     ret = req.cpp_info.components[comp_name].get_property("pkg_config_name", "PkgConfigDeps")
@@ -48,11 +48,11 @@ class PkgConfigDeps(object):
         self._conanfile = conanfile
 
     @staticmethod
-    def _get_composed_pkg_config_name(pkg_name, comp_name):
+    def _get_pc_name(pkg_name, comp_name):
         """Build a composed name for all the components and its package root name"""
         return "%s-%s" % (pkg_name, comp_name)
 
-    def _get_requires_names(self, dep, cpp_info=None):
+    def _get_requires_names(self, name, cpp_info):
         """
         Get all the pkg-config valid names from the requires ones given a dependency and
         a CppInfo object.
@@ -60,38 +60,38 @@ class PkgConfigDeps(object):
         Note: CppInfo could be coming from one Component object instead of the dependency
         """
         ret = []
-        cpp_info = cpp_info or dep.cpp_info
         for req in cpp_info.requires:
-            # FIXME: this str(dep.ref.name) is only needed for python2.7 (unicode values).
-            #        Remove it for Conan 2.0
-            pkg_name, comp_name = req.split("::") if "::" in req else (str(dep.ref.name), req)
+            pkg_name, comp_name = req.split("::") if "::" in req else (name, req)
             # FIXME: it could allow defining requires to not direct dependencies
             req_conanfile = self._conanfile.dependencies.host[pkg_name]
-            comp_alias_name = get_component_alias(req_conanfile, comp_name)
-            ret.append(self._get_composed_pkg_config_name(pkg_name, comp_alias_name))
+            comp_alias_name = get_component_name(req_conanfile, comp_name)
+            ret.append(self._get_pc_name(pkg_name, comp_alias_name))
         return ret
 
-    def get_components_content(self, dep):
+    def get_components_files_and_content(self, dep):
         """Get all the *.pc files content for the dependency and each of its components"""
         pc_files = {}
-        pkg_name = get_package_namespace(dep)
+        pkg_name = get_package_name(dep)
         comp_names = []
         pc_gen = _PCFilesTemplate(self._conanfile, dep)
         # Loop through all the package's components
         for comp_name, comp_cpp_info in dep.cpp_info.get_sorted_components().items():
-            comp_name = get_component_alias(dep, comp_name)
+            comp_name = get_component_name(dep, comp_name)
             comp_names.append(comp_name)
-            comp_requires_names = self._get_requires_names(dep, comp_cpp_info)
+            # FIXME: this str(dep.ref.name) is only needed for python2.7 (unicode values).
+            #        Remove it for Conan 2.0
+            comp_requires_names = self._get_requires_names(str(dep.ref.name), comp_cpp_info)
             # Get the *.pc file content for each component
-            pkg_comp_name = self._get_composed_pkg_config_name(pkg_name, comp_name)
-            pc_files.update(pc_gen.get_long_pc_file_content(comp_requires_names, name=pkg_comp_name,
-                                                            cpp_info=comp_cpp_info))
+            pkg_comp_name = self._get_pc_name(pkg_name, comp_name)
+            pc_files.update(pc_gen.get_pc_filename_and_content(comp_requires_names,
+                                                               name=pkg_comp_name,
+                                                               cpp_info=comp_cpp_info))
         # After looping through all the package's components, we have to check if the package
         # was between these components, if not, we'll have to create a *.pc file with a short content
         # for this one
         if pkg_name not in comp_names:
-            pkg_requires = [self._get_composed_pkg_config_name(pkg_name, i) for i in comp_names]
-            pc_files.update(pc_gen.get_short_pc_file_content(pkg_requires))
+            pkg_requires = [self._get_pc_name(pkg_name, i) for i in comp_names]
+            pc_files.update(pc_gen.get_wrapper_pc_filename_and_content(pkg_requires))
         return pc_files
 
     @property
@@ -101,11 +101,13 @@ class PkgConfigDeps(object):
         host_req = self._conanfile.dependencies.host
         for require, dep in host_req.items():
             if dep.cpp_info.has_components:
-                pc_files.update(self.get_components_content(dep))
+                pc_files.update(self.get_components_files_and_content(dep))
             else:  # Content for package without components
                 pc_gen = _PCFilesTemplate(self._conanfile, dep)
-                requires = self._get_requires_names(dep)
-                pc_files.update(pc_gen.get_long_pc_file_content(requires))
+                # FIXME: this str(dep.ref.name) is only needed for python2.7 (unicode values).
+                #        Remove it for Conan 2.0
+                requires = self._get_requires_names(str(dep.ref.name), dep.cpp_info)
+                pc_files.update(pc_gen.get_pc_filename_and_content(requires))
         return pc_files
 
     def generate(self):
@@ -121,9 +123,9 @@ class _PCFilesTemplate(object):
     def __init__(self, conanfile, dep):
         self._conanfile = conanfile
         self._dep = dep
-        self._name = get_package_namespace(dep)
+        self._name = get_package_name(dep)
 
-    long_content_template = textwrap.dedent("""\
+    pc_file_template = textwrap.dedent("""\
     prefix={{prefix_path}}
     {% for name, path in libdirs %}
     {{ name + "=" + path }}
@@ -146,7 +148,7 @@ class _PCFilesTemplate(object):
     {% endif %}
     """)
 
-    short_content_template = textwrap.dedent("""\
+    wrapper_pc_file_template = textwrap.dedent("""\
     Name: {{name}}
     Description: {{description}}
     Version: {{version}}
@@ -155,7 +157,7 @@ class _PCFilesTemplate(object):
     {% endif %}
     """)
 
-    def get_long_pc_file_content(self, requires, name=None, cpp_info=None):
+    def get_pc_filename_and_content(self, requires, name=None, cpp_info=None):
 
         def _concat_if_not_empty(groups):
             return " ".join(
@@ -217,11 +219,11 @@ class _PCFilesTemplate(object):
             "cflags": get_cflags(includedirs),
             "requires": requires
         }
-        template = Template(self.long_content_template, trim_blocks=True, lstrip_blocks=True,
+        template = Template(self.pc_file_template, trim_blocks=True, lstrip_blocks=True,
                             undefined=StrictUndefined)
         return {dep_name + ".pc": template.render(context)}
 
-    def get_short_pc_file_content(self, requires, name=None):
+    def get_wrapper_pc_filename_and_content(self, requires, name=None):
         dep_name = name or self._name
         context = {
             "name": dep_name,
@@ -229,6 +231,6 @@ class _PCFilesTemplate(object):
             "version": self._dep.ref.version,
             "requires": requires
         }
-        template = Template(self.short_content_template, trim_blocks=True, lstrip_blocks=True,
+        template = Template(self.wrapper_pc_file_template, trim_blocks=True, lstrip_blocks=True,
                             undefined=StrictUndefined)
         return {dep_name + ".pc": template.render(context)}
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -110,27 +110,24 @@ class CompletePCFile(object):
     def template():
         return textwrap.dedent("""\
         prefix={{prefix_path}}
-        {%- for name, path in libdirs.items() %}
-            {{ name + "=" + path }}
+        {% for name, path in libdirs.items() %}
+        {{ name + "=" + path }}
         {% endfor %}
-        {%- for name, path in includedirs.items() %}
-            {{ name + "=" + path }}
+        {% for name, path in includedirs.items() %}
+        {{ name + "=" + path }}
         {% endfor %}
         {% if pkg_config_custom_content %}
         # Custom PC content
         {{pkg_config_custom_content}}
         {% endif %}
 
-        Name: {{dep_name}}
+        Name: {{name}}
         Description: {{description}}
-        Version: {{dep_version}}
+        Version: {{version}}
         Libs: {{ libs }}
-        Cflags: {{ flags }}
-        {% if public_deps %}
-        Requires: {% for dep in public_deps %}
-                {{ dep }}
-            {%- if not loop.last %},{% endif %}
-            {% endfor %}
+        Cflags: {{ cflags }}
+        {% if requires|length %}
+        Requires: {% for dep in requires %}{{ dep }}{%- if not loop.last %},{% endif %}{% endfor %}
         {% endif %}
         """)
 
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -17,16 +17,15 @@
...
 from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
 from conans.errors import ConanException
 from conans.util.files import save
 
 
-def _concat_if_not_empty(groups):
-    return " ".join([param for group in groups for param in group if param and param.strip()])
-
-
 def get_target_namespace(req):
```
tapia: PkgConfig doesn't have targets, does it? Maybe this should be renamed to `get_package_namespace`?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -47,158 +46,191 @@ class PkgConfigDeps(object):
...
-            prefix = ""
-            if not os.path.isabs(directory):
-                prefix = "${prefix}/"
-            elif directory.startswith(prefix_path):
-                prefix = "${prefix}/"
-                directory = os.path.relpath(directory, prefix_path).replace("\\", "/")
-            lines.append("%s=%s%s" % (name, prefix, directory))
-            varnames.append(name)
-        return lines, varnames
+                requires = self._get_requires_names(dep, dep.cpp_info)
```
memsharded: Just pass one ``dep`` argument, as the function can get inside the ``dep.cpp_info``

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -47,158 +46,191 @@ class PkgConfigDeps(object):
...
+            "libs": get_libs(libdirs),
+            "cflags": get_cflags(includedirs),
+            "requires": requires
+        }
+        template = Template(_CompletePCFile.template(), trim_blocks=True, lstrip_blocks=True,
+                            undefined=StrictUndefined)
+        return {dep_name + ".pc": template.render(context)}
+
+
+class _SimplePCFile(object):
```
memsharded: These classes with everything ``staticmethod`` sounds a bit OO abuse.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -47,158 +46,191 @@ class PkgConfigDeps(object):
...
+                [['-I"${%s}"' % n for n, _ in includedirs_],
+                 cpp_info.cxxflags,
+                 cpp_info.cflags,
+                 ["-D%s" % d for d in cpp_info.defines]])
+
+        def get_formmatted_dirs(field, folders, prefix_path_):
+            ret = []
+            for i, directory in enumerate(folders):
+                directory = os.path.normpath(directory).replace("\\", "/")
+                n = field if i == 0 else "%s%d" % (field, (i + 1))
```
memsharded: This sounds a bit too much logic in the data preparation that could be moved to the rendering jinja part. How the file is rendered based on the input data is part of the template.

conan/tools/gnu/pkgconfigdeps.py
franramirez688: Yes, but I'll have to pass it as `kwarg` because, at some point, I need to pass `dep` and `component.cpp_info`

memsharded: Ok, then you can leave it, I hadn't seen the case for components.

conan/tools/gnu/pkgconfigdeps.py
franramirez688: There are other internal functions that are using those names like `get_cflags()` and `get_libs()` so it could be a mess to manage those names by the templates.

conans/test/functional/toolchains/gnu/test_pkgconfigdeps.py
```diff
@@ -53,6 +53,7 @@ def package_info(self):
         Version: 0.1
         Libs: -L"${libdir}" -L"${libdir2}"%s
         Cflags: -I"${includedir}\"""" % expected_rpaths)
+
```
tapia, line 53: Doesn't all this refactor need a single test?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,183 @@ def __init__(self, conanfile):
...
+                                         gnudeps_flags.framework_paths])
+
+        def get_cflags(includedirs_):
+            return _concat_if_not_empty(
+                [['-I"${%s}"' % n for n, _ in includedirs_],
+                 cpp_info.cxxflags,
+                 cpp_info.cflags,
+                 ["-D%s" % d for d in cpp_info.defines]])
+
+        def get_formmatted_dirs(field, folders, prefix_path_):
```
tapia: Typo: `formatted`, with a single m :-)

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,183 @@ def __init__(self, conanfile):
...
+        # Loop through all the dependency components
+        for comp_name, comp_cpp_info in dep.cpp_info.get_sorted_components().items():
+            comp_name = get_component_alias(dep, comp_name)
+            comp_names.append(comp_name)
+            comp_requires_names = self._get_requires_names(dep, comp_cpp_info)
+            pkg_comp_name = self._get_composed_pkg_config_name(pkg_name, comp_name)
+            # Save the *.pc file for this component
+            pc_files.update(pc_gen.complete_pc_content(comp_requires_names, name=pkg_comp_name,
+                                                       cpp_info=comp_cpp_info))
+        # Adding the dependency *.pc file (including its components as requires if any)
```
tapia: This comment needs to be rewritten, I don't understand what the part between parenthesis means. Maybe something like "which will include the 'requires' statement for the files generated in the previous loop"?

Also, the `complete_pc` and `simple_pc` names are quite confusing, it looks like they are two versions of the same file. Please give them a more descriptive name. If I understand correctly what they really are, how about `component_pc` and `main_package_pc` or something like that?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,183 @@ def __init__(self, conanfile):
...
-                        dep.package_folder, dep.ref.version)
-                # Adding the pkg *.pc file (including its components as requires if any)
-                comp_gennames = [comp_genname for comp_genname, _, _ in components]
-                if pkg_genname not in comp_gennames:
-                    pkg_requires = (self._get_composed_require_name(pkg_genname, i)
-                                    for i in comp_gennames)
-                    ret["%s.pc" % pkg_genname] = self._global_pc_file_contents(pkg_genname,
-                                                                               dep,
-                                                                               pkg_requires)
+                pc_files.update(self.get_components_pc_files(dep))
```
tapia: If `get_components_pc_files()` generates PC files, why don't you move it to the PCGenerator class, like `complete_pc_content()` and `simple_pc_content()`?

conans/test/functional/toolchains/gnu/test_pkgconfigdeps.py
franramirez688: Actually, this refactor is completely covered by the current tests battery.

conan/tools/gnu/pkgconfigdeps.py
franramirez688: I see your point but I don't think so. Perhaps the problem is the naming and nothing else. Let me rethink these method names because I believe this comment is related to the previous one. Let's discuss it after updating the PR 😉

tapia: Sure 🙂

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
 
-    def _get_require_comp_name(self, dep, req):
-        # FIXME: this str() is only needed for python2.7 (unicode values). Remove it for Conan 2.0
-        pkg_name = str(dep.ref.name)
-        pkg, comp_name = req.split("::") if "::" in req else (pkg_name, req)
-        # FIXME: it could allow defining requires to not direct dependencies
-        req = self._conanfile.dependencies.host[pkg]
-        cmp_name = get_component_alias(req, comp_name)
-        return self._get_composed_require_name(pkg, cmp_name)
+    def _get_requires_names(self, dep, cpp_info=None):
```
lasote: From `dep` you are using only the `dep.ref.name` and getting the `cppinfo` so it might be less confusing to have `_get_requires_names(self, name, cpp_info)` (cppinfo always required) ? 

And finally, `_get_requires_pc_names(self, name, cpp_info)` would be more consistent with my previous comment.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -17,17 +17,16 @@
...
 from conans.errors import ConanException
 from conans.util.files import save
 
 
-def _concat_if_not_empty(groups):
-    return " ".join([param for group in groups for param in group if param and param.strip()])
-
-
-def get_target_namespace(req):
+def get_package_namespace(req):
```
lasote: I wouldn't call it `namespace` either, it looks like it responds to some special property and the concept of namespace is not really a thing here in package config.
`get_package_name` should be enough.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -36,7 +35,7 @@ def get_component_alias(req, comp_name):
     if comp_name not in req.cpp_info.components:
```
lasote, line 35: `get_component_name` to be consistent with `get_package_name`

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
         self._conanfile = conanfile
 
     @staticmethod
-    def _get_composed_require_name(pkg_name, comp_name):
+    def _get_composed_pkg_config_name(pkg_name, comp_name):
```
lasote: `_get_pc_name` sounds friendly.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
-            if "::" in require:  # Points to a component of a different package
-                pkg, cmp_name = require.split("::")
-                req = dep.dependencies.direct_host[pkg]
-                public_comp_deps.append(
-                    self._get_composed_require_name(pkg, get_component_alias(req, cmp_name))
-                )
-            else:  # Points to a component of same package
-                public_comp_deps.append(get_component_alias(dep, require))
-        return public_comp_deps
+    def get_components_content(self, dep):
```
lasote:
```suggestion
    def get_components_files_and_contents(self, dep):
```

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
+            "version": version,
+            "libs": get_libs(libdirs),
+            "cflags": get_cflags(includedirs),
+            "requires": requires
+        }
+        template = Template(self.long_content_template, trim_blocks=True, lstrip_blocks=True,
+                            undefined=StrictUndefined)
+        return {dep_name + ".pc": template.render(context)}
+
+    def get_short_pc_file_content(self, requires, name=None):
```
lasote:
```suggestion
    def get_wrapper_pc_filename_and_content(self, requires, name=None):
```

conan/tools/gnu/pkgconfigdeps.py
lasote: > This sounds a bit too much logic in the data preparation that **could be moved to the rendering jinja part**.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
+            comp_name = get_component_alias(dep, comp_name)
+            comp_names.append(comp_name)
+            comp_requires_names = self._get_requires_names(dep, comp_cpp_info)
+            # Get the *.pc file content for each component
+            pkg_comp_name = self._get_composed_pkg_config_name(pkg_name, comp_name)
+            pc_files.update(pc_gen.get_long_pc_file_content(comp_requires_names, name=pkg_comp_name,
+                                                            cpp_info=comp_cpp_info))
+        # After looping through all the package's components, we have to check if the package
+        # was between these components, if not, we'll have to create a *.pc file with a short content
+        # for this one
```
lasote: Not sure about this logic or rule. 
So... we have a package `foo` declaring components `c1` and `c2`, and it will generate the following files:

```
foo-c1.pc
foo-c2.pc
foo.pc => Including `foo-c1.pc` and `foo-c2.pc`
```
Right?

But, if the author also declares a component `foo` we end with:

```
foo-c1.pc
foo-c2.pc
foo-foo.pc 
```

My question is, What is the problem of always generating a `foo.pc` file aggregating everything? I might be missing something...

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
+
+
+class _PCFilesTemplate(object):
+
+    def __init__(self, conanfile, dep):
+        self._conanfile = conanfile
+        self._dep = dep
+        self._name = get_package_namespace(dep)
+
+    long_content_template = textwrap.dedent("""\
```
lasote:
```suggestion
    pc_file_template = textwrap.dedent("""\
```

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
+    Description: {{description}}
+    Version: {{version}}
+    Libs: {{ libs }}
+    Cflags: {{ cflags }}
+    {% if requires|length %}
+    Requires: {% for dep in requires %}{{ dep }}{%- if not loop.last %} {% endif %}{% endfor %}
+    {% endif %}
+    """)
+
+    short_content_template = textwrap.dedent("""\
```
lasote:
```suggestion
    wrapper_pc_template = textwrap.dedent("""\
```

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,187 @@ def __init__(self, conanfile):
...
+    Name: {{name}}
+    Description: {{description}}
+    Version: {{version}}
+    {% if requires|length %}
+    Requires: {% for dep in requires %}{{ dep }}{%- if not loop.last %} {% endif %}{% endfor %}
+    {% endif %}
+    """)
+
+    def get_long_pc_file_content(self, requires, name=None, cpp_info=None):
```
lasote:
```suggestion
    def get_pc_filename_and_content(self, requires, name=None, cpp_info=None):
```

lasote (changes requested):

conan/tools/gnu/pkgconfigdeps.py
tapia: This is not really solved. You're still using names like `_get_composed_pkg_config_name()` or `get_long_pc_file_content()`, that don't give any clue about what they're actually doing.

conan/tools/gnu/pkgconfigdeps.py
tapia: If I'm not wrong, this is only used for component's files. I'd call it `component_pc_file_template`.

conan/tools/gnu/pkgconfigdeps.py
tapia: Same thing here. This is only used for the general pc file. How about `package_pc_file_template`?

conan/tools/gnu/pkgconfigdeps.py
tapia: This would become `def get_component_pc_filename_and_content()`

conan/tools/gnu/pkgconfigdeps.py
tapia: And this would become `def get_package_pc_filename_and_content()`

tapia (commented):
To summarize all my proposals:

`long_content_template` -> `component_pc_file_template`
`short_content_template` -> `package_pc_file_template`
`get_long_pc_file_content()` -> `get_component_pc_filename_and_content()`
`get_short_pc_file_content()` -> `get_package_pc_filename_and_content()`
`_get_composed_pkg_config_name()` -> `_get_pc_name()`

conan/tools/gnu/pkgconfigdeps.py
franramirez688: This is used for components and non-components flow

conan/tools/gnu/pkgconfigdeps.py
franramirez688: The other template is being used for the main package and its components so the `package_pc_file_template` is confusing

conan/tools/gnu/pkgconfigdeps.py
franramirez688: It's not only related to components. Have a look at https://github.com/conan-io/conan/pull/9806/files#diff-4fc089c803914bd943aa1fb943fad3f054561571230e8c71540704af96af5a6eR105

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -48,50 +48,50 @@ def __init__(self, conanfile):
...
-            pc_files.update(pc_gen.get_long_pc_file_content(comp_requires_names, name=pkg_comp_name,
-                                                            cpp_info=comp_cpp_info))
+            pkg_comp_name = self._get_pc_name(pkg_name, comp_name)
+            pc_files.update(pc_gen.get_pc_filename_and_content(comp_requires_names,
+                                                               name=pkg_comp_name,
+                                                               cpp_info=comp_cpp_info))
         # After looping through all the package's components, we have to check if the package
         # was between these components, if not, we'll have to create a *.pc file with a short content
         # for this one
         if pkg_name not in comp_names:
```
lasote: I think this still makes no sense. So if I declared a component "foo" and the package name is "foo" you don't generate a global `foo.pc`? What is the point? It would only make sense if the component is not scoped (foo_foo.pc => foo.pc) but it is always scoped, right?

franramirez688: I even think this is a bug right now because, if we check legacy `PkgConfig` it was always creating a `foo.pc` regardless it was declared as a component or not. Now, we are creating either `foo-foo.pc` in case of being as a component or `foo.pc` in the opposite case. Then I agree with you that we should create always a `foo.pc`.

memsharded: Agree.

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,197 @@ def __init__(self, conanfile):
...
+
+    def __init__(self, conanfile, dep):
+        self._conanfile = conanfile
+        self._dep = dep
+        self._name = get_package_name(dep)
+
+    pc_file_template = textwrap.dedent("""\
+
+    {%- macro get_libs(libdirs, cpp_info, gnudeps_flags) -%}
+    {%- for _ in libdirs -%}
```
memsharded, line 128: All the ``-`` in the template can be avoided using the jinja ``trim_blocks=True, lstrip_blocks=True,`` will take care of blanks, stripping them

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -49,156 +48,197 @@ def __init__(self, conanfile):
...
+    Description: {{ description }}
+    Version: {{ version }}
+    {% if requires|length %}
+    Requires: {{ requires|join(' ') }}
+    {% endif %}
+    """)
+
+    def get_pc_filename_and_content(self, requires, name=None, cpp_info=None):
+
+        def get_formatted_dirs(folders, prefix_path_):
```
memsharded, line 196: It is likely that in some situations we want pkg_config to run in a Windows subsystem, and then the ``subsystem_path()`` path formatting is necessary. It is ok as-is, but we might want to run subsystems testing for pkg_config too.

memsharded (approved):

conan/tools/gnu/pkgconfigdeps.py
franramirez688: What I'm doing here it's to avoid this effect:

```
-L"${libdir1}"
-L"${libdir2}"
-L"${libdir3}"
```
I want all the values inline. Perhaps, I don't need all the signs but I put them just in case.

czoido (approved):

PR #9811, opened 2021-10-15
[develop2] remove legacy vcvars and intel tools

memsharded:
Changelog: Remove: Getting rid of the legacy VS tools


conan/tools/microsoft/visual.py
```diff
@@ -1,9 +1,9 @@
 import os
 import textwrap
 
+from conans.client.conf.detect_vs import vs_installation_path
```
memsharded, line 4: I have a problem with this: the ``Visual`` integration, used by lots of ``conan.tools`` rely on the ``vs_installation_path``, which relies on dynamic detection, running ``vswhere()``, everytime it is invoked.

Our design principles would move this to a ``[conf]`` item, but this could be a bit of friction for onboarding, more complexity in defining or auto-detection of profiles, so I am not convinced to follow this path for this specific one.

Wdyt @lasote?

lasote: Yeah, as we talked maybe is too much to force declaring the path (Who knows where is it and what is the correct one?) and the detection didn't cause issues thus far.

memsharded: Yep, https://github.com/conan-io/conan/pull/9827 implements this already in Conan 1.42

franramirez688:
@memsharded 

I'm wondering why we are keeping the `intel` compiler in the `settings.yml` file and all the if-statements like `compiler == "intel"` if we're getting rid of this legacy Intel support. Perhaps, I'm missing something but the new `conan.tools.intel.intel_cc` is only managing the new `Intel oneAPI Toolkit` compilers and not the legacy one.

memsharded:
> 'm wondering why we are keeping the intel compiler in the settings.yml file and all the if-statements like compiler == "intel" if we're getting rid of this legacy Intel support. Perhaps, I'm missing something but the new conan.tools.intel.intel_cc is only managing the new Intel oneAPI Toolkit compilers and not the legacy one.


~~Yes, the legacy settings can be removed. Lets do it.~~
EDIT: actually, better wait to remove the "base.compatible" mechanism, this is a bit more generic than intel. I think this should go to the Tribe 2.0

franramirez688 (approved):

PR #9827, opened 2021-10-19
improving vcvars install_path and removed legacy intel support in VCVars

memsharded:
Changelog: Feature: Define new ``conf["tools.microsoft.msbuild:install_path"]`` for the MSBuild toolchain and remove generation of intel_vars files by ``VCVars``.
Docs: https://github.com/conan-io/docs/pull/2281


franramirez688 (approved):

lasote (approved):
Everything is fine but I wonder if the Visual Studio installation path (or build tools or whatever is valid there) is something obvious or not really. Would it be useful to write a real example path in the docs or not really?

PR #9830, opened 2021-10-19
Proposing a new RecipeReference model, to gradually migrate

memsharded:
Changelog: Omit
Docs: Omit

Some design ideas:

- Make it mutable, so it can be centralized and shared, instead of passed by value and copied
- Validation should probably be orthogonal, and applied only at certain critical points
- New model should provide an incremental migration, with conversion to/from the legacy model
- The new model should enable the new use syntax, like using ``pkg/version`` instead of ``pkg/version@``.
- New model should be clearly differentiating "requirements", like ``pkg/[>1.0]`` including version ranges, from actual references, which use fixed, defined versions.
- Integrate timestamps in the model
- Allow ``pkg/version@user`` definitions, without channel
- Try to define an integrated string format for timestamps ``pkg/version@user/channel#rev%timestamp``. Being timestamp an integer. Or should we use human readable dates for text serialization always?

To be processed after @lasote Python API proposals, this could be an object that maybe could be in the API interface.


conans/model/ref.py
```diff
@@ -146,6 +146,91 @@ def validate_revision(revision):
...
+        return result
+
+    def __lt__(self, ref):
+        # The timestamp goes before the revision for ordering revisions chronologically
+        return (self.name, self.version, self.user, self.channel, self.timestamp, self.revision) \
+               < (ref.name, ref.version, ref.user, ref.channel, ref.timestamp, ref.revision)
+
+    def __eq__(self, other):
+        # TODO: In case of equality, should it use the revision and timestamp?
+        raise Exception("WHO IS COMPARING RECIPE REFERENCES?")
```
lasote: LOL

conans/model/ref.py
```diff
@@ -146,6 +146,91 @@ def validate_revision(revision):
...
+        # return not self.__eq__(other)
+
+    def __hash__(self):
+        raise Exception("WHO IS COMPARING RECIPE REFERENCES?")
+        # return hash((self.name, self.version, self.user, self.channel, self.revision))
+
+    @staticmethod
+    def loads(text):  # TODO: change this default to validate only on end points
+        try:
+            # revision
```
lasote:
```suggestion
            # timestamp
```

lasote (commented):
We could start introducing it, I don't see a problem.

conans/model/ref.py
memsharded: Comparing by equality default comparison has bit me more times that I would like 😂

czoido:
Looks good

czoido (approved):

PR #9833, opened 2021-10-19
Fix pkgconfig tests for Windows

### LINKED ISSUE #9483: [bug] Meson with pkg-config in Windows
https://github.com/conan-io/conan/pull/9471 disabled these tests, because the new Meson integration with pkg-config doesn't work.

It seems that it never worked, apparently the tests were not running in Windows, because of Meson not correctly detected in CI.

franramirez688:
Changelog: Fix: Recovering `pkg-config` tests into Windows platform.
CI: Needs to be installed in Windows machine `pkg-config` tool through chocolatey: `choco install pkgconfiglite --version 0.28`
Closes: https://github.com/conan-io/conan/issues/9483
Docs: omit

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -84,12 +84,6 @@ tools_locations = {
         "exe": "mingw32-make",
         "system": {"path": {'Windows': "C:/msys64/mingw32/bin"}},
     },
-    'mingw32_g_plusplus': {
-        "platform": "Windows",
-        "default": "system",
-        "exe": "g++",
-        "system": {"path": {'Windows': "C:/msys64/mingw32/bin"}},
-    },
     'mingw64': {
         "platform": "Windows",
         "default": "system",
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -39,11 +39,15 @@ tools_locations = {
                       "15": {},
                       "16": {"disabled": True},
                       "17": {"disabled": True}},
-    'pkg_config': {"exe": "pkg-config",
-                   "default": "system",
-                   # pacman -S pkg-config inside msys2-mingw64
-                   "system": {"path": {'Windows': "C:/msys64/usr/bin"}},
-                   },
+    'pkg_config': {
+        "exe": "pkg-config",
+        "default": "0.28",
+        "0.28": {
+            "path": {
+                # Using chocolatey in Windows -> choco install pkgconfiglite --version 0.28
+                'Windows': "C:/ProgramData/chocolatey/lib/pkgconfiglite/tools/pkg-config-lite-0.28-1/bin"
+            }
+        }},
     'autotools': {"exe": "autoconf"},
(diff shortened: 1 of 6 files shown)


conans/test/conftest.py
```diff
@@ -80,6 +84,12 @@
         "exe": "mingw32-make",
         "system": {"path": {'Windows': "C:/msys64/mingw32/bin"}},
     },
+    'mingw32_g_plusplus': {
```
memsharded: This shouldn't be necessary. We already have mingw32 enviroments, why is this special?

conans/test/conftest.py
```diff
@@ -39,11 +39,15 @@
...
-                   # pacman -S pkg-config inside msys2-mingw64
-                   "system": {"path": {'Windows': "C:/msys64/usr/bin"}},
-                   },
+    'pkg_config': {
+        "exe": "pkg-config",
+        "default": "0.28",
+        "0.28": {
+            "path": {
+                # Using chocolatey in Windows -> choco install pkgconfiglite --version 0.28
+                'Windows': "C:/ProgramData/chocolatey/lib/pkgconfiglite/tools/pkg-config-lite-0.28-1/bin"
```
memsharded, line 48: Already installed in CI then?

memsharded (approved):
Looking good. I'd probably try to avoid the ``mingw32_g_plusplus`` specific tool.

conans/test/conftest.py
franramirez688: Then, would it be enough to put the `mingw32` tool or I don't need anything? I was afraid of that tool it's using only `mingw32-make`

memsharded: It should contain g++ too, it is a full mingw env, shouldn't be an issue.

conans/test/functional/build_helpers/pkg_config_test.py
```diff
@@ -121,7 +124,10 @@ def build(self):
                             'PKG_CONFIG_PATH': "%s" % self.deps_cpp_info["libB"].rootpath}
 
                     with tools.environment_append(vars):
-                       self.run('g++ main.cpp $(pkg-config %s libB --libs --cflags) -o main' % args)
+                        # FIXME: windows is not able to catch the output, "$()" does not exist in cmd
+                        self.run("pkg-config %s libB --libs --cflags >> output.txt" % args)
+                        with open("output.txt") as f:
+                            self.run('g++ main.cpp %s -o main' % f.readline().strip())
```
tapia, line 130: Probably not important, being this a test, but you're redirecting the `pkg-config` output to `output.txt` with `>>`, meaning that you're appending, and the previous content won't be deleted. But then you're only reading the first line of the file (`f.readline()`), so it will only get the correct line if `output.txt` was previously empty.

TLDR: the redirection should be `>`, not `>>`

tapia (approved):

PR #9838, opened 2021-10-20
2.0 remotes + users API and refactored remotes management 

lasote:
- Created "subapis" for `remotes`, `search` and `list`. Usage of "subapis":
  
    ```
    remotes = conan_api.remotes.list()
    ```
 
- Removed "conan user" command and introduced it as a subcommand of "conan remote":
    - conan remote login _remote_ _username_ [-p password]
    - conan remote logout _remote_
    - conan remote list-users
    - conan remote set-user _remote_ _username_

- Unified `-r` argument usage in these commands to always allow patterns (so -r "*" means all) 
- Centralized selection of the remotes from command (with -r) into the `conan_app`, also the `update`. So we stopped passing "remotes" and "remote" and "update" arguments through all Conan app classes.
- Design principles followed for the API:
     - The API is not capturing exceptions.
     - The API only returns basic types (dict, list, str) and objects at the `conans.cli.api.model` module. (Now only Remote)
     - The API methods only receive a single remote (except commands that perform graph computation etc like a conan install). If the command affect several remotes, the command should iterate.
     - The Command doesn't use "internals" of the application but only the conan_api


conans/client/cmd/download.py
```diff
@@ -1,18 +1,26 @@
...
 
-def download(app, ref, package_ids, remote, recipe, remotes):
+def download(app, ref, package_ids, recipe):
     remote_manager, cache, loader = app.remote_manager, app.cache, app.loader
     hook_manager = app.hook_manager
     assert(isinstance(ref, ConanFileReference))
     scoped_output = ScopedOutput(str(ref), ConanOutput())
+    remote = app.selected_remote
+    if not remote:
+        # FIXME: Probably this shouldn't be done, the "default" remote concept when no remote is
```
memsharded, line 16: Yes, I am fine raising here an error, ``download`` needs explicit ``-r`` remote

conans/client/conan_api.py
```diff
@@ -637,100 +630,26 @@ def export(self, path, name, version, user, channel, cwd=None,
...
                parallel_upload=False):
         """ Uploads a package recipe and the generated binary packages to a specified remote
         """
         app = ConanApp(self.cache_folder)
-        uploader = CmdUpload(app.cache, app.remote_manager, app.loader,
-                             app.hook_manager)
-        remotes = app.load_remotes(remote_name=remote_name)
-        uploader.upload(pattern, remotes, package, all_packages, confirm,
+        # FIXME: remote_name should be remote
+        app.load_remotes([Remote(remote_name, None)])
```
memsharded, line 644: This ``app.load_remotes([Remote(remote_name, None)])`` is confusing, yes.

memsharded (approved):
I like this, this is great.

It is pending the clarification about the previous ``--update`` and what we want for 2.0, but this PR can be merged.

conans/client/conan_api.py
lasote: Yup, I hate it but while migrating from the old conan_api to the new one that will be solved when the time comes.

czoido (approved):

franramirez688 (approved):

PR #9839, opened 2021-10-20
Bugfix: add test to illustate failure when options differ between requires and build requires

SSE4:
we have found conan client may fail if options are different between requirements and build requirements:
https://github.com/conan-io/conan-center-index/pull/7747#issuecomment-947524266

I think the cause of the issue is: https://github.com/conan-io/conan/blob/6d7642d4d3bd911892eb0cf91161148f22968b32/conans/client/graph/graph_binaries.py#L311

Changelog: omit
Docs: omit

### DIFF
--- a/conans/test/integration/build_requires/profile_build_requires_test.py
+++ b/conans/test/integration/build_requires/profile_build_requires_test.py
@@ -266,6 +266,29 @@ nonexistingpattern*: SomeTool/1.2@user/channel
         self.assertIn("MyLib/0.1@lasote/stable: Hello world from python tool!", client.out)
         self.assertNotIn("Project: Hello world from python tool!", client.out)
 
+    def test_build_requires_options_different(self):
+        client = TestClient()
+
+        conanfile_openssl_1_1_1 = GenConanfile("openssl", "1.1.1")
+        conanfile_openssl_3_0_0 = GenConanfile("openssl", "3.0.0")\
+            .with_option("no_fips", [True, False])\
+            .with_default_option("no_fips", True)
+        conanfile_cmake = GenConanfile("cmake", "0.1")\
+            .with_requires("openssl/1.1.1")
+        conanfile_consumer = GenConanfile("consumer", "0.1")\
+            .with_build_requires("cmake/0.1")\
+            .with_requires("openssl/3.0.0")
+
+        client.save({"conanfile_openssl_1_1_1.py": conanfile_openssl_1_1_1,
+                     "conanfile_openssl_3_0_0.py": conanfile_openssl_3_0_0,
+                     "conanfile_cmake.py": conanfile_cmake,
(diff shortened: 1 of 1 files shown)


SSE4:
checked `develop2` locally, it passes the same test, so no issue with v2

SSE4:
I was doing a `git bisect` and identified a commit fixing an issue in `develop2`:
https://github.com/conan-io/conan/commit/6a5bd19bb8ba034f0f83eda6b64c3f3707339705
the commit is quite big, and as it was squashed, so I can't say the exact line fixed it.
still, it identifies the only commit to be back-ported into `develop` to get the fix.

UPD: indeed, it removes the line doing `conanfile.build_requires_options = conanfile.options.values`.
@memsharded do you think this particular change can be isolated/extracted and back-ported into `develop1`?

memsharded:
I have checked this for a second time (sorry for the delay, I checked it some time ago, but didn't finalize report), it really looks like it is impossible to backport any of that from 2.0 to 1.X without breaking.

jgsogo:
For anyone arriving at this issue, let's provide **a workaround for Conan v1** (see below).

### Description of the scenario

If the **same recipe name is used in _host_ and _build_ contexts**, but those contexts are using **different recipes with different options** (number or name, not values), for example:

**name/v1/conanfile.py**
```python
class Recipe(ConanFile):
   name = "name"
   version = "v1"

   options = {
      "opt1": [True, False]
   }
```

**name/v2/conanfile.py**
```python
class Recipe(ConanFile):
   name = "name"
   version = "v2"

   options = {
      "opt2": [True, False]
   }
```

And the consumer:

**conanfile.txt**
```ini
[requires]
name/v1

[build_requires]
name/v2
``` 

Conan v1 will compute only one set of options and it will try to apply the same set to both recipes. Conan will raise an error saying that some of those options cannot be found in the recipe (indeed, `opt2` is not available in `name/v1`).


### Workaround for Conan v1

The easiest way to workaround this issue is to match the list of options between the two versions, by adding dummy (unused) options to their recipes:

**name/v1/conanfile.py**
```python
class Recipe(ConanFile):
   name = "name"
   version = "v1"

   options = {
      "opt1": [True, False],
      "opt2": [True, False],  # <-- not used in this version
   }
   
   def package_id(self):
      del self.options.opt2  # <-- removed from package-id computation
```

**name/v2/conanfile.py**
```python
class Recipe(ConanFile):
   name = "name"
   version = "v2"

   options = {
      "opt1": [True, False],  # <-- not used in this version
      "opt2": [True, False]
   }
   
   def package_id(self):
      del self.options.opt1  # <-- removed from package-id computation
```

memsharded:
It will not be possible to provide a fix for 1.X, and this will need to wait until 2.0 for a solution.
Hopefully the workaround above is enough in the meantime. Thanks all for the feedback.

PR #9840, opened 2021-10-20
[develop2] Simplifying the Options model

memsharded:
Changelog: Omit
Docs: Omit

New proposal to simplify Options:

- Remove the ``OptionValues``, ``PkgOptionValues``, ``PkgOptionValue`` classes that duplicate too much of the model, making it way more complex.
- Isolated the implementation, the only public interface to options is now the ``Options`` class
- Some simplifications in the ``Options`` class and its collaborators
- Changed how options are propagated in the graph, separating the options that are modified by downstream dependencies from the options that are self-defined or defined by profiles, because that massively affects things like build-order
- This change in propagation should help to prepare for better propagation in the "build" context. This might still need some new model, like ``build_options`` in recipes, but that will follow in another PR
- Changes in the ``conaninfo.txt`` now are very minimal, just derived from the fact that not all options from dependencies are stored there now. This will change, but in a following PR, to keep this more focused.

Close https://github.com/conan-io/conan/issues/9900


lasote (approved):
👍 🥇

czoido (approved):

PR #9842, opened 2021-10-21
test that captures with FIXME current issue with per-package settings

memsharded:
Changelog: Omit
Docs: Omit

Probably something that we want to fix for 2.0?


memsharded:
Implemented in https://github.com/conan-io/conan/pull/11202, directly for Conan 2.0

PR #9844, opened 2021-10-21
Recovered test for CMakeDeps

lasote:
Recover test Overriding with a component "rename" the global scope.
Also completing the "Help template" of the conftest to include a default cmake version and avoid unexpected results when using `contest_user.py`

Changelog: omit
Docs: omit

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -18,6 +18,7 @@ To override these locations with your own in your dev machine:
 tools_locations = {
     'svn': {"disabled": True},
     'cmake': {
+        "default": "3.19",
         "3.15": {},
         "3.16": {"disabled": True},
         "3.17": {"disabled": True},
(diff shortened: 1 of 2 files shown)


conans/test/conftest.py
```diff
@@ -18,6 +18,7 @@
 tools_locations = {
     'svn': {"disabled": True},
     'cmake': {
+        "default": "3.19",
```
memsharded, line 21: Our testing default shouldn't be 3.19, but 3.15

lasote: This is inside a comment teaching how to create the `conftest_user.py`. If you don't adjust your default, it skips cmake tests for no apparent reason (as it was my case), because the "base" default is 3.15 and maybe you have explicitly skipped the 3.15 because you don't have it installed.

memsharded (approved):

czoido (approved):

Issue #9848, opened 2021-10-22
[feature] introduce alias for PkgConfigDeps


lasote:
**Issue**: The current PkgConfigDeps scopes the `pc` files with the name of the package, but people, specially from open source and conan-center, want to consume "crypto.pc" not "openssl-crypto.pc", otherwise every consumer would need to modify the build system script.

**Proposal**

1. Keep generating files scoped, always, and a global one that include all the components:

With components:
```
self.cpp_info.components["crypto"] => "openssl-crypto.pc"
self.cpp_info.components["ssl"] => "openssl-ssl.pc"
self.cpp_info.components["tls"] => "openssl-tls.pc"
self.cpp_info.components["openssl"] => "openssl-openssl.pc"
"openssl.pc" => include all components
```
 
Without components:

```
"openssl.pc" => Declares the information using the same template of the components
```

 2. Enable a new property `pkg_config_aliases` similar to `cmake_target_aliases` (@SSE4) only when the user **really want** to declare a `pc` file without scope. This will be done in conan center in very special cases where the `pc` file is kind of a "standard". Any alias to `core` or things like that should be avoided.
 
```
self.cpp_info.components["crypto"].set_property("pkg_config_aliases", ["crypto"]) => "crypto.pc" includes "openssl-crypto.pc"
self.cpp_info.set_property("pkg_config_aliases", ["bar"]) => "bar.pc" includes "openssl.pc"
```

3. There is another special case where the recipe wants to declare a `foo.pc` but including only specific component/s, not to all of them, for example, the "core" component but not some optionals. In that case:

```
self.cpp_info.components["core"] => "foo-core.pc"
self.cpp_info.components["other"] => "foo-other.pc"
self.cpp_info.components["core"].set_property("pkg_config_aliases", ["foo"]) => "foo.pc" include to "foo-core.pc"  

In this case the `foo.pc` overwrite the global `foo.pc`.

franramirez688:
For now, it's blocked by https://github.com/conan-io/conan/pull/9851

franramirez688:
Actually, it was not blocked by that MR. It can be developed without having it merged

PR #9851, opened 2021-10-22
Component with same name as package

franramirez688:
Changelog: Feature: Let users define a new Component with the same name as the package.

Well, if you do that, Conan will raise this error:
```
ERROR:
	ConanException: pkg/0.1 package_info(): Component name cannot be the same as the package name: 'pkg'
```

This PR is aimed to solve this problem in relation to these comments:

* https://github.com/conan-io/conan/pull/9806#discussion_r733364484
* https://github.com/conan-io/conan/pull/9806#discussion_r734408718

Related PR: https://github.com/conan-io/conan/pull/9806

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -449,13 +449,6 @@ class CppInfo(_CppInfo):
                                          "but %s" % (comp_require, reason))
 
         if self.components:
-            # Raise on component name
-            for comp_name, comp in self.components.items():
-                if comp_name == package_name:
-                    raise ConanException(
-                        "Component name cannot be the same as the package name: '%s'"
-                        % comp_name)
-
             # check that requires are used in components and check that components exists in requires
             requires_from_components = set()
             for comp_name, comp in self.components.items():
--- a/conans/test/integration/generators/package_info/package_info_test.py
+++ b/conans/test/integration/generators/package_info/package_info_test.py
@@ -209,49 +209,6 @@ class HelloConan(ConanFile):
         self.assertIn("dep.defines: ['definedep1']", client.out)
         self.assertIn("dep.include_paths: ['include', 'includedep2']", client.out)
 
-    def test_package_info_raise_components(self):
-        conanfile = textwrap.dedent("""
-            from conans import ConanFile
-
-            class MyConan(ConanFile):
-
-                def package_info(self):
-                    self.cpp_info.defines.append("defint")
-                    self.cpp_info.components["int1"].libs.append("libint1")
-        """)
-        client = TestClient()
-        client.save({"conanfile.py": conanfile})
-        client.run("create conanfile.py dep/1.0@us/ch", assert_error=True)
-        self.assertIn("dep/1.0@us/ch package_info(): self.cpp_info.components cannot be used "
-                      "with self.cpp_info global values at the same time", client.out)
-
-        conanfile = textwrap.dedent("""
-            from conans import ConanFile
-
-            class MyConan(ConanFile):
-
-                def package_info(self):
-                    self.cpp_info.release.defines.append("defint")
-                    self.cpp_info.components["int1"].libs.append("libint1")
-        """)
-        client.save({"conanfile.py": conanfile})
-        client.run("create conanfile.py dep/1.0@us/ch", assert_error=True)
-        self.assertIn("dep/1.0@us/ch package_info(): self.cpp_info.components cannot be used "
-                      "with self.cpp_info configs (release/debug/...) at the same time", client.out)
-
-        conanfile = textwrap.dedent("""
-                    from conans import ConanFile
-
-                    class MyConan(ConanFile):
-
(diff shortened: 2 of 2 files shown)


jgsogo:
It would be helpful if you can summarize the scope of this PR. Besides the changes associated with `pkg_config`, how does it affect other generators? is this just to avoid the workaround? does it _fixes_ (breaking) the issue with `cmake_find_package` generator where the global target overrides (again) the overriding target?

franramirez688:
> It would be helpful if you can summarize the scope of this PR. Besides the changes associated with `pkg_config`, how does it affect other generators? is this just to avoid the workaround? does it _fixes_ (breaking) the issue with `cmake_find_package` generator where the global target overrides (again) the overriding target?

Sure @jgsogo. I wanted to open a discussion about this logic. In summary, this PR is deleting the mechanism which is avoiding the creation of `components` with the same name as the package. I don't even know how it could affect other generators and that's the reason why I'm opening this PR. Actually, I realized about it due to I wanted to create a test for `foo-foo.pc` and I couldn't because of this piece of code, and, in my opinion, that's weird.
So, I guess, the reason is, as you said before:

> `cmake_find_package` generator where the global target overrides (again) the overriding target

Are we managing it because of it? or are there more reasons that I'm missing? Would it make sense each generator should deal with it in case of having any kind of collisions?

danimtb:
This restriction was introduced in order to avoid breaking recipes that were relying on the name of the main target. At that moment we did not have features such as `names` or `filenames`, not even `set_property()`. I'll explain why it was restricted with an example (I am talking about `cmake_find_package` here):
 
Let's say we got the recipe for `openssl` that does not declare components in `cpp_info` -> the generated CMake target will be ``openssl::openssl``. Any consumer could link to ``openssl::openssl`` when requiring the package (1).

In the case that `openssl` recipe started to use components, it could break consumers by not providing an ``openssl::openssl`` target, so to maintain compatibility, if a recipe declared any target, there would be always an ``openssl::openssl`` target generated that would include all the components to cover the use case described in (1).

I reckon this was done that way because the key of the `self.cpp_info.components` dict was the one used as name of the target. Probably now with the flexibility of properties, it could be possible to remove this restriction, making sure that the target ``openssl::openssl`` is always generated (?)

lasote (changes requested):
A bit more context. This change would be needed for point 3) of https://github.com/conan-io/conan/issues/9848

I think this change is ok because:
 - If now Conan raises an exception and we remove the exception we are not breaking anything by definition. The only case would be someone relying on the Exception for something but that would be very weird and we will be able to manage it. So I think we cannot consider this "breaking" by any means.
 
 - We should unify the criteria, If a user changes a recipe introducing a new component with the name of the package, that should be interpreted by the generator as "Override the global target and use the component".  Maybe we should add a test to double-check that for the `CMakeDeps` and maybe the `cmake_find_package_multi` if possible.
 
 - If any generator couldn't manage a component named like the global (I don't know if that's the case) the generator can raise an error. But not the model. **So maybe we want to raise on every generator except the new ones moving that check removed to a function.** (to simplify the change and minimize the scope of this PR)

jgsogo:
I totally support the change. It would simplify some existing recipes, and avoid the current workaround we are doing to override the global target 🙌 


Nevertheless (and following Dani's comment), if we start to have a component with the same name of the package (recipes evolve to _accept_ this PR), it could break consumers that are already doing something like this to propagate information:

```python
def package_info(self):
    self.cpp_info.components["xxxx"].requires = ["openssl::openssl"]
```

This will no longer be all the `openssl` package, but only the things included in the _new_ non-forbidden component. It can break someone, but IMHO it's worth it (they should be using revisions, mirror recipes,...)

lasote:
Yeah, but you know, if we modify the recipe and remove the `.libs` list we are also breaking the consumers... that's ok.

lasote:
> I totally support the change. It would simplify some existing recipes, and avoid the current workaround we are doing to override the global target 🙌

About that, we need to decide the scope of this change, I think we should focus on the new generators only and for the old ones keep raising individually. I think that is the most conservative and the most aligned that what we are doing. IMHO we should focus on having recipes 2.0 compatible and forget about old integrations. @memsharded

lasote:
After the latest changes in the properties for CMakeDeps and PkgConfigDeps I would say I would move this to the 2.0 branch. About my previous comment:

> We should unify the criteria, If a user changes a recipe introducing a new component with the name of the package, that should be interpreted by the generator as "Override the global target and use the component". Maybe we should add a test to double-check that for the CMakeDeps and maybe the cmake_find_package_multi if possible.

I don't agree anymore, the component definition should be independent, a component could have the same name as the main package and then receive a rename with a property avoiding the collision, what we have to check is, in case of collisions, what would happen.

franramirez688:
This PR does not make sense for Conan v2.x because this check does not exist. Closing the PR.

PR #9852, opened 2021-10-22
Replace cpp.source + cpp.build with cpp.local

lasote:
Changelog: omit
THIS HAS BEEN DISCARDED AND WILL BE REVERTED
Docs: omit

### DIFF
--- a/conan/tools/files/packager.py
+++ b/conan/tools/files/packager.py
@@ -1,0 +1,73 @@
+import os
+
+from conans.client.file_copier import FileCopier
+from conans.errors import ConanException
+
+
+class _AutoPackagerPatternEntry(object):
+
+    def __init__(self):
+        self.include = ["*.h", "*.hpp", "*.hxx"]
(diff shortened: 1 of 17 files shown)


conan/tools/files/packager.py
```diff
@@ -0,0 +1,73 @@
...
+        self.include = ["*.h", "*.hpp", "*.hxx"]
+        self.lib = ["*.so", "*.so.*", "*.a", "*.lib", "*.dylib"]
+        self.bin = ["*.exe", "*.dll"]
+        self.src = []
+        self.build = []
+        self.res = []
+        self.framework = []
+
+
+class AutoPackager(object):
```
lasote, line 19: Not especially passionate about this class either, but better located.

memsharded (approved):
Yes, I like this, and I think this will be more intuitive for users

memsharded (approved):

czoido (approved):
I think the layouts are now more intuitive with these changes 👍

PR #9858, opened 2021-10-24
meson 0.60 broke preprocessor defs

memsharded:
Changelog: Fix: Change ``MesonToolchain`` to define ``preprocessor_definitions`` as ``[constant]``, because Meson 0.60 now raises errors on it defined in ``[built-in options]``.
Docs: Omit


memsharded (commented):
I am going to merge this, as both "develop" and "develop2" and all PRs to both are currently broken, and this fixes it. Note the fix works also in <0.60 like 0.59 which seems to still be installed in Linux agents (0.60 seems to have been updated in Win and OSX). Please review anyway, just in case.

PR #9859, opened 2021-10-24
Implement middleware support (conan-io#9636)

### LINKED ISSUE #9636: [feature] Middleware for customization of all recipes

gmeeker:
Fixes #9636
This is a re-implementation that supercedes #9637 (depending on feedback).

Changelog: (Feature): Middleware for customization of all recipes

I have not implemented tests or updated docs. This is a prototype to facilitate discussion of #9636 and may change drastically or re-implemented entirely.

### DIFF
--- a/conans/client/cache/cache.py
+++ b/conans/client/cache/cache.py
@@ -33,6 +33,7 @@ PROFILES_FOLDER = "profiles"
 HOOKS_FOLDER = "hooks"
 TEMPLATES_FOLDER = "templates"
 GENERATORS_FOLDER = "generators"
+MIDDLEWARE_FOLDER = "middleware"
 
 
 def _is_case_insensitive_os():
@@ -195,6 +196,10 @@ class ClientCache(object):
         return os.path.join(self.cache_folder, GENERATORS_FOLDER)
 
     @property
+    def middleware_path(self):
+        return os.path.join(self.cache_folder, MIDDLEWARE_FOLDER)
+
+    @property
     def default_profile_path(self):
         if os.path.isabs(self.config.default_profile):
             return self.config.default_profile
@@ -245,6 +250,17 @@ class ClientCache(object):
                 if os.path.isfile(generator) and generator.endswith(".py"):
                     generators.append(generator)
         return generators
+
+    @property
+    def middleware(self):
+        """Returns a list of middleware paths inside the middleware folder"""
+        middleware = []
+        if os.path.exists(self.middleware_path):
+            for path in os.listdir(self.middleware_path):
+                middleware_path = os.path.join(self.middleware_path, path)
+                if os.path.isfile(middleware_path) and middleware_path.endswith(".py"):
+                    middleware.append(middleware_path)
+        return middleware
 
     def delete_empty_dirs(self, deleted_refs):
         """ Method called by ConanRemover.remove() to clean up from the cache empty folders
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -23,6 +23,7 @@ from conans.client.cmd.user import user_set, users_clean, users_list, token_pres
 from conans.client.conanfile.package import run_package_method
 from conans.client.conf.required_version import check_required_conan_version
 from conans.client.generators import GeneratorManager
+from conans.client.middleware import MiddlewareManager
 from conans.client.graph.graph import RECIPE_EDITABLE
 from conans.client.graph.graph_binaries import GraphBinariesAnalyzer
 from conans.client.graph.graph_manager import GraphManager
@@ -201,8 +202,11 @@ class ConanApp(object):
         self.python_requires = ConanPythonRequire(self.proxy, self.range_resolver,
                                                   self.generator_manager)
         self.pyreq_loader = PyRequireLoader(self.proxy, self.range_resolver)
+        self.middleware_manager = MiddlewareManager(self.proxy, self.range_resolver)
         self.loader = ConanFileLoader(self.runner, self.out, self.python_requires,
-                                      self.generator_manager, self.pyreq_loader, self.requester)
+                                      self.generator_manager, self.pyreq_loader, self.requester,
+                                      middleware_manager=self.middleware_manager)
+        self.loader.load_middleware_from_cache(self.cache)
 
         self.binaries_analyzer = GraphBinariesAnalyzer(self.cache, self.out, self.remote_manager)
         self.graph_manager = GraphManager(self.out, self.cache, self.remote_manager, self.loader,
@@ -215,6 +219,7 @@ class ConanApp(object):
         self.python_requires.enable_remotes(update=update, check_updates=check_updates,
                                             remotes=remotes)
         self.pyreq_loader.enable_remotes(update=update, check_updates=check_updates, remotes=remotes)
+        self.middleware_manager.enable_remotes(update=update, check_updates=check_updates, remotes=remotes)
         return remotes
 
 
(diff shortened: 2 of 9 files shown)


memsharded:
Hi @gmeeker 

First, sorry we didn't have the time to have a look at it back then. Our backlog is huge and managing priorities is difficult. It seems that this proposal was too big and had too many implications, we never had the time to review it.

I have been reading this PR, and I am afraid this is not going to fly. There are currently 2 approaches that allows to hook into the recipes functionality:

- For your own recipes, the recommended approach is to use the functionality of a base class or helper methods inside a ``python_requires``. This allows full control of what is being done in recipes, while keeping excellent tracking on the effect on the binaries
- For your own recipes, but also for others recipes, it is possible to run code on ``hooks``

The inconvenience of the first one is that you need to plan it in advance, it is not possible to do it after the recipes have been created. But in our experience this shouldn't be a big issue, tons of users, specially the users doing these kind of things, they already have such ``python-requires``, because they need them for other purposes, including not repeating themselves.

The second one, the example in https://github.com/gmeeker/conan-multi-build-hook/blob/main/conan-multi-build.py is provided to highlight the complexity. This is more than expected, the use case of multi-build is simply too much to not be done as a core-thing. Also, Conan 2.0 has moved forward removing some trailing legacy things of 1.X related to multi-config binaries. Conan 2.0 binaries, as managed by the Conan built-in integrations will be single-config. Creating multi-config binaries, like fat binaries, is something that will be done with ``deployers``, for the final "releasing outside of Conan" stage.

I think that the use-case for signing executables is probably quite simple in ``hooks``, not being complex, and providing a good solution for the use case.

In general, middleware classes like the proposed ones will be at the very least challenging. Implementing a class that does something in the "build()`` method, will need to handle absolutely every build system out there. How does a MiddleWare that uses some CMake logic will inject to the ``boost`` or to the ``openssl`` recipes? The MiddleWare approach will really not scale beyond some simple use cases that are already addressed by hooks.

We have also learned the importance of tracking and reproducibility. This kind of functionality that lives only in the cache, and as such is not versioned directly by Conan, introduces a new dimension of non-reproducible builds. We strongly recommend ``python_requires`` because:
- They can be evolved without breaking, as every Conan recipe can require a different version to not break.
- There is absolutely full control and reproducibility, as ``python_requires`` are also in lockfiles
- ``python_requires`` affect also the ``package_id`` computations, being possible to decide what needs to be re-built from source just by bumping the right version in the ``python_requires`` package.

Anything that is stored in the cache is not versioned by Conan, we have also not implemented in Conan 2.0 the in-cache generators until last release, when we had to do it for some extreme cases, and even in those cases we will recommend that such generators are not for being used by recipes, but being used only by end-consumers from the command-line (that is, the actual builds of the packages shouldn't be using or affected by in-cache generators, only by ``python_requires`` generators)


I am really sorry to tell this, after all the work you did back then, but at this moment and with the experience we have also learned since then, we think this cannot be moved forward, the added complexity would be too much, the posterior testing, documentation, support, and maintenance would also bee too high for the value, while it could also bring a new family of issues. Thanks very much again, even if this couldn't be done, we really appreciate it!

PR #9866, opened 2021-10-26
Fix test_token_expired

czoido:
Changelog: omit
Docs: omit

Looks like this test was failing randomly maybe because sometimes it did not expire the token, let's see if this fixes that.

### DIFF
--- a/conans/test/integration/remote/auth_test.py
+++ b/conans/test/integration/remote/auth_test.py
@@ -177,7 +177,7 @@ def test_token_expired():
     server_folder = temp_folder()
     server_conf = textwrap.dedent("""
        [server]
-       jwt_expire_minutes: 0.001
+       jwt_expire_minutes: 0.01
        authorize_timeout: 0
        disk_authorize_timeout: 0
        disk_storage_path: ./data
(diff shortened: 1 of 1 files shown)


conans/test/integration/remote/auth_test.py
```diff
@@ -177,7 +177,7 @@ def test_token_expired():
     server_folder = temp_folder()
     server_conf = textwrap.dedent("""
        [server]
-       jwt_expire_minutes: 0.01
+       jwt_expire_minutes: 0.001
```
memsharded: I had this value before! It was also failing, in this case, the first upload was failing because it expired too soon, and the upload wasn't fast enough to not get expired the token

memsharded (changes requested):

conans/test/integration/remote/auth_test.py
memsharded: What I would try is to increment the sleep below.

czoido: ok, let's try

memsharded (approved):

PR #9869, opened 2021-10-26
Remove folders.package

lasote:
Changelog: Fix: Removed the new layout `folders.package` property as it was not clear the value/usage.
Changelog: Fix: The "conan package" method now raises an exception when declaring the `layout()` method. This is part of the migration to Conan 2.0 where the local method "conan package" is removed.

### DIFF
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
@@ -663,8 +663,10 @@ class TestComponentsCMakeGenerators:
                     cmake = CMake(self)
                     cmake.configure(build_script_folder="src")
                     cmake.build()
-                    self.run("./main", env="conanrun")
-            """.format("CMakeDeps"))
+                    local_path = os.path.relpath(self.cpp.local.bindirs[0], self.folders.build)
+                    cmd = os.path.join(local_path, "main")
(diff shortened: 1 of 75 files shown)


conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
```diff
@@ -663,8 +663,7 @@ def build(self):
                     cmake = CMake(self)
                     cmake.configure(build_script_folder="src")
                     cmake.build()
-                    cmd = os.path.join(self.cpp.build.bindirs[0], "main")
-                    self.run(cmd, env="conanrun")
+                    self.run("./main", env="conanrun")
```
memsharded: Why this doesn't need layout anymore?

memsharded (commented):
I don't love it either, but not sure what direction to follow

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
lasote: Probably this is needed but the test is "slow". This is the link I don't like and the same you were working last night.

lasote: Yeah, is the same test you fixed in develop. I detected it in local because I was running the whole test suite.

PR #9876, opened 2021-10-26
BazelDeps build requires support

### LINKED ISSUE #9875: [feature] Support build required dependencies in BazelDeps
Support build required dependencies in BazelDeps so that binaries at the end could be used in Bazel.

fanStefan:
Changelog: Feature: Add build_requires support in BazelDeps generators.
Docs: omit

Closes: #9875 

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -13,7 +13,8 @@ class BazelDeps(object):
 
     def generate(self):
         local_repositories = []
-        for build_dependency in self._conanfile.dependencies.build.values():
+
+        for build_dependency in self._conanfile.dependencies.direct_build.values():
             content = self._get_build_dependency_buildfile_content(build_dependency)
             filename = self._save_dependendy_buildfile(build_dependency, content)
 
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -13,6 +13,13 @@ class BazelDeps(object):
 
     def generate(self):
         local_repositories = []
+        for build_dependency in self._conanfile.dependencies.build.values():
+            content = self._get_build_dependency_buildfile_content(build_dependency)
+            filename = self._save_dependendy_buildfile(build_dependency, content)
+
+            local_repository = self._create_new_local_repository(build_dependency, filename)
+            local_repositories.append(local_repository)
+
         for dependency in self._conanfile.dependencies.host.values():
             content = self._get_dependency_buildfile_content(dependency)
             filename = self._save_dependendy_buildfile(dependency, content)
@@ -27,6 +34,18 @@ class BazelDeps(object):
         filename = 'conandeps/{}/BUILD'.format(dependency.ref.name)
         save(filename, buildfile_content)
         return filename
+
+    def _get_build_dependency_buildfile_content(self, dependency):
+        filegroup = textwrap.dedent("""
+            filegroup(
+                name = "{}_binaries",
+                data = glob(["**"]),
+                visibility = ["//visibility:public"],
+            )
+
+        """).format(dependency.ref.name)
+
+        return filegroup
 
     def _get_dependency_buildfile_content(self, dependency):
         template = textwrap.dedent("""
--- a/conans/test/unittests/tools/google/test_bazeldeps.py
+++ b/conans/test/unittests/tools/google/test_bazeldeps.py
@@ -22,7 +22,7 @@ def test_bazeldeps_dependency_buildfiles():
     conanfile_dep.cpp_info = cpp_info
     conanfile_dep._conan_node = Mock()
     conanfile_dep._conan_node.ref = ConanFileReference.loads("OriginalDepName/1.0")
-    conanfile_dep.package_folder = "/path/to/folder_dep"
+    conanfile_dep.folders.set_base_package("/path/to/folder_dep")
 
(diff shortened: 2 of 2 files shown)


conan/tools/google/bazeldeps.py
```diff
@@ -28,6 +35,18 @@ def _save_dependendy_buildfile(self, dependency, buildfile_content):
         save(filename, buildfile_content)
         return filename
 
+    def _get_build_dependency_buildfile_content(self, dependency):
+        filegroup = textwrap.dedent("""
+            filegroup(
+                name = "{}_binaries",
+                data = glob(["**"]),
+                visibility = ["//visibility:public"],
```
memsharded, line 44: without having any idea about Bazel: is it possible that we want some other visibilty here? Typically the ``build_requires`` are not propagated to downstream consumers. If this was a transitive dependency, maybe it doesn't need to be visible to the downstream?

conan/tools/google/bazeldeps.py
```diff
@@ -13,6 +13,13 @@ def __init__(self, conanfile):
 
     def generate(self):
         local_repositories = []
+        for build_dependency in self._conanfile.dependencies.build.values():
```
memsharded: For ``build_requires`` it is typical that only direct requires executables are available, for example if you have a build_require to "cmake", and it has a requires to ``openssl``, the consumer should only have access to cmake.exe, but not necessary to openssl.exe (a different story is shared libs). Maybe want to use ``direct_build`` here?

memsharded (approved):
Looking good. I have no much idea about Bazel, but it seems you do, the implementation is clean, it is unit-tested, and I guess that you have tried manually with something bigger and a bit more real? In that case, this could be merged soon, and we could keep iterating the Bazel integration if necessary. Thanks very much for contributing this.

conan/tools/google/bazeldeps.py
fanStefan: Then this definitely need to be `direct_build`. It will be changed. Thanks for the explanation.

conan/tools/google/bazeldeps.py
fanStefan: With `visibility` attribute of the bazel targets, in this case `{conan_package_name}_binaries` target, we will define which bazel packages in the repository can use it as dependency. Be aware that bazel package are something completely different than conan package. So this should always be `//visibility:public`, because we don't know in which bazel packages user want to use it as dependency.

fanStefan:
> Looking good. I have no much idea about Bazel, but it seems you do, the implementation is clean, it is unit-tested, and I guess that you have tried manually with something bigger and a bit more real? In that case, this could be merged soon, and we could keep iterating the Bazel integration if necessary. Thanks very much for contributing this.

Yes, it was tested manually, but there is one thing that maybe is not that great and needs rework. If same conan package is in `requires` and `build_requires` sections, than generator will not generate anything for  `build_requires` package.
Is it common to have same package it both sections?

Adding `filegroup` for `build_requires` packages will expose conan package content to bazel. I think that it is enough at this moment and later we can change if needed.

You are welcome.

memsharded:
> Yes, it was tested manually, but there is one thing that maybe is not that great and needs rework. If same conan package is in requires and build_requires sections, than generator will not generate anything for build_requires package.
Is it common to have same package it both sections?

Yes, it is possible, typical use case, ``protobuf`` as build_require, to be able to use ``protoc`` compiler (in the build context, build profile), and ``protobuf`` as ``requires`` to be able to link with the ``protobuf`` library (in the host context).

> Adding filegroup for build_requires packages will expose conan package content to bazel. I think that it is enough at this moment and later we can change if needed.

Sure, no need to be perfect, it can be changed and improved later.

conan/tools/google/bazeldeps.py
memsharded: Great. We are about to close 1.42, this might go in if done quickly, otherwise (or in case someone else raises some issues),  it would need to go to 1.43.

fanStefan: I would leave it for 1.43.

fanStefan: When package is present in `[build_requires]` section only, `direct_build` contains expected value.
If the same package is present in both `[build_requires]` and `[requires]` sections, `direct_build` is empty.
Is this expected behavior?

memsharded: Are you sure you are providing both the "build" and the "host" profile? You should pass them both, even if they are the same.

fanStefan: I don't provide any profile. Thanks for the hint, I will take a look on this.

tapia: @fanStefan please mention me when you finish your rework on this, so we can review and merge your contribution.

Thank you!

memsharded: Hi @fanStefan. Did you manage to check that of the profiles? 
I would like to merge this in 1.43 if possible. Thanks!

fanStefan: Unfortunately I can't find time to work on this. Can we prolong the merging of this one, to the one of next milestones?

memsharded: Yes, no prob, it can wait. If for next release not possible, maybe @tapia lets try to move it forward and finish it ourselves?

tapia: Sure, we can delay it until 1.44, and take over if @fanStefan can't find the time.

PR #9879, opened 2021-10-27
Fix some tests to use MSVC

### LINKED ISSUE #9683: Update tests to use msvc setting
Instead of Visual Studio

Not necessary to migrate the whole test suite. Just add or update some tests to have better coverage of the new ``msvc`` compiler.

There are already some tests that use it like ``test_locally_build_msvc_toolset()`` or ``test_toolchain_win_vs2017()``. The goal is to check relevant cases, like standard CMake build, and parameterize or add new tests.

Only ``conan.tools`` new integrations

tapia:
Changelog: omit
Docs: omit

Closes #9683

### DIFF
--- a/conans/test/unittests/client/settings_preprocessor_test.py
+++ b/conans/test/unittests/client/settings_preprocessor_test.py
@@ -2,6 +2,8 @@
 
 import unittest
 
+import pytest
+
 from conans import Settings
 from conans.client.conf import get_default_settings_yml
 from conans.client.settings_preprocessor import preprocess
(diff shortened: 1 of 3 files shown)


conans/test/unittests/client/settings_preprocessor_test.py
```diff
@@ -64,3 +65,32 @@ def test_custom_base_runtime_set(self):
         self.settings.compiler.runtime = "MT"
         preprocess(self.settings)
         self.assertEqual(self.settings.compiler.runtime, "MT")
+
+class SettingsCompilerMSVCPreprocessorTest(unittest.TestCase):
```
memsharded: Do not do unittest anymore for new tests, please try to do only ``pytest`` ones.

memsharded (commented):
You need to merge/rebase continuously from ``develop`` branch, at least before doing the PR. It evolves quick, and your branch gets easily outdated. I have done some minor changes and merged.

PR #9883, opened 2021-10-27
[develop2][refactor]use db two tables

memsharded:
Changelog: Omit
Docs: Omit

Change the DB to use Recipes and Packages tables. Should be a pure refactor.


conans/model/ref.py
```diff
@@ -1,4 +1,5 @@
 import re
+import traceback
```
lasote, line 2: ?

lasote (approved):

czoido (approved):

conans/model/ref.py
memsharded: I will remove it later

PR #9891, opened 2021-10-28
Improve 'conan search' error message

tapia:
Changelog: Fix: Show a more specific error message on 'conan search' when a package version is not valid
Docs: omit


conans/model/ref.py
```diff
@@ -177,7 +197,7 @@ def _validate(self):
         if self.name is not None:
             ConanName.validate_name(self.name, reference_token="package name")
         if self.version is not None:
-            ConanName.validate_name(self.version, True, reference_token="package version")
+            ConanName.validate_name(self.name, self.version, reference_token="package version")
```
memsharded: This is ugly, because it is breaking the pattern of ``validate_name`` receiving as first argument what needs to be validated, and then arguments of how it should be validated. I don't care much, because this will be changed soon in 2.0, but not pretty.

tapia: I agree it could be better, but I don't like the other way either. A function called `validate_name`, but what is validating maybe is a name, maybe it's not. With this change at least the first parameter is always the package's name, and you get also the version if it's necessary.

Maybe we should split this function into `validate_name` and `validate_version`?

memsharded (approved):

conans/model/ref.py
tapia: I just did the split, it looks much better now 👍

lasote (changes requested):
I like the split but is still confusing, check the comment.

conans/model/ref.py
```diff
@@ -129,17 +142,27 @@ def validate_string(value, reference_token=None):
...
         if name == "*":
             return
         if ConanName._validation_pattern.match(name) is None:
-            if version and ((name.startswith("[") and name.endswith("]")) or
-                            (name.startswith("(") and name.endswith(")"))):
-                return
             ConanName.invalid_name_message(name, reference_token=reference_token)
 
+    @staticmethod
+    def validate_version(name, version):
```
lasote:
```suggestion
    def validate_version(version, pkg_name):
```
Very confusing. Put the version as the first argument.

tapia: Done

memsharded (approved):

conans/model/ref.py
```diff
@@ -117,6 +117,19 @@ def invalid_name_message(value, reference_token=None):
         )
         raise InvalidNameException(message)
 
+    @staticmethod
+    def invalid_version_message(name, version):
```
franramirez688: If the function's goal is to raise an exception, shouldn't `def raise_invalid_version_error` be a better name? Or something like that

tapia: Done

franramirez688 (approved):

czoido (approved):

PR #9894, opened 2021-10-28
making the filename property independent

memsharded:
Changelog: Fix: Do not fall back for the property "filename" to target properties, only to legacy "names" definition.
Docs: omit

#tags: slow


danimtb:
`cpp_info.set_property()` is not extended right now across recipes in conan-center-index, so this should be fine

czoido (approved):

danimtb (approved):

SSE4 (approved):

PR #9896, opened 2021-10-28
Bugfix: missing framework for xcode generator with no compiler setting

### LINKED ISSUE #9878: [bug] conanbuildinfo.xcconfig is missing dependency frameworks
When I build a recipe depending (directly, but mostly transitively) on a number of other recipes which generate iOS framework artifacts, there's a difference in behavior between 1.23.0 and 1.41.0.

The difference seems to be in the generated conanbuildinfo.xcconfig. HEADER_SEARCH_PATHS and LIBRARY_SEARCH_PATHS seem to be similar enough to not be a problem. However, OTHER_LDFLAGS under 1.23.0 contains every framework dependency as `-framework MyFramework`. In 1.41.0 they're all gone, causing a failure to link.

conanbuildinfo.txt still contains all the relevant frameworks in the `[frameworks]` section in both versions.

SSE4:
closes: #9878 

Changelog: Bugfix: Missing framework for Xcode generator with no compiler setting.
Docs: omit

### DIFF
--- a/conans/client/generators/xcode.py
+++ b/conans/client/generators/xcode.py
@@ -29,7 +29,8 @@ FRAMEWORK_SEARCH_PATHS = $(inherited) {rootpaths} {framework_paths}
         self.cxx_compiler_flags = " ".join(deps_cpp_info.cxxflags)
         self.linker_flags = " ".join(deps_cpp_info.sharedlinkflags)
         self.rootpaths = " ".join('"%s"' % d.replace("\\", "/") for d in deps_cpp_info.rootpaths)
-        self.frameworks = " ".join(format_frameworks(deps_cpp_info.frameworks, self.conanfile.settings))
+        self.frameworks = " ".join(["-framework %s" % framework
+                                    for framework in deps_cpp_info.frameworks])
         self.framework_paths = " ".join(deps_cpp_info.framework_paths)
         self.system_libs = " ".join(['-l%s' % lib for lib in deps_cpp_info.system_libs])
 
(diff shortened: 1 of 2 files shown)


conans/client/generators/xcode.py
```diff
@@ -29,7 +30,10 @@ def __init__(self, conanfile):
         self.cxx_compiler_flags = " ".join(deps_cpp_info.cxxflags)
         self.linker_flags = " ".join(deps_cpp_info.sharedlinkflags)
         self.rootpaths = " ".join('"%s"' % d.replace("\\", "/") for d in deps_cpp_info.rootpaths)
-        self.frameworks = " ".join(format_frameworks(deps_cpp_info.frameworks, self.conanfile.settings))
+
+        settings = Settings({"compiler": ["apple-clang"]})
```
memsharded: This is very hacky, simply overriding this with a local compiler is very dirty.
If the ``format_frameworks`` has this implementation:
```python
def format_frameworks(frameworks, settings):
    """
    returns an appropriate compiler flags to link with Apple Frameworks
    or an empty array, if Apple Frameworks aren't supported by the given compiler
    """
    compiler = settings.get_safe("compiler")
    compiler_base = settings.get_safe("compiler.base")
    if (str(compiler) not in GCC_LIKE) and (str(compiler_base) not in GCC_LIKE):
        return []
    return ["-framework %s" % framework for framework in frameworks]
```
With your hack, this means that it will be compiler independent, so it would be good to do just:
```
self.frameworks = " ".join(["-framework %s" % framework for framework in frameworks])
```

If it is safe to assume that in Xcode, the frameworks will always be there, which probably doesn't hold for cross-building (is cross-building to some platform that won't accept frameworks possible with Xcode?)

memsharded (changes requested):

conans/client/generators/xcode.py
SSE4: so far, I didn't see such cross-building. AFAIK it allows only cross-building to apple platforms (iOS, tvOS, watchOS). I think at least for now, we can assume Xcode supports frameworks, and it only supports `apple-clang` compiler. or at least we can say conan supports only that use-case.

PR #9899, opened 2021-10-28
conan_server own package

lasote:
- `pip install -e .` only installs conan (without conan_server requirements)
- ⚠️ To install `conan_server` tool in development we cannot (apparently) use `pip install -e` but we can do `python setup_server.py install`
- The deployment can be done the same way but split, something like:
  ```
   $ > rm -rf dist
   $ > python3 setup.py bdist_wheel 
   $ > twine upload dist/* 
   $ > rm -rf dist
   $ > python3 setup_server.py bdist_wheel 
   $ > twine upload dist/* 
   ```
   This has to be confirmed by the release manager @czoido (note that I'm not sure if we are using bdist_wheel or sdist, keep using the same)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,9 +28,11 @@ def get_requires(filename):
 project_requirements = get_requires("conans/requirements.txt")
 dev_requirements = get_requires("conans/requirements_dev.txt")
 # The tests utils are used by conan-package-tools
-exclude_test_packages = ["conans.test.{}*".format(d)
+excluded_test_packages = ["conans.test.{}*".format(d)
                          for d in os.listdir(os.path.join(here, "conans/test"))
                          if os.path.isdir(os.path.join(here, "conans/test", d)) and d != "utils"]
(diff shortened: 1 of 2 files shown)


czoido:
We are using `python3 setup.py sdist` I'll make some tests with the release job.

memsharded (approved):

czoido:
For the moment until we discover a better solution or maybe after we change the whole conan project layout, we have to rename the setup.py to create the package:
```
rm -rf ./dist && python3 setup.py sdist && python3 -m pip install dist/conan-2.0.0.dev0.tar.gz  
rm -rf ./dist && rm setup.py && mv setup_server.py setup.py && python3 setup.py sdist && pip install dist/conan-server-2.0.0.dev0.tar.gz
```
Also, the executable for the package `conan-server` will be `conan_server` for the moment, it may change in the future to `conan-server`

czoido (approved):

PR #9904, opened 2021-10-28
Implement build variants support

### LINKED ISSUE #9903: [feature] Build variants to manage iOS/OSX universal binaries
Although using CMake with Xcode supports universal binaries, many libraries are using other build systems.  This has been discussed in https://github.com/conan-io/conan-extensions/issues/59 but it's complicated and specific to Apple.  It's also apparently an issue for Android as discussed in https://github.com/conan-io/conan/issues/6384 (I'm not familiar with Android.)

This proposal, however, is to create a general purpose framework for calling the build tools multiple times, and then combining them (e.g. with 'lipo' on iOS/OSX).  Using pyreq it could look like this:

```
from conans import ConanFile, Variants

class ExampleVariants(Variants):
    name = "variants_pyreq"
    version = "1.0"

    def conanfile(self):
        return super(ExampleVariants, self)

    def package(self):
        self.package_variants()
        print("package: running lipo")
        # self.run("lipo ...")

class ExampleVariantsConanFile(ConanFile):
    name = "variants_pyreq"
    version = "1.0"
```

```
from conans import ConanFile, Variants

# using pyreq makes the class hierarchy like this:
# class LipoConan(Variants, ConanFile):

class ExampleConan(ConanFile):
    name = "variants_example"
    version = "1.0"
    settings = "arch", "os"
    python_requires = "variants_pyreq/1.0@user/testing"
    python_requires_extend = "variants_pyreq.ExampleVariants"

    def configure(self):
        """ usual configure here """
        # self.set_variants("x86_64 armv8")
        # self.set_variants(self.settings.multiarch)
        self.set_variants([
            {"arch": "x86_64", "os.version": "10.13", "display_name": "Intel", "folder": "x86_64"},
            {"arch": "armv8", "os.version": "11.0", "display_name": "M1", "folder": "armv8"},
        ])

    def build_variant(self):
        print("build", self.settings.arch)

    def package_variant(self):
        print("package", self.settings.arch)
```

Note that Variants is a separate mix-in class, not a subclass of ConanFile.  This works better for pyreq and also middleware (see below).  Variants is only responsible for cloning the ConanFile instance for each variant and run build(), package() and test() in a subfolder (currently "variants/x86_64", etc.)  It also handles settings for each variant, including adapting package_id().  Options are not handled per variant, although maybe we should consider this?  Some libraries have processor specific options for SSE, etc.  See the docstrings for Variants for more details, and how super() and MRO affect this.

My current implementation contains few changes outside of the new Variants flag, no new settings, and nothing OS specific.  Perhaps lipo support could be included in conan eventually, but that's beyond the scope of this feature request.  A more complete example using lipo for Apple is here:
https://github.com/gmeeker/conan-lipo-middleware/pyreq

Note, however, that each recipe must be customized to support variants.  https://github.com/conan-io/conan/issues/9636 will address this with middleware to adapt existing recipes (like CCI) and can work with variants.  It turned out to be the more complicated of the two, and variants is likely to get more attention.

gmeeker:
Fixes #9903

Changelog: (Feature): General purpose build variants to manage iOS/OSX universal binaries

I have not implemented tests or updated docs.  Awaiting feedback because this is a prototype that is likely to change.

### DIFF
--- a/conans/__init__.py
+++ b/conans/__init__.py
@@ -5,6 +5,7 @@ from conans.client.build.msbuild import MSBuild
 from conans.client.build.visual_environment import VisualStudioBuildEnvironment
 from conans.client.run_environment import RunEnvironment
 from conans.model.conan_file import ConanFile
+from conans.model.conan_variants import Variants
 from conans.model.options import Options
 from conans.model.settings import Settings
 from conans.util.files import load
--- a/conans/client/generators/__init__.py
+++ b/conans/client/generators/__init__.py
@@ -144,7 +144,7 @@ class GeneratorManager(object):
             raise ConanException("Internal Conan error: Generator '{}' "
                                  "not commplete".format(generator_name))
 
-    def write_generators(self, conanfile, old_gen_folder, new_gen_folder, output):
+    def _write_generators(self, conanfile, old_gen_folder, new_gen_folder, output):
         """ produces auxiliary files, required to build a project or a package.
         """
         _receive_conf(conanfile)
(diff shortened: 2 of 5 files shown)


memsharded:
Please see my comment in https://github.com/conan-io/conan/issues/9903#issuecomment-954238276

I think this approach will not fly. The comment suggests moving this to the deploy stage, not in the build/package stage.
The main problem is not only that this will introduce some relevant complexity, but the approach is fundamentally flawed:
When you are evaluating a recipe, that recipe is already instantiated in a dependency graph with one given configuration (settings + options), and just 1 single configuration. Therefore, you cannot do several builds for different architectures and platforms in the same place, because the dependencies for those architectures and platforms will not be there, they can't be there. The only possibility is that all other used dependencies also followed the same approach, effectively deleting all the binary information, and that is quite a viral behavior that won't scale.

I'd say that this PR should be closed, and leveraging the Conan 2.0 new custom modular commands, start working on a deploy-aggregator with lipo command.

gmeeker:
I split this off https://github.com/conan-io/conan/issues/9636 where the idea was to alter all recipes to run lipo using middleware.  Yes, this is pretty flawed on it's own.  That PR's even more complex than this one, of course.

There are cases where running lipo at deployment is different than a fat package, e.g. a signed binary will be need to be re-signed and bundles are more complicated.  Maybe it's not enough of a distinction given the complexity, but I've been considering how to deploy signed fat binaries.  It sounds like this is beyond the scope of conan.

I'll take a look at 2.0.  It sounds like a simpler PR for sure.

Issue #5163, opened 2019-05-17
Return errorcode 6 when settings are constrained to not support the current environment


monsdar:
There are two approaches to restrict the environment a recipe is working for: [Constrain Settings and Options](https://docs.conan.io/en/latest/mastering/conditional.html?highlight=conaninvalidconfiguration#constrain-settings-and-options)

When using the first approach and raising a `ConanInvalidConfiguration` Conan returns with errorcode 6. This could be further processed by scripts/CI perfectly. However the second approach with constraining the settings does not return with errorcode 6, but results in an [error](http://docs.conan.io/en/latest/faq/troubleshooting.html#error-invalid-setting) and errorcode 1 instead.

Is there a reason why both return with different behaviour? It would be perfect if both would return with errorcode 6, so developers could use both methods and scripts/CI can react on that in the same way.

lasote:
I think you are right, make sense to return 6 too. Do you agree @memsharded ?
I've open a PR with the fix and assigned to the release.

fourbft:
Conan version 1.6.1 handles constrained settings correctly now, when calling "create" (returns error code 6 on unsupported environments), but "info" still returns error code 1 in such a case.

monsdar:
@lasote or @memsharded: Could you reopen the issue? It seems that it's only partially implemented, at least `conan info` is missing this return code.

monsdar:
I did a PR which right now only tests for the issue...

PR #9905, opened 2021-10-28
Add return code 7 (Invalid system requirements)

n-bes:
Changelog: Fix: Respect error code 6 in some situations.
Docs: omit

In my case, i want to filter builds (with incorrect environment) which return code is 6. Based on https://github.com/conan-io/conan/issues/5163, I think that it can be merged.

### DIFF
--- a/conans/client/tools/system_pm.py
+++ b/conans/client/tools/system_pm.py
@@ -5,7 +5,7 @@ import six
 from conans.client.runner import ConanRunner
 from conans.client.tools.oss import OSInfo, cross_building, get_cross_building_settings
 from conans.client.tools.files import which
-from conans.errors import ConanException, ConanInvalidConfiguration
+from conans.errors import ConanException, ConanInvalidSystemRequirements
 from conans.util.env_reader import get_env
 from conans.util.fallbacks import default_output
 
@@ -116,7 +116,7 @@ class SystemPackageTool(object):
             if mode == "verify" and not self._installed(packages):
                 self._output.error("The following packages need to be installed:\n %s"
                                    % "\n".join(packages))
-                raise ConanInvalidConfiguration("Aborted due to CONAN_SYSREQUIRES_MODE=%s. "
+                raise ConanInvalidSystemRequirements("Aborted due to CONAN_SYSREQUIRES_MODE=%s. "
                                      "Some system packages need to be installed" % mode)
 
         if not force and self._installed(packages):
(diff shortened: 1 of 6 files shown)


conans/client/tools/system_pm.py
```diff
@@ -116,7 +116,7 @@ def install(self, packages, update=True, force=False, arch_names=None):
             if mode == "verify" and not self._installed(packages):
                 self._output.error("The following packages need to be installed:\n %s"
                                    % "\n".join(packages))
-                raise ConanException("Aborted due to CONAN_SYSREQUIRES_MODE=%s. "
+                raise ConanInvalidConfiguration("Aborted due to CONAN_SYSREQUIRES_MODE=%s. "
```
memsharded: Could you please elaborate why do you need this to be a "6" return code?
Because ``ConanInvalidConfiguration`` is defined for "impossible" configurations of **settings** and **options**, for example, this package cannot build in Windows (only build in Linux), or it is impossible to build this specific library for arm, etc. 

This doesn't seem the case, because if the CONAN_SYSREQUIRE_MODE would have other value, then the resulting binary would be correct for the given settings.

memsharded (commented):
Thanks for your contribution. 
However I have some doubts that this can be done with ``ConanInvalidConfiguration``, and I would need to understand better the use case.

conans/client/tools/system_pm.py
n-bes: I chose ```ConanInvalidConfiguration``` as more suitable for this situation. It can be any else.

In my case, i want to split real errors and errors which can be fixed:

``` sh
# libdaemon failed with error code 1:
Daemon returned 0 as return value.
Failed to kill daemon: Timer expired
WARN: Remotes registry file missing, creating default one in /root/.conan/remotes.json
ERROR: libdaemon/0.14 (test package): Error in test() method, line 17
	self.run(bin_path + " -k", run_environment=True)
	ConanException: Error 1 while executing bin/test_package -k
```

``` sh
# ffmpeg  failed with error code 1:

ERROR: The following packages need to be installed:
 libva-dev
vaapi/system: ERROR: while executing system_requirements(): Aborted due to CONAN_SYSREQUIRES_MODE=verify. Some system packages need to be installed
```

memsharded: Ok, I see. Then I guess something like ``ConanInvalidSystemRequirements``, with a dedicated error code would make more sense. Wdyt @lasote ?

lasote: Yes, it could be another exception and with a different return code. But I still need a bit more context about why we need it, sounds good but not fully convinced.

n-bes: I'm preparing table to find broken recipes and I want to focus on recipes, not an environment.

For now, my table contains only thousands builds, but for millions it can simplified my analyze. It also not a problem for me have patched version of ```conan``` (i mean reject PR).
 
```sql
SELECT result_code, COUNT(result_code)
FROM build
GROUP BY result_code;

result_code  COUNT(result_code)
-----------  ------------------
0            2207              
1            148               
6            41                

```

n-bes: > Yes, it could be another exception and with a different return code.

@lasote New return code ```ERROR_INVALID_SYSTEM_REQUIREMENTS  (7)``` added

lasote (approved):
👍

memsharded (approved):

PR #9918, opened 2021-10-31
Remove full_options, full_settings, full_requires from conaninfo.txt

memsharded:
Follow up of https://github.com/conan-io/conan/pull/9840 (already merged)

Removing ``full_options`` , ``full_settings``, etc. from ``conaninfo.txt``. That will change some computed package revisions in a few tests.


conans/model/info.py
```diff
@@ -530,9 +517,10 @@ def serialize_min(self):
         """
         This info will be shown in search results.
         """
+        # Lets keep returning the legacy "full_requires" just in case some client uses it
         conan_info_json = {"settings": dict(self.settings.serialize()),
                            "options": dict(self.options.serialize())["options"],
-                           "full_requires": self.full_requires.serialize()
+                           "requires": self.requires.serialize()
```
memsharded, line 523: Not an issue for compatibility, the client has had a ``requires->full_requires`` fallback for a long time.

memsharded (commented):
The ``"options": dict(self.options.serialize())["options"],`` is what might slightly change:

- Now, it can add dependencies options like ``mydep:myoption=value`` to the returned values. In the past this didn't happen, because those values were removed, and if ``package_id()`` manipulation of ``self.info.options`` added them, they were not returned by ``conan search``.

lasote (approved):

PR #9924, opened 2021-11-02
Avoid creating entries in DB for temporary export without revision yet

memsharded:
Changelog: Omit
Docs: Omit


conan/cache/cache.py
```diff
@@ -51,35 +43,30 @@ def base_folder(self):
...
 
-    def create_tmp_reference_layout(self, ref: ConanReference):
-        assert not ref.rrev, "Recipe revision should be unknown"
-        # even if they are temporal, we have to assign unique revisions to temporal references
-        # until we know the calculated revision so that we don't store multiple references
-        # with same rrev or prev
-        temp_rrev = f"{RREV_UNKNOWN}-{str(uuid.uuid4())}"
-        ref = ConanReference(ref.name, ref.version, ref.user, ref.channel, temp_rrev,
-                             ref.pkgid, ref.prev)
+    def create_export_recipe_layout(self, ref: ConanReference):
```
memsharded, line 65: Tried to add the ``export`` word here, to explicit the use case, this only happens when exporting a recipe to the cache. As the implementation shows, this has not entry in the DB now

conan/cache/cache.py
```diff
@@ -95,15 +82,6 @@ def create_tmp_package_layout(self, pref: ConanReference):
         self._create_path(package_path)
         return PackageLayout(pref, os.path.join(self.base_folder, package_path))
 
-    def create_reference_layout(self, ref: ConanReference):
```
memsharded, line 98: This was not called externally, so I have integrated it in the ``get_or_create_reference_layout()`` (it could have been made private, but as it is not very long, and only used once, I integrated it)

conan/cache/cache.py
```diff
@@ -191,9 +147,8 @@ def _move_prev(self, old_pref: ConanReference, new_pref: ConanReference):
 
         return new_path
 
-    def update_recipe(self, old_ref: ConanReference, new_ref: ConanReference = None,
-                         new_path=None, new_timestamp=None):
-        self._db.update_recipe(old_ref, new_ref, new_path, new_timestamp)
+    def update_recipe_timestamp(self, ref: ConanReference,  new_timestamp=None):
```
memsharded, line 150: The only scenario of recipe_reference update in the DB is when there is an update of timestamp due to some server updates. Nothing else should/can be updated, so better limit scope.

conan/cache/db/recipes_table.py
```diff
@@ -53,38 +53,27 @@ def get(self, ref: ConanReference):
             raise ConanReferenceDoesNotExistInDB(f"No entry for recipe '{ref.full_reference}'")
         return self._as_dict(self.row_type(*row))
 
-    def save(self, path, ref: ConanReference, timestamp):
-        # we set the timestamp to 0 until they get a complete reference, here they
-        # are saved with the temporary uuid one, we don't want to consider these
-        # not yet built packages for search and so on
+    def create(self, path, ref: ConanReference, timestamp):
```
memsharded, line 56: Try to explicit say this is only a "creation" scenario, not possible an "update" calling it. Reference should not exist in DB yet

lasote (approved):

czoido (approved):

PR #9927, opened 2021-11-02
removing python 2 legacy

memsharded:
- Mostly Python 2 legacy
- Some minor cleaning in ``self.run()``
- Removing legacy ``os_build`` messages
- Python version checks
- Replaced custom ``which()`` with ``shutil.which()``


lasote (approved):

lasote:
🥳

PR #9929, opened 2021-11-03
Make the vs environment take effect when generating the configuration…

deadash:
When meson generates the configuration, append_vcvars does not work. This pr submission is to make this parameter work.

The benefits are as follows:
- [x] A 32-bit program can be generated when vs is backend, which was not possible before
- [x] In the old version of meson, this allows meson to find vs

Signed-off-by: deadash <dead.ash@hotmail.com>

Changelog: (Feature): Describe here your pull request

### DIFF
--- a/conans/client/build/meson.py
+++ b/conans/client/build/meson.py
@@ -164,7 +164,11 @@ class Meson(object):
             build_type
         ])
         command = 'meson "%s" "%s" %s' % (source_dir, self.build_dir, arg_list)
-        with environment_append({"PKG_CONFIG_PATH": pc_paths}):
+        envs = {"PKG_CONFIG_PATH": pc_paths}
+        if self._append_vcvars and self.backend.startswith("vs"):
+            vcvars_dict = tools.vcvars_dict(self._settings, output=self._conanfile.output)
+            envs.update(vcvars_dict)
+        with environment_append(envs):
             self._run(command)
 
     @property
--- a/conans/test/functional/build_helpers/meson_test.py
+++ b/conans/test/functional/build_helpers/meson_test.py
@@ -30,4 +30,5 @@ class MesonTest(unittest.TestCase):
 
         client.save({"conanfile.py": conanfile_vcvars})
         client.run('create . pkg/1.0@ -e PATH="MyCustomPath"')
-        self.assertIn("pkg/1.0: PATH ENV VAR: MyCustomPath;", client.out)
+        self.assertIn("pkg/1.0: PATH ENV VAR: ", client.out)
+        self.assertIn("MyCustomPath;", client.out)


deadash:
The test is adjusted because the vc environment must be placed in the front, if it is placed in the back, there will be a problem of meson judgment error.

conans/test/functional/build_helpers/meson_test.py
```diff
@@ -30,4 +30,5 @@ def run(self, cmd):
 
         client.save({"conanfile.py": conanfile_vcvars})
         client.run('create . pkg/1.0@ -e PATH="MyCustomPath"')
-        self.assertIn("pkg/1.0: PATH ENV VAR: MyCustomPath;", client.out)
+        self.assertIn("pkg/1.0: PATH ENV VAR: ", client.out)
```
memsharded, line 33: How is this test different than before? Is this change necessary?

conans/client/build/meson.py
```diff
@@ -164,7 +164,11 @@ def configure(self, args=None, defs=None, source_dir=None, build_dir=None,
             build_type
         ])
         command = 'meson "%s" "%s" %s' % (source_dir, self.build_dir, arg_list)
-        with environment_append({"PKG_CONFIG_PATH": pc_paths}):
+        envs = {"PKG_CONFIG_PATH": pc_paths}
+        if self._append_vcvars and str.startswith(self.backend, "vs"):
```
memsharded: why not ``self.backend.startswith("vs")``? This is the usual syntax we use in the codebase.

conans/client/build/meson.py
```diff
@@ -164,7 +164,11 @@ def configure(self, args=None, defs=None, source_dir=None, build_dir=None,
...
+            vcvars_dict = tools.vcvars_dict(self._settings, output=self._conanfile.output)
```
memsharded, line 169: This change might be breaking. It introduces the usage of vcvars in an scenario that was not done before.

memsharded (commented):
The whole helper is disabling, harcoded in code the usage of other backends:
```python
    def _run_ninja_targets(self, args=None, build_dir=None, targets=None):
        if self.backend != "ninja":
            raise ConanException("Build only supported with 'ninja' backend")
```
So this will not work. Adding tests trying to change the backed would fail.

Furthermore, this is the legacy integration. Changes and improvements should be targeted to the new ones: https://docs.conan.io/en/latest/reference/conanfile/tools.html

conans/client/build/meson.py
deadash: Yes, there is no difference between the two, then do as you say.

I do this just for the convenience of automatic coding.

conans/test/functional/build_helpers/meson_test.py
deadash: Previously, the environment variable was inserted in the middle, and the vs environment variable must be inserted earlier, otherwise it will be destroyed by the other environment of the system.

The resulting effect is
```
pkg/1.0: PATH ENV VAR: [vs env];bla bla bla;MyCustomPath;
```
Instead of expected
```
pkg/1.0: PATH ENV VAR: MyCustomPath;[vs env];bla bla bla;
```

conans/client/build/meson.py
deadash: This change is because when meson uses vs as the backend to generate vs solution, it is in the generation configuration stage. It happens that if vcvars is not attached, meson will find vcvars by itself, and then the 32-bit arch set will be invalid. In ninja, Environment variables are used in the generation phase, so ninja inserts vcvars when building.

E.g:
Set compiler.arch=x86
Then using vs as the backend, the 64-bit program is generated instead of the expected 32-bit program.
If you add this patch, VS will generate a 32-bit program.

deadash:
@memsharded 

I tried a new compilation tool, which does not support vs as a backend. And there are some problems.

If you fully support vs as a backend, there are a lot of codes that need to be modified, such as whether to compile using meson completely, or only vs. using meson, and ninja uses ninja_target to compile.

The patch here does not adapt to this problem, but only fixes the problem when vs is used as a backend.

Just replace `meson.build()` with `meson._run_meson_command('compile', None, build_dir=self.build_folder)`

deadash:
@memsharded Is there any reason why it cannot be merged? This pr will be very helpful for using meson instead of mesontoolchain.

memsharded:
This PR has become outdated, I am closing it. 

Conan 2.0 has already removed the ``Meson`` legacy helper, and only the new ``from conan.tools.meson import Meson`` keeps there. Also Meson 1.0 is there, so probably some other things have changed too.

PR #9930, opened 2021-11-03
some new Version improvements for build and pre-releases

memsharded:
New ``RecipeReference`` is using a new ``Version`` class:

- The legacy ``Version`` class we have is quite dirty
- The new ``Version`` that was added to ``tools.Version`` for recipes usage and comparison is based on the dependency pip "node-semver" which seems to be entering maintenance mode, not much activity (17 stars in Github)
- Users seems to be demanding more flexibility on versions, having more than 3 main digits.

So I am evaluating the effort to maintain our own ``Version`` class for everything. I have fully read ``node-semver`` code. So far seems quite doable.

### DIFF
--- a/conans/model/recipe_ref.py
+++ b/conans/model/recipe_ref.py
@@ -1,4 +1,3 @@
-import re
 from functools import total_ordering
 
 from conans.errors import ConanException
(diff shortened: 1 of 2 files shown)


lasote (approved):
Looks good. This can be introduced and keep evaluating if we want to take over the dependency.

PR #9932, opened 2021-11-03
Fixes wrong parameter name in the find_library() function

### LINKED ISSUE #9931: [bug] Wrong parameter name used in find_library() function in cmakedeps_macros.cmake
When using the cmake function find_library() it accepts a parameter named "NAMES" but in cmakedeps_macro.cmake, generated by conan/tools/cmake/cmakedeps/templates/macros.py , the parameter name "NAME" is used by mistake.
See also: https://cmake.org/cmake/help/latest/command/find_library.html

As the NAMES parameter name is kind of optional, it currently works as it is but the find_library() function not only searches for the actual library by its name, but it also looks for a library called NAME.
This can bee seen, when cmake is run with the --debug-find argument.

I will provide a pull request for the fix.

akleber:
Changelog: Fix: Fix wrong parameter name in CMakeDeps find_library function.
Docs: omit

Fixes #9931


memsharded (approved):
Many thanks for fixing this!

PR #9933, opened 2021-11-03
Removed PackageReference in favor of PkgReference

lasote:
- NOT migrated to RecipeReference
- NOT migrated ConanReference
- Changed little interfaces with a clearer name, but many more and deeper pending (decide when a function completes a reference/when returns a modified reference)
- There are still many places where copies of references are done and probably can be avoided just by modifying the references.
- We have to revisit all the remote/rest API methods and naming to clean the mess of confusing naming like "prev" being revision str or a complete reference, package_ids containing "#" with revision...


conans/client/cmd/export_pkg.py
```diff
@@ -72,7 +72,7 @@ def export_pkg(app, ref, source_folder, build_folder, package_folder,
         else:
             prev = run_package_method(conanfile, package_id, hook_manager, conan_file_path, ref)
 
-    pref = PackageReference(pref.ref, pref.id, prev)
+    pref.revision = prev
```
memsharded: This is what can fail, because if you change by reference, the ``assign_prev`` doesn't have the old and the new reference to do the move (as the pref.revision also changes layout.pref.revision)

conans/model/package_ref.py
```diff
@@ -44,12 +44,15 @@ def __lt__(self, ref):
...
+        #    at "graph_binaries" to check: cache_latest_prev != pref
+        #    at "installer" to check: if pkg_layout.reference != pref (probably just optimization?)
+        #    at "revisions_test"
+        return self.ref == other.ref and self.revision == other.revision
 
     def __hash__(self):
-        raise Exception("WHO IS COMPARING RECIPE REFERENCES?")
-        # return hash((self.name, self.version, self.user, self.channel, self.revision))
+        # Used in dicts of PkgReferences as keys like the cached nodes in the graph binaries
+        return hash((hash(self.ref), self.package_id, self.revision))
```
memsharded: I think ``hash(self.ref)`` is not needed, hash of a tuple hashes its elements

memsharded (approved):
Fantastic

PR #9945, opened 2021-11-04
[bugfix] PkgConfigDeps and dependencies requires

### LINKED ISSUE #9939: [question] Indirect dependencies missing when using PkgConfigDeps
There's an issue regarding **pkg config** I'm fighting with:

I have libA->libB->libC

- libA requires libB
- libB requires libC

all regular `requires` not `build_requires`

When building libA with the pkg_config generator I get the following generated .pc file for libB
```
prefix=/.conan/data/libB/1.0.0_test/company/testing/package/13ca7c3421ced37746801df275efb2bb7f8fca28
libdir=${prefix}/lib
includedir=${prefix}/include

Name: libB
Description: Library B
Version: 1.0.0_test
Libs: -L"${libdir}" -lds  -lds-tests  -Wl,-rpath,"${libdir}"
Cflags: -I"${includedir}"
Requires: libC
```

After switching to PkgConfigDeps, the generated .pc file changes:
```
prefix=/.conan/data/libB/1.0.0_test/company/testing/package/13ca7c3421ced37746801df275efb2bb7f8fca28
libdir=${prefix}/lib
includedir=${prefix}/include

Name: libB
Description: Library B
Version: 1.0.0_test
Libs: -L"${libdir}" -lds  -lds-tests  -Wl,-rpath,"${libdir}"
Cflags: -I"${includedir}"
```
The **Requires** attribute is missing for the indirect dependency.
Can someone help me out with this?

Thanks very much

franramirez688:
Changelog: Bugfix: `PkgConfigDeps` was not adding correctly the `Requires` for all the package dependencies.
Closes: https://github.com/conan-io/conan/issues/9939
Docs: omit

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -60,12 +60,22 @@ class PkgConfigDeps(object):
         Note: CppInfo could be coming from one Component object instead of the dependency
         """
         ret = []
-        for req in cpp_info.requires:
-            pkg_name, comp_name = req.split("::") if "::" in req else (name, req)
-            # FIXME: it could allow defining requires to not direct dependencies
-            req_conanfile = self._conanfile.dependencies.host[pkg_name]
-            comp_alias_name = get_component_name(req_conanfile, comp_name)
-            ret.append(self._get_pc_name(pkg_name, comp_alias_name))
+        if cpp_info.requires:
+            for req in cpp_info.requires:
+                pkg_name, comp_name = req.split("::") if "::" in req else (name, req)
+                # FIXME: it could allow defining requires to not direct dependencies
+                req_conanfile = self._conanfile.dependencies.host[pkg_name]
+                comp_alias_name = get_component_name(req_conanfile, comp_name)
+                ret.append(self._get_pc_name(pkg_name, comp_alias_name))
+        else:
+            # Let's try to get the transitive dependencies for the given package name
+            try:
+                dep_conanfile = self._conanfile.dependencies.host[name]
+            except KeyError:
+                pass
+            else:
+                ret = [get_package_name(req)
+                       for req in dep_conanfile.dependencies.direct_host.values()]
         return ret
 
     def get_components_files_and_content(self, dep):
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -68,7 +68,7 @@ class PkgConfigDeps(object):
                 comp_alias_name = get_component_name(req_conanfile, comp_name)
                 ret.append(self._get_pc_name(pkg_name, comp_alias_name))
         else:
-            # Let's try to get the transitive dependencies for the given package name
+            # Let's try to get all the requires for the given package name
             try:
                 dep_conanfile = self._conanfile.dependencies.host[name]
             except KeyError:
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -60,12 +60,22 @@ def _get_requires_names(self, name, cpp_info):
         Note: CppInfo could be coming from one Component object instead of the dependency
         """
         ret = []
-        for req in cpp_info.requires:
-            pkg_name, comp_name = req.split("::") if "::" in req else (name, req)
-            # FIXME: it could allow defining requires to not direct dependencies
-            req_conanfile = self._conanfile.dependencies.host[pkg_name]
-            comp_alias_name = get_component_name(req_conanfile, comp_name)
-            ret.append(self._get_pc_name(pkg_name, comp_alias_name))
+        if cpp_info.requires:
```
memsharded: But ``.requires`` is mandatory and should be correctly defined for this to work. Is this an issue that the original package didn't declare ``requires``?

franramirez688: It's mandatory but its value could be `[]`, right?

memsharded: But if it is ``[]`` it means that this component or thing doesn't depend on any transitive things, and as such, the generated ``.pc`` (or .cmake or whatever), will not transitively depend on anything.

franramirez688: Then, if I have this scheme:

```
libA->libB->libC

libA requires libB
libB requires libC
```
why is the result of `cpp_info.requires` equal to `[]` when that `cpp_info` belongs to `libB`?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -60,12 +60,22 @@ def _get_requires_names(self, name, cpp_info):
...
+            for req in cpp_info.requires:
+                pkg_name, comp_name = req.split("::") if "::" in req else (name, req)
+                # FIXME: it could allow defining requires to not direct dependencies
+                req_conanfile = self._conanfile.dependencies.host[pkg_name]
+                comp_alias_name = get_component_name(req_conanfile, comp_name)
+                ret.append(self._get_pc_name(pkg_name, comp_alias_name))
+        else:
```
lasote: I think this `else` shouldn't apply when the `cpp_info` is from a component (called from `comp_requires_names = self._get_requires_names(str(dep.ref.name), comp_cpp_info)`) 
Probably move this `else` code to (replace?) the second call to the function: `requires = self._get_requires_names(str(dep.ref.name), dep.cpp_info)`

lasote (changes requested):

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -61,13 +61,37 @@ def _get_requires_names(self, name, cpp_info):
...
+        dep_name = str(dep.ref.name)
+        # At first, let's check if we have defined some component requires, e.g., "pkg::cmp1"
+        requires = self._get_component_requires_names(dep_name, dep.cpp_info)
+        # If we have found some component requires it would be enough
+        if requires:
+            return requires
+        try:
+            # If no requires were found, let's try to get all the direct
+            # dependencies, e.g., requires = "other_pkg/1.0"
+            dep_conanfile = self._conanfile.dependencies.host[dep_name]
```
lasote: This is weird, probably `dep_conanfile == dep`. Why don't use directly `dep.direct_host.values()`? is that not working?

lasote (approved):

memsharded (approved):

PR #9948, opened 2021-11-04
Feature/develop2 profiles api

memsharded:
New ``profiles`` API


conans/cli/commands/profile.py
```diff
@@ -0,0 +1,131 @@
+from conans.cli.command import conan_command, conan_subcommand, Extender, COMMAND_GROUPS
+from conans.cli.commands import json_formatter
+from conans.cli.output import ConanOutput
+
+
+def profiles_cli_output(profiles):
+    host, build = profiles
+    output = ConanOutput()
```
czoido: Shouldn't this be `cli_out_write` ? I think `ConanOutput` is writting to the **stderr**, not **stdout**.

czoido:
It's not critical, but maybe I would change the UI for `conan profile show`, make the profile positional and add the context as an argument (with default value `host` maybe)?
I find a little bit strange doing something like `conan profile show -pr:h=default`

conans/cli/commands/profile.py
memsharded: Yes, you are right, lets fix it.

memsharded:
>  I find a little bit strange doing something like conan profile show -pr:h=default

The rational of this is to show the profiles that will result from some command line arguments, including something like ``conan profile show -pr=default -s build_type=Debug`` for example. Using exactly the same command line arguments that are passed to other commands.

I also thought about splitting, doing 1 profile either build or host (it is relevant too, might need some arg like ``--build`` to specify), at a time, but the UI seemed way more ad-hoc for this particular command, so I didn't like it much. Lets review it together.

czoido (approved):

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,60 @@
...
+from conans.model.profile import Profile
+from conans.util.files import save
+
+
+class ProfilesAPI:
+
+    def __init__(self, conan_api):
+        self.conan_api = conan_api
+
+    def get_profile(self, profiles=None, settings=None, options=None, conf=None,
```
lasote:
```suggestion
    def combine(self, profiles=None, settings=None, options=None, conf=None,
```

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,60 @@
...
+                    cwd=None, build_profile=False):
+        cache = ClientCache(self.conan_api.cache_folder)
+        loader = ProfileLoader(cache)
+        env = None  # TODO: Not handling environment
+        profile = loader.from_cli_args(profiles, settings, options, env, conf, cwd, build_profile)
+        return profile
+
+    def profile_path(self, profile, cwd=None):
```
lasote:
```suggestion
    def get_path(self, profile, cwd=None):
```

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,60 @@
...
+            for current_directory, _, files in os.walk(profiles_path, followlinks=True):
+                for filename in files:
+                    rel_path = os.path.relpath(os.path.join(current_directory, filename),
+                                               profiles_path)
+                    profiles.append(rel_path)
+
+        profiles.sort()
+        return profiles
+
+    def detect_profile(self, profile_name=None, force=False):
```
lasote:
```suggestion
    def detect(self, profile_name=None, force=False):
```

conans/cli/api/subapi/profiles.py
lasote: I don't like the API having the `profiles` with `None` and then having to say, hey, but this is supposed to be a `build` profile, so read me the build default... too much, road to messy API again.
I would prefer to have a `def get(self, profile_name)`, even a `get_build_default()` and `get_host_default()` and pass explicit argument to a `def combine(self, profiles, settings=None, options=None....)`.
That would probably require to refactor the `loader.from_cli_args` to extract the `calculation` of the default profiles.

lasote: If the `Profile` object is managed in the API (I totally agree) I would move it to the `model.py` (or create a python package if better)

conans/cli/api/subapi/profiles.py
lasote: I would rather have a `def detect(self) -> Profile` method and a `def save(profile, file_name, force=False)`

lasote (changes requested):

conans/cli/api/subapi/profiles.py
memsharded: Agree with the first, reviewed, lets see if now is better.
The second one must wait, current ``Profile`` is too dirty, that requires more work to do, too much here.

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,77 @@
...
+        it loads the "profiles", composing them in order (last profile has priority), and finally
+        adding the individual settings, options (priority over the profiles)
+        """
+        assert profiles and isinstance(profiles, list), "Please provide at least 1 profile"
+        loader = ProfileLoader(self._cache)
+        env = None  # TODO: Not handling environment
+        profile = loader.from_cli_args(profiles, settings, options, env, conf, cwd)
+        return profile
+
+    def get_path(self, profile, cwd=None, exists=True):
```
lasote, line 41: `profile` or `profile_name`?

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def get_default_build(self):
+        """
+        @return: the path to the default "build" profile, either in the cache or as defined by
+        the user in configuration
+        """
+        loader = ProfileLoader(self._cache)
+        return loader.get_default_build()
+
+    def get_profile(self, profiles, settings=None, options=None, conf=None, cwd=None):
```
lasote, line 30: `profiles` or `profiles_paths`?

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,77 @@
...
+from conans.client.profile_loader import ProfileLoader
+from conans.model.profile import Profile
+
+
+class ProfilesAPI:
+
+    def __init__(self, conan_api):
+        self._cache = ClientCache(conan_api.cache_folder)
+
+    def get_default_host(self):
```
lasote, line 14:
```suggestion
    def get_default_host_path(self):
```
?

conans/cli/api/subapi/profiles.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        @return: the path to the default "host" profile, either in the cache or as defined by
+        the user in configuration
+        """
+        loader = ProfileLoader(self._cache)
+        return loader.get_default_host()
+
+    def get_default_build(self):
```
lasote, line 22:
```suggestion
    def get_default_build_path(self):
```
?

lasote (approved):

Issue #9949, opened 2021-11-04
[bug] XcodeDeps generates bad xcconfig variables with dash-cased pkg and armv8 arch


jevarg:
# XcodeDeps generates bad xcconfig variables with dash-cased pkg and armv8 arch

### Environment Details (include every applicable attribute)
  * Conan version: 1.42

### Steps to reproduce (Include if Applicable)
- Create  a conanfile with a requirement on a dash-cased package: `requires =  "test-lib/x.y.z"`
- Set the XcodeDeps generator: `generators = "XcodeDeps"`
- Install armv8 deps `conan install . -s arch=armv8`

### Logs (Executed commands with output) (Include/Attach if Applicable)
Here is the content of one of the generated `.xcconfig` file:
![Screenshot 2021-11-04 at 17 05 46](https://user-images.githubusercontent.com/2744572/140375936-d43f1432-aca3-4922-96ff-02e8b12b62da.png)

Possible fixed naming using snake_case:
![Screenshot 2021-11-04 at 17 08 30](https://user-images.githubusercontent.com/2744572/140376161-1ccc0e4c-920e-4b74-8d05-107d26d8f077.png)

On top of that, if I try to build with Xcode with an arm64 iOS target, those variables will never be used as Xcode uses `arm64` instead of `armv8` that is generated by conan with XcodeDeps.

czoido:
Hi @jevarg 
Thanks a lot for reporting, and also thanks a lot for contributing the fix, we'll review it. to be merged ASAP.

memsharded:
Fixed in #9955, will be released soon in 1.42.1

PR #9950, opened 2021-11-04
Fix/xcodedeps pkg name and arm64

jevarg:
Changelog: (Bugfix): Fixes XcodeDeps bad xcconfig generation when using either dash-case-named packages or `armv8` arch.

Here is the link to the open issue: #9949

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -70,6 +70,10 @@ class XcodeDeps(object):
         self._conanfile = conanfile
         self.configuration = conanfile.settings.get_safe("build_type")
         self.architecture = conanfile.settings.get_safe("arch")
+
+        if self.architecture == "armv8":
+            self.architecture = "arm64" # Xcode uses arm64 instead of armv8
+
         # TODO: check if it makes sense to add a subsetting for sdk version
         #  related to: https://github.com/conan-io/conan/issues/9608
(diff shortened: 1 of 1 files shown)


conan/tools/apple/xcodedeps.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, conanfile):
         self._conanfile = conanfile
         self.configuration = conanfile.settings.get_safe("build_type")
         self.architecture = conanfile.settings.get_safe("arch")
+
+        if self.architecture == "armv8":
+            self.architecture = "arm64" # Xcode uses arm64 instead of armv8
```
czoido, line 75: We already have a conversion method, please check: `conan.tools.apple.apple.to_apple_arch`

czoido:
Thanks a lot for the contribution, I'm pointing this to the release/1.42 branch to be released in 1.42.1 version

czoido:
Hi @jevarg, could you please create a new PR branched from branch release/1.42 so this changes can be release in 1.42.1?

conan/tools/apple/xcodedeps.py
jevarg: Oh thanks for that, will change it

jevarg:
> Hi @jevarg, could you please create a new PR branched from branch release/1.42 so this changes can be release in 1.42.1?

Yep, sorry about the delay, will do it now

PR #9956, opened 2021-11-04
[develop2] Settings refactors

memsharded:
Minor refactors:

- Renamed ``profile.processed_settings`` => ``profile.settings``
- Minor change how settings_yaml is loaded from cache


memsharded:
Closing it, too outdated and conflicted.

It seems that we might still want to think about the ``Profile`` public API, and how the ``processed_settings`` vs ``settings`` are managed, but better start fresh, not worth rescuing this one.

PR #9970, opened 2021-11-05
introducing new version_range model

memsharded:
It is missing the parsing of different version ranges models like ``~1.0`` and ``1.*``, but to see more or less the structure of the solution


czoido (approved):

franramirez688 (approved):

conans/model/recipe_ref.py
```diff
@@ -16,6 +17,27 @@ def __init__(self, value):
         self._pre = None
         self._build = None
 
+    def bump(self, index):
+        if self._items is None:  # the indicator parse is needed is empty items
```
lasote, line 29:
```suggestion
        if self._items is None:  # the indicator parse is needed if empty items
```

conans/model/recipe_ref.py
```diff
@@ -16,6 +17,27 @@ def __init__(self, value):
         self._pre = None
         self._build = None
 
+    def bump(self, index):
```
lasote, line 20:
```suggestion
    def bump(self, index):
        """
           Increments by 1 the version field at the specified index, setting to 0 the fields on the right.
           2.5 => bump(1) => 2.6
           1.5.7 => bump(0) => 2.0.0
        """
```

lasote (approved):
I would add a simple unit test for the bump.
Can be merged.

memsharded:
> I would add a simple unit test for the bump.

I have realized that bumping is not completely clear for prereleases and build, I will check it and add tests

PR #9980, opened 2021-11-08
[Bugfix] `CMakeDeps` and options for linking

### LINKED ISSUE #9936: [bug] CMakeDeps generated .cmake config files doesn't put linker flags in quotation
  * Operating System+version: Ubuntu 16.04
  * Compiler+version: GCC 9.3
  * Conan version: 1.40.1
  * Python version: 3.5.2

Create a package that specifies linker flags, e.g: MyLib

```
def package_info(self):
    self.cpp_info.libs = ["mylib"]
    self.cpp_info.sharedlinkflags = ["-z now", "-z relro"]
    self.cpp_info.exelinkflags = ["-z now", "-z relro"]
```

Then for a package that depends on MyLib we rely on CMakeDeps generator

```
[requires]
mylib/1.0.0

[generators]
CMakeDeps
```

It then has a CMakeLists.txt that uses the library:

```
project(CMakeDepsTest C)

# Prefer -config.cmake files to be ConanDeps friendly
set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
list(APPEND CMAKE_PREFIX_PATH "${CMAKE_BINARY_DIR}")

find_package(MYLIB)

add_executable(mytest main.c)
target_link_libraries(mytest mylib::mylib)
```

When building this test application the linker will fail with errors:
```
/usr/bin/ld: warning: -z --library=now ignored.
/usr/bin/ld: cannot find -lrelro
```

Inspecting the linker command line it's clear that the linker args got passed wrong, i.e. CMake thinks they are a list of libraries to be linked in. The reason for this is that the generated file `mylib-release-x86_64-data.cmake` specifies:
```
set(mylib_SHARED_LINK_FLAGS_RELEASE -z now;-z relro)
```
which is treated as a list "-z;now;-z;relro;". Now this further goes wrong because the linker flags are added as libraries in the file `mylib-Target-release.cmake`:
```
set_property(TARGET intel-mpi::intel-mpi
             $<$<CONFIG:Release>:${intel-mpi_LIBRARIES_TARGETS_RELEASE}
                                           ${intel-mpi_LINKER_FLAGS_RELEASE}> APPEND)
```
and not populating the `INTERFACE_LINK_OPTIONS` property.

The solution is clear:
1) Make sure `mylib_SHARED_LINK_FLAGS_RELEASE` is set as a list of strings or a single string
2) Set linker flags by populating the `INTERFACE_LINK_OPTIONS` property instead of appending them to the `INTERFACE_LINK_LIBRARIES`.

franramirez688:
Changelog: Bugfix: CMakeDeps generated `*-data.cmake` was not including properly the set of link flags.
Changelog: Bugfix: CMakeDeps was not populating `INTERFACE_LINK_OPTIONS` to each target.
Closes: https://github.com/conan-io/conan/issues/9936
Docs: omit


conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_and_linker_flags.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.skipif(platform.system() != "Linux", reason="Only Linux")
+@pytest.mark.tool_cmake
+def test_shared_link_flags():
+    """
+    Testing CMakeDeps and linker flags injection
+
+    Issue: https://github.com/conan-io/conan/issues/9936
+    """
+    conanfile = textwrap.dedent("""
```
memsharded, line 18: The main purpose of ``textwrap.dedent`` is to be able to indent the text (typically 1 tab more than the variable ``conanfile`` that declared it),

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -215,8 +218,8 @@ def join_paths_single_var(values):
         # linker flags without magic: trying to mess with - and / =>
         # https://github.com/conan-io/conan/issues/8811
         # frameworks should be declared with cppinfo.frameworks not "-framework Foundation"
-        self.sharedlinkflags_list = join_flags(";", cpp_info.sharedlinkflags)
-        self.exelinkflags_list = join_flags(";", cpp_info.exelinkflags)
+        self.sharedlinkflags_list = join_flags(";", cpp_info.sharedlinkflags, as_string=True)
```
memsharded: This is a bit confusing. The ``self.xxxx_list`` variable was created to explicitly represent it contains a list of flags, not a unique string. Maybe we want to return to the ``sharedlinkflags`` variable? This was added because some CMake was complaining about receiving a string and it always needed a list, but better double check it.

franramirez688: Yes, but the thing is that it's still a list of flags but passed as a string. I mean, internally, CMake is reading it as a list of flags -> `"oneflag;otherflag;-z flag"`.
I don't know what it was doing before but this is the recommended way to pass it, isn't it? For me, what could be confusing is the parameter `as_string`, any suggestion?

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_and_linker_flags.py
franramirez688: Agree, bad copy&paste from other test 😓

conan/tools/cmake/cmakedeps/templates/target_data.py
lasote: I would remove the `as_string` and quote the final string in the assignation directly.

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -215,8 +218,8 @@ def join_paths_single_var(values):
...
+        self.exelinkflags_list = join_flags(";", cpp_info.exelinkflags, as_string=True)
```
lasote:
```suggestion
        self.exelinkflags_list = '"{}"'.format(join_flags(";", cpp_info.exelinkflags))
```

conan/tools/cmake/cmakedeps/templates/target_data.py
lasote:
```suggestion
        self.sharedlinkflags_list = '"{}"'.format(join_flags(";", cpp_info.sharedlinkflags))
```

lasote (approved):

lasote (approved):

PR #9982, opened 2021-11-08
RecipeReference + Interfaces renaming and improving

lasote:
- Introduced a private `_VersionRepr` at `conans/model/info.py` to manage the package_id modes.
- Removed `ConanFileReference`
- Added `repr_notime()` to both `RecipeReference` and `PackageReference` given that the `repr()` contains the timestamp.
- Renamed methods from proxy to rest_api and returning references when possible.
- A lot of work to do still..


conans/model/requires.py
```diff
@@ -2,7 +2,7 @@
 
 from conans.errors import ConanException
 from conans.model.pkg_type import PackageType
-from conans.model.ref import ConanFileReference
+from conans.model.recipe_ref import RecipeReference
```
memsharded, line 5: This is something that I want to differentiate, between a ``RecipeReference``, which is exact, and a ``RequireReference`` which can contain a version range.

memsharded (approved):
There are a couple of very minor things to keep iterating later, but I think this is already a good move and I vote for merging this asap.

PR #9984, opened 2021-11-08
let CMakeToolchain not crash if build_type not defined

memsharded:
Changelog: Fix: CMakeToolchain will not crash if build_type not defined.
Docs: Omit

@KerstinKeller

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -8,7 +8,7 @@ from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
 from conan.tools._compilers import architecture_flag, use_win_mingw
-from conan.tools.cmake.utils import is_multi_configuration, get_file_name
+from conan.tools.cmake.utils import is_multi_configuration
 from conan.tools.files import save_toolchain_args
 from conan.tools.intel import IntelCC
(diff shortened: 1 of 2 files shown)


lasote (commented):
Missing a bit of context, there is an issue we can link? I assume there is something to discuss (look into label).

memsharded:
> Missing a bit of context, there is an issue we can link? I assume there is something to discuss (look into label).

Yes, this came from slack conversation with user (and Tribe member):

 - They have multi-config packages, containing both Debug and Release
 - They don't use CMakeDeps, because they use cmake_paths to get the config.cmake files from inside the packages
 - The new ``CMakeToolchain`` was raising errors if they don't define the ``build_type`` setting.

Even if the tribe proposal said that ``cpp_info.configs`` is being removed and the plan is that users should migrate to single-config packages, there still will be some users in this situation. It seems that this PR can help them, but probably we want to carefully consider the implications.

KerstinKeller:
The problem can also be fixed by removing the `vs_runtime` block for generation
```
tc.blocks.remove("vs_runtime")
```
For my usecase it is good enough, so I guess, for us, strictly speaking the PR is not neccessary.

lasote (approved):

PR #9990, opened 2021-11-09
- Feature: add backend support to MesonToolchain

### LINKED ISSUE #9922: [question] Why use ninja in meson to replace meson's built-in compilation
Use meson to compile, only support backend to ninja, if I modify it to not support anything else, such as vs, I also need to use msbuild to compile.

The following is the code

```
def build(self, args=None, build_dir=None, targets=None):
    if not self._conanfile.should_build:
        return
    conan_v2_error("build_type setting should be defined.", not self._build_type)
    self._run_ninja_targets(args=args, build_dir=build_dir, targets=targets)
```

If I change the compilation of ninja to meson, it can also be successful, as follows

```
self._run_meson_command(subcommand='compile', args=args, build_dir=build_dir)
```

I omitted the generation target here, meson is supported.

Using ninja also has the problem that the meson file cannot be changed when the meson file is changed.

If I use vs as backend, there will be errors, as follows

```
def _run_ninja_targets(self, args=None, build_dir=None, targets=None):
    if self.backend != "ninja":
        raise ConanException("Build only supported with 'ninja' backend")
```

SSE4:
Changelog: Feature: Add backend support to MesonToolchain.
Docs: https://github.com/conan-io/docs/pull/2295

Closes: #9922 

### DIFF
--- a/conans/test/functional/toolchains/meson/test_backend.py
+++ b/conans/test/functional/toolchains/meson/test_backend.py
@@ -1,0 +1,55 @@
+import os
+import platform
+import sys
+import textwrap
+
+import pytest
+
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.utils.tools import TestClient
+
+@pytest.mark.tool_meson
+@pytest.mark.skipif(sys.version_info.major == 2, reason="Meson not supported in Py2")
+@pytest.mark.skipif(platform.system() != "Windows", reason="requires Windows")
+def test_cross_x86():
+    conanfile_py = textwrap.dedent("""
+        from conans import ConanFile, tools
+        from conan.tools.meson import Meson, MesonToolchain
+
+
+        class App(ConanFile):
+            settings = "os", "arch", "compiler", "build_type"
+            options = {"shared": [True, False], "fPIC": [True, False]}
+            default_options = {"shared": False, "fPIC": True}
+
+            def config_options(self):
+                if self.settings.os == "Windows":
+                    del self.options.fPIC
+
+            def generate(self):
+                tc = MesonToolchain(self, backend='vs')
+                tc.generate()
+
+            def build(self):
+                meson = Meson(self)
+                meson.configure()
+                meson.build()
+        """)
+    meson_build = textwrap.dedent("""
+        project('tutorial', 'cpp')
+        executable('demo', 'main.cpp')
+        """)
+    main_cpp = gen_function_cpp(name="main")
+    client = TestClient()
+    client.save({"conanfile.py": conanfile_py,
+                 "meson.build": meson_build,
+                 "main.cpp": main_cpp})
+    client.run("install .")
+    content = client.load("conan_meson_native.ini")
+    assert "backend = 'vs'" in content
+    client.run("build .")
+    client.run_command(os.path.join("build", "demo"))
+
+    assert "main _M_X64 defined" in client.out
+    assert "main _MSC_VER19" in client.out
+    assert "main _MSVC_LANG2014" in client.out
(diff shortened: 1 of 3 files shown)


lasote:
Could we add a test using the visual backend?

SSE4:
> Could we add a test using the visual backend?

sure, I've added some test

conans/test/functional/toolchains/meson/test_backend.py
```diff
@@ -0,0 +1,55 @@
...
+    client.save({"conanfile.py": conanfile_py,
+                 "meson.build": meson_build,
+                 "main.cpp": main_cpp})
+    client.run("install .")
+    content = client.load("conan_meson_native.ini")
+    assert "backend = 'vs'" in content
+    client.run("build .")
+    client.run_command(os.path.join("build", "demo"))
+
+    assert "main _M_X64 defined" in client.out
```
memsharded: It is very likely that this would be identical irrespective of the backend. If it is, and we cannot test the backend this way, then maybe move this test to "integration" make it faster (no need to compile anything), and just check the generated ``conan_meson_native.ini``?

memsharded (approved):

conans/test/functional/toolchains/meson/test_backend.py
memsharded: Please check this @SSE4 if possible, to get this merged to 1.43

SSE4: I am not sure what do you mean by `we cannot test the backend this way`.
we can test, and it's already tested that way. we build an executable and verify it can be run successfully, and was compiled for the proper architecture.

SSE4: `It is very likely that this would be identical irrespective of the backend.`
I am also not sure about this one. in case of ninja (default), there is `ninja` build generated and built via ninja.
in case of vs backend, it generates vcxproj file and built it via msbuild.
in second case you can use an IDE to open and edit the project. so it's not `identical`.

memsharded: The output of running the compiled executable depends only in the compiler (and the conan settings), it should be the same with either backend. So this test output checks will still pass even if the backend was not ``vs``. The only valid check in this test is the one checking ``backend = "vs"``, and the test can stop there, unless some other check is done, like checking the output of Meson for something like "Using the Visual Studio backend".

SSE4: that only checks valid .ini file was generated. it doesn't check if it builds successfully.
but if you insist, I am moving it.

memsharded: This test is specific about the ``vs`` backend in Meson. The current test checks would still pass if the .ini files contains a useless comment with ``backend = "vs"``. What I am saying is that the test should somehow check that the VS is currently in use, and there is nothing there. The test could be using the Ninja backend and still the exe output will be the same. It would be better if there were some check that check some difference between the Ninja and VS backends, otherwise, building and executing the exe doesn't prove a lot. It proves things, but those things should have been covered by other tests?

SSE4: > What I am saying is that the test should somehow check that the VS is currently in use, and there is nothing there.

I doubt it's worth implementing that check, to be honest. the only way to check is to verify which process is executed during the `meson build` (like `msbuild.exe` or `devenv.exe` or whatever), and this process is launched from another child process of the conan client, so it's tricky to handle that.

memsharded: Restored the check. Added ``assert "Auto detected Visual Studio backend" in client.out`` to the test.
Now, there is some good evidence that Meson is actually using the Visual Studio backend.

memsharded: Sorry, I messed with the branches. I will restore things.

SSE4: okay, it might be enough for our purposes. it only verifies that meson prints something (e.g. claims that it uses Visual Studio backend), but we still don't know does it actually use it. but even if it doesn't, I'd say it's meson bug and not our problem :)

memsharded: Yes, that would be a Meson bug. Our job ends verifying that Meson uses that VS backend, and that check should be enough with it.

PR #9992, opened 2021-11-09
- Bugfix: use new tools.cross_building in MesonToolchain 

SSE4:
Changelog: Bugfix: Use new `tools.cross_building` in MesonToolchain .
Docs: omit

Related: #9922 
see an explanation in comment: https://github.com/conan-io/conan/issues/9922#issuecomment-964109194


lasote (approved):

lasote (commented):
We would need a test using both profiles and using correctly the `conan_meson_cross.ini` (maybe testing the x86 case like the user was trying to do?)

SSE4:
> We would need a test using both profiles and using correctly the `conan_meson_cross.ini` (maybe testing the x86 case like the user was trying to do?)

sure, I have added some test

lasote (approved):

memsharded (approved):

PR #9995, opened 2021-11-09
Remove ConanReference in DB layer

memsharded:
Several design small refactors:
- ClientCache should not contain DB logic (does many pure forwards to DataCache)
- DataCache should not contain DB logic, as it is a wrapper of the pkg cache, managing both paths/layouts and the DB (does many pure forwards to CacheDatabase )
- CacheDatabase is the one that should contain the DB logic

Decisions:

- Timestamp representation: Internally stored as floats. The HTTP server api transfers ISO format
- str() = friendly simple
- repr() = full with everything
- repr_notime() = omit time 
- repr_humantime() = time in readable format


conans/server/revision_list.py
```diff
@@ -15,21 +14,10 @@ def __init__(self):
     @staticmethod
     def loads(contents):
         ret = RevisionList()
-        ret._data = [_RevisionEntry(e["revision"], RevisionList._fix_timestamp(e["time"]))
+        ret._data = [_RevisionEntry(e["revision"], e["time"])
```
memsharded: This doesn't migrate cleanly for 2.0 in the server, it requires a clean server to work.

czoido (approved):

conans/server/revision_list.py
lasote: Probably better to be friendly if doesn't require much code

lasote (changes requested):
Keep server compatible

lasote (approved):

PR #10003, opened 2021-11-11
removing dead new -t files

memsharded:


conans/client/cmd/new.py
```diff
@@ -315,17 +255,11 @@ def _get_files_from_template_dir(template_dir, name, version, package_name, defi
     return out_files
 
 
-def cmd_new(ref, header=False, pure_c=False, test=False, exports_sources=False, bare=False,
+def cmd_new(name, version, header=False, pure_c=False, test=False, exports_sources=False, bare=False,
```
lasote, line 258: So, the `test` argument is useless? and the `-t` from the command also.

memsharded: Yes at the moment it does nothing with templates. Probably we want to keep it and let templates understand the ``-t`` to add or not the ``test_package`` code.

conans/client/cmd/new.py
```diff
@@ -315,17 +255,11 @@ def _get_files_from_template_dir(template_dir, name, version, package_name, defi
...
-        name, version, user, channel, revision = get_reference_fields(ref, user_channel_input=False)
-        # convert "package_name" -> "PackageName"
-        package_name = re.sub(r"(?:^|[\W_])(\w)", lambda x: x.group(1).upper(), name)
-    except ValueError:
-        raise ConanException("Bad parameter, please use full package name,"
-                             "e.g.: MyLib/1.2.3@user/testing")
 
-    # TODO: Validate it is a valid reference
+    # convert "package_name" -> "PackageName"
+    package_name = re.sub(r"(?:^|[\W_])(\w)", lambda x: x.group(1).upper(), name)
```
franramirez688, line 262: Should we have this pattern "compiled" in another common file?

memsharded: It doesn't seem to have enough entity or relevance, quite specific for this use case. Better local than excessive premature generalization.

PR #10005, opened 2021-11-11
Support cmake_target_namespace in legacy cmake_find_package, cmake_find_package_multi generators

### LINKED ISSUE #9946: [question] cmake_target_namespace support in cmake_find_package(_multi) generators?
Do you plan tu support `cmake_target_namespace` property in `cmake_find_package` & `cmake_find_package_multi` generators? Currently it doesn't work (see https://github.com/conan-io/conan-center-index/pull/7890#discussion_r739556016), but it prevents to use this property in CCI since we still need to support those generators.

czoido:
Changelog: omit
Docs: omit

This was reverted by https://github.com/conan-io/conan/pull/10099

Also check: https://github.com/conan-io/conan/pull/10098

Related discussion: https://github.com/conan-io/conan/pull/10077

Closes: https://github.com/conan-io/conan/issues/9946

#TAGS: slow

### DIFF
--- a/conans/client/generators/cmake_find_package.py
+++ b/conans/client/generators/cmake_find_package.py
@@ -227,10 +227,13 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
         for pkg_name, cpp_info in self.deps_build_info.dependencies:
             pkg_filename = self._get_filename(cpp_info)
             pkg_findname = self._get_name(cpp_info)
+            pkg_namespace = self._get_namespace(cpp_info)
+            pkg_namespace = pkg_namespace or pkg_findname
             ret["Find%s.cmake" % pkg_filename] = self._find_for_dep(
                 pkg_name=pkg_name,
                 pkg_findname=pkg_findname,
                 pkg_filename=pkg_filename,
+                pkg_namespace=pkg_namespace,
                 cpp_info=cpp_info
             )
         return ret
@@ -245,7 +248,7 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
             ret.append((comp_genname, deps_cpp_cmake))
         return ret
 
-    def _find_for_dep(self, pkg_name, pkg_findname, pkg_filename, cpp_info):
+    def _find_for_dep(self, pkg_name, pkg_findname, pkg_filename, pkg_namespace, cpp_info):
         # return the content of the FindXXX.cmake file for the package "pkg_name"
         self._validate_components(cpp_info)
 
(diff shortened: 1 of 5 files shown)


conans/model/build_info.py
```diff
@@ -228,6 +228,13 @@ def get_name(self, generator, default_name=True):
         return self.get_property(property_name, generator) \
                or self.names.get(generator, self._name if default_name else None)
 
+    # To provide compatibility for legacy generators with cmake_target_namespace
+    # and make migration easier
+    def get_namespace(self, generator):
+        if generator == "cmake_find_package" or "cmake_find_package_multi":
```
memsharded: This seems a bug, I guess you want to do:
```suggestion
        if generator == "cmake_find_package" or generator == "cmake_find_package_multi":
```
If a bug, check why it didn't fail in tests.

conans/client/generators/cmake_find_package.py
```diff
@@ -76,7 +76,7 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
 
         if({{ pkg_name }}_FIND_COMPONENTS)
             foreach(_FIND_COMPONENT {{ '${'+pkg_name+'_FIND_COMPONENTS}' }})
-                list(FIND {{ pkg_name }}_COMPONENTS "{{ pkg_name }}::${_FIND_COMPONENT}" _index)
+                list(FIND {{ pkg_name }}_COMPONENTS "{{ pkg_namespace }}::${_FIND_COMPONENT}" _index)
```
memsharded: Is there any difference between ``pkg_namespace`` and ``namespace``. In contents, it seems the value is the same one? If yes, then maybe we want to call it the same in the templates?

conans/model/build_info.py
czoido: Yes, sorry 🤦‍♂️
It's ok that it did not fail, this is just to filter who will listen to the property with the old cpp_info. It always entered the line where namespace was assigned but returning None if it was not declared.

conans/client/generators/cmake_find_package.py
czoido: Yes, it's the same, I'll change it

memsharded (approved):

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -289,6 +289,95 @@ def package_info(self):
...
+            settings = "build_type"
+            name = "greetings"
+            version = "1.0"
+            requires = "hello/1.0"
+            def package_info(self):
+                self.cpp_info.components["greetingshello"].requires = ["hello::helloworld"]
+                {}
+        """)
+
+    client.save({"hello.py": hello.format('self.cpp_info.set_property("cmake_target_namespace", "hello_namespace")'),
```
SSE4, line 322: can we also check `cmake_target_namespace` property works on component level?

czoido: From the documentation https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html#properties I think it should not work at component level, just at the root cpp_info level.

danimtb (approved):

SSE4 (approved):

franramirez688 (approved):

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
danimtb: I think that if you want to change the namespace at component level you should use the alias feature right?

SSE4: for sure, you may use an alias. it's just misleading and confusing for me that some properties work on component level, but some don't.

PR #10007, opened 2021-11-12
Remove legacy activation of environment

memsharded:
- Removed Conan integrated cacert. Lets try to run with the Python and system ones, as we agreed after the latest cacert incident
- Moving the ``conans.tools.environment_append`` to internal usage only in ``conans.util.env.environment_update`` with much simpler behavior. Users in ``tools`` should use ``Environment()``
- Stop the injection of conan.conf env-vars to every command invocation via the ``api_method`` decorator
- Moved retry, retry wait to new config. Separated retries for core, upload, download and requests, different things
- New ``config.get(conf_name, type, default)`` helper to simplify access to conf
- Definition of proxies and all network configuration via new conf

Questions:

- Should ``core.download:xxx`` and ``core.upload:xxx`` also be ``core.net.download:xxx``?


conans/cli/api/subapi/__init__.py
```diff
@@ -34,12 +32,7 @@ def wrapper(subapi, *args, **kwargs):
             log_command("{}.{}".format(subapi_name, f.__name__), kwargs)
             # FIXME: Not pretty to instance here a ClientCache
             # FIXME: Remove this when everything is a subapi
```
lasote:
```suggestion

```

lasote (approved):
Just leave me a comment about why you don't consider `COMPRESSION_LEVEL` useful anymore as config and feel free to merge.
Great work!

PR #10008, opened 2021-11-12
fix unescaped double-quotes for defines

Enhex:
Changelog: Bugfix: Fix unescaped double-quotes for defines in ``Premake`` generator.
Docs: Omit

`boost/1.77.0` generates unescaped double-quotes:
```lua
conan_defines = {"BOOST_STACKTRACE_ADDR2LINE_LOCATION="/usr/bin/addr2line"",
```
the Premake generator didn't escape double-quotes for defines, so this PR fixes that.

### DIFF
--- a/conans/client/generators/premake.py
+++ b/conans/client/generators/premake.py
@@ -13,7 +13,7 @@ class PremakeDeps(object):
                                     for p in deps_cpp_info.bin_paths)
         self.libs = ", ".join('"%s"' % p.replace('"', '\\"') for p in deps_cpp_info.libs)
         self.system_libs = ", ".join('"%s"' % p.replace('"', '\\"') for p in deps_cpp_info.system_libs)
-        self.defines = ", ".join('"%s"' % p for p in deps_cpp_info.defines)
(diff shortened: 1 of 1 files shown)


memsharded:
Please recall that ``Premake`` is considered a legacy generator. It needs to be modernized into https://docs.conan.io/en/latest/reference/conanfile/tools.html first in order to be ready to move forward to 2.0

Enhex:
I'm unfamiliar with the Conan 2.0 tools, is there any guide that I can use for creating a Premake one?

memsharded:
> I'm unfamiliar with the Conan 2.0 tools, is there any guide that I can use for creating a Premake one?

The source code for the other generators in the new ``conan.tools.cmake``, (note the singular "conan") and ``conan.tools.microsoft`` etc are the ones to check. There is no explicit guide but:

- They should use ``self.dependencies``, not ``self.deps_cpp_info``
- they don't extend any class, implement ``generate()`` method
- If possible they should generate 1 file (or set of files) per dependency
- If possible they should handle multi-configuration

PR #10014, opened 2021-11-15
Improved: Search + List API and [list+search] commands

lasote:
- Moved filtering packages (by query) to the API (not remote manager, not rest_client, not both)
- New API "references" from the old "list" API (name to be consolidated later when more usage is needed)
- Refactor of the "search" API to be more consistent with the rest, Also naming.
- APIS return references and a couple of new API models.
- The APIS are now more prepared for "conan remove" and "conan upload" (future Prs)
- Refactor the "list command" to avoid using custom dictionaries and manage objects from the API models.
- JSON serialization with some required tricks to manage custom classes.


conans/cli/commands/remove.py
```diff
@@ -0,0 +1,10 @@
+from conans.cli.api.conan_api import ConanAPIV2
+from conans.cli.command import conan_command, COMMAND_GROUPS, Extender
+
+
+@conan_command(group=COMMAND_GROUPS['creator'])
+def remove(conan_api: ConanAPIV2, parser, *args, **kwargs):
```
memsharded, line 6: Not sure if ``remove`` is creator only, it is not that unusual that consumers want to remove installed packages, clean the cache, etc.

conans/client/remote_manager.py
```diff
@@ -199,10 +199,14 @@ def search_recipes(self, remote, pattern, ignorecase=True):
...
 
-    def search_packages(self, remote, ref, query):
-        packages = self._call_remote(remote, "search_packages", ref, query)
-        packages = filter_packages(query, packages)
-        return packages
+    def search_packages(self, remote, ref):
+        infos = self._call_remote(remote, "search_packages", ref)
+        ret = OrderedDict()
+        for package_id, data in infos.items():
+            # FIXME: we don't have the package reference, it uses the latest, we could check
```
memsharded, line 206: Lets try to improve server side if possible.

conans/cli/api/subapi/references.py
```diff
@@ -0,0 +1,61 @@
...
+        if remote:
+            ret = app.remote_manager.get_latest_recipe_reference(ref, remote=remote)
+        else:
+            ret = app.cache.get_latest_recipe_reference(ref)
+
+        return ret
+
+    def latest_package(self, pref: PkgReference, remote: Remote=None) -> PkgReference:
+        assert pref.ref.revision is not None, "latest_package_reference: the recipe revision is None"
+        assert pref.revision is None, "latest_package_reference: "
```
memsharded: Incomplete assertion message

memsharded (approved):
I like it.

Do you have some specific example usage of the "ReferencesAPI" ``latest_recipe()`` and ``latest_package()``?

I still wonder where the ``search_packages`` should be. I liked it in the ``list`` api, because in that way it doesn't imply and real search. Otherwise users think and try to get: "give me all recipes that have a "windows, compiler=msvc" binary available.

conans/client/remote_manager.py
lasote: Already added to the doc of Artifactory improvements.

lasote:
> Do you have some specific example usage of the "ReferencesAPI" latest_recipe() and latest_package()?

Honestly, I still don't. I made them for some reason I cannot recall, probably something in the remove or upload. As I'm not using them yet I'll remove them.

> I still wonder where the search_packages should be. I liked it in the list api, because in that way it doesn't imply and real search. 

I'm thinking about an alternative. Not convinced either but liked it more than in the "list". I think the "query" is a search. Let me try something different.
What I'm sure I don't like is the "list" name because I'm sure we will need things that are not from the "list" command, like the `latest` for example.

memsharded:
> > Do you have some specific example usage of the "ReferencesAPI" latest_recipe() and latest_package()?
> 
> Honestly, I still don't. I made them for some reason I cannot recall, probably something in the remove or upload. As I'm not using them yet I'll remove them.
> 
> > I still wonder where the search_packages should be. I liked it in the list api, because in that way it doesn't imply and real search.
> 
> I'm thinking about an alternative. Not convinced either but liked it more than in the "list". I think the "query" is a search. Let me try something different. What I'm sure I don't like is the "list" name because I'm sure we will need things that are not from the "list" command, like the `latest` for example.

I am completely fine with this. If your gut tells you this is the right approach, lets do it, lets move forward and we can check again when we have more evidence of usage from other install/create, etc cases.

lasote:
Check the latest commit, `references` back to `list` but now containing a `list.packages_configurations` that doesn't have the `query` as an argument. On the other hand a `filter_packages_configurations` in the `search` endpoint accepts the return of the `list.packages_configurations` and a `query`.

We will probably change all this more than 1 more time ;) so as you said, lets move forward and check later.

memsharded (approved):
Yes, I think this is better. Good job, can be merged!

PR #10016, opened 2021-11-15
trying new ConanFile import location from conan import ConanFile

memsharded:
Changelog: Feature: Enable ``from conan import ConanFile`` to prepare for future namespace.
Docs: https://github.com/conan-io/docs/pull/2318

Just to discuss:
- New import?
- New ``class Recipe:`` convention?

### DIFF
--- a/conan/__init__.py
+++ b/conan/__init__.py
@@ -1,0 +1,1 @@
+from conans.model.conan_file import ConanFile
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -5,8 +5,7 @@ from contextlib import contextmanager
 import six
 from six import string_types
 
-from conan.tools.env import Environment
-from conan.tools.env.environment import environment_wrap_command
+
 from conans.client import tools
 from conans.client.output import ScopedOutput
 from conans.client.tools.env import environment_append, no_op, pythonpath
@@ -155,7 +154,7 @@ class ConanFile(object):
         self.compatible_packages = []
         self._conan_using_build_profile = False
         self._conan_requester = None
-
+        from conan.tools.env import Environment
         self.buildenv_info = Environment()
         self.runenv_info = Environment()
         # At the moment only for build_requires, others will be ignored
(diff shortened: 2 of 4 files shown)


lasote:
I would say the new import has to exist, nothing should be left in "conans" BUT as the "opt-in" for possible breaking changes I like the `class Recipe` , it looks more friendly and explicit.

memsharded:
> I would say the new import has to exist, nothing should be left in "conans" BUT as the "opt-in" for possible breaking changes I like the class Recipe , it looks more friendly and explicit.

Then, should we still make the ``from conan import ConanFile`` available? Maybe not, if the naming convention is good enough, I would go for it, as the way forward in 2.0, and avoid 2 different ways of doing the same. My only (minor) concern is a good naming and if that name could have potential conflicts (users having ``Recipe`` generic classes in their python_requires, for example)

Possible names:
- ``class Recipe:``
- ``class ConanRecipe:``
- ``class Conanfile``: # we can play here with the casing, ConanFile vs Conanfile, could work
- Any other idea?

lasote:
As we talked, we could introduce the "class Recipe" as the new way to declare the recipe (without imports) but to keep compatibility, check first that there is no other conanfile subclass just in case we break some user naming the classes as "Recipe"

conans/client/loader.py
```diff
@@ -373,6 +373,8 @@ class defining the Recipe, but also process possible existing generators
                 raise ConanException("More than 1 conanfile in the file")
         elif issubclass(attr, Generator) and attr != Generator:
             generator_manager.add(attr.__name__, attr, custom=True)
+        elif attr.__name__ == "Recipe":
+            result = type('Recipe', (attr, ConanFile, object), {})
```
memsharded: This seems to work, I have run some tests forcing ``GenConanfile`` to always use this way, and the only tests the break seems the ones that change the recipe revision (obviously, conanfile.py is different).

However, this feels a bit like too magic. I am creating a completely new type, ``Recipe`` that derives from the user ``Recipe`` class written in the ``conanfile.py`` (but still a different entity, I am not just injecting the ``ConanFile`` base class).

So not sure what would be the best approach:
- To use explicit ``from conan import ConanFile  & class Pkg(ConanFile)``
- To use magical ``class Recipe:``

PR #10028, opened 2021-11-16
proposing 19.2X versions for msvc

### LINKED ISSUE #10017: [bug] msvc version model might be flawed
The current ``19.3X`` series is represented by ``19.3``, which is lower than ``19.27``, making it broken for comparisons.

cc/ @mpusz

memsharded:
Changelog: Omit
Docs: Omit

Fix: ``msvc`` general 19.1, 19.2, 19.3 versions are generalized as 19.1X, 19.2X, 19.3X. Close https://github.com/conan-io/conan/issues/10017

#tags: slow


memsharded:
cc/ @mpusz (please have a look if possible)

SSE4:
what happens if versions ever go above 100, e.g. `19.111`, should it be considered `19.1X` or `19.11X` or both?
it seems to be another source of confusion.

memsharded:
> what happens if versions ever go above 100, e.g. 19.111, should it be considered 19.1X or 19.11X or both?
it seems to be another source of confusion.

19.1X is the identifier of all possible series of 19.1XYZ, it will work with all. If Microsoft changes their versioning approach, that is a different story anyway, but if they don't the current proposal will keep working.

mpusz:
I think that solves the previous issues and confusion. Thanks @memsharded!

lasote (approved):

Issue #4702, opened 2019-03-08
Use same artifactory credential to download sources in conan recipes


derived-coder:
Conan stores the credentials of artifactory in order to download recipes.
My use case is, that I have source code in artifactory that the conan recipes is downloading via `tools.get`
The artifactory needs also credentials for downloading the stuff. It would be nice, if conan can reuse these credentials in order to get these artifacts.

uilianries:
Hi @derived-coder !

I'm curious, how are you dealing with this?

memsharded:
Yes, this seems to be a valid use case.

It might be a difficult feature to implement, because what conan stores is not the credentials, but stores the user and a JWT token that works with the Conan API. I have no idea if this token could also work for other repository types, so this deserves some investigation first.

uilianries:
I think it can be implemented by Conan client ...

The `tools.get` has the parameter `auth` which can receive CONAN_LOGIN_USERNAME and CONAN_PASSWORD. However, we could pass the conan user instance to re-use both username and password IF the remote address matches, e.g. using the same Artifactory instance for source downloading and package uploading.

We could add a new parameter:

    tools.get(url, sha256, ..., remote_credentials="bincrafters")

I'm saying, to download this source, if an authentication is required, then use the same that for the remote "bincrafters", which was computed by `conan user` command early.

The current option is not far from this approach, but requires a few steps more:

    username = os.getenv("CONAN_LOGIN_USERNAME", "user")
    password = os.getenv("CONAN_PASSWORD", None)
    tools.get(url, sha256, auth=(username, password))

memsharded:
The problem I see with this approach is that there is no guarantee that the access tokens of the Artifactory Conan repo would be valid for another repo, specially if the other repo is another type, like a generic repo. Yes, in many cases the permissions might be good, but security policies in companies might differ. What the solution would look like for this case?

uilianries:
> The problem I see with this approach is that there is no guarantee that the access tokens of the Artifactory Conan repo would be valid for another repo, specially if the other repo is another type, like a generic repo.

I see. As generic repo can't be added as Conan remote, it won't work for this approach.

>  in many cases the permissions might be good, but security policies in companies might differ

Do you mean sniffing passwords? Even `auth` is not safe, which is the regular approach.

> What the solution would look like for this case?

New repo category? Add support for generic repo, for Artifactory and Bintray only, by a new argument:

    conan remote add sources --generic http://192.168.0.24/api/generic/company/sources
[...]

PR #10035, opened 2021-11-17
POC generic remotes

lasote:
- A remote can be added with the same "conan remote add" command but with a `--generic` argument.
- As the `tools.download` (only supported yet) needs access to the remotes I gave access in the recipe to a list at `self.generic_remotes`. <= `discuss`, could be private, could be the cache, could be the remote registry,...
- You can call the download like this: `download(self, "/relative/path/file.txt", "file.txt", remote="foo")`. The URL should be relative to the remote one.
- It is obvious looking now at the remotes API that the "dual" storage of credentials is a mess. Maybe we could store them all just in the plain `json` or all in the database. <= `discuss`. Anyway, I think it is a good approach that should be generalized in any case to have the `username` and `password/token` at the `Remote` model from the API because it simplifies the code and the API.
- My opinion about this POC: If we are going to unify the credentials storage, it is ok, not complex, not very invasive, otherwise I don't think we should introduce this, or we will regret it later because of the mess.

Closes https://github.com/conan-io/conan/issues/4702

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -97,7 +97,7 @@ def ftp_download(conanfile, ip, filename, login='', password=''):
 
 
 def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
-             auth=None, headers=None, md5='', sha1='', sha256=''):
+             auth=None, headers=None, md5='', sha1='', sha256='', remote=None):
     """Retrieves a file from a given URL into a file with a given filename.
        It uses certificates from a list of known verifiers for https downloads,
        but this can be optionally disabled.
@@ -116,8 +116,13 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
     :param md5: MD5 hash code to check the downloaded file
     :param sha1: SHA-1 hash code to check the downloaded file
     :param sha256: SHA-256 hash code to check the downloaded file
+    :param remote: Name of a generic remote configured. If specified, the url should be relative
+                   to the configured remote URL, e.g: "/files/foo.txt".
     :return: None
     """
+    if auth and remote:
+        raise ConanException("The 'auth' parameter cannot be used together if 'remote'")
+
     # TODO: Add all parameters to the new conf
     requester = conanfile._conan_requester
     config = conanfile.conf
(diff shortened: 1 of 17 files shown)


conan/tools/files/files.py
```diff
@@ -135,6 +140,22 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
 
     def _download_file(file_url):
         # The download cache is only used if a checksum is provided, otherwise, a normal download
+        auth = None
+        if remote:
+            #  A generic repository name has been provided, read the remote, complete the URL
+            #  and use the credentials if present
+            generic_remotes = [r for r in conanfile.generic_remotes
```
memsharded, line 147: Maybe we should gather all implementation collaborators (I think the http requester still travels like this too), in something like ``conanfile._conan_things.generic_remotes``?

conans/cli/api/subapi/remotes.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, conan_api):
         self.conan_api = conan_api
 
     @api_method
-    def list(self, pattern=None, only_active=False):
```
memsharded, line 16: Agree with this simplification

conans/cli/conan_app.py
```diff
@@ -52,7 +52,8 @@ def __init__(self, cache_folder):
         self.range_resolver = RangeResolver(self)
 
         self.pyreq_loader = PyRequireLoader(self.proxy, self.range_resolver)
-        self.loader = ConanFileLoader(self.runner, self.pyreq_loader, self.requester)
+        self.loader = ConanFileLoader(self.runner, self.pyreq_loader,
```
memsharded, line 55: This is the thing that would benefit from the ``_conan_collaborators`` mechanism

conans/cli/api/subapi/remotes.py
```diff
@@ -64,20 +62,48 @@ def rename(self, remote: Remote, new_name: str):
...
 
     @api_method
     def login(self, remote: Remote, username, password):
         app = ConanApp(self.conan_api.cache_folder)
-        app.remote_manager.authenticate(remote, username, password)
+        # FIXME: This is weird, probably the storage of users should all be in the json
+        if remote.generic:
+            remote.username = username
+            remote.password = password
+            app.cache.remotes_registry.update(remote)
```
memsharded, line 84: It would be great if for Artifactory generic repos, we could connect, negotiate the token and store only the token, as with Conan repos. Don't know if possible, or too complicated at this point, not critical.

conans/client/loader.py
```diff
@@ -199,6 +200,10 @@ def _initialize_conanfile(conanfile, profile):
         conanfile.initialize(tmp_settings, profile.buildenv)
         conanfile.conf = profile.conf.get_conanfile_conf(ref_str)
 
+        # Initialize the generic remotes
+        if self._remotes_registry:
+            conanfile.generic_remotes = [r for r in self._remotes_registry.list() if r.generic]
```
memsharded, line 205: I'd rather pass here the ``generic_remotes`` already processed in the origin, and not listing and filtering at each conanfile loading.

memsharded (approved):
I think this is overall looking good, and agree with your idea of unifying credentials, it looks necessary. Lets discuss the details.

conans/cli/api/subapi/remotes.py
lasote: But first, of that, we need to choose:
  a)  Need to consider that the "--generic"  is only "artifactory".
  b)  Introduce the `--artifactory-generic` type.

Because now you can use this feature for any remote that accepts basic http.

memsharded: Sure, makes sense. Lets start with the fully generic, then we'll see.

lasote:
We've been talking and we don't see this clear:
- The "remote" name in the recipe couples a lot the recipe to the conan remote configuration.
- Looks like it is more needed a "credentials storage" rather than a "remote generic".
- We decided that is worthing to explore the old idea of the sources cache (that could be applied to conancenter too) and maybe we can manage the credentials there or see if the idea of the "credentials storage" is complementary.

Issue #9565, opened 2021-09-08
[bug] [windows] virtualrunenv: The input line is too long


Artalus:
### Environment Details (include every applicable attribute)
  * Operating System+version: Windows
  * Compiler+version: MSVC 142
  * Conan version: 1.34.1
  * Python version: 3.8

### Steps to reproduce (Include if Applicable)
Use conanfile with lots of dependencies.
<details>
  <summary>*conanfile.txt*</summary>

```
[requires]
glib/2.68.3
openssl/1.1.1i

aeron/1.31.2
ags/6.0
alac/cci.20121212
andreasbuhr-cppcoro/cci.20210113
argtable3/3.2.0
arcus/4.9.1
argtable2/2.13
arrow/2.0.0
amqp-cpp/4.3.11
atk/2.36.0
assimp/5.0.1
aws-c-event-stream/0.1.5
aws-c-common/0.4.25
backward-cpp/1.5
argon2/20190702
baical-p7/5.6
bdwgc/8.0.4
bliss/0.73
aws-sdk-cpp/1.8.130
box2d/2.4.1
aws-checksums/0.1.5
asmjit/cci.20210306
butteraugli/cci.20190319
asyncplusplus/1.1
botan/2.18.1
brotli/1.0.9
boost/1.76.0
capstone/4.0.2
cargs/1.0.0
c-ares/1.17.1
cgns/3.4.1
chipmunk2d/7.0.3
cassandra-cpp-driver/2.15.3
celero/2.6.0
ccfits/2.5
cgltf/1.9
cfitsio/3.490
cmp/19
ceres-solver/2.0.0
cjson/1.7.14
charls/2.1.0
civetweb/1.14
clipper/6.4.2
cppbenchmark/cci.20201029
cnpy/cci.20180601
cppserver/1.0.0.0
cpp-optparse/cci.20171104
corrade/2020.06
cpu_features/0.6.0
cppunit/1.15.1
cppcommon/cci.20201104
cunit/2.1.3
crc32c/1.1.1
cspice/0066
cpprestsdk/2.10.18
cwalk/1.2.5
cryptopp/8.5.0
dav1d/0.8.1
date/3.0.1
dcmtk/3.6.6
duktape/2.5.0
double-conversion/3.1.5
cpr/1.6.2
docopt.cpp/0.6.3
```  
</details>

Install binaries for that conanfile:
```
[...]

PR #10037, opened 2021-11-17
added test that fails with too many deps

memsharded:
Changelog: Omit
Docs: Omit

Reproducing the issue in https://github.com/conan-io/conan/issues/9565


memsharded:
I have tried to fix it by moving to powershell files, but still the same issue.

Do we still want the test as a "negative" test in our suite, to proof that this is raising an error?

franramirez688 (approved):

PR #10039, opened 2021-11-17
Cleaning api.v1 code from client and server, mostly not even tested anymore

lasote:
- Removed v1 from client and server, it is not tested nor used anymore.
- This will make the Conan 2.0 server "only revisions" compatible.
- Cleaned the `user_download` argument at the downloaders that was cleaning the "signature" that we used in "v1".
- IMPORTANT: Artifactory is not returning the capabilities at "v2/ping" so the client is asking "v1/ping" again, because of that, the server has to implement (only) "v1/ping" so the client can speak with it.


memsharded (approved):

lasote:
Don't merge yet, I'm checking some issues.

memsharded (approved):
Great cleanup.
Lets make sure it is not breaking against Artifactory (besides the search-conaninfo.txt parsing already broken)

PR #10043, opened 2021-11-18
[feature] New proposal about `pkg_config_name` usage

franramirez688:


franramirez688:
Using `pkg_config_name` as an `alias` simplifies much more the code and helps us to solve issues about a non-desired usage of the default `scope`. For instance, in the case of `openssl/3.x.x`:
```python
  def package_info(self):
        self.cpp_info.set_property("pkg_config_name", "openssl")
        self.cpp_info.components["crypto"].set_property("pkg_config_name", "libcrypto")
        self.cpp_info.components["ssl"].set_property("pkg_config_name", "libssl")
```
These changes will produce a result like:
```
openssl.pc
libcrypto.pc
libssl.pc
```
Using the current `develop` branch:
```
openssl.pc
openssl-libcrypto.pc
openssl-libssl.pc
```

SSE4:
but I think we have agreed to specify a list, isn't it?

franramirez688:
> but I think we have agreed to specify a list, isn't it?

Yes, I'm working on the changes. I just opened a PR because I had already prepared this approach

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -113,12 +170,19 @@ def content(self):
         pc_files = {}
         host_req = self._conanfile.dependencies.host
         for require, dep in host_req.items():
+            # Restart the aliases cache per dependency
+            self._pkg_config_name_aliases = dict()
```
memsharded, line 174: this sounds a bit fragile. Maybe it would be better if it is not a class attribute, but passed around as argument or returned from methods.

memsharded (approved):
Looks good, though I'd say that the code of PkgConfigDeps is becoming a bit obscure and difficult to follow, but nothing to worry much atm.

conan/tools/gnu/pkgconfigdeps.py
franramirez688: Agree. I will come up with another approach.

franramirez688:
Closing it. New PR opened https://github.com/conan-io/conan/pull/10084

PR #10044, opened 2021-11-18
New install command and API

czoido:
Changelog: Feature: New install command for Conan V2
Changelog: Feature: Add install and graph Conan V2 APIs
Docs: TODO

This is a first step to start migrating things...
Just moving things, updating tests, still lots of things to do and decide. Refine arguments to pass to the API so that they make sense for users and are easy to understand. Moving most of the deps_install to the API but still some repeated code will be there until we migrate all the commands that load the graph or install binaries...
Opening the PR to check to check all test suite, then will refine everything, still very rough.


conans/cli/api/subapi/graph.py
```diff
@@ -0,0 +1,75 @@
...
+        deps_graph.report_graph_error()
+
+        if graph_lock:
+            graph_lock.update_lock(deps_graph)
+
+        return deps_graph
+
+    # should this be an API method?
+    @api_method
+    def get_graph_lock(self, lockfile):
```
memsharded: I think we can safely call things "lockfile" now, everywhere, like ``get_lockfile`` here.

memsharded: No, this doesn't read like a GraphAPI api method

conans/cli/api/subapi/profiles.py
```diff
@@ -26,6 +26,16 @@ def get_default_build(self):
         loader = ProfileLoader(self._cache)
         return loader.get_default_build()
 
+    def get_profiles(self, args):
```
memsharded: This doens't seem very good. This is very cli-args specific, having something in the API layer to parse CLI  args is not a good pattern.

It is totally fine to have reusable pieces in the command layer, that parse args, or add common args. But all the functionality mapping from cli args to domain variables should belong to the command layer, and the API should receive regular variables.

conans/cli/api/subapi/install.py
```diff
@@ -0,0 +1,146 @@
...
+                # The conanfile loaded is a virtual one. The one w deploy is the first level one
+                neighbours = deps_graph.root.neighbors()
+                deploy_conanfile = neighbours[0].conanfile
+                if hasattr(deploy_conanfile, "deploy") and callable(deploy_conanfile.deploy):
+                    run_deploy(deploy_conanfile, install_folder)
+
+        return deps_graph
+
+    # maybe this is not wanted as API method, for the moment just leaving this to not
+    # explode things too much
```
memsharded, line 100: Yes, certainly we don't want this as API method, but it is fine as-is at the moment if it helps to move forward.

conans/client/tools/system_pm.py
```diff
@@ -355,7 +355,7 @@ def update(self):
         self._run("brew update")
 
     def install(self, package_name):
-        self._run("brew install %s" % package_name)
+        self._run("brew install --reference=%s" % package_name)
```
memsharded: search & (wrong replace)?

conans/test/functional/command/config_install_test.py
```diff
@@ -488,7 +488,7 @@ def my_download(obj, url, file_path, **kwargs):  # @UnusedVariable
             self._create_zip(file_path)
 
         with patch.object(FileDownloader, 'download', new=my_download):
-            self.client.run("config install %s" % fake_url_with_credentials)
+            self.client.run("config install --reference=%s" % fake_url_with_credentials)
```
memsharded: Same search & wrong replace

memsharded (approved):
Great!

conans/cli/api/subapi/graph.py
```diff
@@ -0,0 +1,67 @@
...
+from conans.cli.output import ConanOutput
+from conans.errors import ConanException
+
+
+class GraphAPI:
+
+    def __init__(self, conan_api):
+        self.conan_api = conan_api
+
+    # Maybe we don't want to pass Profile objects to the API?
```
lasote, line 12: Yes, we want. I would love to see it (not in this PR) moved to the `conans.cli.api.model` module.

conans/cli/api/subapi/graph.py
```diff
@@ -0,0 +1,67 @@
...
+        @param require_overrides:
+        @param remote_name:
+        @param update:
+        """
+        if path and reference:
+            raise ConanException("Both path and reference arguments were provided. Please provide "
+                                 "only one of them")
+
+        app = ConanApp(self.conan_api.cache_folder)
+        # FIXME: remote_name should be remote
```
lasote: Yes, please, do it so we don't forget. It should be easy, just call  `r = conan_api.remotes.get(args.remote)` before calling this.

conans/cli/api/subapi/graph.py
```diff
@@ -0,0 +1,67 @@
...
+        app.load_remotes([Remote(remote_name, None)], update=update)
+
+        assert profile_host is not None
+        assert profile_build is not None
+
+        if path and reference:
```
lasote, line 43: not to fix it now, because I see that the issue is at the `graph_manager.load_graph(ref_or_path` but we have to keep in mind avoiding API methods with this kind of argument selection.

conans/cli/api/subapi/install.py
```diff
@@ -0,0 +1,143 @@
...
+        @param conanfile_folder: Folder where the conanfile is located
+        @param build_modes:
+        @param generators:
+        @param no_imports:
+        @param create_reference:
+        @param update:
+        @param test:
+        """
+        app = ConanApp(self.conan_api.cache_folder)
+        # FIXME: remote_name should be remote
```
lasote: Idem

conans/cli/commands/install.py
```diff
@@ -0,0 +1,122 @@
...
+    reference. If any requirement is not found in the local cache, it will
+    retrieve the recipe from a remote, looking for it sequentially in the
+    configured remotes. When the recipes have been downloaded it will try
+    to download a binary package matching the specified settings, only from
+    the remote from which the recipe was retrieved. If no binary package is
+    found, it can be built from sources using the '--build' option. When
+    the package is installed, Conan will write the files for the specified
+    generators.
+    """
+    parser.add_argument("path", nargs="?", help="Path to a folder containing a recipe (conanfile.py "
```
lasote, line 60: This is very personal, but I don't like a positional argument being optional. Probably the same "split" we have to do in the API and in the graph collaborators between the "conanfile path" and a "reference" should be done at command level... but I also know that a subcommand here is not nice... so this is gonna be challenging.

As this has to be seen globally with the rest of the commands and we still don't even know how to do the "remove" and the "upload" we can decide later.

conans/cli/commands/install.py
```diff
@@ -0,0 +1,122 @@
...
+                        help='Provide a user')
+    parser.add_argument("--channel", action=OnceArgument,
+                        help='Provide a channel')
+
+    parser.add_argument("--reference", action=OnceArgument,
+                        help='Provide a package reference instead of a conanfile')
+
+    parser.add_argument("-g", "--generator", nargs=1, action=Extender,
+                        help='Generators to use')
+    parser.add_argument("-if", "--install-folder", action=OnceArgument,
```
lasote, line 78: As a note, we should make this install_folder disappear soon.

lasote (approved):
Change if possible the `remote_name` arguments and can be merged. Great work.

PR #10049, opened 2021-11-19
Fix the EnvVar.save_ps1() method

tapia:
Changelog: Fix: Fix the `EnvVar.save_ps1()` method.
Docs: omit

### DIFF
--- a/conans/test/integration/environment/test_env.py
+++ b/conans/test/integration/environment/test_env.py
@@ -491,3 +491,39 @@ def test_multiple_deactivate():
     assert 2 == str(out).count("Restoring environment")
     assert "VAR1=!!" in out
     assert "VAR2=!!" in out
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Path problem in Windows only")
+@pytest.mark.parametrize("num_deps", [3, 60])
+def test_massive_paths(num_deps):
+    client = TestClient()
+    compiler_bat = "@echo off\necho MYTOOL {}!!\n"
+    conanfile = textwrap.dedent("""\
+        from conans import ConanFile
+        class Pkg(ConanFile):
+            exports = "*"
+            def package(self):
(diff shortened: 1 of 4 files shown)


conans/test/unittests/tools/env/test_env.py
```diff
@@ -201,6 +201,7 @@ def test_env_files():
                "MyPath3": "OldPath3",
                "MyPath4": "OldPath4",
                }
+    prevenv.update(dict(os.environ.copy()))
```
tapia: Note for reviewers: some environment variables, like PATH are necessary for the ps1 test. If they don't exist, the powershell.exe command will not be found or will exit with errors

conans/test/integration/environment/test_env.py
```diff
@@ -491,3 +491,39 @@ def generate(self):
     assert 2 == str(out).count("Restoring environment")
     assert "VAR1=!!" in out
     assert "VAR2=!!" in out
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Path problem in Windows only")
+@pytest.mark.parametrize("num_deps", [3, 60])
```
memsharded: Leave this test out of the new PS functionality, this is still not working, and not addressed yet.

conans/test/unittests/tools/env/test_env.py
memsharded: Yes, but this could be affecting the previous test functionality. I think it is better to decouple the PS test then, and leave the previous .bat and .sh without requiring this. Or make it cleaner with a parameterized test.

tapia: Done

conans/test/integration/environment/test_env.py
tapia: Done

memsharded (approved):

PR #10052, opened 2021-11-19
remove legacy Version classes, use new one

memsharded:
The ``Version`` class itself is not modified, simply taken from the ``recipe_ref.py`` and put it ``version.py``, overwriting what was there before.

Next steps:
- Remove the dependency to node-semver, and use our own VersionRange
- Implement a ``RequireReference`` model that can contain a VersionRange


conans/client/tools/oss.py
```diff
@@ -202,69 +202,69 @@ class _OSVERSIONINFOEXW(ctypes.Structure):
...
+        elif version.major == 6:
             return "squeeze"
-        elif version.major() == "5.Y.Z":
+        elif version.major == 5:
             return "lenny"
-        elif version.major() == "4.Y.Z":
+        elif version.major == 4:
             return "etch"
-        elif version.minor() == "3.1.Z":
+        elif version.minor == 1:
```
madebr: Don't you want to compare major version as well?
```suggestion
        elif (version.major, version.minor) == (3, 1):
```

conans/client/tools/oss.py
```diff
@@ -202,69 +202,69 @@ class _OSVERSIONINFOEXW(ctypes.Structure):
...
             return "sarge"
-        elif version.minor() == "3.0.Z":
+        elif version.minor == 0:
```
madebr:
```suggestion
        elif (version.major, version.minor) == (3, 0):
```

conans/client/tools/oss.py
```diff
@@ -202,69 +202,69 @@ class _OSVERSIONINFOEXW(ctypes.Structure):
...
             return "Windows 7"
-        elif version.minor() == "6.2.Z":
+        elif version.major == 6 and version.minor == 2:
             return "Windows 8"
-        elif version.minor() == "6.3.Z":
+        elif version.major == 6 and version.minor == 3:
             return "Windows 8.1"
-        elif version.minor() == "10.0.Z":
+        elif version.major == 10:
             return "Windows 10"
```
madebr, line 235:
```suggestion
        elif version.major == 10:
            return "Windows 10"
        elif version.major == 11:
            return "Windows 11"
```
(or perhaps these suggestions do not belong in this pr)

memsharded:
Yeah, thanks for the suggestions, they are ok. I was just doing some quick migration to the new Version, and took the shortest path that didn't break, but it is certainly better to check majors too, just in case.

conans/client/tools/oss.py
```diff
@@ -202,69 +202,71 @@ class _OSVERSIONINFOEXW(ctypes.Structure):
...
+        elif version.major == 11:
```
madebr, line 236: Looking deeper into this, I'm not sure about this anymore.

memsharded: Not a big deal though. We will have a look to this legacy code here, and probably hide or even remove some of it. I have been having a look at conan-center-index, and it is not very clear what is the usage of the Windows version detection (if the Windows version is relevant, it should probably go in a subsetting, but not automatically detected by a recipe. This is the criteria we are following for 2.0)

conans/client/tools/oss.py
```diff
@@ -275,9 +277,9 @@ def get_freebsd_version():
     def get_solaris_version_name(version):
         if not version:
             return None
-        elif version.minor() == "5.10":
+        elif version.minor == "10":
```
lasote: Shouldn't we keep the major just in case the "6" comes?

lasote (approved):
Great!

conans/client/tools/oss.py
memsharded: Do you really think that a Solaris major version is going to be released? 😄 

And if it is released, still a fix would be necessary to add the new model anyway. But sure, no problem, I can add a check on the major too.

memsharded: Fixed, added 5 major check

PR #10054, opened 2021-11-19
better printing of lockfile/computed option mismatch

rl-enjoyer:
As it stands, if the lockfile locked options dont meet the computed ones, both option configurations are just dumped to stderr in the exception. It makes it hard to parse with your eyeballs in the case where there are a ton of options. With this change, only the difference in options is printed, along with for which configurations the differences are present. So, you just see what was different, and where it was different. 

For example if the lockfile has foo:shared=true and bar:thing=18, and the computed options are foo:shared=true and bar:baz=20, you will see the differences of bar:thing=18 and bar:baz=20 printed first. Then bar:thing=18 since it was the part of the difference in the lockfile. then bar:baz=20 since it was the part of the difference in the computed options. 

This is my first contribution so I really don't know what I'm doing, but I didn't make an issue for it because seems like the PR just does a better job of describing it, and also its small. Also I didn't open a docs issue because I didn't figure anything would need to be documented for this. 

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -399,11 +399,24 @@ class GraphBinariesAnalyzer(object):
             if (node.graph_lock_node is not None and
                     node.graph_lock_node.options is not None and
                     node.conanfile.options.values != node.graph_lock_node.options):
+
+                toset = lambda a: set(str(a).split())
+
+                computed = toset(node.graph_lock_node.options)
+                locked = toset(node.conanfile.options.values)
+
+                difference = computed ^ locked
+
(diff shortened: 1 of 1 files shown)


rl-enjoyer:
Looks like I need to also update conans/test/integration/graph_lock/dynamic_test.py as it checks this exception string

memsharded:
Hi @NolanC33 

Thanks for contributing this.

FYI, we are following a very different approach for lockfiles in Conan 2.0, which is already at alpha stage. In case this is too much work now, maybe you want to know that this will be removed in 2.0, because it will not be necessary.

PR #10058, opened 2021-11-20
[develop2] remove dep to node-semver

memsharded:
- Implemented prerelease management in our own VersionRange
- Removed node-semver dependency, replace usages for our own VersionRange


lasote (approved):
Cool!

PR #10064, opened 2021-11-22
replace legacy non_interactive => core:non_interactive

memsharded:


czoido:
I think this config should also be added to the `DEFAULT_CONFIGURATION` in `conans/model/conf.py`

czoido (approved):

PR #10065, opened 2021-11-23
modern tools.gnu.PkgConfig

memsharded:
Changelog: Feature: Modern ``tools.gnu.PkgConfig`` to supersede legacy ``tools.PkgConfig``. Includes management of `PKG_CONFIG_PATH` and mapping to a ``cpp_info`` structure
Docs: https://github.com/conan-io/docs/pull/2310

I am dropping unused or untested options, abstracting the model, and providing a ``cpp_info`` mapping that seems will be useful in ConanCenter recipes cc/ @madebr @spaceim

### DIFF
--- a/conan/tools/gnu/pkgconfig.py
+++ b/conan/tools/gnu/pkgconfig.py
@@ -1,0 +1,83 @@
+import subprocess
+
+from conan.tools.env import Environment
+from conans.errors import ConanException
+from conans.util.runners import check_output_runner
+
+
+class PkgConfig:
+
+    def __init__(self, conanfile, package, pkg_config_path=None):
+        """
+        :param package: library (package) name, such as libastral
+        """
+        self._conanfile = conanfile
+        self._package = package
+        self._info = dict()
+        self._pkg_config_path = pkg_config_path
+
+    def _parse_output(self, option):
+        executable = self._conanfile.conf["tools.gnu:pkg_config"] or "pkg-config"
+        command = [executable, '--' + option, self._package, '--print-errors']
+        try:
+            env = Environment()
+            if self._pkg_config_path:
+                env.prepend_path("PKG_CONFIG_PATH", self._pkg_config_path)
+            with env.vars(self._conanfile).apply():
+                return check_output_runner(command).strip()
+        except subprocess.CalledProcessError as e:
+            raise ConanException('pkg-config command %s failed with error: %s' % (command, e))
+
+    def _get_option(self, option):
+        if option not in self._info:
+            self._info[option] = self._parse_output(option)
+        return self._info[option]
+
+    @property
+    def includedirs(self):
+        return [include[2:] for include in self._get_option('cflags-only-I').split()]
+
+    @property
+    def cflags(self):
+        return [flag for flag in self._get_option('cflags-only-other').split()
+                if not flag.startswith("-D")]
+
+    @property
+    def defines(self):
+        return [flag for flag in self._get_option('cflags-only-other').split()
+                if flag.startswith("-D")]
+
+    @property
+    def libdirs(self):
+        return [lib[2:] for lib in self._get_option('libs-only-L').split()]
+
+    @property
+    def libs(self):
+        return [lib[2:] for lib in self._get_option('libs-only-l').split()]
+
+    @property
+    def linkflags(self):
+        return self._get_option('libs-only-other').split()
+
+    @property
+    def provides(self):
+        return self._get_option('print-provides')
+
+    @property
+    def version(self):
+        return self._get_option('modversion')
+
+    def cpp_info(self, cpp_info, is_system=True):
+        if not self.provides:
+            raise ConanException("PkgConfig error, '{}' files not available".format(self._package))
+        if is_system:
+            cpp_info.system_libs = self.libs
+        else:
+            cpp_info.libs = self.libs
+        cpp_info.libdirs = self.libdirs
+        cpp_info.sharedlinkflags = self.linkflags
+        cpp_info.exelinkflags = self.linkflags
+        cpp_info.defines = self.defines
+        cpp_info.includedirs = self.includedirs
+        cpp_info.cflags = self.cflags
+        cpp_info.cxxflags = self.cflags
--- a/conan/tools/gnu/pkgconfig.py
+++ b/conan/tools/gnu/pkgconfig.py
@@ -77,13 +77,15 @@ class PkgConfig:
                 self._variables[name] = self._parse_output('variable=%s' % name)
         return self._variables
 
-    def cpp_info(self, cpp_info, is_system=True):
+    def fill_cpp_info(self, cpp_info, is_system=True, system_libs=None):
         if not self.provides:
             raise ConanException("PkgConfig error, '{}' files not available".format(self._package))
         if is_system:
             cpp_info.system_libs = self.libs
         else:
-            cpp_info.libs = self.libs
+            system_libs = system_libs or []
+            cpp_info.libs = [lib for lib in self.libs if lib not in system_libs]
+            cpp_info.system_libs = [lib for lib in self.libs if lib in system_libs]
         cpp_info.libdirs = self.libdirs
         cpp_info.sharedlinkflags = self.linkflags
         cpp_info.exelinkflags = self.linkflags
--- a/conan/tools/gnu/pkgconfig.py
+++ b/conan/tools/gnu/pkgconfig.py
@@ -13,8 +13,9 @@ class PkgConfig:
         """
         self._conanfile = conanfile
         self._package = package
-        self._info = dict()
+        self._info = {}
         self._pkg_config_path = pkg_config_path
+        self._variables = {}
 
     def _parse_output(self, option):
         executable = self._conanfile.conf["tools.gnu:pkg_config"] or "pkg-config"
@@ -66,6 +67,15 @@ class PkgConfig:
     @property
     def version(self):
         return self._get_option('modversion')
+
+    @property
+    def variables(self):
+        if self._variables is None:
+            variable_names = self._parse_output('print-variables').split()
+            self._variables = {}
+            for name in variable_names:
+                self._variables[name] = self._parse_output('variable=%s' % name)
+        return self._variables
 
     def cpp_info(self, cpp_info, is_system=True):
         if not self.provides:
--- a/conan/tools/gnu/pkgconfig.py
+++ b/conan/tools/gnu/pkgconfig.py
@@ -45,7 +45,7 @@ class PkgConfig:
 
     @property
     def defines(self):
-        return [flag for flag in self._get_option('cflags-only-other').split()
+        return [flag[2:] for flag in self._get_option('cflags-only-other').split()
                 if flag.startswith("-D")]
 
     @property
--- a/conan/tools/gnu/pkgconfig.py
+++ b/conan/tools/gnu/pkgconfig.py
@@ -15,7 +15,7 @@ class PkgConfig:
         self._package = package
         self._info = {}
         self._pkg_config_path = pkg_config_path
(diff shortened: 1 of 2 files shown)


conans/test/functional/tools/test_pkg_config.py
```diff
@@ -0,0 +1,61 @@
...
+        assert pkg_config.libdirs == ['/usr/local/lib/libastral']
+        assert pkg_config.linkflags == ['-Wl,--whole-archive']
+        assert pkg_config.variables['prefix'] == '/usr/local'
+
+        cpp_info = NewCppInfo()
+        pkg_config.cpp_info(cpp_info)
+
+        assert cpp_info.includedirs == ['/usr/local/include/libastral']
+        assert cpp_info.defines == ['-D_USE_LIBASTRAL']
+        assert cpp_info.system_libs == ['astral']
```
madebr: The `is_system` argument is a good starter, but feels a bit too too "hard".
Perhaps the method can accept a list argument to differentiate between ordinary and system libraries?
e.g. if a lib is in `["dl", "m", "pthread", "rt"]` then it is a system library, else it is an ordinary lib.

This would require the logic to have access to the current `settings.os`.

conans/test/functional/tools/test_pkg_config.py
```diff
@@ -0,0 +1,61 @@
...
+        pkg_config = PkgConfig(conanfile, "libastral", pkg_config_path=tmp_dir)
+
+        assert pkg_config.provides == "libastral = 6.6.6"
+        assert pkg_config.version == "6.6.6"
+        assert pkg_config.includedirs == ['/usr/local/include/libastral']
+        assert pkg_config.defines == ['-D_USE_LIBASTRAL']
+        assert pkg_config.libs == ['astral']
+        assert pkg_config.libdirs == ['/usr/local/lib/libastral']
+        assert pkg_config.linkflags == ['-Wl,--whole-archive']
+        assert pkg_config.variables['prefix'] == '/usr/local'
```
madebr, line 51: Testing custom variables more profoundly would be useful, imho.

conan/tools/gnu/pkgconfig.py
```diff
@@ -0,0 +1,93 @@
...
+    @property
+    def variables(self):
+        if self._variables is None:
+            variable_names = self._parse_output('print-variables').split()
+            self._variables = {}
+            for name in variable_names:
+                self._variables[name] = self._parse_output('variable=%s' % name)
+        return self._variables
+
+    def cpp_info(self, cpp_info, is_system=True):
```
madebr: The method name is misleading imho.
It feels like this method returns a cpp_info, not fill one in.
Also, this method overwrites a cpp_info object, not append to it.
So why not always return a new one?

conan/tools/gnu/pkgconfig.py
```diff
@@ -0,0 +1,93 @@
...
+from conan.tools.env import Environment
+from conans.errors import ConanException
+from conans.util.runners import check_output_runner
+
+
+class PkgConfig:
+
+    def __init__(self, conanfile, package, pkg_config_path=None):
+        """
+        :param package: library (package) name, such as libastral
```
franramirez688: Perhaps, it's clearer to name it like `reference_name` or even `reference` and perform the split later. I believe that `package` sounds weird to me.

conans/test/functional/tools/test_pkg_config.py
memsharded: Why would need access to ``settings.os``? The new approach has it, because it has access to ``self._conanfile``, as every other new ``conan.tool.xxx``, but not fully clear how the ``settings.os`` could help here.

The argument for filtering some system libs sounds good, I'll try to add it.

conan/tools/gnu/pkgconfig.py
memsharded: This is because it can be complicated to assign it to components specification. Also, this approach works with legacy ``CppInfo`` class and new 2.0 ``NewCppInfo`` class. Returning any of those would probably break the other in the migration.

I can rename it to ``fill_cpp_info``

conans/test/functional/tools/test_pkg_config.py
madebr: e.g. to add `["user32", "shlwapi", "ws2_32", ...]` to system_libs on Windows and `["dl", "m", "pthread", "rt", ...]` on Linux systems.
Maintaining such a system might become a nightmare though.

conan/tools/gnu/pkgconfig.py
madebr: What is the syntax to assign it to components? Or the goal, if it has not been implemented yet?

`fill_cpp_info` sounds good. I'll leave the subjective decisions up to the project lead.

conan/tools/gnu/pkgconfig.py
```diff
@@ -0,0 +1,93 @@
...
+        return [flag for flag in self._get_option('cflags-only-other').split()
+                if flag.startswith("-D")]
+
+    @property
+    def libdirs(self):
+        return [lib[2:] for lib in self._get_option('libs-only-L').split()]
+
+    @property
+    def libs(self):
+        return [lib[2:] for lib in self._get_option('libs-only-l').split()]
```
franramirez688, line 57: I guess that this `.split()` will fail in the case of having paths with spaces. Could you add a test using that?

conans/test/functional/tools/test_pkg_config.py
memsharded: But isn't it just a filter? I have proposed ``pkg_config.fill_cpp_info(cpp_info, system_libs=["user32", "ws2_32", "m", "pthread", "rt", ...]``. I'd prefer not to have this encoded in Conan, we are trying to move away of this kind of hardcoded knowledge of data in helpers, and most of these things should be provided by recipes or conf. 

Because if we do it in Conan, it is always an infinite number of tickets, requesting: "hey, this system library "xxx" is not being correctly assigned to ``system_libs`` and it is returned as ``lib``. And if we add that library, maybe for another user that library will be in a Conan package because they are not depending on the system one, but on the Conan package one, and then we have a blocker.

conan/tools/gnu/pkgconfig.py
memsharded: This should work as:

```python
for name in components_names:
    pkg_config=PkgConfig(self, name)
    pkg_config.fill_cpp_info(self.components[name])
```

Because the ``self.components[name]`` is a ``CppInfo`` compatible object

conans/test/functional/tools/test_pkg_config.py
madebr: Completely agree. A filter is a good middle ground.

conan/tools/gnu/pkgconfig.py
madebr: Thanks. That clears things up.

conan/tools/gnu/pkgconfig.py
madebr: This goes for all `.split()`'s in this class.
Perhaps `shlex.split` can come to the rescue?

memsharded: I think if we agree on the interfaces and general approach and behavior, we can move this PR, and I suggest @franramirez688 doing some extra tests with path with spaces and fix the implementation as necessary (this was apparently not covered by previous functionality either)

franramirez688 (approved):

conan/tools/gnu/pkgconfig.py
franramirez688: Fine for me

conan/tools/gnu/pkgconfig.py
memsharded: Recovering the name ``library``, as it was the original one, aligned with official ``pkg_config`` man docs.

conan/tools/gnu/pkgconfig.py
```diff
@@ -0,0 +1,95 @@
...
+    @property
+    def variables(self):
+        if self._variables is None:
+            variable_names = self._parse_output('print-variables').split()
+            self._variables = {}
+            for name in variable_names:
+                self._variables[name] = self._parse_output('variable=%s' % name)
+        return self._variables
+
+    def fill_cpp_info(self, cpp_info, is_system=True, system_libs=None):
```
SSE4, line 80: I personally don't like a mutating function, it's better if it returns CppInfo constructed from scratch which could be appended to the existing CppIInfo constructing a new object. something like:
```
self.cpp_info = PkgConfig("alsa").cpp_info
self.cpp_info += PkgConfig("gl").cpp_info
self.cpp_info.extend(PkgConfig("drm").cpp_info)
```

conan/tools/gnu/pkgconfig.py
```diff
@@ -0,0 +1,95 @@
...
+        return self._get_option('modversion')
+
+    @property
+    def variables(self):
+        if self._variables is None:
+            variable_names = self._parse_output('print-variables').split()
+            self._variables = {}
+            for name in variable_names:
+                self._variables[name] = self._parse_output('variable=%s' % name)
+        return self._variables
```
SSE4, line 78: is it possible to modernise, so it works like `CppInfo.variables.prefix` instead of `CppInfo.variables["prefix"]`?

conan/tools/gnu/pkgconfig.py
memsharded: Yes, but as stated above, this can't be done at this moment without breaking, because there are different ``CppInfo`` object, cannot return one without breaking.

conans/test/functional/tools/test_pkg_config.py
```diff
@@ -0,0 +1,61 @@
...
+        assert pkg_config.libs == ['astral', 'm']
+        assert pkg_config.libdirs == ['/usr/local/lib/libastral']
+        assert pkg_config.linkflags == ['-Wl,--whole-archive']
+        assert pkg_config.variables['prefix'] == '/usr/local'
+
+        cpp_info = NewCppInfo()
+        pkg_config.fill_cpp_info(cpp_info, is_system=False, system_libs=["m"])
+
+        assert cpp_info.includedirs == ['/usr/local/include/libastral']
+        assert cpp_info.defines == ['-D_USE_LIBASTRAL']
```
SSE4: does it need to have `-D` here?

conan/tools/gnu/pkgconfig.py
memsharded: The ``getattr`` accessors we have in other locations of the codebase are an actual nightmare to debug and maintain. Unless strictly necessary, I won't be adding this syntax to new Conan helpers, and keep the native Python dict [] access

conan/tools/gnu/pkgconfig.py
SSE4: I don't see how is it breaking if it's a completely new tool in a completely namespace 🤔

conans/test/functional/tools/test_pkg_config.py
memsharded: Yes, good point, lets remove it

conan/tools/gnu/pkgconfig.py
memsharded: Because the ``cpp_info`` is not completely new. There is a new wrapper ``NewCppInfo`` that will use the newer integrations and Conan 2.0, but its definition in 1.X is a ``CppInfo`` class, which is a different one. 

And you cannot assign to a component ``CppInfo``, this is the current 1.X definition of Components:
```python
 self.components = DefaultOrderedDict(lambda: Component(self.rootpath,
                                                               self.version, self._default_values))
```

TL:DR: ``PkgConfig`` is new, in a new space. ``cpp_info`` contains lots of legacy, cannot be done.

SSE4: okay :(

PR #10076, opened 2021-11-24
[2.0] New Graph model review, part 1

lasote:
- Reviewed "package_type" tribe feedback: Added `build-scripts` that will help to declare the requirement traits when consumed: `run=False`, `libs=False`, `headers=False`, `visible=False`.
- Introduced Require private fields for the traits, that are kept to `None` so we can know if they have been explicitly declared in the `requirement` declaration or not, so the `package type` can fill the `None` fields only. This way, the consumer always rules.
- If a recipe declares a generic `build_require` now it has a default `run=False`. 
- If a recipe want's to require a build tool it has to use `tool_require` instead that has the `run=True`.
- The `virtualrunenv` generator will only get the `runenv` information from the packages required with `run=True`, so now, by default, if a conanfile `A` declares a `build_require` to `B`, the `runinfo` from `B` will be only used if `B` has a package_type="application" or `B` is shared library package (of course, also if the `build_require` is changed to `tool_require`).
- The `package_id_mode` deductions based on the `package_type` now are done in the `GraphBuilder` and not during the package_id computation, so the graph information is more complete in that previous stage.
- The access to `self.dependencies.get("foo", trait1=foo, trait2=var)` or `self.dependencies["foo"]` are by default getting `build=False` requires. (regular ones).
- The access to `self.dependencies.get("foo", trait1=foo, trait2=var)` will use the specified traits to filter the dependencies, in case there are more than one, it will raise a ConanException showing the found ones. It is not based anymore on the `__eq__` implementation of the `Requirement` class.
- Enabled to declare `build_requires` and `tool_requires` at the `conanfile.requirements` method, not being necessary (but kept for compatibility) the usage of `conanfile.build_requirements` for that.
- Updated `GenConanfile` to generate by default the `build_requirement` and `tool_requirement` at the `requirements` conanfile method.
- Both the `profile` and the `conanfile.txt` don't accept the `build_require` entry anymore, only `tool_require`.


conans/model/requires.py
```diff
@@ -8,28 +8,134 @@
...
+        self._transitive_libs = transitive_libs
+        self._test = test
+        self._package_id_mode = package_id_mode
+        self._force = force
+        self._override = override
+        self._direct = direct
+
+    @staticmethod
+    def _default_if_none(field, default_value):
+        return field if field is not None else default_value
```
memsharded, line 33: There might be a subtle corner case here:
- The propagation is done 2 times, first one it checks for diamonds, previously existing nodes in the graph, with the defined ``requires`` traits, but still without been completed by ``pkg_type``. Second one, after the dependency package has been evaluated and exists, the require is adjusted with the ``pkg_type``, completing Nones
- Previously, if a conflict was detected in this second propagation was because of a runtime (trait ``run``) discrepancy, and the graph was reporting a ``GraphError.runtime`` error.
- Now, this would be generalized for other potential ``None`` values, and the error message could be confusing.

memsharded (approved):
Amazing job!

czoido (approved):

Issue #9943, opened 2021-11-04
[bug] conan 1.42.0: CMakeDeps does not inherit anymore from cmake_find_package_multi names


SpaceIm:
### Environment Details (include every applicable attribute)
  * Operating System+version: macOS Big Sur
  * Compiler+version: Apple Clang 13
  * Conan version: 1.42.0
  * Python version: 3.9.7

### Steps to reproduce (Include if Applicable)

- create a recipe where `self.cpp_info.names["cmake_find_package_multi"]` in `package_info()` has a different name than recipe name (a simple one is `utfcpp` recipe).
- consume this recipe in a project and use `CMakeDeps` generator (you can replace cmake_find_package_multi by CMakeDeps in test_package of utfcpp recipe and run conan create).

=> generated config file and target will have default value (recipe name), so `find_package()` and `target_link_libraries()` might fail.

It has been seen in https://github.com/conan-io/conan-center-index/issues/7887

Workaround: migrate to `set_property()` (which has an unwanted side effect in CONAN_PKG targets of `cmake` generator: https://github.com/conan-io/conan-center-index/issues/7925, so the combination of those 2 issues is really annoying).

memsharded:
Hi @SpaceIm 

yes, this is known, not a bug. ``CMakeDeps`` will not follow any of the legacy methods, not the ``.names`` field, not the ``cmake_find_package_multi`` one. It will only use the new ``set_property()`` mechanism, this is necessary to move forward.

Adding the properties in recipes is the way to go. 
If the ``set_property`` is breaking ``cmake`` generator, then we should have a look at this and try to fix it.

SpaceIm:
I understand, but to move forward in CCI recipes, I would have expected:
- a new CCI hook preventing to use `.names`
- `CMakeDeps` compatibility with `.names["cmake_find_package_multi"]` until conan v2 to ensure smooth and transparent transition of consumers to `CMakeDeps` generator (it will take weeks or months to update CCI recipes).

memsharded:
> CMakeDeps compatibility with .names["cmake_find_package_multi"] until conan v2 to ensure smooth and transparent transition of consumers to CMakeDeps generator (it will take weeks or months to update CCI recipes).

Yes, we tried so far, but the issue is that CMakeDeps have moved internally to use all the new ``dependencies`` and new ``CppInfo`` model, that no longer has ``names`` field, it is simply not there (and can't be easily introduced without extending the legacy into 2.0, which is what is not desired).

[...]

Issue #9000, opened 2021-05-25
[feature] Allow to generate more than one Find/config CMake file per recipe


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Currently, for a given recipe, `cmake_find_package` and `cmake_find_package_multi` can generate one and only one CMake Find/config file, based on the value of `self.cpp_info.filenames["cmake_find_package"]`/`self.cpp_info.filenames["cmake_find_package_multi"]` in `package_info()`.

Unfortunately, some libraries provide several config files for their components. In CCI, at least these recipes can't properly model official config files:

- mongo-c-driver
- mongo-cxx-driver
- spirv-tools
- google-cloud-cpp

A simple solution would be to allow `self.cpp_info.filenames["cmake_find_package"]` and `self.cpp_info.filenames["cmake_find_package_multi"]` to take a tuple, and duplicate the same Find/config files when conan generators are executed. `mongo-cxx-driver` recipe relies on this trick to duplicate the generated config file of `mongo-c-driver` at build time to satisfy its build system without the need to patch it: https://github.com/conan-io/conan-center-index/blob/master/recipes/mongo-cxx-driver/all/conanfile.py#L146-L151

It would also be nice to support this feature in the new generator `CMakeDeps` for conan 2.0

lasote:
Could you help me providing a list of "config file name" + "namespace name" + "target name" that the libraries you mentioned originally generate?

SpaceIm:
@lasote sorry for the late answer

An example is `mongo-c-driver`, it generates these 2 official config files:

- config file 1: `mongoc-1.0-config.cmake`
   target: `mongo::mongoc_shared` if shared else `mongo::mongoc_static`
- config file 2: `bson-1.0-config.cmake`
   target: `mongo::bson_shared` if shared else `mongo::bson_static`

Here is the best we have been able to do in CCI: https://github.com/conan-io/conan-center-index/blob/c3bdc181e09c8cef14320ed628eb147cc7d7905f/recipes/mongo-c-driver/all/conanfile.py#L180-L231

prince-chrismc:
Another example in hiredis https://github.com/redis/hiredis/blob/master/CMakeLists.txt

SpaceIm:
@memsharded @lasote @czoido Do you know if this feature is considered for CMakeDeps?

lasote:
[...]

PR #10077, opened 2021-11-24
Use absolute target names for cmake targets and new properties and deprecating namespace properties

czoido:
Changelog: Feature: Force to use absolute target names with namespaces always via `cmake_target_name`.
Changelog: Feature: Add compatibility logic between legacy `cpp_info.names` and `cmake_target_name`.
Changelog: Feature: Restrict `cmake_target_name` for components to have the same namespace as the root cpp_info.
Changelog: Feature: Deprecate `cmake_target_namespace` and `cmake_module_target_namespace`.
Docs: TODO

This PR tries to make possible a transition from the old syntax for `cpp_info.names` and the new `cmake_target_name` property for `cmake_find_package[multi]` generators. `CMakeDeps` will also now use always absolute names via `set_property`.

There are some supposition when using `cmake_target_name` (maybe we can get rid of them for 2.0...):
- `cmake_target_name` is absolute in the form `NAMESPACE::NAME`, otherwise an exception raises.
- `cmake_target_name` for **components** has to specify a namespace that is always the same as the root cpp_info namespace, otherwise an expection raises.

~~Also, there's one more change, now when specifying components in find_package those have to be named with the full name: `find_package(hello COMPONENTS hello::say)` not `find_package(hello COMPONENTS say)`~~ --> This one is reverted 

#TAGS: slow

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -30,8 +30,6 @@ class ConfigTemplate(CMakeDepsFileTemplate):
                 "version": self.conanfile.ref.version,
                 "file_name": self.file_name,
                 "pkg_name": self.pkg_name,
-                # TODO: maybe we should change this to file_name instead
-                #  using the namespace of the global target
                 "global_namespace": self.global_target_namespace,
                 "config_suffix": self.config_suffix,
                 "check_components_exist": self.cmakedeps.check_components_exist,
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -31,7 +31,6 @@ class ConfigTemplate(CMakeDepsFileTemplate):
                 "file_name": self.file_name,
                 "pkg_name": self.pkg_name,
                 "config_suffix": self.config_suffix,
-                "target_namespace": self.target_namespace,
                 "check_components_exist": self.cmakedeps.check_components_exist,
                 "targets_include_file": targets_include}
 
@@ -76,9 +75,9 @@ class ConfigTemplate(CMakeDepsFileTemplate):
         {% if check_components_exist %}
         # Check that the specified components in the find_package(Foo COMPONENTS x y z) are there
         # This is the variable filled by CMake with the requested components in find_package
-        if({{ target_namespace }}_FIND_COMPONENTS)
-            foreach(_FIND_COMPONENT {{ '${'+target_namespace+'_FIND_COMPONENTS}' }})
-                if (TARGET {{ target_namespace }}::${_FIND_COMPONENT})
+        if({{ file_name }}_FIND_COMPONENTS)
+            foreach(_FIND_COMPONENT {{ '${'+file_name+'_FIND_COMPONENTS}' }})
+                if (TARGET ${_FIND_COMPONENT})
                     conan_message(STATUS "Conan: Component '${_FIND_COMPONENT}' found in package '{{ pkg_name }}'")
                 else()
                     conan_message(FATAL_ERROR "Conan: Component '${_FIND_COMPONENT}' NOT found in package '{{ pkg_name }}'")
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -30,6 +30,9 @@ class ConfigTemplate(CMakeDepsFileTemplate):
                 "version": self.conanfile.ref.version,
                 "file_name": self.file_name,
                 "pkg_name": self.pkg_name,
+                # TODO: maybe we should change this to file_name instead
+                #  using the namespace of the global target
+                "global_namespace": self.global_target_namespace,
                 "config_suffix": self.config_suffix,
                 "check_components_exist": self.cmakedeps.check_components_exist,
                 "targets_include_file": targets_include}
@@ -77,7 +80,7 @@ class ConfigTemplate(CMakeDepsFileTemplate):
         # This is the variable filled by CMake with the requested components in find_package
         if({{ file_name }}_FIND_COMPONENTS)
             foreach(_FIND_COMPONENT {{ '${'+file_name+'_FIND_COMPONENTS}' }})
-                if (TARGET ${_FIND_COMPONENT})
+                if (TARGET {{ global_namespace }}::${_FIND_COMPONENT})
                     conan_message(STATUS "Conan: Component '${_FIND_COMPONENT}' found in package '{{ pkg_name }}'")
                 else()
                     conan_message(FATAL_ERROR "Conan: Component '${_FIND_COMPONENT}' NOT found in package '{{ pkg_name }}'")
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -22,7 +22,6 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
         deps_targets_names = self.get_deps_targets_names() \
             if not self.conanfile.is_build_context else []
         return {"pkg_name": self.pkg_name,
-                "target_namespace": self.target_namespace,
                 "global_target_name": self.global_target_name,
                 "config_suffix": self.config_suffix,
                 "deps_targets_names": ";".join(deps_targets_names),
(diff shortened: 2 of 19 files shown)


conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -76,9 +75,9 @@ def template(self):
         {% if check_components_exist %}
         # Check that the specified components in the find_package(Foo COMPONENTS x y z) are there
         # This is the variable filled by CMake with the requested components in find_package
-        if({{ target_namespace }}_FIND_COMPONENTS)
-            foreach(_FIND_COMPONENT {{ '${'+target_namespace+'_FIND_COMPONENTS}' }})
-                if (TARGET {{ target_namespace }}::${_FIND_COMPONENT})
+        if({{ file_name }}_FIND_COMPONENTS)
+            foreach(_FIND_COMPONENT {{ '${'+file_name+'_FIND_COMPONENTS}' }})
```
czoido, line 80: This was a bug in CMakeDeps?

conans/client/installer.py
```diff
@@ -672,6 +672,10 @@ def _call_package_info(self, conanfile, package_folder, ref, is_editable):
 
                     conanfile.package_info()
 
+                    # this is necessary to check if the components declared for legacy generators
+                    # have the same namespace as the root cpp_info
+                    conanfile.cpp_info._raise_incorrect_components_names()
```
czoido, line 677: This should not be necessary for 2.0, with CMakeDeps being free to specify any target name for components without restrictions.

conans/test/functional/generators/components/test_components_cmake_generators.py
```diff
@@ -57,16 +57,10 @@ def package_info(self):
                     self.cpp_info.filenames["cmake_find_package"] = "MYG"
                     self.cpp_info.set_property("cmake_file_name", "MYG")
 
-                    self.cpp_info.names["cmake_find_package_multi"] = "MyGreetings"
```
memsharded, line 60: Just to make sure, the ``.names`` approach doesn't dissapear for 1.X and ``cmake_find_package[multi]`` right? Then maybe it is better to leave the test as it was, just to make sure we are not breaking the whole ConanCenter that relies on ``.names``?

memsharded: Maybe just leaving the test as-is, and adding a new, duplicated one with ``cmake_target_name``?

SSE4:
can you explain the reason(s) for the following restrictions? 

> cmake_target_name is absolute in the form NAMESPACE::NAME, otherwise an exception raises.

I think we already know real cases where it shouldn't be in a form `NAMESPACE::NAME`, so I wouldn't set this restriction. it's actually pretty common to have targets without namespaces (e.g. sqlite3). we already know libraries which use other delimiters than colon (e.g. SFML).

> cmake_target_name for components has to specify a namespace that is always the same as the root cpp_info namespace, otherwise an expection raises.

we also know cases where components may use different namespaces (e.g. OpenEXR).

> find_package(hello COMPONENTS hello::say) not find_package(hello COMPONENTS say)

I don't think it should be that way. we already know for transparent integration we need `find_package(Boost COMPONENTS thread)`, and not `find_package(Boost COMPONENTS Boost::thread)`. also as I understand component name doesn't necessarily match target name (SFML case), so it might deserve its own property (with a reasonable default).

/cc @madebr @Croydon @SpaceIM @prince-chrismc @ericLemanissier 

P.S. check also these comments, they might be essential to ensure we're on the same page:
https://github.com/conan-io/conan/pull/8533#issuecomment-972860669
https://github.com/conan-io/conan/pull/8533#issuecomment-972867610
https://github.com/conan-io/conan/issues/9943#issuecomment-977282813

conan/tools/cmake/cmakedeps/templates/config.py
SSE4: what's the bug?

SpaceIm:
Yes, basically we should be able to define:

- a name for the config file (default to recipename). It would be nice to be able to generate several config files by the way (see https://github.com/conan-io/conan/issues/9000).
- a global imported target name (default to recipename::recipename). It can be anything, even names without `::`.
- per component:
  - an imported target name (default to recipename::componentname). Same than global target name, it can be anything.
  - a component name for COMPONENTS part of find_package() (I would say default value should be the last part of imported target name split by `::`, so if there is no `::` it's just the imported target name). There should be a way to override this value (see `SFML` and `OpenCV` libraries, component names are different than associated imported targets).
     The global target may also have this "component" property, not sure.

Please no more tricks with `build_modules` we have in few CCI recipes, `markdown` generator is not able to understand those workarounds, and doesn't display the proper way to inject imported targets without namespace.

>  it's actually pretty common to have targets without namespaces (e.g. sqlite3)

sqlite3 imported target is namespaced.

czoido:
> can you explain the reason(s) for the following restrictions?
> 
> > cmake_target_name is absolute in the form NAMESPACE::NAME, otherwise an exception raises.
> 

This is just a first proposal. Legacy generators are very coupled with the namespace and I did not want to change all the internal code of the legacy generators because I consider it a bit dangerous. Because of that, I made some assumptions for this PR. It makes a bit easier to give compatibility between names and set_property. 
In the future, for `CMakeDeps` this restriction could be removed as it has not to be compatible with the old .names and also I have decoupled the code it generates from the namespace concept. Also, I did not want to introduce in the PR things you could not do with the `.names` model. I think that if we want more power maybe that should be done when we migrate to `CMakeDeps`...

> I think we already know real cases where it shouldn't be in a form `NAMESPACE::NAME`, so I wouldn't set this restriction. it's actually pretty common to have targets without namespaces (e.g. sqlite3). we already know libraries which use other delimiters than colon (e.g. SFML).
> 
> > cmake_target_name for components has to specify a namespace that is always the same as the root cpp_info namespace, otherwise an expection raises.
> 
> we also know cases where components may use different namespaces (e.g. OpenEXR).
> 
> > find_package(hello COMPONENTS hello::say) not find_package(hello COMPONENTS say)
> 

Yes, that will probably be reverted.

> /cc @madebr @Croydon @SpaceIm @prince-chrismc @ericLemanissier
> 
> P.S. check also these comments, they might be essential to ensure we're on the same page: [#8533 (comment)](https://github.com/conan-io/conan/pull/8533#issuecomment-972860669) [#8533 (comment)](https://github.com/conan-io/conan/pull/8533#issuecomment-972867610) [#9943 (comment)](https://github.com/conan-io/conan/issues/9943#issuecomment-977282813)

We are discussing things about this PR and how to make finally the migration to set_property so probably this PR will evolve as we discuss them...

SSE4:
> sqlite3 imported target is namespaced.

yes, sorry, my bad. sqlite3 definetely doesn't use that, it just defines some CMake variables. the list of projects using targets without namespaces is available here: https://github.com/conan-io/conan/issues/7615. it contains popular libraries like `jsoncpp` amongst others.

conan/tools/cmake/cmakedeps/templates/config.py
czoido: That CMake generates a `<PackageName>_FIND_COMPONENTS` when `COMPONENTS` are specified in the `find_package`, but the `<PackageName>`!=`target_namespace`, it is `file_name`

conans/client/generators/cmake_find_package.py
```diff
@@ -74,8 +74,8 @@ class CMakeFindPackageGenerator(GeneratorComponentsMixin, Generator):
 
         set({{ pkg_name }}_COMPONENTS {{ pkg_components }})
 
-        if({{ pkg_name }}_FIND_COMPONENTS)
-            foreach(_FIND_COMPONENT {{ '${'+pkg_name+'_FIND_COMPONENTS}' }})
+        if({{ pkg_filename }}_FIND_COMPONENTS)
+            foreach(_FIND_COMPONENT {{ '${'+pkg_filename+'_FIND_COMPONENTS}' }})
```
czoido, line 78: Same situation as https://github.com/conan-io/conan/pull/10077/files#r756964886 this was not working if cpp_info.filename was different from names

conans/client/generators/cmake_find_package_multi.py
```diff
@@ -184,8 +184,8 @@ class CMakeFindPackageMultiGenerator(CMakeFindPackageGenerator):
             include(${f})
         endforeach()
 
-        if({{ pkg_name }}_FIND_COMPONENTS)
-            foreach(_FIND_COMPONENT {{ '${'+pkg_name+'_FIND_COMPONENTS}' }})
+        if({{ pkg_filename }}_FIND_COMPONENTS)
+            foreach(_FIND_COMPONENT {{ '${'+pkg_filename+'_FIND_COMPONENTS}' }})
```
czoido, line 188: Same situation as https://github.com/conan-io/conan/pull/10077/files#r756964886 this was not working if cpp_info.filename was different from names

conans/test/functional/generators/components/test_components_cmake_generators.py
```diff
@@ -500,15 +499,19 @@ def package_info(self):
         assert ("Component 'greetings::non-existent' not found in 'greetings' "
                 "package requirement" in client.out)
 
-    @pytest.mark.parametrize("generator", ["cmake_find_package_multi", "cmake_find_package"])
-    def test_component_not_found_cmake(self, generator):
+    @pytest.mark.parametrize("generator, filename", [("cmake_find_package_multi", "greetings"),
```
czoido, line 502: I have extended this test to cover cases where filename is not the same as the name of the package, because it was failing for those cases: https://github.com/conan-io/conan/pull/10077/files#r757272609

conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -76,9 +76,9 @@ def template(self):
...
+                if (TARGET {{ global_namespace }}::${_FIND_COMPONENT})
```
lasote, line 81: But this is only `True` if the `cmake_target_name` property for the component has a `namespace` equal to the global `cppinfo` `cmake_target_name`  property. Right? Maybe we are blocking it now to be different (to discuss) but what about later? how could you find the correct target for the component if the input is "the filename" and the "suffix" of the target name? 

Given that the `find_package(Foo COMPONENTS bar1 bar2)` is almost useless here, because the targets are already defined, and it is only used to check if the components are declared... Should we just assume that the call to find_package should be: `find_package(Foo COMPONENTS namespacebar1::bar1 namespacebar2::bar2)` ?

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -153,7 +152,9 @@ def get_required_components_names(self):
         ret = []
         sorted_comps = self.conanfile.cpp_info.get_sorted_components()
         for comp_name, comp in sorted_comps.items():
-            ret.append(self.get_component_alias(self.conanfile, comp_name))
```
lasote, line 156: This is not from this PR but I find the name `get_required_components_names` very confusing (and I named it!) it should be something more like `get_declared_components_names`, right? About the `variable_name`, I found it a bit hidden, couldn't the replacement be done in the template directly?

czoido:
After having a long conversation about this topic with the team, it's clear the current strategy is a mistake and that we have to adopt a new strategy to successfully migrate to the properties model without being forced to add too many restrictions fearing breaking lots of users. 

It's challenging to make a smooth transition to the new properties model, maintaining compatibility between the **legacy CMake generators** (that can listen to `cpp_info.names` and `set_property`) and the **new CMakeDeps** generator at the same time. On one side, we have a high risk of breaking users, and on the other, we are restricting the introduction of new features that will not be compatible with legacy generators.

We have concluded that it would be better (less breaking, while allowing new powerful features in new integrations) for the community to move forward if:

- We make the models for the legacy generators and the new ones completely independent. That means that _cmake_find_package_ and _cmake_find_package_multi_ will completely stop listening to properties, and only listen to `.names` attribute, as they have been doing since the beginning, and that is still what the vast majority of recipes in conan-center-index do (around 650 recipes). This implies having to recover the `.names` attribute in about 30 recipes in c3i that removed this attribute, but will allow to not break `cmake_find_package` generators at all in the migration

- CMakeDeps will **only** listen to `set_property` (mostly `"cmake_target_name"`), that way we will have more freedom to adopt the new feature requests for the model without being stopped by the possibility of breaking lots of users.

That means that both `.names` and `set_property` will coexist in the recipes until we decide to remove `.names` at some point **after** Conan 2.0. Legacy generators' behaviour will be dictated only by the old ``.names`` model and CMakeDeps only by the `set_property` model. 

So the plan for the migration is first recovering the `.names` attributes in the recipes that removed it, leaving the `set_property` that were added there (although when we release conan 1.43 cmake_target_name must be switched to absolute targets). Then, after Conan 1.43 is released, recipes will add `set_property` gradually and we, in the Conan team, will try to make an effort to migrate as many recipes as we can.

We are sorry we have to revert some changes we recommended to do, it is clear that this way was a mistake and is being painful for the community and we believe this new plan will make the migration easier and faster.

/cc @SSE4 @madebr @Croydon @SpaceIm @prince-chrismc @ericLemanissier

czoido:
Just one note about `pkg_config` legacy generator. For this generator we maintain the set_property compatibility so for packages that set the cpp_info for this generator there's no need to bring back properties or modify the `set_property` calls if they are already there.

czoido:
This has been superseded by: https://github.com/conan-io/conan/pull/10098 + https://github.com/conan-io/conan/pull/10099

PR #10083, opened 2021-11-25
Check some required settings in msvc_compatible()

tapia:


conans/model/info.py
```diff
@@ -585,6 +586,11 @@ def msvc_compatible(self):
         runtime = compatible.settings.compiler.runtime
         runtime_type = compatible.settings.compiler.runtime_type
 
+        if version is None:
+            raise undefined_value('settings.compiler.version')
+        if runtime is None:
+            raise undefined_value('settings.compiler.runtime')
```
franramirez688, line 592: Shouldn't it be checked in an upper layer? I mean, it sounds like if they are mandatory settings so it should raise an error before this compatibility check, shouldn't it?

memsharded: That is the thing, they are not always mandatory. Packages can actually skip defining some of them, and cleaning them in ``package_id()`` will still map to a valid binary package. This is the reason the check has been delayed, to improve in the definition of "what cannot be built with this configuration", and "no compatible binary exist for this configuration"


Put in other words: For consuming an app-only package, the ``os`` and ``arch`` should be more than enough. Lets say that recipe ``package_id`` removes ``compiler`` and ``build_type`` in the ``package_id()``, but keeps them as inputs in order to be able to build the package. Then it should only fail if the build() is invoked.

conans/model/info.py
```diff
@@ -585,6 +586,11 @@ def msvc_compatible(self):
         runtime = compatible.settings.compiler.runtime
         runtime_type = compatible.settings.compiler.runtime_type
 
+        if version is None:
```
memsharded, line 589: An accompanying test for this fix would be good.

PR #10084, opened 2021-11-25
Refactored `conan.tools.gnu.pkgconfigdeps` module and added new `pkg_config_aliases` property

### LINKED ISSUE #9848: [feature] introduce alias for PkgConfigDeps

franramirez688:
Changelog: Feature: `pkg_config_name` is used as the main name for a package/component and it will be used as the name for the `*.pc` file.
Changelog: Feature: Added new property`pkg_config_aliases` which admits a  list of strings to define different aliases for any package/component.
Docs: https://github.com/conan-io/docs/pull/2315

Closes: #9848


lasote (approved):
Looks very good! Just a question, all the "wrapper" terms in the PR are referred to "alias", right? Might be clearer, in that case, to call it "alias" instead of "wrapper"?

conan/tools/gnu/pkgconfigdeps/pc_files_creator.py
```diff
@@ -0,0 +1,76 @@
+from conan.tools.gnu.pkgconfigdeps.pc_files_templates import PCFilesTemplate
+from conan.tools.gnu.pkgconfigdeps.pc_info_loader import PCInfoLoader
+
+
+class PCFilesCreator(object):
+
+    def __init__(self, conanfile, dep):
+        self._conanfile = conanfile
```
memsharded: This conanfile seems to be the "consumer" conanfile. And it is only used here for the ``description`` field. Shouldn't the description field be from the actual ``dep``? Maybe it is even possible to remove it completely, and have the ``description`` almost empty or automated to something like "Conan package: xxxx"

conan/tools/gnu/pkgconfigdeps/pc_info_loader.py
```diff
@@ -0,0 +1,139 @@
...
+        a CppInfo object.
+
+        Note: CppInfo could be coming from one Component object instead of the dependency
+        """
+        ret = []
+        for req in cpp_info.requires:
+            pkg_ref_name, comp_ref_name = req.split("::") if "::" in req \
+                                                          else (self._pkg_ref_name, req)
+            # FIXME: it could allow defining requires to not direct dependencies
+            req_conanfile = self._conanfile.dependencies.host[pkg_ref_name]
```
memsharded: Are you sure these are the right dependencies? Shouldn't it be the ``dep`` dependencies, instead of the root consumer ``conanfile.py``?

memsharded (commented):
A bit too unnecessarily OO for my taste, but looking much cleaner, so great job!

Please have a look to the comments, I have a couple of concerns, if it happens that I am right, then the thing is that the tests are not comprehensive enough, and maybe more tests with transitive deps are needed.

franramirez688:
@memsharded 

I have changed the model to a more functional one, I think it's even clearer now.

lasote (approved):

memsharded (approved):
Even better, good job!

PR #10090, opened 2021-11-26
[2.0] add Premake support

Enhex:
Changelog: Omit
Docs: Omit

This adds some basics for modern Premake tools, but still missing testing to check everything works, so lets wait until next release to document it.

no documentation so it's been guesswork, let me know if anything is missing.

### DIFF
--- a/conan/tools/premake/premake.py
+++ b/conan/tools/premake/premake.py
@@ -15,7 +15,7 @@ class Premake(object):
                         '16': '2019'}
             premake_command = "premake5 vs%s" % _visuals.get(str(self.settings.compiler.version), "UnknownVersion %s" % str(self.settings.compiler.version))
             self.run(premake_command)
-        else if "msvc" in self.settings.compiler:
+        elif "msvc" in self.settings.compiler:
             _visuals = {'14.0': '2005',
                         '15.0': '2008',
                         '16.0': '2010',
--- a/conan/tools/premake/premake.py
+++ b/conan/tools/premake/premake.py
@@ -15,5 +15,25 @@ class Premake(object):
                         '16': '2019'}
             premake_command = "premake5 vs%s" % _visuals.get(str(self.settings.compiler.version), "UnknownVersion %s" % str(self.settings.compiler.version))
             self.run(premake_command)
+        else if "msvc" in self.settings.compiler:
+            _visuals = {'14.0': '2005',
+                        '15.0': '2008',
+                        '16.0': '2010',
+                        '17.0': '2012',
+                        '18.0': '2013',
+                        '19.0': '2015',
+                        # add non-trailing 0 variation manually
+                        '19.1': '2017',
+                        '19.2': '2019'}
+            # generate VS2017 versions
+            for i in range(0,7):
+                ver = '19.1' + str(i)
+                _visuals[ver] = '2017'
+            # generate VS2019 versions
+            for i in range(0,10):
+                ver = '19.2' + str(i)
+                _visuals[ver] = '2019'
+            premake_command = "premake5 vs%s" % _visuals.get(str(self.settings.compiler.version), "UnknownVersion %s" % str(self.settings.compiler.version))
+            self.run(premake_command)
         else:
             self.run("premake5 gmake2")
--- a/conan/tools/premake/premake.py
+++ b/conan/tools/premake/premake.py
@@ -1,0 +1,19 @@
+class Premake(object):
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+
+    # automatically chooses premake action based on used compiler
+    def configure(self):
+        if "Visual Studio" in self.settings.compiler:
+            _visuals = {'8': '2005',
+                        '9': '2008',
+                        '10': '2010',
+                        '11': '2012',
+                        '12': '2013',
+                        '14': '2015',
+                        '15': '2017',
+                        '16': '2019'}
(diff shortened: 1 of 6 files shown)


conan/tools/premake/premake.py
```diff
@@ -0,0 +1,39 @@
+class Premake(object):
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+
+    # automatically chooses premake action based on used compiler
+    def configure(self):
+        if "Visual Studio" in self.settings.compiler:
```
memsharded, line 7: This is typically done in ``PremakeToolchain`` that is called at ``generate()`` time.
This is the feature that allows using ``Premake`` as developer too ``conan install`` + ``premake``...

Good enough for this first iteration, but something to do before making this integration stable.

conan/tools/premake/premakedeps.py
```diff
@@ -0,0 +1,106 @@
...
+                ['conan_build_type = "{0}"'.format(str(self._conanfile.settings.build_type)),
+                'conan_arch = "{0}"'.format(str(self._conanfile.settings.get_safe("arch"))),
+                ""]
+        )
+
+        deps = PremakeTemplate(self._get_cpp_info())
+        all_flags = template.format(dep="", deps=deps)
+        sections.append(all_flags)
+        sections.append(
+            "function conan_basic_setup()\n"
```
memsharded, line 74: If there is some mechanism in Premake that allows injecting this on the command line, instead of later requiring to modify user build scripts, should be leveraged.

memsharded (commented):
Some things to keep working on, but not necessarily right now, can be done in future PRs (maybe just add FIXME comments in code in the sections.

What would be necessary is at least some integration test that validates the files are correctly generated. Please tell if you need help with this. Thanks!

conan/tools/premake/premakedeps.py
Enhex: from what I know it Premake doesn't have a feature for injecting code.

conan/tools/premake/premake.py
Enhex: from what i understood `[tool]Toolchain` is for setting up configuration and `[tool]` is for helper functions to call the tool from python.
unlike CMake Premake doesn't have separate "configure" and "generate" steps, and it takes the target build system as a command line argument when called.
so it kinda falls into both categories of config in the sense of choosing build system and helper for calling premake.
I think it makes more sense for it to belong with the helper because that's the way premake is intended to be used (it doesn't look for config files when called).

if the user want to manually choose the build system he can manually craft a premake command instead of using the helper.

Enhex:
I couldn't get pytest to run on my machine (getting 100's of errors) so guesswork is involved and i might got things wrong.

conan/tools/premake/premake.py
memsharded: Understood, yes, I assumed by a very quick read that it was using something like vcvars, but it is not the case. Looking good.

conan/tools/premake/premakedeps.py
```diff
@@ -0,0 +1,106 @@
...
+        self._conanfile = conanfile
+
+    def generate(self):
+        # Current directory is the generators_folder
+        generator_files = self.content
+        for generator_file, content in generator_files.items():
+            save(generator_file, content)
+
+    def _get_cpp_info(self):
+        ret = CppInfo('','')
```
memsharded: This seems it could be better a ``NewCppInfo()``, that doesn't require args. That is the helper that we use for new integrations to clean things for 2.0

conan/tools/premake/premakedeps.py
```diff
@@ -0,0 +1,106 @@
...
+            "    links{conan_frameworks}\n"
+            "    defines{conan_defines}\n"
+            "    bindirs{conan_bindirs}\n"
+            "end\n")
+        ret[BUILD_INFO_PREMAKE] = "\n".join(sections)
+
+        template_deps = template + 'conan_sysroot{dep} = "{deps.sysroot}"\n'
+
+        # Iterate all the transitive requires
+        for require, dep in list(host_req.items()) + list(build_req.items()) + list(test_req.items()):
```
memsharded: It is very likely that ``build_req`` shouldn't be added the same as the others. In 2.0, build_req are only executable tools, they do not have headers, libraries that can be linked. Mostly managed by ``VirtualBuildEnv``. If added to the generator, they are typically added differently, with some mechanism that only put executables in the path or something like that, but making sure that libs/headers aren't cause they might produce errors because they will be in a different context (the build context)

memsharded (approved):

conan/tools/premake/premake.py
```diff
@@ -0,0 +1,39 @@
...
+            _visuals = {'8': '2005',
+                        '9': '2008',
+                        '10': '2010',
+                        '11': '2012',
+                        '12': '2013',
+                        '14': '2015',
+                        '15': '2017',
+                        '16': '2019'}
```
memsharded, line 15: Many thanks for contributing this PR. It has been merged, it will be in next 1.44.
If you keep using it, please try to add VS 2022 support as soon as you upgrade to it, there are users already using VS2022 with other generators.

Enhex: according to [Premake's documentation](https://premake.github.io/docs/Using-Premake) there's no VS 2022 generator yet.

memsharded: Oh, I see. Good, then users will need to wait to Premake to add it, lets wait then.

memsharded:
Please have a look to  #10216 when possible

PR #10091, opened 2021-11-26
changed cwd for source() method

memsharded:
Changelog: Feature: When the layout is declared, the cwd() in the source() method will follow the declared `self.folders.source` but the `exports_sources` will still be copied to the base source folder.
Docs: https://github.com/conan-io/docs/pull/2486

#tags: slow


conans/test/functional/tools/test_files.py
```diff
@@ -116,6 +116,61 @@ def build(self):
     assert 'mypkg/1.0: Apply patch (security)' in str(client.out)
 
 
+def test_patch_real():
+    conanfile = textwrap.dedent("""
```
memsharded, line 121: Please @lasote review this new test:

- Seems to work well, I'd say still intuitive
- The only thing I slightly struggled is the new ``patch()`` function assuming it works on the ``conanfile.source_folder`` to apply patches to ``build()``. I needed to pass ``base_path=self.build_folder``, but that was unexpected I assumed it was operating on ``cwd``

lasote: Indeed that was unexpected. I always though about patches in "source" but probably the line `root = os.path.join(conanfile.source_folder, base_path) if base_path else conanfile.source_folder` should be changed.

memsharded: The real ``no_copy_source`` is challenging in any case, cannot be done with the current code in ``develop``. Changed so it is at least supported in someway.

danimtb (approved):

memsharded:
@uilianries @SSE4 this PR contains the test ``test_third_party_patch_flow()`` that shows how can patches be created without touching the Conan cache, with a more intuitive developer flow. Please review it, and if there is something I have not considered, I would love to know it. Thanks

conans/test/functional/test_third_party_patch_flow.py
```diff
@@ -0,0 +1,199 @@
...
+    client.run("source .")
+    client.run("build .")
+    assert "conanfile.py (mypkg/1.0): Calling build()" in client.out
+
+    # of course create should work too
+    client.run("create .")
+    assert "mypkg/1.0: Created package" in client.out
+
+
+def test_third_party_overwrite_build_file():
```
memsharded, line 108: @lasote I have added these 2 more "flow" tests, to illustrate how this would behave for downloads and git clones. There is one workaround to do , but might be still worth it the approach.

SSE4:
it's not very well explained in the description, so I wanna ask some potentially dumb questions.

1. why does source need to deal with `cwd()` at all? for most consumers, it's probably natural to expect that `cwd` matches the `cwd` of the shell where conan command was invoked, so it may access files relative to the current directory. if I want to access files in other directory, I always may use `tools.chdir(self.source_folder)` or `tools.chdir(self.build_folder)` or other.

2. from my point of view, it's very inconvenient to have patches applied in the source method. we already agreed previously that sources should be immutable (at least in CCI), and should match vanilla upstream sources, so you can always inspect the initial state without a need to revert any patches or replace_in_file calls. we already know the way conan caches sources, it might be very problematic to e.g. apply a different set of mutually exclusive patches for different platforms. there are also other annoying UX side-effects, such as inability to make patches optional, force removal of source folder if patch fails to apply for whatever reason, etc. right now, there is a natural expectation that `source` method is about transferring data to the machine (usually, download + check sha + unzip), and `build` method is about following the upstream build instructions, and if build instructions include applying patches or modifying patches, they are also natural and integral parts of the build process.

conans/client/source.py
```diff
@@ -141,13 +140,15 @@ def _run_source(conanfile, conanfile_path, hook_manager, reference, cache,
...
                     _clean_source_folder(src_folder)
-                with conanfile_exception_formatter(conanfile.display_name, "source"):
-
-                    with conan_v2_property(conanfile, 'settings',
-                                           "'self.settings' access in source() method is deprecated"):
-                        with conan_v2_property(conanfile, 'options',
-                                               "'self.options' access in source() method is deprecated"):
-                            conanfile.source()
+                mkdir(conanfile.source_folder)
+                with tools.chdir(conanfile.source_folder):  # Lets see how many things this breaks
```
SSE4: I think you may comma-separate context managers instead of nesting them

conans/test/functional/test_third_party_patch_flow.py
```diff
@@ -0,0 +1,199 @@
...
+            def build(self):
+                for i in (1, 2, 3):
+                    if "mistake{}".format(i) in load(os.path.join(self.source_folder, "myfile.cpp")):
+                        raise Exception("MISTAKE{} BUILD!".format(i))
+        """)
+
+    client = TestClient()
+    client.save({"conanfile.py": conanfile,
+                 "conandata.yml": ""})
+    client.run("install .")
```
SSE4, line 45: I miss where is layout defined here, there is no command-line argument like `--layout` or individual directories like `-if`/`-bf`/`-sf`/`-pf`.
every command just passes an unnamed positional argument `.`, what are these directories in terms of layout?
as I can see, the layout is defined inside the recipe, why not outside of it?
in this case, the client seems to be required to have knowledge that this particular recipe uses `src` directory as a source folder.
how should I write my commands below (lines 55-56) so that they work on arbitrary recipes (the one which may define `self.folder.source = "non-standard"`?

conans/test/functional/test_third_party_patch_flow.py
```diff
@@ -0,0 +1,199 @@
...
+    client.run_command("cd src && git diff > ../patches/patch1")
+    client.run_command("cd src && git add . && git commit -m patch1")
+    conandata = textwrap.dedent("""
+        patches:
+          "1.0":
+            - patch_file: "patches/patch1"
+        """)
+    client.save({"conandata.yml": conandata})
+
+    client.run("source .")
```
SSE4, line 64: I would love to see the check it (`source` command) doesn't lose any of my local modifications made into the `src` sub-directory, if any.
right now it saves entire changes to the patch, it would be better if the test only saved some changes and some were just left as uncommitted or staged.

memsharded:
> why does source need to deal with cwd() at all? for most consumers, it's probably natural to expect that cwd matches the cwd of the shell where conan command was invoked, so it may access files relative to the current directory. if I want to access files in other directory, I always may use tools.chdir(self.source_folder) or tools.chdir(self.build_folder) or other.

``source()`` doesn't need to deal with cwd at all. It doesn't do it. With the ``layout()`` all the paths are defined by the layout, not local. If you mean about this modification:
```
root = os.getcwd() if base_path is None else os.path.join(os.getcwd(), base_path)
```

The ``cwd`` of the ``apply_conandata_patches()`` is the current directory, already defined by the calling function. For ``source()`` it will be the ``self.source_folder``, and for ``build()`` it will be the ``self.build_folder()``. It is just a mechanism to avoid over-specification the current default.


> from my point of view, it's very inconvenient to have patches applied in the source method. we already agreed previously that sources should be immutable (at least in CCI), and should match vanilla upstream sources, so you can always inspect the initial state without a need to revert any patches or replace_in_file calls. we already know the way conan caches sources, it might be very problematic to e.g. apply a different set of mutually exclusive patches for different platforms. there are also other annoying UX side-effects, such as inability to make patches optional, force removal of source folder if patch fails to apply for whatever reason, etc. right now, there is a natural expectation that source method is about transferring data to the machine (usually, download + check sha + unzip), and build method is about following the upstream build instructions, and if build instructions include applying patches or modifying patches, they are also natural and integral parts of the build process.

The proposal doesn't involve the cache at all, just a local user folder (``we already know the way conan caches sources``, 
There is no caching of Conan sources in the proposed flow, every thing is local). You say that you want those sources to be immutable, but at the same time you are doing changes to that source code. I don't get it. The purpose of doing a ``git init`` in the source folder is exactly capturing the initial state, and making all modifications and changes traceable and diff-able. If you are using git you can always inspect the initial state, that is not an issue at all.

``apply a different set of mutually exclusive patches for different platforms`` how would you do this in practice in the ``build()`` folder? because if you are in the same machine, you are not going to easily build that for different platforms there, you are still going to have a collision or something in the local ``build`` folder. And how you do this in ``conan-center-index`` is it really a thing? Can you please point me to some examples in CCI, because I cannot find this use case of mutually exclusive patches for different platforms there. How do you make in CCI patches optional (``such as inability to make patches optional``)?

SSE4:
> because if you are in the same machine, you are not going to easily build that for different platforms there, you are still going to have a collision or something in the local build folder. 

as far as I understand, the `source_folder` has to be the same for all the configurations (package IDs), it's shared between all of them. otoh, `build_folder` is specific for the given configuration, each package ID always gets its own, unique build folder. there is no collision in this case.

> And how you do this in conan-center-index is it really a thing?

there are many conditional patches in CCI ATM, some examples:
https://github.com/conan-io/conan-center-index/blob/1320e914c6fc917da7053433b663388fee607bdc/recipes/liquid-dsp/all/conanfile.py#L75
https://github.com/conan-io/conan-center-index/blob/0f4be54da887bfe38558093691b20f24300d0dd6/recipes/openssl/1.x.x/conanfile.py#L733
https://github.com/conan-io/conan-center-index/blob/0f4be54da887bfe38558093691b20f24300d0dd6/recipes/openssl/1.x.x/conanfile.py#L738
(not a best practice, though, but happens)

> How do you make in CCI patches optional

same as above, add an option and condition to skip patches

> You say that you want those sources to be immutable, but at the same time you are doing changes to that source code.

they are mutable in the development flow, because I was unable to find a better alternative. in a normal flow, `source folder` always contains vanilla sources in case of CCI. all the patches and `replace_in_file` calls happen within `build` method. there are very few exceptions, especially for large sources such as `boost`, which use `no_copy_source` for performance reasons, but that's an exceptional case. for 99% of packages we really want immutable sources.

SSE4:
> The purpose of doing a `git init` in the source folder is exactly capturing the initial state, and making all modifications and changes traceable and diff-able. If you are using git you can always inspect the initial state, that is not an issue at all.

for me, as a contributor who is initially adding all of these patches - yes, I am using git to add these patches, and I can track all the changes.

but for one who consumes an existing recipe with an existing set of patches such as an example:
```
           patches:
             "1.0":
               - patch_file: "patches/patch1"
               - patch_file: "patches/patch2"
               - patch_file: "patches/patch3"
```
```
            def source(self):
                # emulate a download from web site
                save("myfile.cpp", "mistake1\nsomething\nmistake2\nmistake3\nsome\n")
                apply_conandata_patches(self)
```
I don't see an easy way to trace an initial (vanilla) state, or in other words, get the state of sources right before the `apply_conandata_patches` call. that's due to the fact `source` method is atomic, and you can't just stop in the middle of its execution and inspect the sources. the only way is to modify `conandata.yml` and comment out all the patches (and it means introducing a new recipe revision, which might not be desired).
that's complicated workflows, as if you encounter an error, you don't know if it's from the original set of sources or from the patched one. you just see something like `error at Makefile line 666` and you don't know if line 666 was patched or not - you either need to study patches + source method, or go to browse sources somewhere else (e.g. on GitHub).

but it doesn't necessarily have to go to build method, there are other ways, e.g. you can add an argument `conan source . --vanilla`, which will force all calls like `apply_conandata_patches` or `replace_in_file` to be no-op effectively, in order to get vanilla sources. that's another possible proposal.

memsharded:
> as far as I understand, the source_folder has to be the same for all the configurations (package IDs), it's shared between all of them. otoh, build_folder is specific for the given configuration, each package ID always gets its own, unique build folder. there is no collision in this case.

The source folder **in the cache** will be the same.

The provided example flow **do not use** the source folder in the cache for anything

SSE4:
> The source folder **in the cache** will be the same.
> 
> The provided example flow **do not use** the source folder in the cache for anything

that's totally fine, but with local flow, without cache, what should I do?
option A:
```
conan source <source_folder>
conan build <build_folder1> --profile <profile1>
conan build <build_folder2> --profile <profile2>
```
option B:
```
conan source <source_folder1>
conan source <source_folder2>
conan build <build_folder1> --profile <profile1>
conan build <build_folder2> --profile <profile2>
```
in other words, should 2 different package IDs in local flow use 2 completely different layouts, or should they share the same source folder?

conans/test/functional/test_third_party_patch_flow.py
```diff
@@ -0,0 +1,199 @@
...
+            version = "1.0"
+            exports_sources = "CMakeLists.txt"
+
+            def layout(self):
+                self.folders.source = "src"
+                self.folders.build = "build"
+
+            def source(self):
+                # git clone: WILL FAIL because folder not empty, because UGLY local exports already
+                [shutil.rmtree(t) if os.path.isdir(t) else os.remove(t) for t in os.listdir(".")]
```
lasote: But I don't know why do you need here to remove anything if you didn't export anything to the src folder.

memsharded: I have added a check to try to explain this, see next commit. The thing is that local ``conan source`` does a copy of CMakeLists.txt and conandata.yml inside the "src" folder.

conans/test/functional/test_third_party_patch_flow.py
lasote: Please, you should take a look at least at the highlight features we release, the layout feature has some time now. This PR is not about discussing the layouts feature.

conans/test/functional/test_third_party_patch_flow.py
```diff
@@ -0,0 +1,203 @@
...
+
+            def layout(self):
+                self.folders.source = "src"
+                self.folders.build = "build"
+
+            def source(self):
+                # git clone: WILL FAIL because folder not empty, because UGLY local exports already
+                self.output.info("CWD: {{}}!".format(os.getcwd()))
+                self.output.info("FILES: {{}}!".format(sorted(os.listdir("."))))
+                [shutil.rmtree(t) if os.path.isdir(t) else os.remove(t) for t in os.listdir(".")]
```
lasote: Now I understand... but this is creepy and a mess. The source/exports flow is obviously not prepared yet.

lasote: Maybe we need an export layout, relative to the `base_source`, so in this case you can:

 `self.folders.export = ""` 

... and avoid the cleaning?

lasote: Or should we avoid copying the exports in the source method? Is that in develop2? I'm a bit lost

lasote (approved):

PR #10093, opened 2021-11-26
further improving GraphAPI and InstallAPI

memsharded:
Splitting:

- The ``load_root_node`` which is the ugliest part of computing the graph, better isolate it (maybe further split in the feature for the different cases (``reference``, ``path``, etc)
- The ``binary_analyzing``

### DIFF
--- a/conans/cli/api/subapi/graph.py
+++ b/conans/cli/api/subapi/graph.py
@@ -13,13 +13,17 @@ class GraphAPI:
     @api_method
     def load_root_node(self, reference, path, profile_host, profile_build, lockfile, root_ref,
                        create_reference=None, is_build_require=False,
-                       require_overrides=None):
+                       require_overrides=None, remote=None, update=None):
 
         if path and reference:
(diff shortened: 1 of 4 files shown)


conans/cli/api/subapi/graph.py
```diff
@@ -9,59 +10,63 @@ class GraphAPI:
...
+            out.writeln("")
+            app.range_resolver.clear_output()
 
         if lockfile:
             lockfile.update_lock(deps_graph)
 
         return deps_graph
+
+    @api_method
+    def analyze_binaries(self, deps_graph, build_mode=None, remote=None, update=None):
```
lasote: Docstring

lasote (approved):

conans/cli/api/subapi/graph.py
memsharded: Done, in this and other methods too, completing params.
Some that are still too unstable, and very unlikely to remain, I have put TODOs.

lasote (approved):

PR #10094, opened 2021-11-26
initializing Options in the conanfile constructor, not delayed

memsharded:
At the moment it is delayed to avoid ``python_requires_extend`` extending model over the ``options`` primitives, not the ``Options`` object.


lasote:
> At the moment it is delayed to avoid python_requires_extend extending model over the options primitives, not the Options object.

So? we don't care anymore or...

memsharded:
> So? we don't care anymore or...

Yes, we care, the PR is proposing a ``Options.update()`` method that allows this functionality, and some ``inspect`` modifications to print the ``Options`` as ``conan inspect`` expects. The main difference is that now it has to be provided by ``Options``, instead of being a Python builtins user operation merging dicts, lists, or whatever. It is kind of a trade-off:

- Do the same with ``Options`` as we do with ``Requirements``: Transform the user class attributes from python builtins to Conan model at ``__init__()`` time. Advantage: less spaguetti code, no need to call ``conanfile.initialize()`` or something like that after the execution of the ``python_requires_extend``. Disadvantage: Need to do the operation of merging options with Conan ``Options`` provided functionality, instead of builtin Python functionality. The ``inspect`` functionality needs some awareness of how to print the ``Options`` attribute differently.
- Keep ``Options`` initialization delayed (the opposite of above)

conans/client/conan_api.py
```diff
@@ -153,7 +153,11 @@ def inspect(self, path, attributes, remote_name=None):
         for attribute in attributes:
             try:
                 attr = getattr(conanfile, attribute)
-                result[attribute] = attr
+                if attribute == "options":
+                    result[attribute] = {k: v._possible_values
+                                         for k, v in attr._package_options._data.items()}
```
lasote: bit dirty. Could we have a method or property to get the items directly for representation?

lasote (commented):
I think it is more consistent now and a bit more clear and self-contained.

conans/client/conan_api.py
memsharded: Fixed

lasote (approved):

lasote:
I solved a conflict, you can merge when green.

PR #10096, opened 2021-11-27
refactoring settings definition out of loader into graph_manager

memsharded:
Should be pure refactor. Only some unittests changed because of internal interfaces changing.

The assignment of settings is done too inside the ``ConanFileLoader`` (when we only had 1 profile), but also in other places, making it very spread distributed and difficult to track. The pattern-matching of package-settings, and other profile => conanfile assignment is being centralized in one place.


lasote (approved):
Nice refactor. Still a draft?

memsharded:
> Nice refactor. Still a draft?

Yes, I wanted to at least add some more comments, TODOs, etc. I will add them and mark it as ready if you like the overall idea (there is still ton of work to do to have something nice here)

PR #10099, opened 2021-11-29
Remove cmake_target_namespace and cmake_module_target_namespace properties in CMakeDeps generator

czoido:
Changelog: Feature: Remove `cmake_target_namespace` and `cmake_module_target_namespace` properties.
Changelog: Feature: Allow `CMakeDeps` to set `cmake_target_name` property as an absolute target.
Changelog: Feature: Add warning in `CMakeDeps` generated CMake files when target names collide.
Docs: https://github.com/conan-io/docs/pull/2316

Now `CMakeDeps` will allow setting an absolute name for the targets. It's also possible to name the target for the root cpp_info and components without namespaces.
If no `cmake_target_name` is set, then the default generated target is: `pkg_name::pkg_name` for the root cpp_info and `pkg_name::component_name` for components (note it will not try to extract the namespace from the root cpp_info to compose the component target name)

supersedes: https://github.com/conan-io/conan/pull/10077

#TAGS: slow


conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -82,35 +78,32 @@ def config_suffix(self):
...
-            ret = req.cpp_info.components[comp_name].get_property("cmake_module_target_name",
-                                                                  "CMakeDeps")
+            ret = req.cpp_info.components[comp_name].get_property("cmake_module_target_name", "CMakeDeps")
             if ret:
                 return ret
         ret = req.cpp_info.components[comp_name].get_property("cmake_target_name", "CMakeDeps")
-        return ret or comp_name
+        # If we don't specify a property for the name it will fallback to the pkg_name::comp_name
+        # note that it wont take cmake_target_name namespace as first component
+        # of the absolute target!
```
lasote:
```suggestion
        # If we don't specify the `cmake_target_name` property for the component it will 
        # fallback to the pkg_name::comp_name, it wont use the root cpp_info cmake_target_name 
        # property because that is also an absolute name (Greetings::Greetings), it is not a namespace 
        # and we don't want to split and do tricks.
```

conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -17,13 +17,9 @@ def __init__(self, cmakedeps, require, conanfile, find_module_mode=False):
     def pkg_name(self):
         return self.conanfile.ref.name + self.suffix
 
-    @property
-    def target_namespace(self):
-        return self.get_target_namespace(self.conanfile) + self.suffix
-
     @property
     def global_target_name(self):
```
lasote: Given the changes we are introducing, I would rename this method to `root_target_name`

conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -82,35 +78,32 @@ def config_suffix(self):
...
-
-        ret = req.cpp_info.get_property("cmake_target_namespace", "CMakeDeps")
-        return ret or self.get_global_target_name(req)
+    @staticmethod
+    def _get_target_default_name(req, component_name="", suffix=""):
+        return "{name}{suffix}::{cname}{suffix}".format(cname=component_name or req.ref.name,
+                                                        name=req.ref.name, suffix=suffix)
 
-    def get_global_target_name(self, req):
+    def get_global_target_name(self, req, suffix=""):
```
lasote: And I would rename this method to `get_root_target_name`

conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -82,35 +78,32 @@ def config_suffix(self):
...
+        return ret or self._get_target_default_name(req, suffix=suffix)
 
     def get_component_alias(self, req, comp_name):
         if comp_name not in req.cpp_info.components:
             # foo::foo might be referencing the root cppinfo
             if req.ref.name == comp_name:
-                return self.get_target_namespace(req)
+                return self.get_global_target_name(req)
             raise ConanException("Component '{name}::{cname}' not found in '{name}' "
                                  "package requirement".format(name=req.ref.name, cname=comp_name))
```
lasote: I would try to move these lines to the `get_required_components_cpp` because that can only happen when we process the "requires" and it is quite confusing in other places.

lasote (approved):

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -21,12 +21,17 @@ def filename(self):
     def context(self):
         deps_targets_names = self.get_deps_targets_names() \
             if not self.conanfile.is_build_context else []
+
+        components_targets_names = self.get_declared_components_targets_names()
+        components_variables_names = [components_variables_name.replace("::", "_") for
+                                      components_variables_name in components_targets_names]
+        components_names = list(zip(components_targets_names, components_variables_names))
```
franramirez688: This is the same, isn't it?

```suggestion
        components_names = [(components_variables_name, components_variables_name.replace("::", "_")) for components_variables_name in components_targets_names]
```

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
```diff
@@ -778,5 +777,68 @@ def build(self):
...
+                cmake.build()
+        """)
+
+    cmakelists = textwrap.dedent("""cmake_minimum_required(VERSION 3.15)
+        project(Consumer)
+        find_package(CURLFILENAME CONFIG REQUIRED COMPONENTS libcurl libcurl2)
+        """)
+
+    client.save({"consumer/conanfile.py": conanfile, "consumer/CMakeLists.txt": cmakelists})
+    client.run("create consumer")
```
franramirez688, line 849: I guess it's enough to see that all the commands run fine so that's the reason why we don't need any `assert` here but it'd be great if you add a little explanation as docstring to understand the whole context and what we expect

franramirez688 (approved):

lasote:
I was reviewing [this](https://github.com/conan-io/conan/pull/9851) pull request proposing to remove the restriction of having a component with the name of the package, and I have some related questions:
What would happen if a user declares a component with a `cmake_target_name` and the target name collides with the root one? Or with another component? What about the aliases? Do we have tests? Is it undefined behavior?

czoido:
> I was reviewing [this](https://github.com/conan-io/conan/pull/9851) pull request proposing to remove the restriction of having a component with the name of the package, and I have some related questions: What would happen if a user declares a component with a `cmake_target_name` and the target name collides with the root one? Or with another component? What about the aliases? Do we have tests? Is it undefined behavior?

For the aliases there's a check and a warning that the alias won't be declared if already declared: https://github.com/SSE4/conan/blob/16556f34e1e70e07e6d20188193df16adaed7c59/conan/tools/cmake/cmakedeps/templates/targets.py#L85

For `cmake_target_name` I think we should probably raise if the target names collide

lasote:
I would say, Let's raise in both cases. We can always revert if we realize later that something is being re-declared or something like that.

czoido:
> I would say, Let's raise in both cases. We can always revert if we realize later that something is being re-declared or something like that.

Ok, I'll introduce the changes to raise if targets collide (if necessary) in this PR ~~and let the changes for cmake_target_aliases for another PR.~~ (also adding a FATAL error for aliases)

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py
```diff
@@ -119,3 +119,65 @@ def package_info(self):
...
+        def package_info(self):
+            self.cpp_info.components["buy"].set_property("cmake_target_aliases", ["hello::buy"])
+    """)
+
+    cmakelists = textwrap.dedent("""
+    cmake_minimum_required(VERSION 2.8)
+
+    find_package(hello REQUIRED)
+    get_target_property(link_libraries hola::adios INTERFACE_LINK_LIBRARIES)
+    message("hola::adios link libraries: ${link_libraries}")
```
lasote:
```suggestion
```

lasote (approved):

lasote (approved):
Note for the record: We decided not to raise with a target collision because we can't, there is a "feature" that allows declaring a generic target "nonstd::nonstd" for example, that will aggregate the target properties for all the declared targets in the dependencies for that name. We are afraid this is used in more than one company also.

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -78,25 +83,25 @@ def template(self):
...
-              set({{ pkg_name }}_{{ comp_name }}_COMPILE_OPTIONS_CXX{{ config_suffix }} "{{ cpp.cxxflags_list }}")
-              set({{ pkg_name }}_{{ comp_name }}_LIBS{{ config_suffix }} {{ cpp.libs }})
-              set({{ pkg_name }}_{{ comp_name }}_SYSTEM_LIBS{{ config_suffix }} {{ cpp.system_libs }})
-              set({{ pkg_name }}_{{ comp_name }}_FRAMEWORK_DIRS{{ config_suffix }} {{ cpp.framework_paths }})
-              set({{ pkg_name }}_{{ comp_name }}_FRAMEWORKS{{ config_suffix }} {{ cpp.frameworks }})
-              set({{ pkg_name }}_{{ comp_name }}_DEPENDENCIES{{ config_suffix }} {{ cpp.public_deps }})
-              set({{ pkg_name }}_{{ comp_name }}_LINKER_FLAGS{{ config_suffix }}
+              set({{ pkg_name }}_COMPONENTS{{ config_suffix }} {{ components_names }})
+
+              {%- for comp_variable_name, comp_target_name, cpp in components_cpp %}
```
memsharded, line 88: The iteration in above files is ``{%- for comp_target_name, comp_variable_name in components_names %}``, maybe good to use the same order and swap?

conan/tools/files/cpp_package.py
```diff
@@ -47,7 +47,11 @@ def package_info(self, conanfile):
             conanfile.cpp_info.components[cname].libs = component.libs
             conanfile.cpp_info.components[cname].requires = component.requires
             for generator, gname in component.names.items():
-                conanfile.cpp_info.components[cname].set_property("cmake_target_name", gname, generator)
+                # TODO: probably we want to revisit all this stuff with the new absolute
+                #  target names via set_property
+                conanfile.cpp_info.components[cname].set_property("cmake_target_name",
+                                                                  "{}::{}".format(conanfile.cpp_info.name, gname),
```
memsharded: Is there a way we could do this without relying on ``cpp_info.name``? Because this will break merging to ``develop2``

This ``cpp_package`` is a new feature, should be ok if doing only the new thing, or supporting only CMakeDeps.

memsharded (approved):

memsharded (approved):

Issue #10104, opened 2021-11-30
[question] [2.0] Options are frozen when calling `configure`


babariviere:
### Environment Details (include every applicable attribute)
  * Operating System+version: Nixos-21.11.20211122.263ef4c
  * Compiler+version: `gcc (GCC) 10.3.0`
  * Conan version: `Conan version 2.0.0-alpha1`
  * Python version: `Python 3.8.2`

### Steps to reproduce (Include if Applicable)

1. `conan new hello/0.1 --template=cmake_lib`
2.  Apply this patch:
```diff
--- a/conanfile.py	2021-11-30 10:24:55.235810970 +0100
+++ b/conanfile.py	2021-11-30 10:22:49.318555176 +0100
@@ -16,8 +16,18 @@
 
     # Binary configuration
     settings = "os", "compiler", "build_type", "arch"
-    options = {"shared": [True, False], "fPIC": [True, False]}
-    default_options = {"shared": False, "fPIC": True}
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "zlib": [True, False],
+        "without_zlib": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "zlib": True,
+        "without_zlib": False,
+    }
 
     # Sources are located in the same place as this recipe, copy them to the recipe
     exports_sources = "CMakeLists.txt", "src/*"
@@ -33,6 +43,11 @@
         tc = CMakeToolchain(self)
         tc.generate()
 
+    def configure(self):
+        if self.options.get_safe("without_zlib", False):
+            del self.options.zlib  # fails silently
+            self.options.zlib = False  # return an exception
+
     def build(self):
         cmake = CMake(self)
         cmake.configure()
```

<details>
 <summary>Full conanfile.py</summary>

```python
from conans import ConanFile
from conan.tools.cmake import CMakeToolchain, CMake
from conan.tools.layout import cmake_layout


class HelloConan(ConanFile):
    name = "hello"
    version = "0.1"

    # Optional metadata
    license = "<Put the package license here>"
    author = "<Put your name here> <And your email here>"
    url = "<Package recipe repository url here, for issues about the package>"
    description = "<Description of Hello here>"
    topics = ("<Put some tag here>", "<here>", "<and here>")

    # Binary configuration
    settings = "os", "compiler", "build_type", "arch"
    options = {
        "shared": [True, False],
        "fPIC": [True, False],
        "zlib": [True, False],
[...]

PR #10109, opened 2021-12-01
Feature/develop2 relax options freeze

memsharded:
Solve https://github.com/conan-io/conan/issues/10104#issuecomment-983499654

The current code in 2.0 seems to be too strict on the freezing on options values. The final goal is to make sure the values defined in the profiles always have priority over the ones defined in recipes. It is very bad that I declare ``boost:static=False`` in my profile and for some reason some recipe can decide to change it to ``boost:static=True``, without erroring. This is what is possible with Conan 1.X and we are trying to avoid here.

However, the current ``freeze`` seems to strict, and I am relaxing here to allow:
- Removing options that have not been assigned a value yet (``None`` value)
- Assigning values to options that have not been defined yet (``None`` value)
- 
This is an exploratory PR to discuss, lets see what makes sense

### DIFF
--- a/conans/model/options.py
+++ b/conans/model/options.py
@@ -87,8 +87,6 @@ class _PackageOptions:
 
     def clear(self):
         # for header_only() clearing
-        if self._freeze:
-            raise ConanException(f"Incorrect attempt to modify options.clear()")
         self._data.clear()
 
     def freeze(self):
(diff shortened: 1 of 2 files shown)


lasote (commented):
Looks good and the behavior too.

PR #10089, opened 2021-11-26
Updated the ConanException in installer.py to improve the error messa…

abrousseau001:
Changelog: Fix: Updated the ConanException in installer.py to improve the error message handling.
Docs: https://github.com/conan-io/docs/pull/2305

Fixes: #9804

- [x] Refer to the issue that supports this Pull Request.
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [x] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #10113, opened 2021-12-01
printing only 1 ref at missing binary message

memsharded:
Changelog: Omit
Docs: Omit

Changes in https://github.com/conan-io/conan/pull/10089 can be confusing when more than 1 binary is missing. This should fix it.

Please review @abrousseau001

### DIFF
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -373,7 +373,7 @@ class BinaryInstaller(object):
             Or try to build locally from sources with '%s'
 
             More Info at 'https://docs.conan.io/en/latest/faq/troubleshooting.html#error-missing-prebuilt-package'
-            ''' % (missing_pkgs, missing_pkgs, build_str)))
+            ''' % (missing_pkgs, ref, build_str)))
 
     def _download(self, downloads, processed_package_refs):
         """ executes the download of packages (both download and update), only once for a given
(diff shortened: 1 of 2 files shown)


abrousseau001:
@memsharded I will take a look at it this evening.

abrousseau001:
@memsharded and @czoido, I have reviewed the change and it makes sense. I see that the unit test was updated as well to match that the first missing package is what we are using for the error message. 

If we instead wanted to change the message to include the 'conan search' string that would apply to each package, we would need to adjust the logic to iterate over the missing_pkgs variable to build a string that would fall in place. I will leave that decision to you. Otherwise, it is good to go for the hotfix and I apologize for missing that.

memsharded:
> If we instead wanted to change the message to include the 'conan search' string that would apply to each package, we would need to adjust the logic to iterate over the missing_pkgs variable to build a string that would fall in place. I will leave that decision to you. Otherwise, it is good to go for the hotfix and I apologize for missing that.

Nop, I think it is better to show just 1 line of the ``conan search`` as example, the user can deduce to apply it for the other missing refs. But displaying all would be too noisy IMHO.

Don't worry! 😄 These things happen all the time to everyone, thanks very much again for moving forward and contributing the feature! Lets merge it.

Issue #9396, opened 2021-08-06
[feature] Add version() for git scm


dariusarnold:
My proposal is to add the method version() to tools.Git. Calling this method would return the Major.Minor.Patch of the found git executable (maybe as a named tuple). These can be used to customize the git commands which are used. As an example, partial cloning (cloning only some files/directories from a repo to save time) is available since Git 2.19.0. If the executable is new enough, partial cloning could be used. Otherwise the conanfile.py would fallback to cloning the while repository.
It is possible to do this manually by running "git --version" using git.run() and parsing the output, but this could be said about many of the methods provided by tools.Git.

Is there any interest in this?

- [-] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Yes, I understand the use case, I wouldn't oppose to such a method, sounds relatively low risk, and could help users, specifically those that need to deal with quite older git versions.

ghost:
I'm interested in taking this on, can sb assign me or @dariusarnold  will work on it?

dariusarnold:
You are free to take it. I just suggested it without a concrete plan to start working on it.

For feature parity maybe it would be useful to implement it for svn as well, but I dont know much about it.

memsharded:
> For feature parity maybe it would be useful to implement it for svn as well, but I dont know much about it.

Would be good, but not strictly necessary, only if SVN users demand it. So don't feel obligated if you don't have svn or knowledge about it.

aniketh-varma:
Hi, I would like to work on this issue. Since this is my first open-source contribution, I would be grateful if you can also guide me on doing the same.

memsharded:
@Alzamer had it assigned first. Can you please @Alzamer confirm that you are working on it?

ghost:
@memsharded @aniketh-varma Yes, I'm wrapping my head around it - I have been quite busy lately, hence I've just started

memsharded:
@aniketh-varma please have a look at other "good first issue" in the repo.

aniketh-varma:
> @aniketh-varma please have a look at other "good first issue" in the repo.

Sure!

ghost:
@aniketh-varma Sorry for the confusion, but I won't be able to end this task in a reasonable period of time - take it if you still interested

abrousseau001:
[...]

PR #10114, opened 2021-12-01
Feature - Add version method for git scm per #9396

abrousseau001:
Changelog: Feature: Add ``Git.version`` property to check the current git version, aligned with ``SVN.version``.
Docs: https://github.com/conan-io/docs/pull/2338

Implements feature request #9396 

This was left as just the entire return string instead of the major.minor.patch as was suggested by feature request to conform with what was already being done with the SVN side of things.

A simple test to check if the return of the version is not the ConanException error indicating the executable was missing from the path

### DIFF
--- a/conans/client/tools/scm.py
+++ b/conans/client/tools/scm.py
@@ -148,6 +148,13 @@ class Git(SCMBase):
     def _configure_ssl_verify(self):
         return "-c http.sslVerify=%s " % ("true" if self._verify_ssl else "false")
 
+    @property
+    def version(self):
+        if not hasattr(self, '_version'):
+            version = Git.get_version()
+            setattr(self, '_version', version)
+        return getattr(self, '_version')
+
     def run(self, command):
         command = self._configure_ssl_verify + command
         return super(Git, self).run(command)
--- a/conans/test/functional/scm/scm_test.py
+++ b/conans/test/functional/scm/scm_test.py
@@ -627,6 +627,10 @@ class ConanLib(ConanFile):
         content = load(exported_conanfile)
         self.assertIn(commit, content)
 
+    def test_git_version(self):
+        git = Git()
+        self.assertNotIn("Error retrieving git", git.version)
+
 
 @pytest.mark.tool_svn
 class SVNSCMTest(SVNLocalRepoTestCase):


abrousseau001:
I did not open another PR in the documentation as the version string is not documented for SVN. It seems only very high level information is given to the using of source control.

memsharded (commented):
I am a bit confused by this PR (more exactly, about the request). Apparently, ``Git`` already contains a ``get_version()`` method, so the main issue was that it was not public nor documented?

In that case, isn't it enough to document the method? Caching the result is not bad, but as this caching is per-instance, it wouldn't even benefit version checking from multiple recipes.

Am I missing something? 
Thanks for contributing this!

abrousseau001:
@memsharded Git itself did not have a get_version() method. That is being called from the parent class SCMBase. I modelled the way we were doing SVN to match for Git. Neither this pull request nor SVN's version property are documented. If we want to change the way this is working, I would suggest thinking a bit more on how we should implement the concrete implementations of Git and SVN as singleton classes.

memsharded:
> Git itself did not have a get_version() method. 

But ``Git`` extends from SCMBase: ``class Git(SCMBase):``, yo, yes, it has a ``get_version()`` method, inherited from ``SCMBase``. There is even a test that validates it works:

```python
    @patch('subprocess.Popen')
    def test_version(self, mocked_open):
        mocked_open.return_value.communicate.return_value = ('git version 2.21.0'.encode(), None)
        version = Git.get_version()
        self.assertEqual(version, "2.21.0")
```

abrousseau001:
@memsharded I understand that, but why does SVN implement it the same way as a property? I think that we should either remove the SVN property and allow the SCMBase handle the version or implement this pull request in the same way. Either way is good with me, but consistency is best.

memsharded:
> @memsharded I understand that, but why does SVN implement it the same way as a property? I think that we should either remove the SVN property and allow the SCMBase handle the version or implement this pull request in the same way. Either way is good with me, but consistency is best.

Oh! Now I see what you mean, sorry I didn't understand that. Yes, this is inconsistent, and I agree that keeping consistency is best. I see SVN is actually documenting ``version()`` as a method 🤦 not a property, so we could declare a bug in the docs, and change the docs to "property" instead (while documenting the new property in ``Git`` as property too).

memsharded:
@abrousseau001 Added https://github.com/conan-io/docs/pull/2338 PR to docu for you

PR #10123, opened 2021-12-03
New command and api for export and export-pkg

memsharded:


conans/cli/api/conan_api.py
```diff
@@ -32,18 +33,14 @@ def __init__(self, cache_folder=None):
...
-
         # Get latest refs and list refs of recipes and packages
         self.list = ListAPI(self)
-
         self.profiles = ProfilesAPI(self)
-
         self.install = InstallAPI(self)
-
         self.graph = GraphAPI(self)
+        self.export = ExportAPI(self)
```
lasote, line 43: I don't like the semantics of `api.export.export()`. Probably the same for the install. I don't know if we might move them all to a `command` subapi.

lasote (approved):

conans/cli/api/conan_api.py
memsharded: Agree. There is still a global discussion about the API and the CLI for different commands. Here I am trying just to "modernize" a bit the code, moving it to another subapi would be straightforward.

czoido (approved):

PR #7760, opened 2020-09-28
Feature: add tests for compression and links

SSE4:
closes: #7295

@PYVERS: Macos@py27, Windows@py36, Linux@py27
@tags: slow, svn

#PYVERS: Macos@py27, Windows@py36, Linux@py27
#tags: slow, svn

~~finally made it work locally except timestamps which sometimes don't match (investigating...)~~

cases:
- simple archive
- valid file symbolic link
- valid file symbolic link (absolute)
- valid file symbolic link (relative)
- valid file symbolic link (outside of the archive)
- valid transitive symbolic link
- valid directory symbolic link
- broken symbolic link
- cyclic symbolic link
- symbolic link to empty directory
- valid hard-link
- more cases? (please comment)

compressions:
- zip
- tar
- tar.gz
- tar.bz2
- tar.lzma (python3 only)
- add more? (will require third-party modules for zstd, lzo, lz4, 7z, etc.)

do we need tests for compression ratio and speed?

Changelog: omit
Docs: omit

- [x] Refer to the issue that supports this Pull Request.
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #10125, opened 2021-12-03
2.0 symlinks improving

lasote:
- FileCopier keeps the symlinks, even if they are broken, even if they are absolute.
- The `merge_directories` calls now are also done with a FileCopier (preserving also symlinks)
- The `gather_files` will return the list of files of a folder and a list of symlinked folders in order to copy/create them properly in the destination if needed.
- The `FileTreeManifest` will hash the paths of the symlinks and the `md5` of the path pointed (not contents) so no matter if the symlink is broken, absolute, or relative, the hash is always complete.
- Removed the `tools.fix_symlinks`, verified with a test that the functionality is covered with the provided tools:
   - absolute_to_relative_symlinks(self, folder)
   - remove_external_symlinks(self, folder)
   - remove_broken_symlinks(self, folder)
- MISSING: 
       - ~~"Resolve" a symlink to a file => transform symlink to the real file.~~ These "resolve" tools are harder and potentially buggyer than expected. I think they are also speculative so let`s do it if requested.
       - ~~"Resolve/unroll" a symlink to a folder => copy the contents of the folder and transform the symlink to a folder.~~
   -  ~~Verify why the tgz (in an introduced test) is not complaining about compressing a symlink to an absolute folder. That might be contradicting the results of this test: https://github.com/conan-io/conan/pull/7760/files#diff-cc1fd3971722453800512735529582a2ea31b054889d0d312625b3cdae4b72d3R567-R583~~ The tgz containing an absolute path entry is not the same that a relative path symlink pointing to an absolute path. This is managed nicely.
   - ~~Add a test with a symlink containing "../../" to verify the behavior documented here: https://github.com/conan-io/conan/pull/7760/files#diff-cc1fd3971722453800512735529582a2ea31b054889d0d312625b3cdae4b72d3R585-R589~~ The tgz containing a path entry with ".." is not the same that a symlink with "..". This is managed nicely.
   - ~~Work for other PRs: Transform the FileCopier into a public class (tool) with the "symlink" tools inside.~~ Let`s do it later in another PR.


conans/client/cmd/export.py
```diff
@@ -246,6 +246,10 @@ def _classify_patterns(patterns):
             excluded.append(p[1:])
         else:
             included.append(p)
+
+    # FIXME: This is weird, as the test_package folder could be named different... do we want it?
```
memsharded: Actually, current implementation only excludes ``test_package/build``, not ``test_package``

conans/model/manifest.py
```diff
@@ -108,17 +98,27 @@ def create(cls, folder, exports_sources_folder=None):
...
         files, _ = gather_files(folder)
+        # The folders symlinks are discarded for the manifest
         for f in (PACKAGE_TGZ_NAME, EXPORT_TGZ_NAME, CONAN_MANIFEST, EXPORT_SOURCES_TGZ_NAME):
             files.pop(f, None)
 
         file_dict = {}
         for name, filepath in files.items():
-            file_dict[name] = md5sum(filepath)
+            if not os.path.exists(filepath):  # Broken symlink
+                # FIXME: What to hash? the path?
```
memsharded: Maybe hashing an empty string is enough, the hash already contains the file path:
```python
 s = ["%s: %s" % (f, fmd5) for f, fmd5 in sorted(self.file_sums.items())]
        s.append("")
        return md5("\n".join(s))```

Comment (SSE4) [reply conans/model/manifest.py]: we definitely shouldn't hash a symlink target contents, as it may change (e.g. symlink to `/usr/bin/perl` will be different for different distros), or it may be even infinite (symlinks to files like `/dev/urandom` or `/dev/zero`).
I would just hash the path (path to the symlink target) itself to ensure symlink stays the same.

**Comment by SSE4** [comment]
I personally don't like the generic name `fix_symlinks` (it's hard to guess what exactly does it do just from the name) and would suggest removing it from 2.0 entirely, replacing by more specific functions, as needed, like:
- `remove_broken_symlinks`
- `convert_abs_symlinks_to_relative`
- `convert_relative_symlinks_to_abs`
- `remove_cyclic_symlinks`
- `convert_symlinks_to_copies`
- `remove_external_symlinks`
- etc

**Comment by lasote** [comment]
> I personally don't like the generic name fix_symlinks (it's hard to guess what exactly does it do just from the name) and would suggest removing it from 2.0 entirely, replacing by more specific functions, as needed, like

Sure, the plan was to provide new tools (with new names) once we agree on what it's needed. That `fix_symlinks` was already there and will be renamed/refactored.

File: conans/client/cmd/export.py (hunk 1 above)
Comment (lasote) [reply conans/client/cmd/export.py]: So, I would say, better to be explicit and remove this.

Comment (memsharded) [reply conans/client/cmd/export.py]: Yes, agree. Users should be careful and not ``export`` the ``test_package`` folder. The only problem is that it might not be very intuitive on the UI. The ``export()`` method can do a controlled copy, and use ``FileCopier.copy(exclude=...)`` or something like that, but with the attribute ``exports = "*"``, that will be bad. But yes, lets remove it. Also we have harcoded ``if basename in [".git", ".svn"]:`` to exclude those subfolders, which are annoying in the 99% of cases to export. But some users asked sometime ago how to export the .git repo folder.

File: conans/model/manifest.py (hunk 2 above)
Comment (lasote) [reply conans/model/manifest.py]: I've hashed the path where the symlink is pointing, after all, that's the content of the file, right? No matter if it is broken or not.

Comment (SSE4) [reply conans/model/manifest.py]: yes, that's correct

File: conan/tools/files/__init__.py [hunk 3]
```diff
@@ -1,5 +1,6 @@
 from conan.tools.files.files import load, save, mkdir, ftp_download, download, get, rename, \
-    load_toolchain_args, save_toolchain_args, chdir
+    load_toolchain_args, save_toolchain_args, chdir, absolute_to_relative_symlinks, \
```
memsharded: maybe group all symlinks tools in some ``conan.tools.files.symlinks`` or something?

conan/tools/files/files.py
```diff
@@ -237,3 +237,56 @@ def chdir(conanfile, newdir):
...
+
+def get_symlinks(base_folder):
+    """Return the absolute path to the symlink files in base_folder"""
+    for (root, dirnames, filenames) in os.walk(base_folder):
+        for el in filenames + dirnames:
+            fullpath = os.path.join(root, el)
+            if not os.path.islink(fullpath):
+                continue
+            else:
+                yield fullpath
```
memsharded:
```suggestion
            if os.path.islink(fullpath):
                yield fullpath
```

conan/tools/files/files.py
```diff
@@ -237,3 +237,56 @@ def chdir(conanfile, newdir):
...
+def _path_inside(base, folder):
+    base = os.path.abspath(base)
+    folder = os.path.abspath(folder)
+    return os.path.commonprefix([base, folder]) == base
+
+
+def absolute_to_relative_symlinks(conanfile, base_folder):
+    """Convert the symlinks with absolute paths to relative if they are pointing to a file or
+    directory inside the 'base_folder'. Any absolute symlink pointing outside the 'base_folder'
+    will be discarded."""
```
memsharded:
```suggestion
    will be ignored"""
```

conans/model/manifest.py
```diff
@@ -17,30 +15,22 @@ def discarded_file(filename):
 
 def gather_files(folder):
     file_dict = {}
-    symlinks = {}
-    for root, dirs, files in walk(folder):
+    symlinked_folders = {}
+    for root, dirs, files in walk(folder, followlinks=True):
```
memsharded: Are we sure this will not package system files? For example if a package defines some symlink from a folder to ``/usr/include``, then this ``followlinks`` might iterate the files there?

conans/model/manifest.py
```diff
@@ -108,17 +98,23 @@ def create(cls, folder, exports_sources_folder=None):
...
         files, _ = gather_files(folder)
+        # The folders symlinks are discarded for the manifest
         for f in (PACKAGE_TGZ_NAME, EXPORT_TGZ_NAME, CONAN_MANIFEST, EXPORT_SOURCES_TGZ_NAME):
             files.pop(f, None)
 
         file_dict = {}
         for name, filepath in files.items():
-            file_dict[name] = md5sum(filepath)
+            # For a symlink: md5 of the pointing path, no matter if broken, relative or absolute.
+            value = md5(os.readlink(filepath)) if os.path.islink(filepath) else md5sum(filepath)
```
memsharded, line 80: why? this would be pretty useless to the manifest, if the file is not packaged, why computing its md5? Why not just blank or the ``md5`` of the ``source`` path (not the target one).

conan/tools/files/__init__.py
lasote: yeah, I thought about it, let's do it

conans/model/manifest.py
lasote: Well, I don't think is useless. The file is the symlink itself, and the content is the path (md5 of the path string, not contents of what is pointing to) where it is pointing. I think it is relevant if you are packaging a different symlink, pointing to a different location.

conans/model/manifest.py
lasote: I think this should avoid that:
```
if os.path.islink(root):
            dirs[:] = []
```
but I will make sure

lasote: anyway the `followlinks=True` is not needed so I'm removing it.

conans/model/manifest.py
memsharded: But that is the thing. If you have a ``pkg/lib/somelib.so -> /usr/lib/somelib.so``, then the system ``somelib.so`` might change from computer to computer, so the Manifest ``conan_manifest.txt`` will be broken, with an incorrect md5, and an irreproducible/uncheckable/wrong package revision. The idea is that we can validate the contents of a Conan package validating its manifest.

conans/util/files.py
```diff
@@ -386,55 +386,36 @@ def exception_message_safe(exc):
...
+    own files that shouldn't be uploaded
+    """
+    return filename == ".DS_Store" or filename.startswith("__conan")
+
+
+def gather_files(folder):
+    file_dict = {}
+    symlinked_folders = {}
+    for root, dirs, files in walk(folder):
+        dirs[:] = [d for d in dirs if d != "__pycache__"]  # Avoid recursing pycache
```
memsharded: I wonder if it is a good moment to remove this hardcoded filter. In the same way we don't filter anymore ``.pyc`` or ``test_package``?

memsharded (approved):

conans/model/manifest.py
lasote: The manifest won't be broken, it will be exactly the same, encoding the string `/usr/lib/somelib.so`, if the symlinks point to the system, and the system changes, of course, the final package will be different, in terms of linking with another different file, but it will be different anyway even if you hash an empty string.
So, if the manifest "is broken" in this case, then is "broken" in any case with external symlinks.

conans/util/files.py
lasote: What about the `discarded_file()` filters? (`*.pyo`, `*.pyc`, `.DS_Store`, `__conan`) (And if we remove all, we could remove the `keep_python = get_env("CONAN_KEEP_PYTHON_FILES", False)`)

conans/model/manifest.py
memsharded: Oh, I see! I was confused by ``md5`` vs ``md5sum``. 
Now makes sense, thanks.

conans/util/files.py
memsharded: all the Python .pyc, .pyo have been already removed from ``discarded_file()``, right? It only keeps the .DS_Store which is really annoying in OSX. That is the thing, it seems a bit assymetric to keep filtering ``__pycache__``, but allow ``.pyc``?

lasote: Yeah, ok. I'll remove it. This is develop2, should we remove the `CONAN_KEEP_PYTHON_FILES` then?

lasote: Sorry, I was looking at "develop" branch by mistake, sure, let's remove it.

czoido (approved):

PR #6633, opened 2020-03-05
Add a test that checks dynamic imports

Minimonium:
Changelog: Bugfix: Fix dynamic imports usage with config installation in the Python API
Docs: Omit

Fixes #6620

Issue #4687, opened 2019-03-06
Share code between 'hook_manager' and 'loader' to dynamic load python modules


jgsogo:
The code to load a conanfile or a hook is exactly the same (the only difference is that the former one uses `imp.load_sources` while the other uses `__import__`). It can be refactored to a common function.

memsharded:
Closed by https://github.com/conan-io/conan/pull/10127

Issue #3441, opened 2018-08-30
imp.load_source replacement


memsharded:
That function is already deprecated/undocumented.

https://github.com/conan-io/conan/pull/3340 tried to replace it with __import__, but failing in OSX Py3.6. 
This has to be revisited and probably fixed for 2.0.

danimtb:
Could be also replace with:

```
from importlib.machinery import SourceFileLoader

loaded = SourceFileLoader(filename, conanfile_path).load_module()
```

We we drop support for python 2

danimtb:
SourceFileLoader().load_module() seems to be deprecated too.
https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module

This could be the solution for python 3:

```
from importlib.machinery import SourceFileLoader
import types
loader = SourceFileLoader(filename, plugin_path)
loaded = types.ModuleType(loader.name)
loader.exec_module(loaded)
```

alalazo:
@memsharded Just stumbled on this issue. As I faced exactly the same problem contributing to a different package manager, was wondering if this snippet might be useful to you:

 https://github.com/epfl-scitas/spack/blob/af6a3556c4c861148b8e1adc2637685932f4b08a/lib/spack/llnl/util/lang.py#L595-L622

sztomi:
I've run into this while trying to parallelize recipe exporting with the following script:

```py
import sys
from pathlib import Path
from concurrent.futures import as_completed, ThreadPoolExecutor

from conans.client.conan_api import ConanAPIV1, _get_conanfile_path, cmd_export


def export_all(path: Path, namespace: str):
  conan = ConanAPIV1()
  conan.create_app()

  user, channel = namespace.split("/")
  sys.path.insert(0, str(Path.cwd().absolute() / "plexconantool"))

  futures = []
  path = Path(path)
  conanfiles = list(sorted(
    _get_conanfile_path(str(subdir.absolute()), cwd=None, py=True)
    for subdir in path.iterdir()
  ))

  with ThreadPoolExecutor(max_workers=12) as pool:
    for cf in conanfiles:
      ftr = pool.submit(
        cmd_export,
        conan.app,
        cf,
        name=None,
        version=None,
        user=user,
        channel=channel,
        keep_source=False,
        graph_lock=None,
        ignore_dirty=False,
      )
      futures.append(ftr)

  for ftr in as_completed(futures):
    ftr.result()


if __name__ == "__main__":
  from argparse import ArgumentParser

[...]

Issue #6620, opened 2020-03-03
A sporadic import failure in the CI when adding a global hook


Minimonium:
Prerequisite for the #6434 

When adding a hook execution into the api layer - a seemingly unrelated test started to fail with a failure to import a module stored in the `python` cache subfolder:
https://conan-ci.jfrog.info/blue/organizations/jenkins/ConanTestSuite/detail/PR-6434/1/pipeline

The test under question:
https://github.com/conan-io/conan/blob/develop/conans/test/functional/command/config_install_test.py#L187

I was able to reproduce the problem locally. Debugging shown that despite all files being identically present in both with and without the hook situations - the import is failing. Chipping the problem down shown that the problem is somehow related to the `_load_module_from_file` being present.

After some investigation, I discovered the https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches - which sounds like the tool for the job since in tests we do append the `sys.path` with folders created within the same Python interpreter instance, which sounds pretty dynamic to me. 
Also, a nice write up on the problem: http://ballingt.com/import-invalidate-caches/

And it did fix the problem for the Python 3.8 interpreter. But it failed on Python 2.7 since it's not implemented in there and the Python 3.6, for some other weird reason. 

After additional investigation, I have found this Python issue: https://bugs.python.org/issue33169 

> Note that it's solved for Python 3.7+.

The most interesting to our use case note: 

> importlib.invalidate_caches() doesn't clear the negative cache in sys.path_importer_cache.

And indeed! At the moment when a sporadic import fails - the `path_import_cache` contains a `None` value for the temp folder key. Here is a description of how they have reproduced it:
https://github.com/python/mypy/pull/4811

In #6434 I copied the "fixed" cache invalidation from the 3.7+:
https://github.com/python/cpython/blob/521995205a2cb6b504fe0e39af22a81f785350a3/Lib/importlib/_bootstrap_external.py#L1180-L1186

And for Python 2.7 I've used the proposed solution from the issue above:
https://github.com/python/mypy/blob/a42ee8f0697439eac3a86644729750c472438b98/mypy/test/teststubgen.py#L147-L153

After that, we've got a green CI.

Sadly, I wasn't able to make a minimal example of the problem since the caching mechanism in Python is not known to me. So anyone interested would need to debug the #6434 (comment out the currently proposed solution for cache invalidation).

[...]

PR #10127, opened 2021-12-03
upgraded to importlib

memsharded:
Includes the ``invalidate_caches()`` call

Close https://github.com/conan-io/conan/issues/4687
Close https://github.com/conan-io/conan/issues/3441
Close https://github.com/conan-io/conan/pull/6633

@Minimonium this should address previous concerns about python loading, note ``invalidate_caches()`` is called in the constructor of ``ConanFileLoader()``, a new one will be created with each ``ConanApp()`` which is now explicitly created inside each ``api`` call, so I'd say we are ok now.

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -1,4 +1,4 @@
-import imp
+from importlib import machinery, invalidate_caches
 import inspect
 import os
 import sys
@@ -9,8 +9,7 @@ import yaml
 from conans.client.conf.required_version import validate_conan_version
 from conans.client.loader_txt import ConanFileTextLoader
 from conans.client.tools.files import chdir
-from conans.errors import ConanException, NotFoundException, ConanInvalidConfiguration, \
-    conanfile_exception_formatter
+from conans.errors import ConanException, NotFoundException, conanfile_exception_formatter
 from conans.model.conan_file import ConanFile
 from conans.model.options import Options
 from conans.model.recipe_ref import RecipeReference
@@ -25,6 +24,7 @@ class ConanFileLoader:
         self._pyreq_loader = pyreq_loader
         self._cached_conanfile_classes = {}
         self._requester = requester
+        invalidate_caches()
 
     def load_basic(self, conanfile_path, graph_lock=None, display=""):
         """ loads a conanfile basic object without evaluating anything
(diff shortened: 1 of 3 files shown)


Minimonium:
Great! 😄 

Be aware of the pre 3.7.0 bug in invalidate_cache tho! 
https://github.com/conan-io/conan/pull/6633/files#diff-c15f4aacf65aa58b31a041bac63e4dd7337355d9ec44a8f32b2a5815d5eb5ea9R210

memsharded:
Thanks for the pointer to https://github.com/conan-io/conan/pull/6633/files#diff-c15f4aacf65aa58b31a041bac63e4dd7337355d9ec44a8f32b2a5815d5eb5ea9R210

But I have some questions:

- Shouldn't the clean of the ``sys.path_importer_cache`` be done added (instead of replacing) the ``invalidate-caches()``?
- It would be great to have a failing test for this bug. The test provided in your previous PR doens't make sense anymore because we removed the global importing of files from the cache. Could you provide some hint I could write a test that fails for this?

Thanks!

Minimonium:
It's been quite a long time since then, so I may miss the fine details, sorry about that. 

I'm pretty sure the only reasoning to branch them out was the core issue of the lack of invalidation of the "negative cache". At the time it was "either a small-scope hack or a conventional tool" for the same purpose. I think before opt-in to either there must be a clear understanding of use-cases where it does help. 

The issue at the time was specifically with cache reuse inside a single Python invocation, a call to user code in the middle that introduces cache side-effects like `conan export .`, [specifically the "negative cache" issue](https://github.com/conan-io/conan/issues/6620#issuecomment-595429710). An example of that was a naive implementation of the automatic config installation because it contained user-provided hooks, which are modules that Conan explicitly did load.

memsharded:
Ok I see.

I think the best would be to write some tests when the ``config install`` functionality is moved to the new PythonAPI, then we will be able to write tests exactly as the potential user would do, alternating ``config install`` calls for hooks with other calls in the PythonAPI. That will surface potential real issues there.

As this is already looking as some progress, I'd probably don't wait to merge it till then, I think this could be merged asap.

conans/client/loader.py
```diff
@@ -1,4 +1,4 @@
-import imp
+from importlib import invalidate_caches, util
```
franramirez688: `util` import is a generic name, please, rename it as `imp_util` or another more specific one

franramirez688 (approved):

conans/client/loader.py
memsharded: done

czoido (approved):

PR #10129, opened 2021-12-05
Change error to info and clarify wording

### LINKED ISSUE #8035: [suggestion] failure to auto-detect a compiler for the default profile should not report 'ERROR:'
Conan outputs that there has been an error when no compiler is found during auto-detect:

```
conan profile new default --detect 
Profile created with detected settings: C:\Jenkins\workspace\my_job\.conan\profiles\default
ERROR: Unable to find a working compiler
```

This happens often because my compiler is not installed on the machine - I am using a `build_requires` package for my cross-compiler toolchain and `conan config install ...` to get the profile which adds it.  But I do still want to auto-detect to get the correct values for `os`, `arch`, etc. based on the build machine.

I find this message misleading and developers often mistake this for the reason why jobs fail, rather than looking further down the build log...  and it only becomes more confusing when creating a package which does not have a need for the `compiler` setting at all (for example, when packaging `build_requires`).

I would like to see it re-worded or even removed.  I guess any problems would become clear on `conan install` anyway.

Thanks!

bionfinc:
Changelog: Omit
Docs: Omit
fixes #8035

### DIFF
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -223,7 +223,7 @@ def _detect_compiler_version(result, output, profile_path):
     except Exception:
         compiler, version = None, None
     if not compiler or not version:
-        output.error("Unable to find a working compiler")
+        output.info("No compiler was detected (one may not be needed)")
         return
 
     # Visual Studio 2022 onwards, detect as a new compiler "msvc"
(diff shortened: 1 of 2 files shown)


memsharded:
Thanks for your contribution @bionfinc 

It seems your change broke a test (https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-10129/1/pipeline), because the mock is not supporting that ``info`` method. Could you please have a look to the test or ask for help if necessary?

bionfinc:
Thank you @memsharded, I apologize it didn't occur to me that this might break a test. I believe all the tests are passing now

czoido (approved):

PR #10135, opened 2021-12-07
minor cleanings and refactors

memsharded:
- Avoid converting to ``str()`` in all ``conanfile_exception_formatter()``, so we will be able to read inside some ``conanfile.conf`` if necessary (trace verbose?)
- Avoid ``dedent()`` and ``normalize()`` of default ``settings.yml``
- Removed unused XXXX_FOLDER (and move to SERVER_XXX_FOLDER when applies, to make it clear the difference)
- Removed some xfailed tests, so broken and complicated, very difficult to recover
- Removed 1.X ConanV2Exception
- Removed allowing ``scm`` different than dictionary


conans/client/cache/cache.py
```diff
@@ -291,7 +290,9 @@ def reset_default_profile(self):
 
     def initialize_settings(self):
         if not os.path.exists(self.settings_path):
-            save(self.settings_path, normalize(get_default_settings_yml()))
+            settings_yml = default_settings_yml
```
franramirez688, line 293: Could we make the same for `get_default_client_conf` to simplify the `normalize(get_default_client_conf()`?

franramirez688 (approved):
Well done! 👍 
Only one comment regarding a similar possible refactor for `normalize(get_default_client_conf()`

conans/client/cache/cache.py
memsharded: Not worth it, the ``conan.conf`` is going to dissapear, be replaced by new ``global.conf`` and profiles conf. It will eventually be completely removed when the remaining configs in it are migrated to the new stuff.

PR #10150, opened 2021-12-09
[fix] Removed CMake warning message

### LINKED ISSUE #10132: [bug] CMakeToolchain/CMakeDeps generates CMake warnings when dependency was already found via find_package
  * Operating System+version: Linux Mint 20.1 Ulyssa
  * Compiler+version: gcc 11.1.0
  * Conan version: 1.43.0
  * Python version: 3.8.10

Use the CMakeToolchain and CMakeDeps generator.
Use the generated conan_toolchain.cmake file then `find_package` for a dependency and then `find_package` for a transitive dependency. E.g.
```cmake
find_package(PNG 1.6 REQUIRED)
find_package(Freetype 2.10 REQUIRED)
```

In this example, Freetype depends on PNG.

This could possibly be avoided by omitting the `find_package` for `PNG` (in this case), but that's not a good solution in case the consumer depends on `PNG` (or other transitively dependent library) and can't depend on the library being included transitively.

Also, adding `QUIET` to the `find_package` calls does not suppress the warning because they are emitted directly via `message(WARNING)` and so cannot be suppressed with that flag.

```
CMake Warning at cmake-build-debug/zlibTargets.cmake:24 (message):
  Target name 'zlib::zlib' already exists.
Call Stack (most recent call first):
  cmake-build-debug/zlib-config.cmake:11 (include)
  [omitted path to cmake]/cmake-3.21/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  cmake-build-debug/libpng-config.cmake:16 (find_dependency)
  [omitted path to cmake]/cmake-3.21/Modules/CMakeFindDependencyMacro.cmake:47 (find_package)
  cmake-build-debug/freetype-config.cmake:16 (find_dependency)
  CMakeLists.txt:83 (find_package)
  ```

### PR DESCRIPTION
**Pull request by franramirez688**
Changelog: Fix: Deleted CMake warning for already existing targets.
Closes: https://github.com/conan-io/conan/issues/10132
Docs: omit

### CODE REVIEW

File: conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aliases.py [hunk 1]
```diff
@@ -148,7 +148,7 @@ def package_info(self):
     client.save({"conanfile.py": consumer, "CMakeLists.txt": cmakelists})
     client.run("create .")
 
-    assert "Target name 'hello::hello' already exists." in client.out
+    assert "Target name 'hello::hello' already exists." not in client.out
```
franramirez688, line 151: I was doubting if keeping these tests or not... I think it could be helpful to have them. Wdyt?

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
```diff
@@ -899,4 +899,4 @@ def build(self):
     client.save({"conanfile.py": consumer, "CMakeLists.txt": cmakelists})
     client.run("create .")
 
-    assert "Target name 'mypkg::name' already exists." in client.out
+    assert "Target name 'mypkg::name' already exists." not in client.out
```
franramirez688: I was doubting if keeping these tests or not... I think it could be helpful to have them. Wdyt?

memsharded (approved):

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components_names.py
memsharded: Negative "not in" checks are low value, because of false positives. The message can change to "hello::hello exists" and it would still be true. Not harm to leave them though, you can leave them, no problem.

franramirez688: You're right. Then, let's get rid of it.

PR #10156, opened 2021-12-10
[develop2] conditional sub-option removal or set test

memsharded:
Response to https://github.com/conan-io/conan/issues/10104#issuecomment-991071661

### DIFF
--- a/conans/test/integration/options/options_test.py
+++ b/conans/test/integration/options/options_test.py
@@ -257,3 +257,35 @@ equal:opt=a=b
         client.save({"conanfile.py": consumer})
         client.run("create . pkg/0.1@user/testing")
         self.assertIn("pkg/0.1@user/testing: Created package ", client.out)
+
+    def test_define_nested_option_not_freeze(self):
+        c = TestClient()
+        conanfile = textwrap.dedent("""
+            from conan import ConanFile
+            class Pkg(ConanFile):
+                options = {"without_stacktrace": [True, False],
+                           "with_stacktrace_backtrace": [True, False]}
+                default_options = {"without_stacktrace": True}
+                def configure(self):
+                    if self.options.without_stacktrace:
+                        del self.options.with_stacktrace_backtrace
+                    else:
+                        self.options.with_stacktrace_backtrace = True
+
+                def build(self):
+                    s = self.options.without_stacktrace
+                    self.output.info("without_stacktrace: {}".format(s))
+
+                    if "with_stacktrace_backtrace" in self.options:
+                        ss = self.options.get_safe("with_stacktrace_backtrace")
+                        self.output.info("with_stacktrace_backtrace: {}".format(ss))
+                    else:
+                        self.output.info("with_stacktrace_backtrace success deleted!")
+            """)
+        c.save({"conanfile.py": conanfile})
+        c.run("create . pkg/0.1@")
+        assert "pkg/0.1: without_stacktrace: True" in c.out
+        assert "pkg/0.1: with_stacktrace_backtrace success deleted!" in c.out
+        c.run("create . pkg/0.1@ -o pkg:without_stacktrace=False")
+        assert "pkg/0.1: without_stacktrace: True" in c.out
+        assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out
--- a/conans/test/integration/options/options_test.py
+++ b/conans/test/integration/options/options_test.py
@@ -287,5 +287,5 @@ equal:opt=a=b
         assert "pkg/0.1: without_stacktrace: True" in c.out
         assert "pkg/0.1: with_stacktrace_backtrace success deleted!" in c.out
         c.run("create . pkg/0.1@ -o pkg:without_stacktrace=False")
-        assert "pkg/0.1: without_stacktrace: True" in c.out
+        assert "pkg/0.1: without_stacktrace: False" in c.out
         assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out


conans/test/integration/options/options_test.py
```diff
@@ -257,3 +257,35 @@ def package_id(self):
...
+                    else:
+                        self.output.info("with_stacktrace_backtrace success deleted!")
+            """)
+        c.save({"conanfile.py": conanfile})
+        c.run("create . pkg/0.1@")
+        assert "pkg/0.1: without_stacktrace: True" in c.out
+        assert "pkg/0.1: with_stacktrace_backtrace success deleted!" in c.out
+        c.run("create . pkg/0.1@ -o pkg:without_stacktrace=False")
+        assert "pkg/0.1: without_stacktrace: False" in c.out
+        assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out
```
babariviere, line 291: Replace `assert ... in ...` with `self.assertIn` as in the other test cases?

memsharded: Nop. ``assert`` is new ``pytest`` style framework, which is replacing ``unittests``, which are the one using ``self.assertIn``, that will dissapear as we keep moving tests to ``pytest``

babariviere:
@memsharded What should we do if the user specify `with_stacktrace_backtrace` via conan command line or via the profile?

The current behavior of the `boost` package is to delete `with_stacktrace_backtrace` when `without_stacktrace` is set to True.

IMO, the user should only provide `without_stacktrace`, but the error message is not clear enough. So I propose something like this:

```python
from conan import ConanFile
class Pkg(ConanFile):
    options = {"without_stacktrace": [True, False],
               "with_stacktrace_backtrace": [True, False]}
    default_options = {"without_stacktrace": False}

    def configure(self):
        if self.options.with_stacktrace_backtrace != None:
            raise ConanException("pkg: with_stacktrace_backtrace shouldn't be set, use without_stacktrace instead")
        elif self.options.without_stacktrace and not self.options.with_stacktrace_backtrace:
            del self.options.with_stacktrace_backtrace
        else:
            self.options.with_stacktrace_backtrace = True
    def build(self):
        s = self.options.without_stacktrace
        self.output.info("without_stacktrace: {}".format(s))
        if "with_stacktrace_backtrace" in self.options:
            ss = self.options.get_safe("with_stacktrace_backtrace")
            self.output.info("with_stacktrace_backtrace: {}".format(ss))
        else:
            self.output.info("with_stacktrace_backtrace success deleted!")

```

With this, we can have this test case:

```python
        c.save({"conanfile.py": conanfile})
        c.run("create . pkg/0.1@")
        assert "pkg/0.1: without_stacktrace: False" in c.out
        assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out
        c.run("create . pkg/0.1@ -o pkg:without_stacktrace=False")
        assert "pkg/0.1: without_stacktrace: False" in c.out
        assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out
        c.run("create . pkg/0.1@ -o pkg:without_stacktrace=True")
        assert "pkg/0.1: without_stacktrace: True" in c.out
        assert "pkg/0.1: with_stacktrace_backtrace success deleted!" in c.out
        with pytest.raises(Exception):
            c.run("create . pkg/0.1@ -o pkg:with_stacktrace_backtrace=False")
            assert "pkg/0.1: without_stacktrace: False" in c.out
            assert "pkg/0.1: with_stacktrace_backtrace: True" in c.out
```

(I didn't check on how to check for the message output on the last test case)

What do you think?
I know it's only a test case, but it can be a good reference for future recipes if it's good enough.

memsharded:
Yes, maybe is enough with a 1 liner assert:

```python
def configure(self):
        assert self.options.with_stacktrace_backtrace == None, "don't set with_stacktrace_backtrace, use without_stacktrace"
        if self.options.without_stacktrace:  # condition can be simplified, we already checked is not None
            del self.options.with_stacktrace_backtrace
        else:
            self.options.with_stacktrace_backtrace = True
```

In any case, I am missing something here. If ``with_stacktrace_backtrace`` is not meant to be configurable, why making it an option anyway? It doesn't make sense, it  can be directly deduced when necessary from ``without_stacktrace``?

babariviere:
@memsharded This would involve a refactor of the boost recipe. I don't know why it was done this way.
So, should we just accept that packages must be refactor, and thus, options will have to change? (in this case, remove all options starting `without_` in boost).

memsharded:
> @memsharded This would involve a refactor of the boost recipe. I don't know why it was done this way.
So, should we just accept that packages must be refactor, and thus, options will have to change? (in this case, remove all options starting without_ in boost).

Recipes will need to refactor anyway to be ready for 2.0, to migrate to new ``conan.tools.xxxx`` integrations, etc.
I think so, if the options are really not options (because users cannot change ever their values), then, they shouldn't be options, and the recipe should be fixed. We are in the moment of improving things and fixing lots of legacy and bad defaults for 2.0, we are fixing a lot of ugly stuff with options, so far this seems it might be a bit of abuse of the options feature that would disallow such improvements, so unless I am missing some use case, I'd say yes, the recipes should be improved.

babariviere:
Good point. Thanks, that was very instructive!

memsharded:
@babariviere I think I will merge this test to capture the behavior and possibility into alpha3, if you happen to come up with new insights or feedback, I think the best would be to open a new issue (as this PR will be merged)

babariviere:
@memsharded Thanks! Will do more feedback if I have any. 😄

PR #10157, opened 2021-12-10
Feature/develop2 dates

memsharded:
As ``python-dateutil`` was used exactly for 1 line in our codebase, I was trying to get rid of it. But happens that parsing the ISO format is complicated enough, so I didn't succeed at first try. This PR just do some cleanings and minor changes:

- Removing legacy tests, so broken that they will not be fixed, need to write new ones
- Removing ``conan_server`` revisions migration in storing revisions timestamps
- Using ``.repr_humantime()`` in a couple of cases
- Removing dead ``iso8601_to_str()``
- Marking as private one function.

I am still not very happy with the fact that servers convert their internally stored timestamps to ISO format to send it via HTTP API to be converted back to timestamps in the client side. Adds complexity, dependencies (python-dateutil) and unnecessary computation cost on both sides, for the extremely hypothetical scenario of someone analyzing the HTTP traffic and wanting to read a ISO date instead of a timestamp. Still convinced about it @lasote?


lasote:
>  Still convinced about it @lasote? 

Well, I have no new arguments more than you asking again:
   - The `unnecessary computation cost on both sides`... converting a date is not mining bitcoins..........
   - The complexity: You said it was one line I would say two and pretty encapsulated in the REST layer.
   - The external dependency: Yes. Looks like Python 3.7 have already something for that, so we cannot use it yet. But it is a bit annoying that we want to implement everything ourselves when we are developing a package manager. 
  
Said that I don't really care if we get rid of it. Just saying that it is the facto standard to use in HTTP communication.

lasote (approved):

memsharded:
> Said that I don't really care if we get rid of it. Just saying that it is the facto standard to use in HTTP communication.

Ok, lets leave it as-is, I think it will be worth re-considering when changes in the HTTP api in the server side are proposed.

PR #10159, opened 2021-12-11
simple upgrade to jinja v3

### LINKED ISSUE #9907: [feature] Compatibility with Jinja2 V3

Hi,

Can you add support for Jinja2 V3? Because today, all projects with Jinja2 V3 have problems with conan:

```
conan 1.40.3 requires Jinja2<3,>=2.9, but you have jinja2 3.0.2 which is incompatible.
```

Thanks.

nilsnolde:
Changelog: Feature: Upgrade Conan python jinja requirement to v3.x.
Docs: omit
Fixes #9907 

#tags: slow

Makes conan more compatible with other libs requiring `jinja` in the same environment. The major version bump seems to e mostly stemming from dropping py2 & 3.5 support, didn't see any other breaking change.

### DIFF
--- a/conans/requirements.txt
+++ b/conans/requirements.txt
@@ -10,5 +10,5 @@ node-semver==0.6.1
 distro>=1.0.2, <=1.6.0
 pygments>=2.0, <3.0
 tqdm>=4.28.1, <5
-Jinja2>=2.9, <3
+Jinja2>=2.9, <4.0.0
 python-dateutil>=2.7.0, <3


nilsnolde:
I did run the tests, but even one of the first ones wasn't running successfully locally (`conans/test/functional/tools_versions_test.py`, also before the change), my hopes are high that this is the only required change though and CI will pass on first try.

memsharded:
Good to know that this is not breaking, I expected a major version to break something. Also, a bit surprising that our Py27 builds keep building with it, is it possible that it is working for Py27 even if they removed support?

I have added the ``tags: slow`` thing, so more thorough tests are executed.

memsharded:
This is the reason:

> Collecting Jinja2<4.0.0,>=2.9
> Using cached Jinja2-2.11.3-py2.py3-none-any.whl (125 kB)


Not sure how to force refreshing it in CI @czoido any suggestion?

czoido:
> This is the reason:
> 
> > Collecting Jinja2<4.0.0,>=2.9
> > Using cached Jinja2-2.11.3-py2.py3-none-any.whl (125 kB)
> 
> Not sure how to force refreshing it in CI @czoido any suggestion?

I think that's expected, It's taking a cached Jinja but it's resolving the ranges right, takes the last Jinja2 [version](https://pypi.org/project/Jinja2/#history) compatible with Python2 in the range `Jinja2>=2.9, <4.0.0` which is `2.11.3`. We should probably add the `--no-cache-dir` to the pip install in ci but the results would be the same.

memsharded:
> I think that's expected, It's taking a cached Jinja but it's resolving the ranges right, takes the last Jinja2 version compatible with Python2 in the range Jinja2>=2.9, <4.0.0 which is 2.11.3. We should probably add the --no-cache-dir to the pip install in ci but the results would be the same.

Why would be the same? If pip install Jinja v3, which might have breaking changes for Conan, then CI might fail right?

czoido:
> > I think that's expected, It's taking a cached Jinja but it's resolving the ranges right, takes the last Jinja2 version compatible with Python2 in the range Jinja2>=2.9, <4.0.0 which is 2.11.3. We should probably add the --no-cache-dir to the pip install in ci but the results would be the same.
> 
> Why would be the same? If pip install Jinja v3, which might have breaking changes for Conan, then CI might fail right?

The CI is running Jinja2 3.0.3 for all nodes but the Linux Py2.7 one, it looks like Jinja2 v3 is not breaking us.

memsharded:
Oh, perfect. You are right, my bad then I was only checking the first config, which yes, happens to be py27. Thanks very much @czoido

memsharded (approved):

czoido (approved):

PR #10165, opened 2021-12-13
fixing definition of CXX11_ABI for gcc-like

memsharded:
Changelog: Bugfix: Fix the definition of D_GLIBCXX_USE_CXX11_ABI in gcc-like compilers for ``CMakeToolchain`` and ``AutotoolsToolchain``. Define it only to ``D_GLIBCXX_USE_CXX11_ABI=0`` for new compilers, assuming that the default is alread 1.
Docs: Omit

Address https://github.com/conan-io/conan-center-index/pull/8389

Lets see if something breaks, need to add tests.

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -187,11 +187,9 @@ class GLibCXXBlock(Block):
                    }.get(libcxx)
             if lib:
                 lib = "-library={}".format(lib)
-        elif compiler == "gcc":
-            # we might want to remove this "1", it is the default in most distros
-            if libcxx == "libstdc++11":
-                glib = "1"
-            elif libcxx == "libstdc++":
+
+        if compiler in ['clang', 'apple-clang', 'gcc']:
+            if libcxx == "libstdc++":
                 glib = "0"
         return {"set_libcxx": lib, "glibcxx": glib}
 
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -191,6 +191,8 @@ class GLibCXXBlock(Block):
         if compiler in ['clang', 'apple-clang', 'gcc']:
             if libcxx == "libstdc++":
                 glib = "0"
+            elif libcxx == "libstdc++11" and self._conanfile.conf["tools.gnu:define_libcxx11_abi"]:
+                glib = "1"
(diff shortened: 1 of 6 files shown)


conan/tools/cmake/toolchain.py
```diff
@@ -187,11 +187,9 @@ def context(self):
...
             if lib:
                 lib = "-library={}".format(lib)
-        elif compiler == "gcc":
-            # we might want to remove this "1", it is the default in most distros
-            if libcxx == "libstdc++11":
-                glib = "1"
-            elif libcxx == "libstdc++":
+
+        if compiler in ['clang', 'apple-clang', 'gcc']:
+            if libcxx == "libstdc++":
```
memsharded, line 192: Lets be aware that from https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

```
Using the default configuration options for GCC the default value of the macro is 1 which causes the new ABI to be active, so to use the old ABI you must explicitly define the macro to 0 before including any library headers. (Be aware that some GNU/Linux distributions configure GCC 5 differently so that the default value of the macro is 0 and users must define it to 1 to enable the new ABI.)
```
So if we go this way, users from those distros will need to customize the ``CMakeToolchain``, something like this:

```python
def generate(self):
    tc = CMakeToolchain(self)
    tc.blocks["libcxx"].values["glibcxx"] = "1" if self.settings.libcxx == "libstdc++11" else "0"
    tc.generate()
```

There were requests to eliminate the default ``GLIBCXX_USE_CXX11_ABI=1``, as it was raising some warnings (undefined in some compilers, but not errors), so it seems there is no perfect solution for 100% of users.

lasote (approved):
I would add a test with the `generate` you suggest in case someone uses a custom distro. Also, we have to document this default.

memsharded:
> I would add a test with the generate you suggest in case someone uses a custom distro. Also, we have to document this default.

I have added tests for both ``CMakeToolchain`` and ``AutotoolsToolchain``, probably needs to be added to Meson, but I see no trace of libcxx there, so that would need more time.

I have also added a new conf ``tools.gnu:define_libcxx11_abi``, as it seems it would be very inconvenient for users of ConanCenter recipes, if they don't want to fork and modify those recipes, for example. What do you think?

lasote (approved):
👍  Could be merged

PR #10166, opened 2021-12-13
[develop2][poc] Checking the ``[system_tools]`` idea

memsharded:
Changelog: Feature: Add `[system_tools]` section to profiles to use your own installed tools instead of the packages declared in the requires.

Very preliminary idea, just toying to see what could make sense


lasote (approved):
This is an interesting idea and simple in code. But probably we are missing corner cases or missing use cases. This would be the typical "experimental" feature I would introduce to test if useful and complete enough but I'm not sure how should we proceed with that kind of experimental feature in an Alpha.

memsharded:
> This is an interesting idea and simple in code. But probably we are missing corner cases or missing use cases. This would be the typical "experimental" feature I would introduce to test if useful and complete enough but I'm not sure how should we proceed with that kind of experimental feature in an Alpha.

Then I suggest, as a first step, try to build a bit more exhaustive test suite for this feature, trying to raise possible corner cases and issues. I'll work on that.

conans/test/integration/graph/test_deferred.py
```diff
@@ -0,0 +1,92 @@
...
+from conans.util.files import save
+
+
+class TestToolRequires:
+    def test_deferred(self):
+        client = TestClient()
+        client.save({"conanfile.py": GenConanfile().with_tool_requires("tool/1.0"),
+                     "profile": "[deferred_requires]\ntool/1.0"})
+        client.run("create . pkg/1.0@ -pr=profile")
+        assert "tool/1.0:357add7d387f11a959f3ee7d4fc9c2487dbaa604 - Deferred" in client.out
```
lasote: How is the package id calculated for a deferred node?

memsharded: Same as others. As the package recipe is basically empty (no settings, no options, no dependencies), the package_id of the deferred will always be the same, we name it in our tests as: ``NO_SETTINGS_PACKAGE_ID = "357add7d387f11a959f3ee7d4fc9c2487dbaa604"``

conans/client/graph/graph_builder.py
```diff
@@ -212,15 +214,33 @@ def _resolve_recipe(self, ref, graph_lock):
...
+
+        if resolved is None:
+            try:
+                # TODO: If it is locked not resolve range
+                # TODO: This range-resolve might resolve in a given remote or cache
+                # Make sure next _resolve_recipe use it
+                self._resolver.resolve(require, str(node.ref), self._remotes, self._update)
+                resolved = self._resolve_recipe(require.ref, graph_lock)
+            except ConanException as e:
+                raise GraphError.missing(node, require, str(e))
```
memsharded, line 243: All this part doesnt change, just putting it inside the ``if resolved is None``

conans/client/profile_loader.py
```diff
@@ -261,6 +261,12 @@ def get_profile(profile_text, base_profile=None):
         options = Options.loads(doc.options) if doc.options else None
         tool_requires = _ProfileValueParser._parse_tool_requires(doc)
 
+        if doc.system_tools:
+            system_tools = [RecipeReference.loads(r.strip())
```
AbrilRBS, line 265: For anyone wondering as I did if this could be improved, not until we support 3.8 with https://peps.python.org/pep-0572/ , which would let us write something like:
```py
system_tools = [RecipeReference.loads(stripped) for r in doc.system_tools.splitlines() if (stripped := r.strip())]
```

AbrilRBS (approved):
Looks really nice, and as discussed, will help a ton

czoido (approved):

franramirez688 (approved):

PR #10171, opened 2021-12-13
`test_type` mechanism deprecation

### LINKED ISSUE #9606: [question] How to find out the name of the package to be tested in test_package?
In a `test_package` I can access all dependencies via `self.deps_cpp_info`, but is there any way I can determine which one is the subject under test?

### LINKED ISSUE #10073: [feature] test package conanfile, `test_type` mechanism deprecation
Conan has the `test_type` mechanism so we can specify in the conanfile of a `test package` the type of requirement that we want to test.
The `test_type` allows `test_type = "build_requires"` or `test_type="requires"` but that doesn't cover the current variety of requirements we could have in 2.0 with the requirements traits. 

So we want to introduce a mechanism in `1.X` to be prepared for the removal of `test_type` in Conan 2.0 to receive (injected) in the conanfile a variable with the reference to be tested (so in case the version is bumped we don't need to update the test_package for example) and then write the requirement explicitly, using that variable. Something like:

```
def MyTestPackage(ConanFile):

      def requirements(self):
          self.build_requires(self.testing_package_reference)

```

lasote:
Changelog: Feature: In the conanfile.py of the test_package, the reference being tested is always available at `self.tested_reference_str`.
Changelog: Feature: Introduced a new `test_type` value `explicit` so a user can declare explicitly the `requires` or `build_requires` manually (using `self.tested_reference_str`), it won't be automatically injected as a require. In Conan 2.0 the `test_type` attribute will be ignored, the behavior will be always explicit, so declaring `test_type="explicit"` will make the test recipe compatible with Conan 2.0.
Docs: https://github.com/conan-io/docs/pull/2341

Close https://github.com/conan-io/conan/issues/10073
Close https://github.com/conan-io/conan/issues/9606

### DIFF
--- a/conans/client/graph/graph_manager.py
+++ b/conans/client/graph/graph_manager.py
@@ -100,7 +100,7 @@ class GraphManager(object):
             if test:
                 conanfile.display_name = "%s (test package)" % str(test)
                 conanfile.output.scope = conanfile.display_name
-                conanfile.testing_reference = str(test)
+                conanfile.tested_reference_str = repr(test)
             run_configure_method(conanfile, down_options=None, down_ref=None, ref=None)
         else:
             conanfile = self._loader.load_conanfile_txt(conanfile_path, profile_host=profile_host)
@@ -248,7 +248,7 @@ class GraphManager(object):
                                                )
         conanfile.display_name = "%s (test package)" % str(test)
         conanfile.output.scope = conanfile.display_name
-        conanfile.testing_reference = str(create_reference)
+        conanfile.tested_reference_str = repr(create_reference)
 
         # Injection of the tested reference
         test_type = getattr(conanfile, "test_type", ("requires", ))
--- a/conans/client/graph/graph_manager.py
+++ b/conans/client/graph/graph_manager.py
@@ -100,7 +100,7 @@ class GraphManager(object):
             if test:
                 conanfile.display_name = "%s (test package)" % str(test)
                 conanfile.output.scope = conanfile.display_name
-                conanfile.testing_package_reference = str(test)
+                conanfile.testing_reference = str(test)
             run_configure_method(conanfile, down_options=None, down_ref=None, ref=None)
         else:
             conanfile = self._loader.load_conanfile_txt(conanfile_path, profile_host=profile_host)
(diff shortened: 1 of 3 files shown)


conans/client/graph/graph_manager.py
```diff
@@ -100,6 +100,7 @@ def load_consumer_conanfile(self, conanfile_path, info_folder,
             if test:
                 conanfile.display_name = "%s (test package)" % str(test)
                 conanfile.output.scope = conanfile.display_name
+                conanfile.testing_reference = str(test)
```
memsharded: Why ``str()`` and not the RecipeReference object? Why not ``repr()`` so we don't lose potential information like the recipe revision?

conans/client/graph/graph_manager.py
```diff
@@ -246,26 +248,32 @@ def _load_root_test_package(self, path, create_reference, graph_lock, profile,
...
-            require = conanfile.requires.get(create_reference.name)
-            if require:
-                require.ref = require.range_ref = create_reference
-            else:
-                conanfile.requires.add_ref(create_reference)
+
+        if "explicit" in test_type:  # 2.0 mode, not automatically add the require, always explicit
+            pass
+        else:
+            if "build_requires" in test_type:
```
memsharded:
```suggestion
        if "explicit" not in test_type:  # 2.0 mode, not automatically add the require, always explicit
            if "build_requires" in test_type:
```

memsharded (commented):
Looks good to me, understood the ``explicit`` thing, yes, it is necessary for not breaking, looks good too.

conans/client/graph/graph_manager.py
lasote: Yes, I was expecting this comment to talk about it:

 If you do a `self.requires(self.testing_reference)` it fails because it expects an `str`. So we would need to enable a `ConanFileReference` as a valid argument. 

Also, I introduced the `str` here because I'm not sure we have documented the `ConanFileReference` (maybe in the `dependencies`?) and it is probably that the users are going to compare the reference so we have to know that `ConanFileReference.loads("foo/1.0") != "foo/1.0"`. So I wanted to talk about this before taking the decission.

lasote:
If we want to differentiate a future "pure reference" argument I cannot come up with something but: `tested_reference_str` I don't like it either but...

conans/model/conan_file.py
```diff
@@ -142,6 +142,9 @@ class ConanFile(object):
     # Run in windows bash
     win_bash = None
 
+    # For test_package, reference being tested
+    tested_reference_str = None
```
memsharded: As this is very specific to only test_package, I am fine if this is dynamic and not a class attribute (but fine with it too)

memsharded (approved):
Checked this approach, indeed no valid for ``python_requires``, which would need extra things, probably a 2.0 thing where the ``pkg_type = "python-require"`` could help a bit

memsharded (approved):

Issue #9713, opened 2021-10-03
[feature] do we need MesonLayout?


SSE4:
from https://github.com/conan-io/conan-center-index/pull/6678#issuecomment-892739876

```
Hooks: It's another "error" which seems to be specific to Linux, where for whatever
 reason libs and pkgconf files are installed under lib/x86_64-linux-gnu instead 
of lib (maybe a regression in MesonToolchain or new Meson helper compared to legacy Meson helper?).
```

we only have some [predefined layouts](https://docs.conan.io/en/latest/reference/conanfile/tools/layout.html#predefined-layouts), e.g. `cmake_layout`, but there is no `meson_layout` right now.
I am not completely familiar with this feature at the moment, so I will need some guidance, at very least:

- does it make sense to add `meson_layout`?
- will it solve the problem reported with different libs/pkgconf installation directories?

- [ ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Yes, if Meson while building uses some kind of predefined folder structure, it would totally make sense to implement a ``meson_layout()`` helper.

> will it solve the problem reported with different libs/pkgconf installation directories?

This is not completely clear, if this is to locate things inside the dependencies folders, then layout() is not the place, as ``layout()`` is intended to model the current project. It can define ``generators_folder`` to say where the generated files (from ``PkgConfigDeps`` for example), should go. Please clarify.

SSE4:
> This is not completely clear, if this is to locate things inside the dependencies folders, then layout() is not the place, as `layout()` is intended to model the current project. It can define `generators_folder` to say where the generated files (from `PkgConfigDeps` for example), should go. Please clarify.

this is not about dependencies, this is about recipe being built itself, as I can say.

e.g. if you create a recipe for `glib`, and do `meson install`, it will install into the `os.path.join(self.package_folder, "lib")` for most of the platforms (e.g. Windows, macOS and 32-bit Linux).
but for Linux x64 (at least for some of distros), it will install it into: `os.path.join(self.package_folder, "lib", "x86_64-linux-gnu")`

previous (old) build helper hard-coded install directory to the`os.path.join(self.package_folder, "lib")`:
https://github.com/conan-io/conan/blob/develop/conans/client/build/meson.py#L47

when we added a new build helper, I was told it has to be managed with layouts:
https://github.com/conan-io/conan/pull/8147#discussion_r535002629

[...]

PR #10174, opened 2021-12-13
started MesonToolchain improvements

### LINKED ISSUE #9710: [feature] MesonToolchain: improve cross-compilation workflow on Apple systems

memsharded:
Changelog: Feature: Improvements in ``MesonToolchain``, including some cross-building functionality.
Docs: Omit

Close #9710

I have started to do some changes to update ``MesonToolchain``, it needs some updates:

- The model shouldn't be using double quoted strings. Users should be able to do ``meson.build_type = "debug"``, and not ``meson.build_type = "'debug'"`` (double quoted).
- Formatting to Meson files (quotes, lists, ints...) should be done only at render time (context) or even better, in the template.
- Using the ``Environment`` to handle the definition of environment instead of the raw ``EnvVars`` should allow avoiding splitting them in the first place. Also avoid doing "to_meson_value" for these.
- Avoiding creating private ``self._backend`` vars that aren't used later at all. Define just the public ones like ``self.backend`` that users can change later in recipes
- It should add support for ``msvc`` new setting.

Some other minor things:
- Using many methods, many of them just 1 line are being called in just one place is too much, requiring a lot of jumping to understand what is being done, more spaguetti code. Better simplify.
- The less code the better. Without becoming cryptic or difficult to read, the whole ``_get_backend()`` method can be collapsed to 2 very simple to read lines in the constructor.

@SSE4 as you have been working in this, if you want to follow up and finish these changes, feel free to continue this work, otherwise I will keep working on it when possible, hopefully 1.44.


SSE4:
well, right now I am more interested and motivated in functional changes for users, than (mostly cosmetic) refactoring.
for instance, I'd like to move forward with #9711, there are some unaddressed questions I would be happy to discuss with you (e.g. https://github.com/conan-io/conan/pull/9711#discussion_r751071077, how can we handle blocks).
so I'd love to see MesonToolchain be usable in practice first (e.g. in the context of CCI).

conan/tools/meson/toolchain.py
```diff
@@ -69,21 +69,49 @@ class MesonToolchain(object):
 
     def __init__(self, conanfile, backend=None):
         self._conanfile = conanfile
-        self._backend = self._get_backend(backend)
-        self._build_type = self._conanfile.settings.get_safe("build_type")
-        self._base_compiler = self._conanfile.settings.get_safe("compiler.base") or \
-                              self._conanfile.settings.get_safe("compiler")
-        self._vscrt = self._conanfile.settings.get_safe("compiler.base.runtime") or \
-                      self._conanfile.settings.get_safe("compiler.runtime")
+        # Values are kept as Python built-ins so users can modify them more easily, and they are
```
SSE4, line 66: these were private (internal) variables, why do we need users to modify them?
in general, I expect users to modify inputs (in our world, settings and conf), instead of messing up with an internal structure of objects. our objects have public interfaces (e.g. `generate` method and few others), but directly modifying object fields doesn't sound like a good idea for me.

conan/tools/meson/toolchain.py
```diff
@@ -319,17 +301,12 @@ def _cross_content(self):
...
-        save(self.native_filename, self._native_content)
-
-    def _write_cross_file(self):
-        save(self.cross_filename, self._cross_content)
+        return Template(self._cross_file_template).render(context)
 
     def generate(self):
         if cross_building(self._conanfile):
-            self._write_cross_file()
+            save(self.cross_filename, self._cross_content)
```
SSE4: for me, it makes code harder to read
you always can inline all the methods, e.g. insert the content of the function into the calling function, but it doesn't always make the code easier to read.
I would say, for me, it's a good practice if the function does just one thing which is easily deduced from the function name. it's okay if some functions take one line or less (like tangent is always just a division of sine by cosine, but I'd like to write just tangent because it's a function on its own). same for meson, I prefer to have methods like `write_native_file` and `write_cross_file` which are self-descriptive, instead of reading the internals. at this level, I don't care how the writing of cross-file is exactly implemented (e.g. does it render a template), I prefer to understand the high level logic (write cross file if cross building, write native file otherwise)

conan/tools/meson/toolchain.py
memsharded: The ``self.backend``, ``self.b_vscrt`` are public variables that can be changed, but are very inconveniently assigned as doubled quoted strings, not regular strings, they should be regular strings.

A design principle of all toolchains is that they should allow easy customization, like:

```python
tc = MesonToolchain(self)
tc.b_vscrt = xxxx if something else yyyy
tc.generate()
```
This is only nice if ``b_vscrt`` is a regular string

memsharded:
> well, right now I am more interested and motivated in functional changes for users, than (mostly cosmetic) refactoring.
for instance, I'd like to move forward with #9711, there are some unaddressed questions I would be happy to discuss with you (e.g. #9711 (comment), how can we handle blocks).

This is beyond cosmetic refactoring. A good design is important, I have changed the title to "improvements", as there are many of those requested changes that are not refactors. For example having the single-quoted strings is something that if not fixed now will be bad if we freeze it as stable in 2.0. Supporting ``msvc`` compiler is also important, as we will start removing ``Visual Studio`` in 2.0 soon. It is important to build cleanly from the foundations. Having duplicated ``self._backend`` and ``self.backend`` hurts the readability and maintenance of the class (the former variable is not needed at all, it can be avoided). Formatting the values in the wrong place hurts the readability and maintenance of the class. Single line methods, called just once in the code highly increase the cognitive overhead of the reader, the same effect can be achieved by renaming some variables. Building a ``_render(template, context)`` abstraction over ``Template(template).render(context)`` is simply too much, it doesn't add any value, but creates more spagetti code. It seems there were also a couple of dead methods, not used at all.

conan/tools/meson/toolchain.py
SSE4: I agree they shouldn't contain extra quotes, that's fine, I am not arguing with that at all
I just don't like to expose too many internal implementation details to the user
e.g. I didn't see anyone actually requested `b_vscrt` to be modifiable directly
we have other (and more important) actual requests for Meson (e.g. cross-building), I'd just keep `b_vscrt` and others to be fully private members for now, as they aren't currently documented anyway, and it gives us the freedom to do breaking changes
in other words, I wouldn't expose and document them right now, until it's actually requested

SSE4: about `self.backend` and `self._backend` - two are redundant and confusing, for sure
but the method `_get_backend` fully mimics `_get_generator` from `CMakeToolchain`
(backend in meson worlds is just like generator in cmake world)
it could be consistent across different toolchains, I think

conans/test/functional/toolchains/meson/test_ios.py
```diff
@@ -56,33 +59,17 @@ def build(self):
...
                 ("arch", self.arch),
                 ("compiler", "apple-clang"),
                 ("compiler.version", "12.0"),
                 ("compiler.libcxx", "libc++")]
 
-    def env(self):
-        cc = self.xcrun.cc
-        cxx = self.xcrun.cxx
-
-        deployment_flag = apple_deployment_target_flag(self.os, self.os_version)
```
memsharded, line 68: Users shouldn't be doing this in their profile code, this is exactly the purpose of ``MesonToolchain``, lets move this logic there.

conans/test/functional/toolchains/meson/test_cross_x86.py
```diff
@@ -29,6 +29,9 @@ def config_options(self):
                 if self.settings.os == "Windows":
                     del self.options.fPIC
 
+            def layout(self):
```
memsharded: Meson needs a clean folder, it was hardcoding "build" in ``Meson`` helper directly. Now it requires defining ``layout()`` better than that.

conan/tools/meson/meson.py
```diff
@@ -1,40 +1,36 @@
...
+        meson_source_folder = self._conanfile.source_folder
+        meson_build_folder = self._conanfile.build_folder
+        if build_script_folder:
+            meson_source_folder = os.path.join(self._conanfile.source_folder, build_script_folder)
 
         cmd = "meson setup"
-        if cross_building(self._conanfile):
-            cmd += ' --cross-file "{}"'.format(MesonToolchain.cross_filename)
+        generators_folder = self._conanfile.generators_folder
+        cross = os.path.join(generators_folder, MesonToolchain.cross_filename)
```
memsharded, line 16: Using the existence of the file, as the decission if it is cross-building or not was already taken by the toolchain.

conans/test/functional/toolchains/meson/test_cross_x86.py
SSE4: see #9713. it's still unclear how to use layout to get reproducible build directories.

conans/test/functional/toolchains/meson/test_ios.py
SSE4: see #9710 (and #9711), it's opened exactly for that.
in general, I am not a fan of doing several unrelated things in one PR. let's continue the discussion on how to do blocks in the dedicated PR.

SSE4:
> This is beyond cosmetic refactoring. A good design is important, I have changed the title to "improvements", as there are many of those requested changes that are not refactors. For example having the single-quoted strings is something that if not fixed now will be bad if we freeze it as stable in 2.0. Supporting `msvc` compiler is also important, as we will start removing `Visual Studio` in 2.0 soon. It is important to build cleanly from the foundations. Having duplicated `self._backend` and `self.backend` hurts the readability and maintenance of the class (the former variable is not needed at all, it can be avoided). Formatting the values in the wrong place hurts the readability and maintenance of the class. Single line methods, called just once in the code highly increase the cognitive overhead of the reader, the same effect can be achieved by renaming some variables. Building a `_render(template, context)` abstraction over `Template(template).render(context)` is simply too much, it doesn't add any value, but creates more spagetti code. It seems there were also a couple of dead methods, not used at all.

I think we need to separate changes, at very least, the ones affecting users (e.g. changing the behavior, like MSVC compiler support) from ones doing "cosmetic" improvements (improving readability and maintenance of the class, but in general invisible to the user, like formatting in wrong places or extracting/inlining methods). otherwise, it's too hard to follow, there is no general goal, there is just a random collection of various unrelated improvements.

conan/tools/meson/toolchain.py
```diff
@@ -299,37 +262,15 @@ def _to_meson_machine(self, machine_os, machine_arch):
...
-        save(self.cross_filename, self._cross_content)
+        return context
 
     def generate(self):
-        if cross_building(self._conanfile):
-            self._write_cross_file()
-        else:
-            self._write_native_file()
+        content = Template(self._meson_file_template).render(self._context)
+        filename = self.native_filename if not self.cross_build else self.cross_filename
```
SSE4, line 235: reversed condition would read better (without `not`)

conan/tools/meson/toolchain.py
```diff
@@ -193,24 +165,22 @@ def _to_meson_cppstd(self, cppstd):
...
-            project_options.append("%s = %s" % (k, self._to_meson_value(v)))
-        project_options = "\n".join(project_options)
+
+        def _to_meson_value(value):
+            # https://mesonbuild.com/Machine-files.html#data-types
+            if isinstance(value, str):
+                return "'%s'" % value
+            elif isinstance(value, bool):
+                return 'true' if value else "false"
+            elif isinstance(value, list):
```
SSE4: it's not the same as the previous implementation, it won't work for `array`, `tuple`, `dict`, only `list`, before it used to check for `Iterable` here

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,159 @@
...
+        fpic = self._conanfile.options.get_safe("fPIC")
+        shared = self._conanfile.options.get_safe("shared")
+        self.b_staticpic = fpic if (shared is False and fpic is not None) else None
         # https://mesonbuild.com/Builtin-options.html#core-options
         # Do not adjust "debug" if already adjusted "buildtype"
-        self.buildtype = self._to_meson_build_type(self._build_type) if self._build_type else None
-        self.default_library = self._to_meson_shared(self._shared) \
-            if self._shared is not None else None
-        self.backend = self._to_meson_value(self._backend)
+        self.default_library = ("shared" if shared else "static") if shared is not None else None
```
SSE4: I don't think we should ever expose all these variables to the public.
they are not supposed to be assigned to the random values from the outside.
they only gain limited set of the values computed from the settings as the input.

memsharded: Made them private. At some point we will want to make it more flexible and customizable, but i agree on hiding them atm

conan/tools/meson/toolchain.py
memsharded: Arguments are internal to the class, and they are all lists. Better simplify it (also removing the six dependency, as it is removed in 2.0, and merging this to develop2 will be easier.

conans/test/functional/toolchains/meson/test_cross_x86.py
memsharded: It is not unclear, it is very well defined that the layout should be defined by ``layout()``, and not rely on ``Meson()`` locally defining a ``build`` folder. Whether we need more of the ``layout()`` and we could use a ``meson_layout()``, is totally possible, but we are certain that we don't want this in the Meson helper.

conan/tools/meson/meson.py
SSE4: I think it is reliable enough. of course, some process could create or delete the file in the middle of conan's execution, but realistically it's a low chance, and anyway, further execution wouldn't be correct.

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,169 @@
...
+        self._default_library = ("shared" if shared else "static") if shared is not None else None
+
+        compiler = (self._conanfile.settings.get_safe("compiler.base") or
+                    self._conanfile.settings.get_safe("compiler"))
+
+        cppstd = self._conanfile.settings.get_safe("compiler.cppstd")
+        self._cpp_std = self._to_meson_cppstd(compiler, cppstd) if cppstd else None
+
+        if compiler == "Visual Studio":
+            vscrt = self._conanfile.settings.get_safe("compiler.base.runtime") or \
```
SSE4: I think we can already drop `compiler.base` here (per our tribe)

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,169 @@
...
+        self._b_staticpic = fpic if (shared is False and fpic is not None) else None
         # https://mesonbuild.com/Builtin-options.html#core-options
         # Do not adjust "debug" if already adjusted "buildtype"
-        self.buildtype = self._to_meson_build_type(self._build_type) if self._build_type else None
-        self.default_library = self._to_meson_shared(self._shared) \
-            if self._shared is not None else None
-        self.backend = self._to_meson_value(self._backend)
+        self._default_library = ("shared" if shared else "static") if shared is not None else None
+
+        compiler = (self._conanfile.settings.get_safe("compiler.base") or
```
SSE4: I think we can already drop `compiler.base` here (per our tribe)

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,169 @@
...
+                    self._conanfile.settings.get_safe("compiler"))
+
+        cppstd = self._conanfile.settings.get_safe("compiler.cppstd")
+        self._cpp_std = self._to_meson_cppstd(compiler, cppstd) if cppstd else None
+
+        if compiler == "Visual Studio":
+            vscrt = self._conanfile.settings.get_safe("compiler.base.runtime") or \
+                    self._conanfile.settings.get_safe("compiler.runtime")
+            self._b_vscrt = {"MD": "md",
```
SSE4: that might use `.lower()`

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,169 @@
...
+            os_build, arch_build, os_host, arch_host = get_cross_building_settings(self._conanfile)
+            self.cross_build["build"] = self._to_meson_machine(os_build, arch_build)
+            self.cross_build["host"] = self._to_meson_machine(os_host, arch_host)
+            if hasattr(conanfile, 'settings_target') and conanfile.settings_target:
+                settings_target = conanfile.settings_target
+                os_target = settings_target.get_safe("os")
+                arch_target = settings_target.get_safe("arch")
+                self.cross_build["target"] = self._to_meson_machine(os_target, arch_target)
+
+        default_comp = ""
```
SSE4: I'd like to avoid anything related to the default comp here, as much as possible.
I think we already agreed to use explicit things rather than autodetection.
many people complain that conan picks the wrong compiler executable (e.g. `/usr/bin/gcc` instead of `gcc-10` or whatever).

for CL, we can make an exception, as we know it's always `cl.exe`.
for GCC/Clang, we basically, don't know, it could be just `gcc.exe`, `gcc-10.exe`, something like `arm-linux-gnueabi-gcc.exe`...
I'd prefer to hard fail instead of silently producing incorrect binaries with the wrong compiler.

and for this kind of thing, I suppose we need to follow blocks approach, as we discuss in #9710

I beg you to exclude this kind of tentative/risky changes from this PR, and continue the discussion we started before

conan/tools/meson/meson.py
```diff
@@ -1,40 +1,36 @@
...
-        build = self._conanfile.build_folder
-        if self._build_folder:
-            build = os.path.join(self._conanfile.build_folder, self._build_folder)
-        return build
-
-    def configure(self, source_folder=None):
-        source = self._conanfile.source_folder
-        if source_folder:
-            source = os.path.join(self._conanfile.source_folder, source_folder)
+    def configure(self,  build_script_folder=None):
```
SSE4: I doubt in meson world they call it `build script folder`, might be misleading for the people

conan/tools/meson/meson.py
```diff
@@ -44,13 +40,13 @@ def build(self, target=None):
...
+        meson_build_folder = self._conanfile.build_folder
+        cmd = 'meson install -C "{}"'.format(meson_build_folder)
         self._conanfile.run(cmd)
 
     def test(self):
-        cmd = 'meson test -v -C "{}"'.format(self._build_dir)
+        meson_build_folder = self._conanfile.build_folder
+        cmd = 'meson test -v -C "{}"'.format(meson_build_folder)
         # TODO: Do we need vcvars for test?
         # TODO: This should use conanrunenv, but what if meson itself is a build-require?
```
SSE4, line 48: it's actually most likely, meson is a build requirement

conan/tools/meson/toolchain.py
memsharded: Not yet. The proposal is ready, but haven't been voted yet.

conan/tools/meson/toolchain.py
memsharded: This is not true. This is not auto-detection, this is a very sane default.
The information is already defined in the ``settings.compiler``, we are not auto-detecting the compiler. 

What doesn't make sense is to always force users to define a ``[buildenv] CC=gcc CXX=g++`` when ``settings.compiler=gcc``. We don't do it in any other build system integration, and it shouldn't be done in Meson either. Assuming a ``gcc/g++``, ``clang/clang++`` is a good default, and the user always has the chance to override it defining the env-var. Otherwise, it is impossible to build a Meson project in Linux or OSX with the default detected profile, and requires all users to manually add CC/CXX env-vars to it.

So this is staying, it must be possible for every build system integration to be able to build a basic project with the default detected profile, for a default system installation of compilers and build systems.

conan/tools/meson/meson.py
memsharded: Suggestion for an alternative name?

conan/tools/meson/meson.py
memsharded: The problem with using here a hardcoded ``if cross_building()`` is that if the ``MesonToolchain`` decides (the user can do it) that it is not cross-building, this will be wrong. The build helpers like ``Meson`` should not decide things just execute what the toolchain has decided to do.

conan/tools/meson/toolchain.py
SSE4: the user might not have a `/usr/bin/gcc` or `/usr/bin/clang` at all.
for the `default` profile, it will pick `/usr/bin/gcc` for `settings.compiler=gcc`.
but it will also pick  `/usr/bin/gcc` for non default (e.g. `armv7`) profile (e.g. `settings.arch=armv7`).
that might not be something desired in this particular case, and in many others (e.g. Android NDK).
there is no check this choice is done for the `default` profile only, it always applies.
yes, it makes some use-cases more smooth, but other use-cases might be silently broken because of that default.
I'd limit these defaults to some known cases only, e.g. :
- apple-clang with XCode, where the compiler should be located via XCRun
- MSVC/Visual Studio, where VCVars is responsible for the compiler discovery
- Android NDK, where we have a clear mapping of settings to the compiler executables

for others, we don't know and can't be sure what is the sane choice. it might be `gcc.exe`, `gcc-10.exe`, `gcc-10.0.exe` or `arm-linux-gnueabi-gcc.exe`.

conan/tools/meson/meson.py
SSE4: could it be problematic in some flows, e.g. if I run several toolchain executions for multiple profiles?

conan/tools/meson/meson.py
SSE4: in [documentation](https://mesonbuild.com/Commands.html#configure), they refer to it as a `builddir`

conan/tools/meson/toolchain.py
memsharded: > but it will also pick /usr/bin/gcc for non default (e.g. armv7) profile (e.g. settings.arch=armv7).
that might not be something desired in this particular case, and in many others (e.g. Android NDK).

Of course the default will not work in cases of cross-compiling, and it will require specifying the specific compiler you want for cross-compile. This doesn't mean that for 75% of the users, doing native builds, we shouldn't define a default. Again, this is the same default as our CMake integration, by default it will pick the system "gcc/g++", and only for cross-compile scenarios you need to specify the CC/CXX env-vars.

SSE4: how about limiting the scope for native file only?

memsharded: Are you aware that if we don't add this, it will be also challenging to build packages using Meson in ConanCenter, where we are using the default profiles? If we change all default profiles in ConanCenter, which are different to the user default profiles, the risk of things not working also increases.

conan/tools/meson/toolchain.py
SSE4: another note (might be slightly unrelated), in CCI users, have started to adopt MSVC compiler already, and now recipes need to use a fairly complicated logic to manage compiler runtime, as settings are very different between old and new compilers. do you have a plan to add some sort of helper, with can return uniform runtime values (e.g. in formats like `MD` or `MultiThreadedDebug`)?

conan/tools/meson/toolchain.py
SSE4: right now, in CCI, the only problematic use case is M1. all other default profiles work.

memsharded: Because the recipes there are not using MesonToolchain yet. If MesonToolchain requires defining CC/CXX in all profiles, for all builds, things will be broken, it will be not possible to update without changing all the profiles in ConanCenter. Recall that it is necessary to move all recipes to the  new integrations to be ready for 2.0.

SSE4: we already tried MesonToolchain in CCI in the past: https://github.com/conan-io/conan-center-index/pull/7360/files https://github.com/conan-io/conan-center-index/pull/6678/files.
the thing is that it requires explicit CC/CXX for cross-building.
but it doesn't require it for native builds. e.g. it is able to deduce `cl` for Visual studio and `gcc` for GCC and `clang` for XCode builds.
profiles don't need to change.
the only one we currently have trouble is M1, but for this one, it needs a bunch of other flags (like `-arch` and `-isysroot`) from the XCrun.

conan/tools/meson/toolchain.py
```diff
@@ -3,187 +3,173 @@
...
+
+        if compiler == "Visual Studio":
+            vscrt = self._conanfile.settings.get_safe("compiler.base.runtime") or \
+                    self._conanfile.settings.get_safe("compiler.runtime")
+            self._b_vscrt = {"MD": "md",
+                            "MDd": "mdd",
+                            "MT": "mt",
+                            "MTd": "mtd"}.get(vscrt, "none")
+        elif compiler == "msvc":
+            # TODO: Fill here the msvc model
```
SSE4:
```suggestion
            # TODO: Fill here the msvc model
            self._b_vscrt = "m{}{}".format(
                    "t" if self.settings.compiler.runtime == "static" else "d",
                    "d" if self.settings.compiler.runtime_type == "Debug" else "",
                )
```

franramirez688 (approved):
Looking good! 👍 

@memsharded I think we should add to the PR description that it closes (perhaps not completely) the issue https://github.com/conan-io/conan/issues/9710, wdyt?

bb010g:
Was making `PkgConfig` (now at `conan.tools.gnu.pkgconfig.PkgConfig`) no longer be accessible via `conan.tools.gnu.PkgConfig` intentional? `PkgConfigDeps` (now at `conan.tools.gnu.pkgconfigdeps.pkgconfigdeps.PkgConfigDeps`) is still accessible via `conan.tools.gnu.PkgConfigDeps`.

PR #10178, opened 2021-12-14
Multiple CMake User Toolchain

### LINKED ISSUE #9724: [feature] How CMakeToolchain could handle multiple user toolchains
Lets say that we have 2 different build_requires, one ``sdk`` and one ``compiler``. Building a package could use both, and each one would want to inject their own ``user_toolchain.cmake``. 

Or they want to propagate information to the consumer in other ways. Challenges:

- Passing information like CC/CXX via environment requires using VirtualBuildEnv, and activating env scripts, which is not the most convenient for IDE developers. Env-vars might not be enough for some cases, and still a user toolchain might be necessary.
- Passing information via ``self.conf_info`` to consumer can work, but at the moment can only provide 1 value for ``user_toolchain``. More than one value would require list aggregation or equivalent functionality

cc/ @Blejdfist

lasote:
Close #9724

Changelog: Feature: The `CMakeToolchain` is now prepared to apply several user toolchains.
Docs: https://github.com/conan-io/docs/pull/2340


conan/tools/conf/conf_aggregator.py
```diff
@@ -0,0 +1,30 @@
...
+
+class ConfAggregator:
+    """Generator that accumulate the configs from the build-require dependencies"""
+    filename = "conf_aggregator.json"
+
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self.data = OrderedDict()
+        # Same behavior than _receive_conf, direct build requires only
+        for dep in self._conanfile.dependencies.direct_build.values():
```
memsharded: I think this is the key, so far there is only 1 conf that makes sense to accumulate in a list, out of 30 that doesn't make sense at all. So this aggregator will be doing the right thing for 1 conf, but the wrong one for 29, so not very clear to me the value. Maybe more than a general Conf aggregator would be a utility that allows to aggregate the specified conf value, just a single one?

lasote: Might be, but still unclear to me that this is the way given my previous comment on the PR.

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -115,3 +115,69 @@ def test_cmake_toolchain_without_build_type():
...
+                user_toolchains = []
+                for dep in self.dependencies.direct_build.values():
+                    ut = dep.conf_info["tools.cmake.cmaketoolchain:user_toolchain"]
+                    if ut:
+                        user_toolchains.append(ut)
+
+                # Modify the context of the user_toolchain block
+                t = CMakeToolchain(self)
+                if {}:
+                    t.blocks["user_toolchain"].values["paths"] = user_toolchains
```
memsharded: Maybe if the block has a explicit ``.user_toolchains`` attribute, the generic context one doesn't need to be documented (or even tested?)

memsharded (approved):
I think this looks better and less complicated than previous approaches.

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
lasote: My point is, you can always do that because of the blocks, right? even if it is not documented you can still watch it, see the key, and alter it.
So, Is it worth it to have the `.user_toolchains` or just the context? As I said in the PR description, I don't think we should keep both.

memsharded: Oh, sorry, I missed the updated description of the PR. 
Yes, I agree we should probably do only 1 way to do this. As the context one is very built-in and cannot be removed, maybe we just want to remove the ``.user_toolchains`` attribute. The truth is that those attributes were added before we had the ``block.values`` helper, that greatly simplify to modify contexts.

PR #10186, opened 2021-12-16
CMakeToolchain: proper support of find_*() & include() commands

### LINKED ISSUE #9427: [question] CMakeToolchain: Support for find_library(), find_path(), find_file(), find_program()

Do you plan to add proper packages paths in `CMakeToolchain` to support:
- `find_library()` of libraries in host context (add `lib_paths` of host context packages in `CMAKE_LIBRARY_PATH`?)
- `find_path()` and `find_file()` of headers in host context (add `include_paths` of host context packages in `CMAKE_INCLUDE_PATH`?)
- `find_program()` of programs in build context (add `bin_paths` of build context packages in `CMAKE_PROGRAM_PATH`?)

Without this support, `CMakeToolchain` is unusable when:
- you package a CMake project which must rely on `find_library()` or `find_path()` to discover dependencies without official CMake Find/config files or pkgconfig files.
- you want to use build requirements in a transparent CMake integration (`virtualrunenv` generator is a workaround, but I find it very hard and unconvenient to use in some IDEs).

### LINKED ISSUE #9735: [question] Unable to find_package with cmake when cross compiling
I've been creating [this](https://notabug.org/williambehrens/AllWpilibConan) for conan, the package is meant to be used next to a cross compiler for the NI roborio, I have the cross compiler installed on my system and I created a conan profile for it:

### roborio.ini
```
toolchain=/usr/arm-frc2021-linux-gnueabi
target_host=arm-frc2021-linux-gnueabi
cc_compiler=gcc
cxx_compiler=g++

[env]
CONAN_CMAKE_FIND_ROOT_PATH=$toolchain  # Optional, for CMake to find things in that folder
CONAN_CMAKE_SYSROOT=$toolchain  # Optional, if we want to define sysroot
CHOST=$target_host
LD=$target_host-ld
AR=$target_host-ar
AS=$target_host-as
RANLIB=$target_host-ranlib
CC=$target_host-$cc_compiler
CXX=$target_host-$cxx_compiler
STRIP=$target_host-strip
RC=$target_host-windres

[settings]
os=Linux
arch=armv7
compiler=gcc
compiler.version=7.3
compiler.libcxx=libstdc++
build_type=Release
 
[options]
# FIXME: Find a way to enable
allwpilib:with_gui = False # Disabled due to missing Xorg package
```
and a cmake toolchain file that looks like this:
### roborio.cmake
```
set(ARM_PREFIX arm-frc2021-linux-gnueabi)

set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR arm)

# In the docker build this folder doesn't exist?
set(CMAKE_SYSROOT /usr/${ARM_PREFIX})

# https://stackoverflow.com/questions/53633705/cmake-the-c-compiler-is-not-able-to-compile-a-simple-test-program
set(CMAKE_C_COMPILER_WORKS 1)
set(CMAKE_CXX_COMPILER_WORKS 1)

set(CMAKE_C_COMPILER /usr/local/bin/${ARM_PREFIX}-gcc)
set(CMAKE_CXX_COMPILER /usr/local/bin/${ARM_PREFIX}-g++)
set(CMAKE_LINKER /usr/local/bin/${ARM_PREFIX}-ld)

set(CMAKE_CXX_FLAGS "-fuse-ld=${CMAKE_LINKER} -m32")

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)

set(CMAKE_MODULE_PATH ${CMAKE_SYSROOT}/usr/share/cmake/Modules)
```
the project itself is stored [here](https://github.com/rr1706/Control2021cpp) in case I missed something important, but after compiling I get a cmake error saying it can't find Findallwpilib.cmake even through it has no trouble finding the files when I'm not cross compiling

### LINKED ISSUE #10167: [bug] CMakeToolchain does not include cpp_info.builddirs from build requirements in CMAKE_PREFIX_PATH CMAKE_MODULE_PATH
When using CMakeToolchain, `cpp_info.builddirs` from all requirements are being added correctly to `CMAKE_PREFIX_PATH` and to `CMAKE_PREFIX_PATH`, However cpp_info.builddirs from **build_requirements** are ignored.

This is problematic, as you could have a conan package that only contains cmake modules, like [extra-cmake-modules](https://conan.io/center/extra-cmake-modules), this package is a conan `build_requirements` and it's `cpp_info.builddirs` should be added `CMAKE_PREFIX_PATH` and `CMAKE_PREFIX_PATH` as well!

  * Operating System+version: windows 10
  * Compiler+version: msvc
  * Conan version: 1.41.0
  * Python version: 3.7

SpaceIm:
Changelog: Fix: Improved CMakeToolchain robustness regarding ``find_file``, ``find_path`` and ``find_program`` commands allowing better cross-build scenarios and better differentiation of the right context where to get, for example,  executables (build vs host).
Docs: https://github.com/conan-io/docs/pull/2383

- In `find_program()`, avoid to find executables of host context before executables of build context.
- support all cross-build scenario (hopefully).
- support `include()` of CMake modules coming from build context.
- remove `CMakeToolchain.find_builddirs` (doesn't make sense anymore).

closes https://github.com/conan-io/conan/issues/9427
closes https://github.com/conan-io/conan/issues/9735
closes https://github.com/conan-io/conan/issues/10167

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -399,23 +399,45 @@ class FindFiles(Block):
         {% if find_package_prefer_config %}
         set(CMAKE_FIND_PACKAGE_PREFER_CONFIG {{ find_package_prefer_config }})
         {% endif %}
-        {% if generators_folder or host_build_paths_root or host_build_paths_noroot or build_build_paths %}
-        set(CMAKE_MODULE_PATH {{ generators_folder }} {{ host_build_paths_root }} {{ host_build_paths_noroot }} {{ build_build_paths }} ${CMAKE_MODULE_PATH})
+
+        # Definition of CMAKE_MODULE_PATH
+        {% if build_build_paths %}
+        # Explicitly defined "buildirs" of "build" context dependencies
+        list(PREPEND CMAKE_MODULE_PATH {{ build_build_paths }})
         {% endif %}
-        {% if generators_folder or host_build_paths_noroot %}
-        set(CMAKE_PREFIX_PATH {{ generators_folder }} {{ host_build_paths_noroot }} ${CMAKE_PREFIX_PATH})
+        {% if host_build_paths_noroot %}
+        # Explicitly defined "builddirs" of "host" dependencies
+        list(PREPEND CMAKE_MODULE_PATH {{ host_build_paths_noroot }})
+        {% endif %}
+        {% if host_build_paths_root %}
+        # The root (which is the default builddirs) path of dependencies in the host context
+        list(PREPEND CMAKE_MODULE_PATH {{ host_build_paths_root }})
+        {% endif %}
+        {% if generators_folder %}
+        # the generators folder (where conan generates files, like this toolchain)
+        list(PREPEND CMAKE_MODULE_PATH {{ generators_folder }})
+        {% endif %}
+
+        # Definition of CMAKE_PREFIX_PATH, CMAKE_XXXXX_PATH
+        {% if host_build_paths_noroot %}
+        # The explicitly defined "builddirs" of "host" context dependencies must be in PREFIX_PATH
+        list(PREPEND CMAKE_PREFIX_PATH {{ host_build_paths_noroot }})
+        {% endif %}
+        {% if generators_folder %}
+        # The Conan local "generators" folder, where this toolchain is saved.
+        list(PREPEND CMAKE_PREFIX_PATH {{ generators_folder }} )
         {% endif %}
         {% if cmake_program_path %}
-        set(CMAKE_PROGRAM_PATH {{ cmake_program_path }} ${CMAKE_PROGRAM_PATH})
+        list(PREPEND CMAKE_PROGRAM_PATH {{ cmake_program_path }})
         {% endif %}
         {% if cmake_library_path %}
-        set(CMAKE_LIBRARY_PATH {{ cmake_library_path }} ${CMAKE_LIBRARY_PATH})
+        list(PREPEND CMAKE_LIBRARY_PATH {{ cmake_library_path }})
         {% endif %}
         {% if is_apple and cmake_framework_path %}
-        set(CMAKE_FRAMEWORK_PATH {{ cmake_framework_path }} ${CMAKE_FRAMEWORK_PATH})
+        list(PREPEND CMAKE_FRAMEWORK_PATH {{ cmake_framework_path }})
         {% endif %}
         {% if cmake_include_path %}
-        set(CMAKE_INCLUDE_PATH {{ cmake_include_path }} ${CMAKE_INCLUDE_PATH})
+        list(PREPEND CMAKE_INCLUDE_PATH {{ cmake_include_path }})
         {% endif %}
 
         {% if cross_building %}
@@ -465,12 +487,16 @@ class FindFiles(Block):
         host_include_paths = []
         for req in host_req:
             cppinfo = req.cpp_info.aggregated_components()
-            host_build_paths_root.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p == ""])
-            host_build_paths_noroot.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            host_build_paths_root.extend([os.path.join(req.package_folder, p)
+                                          for p in cppinfo.builddirs if p == ""])
+            host_build_paths_noroot.extend([os.path.join(req.package_folder, p)
+                                            for p in cppinfo.builddirs if p != ""])
             host_lib_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.libdirs])
             if is_apple_:
-                host_framework_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.frameworkdirs])
-            host_include_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.includedirs])
+                host_framework_paths.extend([os.path.join(req.package_folder, p)
+                                             for p in cppinfo.frameworkdirs])
+            host_include_paths.extend([os.path.join(req.package_folder, p)
+                                       for p in cppinfo.includedirs])
 
         # Read information from build context
         build_req = self._conanfile.dependencies.build.values()
@@ -478,7 +504,8 @@ class FindFiles(Block):
         build_bin_paths = []
         for req in build_req:
             cppinfo = req.cpp_info.aggregated_components()
-            build_build_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs])
+            build_build_paths.extend([os.path.join(req.package_folder, p)
+                                      for p in cppinfo.builddirs])
             build_bin_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
 
         return {
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -8,6 +8,7 @@ from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
 from conan.tools._compilers import architecture_flag, use_win_mingw
+from conan.tools.apple.apple import is_apple_os
 from conan.tools.build import build_jobs
 from conan.tools.cmake.utils import is_multi_configuration
 from conan.tools.cross_building import cross_building
@@ -476,7 +477,7 @@ class FindFiles(Block):
             find_package_prefer_config = "OFF"
 
         os_ = self._conanfile.settings.get_safe("os")
-        is_apple_ = os_ and os_ in ["Macos", "iOS", "watchOS", "tvOS"]
+        is_apple_ = is_apple_os(os_)
 
         # Read information from host context
         host_req = self._conanfile.dependencies.host.values()
@@ -487,16 +488,14 @@ class FindFiles(Block):
         host_include_paths = []
         for req in host_req:
             cppinfo = req.cpp_info.aggregated_components()
-            host_build_paths_root.extend([os.path.join(req.package_folder, p)
-                                          for p in cppinfo.builddirs if p == ""])
-            host_build_paths_noroot.extend([os.path.join(req.package_folder, p)
-                                            for p in cppinfo.builddirs if p != ""])
-            host_lib_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.libdirs])
+            # If the builddir is the package_folder, then it is the default "root" one
+            nf = os.path.normpath(req.package_folder)
+            host_build_paths_root.extend(p for p in cppinfo.builddirs if os.path.normpath(p) == nf)
+            host_build_paths_noroot.extend(p for p in cppinfo.builddirs if os.path.normpath(p) != nf)
+            host_lib_paths.extend(cppinfo.libdirs)
             if is_apple_:
-                host_framework_paths.extend([os.path.join(req.package_folder, p)
-                                             for p in cppinfo.frameworkdirs])
-            host_include_paths.extend([os.path.join(req.package_folder, p)
-                                       for p in cppinfo.includedirs])
+                host_framework_paths.extend(cppinfo.frameworkdirs)
+            host_include_paths.extend(cppinfo.includedirs)
 
         # Read information from build context
         build_req = self._conanfile.dependencies.build.values()
@@ -504,9 +503,8 @@ class FindFiles(Block):
         build_bin_paths = []
         for req in build_req:
             cppinfo = req.cpp_info.aggregated_components()
-            build_build_paths.extend([os.path.join(req.package_folder, p)
-                                      for p in cppinfo.builddirs])
-            build_bin_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
+            build_build_paths.extend(cppinfo.builddirs)
+            build_bin_paths.extend(cppinfo.bindirs)
 
         return {
             "find_package_prefer_config": find_package_prefer_config,
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -428,18 +428,10 @@ class FindFiles(Block):
         {% endif %}
 
         # To support cross building
-        if(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM STREQUAL "ONLY")
-            set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM "BOTH")
-        endif()
-        if(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY STREQUAL "ONLY")
-            set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY "BOTH")
-        endif()
-        if(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE STREQUAL "ONLY")
-            set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
-        endif()
-        if(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE STREQUAL "ONLY")
-            set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE "BOTH")
-        endif()
+        set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM "BOTH")
+        set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY "BOTH")
+        set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
+        set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE "BOTH")
         """)
 
     def context(self):
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -447,42 +447,42 @@ class FindFiles(Block):
 
         # Read information from host context
         host_req = self._conanfile.dependencies.host.values()
-        module_host_paths = []
-        prefix_paths = []
-        bin_host_paths = []
-        lib_paths = []
-        framework_paths = []
-        include_paths = []
+        host_module_paths = []
+        host_prefix_paths = []
+        host_bin_paths = []
+        host_lib_paths = []
+        host_framework_paths = []
+        host_include_paths = []
         for req in host_req:
             cppinfo = req.cpp_info.copy()
             cppinfo.aggregate_components()
-            module_host_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
-            prefix_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            host_module_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            host_prefix_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
             if not cross_building(self._conanfile):
-                bin_host_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
-            lib_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.libdirs])
-            framework_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.frameworkdirs])
-            include_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.includedirs])
+                host_bin_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
+            host_lib_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.libdirs])
+            host_framework_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.frameworkdirs])
+            host_include_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.includedirs])
 
         # Read information from build context
         build_req = self._conanfile.dependencies.build.values()
-        module_build_paths = []
-        bin_build_paths = []
+        build_module_paths = []
+        build_bin_paths = []
         for req in build_req:
             cppinfo = req.cpp_info.copy()
             cppinfo.aggregate_components()
-            module_build_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
-            bin_build_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
+            build_module_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            build_bin_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
 
         return {
             "find_package_prefer_config": find_package_prefer_config,
             "generators_folder": "${CMAKE_CURRENT_LIST_DIR}",
-            "cmake_module_path": self._join_paths(module_host_paths + module_build_paths),
-            "cmake_prefix_path": self._join_paths(prefix_paths),
-            "cmake_program_path": self._join_paths(bin_build_paths + bin_host_paths),
-            "cmake_library_path": self._join_paths(lib_paths),
-            "cmake_framework_path": self._join_paths(framework_paths),
-            "cmake_include_path": self._join_paths(include_paths),
+            "cmake_module_path": self._join_paths(host_module_paths + build_module_paths),
+            "cmake_prefix_path": self._join_paths(host_prefix_paths),
+            "cmake_program_path": self._join_paths(build_bin_paths + host_bin_paths),
+            "cmake_library_path": self._join_paths(host_lib_paths),
+            "cmake_framework_path": self._join_paths(host_framework_paths),
+            "cmake_include_path": self._join_paths(host_include_paths),
         }
 
 
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -394,27 +394,52 @@ class AppleSystemBlock(Block):
         return ctxt_toolchain
 
 
-class FindConfigFiles(Block):
+class FindFiles(Block):
     template = textwrap.dedent("""
         {% if find_package_prefer_config %}
         set(CMAKE_FIND_PACKAGE_PREFER_CONFIG {{ find_package_prefer_config }})
         {% endif %}
-        # To support the generators based on find_package()
-        {% if cmake_module_path %}
-        set(CMAKE_MODULE_PATH {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
+        {% if generators_folder or cmake_module_path %}
+        # To support find_package() of CMake Find files (host context), and include() of CMake modules (host & build context)
+        set(CMAKE_MODULE_PATH {{ generators_folder }} {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
         {% endif %}
-        {% if cmake_prefix_path %}
-        set(CMAKE_PREFIX_PATH {{ cmake_prefix_path }} ${CMAKE_PREFIX_PATH})
+        {% if generators_folder or cmake_prefix_path %}
+        # To support find_package() of CMake config Files from host context
+        set(CMAKE_PREFIX_PATH {{ generators_folder }} {{ cmake_prefix_path }} ${CMAKE_PREFIX_PATH})
+        {% endif %}
+        {% if cmake_program_path %}
+        # To support find_program() of executables from build context
+        set(CMAKE_PROGRAM_PATH {{ cmake_program_path }} ${CMAKE_PROGRAM_PATH})
+        {% endif %}
+        {% if cmake_library_path %}
+        # To support find_library() of libraries from host context
+        set(CMAKE_LIBRARY_PATH {{ cmake_library_path }} ${CMAKE_LIBRARY_PATH})
+        {% endif %}
+        {% if cmake_framework_path %}
+        # To support find_framework() of frameworks from host context
+        set(CMAKE_FRAMEWORK_PATH {{ cmake_framework_path }} ${CMAKE_FRAMEWORK_PATH})
+        {% endif %}
+        {% if cmake_include_path %}
+        # To support find_file() and file_path() of headers from host context
+        set(CMAKE_INCLUDE_PATH {{ cmake_include_path }} ${CMAKE_INCLUDE_PATH})
         {% endif %}
         {% if android_prefix_path %}
         set(CMAKE_FIND_ROOT_PATH {{ android_prefix_path }} ${CMAKE_FIND_ROOT_PATH})
         {% endif %}
 
-        # To support cross building to iOS, watchOS and tvOS where CMake looks for config files
-        # only in the system frameworks unless you declare the XXX_DIR variables
-        {% if cross_ios %}
-            set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE BOTH)
-        {% endif %}
+        # To support cross building
+        if(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM "BOTH")
+        endif()
+        if(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY "BOTH")
+        endif()
+        if(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
+        endif()
+        if(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE "BOTH")
+        endif()
         """)
 
     def context(self):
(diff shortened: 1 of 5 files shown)


SpaceIm:
Two comments regarding existing code I didn't change:
- I don't understand what is the purpose of android part, it seems to specific and dangerous since it manipulates `CMAKE_FIND_ROOT_PATH` which is the top precedence in all paths checked by `find_*()` commands
  => ok, I've removed this part, it's the usual issue due to `CMAKE_FIND_ROOT_PATH_MODE_*` set to `ONLY`.
- `CMAKE_FIND_PACKAGE_PREFER_CONFIG` can be removed (another PR?) now that CMakeDeps can generate Find files (`cmake_find_mode` property).

SpaceIm:
tests fail because:
- they want to be able to find some config files in the root package folder, which is something I don't want `find_package()` can do, because adding root package folder to `CMAKE_PREFIX_PATH` has too many side effects.
  Actually, there should be 2 more properties in `cpp_info` of recipes: `cmake_module_path` (folders of packaged CMake module and Find files) & `cmake_prefix_path` (folders of packaged pkgconfig & CMake Config files), empty by default, and `CMAKE_MODULE_PATH` & `CMAKE_PREFIX_PATH` should be populated with these paths respectively (+ `CMAKE_CURRENT_LIST_DIR`). Indeed `builddirs` is too generic and semantically unrelated to `CMAKE_MODULE_PATH` & `CMAKE_PREFIX_PATH`.
  Anyway, here I expect folders of packaged CMake module and config files to be added explicitly in `builddirs`.
- I have removed `find_builddir` logic, but not related test. I think this switch won't make sense after this PR.

Maybe I should have added the tests I want to succeed in the first place (same recipe in requirements & build requirements with libs & tools packaged, discovery of module files from build requirements).

conan/tools/cmake/toolchain.py
```diff
@@ -425,34 +438,61 @@ def context(self):
...
+            module_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            prefix_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            lib_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.libdirs])
+            framework_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.frameworkdirs])
+            include_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.includedirs])
+
+        # Read information from build context
+        build_req = self._conanfile.dependencies.build.values()
+        bin_paths = []
+        for req in build_req:
```
memsharded, line 504: I am not sure we should be adding by default the paths to the "include", "libs" of ``build_requires``. Consider those are things like ``cmake`` with a transitive dependency of ``openssl`` and ``zlib``. If you do a find_library(openssl) in your code, almost never you will want to find the cmake->openssl build context library.

This is the reason why we don't create ``xxx-config.cmake`` by default for ``build_requires``. Recipes need to opt-in into that, because it is so easy that the ``zlib-config.cmake`` of build-context will conflict and potentially overwrite the one from the host context. I think exactly the same reasoning apply for findind libraries, includes, frameworks (and mostly everything except executables)

memsharded (commented):
I have my concerns about the extension of this to "dependencies.build". Things liks libs, headers and frameworks probably shouldn't be added by default. And the executables and shared libs don't need to be added to the path, because they will already be in the PATH and LD_LIBRARY_PATH, because they are put there by the ``VirtualBuildEnv`` (which is a bit more general solution, as it works for every build system)

conan/tools/cmake/toolchain.py
SpaceIm: > I have my concerns about the extension of this to "dependencies.build". Things liks libs, headers and frameworks probably shouldn't be added by default

It doesn't add `include` and `lib` paths of `build requires`, only `bin` and folders of CMake modules files (builddirs without root package folder).

Folders of CMake module files from build context must be added to `CMAKE_MODULE_PATH` to fix https://github.com/conan-io/conan/issues/10167.

But if someone uses a recipe which package Find or Config files in `builddirs` (not CCI recipes), yes the ones from `build requires` will be discoverable with `find_package()`, but they won't conflict with host context since they are appended after host context  `builddirs` in `CMAKE_MODULE_PATH` and `CMAKE_PREFIX_PATH`.

SpaceIm: > And the executables and shared libs don't need to be added to the path, because they will already be in the PATH and LD_LIBRARY_PATH, because they are put there by the VirtualBuildEnv (which is a bit more general solution, as it works for every build system)

It is not sufficient for `find_*()` commands:
- `PATH` has very low precedence and might be descoped by `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM ONLY`
- `LD_LIBRARY_PATH` doesn't matter for `find_*()` commands, it's a runtime variable. `VirtualBuildEnv` is required if a tool from build context is executed (at configure or build time), but it shouldn't matter for discovery of tools/libs/files during CMake configuration.

Actually this PR doesn't manipulate `PATH` or `LD_LIBRARY_PATH` at all (I've suggested in https://github.com/conan-io/conan/issues/9427#issuecomment-993685376 to add all `build_env` , but it's not in this PR because I think it can only solves runtime execution of tools at configure time, not build time. So yes `VirtualBuildEnv` is a more general solution to runtime concerns, but clearly not a silver bullet for discoverability).

memsharded: > It doesn't add include and lib paths of build requires, only bin and folders of CMake modules files (builddirs without root package folder).

Oh, yes you are right. I got a bit confused by the indents here in Github reviews with the truncated lines I lost scope.

conan/tools/cmake/toolchain.py
```diff
@@ -425,34 +439,64 @@ def context(self):
         if prefer_config is not None and prefer_config.lower() in ("false", "0", "off"):
             find_package_prefer_config = "OFF"
 
-        os_ = self._conanfile.settings.get_safe("os")
-        android_prefix = "${CMAKE_CURRENT_LIST_DIR}" if os_ == "Android" else None
```
memsharded, line 429: This was necessary at some point, to have the Android build finding the xxx-config.cmake by Conan. Maybe it is no longer needed, but it is difficult to know, because we are not running yet Android build tests in our CI

SpaceIm: Yes, it's the same issue than cross-build to iOS/tvOS/watchOS, but 2 different solutions have been implemented. Here it's fixed by forcing `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` to `"BOTH"` (same fix than iOS/tvOS/watchOS).

Of course, the best solution would be to not have to manipulate `CMAKE_FIND_ROOT_PATH_MODE_*` but other solutions raise other issues (conflicts between host & build context), and I think this one is harmless compare to other solutions.

SpaceIm:
TODO:
- [x] add test to check that executables of build context are found before executables of host context in `find_program()` (native & cross build, recipe in build requirement only & recipe in requirements and build requirements).
- [x] add test to check that CMake modules of build context are accessible with `include()` (native & cross build).

SpaceIm:
FYI: locally, all the tests (I mean all, not only the tests of this PR) fail on Windows/Visual Studio with `'settings.compiler.cppstd' value not defined` when `compiler` is part of settings of a recipe, why does it succeed in CI?

SpaceIm:
~I think I've found a solution allowing to not force any of `CMAKE_FIND_ROOT_PATH_MODE_*`, based on `CMAKE_IGNORE_PATH`.~

SpaceIm:
Outch, `CMAKE_IGNORE_PATH` is dangerous for editable packages, since everything may be more or less in the same folder.

conan/tools/cmake/toolchain.py
```diff
@@ -425,33 +445,43 @@ def context(self):
...
-        return {"find_package_prefer_config": find_package_prefer_config,
-                "cmake_prefix_path": "${CMAKE_CURRENT_LIST_DIR} " + build_paths,
-                "cmake_module_path": "${CMAKE_CURRENT_LIST_DIR} " + build_paths,
-                "android_prefix_path": android_prefix,
-                "cross_ios": cross_ios,
-                "generators_folder": "${CMAKE_CURRENT_LIST_DIR}"}
+            host_module_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs])
+            host_prefix_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p != ""])
+            if not cross_building(self._conanfile):
+                host_bin_paths.extend([os.path.join(req.package_folder, p) for p in cppinfo.bindirs])
```
SpaceIm: Maybe it should be avoided, even for native builds, to enforce recipes and consumers to properly list their build requirements.
Previously, executables of host context were always discoverable, and even before executables of build context (because root package folders of requires were added to `CMAKE_PREFIX_PATH` which has precedence over `PATH`).

conan/tools/cmake/toolchain.py
```diff
@@ -394,29 +395,48 @@ def context(self):
...
+        {% if generators_folder or cmake_prefix_path %}
+        # To support find_package() of CMake config Files from host context
+        set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE "BOTH")
+        set(CMAKE_PREFIX_PATH {{ generators_folder }} {{ cmake_prefix_path }} ${CMAKE_PREFIX_PATH})
         {% endif %}
-        {% if android_prefix_path %}
-        set(CMAKE_FIND_ROOT_PATH {{ android_prefix_path }} ${CMAKE_FIND_ROOT_PATH})
+        {% if cmake_program_path %}
+        # To support find_program() of executables from build context (and host context if not cross-building)
+        set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM "BOTH")
```
lasote: These `BOTH` values worry me a bit. That means it always will go search the system for programs, config files, libraries, and so on, right? I think this should probably be opt-in (when cross-building?) because the general rule (except cross-building, sysroot, etc scenarios) is to look only at the "conan" folders. Am I missing something?

SpaceIm: I can't disagree, I would prefer to not manipulate `CMAKE_FIND_ROOT_PATH_MODE_*`.

Actually, if any of these variables is set to `ONLY`, the associated find command won't look into conan folders, because in this case only paths in `CMAKE_FIND_ROOT_PATH` and `CMAKE_SYSROOT` would be considered (paths in `CMAKE_PREFIX_PATH`, `CMAKE_<LIBRARY/INCLUDE/FRAMEWORK/PROGRAM>_PATH`, `PATH`, `LIB` etc are ignored)

Unfortunately, the current conan paradigm (specifically layout of installed packages, and the fact that tools & libs are packaged together) completely prevents to use `CMAKE_FIND_ROOT_PATH` in a robust way.
See also this comment on Slack: https://cpplang.slack.com/archives/C5Y2GDECX/p1639932998167300

It's worth noting that `CMAKE_FIND_ROOT_PATH_MODE_*` set to `BOTH` doesn't hurt as soon as rerooting is properly done, since system paths have the lowest precedence (see https://github.com/conan-io/conan/issues/9427#issuecomment-995780534). So it can be misleading when someone has forgotten some dependencies, but otherwise it's fine.

SpaceIm: Here what could be done for `ONLY`, and the side effects:

if `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` `ONLY`:
- add `CMAKE_CURRENT_LIST_DIR` & `builddirs` of requires (except root package folders) to `CMAKE_FIND_ROOT_PATH`.

if `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY` or `CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK` or `CMAKE_FIND_ROOT_PATH_MODE_INCLUDE` `ONLY`:
- add `builddirs` (root package folders specifically) of requires to `CMAKE_FIND_ROOT_PATH`
- side effects:
  - make executables living in requires discoverable by `find_program()`, and with the top priority, unless `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is `NEVER`.
  - libs, headers & frameworks of packages with non-standard install layout may not be discoverable with `find_file()`/`file_path()`/`find_library()`, even if `cpp_info.libdirs`, `cpp_info.includedirs`, `cpp_info.frameworkdirs` are properly set.

if `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` `ONLY`:
- add `builddirs` (root package folders specifically)  of build requires to `CMAKE_FIND_ROOT_PATH`
- side effets:
  - make headers, libs and frameworks in build requires discoverable by `find_file()`, `find_path()` and `find_library()`, unless `CMAKE_FIND_ROOT_PATH_MODE_<INCLUDE/LIBRARY/FRAMEWORK>` are `NEVER`.
  - executables of packages with non-standard install layout may not be discoverable with `find_program()`, even if `cpp_info.bindirs` is properly set.

SpaceIm: So, proposal to improve this PR:

- keep current logic but remove unconditional set of `CMAKE_FIND_ROOT_PATH_MODE_*` variables.
- moreover, if cross-building:
  - if `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` is `ONLY`:
    - add `CMAKE_CURRENT_LIST_DIR` & `builddirs` of requires (except root package folders) to `CMAKE_FIND_ROOT_PATH`.
  - if concatanation of `bindirs` of build_requires is empty or `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is `NEVER`:
    - if `CMAKE_FIND_ROOT_PATH_MODE_INCLUDE` is `ONLY` or `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY` is `ONLY` or `CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK` is `ONLY`:
      - add `builddirs` of requires to `CMAKE_FIND_ROOT_PATH` (actually root package folder is important here).
  - else:
    - if any of `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM`, `CMAKE_FIND_ROOT_PATH_MODE_INCLUDE`, `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`, `CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK` is `ONLY`, force the variable to `BOTH` instead of `ONLY`.
  - if anything has been added to `CMAKE_FIND_ROOT_PATH`, `CMAKE_CURRENT_LIST_DIR` should be the most left value. Maybe we want also to append in `CMAKE_FIND_ROOT_PATH` (not prepend), to let consumers put their own SDK in `CMAKE_FIND_ROOT_PATH` at the highest precedence.

It should avoid to manipulate `CMAKE_FIND_ROOT_PATH_MODE_*` variables. Indeed, the only modification remaining happens when `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is not `NEVER` when cross-building, which is quite unusual in cross-build toolchains. And no more modification of `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE`.

SpaceIm:
@memsharded the test `test_apple_own_framework_cross_build()` doesn't use 2 profiles to cross build:
https://github.com/conan-io/conan/blob/05e8fb2c77fbcf193750e7209c7ffdfac1a9c5de/conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py#L191-L238

So `conan.tools.cross_building` fails to detect cross-build (maybe a bug), and therefore it breaks the last logic I've implemented.

Adding `pr:b default` in those 2 lines of the test fixes the issue: https://github.com/conan-io/conan/blob/05e8fb2c77fbcf193750e7209c7ffdfac1a9c5de/conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py#L194-L195

The question is: Is `CMakeToolchain` supposed to be robust to cross-build with 1 profile?

conan/tools/cmake/toolchain.py
```diff
@@ -394,28 +395,84 @@ def context(self):
...
         {% endif %}
-        # To support the generators based on find_package()
-        {% if cmake_module_path %}
-        set(CMAKE_MODULE_PATH {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
+        {% if generators_folder or host_build_paths_root or host_build_paths_noroot or build_build_paths %}
+        set(CMAKE_MODULE_PATH {{ generators_folder }} {{ host_build_paths_root }} {{ host_build_paths_noroot }} {{ build_build_paths }} ${CMAKE_MODULE_PATH})
         {% endif %}
-        {% if cmake_prefix_path %}
-        set(CMAKE_PREFIX_PATH {{ cmake_prefix_path }} ${CMAKE_PREFIX_PATH})
+        {% if generators_folder or host_build_paths_noroot %}
```
lasote: Why `host_build_paths_noroot` should be set to `CMAKE_PREFIX_PATH`? just in case a package contains `config` files?  In that case, what if a user wanted to package a `config` file located in the root?

conan/tools/cmake/toolchain.py
```diff
@@ -394,28 +395,84 @@ def context(self):
...
+class FindFiles(Block):
     template = textwrap.dedent("""
         {% if find_package_prefer_config %}
         set(CMAKE_FIND_PACKAGE_PREFER_CONFIG {{ find_package_prefer_config }})
         {% endif %}
-        # To support the generators based on find_package()
-        {% if cmake_module_path %}
-        set(CMAKE_MODULE_PATH {{ cmake_module_path }} ${CMAKE_MODULE_PATH})
+        {% if generators_folder or host_build_paths_root or host_build_paths_noroot or build_build_paths %}
+        set(CMAKE_MODULE_PATH {{ generators_folder }} {{ host_build_paths_root }} {{ host_build_paths_noroot }} {{ build_build_paths }} ${CMAKE_MODULE_PATH})
```
lasote: Priority between HOST context and BUILD context modules? Should be reversed? Probably there is no good answer for this and we should explicitly declare somehow before calling "generate"?

conan/tools/cmake/toolchain.py
```diff
@@ -394,28 +395,84 @@ def context(self):
...
-        {% if android_prefix_path %}
-        set(CMAKE_FIND_ROOT_PATH {{ android_prefix_path }} ${CMAKE_FIND_ROOT_PATH})
+        {% if cmake_program_path %}
+        set(CMAKE_PROGRAM_PATH {{ cmake_program_path }} ${CMAKE_PROGRAM_PATH})
+        {% endif %}
+        {% if cmake_library_path %}
+        set(CMAKE_LIBRARY_PATH {{ cmake_library_path }} ${CMAKE_LIBRARY_PATH})
+        {% endif %}
+        {% if is_apple and cmake_framework_path %}
+        set(CMAKE_FRAMEWORK_PATH {{ cmake_framework_path }} ${CMAKE_FRAMEWORK_PATH})
```
lasote: I read about the `CMAKE_APPBUNDLE_PATH`. Is that something we want to discuss? has it any sense?

conan/tools/cmake/toolchain.py
```diff
@@ -394,28 +395,84 @@ def context(self):
...
+            set(CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK "BOTH")
+        endif()
         {% endif %}
-        """)
+        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_INCLUDE)
+            set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
+        endif()
+
+        if(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH ${CMAKE_FIND_ROOT_PATH} {{ generators_folder }} {{ host_build_paths_noroot }})
```
lasote: is this trick necessary? why not set `BOTH`? I'm sure there is an explanation.

conan/tools/cmake/toolchain.py
```diff
@@ -394,28 +395,84 @@ def context(self):
...
+        endif()
+        if(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM STREQUAL "NEVER")
```
lasote: Same question for all "ifs". Sounds very complex :(

conan/tools/cmake/toolchain.py
```diff
@@ -426,32 +483,46 @@ def context(self):
...
-        else:
-            build_paths = ""
-
-        return {"find_package_prefer_config": find_package_prefer_config,
-                "cmake_prefix_path": "${CMAKE_CURRENT_LIST_DIR} " + build_paths,
-                "cmake_module_path": "${CMAKE_CURRENT_LIST_DIR} " + build_paths,
-                "android_prefix_path": android_prefix,
-                "cross_ios": cross_ios,
-                "generators_folder": "${CMAKE_CURRENT_LIST_DIR}"}
+            host_build_paths_root.extend([os.path.join(req.package_folder, p) for p in cppinfo.builddirs if p == ""])
```
lasote: I think I agree we should stop adding "." as a default builddir and builddir should have effect only in `CMAKE_MODULE_PATH`, not in `CMAKE_PREFIX_PATH`. And I get why you do this, mostly because we cannot break now the model removing the "." default. But even though all this sounds complicated to understand and maintain. I'm wondering if we might postpone these fixes to 2.0 and introduce some breaking changes to the model to improve it.

lasote (commented):
I understand these issues (and agree with them) and I see a lot of value in this PR to understand the issues we currently have. But still not convinced about the way to resolve them. Should we use more properties to declare specific things like `config file paths`? and maybe more things?

Should we do this in Conan 1.X even if maybe the better fix (not either sure) is to change the cppinfo model or defaults? 

Also, I think this PR does some "tricks" like changing the ROOT_PATH_MODE variables and filling `CMAKE_FIND_ROOT_PATH`, probably there is not single answer for all cases, I'm afraid it might depend on every project/package/toolchain, in some cases, I would need to find some things in my packages, sometimes I don't, sometimes from build, sometimes from host...All these questions should be answered by the user, the challenge is how to do it understandable and maintainable.

conan/tools/cmake/toolchain.py
SpaceIm: I guess `CMAKE_APPBUNDLE_PATH` could be populated by `bindirs` from build context for `find_program()`.

conan/tools/cmake/toolchain.py
SpaceIm: Sure, the toolchain could force `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` to BOTH. But I would like to avoid to override this variable if set by user toolchain.

conan/tools/cmake/toolchain.py
SpaceIm: > Why host_build_paths_noroot should be set to CMAKE_PREFIX_PATH? just in case a package contains config files?

Yes

>  In that case, what if a user wanted to package a config file located in the root?

It wouldn't be discoverable, and this toolchain can't allow to add root package folder of requires to `CMAKE_PREFIX_PATH`, otherwise executables in `<root>/bin` of requires are discoverable before executables of build_requires (`CMAKE_PREFIX_PATH` is searched before `CMAKE_PROGRAM_PATH`, and `CMAKE_PROGRAM_PATH` before `PATH`).
Though, it's really unusual to package CMake config file in root folder, I've never seen that.

conan/tools/cmake/toolchain.py
lasote: I'm wondering if the user toolchain is setting `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE STREQUAL "ONLY"`, wouldn't be mostly the same modifying the `CMAKE_FIND_ROOT_PATH`?. Why is that ok but not overriding the `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` variable?

SpaceIm: Not sure to understand the questions. Would you like to set `CMAKE_FIND_ROOT_PATH` with generator folder and non-root builddirs of requires, regardless of `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` value?

conan/tools/cmake/toolchain.py
SpaceIm: Yes it's complex, I'll try to explain the all logic when I have time, it's not obvious. I try to address https://github.com/conan-io/conan/pull/10186#discussion_r772198971.
If you want something simple, I can fallback to something like this:

```python
    template = textwrap.dedent("""
        ...

        {% if cross_building %}
        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_PACKAGE OR CMAKE_FIND_ROOT_PATH_MODE_PACKAGE STREQUAL "ONLY")
            set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE "BOTH")
        endif()
        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_PROGRAM OR CMAKE_FIND_ROOT_PATH_MODE_PROGRAM STREQUAL "ONLY")
            set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM "BOTH")
        endif()
        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_LIBRARY OR CMAKE_FIND_ROOT_PATH_MODE_LIBRARY STREQUAL "ONLY")
            set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY "BOTH")
        endif()
        {% if is_apple %}
        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK OR CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK STREQUAL "ONLY")
            set(CMAKE_FIND_ROOT_PATH_MODE_FRAMEWORK "BOTH")
        endif()
        {% endif %}
        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_INCLUDE OR CMAKE_FIND_ROOT_PATH_MODE_INCLUDE STREQUAL "ONLY")
            set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
        endif()
        {% endif %}
    """)
```

conan/tools/cmake/toolchain.py
lasote: No, just saying that you are protecting the user set value from `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` because `I would like to avoid to override this variable if set by user toolchain.` but then you are modifying the `CMAKE_FIND_ROOT_PATH`, so I'm wondering if it is not kind of "cheating".

lasote: In other words, why is so important to keep the user value of `CMAKE_FIND_ROOT_PATH_MODE_PACKAGE` (and others)?

SpaceIm: Ok, so when `CMAKE_FIND_ROOT_PATH_MODE_*` is `ONLY`, you have no other choice than manipulating `CMAKE_FIND_ROOT_PATH` because all other variables are ignored. `CMAKE_FIND_ROOT_PATH` is like `CMAKE_PREFIX_PATH` but with higher precedence, so no it's not cheating.

Why is it important to honor `CMAKE_FIND_ROOT_PATH_MODE_*` of users? Principle of least surprise.
But as I said before, if `ONLY` is overridden by `BOTH`, it should not break consumers (other proposal in https://github.com/conan-io/conan/pull/10186#discussion_r774247442), because system paths have the lowest priority in search paths of `find_*()` commands, and `BOTH` still look into `CMAKE_FIND_ROOT_PATH` first if the user has injected some stuff not managed by conan.

lasote: Then I would simplify and just override to `BOTH`. 

> Ok, so when CMAKE_FIND_ROOT_PATH_MODE_* is ONLY, you have no other choice than manipulating CMAKE_FIND_ROOT_PATH because all other variables are ignored. CMAKE_FIND_ROOT_PATH is like CMAKE_PREFIX_PATH but with higher precedence, so no it's not cheating.

Yes, I got that, but:

> Why is it important to honor CMAKE_FIND_ROOT_PATH_MODE_* of users? Principle of least surprise.

Not sure that changing `CMAKE_FIND_ROOT_PATH` is `least surprise` either, so I vote for simplifying and just overriding to BOTH when cross_build.

conan/tools/cmake/toolchain.py
lasote: Let's keep it out for now. Better to add it if requested at some point.

conan/tools/cmake/toolchain.py
SpaceIm: simplified in https://github.com/conan-io/conan/pull/10186/commits/1e89fe920cc491611dbff41b377e29100e323219

`CMAKE_FIND_ROOT_PATH_MODE_*` of user honored unless it's `ONLY`.

conan/tools/cmake/toolchain.py
```diff
@@ -395,28 +395,56 @@ def context(self):
...
+        endif()
         {% endif %}
-        """)
+        if(NOT DEFINED CMAKE_FIND_ROOT_PATH_MODE_INCLUDE OR CMAKE_FIND_ROOT_PATH_MODE_INCLUDE STREQUAL "ONLY")
+            set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE "BOTH")
+        endif()
+        {% endif %}
+    """)
+
+    @staticmethod
```
lasote: Hi @SpaceIm, 
I have been talking with the team and discussing how to introduce this PR and the migration to 2.0.  We all understood the issue and agree we must do something.

We have a question regarding this block of code of the `CMAKE_FIND_ROOT_PATH_MODE_*` stuff. We want to simplify it because it is a bit complex and could be problematic. We thought that maybe we could append `{{ generators_folder }} {{ host_build_paths_noroot }}` to the `CMAKE_FIND_ROOT_PATH` always without checking the value of the `CMAKE_FIND_ROOT_PATH_MODE_*` variables? WDYT?

SpaceIm: It would break `find_program()` of buid context executables if:
- the same recipe is in build_requires and requires
- `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is `ONLY`

SpaceIm: Too fragile.

It would break in cross-build scenario for these cases:
- `find_program()` of build context executables when `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` is `ONLY` (quite uncommon in toolchain files)
- `find_library()` when `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY` is `ONLY` (very common in cross-build toolchain files, or even set automatically to this value by CMake if not set in the toolchain file and cross-build detected for iOS or Android).
- `find_file()` and `find_path()` when `CMAKE_FIND_ROOT_PATH_MODE_INCLUDE` is `ONLY` (same comment than `CMAKE_FIND_ROOT_PATH_MODE_LIBRARY`).

You can easily test your suggestion in this PR, I've added 2 flavours of test: `CMAKE_FIND_ROOT_PATH_MODE_*` not set by a toolchain file (or in cmake command line), or `CMAKE_FIND_ROOT_PATH_MODE_*` set to usual values when cross-building (`ONLY` for all of them, except `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` set to `NEVER`).
I've tested this modification, and as expected it breaks tests of `find_library()` and `find_file()`/`find_path()` if cross-building for iOS.

lasote: Oh, I see, thanks.
So, the difference between that code block and setting everything to "BOTH" is protecting the user's toolchain choice when "NEVER", right?

SpaceIm: Yes, this toolchain doesn't change `CMAKE_FIND_ROOT_PATH_MODE_*` if user toolchain has set one of these variables to `NEVER` or already `BOTH`.
Because all the logic in this PR is very robust to `NEVER` (I think it's the default behavior for native build in CMake under the hood) so it's fine for this value. Also if user has set `CMAKE_FIND_ROOT_PATH_MODE_PROGRAM` to `NEVER`, he may want to prevent `find_program()` to find programs in its cross-build sysroot, so it's better to honor this value.
`BOTH` is also fine if the value was `ONLY` in user toolchain (if it was `ONLY` we want `BOTH` to still look into `CMAKE_FIND_ROOT_PATH` because user wanted this behavior by setting to `ONLY`), because this toolchain doesn't manipulate `CMAKE_FIND_ROOT_PATH`. Of course if user toolchain injects stuff in `CMAKE_FIND_ROOT_PATH` which could take precedence over conan packages, it's the responsability of the consumer. If consumer finds build context libraries or headers (in the system probably) due to `BOTH`, it's likely because some requires are missing, `ONLY` is some sort of debugging helper but when you have all your requires properly set, `BOTH` instead of `ONLY` won't break anything.

lasote (approved):

SpaceIm:
some conflicts to solve

memsharded:
I am contributing a fix to the broken test in https://github.com/conan-io/conan/pull/10186/commits/6bb0be80e0fa87322572abf97a4b3c3cd9291934.

The test was failing because we added this new test to run CMake with ``-werror -warn``, and this was detecting a bunch of warnings because of usage of uniniatilized variables. I am changing those lines to ``list(PREPEND ...)`` to avoid that warning, and  I also took a chance and splitted the CMAKE_MODULE_PATH and CMAKE_PREFIX_PATH in multiple lines, adding a comment to each one. I think it will be useful for readers trying to understand what the toolchain is doing and why is adding some folders to those variables (the explanation in each line comment might be poor, it would be great to add the full explanation: "adding to CMAKE_PREFIX_PATH so we can locate the ....config.cmake inside, etc, etc")

memsharded (approved):

memsharded:
I had to push a few changes more, after merging latest "develop" state in https://github.com/conan-io/conan/pull/10186/commits/5d19b4f7921ee1ac5bd7930bd3f0d34ff31a931c.

The most important thing that changed is that now the ``cppinfo`` directories are absolute, everything was almost good, except the comparison of ``builddirs == ""``, that was broken. Now it seems to be fixed.

PR #2797, opened 2018-04-24
Fix abs paths with pkg_config generator

danimtb:
- [x] Refer to the issue that supports this Pull Request: closes #2756
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://raw.githubusercontent.com/conan-io/conan/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [x] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. Also adding a description of the changes in the ``changelog.rst`` file. https://github.com/conan-io/docs

Issue #2756, opened 2018-04-13
Invalid formatting of libdir/includedir in pkg_config generator [bug]


coffee-lord:
Version: conan-1.2.3
OS: Linux

https://github.com/conan-io/conan/blob/ac25c8994878706bac23efd009ae74bc294c1add/conans/client/generators/pkg_config.py#L28-L36

`pkg_config` generator doesn't check if libdir or includedir are absolute paths which leads to the following (invalid) `.pc` being generated for `boost_math/1.66.0@bincrafters/stable`:

```
prefix=/home/user/.conan/data/boost_math/1.66.0/bincrafters/stable/package/937af278b1c9d86290dddaf1322dc99fb7d75862
libdir=${prefix}//home/user/.conan/data/boost_level8group/1.66.0/bincrafters/stable/package/afc149253b16bde7d64e527e0dbfa27b203330c8/math/lib
includedir=${prefix}//home/user/.conan/data/boost_level8group/1.66.0/bincrafters/stable/package/afc149253b16bde7d64e527e0dbfa27b203330c8/math/include
...
```

danimtb:
Hi @coffee-lord !

Thanks you so much fot the detailed report. Seems it is a bug with absolute paths as you said, mainly affecting recipes with `short_paths=True` like boost ones.

We will try to fix it for the next release.
Thanks!

coffee-lord:
@danimtb While we're on the subject, pkg_config generator also doesn't emit `-Wl,-rpath=` flags. Is that intentional?

danimtb:
Hi @coffee-lord 

Yes, those flags are missing from the logic of the generator. Let's update it with those flags now that we have a PR open for this.

Thanks!

coffee-lord:
With regards to #2797:

**1. `-L${libdir}` is added unconditionally**

https://github.com/conan-io/conan/blob/8f9f56496249247ec619d9e973b6b6a73031e433/conans/client/generators/pkg_config.py#L45-L48

Which leads to "Variable 'libdir' not defined" error in pkg-config when `dir_lines` is empty.

```
prefix=...
includedir=...

Name: boost_conversion
Description: Conan package: boost_conversion
Version: 1.66.0
Libs: -L${libdir} ...
```

**2. Irrelevant rpaths**

[...]

PR #10192, opened 2021-12-17
Remove -Wl,-rpath from `AutotoolsDeps`/`GnuDeps`/`PkgConfigDeps`

### LINKED ISSUE #7878: [feature] pkg_config/PkgConfigDeps generators: -Wl,-rpath="${libdir}" should be removed

Currently (conan 1.30.1, but for a long time), pkg_config generator creates a pkg-config file like this on Linux:

```
prefix=/home/spaceim/.conan/data/h3/3.7.0/_/_/package/7fe67a73ac24eef3dbbe871cec7ad5a0356f543a
libdir=${prefix}/lib
includedir=${prefix}/include

Name: h3
Description: Hexagonal hierarchical geospatial indexing system.
Version: 3.7.0
Libs: -L${libdir} -lh3  -lm  -Wl,-rpath="${libdir}"
Cflags: -I${includedir} -DH3_PREFIX=
```

When this pkg-config file is consumed in another project to produce an executable or shared lib, rpath is hardcoded in this runtime file. It leads to some inconsistent behaviour where some paths are in rpath and others not depending on whether the lib has been consumed through pkgconf or not in build system of a recipe.
Moreover, those kinds of absolute RPATH hardcoded in runtimes don't make sense at all for binaries produced on another machine.

"Official" pkg-config files usually never have `-Wl,-rpath` in `Libs`.

SpaceIm:
Changelog: Fix: remove rpath from .pc files generated by `pkg_config` & `PkgConfigDeps` generators.
Docs: https://github.com/conan-io/docs/pull/2339

closes https://github.com/conan-io/conan/issues/7878

### DIFF
--- a/conans/client/generators/compiler_args.py
+++ b/conans/client/generators/compiler_args.py
@@ -2,7 +2,7 @@ from conans.client.build.compiler_flags import (architecture_flag, build_type_de
                                                 build_type_flags, format_defines,
                                                 format_include_paths, format_libraries,
                                                 format_library_paths, libcxx_define, libcxx_flag,
-                                                rpath_flags, sysroot_flag,
+                                                sysroot_flag,
                                                 visual_linker_option_separator, visual_runtime,
                                                 format_frameworks, format_framework_paths)
 from conans.client.build.cppstd_flags import cppstd_flag_new as cppstd_flag
@@ -67,7 +67,6 @@ class CompilerArgsGenerator(Generator):
         if not hasattr(self.conanfile, 'settings_build'):
             os_build = os_build or self.conanfile.settings.get_safe("os")
 
-        flags.extend(rpath_flags(self._settings, os_build, self._deps_build_info.lib_paths))
         flags.extend(format_library_paths(self._deps_build_info.lib_paths, self._settings))
         flags.extend(format_libraries(self._deps_build_info.libs, self._settings))
         flags.extend(format_libraries(self._deps_build_info.system_libs, self._settings))
--- a/conans/client/generators/compiler_args.py
+++ b/conans/client/generators/compiler_args.py
@@ -2,7 +2,7 @@ from conans.client.build.compiler_flags import (architecture_flag, build_type_de
                                                 build_type_flags, format_defines,
                                                 format_include_paths, format_libraries,
                                                 format_library_paths, libcxx_define, libcxx_flag,
-                                                sysroot_flag,
+                                                rpath_flags, sysroot_flag,
                                                 visual_linker_option_separator, visual_runtime,
                                                 format_frameworks, format_framework_paths)
 from conans.client.build.cppstd_flags import cppstd_flag_new as cppstd_flag
@@ -67,6 +67,7 @@ class CompilerArgsGenerator(Generator):
         if not hasattr(self.conanfile, 'settings_build'):
             os_build = os_build or self.conanfile.settings.get_safe("os")
 
+        flags.extend(rpath_flags(self._settings, os_build, self._deps_build_info.lib_paths))
         flags.extend(format_library_paths(self._deps_build_info.lib_paths, self._settings))
         flags.extend(format_libraries(self._deps_build_info.libs, self._settings))
         flags.extend(format_libraries(self._deps_build_info.system_libs, self._settings))
--- a/conans/client/generators/pkg_config.py
+++ b/conans/client/generators/pkg_config.py
@@ -1,6 +1,6 @@
 import os
 
-from conans.client.build.compiler_flags import rpath_flags, format_frameworks, format_framework_paths
+from conans.client.build.compiler_flags import format_frameworks, format_framework_paths
 from conans.client.tools.oss import get_build_os_arch
 from conans.model import Generator
 from conans.model.conan_generator import GeneratorComponentsMixin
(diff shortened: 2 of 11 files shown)


conans/client/generators/pkg_config.py
```diff
@@ -16,7 +16,7 @@
 Name: my-project
 Description: Some brief but informative description
 Version: 1.2.3
-Libs: -L${libdir} -lmy-project-1 -linkerflag -Wl,-rpath=${libdir}
```
memsharded: The thing is that if this is here, it was probably requested by some users, and it might be breaking for them. Maybe it would be better to not touch the ``pkg_config`` and only change ``PkgConfigDeps``

Also taking into account that other generators, like autotools (both the legacy and the new ``conan.tools.gnu`` ones) also apparently add these flags.

memsharded (commented):
If you could elaborate a bit further why these flags have to be removed from ``pkg_config`` only, that would be very helpful. These flags are also being applied to autotools helpers (both legacy and new), and to general compiler_flags, that could be used by other user integrations.

conans/client/generators/pkg_config.py
SpaceIm: It's explained in the linked issue, those flags are never added to official pkgconfig files nor CMake config files. If it breaks some people, they are doing something wrong.
Adding these flags has side effects in executables & shared libs built with dependencies discovered through these pkgconfig files.

SpaceIm: I've found this PR adding rpath https://github.com/conan-io/conan/pull/2797 among other modifications (seems to be the first PR after first introduction of `pkg_config` generator), there is a request to add it in this comment https://github.com/conan-io/conan/issues/2756#issuecomment-383908889 without much rational.

/cc @coffee-lord and @danimtb if you can remember why `-Wl,-rpath` has been added to .pc files.

SpaceIm: I guess that it should also be removed from `AutoToolsBuildEnvironment` helper and `AutotoolsDeps` generator, but I didn't want to change too many conan tools in the same PR.
It's worth noting that it's optional in `AutoToolsBuildEnvironment`, defaulted to False, and CCI recipes never enable rpath in this helper.

memsharded: > It's explained in the linked issue, those flags are never added to official pkgconfig files nor CMake config files. If it breaks some people, they are doing something wrong.

It is not enough to say that they are doing something wrong. Conan is doing it, and it has been doing it for years. If they rely on it (and I can guarantee some people will) and Conan removes it, it will be breaking behavior, and we cannot do that. Unless it is declared a clear bug (which I doubt can be done), then it is just not the best default behavior, but that cannot be used to break (in the same way we haven't been able to change libcxx=libstdc++11 default).

> It's worth noting that it's optional in AutoToolsBuildEnvironment, defaulted to False, and CCI recipes never enable rpath in this helper.

Yes, same, this probably cannot be removed without being breaking, even if not used in CCI, it can be used by users own recipes.

What is clear, is that if this is to be removed, we should make sure it is being removed from ``GnuDepsFlags`` too and not only ``PkgConfigDeps``

SpaceIm: It's a bug because `pkg_config` and `PkgConfigDeps` aim to emulate pkgconfig files, and no one adds -Wl-rpath in pkgconfig files generated for a lib (I refer to official pkgconf files generated by build systems), because it doesn't make sense to make this decision for consumers.
So it's different than `AutoToolsBuildEnvironment` which is pure conan construction (and I recommend to not add `-Wl,-rpath` to `GnuDepsFlags` or `AutotoolsDeps`, or if you want to keep this feature, add an option off by default).

As a consequence, when you consume these pkgconfig files in your build system, it will hardcodes path of dependencies in rpath of generated runtime files. It should not be the default behavior, specially for conan whose purpose is to distribute relocatable files.


Now I understand that it could break some people, not sure to understand why someone would rely on this flags coming from pkgconfig files, instead of expecting what a normal pkgconfig file should provide. So how to procede? Add an option in `PkgConfigDeps`? What about `pkg_config`?

memsharded: That decision was made for consumers 3 years ago. Maybe it was not the right decision, but it was done, as a feature.
I am very happy to remove it from ``PkgConfigDeps``, and we can try to remove it from ``pkg_config`` but if users complain that we are breaking something that was working in the past, then, they will not care that the official pkgconfig files by build systems do differently, they have been trusting Conan to do this for them for years, and now we will break their builds.

You can say exactly the same for ``libcxx=libstdc++`` by default, that it is a bug, because it should be ``libstdc++11++ which is the compiler default. But the truth is that Conan was working well for them, building and running correctly, and if we change that, their builds will completely break. The thing is that it is not intrinsically a bug, but a bad default, which is not the same.

I 100% understand your point, but honestly, I am very tired of users complaining because we "broke" something that was working in the past, I'd prefer to avoid that possibility as much as possible, because it is very time and energy consuming for us (that typically implies having to do a patch release).

SpaceIm: Ok, so would it be ok to remove it in new generators of conan v2 (not stable yet), or add an option off by default?

memsharded: It can be removed from new generators without needing to add an option. Lets do the right thing this time, and if someone request this feature for the new generators, we will make sure to understand the use case and see if it really makes sense, and it is the correct place (in any case, it would be more a "toolchain" thing IMO)

lasote (approved):

conans/client/generators/compiler_args.py
```diff
@@ -2,7 +2,7 @@
                                                 build_type_flags, format_defines,
                                                 format_include_paths, format_libraries,
                                                 format_library_paths, libcxx_define, libcxx_flag,
-                                                rpath_flags, sysroot_flag,
+                                                sysroot_flag,
```
memsharded: compiler_args is a legacy generator, not a new one, better leave it as-is

memsharded (commented):
Almost there, please check the compiler-args issue, and lets merge it for next 1.44

memsharded:
Many thanks for contributing this!

Issue #10193, opened 2021-12-17
[bug] circular dependency of build tools if used in build profile


thorsten-klein:
### Environment Details (include every applicable attribute)
  * Operating System+version: Ubuntu
  * Compiler+version: gcc-10
  * Conan version: 1.43.0
  * Python version: 3.8.10

### Steps to reproduce (Include if Applicable)
myprofile
```
[settings]
arch=x86_64

compiler=clang
compiler.libcxx=libstdc++11
compiler.version=13

os=Linux

[options]

[build_requires]
build_tool_1/1.0.0@dummy/dummy
build_tool_2/1.0.0@dummy/dummy

[env]
MYPROFILEVAR="myval"

[conf]
```

conanfile_build_tool_1.py
```
from conans import ConanFile, CMake
import os

class pkg(ConanFile):
    name = "build_tool_1"
    version = "1.0.0"
    settings = "os", "arch"
```

conanfile_build_tool_2.py
```
from conans import ConanFile, CMake
import os

class pkg(ConanFile):
    name = "build_tool_2"
    version = "1.0.0"
    settings = "os", "arch"

    def build_requirements(self):
        self.build_requires("build_tool_1/1.0.0@dummy/dummy")
```

conanfile_consumer.py
```
from conans import ConanFile, CMake
import os

class pkg(ConanFile):
    name = "consumer"
    version = "1.0.0"
    settings = "os", "arch", "compiler", "build_type"

```

steps to reproduce:
```
conan export conanfile_build_tool_1.py dummy/dummy
conan export conanfile_build_tool_2.py dummy/dummy

conan create  conanfile_consumer.py -pr:b=./myprofile -pr:h=./myprofile --build=consumer --build=build_tool_1 --build=build_tool_2
issue:
```

### Logs (Executed commands with output) (Include/Attach if Applicable)

```
ERROR: Loop detected in context build: 'build_tool_1/1.0.0@dummy/dummy' requires 'build_tool_2/1.0.0@dummy/dummy' which is an ancestor too
```

thorsten-klein:
Suggestion: Solve the circular dependency in the profile via
```
!build_tool_1/*: build_tool_2/1.0.0@dummy/dummy
```

Currently fnmatch is used here, but this does not allow to negotiate the expression. Therefore lets handle the ! in the first position explicitly. I will create some patch.

AbrilRBS:
Closing as solved with the proposed solution now that it has been working for a good while :)

PR #10197, opened 2021-12-17
allow negotiated conditions for build_requires in profiles

thorsten-klein:
Changelog: (Bugfix): Allow negotated conditions for build_requires in profiles (#10193)
Docs: https://github.com/conan-io/docs/pull/2328

### DIFF
--- a/conans/client/graph/graph_manager.py
+++ b/conans/client/graph/graph_manager.py
@@ -345,6 +345,7 @@ class GraphManager(object):
                     continue
                 if ((node.recipe == RECIPE_CONSUMER and pattern == "&") or
                         (node.recipe != RECIPE_CONSUMER and pattern == "&!") or
+                        (pattern.startswith("!") and not fnmatch.fnmatch(str_ref, pattern[1:])) or
                         fnmatch.fnmatch(str_ref, pattern)):
                     for build_require in build_requires:
                         br_key = (build_require.name, default_context)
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Hi @thorsten-klein 

I have reviewed this, and contributed a commit, simplifying a bit the test and removing some unnecessary bits.

I have checked the proposal, and I am not sure it will be useful in the general case. The issue with ``!tool1*:tool2/1.0`` is that it is very difficult to control to what packages the ``tool2`` package will be applied as build_require. Once you negate ``!tool1``, it will be applied to all packages, except ``tool1``. If there is some other package that shouldn't get it as build_require, there is no way with this approach to represent it.

With the positive approach, as it is additive, it can be represented, one condition at a time. In the provided test it is as simple (without requiring any other change) with specifying the profile:

```
            [build_requires]
            tool2*:tool1/1.0
            consumer*:tool2/1.0
```
will make the test pass.

We are considering other approaches of specifying build_requires for Conan 2.0, that might help with this case.

thorsten-klein:
Thanks a lot! :)
Yes, this is the disadvantage of the approach: in case of more than 2 tools it does not scale anymore. I also opened some issue for conan 2. I suggest to use some &+ pattern.

As you wrote consumer*, this does not work for bigger projects, since we do not know the name of all requirements. So we cannot find a pattern matching them all. 

What do you think about a &+ pattern?

thorsten-klein:
How shall we proceed with this PR? Do we merge it this way nevertheless? Or what is your plan?

thorsten-klein:
How shall we proceed with this PR? Do we merge it this way nevertheless? Or what is your plan?

thorsten-klein:
Outdated. No need to fix it anymore.

PR #10199, opened 2021-12-17
Extending loop checking beyond visibility to avoid infinite loops

memsharded:
While working on https://github.com/conan-io/conan/issues/10193, realized that it is easy that build_requires go into infinite loops in 2.0 dependency resolution. This PR aims to improve this.

### DIFF
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -151,6 +151,16 @@ class Node(object):
         source_node = dependant.src
         return source_node.check_downstream_exists(down_require)
 
+    def check_loops(self, new_node):
+        if self.ref == new_node.ref:
+            return self
+        if not self.dependants:
+            return
+        assert len(self.dependants) == 1
+        dependant = self.dependants[0]
+        source_node = dependant.src
+        return source_node.check_loops(new_node)
+
     @property
     def package_id(self):
         return self._package_id
--- a/conans/test/integration/graph/core/test_build_requires.py
+++ b/conans/test/integration/graph/core/test_build_requires.py
@@ -131,12 +131,6 @@ class BuildRequiresGraphTest(GraphManagerTest):
         _check_transitive(lib, [(cmake2, False, False, True, True)])
         _check_transitive(cmake2, [(cmake1, False, False, True, True)])
 
-
-
-
-
-
-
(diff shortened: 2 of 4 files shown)


conans/client/graph/graph.py
```diff
@@ -151,6 +151,16 @@ def check_downstream_exists(self, require):
         source_node = dependant.src
         return source_node.check_downstream_exists(down_require)
 
+    def check_loops(self, new_node):
+        if self.ref == new_node.ref:
+            return self
+        if not self.dependants:
+            return
+        assert len(self.dependants) == 1
```
lasote, line 159: Is this because iterating "Depth First"?

conans/test/integration/graph/core/test_build_requires.py
```diff
@@ -585,3 +579,44 @@ def build_requirements(self):
...
+        # Build requires always apply to the consumer
+        self.assertEqual(2, len(deps_graph.nodes))
+        app = deps_graph.root
+        tool = app.dependencies[0].dst
+
+        self._check_node(app, "app/0.1", deps=[tool], dependents=[])
+        self._check_node(tool, "cmake/0.1#123", deps=[], dependents=[app])
+
+    def test_indirect_loop_error(self):
+        # app -(br)-> gtest/0.1 -(br)-> cmake/0.1 -(br)->gtest/0.1 ....
```
lasote, line 604: Related to my comment, should we add a test for:

```
# app -(br)-> gtest/0.1 -(br)-> cmake/0.1 -(br)->gtest/0.1 ....
                       \ (br)-> foo/0.1  --(br) / (edited) 
```

lasote (approved):

conans/client/graph/graph.py
memsharded: Yes, this is only called when a new node is added to the graph, that only happens before having diamonds downstream (and then, only 1 dependant always)

conans/test/integration/graph/core/test_build_requires.py
memsharded: Trying to add this structure, but that can never happen. build-requires do not close the diamond, because they are not visible downstream. And in any case, only the first branch is expanded first, causing exactly the same scenario and resulting graph than the current existing test.

PR #10207, opened 2021-12-18
convert NewCppInfo folders to absolute

memsharded:
Changelog: Fix: Convert `NewCppInfo` folders to absolute.
Docs: Omit

Previous to be able to define ``--source-folder`` and ``--build-folder`` for editables and ``layout()`` it is necessary to convert the paths in ``NewCppInfo`` to absolute before they get to generators:
- It simplifies the job of consumers and generators, folders are absolute, no need to call ``.set_relative_base_folder()`` needing to understand what folder it would be.
- Generators like ``CMakeDeps`` that want to factorize to a common xxx_PACKAGE_FOLDER can always check the common prefix and extract it
- This change seems to be unveiling a couple of bugs: ``cpp_info.objects``, not handled by ``set_relative_folders()`` (it would never work with editables or src/build layouts), and ``cmake_build_modules`` not being handled either by ``set_relative_folders()``. This one is more difficult to fix, needs some extra model to be solved.

#tags: slow

### DIFF
--- a/conans/model/new_build_info.py
+++ b/conans/model/new_build_info.py
@@ -247,29 +247,6 @@ def from_old_cppinfo(old):
     ret = NewCppInfo()
     ret.merge(old)
     ret.clear_none()
-
-    def _copy_build_modules_to_property(origin, dest):
-        current_value = dest.get_property("cmake_build_modules") or []
-        if isinstance(origin.build_modules, list):
-            current_value.extend([v for v in origin.build_modules if v not in current_value])
-        else:
-            multi = origin.build_modules.get("cmake_find_package_multi")
-            no_multi = origin.build_modules.get("cmake_find_package")
-            if multi:
-                current_value.extend([v for v in multi if v not in current_value])
-            if no_multi:
-                current_value.extend([v for v in no_multi if v not in current_value])
-        dest.set_property("cmake_build_modules", current_value, path=True)
-
-    # Copy the build modules as the new recommended property
-    if old.build_modules:
-        _copy_build_modules_to_property(old, ret)
-
-    for cname, c in old.components.items():
-        if c.build_modules:
-            # The build modules properties are applied to the root cppinfo, not per component
-            # because it is something global that makes no sense to be set at a component
-            _copy_build_modules_to_property(c, ret)
     return ret
 
 
--- a/conans/model/new_build_info.py
+++ b/conans/model/new_build_info.py
@@ -47,10 +47,10 @@ class _NewComponent(object):
             return []
         return [r for r in self.requires if "::" not in r]
 
-    def set_property(self, property_name, value, path=False):
+    def set_property(self, property_name, value, is_path=False):
         if self._generator_properties is None:
             self._generator_properties = {}
-        self._generator_properties[property_name] = (value, path)
+        self._generator_properties[property_name] = (value, is_path)
 
     def get_property(self, property_name):
         if self._generator_properties is None:
@@ -75,8 +75,6 @@ class NewCppInfo(object):
         # Main package is a component with None key
         self.components[None] = _NewComponent()
         self._aggregated = None  # A _NewComponent object with all the components aggregated
-        self._sorted_components = None  # Caching the sorted components
-        self._defined_base_folder = False  # To prevent errors and double definition
 
     def __getattr__(self, attr):
         return getattr(self.components[None], attr)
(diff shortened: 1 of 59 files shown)


conans/model/new_build_info.py
```diff
@@ -47,21 +47,21 @@ def required_component_names(self):
             return []
         return [r for r in self.requires if "::" not in r]
 
-    def set_property(self, property_name, value, generator=None):
+    def set_property(self, property_name, value, generator=None, path=False):
```
lasote: I would prefer `is_path`, but anyway, why is not the consumer of the property the one "adapting" the values?

conans/test/integration/generators/cpp_info_set_generator_properties_test.py
```diff
@@ -78,7 +78,7 @@ def package(self):
             def package_info(self):
                 self.cpp_info.set_property("cmake_file_name", "AnotherFileName")
                 self.cpp_info.components["mycomponent"].set_property("cmake_target_name", "mycomponent-name-but-different")
-                self.cpp_info.components["mycomponent"].set_property("cmake_build_modules", [os.path.join("lib", "non-existing.cmake")])
+                self.cpp_info.components["mycomponent"].set_property("cmake_build_modules", [os.path.join("lib", "non-existing.cmake")], path=True)
```
lasote: I don't like this much. The user will forget to set the `path=True`. We should try to do the path conversions in the consumer (CMakeDeps generator)

conans/model/new_build_info.py
memsharded: Because that is a big challenge. 
The consumers receive all the ``includedirs`` and ``properties`` already merged. When the installer processes the ``layout()`` and take into accounts editables, the ``cpp.source`` and ``cpp.build`` are merged into the final ``cpp.package`` (or ``cpp_info``). If we want the consumer to be able to handle that we would need:
- To add to every single item (every dir in ``includedirs``, ``libdirs``, and every property), the information if it is a ``source``, ``build`` or ``package`` item
- To make accessible to the consumer the final ``build_folder``, ``source_folder``, and ``package_folder`` (the later is always available, but the others are not for normal packages, they are set to None in ``installer``, because the build folder and source folder don't even exist, unless they are editable
- To let the consumer know which packages are in editable mode.
- Then, every consumer needs to be composing with ``set_relative_folder``, passing the right folder depending if it is editable or not, and using the right folder depending on the origin. If this functionality can be wrapped in a single call for convenience (a call that must be called by every generator in order to work), then it makes no sense to let the consumers call it, and not have it called automatically by Conan, so it is already processed when it gets to the generators.

lasote: I see. I'm afraid we are digging the hole again. Anyway, please, call it `is_path`

lasote (approved):

drmacdon:
The current cmake/clion layout() examples construct the build folder name as a concatenation using the build_type keeping everything relative. When running "conan install -if xxx" that install/build folder is unavailable to the conan recipe layout(), in my case I would have been happy if conan created xxx and gave me that as the relative build folder to start my layout, actually think I'd use the default ;-). I need some way to create multiple build folders where the user controls the build folder name not the recipe, but with layout() I feel like I lost that ability. Will these changes you are working on help?

memsharded:
> The current cmake/clion layout() examples construct the build folder name as a concatenation using the build_type keeping everything relative. When running "conan install -if xxx" that install/build folder is unavailable to the conan recipe layout(), in my case I would have been happy if conan created xxx and gave me that as the relative build folder to start my layout, actually think I'd use the default ;-). I need some way to create multiple build folders where the user controls the build folder name not the recipe, but with layout() I feel like I lost that ability. Will these changes you are working on help?

Hi @drmacdon 

I am not sure exactly what is your issue, but I can clarify that this work here is the base for https://github.com/conan-io/conan/pull/10274, which recovers the ``--build-folder`` functionality for ``layout()`` method. Is this what you are looking for?

drmacdon:
Possibly, but I tried to use your latest repo changes (0ee5e) to see if it would give me what I needed but it still shows up as empty string or None which means I don't fully understand. I am not using editable mode references, I am using V2 CMakeDeps, CMakeToolchain. With conan version 1.45.0-dev, I added a print statement and tried to pass the bf/if but didn't get what I thought I would.  

```python
    conan install -bf cmake-build-target1 -if cmake-build-target1 -pr:b default -pr:h target1 .

    def layout(self):
        print("self.folders.build: {}".format(self.folders.build))                  # returned empty 
        print("self.folders.build_folder: {}".format(self.folders.build_folder))    # returned None
        print("self.folders.source_folder: {}".format(self.folders.source_folder))  # returned None
  ```
---
To provide context on my question, using the current conan version 1.44.0 with a custom layout I end up with the problem below which I'm hoping your changes with layout are going to assist.

**Problem**: *Having multiple build folders and a sane conan generation folder*
```bash
 conan install -if cmake-build-target1 -pr:b default -pr:h target1 .
```
With no layout, ```self.folders.generators``` assumes the CLI provided folder ```cmake-build-target1```, and I end up with 93 files generated in my root build folder for all the find packages I'm currently using which is nasty. I like the example layouts generating into conan sub-folder. But if I just specify 
```
        self.folders.generators = "conan"
```
it is not relative to the CLI install folder ```cmake-build-target1``` and instead conan creates two folders off my source tree. What I'm looking for is how to obtain the CLI provided ```cmake-build-target1``` folder so I can make my generators relative to it instead of relative to my recipe. Ideally, the default would have placed them into their own sub-folder instead of placing into the build folder which on large projects is already crowded enough.

I can not use the example ```cmake_toolchain``` route of ignoring the user CLI provided build folder because I like having different profile targets in different build folders. Using the example layout, I would have to delete the ```cmake-build-debug``` each time I wanted to switch profiles. So I'm stuck letting conan pollute the root build folder and can't standardize on using ```-DCMAKE_TOOLCHAIN_FILE="conan/conan_toolchain.cmake"``` for recipes. 

---

**Inconvenience**: *Source layout not found*

```bash
conan create . test/1.0@ -pr:b default -pr:h target1 
```
The default ```self.folders.source``` assumes the build/$PACKAGE_ID folder and is unable to locate the CMakeLists.txt. Because I used a ```scm {"subfolder": name}``` and the default layout does not appear to take that into account although in theory it could.  It's not a problem, just something I think conan could have deduced. 
```
    def layout(self):
        self.folders.source = self.name
```
---
**Usability issue**: *Attempting to find my source folder, I managed to have conan corrupt my ```~/.conan/data```.*
```bash
conan search protobuf
ERROR: Specify the 'user' and the 'channel' or neither of them
```
Caused by a layout which caused conan to create a 'test/1.0/_/source' folder which causes later commands to fail until it is deleted. It is quite confusing to understand what the CLI is trying to say about the search operation above and I technically didn't create that folder albeit my settings must have.
```python
   conan create . test/1.0@ -pr:b default -pr:h target1 

   def layout(self):
        self.folders.source = os.path.dirname(__file__)
```

memsharded:
Hi @drmacdon 

This PR has been closed, not the best place to continue a conversation. Please open a new issue to discuss your use case.
In any case, I think that the functionality of ``layout()`` is being misinterpreted. You cannot read its values in ``layout()`` because the ``layout()`` is the place for defining its values, not for reading them. Doing any action on the ``layout()`` method is not allowed.

Using something like ``self.folders.source = os.path.dirname(__file__)`` is also not recommended, because that can be an absolute path to where the ``conanfile.py`` is, which in the cache will be a different folder``. The ``layout()`` is intended to define "relative" layouts, not absolute.

Please open a new ticket and we can follow up there, thanks!

PR #10213, opened 2021-12-20
Custom CMakeToolchain blocks without subclass

### LINKED ISSUE #10118: [feature] Improve custom CMakeToolchain blocks extension
From https://github.com/conan-io/docs/pull/2312, lets avoid the ``import CMakeToolchainBlock`` altogether, using Python dynamism.

lasote:
Changelog: Feature: Adding a new Block to the CMakeToolchain now doesn't require inheriting `CMakeToolchainBlock`. 
Docs: https://github.com/conan-io/docs/pull/2337
Closes #10118

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -654,7 +654,7 @@ class ToolchainBlocks:
         del self._blocks[name]
 
     def __setitem__(self, name, block_type):
-        if Block not in block_type.__mro__:  # tuple of classes that are considered base classes.
+        if not issubclass(block_type, Block):
             # Create a new class inheriting Block with the elements of the provided one
             block_type = type('proxyUserBlock', (Block,), dict(block_type.__dict__))
 
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -654,10 +654,8 @@ class ToolchainBlocks:
         del self._blocks[name]
 
     def __setitem__(self, name, block_type):
-        if not issubclass(block_type, Block):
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/toolchain.py
```diff
@@ -654,7 +654,9 @@ def remove(self, name):
         del self._blocks[name]
 
     def __setitem__(self, name, block_type):
-        self._blocks[name] = block_type(self._conanfile, self._toolchain)
+        # Create a new class inheriting Block with the elements of the provided one
+        new_class = type('proxyUserBlock', (Block,), dict(block_type.__dict__))
```
lasote: This should be enough, right?

memsharded: Yes, I think so, just need to fix the Block test that is doing ``super()``

lasote: Please, @memsharded  check the last commit. Is it worth it or should we drop the possibility of inheritance from blocks completely? The test failing was doing:

```
    class MyBlock(**GenericSystemBlock**):
        template = "Hello {{build_type}}!!"

        def context(self):
            c = super(MyBlock, self).context()
            c["build_type"] = c["build_type"] + "Super"
            return c

    toolchain.blocks["generic_system"] = MyBlock

```

memsharded: Good point, I'd say keep it simple at the moment, lets remove the inheritance case, sounds that with .values[] it is easier to achieve the same, and if users request it, we will re-consider.

memsharded (approved):

PR #10216, opened 2021-12-20
some minor fixes to premakedeps

memsharded:
Changelog: Omit
Docs: Omit

Follow up of https://github.com/conan-io/conan/pull/10090

Hi @Enhex 

I have seen some minor things in the contributed code, that I think deserves a bit more work:

- The ``Premake`` helper is mostly broken, needs to get some functional testing
- The ``PremakeDeps`` was doing some legacy things, like trying to define build_type or architecture. That is not the responsibility of "xxxDeps" anymore.
- I have added a couple of "fixmes" it seems the generation of file per dependency is missing things (for example, components are not being aggregated).
- Modernized the name to ``conandeps.premake.lua``
- Done some minor Pep8 cleanings (recall that we are also using a width of 100 chars)

Overall, this is a good first step, but this integration still needs some work to take it to production level, including some good functional testing, so it is necessary to keep pushing it. I recommended not to document it, until it is matured a little bit more.

### DIFF
--- a/conan/tools/premake/premake.py
+++ b/conan/tools/premake/premake.py
@@ -4,7 +4,7 @@ class Premake(object):
 
     # automatically chooses premake action based on used compiler
     def configure(self):
-        if "Visual Studio" in self.settings.compiler:
+        if "Visual Studio" in self._conanfile.settings.compiler:
             _visuals = {'8': '2005',
                         '9': '2008',
                         '10': '2010',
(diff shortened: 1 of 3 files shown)


Enhex:
to make sure, to my understanding the work i should do is:
- "include all" file should include the generated individual dep files instead of baking them in.
- `conan_build_type` and `conan_arch` should be part of PremakeToolchain.
- add tests, though I recall I couldn't get the testing framework to properly run on my machine so I might need help with that.

memsharded:
Hi @Enhex 

Yes, the priority is making sure whatever is in the codebase correctly works, before adding more functionality. Then, please, do not add ``PremakeToolchain`` until tests have been added. Same for the "include all" functionality or file (as the addition in Premake is explicit, it might be everything in the same file).

So lets focus on tests first then. I can setup a test for you in next iteration. Lets merge this for 1.44 and open a new issue for 1.45 to add these tests and improve the integration.

PR #10227, opened 2021-12-22
Fix module CMakeDeps vars and call find_dependency for transitive modules

### LINKED ISSUE #10224: CMakeDeps should use the module file name to create Find<pkg>.cmake variables
CMakeDeps is creating some variables in module mode using the name of the package. Instead of that, using the file name is probably more expected. Now, for example, for **libcurl**, it would generate `libcurl_FOUND` and `libcurl_VERSION`. Probably `CURL_FOUND` and `CURL_VERSION` is a bit more expected ([but not perfect](https://cmake.org/cmake/help/latest/module/FindCURL.html)). As commented in the linked discussion, it's impossible for Conan to address all the cases for the declared variables in the official CMake modules, but at least we can try to provide those.

Related to: https://github.com/conan-io/conan-center-index/pull/8294#issuecomment-993690510
and: https://github.com/conan-io/conan-center-index/pull/8425#issuecomment-993726762

czoido:
Changelog: Fix: Fix variable names set by `CMakeDeps` modules.
Changelog: Fix: Call to `find_dependency` in module mode to find transitive dependencies.
Changelog: Feature: Add `<PackageName>_LIBRARIES`, `<PackageName>_INCLUDE_DIRS`, `<PackageName>_INCLUDE_DIR`, `<PackageName>_DEFINITIONS` and `<PackageName>_VERSION_STRING` variables in CMakeDeps.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/10224

When only `cmake_find_mode == "module"` was used in CMakeDeps, transitive dependencies were not correctly found. Also, related with that variable names were not generated correctly.

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -63,15 +63,14 @@ class ConfigTemplate(CMakeDepsFileTemplate):
             find_dependency({{ '${_DEPENDENCY}' }} REQUIRED MODULE)
         endforeach()
 
-        {% endif %}
+        {% else %}
 
         foreach(_DEPENDENCY {{ '${' + file_name + '_FIND_DEPENDENCY_NAMES' + '}' }} )
-            # if we did not set the <dependency>_FOUND var in a Find<Package>.cmake MODULE
-            # let's try to find the CONFIG files
-            if(NOT {{ '${_DEPENDENCY}' }}_FOUND)
-                find_dependency({{ '${_DEPENDENCY}' }} REQUIRED NO_MODULE)
-            endif()
+            find_dependency({{ '${_DEPENDENCY}' }} REQUIRED NO_MODULE)
         endforeach()
+
+        {% endif %}
+
 
         # Only the first installed configuration is included to avoid the collision
         foreach(_BUILD_MODULE {{ '${' + pkg_name + '_BUILD_MODULES_PATHS' + config_suffix + '}' }} )
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -71,6 +71,9 @@ class ConfigTemplate(CMakeDepsFileTemplate):
             endif()
         endforeach()
 
+        set({{ file_name }}_INCLUDE_DIRS {{ '${' + pkg_name + '_INCLUDE_DIRS' + config_suffix + '}' }} )
+        set({{ file_name }}_LIBRARIES {{ '${' + pkg_name + '_LIBRARIES' + config_suffix + '}' }} )
+
         # Only the first installed configuration is included to avoid the collision
         foreach(_BUILD_MODULE {{ '${' + pkg_name + '_BUILD_MODULES_PATHS' + config_suffix + '}' }} )
             conan_message(STATUS "Conan: Including build module from '${_BUILD_MODULE}'")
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -45,22 +45,29 @@ def template(self):
...
-        include(${CMAKE_CURRENT_LIST_DIR}/{{ targets_include_file }})
-        include(CMakeFindDependencyMacro)
+        foreach(_DEPENDENCY {{ '${' + file_name + '_FIND_DEPENDENCY_NAMES' + '}' }} )
+            find_dependency({{ '${_DEPENDENCY}' }} REQUIRED MODULE)
+        endforeach()
+
+        {% endif %}
 
-        foreach(_DEPENDENCY {{ '${' + pkg_name + '_FIND_DEPENDENCY_NAMES' + '}' }} )
+        foreach(_DEPENDENCY {{ '${' + file_name + '_FIND_DEPENDENCY_NAMES' + '}' }} )
```
memsharded: A bit difficult to understand, why this is not in the ``else is_module``, without the ``if _FOUND``? If ``is_module``, then does it make sense to try to find the ``NO_MODULE``? Probably I don't understand how it works, but just in case.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
```diff
@@ -111,3 +111,53 @@ def test_reuse_with_modules_and_config(client):
...
     client.run("build . -if=install")
+
+
+@pytest.mark.parametrize("find_mode", ["both", "config", "module"])
+def test_transitive_modules_found(find_mode):
+    """
+    related to https://github.com/conan-io/conan/issues/10224
+    modules files variables were set with the pkg_name_FOUND or pkg_name_VERSION
+    instead of using filename_*, also there was missing doing a find_dependency of the
+    requires packages to find_package transitive dependencies
```
memsharded, line 122: Nice test with clear explanation 👍

conan/tools/cmake/cmakedeps/templates/config.py
czoido: This was thinking in the cmake_find_mode="both", but probably it makes no sense, I'll have a second thought on this

lasote (approved):

memsharded (approved):

conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -47,25 +47,33 @@ def template(self):
...
+            {% if is_module %}
+                find_dependency({{ '${_DEPENDENCY}' }} REQUIRED MODULE)
+            {% else %}
                 find_dependency({{ '${_DEPENDENCY}' }} REQUIRED NO_MODULE)
+            {% endif %}
             endif()
         endforeach()
 
+        set({{ file_name }}_INCLUDE_DIRS {{ '${' + pkg_name + '_INCLUDE_DIRS' + config_suffix + '}' }} )
+        set({{ file_name }}_LIBRARIES {{ '${' + pkg_name + '_LIBRARIES' + config_suffix + '}' }} )
```
SpaceIm: I don't know if they are defined elsewhere, but it would be very nice to add:
`<file_name>_INCLUDE_DIR>` (the no `S` flavour of `<file_name>_INCLUDE_DIRS>`)
`<file_name>_DEFINITIONS>`
`<file_name>_VERSION_STRING>`

Grep `_create_cmake_module_variables` in CCI, you'll see that these 3 variables are recurrent.

franramirez688 (approved):

conan/tools/cmake/cmakedeps/templates/config.py
memsharded: Can we easily add these @czoido ?

czoido: Yes, I'll add those 👍

PR #10235, opened 2021-12-28
Fixes for BazelDeps

photex:
- Correctly format paths on Windows
- Respect 'install_folder' provided to `conan install`

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -1,3 +1,5 @@
+import os
+import platform
 import textwrap
 
 from jinja2 import Template
@@ -32,7 +34,9 @@ class BazelDeps(object):
         self._save_main_buildfiles(content)
 
     def _save_dependendy_buildfile(self, dependency, buildfile_content):
-        filename = 'conandeps/{}/BUILD'.format(dependency.ref.name)
+        filename = '{}/conandeps/{}/BUILD'.format(
+            self._conanfile.install_folder, dependency.ref.name)
+        filename = filename.replace(os.sep, '/')
         save(filename, buildfile_content)
         return filename
 
@@ -43,7 +47,6 @@ class BazelDeps(object):
                 data = glob(["**"]),
                 visibility = ["//visibility:public"],
             )
-
         """).format(dependency.ref.name)
 
         return filegroup
@@ -55,7 +58,7 @@ class BazelDeps(object):
             {% for lib in libs %}
             cc_import(
                 name = "{{ lib }}_precompiled",
-                static_library = "{{ libdir }}/lib{{ lib }}.a"
+                static_library = "{{ libdir }}/{{ lib }}"
             )
             {% endfor %}
 
@@ -85,6 +88,10 @@ class BazelDeps(object):
 
         """)
 
+        system = platform.system()
+        if not system:
+            system == 'Linux'
+
         cpp_info = dependency.cpp_info.aggregated_components()
 
         if not cpp_info.libs and not cpp_info.includedirs:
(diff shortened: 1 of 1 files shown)


photex:
Ahoy! I see that I have some steps to follow to get this PR in shape.

conan/tools/google/bazeldeps.py
```diff
@@ -32,7 +33,8 @@ def generate(self):
         self._save_main_buildfiles(content)
 
     def _save_dependendy_buildfile(self, dependency, buildfile_content):
-        filename = 'conandeps/{}/BUILD'.format(dependency.ref.name)
+        filename = '{}/conandeps/{}/BUILD'.format(self._conanfile.install_folder, dependency.ref.name)
```
memsharded: If we can avoid having this path absolute to ``install_folder``, better, because the generation should already happen in the right location.

photex: If I run `conan install . -if .conan` then currently it hardcodes the path to 'conandeps' but it is actually written to '.conan/conandeps'.

I assumed the absolute path was fine since the path to the package is already absolute.

photex:
I've also noticed that the generated library targets are also not correct for Windows, I'll update the PR shortly.

I've created an issue to link to this at #10236

conan/tools/google/bazeldeps.py
memsharded: It might be fine atm, but I am thinking about the path to migrate to 2.0, in which the ``install_folder`` dissapears. But it is fine at the moment, I think this can be merged now and we'll check when we finally remove ``install_folder``

memsharded (approved):

memsharded:
We are going to branch right now for 1.44, I'll target this for 1.45.

> I've also noticed that the generated library targets are also not correct for Windows, I'll update the PR shortly.

So, there is still something missing?

BTW, there is already some tests in ``class BazelToolchainTest(Base):`` that seems to be running in Windows so far (it was only removed in OSX, because of weird CI random issues). It would be great to understand why it was passing in Windows, and maybe add some red/green test there. Not critical if the tests remains green with your changes, but could be useful.

conan/tools/google/bazeldeps.py
photex: Understood. I just wanted to patch this up and get something that works for me with Conan as it is today. 
I haven’t used Conan 2 yet.

photex:
What I’ve found and would still like to add is a fix for include paths using back slashes still.

memsharded:
> What I’ve found and would still like to add is a fix for include paths using back slashes still.

Of course. I am not even saying that this couldn't be merged now with the ``install_folder``, I just wanted to explain a bit the current design and how we plan to move it into the future, in case it was possible to change it in other way. As it is experimental, there is some space to evolving things in a pragmatic way, no need to be perfect since minute 0. Typically it is more important to start adding some tests that make sense and help to maintain things working in the future, than a perfect solution today without tests.

So as you prefer, if you are ok, I can merge it now for next 1.45 release. If you want to contribute a fix for the windows backslash, and treat the other issue later separately, that is good too.

photex:
Ok I've fixed two other issues with this and I was able to build a test app.

A good example to look at is the sdl package, it sets imported libraries, headers, and includes. 

I understand what you're saying now, and yes, I think it would be great to merge this for 1.45, then perhaps I can learn more about Conan 2. :+1:

photex:
@memsharded almost forgot. This last commit changes the name of the precompiled target. I don't have enough context to know a better way to make this work. 

Previously the library names were being hardcoded with a unix prefix and suffix. Could one of the reasons this might have been working in CI was that Mingw was being used instead of Msvc?

photex:
Also, I see the way I'm checking for the platform is breaking on CI for a linux target.

lasote:
Hi @photex ! 
I've been doing some improvements to the Bazel integration [here](https://github.com/conan-io/conan/pull/10812), I would say this PR is already covered and tested there. Thank you very much for your help, the feedback will be very welcome if you give it a try, it will be released at 1.47.

Issue #10237, opened 2021-12-28
[feature] [2.0] Explicit test package reference only


lasote:
When develop (1.44) is merged:

- Remove test_type
- Remove automatic injection of the reference being tested
- Change the "conan new" templates to use the new variable `self. tested_reference_str`

memsharded:
> Remove automatic injection of the reference being tested

I guess you mean the automatic injection of the ``require`` being tested, and keep the injection of the reference being tested

memsharded:
Closed by #10240

PR #10240, opened 2021-12-28
[develop2] remove test_type

memsharded:
Close https://github.com/conan-io/conan/issues/10237


conans/assets/templates/new_v2_cmake.py
```diff
@@ -59,6 +59,9 @@ class {package_name}TestConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
     generators = "CMakeDeps", "CMakeToolchain"
 
+    def requirements(self):
```
memsharded, line 62: I don't love the fact that it is not possible to do ``requires = something``, and this is still not a possibility for python_requires. But even with that, it still sounds better to be explicit and inject only the reference and not magically the requires.

lasote (approved):

Issue #8366, opened 2021-01-20
[question] How to build a binary package from reference with self.develop == True


ytimenkov:
This mostly used for tools packages (like CMake): they need certain settings (namely `compiler` and `build_type`) during the build but not when consuming.

Even looking at the CMakes recipe at CCI it adjusts `package_id` by removing `compiler` setting. However  this is not enough because keeping compiler in `settings` will require specifying it at consume time because otherwise running `conan install cmake/version` will fail if default profile doesn't have any `compiler`. (Otherwise it means "I want CMake build with compiler X", which is not really relevant).

In my case default profile doesn't have compiler because it is not present in the build image by default (and comes as a Conan package) and I prefer to avoid human mistakes when one setting is overridden by different profiles depending on the order on the command line.

I work this around by removing those settings in the `configure()` method:
```python
    def configure(self):
        if not self.develop:
            self.output.info("Consuming package, removing build-time settings")
            del self.settings.build_type
            del self.settings.compiler
```

But `self.develop` is defined only when using `conan create` command.

Building a reference is also widely shown in lockfiles examples in form `conan install libd/0.1@user/testing --build=libd/0.1@user/testing`.

I work this around by always running `conan create --not-export` on a path in the cache (this one returned from `conan info --paths`), but now it starts feeling too hackish (not to mention that requiring a valid set of settings to just get path to the export folder is an overkill).

Another option could be to allow `None` but then I would need to signal that I do need a compiler at build time (which is now handled by throwing `ConanInvalidConfiguration` in methods invoked by `conan info`).

jgsogo:
Hi! I think you found a corner-case that totally makes sense. I've create a PR with this scenario here https://github.com/conan-io/conan/compare/develop...jgsogo:issue/8366-compiler-pkgid?expand=1

Your question is more about a workaround (IMHO yours is probably better than the one I could figure out). As you said, the workflow using `conan install --build` is perfectly valid and useful in many scenarios, IMO we need to provide an answer to this failure.

Thanks for pointing it out.

ytimenkov:
I also thought that "create from reference" command could be a better place for `--lockfile-node-id` option (which as I understand also depends on `--build <ref>` option).

memsharded:
> Another option could be to allow None but then I would need to signal that I do need a compiler at build time (which is now handled by throwing ConanInvalidConfiguration in methods invoked by conan info).

I am considering this option like maybe the best approach, at least to become the default in Conan 2.0:
[...]

PR #10241, opened 2021-12-28
exploring new compatibility

memsharded:
Very preliminary idea of how the binary compatibility could look like


lasote (commented):
cool

ytimenkov:
Interesting approach. As I understand this file is a part of `conan install` bundle, right?

I was thinking if it makes sense to pass context (build or host) to the `compatibility` function? This way it would be possible to have more flexibility for build requirements (case like in #8366).

It wouldn't solve that issue completely, but could be something... There I would like to not define compiler at all for tools (build context) and then I need to know which compilers were used to build to add them into the compatibility least. This probably won't scale well if too many combinations are pushed.

memsharded:
> Interesting approach. As I understand this file is a part of conan install bundle, right?

Yes, the idea is that you can share it, update it, etc with ``conan config install too``.

> I was thinking if it makes sense to pass context (build or host) to the compatibility function? This way it would be possible to have more flexibility for build requirements (case like in #8366).

``conanfile`` already contains the ``context`` field, you can do ``conanfile.context`` inside ``compatibility()`` to get it. You should also get there ``conanfile.settings`` and ``conanfile.settings_build``.

> It wouldn't solve that issue completely, but could be something... There I would like to not define compiler at all for tools (build context) and then I need to know which compilers were used to build to add them into the compatibility least. This probably won't scale well if too many combinations are pushed.

You mean that the proposal doesn't solve the issue completely? I can see that, but at the moment seems like a good step forward. But I don't get the rest of your comment about compilers, but I suspect it is about the complexity of actually defining compatibility in the general case, which is something that I am concerned we cannot provide, and each user will still have to define their compatibility for their compilers and their needs.

ytimenkov:
> But I don't get the rest of your comment about compilers, but I suspect it is about the complexity of actually defining compatibility in the general case, which is something that I am concerned we cannot provide, and each user will still have to define their compatibility for their compilers and their needs.

By "complexity" I meant a computational complexity - a number of elements to go through: at least O(N*M) where N is a number of "compatible" packages, M - number of dependencies.

To put it simply: how to express: "I need package built with any compiler as long as it runs" (for example a flatc compiler or other tool)?

With "compatible packages" approach I need to return whole set of compilers were ever used: "compiler=gcc, version=6", "compuler=gcc, version=7" and so on. This list can grow long over time. If other settings are added (like OS version) we get Cartesian product.

This is because how compatible packages work (as I understand): Conan just goes over a list of hashes and tries to download.

Of course there is an option to define compatibility inside the recipe (e.g. as always, by deleting a setting), but this is a loss of information and I do feel that there is something in the out of recipe approach. Especially with recipes being described more semantically.

One option to overcome this explosion of packages to check would be to allow plugin to specify a query instead of a list. Like one used in `conan search --query`. Even with the current functionality a "any compiler" query can be constructed in a generic way by iterating over settings and just filtering out which are wildcards. (Unless querying is supported only in local cache...).

On top of this plugin can order returned list to select the best matching one.

> You mean that the proposal doesn't solve the issue completely? I can see that, but at the moment seems like a good step forward.

Yes, that is exactly what I tried to say 😊

memsharded:
> By "complexity" I meant a computational complexity - a number of elements to go through: at least O(N*M) where N is a number of "compatible" packages, M - number of dependencies.

Yes, this is true, but it is still polynomial, and I think doable. The main limit right now is architectural, not computational, and it is that the check against servers is done 1 package_id at a time. While this will certainly limit the number of compatibility checks that can be done in practice, it is something that we can iterate later (might need a v3 API with servers). But if both the functionality and the UX works well, it is something that probably is worth investing. The rest of the computation is just computing (for each package) N hashes for N package-ids (very fast), and finding the first existing in the existing ones (in the cache or in the server). Only once the ``package_id`` has been matched, the download will be done. So if we remove the "try to download one at a time", this is most likely doable fast enough without issues.

> Even with the current functionality a "any compiler" query can be constructed in a generic way by iterating over settings and just filtering out which are wildcards. (Unless querying is supported only in local cache...).

Yes, exactly, I think at this stage we want a useful mechanism to define any possible rule of compatibility. If this really catches up and becomes more widely used, we will learn about common "any compiler for C packages" rules and the like, and probably make them built-in. The idea is that the list returned by ``compatibility()`` is always ordered and prioritized exactly in that order, so users can define if they prefer to fallback to a "Release" package if they are in "Debug" mode, or they prefer to fallback to a "Debug" package built with the previous compiler version, for example.

czoido (approved):

franramirez688 (commented):
Looking good! 👍

lasote (commented):
I think we might want to have a "compatibility" folder in an "extensions" one, just in case we want to consider activating more than one "compatibility python file" in the future (maybe using a list in the conf, ordered). Might it have sense? I could think of a hypothetical "compatibility/intel.py" and "compatibility/sanitizers.py" and so on, so the compatible packages modules could append "configurations" to a list.

memsharded:
> I think we might want to have a "compatibility" folder in an "extensions" one, just in case we want to consider activating more than one "compatibility python file" in the future (maybe using a list in the conf, ordered). Might it have sense? I could think of a hypothetical "compatibility/intel.py" and "compatibility/sanitizers.py" and so on, so the compatible packages modules could append "configurations" to a list.

the problem with this is that still a priority among different files need to be defined. The final result should be a ordered list of combinations/variations to try. So if there are several files, there should be a meta-file that shows how to combine them. Maybe some kind of ``itertools.product`` in a given order.

conans/test/integration/package_id/test_cache_compatibles.py
```diff
@@ -0,0 +1,30 @@
...
+from conans.util.files import save
+
+
+class TestCacheCompatibles:
+
+    def test_compatible_build_type(self):
+        client = TestClient()
+        compatibles = textwrap.dedent("""\
+            def compatibility(conanfile):
+                result = []
```
memsharded, line 16: - Make sure that we can split the file in several and import them
- Make sure that we can use the recipe reference as data to compatibility
- Move it to extensions folder

memsharded: Done the 3 things.

Regarding the extensions folder it is in ``extensions/plugins/binary_compatibility.py``

conans/client/graph/graph_binaries.py
```diff
@@ -242,6 +221,33 @@ def _process_node(self, node, build_mode):
...
+            return
+
+        # Obtain the cache_latest valid one, cleaning things if dirty
+        while True:
+            cache_latest_prev = self._cache.get_latest_package_reference(node.pref)
+            if cache_latest_prev is None:
+                break
+            package_layout = self._cache.pkg_layout(cache_latest_prev)
+            if not self._evaluate_clean_pkg_folder_dirty(node, package_layout):
+                break
```
lasote: This "dirty" cleaning looks weird. Is this expected to have a dirty package? Why checking here? Are we checking the dirtyness everytime we access a package in the cache?

lasote (approved):

conans/client/graph/graph_binaries.py
memsharded: This is not really new, mostly refactored.
Yes, the checks need to be here, it is where you first check if a given package binary exists in the cache. If that package was dirty from a previous build or whatever, it needs to be cleared. A different story is that now with revisions it is less likely because they are built in a different place in the cache, then moved to the final folder.

conans/client/graph/compute_pid.py
```diff
@@ -78,3 +74,8 @@ def run_package_id(conanfile):
                     conanfile.info.invalid = BINARY_INVALID, str(e)
                 except ConanErrorConfiguration as e:
                     conanfile.info.invalid = BINARY_ERROR, str(e)
+
+    # Once we are done, call package_id() to narrow and change possible values
+    with conanfile_exception_formatter(conanfile, "package_id"):
```
memsharded, line 73: I have tested this order change (1.X package_id + validate, 2.0 validate + package_id) in Conan 1.X, it only breaks 1 test using ``compatible_packages``, which would be fixed if we backport the ``compatibility()`` method

conans/test/integration/package_id/test_validate.py
```diff
@@ -71,6 +69,7 @@ def package_id(self):
         client.assert_listed_binary({"pkg/0.1": ("cf2e4ff978548fafd099ad838f9ecb8858bf25cb",
                                                  "Invalid")})
         client.run("graph info --requires=pkg/0.1@ -s os=Windows")
+        print(client.out)
```
lasote: 👁️

conans/client/graph/compute_pid.py
```diff
@@ -63,11 +64,6 @@ def compute_package_id(node, new_config):
 
 
 def run_package_id(conanfile):
```
lasote: Rename to `run_package_id_validate`

lasote (approved):
Looking good! Maybe a test of a validate discarding a compatible package?

conans/client/graph/compute_pid.py
memsharded: Renamed

memsharded:
> Looking good! Maybe a test of a validate discarding a compatible package?

Added a couple of tests:
- One test with ``validate()`` rejecting invalid cppstd
- Another test with ``package_id()`` cleaning ``cppstd`` from the package_id

Still missing:
- Managing automatically when ``configure()`` removes settings or options

lasote (approved):

franramirez688 (approved):

memsharded:
Ok, lets move forward this for alpha.6. Still a couple of open items, but seems a good first step.

- Need to modernize ``Settings``, hopefully it can mimic a bit better what ``Options`` did, and not separate the ``values`` from it
- The current default proposed ``compatibility.py`` is just a POC, it only covers applications, and is not complete. It needs more work.
- Lets try to define ``self.settings`` and ``self.options`` in ``validate()`` and ``package_id()`` to be the ``info`` values. That would simplify recipes syntax and make it way more intuitive.
- Probably need helpers to query possible values in ``settings.yml`` for ``compatibility.py``
- It would be very useful if removed settings/options in ``configure()`` would also be filtered without needed to explicit the removal again in ``package_id()``.

I will also draft a proposal for the Tribe to discuss this.

PR #10244, opened 2021-12-29
Feature/develop2 cli create

memsharded:
Moving ``conan create`` to new CLI

- This is followed by another (relatively simple) PR to move the ``conan test`` command: https://github.com/conan-io/conan/pull/10254
- The vast majority of changes are just ``conan create . --name=pkg --version=0.1`` test changes
- I haven't managed to completely clean some of the legacy code, as it is still used by other commands (lock, test...) that haven't been updated yet.
- I haven't changed any behavior (with the options default scoping) yet. I think there are things that should be discussed (for example I would like to remove the ``self.develop`` and ``self.in_conan_cache``, they read an anti-pattern to me in 2.0


conans/cli/commands/create.py
```diff
@@ -0,0 +1,168 @@
...
+from conans.cli.output import ConanOutput
+from conans.client.conanfile.build import run_build_method
+from conans.client.graph.printer import print_graph
+from conans.errors import ConanException, conanfile_exception_formatter
+from conans.util.files import chdir, mkdir
+
+
+@conan_command(group=COMMAND_GROUPS['creator'])
+def create(conan_api, parser, *args):
+    """
```
memsharded, line 20: @lasote this is the main method to read. This PR is basically moving the legacy ``create`` here, using the new PythonAPI, and executing things a bit more "flat". This is a large command method, but I prefer it this way at the moment, being explicit.

memsharded (commented):
Didn't change the ``conan test`` command yet.

But already did the first initial refactor of loading of the "root node" for create (without removing code used by other places, but at least introducing the clean new code)

lasote (approved):

lasote (approved):

PR #10254, opened 2021-12-31
[develop2] New "test" command cli

memsharded:


conans/cli/commands/test.py
```diff
@@ -0,0 +1,107 @@
...
+        conanfile.folders.set_base_source(conanfile_folder)
+        conanfile.folders.set_base_package(conanfile_folder)
+        conanfile.folders.set_base_generators(conanfile_folder)
+        conanfile.folders.set_base_install(conanfile_folder)
+    else:
+        conanfile.folders.set_base_build(conanfile_folder)
+        conanfile.folders.set_base_source(conanfile_folder)
+        conanfile.folders.set_base_package(conanfile_folder)
+        conanfile.folders.set_base_generators(conanfile_folder)
+        conanfile.folders.set_base_install(conanfile_folder)
```
lasote: Empty diff?

lasote (approved):

conans/cli/commands/test.py
memsharded: This is because the command has removed at the moment the ``--test-build-folder`` argument, indeed. The idea is that maybe the standard "--build-folder`` would apply for this case if necessary. (The ``--test-build-folder`` would only be needed for ``conan create``, but we could use ``--build-folder`` as well?) 

I will remove the non-sense if(), and we will see how the general ``--build-folder`` moves forward, but better in a new thread?

PR #10255, opened 2021-12-31
new CLI for new command

memsharded:
- Updates to new CLI the ``new`` command.
- Greatly simplified it, now it is only based in jinja templates (also user custom ones)
- Adapted the ``cmake_lib`` and ``cmake_exe`` "conan new" templates
- The syntax is ``conan alias <template-name> --template_arg1=value1 --template_arg2=value2....``
- Removed the ``conan alias`` command. Replaced it with a ``conan new alias`` + ``conan export``, seemed having a command for this was too much.
- Added an exclusion mechanism to avoid rendering some files, like binaries (close https://github.com/conan-io/docs/issues/2165)

Close https://github.com/conan-io/docs/issues/2165


conans/cli/api/subapi/new.py
```diff
@@ -0,0 +1,78 @@
...
+    _NOT_TEMPLATES = "not_templates"
+
+    def __init__(self, conan_api):
+        self.conan_api = conan_api
+
+    @api_method
+    def new(self, template, definitions):
+        app = ConanApp(self.conan_api.cache_folder)
+
+        # Find the template
```
lasote: All these lines about finding the template shouldn't be here. API methods should be more "atomic" regarding the arguments. Probably this function should receive the list of file contents directly? Then have other/s to "scan" them (to the cache directory or the predefined one): `scan_cache()`, `get_preset_files()`, `scan_folder()`
Also, the _NOT_TEMPLATES doesn't sound good, (and not tested?), probably can be replaced with a more explicit argument with excluded patterns in the `scan_folder()`?
So, the `def new(self, template, definitions):` might be more like  `def render(self, files, definitions):` and then 
 it would be almost a loop of jinja renders.

lasote (changes requested):

conans/cli/api/subapi/new.py
memsharded: But we would still move the logic to another ``NewAPI`` method, right? It would be a bit challenging to implement the logic in the command side only, there is a lot of business logic in it. So I agree to split this in 2 methods: 1 for locating the files, and another for rendering.

The problem with ``not_templates`` is that those are files that should still be copied, they need to be loaded and saved, they are part of the template. Just they don't need to be passed through Jinja, that can crash or corrupt the file. I will add testing for it.

memsharded: There is already a test for ``not_templates``, it is called: ``test_template_image_files``. Test extracted from user issue: https://github.com/conan-io/docs/issues/2165

memsharded: Done, please check.

conans/cli/commands/new.py
```diff
@@ -0,0 +1,42 @@
...
+        if os.path.exists(path) and not args.force:
+            raise ConanException(f"File '{f}' already exists, and --force not defined, aborting")
+    # And respecting binary/text encodings
+    for f, v in sorted(files.items()):
+        path = os.path.join(cwd, f)
+        if isinstance(v, str):
+            save(path, v)
+        else:
+            open(path, "wb").write(v)
+        ConanOutput().success("File saved: %s" % f)
```
franramirez688: Should it be enough to instantiate one `ConanOutput` object outside the loop?

conans/test/utils/tools.py
```diff
@@ -610,6 +610,16 @@ def export(self, ref, conanfile=GenConanfile(), args=None):
         tmp.revision = rrev
         return tmp
 
+    def alias(self, source_target):
```
franramirez688: I think it would be easier to define two params:

```suggestion
    def alias(self, source, target):
    """A beautiful docstring here"""
```

Given the fact that it's a helper and it's doing more than a simple `self.run(f"new alias ...")`, I guess two params and a docstring would be fine.

conans/test/integration/command/new_test.py
```diff
@@ -1,265 +1,120 @@
...
-from conans.util.files import load
 
 
-class NewCommandTest(unittest.TestCase):
+class TestNewCommand:
+    def test_new_cmake_lib(self):
+        client = TestClient()
+        client.run("new cmake_lib --name=pkg --version=1.3")
+        conanfile = client.load("conanfile.py")
+        assert "CMakeToolchain" in conanfile
```
franramirez688, line 17: Perhaps, it could be safer to check the whole line `tc = CMakeToolchain(self)`

conans/cli/api/subapi/new.py
lasote: `get_template_files` is still doing too many things. I still think that there are three potential methods there: `scan_cache_templates(), get_preset_template_files(), get_template_files(FOLDER)`, or maybe only 2 if we provide the path to the cache folder templates: `get_preset_files(), get_template_files(FOLDER)` + `.template_cache_folder`

conans/cli/api/subapi/new.py
```diff
@@ -0,0 +1,81 @@
...
+        excluded = template_files.pop(self._NOT_TEMPLATES, None)
+        excluded = [] if not excluded else [s.strip() for s in excluded.splitlines() if s.strip()]
+
+        result = {}
+        definitions["conan_version"] = __version__
+        for k, v in template_files.items():
+            if not any(fnmatch.fnmatch(k, exclude) for exclude in excluded):
+                k = self._render_template(k, definitions)
+                v = self._render_template(v, definitions)
+            if v:
```
lasote: `k` and `v` are rendered? I'm missing something...

franramirez688 (approved):

conans/cli/api/subapi/new.py
```diff
@@ -0,0 +1,81 @@
...
+        if template_files is None:
+            template_files = new_templates.get(template)
+
+        if not template_files:
+            raise ConanException("Template doesn't exist or not a folder: {}".format(template))
+
+        return template_files
+
+    def render(self, template_files, definitions):
+        excluded = template_files.pop(self._NOT_TEMPLATES, None)
```
lasote: Checking here again the `_NOT_TEMPLATES` is a bit messy. Something is wrong.
What if these `get_xx_files` returns a tuple with (template_paths, files_paths) (instead of contents) and this `render()` only accept the template_paths? maybe the files_paths can just be copied.

conans/cli/api/subapi/new.py
memsharded: Yes. The filenames are templatized like ``src/{{name}}.h`` too.

conans/cli/api/subapi/new.py
memsharded: Yes, that makes sense. The problem is the asymmetry for built-in templates, so it is better if the method return the contents of the files already, but as long as the files are splitted into "files-to-render" and "files-to-not-render", that would be fine?

conans/test/integration/command/new_test.py
memsharded: This is just a very quick test to check the file is not empty. There are full tests building the result, this is not an issue.

conans/test/utils/tools.py
memsharded: Done

conans/cli/api/subapi/new.py
```diff
@@ -3,46 +3,46 @@
...
-            folder_template = template
-        else:
-            folder = os.path.join(app.cache.cache_folder, "templates", "command/new", template)
-            if os.path.exists(folder):
-                folder_template = folder
+    def get_template(self, template_name):
+        """ Load a template from a user absolute folder
+        """
+        if os.path.isdir(template_name):
+            return self._read_files(template_name)
```
lasote:
```suggestion
    def get_template(self, template_folder):
        """ Load a template from a user absolute folder
        """
        if os.path.isdir(template_folder):
            return self._read_files(template_folder)
```

conans/cli/api/subapi/new.py
```diff
@@ -3,46 +3,46 @@
...
+    @api_method
+    def get_home_template(self, template_name):
+        """ load a template from the Conan home templates/command/new folder
+        """
+        folder_template = os.path.join(self.conan_api.home_folder, "templates", "command/new",
+                                       template_name)
+        if os.path.isdir(folder_template):
+            return self._read_files(folder_template)
 
+    def _read_files(self, folder_template):
```
lasote:
```suggestion
        template_folder = os.path.join(self.conan_api.home_folder, "templates", "command/new",
                                       template_name)
        if os.path.isdir(template_folder):
            return self._read_files(template_folder)

    def _read_files(self, template_folder):
```

conans/cli/api/subapi/new.py
```diff
@@ -3,46 +3,46 @@
...
-        if os.path.isdir(template):
-            folder_template = template
-        else:
-            folder = os.path.join(app.cache.cache_folder, "templates", "command/new", template)
-            if os.path.exists(folder):
-                folder_template = folder
+    def get_template(self, template_name):
+        """ Load a template from a user absolute folder
+        """
+        if os.path.isdir(template_name):
```
lasote: What if it is not?

memsharded: It simply returns None

lasote (approved):

franramirez688 (approved):

czoido (approved):
Maybe I would add some comment code in the templates so that if someone does not know which parameters you can set, you find some information in the template. So for example if you make `conan new alias` without arguments you find some help in the generated file about which variables you have to define:

```
class AliasConanfile(ConanFile):
    # use -d name="<name>" to set the package name
    name = "<name>"
    # ...
    version = "<version>"
    # ...
    alias = "<name>/<target>"
    revision_mode = "hash"
```
or maybe adding a docstring to the class and outputting it if no parameters are passed to the template? I think that would be great but I'm ok with this being merged as is.

memsharded:
>  use -d name="<name>" to set the package name

I have tried this, as user, and found that when I open the conanfile and I see:
```python
class xxxPackageRecipe(ConanFile):
    # Use "conan new cmake_lib -d name=xxx -d version=yyy" to define name & version
    name = "xxx"
    version = "yyy"
```
And then it is a bit confusing, because it seems that now I should call ``conan new`` again? Then tried with docstring:

```python
class MyPkgRecipe(ConanFile):
    """ This template defines the following:
      -d name=<name> to define the name="<name>" attribute
      -d version=<version> to define the version="<version>" attribute
      -d class_name=<class_name>, to define the class name class <class_name>Recipe(ConanFile)
    Generated source files will be named after the "name" of the package, like <name>.cpp, <name>.h
    And the generated library will also be called "name(.lib|.a|...)"
    """
    name = "xxx"
    version = "yyy"
```
And it seemed a bit verbose, and will lose focus of the people learning, because it is a bunch of text to read, to just learn that they should use the ``-d name=<name>`` arguments, but if they are reading that, it is that they were already executing the tutorial, and in the vast majority of cases, they would have already run ``conan new cmake -d name=hello -d version=0.1``.

So I'd say that it is probably a bit more distracting than helpful? I am merging this as-is, but I have changes in a branch, so if still think it is worth, I'll submit those changes too

PR #10277, opened 2022-01-05
removed legacy apple tools (but new ones need to be modernized yet)

memsharded:
Basically, replacing the legacy ``tools.apple...`` with ``conan.tools.apple`` and some other minor changes:

- Tools not made public yet, the UIs need to be revised against real-life use cases and ConanCenter.
- Now, XCRun is going to be used only for testing purposes.
- Removing the ``sdk=None`` value for all the Apple OSs from ``settings.yml``. For instance:
   ```      
     Macos:
            version: [None, "10.6", "10.7", "10.8", "10.9", "10.10", "10.11", "10.12", "10.13", "10.14", "10.15", "11.0", "12.0", "13.0"]
        iOS:
            sdk: ["iphoneos", "iphonesimulator"]
    . . .
   ```
Now it must be explicitly written in the user profiles via ``os.sdk`` (except for ``Macos`` because it's automatically put as ``macosx``).
- Simplified and refactored functions like ``apple_min_version_flag()``.


conan/tools/apple/apple.py
```diff
@@ -18,15 +18,15 @@ def to_apple_arch(arch, default=None):
             'armv7k': 'armv7k'}.get(arch, default)
 
 
-def get_apple_sdk_name(conanfile):
+def get_apple_sdk_name(settings):
```
memsharded: If this is a tool that could be called from the outside, it better receive ``conanfile`` as argument, in case we need to inject some ``[conf]`` at some point. If it is private at the moment, good, it can use ``settings``, but lets recall not to make this public or document it.

memsharded (commented):
LGTM!

lasote (approved):

PR #10286, opened 2022-01-07
[develop2] new CLI and API for ``conan config``

memsharded:
New CLI and API for conan config


conans/cli/commands/config.py
```diff
@@ -0,0 +1,111 @@
...
+    Computes a dependency graph, without  installing or building the binaries
+    """
+
+
+def json_build_order(build_order):
+    return json.dumps(build_order, indent=4)
+
+
+@conan_subcommand()
+def config_remote_install(conan_api, parser, subparser, *args):
```
memsharded: The names ``conan config remote-install``, ``conan config remote-reinstall``, etc, are to be defined, I couldn't think of a better name, but I know we want a name to differentiate them from ``conan config list`` (that lists [conf]).

Thought about sub-sub-commands, but seemed too much (requires adding infra from our cli code)

conans/client/cache/cache.py
```diff
@@ -265,25 +263,13 @@ def generators(self):
         return generators
 
     def initialize_config(self):
+        # TODO: This is called by ConfigAPI.init(), maybe move everything there?
         if not os.path.exists(self.conan_conf_path):
-            save(self.conan_conf_path, normalize(get_default_client_conf()))
-
-    def reset_config(self):
-        if os.path.exists(self.conan_conf_path):
```
memsharded, line 272: This belongs to ``ConfigAPI``, not to ``Cache``.

I have left only the ``initialize_config()``, as too many tests and other behavior relies on it, it would be a big change (we can do it later if we think so)

conans/cli/commands/config.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    home = conan_api.config.home()
+    ConanOutput().info(f"Current Conan home: {home}")
+    return home
+
+
+@conan_subcommand()
+def config_init(conan_api, parser, subparser, *args):
+    """
+    Initialize Conan home configuration: settings, conf and remotes
```
lasote: I see this as a `reset` more than an `init`.

conans/cli/commands/config.py
```diff
@@ -0,0 +1,111 @@
...
+    subparser.add_argument("item", type=int,
+                           help='Remove configuration origin by index in list (index '
+                                'provided by --list argument)')
+    args = parser.parse_args(*args)
+    conan_api.config.remote_remove(index=args.item)
+    ConanOutput().success("Removed remote-install configuration")
+
+
+@conan_subcommand(formatters={"json": format_remote_install_configs})
+def config_list(conan_api, parser, subparser, *args):
```
lasote: I don't think I would search the list of the available configs in a command. For me this is documentation.

lasote (changes requested):
I still think this command is quite confusing. What about:

```
    home                Gets the Conan home folder
    reset               Reset Conan home configuration: settings, conf, and remotes.
    install             Installs the configuration from the configured origins
    origin-add          Adds an origin to get configuration (remotes, profiles, conf) from git, HTTP or folder
    origin-remove       Remove the origin from the list
    origin-list         Returns the defined origins of configuration
```

That would require a two step command to install configuration (origin-add + install) but `reinstall` become unnecessary and it feels more natural.

conans/cli/commands/config.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    configs = conan_api.config.remote_list()
+    for index, remote_config in enumerate(configs):
+        ConanOutput().info("%s: %s" % (index, remote_config))
+
+
+@conan_subcommand()
+def config_remote_remove(conan_api, parser, subparser, *args):
+    """
+    Returns the defined origins of configuration
```
lasote: Change the docstring.

conans/cli/commands/config.py
memsharded: It was originally called "init", because the actual use case is users that want to have the settings.yml, conan.conf in the Conan home, to read them or do something with them, in CI, after installing Conan, so does files don't exist at all. The command main use was to force the creation and initialization, and then a secondary use case was to reset them to defaults.

conans/cli/commands/config.py
memsharded: Documentation explaining the internal format of the Conan home "origins" json file, so they can read it and deduce the configuration index order they want to remove? One of the main uses of this command is to know which integer index you need to pass to "remove" subcommand.

memsharded:
> That would require a two step command to install configuration (origin-add + install) but reinstall become unnecessary and it feels more natural.

I don't dislike the idea of decoupling the definition from the execution, but I see a couple of edges:

- The ``config install`` command, always reinstall all the previous configuration origins too? It is not possible to execute just 1 "origin" (specially if some previous origin fails, because it is not reachable or not exist anymore)?
- UX wide, it will be a bit verbose for company users, which more common use case is to bundle everything in 1 config install, so not having a command to do it in a single step... I can see the feature requests coming asking for it.

conans/cli/commands/config.py
```diff
@@ -0,0 +1,111 @@
...
+from conans.cli.command import conan_command, COMMAND_GROUPS, conan_subcommand
+from conans.cli.output import ConanOutput
+from conans.model.conf import DEFAULT_CONFIGURATION
+from conans.util.config_parser import get_bool_from_text
+
+
+@conan_command(group=COMMAND_GROUPS['consumer'])
+def config(conan_api, parser, *args):
+    """
+    Computes a dependency graph, without  installing or building the binaries
```
czoido: This sneaked in from `conan graph`

conans/cli/commands/config.py
czoido: If we go forward with this, as `remote-install` also manages local files, maybe it should be just `install` ?

czoido:
I also think that there's something not natural with this command. IMHO I would move the config origins outside of conan and let the users manage that... if that's too radical maybe we could split functionalities, one for managing files in the cache and other for managing origins. I'm afraid also we could end up with tons of commands in Conan 2.0, maybe we can discuss this further and come up with more ideas.

memsharded:
> I also think that there's something not natural with this command. IMHO I would move the config origins outside of conan and let the users manage that... if that's too radical maybe we could split functionalities, one for managing files in the cache and other for managing origins. I'm afraid also we could end up with tons of commands in Conan 2.0, maybe we can discuss this further and come up with more ideas.

the main problem is the (relatively very demanded and used) scheduled install, in which the automatic call of re-installing from defined origins happen every X time (as defined by users in their conf). If Conan doesn't manage built-in the list of origins, this feature is not possible at all.

We could be tempted to move this to another "command layer", and leave the core Conan without this. But specifically the auto-scheduled ``conan config install`` cannot happen as a custom command, it needs to be implemented built-in. At the very least, we should add a new "init-conan" hook or something that allows to run things when Conan loads, but that will not be very elegant either.

czoido (approved):

lasote (approved):

PR #10292, opened 2022-01-10
fix CMakeToolchain warnings and werror

### LINKED ISSUE #10288: [bug] Generated CMake files cause errors with `-Werror=dev --warn-uninitialized`
Conan generated CMake files cause errors when the project is configured with `-Werror=dev --warn-uninitialized`.

  * Compiler+version: VS2019
  * Conan version: 1.44.0
  * Python version: 3.10.0

`find_package()` a dependency after installing it with `CMakeToolchain` and `cmake_find_package_multi` generators.

Example log attached for when using the toolchain to `find_package(freetype)`. It's just a lot of uninitialized variable use.

[log.log](https://github.com/conan-io/conan/files/7833584/log.log)

memsharded:
Changelog: Fix: Fix the ``CMakeToolchain`` generated code, so it doesnt fail for ``-Werror --warn-unitilized``. 
Docs: Omit

Close https://github.com/conan-io/conan/issues/10288

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
@@ -190,17 +190,22 @@ def test_cmaketoolchain_no_warnings():
         from conans import ConanFile
         class Conan(ConanFile):
             settings = "os", "compiler", "arch", "build_type"
-            generators = "CMakeToolchain"
+            generators = "CMakeToolchain", "CMakeDeps"
+            requires = "dep/0.1"
         """)
     consumer = textwrap.dedent("""
        set(CMAKE_CXX_COMPILER_WORKS 1)
        set(CMAKE_CXX_ABI_COMPILED 1)
        project(MyHello CXX)
        cmake_minimum_required(VERSION 3.15)
+
+       find_package(dep CONFIG REQUIRED)
        """)
-    client.save({"conanfile.py": conanfile,
+    client.save({"dep/conanfile.py": GenConanfile("dep", "0.1"),
+                 "conanfile.py": conanfile,
                  "CMakeLists.txt": consumer})
 
+    client.run("create dep")
     client.run("install .")
     client.run_command("cmake . -DCMAKE_TOOLCHAIN_FILE=./conan_toolchain.cmake "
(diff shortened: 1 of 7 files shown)


memsharded:
After user @friendlyanon feedback, I have added checks for CMakeDeps too.

But now I find errors with ``dep_COMPONENT_NAMES`` and ``dep_FIND_DEPENDENCY_NAMES``, which requires some extra investigation, because I don't get it yet why is it complaining. 

Update: managed to solve the issue. The thing is that ``list(APPEND)`` will not initialize the variable if not appending something, but it will create and initialize de variable if previously uninitialized.

czoido (approved):

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -177,3 +177,37 @@ def build(self):
...
+        class Conan(ConanFile):
+            settings = "os", "compiler", "arch", "build_type"
+            generators = "CMakeToolchain", "CMakeDeps"
+            requires = "dep/0.1"
+        """)
+    consumer = textwrap.dedent("""
+       set(CMAKE_CXX_COMPILER_WORKS 1)
+       set(CMAKE_CXX_ABI_COMPILED 1)
+       project(MyHello CXX)
+       cmake_minimum_required(VERSION 3.15)
```
friendlyanon, line 200: The order of command calls isn't right here. `cmake_minimum_required` also sets up policies according to the provided version number. The order here results in running the toolchain with no policies. The toolchain is run via `project` while enabling a language for the first time. This should be:

```cmake
set(CMAKE_CXX_COMPILER_WORKS 1)
set(CMAKE_CXX_ABI_COMPILED 1)
project(MyHello CXX)
```

memsharded: Good to know, but not critical for this PR (it has already been merged), right? The test failed without the fixes, then passed with them. It it jsut an optimization to try to save some seconds of CI time.

friendlyanon: It has been merged, yes. Just letting you know that `cmake_minimum_required` should really be the very first command call in an entrypoint.

PR #10300, opened 2022-01-10
[feature] Mandatory `os.sdk` for CMakeToolchain (only `AppleSystemBlock` block)

### LINKED ISSUE #10275: [bug][2.0] CMakeToolchain does not respect os.sdk when targeting armv8 iOS simulator
Hello,

I'm trying to cross-build some iOS recipes, and I think there's a bug in CMakeToolchain. I use this profile:

```ini
[settings]
os=iOS
os.version=9.0
os.sdk=iphonesimulator
build_type=Release
arch=armv8
compiler=apple-clang
compiler.version=13.0
compiler.libcxx=libc++
compiler.cppstd=17
```

The `test_package` step fails with the following error:

```
ld: in /Users/theo/.conan2/p/f92ba1e43ba51502/p/lib/libsodium.a(libsodium_la-core.o), building for iOS, but linking in object file built for iOS Simulator, for architecture arm64
```

The generated `conan_toolchain.cmake` contains the following line:

>set(CMAKE_OSX_SYSROOT iphoneos CACHE STRING "" FORCE)

This does not happen when setting `arch=x86_64`, `CMAKE_OSX_SYSROOT` being correctly set to `iphonesimulator` in this case.

franramirez688:
Changelog: Feature: `os.sdk` field is mandatory for CMakeToolchain and OS in `('Macos', 'iOS', 'watchOS', 'tvOS')`.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/10275

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -350,25 +350,21 @@ class AppleSystemBlock(Block):
                 "armv8": "arm64",
                 "armv8_32": "arm64_32"}.get(arch, arch)
 
-    # TODO: refactor, comes from conans.client.tools.apple.py
     def _apple_sdk_name(self):
-        """returns proper SDK name suitable for OS and architecture
-        we're building for (considering simulators)"""
-        arch = self._conanfile.settings.get_safe('arch')
+        """
+        Returns the 'os.sdk' (SDK name) field value. Every user should specify it because
+        there could be several ones depending on the OS architecture.
+
+        Note: In case of MacOS it'll be the same for all the architectures.
+        """
         os_ = self._conanfile.settings.get_safe('os')
         os_sdk = self._conanfile.settings.get_safe('os.sdk')
         if os_sdk:
             return os_sdk
-        elif arch.startswith('x86'):
-            return {'Macos': 'macosx',
-                    'iOS': 'iphonesimulator',
-                    'watchOS': 'watchsimulator',
-                    'tvOS': 'appletvsimulator'}.get(os_)
+        elif os_ == "Macos":  # it has only a single value for all the architectures for now
+            return "macosx"
         else:
-            return {'Macos': 'macosx',
-                    'iOS': 'iphoneos',
-                    'watchOS': 'watchos',
-                    'tvOS': 'appletvos'}.get(os_)
+            raise ConanException("Please, specify a suitable value for os.sdk.")
 
     def context(self):
         os_ = self._conanfile.settings.get_safe("os")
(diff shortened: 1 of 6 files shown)


conan/tools/cmake/toolchain.py
```diff
@@ -356,7 +356,10 @@ def _apple_sdk_name(self):
         we're building for (considering simulators)"""
         arch = self._conanfile.settings.get_safe('arch')
         os_ = self._conanfile.settings.get_safe('os')
-        if arch.startswith('x86'):
+        os_sdk = self._conanfile.settings.get_safe('os.sdk')
+        if os_sdk:
+            return os_sdk
+        elif arch.startswith('x86'):
```
memsharded: Better force the definition of ``os.sdk``?

The thing is that we don't want to auto-deduce it is a simulator. The binary is different, it should have a different package_id, and for that it should have the setting properly defined. It might not be robust relying on the architecture to deduce this. 

It might be necessary to implement the logic of ``Macos`` => ``macosx`` as there is no other sdk for it, but for iOS, watchOS, tvOS, it should be mandatory to be defined in settings.

franramirez688: Yeah, you're right but it sounds like a breaking change, doesn't it? Should we make that change in another PR?

franramirez688: Or could we even introduce that change for Conan 2.x?

memsharded: This is ``CMakeToolchain``, which is experimental. It might be fine to make it only work with the 2.0-ready behavior, and fail otherwise, even if it is "breaking" (I don't think it is a huge issue, as it boils down to add the sdk in your profiles or command line, not needing to modify recipes, and it is not that CMakeToolchain for iOS is that extended yet either.

franramirez688: Definitely, yes, let's do that for all the cases 1.x and 2.x.

memsharded (approved):

memsharded (approved):

conan/tools/cmake/toolchain.py
```diff
@@ -350,22 +350,21 @@ def _get_architecture(self):
...
+        """
         os_ = self._conanfile.settings.get_safe('os')
-        if arch.startswith('x86'):
-            return {'Macos': 'macosx',
-                    'iOS': 'iphonesimulator',
-                    'watchOS': 'watchsimulator',
-                    'tvOS': 'appletvsimulator'}.get(os_)
+        os_sdk = self._conanfile.settings.get_safe('os.sdk')
+        if os_sdk:
+            return os_sdk
```
memsharded, line 363: this is functionally correct, but I am missing the path forward 2.0. If ``os.sdk`` will not exist for ``os=Macos`` (because it always uses the macosx sdk), then, shouldn't we make sure that the user is NOT defining the os.sdk to be ready?

PR #10310, opened 2022-01-11
Validate if settings.compiler is related to Visual Studio

### LINKED ISSUE #10306: [feature] Tool: is_msvc helper
Our friend and great contributor SpaceIm is using a new detection which I believe could be part of the mainstream:

```python
@property
def _is_msvc(self):
    return str(self.settings.compiler) in ["Visual Studio", "msvc"]
```

This property can be largely re-used, when checking on `validate()` or any other condition.

uilianries:
Add a new helper to simplify and avoid method copies in Conan Center Index

Changelog: Feature: Add `is_msvc` to validate if `settings.compiler` is `Visual Studio` and `msvc` compilers.
Docs: https://github.com/conan-io/docs/pull/2353

closes #10306

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -172,3 +172,15 @@ def _vcvars_vers(conanfile, compiler, vs_version):
         # The equivalent of compiler 192 is toolset 14.2
         vcvars_ver = "14.{}".format(compiler_version[-1])
     return vcvars_ver
+
+
+def is_msvc(conanfile):
+    """ Validate if current compiler is 'Visual Studio' or 'msvc'
+    :param conanfile: ConanFile instance
+    :return: True, if the host compiler is related to Visual Studio, otherwise, False.
+    """
+    settings = conanfile.settings
+    compiler = settings.get_safe("compiler")
+    if not compiler:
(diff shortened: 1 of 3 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -172,3 +172,15 @@ def _vcvars_vers(conanfile, compiler, vs_version):
...
+def is_msvc(conanfile):
+    """ Validate if current compiler is 'Visual Studio' or 'msvc'
+    :param conanfile: ConanFile instance
+    :return: True, if the host compiler is related to Visual Studio, otherwise, False.
+    """
+    settings = conanfile.settings
+    compiler = settings.get_safe("compiler")
+    if not compiler:
+        raise ConanException("settings.compiler is mandatory to check Visual Studio relation.")
+    return str(compiler) in ["Visual Studio", "msvc"]
```
memsharded: In theory this would always be already a ``str()``, no need to force it.

memsharded (approved):
Looks good.

Just a clarification, this is not a "detection" of compiler, just a check. We use the term "detect" to "guessing" the compiler (like in the profile --detect)

conan/tools/microsoft/visual.py
uilianries: Let me fix it, less code is better.

PR #10315, opened 2022-01-11
Add VxWorks to os in default settings.yml. (#10313)

### LINKED ISSUE #10313: [feature] Add VxWorks to os in default settings.yml.
I would be nice to have proper support for VxWorks (at least version 7).
Currently VxWorks (SR 21.11) supports gcc, llvm. So all compilers are currently already mentioned in the settings.yml.

More details can be found here:
https://www.windriver.com/themes/Windriver/pdf/vxworks-7-datasheet.pdf

wackinger:
Closes https://github.com/conan-io/conan/issues/10313

Changelog: Feature: Add VxWorks to OSs in default settings.yml.
Docs: https://github.com/conan-io/docs/pull/2355

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
@@ -1,0 +1,62 @@
+import platform
+import textwrap
+
+import pytest
+
+from conans.test.assets.cmake import gen_cmakelists
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.utils.tools import TestClient
+from conans.util.files import save
+
+
+@pytest.fixture
+def client():
+    c = TestClient()
+    save(c.cache.new_config_path, "tools.env.virtualenv:auto_use=True")
+    clang_profile = textwrap.dedent("""
+        [settings]
+        arch=armv7
+        build_type=RelWithDebInfo
+        compiler=clang
+        compiler.libcxx=libstdc++11
+        compiler.version=12
+        os=VxWorks
+        os.version=7
+
+        [buildenv]
+        CC=clang
+        CXX=clang++
+        """)
+    conanfile = textwrap.dedent("""
+        import os
+        from conans import ConanFile
+        from conan.tools.cmake import CMake
+        from conan.tools.layout import cmake_layout
+        class Pkg(ConanFile):
+            settings = "os", "compiler", "build_type", "arch"
+            exports_sources = "*"
+            generators = "CMakeToolchain"
+
+            def layout(self):
+                cmake_layout(self)
+
+            def build(self):
+                cmake = CMake(self)
+                cmake.configure()
+                cmake.build()
+                cmd = os.path.join(self.cpp.build.bindirs[0], "my_app")
+                self.run(cmd, env=["conanrunenv"])
+        """)
+    c.save({"conanfile.py": conanfile,
+            "clang": clang_profile,
+            "CMakeLists.txt": gen_cmakelists(appname="my_app", appsources=["src/main.cpp"]),
+            "src/main.cpp": gen_function_cpp(name="main")})
+    return c
+
+
+@pytest.mark.tool_cmake
+@pytest.mark.tool_clang(version="12")
+def test_clang_cmake_ninja(client):
+    client.run("create . pkg/0.1@ -pr=clang -c tools.cmake.cmaketoolchain:generator=Ninja")
+    assert 'cmake -G "Ninja"' in client.out
+    assert "main __clang_major__12" in client.out
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
@@ -45,10 +45,76 @@ def client():
                 cmake.configure()
                 cmake.build()
                 cmd = os.path.join(self.cpp.build.bindirs[0], "my_app")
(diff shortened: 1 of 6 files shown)


memsharded (commented):
This is fine, adding VxWorks make sense.

There is still a couple of tests broken, because they list the available OSs in settings.yml, and that output changed now. Please see if you can fix them, otherwise we will help.

What would be very nice to have in the codebase is a full functional test that does some build of a package to VxWorks.  Ideally with the most mainstream tooling (if building for VxWorks is commonly done with CMake, then using CMake). This test can be excluded from our CI with ``pytest.mark.tool_vxworks`` (check the ``conftest.py`` file in the ``test`` folder), but it should work locally when the vxworks tools are installed. This will serve as a documentation and also utility to debug and support users using VxWorks. Not necessary to add it in this PR, but if there is experience out there using Conan for VxWorks, it would be great to incorporate at least the basic knowledge in the codebase to be future-proof.

memsharded (approved):
I am going to approve this atm, and will comment on the docs PR.

I still would like to really test the things that are stated in the docs PR. It should be possible to write a unit-test that covers using ``os=VxWorks`` (and the necessary toolchain, via profiles, etc), that really builds and check the binaries, and that any user with the toolchain installed could execute. The ``conftest.py`` file can explain the configuration of the toolchain. We have struggled in the past to maintain insufficiently tested integrations, and having these tests would really help.

So if you please would consider adding this in the future, it would be fantastic, open an issue to discuss, and we will give guidelines.

wackinger:
Great to hear. Thanks.
I'm still trying to find my way though the tests to find the right location to add a proper test.

wackinger:
I'd love to provide sufficient tests, but am not familiar with the conan code base, yet. Could you perhaps give me a little hint on where to start to implement proper tests and what the proper preconditions would be? A pointer to an existing "exotic" toolchain/os test would be a great start. Thanks!

memsharded:
Sure, I think the best would be something like this, testing clang in Windows: https://github.com/conan-io/conan/blob/develop/conans/test/functional/toolchains/cmake/test_cmake_toolchain_win_clang.py

- Very simple conanfile, building one executable
- The ``@pytest.mark.tool_clang(version="12")`` is what inhibits our CI to fail, as clang 12 not installed there. But configuring my ``conftest_user.py`` file, I can activate it
- As the ``self.run(cmd, env=["conanrunenv"])`` will not be possible, it can be replaced with a simple check that the binary file exists (os.path.exists), or if there is some possibility to check the binary architecture with ``objdump`` or equivalent, that would be even better.
- No need to do several tests 1 single test for 1 configuration is ok.
- Depending on the build system, put in one place or another, under ``conans/test/functional/toolchains`` (dont know if vxworks uses always makefiles)

conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
```diff
@@ -0,0 +1,62 @@
...
+            "src/main.cpp": gen_function_cpp(name="main")})
+    return c
+
+
+@pytest.mark.tool_cmake
+@pytest.mark.tool_clang(version="12")
+def test_clang_cmake_ninja(client):
+    client.run("create . pkg/0.1@ -pr=clang -c tools.cmake.cmaketoolchain:generator=Ninja")
+    assert 'cmake -G "Ninja"' in client.out
+    assert "main __clang_major__12" in client.out
```
memsharded: In theory the binary created will not run in the current platform, is not executable, this shouldn't be possible.

conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
```diff
@@ -0,0 +1,62 @@
...
+
+            def layout(self):
+                cmake_layout(self)
+
+            def build(self):
+                cmake = CMake(self)
+                cmake.configure()
+                cmake.build()
+                cmd = os.path.join(self.cpp.build.bindirs[0], "my_app")
+                self.run(cmd, env=["conanrunenv"])
```
memsharded: This self.run() should fail if the binary was actually cross-built for vxwork os. It should be replaced for some check of the binary.

conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
```diff
@@ -0,0 +1,62 @@
...
+
+@pytest.fixture
+def client():
+    c = TestClient()
+    save(c.cache.new_config_path, "tools.env.virtualenv:auto_use=True")
+    clang_profile = textwrap.dedent("""
+        [settings]
+        arch=armv7
+        build_type=RelWithDebInfo
+        compiler=clang
```
memsharded, line 20: Is ``clang`` the compiler used to cross-compile to ``vxworks``?

For cross compiling, it is typically necessary to provide a ``CC=full/path/to/cross/compiler`` env-var

memsharded (commented):
It seems the test is doing just a normal native "clang" compilation, not cross-compiling to vxworks.

conans/test/functional/toolchains/cmake/test_cmake_toolchain_vxworks_clang.py
wackinger: clang and ldarm are available in the path in my docker container.

memsharded: Ok, good, I was mostly wondering if ``clang`` is the compiler actually used to compile to vxworks, and no other compiler should go here.

czoido (approved):

Issue #10282, opened 2022-01-06
[question] cmake_build_modules behavior in conan 1.43.0 and conan 1.43.2


rajatgirotra:
I am using conan 1.43.2
I have built conan package for google protobuf for my platform using this recipe in conan central.
https://github.com/conan-io/conan-center-index/blob/master/recipes/protobuf/all/conanfile.py
the build recipe sets "cmake_build_modules" in package_info().
`self.cpp_info.components["libprotobuf"].set_property("cmake_build_modules", build_modules)`

// create command
`bash$> conan create .  x.y.z@myuser/mychannel -pr debug64 -pr:b debug64`

In my consumer conanfile.py, I am using CMakeToolchain and CMakeDeps generators

```
class MyProjetConan(ConanFile):
    ....
    requires = "protobuf/x.y.z@myuser/mychannel"

    def generate(self):
        tc = CMakeToolchain(self)
        tc.variables["CMAKE_VERBOSE_MAKEFILE"] = True
        tc.generate()
        deps = CMakeDeps(self)
        deps.generate()

```
I am consuming like this
`bash$ /myproject_root/build/> conan install .. -pr debug64 -pr:b debug64`

However, the build modules (*.cmake files) aren't set in the generated cmake files with conan 1.43.2
```
bash$ /myproject_root/build/> grep protobuf_BUILD_MODULES_PATHS_DEBUG protobuf*
protobuf-debug-x86_64-data.cmake:set(protobuf_BUILD_MODULES_PATHS_DEBUG )  # Not Set!!
```

If I downgrade to conan 1.43.0, then it works.
```
bash$ /myproject_root/build/> conan install .. -pr debug64 -pr:b debug64  # with conan 1.43.0
bash$ /myproject_root/build/> grep protobuf_BUILD_MODULES_PATHS_DEBUG protobuf*
protobuf-debug-x86_64-data.cmake:set(protobuf_BUILD_MODULES_PATHS_DEBUG "${protobuf_PACKAGE_FOLDER_DEBUG/lib/cmake/protobuf/protobuf-generate.cmake" .... )
```

Is it a bug or am I using host and build profiles incorrectly? I am building and consuming on the same physical host.

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Hi @rajatgirotra 

[...]

PR #10326, opened 2022-01-12
CMakeDeps and cmake_build_modules property from components

lasote:
Changelog: Bugfix: The `CMakeDeps` generator now uses the property `cmake_build_modules` declared in components of the required packages not only in the root cpp_info.
Docs: omit

Closes https://github.com/conan-io/conan/issues/10282

### DIFF
--- a/conans/model/new_build_info.py
+++ b/conans/model/new_build_info.py
@@ -182,6 +182,8 @@ class NewCppInfo(object):
                         current_values = result.get_init("requires", [])
                         current_values.extend(component.requires)
 
+                # We copy the properties from the root object, even if we have components
+                result._generator_properties = copy.copy(self._generator_properties)
                 # FIXME: What to do about sysroot?
             else:
                 result = copy.copy(self.components[None])
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_build_modules.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_build_modules.py
@@ -1,272 +1,160 @@
-import os
-import platform
 import textwrap
-import unittest
 
 import pytest
 
-from conans.model.ref import ConanFileReference, PackageReference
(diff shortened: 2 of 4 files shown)


conans/model/new_build_info.py
```diff
@@ -182,6 +182,8 @@ def aggregated_components(self):
                         current_values = result.get_init("requires", [])
                         current_values.extend(component.requires)
 
+                # We copy the properties from the root object, even if we have components
+                result._generator_properties = copy.copy(self._generator_properties)
```
lasote, line 186: Could this be breaking? I think it shouldn't, the aggregated_components is used only by the new generators, and it makes sense that aggregating the components in a single cppinfo, keeps the root properties. The aggregation cannot be done for the component's properties because they cannot be aggregated but keeping the root one make sense. right?

memsharded: Yes, I think it makes sense, and it should be safe, cause only used in new integrations.
So it seems it is possible to define both root properties and component properties, but it is not possible to define root xxxdirs and component xxxdirs, right?

memsharded (approved):

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_build_modules.py
```diff
@@ -1,272 +1,166 @@
...
-            import os
-            from conans import ConanFile
+            def package_info(self):
+                module = os.path.join("share", "cmake", "target-alias.cmake")
+                {}
+        """)
+    if use_components:
+        info = """
+        self.cpp_info.name = "namespace"
+        self.cpp_info.filenames["CMakeDeps"] = "hello"
```
czoido:
```suggestion
```

czoido (approved):

conans/model/new_build_info.py
lasote: yes.

PR #9670, opened 2021-09-25
Typo in toolchain cppstd message

puetzk:
Remove a stray extra `}` from the message "Conan C++ Standard 14 with extensions OFF}"

Changelog: Omit
Docs: Omit

- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.

PR #10337, opened 2022-01-13
Fix cppstd info message

jkleve:
Changelog: Fix: Fix extra `}` characters in cppstd info message.
Docs: omit

The message that indicates the c++ standard was modified has one too many '}' characters giving the text 

```sh
-- Conan toolchain: C++ Standard 17 with extensions OFF}
```

No related issue that I saw.
Related PR: #9670


memsharded:
Thanks for fixing this!

PR #10339, opened 2022-01-13
fix -debug in AutotoolsToolchain

### LINKED ISSUE #10338: [bug] AutotoolsToolchain doesn't pass -debug to msvc linker
I was building a package with `-s build_type=Debug` and tried to debug the generated executable but the debugger complained that "Module was built without symbols". 
After a short investigation it seemed that `-debug` flag should have been added to `LDFLAGS`.
Adding the following to the recipe enabled me to debug the executable:

```
tc = AutotoolsToolchain(self)
if self._is_msvc and self.settings.build_type in ["Debug", "RelWithDebInfo"]:
    tc.ldflags.append("-debug")
tc.generate()
```

I tried to look at the code of the legacy `AutotoolsEnvironment` build helper and couldn't see it passing this flag either but I also couldn't find any issue reporting that so far so I'm not sure if it's broken as well.

  * Compiler+version: Visual Studio 16
  * Conan version: 1.44.0

This is part of the generated conanautotoolstoolchain.bat file before the workaround
```
set CPPFLAGS=%CPPFLAGS%
set CXXFLAGS=%CXXFLAGS% -Zi -Ob0 -Od
set CFLAGS=%CFLAGS% -Zi -Ob0 -Od
set LDFLAGS=%LDFLAGS%
```

and after
```
set CPPFLAGS=%CPPFLAGS%
set CXXFLAGS=%CXXFLAGS% -Zi -Ob0 -Od
set CFLAGS=%CFLAGS% -Zi -Ob0 -Od
set LDFLAGS=%LDFLAGS% -debug
```

memsharded:
Changelog: Fix: Add ``-debug`` to `LDFLAGS` in ``AutotoolsToolchain`` when necessary. 
Docs: Omit

Fix https://github.com/conan-io/conan/issues/10338

### DIFF
--- a/conan/tools/_compilers.py
+++ b/conan/tools/_compilers.py
@@ -56,6 +56,25 @@ def architecture_flag(settings):
     return ""
 
 
+def build_type_link_flags(settings):
+    """
+    returns link flags specific to the build type (Debug, Release, etc.)
+    [-debug]
+    """
+    compiler = settings.get_safe("compiler")
+    build_type = settings.get_safe("build_type")
+    if not compiler or not build_type:
+        return []
+
+    # https://github.com/Kitware/CMake/blob/d7af8a34b67026feaee558433db3a835d6007e06/
+    # Modules/Platform/Windows-MSVC.cmake
+    if compiler in ["clang", "gcc", "apple-clang", "qcc", "mcst-lcc"]:
+        if build_type in ("Debug", "RelWithDebInfo"):
+            return ["-debug"]
+
+    return []
+
+
 def build_type_flags(settings):
     """
     returns flags specific to the build type (Debug, Release, etc.)
@@ -65,7 +84,7 @@ def build_type_flags(settings):
     build_type = settings.get_safe("build_type")
     vs_toolset = settings.get_safe("compiler.toolset")
     if not compiler or not build_type:
-        return ""
+        return []
 
     # https://github.com/Kitware/CMake/blob/d7af8a34b67026feaee558433db3a835d6007e06/
     # Modules/Platform/Windows-MSVC.cmake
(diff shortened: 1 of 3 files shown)


conan/tools/_compilers.py
```diff
@@ -56,6 +56,25 @@ def architecture_flag(settings):
...
+    if not compiler or not build_type:
+        return []
+
+    # https://github.com/Kitware/CMake/blob/d7af8a34b67026feaee558433db3a835d6007e06/
+    # Modules/Platform/Windows-MSVC.cmake
+    if compiler in ["msvc", "Visual Studio"]:
+        if build_type in ("Debug", "RelWithDebInfo"):
+            return ["-debug"]
+
+    return []
```
franramirez688, line 75: Perhaps, we could keep it as simple like this one for now:

```python
def build_type_link_flags(settings):
    """
    returns link flags specific to the build type (Debug, Release, etc.)
    [-debug]
    """
    compiler = settings.get_safe("compiler")
    build_type = settings.get_safe("build_type")
    # https://github.com/Kitware/CMake/blob/d7af8a34b67026feaee558433db3a835d6007e06/
    # Modules/Platform/Windows-MSVC.cmake
    if compiler in ["msvc", "Visual Studio"] and build_type in ("Debug", "RelWithDebInfo"):
        return ["-debug"]
    else:
        return []
```

franramirez688 (approved):

conan/tools/_compilers.py
memsharded: Yes, could be. Current structure is parallel to the other "flags" methods, I mostly copied ``build_type_flags()``, probably better to keep the same method structure? Even if this is a bit simpler now, it might get other compiler flags or something like that?

PR #10363, opened 2022-01-18
remove get and inspect commands

memsharded:
Remove both the ``conan get`` and ``conan inspect`` commands in 2.0.

It is likely that we will need some of its functionality, but better re-think from scratch, and with fresh feedback, because they are kind of problematic (like inspect changing the current cache state)


franramirez688 (approved):

franramirez688:
@memsharded Please, solve the conflicts to merge the PR.

PR #10380, opened 2022-01-19
Redesign SystemPackageTool interface

### LINKED ISSUE #1659: Redesign SystemPackageTool interface
From the addition of Chocolatey to SystemPackageTool (https://github.com/conan-io/conan/pull/1657), which could break existing recipes, I think there is something a bit weird in SystemPackageTool

The thing is that it can be used as:

```python
pack_name = "some_pkg"
installer = SystemPackageTool()
installer.install(pack_name)
```
That will try to use a different system package manager in every platform, like apt in debian, yum, pkgutil.... And will fall back to the ``NullTool`` if not existing. For example in Windows, and the above code will be a no-op in Windows.

So adding Chocolatey as another tool for Windows had potential to break things.

Typically the complexity of the decission to install a package is given on the previous package name selection:

```python
pack_name = None
    if os_info.linux_distro == "ubuntu":
        if os_info.os_version > "12":
            pack_name = "package_name_in_ubuntu_10"
        else:
            pack_name = "package_name_in_ubuntu_12"
    elif 
```

Maybe it makes more sense something like:

```python
      apt.install(<some pkgs>)
      yum.install(<some pkgs>)
      choco.install(<some pkgs>)
```
So extensibility is under user control. The above will be no-ops if the system it is running doesn't support that package managers.

Wdyt? @uilianries @sixten-hilborn @lasote

czoido:
Changelog: Feature: New tools in `conan.tools.system` for invoking system package managers in recipes.
Docs: https://github.com/conan-io/docs/pull/2379

To discuss:
- Check if we want to remove the guard to not running twice the system_requirements method. Add a conf for this?
- Taking into account build_requires
- Adding distro -> pm mapping
...

Closes: https://github.com/conan-io/conan/issues/1659


conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,175 @@
...
+                                   self._arch_names.get(self._arch))
+        return package
+
+    @property
+    def sudo_str(self):
+        sudo = "sudo " if self._sudo else ""
+        askpass = "-A " if self._sudo and self._sudo_askpass else ""
+        return "{}{}".format(sudo, askpass)
+
+    def method_decorator(wrapped):
```
memsharded: I think at some point, this should fail. If you use one of this, the mode is not "install" but only "check", and the packages are not installed in the system, it should raise an exception, not just warn?

We might want to re-design this a bit more decoupled, having a method that only collects the needs (like ``requirements()`` do for Conan dependencies. But the execution only happens later. This way we can collect, list, report on system requirements easily (besides installing them). This would be a different feature, probably out of the scope of this, but just an idea, this has been requested in the past by users (wanting to know, for a full dependency graph, what "system-requirements" are needed, etc, before installing them)

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,175 @@
...
+        def wrapper(self, *args, **kwargs):
+            method_name = wrapped.__name__
+            if self._active_tool == self.__class__.tool_name:
```
memsharded: Seems like the activation-deactivation of the tool should be a different responsibility (other decorator?) than checking the mode.

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,175 @@
...
+                if method_name in self.install_methods and self._mode == self.mode_check:
+                    self._conanfile.output.warn("Can't {}. Please update packages manually or set "
+                                                "'tools.system.package_manager:mode' to "
+                                                "'install'".format(method_name))
+                else:
+                    return wrapped(self, *args, **kwargs)
+            else:
+                self._conanfile.output.warn("Ignoring call to {}.{}. Set "
```
memsharded: This seems will be very noisy when a recipe adds a few managers for different platforms, all the others not active will be constantly warning? Maybe better remove this warning.

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,175 @@
...
+    update_command = ""
+    check_command = ""
+
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._active_tool = self._conanfile.conf["tools.system.package_manager:tool"]
+        self._sudo = self._conanfile.conf["tools.system.package_manager:sudo"]
+        self._sudo_askpass = self._conanfile.conf["tools.system.package_manager:sudo_askpass"]
+        self._mode = self._conanfile.conf["tools.system.package_manager:mode"] or self.mode_check
+        self._arch = self._conanfile.settings.get_safe("arch")
```
memsharded: This probably depends on the context, and needs to change for cross-building and ``build_requires`` that contain a ``system_requirements()``.

memsharded (commented):
Design and approach is looking good, good job.

conan/tools/system/package_manager.py
czoido: I changed it to raise an exception now, I also added the option of doing a check of missing packages before installing...

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,208 @@
+import platform
+
+from conans.client.graph.graph import CONTEXT_BUILD
+from conans.errors import ConanException
+
+
+class SystemPackageManagerTool(object):
```
memsharded: If not intended to be public, it is good to make it ``_System..`` private

conans/test/integration/tools/system/package_manager.py
```diff
@@ -0,0 +1,198 @@
+import mock
```
memsharded: File must be named ``test_package_manager.py`` otherwise it doesn't run in CI?

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        self._sudo = self._conanfile.conf["tools.system.package_manager:sudo"]
+        self._sudo_askpass = self._conanfile.conf["tools.system.package_manager:sudo_askpass"]
+        self._mode = self._conanfile.conf["tools.system.package_manager:mode"] or self.mode_check
+        self._arch = self._conanfile.settings_build.get_safe('arch') \
+            if self._conanfile.context == CONTEXT_BUILD else self._conanfile.settings.get_safe('arch')
+        self._arch_names = {}
+        self._arch_separator = ""
+
+    def get_default_tool(self):
+        os_name = platform.system()
```
memsharded, line 27: It is possible that at some point we will want to check that it matches ``self.settings`` too, but it is ok as-is now.

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,208 @@
...
+            if method_name in self.install_methods and self._mode == self.mode_check:
+                raise ConanException("Can't {}. Please update packages manually or set "
+                                     "'tools.system.package_manager:mode' to "
+                                     "'install'".format(method_name))
+            else:
+                return wrapped(self, *args, **kwargs)
+
+        return wrapper
+
+    @check_enabled_tool  # noqa
```
memsharded: I am wondering if it would be easier and simpler (no need of decorator and no need to decorate all methods) to have instead a ``def run()`` command that is the one controlled by the enable/disable tool.

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,208 @@
...
+        def wrapper(self, *args, **kwargs):
+            if self._active_tool == self.__class__.tool_name:
+                return wrapped(self, *args, **kwargs)
+
+        return wrapper
+
+    def check_mode(wrapped):
+        def wrapper(self, *args, **kwargs):
+            method_name = wrapped.__name__
+            if method_name in self.install_methods and self._mode == self.mode_check:
```
memsharded: I am not sure if I fully understand the logic. I think that the tools, even in check mode if possible should check if the requested packages are installed, and if they are, it shouldn't be an error, it should only raise if it really needs to install those packages.

I don't see how this implements this behavior, but maybe I am missing something.

conans/test/integration/tools/system/package_manager.py
czoido: 🤦

conan/tools/system/package_manager.py
czoido: I have updated the logic and removed the decorator for the mode check, now:
- In `check` mode:
   - `update` fails
   - `install` will check if missing packages by default (you can deactivate this) and if there are missing packages will fail, if not it will continue
- In `install` mode:
   - `update` tries to update
   - `install` install will check if missing packages by default (you can deactivate this) and if there are missing packages will try to install them

memsharded (approved):
Great!

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,212 @@
...
+            self.update()
+
+        if check:
+            packages = self.check(packages)
+
+        if self._mode == self.mode_check and packages:
+            raise ConanException("System requirements: '{}' are missing but can't install "
+                                 "because tools.system.package_manager:mode is '{}'."
+                                 "Please update packages manually or set "
+                                 "'tools.system.package_manager:mode' "
```
lasote, line 84: This is not the first time I wonder if a user will know that we are talking about a line in the `global.conf`. Should we be more explicit? I know the config could come from a requirement but...

lasote (approved):

conan/tools/system/package_manager.py
memsharded: It can come from a profile ``[conf]`` section too, it is probably the best place for this one

conan/tools/system/package_manager.py
```diff
@@ -0,0 +1,212 @@
...
+                                 "to '{}'".format(", ".join(packages),
```
memsharded:
```suggestion
            raise ConanException("System requirements: '{}' are missing but can't install "
                                 "because tools.system.package_manager:mode is '{}'."
                                 "Please update system packages manually or set in your profile [conf] section"
                                 "'tools.system.package_manager:mode={}'".format(", ".join(packages),
```

PR #10381, opened 2022-01-19
move editable to new cli

memsharded:


czoido (approved):
Looks good, just pending of conflict resolving

PR #10388, opened 2022-01-20
Add quotes around `conan_message` output variable so it is not modified

jgsogo:
Changelog: Fix: Add quotes around `conan_message` output variable so it is not modified.
Docs: omit

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -28,7 +28,7 @@ class MacrosTemplate(CMakeDepsFileTemplate):
         return textwrap.dedent("""
         function(conan_message MESSAGE_TYPE MESSAGE_CONTENT)
             if(NOT CONAN_CMAKE_SILENT_OUTPUT)
-                message(${MESSAGE_TYPE} ${MESSAGE_CONTENT})
+                message(${MESSAGE_TYPE} "${MESSAGE_CONTENT}")
             endif()
         endfunction()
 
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -28,7 +28,7 @@ def template(self):
         return textwrap.dedent("""
         function(conan_message MESSAGE_TYPE MESSAGE_CONTENT)
             if(NOT CONAN_CMAKE_SILENT_OUTPUT)
-                message(${MESSAGE_TYPE} ${MESSAGE_CONTENT})
+                message(${MESSAGE_TYPE} "${MESSAGE_CONTENT}")
```
jgsogo, line 31: Without quotes, it removes the `;` from lists... which can be very confusing when you are focused on an issue related to a list of paths 😄

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -212,7 +212,6 @@ def join_paths_single_var(values):
             return '"%s"' % ";".join(p.replace('\\', '/').replace('$', '\\$') for p in values)
 
         self.include_paths = join_paths(cpp_info.includedirs)
-        self.include_path = join_paths_single_var(cpp_info.includedirs)
```
jgsogo, line 215: This one is not used in the template (but I didn't check the if it is used in the rest of the codebase)

memsharded: Indeed seems unused, no prob

memsharded (approved):

lasote:
Relaunched CI, failed because of the weird language change in Windows CI.

lasote (approved):

PR #10389, opened 2022-01-20
fixing transitive headers and libs for known consuming packages

memsharded:
Minor fix, implementing headers and libs transitivity **removal** by default when a package is consumed by a known type in "app", "library". That is, headers and libs do not propagate downstream by default, irrespective of the dependency type when the consumer is an application or a library.


conans/model/requires.py
```diff
@@ -270,6 +270,10 @@ def transform_downstream(self, pkg_type, require, dep_pkg_type):
             # Unknown, default. This happens all the time while check_downstream as shared is unknown
```
lasote, line 270: Should we assert that the `dep_pkg_type` is None?

lasote (approved):

conans/model/requires.py
memsharded: It can be other things != None, like an APP, or build-scripts.
What we really know is that if the consumer package is a library or an app/library, we don't want to propagate headers or libs downstream

PR #10391, opened 2022-01-20
fix: use package folder for Bazel dependency paths

bb010g:
Changelog: Feature: Use `.bazel` suffix for generated Bazel files.
Changelog: Fix: Use install folder for Bazel dependency paths.
Docs: https://github.com/conan-io/docs/pull/2381

Builds for subdirectories wouldn't work when `--install-folder` was set to anything besides the current directory without this.

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -1,3 +1,4 @@
+import os
 import textwrap
 
 from jinja2 import Template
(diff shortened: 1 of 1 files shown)


conan/tools/google/bazeldeps.py
```diff
@@ -13,26 +14,27 @@ def __init__(self, conanfile):
 
     def generate(self):
         local_repositories = []
+        conandeps = os.path.join(self._conanfile.install_folder, 'conandeps')
```
memsharded: We are using the ``conanfile.generators_folder`` for the location of the generated files, as the ``install_folder`` will be deprecated.

It is also possible to define the relative folders layout in a ``layout()`` method.

memsharded (commented):
Thanks for contributing this.

The CI is read now, for an unrelated glitch in the Windows machine, that we need to investigate. Don't worry about it, we will relaunch this PR when it is fixed.

conan/tools/google/bazeldeps.py
bb010g: I noticed in https://github.com/conan-io/conan/pull/8246 that `conanfile.*_folder` seemed to be preferred over `conanfile.layout.*_folder`. Is this still the case?

https://github.com/conan-io/conan/blob/01f4aecbfe1a49f71f00af8f1b96b9f0174c3aad/conan/tools/cmake/cmake.py#L62-L63

Related question: Is there a location available at `generate()` time for the directory that would contain the `WORKSPACE.bazel` file? Normally, it's a sibling of the conanfile (and this might be a hard requirement for now?), and it is a child of the source location, but `source_folder` is not available (is `None`) at `generate()` time. It might be the case that passing `--source-folder <source_folder>` should be allowed for `conan install` due to Bazel's dependence on a workspace folder for configuration. I wanted to have this be more along the lines of
```py
    def generate(self):
        local_repositories = []
        conandeps = os.path.join(self._conanfile.generators_folder, 'conandeps')
        if conandeps.starts_with(self._conanfile.bazel_workspace_folder):
           conandeps = os.relpath(conandeps, self._conanfile.bazel_workspace_folder)
```
but I couldn't figure out a proper value to substitute for `self._conanfile.bazel_workspace_folder`.

As an example, without this PR, `conan install --profile:build default .` will generate a `./conandeps/dependencies.bzl` like:
```bzl

def load_conan_dependencies():
    
    native.new_local_repository(
        name="hello",
        path="/Users/bb010g/.conan/data/hello/1.0.0/_/_/package/ffffffffffffffffffffffffffffffffffffffff",
        build_file="conandeps/hello/BUILD.bazel",
    )

```
With this patch stack, I desire that `conan install --install-folder conaninfo --profile:build default .` will generate a `./conaninfo/conandeps/dependencies.bzl` like:
```bzl

def load_conan_dependencies():
    
    native.new_local_repository(
        name="hello",
        path="/Users/bb010g/.conan/data/hello/1.0.0/_/_/package/ffffffffffffffffffffffffffffffffffffffff",
        build_file="conaninfo/conandeps/hello/BUILD.bazel",
    )

```
However, with this patch stack as it currently is, `conan install --install-folder conaninfo --profile:build default .` will generate a `./conaninfo/conandeps/dependencies.bzl` like:
```bzl

def load_conan_dependencies():
    
    native.new_local_repository(
        name="hello",
        path="/Users/bb010g/.conan/data/hello/1.0.0/_/_/package/ffffffffffffffffffffffffffffffffffffffff",
        build_file="/Users/bb010g/Documents/my-example-conan-package/conaninfo/conandeps/hello/BUILD.bazel",
    )

```

memsharded (approved):

PR #10392, opened 2022-01-21
remove ConanRunner

memsharded:


lasote (approved):
Willing that py3 solves the old issue we had with the pipes, will see.
We should write something in the migration guide. 
We should think about what we want to change in the log system (tracer included) and here maybe log the run commands (not the output) to the log. Probably is a good idea to have different log keys, not just "conan", so people could filter only the entries for run commands, or any other thing we enable.

memsharded:
> Willing that py3 solves the old issue we had with the pipes, will see.

Well, the proposed solution, for production is not piped, it is passing the streams directly to the child process. That should avoid any pipe locking issue.

Only in testing and when the user provides a StringIO buffer to capture, the output will be piped to that.

> We should think about what we want to change in the log system (tracer included) and here maybe log the run commands (not the output) to the log. Probably is a good idea to have different log keys, not just "conan", so people could filter only the entries for run commands, or any other thing we enable.

Yes, I think this is part of a global effort we should do on standardizing on the output/logging levels and information.

Issue #10362, opened 2022-01-18
[feature] default cppinfo.builddirs


lasote:
- Remove the default "" from cppinfo.buildirs
- Clean the CMakeDeps code so it is not necessary to clear the root folders to avoid setting them to the CMAKE_PREFIX_PATH (PR SpaceIm)
- Document the migration path: Remove the '' from builddirs or use assign and not append.

czoido:
Closed by https://github.com/conan-io/conan/pull/10394

PR #10394, opened 2022-01-21
[2.0 alpha3] Remove default builddirs + simplified CMakeToolchain

lasote:
~~I'll open a PR~~ https://github.com/conan-io/docs/pull/2365 in the doc to complete the migration guide regarding the builddirs.
We will document at the 2.0 docs that the inclusion of [''] at the builddirs is discouraged. (e.g. once the root dir is added to CMAKE_PREFIX_PATH, cmake will find executables from host context instead of build context)

Closes https://github.com/conan-io/conan/issues/10362

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -383,17 +383,8 @@ class FindFiles(Block):
         {% endif %}
 
         # Definition of CMAKE_MODULE_PATH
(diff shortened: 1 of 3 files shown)


memsharded (approved):
Amazing, love this, very necessary.

czoido (approved):

PR #10401, opened 2022-01-23
removing legacy SystemPM and OSInfo

memsharded:


memsharded (commented):
Realized while doing this that there are a couple of tests with:
```python
@pytest.mark.skip(reason="brew update takes a lot of time")
```
@czoido , this is not a good sign, we might need the caching of state, or maybe avoid the update for brew and run only the check/install?

czoido:
> ```python
> brew update takes a lot of t
> ```

I have to make more functional tests, but  I think that this only happens the first time you do an update since a long time. In local it's running now in a couple of seconds but the Mac node in the CI was taking a lot of time and then I skipped them.

czoido (approved):
LGTM

PR #10414, opened 2022-01-24
fix: use  `Folders.package` in `Folders.package_folder`

bb010g:
Currently, installing directly from a configured build tool uses `folders._base_package`, which is often the root folder. Relevant cache cases should set `folders.package = None` before getting `folders.package_folder` or otherwise find a way to only use `folders._base_package`.

Changelog:
* Feat: Use `Folders.package_folder` in `ConanFile.package_folder`
* Fix: Respect `Folders.package` in `Folders.package_folder`

### DIFF
--- a/conans/client/cmd/export_pkg.py
+++ b/conans/client/cmd/export_pkg.py
@@ -61,6 +61,7 @@ def export_pkg(app, recorder, full_ref, source_folder, build_folder, package_fol
         conanfile_folder = os.path.dirname(source_conanfile_path)
         conanfile.folders.set_base_folders(conanfile_folder, output_folder=None)
         conanfile.folders.set_base_package(dest_package_folder)
+        conanfile.folders.package = None
     else:
         conanfile.folders.set_base_build(build_folder)
         conanfile.folders.set_base_source(source_folder)
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -570,6 +570,7 @@ class BinaryInstaller(object):
                                                    % (str(pref), package_folder))
             # Call the info method
             conanfile.folders.set_base_package(package_folder)
+            conanfile.folders.package = None
             conanfile.folders.set_base_source(None)
             conanfile.folders.set_base_build(None)
             conanfile.folders.set_base_install(None)
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -271,7 +271,7 @@ class ConanFile(object):
 
     @property
     def package_folder(self):
-        return self.folders.base_package
+        return self.folders.package_folder
 
     @property
     def install_folder(self):
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -102,8 +102,11 @@ class Folders(object):
 
     @property
     def package_folder(self):
-        """For the cache, the package folder is only the base"""
-        return self._base_package
+        if self._base_package is None:
+            return None
+        if not self.package:
+            return self._base_package
+        return os.path.join(self._base_package, self.package)
 
     @property
     def generators_folder(self):


memsharded (commented):
There is a reason we decided not to apply the layout model to the "package" functionality, it is that we are removing the support for local ``package`` in user space, as it didn't have much utility beyond debugging the ``package()`` method, and that can still be done using the ``conan export-pkg`` command. Actually there is also a ConanException when using ``package`` and the recipe defines a ``layout()`` method, to prevent this:

```python
raise ConanException("The usage of the 'conan package' local method is disabled when "
                                 "using layout(). Use 'export-pkg' to test if the recipe is "
                                 "packaging the files correctly or use the cpp.info.local object "
                                 "if you are going to use this package as editable package.")
```



So I think this is not desired, but lets check with @lasote

bb010g:
@memsharded Without this, a `hello/conanfile.py` snippet
```py
from conan import ConanFile
from conan.tools.meson import Meson, MesonToolchain

class HelloConan(ConanFile):
    def layout(self):
        self.folders.generators = "conangen"
        self.folders.source = "src"
        self.folders.build = "build"
        self.folders.package = "dist"

    def generate(self):
        toolchain = MesonToolchain(conanfile=self)
        toolchain.generate()

    def build(self):
        meson = Meson(conanfile=self)
        meson.configure()
        meson.build()
```
will always make `conan install --profile:build --default . && conan build .` run the Meson configure cmd `meson setup --native-file "…/hello/conangen/conan_meson_native.ini" "…/hello/build" "…/hello/src" -Dprefix="…/hello"`. Passing `-Dprefix="…/hello/dist"` is not possible without either manually running `meson setup -Dprefix="…/hello/dist"` afterwards & becoming out of sync with Conan, avoiding `conan build` altogether & running `meson setup` manually, or applying this patch. I don't care what Meson's `prefix` variable is for cache builds; I very much care what it is when I'm running `rm -r dist; meson install -C build` during development, especially now that `conan package` is deprecated.

https://github.com/conan-io/conan/blob/1940331e56454c7bfae2b17809f5d176a5b6ebbc/conan/tools/meson/meson.py#L23-L24

https://github.com/conan-io/conan/blob/1940331e56454c7bfae2b17809f5d176a5b6ebbc/conan/tools/cmake/cmake.py#L81-L83

If this isn't Conan's answer for selecting the directory used by its tools as an install prefix for local development, Conan needs to figure out what that answer is and start respecting it. Otherwise, if these tools act as they should *and* a `folders.package_folder` that's just `folders.base_package` is correct, Conan intentionally lacks a way to specify the install prefix for local development, and I don't get what is gained by that.

Additionally, all CI tests pass, so I think the behavior with regards to the cache is fine? `test/functional/layout/test_local_commands.py` looks to test this behavior explicitly? 

<https://github.com/conan-io/conan/blob/1940331e56454c7bfae2b17809f5d176a5b6ebbc/conans/test/functional/layout/test_local_commands.py#L298-L306>

Side note: That exception message mentions using `cpp.local`, but `cpp.local` was removed (reverted) in <https://github.com/conan-io/conan/pull/9882>.

lasote:
> test/functional/layout/test_local_commands.py looks to test this behavior explicitly?

That test (the only one assigning a `folders.package`) is skipped with `@pytest.mark.xfail`. 

I think we removed that support because we don't want the package layout to be customized in the cache and it was weird to have a different behavior locally. Actually, I bet that accepting this PR will customize the package in the cache too, it is not that straightforward to skip that behavior only for the cache. Said that I think you are right! we thought, ok, if we are removing the "conan package" command, this is useless, but we forgot about the build helpers and the prefix path and I think that is a legit use case.

I would say we have to think about this again and try to come up with something less weird, less fragile, and more explicit than having the "folders.package" only working for local "randomly".

bb010g:
Resetting `conanfile.folders.package` in `export_pkg()` like so has been working so far:

```py
    if hasattr(conanfile, "layout"):
        conanfile_folder = os.path.dirname(source_conanfile_path)
        conanfile.folders.set_base_build(conanfile_folder)
        conanfile.folders.set_base_source(conanfile_folder)
        conanfile.folders.set_base_package(dest_package_folder)
        conanfile.folders.package = None
        conanfile.folders.set_base_install(conanfile_folder)
        conanfile.folders.set_base_generators(conanfile_folder)
    else:
```

If you combine this with an e.g. `meson.configure()` (that knows when to `--reconfigure`) before `meson.install()` in `ConanFile.package()`, `conan export-pkg` works as expected in what I've tried so far.

memsharded:
This is related to https://github.com/conan-io/conan/pull/10672

memsharded:
We are discussing this and the proposed approach will be:

- The local flow for new tools and using ``layout()`` will not define ``package_folder``. Then the build system integrations will not define prefixes for installation
- Users will be able to define those prefixes when they call their build system, like ``cmake --install --prefix=mydecidedfolder``. They can also install their app in the system if that is the build system default, and users want to do that.
- Basically Conan will do nothing when working in local flow regarding install prefixes, and accordingly, it will not need the ``package_folder`` 

We will try in another PR this approach, making ``package_folder=None`` for the local flow, and see what happens.

memsharded:
This PR is now a bit outdated.
As explained above, the parametrization of the ``package_folder`` is not something intended to be covered by the ``layout()``, because it doesn't fit the model. In local flow, the ``package_folder`` will not be defined (None), and in the cache, it cannot be parameterized, it is always the same "package" folder in the cache.

Closing the PR, thanks very much again for your contribution!

Issue #10411, opened 2022-01-24
[feature] List required_conan_version when running inspect and info commands


uilianries:
The `required_conan_version` can be a global configuration and listed by `conan config` command, but when it's declared in a recipe, both commands `inspect` and `info` can not list it because it's not part of `ConanFile` object.

https://docs.conan.io/en/latest/reference/conanfile/other.html#requiring-a-conan-version-for-the-recipe

The idea is listing the `required_conan_version` when running:

    conan inspect conanfile.py
    required_conan_version: >=1.44.0

    conan info conanfile.py
    Require Conan Version: >=1.44.0

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

PR #10431, opened 2022-01-26
List required_conan_version on inspect command

uilianries:
We have some ideas for C3i which require inspecting `required_conan_version`, also it's a variable which users can't check without reading the file.

This PR updates the command `inspect` only, but I'm intended to update `info` in a separated PR.

Changelog: Feature: Inspect required_conan_version value
Docs: https://github.com/conan-io/docs/pull/2371

Related to #10411

### DIFF
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -33,7 +33,7 @@ from conans.client.graph.range_resolver import RangeResolver
 from conans.client.hook_manager import HookManager
 from conans.client.importer import run_imports, undo_imports
 from conans.client.installer import BinaryInstaller
-from conans.client.loader import ConanFileLoader
+from conans.client.loader import ConanFileLoader, load_required_conan_version
 from conans.client.manager import deps_install
 from conans.client.migrations import ClientMigrator
 from conans.client.output import ConanOutput, colorama_initialize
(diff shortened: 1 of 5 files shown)


memsharded (commented):
I am not convinced about this feature. Reads a complex implementation, with a dedicated loading and python parsing to extract just this element data.

Furthermore, we have removed ``conan inspect`` from conan 2.0, because we are not convinced about its implementation either, specially when loading from a remote. This makes the issue bigger, so doesn't look good, at this moment this won't be merged.

Can you elaborate on the ideas and the need for knowing the required conan version at c3i?

uilianries:
> I am not convinced about this feature. Reads a complex implementation, with a dedicated loading and python parsing to extract just this element data.
> 
> Furthermore, we have removed `conan inspect` from conan 2.0, because we are not convinced about its implementation either, specially when loading from a remote. This makes the issue bigger, so doesn't look good, at this moment this won't be merged.
> 
> Can you elaborate on the ideas and the need for knowing the required conan version at c3i?

Please, read https://github.com/conan-io/c3i_jenkins/issues/1248

uilianries:
Besides C3i need, let's say you want to know which minimal Conan version your recipe needs. How do you check it without reading the recipe? You basically can't. We have a feature for conan config, where we use a minimal global version, but we can't do it individually.

memsharded:
> Besides C3i need, let's say you want to know which minimal Conan version your recipe needs. How do you check it without reading the recipe? 

Yes, you read the recipe, what is the big issue here? I still don't fully see the use case yet. The Conan version in c3i is fixed. For users, we explicitly say that only latest Conan guarantees compatibility for ConanCenter recipes in the general case.

uilianries:
Okay, so let's close this one. At least, C3i is fixed and no one requested it so far, besides us.

PR #10433, opened 2022-01-26
POC: Remotes from local file tree

lasote:
- The remote points to a local folder (can be a clone, or whatever):
    - `conan remote add ccc-sources file:///Users/luism/workspace/conan-center-index`
- The remote is usable, search is working too.
- The upload/remove management is disabled and a package get will always raise PackageNotFound. 
- Every time the client request a recipe, a recipe is exported in a conan cache is kept at the ``{clone folder}/.cache/conan_remote``.
- If the clone is pulled, and a recipe is updated, you "conan install" that recipe with --update so you will get a different revision. Both revisions will be kept until the user remove the remote cache folder. After a clean, only the latest will be available.


tsondergaard:
Maybe it is worth considering a slightly more general solution where remotes can also be scripts. With a file:// url there would have to be a lot of assumptions about the layout. See my feature request #11646, which I have now closed after this PR was pointed out to me.

memsharded:
> Maybe it is worth considering a slightly more general solution where remotes can also be scripts. With a file:// url there would have to be a lot of assumptions about the layout. See my feature request https://github.com/conan-io/conan/issues/11646, which I have now closed after this PR was pointed out to me.

We have considered different approaches, and it seems that some assumptions about the layout are necessary, it is not possible to be that generic, one of the reasons is the resolution of versions. Having a layout that defines the different existing versions to be able to resolve them, resolve version ranges, see if a given version exist, etc, before even fetching it is necessary.

memsharded:
Superseded by https://github.com/conan-io/conan/pull/13930

PR #10434, opened 2022-01-26
cmake-toolchain: set CMAKE_SYSTEM_PROCESSOR in Apple block

theodelrieu:
Hello,

`CMAKE_SYSTEM_PROCESSOR` is not set by `CMakeToolchain` when building on Apple, I'm not sure if this is the right place to set it.

It is mentioned in `GenericSystemBlock`, but after debugging a bit, the values returned by `self._get_cross_build()` are all `None`.

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -311,6 +311,9 @@ class AppleSystemBlock(Block):
         {% if CMAKE_SYSTEM_VERSION is defined %}
         set(CMAKE_SYSTEM_VERSION {{ CMAKE_SYSTEM_VERSION }})
         {% endif %}
+        {% if CMAKE_SYSTEM_PROCESSOR is defined %}
+        set(CMAKE_SYSTEM_PROCESSOR {{ CMAKE_SYSTEM_PROCESSOR }})
+        {% endif %}
         # Set the architectures for which to build.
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Are you sure that using different architecture for host and build profiles it was not defined? I am concerned that this can conflict with the ``GenericSystemBlock`` one in some cases.

Also a bit more detailed why/when this is needed, problems that it is solving, etc. So far we have apple OSX, iOS and M1 cross-builds in our testing, and all seems to be working fine.

It would also be needed at least some unit testing. We can help with that once the problem and scope is clear.

theodelrieu:
>Also a bit more detailed why/when this is needed, problems that it is solving, etc.

We have a library that checks `CMAKE_SYSTEM_PROCESSOR` to determine if `thread_local` should be used or not (CMake has `cxx_feature` stuff, but `thread_local` detection is buggy).

When cross-building for 32-bit x86 simulator, the library did define `thread_local`, which is not supported on this platform. I realised that `CMAKE_SYSTEM_PROCESSOR` was empty.

>I am concerned that this can conflict with the GenericSystemBlock one in some cases.

True, I guess finding out why `_get_cross_build` returns empty values would be a cleaner way to tackle the problem 😅

memsharded:
Superseded by https://github.com/conan-io/conan/pull/10856, which is targeted for next 1.47.
Thanks very much for raising this!

PR #10442, opened 2022-01-27
Test Shared OSX exe

lasote:
Added tests about shared libraries, rpaths at OSX


lasote:
This is ready, setting @executable_path at DYLD_LIBRARY_PATH fails in CI but there are tests with some other alternatives.

conans/test/utils/tools.py
```diff
@@ -559,7 +559,7 @@ def _handle_cli_result(self, command, assert_error, error):
                 output_header='{:-^80}'.format(" Output: "),
                 output_footer='-' * 80,
                 cmd=command,
-                output=str(self.stderr) + str(self.stdout)
+                output=str(self.stderr) + str(self.stdout) + "\n" + str(self.out)
```
memsharded, line 567: Why is this necessary? It shouldn't, as ``self.out = self.stderr + self.stdout``?

conans/test/functional/toolchains/cmake/test_shared_cmake.py
```diff
@@ -39,3 +43,156 @@ def test_shared_cmake_toolchain_test_package():
...
+
+                    def layout(self):
+                        cmake_layout(self)
+
+                    def build(self):
+                        cmake = CMake(self)
+                        cmake.configure()
+                        cmake.build()
+
+                    def imports(self):
```
memsharded: This is probably dead now that we removed imports?

conans/test/functional/toolchains/cmake/test_shared_cmake.py
```diff
@@ -39,3 +43,156 @@ def test_shared_cmake_toolchain_test_package():
     client.save(files)
     client.run("create . -o hello:shared=True")
     assert "hello: Release!" in client.out
+
+
+@pytest.fixture()
+def test_client_shared():
+    client = TestClient()
+    files = pkg_cmake("hello", "0.1")
+    files.update(pkg_cmake_test("hello"))
```
memsharded: This might be the ``conan new cmake_lib`` template already?

conans/test/functional/toolchains/cmake/test_shared_cmake.py
```diff
@@ -39,3 +43,156 @@ def test_shared_cmake_toolchain_test_package():
...
+                    def layout(self):
+                        cmake_layout(self)
+
+                    def build(self):
+                        cmake = CMake(self)
+                        cmake.configure()
+                        cmake.build()
+                        cmake.install()
+
+                    def imports(self):
```
memsharded: Same

conans/test/utils/tools.py
lasote: I think the `run_command` (not the `run`) is not showing correctly the output error because it is doing: `with redirect_output(self.out):`.
Probably it has to be fixed otherwise.

conans/test/functional/toolchains/cmake/test_shared_cmake.py
lasote: only changed the cpp file name, the output of the exe... half an hour debugging GRRRR!

memsharded (approved):
I am approving this, but I would like to suggest there are several potential improvements, specially regarding testing time:

- The ``fixture()`` can be put at scope ``module``. So it will build only once, and as all the other tests seems to be just consuming that shared dylib, it should be ok, and much faster.
- Put all tests under a ``TestOSXShared`` class, and do the ``skipif()`` for the class.
- Avoid ``*.dll`` management, not used.
- Use ``conan test`` command or ``conan create --build=missing`` to avoid rebuild if not necessary.

Issue #10417, opened 2022-01-25
[feature] Support for multiple remotes in Conan 2.0


czoido:
Currently you can't do something like `conan install -r remote1 -r remote2` just forcing one remote or checking in all the remotes.
Check how to give support for this for several commands.

Related to: https://github.com/conan-io/conan/issues/7998

memsharded:
Closed by https://github.com/conan-io/conan/pull/10443

PR #10443, opened 2022-01-27
Accept several remotes as arguments for some commands (-r remote1 -r remote2)

czoido:
In Conan 1.X for some commands like `create` and `install` you can just specify one remote or not adding the `-r`  argument, meaning, check in all enabled remotes. 
Now, with this change you can do a `conan install -r remote1 -r remote2` meaning that it will only check in those remotes.

Closes: https://github.com/conan-io/conan/issues/10417


conans/test/integration/remote/multi_remote_test.py
```diff
@@ -137,7 +137,7 @@ def test_conan_install_update(self):
                       client.out)
 
         client.run("install --reference=hello0/0.0@lasote/stable --update -r default")
-        self.assertIn("hello0/0.0@lasote/stable#4963e851819c02f9578b9d6bcecd01aa - Cache",
+        self.assertIn("hello0/0.0@lasote/stable#4963e851819c02f9578b9d6bcecd01aa - Newer",
```
czoido, line 140: This test was returning cache because it was searching in all remotes, including local, but that was a bug, if you add the argument -r default it should only check in default, not in all remotes.

conans/cli/api/subapi/remotes.py
```diff
@@ -31,9 +31,12 @@ def list(self, pattern=None, only_active=False):
         return remotes
 
     @api_method
-    def get(self, remote_name):
+    def get(self, remote_names=None):
```
lasote: I think a `get` is for 1 element. You can perfectly iterate to get N names calling N times.

conans/client/conan_api.py
```diff
@@ -103,7 +103,8 @@ def download(self, reference, remote_name=None, packages=None, recipe=False):
                         raise ConanException("It is needed to specify the recipe revision if you "
                                              "specify a package revision")
             # FIXME: remote_name should be remote
```
lasote, line 105: Might be a good time to fix the argument and pass a Remote

conans/test/integration/remote/multi_remote_test.py
lasote: Yes, I observed that bug lately

conans/client/cmd/download.py
```diff
@@ -13,12 +13,16 @@ def download(app, ref, package_ids, recipe):
     hook_manager = app.hook_manager
     assert(isinstance(ref, RecipeReference))
     scoped_output = ScopedOutput(str(ref), ConanOutput())
-    remote = app.selected_remote
-    if not remote:
+    # TODO: for download we should only accept one remote ?
```
lasote, line 16: I would say yes, the command is a bit weird, downloading from N is even more weird.

conans/cli/commands/remove.py
```diff
@@ -31,7 +31,7 @@ def remove(conan_api: ConanAPIV2, parser, *args):
     args = parser.parse_args(*args)
```
lasote: I don't see it necessary, but I see more value in a "multiple remote remove" than in a "multiple remote download", said that, I would keep it single remote.

lasote (changes requested):
Looking good!

conans/client/conan_api.py
czoido: But that should be done when we migrate conan download API to V2, right?

lasote: oh! right, this is the old.

conans/cli/api/subapi/remotes.py
```diff
@@ -35,6 +35,14 @@ def get(self, remote_name):
         app = ConanApp(self.conan_api.cache_folder)
         return app.cache.remotes_registry.read(remote_name)
 
+    @api_method
+    def get_remotes(self, remote_names=None):
+        app = ConanApp(self.conan_api.cache_folder)
+        if remote_names:
+            return [self.get(remote_name) for remote_name in remote_names]
+        elif remote_names is None:  # if we don't pass any remotes we want to retrieve only the enabled ones
+            return app.cache.remotes_registry.list(filter_disabled=True)
```
lasote: I don't think this code belongs to the remotes API, the `# if we don't pass any remotes we want to retrieve only the enabled ones` is logic from the command. We might just share a function in the `cli/commands/__init__` or so.

lasote (changes requested):

conans/cli/api/subapi/remotes.py
czoido: yes, this does not look good here, I'll change it

conans/cli/common.py
```diff
@@ -127,3 +127,10 @@ def get_lockfile(lockfile, strict=False):
         graph_lock.strict = strict
         ConanOutput().info("Using lockfile: '{}'".format(lockfile))
     return graph_lock
+
+
+def get_multiple_remotes(conan_api, remote_names=None):
```
memsharded, line 133: I know this function has been moved, from the api, but the access to ``conan_api.cache.remotes_registry.list()`` in this layer is a smell, it should be a call to the api

conans/client/graph/range_resolver.py
```diff
@@ -67,19 +67,17 @@ def _resolve_local(self, search_ref, version_range):
     def _search_and_resolve_remotes(self, search_ref, version_range):
         results = []
         remotes = self._conan_app.enabled_remotes
```
memsharded: Why here it is ``enabled`` and not ``select_remotes``?

memsharded (commented):
Looking good

conans/cli/common.py
lasote: Yes, the `conan_api.remotes.list` could be used.

conans/client/graph/range_resolver.py
czoido: sorry, that was a mistake, that should be `selected_remotes` 👍

memsharded (approved):

conans/cli/commands/remove.py
memsharded: Yes, better single remote, both for download and for remove

lasote (approved):

PR #10464, opened 2022-01-30
[develop2] remove legacy --install-folder

memsharded:
The --install-folder was used for files not generated anymore, this is being cleaned.
Now it is superseded by ``--output-folder`` as base folder for layout()


czoido (approved):

lasote (approved):
👏  Today is a great day when after huge effort and N features, we finally remove the install folder

PR #10468, opened 2022-01-31
Bugfix/make 'conan new' templates work for special chars in package name

domfellner:
Changelog: Bugfix: Avoid issues of ``conan new`` templates with packages containing special characters like ``-.+``
Docs: Omit

- Replace leftovers of 'package_name' in templates (Conan 1.x) with 'class_name', defaulting to patched 'name')
- Make 'conan new' templates work for special chars in package name ("+-._" are allowed)

This is targeting develop2 and I didn't find corresponding docs, so I omitted them. Closed the former docs PR targeting develop

### DIFF
--- a/conans/cli/api/helpers/new/cmake_exe.py
+++ b/conans/cli/api/helpers/new/cmake_exe.py
@@ -13,7 +13,7 @@ class {{class_name or name}}Recipe(ConanFile):
     license = "<Put the package license here>"
     author = "<Put your name here> <And your email here>"
     url = "<Package recipe repository url here, for issues about the package>"
-    description = "<Description of {package_name} here>"
+    description = "<Description of {{ name }} package here>"
     topics = ("<Put some tag here>", "<here>", "<and here>")
 
     # Binary configuration
@@ -56,7 +56,7 @@ from conan import ConanFile
 from conan.tools.build import cross_building
 
 
-class {{package_name}}TestConan(ConanFile):
+class {{class_name}}TestConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
 
     def requirements(self):
--- a/conans/cli/api/helpers/new/cmake_exe.py
+++ b/conans/cli/api/helpers/new/cmake_exe.py
@@ -4,7 +4,7 @@ conanfile_exe = '''from conan import ConanFile
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
 
 
-class {{class_name or name}}Recipe(ConanFile):
+class {{class_name or name.replace("-", "_").replace("+", "_").replace(".", "_")}}Recipe(ConanFile):
     name = "{{name}}"
     version = "{{version}}"
     package_type = "application"
@@ -56,7 +56,7 @@ from conan import ConanFile
 from conan.tools.build import cross_building
 
 
-class {{class_name}}TestConan(ConanFile):
+class {{class_name or name.replace("-", "_").replace("+", "_").replace(".", "_")}}TestConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
 
     def requirements(self):
--- a/conans/cli/api/helpers/new/cmake_lib.py
+++ b/conans/cli/api/helpers/new/cmake_lib.py
@@ -168,7 +168,7 @@ from conan.tools.layout import cmake_layout
 from conan.tools.build import cross_building
 
 
-class {{package_name}}TestConan(ConanFile):
+class {{class_name}}TestConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
     generators = "CMakeDeps", "CMakeToolchain"
 
--- a/conans/cli/api/helpers/new/cmake_lib.py
+++ b/conans/cli/api/helpers/new/cmake_lib.py
@@ -2,7 +2,7 @@ conanfile_sources_v2 = '''from conan import ConanFile
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
 
 
-class {{class_name or name}}Recipe(ConanFile):
+class {{class_name or name.replace("-", "_").replace("+", "_").replace(".", "_")}}Recipe(ConanFile):
     name = "{{name}}"
     version = "{{version}}"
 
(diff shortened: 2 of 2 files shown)


conans/assets/templates/new_v2_cmake.py
```diff
@@ -204,21 +204,23 @@ def test(self):
...
 
 int main() {{
-    {name}();
+    {symbol_name}();
 }}
 """
 
 
-def get_cmake_lib_files(name, version, package_name="Pkg"):
+def get_cmake_lib_files(name, version, package_name="Pkg", symbol_name="pkg"):
```
memsharded: The default ``symbol_name="pkg"`` is dangerous. It shouldn't have a default.

memsharded (changes requested):
This is modifying code that has already been ported to 2.0 (develop2) branch, under a new place, new cli syntax.

We'd prefer to focus the work there. Merging this will create too many conflicts while merging develop->develop2, and this PR is not a critical functionality, so if you could target this for 2.0 "develop2" branch that would be very appreciated.

conans/assets/templates/new_v2_cmake.py
domfellner: I didn't want to break the API of this function. Should I add it as a mandatory parameter before package_name (I wondered this has a default anyway)?

domfellner: Thanks for the fast feedback anyway.

Sorry, I cannot find where the rendering context is defined in the develop2 branch.

Is this implemented yet? The template looks a bit WIP there, containing

```class {{class_name or name}}Recipe(ConanFile):```

and I cannot find the definition of `class_name` anywhere.

Maybe these variables should be renamed according to their formatting like upper_camel_case_name and lower_camel_case_name? This would be more verbose than object_name.

Comment (memsharded) [reply conans/assets/templates/new_v2_cmake.py]: > I didn't want to break the API of this function. Should I add it as a mandatory parameter before package_name (I wondered this has a default anyway)?

It is purely internal function, probably called just once in code. No probe to change it.

Comment (memsharded) [reply conans/assets/templates/new_v2_cmake.py]: > Is this implemented yet? The template looks a bit WIP there, containing

Yes, the ``conan new cmake_lib -d name=pkg -d version=0.1`` command is fully operational in alpha3 and in develop2 branch. 
The templates are jinja templates, ``-d class_name=Something`` will dynamically inject it.

Comment (domfellner) [reply conans/assets/templates/new_v2_cmake.py]: Ah I see, I need to manually inject the definitions. It was filled automatically in Conan 1.x, but it makes sense to make this explicit.

Comment (memsharded) [reply conans/assets/templates/new_v2_cmake.py]: Yes, we are moving the whole "new" functionality to pure jinja templates, and a thin command line layer.

Comment (domfellner) [reply conans/assets/templates/new_v2_cmake.py]: OK have changed the commit. It's not really worth the effort any more 🤷 I'll close the corresponding docs PR

**Review by memsharded — COMMENTED** [review]
I would like to ask, because I don't fully get the purpose of this. Is this just to have a prettier define, like ``HELLO_EXPORT`` instead of ``hello_EXPORT``? Because I am fine with doing ``{{name.upper()}}`` that would be simpler. Because the thing is that users learning the ``-d symbol_name=HELLO`` variable and purpose is unlikely. 

Recall that this ``conan new`` templates are purely for demoing, testing, and quickly starting things, by no mean it intends to be production C++ code, and it should be replaced anyway with real code.

**Comment by domfellner** [comment]
> I would like to ask, because I don't fully get the purpose of this. Is this just to have a prettier define, like `HELLO_EXPORT` instead of `hello_EXPORT`? Because I am fine with doing `{{name.upper()}}` that would be simpler. Because the thing is that users learning the `-d symbol_name=HELLO` variable and purpose is unlikely.
> 
> Recall that this `conan new` templates are purely for demoing, testing, and quickly starting things, by no mean it intends to be production C++ code, and it should be replaced anyway with real code.

My initial intention was that I created a package "foo-bar", as our package name convention works with "-" as separator. This leads to `foo-bar_EXPORT` etc. which breaks the build. I regarded this as inconvenient and frustrating for new users and wanted to ensure the buildability . As "+-." are allowed in names, these characters needed to be removed from the variable before replacement in the template, but I fear this is only implementable with three `replace()` calls. Maybe that's the way to go?
I agree that this is rather a corner use case and this new variable is inconvenient as well.

As I wrote it actually happened here and if I was a Conan beginner, I would have been frustrated. But we can also just leave it as it is as most users will not come across this anyway.

**Comment by memsharded** [comment]
> As I wrote it actually happened here and if I was a Conan beginner, I would have been frustrated. But we can also just leave it as it is as most users will not come across this anyway.

Now I completely get the source cause which I think you are right. It is important. And it is important to solve it in an automated way, because other users will be frustrated until they learn they have another ``-d something=xxx`` to fix it.

I think we can safely do a ``replace("-","_").upper()`` in the jinja template (yeah, no big problem if we need to do 3 replaces for +. too, it is more important the UX than the template being super pretty), so it will be solved automatically for everyone. What do you think?

File: conans/cli/api/helpers/new/cmake_exe.py [hunk 2]
```diff
@@ -13,7 +13,7 @@ class {{class_name or name}}Recipe(ConanFile):
     license = "<Put the package license here>"
     author = "<Put your name here> <And your email here>"
     url = "<Package recipe repository url here, for issues about the package>"
-    description = "<Description of {package_name} here>"
+    description = "<Description of {{ name }} package here>"
```
domfellner, line 16: adapted this to cmake_lib.py

conans/cli/api/helpers/new/cmake_lib.py
```diff
@@ -60,20 +60,20 @@ def package_info(self):
 """
 
 source_h = """#pragma once
-
+{% set define_name = name.replace("-", "_").replace("+", "_").replace(".", "_").upper() %}
```
domfellner, line 64: This might serve as a hint that there is more to jinja than mere replacing, exposing the {% ... %} syntax

domfellner:
Okay I tried my luck. I'll pay for another round ;)

memsharded (approved):
Looking good, thanks!

memsharded:
Merged, thanks for contributing, will be released in next 2.0.0-alpha4

PR #10480, opened 2022-02-01
Fix Test System

lasote:
- The `pytest.mark.tool_XXX` was not working correctly when declared a fixture in the `scope="module"` because our `tool_XXX` mechanism was using an `autouse` fixture that was not applied when living in a different scope.
- Replaced with `pytest_runtest_setup` + `pytest_runtest_teardown` at `conftest.py`
- A huge amount of tests were not adding the `tool_XXX` needed.
- New syntax of tools `@pytest.mark.tool("cmake", "3.7")` with the version as the second argument being optional.

All this has been unveiled because my pycharm stopped adding the `/usr/local/bin` to the PATH so all the brew tools where only located when correctly tagged and when no `scope="module"`.

### DIFF
--- a/conans/test/functional/layout/test_editable_cmake.py
+++ b/conans/test/functional/layout/test_editable_cmake.py
@@ -68,11 +68,12 @@ def editable_cmake(generator, build_folder=None):
 
 @pytest.mark.skipif(platform.system() != "Windows", reason="Only windows")
 @pytest.mark.parametrize("generator", [None, "MinGW Makefiles"])
-@pytest.mark.tool_mingw64
+@pytest.mark.tool("mingw64")
 def test_editable_cmake_windows(generator):
     editable_cmake(generator)
 
 
+@pytest.mark.tool("cmake")
 def test_editable_cmake_windows_folders():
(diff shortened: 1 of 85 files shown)


conans/test/functional/layout/test_editable_cmake.py
```diff
@@ -68,26 +68,27 @@ def build_pkg(msg):
...
+@pytest.mark.tool("cmake")
 def test_editable_cmake_windows_folders():
     build_folder = temp_folder()
     editable_cmake(generator=None, build_folder=build_folder)
 
 
 @pytest.mark.skipif(platform.system() != "Linux", reason="Only linux")
 @pytest.mark.parametrize("generator", [None, "Ninja", "Ninja Multi-Config"])
-@pytest.mark.tool_cmake(version="3.17")
+@pytest.mark.toolcmake(version="3.17")
```
memsharded: Broken?

memsharded (approved):
Marking approach reads better and more correctly implemented, nice!

czoido (approved):
🙌

Issue #10481, opened 2022-02-01
BazelDeps failing when using absolute (system) folders


memsharded:
Looks like for some packages, we have:

```
p= /usr/include/freetype2 base_path= /home/mboujema/.conan/data/xorg/system/_/_/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
```
assertions fails then:
```
        def _relativize_path(p, base_path):
            # TODO: Very fragile, to test more
            assert os.path.isabs(p), "{} is not absolute".format(p)
            assert p.startswith(base_path)

            return p[len(base_path):].replace("\\", "/").lstrip("/")
 ```

I'm guessing we need to remove the preffix only if its there

_Originally posted by @mohabouje in https://github.com/conan-io/conan/issues/10476#issuecomment-1027035172_

**Comment by mohabouje** [comment]
Could be related:
```
conanfile.py (tijara-conan-integration/0.0.1): ERROR: Traceback (most recent call last):
  File "/home/mboujema/.local/lib/python3.9/site-packages/conans/client/generators/__init__.py", line 174, in write_generators
    generator.generate()
  File "/home/mboujema/.local/lib/python3.9/site-packages/conan/tools/google/bazeldeps.py", line 27, in generate
    content = self._get_dependency_buildfile_content(dependency)
  File "/home/mboujema/.local/lib/python3.9/site-packages/conan/tools/google/bazeldeps.py", line 129, in _get_dependency_buildfile_content
    "libdir": cpp_info.libdirs[0],
IndexError: list index out of range
```

**Comment by mohabouje** [comment]
> Could be related:
> 
> ```
> conanfile.py (tijara-conan-integration/0.0.1): ERROR: Traceback (most recent call last):
>   File "/home/mboujema/.local/lib/python3.9/site-packages/conans/client/generators/__init__.py", line 174, in write_generators
>     generator.generate()
>   File "/home/mboujema/.local/lib/python3.9/site-packages/conan/tools/google/bazeldeps.py", line 27, in generate
>     content = self._get_dependency_buildfile_content(dependency)
>   File "/home/mboujema/.local/lib/python3.9/site-packages/conan/tools/google/bazeldeps.py", line 129, in _get_dependency_buildfile_content
>     "libdir": cpp_info.libdirs[0],
> IndexError: list index out of range
> ```

Simple example to replicate this:

```
from conans import ConanFile
from conan.tools.google import BazelToolchain, BazelDeps


[...]

Issue #10476, opened 2022-02-01
[bug] Error while compiling with bazel: Error in glob: pattern cannot be absolute


mohabouje:
I reported a bug yesterday: https://github.com/conan-io/conan/issues/10471#issuecomment-1026297628

I've installed master with the fix and it seems to solve that issue.

I'm running into a different issue when using bazel now, It may not be related to that specific commit but its something that breaks in the commits in master planned for version 1.45, as it works just fine in version 1.44.1

### Environment Details (include every applicable attribute)
  * Operating System+version: Fedora release 34 (Thirty Four)
  * Compiler+version: gcc (GCC) 11.0.1 20210324 (Red Hat 11.0.1-0)
  * Conan version: Conan version 1.44.1 
  * Python version: Python 3.9.2

### Steps to reproduce (Include if Applicable)
```
from conans import ConanFile
from conan.tools.google import BazelToolchain, BazelDeps


class ExampleConanIntegration(ConanFile):
    name = "example"
    version = "0.0.1"
    generators = ['BazelDeps', 'BazelToolchain']
    requires = [
        'glew/2.2.0'
    ]

    def configure(self):
          pass

    def imports(self):
          pass
```
Then run:

```
conan install . --build=missing -s compiler=gcc -s compiler.version=11 -s compiler.libcxx=libstdc++11 -s build_type=Debug
```
Try to use the dependencies on any bazel project:

```
bazel build //...
```
It wont compile anymore

### Logs (Executed commands with output) (Include/Attach if Applicable)

```
ERROR: /home/mboujema/Developer/tijara/conandeps/glew/BUILD.bazel:5:10: //conandeps/glew:GLEWd_precompiled: invalid label '/home/mboujema/.conan/data/glew/2.2.0/_/_/package/44741af961cbe3bae7bd261b66459e5b6dc65693/lib/libGLEWd.a' in attribute 'static_library' in 'cc_import' rule: invalid target name '/home/mboujema/.conan/data/glew/2.2.0/_/_/package/44741af961cbe3bae7bd261b66459e5b6dc65693/lib/libGLEWd.a': target names may not start with '/'
ERROR: Traceback (most recent call last):
        File "/home/mboujema/Developer/tijara/conandeps/glew/BUILD.bazel", line 14, column 16, in <toplevel>
                hdrs = glob(["/home/mboujema/.conan/data/glew/2.2.0/_/_/package/44741af961cbe3bae7bd261b66459e5b6dc65693/include/**"]),
[...]

PR #10484, opened 2022-02-01
 Fix: Bazel builds generator

mohabouje:
Changelog: Fix: BazelDeps was generating invalid bazel files cause the static_library paths were absolute and not relative.
Changelog: Fix: BazelDeps was crashing when generating packages without libs cause the context was not checking the array size.
Docs: Omit

* BazelDeps was generating invalid bazel files cause the static_library paths were absolute and not relative
* BazelDeps was crashing when generating packages without libs cause the context was not checking the array size

I've tested this in my project using major libs like boost, qt, fmt etc. It works just fine with the latest commit

-  https://github.com/conan-io/conan/issues/10481
-  https://github.com/conan-io/conan/issues/10476

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -126,7 +126,7 @@ class BazelDeps(object):
         if len(cpp_info.libdirs) != 0:
             lib_dir = _relativize_path(cpp_info.libdirs[0], package_folder)
 
-        shared_library = dependency.options.get_safe("shared")
+        shared_library = dependency.options.get_safe("shared") if dependency.options else False
         context = {
             "name": dependency.ref.name,
             "libs": cpp_info.libs,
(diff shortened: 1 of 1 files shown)


conan/tools/google/bazeldeps.py
```diff
@@ -121,16 +121,23 @@ def _relativize_path(p, base_path):
             linkopts.append('"-l{}"'.format(linkopt))
         linkopts = ', '.join(linkopts)
 
+
+        lib_dir = 'lib'
+        if len(cpp_info.libdirs) != 0:
+            lib_dir = _relativize_path(cpp_info.libdirs[0], package_folder)
+
+        shared_library = dependency.options.get_safe("shared")
```
memsharded: Not all dependencies might have options. 
```suggestion
        shared_library = dependency.options.get_safe("shared") if dependency.options else False
```

mohabouje: Done

memsharded (approved):

memsharded:
Thanks for your contribution! It will be released in 1.46

PR #10494, opened 2022-02-02
package() and exports() with explicit copy

lasote:
- `package()` and `exports()` without `self.copy` with explicit copy tool and available explicit folders.
- Not modified "deploy" nor "imports", both keep the `self.copy`


conans/client/cmd/export.py
```diff
@@ -47,8 +48,10 @@ def cmd_export(app, conanfile_path, name, version, user, channel, graph_lock=Non
     mkdir(export_folder)
     mkdir(export_src_folder)
     origin_folder = os.path.dirname(conanfile_path)
-    export_recipe(conanfile, origin_folder, export_folder)
-    export_source(conanfile, origin_folder, export_src_folder)
+    setattr(conanfile, "conanfile_folder", origin_folder)
```
memsharded: Why a ``setattr``? Reads weird, why not ``conanfile.conanfile_folder = origin_folder``?

Also, why not using the already existing ``recipe_folder``?

conans/client/source.py
```diff
@@ -67,8 +67,10 @@ def get_sources_from_exports():
         if conanfile_folder != src_folder:
             _run_local_scm(conanfile, conanfile_folder, src_folder)
             conanfile.output.info("Executing exports to: %s" % src_folder)
-            export_recipe(conanfile, conanfile_folder, src_folder)
-            export_source(conanfile, conanfile_folder, src_folder)
+            setattr(conanfile, "conanfile_folder", conanfile_folder)
```
memsharded: Same as above

memsharded (commented):
Very nice, so explicit, great!

Just check the thing about "recipe_folder" please.

memsharded (approved):

PR #10497, opened 2022-02-02
Feature: Improve CMakeDeps config files (determine library type, fixes #10181)

### LINKED ISSUE #10181: [bug] `cmakedeps_macros.cmake` should use IMPORTED_IMPLIB on Windows
  * Conan version: 1.43

# `cmakedeps_macros.cmake`:

Line 39:

Instead of this

```cmake
add_library(${_LIB_NAME} UNKNOWN IMPORTED) // unknown is not good here anymore.
```

One should use something like this:
```cmake
add_library(${_LIB_NAME} SHARED IMPORTED) # Depening on what it is...
if(WINDOWS)
   set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_IMPLIB ${CONAN_FOUND_LIBRARY})
   set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_FOUND_LIBRARY}/../lib/lib.dll)
else()
endi()
```

This is necessary to make `$<TARGET_RUNTIME_DLLS:tgt>` work correctly. 
- https://gitlab.kitware.com/cmake/cmake/-/issues/22845
- See Question: https://github.com/conan-io/conan/issues/10120

gabyx:
Changelog: Bug: Runtime libraries should be set on Windows. 

Close #10181

WIP: I still need to look through the CMake again, to make sure everything is correct. Some of our DevOps are heavily reviewing CMake at the moment. Maybe they have some time.

- [x] ~I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one.~

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -131,7 +131,7 @@ class MacrosTemplate(CMakeDepsFileTemplate):
             # Set the libraries.
             set_target_properties(${target_name} PROPERTIES IMPORTED_LOCATION ${imported_loc})
             if(NOT imported_imp_lib STREQUAL "")
-                set_target_properties(${target_name} PROPERTIES IMPORTED_LOCATION ${imported_imp_lib})
+                set_target_properties(${target_name} PROPERTIES IMPORTEDIMP_LIB ${imported_imp_lib})
             endif()
 
         endfunction()
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -85,6 +85,8 @@ class MacrosTemplate(CMakeDepsFileTemplate):
         function(conan_define_package_library_target package_bindirs library_name target_name library_path)
 
             conan_determine_library_type(library_path type)
+
+            conan_message(DEBUG "Added library target '${target_name}'")
             add_library(${target_name} ${type} IMPORTED)
 
             set(imported_loc "")
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -87,8 +87,8 @@ class MacrosTemplate(CMakeDepsFileTemplate):
             conan_determine_library_type(library_path type)
             add_library(${target_name} ${type} IMPORTED)
 
-            set(imported_loc)
-            set(imported_imp_lib)
+            set(imported_loc "")
+            set(imported_imp_lib "")
             if(WIN32)
                 if(type STREQUAL "SHARED")
                     # Find a corresponding DLL in the binary directory (if possible)
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -31,6 +31,107 @@ class MacrosTemplate(CMakeDepsFileTemplate):
                 message(${MESSAGE_TYPE} "${MESSAGE_CONTENT}")
             endif()
         endfunction()
+        function(conan_determine_windows_library_type library_path out_type)
+            set(${out_type} UNKNOWN PARENT_SCOPE)
+
+            if(EXISTS "${library_path}")
+                if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.20.0)
+                    cmake_path(GET library_path EXTENSION ext)
+                    cmake_path(GET library_path STEM name)
+                else()
+                    get_filename_component(ext "${library_path}" EXT)
+                    get_filename_component(name "${library_path}" NAME_WE)
+                endif()
+
+                # Is the .lib file a dynamic library.
+                file(STRINGS "${library_path}" match REGEX "${name}\\.[dll|DLL]" LIMIT_COUNT 1)
+                if(NOT match STREQUAL "")
+                    set(${out_type} SHARED PARENT_SCOPE)
+                endif()
+            endif()
+        endfunction()
+
+        function(conan_determine_unix_library_type library_path out_type)
+            set(${out_type} UNKNOWN PARENT_SCOPE)
+            if(EXISTS "${library_path}")
+                if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.20.0)
+                    cmake_path(GET library_path EXTENSION ext)
+                else()
+                    get_filename_component(ext "${library_path}" EXT)
+                endif()
+
+                # Does it match a common Unix library extension.
+                # - A.so.1.3.0, A.so, A.dylib, A.dylib.1.3.0
+                if(ext MATCHES ".*\\.(a)(\\..+)$")
+                    set(${out_type} STATIC PARENT_SCOPE)
+                elseif(ext MATCHES ".*\\.(so|dylib)(\\..+)$")
+                    set(${out_type} SHARED PARENT_SCOPE)
+                endif()
+            endif()
+        endfunction()
+
+        function(conan_determine_library_type library_path out_type)
+            set(type UNKNOWN)
+            if(WIN32)
+                conan_determine_windows_library_type(${library_path} type)
+            else(UNIX)
+                conan_determine_unix_library_type(${library_path} type)
+            endif()
+
+            set(${out_type} ${type} PARENT_SCOPE)
+        endfunction()
+
+        function(conan_define_package_library_target package_bindirs library_name target_name library_path)
+
+            conan_determine_library_type(library_path type)
+            add_library(${target_name} ${type} IMPORTED)
+
+            set(imported_loc)
+            set(imported_imp_lib)
+            if(WIN32)
+                if(type STREQUAL "SHARED")
+                    # Find a corresponding DLL in the binary directory (if possible)
+                    # We search for the exact same name but with a .DLL/.dll ending
+                    # inside the package bin directory.
+                    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.20.0)
+                        cmake_path(GET library_path FILENAME name)
+                    else()
+                        get_filename_component(name "${library_path}" NAME)
+                    endif()
+
+                    string(REPLACE ".lib" ".dll" runtimeLib1 "${name}")
+                    string(REPLACE ".lib" ".DLL" runtimeLib2 "${name}")
+
+                    foreach(runtimeLib IN ITEMS "${runtimeLib1}" "${runtimeLib2}")
+                        find_file(found NAMES "${runtimeLib}" PATHS ${package_bindirs}
+                                NO_DEFAULT_PATH
+                                NO_PACKAGE_ROOT_PATH
+                                NO_CMAKE_PATH
+                                NO_CMAKE_ENVIRONMENT_PATH
+                                NO_CMAKE_SYSTEM_PATH
+                                NO_CMAKE_FIND_ROOT_PATH
+                                NO_CACHE)
+                        if(found)
+                            set(imported_imp_lib ${library_path})
+                            set(imported_loc ${found})
+                            conan_message(DEBUG "Runtime Library ${library_name} found '${found}'")
+                            break()
+                        endif()
+                    endforeach()
+                else()
+                    set(imported_loc ${library_path})
+                endif()
+            else(UNIX)
+                set(imported_loc ${library_path})
+            endif()
+
+            # Set the libraries.
+            set_target_properties(${target} PROPERTIES IMPORTED_LOCATION ${imported_loc})
+            if(NOT imported_imp_lib STREQUAL "")
+                set_target_properties(${target} PROPERTIES IMPORTED_LOCATION ${imported_imp_lib})
+            endif()
+
+        endfunction()
 
        macro(conan_find_apple_frameworks FRAMEWORKS_FOUND FRAMEWORKS FRAMEWORKS_DIRS)
            if(APPLE)
@@ -47,7 +148,7 @@ class MacrosTemplate(CMakeDepsFileTemplate):
            endif()
        endmacro()
 
-       function(conan_package_library_targets libraries package_libdir deps out_libraries out_libraries_target config_suffix package_name)
+       function(conan_package_library_targets libraries package_bindirs package_libdir deps out_libraries out_libraries_target config_suffix package_name)
            set(_out_libraries "")
            set(_out_libraries_target "")
            set(_CONAN_ACTUAL_TARGETS "")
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -65,8 +165,7 @@ def template(self):
                    set(_LIB_NAME CONAN_LIB::${package_name}_${_LIBRARY_NAME}${config_suffix})
                    if(NOT TARGET ${_LIB_NAME})
                        # Create a micro-target for each lib/a found
-                       add_library(${_LIB_NAME} UNKNOWN IMPORTED)
-                       set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_FOUND_LIBRARY})
+                       conan_define_package_library_target(${package_bindirs} ${_LIBRARY_NAME} ${_LIB_NAME} ${CONAN_FOUND_LIBRARY})
```
gabyx: That needs to be "${package_bindirs}" which is fixed in the refactoring after.

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -26,66 +26,179 @@ def context(self):
...
+                            set(imported_imp_lib ${library_path})
+                            set(imported_loc ${found})
+                            conan_message(DEBUG "Runtime Library ${library_name} found '${found}'")
+                        endif()
+                    endforeach()
+                else()
+                    set(imported_loc ${library_path})
+                endif()
+            else(UNIX)
+                set(imported_loc library_path)
```
A-A-G: ${library_path}

gabyx: Thx

SimonHeybrock:
I gave this branch a try with `IMPORTED_RUNTIME_ARTIFACTS` (see #10788), and the issue remains the same (only tried with TBB). I am not sure if the change here is supposed to fix this, or if the problem is not completely related?

memsharded:
I am checking this implementation, and it is true that this seems challenging on the CMake side. I am thinking if it would make sense to implement it in Python in the ``cpp_info`` side, which might also be a valuable information for other build systems.

I will discuss this possibility with the team.

gabyx:
Could you check if the target has been labled not UNKNOWN? Did not have time to test this PR completely. The generator expression: $<TARGET_RUNTIME_DLLS:tgt>

should give you all DLLs on Windows. On Linux, I guess its the right way how you do it. Do you have a minimal working example in CMake I could test? 

Von meinem iPhone gesendet

SimonHeybrock:
Here is a minimal working example:

In a directory add these files:

`CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.21)
project(dummy LANGUAGES CXX)
set(CMAKE_BUILD_TYPE
    "Release"
    )

if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/conan.cmake")
  message(
    STATUS
      "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
  file(
    DOWNLOAD
    "https://raw.githubusercontent.com/conan-io/cmake-conan/0.17.0/conan.cmake"
    "${CMAKE_CURRENT_BINARY_DIR}/conan.cmake"
    EXPECTED_HASH
      SHA256=3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef
    TLS_VERIFY ON)
endif()

include(${CMAKE_CURRENT_BINARY_DIR}/conan.cmake)
list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_BINARY_DIR})
list(APPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_BINARY_DIR})

conan_cmake_autodetect(conan_settings)
conan_cmake_install(PATH_OR_REFERENCE ${CMAKE_CURRENT_SOURCE_DIR} SETTINGS
                    ${conan_settings} BUILD outdated)

find_package(TBB CONFIG REQUIRED)
get_target_property(TBB_LIBRARY_TYPE TBB::tbb TYPE)
message(STATUS "TBB_LIBRARY_TYPE: ${TBB_LIBRARY_TYPE}")
install(IMPORTED_RUNTIME_ARTIFACTS TBB::tbb RUNTIME DESTINATION ${PYTHONDIR})

message(STATUS "Using TBB from: ${TBB_DIR}")
```

`conanfile.txt`:
```
[requires]
onetbb/2021.3.0

[generators]
cmake_find_package_multi
```

```sh
mkdir build
mkdir install
cd build
cmake -DCMAKE_INSTALL_PREFIX=../install ..
cmake --build . --target install
ls -R ../install/
```

Using, say, `tbb-devel` from `conda-forge`, removing the `onetbb` line in `conanfile.txt` this works. With TBB from `conan-1.46.0` (tried on Ubuntu) I get the error:

```
-- TBB_LIBRARY_TYPE: INTERFACE_LIBRARY
CMake Error at CMakeLists.txt:31 (install):
  install IMPORTED_RUNTIME_ARTIFACTS given target "TBB::tbb" which is not an
  executable, library, or module.
```

gabyx:
Ok, the following works. I hade a few errors and typos in the PR still. (CMake is an ugly language...)

**Note** @SimonHeybrock You need the [`CMakeDeps`](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html) generator.

It works, but only with the hidden target `CONAN_LIB::onetbb_tbb_RELEASE`. The `onetbb::onetbb` is a `INTERFACE_TARGET`. I guess getting to the internal `CONAN_LIB::onetbb_tbb_RELEASE` is not that easy starting from `onetbb::onetbb` since that involves traversing the `LINK_LIBRARIES` properties by evaluating generator expressions... etc... (it can be done but is ugly...)

```
mkdir build && cd build &&
cmake --log-level debug -DCMAKE_INSTALL_PREFIX=../install ..
```

`CMakeLists.txt:`

```cmake
cmake_minimum_required(VERSION 3.21)
project(dummy LANGUAGES CXX)
set(CMAKE_BUILD_TYPE
    "Release"
    )

if(NOT EXISTS "${CMAKE_CURRENT_BINARY_DIR}/conan.cmake")
  message(
    STATUS
      "Downloading conan.cmake from https://github.com/conan-io/cmake-conan")
  file(
    DOWNLOAD
    "https://raw.githubusercontent.com/conan-io/cmake-conan/0.17.0/conan.cmake"
    "${CMAKE_CURRENT_BINARY_DIR}/conan.cmake"
    EXPECTED_HASH
      SHA256=3bef79da16c2e031dc429e1dac87a08b9226418b300ce004cc125a82687baeef
    TLS_VERIFY ON)
endif()

include(${CMAKE_CURRENT_BINARY_DIR}/conan.cmake)
list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_BINARY_DIR})
list(APPEND CMAKE_PREFIX_PATH ${CMAKE_CURRENT_BINARY_DIR})

conan_cmake_autodetect(conan_settings)
conan_cmake_install(PATH_OR_REFERENCE ${CMAKE_CURRENT_SOURCE_DIR} SETTINGS
                    ${conan_settings} BUILD outdated)

find_package(onetbb CONFIG REQUIRED)
install(IMPORTED_RUNTIME_ARTIFACTS CONAN_LIB::onetbb_tbb_RELEASE
        RUNTIME DESTINATION ${CMAKE_CURRENT_BINARY_DIR}/externals)
```

`coannefile.py`
```python
from conans import ConanFile
from conan.tools.cmake import CMakeDeps

required_conan_version = ">=1.43.0"

class TestLib(ConanFile):
  name = "base"

  description = """This is a test project with a library base::io
    and base::math and an executable cli."""

  license = "MIT"
  build_policy = "missing"  # if this package is build by default if missing.
  version = "1.0.0"

  def requirements(self):
    self.requires("onetbb/2021.3.0")

  def generate(self):
    deps = CMakeDeps(self)
    deps.generate()
```

@memsharded: I think the refactoring in the PR is valuable eitherway if this solution gets in or not.
I guess one can later on replace/enhance it with the direct way by providing `STATIC`/`SHARED` settings from the dependencies `conanfile.py` etc...
Do you have any news on that?

SimonHeybrock:
@gabyx I am trying this now, but I get exactly the same error (about `CONANLIB::onetbb_tbb_RELEASE`). Do I need this branch, or does it work with the released conan?

SimonHeybrock:
Scratch the above. Running cmake with *this branch* works. However, the `install` target does not appear to install all required files? Only `libtbb.so` (which is a link to `libtbb.so.12`, which does not exist) is installed, i.e., I only get a dead link in the install folder after running `cmake --build . --target=install`.

gabyx:
> @gabyx I am trying this now, but I get exactly the same error (about `CONANLIB::onetbb_tbb_RELEASE`). Do I need this branch, or does it work with the released conan?

Of course it does only work with this branch:

` pip install git+https://..gabyx...@feature/improve-cmakedeps`

gabyx:
> Scratch the above. Running cmake with _this branch_ works. However, the `install` target does not appear to install all required files? Only `libtbb.so` (which is a link to `libtbb.so.12`, which does not exist) is installed, i.e., I only get a dead link in the install folder after running `cmake --build . --target=install`.

Oh really, hm... thats weird because it should be CMake's job to analyze the target (`ldd` and sorts) and follow symlinks (?). Actually this branch only helps, because the target type is no more `UNKNOWN` and CMake can do its things with the functionality of recursively detection the dependencies. One can also use the following for a built executable for example:

https://cmake.org/cmake/help/latest/command/install.html#installing-targets
```cmake
install(TARGET executableTarget RUNTIME_DEPENDENCIES DESTINATION ${runtimeDir}))
```

SimonHeybrock:
I tried the latter, and that does not install *anything* related to TBB or onetbb, not even the dangling symlink.

gabyx:
@SimonHeybrock You need to have a target which is linked with tbb. But the install will fail, because the search paths to find the tbb lib are not set (e.g. `~/.conan/cache/...`) which is possible by reading from the the output of `virtualrunenv` generator of `conan install -g=virtualrunenv` and using this in the add. settings `DIRECTORIES` for `RUNTIME_DEPENDENCIES` in `install()`

SimonHeybrock:
@gabyx Thanks for your effort, but I don't think I am following any more. How would I use this in the cmake-conan example?

gabyx:
I try to clarify again:

Status quo: Conan config files (.cmake) are setting up targets with Type `UNKNOWN`. That makes it hard or impossible to use certain features like `$<TARGET_RUNTIME_DLLS:tgt>`.

- This PR corrects this by setting up the targets as `STATIC` or `SHARED` depending on what CMake detects on the file extension (`.so,.dll` etc, that should be provided somewhen by conan itself...). Also it sets correct `.lib`/`.dll` to target properties on Windows, which it did not before.

- @SimonHeybrock: I guess your motivation was just to get dependencies installed too, this can be done independently of this PR, by `install(TARGETS ...  RUNTIME_DEPENDENCIES)` or `file(GET_RUNTIME_DEPENDENCIES)` (see my example above). This functionality acts on the binary of a target/file to recursively determine all shared librbaries which this binary links. This is related to this PR as targets probably need to be setup correctly, to make these function work correctly or more proper. These functions can use additional search `DIRECTORIES ...`, thats where you would enter the output from `virtualrunenv` generator from Conan, we use this inside `conanfile.py::generat(self)` :

```python
def generate_run_environment(conanFile: ConanFile):
  """ Generates the virual run environment scripts together with an
      `conanrun.env` file which can be used in debuggers supporting env. files.
  """
  # Write 'conanrun.env' file.
  runEnv = VirtualRunEnv(conanFile)
  with open(path.join(conanFile.folders.generators, "cononrun.env"), "w") as f:
    for k, v in runEnv.vars(scope="run").items():
      f.write(f"{k}=\"{v}\"\n")

  # Write other run environment scripts.
  runEnv.generate()
```

This will give you a `conanrun.env` file (you can generate also `json` in the above python) (on `conan install`) which you can parse and use in CMake for exactly this `DIRECTORIES` option to make searching shared libs successful.

SimonHeybrock:
I wanted to report back, to confirm that this works now with this branch. The crucial points for me were (using cmake):

- Use `install` with `RUNTIME_DEPENDENCIES` option. I could not get `install(IMPORTED_RUNTIME_ARTIFACTS)` to work.
- On Linux and MacOS this is sufficient, provided that the rpath of the installed targets is setup correctly. The dependencies get found and installed with the library.
- On Windows rpath is not used. As @gabyx pointed out we can use the `DIRECTORIES` parameter to `RUNTIME_DEPENDENCIES`. @gabyx suggested to use `conan install -g=virtualrunenv` to find the path. Here is a simpler approach (at least simpler for me with cmake):
  - Add the `deploy` generator (e.g., to `conanfile.txt`). This will install the dependencies into `${CMAKE_BINARY_DIR}`
  - Use `install` with `RUNTIME_DEPENDENCIES DIRECTORIES ${CMAKE_BINARY_DIR}/onetbb/bin  ${CMAKE_BINARY_DIR}/onetbb/lib` (replace `onetbb` with the name of your dependency). Not sure if we actually need both `bin` and `lib`.
- To avoid `RUNTIME_DEPENDENCIES` installing a lot of system libraries (which cases trouble, e.g., when building a wheel for PyPI), I used a combination of the `PRE_INCLUDE_REGEXES` and `PRE_EXCLUDE_REGEXES` parameters of `RUNTIME_DEPENDENCIES`.

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -26,66 +26,182 @@ def context(self):
...
+                endif()
+            else(UNIX)
+                set(imported_loc ${library_path})
+            endif()
+
+            # Set the libraries.
+            set_target_properties(${target_name} PROPERTIES IMPORTED_LOCATION ${imported_loc})
+            if(NOT imported_imp_lib STREQUAL "")
+                set_target_properties(${target_name} PROPERTIES IMPORTED_LOCATION ${imported_imp_lib})
+            endif()
```
gabyx: @lasote Thats the part I mean.

SimonHeybrock:
What is the status of this? It would be useful if such a change could make it into conan.

Thanks,
Simon

lasote:
It has been replaced with [this approach](https://github.com/conan-io/conan/pull/11111) using Conan 2.0 features

SimonHeybrock:
I just tried conan-1.51.3 and the solution in my comment https://github.com/conan-io/conan/pull/10497#issuecomment-1080355878 works out of the box. Apparently changes unrelated to the branch in this PR fixed the issue.

lasote:
Yes, the underlying issue was fixed with other changes too. Thanks!

Issue #10506, opened 2022-02-03
[feature] "conan deploy" command


lasote:
See (*1) from https://github.com/conan-io/conan/issues/10505
Kind of the same idea behind https://github.com/conan-io/conan/pull/9273/files

memsharded:
Implemented by https://github.com/conan-io/conan/pull/10507

Issue #10479, opened 2022-02-01
[question] Ship buildable-from-sources projects


vdsbenoit:
### Context
We use conan to package our company libraries. We have a private conan registry where we store them. Some of our libraries depends on others of our private libraries. 

### Issue
In some cases, we need to share projects with collaborators who are external to the company. These collaborators do not have access to our conan registry. 

### Overall question
If we need to share a project that depends on some other internal projects (i.e. conan dependencies) to an external collaborator, how could we ship our project in a self-contained way? I mean, how could we enable our collaborators to build from source ? 

Prior using conan, we used to attach the dependencies inside the repository and configure CMake to use them. 

### External registry
We could upload our libraries to an external private registry the collaborators have access to.

**Outcome:** some legal reasons, we cannot upload our private conan packages to an external registry. 

### conan install & conan build
I thought about installing the dependencies of my project in a `.build` directory:
````bash
conan install . --install-folder .build 
````
Then ship my project to our collaborators with the `.build` directory and ask them to run
````bash
conan build . --build-folder .build
````

**Outcome:** it does not work. The files in `.build` point to file from the conan cache, which is local.

### Deploy generator
I then thought the [deploy generator](https://docs.conan.io/en/latest/reference/generators/deploy.html) could be a good track. Thus I tried 
````bash
conan install . --install-folder .build -g=deploy
````

**Outcome:** it does not work. The `.build/conanbuildinfo.cmake` file still contains paths to the conan cache, which is local.

### VirtualEnv
I read about the [Virtual Environments](https://docs.conan.io/en/latest/mastering/virtualenv.html#virtual-environment-generator).

**Outcome:** the virtual environments do not bundle the packages. They point to the conan cache too.

### Conclusion

A solution could be an option that creates a kind of self-contained virtual env. I mean something similar to the current virtual env but that would point to binaries included in the virtual env directory.

## Other existing GitHub issues

I have searched for other issues on this topic. I found #1061. However, the goal there is to ship self-contained binaries, not the source code. 

---

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Hi @vdsbenoit 

We are aware of this. One of the root causes is that there is no general, robust, collision-free "deploy" output. The only place where binaries will certainly not conflict is the Conan cache.

[...]

PR #10507, opened 2022-02-03
New Conan 2.0 deployers

memsharded:
Close https://github.com/conan-io/conan/issues/10506

New approach for deploy of files:

- ``conan install --deploy=mydeploy`` will search for a ``mydeploy.py`` in this order:
  - First in local folder. ``--deploy=folder/mydeploy`` is valid
  - Then in local cache ``extensions/deploy`` folder
  - The .py extension can be added to command line too
- ``conan install --deploy=conan_xxxx`` is for built-in deployers. ``conan_full_deploy`` one is proposed
- More than 1 deployer can be specified
- Deployers always run before generators, and only in the consumer side
- Deployers can change the folder, so generators that execute after will point to the deployed folder, not the cache
- Deployers might be challenging to code that take into account packages in editable mode. Initially deployers will not work with editables.
-


conans/test/functional/command/test_install_deploy.py
```diff
@@ -0,0 +1,26 @@
+from conans.test.assets.cmake import gen_cmakelists
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.utils.tools import TestClient
+
+
+def test_install_deploy():
```
lasote, line 14: missing cmake tool

lasote, line 14:
```suggestion
@pytest.mark.tool("cmake")
def test_install_deploy():
```

conans/test/functional/command/test_install_deploy.py
```diff
@@ -0,0 +1,26 @@
+from conans.test.assets.cmake import gen_cmakelists
```
lasote, line 6:
```suggestion
@pytest
import pytest

from conans.test.assets.cmake import gen_cmakelists
```

conans/cli/api/subapi/install.py
```diff
@@ -59,3 +62,30 @@ def install_consumer(deps_graph, generators=None, no_imports=False, deploy=False
...
+
+        conanfile.folders.set_base_source(source_folder)
+        conanfile.folders.set_base_imports(output_folder)
+        conanfile.folders.set_base_generators(output_folder)
+
+        if type(conanfile).system_requirements != ConanFile.system_requirements:
+            call_system_requirements(conanfile)
+
+        for r, d in conanfile.dependencies.items():
+            new_folder = os.path.join(output_folder, d.ref.name)
```
lasote: To make it collision-safe and multi-config?
```suggestion
            new_folder = os.path.join(output_folder, d.ref.name, str(d.ref.version))
            if conanfile.settings.get_safe("arch"):
                new_folder = os.path.join(new_folder, str(conanfile.settings.arch))
            if conanfile.settings.get_safe("build_type"):
                new_folder = os.path.join(new_folder, str(conanfile.settings.build_type))
```

conans/cli/api/subapi/install.py
```diff
@@ -59,3 +62,30 @@ def install_consumer(deps_graph, generators=None, no_imports=False, deploy=False
             deploy_conanfile.folders.set_base_imports(output_folder)
             if hasattr(deploy_conanfile, "deploy") and callable(deploy_conanfile.deploy):
                 run_deploy(deploy_conanfile, output_folder)
+
+    @staticmethod
+    def deploy_consumer(deps_graph, generators=None, source_folder=None, output_folder=None):
+        """ Once a dependency graph has been installed, there are things to be done, like invoking
+        generators for the root consumer, or calling imports()/deploy() to copy things to user space.
+        This is necessary for example for conanfile.txt/py, or for "conan install <ref> -g
+        """
```
lasote: Copy pasted?

conans/cli/api/subapi/install.py
memsharded: This is the POC. The idea is that this will not be built-in here, but on the user side, something in the line of the ``layout()`` and we could provide something like ``cmake_layout()``, as ``typical_deploy()``. This is basically one of the points that we should discuss around this feature.

vdsbenoit:
> What happens with multi-config, like Debug/Release? It is necessary to define some extra logic for this.

I take the liberty to share some suggestions. Feel free to do what you want with it. 

### Mutli-config

Couldn't it be left at the developers' convenience ? I mean, the developers could explicitly specify the location where to install the binaries.

I actually [faced this situation](https://github.com/conan-io/conan/issues/10479) (i.e. trying to install the dependencies of my project in a self-contained way). I was about to do what follows, before I realize the `conanbuildinfo.cmake` created by the deploy generator actually points to to the conan cache.

````bash
conan install . -g deploy --install-folder .build/Release-linux-x64-gcc7 --profile linux-gcc7 --build_type=Release
conan install . -g deploy --install-folder .build/Debug-linux-x64-gcc7 --profile linux-gcc7 --build_type=Debug
````

In order to have a `.build` directed structured by configuration. I would then run my build with the appropriate conan build command, for instance:
````bash
# for the Release configuration
conan build . --build-folder .build/Release-linux-x64-gcc7
# for the Debug configuration
conan build . --build-folder .build/Debug-linux-x64-gcc7
````
That would enable me to share my repository with a collaborator who does not have access to the the dependencies (e.g. private conan repository).


### Cross compilation
Regarding cross-compilation, let's take the same example as above and assume that `protobuf` is defined both in `tool_requires` and `requires`.

````bash
conan install . -g deploy --install-folder .build/Release-android-armv8-clang7 --profile:build default --profile:host android-armv8-clang7 --build_type=Release
````

Conan could make 2 subdirectories:
````
.build/Release-android-armv8-clang7/BUILD/protobuf
.build/Release-android-armv8-clang7/HOST/protobuf
````
or conan could add a suffix to the directories containing the dependencies:
````
.build/Release-android-armv8-clang7/protobuf-BUILD
.build/Release-android-armv8-clang7/protobuf-HOST
````

**Comment by memsharded** [comment]
> Couldn't it be left at the developers' convenience ? I mean, the developers could explicitly specify the location where to install the binaries.

Yes, I think this is exactly the key that will enable this possibility that it has not been possible so far: the deploy "layout" should be defined by the users, and not built-in by Conan. The UI/UX to do so, is something to discuss, but you are right that is possible the core idea.

**Comment by memsharded** [comment]
As the next step, I am proposing:

- Making the "deploy" logic external. In the test, just a ``--deploy=deploy.py`` file in the user folder.
- We could of course have some built-in default one (or several), that could be be called by ``--deploy=some_builtin_name``.
- We can have them installed with ``conan config install`` in the cache too.

Feedback of this new change welcome

**Comment by lasote** [comment]
> - We can have them installed with ``conan config install`` in the cache too.

Do we currently have a folder where to look for raw python files? or only hooks at the moment?

**Review by lasote — COMMENTED** [review]
I like the general idea, very flexible.

**Comment by memsharded** [comment]
Note this approach is super close to generators (and custom generators, which is something to discuss for some of the non ported yet 1.X generators, ``ycm``, ``markdown``), but with a couple of differences:

- It is critical that this run **before** other generators like ``CMakeDeps``, because they can change the final folder where the binaries are
- The new ``conanfile.set_deploy_folder()`` is key to this approach, as it is the piece that allows to relocate the final artifacts in the user folder

We have consider the name ``extensions`` to name things like these deployers or the ``compatibility.py`` user logic (in the cache). What will be the command line? Some options:

```bash
conan install ...  --deploy=ycm --deploy=standard_copy_deploy  # Note that copy_deploy should still run before ycm
conan install ...   --generator=ycm --deploy=standard_copy_deploy
```

Some ideas:
- more than 1 deployer and more than 1 generator can be allowed
- custom generators are only valid in the end-consumer case, for command line. Not in recipes by name
- deployers by default will be looked for in .conan/extensions/deploy/mydeploy.py
- generators could be in .conan/extensions/generators/ycm.py

vdsbenoit:
> Making the "deploy" logic external. In the test, just a --deploy=deploy.py file in the user folder.

What would be the content of the `deploy.py` file? The content of what would have been in the `deploy()` method of the `conanfile.py` ?

Would this solution be compatible with the `deploy` generator? (i.e. providing an `install-folder` for the package content & its dependencies). 

If yes, would the paths in the `conanbuildinfo.cmake` file point to the installation folder instead of the conan cache ?

memsharded:
> What would be the content of the deploy.py file? The content of what would have been in the deploy() method of the conanfile.py ?

Not exactly, we are making things more explicit here, check the unit test:

```python
import os, shutil

def deploy(conanfile, output_folder):
    for r, d in conanfile.dependencies.items():
        new_folder = os.path.join(output_folder, d.ref.name)
        shutil.copytree(d.package_folder, new_folder)
        d.set_deploy_folder(new_folder)
```


> Would this solution be compatible with the deploy generator? (i.e. providing an install-folder for the package content & its dependencies).

Yes, it will actually use the ``--output-folder`` cli arg. Not compatible with the ``deploy`` generator, but basically the above implements the behavior of the current deploy generator

> If yes, would the paths in the conanbuildinfo.cmake file point to the installation folder instead of the conan cache 

``conanbuildinfo.cmake`` is removed, but the new generators like ``CMakeDeps`` and ``CMakeToolchain``, yes, they will point to the copy deployed to the ``output-folder``. The unit test is doing a real build against it, using the generated files from that integration, it is a fully functional test of real building against the deployed copy.

vdsbenoit:
Alright, thank you for the clarification. That looks good to me !

conans/cli/commands/install.py
```diff
@@ -143,7 +143,8 @@ def install(conan_api, parser, *args):
     parser.add_argument("-of", "--output-folder",
                         help='The root output folder for generated and build files')
     parser.add_argument("-sf", "--source-folder", help='The root source folder')
-
+    parser.add_argument("--deploy", action=Extender,
+                        help='Deploy using the provided deployer to the output folder')
```
franramirez688, line 147: I think we should add examples of possible valid `deployer` values, e.g., `conan_full_deploy`, `mydeploy.py`, `my/path/to/my/deployer`, etc.

conans/cli/api/subapi/install.py
```diff
@@ -40,13 +42,64 @@ def install_consumer(deps_graph, generators=None, source_folder=None, output_fol
...
             call_system_requirements(conanfile)
+
+
+def _do_deploys(conan_api, conanfile, deploy, output_folder):
+    # Handle the deploys
+    cwd = os.getcwd()
+    for d in deploy or []:
+        # First, find the file containing the deployer, could be local or in the cache
+        if d.startswith("conan"):  # built-in!
+            deployer = {"conan_full_deploy": conan_full_deploy}[d]
```
lasote: I don't understand why that `startswith` and the dictionary, why not just check with `d=="conan_full_deploy"`?

lasote (approved):

conans/cli/api/subapi/install.py
memsharded: I wanted to have an explicit way of differentiating a built-in one of an external one, and reserve the name prefix ``conan_`` for it, with the current implementation it will fail for any ``--deploy=conan_xxxx`` if users decide to name ``conan_xxxx`` to their own deployers.

conans/cli/api/subapi/install.py
```diff
@@ -38,9 +42,64 @@ def install_consumer(deps_graph, generators=None, source_folder=None, output_fol
...
```
franramirez688: what about if instead of using `conan_full_deploy` we'd use `default`? I think it's enough to put `default` to suppose that we want to use the Conan one, isn't it?

conans/model/build_info.py
```diff
@@ -363,6 +363,19 @@ def set_relative_base_folder(self, folder):
...
+        for component in self.components.values():
+            for varname in _DIRS_VAR_NAMES:
+                origin = getattr(component, varname)
+                if origin is not None:
+                    new_ = []
+                    for el in origin:
+                        rel_path = os.path.relpath(el, package_folder)
+                        new_.append(os.path.join(deploy_folder, rel_path))
+                    origin[:] = new_
+                # TODO: Missing properties
```
franramirez688, line 377: I think we should refactor this method and the `set_relative_base_folder()` one in the future. They're very similar and we could have a single method to set a general base folder.

conans/test/functional/command/test_install_deploy.py
```diff
@@ -0,0 +1,129 @@
...
+    """ check that we can also deploy a reference
+    """
+    c = TestClient()
+    c.save({"conanfile.py": GenConanfile("pkg", "1.0").with_package_file("include/hi.h", "hi")})
+    c.run("create .")
+
+    c.run("install  --reference=pkg/1.0 --deploy=conan_full_deploy --output-folder=output")
+    # NOTE: Full deployer always use build_type/arch, even if None/None in the path, same structure
+    header = c.load("output/host/pkg/1.0/None/None/include/hi.h")
+    assert "hi" in header
```
franramirez688, line 129: Should we have a test without passing the `--output-folder` param?

conans/cli/api/subapi/install.py
memsharded: I think we will have a couple of different deployers, like one "conan_runtime_deploy" to install only the shared libs and exes, or a "conan_license_deploy" to collect licenses.

conans/cli/commands/install.py
memsharded: But adding examples in the cli contextual reference would be a bit noisy? I can add them, sure, just wondering if it is really expected or more a docs thing

conans/model/build_info.py
memsharded: They *look like* very similar but they are not. This one is making the folder relative, then absolute to the new one. It takes 2 arguments, while the other take 1. Not sure if refactoring and making common would end in a "frankestein monster" method instead, driven by "ifs" instead by semantic method call.

conans/cli/api/subapi/install.py
```diff
@@ -38,9 +42,64 @@ def install_consumer(deps_graph, generators=None, source_folder=None, output_fol
...
+        build_type = str(d.info.settings.build_type)
+        arch = str(d.info.settings.arch)
+        if build_type:
+            folder_name = os.path.join(folder_name, build_type)
+        if arch:
+            folder_name = os.path.join(folder_name, arch)
+        new_folder = os.path.join(output_folder, folder_name)
+        # TODO: This deployer doesn't work with editable (and it would be difficult)
+        #  we could raise an error if editable, but not easy to know it, not in ConanFileInterface
+        shutil.copytree(d.package_folder, new_folder)
```
czoido: Running the deploy twice, over the same location is going to fail because the folders already exist

conans/cli/api/subapi/install.py
lasote: What about rising an Exception saying "The name conan is reserved", that could be useful for the user and for the developer :P

czoido: If we have an order of resolution of the deployers: first local folder, then local cache in `/extensions/deploy` do we really want to reserve `conan_` for builtin? Why not call it `full_deploy` and in case someone adds `full_deploy.py` to local folder resolve in that order, with preference to the local `full_deploy.py` over the builtin...
WDYT?

conans/cli/api/subapi/install.py
memsharded: Yes, good point. And probably that is what we want unless you pass a ``--force`` arg? We have to define the UX around this use case, indeed.

conans/cli/api/subapi/install.py
memsharded: Not sure about that. If you have your own custom deployer, why name it exactly the same as one documented on in the Conan docs, but it will have a different behavior? For me it is only going to cause more struggles trying to understand what is happening, someone reports: "I have an issue, I am running ``conan install ... --deploy=full_deploy`` and it is failing..." And now we don't know if they are replacing the built-in one with a local or a cache deployer of their own.

conans/cli/api/subapi/install.py
czoido: Yes, I think we should decide what Conan should do in these cases and make it behave coherently for all commands...

conans/cli/api/subapi/install.py
czoido: Yes, creating a deployer which name is exactly the same as the builtin is something a user should not do... 
The thing is that for other extensions/templates we don't add the `conan_` prefix to reserve, like for example using `conan_cmake_lib` or `conan_meson_exe` for `conan new`... if we are reserving that prefix for builtins, maybe we should do for everything, but if we have a clear, documented order of resolution I don't think it's necessary.
And you can always protect your own deployers naming them with `user_deploy_full.py`

conans/cli/api/subapi/install.py
franramirez688: Indeed, you can specify it in the docs, but I think it could be important to put a couple of examples here just to demonstrate what are the most common values.

conans/model/build_info.py
franramirez688: agree

conans/cli/commands/install.py
franramirez688: Indeed, you can specify it in the docs, but I think it could be important to put a couple of examples here just to demonstrate what are the most common values.

conans/cli/api/subapi/install.py
memsharded: It is a bit variable:

- For files, like the ones generated by generators like CMakeDeps, they simply overwrite each time, and it is necessary they do, otherwise in practice it would be really bad UX, and --force would be mandatory
- The ``conan new`` command doesn't overwrite, unles ``--force``, because it is likely that user did local changes to the created files that would be lost otherwise. And it doesn't seem a common use case to be calling ``conan new`` repeteadly
- ``conan profile detect`` also will fail if the file already exist, for the same reason: it could be losing user changes

For this case, as all files are just a copy from the conan cache, I'd say overwriting by default should be ok, but lets discuss it.

czoido: Yes, for this case it should probably behave like the generators. In case you don't want to behave that way the deploy you can always implement other logic in your custom deployer...

czoido (approved):
Very cool feature!

conans/cli/api/subapi/install.py
memsharded: Ok, implementing overwrite of existing deployed files (this is responsibility of the deployer, doing it in the ``conan_full_deploy`` deployer.

conans/cli/api/subapi/install.py
memsharded: Renamed to "full_deploy" and changed the order

franramirez688 (approved):

conans/cli/api/subapi/install.py
```diff
@@ -89,17 +100,29 @@ def conan_full_deploy(conanfile, output_folder):
...
-    conanfile.output.info(f"Conan built-in full deployer to {output_folder}")
-    for r, d in conanfile.dependencies.items():
-        folder_name = os.path.join(d.context, d.ref.name, str(d.ref.version))
-        build_type = str(d.info.settings.build_type)
-        arch = str(d.info.settings.arch)
+    root_node = graph.root
+    root_conanfile = root_node.conanfile
+
+    root_conanfile.output.info(f"Conan built-in full deployer to {output_folder}")
+    for node in graph.nodes:
```
lasote: We cannot expect the user to iterate the graph nodes to do the deployers, we need an interface through conanfile.dependencies and skip that logic of virtual, consumers, binary skip and editables.

conans/model/conan_file.py
```diff
@@ -237,3 +237,7 @@ def test(self):
 
     def __repr__(self):
         return self.display_name
+
+    def set_deploy_folder(self, deploy_folder):
```
lasote, line 241: I like better. 👍

conans/test/functional/command/test_install_deploy.py
```diff
@@ -20,8 +20,8 @@ def test_install_deploy():
         import os, shutil
 
         # USE **KWARGS to be robust against changes
-        def deploy(conanfile, output_folder, **kwargs):
-            for r, d in conanfile.dependencies.items():
+        def deploy(graph, output_folder, **kwargs):
```
lasote: definitely not suitable to expose the graph.

lasote (changes requested):

lasote (approved):

PR #10516, opened 2022-02-04
[2.0] Remove imports deploy

### LINKED ISSUE #10505: [feature] 2.0 Remove "imports" and "deploy" mechanism
We think that the imports use cases can be solved with other mechanisms:

`imports()` method:
1. Import a tool from a dependency to run it, like cmake: Superseded by virtual environments.
2. Import shared libraries to run something in the build method. Superseded by virtual environments, and also broken because if you want to build a package locally you want to import the artifacts at "generate" time, not just before calling "build()"
3. Import any asset from dependencies necessary for building (same case 2)

**Both could be done using the new `copy` tool and iterating the dependencies. Include a POC to see how it could look (access to bindirs, libdirs...) and maybe propose some tool to help?**

`imports()` command:
1. Bring some things to the local directory, shared, or tools or whatever. But we were missing the very requested feature of installing the packages from the dependencies in a local directory but pointing the generators to these local folders, so you can distribute all as a zip (*1)

`deploy()` method
Kind of mixed-use case where you also can copy files from the package itself. 

**We will try to implement something in the direction of (\*1) in a different issue**

lasote:
- Removed imports() method, import layout, the import command
- Removed deploy() method and mechanism.

Close #10505 

PENDING TO DOCUMENT: 
  - Migration guide of import => use copy and iterate dependencies, examples. 
```python
def generate(self):
    # Simulate an "imports"
    for dep in self.dependencies.values():
        dest_folder = os.path.join(self.build_folder, dep.ref.name)
        copy(self, "myfile.txt", dep.package_folder, dest_folder)

```
  - Migration guide of deploy? Probably the @memsharded PR.


conans/client/loader_txt.py
```diff
@@ -8,7 +7,7 @@ class ConanFileTextLoader(object):
     def __init__(self, input_text):
         # Prefer composition over inheritance, the __getattr__ was breaking things
         self._config_parser = ConfigParser(input_text,  ["requires", "generators", "options",
-                                                         "imports", "tool_requires"],
+                                                         "tool_requires"],
```
memsharded: Maybe don't remove this one, so parsing doesn't crash for 1.X-2 compatible conanfile.txt?

conans/client/manager.py
```diff
@@ -3,26 +3,22 @@
...
-from conans.client.importer import run_imports, run_deploy
 from conans.client.installer import BinaryInstaller, call_system_requirements
 from conans.model.conan_file import ConanFile
-from conans.model.recipe_ref import RecipeReference
 
 # FIXME: this is duplicated in the new API until all commands that use this function are migrated
 #  this should be replaced by a call to APIV2.graph.load_graph + APIV2.install.install_binaries
 
 
 def deps_install(app, ref_or_path, base_folder, profile_host, profile_build,
```
memsharded: deps_install method might be not used anymore?

conans/model/conan_file.py
```diff
@@ -165,10 +165,6 @@ def package_folder(self):
     def generators_folder(self):
         return self.folders.generators_folder
 
-    @property
-    def imports_folder(self):
```
memsharded, line 169: I think this might be a bit painful for 1.X -> 2 migrations, but lets move forward and we will see.

conans/test/integration/lockfile/test_ci.py
```diff
@@ -14,12 +15,20 @@ class Pkg(ConanFile):
         settings = "os"
         {requires}
         exports_sources = "myfile.txt"
-        keep_imports = True
-        def imports(self):
-            self.copy("myfile.txt", folder=True)
+
+        def generate(self):
+            # Simulate an "imports"
+            for dep in self.dependencies.values():
```
memsharded, line 21: Nice

memsharded (approved):
Fantastic

czoido (approved):

conans/model/conan_file.py
lasote: I think, migrating the copy tool and with "generate" it shouldn't be that hard.

memsharded (approved):

Issue #10503, opened 2022-02-03
[feature] scm migrate to "conan" namespace


lasote:
- Simplify architecture (superclasses and other magic)
- Remove the "no-auto" use case: Defining fixed fields just to avoid doing a manual clone/checkout. Deprecate in the migration guide the usage of revision != "auto" or whatever is needed.
- Expose simple tools (moving to conan) for Git, SVN, and internally, for the scm feature, use whatever we need (do not break).

memsharded:
This has been already done: Conan 1.45 got the new ``tools.scm.Git`` helper that will implement the ``scm`` behavior. All the other ``scm`` magic and classes has been removed.

PR #10520, opened 2022-02-04
remove local scm checkout, should be source()

memsharded:
As we have discussed, the local scm checkout doesn't make sense:

- scm is for "auto" capture only, and later retrieval
- otherwise use the ``source()`` method with some conandata.yml data.
Related https://github.com/conan-io/conan/issues/10503

### DIFF
--- a/conans/client/cmd/export.py
+++ b/conans/client/cmd/export.py
@@ -119,7 +119,7 @@ def _capture_scm_auto_fields(conanfile, conanfile_dir, recipe_layout, ignore_dir
     captured = scm_data.capture_origin or scm_data.capture_revision
 
     if not captured:
-        # We replace not only "auto" values, also evaluated functions (e.g from a python_require)
+        
         _add_scmdata_to_conandata_yml(recipe_layout, scm_data)
         return scm_data, None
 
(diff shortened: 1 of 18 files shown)


conans/client/cmd/export.py
```diff
@@ -120,7 +120,7 @@ def _capture_scm_auto_fields(conanfile, conanfile_dir, recipe_layout, ignore_dir
 
     if not captured:
         # We replace not only "auto" values, also evaluated functions (e.g from a python_require)
```
lasote:
```suggestion
        
```

conans/client/loader.py
```diff
@@ -50,18 +50,23 @@ def load_basic_module(self, conanfile_path, graph_lock=None, display=""):
...
-                    conanfile.scm.update(scm_data)
+
+            if hasattr(conanfile, "scm"):
+                if "scm" not in conanfile.__class__.__dict__:
+                    # If the scm is inherited, create my own instance
+                    conanfile.scm = conanfile.scm.copy()
+
+                if conanfile.scm.get("revision", "auto") != "auto":
+                    raise ConanException("'scm' can only be used for 'auto'. For fixed revisions use"
+                                         " the 'source()' method (and maybe conandata.yml file)")
```
lasote, line 64: If always has to be "auto", why accept a "revision" field and not just fail?

lasote (approved):

conans/client/loader.py
memsharded: To be compatible with 1.X

PR #10525, opened 2022-02-05
Add watcom compiler + Dos os

madebr:
This pr adds support for the watcom compiler + building for Dos.
With little modification to the recipes at conan-center-index, it can build binaries for dos/windows.
See this WIP branch: https://github.com/madebr/conan-center-index/tree/watcom-support

I don't know the exact procedure to modify settings.yml attributes, so I modified it in 2 places: in the method `get_default_settings_yml` and in the settings migration script.

I added some screenshots as "proof".

<details>

<summary>Screenshot proof</summary>

e.g. zlib on DOSBOX:
Screenshot from 2022-02-05 20-38-56
cJSON:
Screenshot from 2022-02-05 20-37-51

</details>

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -66,6 +66,7 @@ _t_default_settings_yml = Template(textwrap.dedent("""
         baremetal:
         VxWorks:
             version: ["7"]
+        Dos:
     arch: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv4, armv4i, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7, xtensalx6, xtensalx106]
     compiler:
         sun-cc:
@@ -142,6 +143,9 @@ _t_default_settings_yml = Template(textwrap.dedent("""
                     <<: *gcc
                     threads: [None]
                     exceptions: [None]
+        watcom:
+            version: ["1.0", "1.1", "1.2", "1.3", "1.4", "1.5", "1.6", "1.7", "1.8", "1.9",
+                      "2.0"]
 
     build_type: [None, Debug, Release, RelWithDebInfo, MinSizeRel]
 
--- a/conans/client/generators/cmake_common.py
+++ b/conans/client/generators/cmake_common.py
@@ -542,6 +542,11 @@ class CMakeCommonMacros:
                        conan_error_compiler_version()
                     endif()
                 endif()
+            elseif(CONAN_COMPILER STREQUAL "watcom")
+                conan_split_version(${CONAN_COMPILER_VERSION} CONAN_COMPILER_MAJOR CONAN_COMPILER_MINOR)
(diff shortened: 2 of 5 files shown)


conans/client/conf/__init__.py
```diff
@@ -66,6 +66,7 @@
         baremetal:
         VxWorks:
             version: ["7"]
+        Dos:
```
SSE4, line 69: should it be MS-DOS instead of generic DOS which is just Disk Operation System (AFAIK there were many implementations, such as TR-DOS, DR-DOS, MSX-DOS, etc, and I am not sure if they were binary compatible)

memsharded (commented):
At this stage (2.0 alpha, getting closer to beta), we will not add new compilers to the legacy integrations, unless it is very requested. (New OSs is a bit easier, because it is less used in build system integrations).

To have new compilers, we need to make sure they work with the new integrations. We are trying not to add new tools unless we have tests in the suite:
- A test with the ``conan new --template=cmake_lib`` (or alternatively something with Autotools or other build system modern integration) should be done, passing a profile that defines the new compiler settings
- The test can use ``@pytest.mark.tool_watcom`` to make sure it doesn't run in CI and define the tool in the ``conftest_user.py`` file (as defined in conftest.py). Some links in the test with instructions how to install the tools for the tests would be appreciated.
- If that works, we won't oppose to add things to the legacy integrations, if they are very low risk (this PR changes seem low risk, we should be good there)

conans/client/conf/__init__.py
madebr: There is also Freedos, which is still maintained.
I'm proposing a simple Dos here as a common base.
I don't know exactly, but I suppose there is some compatibility.
I think some/a lot syscalls are shared

A subsystem-like option might be needed, as you can build dos applications in multiple configurations:
e.g. LE executable, LX executable, dos4g extender, dos32a extender, ...
I'm going to ask watcom upstream what they think of this.

SSE4: from what I remember from these times, there were some interrupts provided by MS-DOS (e.g. int 21h), but these were way too limited, and almost every program practically used BIOS interrupts (such as int 10h).
outside of MS-DOS, it's hard to say how these interrupts are compatible, maybe need to check something like Interrup List book.
for binaries, yes, the most common formats were MZ and COM. LE/LX formats were designed for OS/2, and NE was designed for Win 3.x, but they were used by various DOS extenders (probably, they all were incompatible).
it has to be checked in watcom compiler docs what exactly can it generate.

madebr: One thing I'm also unsure about is the architecture.
Watcom can build 16- and 32-bit executables. So should such an option be added? And if it is needed, how should we name it?
Here's a quick overview of what archs are supported + compiler flags to use: https://wiki.archlinux.org/title/Open_Watcom

SSE4: yes, we probably will have to add x86_16 (or whatever name can be used to distinguish 16-bit from 32-bit x86, quick googling doesn't find any widespread name)
dos4gw (and other extenders) probably should be sub-systems.

so it may look like:
```
os: MSDOS
arch: x86, x86_16
subsystem: [None, "dos4gw"]
```

memsharded:
This PR is outdated, and also it seems there is not enough evidence of the utility to add these to the default ``settings.yml``. In Conan 2.0 adding custom settings is even easier with ``settings_user.yml``. And if any changes are necessary in any integration, it should be targeted to the new build system integrations, not the legacy one. Closing it, thanks very much for your contribution!

PR #10532, opened 2022-02-07
File tools at "conan." namespace

### LINKED ISSUE #10502: [feature] Move file tools to "conan"
There are still some "file" related tools not migrated to "conan".
- Some of them are probably overacting regarding the encoding, probably in some situations, inherited from python2. As we really don't know what is py2 legacy and what is necessary, try to simplify is something looks weird or too much, maybe always, all, utf-8?
- Some others maybe are not very used, take a look at conan-center and if they are not used better to postpone and migrate later.

lasote:
Changelog: omit
Docs: omit
Close https://github.com/conan-io/conan/issues/10502

### DIFF
--- a/conan/tools/files/__init__.py
+++ b/conan/tools/files/__init__.py
@@ -1,6 +1,5 @@
 from conan.tools.files.files import load, save, mkdir, ftp_download, download, get, rename, \
-    load_toolchain_args, save_toolchain_args, chdir, unzip, replace_in_file, collect_libs, \
-    check_md5, check_sha1, check_sha256
+    chdir, unzip, replace_in_file, collect_libs, check_md5, check_sha1, check_sha256
 
 from conan.tools.files.patches import patch, apply_conandata_patches
 from conan.tools.files.cpp_package import CppPackage
--- a/conan/tools/files/__init__.py
+++ b/conan/tools/files/__init__.py
@@ -1,5 +1,7 @@
 from conan.tools.files.files import load, save, mkdir, ftp_download, download, get, rename, \
-    load_toolchain_args, save_toolchain_args, chdir
+    load_toolchain_args, save_toolchain_args, chdir, unzip, replace_in_file, collect_libs, \
+    check_md5, check_sha1, check_sha256
+
 from conan.tools.files.patches import patch, apply_conandata_patches
 from conan.tools.files.cpp_package import CppPackage
 from conan.tools.files.packager import AutoPackager
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -1,23 +1,25 @@
 import configparser
 import errno
+import gzip
+import hashlib
 import os
 import platform
 import subprocess
+import sys
 from contextlib import contextmanager
+from fnmatch import fnmatch
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
 from conans.client.downloaders.download import run_downloader
-from conans.client.tools.files import unzip, which
+from conans.client.tools.files import which
 from conans.errors import ConanException
-from conans.util.files import decode_text, to_file_bytes
 
 
-def load(conanfile, path, binary=False, encoding="auto"):
+def load(conanfile, path) -> bytes:
     """ Loads a file content """
     with open(path, 'rb') as handle:
         tmp = handle.read()
-        # TODO: Get rid of encoding auto-detection
-        return tmp if binary else decode_text(tmp, encoding)
+        return tmp
 
 
 def save(conanfile, path, content, append=False):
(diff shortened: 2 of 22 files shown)


conan/tools/files/files.py
```diff
@@ -242,3 +245,250 @@ def chdir(conanfile, newdir):
...
+    :param file_path: Path to the file
+    :param search: Pattern to search, can be a str or a bytes() with a specific encoding
+    :param replace: string to replace the matches, can be a str or a bytes() with a specific encoding
+    :param strict: Raise in case "search" is not found in the file contents
+    :return:
+    """
+    if isinstance(search, str):
+        search = bytes(search, "utf-8")
+    if isinstance(replace, str):
+        replace = bytes(replace, "utf-8")
```
lasote: To discuss

conan/tools/files/files.py
```diff
@@ -41,7 +43,9 @@ def save(conanfile, path, content, append=False):
                 raise
 
     with open(path, mode) as handle:
-        handle.write(to_file_bytes(content, encoding="utf-8"))
+        if not isinstance(content, bytes):
```
lasote, line 50: To discuss: This allows to write directly bytes or str with an encoding.

conans/test/unittests/tools/files/collect_lib_test.py
```diff
@@ -0,0 +1,72 @@
+import os
+
+from conan.tools.files import check_md5, check_sha1, check_sha256, collect_libs
+from conans.model.build_info import CppInfo
+from conans.test.utils.mocks import ConanFileMock
+from conans.test.utils.test_files import temp_folder
+from conans.tools import save
+
+
+def test_collect_libs():
```
lasote, line 10: "Adapted" from old test

conan/tools/files/__init__.py
```diff
@@ -1,5 +1,7 @@
 from conan.tools.files.files import load, save, mkdir, ftp_download, download, get, rename, \
-    load_toolchain_args, save_toolchain_args, chdir
+    load_toolchain_args, save_toolchain_args, chdir, unzip, replace_in_file, collect_libs, \
```
memsharded: I'd say lets keep ``load_toolchain_args`` and ``save_toolchain_args`` a implementation detail of the current integrations. Users can save any file they want and recover it easily in ``build()``, no need to standardize.

I am considering exploring the idea of moving to a json compatible with CMakePresets.json, and getting the information from there, from recent feedback it seems doable and it will be much better integrated with IDEs and other stuff

conan/tools/files/files.py
```diff
@@ -1,26 +1,28 @@
...
 import platform
 import subprocess
+import sys
 from contextlib import contextmanager
+from fnmatch import fnmatch
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
 from conans.client.downloaders.download import run_downloader
-from conans.client.tools.files import unzip, which
+from conans.client.tools.files import which
```
memsharded: which in Conan 2.0 is going to be just ``shutil.which()``. Do we still want a wrapper? (Maybe necessary for Py2 only?)

conans/test/unittests/tools/files/checksums_test.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+from conan.tools.files import check_md5, check_sha1, check_sha256
+from conans.test.utils.test_files import temp_folder
+from conans.tools import save
```
memsharded: Better not use ``from conans.tools import save`` for tests, as it will break when merged to ``develop2``. Better use ``conans.utils.files.save``

conans/test/unittests/tools/files/collect_lib_test.py
```diff
@@ -0,0 +1,72 @@
+import os
+
+from conan.tools.files import check_md5, check_sha1, check_sha256, collect_libs
+from conans.model.build_info import CppInfo
+from conans.test.utils.mocks import ConanFileMock
+from conans.test.utils.test_files import temp_folder
+from conans.tools import save
```
memsharded: same

conans/test/unittests/tools/files/test_zipping.py
```diff
@@ -0,0 +1,31 @@
+import os
+import zipfile
+from os.path import basename
+
+import pytest
+
+from conan.tools.files import unzip
+from conans.test.utils.mocks import MockConanfile
+from conans.test.utils.test_files import temp_folder
+from conans.tools import save
```
memsharded: same

conan/tools/files/__init__.py
lasote: Sure, I just kept them because they were already there.

conan/tools/files/files.py
lasote: I assume, the new tools at "conan" will be python3 (because conan2 is python3) so I'll remove the import (I'll keep the implementation to not break) and use `shutil.which()` here.

memsharded (approved):

PR #10536, opened 2022-02-08
Add missing tools to develop

czoido:
Changelog: Feature: Port `conan.tools.Version` to Conan 1.X.
Changelog: Feature: Port `conan.tools.build.cross_building` to Conan 1.X.
Docs: https://github.com/conan-io/docs/pull/2434

Later moved Version to conan.tools.scm.Version in https://github.com/conan-io/conan/pull/10712

There were some tools that are not ported to Conan 1.X, [this PR](https://github.com/czoido/conan-center-index/pull/4) was failing because of that.


conan/tools/build/flags.py
```diff
@@ -0,0 +1,457 @@
+from conan.tools.apple.apple import to_apple_arch
+from conans.model.version import Version
+
+
+def architecture_flag(settings):
```
memsharded: Just to make sure, most of these are internal in 2.0 to the toolchains and other helpers. If they are backported to 1.X we should make sure:

- They are not documented as public, they are not exported via ``from conan.tools.xxxx``
- I would expect the consumers (toolchains) using these to have been changed.
- Recall that it is possible that we want to pass ``conanfile`` as the first argument always to all public tools, so this might need a refactor for that.

memsharded (commented):
I think I am missing something. https://github.com/czoido/conan-center-index/pull/4/files doesn't seem to need all these tools. Aren't those insufficiently tested internally in the test-suite?

czoido:
> I think I am missing something. https://github.com/czoido/conan-center-index/pull/4/files doesn't seem to need all these tools. Aren't those insufficiently tested internally in the test-suite?

No, I think we don't need all of these, I'll remove some

memsharded (approved):

PR #10537, opened 2022-02-08
[feature] `ConfDefinition` and `Conf` enhancements.

franramirez688:
Changelog: Feature: Adding more functionality to `ConfDefinition` and `Conf`, something similar to `ProfileEnvironment` and `Environment` ones.

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -75,7 +75,6 @@ def apple_min_version_flag(conanfile):
 
 def apple_sdk_path(conanfile):
     sdk_path = conanfile.conf.get("tools.apple:sdk_path")
-    # It's better to keep the default return later in this case
     if not sdk_path:
         sdk_path = XCRun(conanfile.settings).sdk_path
     return sdk_path
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -74,7 +74,8 @@ def apple_min_version_flag(conanfile):
 
 
 def apple_sdk_path(conanfile):
-    sdk_path = conanfile.conf["tools.apple:sdk_path"]
+    sdk_path = conanfile.conf.get("tools.apple:sdk_path")
+    # It's better to keep the default return later in this case
     if not sdk_path:
         sdk_path = XCRun(conanfile.settings).sdk_path
     return sdk_path
--- a/conan/tools/build/cpu.py
+++ b/conan/tools/build/cpu.py
@@ -8,7 +8,7 @@ def build_jobs(conanfile):
     njobs = conanfile.conf.get("tools.build:jobs",
                                default=_cpu_count(),
                                check_type=int)
-    return njobs or None
+    return njobs
 
 
 def _cpu_count():
--- a/conan/tools/build/cpu.py
+++ b/conan/tools/build/cpu.py
@@ -5,13 +5,10 @@ from conans.util.files import load
 
 
 def build_jobs(conanfile):
-    njobs = conanfile.conf["tools.build:jobs"]
-    if njobs is not None:
-        result = int(njobs)
-        if result == 0:
-            return None
-        return result
-    return _cpu_count()
+    njobs = conanfile.conf.get("tools.build:jobs",
+                               default=_cpu_count(),
+                               check_type=int)
+    return njobs or None
 
 
 def _cpu_count():
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -473,8 +473,9 @@ class FindFiles(Block):
         # To find the generated cmake_find_package finders
         # TODO: Change this for parameterized output location of CMakeDeps
         find_package_prefer_config = "ON"  # assume ON by default if not specified in conf
-        prefer_config = self._conanfile.conf.get("tools.cmake.cmaketoolchain:find_package_prefer_config")
-        if prefer_config in (False, 0) or (isinstance(prefer_config, str) and prefer_config.lower() in ("false", "off")):
+        prefer_config = self._conanfile.conf.get("tools.cmake.cmaketoolchain:find_package_prefer_config",
+                                                 check_type=bool)
+        if prefer_config is False:
             find_package_prefer_config = "OFF"
 
         os_ = self._conanfile.settings.get_safe("os")
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -194,7 +194,7 @@ class GLibCXXBlock(Block):
         if compiler in ['clang', 'apple-clang', 'gcc']:
             if libcxx == "libstdc++":
                 glib = "0"
-            elif libcxx == "libstdc++11" and self._conanfile.conf["tools.gnu:define_libcxx11_abi"]:
+            elif libcxx == "libstdc++11" and self._conanfile.conf.get("tools.gnu:define_libcxx11_abi"):
                 glib = "1"
         return {"set_libcxx": lib, "glibcxx": glib}
 
@@ -308,7 +308,7 @@ class AndroidSystemBlock(Block):
         #  https://developer.android.com/ndk/guides/cpp-support
         libcxx_str = str(self._conanfile.settings.compiler.libcxx)
 
-        android_ndk_path = self._conanfile.conf["tools.android:ndk_path"]
+        android_ndk_path = self._conanfile.conf.get("tools.android:ndk_path")
         if not android_ndk_path:
             raise ConanException('CMakeToolchain needs tools.android:ndk_path configuration defined')
         android_ndk_path = android_ndk_path.replace("\\", "/")
@@ -473,8 +473,8 @@ class FindFiles(Block):
         # To find the generated cmake_find_package finders
         # TODO: Change this for parameterized output location of CMakeDeps
         find_package_prefer_config = "ON"  # assume ON by default if not specified in conf
-        prefer_config = self._conanfile.conf["tools.cmake.cmaketoolchain:find_package_prefer_config"]
-        if prefer_config is not None and prefer_config.lower() in ("false", "0", "off"):
+        prefer_config = self._conanfile.conf.get("tools.cmake.cmaketoolchain:find_package_prefer_config")
+        if prefer_config in (False, 0) or (isinstance(prefer_config, str) and prefer_config.lower() in ("false", "off")):
             find_package_prefer_config = "OFF"
 
         os_ = self._conanfile.settings.get_safe("os")
@@ -531,7 +531,7 @@ class UserToolchain(Block):
 
     def context(self):
         # This is global [conf] injection of extra toolchain files
-        user_toolchain = self._conanfile.conf["tools.cmake.cmaketoolchain:user_toolchain"]
+        user_toolchain = self._conanfile.conf.get("tools.cmake.cmaketoolchain:user_toolchain")
         toolchains = [user_toolchain.replace("\\", "/")] if user_toolchain else []
         return {"paths": toolchains if toolchains else []}
 
@@ -657,13 +657,13 @@ class GenericSystemBlock(Block):
         return None
 
     def _get_cross_build(self):
-        user_toolchain = self._conanfile.conf["tools.cmake.cmaketoolchain:user_toolchain"]
+        user_toolchain = self._conanfile.conf.get("tools.cmake.cmaketoolchain:user_toolchain")
         if user_toolchain is not None:
             return None, None, None  # Will be provided by user_toolchain
 
-        system_name = self._conanfile.conf["tools.cmake.cmaketoolchain:system_name"]
-        system_version = self._conanfile.conf["tools.cmake.cmaketoolchain:system_version"]
-        system_processor = self._conanfile.conf["tools.cmake.cmaketoolchain:system_processor"]
+        system_name = self._conanfile.conf.get("tools.cmake.cmaketoolchain:system_name")
+        system_version = self._conanfile.conf.get("tools.cmake.cmaketoolchain:system_version")
+        system_processor = self._conanfile.conf.get("tools.cmake.cmaketoolchain:system_processor")
 
         settings = self._conanfile.settings
         if hasattr(self._conanfile, "settings_build"):
@@ -869,7 +869,7 @@ class CMakeToolchain(object):
         return content
 
     def generate(self):
-        toolchain_file = self._conanfile.conf["tools.cmake.cmaketoolchain:toolchain_file"]
+        toolchain_file = self._conanfile.conf.get("tools.cmake.cmaketoolchain:toolchain_file")
         if toolchain_file is None:  # The main toolchain file generated only if user dont define
             save(self.filename, self.content)
         # If we're using Intel oneAPI, we need to generate the environment file and run it
@@ -896,7 +896,7 @@ class CMakeToolchain(object):
         conanfile = self._conanfile
 
         # Downstream consumer always higher priority
-        generator_conf = conanfile.conf["tools.cmake.cmaketoolchain:generator"]
+        generator_conf = conanfile.conf.get("tools.cmake.cmaketoolchain:generator")
         if generator_conf:
             return generator_conf
 
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -108,16 +108,9 @@ class MSBuildDeps(object):
         self.platform = {'x86': 'Win32',
                          'x86_64': 'x64'}.get(str(conanfile.settings.arch))
         # ca_exclude section
-        self.exclude_code_analysis = None
-        ca_exclude = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis")
-        if ca_exclude is not None:
-            # TODO: Accept single strings, not lists
-            if isinstance(ca_exclude, list):
-                self.exclude_code_analysis = ca_exclude
-            else:
-                raise ConanException("tools.microsoft.msbuilddeps:exclude_code_analysis must be a"
-                                     " list of package names patterns like ['pkga*']")
-
+        # TODO: Accept single strings, not lists
+        self.exclude_code_analysis = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis",
+                                                              check_type=list)
         check_using_build_profile(self._conanfile)
 
     def generate(self):
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -112,7 +112,7 @@ class MSBuildDeps(object):
         ca_exclude = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis")
         if ca_exclude is not None:
             # TODO: Accept single strings, not lists
-            if isinstance(self.exclude_code_analysis, list):
+            if isinstance(ca_exclude, list):
                 self.exclude_code_analysis = ca_exclude
             else:
                 raise ConanException("tools.microsoft.msbuilddeps:exclude_code_analysis must be a"
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -111,6 +111,7 @@ class MSBuildDeps(object):
         self.exclude_code_analysis = None
         ca_exclude = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis")
         if ca_exclude is not None:
+            # TODO: Accept single strings, not lists
             if isinstance(self.exclude_code_analysis, list):
                 self.exclude_code_analysis = ca_exclude
             else:
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -109,11 +109,11 @@ class MSBuildDeps(object):
                          'x86_64': 'x64'}.get(str(conanfile.settings.arch))
         # ca_exclude section
         self.exclude_code_analysis = None
-        ca_exclude = self._conanfile.conf["tools.microsoft.msbuilddeps:exclude_code_analysis"]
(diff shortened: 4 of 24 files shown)


conans/test/unittests/model/test_conf.py
```diff
@@ -9,8 +9,9 @@
 @pytest.fixture()
 def conf_definition():
     text = textwrap.dedent("""\
-        tools.microsoft.msbuild:verbosity=minimal
-        user.company.toolchain:flags=someflags""")
+    tools.microsoft.msbuild:verbosity=minimal
```
memsharded: The previous indent wasn't bad, it was actually the one closer to what "black" formatter is doing (which we might probably adopt at some point). In any case it is a change that doesn't help in the review, as it seems initially you are changing tests (more than necessary)

conans/test/unittests/model/test_conf.py
```diff
@@ -85,3 +89,28 @@ def test_parse_spaces():
...
+    user.company.toolchain:flags+=moreflags
+    # tools.cmake:njobs=5
+    # tools.flags:cc=-first 1 -second 2
+    # tools.flags:cc+=-third 3
+    # tools.flags:cxx=-first 1
+    # tools.cmake:njobs=!
+    # tools.flags:cxx=+-second 2
+    # tools.flags:cxx=+-third 3
+    """)
+    c2 = ConfDefinition()
```
memsharded, line 175: We need more tests of access, checking what is the final value (via apply() or get()) of a given conf.

memsharded (commented):
This is looking very good.
I can see there is some duplication with Environment, but I really appreciate the clean approach of introducing the functionality and tests first, without touching anything else, and I prefer considering any common refactoring later, good job.

franramirez688:
@memsharded Exactly. I preferred to duplicate everything as necessary and refactor later, having some common `BaseEnvModel` or whatever class to inherit from it. Let's go step by step 😁

conans/test/unittests/model/test_conf.py
```diff
@@ -85,3 +89,65 @@ def test_parse_spaces():
...
+    c.update_conf_definition(c2)
+    c_str = repr(c)
+    assert "None: Conf: OrderedDict([('user.company.toolchain:flags', ConfValues: oneflag secondflag), " \
+           "('tools.microsoft.msbuild:verbosity', ConfValues: )])" in c_str
+    assert "'zlib': Conf: OrderedDict([('user.company.toolchain:flags', ConfValues: z1flag z2flag)])" in c_str
+    assert "'openssl': Conf: OrderedDict([('user.company.toolchain:flags', ConfValues: oflag)])" in c_str
+    assert c["tools.microsoft.msbuild:verbosity"] is None
+    assert c["user.company.toolchain:flags"] == ["oneflag", "secondflag"]
+    assert c["zlib:user.company.toolchain:flags"] == ["z1flag", "z2flag"]
+    assert c["openssl:user.company.toolchain:flags"] == "oflag"  # FIXME: Backward compatibility
```
memsharded: I think this is problematic. That means that every consumer will actually need to check if things are a list, a single value, or None.

We might prefer to normalize this, and we talked to normalize access to the conf (by consumers) with a ``conf.get()`` method. Probably this method can implement the logic:

```python
conf.get("msbuild:verbosity", accept_list=False)  # will raise if the internal list contains more than 1 value, no sense to append or prepend to msbuild:verbosity. 
# we probably want to make ``acept_list=False`` the default (and pick a better name)
conf.get("cmaketoolchain:user_toolchain", accept_list=True) # we know there can be multiple toolchains
```

memsharded (approved):
I think we can move this PR forward, but as long as it follows a better ``conf.get()`` (and remove the ``[]`` access, or keep it for 1 release for transitioning)

conans/test/unittests/model/test_conf.py
franramirez688: Yes, but I did not want to force it for now. I checked what is already in Conan2.0 and my idea is to bring it back here. Next iteration, I'm going to add that `conf.get()` and I want to discuss if we should get rid of the `[]` access or not for 1.x version.

lasote (approved):

conans/model/conf.py
```diff
@@ -1,43 +1,43 @@
...
+    "core:required_conan_version": (str, "Raise if current version does not match the defined range."),
+    "core.package_id:msvc_visual_incompatible": (str, "Allows opting-out the fallback from the new msvc compiler to the Visual Studio compiler existing binaries"),
+    "core:default_profile": (str, "Defines the default host profile ('default' by default)"),
+    "core:default_build_profile": (str, "Defines the default build profile (None by default)"),
+    "tools.android:ndk_path": (str, "Argument for the CMAKE_ANDROID_NDK"),
+    "tools.build:skip_test": (str, "Do not execute CMake.test() and Meson.test() when enabled"),
+    "tools.build:jobs": (str, "Default compile jobs number -jX Ninja, Make, /MP VS (default: max CPUs)"),
+    "tools.cmake.cmaketoolchain:generator": (str, "User defined CMake generator to use instead of default"),
+    "tools.cmake.cmaketoolchain:find_package_prefer_config": (str, "Argument for the CMAKE_FIND_PACKAGE_PREFER_CONFIG"),
+    "tools.cmake.cmaketoolchain:toolchain_file": (str, "Use other existing file rather than conan_toolchain.cmake one"),
```
memsharded: Having the types here is unexpected, and I think it makes the thing a bit unnecessarily complicated. 
Taking into account that user confs will not have the possibility to define the type, I would recommend a system that doesn't require the  type to be defined anywhere, but let the user define the type with their inputs.

Part of the surprise is that if this would be a valid feature, I would expect to see:
- ``user_toolchain`` to be a list, not str. At least to have something that is not a ``str``, otherwise, if everything is a ``str`` why defining it?
- To see tests where introducing the wrong type fails.

I am not saying to add that, I am suggesting that possibly it is not necessary, and things can be simpler.

conans/model/conf.py
```diff
@@ -47,117 +47,391 @@ def _is_profile_module(module_name):
...
+        self._values.pop(conf_name, None)
+        return value
+
+    def _get_conf_value(self, name, value):
+        """
+        Get a valid _ConfValue object based on the built-in Conf type declared
+        or the value-like object. For now, we only manage lists or strings.
+        """
+        if (name in self._values and isinstance(self._values[name], _ConfListValue)) \
+           or BUILT_IN_CONFS.get(name, [None])[0] is list or isinstance(value, list):
```
memsharded: This is quite complex to understand, for me it is a symptom of the above decision to define and try to validate types.

memsharded (changes requested):
To discuss, please see comments, I am not sure this is the approach we want to follow.

franramirez688:
If you like more this approach, I'll have to add more unit tests.

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -109,11 +109,12 @@ def __init__(self, conanfile):
                          'x86_64': 'x64'}.get(str(conanfile.settings.arch))
         # ca_exclude section
         self.exclude_code_analysis = None
-        ca_exclude = self._conanfile.conf["tools.microsoft.msbuilddeps:exclude_code_analysis"]
+        ca_exclude = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis")
         if ca_exclude is not None:
             # TODO: Accept single strings, not lists
-            self.exclude_code_analysis = eval(ca_exclude)
-            if not isinstance(self.exclude_code_analysis, list):
+            if isinstance(ca_exclude, list):
```
memsharded: Looks good at the moment, this is a good example of the ``conf.get("xxx", type=list)`` utility, that will allow to simplify code of consumers. Can be done later in another PR, no prob

conans/model/conf.py
```diff
@@ -47,117 +47,402 @@ def _is_profile_module(module_name):
...
+        return repr(self._value)
+
+    @property
+    def value(self):
+        return self._value
+
+    def copy(self):
+        return self.__class__(self._name, self._value)
+
+    def dumps(self):
```
memsharded, line 75: Probably can be dropped (these 4 methods that rais NotImplementedError), boilerplate exclusively to protect developers, not really for users.

conans/model/conf.py
```diff
@@ -47,117 +47,402 @@ def _is_profile_module(module_name):
...
 
     def __init__(self):
-        self._values = {}  # property: value
+        # It being ordered allows for Windows case-insensitive composition
+        self._values = OrderedDict()  # {var_name: [] of values, including separators}
+
+    def __bool__(self):
+        return bool(self._values)
+
+    __nonzero__ = __bool__
```
memsharded, line 154:
```suggestion
    # TODO: Python2, remove in 2.0
    __nonzero__ = __bool__
```

conans/model/conf.py
```diff
@@ -47,117 +47,402 @@ def _is_profile_module(module_name):
...
+    def __repr__(self):
+        return "Conf: " + repr(self._values)
+
+    def __eq__(self, other):
+        """
+        :type other: Conf
+        """
+        return other._values == self._values
+
+    def __ne__(self, other):
```
memsharded, line 166:
```suggestion
    # TODO: Python2, remove in 2.0
    def __ne__(self, other):
```

memsharded (approved):
Good job!
I think this can be merged, but lets get another review too.

conan/tools/microsoft/msbuilddeps.py
franramirez688: In this case, it would be enough with that `conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis")`, without having to convert it because the user knows that this field is a `list`:

```
[conf]
tools.microsoft.msbuilddeps:exclude_code_analysis = ['pkga*']
```
So, this refactor brings us:

```python
conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis") == ["pkga*"]
```

memsharded: But you still have the ``if type is not list => raise ConanException("hey I expected a list")`` in your ``msbuilddeps.py`` code.
I am saying this is not necessary here, but the user code in ``msbuilddeps.py`` should be just:
```python
self.exclude_code_analysis = conf.get("tools...:exclude_code_analysis", type=list)
```
And that would already raise if it is not a list. Because that check will be common to other places in code, it is more logical to have it available inside ``conf.get()``

conans/model/conf.py
```diff
@@ -47,117 +47,404 @@ def _is_profile_module(module_name):
...
+        :param conf_name: conf name
+        :param default: default value in case of conf does not have the conf_name key
+        :param cast: any callable to apply any kind of transformation to final conf value
+        """
+        conf_value = self._values.get(conf_name)
+        if conf_value:
+            v = conf_value.value
+            if cast is not None:
+                try:
+                    v = cast(v)
```
franramirez688: @memsharded @czoido
In my opinion, this `cast` or `type` (the name can change) is a bit useless. For instance:

```python
# we have this example
conf.get("xxxxxx") == "True"
assert conf.get("xxxxxx", cast=bool) is True  # that's OK

# Now, let's imagine the other case
conf.get("xxxxxx") == "False"
assert conf.get("xxxxxx", cast=bool) is True  # that's OK but weird, isn't it?
```
So, from my point of view, what I expect would be:
```python
# we have this example
conf.get("xxxxxx") == "True"
assert conf.get("xxxxxx", cast=eval) is True  # that's OK

# Now, let's imagine the other case
conf.get("xxxxxx") == "False"
assert conf.get("xxxxxx", cast=eval) is False  # that's OK
```
But it looks a bit odd... What do you think about it? Should this `get-cast` be smarter?

conan/tools/microsoft/msbuilddeps.py
franramirez688: Oh, I got it. Then, I'd prefer to see something like:
```python
self.exclude_code_analysis = conf.get("tools...:exclude_code_analysis", check_type=list)
```
I think it's clearer, isn't it?

conans/model/conf.py
franramirez688: Or even get rid of it? 😄

conans/model/conf.py
franramirez688: Exactly, I agree 😉

conan/tools/microsoft/msbuilddeps.py
memsharded: Yes, ``check_type`` is even better 👍

conans/model/conf.py
memsharded: Yes, I don't think a cast here is that useful. If we are doing ``parsed_value = eval(value)`` while parsing user input, and that converts directly to Python primitive types, then probably what is better is to remove the ``str()`` conversion done in ``_ConfStrValue`` and move the primitive type (could be directly a boolean) from the user input totally transparently to the user ``conf.get()`` configuration.

franramirez688: removed 👍

conans/test/unittests/model/test_conf.py
```diff
@@ -85,3 +90,103 @@ def test_parse_spaces():
...
+    """)
+    c = ConfDefinition()
+    c.loads(text)
+    text = textwrap.dedent("""\
+        user.company.cpu:jobs=5
+        user.company.build:ccflags=--m otherflag
+        zlib:user.company.check:shared=!
+        zlib:user.company.check:shared_str="False"
+        user.company.list:objs+=[5, 6]
+        user.company.list:objs=+0
```
lasote, line 171: Could this be confusing? When is append and when it is an extend?

czoido:
How can we set a list configuration from the command line? I can't see any tests about that

conans/test/unittests/model/test_conf.py
lasote: What happens if you do += to a dictionary?

conan/tools/apple/apple.py
```diff
@@ -74,7 +74,8 @@ def apple_min_version_flag(conanfile):
 
 
 def apple_sdk_path(conanfile):
-    sdk_path = conanfile.conf["tools.apple:sdk_path"]
+    sdk_path = conanfile.conf.get("tools.apple:sdk_path")
+    # It's better to keep the default return later in this case
```
czoido: What do you mean with this comment?

conans/test/unittests/model/test_conf.py
franramirez688: Adding more tests 😉

conan/tools/apple/apple.py
franramirez688: I meant that we don't want to run the `xcrun` tool always without knowing if it was given a `sdk_path` previously via `conf`.

conans/test/unittests/model/test_conf.py
franramirez688: Anyway, let me explain it. `append` and `extend` have the same syntax as environment:

`+= -> append`
`=+ -> prepend`

For now, only lists are able to do it. If you try it with a dict-like object, it will fail.

lasote: Yes, my point is, if you have a list `[1,2,3]` and you do `user.company.list:objs=+0` you get `[0,1,2,3]`. That is totally expected but, if you have `[0,1,2,3]` and you do `user.company.list:objs+=[5, 6]`, what you get?

a) `[0,1,2,3,4,5,6]`
b) `[0,1,2,3,4,[5,6]]`

What if I append a dictionary? `user.company.list:objs+={foo: 5}`

a) `[0,1,2,3,4, {foo: 5}]`

So, looks like if you are appending a list, you extend, but if you append another thing, you append.

> For now, only lists are able to do it. If you try it with a dict-like object is going to fail.

Good, I understand. I would like to see a test.

conan/tools/build/cpu.py
```diff
@@ -5,13 +5,10 @@
...
-    if njobs is not None:
-        result = int(njobs)
-        if result == 0:
-            return None
-        return result
-    return _cpu_count()
+    njobs = conanfile.conf.get("tools.build:jobs",
+                               default=_cpu_count(),
+                               check_type=int)
+    return njobs or None
```
memsharded: This is confusing, why ``njobs or None``? Shouldn't be it at least cpu_count?
Or is it just in case someone define "0"?

conan/tools/cmake/toolchain.py
```diff
@@ -473,8 +473,8 @@ def context(self):
         # To find the generated cmake_find_package finders
         # TODO: Change this for parameterized output location of CMakeDeps
         find_package_prefer_config = "ON"  # assume ON by default if not specified in conf
-        prefer_config = self._conanfile.conf["tools.cmake.cmaketoolchain:find_package_prefer_config"]
-        if prefer_config is not None and prefer_config.lower() in ("false", "0", "off"):
+        prefer_config = self._conanfile.conf.get("tools.cmake.cmaketoolchain:find_package_prefer_config")
+        if prefer_config in (False, 0) or (isinstance(prefer_config, str) and prefer_config.lower() in ("false", "off")):
```
memsharded: A bit dirty. I'd probably add a ``get(.., check_type=bool)`` smart converter or a ``conf.get_bool()``, or something that implements our "boolean" conf convention for ``off, false, 0`` etc

conan/tools/microsoft/visual.py
```diff
@@ -54,18 +54,21 @@ def generate(self, scope="build"):
...
-        toolset_override = conanfile.conf["tools.microsoft.msbuild:vs_version"]
+        toolset_override = conanfile.conf.get("tools.microsoft.msbuild:vs_version")
         if toolset_override:
             visual_version = toolset_override
         else:
             visual_version = msvc_version_to_vs_ide_version(compiler_version)
     else:
         visual_version = compiler_version
-    return visual_version
+    return str(visual_version)
```
memsharded: This must be a string at this point, right? Why conversion?

memsharded (approved):
Good job!

conan/tools/build/cpu.py
franramirez688: Yes, the second one, because it had this original code:

```python
     result = int(njobs)
     if result == 0:
          return None
```
if someone defines `0` then we return `None`.

conan/tools/microsoft/visual.py
franramirez688: this `conanfile.conf.get("tools.microsoft.msbuild:vs_version")` returns an integer:

```
tools.microsoft.msbuild:vs_version = 17
```
Internally, it's managed as an integer. If you want to force it to be a string value, you'll have to write:

```
tools.microsoft.msbuild:vs_version = "17"
```
So I can change that line and apply the conversion to the original one:
```python
toolset_override = str(conanfile.conf.get("tools.microsoft.msbuild:vs_version", default=""))
```

conan/tools/apple/apple.py
czoido: I think the comment is confusing and does not make much sense outside the context of this PR, I think it would be better to remove it or extend add more detailed information of why that is important...

conans/test/integration/configuration/conf/test_conf.py
```diff
@@ -153,3 +153,22 @@ def test_composition_conan_conf_overwritten_by_cli_arg(client):
...
+    """
+    conf = textwrap.dedent("""\
+        tools.build.flags:ccflags=["-Wall"]
+        """)
+    save(client.cache.new_config_path, conf)
+    client.run('install . -c "tools.build.flags:ccflags+=[\'-Werror\']" '
+               '-c "tools.microsoft.msbuildtoolchain:compile_options={\'ExceptionHandling\': \'Async\'}"')
+
+    assert "tools.build.flags:ccflags$['-Wall', '-Werror']" in client.out
+    assert "tools.microsoft.msbuildtoolchain:compile_options${'ExceptionHandling': 'Async'}" in client.out
```
franramirez688, line 174: @czoido testing the Conf for list/dict types via cli

conans/test/unittests/model/test_conf.py
```diff
@@ -85,3 +90,139 @@ def test_parse_spaces():
...
+    assert c.get(text.split("=")[0]) == expected
+
+
+@pytest.mark.parametrize("text1, text2, expected", [
+    ("user.company.list:objs=[2, 3]", "user.company.list:objs=+[0, 1]", [0, 1, 2, 3]),
+    ("user.company.list:objs=[2, 3]", "user.company.list:objs+=[4, 5]", [2, 3, 4, 5]),
+    ("user.company.list:objs=[2, 3]", "user.company.list:objs+={'a': 1}", [2, 3, {'a': 1}]),
+    ("user.company.list:objs=[2, 3]", "user.company.list:objs=+start", ["start", 2, 3]),
+    ("user.company.list:objs=[2, 3]", "user.company.list:objs=[0, 1]", [0, 1]),
+])
```
franramirez688, line 121: @lasote Added more tests here and below.

conan/tools/microsoft/visual.py
memsharded: That is the thing, we can make this incorrect, and fail if it is not a string? All versions are always strings, not integer. Defining a version as an integer is not great, but that opens the door to define a version as a float like ``...:vs_version=17.3``, which looks correct on the surface, but managing such version as a float is really bad. So maybe we want to ban defining versions as integers too?

conan/tools/microsoft/visual.py
```diff
@@ -54,11 +54,14 @@ def generate(self, scope="build"):
...
 def vs_ide_version(conanfile):
+    """
+    Get the VS IDE version as string
+    """
     compiler = conanfile.settings.get_safe("compiler")
     compiler_version = (conanfile.settings.get_safe("compiler.base.version") or
                         conanfile.settings.get_safe("compiler.version"))
     if compiler == "msvc":
-        toolset_override = conanfile.conf["tools.microsoft.msbuild:vs_version"]
+        toolset_override = conanfile.conf.get("tools.microsoft.msbuild:vs_version", check_type=str)
```
franramirez688, line 64: @memsharded I have added a kind of several smart conversions for `str` and `bool` types so, if you specify the `check_type`, they'll help the user to ensure that the value gotten is a string or a boolean one.

conan/tools/build/cpu.py
franramirez688: It seemed that all the places where it was being used were prepared to check if it was returning more than 0 jobs. Removed 👍

czoido (approved):

PR #10546, opened 2022-02-09
new imports functionality via generate() tested for editable

memsharded:
as we are removing the imports explicit functionality, test if the editable
can still work for the "imports" case. It seem possible if:

- "dep" package, both in cache and editable mode, defines correctly its layout
- Tricky: "dep" package musth use self.source_folder+res instead of layout to package()
- IMPORTANT: Consumers should use dep.cpp_info.resdirs[0], not dep.package_folder

### DIFF
--- a/conans/test/integration/editable/test_editable_import.py
+++ b/conans/test/integration/editable/test_editable_import.py
@@ -1,0 +1,58 @@
+import os
+import shutil
+import textwrap
+
+from conans.test.utils.tools import TestClient
+
+
+class TestEditableImport:
+
+    def test_copy_from_dep_in_generate(self):
+        """ as we are removing the imports explicit functionality, test if the editable
+        can still work for the "imports" case. It seem possible if:
+
+        - "dep" package, both in cache and editable mode, defines correctly its layout
+        - Tricky: "dep" package musth use self.source_folder+res instead of layout to package()
+        - IMPORTANT: Consumers should use dep.cpp_info.resdirs[0], not dep.package_folder
+        """
+        t = TestClient()
+        dep = textwrap.dedent("""
+            import os
+            from conan import ConanFile
+            from conan.tools.files import copy
+            class Pkg(ConanFile):
+                exports = "*"
+                def layout(self):
+                    self.folders.source = "src"
+                    self.cpp.source.resdirs = ["res"]
+                def package(self):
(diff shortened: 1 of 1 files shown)


lasote (approved):
I think there is no problem with this example and some interesting conclusions.
Regarding your comments:

> "dep" package, both in cache and editable mode, defines correctly its layout

Sure, nothing new

> Tricky: "dep" package musth use self.source_folder+res instead of layout to package()

Why is exactly tricky? The copy from self.source_folder is necessary to package to the cache only, but how else could it be? 

> IMPORTANT: Consumers should use dep.cpp_info.resdirs[0], not dep.package_folder

This is the most important conclusion,  as the cpp_info directories are always absolute (since a couple of weeks or so), it makes perfect sense that, if you are trying to import contents from a dependency (no matter if editable or cache) use the declaration of the `cppinfo` directories. It is probably more obvious if you try to import the shared libraries or a tool, you'll need to use the `cppinfo.libdirs` or `cppinfo.bindirs`, How else would you know where to locate them?. The lesson learned is, using "dep.package_folder" for an import is not accurate and won't work with editable packages.

memsharded:
> Why is exactly tricky? The copy from self.source_folder is necessary to package to the cache only, but how else could it be?

I had to do some failed attemps to use ``self.cpp.source.resdirs`` (trying to be symmetric with the consumer case), until realized that it was impossible, because at the ``package()`` step, the absolute paths for ``resdirs`` haven't been computed yet (they are made absolute when ``package_info()`` is called, but that happens after ``package()``. So even if ``resdirs`` has been defined, I need to "hardcode" the ``os.path.join(self.source_folder, "res")`` thing:

```python
def package(self):
       # resdir = self.cpp.source.resdirs[0]  # this FAILED; as resdirs is still relative here, not found
       resdir = os.path.join(self.source_folder, "res")
```

lasote:
I understand, but, typically (except layout declaration) these directories are declared at the package_info. For me that feels natural, you describe with the cppinfo where are the files located, not the opposite. With the old self.copy we always indicate the directory "manually":
`self.copy("*.dll", src="CPP/7zip", dst="bin", keep_path=False)`
we indicate "bin" not self.cpp_info.bindirs[0], that is for the consumers.

czoido (approved):

PR #10547, opened 2022-02-10
2.0 moved download to cli

lasote:
Changes to "conan download": 

- "conan download" moved to cli and new api
- "conan download <expression>" requires a single "-r", allowing multiple could be messing resolving the expression in multiple remotes and then downloading from the resolved one.
- "conan download <ref>" will download only the recipe
- "conan download lib/1.0:package_id" is not valid, a recipe revision has to be specified
- "conan download lib/1.0#latest" is valid
- "conan download lib/1.0#* -p 'os=Windows`" is valid

Changes to "conan list packages":

- "conan list packages lib/1.0#latest" is now valid


memsharded (approved):
Looks great

czoido (approved):

PR #10548, opened 2022-02-10
Remove/develop2 tools files legacy

memsharded:


conans/test/unittests/client/tools/files/remove_files_by_mask_test.py
```diff
@@ -1,14 +1,14 @@
...
-from conans.client.tools.files import chdir
+import pytest
+
 from conans.test.utils.test_files import temp_folder
-from conans.util.files import save_files
+from conans.util.files import save_files, chdir
 
 
 class RemoveFilesByMaskTest(unittest.TestCase):
+    @pytest.mark.xfail(reason="remove_files_by_mask not migrated yet")
```
memsharded, line 11: @lasote expected, right? We agreed to not implement this initially, but to replace with a search+remove?

conans/test/unittests/util/conanfile_tools_test.py
```diff
@@ -55,17 +55,18 @@ def test_untar(self):
...
 
     def test_replace_in_file(self):
         tmp_dir = temp_folder()
         text_file = os.path.join(tmp_dir, "text.txt")
         save(text_file, "ONE TWO THREE")
-        replace_in_file(text_file, "ONE TWO THREE", "FOUR FIVE SIX")
+        replace_in_file(ConanFileMock(), text_file, "ONE TWO THREE", "FOUR FIVE SIX")
         self.assertEqual(load(text_file), "FOUR FIVE SIX")
 
+    @pytest.mark.xfail(reason="_add_write_permissions has been removed, this no longer pass")
```
memsharded, line 69: @lasote not sure about this, is something that we want not to implement? It seems there was a use case there for this.

conans/test/unittests/client/tools/files/remove_files_by_mask_test.py
lasote: I think we finally agree that the weird was the name, we have to replace it with a simple `def remove(conanfile, pattern, folder)`

memsharded: Ok, so it is good to mark this test as xfail at this moment to remove the legacy, and we will recover it when the new ``remove()`` is implemented

conans/test/unittests/util/conanfile_tools_test.py
lasote: I really don't know, but I think that if you have no write permissions, you can assign them manually (maybe another tool?) and then replace, right?

lasote (approved):

conans/test/unittests/util/conanfile_tools_test.py
memsharded: I guess one problem could be the "CONAN_READ_ONLY_CACHE" behavior that puts the whole cache in read-only mode, so you might find that your ``replace_in_file()`` starts to fail when it was not failing before, because something changed in the system. I honestly don't know what to do with the CONAN_READ_ONLY_CACHE feature, I understand the use case, and it is typical of other package managers like pip to put things in read-only in the installation, but still reads dirty for us, specially when later you want to deploy locally, and we are reverting the permissions back.

PR #10552, opened 2022-02-10
changing CMakeToolchain message to Verbose

### LINKED ISSUE #10550: nitpicks on "Using conan toolchain" log message
  * Compiler+version: gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
  * Conan version: 1.45.0
  * Python version: 3.8.10

I have some complicated projects (using winegcc) that do pretty custom probing of the underlying gcc compiler prior to letting CMake do its normal CMakeDetect*Compiler with winegcc/wineg++. This is done using a similar techniques to [CheckLanguage](https://cmake.org/cmake/help/latest/module/CheckLanguage.html) - configuring a sub-project and then pulling certain variables back into the toplevel by reading a result.cmake. With the conan_toolchain.cmake, this results in a somewhat annoying degree of log-spam from messages.

> Using Conan toolchain: .../conan_toolchain.cmake.

https://github.com/conan-io/conan/blob/03801e6a395dd971c8778a743774deb2fc8c180e/conan/tools/cmake/toolchain.py#L798-L800

I have a couple of minor quibbles with this message:

1. It prints `${CMAKE_TOOLCHAIN_FILE}` rather than `${CMAKE_CURRENT_LIST_FILE}` - so if conan's toolchain was ever composed into/incuded by something else, it will print the name of that something else (presumably defeating the point of this report identifying the conan-generated file).

2. It's kind of weird in that it tests if(CMAKE_TOOLCHAIN_FILE) after it clearly *has* been included somehow - I *think* this was probably intended as a way to avoid repeating the message when CMakeDetermine*Compiler* does various try_compile calls; these get the toolchain file by reusing the same CMAKE_PLATFORM_INFO_DIR}/CMakeSystem.cmake, rather than via CMAKE_TOOLCHAIN_FILE directly. But if that is indeed the goal, it would might be clearer to put this messaging in with TryCompileBlock and just have that be an if/else reporting on whether/when the toolchain is used at top-level vs inside a try_compile. 

2. the [`message([mode])`](https://cmake.org/cmake/help/latest/command/message.html) parameter to is not specified, leaving it to default to `NOTICE`. That means it's treated as 
   > Important message printed to stderr to attract user's attention.

   and thus escapes from things that capture messages using OUTPUT_VARIABLE (because `NOTICE` goes to stderr). I think this kind "yep, I'm using the file you said to use" chatter should be at most `STATUS`, and personally I'd vote for dropping it to `VERBOSE`)

memsharded:
Changelog: Fix: Change the ``CMakeToolchain`` message to use ``CMAKE_CURRENT_LIST_FILE``. 
Docs: Omit
Close https://github.com/conan-io/conan/issues/10550

### DIFF
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -795,7 +795,7 @@ class CMakeToolchain(object):
         #   CMAKE_CXX_FLAGS. See https://github.com/android/ndk/issues/323
         include_guard()
 
-        message(VERBOSE "Using Conan toolchain: ${CMAKE_CURRENT_LIST_FILE}")
+        message(STATUS "Using Conan toolchain: ${CMAKE_CURRENT_LIST_FILE}")
 
         if(${CMAKE_VERSION} VERSION_LESS "3.15")
             message(FATAL_ERROR "The 'CMakeToolchain' generator only works with CMake >= 3.15")
--- a/conan/tools/cmake/toolchain.py
+++ b/conan/tools/cmake/toolchain.py
@@ -795,9 +795,7 @@ class CMakeToolchain(object):
         #   CMAKE_CXX_FLAGS. See https://github.com/android/ndk/issues/323
         include_guard()
 
-        if(CMAKE_TOOLCHAIN_FILE)
-            message("Using Conan toolchain: ${CMAKE_TOOLCHAIN_FILE}.")
-        endif()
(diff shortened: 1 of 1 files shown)


memsharded:
I am reverting the mode to STATUS, I think it will be better if by default the message shows, without requiring verbose, as that really helps in triaging and doing support, and also increasing user awareness of what is happening.

puetzk:
`NOTICE` was excessive, but I'm fine with `STATUS`.

PR #10564, opened 2022-02-13
Quotes are needed, implementation copied from `cmake_find_package`

jgsogo:
Trying some `exelinkflags` like `-s FULL_ES3=1` when building with emsdk


memsharded (commented):
If this is failing for some case, please:

- Add a test that covers it. 
- And what is more important, extend the test to cover CMakeDeps, as ``cmake_find_package_multi`` has already been removed in 2.0

jgsogo:
👍  I wanted to check first if it was tested somewhere else in the other direction...

jgsogo:
The main issue here is how CMake pass these flags to the linker. Without quotes, the linker tries to link with `-lFULL_ES3=1`. These tests are not checking the linker line... maybe I should do it to be sure how it interprets quotes, `;` and blanks. How do you feel about checking the files generated by CMake for this use case?

conans/test/unittests/tools/cmake/test_cmakedeps.py
```diff
@@ -163,3 +163,41 @@ def test_component_name_same_package():
...
+        req = Requirement(ConanFileReference.loads("OriginalDepName/1.0"))
+        mock_deps.return_value = ConanFileDependencies({req: ConanFileInterface(conanfile_dep)})
+
+        cmakedeps = CMakeDeps(conanfile)
+        files = cmakedeps.content
+        data_cmake = files["mypkg-release-x86-data.cmake"]
+        assert 'set(mypkg_mypkg_cmp1_SHARED_LINK_FLAGS_RELEASE "-NODEFAULTLIB;-OTHERFLAG")' in data_cmake
+        assert 'set(mypkg_mypkg_cmp1_EXE_LINK_FLAGS_RELEASE "-OPT:NOICF;-s FULL_ES3=1")' in data_cmake
+        assert 'set(mypkg_mypkg_cmp1_OBJECTS_RELEASE "${mypkg_PACKAGE_FOLDER_RELEASE}/myobject.o")' \
+               in data_cmake
```
franramirez688: This test is duplicated. Have a look at this PR (merged) https://github.com/conan-io/conan/pull/9980/files

conans/test/unittests/client/generators/cmake_find_package_multi_test.py
```diff
@@ -44,12 +44,36 @@ def test_cmake_find_package_multi_links_flags():
...
+    cpp_info.components["cmp1"].exelinkflags = ["-OPT:NOICF", "-s FULL_ES3=1"]
+    conanfile.deps_cpp_info.add("mypkg", DepCppInfo(cpp_info))
+
+    gen = CMakeFindPackageMultiGenerator(conanfile)
+    files = gen.content
+    d = files["mypkgTarget-release.cmake"]
+
+    assert '"$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,EXECUTABLE>:-OPT:NOICF;-s FULL_ES3=1>"' in d
+    assert '"$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,MODULE_LIBRARY>:-NODEFAULTLIB;-OTHERFLAG>"' in d
+    assert '"$<$<STREQUAL:$<TARGET_PROPERTY:TYPE>,SHARED_LIBRARY>:-NODEFAULTLIB;-OTHERFLAG>"' in d
```
franramirez688, line 79: This test is fine, but I think it could be better to have a real one. We could add it in `conans/test/functional/generators/cmake_find_package_multi_test.py` and check that it's correctly running with some real link flags like the PR link pasted in the previous comment.

franramirez688 (changes requested):

conans/test/unittests/client/generators/cmake_find_package_multi_test.py
jgsogo: Probably I can add some actual `create` here `conans/test/functional/generators/cmake_find_package_multi_test.py`, using flags that should be available in any regular compiler... I'm not going to add a emsdk build just because of this 😄

jgsogo:
I've added a test that fails, @franramirez688 , but I need to have a closer look at it. Not sure about it.

jgsogo:
After having a quick look to some compilers:

**Macos**
```
clang --help
  -Xlinker <arg>          Pass <arg> to the linker
  -z <arg>                Pass -z <arg> to the linker
```

**Linux**
```
gcc --help
  -Xlinker <arg>           Pass <arg> on to the linker.
```

```
clang --help
  -Xlinker <arg>          Pass <arg> to the linker
  -z <arg>                Pass -z <arg> to the linker
```

So, in theory, when we have `cpp_info.sharedlinkflags`, `exelinkflags`,... what should we provide? the full `-z xxx` or just `xxx`? Should Conan take care of adding `-z` or `-Xlinker` to those arguments?

```python
        def package_info(self):
            self.cpp_info.sharedlinkflags = ["-z now", "-z relro"]
            self.cpp_info.exelinkflags = ["-z now", "-z relro"]
```

This is important because if we add quotes around these arguments, the compiler sees something like `compiler "-z now"` and doesn't realize it is the argument `-z` but a single information unit. On the other hand if we need to pass something like `-s FULL_ES3=1` to the linker we would need `compiler -Xlinker "-s XXXX"` so quotes ensure that it is a single argument to pass to the linker.

I'm not sure we are doing this right, and probably (as a workaround) we should be using:
```
            self.cpp_info.sharedlinkflags = ["-z", "now", "-z",  "relro"]
```
so we can also use:
```
            self.cpp_info.sharedlinkflags = ["-z", "-s XXXX"]. # Do we need the -z??
```
and they are converted into the right number of arguments.

---

Of course, this needs further investigation, tests and examples.

memsharded:
we are no longer modifying cmake_find_xxxx generators unless very broken. This has risk and it is not worth.

Please target the issues or changes to ``CMakeDeps`` instead.

PR #10573, opened 2022-02-14
adding cli-args to CMake new helper calls

### LINKED ISSUE #10515: [question] How to pass cmake configuration option with modern conan?
I need to pass cmake configuration option to be able to modify the project version. Unfortunately, by using toolchain variables, those variable aren't already loaded when the project() cmake function is called.

How could I achieve that?

- [ x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

### LINKED ISSUE #10562: [feature] Add ability to pass arguments to ctest in conan.tools.cmake.CMake.test
The `CMake.test()` interface in `conan.tools.cmake` is much more limited in comparison to that of `conans.CMake.test()`. In `conan.tools.cmake.CMake` (the new version) we have the following definition:

```python
    def test(self, build_type=None, target=None, output_on_failure=False):
        if self._conanfile.conf["tools.build:skip_test"]:
            return
        if not target:
            is_multi = is_multi_configuration(self._generator)
            target = "RUN_TESTS" if is_multi else "test"
        self._build(build_type=build_type, target=target)
```
Whereas the old implementation, in `conans.CMake` was:

```python
    def test(self, args=None, build_dir=None, target=None, output_on_failure=False):
        if not self._conanfile.should_test or not get_env("CONAN_RUN_TESTS", True) or \
           self._conanfile.conf["tools.build:skip_test"]:
            return
        if not target:
            target = "RUN_TESTS" if self.is_multi_configuration else "test"

        test_env = {'CTEST_OUTPUT_ON_FAILURE': '1' if output_on_failure else '0'}
        if self.parallel:
            test_env['CTEST_PARALLEL_LEVEL'] = str(cpu_count(self._conanfile.output))
        with environment_append(test_env):
            self._build(args=args, build_dir=build_dir, target=target)
```
Specifically, it looks like the `args` keyword argument has been removed from the interface of both `test()` and `_build()`. This is limiting, because I want to be able to customize the way my unit test output displays for greater granularity so that I can see which specific tests failed and what the error was - mostly by passing the `-V` flag for additional verbosity, but in principle I think the ideal case here would be to be able to utilise any of the `ctest` command line options.

Is there a reason this has been removed in the new interface, and if not, can it be re-added?

For completeness, the way I have been invoking unit tests in my pre-conan 2.0 recipes has been:

```python
    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()
        if self.options.with_tests:
            os.environ["GTEST_COLOR"] = "1"
            cmake.test(args=["--", "ARGS=-V"], output_on_failure=True)
```

memsharded:
Changelog: Feature: Add CMake variables, cli arguments and native build system arguments to new ``conan.tools.cmake.CMake`` helper. 
Docs: https://github.com/conan-io/docs/pull/2424
Close https://github.com/conan-io/conan/issues/10515. 
Close https://github.com/conan-io/conan/issues/10562

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -60,7 +60,7 @@ class CMake(object):
 
         self._cmake_program = "cmake"  # Path to CMake should be handled by environment
 
-    def configure(self, build_script_folder=None):
+    def configure(self, variables=None, build_script_folder=None):
         cmakelist_folder = self._conanfile.source_folder
         if build_script_folder:
             cmakelist_folder = os.path.join(self._conanfile.source_folder, build_script_folder)
@@ -83,7 +83,10 @@ class CMake(object):
             pkg_folder = self._conanfile.package_folder.replace("\\", "/")
             arg_list.append('-DCMAKE_INSTALL_PREFIX="{}"'.format(pkg_folder))
         if platform.system() == "Windows" and self._generator == "MinGW Makefiles":
+            # It seems this doesn't work in the toolchain file, it needs to be here in command line
             arg_list.append('-DCMAKE_SH="CMAKE_SH-NOTFOUND"')
+        if variables:
+            arg_list.extend(["-D{}={}".format(k, v) for k, v in variables.items()])
         arg_list.append('"{}"'.format(cmakelist_folder))
 
         command = " ".join(arg_list)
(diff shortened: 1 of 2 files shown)


lasote (commented):
~About the `cmake.configure(variables={"MYVAR": "MYVALUE"})`, shouldn't they be set in the cmake toolchain? Wouldn't this go against dummy builds?~
I'm reading the associated issue and now I get it. I would make sure that we explain it correctly in the documentation, these variables should be used only when you have no other choice.

lasote (approved):

PR #10574, opened 2022-02-14
2.0 remove develop and dev reference

lasote:
I've tried to make it more clear, assign a private variable `_conan_is_consumer` when creating a root node and when it is required by a test_package or a virtual. Not very happy but I like it a bit better.
Suggestions are welcome, I didn't find a way to achieve it without saving the state in the conanfile given that both the test_package and the package being tested are "consumers" and also a virtual when "conan install ." etc.


memsharded (approved):
Yes, I agree, even if not super great, it is already better than the user visible ``develop`` variable, and this is more direct.

If we could get rid of the "consumer" definition, like always forcing a package name or something, then this problem might be completely gone.

PR #10578, opened 2022-02-14
Feature/develop2 editable create build

memsharded:
- Implement tests that capture current behavior, which is NOT failing for other commands (conan list, conan export-pkg, conan upload, etc), as those are valid operations over cache packages
- Allows building editable packages directly from consumers ``conan install --build=editable``. Needs to be explicit, ``--build=missing`` will not make it, as it is not possible to differentiate if the package is built or not. Maybe we want to add some ``--build=editable`` too, or assume ``missing``=>build for editable, which could make sense. TO DISCUSS


lasote (approved):
Cool. I would add (later if requested) the `--build editable` I don't think a `--build missing` should mean always "build the editables too". But we can go with the name meanwhile.

memsharded:
> Cool. I would add (later if requested) the --build editable I don't think a --build missing should mean always "build the editables too". But we can go with the name meanwhile.

Actually, I was playing with the feature a bit, and it was really poor UX. ``--build=missing`` didn't  mean to build the editables, it didn't, so the only way to play with the feature was building all, which includes dependencies in the cache too, or doing it explicitly ``--build=mypkg1...``. 

It was just a very few lines, so I finally added the ``--build=editable`` flag, and added tests for it.

czoido (approved):

Issue #10563, opened 2022-02-13
[bug] Invalid defines in Bazel build file when the define is a string type


mohabouje:
Running into this issue on the latest commit in dev branch when trying to use Dear ImGUI:

```
    defines = [
        "IMGUI_USER_CONFIG='imgui_user_config.h'"
    ],
```
That's an invalid definition in Bazel. 

The expected (works fine for me, probably there are other ways to do this) is:
```
    defines = [
        "IMGUI_USER_CONFIG=\\\"imgui_user_config.h\\\""
    ],
```

For now, it's not a big deal as I end up doing:

```
    defines = [
        "IMGUI_USER_CONFIG=\\\"imconfig.h\\\""
    ],
    copts = ["-w"],
```

disabling the GCC warnings.

The fix seems trivial when exporting defines from the Conan recipe, if it's a string, convert it to this format. 

I will prepare a pull request if I have some time this week.

memsharded:
Hi @mohabouje 

Yes, it seems relatively easy. If you can submit that PR, fantastic, if not, just tell, I think we might be able to try it to fix this for 1.46. Thanks for reporting and offering to help.

memsharded:
This has been fixed by https://github.com/conan-io/conan/pull/10812

PR #10581, opened 2022-02-15
Fixing string-alike defines in BazelDeps generator

mohabouje:
Originally:

    defines = [
        "IMGUI_USER_CONFIG='imgui_user_config.h'"
    ],

This is an invalid bazel define, it will read a define with no value. A valid alternative:

    defines = [
        "IMGUI_USER_CONFIG=\\\"imgui_user_config.h\\\""
    ],

The list of defined does not provide information about the type, we can assume that defines that contains `"xxx"` are string and perform a correct replacement.

Fix for: https://github.com/conan-io/conan/issues/10563

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -103,7 +103,7 @@ class BazelDeps(object):
                 return p[len(base_path):].replace("\\", "/").lstrip("/")
             return p.replace("\\", "/").lstrip("/")
 
-        # TODO: This only wokrs for package_folder, but not editable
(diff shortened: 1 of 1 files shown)


lasote:
Thanks for your help, I verified with a test (WIP at https://github.com/conan-io/conan/pull/10812) that you are totally right. As there are other things being improved in the Bazel integration of Conan and given that are some unit tests here failing that have to be fixed, I will take care of this issue in that PR. This will be released in the next 1.47, thanks again for your help!!!

mohabouje:
More than happy to help. Looking forward to try this new version

PR #10586, opened 2022-02-15
new update_conandata() helper

memsharded:
Changelog: Feature: New ``from conan.tools.files import update_conandata()`` helper to add data to ``conandata.yml`` in the ``export()`` method.
Docs: https://github.com/conan-io/docs/pull/2422

### DIFF
--- a/conan/tools/files/__init__.py
+++ b/conan/tools/files/__init__.py
@@ -6,3 +6,4 @@ from conan.tools.files.cpp_package import CppPackage
 from conan.tools.files.packager import AutoPackager
 from conan.tools.files.symlinks import symlinks
 from conan.tools.files.copy_pattern import copy
+from conan.tools.files.conandata import update_conandata
--- a/conan/tools/files/conandata.py
+++ b/conan/tools/files/conandata.py
@@ -1,0 +1,30 @@
+import os
+
+import yaml
+
+from conans.errors import ConanException
+from conans.util.files import load, save
+
+
+def update_conandata(conanfile, data):
+    """
+    this only works for updating the conandata on the export() method, it seems it would
+    be plain wrong to try to change it anywhere else
+    """
+    if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
+        raise ConanException("The 'update_conandata()' can only be used in the 'export()' method")
+    path = os.path.join(conanfile.export_folder, "conandata.yml")
+    conandata = load(path)
+    conandata = yaml.safe_load(conandata)
+
+    def recursive_dict_update(d, u):
+        for k, v in u.items():
+            if isinstance(v, dict):
+                d[k] = recursive_dict_update(d.get(k, {}), v)
+            else:
+                d[k] = v
+        return d
+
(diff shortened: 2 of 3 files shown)


santhonisz:
Hi @memsharded,

This looks to be a very useful function and one we were literally about to look at implementing ourselves. However can I ask you to consider an additional usage scenario: 

We would like to capture to some data in the `conandata.yml` for all our package recipes (literally 100s) when running in CI, and so were planning on using a `post_export` hook. I've checked and the passed `conanfile` arg doesn't have the `export_folder` attribute available for this hook type, hence this new function wouldn't be usable in this scenario. 

Does it sound reasonable to add support for this use case?

memsharded:
(BTW, this new helper is being added as it will be necessary to migrate some features to 2.0). 

> Does it sound reasonable to add support for this use case?

Yes, it could be possible, but I have had a look and seems it could be a bit dirty now in the 1.X, we can try to have a second look, when we revisit the hooks for Conan 2.0, we want to understand better their usage. Can you please describe a bit more what information you are adding to conandata.yml by the hook? Why in a hook and not directly in the recipe. As a rule of thumb:
- Anything that is necessary for the recipe and package to work correctly, should be in the recipe, it shouldn't rely on hooks to be installed to operate
- Hooks can be used as an extra thing, to augment, check, validate, run tools, etc, but if we drop them, the recipes and packages should still be valid.

santhonisz:
We currently only use hooks when running in our CI environment. They are stored in a dedicated git config repo and installed via Conan Package Tool's `CONAN_CONFIG_URL` env var. We primarily use them to perform validation against our organisations requirements in regards to our (private) packages e.g. the `revision_mode` attribute must be set to `scm`, the `license` attribute must be set etc. The advantage of doing this in a hook is that we can perform such additional actions against any package being built, while for developers the recipe still functions perfectly well even though they will not be using said hooks.

In the case at hand, we are now wanting to store some CI-related information in the `conandata.yml` for all packages. This information is primarily SCM-related which, among other things, may be used by recipes in their `source` method at a later date (much like your unit test). We expect the usage of the `source` method in their recipes to be opt-in for our teams (we currently do not build "missing" packages either locally or in CI), however we see real value in capturing this information across the board. A hook allows us to do this for all packages built in CI without having to update each individual recipe, of which there are 100s. 

Hope that clarifies things a bit.

lasote (approved):

memsharded:
> In the case at hand, we are now wanting to store some CI-related information in the conandata.yml for all packages. This information is primarily SCM-related which, among other things, may be used by recipes in their source method at a later date (much like your unit test). We expect the usage of the source method in their recipes to be opt-in for our teams (we currently do not build "missing" packages either locally or in CI), however we see real value in capturing this information across the board. A hook allows us to do this for all packages built in CI without having to update each individual recipe, of which there are 100s.

Thanks very much @santhonisz for the details.

As this is closed, I am going to create an issue for future consideration then, now it would be dirty to make the ``export_folder`` available to the hooks, but after 2.0 cleaning and refactoring, it might be more doable.

Issue #10248, opened 2021-12-30
[feature] Unify patterns


lasote:
Review all application to see where Conan accept inputs with patterns and unify the behavior. Some of the "inconsistencies" detected right now:

- `--build`: Accepts * at any point of the string matching the reference, also the NAME, also the NAME/VERSION (without using *)
- `-o`: Accepts literal  "*" meaning "all", also NAME but not NAME/VERSION. You cannot use `foo*` or similar
- `-s`: Accepts * at any point of the string matching the reference, also the NAME, and also the NAME/VERSION (without using *)

The behavior should be: A valid reference or `fnmatch`.  That means `-ozlib:shared=True` is not valid anymore and it should be `-o "zlib/*:shared=True`). But `-s zlib/1.2.11:shared=True` would be still valid, as it is a valid reference.

memsharded:
Try to remove the dependency to ``self.develop`` for the "consumer" patterns ``&``

lasote:
Review the "!" pattern, it is requested.
Related:
https://github.com/conan-io/conan/pull/10197

lasote:
Review also patterns in "conf" and "env"

memsharded:
Implemented in https://github.com/conan-io/conan/pull/10587

PR #8483, opened 2021-02-10
Exclude a package when building all from sources

uilianries:
We have 3 different ways to execute it, to illustrate:

Build all from sources, except zlib:

    conan install boost/1.75.0@ --build=!zlib --build=*
    conan install boost/1.75.0@ --build=!zlib --build

Do not build anything from sources. --build= means '' package, which is invalid.

    conan install boost/1.75.0@ --build=!zlib --build=

Changelog: Feature: Skip package when building all package from sources at once using `--build=!<package>` syntax.
Docs: https://github.com/conan-io/docs/pull/2023
closes #8442

- [x] Refer to the issue that supports this Pull Request.
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

PR #10587, opened 2022-02-15
Unify reference pattern matching

lasote:
At:

- `-s`
- `-o`
- `--build`
- `[conf]` and `[buildenv]` at profiles

Internal options management changed significantly, the key is not the name anymore but a reference. The old conf access now allow to do (collateral effect):

`self.options["zlib/*"].shared = self.options.shared`

Close https://github.com/conan-io/conan/issues/10248


conans/test/functional/configuration/profile_test.py
```diff
@@ -743,16 +743,20 @@ def configure(self):
...
+    client.run("create . -s foo*:build_type=Debug")
     assert "I'm foo and my build type is Debug" in client.out
 
-    client.run("create . -s foo:build_type=Debug -s &:build_type=Release")
+    client.run("create . -s foo*:build_type=Debug -s &:build_type=Release")
     assert "I'm foo and my build type is Release" in client.out
 
-    # The order doesn't matter
-    client.run("create . -s &:build_type=Release -s foo:build_type=Debug ")
+    # The order DOES matter !!! This behavior changed!
```
lasote: Notice this. Previously the order didn't matter, now it is applied in order.

memsharded:
There was a pending check in https://github.com/conan-io/conan/pull/8483#pullrequestreview-588483508, that couldn't be addressed in 1.X, we probably want to take it into account for 2.0 (basically the "negation" of patterns, and how that works together with general patterns as ``--build``).

lasote:
The last commit made that `--build` is equal to `--build=*`. On the other hand `--build=` is still incorrect because it is an empty value not matching any reference. We could change that too but I think that is ok.

lasote:
Approached https://github.com/conan-io/conan/pull/10197 too

memsharded:
Note: at some point we were talking about making it even more explicit: ``--build-policy=missing --build=pattern``, to dissambiguate between policies and real patterns

conans/cli/command.py
```diff
@@ -38,6 +39,15 @@ def __call__(self, parser, namespace, values, option_strings=None):  # @UnusedVa
...
                 dest.append(values)
+        else:  # When "--argument" with no value is specified
+            if self.default_when_none:
+                dest.append(self.default_when_none)
+
+
+class ExtenderDefaultStar(Extender):
+
+    # If --build is specified, it will take a "*" value
+    default_when_none = "*"
```
memsharded: This might be related to my note, are we sure we don't want to force it to be explicit in 2.0? Like ``--build=*``? or ``--build-policy=all``?

conans/model/options.py
```diff
@@ -270,16 +278,23 @@ def __setattr__(self, attr, value):
...
         self._package_options.__delattr__(field)
 
-    def __getitem__(self, item):
+    def __getitem__(self, ref):
         # To access dependencies options like ``options["mydep"]``. This will no longer be
         # a read case, only for defining values. Read access will be via self.dependencies["dep"]
-        return self._deps_package_options.setdefault(item, _PackageOptions())
-
-    def scope(self, name):
+        if isinstance(ref, str):
```
memsharded: I wonder if we could directly remove this, and raise? ``self.dependencies`` is in 1.X, there is no reason to keep this in 2.0?

conans/model/recipe_ref.py
```diff
@@ -109,3 +110,23 @@ def loads(rref):  # TODO: change this default to validate only on end points
...
+                      fnmatch.fnmatchcase(self.repr_notime(), pattern))
+        if negate:
+            return not condition
+        return condition
+
+
+def ref_matches(ref, pattern, is_consumer):
+    if not ref or not str(ref):
+        assert is_consumer
+        ref = RecipeReference.loads("*/*")
```
memsharded: Not sure about this ``RecipeReference.loads()`` with something that will fail if we start to add some validation there (it was removed in the refactor, but at some point it might make sense to do some checks). If it is going to match always, can't we return True directly?

conans/test/functional/configuration/profile_test.py
```diff
@@ -694,7 +694,7 @@ def configure(self):
     assert "I'm None and my shared is False" in client.out
 
     # Now the dependency by name
-    client.run("install . -s dep:build_type=Debug -o dep:shared=True")
+    client.run("install . -s dep/*:build_type=Debug -o dep*:shared=True")
```
memsharded:
```suggestion
    client.run("install . -s dep/*:build_type=Debug -o dep/*:shared=True")
```

I guess it worked because there were no ``dep_other`` package? Was this intended?

conans/test/functional/configuration/profile_test.py
```diff
@@ -743,16 +743,20 @@ def configure(self):
...
     assert "I'm foo and my build type is Debug" in client.out
 
-    client.run("create . -s foo:build_type=Debug -s &:build_type=Release")
+    client.run("create . -s foo*:build_type=Debug -s &:build_type=Release")
     assert "I'm foo and my build type is Release" in client.out
 
-    # The order doesn't matter
-    client.run("create . -s &:build_type=Release -s foo:build_type=Debug ")
+    # The order DOES matter !!! This behavior changed!
+    client.run("create . -s foo*:build_type=Debug -s &:build_type=Release")
```
memsharded: Isn't this identical to line 749?
Also this is one of the lines that is difficult to understand. Why using ``&`` if we have the package name? For convenience?

conans/test/integration/command_v2/test_info_build_order.py
```diff
@@ -134,7 +134,7 @@ def test_info_build_order_options():
                         "context": "build",
                         "binary": "Build",
                         "options": [
-                                "tool:myopt=1"
+                                "tool/0.1#b4c19a1357b43877a2019dd2804336a9:myopt=1"
```
memsharded: It seems the revision is not necessary and ``--build=tool/0.1`` or ``-o tool/0.1:myopt=1`` will work perfectly too. Do we want to simplify here?

conans/test/integration/configuration/conf/test_conf.py
```diff
@@ -109,7 +109,7 @@ def test_new_config_file(client):
     conf = textwrap.dedent("""\
         tools.microsoft.msbuild:verbosity=Minimal
         user.mycompany.myhelper:myconfig=myvalue
-        *:tools.cmake.cmake:generator=X
+        */*:tools.cmake.cmake:generator=X
```
memsharded: Doesn't the ``*`` already match everything? Is this change necessary?

memsharded (commented):
There is a lot to digest here in this PR. 
I would need at least another pass to start understanding, but sending already some comments to start talking.

In any case, it seems definitely a very good step, and this is good job to update all the tests, etc.

conans/cli/command.py
lasote: Yes, we want.

conans/model/options.py
lasote: Probably, I know there is the usage of this but I will check where exactly apart from the tests.

conans/model/recipe_ref.py
lasote: I agree that is ugly. But no, you cannot return True directly, maybe the pattern is `!&` or an invalid one like `potato`.

conans/test/functional/configuration/profile_test.py
lasote: > Why using & if we have the package name? For convenience?

hahaha, yes, you use `&` because you can. In this case you could use the name but the "consumer" mechanism, unless we modify it, allows to use `&` in a create like before.

Yes, the line is repeated. I'll remove it.

conans/test/integration/configuration/conf/test_conf_profile.py
```diff
@@ -87,7 +88,7 @@ def test_cmake_config_package(client):
         compiler.runtime=MD
         build_type=Release
         [conf]
-        dep*:tools.microsoft.msbuild:verbosity=Minimal
+        dep*/*:tools.microsoft.msbuild:verbosity=Minimal
```
lasote: Same. It shouldn't be necessary. I'll check.

conans/model/options.py
lasote: I have removed but there are some tests failing now accessing options in configure like:

```
def configure(self):
            self.options["zlib/*"].shared = self.options.shared
```

lasote: Recovered getattr

conans/model/conf.py
```diff
@@ -53,7 +55,6 @@
     "tools.system.package_manager:mode": "Mode for package_manager tools: 'check' or 'install'",
     "tools.system.package_manager:sudo": "Use 'sudo' when invoking the package manager tools in Linux (False by default)",
     "tools.system.package_manager:sudo_askpass": "Use the '-A' argument if using sudo in Linux to invoke the system package manager (False by default)",
-    "tools.apple.xcodebuild:verbosity": "Verbosity level for xcodebuild: 'verbose' or 'quiet"
```
memsharded: Why removed this conf?

conans/model/options.py
```diff
@@ -199,6 +198,12 @@ def __init__(self, options=None, options_values=None):
                     tokens = k.split(":", 1)
                     if len(tokens) == 2:
                         package, option = tokens
+                        if "/" not in package:
```
memsharded: But if it is a pattern it will work? Having ``pkg*`` will be ok too?
```suggestion
                        if "/" not in package: and "*" not in package:
```

memsharded (approved):
I think it is good to move forward

conans/model/conf.py
lasote: no idea, a bad merge probably. will recover it

memsharded (approved):

PR #10589, opened 2022-02-15
[develop2] removing cache generators, use deployers

memsharded:
The previous in-cache user custom generators were not longer working in 2.0.
There was the idea of defining them in a similar way to the deployers, but while trying to implement that, I have found a few issues:

- It is problematic to have name collisions or hiding. If a user puts a ``CMakeDeps.py`` file in the cache or local folder, is it going to override the built-in ``CMakeDeps`` generator? If that happens it will only happen on the final consumer case, but not in the cache creation or transitive dependency build from source, which is a disaster.
- Permanent overwriting of a built-in generator is even worse, that introduces a variability over the binaries that is completely out of control, and can be very deep. There is a well defined mechanism that is using python_requires, as python_requires affect the ``package_id`` and can be used to model binary evolution. Cache generators are global, uncontrolled state.
- Questions whether they should run after, before the built-in generators. For example the builtin behavior for ``VirtualBuildEnv`` can collect files from different tools into one single ``conanbuild.bat|sh`` script, if we want this to be applied for user generators they should run before the built-in ones.

In essence, it seems that there is nothing (besides the class or function interface) different than the new ``deployers``, which:
- Run before the generators
- Can apply to end consumers by construction, no way they can overwrite or be used in the cache
- They have a different name, which is good to differentiate from the built-in generators too.

So maybe the custom cache (the recipe ones, are python-requires) generators can be dropped and superseded completely by deployers?


lasote (commented):
But the deployers cannot be used in a recipe, right? What if I want to write my `JanderBuild` generator for my build system and use it? EDITED: So this use case is only left for python requires?

memsharded:
> But the deployers cannot be used in a recipe, right? What if I want to write my JanderBuild generator for my build system and use it? EDITED: So this use case is only left for python requires?

Yes, exactly. If you want something that can be used in recipes, the approach is:
- Put the logic in a ``python_require``
- In the consumer ``python_require`` it, and instantiate and use what you want, explicitly in the ``generate()`` method.

The deployers are definitely different in the sense that they cannot be used/consumed by a recipe, only by the final consumer/user. So ``ycm`` is a deployer, a recipe never wants that. ``markdown`` is another deployer, etc. 

It seems it fits reasonably well, the only precaution that needs to be taken into account is call deployers that change the final destination of the package before calling other deployers that will use the final destination of the package for something.

lasote (approved):
Not very passionate about the python requires but I like to simplify the amount of ways to declare a generator.

czoido (approved):

czoido:
We should be very clear in the documentation with this difference from what's a deployer that used to be a generator like `markdown` and what's a generator that goes in a `python_requires` so that people is not confused with this in 2.0

PR #10594, opened 2022-02-16
introducing modern conan.tool.scm.Git helper

memsharded:
Changelog: Feature: Introduce new ``conan.tools.scm.Git`` helper, for direct use in ``export()`` method to capture git url and commit, and to be used in ``source()`` method to clone and checkout a git repo.
Docs: https://github.com/conan-io/docs/pull/2419

This is a first iteration, containing only the capturing code that enables the old ``scm`` flow:

- Very flat design, no inheritance, everything explicit.
- Only the minimum methods to implement ``scm`` old flow
- Powerful flexibility can implement the Git monorepo case

AFTER https://github.com/conan-io/conan/pull/10654

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -4,6 +4,7 @@ import gzip
 import hashlib
 import os
 import platform
+import shutil
 import subprocess
 import sys
 from contextlib import contextmanager
@@ -14,6 +15,7 @@ import six
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
 from conans.client.downloaders.download import run_downloader
 from conans.errors import ConanException
+from conans.util.files import rmdir
 
 if six.PY3:  # Remove this IF in develop2
     from shutil import which
@@ -491,3 +493,20 @@ def collect_libs(conanfile, folder=None):
                     result.append(name)
     result.sort()
     return result
+
+
+def swap_child_folder(parent_folder, child_folder):
+    """ replaces the current folder contents with the contents of one child folder. This
+    is used in the SCM monorepo flow, when it is necessary to use one subproject subfolder
+    to replace the whole cloned git repo
+    """
+    for f in os.listdir(parent_folder):
+        if f != child_folder:
+            path = os.path.join(parent_folder, f)
+            if os.path.isfile(path):
+                os.remove(path)
+            else:
+                rmdir(path)
+    child = os.path.join(parent_folder, child_folder)
+    for f in os.listdir(child):
+        shutil.move(os.path.join(child, f), os.path.join(parent_folder, f))
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -495,6 +495,7 @@ def collect_libs(conanfile, folder=None):
     return result
 
 
+# TODO: Do NOT document this yet. It is unclear the interface, maybe should be split
 def swap_child_folder(parent_folder, child_folder):
     """ replaces the current folder contents with the contents of one child folder. This
     is used in the SCM monorepo flow, when it is necessary to use one subproject subfolder
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -7,7 +7,7 @@ from conans.util.runners import check_output_runner
 
 
 class Git(object):
-    def __init__(self, conanfile, folder):
+    def __init__(self, conanfile, folder="."):
         self._conanfile = conanfile
         self.folder = folder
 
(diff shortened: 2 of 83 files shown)


conans/test/utils/scm.py
```diff
@@ -48,6 +59,17 @@ def create_local_git_repo(files=None, branch=None, submodules=None, folder=None,
     return tmp.replace("\\", "/"), git.get_revision()
 
 
+def git_change_and_commit(files, folder, msg="fix"):
+    save_files(folder, files)
+    cwd = os.getcwd()
+    try:
+        os.chdir(folder)
```
SSE4, line 65: I think `chdir` might be avoided somehow, e.g. using `git -C <folder>` within commands
it should simplify the code, and should left cwd untouched in case conan is crashed/aborted

conans/client/source.py
```diff
@@ -105,7 +105,9 @@ def get_sources_from_exports():
                 # First of all get the exported scm sources (if auto) or clone (if fixed)
                 _run_cache_scm(conanfile, scm_sources_folder, output)
                 # so self exported files have precedence over python_requires ones
-                merge_directories(export_folder, conanfile.folders.base_source)
+                if not hasattr(conanfile, "layout"):
```
memsharded: This was a necessary change to implement the SCM flow, as the conandata.yml happens to be copied the first thing to the cache "source" folder, so trying to do a clone in the cache in that folder fails

conan/tools/scm/git.py
```diff
@@ -0,0 +1,85 @@
...
+            branches = self._run("branch -r --contains {}".format(commit))
+            return "{}/{}".format(remote, branch) in branches
+        except Exception as e:
+            raise ConanException("Unable to check remote commit in '%s': %s" % (self.folder, str(e)))
+
+    def is_dirty(self):
+        status = self._run("status -s").strip()
+        return bool(status)
+
+    def get_url_commit(self, remote="origin", branch="master"):
```
czoido: I think we should change this default to `main`.
Also since git>2.28 there's the `init.defaultBranch` configuration option that we may check using `git config --get init.defaultBranch`

lasote: I don't like the name `get_url_commit`, it reads like `get the URL of the commit` and makes no sense.
About the `remote` and `branch` defaults, maybe we should make it positional (not defaulted)? I would avoid extra commands like the `git config --get init.defaultBranch` suggested, even more, if they depend on the git version.

czoido: Maybe we can avoid using the branch and just use the commit hash? Do we need the branch name?

memsharded: Agree, need to find a better name.

In any case, the branch needs to be worked out. It shouldn't be necessary, and indeed is very problematic for user flows like branching, PRs, in which the branch will not be "main" or "master". So the commit should be looked for directly in the remote without a branch hint, I'll investigate if this is possible.

conan/tools/files/files.py
```diff
@@ -491,3 +493,20 @@ def collect_libs(conanfile, folder=None):
                     result.append(name)
     result.sort()
     return result
+
+
+def swap_child_folder(parent_folder, child_folder):
```
memsharded, line 499: Name to be improved. Suggestions welcome.

conan/tools/scm/git.py
memsharded: I have changed the name to ``get_url_and_commit()`` and removed the ``branch`` argument, now checking if the commit is in any branch. Added a test with an actual branch pushed, to see it still works.

conans/test/utils/scm.py
```diff
@@ -48,6 +59,20 @@ def create_local_git_repo(files=None, branch=None, submodules=None, folder=None,
     return tmp.replace("\\", "/"), git.get_revision()
 
 
+def git_change_and_commit(files, folder, msg="fix"):
```
lasote: I don't like this function name, is confusing.

lasote (approved):

conan/tools/files/files.py
lasote: Maybe a generic `move_folder_contents(dst_folder, folder, clean_first=True)` ?

czoido: I think maybe we can split the operation in two, one for removing all folders but the one we want and other for moving the folder... or `move_folder_contents(dst_folder, folder, clean_first=True)` is more explicit, I'm ok with that too, maybe `clean_other` instead of `clean_first`?

czoido (approved):

conan/tools/files/files.py
memsharded: I am annotating it to NOT document, so we can think about file methods in a more broad perspective, this PR is focused on the Git integration for sources, this is a very minor detail, just to prove it is possible to implement the mono-repo approach

PR #10595, opened 2022-02-16
Add support for gcc 12

voidanix:
Changelog: Feature: Add ``compiler.version`` 12 for GCC in settings.
Docs: Omit

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -80,7 +80,8 @@ _t_default_settings_yml = Template(textwrap.dedent("""
                       "8", "8.1", "8.2", "8.3", "8.4",
                       "9", "9.1", "9.2", "9.3",
                       "10", "10.1", "10.2", "10.3",
-                      "11", "11.1", "11.2"]
+                      "11", "11.1", "11.2",
+                      "12"]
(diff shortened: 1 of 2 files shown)


conans/client/conf/__init__.py
```diff
@@ -80,7 +80,7 @@
                       "8", "8.1", "8.2", "8.3", "8.4",
                       "9", "9.1", "9.2", "9.3",
                       "10", "10.1", "10.2", "10.3",
-                      "11", "11.1", "11.2"]
+                      "11", "11.1", "11.2", "12"]
```
memsharded: Better put it in its own line, in anticipation of future minors
```suggestion
                      "11", "11.1", "11.2", 
                      "12"]
```

memsharded (commented):
When introducing a new compiler version, there is typically one thing to check: what is the default c++ standard (the ``compiler.cppstd`` setting), in case it is changing, not usual, but it could potentially happen. Also if there is some change in the versioning scheme (also very unusual, but just in case)

voidanix:
None of the things you mentioned changed, did check the GCC man page before PR-ing.

memsharded:
Thanks!

PR #10600, opened 2022-02-16
fix meson_layout

memsharded:
Changelog: bugfix: Fix ``meson_layout()`` issue with shared folders.
Docs: Omit

### DIFF
--- a/conan/tools/meson/layout.py
+++ b/conan/tools/meson/layout.py
@@ -1,4 +1,8 @@
+import os
+
+
 def meson_layout(conanfile):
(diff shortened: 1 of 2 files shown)


conan/tools/meson/layout.py
```diff
@@ -1,4 +1,8 @@
+import os
+
+
 def meson_layout(conanfile):
     conanfile.folders.build = "build-{}".format(str(conanfile.settings.build_type).lower())
+    conanfile.folders.generators = os.path.join(conanfile.folders.build, "conan")
```
memsharded, line 6: Missing this was hiding in test the fact that shared libs doesn't work, because dll was in "lib" folder, not in "bin"

PR #10612, opened 2022-02-18
change source folder

memsharded:
Changelog: Fix: Make the ``source()`` method run inside the ``self.source_folder``, in the same way ``build()`` runs in ``self.build_folder``. But only for recipes that define the ``layout()`` method, to not break unless using ``layout()``.
Docs: https://github.com/conan-io/docs/pull/2418

This is my second try to change the source-folder of the ``source()`` method. Please have a look at the tests for more details.

### DIFF
--- a/conans/test/functional/layout/test_in_cache.py
+++ b/conans/test/functional/layout/test_in_cache.py
@@ -26,7 +26,7 @@ def conanfile():
     def source(self):
         self.output.warn("Source folder: {}".format(self.source_folder))
         # The layout describes where the sources are, not force them to be there
-        tools.save("my_sources/source.h", "foo")
+        tools.save("source.h", "foo")
 
     def build(self):
         self.output.warn("Build folder: {}".format(self.build_folder))
@@ -288,7 +288,7 @@ def test_git_clone_with_source_layout():
                    self.folders.source = "src"
 
                def source(self):
-                   self.run('git clone "{}" src')
+                   self.run('git clone "{}" .')
        """).format(repo.replace("\\", "/"))
 
     client.save({"conanfile.py": conanfile,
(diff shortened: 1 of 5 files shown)


conans/test/functional/layout/test_in_cache.py
```diff
@@ -26,7 +26,7 @@ def layout(self):
     def source(self):
         self.output.warn("Source folder: {}".format(self.source_folder))
         # The layout describes where the sources are, not force them to be there
-        tools.save("my_sources/source.h", "foo")
+        tools.save("source.h", "foo")
```
memsharded, line 29: This is one evidence IMHO that this change is necessary. Check the ``build()`` ``tools.save()`` without needing to add the subfolder ``my_build`` specified in the ``layout()``. I think it should be symmetric, if ``build()`` executes in the full ``self.build_folder``, including the ``layout`` part, the ``source()`` method should execute in the ``self.source_folder``, including the ``layout`` part.

conans/test/functional/layout/test_layout_autopackage.py
```diff
@@ -14,19 +14,19 @@ def test_auto_package_no_components():
...
-        tools.save("my_source/source_libs/slibone.a", "")
-        tools.save("my_source/source_libs/slibtwo.a", "")
-        tools.save("my_source/source_libs/bin_to_discard.exe", "")
-        tools.save("my_source/source_bins/source_bin.exe", "")
-        tools.save("my_source/source_frameworks/sframe1/include/include.h", "")
-        tools.save("my_source/source_frameworks/sframe2/include/include.h", "")
-        tools.save("my_source/source_frameworks/sframe1/lib/libframework.lib", "")
-        tools.save("my_source/source_frameworks/sframe2/lib/libframework.lib", "")
-        tools.save("my_source/source_frameworks/sframe2/foo/bar.txt", "")
+        tools.save("source_sources/source_stuff.cpp", "")
```
memsharded, line 17: Same as above, this change makes the same approach between the ``build()`` and ``source()`` methods.

lasote (approved):
I agree this is more consistent. Now that we have more clear that the user must avoid collisions of the exported files in the "source" folder declared by the layout, it should be good. Let's try it.

PR #10619, opened 2022-02-20
let conandata_update create conandata.yml if not exist

memsharded:
Changelog: Omit
Docs: Omit

The conandata_update() is not released yet, it has its own PR and docs.

The tests are just a refactor, moving the existing tests into a class, and the only new test is ``test_conandata_create_if_not_exist()``.

### DIFF
--- a/conan/tools/files/conandata.py
+++ b/conan/tools/files/conandata.py
@@ -14,8 +14,11 @@ def update_conandata(conanfile, data):
     if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
         raise ConanException("The 'update_conandata()' can only be used in the 'export()' method")
     path = os.path.join(conanfile.export_folder, "conandata.yml")
-    conandata = load(path)
-    conandata = yaml.safe_load(conandata)
+    if os.path.exists(path):
+        conandata = load(path)
+        conandata = yaml.safe_load(conandata)
+    else:  # File doesn't exist, create it
+        conandata = {}
 
     def recursive_dict_update(d, u):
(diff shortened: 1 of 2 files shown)


conan/tools/files/conandata.py
```diff
@@ -14,8 +14,11 @@ def update_conandata(conanfile, data):
     if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
         raise ConanException("The 'update_conandata()' can only be used in the 'export()' method")
     path = os.path.join(conanfile.export_folder, "conandata.yml")
-    conandata = load(path)
-    conandata = yaml.safe_load(conandata)
+    if os.path.exists(path):
+        conandata = load(path)
+        conandata = yaml.safe_load(conandata)
+    else:  # File doesn't exist, create it
+        conandata = {}
```
franramirez688, line 21: If it does not exist, do you have to create it? I'm asking because of the comment above. Indeed, the `conandata` is not used anymore, is it?

memsharded: Yes, this is exactly the feature. If you have some ``conandata_update()`` in your recipe, you want that data in your final conandata.yml, even if you don't have an empty conandata.yml besides your conanfile. I don't think we need another method, or even worse, checking the existence of the file in the recipe.

franramirez688: Oh, yeah, I'm sorry, I did not get to see that this function has more code below. My bad, it's completely fine 👍

franramirez688 (approved):

PR #10624, opened 2022-02-21
[develop2] Fix bug in ``override=True`` trait propagation

memsharded:
This is necessary for the 2.0 docs and https://github.com/conan-io/examples2/pull/2


conans/model/requires.py
```diff
@@ -212,6 +212,8 @@ def aggregate(self, other):
         self.libs |= other.libs
         self.run = self.run or other.run
         self.visible |= other.visible
+        # The force trait is also defined from an override
+        self.force |= other.force or other.override
```
memsharded, line 216: This is the core fix of this PR, the rests are tests, output and removing dead code

conans/test/integration/remote/download_retries_test.py
```diff
@@ -58,4 +58,4 @@ def get(self, *args, **kwargs):
                             requester_class=BuggyRequester)
         client.run("install --reference=pkg/0.1@lasote/stable", assert_error=True)
         self.assertEqual(str(client.out).count("Waiting 0 seconds to retry..."), 2)
-        self.assertEqual(str(client.out).count("Error 200 downloading"), 3)
+        self.assertEqual(str(client.out).count("Error 200 downloading"), 4)
```
memsharded, line 61: This increments by 1 because the detailed information about the Graph error includes the error downloading message.

czoido (approved):

PR #10625, opened 2022-02-21
exports are not sources, not copied if layout()

memsharded:
Changelog: Fix: When ``layout()`` is defined, the ``exports`` will not be considered sources anymore, but only the ``exports_sources``. The ``exports`` are used exclusively by the recipe, but not as package source.
Docs: Omit

Continuation of https://github.com/conan-io/conan/pull/10612
Preparation for https://github.com/conan-io/conan/pull/10594


lasote (approved):
Let's do it 🍔

PR #10626, opened 2022-02-21
update jinja2>=3

memsharded:
Changelog: Fix: Update ``jinja2>=3.0``, after the last Jinja2 breaking because of ```Markupsafe`` update, and maintainers saying it is no longer supported, and it can break again anytime.
Docs: Omit

TO DISCUSS: Or should we limit markupsafe<2.1, in the same way we did for the 1.43.4 patch release?


memsharded:
It seems we can't do it without fully removing Py2 support. I know it is very legacy now, but still a minority of users using it, we care for them, we will not fully retire Py2 until 2.0 if possible.

PR #10654, opened 2022-02-24
proposing base_source_folder, and folders.root

memsharded:
Changelog: Feature: Adding ``self.base_source_folder`` for ``exports_sources`` explicit layouts.
Changelog: Feature: Adding ``root`` to layout model to allow conanfile.py in subfolders.
Docs: https://github.com/conan-io/docs/pull/2418

### DIFF
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -26,9 +26,23 @@ class Folders(object):
         self.package = ""
         self.generators = ""
         self.imports = ""
+        self.root = None
 
     def __repr__(self):
         return str(self.__dict__)
+
+    def set_local(self, conanfile_folder, output_folder):
+        # This must be called only after ``layout()`` has been called
+        base_folder = conanfile_folder if self.root is None else \
+            os.path.normpath(os.path.join(conanfile_folder, self.root))
+
+        self._base_source = base_folder
+
+        self._base_install = output_folder or base_folder
+        self._base_build = output_folder or base_folder
+        self._base_package = output_folder or base_folder
+        self._base_generators = output_folder or base_folder
+        self._base_imports = output_folder or base_folder
 
     @property
     def source_folder(self):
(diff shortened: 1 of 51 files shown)


conans/model/layout.py
```diff
@@ -26,10 +26,24 @@ def __init__(self):
         self.package = ""
         self.generators = ""
         self.imports = ""
+        self.root = None
 
     def __repr__(self):
         return str(self.__dict__)
 
+    def set_local(self, conanfile_folder, output_folder):
```
lasote: maybe `set_base_folders`? I don't know why `set_local`

lasote: Also a docstring would be nice. It is not easy to understand the arguments and the sense they have.

conans/model/layout.py
```diff
@@ -26,10 +26,24 @@ def __init__(self):
         self.package = ""
         self.generators = ""
         self.imports = ""
+        self.root = None
```
lasote, line 31:
```suggestion
        # e.g: If the conanfile is in a subfolder of the project self.root = ".."
        self.root = None  
```

lasote (approved):
I would try to make it a bit more self-explanatory because it is not easy to understand at all .

conans/model/layout.py
memsharded: This is only called from local flow, never in the cache. I tried to convey that meaning. I can try to rename to something like ``set_local_base_folders()`` and I will add docstring.

lasote: I don't know if "from where I call it" should be the thing naming the function rather than the thing that the function does.

memsharded: Renamed the method, and added a bunch of docstrings and comments in https://github.com/conan-io/conan/pull/10654/commits/b6971bafccb575bef5e49de6a52ec188b11acf8f, please re-review and let me know

PR #10659, opened 2022-02-24
basic_layout and external_sources for cmake_layout, removed meson_layout

### LINKED ISSUE #10645: [feature] [2.0] predefined layouts when playing locally
- When you work, let's say with CMake, with conan local commands and/or building locally with the build system, if you have a source() method to retrieve the external sources, (e.g conan-center) you may want to adjust the layout as:

```
    def layout(self):
        cmake_layout(self)
        self.folders.source = "src"
```

The `self.folders.source = "src"` is because you don't want to end up with a bunch of cloned/unzipped files at the root directory of your package.

- It happens also with any other build system, for example, using Autotools (we don't have any preset layout for that) you might want to declare something like:

```
    def layout(self):
         self.folders.build = os.path.join("build", str(self.settings.build_type))
         self.folders.generators = os.path.join(self.folders.build, "conan")
         self.folders.source = "src"
```

Because you don't want either the cloned "src" or the build/generators littering your package folder. 

- Given that the previous definition is totally random, not attached to Autotools at all, we should consider to add a `basic_layout()` helper. 
- Given that in both cases we are interested in not littering the repo with the cloned sources, consider an argument to all the layouts `external_sources=True` to assign a "src" folder.
- Check if the Meson layout is generic and it could be renamed.

EDIT: Consider not having the "build_type" setting for a basic layout.

lasote:
Changelog: Feature: Added `basic_layout`, removed `meson_layout` and added argument `src_folder` to `cmake_layout`as a shortcut for adjusting `conanfile.folders.source`.
Docs: https://github.com/conan-io/docs/pull/2426

Close https://github.com/conan-io/conan/issues/10645

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -3,7 +3,7 @@ import os
 from conans.errors import ConanException
 
 
-def cmake_layout(conanfile, generator=None, external_sources=False):
+def cmake_layout(conanfile, generator=None, src_folder="."):
     gen = conanfile.conf.get("tools.cmake.cmaketoolchain:generator", default=generator)
     if gen:
         multi = "Visual" in gen or "Xcode" in gen or "Multi-Config" in gen
@@ -14,12 +14,7 @@ def cmake_layout(conanfile, generator=None, external_sources=False):
         else:
             multi = False
 
-    if not external_sources:
-        # The CMakeLists.txt is typically in the root
-        conanfile.folders.source = "."
-    else:
-        # Cloned/unzipped in a subdirectory
-        conanfile.folders.source = "src"
+    conanfile.folders.source = src_folder
     try:
         build_type = str(conanfile.settings.build_type)
     except ConanException:
@@ -32,10 +27,12 @@ def cmake_layout(conanfile, generator=None, external_sources=False):
         conanfile.folders.build = "cmake-build-{}".format(build_type)
         conanfile.folders.generators = os.path.join(conanfile.folders.build, "conan")
 
-    if not external_sources:
+    if src_folder == ".":
+        # The CMakeLists.txt is in the root, the includes (and sources) typically don't
         conanfile.cpp.source.includedirs = ["src"]
     else:
         conanfile.cpp.source.includedirs = ["."]
+
     if multi:
         conanfile.cpp.build.libdirs = ["{}".format(build_type)]
         conanfile.cpp.build.bindirs = ["{}".format(build_type)]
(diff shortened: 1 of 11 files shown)


memsharded (approved):
Fantastic

czoido:
Why `cmake_layout(self, external_sources=True)`  and not `cmake_layout(self, src_folder="whatever")` ? 🤔

lasote:
That's is a good question @czoido , cc @memsharded

memsharded:
> Why cmake_layout(self, external_sources=True) and not cmake_layout(self, src_folder="whatever")

Indeed a good question. I guess we wanted to emphasize the fact that this argument is intended to work with external sources, which was the problem being addressed. But yes, that would be more generic and flexible, lets change it.

conan/tools/cmake/layout.py
```diff
@@ -3,7 +3,7 @@
 from conans.errors import ConanException
 
 
-def cmake_layout(conanfile, generator=None):
+def cmake_layout(conanfile, generator=None, external_sources=False):
```
memsharded: Lets change it to ``src_folder="."``  as recommended by @czoido

memsharded (changes requested):

conan/tools/cmake/layout.py
```diff
@@ -27,7 +27,12 @@ def cmake_layout(conanfile, generator=None):
         conanfile.folders.build = "cmake-build-{}".format(build_type)
         conanfile.folders.generators = os.path.join(conanfile.folders.build, "conan")
 
-    conanfile.cpp.source.includedirs = ["src"]
+    if src_folder == ".":
+        # The CMakeLists.txt is in the root, the includes (and sources) typically don't
```
memsharded: This is a bit confusing, maybe this means ``includedirs`` is yet another argument ``cmake_layout(..., include_dirs=xxx``? Or that should be an after ``cmake_layout()`` call, directly calling ``self.cpp.source.includedirs = ["xxx"]`` to change it? Why ``src_folder`` yes and ``includedirs`` no?

memsharded (approved):
My comment can be discussed or addressed later, good for me

conan/tools/cmake/layout.py
lasote: That's why I added the comment, it is very confusing and quite random to provide "src"

czoido: I agree, that's confusing, maybe we can also add that as an argument

memsharded: Lets make:
```python
conanfile.cpp.source.includedirs = ["include"]
```
For all cases

lasote: done

memsharded (approved):

czoido (approved):

PR #10664, opened 2022-02-25
Fixed testing get password

lasote:
Some tests using client and server uploading packages, **launched from terminal** with pytest, got freeze because internally python is accessing `/dev/tty` while calling `getpass` instead of using `sys.stdin` (that we have mocked to return a list of strings).

### DIFF
--- a/conans/test/utils/tools.py
+++ b/conans/test/utils/tools.py
@@ -457,6 +457,16 @@ class TestClient(object):
         with mock.patch("conans.client.rest.conan_requester.requests", _req):
             yield
 
+    @contextmanager
+    def mocked_io(self):
+        def mock_get_pass(*args, **kwargs):
+            return self.user_inputs.readline()
+
+        with redirect_output(self.stderr, self.stdout):
+            with redirect_input(self.user_inputs):
+                with mock.patch("getpass.getpass", mock_get_pass):
+                    yield
+
     def run_cli(self, command_line, assert_error=False):
         current_dir = os.getcwd()
         os.chdir(self.current_folder)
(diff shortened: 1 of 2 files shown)


conans/test/utils/tools.py
```diff
@@ -457,6 +457,16 @@ def mocked_servers(self, requester=None):
...
             yield
 
+    @contextmanager
+    def mocked_io(self):
+        def mock_get_pass(*args, **kwargs):
+            return self.user_inputs.readline()
+
+        with redirect_output(self.stderr, self.stdout):
+            with redirect_input(self.user_inputs):
+                with mock.patch("getpass.getpass", mock_get_pass):
```
lasote, line 467: This is the key, we call the mocked stdin

conans/test/utils/tools.py
```diff
@@ -457,6 +457,16 @@ def mocked_servers(self, requester=None):
         with mock.patch("conans.client.rest.conan_requester.requests", _req):
             yield
 
+    @contextmanager
+    def mocked_io(self):
+        def mock_get_pass(*args, **kwargs):
+            return self.user_inputs.readline()
```
lasote, line 463: This `self.user_inputs` is the mock of `sys.stdin`

conans/test/utils/tools.py
```diff
@@ -494,30 +504,29 @@ def run(self, command_line, assert_error=False, redirect_stdout=None, redirect_s
         with environment_update({"NO_COLOR": "1"}):  # Not initialize colorama in testing
             self.stdout = RedirectedTestOutput()  # Initialize each command
             self.stderr = RedirectedTestOutput()
-            with redirect_output(self.stderr, self.stdout):
```
lasote, line 497: Just extracted the 3 contexts to a single one. No code changes.

franramirez688 (approved):

PR #10666, opened 2022-02-25
Merge/develop -> develop2

franramirez688:
Update from latest develop

### DIFF
--- a/conans/test/integration/command/install/install_test.py
+++ b/conans/test/integration/command/install/install_test.py
@@ -374,3 +374,51 @@ class TestCliOverride:
                      "test_package/conanfile.py": GenConanfile().with_test("pass")})
         client.run("create . --name=pkg --version=0.1 --require-override=zlib/2.0")
         assert "zlib/2.0: Already installed" in client.out
+
+
+def test_install_bintray_warning():
+    server = TestServer(complete_urls=True)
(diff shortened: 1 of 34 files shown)


conans/test/integration/command/install/install_test.py
```diff
@@ -374,3 +374,51 @@ def test_create_cli_override(self, client):
                      "test_package/conanfile.py": GenConanfile().with_test("pass")})
         client.run("create . --name=pkg --version=0.1 --require-override=zlib/2.0")
         assert "zlib/2.0: Already installed" in client.out
+
+
+def test_install_bintray_warning():
```
memsharded, line 379: We don't want bintray warnings in 2.0, this can be completely removed. Make sure the potential warning in codebase is also removed.

memsharded (changes requested):
Just an important operational thing: pull-requests over develop2 to solve conflicts don't seem to work well. Future merge will keep raising conflicts. The best way I have found so far to minimize future conflicts is doing the merge directly on the develop2, fixing, commiting when test passes and pushing. Probably there is a better way, or making sure that PRs are fast-forward only, but merge commits are problematic.

lasote:
This has been merged already, fixing the test failing and verifying there is no bintray stuff there.

Issue #10650, opened 2022-02-24
[feature] AutotoolsToolchain : reconsider default_configure_install_args default to True?


SSE4:
from https://github.com/conan-io/conan-center-index/pull/9488

without `default_configure_install_args = True`, Autotools will try to install into `/usr`, and will eventually fails in most of the cases (as it requires root access). or if it doesn't fail, it will pollute the system (potentially, overriding system packages?).
I think, practically, every actual recipe will have to add `default_configure_install_args = True` in order to work correctly (install filess to the package folder).

- [ ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

lasote:
Related https://github.com/conan-io/conan/issues/10430

I think at least it should adjust the ``--prefix``, about the rest of variables, probably too: https://github.com/conan-io/conan/issues/10430 and https://github.com/conan-io/conan/issues/10393

memsharded:
Changed to True in https://github.com/conan-io/conan/pull/10672, will be released in 1.46

Issue #10430, opened 2022-01-26
AutotoolsToolchain could use "res" as datadir instead of "share", to align with ConanCenter


memsharded:
Following conan-io/conan-center-index#9091 (comment), it could be that changing the default toolchain ``datadir`` to "res" instead of "share" would be more aligned with ConanCenter requested layouts, lets check it.

_Originally posted by @memsharded in https://github.com/conan-io/conan/issues/8476#issuecomment-1022130033_

ericLemanissier:
My personal take on this is that using `share` folder would allow to simplify a lot of recipes which have to manually move some resources to the `res` folder, which is non-standard.

lasote:
Looks necessary to adjust in AutotoolsToolchain always the "prefix" and probably also the "lib" to avoid "lib64" and "bin" to avoid "sbin", and of course the "datarootdir" to "res" or "share". 
In the cppinfo model, only one of "res"/"share" should be kept.

Related https://github.com/conan-io/conan/issues/10393

SpaceIm:
I agree with @ericLemanissier, I don't understand why in the first place `res` has been chosen instead of `share`.

memsharded:
Because it was chosen like 6 years ago, and my bias as Windows developer put there ``res``. The ``cpp_info.resdirs`` is called ``resdirs`` no ``sharedirs`` for same reason. There is simply no good default for everyone. Windows devs will probably not love if we change the ``resdirs = ["share"]`` default.

memsharded:
New Autotools now use the defined ``cpp_info.resdirs`` as the default for autotools based install, implemented in https://github.com/conan-io/conan/pull/10672, will be in next 1.46

Issue #10393, opened 2022-01-21
[bug] Libraries in lib64 are not exported to libdir


ashley-b:
I have a library using modern CMake install syntax with no destination set so it uses the system lib path which on my system is lib64.
When I install/consume this recipe I can see the lib is set correctly but the libdir is empty.
I think this is related to the following issue #3369, #3488.

I have found two workarounds, so I keep the CMake script portable.

* Force CMake to install to lib for conan
```
def generate(self):
  tc = CMakeToolchain(self)
  tc.variables["CMAKE_INSTALL_LIBDIR"] = "lib"
  tc.generate()
```
*OR*
* Tell conan were to find it.
```
def package_info(self):
  self.cpp_info.libdirs = ['lib','lib64']
```
I feel this is some thing conan should be able to take care of

### Environment Details
  * Fedora release 35
  * gcc (GCC) 11.2.1 20211203 (Red Hat 11.2.1-7)
  * Conan version:  1.44.1
  * Python 3.10.1

memsharded:
Conan tries to be agnostic about all these platform specific issues, and it should be kept agnostic, and instead provide tools to handle things. Take into account for example that this might happen in Fedora but not Ubuntu, but Conan has no representation or model that you are running in Fedora. Conan wraps the build systems, if the build system is doing a different thing in a different platform, Conan should be informed about it, we have tried Conan "guessing" or trying to be smart about those things in the past, and it always leads to fragile and confusing behavior.

If some system is packaging to a different folder, indeed there are mechanisms to handle that, depends on what you prefer:

- You can tell your build system, CMake in this case to output to a different folder
- You can tell your ``package()`` method to pick from a different folder (if using ``self.copy()``) or doing a rename (if using ``cmake.install()``)
- You can tell your ``package_info()`` the name of ``libdirs``. In this case it could be more conceptually correct to do something like ``if x64-arch: self.cpp_info.libdirs = ["lib64"] else: self.cpp_info.libdirs = ["lib"]``

The latest seems the less clean, and it should probably be addressed in the build and package steps.

ashley-b:
Thanks for the reply.
To give some more back story, This recipe is using the "cmake" generator and it works. It seams to set CMAKE_INSTALL_LIBDIR=lib regardless of arch setting.

Its only since I try migrating my recipe to use the new CMakeToolchain and CMakeDeps generators it stopped setting CMAKE_INSTALL_LIBDIR and as a result is not setting the LIBDIR

[...]

PR #10672, opened 2022-02-28
Output install dirs in new toolchains

lasote:
Changelog: Feature: The `CMakeToolchain` and `AutotoolsToolchain` take into account the `cpp.package` info to set the output directories for libraries, executables, and so on when running `cmake.install` or `make install`.
Docs: https://github.com/conan-io/docs/pull/2425

#10430 
https://github.com/conan-io/conan/issues/10650
https://github.com/conan-io/conan/issues/10393


memsharded (commented):
I would say this is fine, and we should go this way, do we have any other alternative? Up to my understanding this is necessary

czoido (approved):

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -730,3 +730,37 @@ def context(self):
...
+           set("CMAKE_INSTALL_INCLUDEDIR" "{{default_include}}")
+           set("CMAKE_INSTALL_OLDINCLUDEDIR" "{{default_include}}")
+           set("CMAKE_INSTALL_DATAROOTDIR"] "{{default_res}}")
+        """)
+
+    def _get_cpp_info_value(self, name):
+        # Why not taking cpp.build? because this variables are used by the "cmake install"
+        # that correspond to the package folder (even if the root is the build directory)
+        elements = getattr(self._conanfile.cpp.package, name)
+        return elements[0] if elements else None
```
franramirez688, line 766: Would it make sense to have a more specific function on ConanFile class side? Something like:

```python
class ConanFile(object):
    # . . .
    @property
    def default_configure_install_dirs(self):
        # Why not taking cpp.build? because this variables are used by the "cmake install"
        # that correspond to the package folder (even if the root is the build directory)
        p = self.cpp.package
        return {
            "includedirs": p.includedirs[0] if len(p.includedirs) else None,
            "libdirs": p.libdirs[0] if len(p.libdirs) else None,
            "bindirs": p.bindirs[0] if len(p.bindirs) else None,
            "resdirs": p.resdirs[0] if len(p.resdirs) else None,
            "builddirs": p.builddirs[0] if len(p.builddirs) else None,
            "frameworkdirs": p.frameworkdirs[0] if len(p.frameworkdirs) else None
        }
```

franramirez688 (commented):
I think we need it and it's good enough for now 👍

conan/tools/cmake/toolchain/blocks.py
memsharded: Not in ``ConanFile`` public interface, that is for sure, maybe in ``NewCppInfo`` (the indicator that it could belong there is that almost everything in that logic is ``p.xxxx`` with ``p`` being an object of type ``NewCppInfo``.

Probably not worth extracting yet until this is clear and more stable.

memsharded (approved):

SSE4 (approved):
that's nice. probably, meson will need similar changes as well.

conans/test/integration/toolchains/gnu/test_autotools_toolchain.py
```diff
@@ -0,0 +1,56 @@
...
+import pytest
+
+from conans.model.ref import ConanFileReference
+from conans.test.assets.autotools import gen_makefile_am, gen_configure_ac
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.utils.tools import TurboTestClient
+
+
+@pytest.mark.tool_autotools()
+def test_install_output_directories():
```
memsharded: This test should probably go to ``functional`` as it is a full compilation test.

conans/test/integration/toolchains/gnu/test_autotools_toolchain.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    If we change the libdirs of the cpp.package, as we are doing cmake.install, the output directory
+    for the libraries is changed
+    """
+    client = TurboTestClient(path_with_spaces=False)
+    client.run("new hello/1.0 --template cmake_lib")
+    client.run("create .")
+    consumer_conanfile = textwrap.dedent("""
+        from conans import ConanFile
```
memsharded:
```suggestion
        from conan import ConanFile
```

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
```diff
@@ -231,3 +231,33 @@ def generate(self):
     with open(os.path.join(client.current_folder, "conan_toolchain.cmake")) as f:
         contents = f.read()
         assert "/path/to/builddir" in contents
+
+
+def test_install_output_directories():
```
memsharded: functional test, not integration

memsharded (approved):

franramirez688 (approved):

PR #10673, opened 2022-02-28
Improve markdown generator

czoido:
Changelog: Feature: Improve markdown generator instructions.
Docs: https://github.com/conan-io/docs/pull/2423

This started as just adding the language to the code blocks but this generator clearly needed some extra work, there were links to old integrations, it did not render well and there were some mistakes.

image

image

This does not include the XcodeDeps/XcodeToolchain as there's still a pending PR for those.


czoido:
This is what we previously had:

image

memsharded (approved):
Great job, thanks very much!

danimtb (approved):

SpaceIm:
Looks nice.

Maybe not the good place, but I have one comment: I'm a little bit concerned that conan suggests in this file to link to `Poco::Poco`. It's a global target we can't avoid in poco recipe of conan-center recipe, even with `set_property()` , but this target doesn't exist upstream.
I know it could make generators more complex but it would be nice to be able to disable creation of the global target in `package_info()` when there are components (and fix the consequences in generators, because I understand that global target is used under the hood instead of using union of components targets).

SSE4 (approved):
quite long and verbose, but I think it's fine, and helpful for newcomers

jgsogo:
> Looks nice.
> 
> Maybe not the good place, but I have one comment: I'm a little bit concerned that conan suggests in this file to link to `Poco::Poco`. It's a global target we can't avoid in poco recipe of conan-center recipe, even with `set_property()` , but this target doesn't exist upstream. I know it could make generators more complex but it would be nice to be able to disable creation of the global target in `package_info()` when there are components (and fix the consequences in generators, because I understand that global target is used under the hood instead of using union of components targets).

IIRC, generators (CMake ones) use the components of their requirements, that's why you need to list `...requires = "req1::cmp1", "req2::cmp1",...`, the recipe writer chooses which components to consume (targets to link) from the requirements. Of course, it can also be the global one with `req1::req1`. @danimtb did some drawings taken from CMake outputs while developing the feature, maybe they can be found in the original PR.

PR #10677, opened 2022-02-28
Feature/profile include priority

### LINKED ISSUE #10499: [bug] Profile includes prefer $HOME/.conan/profiles folder over the folder of the importing profile
  * Operating System+version: Linux/Ubuntu20
  * Conan version: 1.42.2
  * Python version: 3.9.5

In my env I have had old deprecated profile hanging here
`~/.conan/profiles/linux-base-profile`

```
~/temp/linux-base-profile  # notice the filename matches with the one under conan home
~/temp/linux-profile
    include(linux-base-profile)
```
    
```
$ mkdir ~/temp/build && cd ~/temp/build
$ conan install <foo> --profile ../linux-profile
```

From the output I can see that include() actually loads the settings & options from ~/.conan/profiles/linux-base-profile, not from
~/temp/linux-base-profile?
When I delete the (deprecated) profile from ~/.conan/profiles/ it works as expected.

I would still expect that the search order should have the folder of the importing profile prioritized?

memsharded:
Fix https://github.com/conan-io/conan/issues/10499


memsharded:
After some discussion, we have finally decided:

- We are not changing the behavior. Maybe it is not perfectly intuitive in all cases, but it is more explicit and systematic, and allows to do whatever the user wants.
- Profile paths starting with ``./`` will be considered relative to the current profile (if profile include()) or current working directory (if cli arg)
- Profiles without ``./`` will be searched in the cache/profiles folder.

PR #10687, opened 2022-03-01
adding jinja rendering of global.conf config file

### LINKED ISSUE #10191: [feature] Have a convention in global.conf to specify max number of cores
Recently ``tools.build:jobs`` was changed to default to the maximum number of cores
Maybe it would be better to explicitly allow it, like ``tools.build:jobs=MAX_CORES``. 

This would also apply to ``core.download:parallel`` and ``core.upload:parallel`` (when implemented, not yet available)

memsharded:
Changelog: Feature: Adding jinja rendering of global.conf config file.

Close https://github.com/conan-io/conan/issues/10191

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -1,7 +1,10 @@
 import fnmatch
+import os
+import platform
 from collections import OrderedDict
 
 import six
+from jinja2 import Template
 
 from conans.errors import ConanException
 
@@ -470,6 +473,7 @@ class ConfDefinition:
         return parsed_value
 
     def loads(self, text, profile=False):
+        text = Template(text).render({"platform": platform, "os": os})
         self._pattern_confs = {}
 
         for line in text.splitlines():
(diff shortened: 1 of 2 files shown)


conans/model/conf.py
```diff
@@ -470,6 +473,7 @@ def _get_evaluated_value(__v):
         return parsed_value
 
     def loads(self, text, profile=False):
+        text = Template(text).render({"platform": platform, "os": os})
```
franramirez688, line 476: Should we provide `multiprocessing` for Python 2.x users too? To be able to run a `multiprocessing.cpu_count()`.

franramirez688 (approved):
I think it's fine for now. Anyway, I still have doubts if this solution is overkill to solve the request about being able to pass `MAX_CORES` in an easy way...

conans/model/conf.py
memsharded: No need to support Py2 for these very new (experimental) things focused completely on 2.0

PR #10689, opened 2022-03-01
Revert: package folder for local install

lasote:
Reverted #10655

### DIFF
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -601,8 +601,6 @@ class ConanAPIV1(object):
                                         lockfile=lockfile)
 
             install_folder = _make_abs_path(install_folder, cwd)
-            if output_folder:
-                output_folder = _make_abs_path(output_folder, cwd)
             conanfile_path = _get_conanfile_path(path, cwd, py=None)
 
             remotes = self.app.load_remotes(remote_name=remote_name, update=update)
--- a/conans/client/conan_api.py
+++ b/conans/client/conan_api.py
@@ -601,6 +601,8 @@ class ConanAPIV1(object):
                                         lockfile=lockfile)
 
             install_folder = _make_abs_path(install_folder, cwd)
+            if output_folder:
+                output_folder = _make_abs_path(output_folder, cwd)
             conanfile_path = _get_conanfile_path(path, cwd, py=None)
 
             remotes = self.app.load_remotes(remote_name=remote_name, update=update)
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -467,7 +467,7 @@ class BinaryInstaller(object):
         base_path = package_layout.base_folder()
 
         if hasattr(conanfile, "layout"):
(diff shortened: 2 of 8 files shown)


conans/client/conan_api.py
```diff
@@ -601,8 +601,6 @@ def install(self, path="", name=None, version=None, user=None, channel=None,
                                         lockfile=lockfile)
 
             install_folder = _make_abs_path(install_folder, cwd)
-            if output_folder:
```
memsharded: Maybe this ``make_abs`` was still correct, when it is defined?

memsharded (commented):
There were some changes in between. Like ``set_base_folders()`` now defines ``_base_package`` too for the local flow, which will be used by other commands like ``conan build``. Maybe that should be de-activated and tested too?

SSE4:
what is the motivation to revert this fix? what is the alternative fix?

conans/client/conan_api.py
lasote: Yes. I think this was probably a bug.

lasote:
@SSE4 we consider that a local "conan install" should be done when you install your dependencies and you are developing locally your package (or project) so you should call the build system setting the prefix manually. What if I really want to build a project and install it in another folder or even the system? it is up to the user to call "cmake" or "autotools" setting the prefix. Also, it makes not much sense to set the base package folder to the output folder, where the build files will be, because even an editable package will take the "build" layout to locate the libraries. The 'cmake.install' and `make install` should be done in the `package()` method.

SSE4:
@lasote thank you for the explanation, it makes sense.
however, can we improve an error message at least?
right now (1.45) I am getting:
```
  File "/opt/pyenv/versions/3.7.5/lib/python3.7/site-packages/conans/errors.py", line 34, in conanfile_exception_formatter
    yield
  File "/opt/pyenv/versions/3.7.5/lib/python3.7/site-packages/conans/client/generators/__init__.py", line 245, in write_toolchain
    conanfile.generate()
  File "/host/bincrafters/conan-center-index/recipes/kmod/all/conanfile.py", line 62, in generate
    tc.generate()
  File "/opt/pyenv/versions/3.7.5/lib/python3.7/site-packages/conan/tools/gnu/autotoolstoolchain.py", line 152, in generate
    self.generate_args()
  File "/opt/pyenv/versions/3.7.5/lib/python3.7/site-packages/conan/tools/gnu/autotoolstoolchain.py", line 161, in generate_args
    ["--prefix=%s" % self._conanfile.package_folder.replace("\\", "/"),
AttributeError: 'NoneType' object has no attribute 'replace'
```
(it's with verbose traceback, otherwise it's just AttributeError)
which is not very helpful for the user

lasote:
Sure, but that is more related to: https://github.com/conan-io/conan/pull/10672

franramirez688 (approved):

memsharded:
Superseded by https://github.com/conan-io/conan/pull/10695

PR #10693, opened 2022-03-02
Merge develop->develop2 new Xcode tools

czoido:


conan/tools/apple/apple.py
```diff
@@ -18,19 +18,21 @@ def to_apple_arch(arch, default=None):
             'armv7k': 'armv7k'}.get(arch, default)
 
 
-def get_apple_sdk_name(conanfile):
+def get_apple_sdk_fullname(conanfile):
```
czoido, line 21: changed the name of this function to remark that it returns sdk_name with version if exists

conan/tools/apple/apple.py
```diff
@@ -41,16 +43,24 @@ def apple_min_version_flag(os_version, os_sdk, subsystem):
...
     # FIXME: This guess seems wrong, nothing has to be guessed, but explicit
-    flag = {'macosx': '-mmacosx-version-min',
-            'iphoneos': '-mios-version-min',
-            'iphonesimulator': '-mios-simulator-version-min',
-            'watchos': '-mwatchos-version-min',
-            'watchsimulator': '-mwatchos-simulator-version-min',
-            'appletvos': '-mtvos-version-min',
-            'appletvsimulator': '-mtvos-simulator-version-min'}.get(str(os_sdk))
+    flag = ''
+    if 'macosx' in os_sdk:
```
czoido, line 47: now os_sdk can be macosx11.3 for example

conan/tools/apple/xcodedeps.py
```diff
@@ -86,38 +129,27 @@ def generate(self):
...
         fields = {
             'name': name,
-            'configuration': self.configuration,
-            'architecture': self.architecture,
-            'root_folder': dep.package_folder,
-            'bin_dirs': " ".join('"{}"'.format(os.path.join(dep.package_folder, p)) for p in cpp_info.bindirs),
-            'res_dirs': " ".join('"{}"'.format(os.path.join(dep.package_folder, p)) for p in cpp_info.resdirs),
-            'include_dirs': " ".join('"{}"'.format(os.path.join(dep.package_folder, p)) for p in cpp_info.includedirs),
-            'lib_dirs': " ".join('"{}"'.format(os.path.join(dep.package_folder, p)) for p in cpp_info.libdirs),
+            'bin_dirs': " ".join('"{}"'.format(os.path.join(p)) for p in cpp_info.bindirs),
```
czoido: removing dep.package_folder, now folders are absolute

memsharded: This would be great to do in 1.X if possible, not just in 2.0

memsharded (approved):

conan/tools/apple/xcodedeps.py
franramirez688: Then, we can remove all the `os.path.join`, right?

```suggestion
            'bin_dirs': " ".join('"{}"'.format(p) for p in cpp_info.bindirs),
```

memsharded:
Nop :(

It still gives conflicts. You need to push to develop2 directly

PR #10695, opened 2022-03-02
Fixing xfails of https://github.com/conan-io/conan/pull/10689

### LINKED ISSUE #10691: [bug] editables + layout broken
Review the 4 tests skipped: `@pytest.mark.xfail(reason="Editable + layout are broken,`
The cause: The generators are accessing `dep.package_folder` for several reasons but they shouldn't because the paths in the cpp_info object are already absolute. At most `dep.package_folder` could be used to "extract" the common path to a variable but the generator has to check if `dep.package_folder` is None before.

From https://github.com/conan-io/conan/pull/10689

memsharded:
Changelog: omit
Docs: omit

Close https://github.com/conan-io/conan/issues/10691

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -486,10 +486,11 @@ class FindFiles(Block):
         for req in host_req:
             cppinfo = req.cpp_info.aggregated_components()
             # If the builddir is the package_folder, then it is the default "root" one
-            nf = os.path.normpath(req.package_folder)
-            host_build_paths_root.extend(p for p in cppinfo.builddirs if os.path.normpath(p) == nf)
-            host_build_paths_noroot.extend(p for p in cppinfo.builddirs if os.path.normpath(p) != nf)
-            host_lib_paths.extend(cppinfo.libdirs)
+            if req.package_folder is not None:
+                nf = os.path.normpath(req.package_folder)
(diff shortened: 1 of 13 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -486,7 +486,9 @@ def context(self):
         for req in host_req:
             cppinfo = req.cpp_info.aggregated_components()
             # If the builddir is the package_folder, then it is the default "root" one
-            nf = os.path.normpath(req.package_folder)
+            # The package_folder can be None if editable and layout(), in that case only the
+            # host_build_paths_noroot will be populated
+            nf = os.path.normpath(req.package_folder or "")
             host_build_paths_root.extend(p for p in cppinfo.builddirs if os.path.normpath(p) == nf)
             host_build_paths_noroot.extend(p for p in cppinfo.builddirs if os.path.normpath(p) != nf)
```
lasote: I would see this more clear:

```suggestion
            if req.package_folder:
                nf = os.path.normpath(req.package_folder)
                host_build_paths_root.extend(p for p in cppinfo.builddirs if os.path.normpath(p) == nf)
                host_build_paths_noroot.extend(p for p in cppinfo.builddirs if os.path.normpath(p) != nf)
            else:
                host_build_paths_root = []
                host_build_paths_noroot.extend(p for p in cppinfo.builddirs)
```

lasote (approved):

PR #10716, opened 2022-03-04
[develop2] Simplify source management

memsharded:
- Do not remove files in the process, not necessary (remove ``_clean_source_folder()``
- Do not do a local copy of exports files in local flow.


lasote (approved):
🥇

Issue #10632, opened 2022-02-22
[feature] Rename ``conan install --reference`` to ``conan install --requires=xxx`` and allow multiple


memsharded:
This might align better with the conanfile ``[requires]`` definition, and understand that it allows version ranges for example.

It might be worth considering too the ``--reference=xxxx --build-require`` and move it to ``--build-require=xxxx`` argument (even in combination with the above)

The mapping between the conanfile and the command line will be direct and more understandable?

memsharded:
Implemented in https://github.com/conan-io/conan/pull/10721

PR #10721, opened 2022-03-05
rename reference=>requires

memsharded:
The goal is to have the equivalent to conanfile.txt ``[requires]``, ``[tool_requires]``, etc in cli:
- Renamed ``--reference`` => ``--requires``
- Added ``--tool-requires=xxx``, removed ``install --requires=xxx --build-require``
- Made both multiple
- Both can be used simultaneously 
- ``conan create`` maintains ``--build-require`` as argument (in case there is no ``test_package``), to indicate it is a build-require, still seems necessary (maybe rename it to ``--tool-require`` or ``--is-tool-require``?

Close https://github.com/conan-io/conan/issues/10632


conans/cli/api/subapi/graph.py
```diff
@@ -83,11 +84,16 @@ def load_root_test_conanfile(self, path, tested_reference, profile_host, profile
...
                                     require_overrides=None):
+        if not requires and not tool_requires:
+            raise ConanException("Provide requires or tool_requires")
         app = ConanApp(self.conan_api.cache_folder)
-        profile_host.options.scope(ref.name)
-        conanfile = app.loader.load_virtual([ref],  is_build_require=is_build_require,
+        if requires and len(requires) == 1 and not tool_requires:
+            profile_host.options.scope(requires[0].name)
+        if tool_requires and len(tool_requires) == 1 and not requires:
+            profile_host.options.scope(tool_requires[0].name)
```
lasote: This looks tricky in the API. I get why it is like this, but it makes no sense for the API. Should the profile options be scoped before this call?

conans/cli/api/subapi/graph.py
```diff
@@ -83,11 +84,16 @@ def load_root_test_conanfile(self, path, tested_reference, profile_host, profile
         return root_node
 
     @api_method
-    def load_root_virtual_conanfile(self, ref, profile_host, is_build_require=False,
+    def load_root_virtual_conanfile(self, profile_host, requires=None, tool_requires=None,
```
lasote, line 87: Would make sense a generic build_require instead the tool_requires?

memsharded: I am not sure, it tries to mimic the ``[tool_requires]`` we have in the ``conanfile.txt`` (and ``conanfile.py``). If we are adding a ``--build-requires``, then a explicit equivalent in conanfile should be created?
Actually it would make more sense to have a ``--test-requires``, the only thing is that for consumer ``conan install`` that is equivalent in practice to ``--requires``, so it is not strictly required functionality, just syntactic sugar.

conans/cli/api/subapi/graph.py
memsharded: Yes, I don't love it, I was adding this to not break some existing tests, but could be moved outside the API, yes.

czoido (approved):

conans/cli/api/subapi/graph.py
lasote: Agree

conans/cli/api/subapi/graph.py
memsharded: Moved to ``scope_options()`` helper.

lasote (approved):

PR #10729, opened 2022-03-07
[conf] Defining user_toolchain as list

franramirez688:
Changelog: Feature: Configuration field `tools.cmake.cmaketoolchain:user_toolchain` defined as list-like object

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -16,7 +16,7 @@ BUILT_IN_CONFS = {
     "tools.cmake.cmaketoolchain:generator": "User defined CMake generator to use instead of default",
     "tools.cmake.cmaketoolchain:find_package_prefer_config": "Argument for the CMAKE_FIND_PACKAGE_PREFER_CONFIG",
     "tools.cmake.cmaketoolchain:toolchain_file": "Use other existing file rather than conan_toolchain.cmake one",
-    "tools.cmake.cmaketoolchain:user_toolchain": "Inject existing user toolchain at the beginning of conan_toolchain.cmake",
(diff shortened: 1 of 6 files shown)


conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -157,12 +157,8 @@ class Pkg(ConanFile):
             def generate(self):
                 # Get the toolchains from "tools.cmake.cmaketoolchain:user_toolchain" conf at the
                 # tool_requires
-                user_toolchains = []
-                for dep in self.dependencies.direct_build.values():
-                    ut = dep.conf_info["tools.cmake.cmaketoolchain:user_toolchain"]
-                    if ut:
-                        user_toolchains.append(ut.replace('\\\\', '/'))
-
+                user_toolchains = self.conf.get("tools.cmake.cmaketoolchain:user_toolchain", default=[], check_type=list)
```
memsharded: Nop. Just let the CMakeToolchain internals do the right thing, no longer necessary. You could even remove the ``generate()`` method, leave ``generators="CMakeToolchain"`` and it should work as expected.

conans/model/conf.py
```diff
@@ -67,6 +67,10 @@ def __repr__(self):
 
     @property
     def value(self):
+        if self._value_type is list and _ConfVarPlaceHolder in self._value:
+            v = self._value[:]
+            v.remove(_ConfVarPlaceHolder)
```
memsharded, line 72: This was an uncaught bug with lists, because lists were not enough tested, right? 🍔 😄

franramirez688: Yep 😢

memsharded (approved):

czoido (approved):

Issue #10647, opened 2022-02-23
[bug] Incorrect version in lockfile when deduced from other lockfile on package with non Sermver Versions


KerstinKeller:
I am having two conanfiles, and i want to derive a lockfile for `OtherApp` from the lockfile of `App`.

```
conanfile.py

class App(ConanFile):
    name = "app"
    version = "1.0.0"

    def requirements(self):
        self.requires("openssl/1.1.1i@")
```

```
conanfile.py

class OtherApp(ConanFile):
   name = "otherapp"

    def requirements(self):
        self.requires("openssl/[*]@")
```

`App/1.0.0@` is in the cache, the `conanfile.py` for `OtherApp` is in my local folder.

```
conan lock create --reference=App/1.0.0@ --base --lockfile-out=app.lock
conan lock create conanfile.py --lockfile=app.lock --lockfile-out=otherapp.lock
```

## The bug:
`app.lock` contains `openssl/1.1.1i@` (correct)
`otherapp.lock` contains `openssl/1.1.1@` (incorrect, the `i`postfix is missing)

## What i expect
Both lockfiles contain `openssl/1.1.1i@`.

## Screenshot
comparison of lockfiles
![image](https://user-images.githubusercontent.com/13848742/155364408-dfe81963-c944-41ad-9c51-ed8a480d390c.png)

and on the console:
creation of first lock
![image](https://user-images.githubusercontent.com/13848742/155364731-3ea80239-e2ee-4687-a80b-95eddfdb020e.png)

creation of second lock
![image](https://user-images.githubusercontent.com/13848742/155365153-3d72f434-29f3-4904-ac48-9847eae43c5e.png)

### Environment Details (include every applicable attribute)
  * Conan version: 1.45.0

memsharded:
Hi @KerstinKeller 

I think this is intrinsic to the way lockfiles are extended in Conan 1.X. When you are creating one lockfile passing an initial lockfile as argument, such lockfile should belong to the dependency graph.
You cannot apply the ``App`` lockfile to something that doesn't depend on ``App``. It will be simply disregarded. The locked ``openssl/1.1.1i`` will only be used when ``App`` is in the dependency graph. If you inspect the actual generated ``app.lock`` you will see this relation is explicit.

So when the lockfile is applied to ``otherapp``, it nevers reach the locked openssl, because that only happens if ``App`` is a dependency of ``OtherApp``.

This is one of the things that have greatly improved in 2.0, and it will work as you are expecting.

[...]

PR #10732, opened 2022-03-07
Add testcase showing deriving the locks for unrelated apps works fine…

KerstinKeller:
… with regular semver versions, but not on the non-semver one.

Changelog: (Test): Test that demonstrates behavior described in #10647

### DIFF
--- a/conans/test/integration/graph_lock/test_lock_derive.py
+++ b/conans/test/integration/graph_lock/test_lock_derive.py
@@ -25,13 +25,13 @@ def test_regular_package():
     client.run("lock create --reference=defaults/1.0@ --base --lockfile-out=defaults.lock")
     # derive app.lock from defaults.lock
     client.run("lock create app/conanfile.py --lockfile=defaults.lock --lockfile-out=app.lock")
-    client.run("install app/conanfile.py --lockfile=app.lock", assert_error=True)
+    client.run("install app/conanfile.py --lockfile=app.lock --build=missing")
 
     # check that specified versions from default package are to be installed for app.
     assert "pkga/1.0.0" in client.out
     assert "pkgb/1.1.0" in client.out
 
-         
+
 def test_non_semver_package():
     client = TestClient()
     # TODO: This is hardcoded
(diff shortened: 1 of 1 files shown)


memsharded:
Good news is that porting these tests to the 2.0 syntax, they both work as expected:
```python
def test_regular_package():
    client = TestClient()
    client.save({"pkga/conanfile.py":     GenConanfile("pkga"),
                 "pkgb/conanfile.py":     GenConanfile("pkgb"),
                 # specify versions in default packages
                 "defaults/conanfile.py": GenConanfile("defaults").with_requires("pkga/1.0.0",
                                                                                 "pkgb/1.1.0"),
                 # version range in app
                 "app/conanfile.py":      GenConanfile("app").with_requires("pkga/[*]","pkgb/[*]")})

    # export different version from the libs to the cache
    client.run("export pkga --version=1.0.0")
    client.run("export pkga --version=1.1.0")
    client.run("export pkga --version=1.1.1")
    client.run("export pkgb --version=1.0.0")
    client.run("export pkgb --version=1.1.0")
    client.run("export pkgb --version=1.1.1")
    client.run("export defaults --version=1.0")
    client.run("lock create --reference=defaults/1.0 --lockfile-out=defaults.lock")
    # derive app.lock from defaults.lock
    client.run("lock create app/conanfile.py --lockfile=defaults.lock --lockfile-out=app.lock")
    client.run("install app/conanfile.py --lockfile=app.lock --build=missing")

    # check that specified versions from default package are to be installed for app.
    assert "pkga/1.0.0" in client.out
    assert "pkgb/1.1.0" in client.out


def test_non_semver_package():
    client = TestClient()
    client.save({"pkga/conanfile.py":     GenConanfile("pkga"),
                 # require a non semver version
                 "defaults/conanfile.py": GenConanfile("defaults").with_requires("pkga/1.0.0i"),
                 "app/conanfile.py":      GenConanfile("app").with_requires("pkga/[*]")})

    client.run("export pkga --version=1.0.0")
    client.run("export pkga --version=1.0.0i")
    client.run("export pkga --version=1.1.0")
    client.run("export pkga --version=1.1.1")
    client.run("export defaults --version=1.0")
    client.run("lock create --reference=defaults/1.0 --lockfile-out=defaults.lock")
    client.run("lock create app/conanfile.py --lockfile=defaults.lock --lockfile-out=app.lock")
    client.run("install app/conanfile.py --lockfile=app.lock --build=missing")

    assert "pkga/1.0.0i" in client.out
```
I will check now in 1.X

memsharded:
Bad news: I have tracked the issue down to the third party ``node-semver`` dependency, that basically rejects ``1.0.0i`` as a valid match for the ``[*]`` range.

KerstinKeller:
@memsharded i have added two more testcases to demonstrate our usecase scenario. the second one with transitive dependencies currently fails 
```
E           Exception:
E           lock create app/conanfile.py --lockfile=defaults.lock --lockfile-out=app.lock
E           ERROR: Conflict in pkgb/1.0.0:
E               'pkgb/1.0.0' requires 'pkga/1.0.0' while 'conanfile.py' requires 'pkga/1.1.0'.
E               To fix this conflict you need to override the package 'pkga' in your root package.
E
E           --------------------------------------------------------------------------------
```

Will this behave differently in 2.0? Can I relax the lockfile with regards to `pkgb` dependency to `pkga`? E.g. to reevaluate the transitive dependency?
I tried to pass the `--build` option to the original (default version) lockfile creation, but it doesn't help (the documentation mentions that i can be used to relax lockfiles).
I know I can use `override` in the `requires` of `app`, but I guess I don't want that since it will be a hard overwrite. I still want version checks regarding the original recipes to be in place.

memsharded:
> Will this behave differently in 2.0? Can I relax the lockfile with regards to pkgb dependency to pkga? E.g. to reevaluate the transitive dependency?

Yes, absolutely. Lockfiles in 2.0 will be by default "relaxed", and you can reevaluate transitive dependencies. Only using ``--lockfile-strict`` will enforce it

> I tried to pass the --build option to the original (default version) lockfile creation, but it doesn't help (the documentation mentions that i can be used to relax lockfiles).

The ``--build`` is mostly necessary to include the ``build-requires`` in the lockfile, so they aren't missing later when they are needed. But  I think the docs might not be accurate, I don't think the ``--build`` actually relax the graph and allow changing dependencies (in 1.X)

KerstinKeller:
Thanks for the answer. We will have to see if we can start introducing this workflow at the moment (with explicit overrides), or if we will wait till 2.0.

PR #10739, opened 2022-03-09
custom commands layers

memsharded:
Closes https://github.com/conan-io/conan/issues/10741


czoido:
I was thinking about the syntax and that maybe we could change a couple of things about the custom commands (not all directly related with this)
- Use the syntax `conan frog.command ...` `conan frog.ci.command ...` and we can organize commands inside layers in folders
- Maybe remove the `cmd_` to the custom commands files and instead of loading all custom commands at initializing just load on demand the ones we want to use?

PR #10742, opened 2022-03-09
Missing conf entries for the subsystems

lasote:
Were they intentionally not added? Are they supposed to be at the profile?
Changelog: omit
Docs: omit


memsharded (approved):
> Were they intentionally not added?

No, probably missing.

memsharded (approved):

PR #10744, opened 2022-03-09
fixing package_id bugs and tests

memsharded:
- Some minor cleanings and simplifications on package_id RequirementInfo
- Fixed one bug on package_id computation for applications

### DIFF
--- a/conans/client/graph/compute_pid.py
+++ b/conans/client/graph/compute_pid.py
@@ -23,20 +23,20 @@ def compute_package_id(node, new_config):
     data = OrderedDict()
     build_data = OrderedDict()
     for require, transitive in node.transitive_deps.items():
-        dep_package_id = require.package_id_mode
+        dep_package_id_mode = require.package_id_mode  # the package_id_mode defined as Require trait
         dep_node = transitive.node
 
         if require.build:
-            if dep_package_id:
-                req_info = RequirementInfo(dep_node.pref, dep_package_id)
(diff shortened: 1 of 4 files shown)


conans/client/graph/compute_pid.py
```diff
@@ -25,16 +25,18 @@ def compute_package_id(node, new_config):
     for require, transitive in node.transitive_deps.items():
         dep_package_id = require.package_id_mode
```
lasote: Side note, the `dep_package_id` variable is a very confusing name when it is the package id mode.

conans/model/info.py
```diff
@@ -513,26 +434,28 @@ def package_id(self):
         options and settings
         """
```
lasote, line 435: just to make sure, this is going to change the package_id for the currently built packages (in the v2 repository), right?

lasote (approved):

conans/model/info.py
memsharded: Most likely not, as it didnt change any package_id in the codebase. This change only kicks in when building executables, defined with the ``package_type = "application"``. If I understood correctly, @czoido is faking the cmake package build at the moment, so shouldn't be an issue.

conans/client/graph/compute_pid.py
memsharded: Renamed to ``dep_package_id_mode``

PR #10749, opened 2022-03-10
Add failing test when removing options that already have a value

czoido:
`conan install . -o pkg*:shared=True --build=missing`
over a package that defines something like:
```
def configure(self):
    if self.options.shared:
        del self.options.fPIC
```
fails with:
`ConanException: Incorrect attempt to remove option 'fPIC' with current value 'True' `


memsharded:
I have relaxed the exception, not raising anymore.
Removing with ``del self.options.xxx`` in ``configure()`` will be allowed (what is restricted is assigning values).

lasote (approved):

PR #10753, opened 2022-03-10
removing legacy dead SCM code

memsharded:


conans/cli/api/subapi/upload.py
```diff
@@ -62,12 +62,11 @@ def prepare(self, upload_bundle):
         preparator.prepare(upload_bundle)
 
     @api_method
-    def upload_bundle(self, upload_bundle, remote, force=False):
```
franramirez688, line 65: For me to understand it, why is `force` removed?

memsharded: Only the argument. It was used only to avoid uploading dirty "scm" packages, but scm doesn't exist anymore.
We will block the "dirty" packages earlier in the process (at export) if anything.

franramirez688: Got it! Thanks 😁

franramirez688 (approved):

PR #10770, opened 2022-03-14
CMAKE_MAKE_PROGRAM for mingw

lasote:
Changelog: Fix: Set `-DCMAKE_MAKE_PROGRAM` when the generator is for MinGW and the conf `tools.gnu:make_program` is set.
Docs: omit

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -83,8 +83,13 @@ class CMake(object):
             pkg_folder = self._conanfile.package_folder.replace("\\", "/")
             arg_list.append('-DCMAKE_INSTALL_PREFIX="{}"'.format(pkg_folder))
         if platform.system() == "Windows" and self._generator == "MinGW Makefiles":
-            # It seems this doesn't work in the toolchain file, it needs to be here in command line
+            # It seems these don't work in the toolchain file, they need to be here in command line
             arg_list.append('-DCMAKE_SH="CMAKE_SH-NOTFOUND"')
+            cmake_make_program = self._conanfile.conf.get("tools.gnu:make_program", default=None)
+            if cmake_make_program:
+                cmake_make_program = cmake_make_program.replace("\\", "/")
+                arg_list.append('-DCMAKE_MAKE_PROGRAM="{}"'.format(cmake_make_program))
+
         if variables:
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -736,6 +744,11 @@ def context(self):
 
         system_name, system_version, system_processor = self._get_cross_build()
 
+        cmake_make_program = self._conanfile.conf.get("tools.gnu:make_program",
```
memsharded: Maybe the ``make_program`` will only be set in mingw profiles, nobody will define it if not necessary, no need to check the ``is_mingw()`` and no need to make it none? Maybe the variable is also useful for other compilers like ``qcc`` (qnx)?

Should it have a default ``make`` value or this is only necessary when make is not in the path?

conan/tools/cmake/cmake.py
```diff
@@ -83,8 +83,13 @@ def configure(self, variables=None, build_script_folder=None):
             pkg_folder = self._conanfile.package_folder.replace("\\", "/")
             arg_list.append('-DCMAKE_INSTALL_PREFIX="{}"'.format(pkg_folder))
         if platform.system() == "Windows" and self._generator == "MinGW Makefiles":
-            # It seems this doesn't work in the toolchain file, it needs to be here in command line
+            # It seems these don't work in the toolchain file, they need to be here in command line
```
memsharded, line 86: I wonder if this is because the toolchain doesnt set the variables with "CACHE" status?

lasote: no, already tried.

conan/tools/cmake/cmake.py
```diff
@@ -83,8 +83,13 @@ def configure(self, variables=None, build_script_folder=None):
...
             arg_list.append('-DCMAKE_SH="CMAKE_SH-NOTFOUND"')
+            cmake_make_program = self._conanfile.conf.get("tools.gnu:make_program", default=None)
```
memsharded, line 88: I have had a second look, just in case, and it really seems difficult.
So I am approving and merging this. However, I think it would be better if we started storing this arguments in ``CMakeToolchain`` in the ``conanbuild.conf`` file, or even better, if we start considering moving to a CMakePresets.json compatible format.

memsharded (approved):

PR #10713, opened 2022-03-03
[develop2] proof of concept of adding overrides to lockfiles

memsharded:
Via ``conan lock add`` new command

Close https://github.com/conan-io/conan/pull/10046
Close https://github.com/conan-io/conan/issues/9912

PR #10771, opened 2022-03-14
removing --require-override, now is in lockfiles

memsharded:
After https://github.com/conan-io/conan/pull/10713, the ``--require-override`` cli arg doesn't make sense anymore, removing


franramirez688 (approved):
@memsharded 
For me to know, why does it not make sense anymore? I mean, this arg was applied for install, create, etc. so I think they are different things, aren't they?

memsharded:
> For me to know, why does it not make sense anymore? I mean, this arg was applied for install, create, etc. so I think they are different things, aren't they?

It is replaced with ``conan lock add`` + ``conan install --lockfile=xxx`` flow

PR #10775, opened 2022-03-14
2.0 Fix subsystems run + test

lasote:
Changelog: Bugfix: Fixed crash when running commands inside a windows subsystem (`self.win_bash=True`).
Docs: omit

### DIFF
--- a/conans/test/functional/subsystems_build_test.py
+++ b/conans/test/functional/subsystems_build_test.py
@@ -45,7 +45,7 @@ class TestSubsystems:
         assert "'uname' is not recognized as an internal or external command" in client.out
 
 
-#@pytest.mark.skipif(platform.system() != "Windows", reason="Tests Windows Subsystems")
+@pytest.mark.skipif(platform.system() != "Windows", reason="Tests Windows Subsystems")
 class TestSubsystemsBuild:
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


conans/test/functional/subsystems_build_test.py
```diff
@@ -88,6 +89,64 @@ def test_mingw64(self):
...
+
+                    def test(self):
+                        self.run("app")
+                """)
+        profile = textwrap.dedent("""
+        include(default)
+        [conf]
+        tools.microsoft.bash:subsystem=msys2
+        tools.microsoft.bash:path=bash.exe
+        tools.gnu:make_program=mingw32-make.exe
```
memsharded: These maybe reasonable defaults (once msys2 and gcc are defined)?
I mean, these will probably take these values 95% of the time, so consider simplifying and defaulting to them (conditioned to msys2, gcc in Windows)

memsharded (approved):
Good!

conans/test/functional/subsystems_build_test.py
memsharded: Actually, it seems this is already the default for mingw32-make?
```python
make_program = self._conanfile.conf.get("tools.gnu:make_program",
                                                default="mingw32-make" if self._use_win_mingw() else "make")
```

I'd say we do the same default for ``bash.exe``

lasote: yes, the `mingw32-make` has already defaulted. About the `bash.exe`, it might be a good default, I would change it, but it might be possible (also in the test specifying `tools.microsoft.bash:path=bash.exe`) that it finds another `bash.exe` from `git` or another subsystem, it would be quite confusing but the user can change it anyway...

lasote:
Check again please

memsharded (approved):

Issue #10784, opened 2022-03-15
[feature][develop2] Remove ``user_info``, superseded by ``conf``


memsharded:
The ``conf`` is more structured and powerful, with appends, prepends, and operations on values

memsharded:
Closed by https://github.com/conan-io/conan/pull/10794

PR #10794, opened 2022-03-15
remove UserInfo, use ``conf_info`` instead

memsharded:
Close https://github.com/conan-io/conan/issues/10784

There are open questions regarding the propagation of ``conf_info``: should we propagate it from normal requires?


memsharded:
> There are open questions regarding the propagation of conf_info: should we propagate it from normal requires?

Discussed with @lasote. As the use cases are not that evident, lets do explicit ``self.dependencies["..."].conf_info`` initially for host configurations, and wait for user feedback and cases.

czoido (approved):

franramirez688 (approved):

lasote (approved):

PR #10800, opened 2022-03-16
[conf] Global flags management

### LINKED ISSUE #10452: [feature] New generators and build helpers listening to global flags from `[conf]` section
Depends on: https://github.com/conan-io/conan/issues/10451

For instance, given this global configuration (proposal):

```
[conf]

tools.build:cxxflags+=flag1
tools.build:cflags=+flag4
tools.build:cppflags=+flag3
tools.build:ldflags=flag1 flag2 flag3
tools.build:sharedlinkflags=flag2
tools.build:exelinkflags=flag5
```
New toolchains and/or build helpers from `conan.tools.xxxxx` that use those flags (in their own way) should be able to manage them properly.

### LINKED ISSUE #10805: [feature] Handling fine control of builds: C[XX]FLAGS, etc.
Following a [discussion on the cpplang slack](https://cpplang.slack.com/archives/C41CWV9HA/p1646556536574179), I'm moving discussion here.

The discussion more or less goes like this:

I'm the author of library L1.
In my library L1, I want to use dependencies D1, D2, D3.
D1, D2, D3 consumed via Conan, of course using a `conanfile.[txt|py]`, whatever would be more convenient. 
CMake for building L1.
For whatever reason, it would be beneficial to build D# with flag `-flagD#`.

How can this be achieved with Conan?

Conan doesn't have a mechanism to pass arbitrary additional compilation flags, or to set environment variables for dependencies.

And this is where the problem starts.

Assuming the dependency's build system respects the `CXXFLAGS` environment variable, we could go with that route. But:

- Conan doesn't take `CXXFLAGS` into account when building package IDs, so without `--build always` it would just pick binaries from the repository anyway;
- There isn't a straightforward way to set this environment variable anyway, leading to the following discussion:

The first suggestion was to use profiles, shipped alongside the library L1, but that's a big can of worms:

Profiles would require me to set this section:
```
[settings]
os=
arch=
compiler=
compiler.version=
```
for every compiler I expect users/devs of L1 to use. That's pretty much impossible.
Another option is to _not_ do that and have a simpler profile with just the `[env]` section filled in, but since profiles can't be "merged" at "usage time", users/devs of L1 can't use their compiler settings when `conan [create|install]`-ing:
```
conan install -s compiler="Visual Studio" -s compiler.version="16"  # etc etc etc, even with a profile...
```

Compare this with, for example, how CMake does it:
```cmake
if(MSVC)
  add_compile_options("/fp:fast")
else()
  add_compile_options("-ffast-math")
endif()
```
The `else()` clause is pretty much guaranteed to work on pretty much "all" GCC or Clang versions, easily. 
This also applies to dependencies: For example, if I used `git submodule` + `add_subdirectory`, or even an `ExternalContent` /`FetchContent` workflow  it would be possible to control the way `CMAKE_CXX_FLAGS`  would be initialized indeed for those dependencies.

I guess an option would be to have a `extra_cxx_flags` option for all recipes. If you ever want to consider doing this to CCI, count me in to help. For example, an `extra_cxx_flags` option could translate in `f"-DCMAKE_CXX_FLAGS={extra_cxx_flags}"` argument passed via the CMake build helper. 

The _really_ high level option would be to have something like a `mydependency1.cxxflags=["-ffast-math"] #pseudocode` setting in a consumer `conanfile.txt`, but that's a high dream. 
Imagine this:
```
self.dependencies.["D1"].cxx_flags = ["-f1"] # iterable for append, literal for replace, like profiles do
```

Read the slack discussion for more context, or feel free to ask questions. 

In conclusion, I'd like a system to be able to have this fine control over how my dependencies are built.

franramirez688:
Changelog: Feature: Added mechanism to inject extra flags via `[conf]` into several toolchains like `AutotoolsToolchain`, `MesonToolchain` and `CMakeToolchain`.

* `tools.build:cxxflags`
* `tools.build:cflags`
* ~`tools.build:cppflags`~ -> Changed to `tools.build:defines` (https://github.com/conan-io/conan/pull/10928)
* ~`tools.build:ldflags`~ -> Removed (https://github.com/conan-io/conan/pull/10928)
* `tools.build:sharedlinkflags`
* `tools.build:exelinkflags`

Changelog: Fix: `Conf.get()` always returns `default` value if internal `conf_value.value` is `None`, i.e., it was unset.

Docs: https://github.com/conan-io/docs/pull/2484
Closes: https://github.com/conan-io/conan/issues/10805 https://github.com/conan-io/conan/issues/10452 https://github.com/conan-io/conan/issues/10895

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -14,7 +14,7 @@ from conan.tools.cmake.utils import is_multi_configuration
 from conan.tools.intel import IntelCC
 from conan.tools.microsoft.visual import is_msvc, msvc_version_to_toolset_version
 from conans.errors import ConanException
-from conans.model.conf import ConfDefinition
+from conans.model.conf import Conf
 from conans.util.files import load
 
 
@@ -502,17 +502,10 @@ class CMakeFlagsInitBlock(Block):
         string(APPEND CMAKE_EXE_LINKER_FLAGS_INIT " {{ exelinkflags }}")
     """)
 
-    conan_conf_template = textwrap.dedent("""
-        tools.build:cxxflags=["{cxxflags}"]
-        tools.build:cflags=["{cflags}"]
-        tools.build:sharedlinkflags=["{sharedlinkflags}"]
-        tools.build:exelinkflags=["{exelinkflags}"]
-    """)
-
     def __init__(self, conanfile, toolchain):
         super(CMakeFlagsInitBlock, self).__init__(conanfile, toolchain)
         # Load all the predefined Conan C, CXX, etc. flags
-        self._conan_conf = ConfDefinition()
+        self._conan_conf = Conf()
         self._process_conan_flags()
 
     def _process_conan_flags(self):
@@ -521,13 +514,10 @@ class CMakeFlagsInitBlock(Block):
         glib_flag = self._get_glib_flag()
 
         # Defining all the flags
-        flags = self.conan_conf_template.format(
-            cxxflags=" ".join([glib_flag, arch_flag, mp_flag]),
-            cflags=" ".join([arch_flag, mp_flag]),
-            sharedlinkflags=arch_flag,
-            exelinkflags=arch_flag
-        )
-        self._conan_conf.loads(flags)
+        self._conan_conf.define("tools.build:cxxflags", [glib_flag, arch_flag, mp_flag])
+        self._conan_conf.define("tools.build:cflags", [arch_flag, mp_flag])
+        self._conan_conf.define("tools.build:sharedlinkflags", [arch_flag])
+        self._conan_conf.define("tools.build:exelinkflags", [arch_flag])
 
     def _get_parallel_jobs_flags(self):
         compiler = self._conanfile.settings.get_safe("compiler")
@@ -569,8 +559,8 @@ class CMakeFlagsInitBlock(Block):
         return glib_flag
 
     def context(self):
-        # now, it's time to update the predefined flags with [conf] ones injected by the user
-        self._conan_conf.update_conf_definition(self._conanfile.conf)
+        # Now, it's time to update the predefined flags with [conf] ones injected by the user
+        self._conan_conf.compose_conf(self._conanfile.conf)
         cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list)
         cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list)
         sharedlinkflags = self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -576,10 +576,10 @@ class CMakeFlagsInitBlock(Block):
         sharedlinkflags = self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)
         exelinkflags = self._conanfile.conf.get("tools.build:exelinkflags", default=[], check_type=list)
         return {
-            "cxxflags": cxxflags,
-            "cflags": cflags,
-            "sharedlinkflags": sharedlinkflags,
-            "exelinkflags": exelinkflags,
+            "cxxflags": " ".join(cxxflags),
+            "cflags": " ".join(cflags),
+            "sharedlinkflags": " ".join(sharedlinkflags),
+            "exelinkflags": " ".join(exelinkflags),
         }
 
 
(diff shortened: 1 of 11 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -549,19 +496,90 @@ def context(self):
...
+        return glib_flag
+
+    def _process_conan_flags(self):
+        """Calculating all the flags predefined by Conan"""
+        mp_flag = self._get_parallel_jobs_flags()
+        arch_flag = self._get_arch_flag()
+        glib_flag = self._get_glib_flag()
+
+        # Defining all the flags
+        self._conan_conf.define("tools.build:cxxflags", [arch_flag, glib_flag, mp_flag])
```
memsharded: I am not sure the aggregation has to be via conf, maybe it has to be defined via CMake instead (in the toolchain).
So the toolchain could define in one block/section some flags, as defined by the block (architecture, parallel, etc)

Then, other block could define the profile/conf variables, depending on the action can be added or overwritten.
But I don't love the idea of moving all the logic in "glibc", "architecture", etc blocks to a single block.

Lets see what others think.

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -547,6 +547,54 @@ def context(self):
...
+class ExtraFlagsBlock(Block):
+    """This block is adding flags directly from user [conf] section"""
+
+    template = textwrap.dedent("""
+        {% if cxxflags %}
+        string(APPEND CONAN_CXX_FLAGS
+        {% for cxxflag in cxxflags %}
+        " {{ cxxflag }}"
+        {% endfor %}
+        )
```
czoido: Minor detail, but maybe we could improve formatting of this template, it formats like:

```cmake
string(APPEND CONAN_CXX_FLAGS
" -Wpedantic"
" -fstack-protector-strong"
)
```
that is very different from other variables set in the toolchail

```suggestion
        {% if cxxflags %}
        string(APPEND CONAN_CXX_FLAGS "{% for cxxflag in cxxflags %} {{ cxxflag }}{% endfor %}")
        {% endif %}
```

conans/test/unittests/client/toolchain/autotools/autotools_toolchain_test.py
```diff
@@ -408,3 +408,22 @@ def test_custom_ldflags():
 
     assert "MyFlag" not in env["CXXFLAGS"]
     assert "MyFlag" not in env["CFLAGS"]
+
+
+def test_extra_flags_via_conf():
```
czoido, line 413: Maybe we could also add an integration or functional test?

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -15,33 +15,35 @@ class AutotoolsToolchain:
...
         self.make_args = []
         self.default_configure_install_args = True
 
+        # Flags
+        self.cxxflags = []
+        self.cflags = []
+        self.ldflags = []
+        self.defines = []
+
         # TODO: compiler.runtime for Visual studio?
```
lasote: This comment was already there but please, remove it. It is already implemented

```suggestion
```

conans/model/conf.py
```diff
@@ -228,6 +235,8 @@ def get(self, conf_name, default=None, check_type=None):
                 return self._get_boolean_value(v)
             elif check_type is str and not isinstance(v, str):
                 return str(v)
+            elif v is None:  # value was unset
```
lasote, line 238: 🤔  when does this happen?

lasote (approved):

conans/model/conf.py
franramirez688: Good point! This is happening when you unset a variable and you want to get its value (I added a few lines in `conans/test/unittests/model/test_conf.py` to test it):
```
[conf]
user.company.check:shared=!
```
```python
    assert c.get("user.company.check:shared") is None  # unset value
    assert c.get("user.company.check:shared", default=[]) == []  # returning default
    assert c.get("user.company.check:shared", default=[], check_type=list) == []  # not raising exception
```
Before my change, that `check_type=list` was raising an exception because the value was `None` but I thought it makes sense that you get the `default` value instead of raising the exception.

czoido (approved):

ruilvo:
Question: Is this [conf] section accessible from recipes directly? I was thinking, on recipes that don't use a toolchain or helper, like boost, replicating this manually at recipe level.

memsharded:
> Question: Is this [conf] section accessible from recipes directly? I was thinking, on recipes that don't use a toolchain or helper, like boost, replicating this manually at recipe level.

All the conf is accessible via ``self.conf.get("xxxxx")`` inside recipes, yes. The idea is that the conf in general can be used, you can also define your own ``user.xxxx`` confs in profiles and used them in recipes.

PR #10804, opened 2022-03-17
Slack alert when failed develop2

czoido:


memsharded (commented):
Nice! To which channel? is it hardcoded?

czoido:
> Nice! To which channel? is it hardcoded?

Yes, it's configured inside Jenkins, it will publish in the #conan_ci channel in Conan slack, like the preflight and the docs links.

PR #10812, opened 2022-03-17
Bazel integration improvements

### LINKED ISSUE #10676: [bug] Cannot make work basic Bazel example on Windows 10 with Bazel 5
Firstly, I want to thank you for providing basic support of Bazel, this is great news!

But I cannot make the bazel-example integration sample work - https://docs.conan.io/en/1.44/integrations/build_system/bazel.html .
Looks like some key information is missing for this page and probably BazelDeps generator doesn't provide right output for Bazel 5.

About missing parts:
1. I had to create 'conandeps' folder and put there conanfile.py file.
I didn't know that just running `conan install .` on conanfile.py which is inside the root folder (Bazel workspace) will not create this folder and/or
`load("@//conandeps:dependencies.bzl", "load_conan_dependencies")
load_conan_dependencies()`
will not do some magic to make dependencies work.
Probably some steps to explain this process should be provided here.
2. I had to call `conan install . --build missing`, because boost binaries of version 1.76.0 for my OS and compiler doesn't exist in the ConanCenter.
Probably some link here to some page will make easier this step.
3. I cannot build boost 1.76.0 for my OS and compiler due to some error from b2.exe, so I had to switch boost version to 1.78.0.
Not sure what can be done here, but maybe some note will also help.
4. I had to add BazelDeps generator from this page - https://github.com/conan-io/conan/issues/10399 to conanfile.py to make Bazel/Conan integration work.

After some kind of success I tried to build my project with boost dependency:
`bazel-5.0.0-windows-x86_64.exe build //main:bazel-example`
and started to obtain such errors:
`ERROR: C:/users/denjm/_bazel_denjm/xv66akve/external/boost/BUILD.bazel:7:26: invalid escape sequence: \.. Use '\\' to insert '\'.`

I tried to google the problem, possibly found a solution by setting of `--incompatible_restrict_string_escapes` Bazel flag to false.
But looks like Bazel 5 doesn't support this flag anymore - https://github.com/bazelbuild/bazel/commit/21b4b1a489c991682b9fcc9a10b468b9ea18e931 .

So, I think that BazelDeps generator should be fixed to provide proper Bazel files.

  * Operating System+version: Windows 10 Pro build 19043.1526
  * Compiler+version: Visual Studio 17
  * Conan version: 1.45.0
  * Python version: 3.10.2
  * Bazel version: 5.0.0

1. Download and unzip attached project 
[bazel-example.zip](https://github.com/conan-io/conan/files/8155226/bazel-example.zip)
2. Install dependencies with Conan inside of conandeps folder.
3. Try to build project with Bazel 5:
`bazel-5.0.0-windows-x86_64.exe build //main:bazel-example`

[command.log](https://github.com/conan-io/conan/files/8155209/command.log)

lasote:
Changelog: Fix: Several improvements of the Bazel integration (`Bazel`, `BazelToolchain`, `BazelDeps`), new functional tests in all platforms. 
Changelog: Feature: New templates for the `conan new` command with bazel examples: `bazel_exe` and `bazel_lib`. 
Docs: https://github.com/conan-io/docs/pull/2474

Close #10676

- Added "conan new" templates for bazel.
- Added a "bazel_layout"
- Added a couple of functional tests working and using the confs to pass "rc" file and configs with the properties.
- Transformed `tools.google.bazel:config` into a list.
- Added a transitive dependency, the "lib + test_package" to validate the bugs of the BazelDeps.
- Editable packages are not supported yet, they are using the package_folder to locate the libraries.


conan/tools/google/bazel.py
```diff
@@ -40,5 +39,6 @@ def build(self, args=None, label=None):
     def _get_bazel_project_configuration(self):
         toolchain_file_content = load_toolchain_args(self._conanfile.generators_folder,
                                                      namespace=self._namespace)
-        self._bazel_config = toolchain_file_content.get("bazel_config")
+        configs = toolchain_file_content.get("bazel_configs")
+        self._bazel_config = configs.split(",") if configs else []
```
lasote, line 43: a bit dirty, storing a list type with a comma-separated string.

conan/tools/google/bazeldeps.py
```diff
@@ -115,41 +118,69 @@ def _relativize_path(p, base_path):
...
             "linkopts": linkopts,
             "library_type": "shared_library" if shared_library else "static_library",
-            "extension": "so" if shared_library else "a",
             "dependencies": dependencies,
         }
         content = Template(template).render(**context)
         return content
 
+    def _resolve_real_file(self, libdirs, lib):
+        for libdir in libdirs:
```
lasote, line 162: Note that this is looking for the real files (code extracted from the collect_libs)

franramirez688: I guess we could try another clearer name... Perhaps `def _get_stored_lib_paths` or something similar because `_resolve_real_file` is a bit confusing for me

franramirez688 (approved):

conan/tools/google/bazeldeps.py
lasote: Renamed to `_get_libs_absolute_paths`. Is that better?

lasote: Mhhh, no, these are not absolute. Better `_get_lib_file_paths`?

franramirez688: Fine for me 😁

conan/tools/google/bazel.py
memsharded: Yes, this is a bit dirty. I don't recall why exactly we changed our previous json toolchain_file for the INI one, but it seems the INI file is being a bit limited?

conan/tools/google/bazeldeps.py
memsharded: This could be close to the code that we would want to do to detect the static/shared libs, in order to avoid the UNKNOWN IMPORTED in CMakeDeps. Might be worth refactorizing from here (later, I am commenting here for awareness) if that is the way forward.

conans/test/functional/toolchains/google/test_bazel.py
```diff
@@ -1,125 +1,251 @@
...
-                bazel = Bazel(self)
-                bazel.configure()
-                bazel.build(label="//app:main")
-
-            def package(self):
-                self.copy('*', src='bazel-bin')
+@pytest.fixture(scope="module")
+def bazelrc():
+    return textwrap.dedent("""
+        build:Debug -c dbg --copt=-g
```
memsharded, line 12: Isn't it a bit weird that this is not part of the ``conan new`` template? why not?

conans/test/unittests/tools/google/test_bazeldeps.py
```diff
@@ -34,8 +40,11 @@ def test_bazeldeps_dependency_buildfiles():
         for dependency in bazeldeps._conanfile.dependencies.host.values():
             dependency_content = bazeldeps._get_dependency_buildfile_content(dependency)
             assert 'cc_library(\n    name = "OriginalDepName",' in dependency_content
-            assert 'defines = ["DUMMY_DEFINE=\'string/value\'"],' in dependency_content
-            assert 'linkopts = ["-lsystem_lib1"],' in dependency_content
+            assert """defines = ["DUMMY_DEFINE=\\\\\\"string/value\\\\\\""]""" in dependency_content
```
memsharded, line 43: So many backslashes necessary? Quite ugly...

memsharded (approved):

conans/test/unittests/tools/google/test_bazeldeps.py
lasote: Just the half of them, the rest is to escape the assert. 🤣 
The rest are absolutely necessary.

conan/tools/google/bazel.py
lasote: yes. I don't recall either. @franramirez688  do you?

conan/tools/google/bazeldeps.py
lasote: Yep, I thought the same.

conans/test/functional/toolchains/google/test_bazel.py
lasote: Ok, It might be useful but take into account, that is just an "invented" configuration file for bazel, it is very likely that a bazel user use different values, the "pre-cooked" build types in bazel are "opt", "dbg" and "fastbuild". Also the `--color` and the `withTimeStamps` are totally useless and just make sense in the test.

Also, the bazelrc is a file that makes no sense in the project, but in a common directory (like a profile).

lasote: Do we generate in the "conan new" also some base profiles (Release, Debug) setting the config name matching the bazelrc?

conan/tools/google/bazel.py
lasote: Taking a second look it looks like the thing to store here is directly the command line arguments to call bazel: `--config foo --config bar` like we've done with others? Note, the `--bazelrc=juernes` has to be stored in another variable because the argument goes before the "build" word:

```
command = 'bazel {} build {} {}'.format(
            bazelrc_path,
            bazel_configs,
            label
        )
```

Example Autotools:
``` 
```

franramirez688: I got the original issue: https://github.com/conan-io/conan/issues/9130

franramirez688: I'm checking the ConfigParser documentation and I saw the `converters` keyword. Perhaps, we could add something like this:
```python
    toolchain_config = configparser.ConfigParser(converters={"list": lambda x: x.split(",")})
    # ...
    configs = toolchain_file_content.getlist("bazel_configs", fallback=[])
```
So, adding that `converters` keyword into the `load_toolchain_args` function, we could pass our own `converters` like that one:
```python
def load_toolchain_args(generators_folder=None, namespace=None, converters=None):
```

memsharded: I am still not very convinced. I'd like to try the CMakePresets.json-like file approach, it would be a great thing for modern CMake users, it is already demanded. But it is a json, if we cannot store things in json, this won't be feasible at all? Maybe things can be stored un-scaped in the json?

PR #10814, opened 2022-03-17
[develop2] General conf system for new package_id computation

memsharded:


conans/client/graph/compute_pid.py
```diff
@@ -12,9 +12,13 @@ def compute_package_id(node, new_config):
...
-    # Todo: revise this default too. Should have been defined by requirement traits?
-    default_package_id_mode = new_config.get("core.package_id:default_mode", default="semver_mode")
-    default_python_requires_id_mode = new_config.get("core.package_id:python_default_mode", default="minor_mode")
+
+    unknown_mode = new_config.get("core.package_id:default_unknown_mode", default="semver_mode")
+    lib_mode = new_config.get("core.package_id:default_lib_mode", default="minor_mode")
+    # TODO: Change it to "full_mode" including package_id
+    bin_mode = new_config.get("core.package_id:default_bin_mode", default="recipe_revision_mode")
+    python_mode = new_config.get("core.package_id:default_python_mode", default="minor_mode")
+    build_mode = new_config.get("core.package_id:default_build_mode", default=None)
```
lasote, line 21: This is for the build-requires, right? So by default a "build require" doesn't affect the package_id unless you decide, for example, if you are requiring some build scripts or toolchain altering some flags? Did I get it?

lasote (approved):

conans/client/graph/compute_pid.py
memsharded: Yes, the default for build_requires is to not affect the binaries. You can always opt-in explicitly in recipes, but if you want something general, you can define it here. At the moment this one only affects **direct** tool_requires, because you don't really care ``mypkg -> cmake -> openssl``, the transitive deps of cmake most likely don't affect you, and everything is contained in the cmake version

PR #10817, opened 2022-03-17
[develop2] self.info.options for package_id is validating, and will raise if some value not defined

memsharded:


conans/test/unittests/model/options_test.py
```diff
@@ -11,7 +11,7 @@ class TestOptions:
 
     @pytest.fixture(autouse=True)
     def _setup(self):
-        options = {"static": [True, False], "optimized": [2, 3, 4], "path": "ANY"}
+        options = {"static": [True, False], "optimized": [2, 3, 4], "path": ["ANY"]}
```
memsharded, line 14: This is changing. I am now forcing to declare "ANY" as an element in a list. It simplifies things and is a bit more consistent too.

conans/test/unittests/model/options_test.py
```diff
@@ -163,7 +163,7 @@ def test_basic(self):
 class TestOptionsNone:
     @pytest.fixture(autouse=True)
     def _setup(self):
-        options = {"static": [None, 1, 2], "other": "ANY", "more": ["None", 1]}
+        options = {"static": [None, 1, 2], "other": [None, "ANY"], "more": ["None", 1]}
```
memsharded, line 166: This aligns with the behavior of settings, that need to declare this way to allow undefined value and any value. So far options were behaving differently and "ANY" allowed for None too.

conans/model/info.py
```diff
@@ -450,6 +454,10 @@ def package_id(self):
         result.append("")  # Append endline so file ends with LF
         text = '\n'.join(result)
         package_id = sha1(text.encode())
+        try:
+            self.options.validate()
+        except ConanException as e:
+            self.invalid = BINARY_INVALID, str(e)
```
lasote, line 460: Recall me why you mark this as INVALID, I guess is for compatible packages, but put me an example.

conans/model/options.py
```diff
@@ -18,14 +18,24 @@ def __init__(self, name, value, possible_values=None):
...
+        if scope:
+            return "%s:%s=%s" % (scope, self._name, self._value)
+        else:
+            return "%s=%s" % (self._name, self._value)
+
     def copy_conaninfo_option(self):
         # To generate a copy without validation, for package_id info.options value
-        return _PackageOption(self._name, self._value)
+        assert self._possible_values is not None  # this should always come from recipe, with []
+        return _PackageOption(self._name, self._value, self._possible_values + ["ANY"])
```
lasote, line 38: I don't get this, so for the package id computation, is every option value is ok?

lasote (commented):
I'm not sure I get it.

conans/model/options.py
memsharded: This is before the package_id computation, but the ``package_id()`` method can change values with ``self.info.options.shared = "whatever"``. So the ``ANY`` option must be added to the possibilities if we don't want it to raise an exception.

That still needs a value defined, and will raise if the value is ``None`` (not defined)

conans/model/info.py
memsharded: Yes, it is for compatible packages. Example: 
- ``cmake`` package built with all the settings, including compiler. Settings not removed in the ``package_id`` method.
- ``cmake`` package required with only ``os`` and ``arch`` setting. It will compute a pcakage_id which is invalid, won't be able to be installed or built
- But if you have a ``compatible_packages`` or ``compatibilitly.py`` that defines compatibles different compilers etc, it will work.

conans/test/unittests/model/options_test.py
lasote: Could you add a test with ANY without the [] to see if the error is good enough and "pin" the behavior?

lasote (approved):

PR #10834, opened 2022-03-21
- BugFix: allow tested_reference_str to be None 

### LINKED ISSUE #10714: [question] how to run `conan info` on test_package?
for CI purpose, we need to collect all the `requiremenets` and `build_requirements`.
in reality, `test_package` may have it's own `requirements`/`build_requirements` that package itself doesn't have (that's a common practice in CCI).
but running `conan info` against `test_package` results in the following error:
```
$ conan info test_package/conanfile.py
ERROR: conanfile.py: Error in requirements() method, line 17
	self.requires(self.tested_reference_str)
	AttributeError: 'TestPackageConan' object has no attribute 'tested_reference_str'
```
I don't see any way to set `tested_reference_str`. is it possible to do it from `conan info` at all?
otherwise, the only solution I can imaging is applying `sed` on conanfile, which is kinda... weird

SSE4:
closes: #10714

Changelog: Feature: Allow `tested_reference_str` to be `None`.
Docs: omit

### DIFF
--- a/conans/test/integration/command/info/info_test.py
+++ b/conans/test/integration/command/info/info_test.py
@@ -720,3 +720,30 @@ class TestInfoPythonRequires:
         client.run("info . --json=file.json")
         info = json.loads(client.load("file.json"))
         assert info[0]["python_requires"] == ['tool/0.1#f3367e0e7d170aa12abccb175fee5f97']
+
+class TestInfoTestPackage:
+    # https://github.com/conan-io/conan/issues/10714
+
+    def test_tested_reference_str(self):
+        client = TestClient()
+        client.save({"conanfile.py": GenConanfile()})
+        client.run("export . tool/0.1@")
+
+        conanfile = textwrap.dedent("""
+from conans import ConanFile
+class HelloConan(ConanFile):
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+
+    def build_requirements(self):
+        self.build_requires(self.tested_reference_str)
+
+    test_type = 'explicit'
+""")
+        client.save({"conanfile.py": conanfile})
+
+        for args in ["", " --build-order tool/0.1@", " --build"]:
+            client.run("info . " + args)
+            assert "AttributeError: 'HelloConan' object has no attribute 'tested_reference_str'"\
+                   not in client.out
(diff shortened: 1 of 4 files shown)


lasote (changes requested):
We would like to avoid that new argument. We've been discussing and we might prefer if `self.requires(None)` and `self.build_requires(None)` don't do anything, that is actually more accurate to the current behavior with 1.X, because the whole graph is not computed either, just the test_package and its requires. Probably we need to initialize the `self.tested_reference_str` to None always. WDYT?

SSE4:
> We would like to avoid that new argument. We've been discussing and we might prefer if `self.requires(None)` and `self.build_requires(None)` don't do anything, that is actually more accurate to the current behavior with 1.X, because the whole graph is not computed either, just the test_package and its requires. Probably we need to initialize the `self.tested_reference_str` to None always. WDYT?

I am personally fine with that. it will allow to compute requirements of test_package in our CI, and we already have requirements of the package itself computed. (well, it might not work if test_package changes options of requirements, but it doesn't matter for CCI in this case, as it will fail for a different reason - missing package, cause package built is not one being tested).

SSE4:
@lasote I've changed the implementation according to your suggestion. `tested_reference_str` is initialized to `None`. `self.requires(None)` and `self.build_requires(None)` are no-ops now.

conans/test/integration/command/info/info_test.py
```diff
@@ -720,3 +720,23 @@ class Pkg(ConanFile):
...
+    def test_tested_reference_str(self):
+        client = TestClient()
+        client.save({"conanfile.py": GenConanfile()})
+        client.run("export . tool/0.1@")
+
+        conanfile = GenConanfile()\
+            .with_test_type("explicit")\
+            .with_requirement("placeholder")\
+            .with_build_requirement("placeholder")
+        conanfile = str(conanfile).replace('"placeholder"', 'self.tested_reference_str')
```
memsharded: Not worth the trick. Just write a explicit ``conanfile`` it will be just a couple of lines longer, but way more explicit and understandable.

memsharded (approved):

conans/test/integration/command/info/info_test.py
SSE4: okay, changed

lasote (approved):

Issue #10796, opened 2022-03-16
[question] Set conan_meson_cross.ini [properties] field?


avetiso:
There is a well known case in meson that it will fail to correctly auto detect if the build machine can execute host binaries if the build machine is Macos and the host is iOS because in my case the arch is arm64 in both. The result is that the build fails because Meson trys to execute the binary as part of of its 'compiler sanity check'. This fails of course, and stops the build, even though it was actually successful.

The suggested fix from Meson is to set

```
[properties]
needs_exe_wrapper=true
```

in conan_meson_cross.ini. However, there doesn't seem to be any way to add [properties] to this field from the Meson Toolchain. There's probably a way to do in the actual meson.build files, but I'd imagine for people trying to use conan for cross-platform, it would be useful to be able to set this at the profile level, or at the very least from the ConanFile via the Meson Toolchain.

Any thoughts on this?

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

eli-schwartz:
> There's probably a way to do in the actual meson.build files

Meson is principled here, and actually refuses to allow projects to encode information modifying the conan cross file properties. This does need to be solved as part of conan's build environment setup.

memsharded:
> Meson is principled here, and actually refuses to allow projects to encode information modifying the conan cross file properties.

I am not sure what that means, Meson is agnostic about Conan, it is not that it can refuse to encode information regarding Conan files, it is that it doesn't even know that such files exists. 

If the thing is that adding to ``[properties]`` custom things is necessary, I think that is an easy feature. The ``MesonToolchain`` already has support for ``[project_options]``, it would be identical. Someone wants to contribute it?

avetiso:
I can try to tackle this sometime this weekend, in the meantime simply layering an additional cross file with these properties works.

eli-schwartz:
> I am not sure what that means, Meson is agnostic about Conan, it is not that it can refuse to encode information regarding Conan files, it is that it doesn't even know that such files exists.

It means that conan provides a Meson cross file, and Meson doesn't permit *setting* values of Meson cross files from `meson.build` files because Meson strictly separates build recipes (`meson.build`) from environment setup (cross files).

> If the thing is that adding to `[properties]` custom things is necessary, I think that is an easy feature.

That is indeed what is necessary, and hopefully should be quite easy.

[...]

PR #10846, opened 2022-03-21
Add meson properties field, and specificy needs_exe_wrapper depending…

avetiso:
… on cross conditions.

Changelog: Feature: Add detection in meson toolchain for cross conditions and requirement of needs_exe_wrapper. Users may specify the exe wrapper in their meson.build now.
Docs: omit

Issue: https://github.com/conan-io/conan/issues/10796

Not sure if docs are required, as this is probably expected behavior of the generator for someone who understands what meson is looking for. Happy to write some docs, though.

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -17,6 +17,11 @@ class MesonToolchain(object):
     cross_filename = "conan_meson_cross.ini"
 
     _meson_file_template = textwrap.dedent("""
+    [properties]
+    {% for it, value in properties.items() -%}
+    {{it}} = {{value}}
+    {% endfor %}
+
     [constants]
     preprocessor_definitions = [{% for it, value in preprocessor_definitions.items() -%}
     '-D{{ it }}="{{ value}}"'{%- if not loop.last %}, {% endif %}{% endfor %}]
@@ -96,6 +101,7 @@ class MesonToolchain(object):
         else:
             self._b_vscrt = None
 
+        self.properties = {}
         self.project_options = {}
         self.preprocessor_definitions = {}
         self.pkg_config_path = self._conanfile.generators_folder
@@ -109,6 +115,7 @@ class MesonToolchain(object):
             os_build, arch_build, os_host, arch_host = get_cross_building_settings(self._conanfile)
             self.cross_build["build"] = to_meson_machine(os_build, arch_build)
             self.cross_build["host"] = to_meson_machine(os_host, arch_host)
+            self.properties["needs_exe_wrapper"] = True
             if hasattr(conanfile, 'settings_target') and conanfile.settings_target:
                 settings_target = conanfile.settings_target
                 os_target = settings_target.get_safe("os")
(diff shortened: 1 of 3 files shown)


eli-schwartz:
> Users may specify the exe wrapper in their meson.build now.

Err... I assume you mean something other than the `meson.build` file which is read by `meson setup`?

avetiso:
> > Users may specify the exe wrapper in their meson.build now.
> 
> Err... I assume you mean something other than the `meson.build` file which is read by `meson setup`?

Maybe? That's where I've seen most examples of exe_wrappers being set. Feel free to correct me here.

avetiso:
> > Users may specify the exe wrapper in their meson.build now.
> 
> Err... I assume you mean something other than the `meson.build` file which is read by `meson setup`?

Also, for the case of (arch_build != arch_host), do you think it makes sense to be explicit here re: needs_exe_wrapper when meson probably guess this correctly?

conan/tools/meson/toolchain.py
```diff
@@ -117,6 +123,8 @@ def __init__(self, conanfile, backend=None):
             if is_apple_os(os_host):  # default cross-compiler in Apple is common
                 default_comp = "clang"
                 default_comp_cpp = "clang++"
+            if (arch_host != arch_build) or (os_host == "iOS") or (os_host == "Android" and os_build != "Android"):
```
memsharded: There is a helper that is called ``cross_building()``. Maybe it is better to use it here? This seems a bit ad-hoc, assuming that iOS and Android is cross building. Can you please clarify a bit the logic when this property is needed?

memsharded (commented):
Some tests that covers this would be necessary. I'd recommend an integration test, doesn't need to actually build anything, just check the generated file contents under several conditions.

conan/tools/meson/toolchain.py
avetiso: `needs_exe_wrapper` is required whenever the compiled binary for the host is incapable of running on the build machine. Meson conducts compiler sanity checks and attempts to determine this itself. However, it will fail in certain cases, like armv8 iOS trying to run on Macos armv8. Meson thinks it makes sense to try and run the armv8 iOS binary because the build machine is also armv8, but it fails. This option is needed to suppress that check.

This method will add needs_exe_wrapper in other cases where meson can successfully autodetect, as well. I don't think that hurts anything, but perhaps it is overly explicit. For example, meson can tell that it shouldn't try to run an iOS x86_64 binary even if the option needs_exe_wrapper is not set. So it would be redundant (but still present) in that case. 

I don't have a problem simply using `cross_building()`. I'll go take a look at the integration tests and try to write some meaningful tests tonight.

avetiso: It does look like that branch of code is executed is things are cross_building anyways (`if cross_building(conanfile, skip_x86_64=True`), so I could simply always include `needs_exe_wrapper` if that block of code executes. But I'm not quite sure what `skip_x64_x86` does. Is that simply saying that compiling from 64 bit to 32 bit is not considered crosscompiling?

memsharded: > But I'm not quite sure what skip_x64_x86 does. Is that simply saying that compiling from 64 bit to 32 bit is not considered crosscompiling?

Yes, exactly, that means that a 32 bits ``x86`` binary in Windows runs fine in Windows x86_64. It is generally not considered cross-building in Windows builds.

avetiso:
Test for presence of fields in Android and iOS cross builds, and test for lack of presence in a native build. @memsharded let me know what you think. Seemed to make more sense to put the tests there rather in anything under integration/.

avetiso:
@memsharded any update feedback?

memsharded (approved):

memsharded:
Merged for 1.47, thanks!

PR #10856, opened 2022-03-22
Fix cross-build vars in CMakeToolchain

czoido:
Changelog: Fix: Setting `CMAKE_SYSTEM_PROCESSOR` for Apple cross-compiling including M1.
Changelog: Fix: Do not overwrite values for `CMAKE_SYSTEM_NAME`, `CMAKE_SYSTEM_VERSION` and `CMAKE_SYSTEM_PROCESSOR` set from the [conf] or the user_toolchain.
Changelog: Fix: Fix architecture translation from Conan syntax to build system in CMakeToolchain.
Docs: https://github.com/conan-io/docs/pull/2472

I tried to maintain all the logic like it was before for the “Generic” part and fix setting `CMAKE_SYSTEM_NAME`, `CMAKE_SYSTEM_VERSION` and `CMAKE_SYSTEM_PROCESSOR` for Apple.

**A bit of background on these variables:**

They are typically set when we cross-compile, just setting the `CMAKE_SYSTEM_NAME` variable manually will make that CMake sets `CMAKE_CROSSCOMPILING=True`. If you set `CMAKE_SYSTEM_NAME`, you should also set `CMAKE_SYSTEM_VERSION` and `CMAKE_SYSTEM_PROCESSOR`.

**The Apple case:**

From [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html):

> Apple case is a bit particular. Apple platforms other than macOS are handled differently to other cross compiling 
 scenarios. Rather than relying on CMAKE_SYSTEM_NAME to select the target platform, Apple device builds use 
 CMAKE_OSX_SYSROOT to select the appropriate SDK, which indirectly determines the target platform. 

That means for some cases we could not set `CMAKE_SYSTEM_NAME` and as we are setting `CMAKE_OSX_SYSROOT` everything should be ok. Seems that this holds true for the M1 case, in fact in that case we were not setting `CMAKE_SYSTEM_NAME` before.

Also from [CMake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_CROSSCOMPILING.html):

> Furthermore, when using the Xcode generator, developers can switch between device and simulator builds at build time  rather than having a single choice at configure time, so the concept of whether the build is cross compiling or not is  more complex. Therefore, the use of CMAKE_CROSSCOMPILING is not recommended for projects targeting Apple devices.

Maybe we should not set these variables for the Apple case but we may affect someone using `CMAKE_CROSSCOMPILING` (although it’s not recommended). 

I have also checked the popular [ios-cmake toolchain](https://github.com/leetal/ios-cmake ), and they seem to set those vars for the Apple case (with the same values that we do now) 

#TAGS: svn, slow


memsharded (commented):
One of the major advantages of the AppleBlock is that it is esier to maintain, reason about, or customize, if necessary, it is easy to replace the AppleBlock without touching the rest of cross building. More modular than a very large ``_get_cross_build`` method with all the cross-build logic for all platforms.

Just a balance, I am not saying this is worse than the current situation, I need to wrap my head around it first.

czoido:
> 

- I have added a test (_test_apple_vars_overwrite_user_conf_) that fails in develop and shows that the `[conf]` values for `CMAKE_SYSTEM_NAME/VERSION` are overwritten by the Apple block. I think we should not set the same values in two blocks and also they are more related to cross-building that to Apple.

- There's other thing I have noticed we were doing wrong, for Apple `CMAKE_SYSTEM_VERSION` should be set to the **sdk** version, not the **os** version

- I'm also not sure we are doing things right for M1, I'll try to investigate further and are more tests

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -574,20 +555,95 @@ class TryCompileBlock(Block):
...
+        os_host = self._conanfile.settings.get_safe("os")
+        arch_host = self._conanfile.settings.get_safe("arch")
+        arch_build = self._conanfile.settings_build.get_safe("arch")
+        if self._is_apple_cross_building():
+            return to_apple_arch(arch_host)
+        elif os_host != 'Android' and arch_host != arch_build:
+            return arch_host
+
+    def _get_cross_build(self):
+        user_toolchain = self._conanfile.conf.get("tools.cmake.cmaketoolchain:user_toolchain")
```
memsharded: Maybe (not related to this PR, just a possible improvement) it is worth honoring the other confs, and letting users passing both the ``user_toolchain`` and the other ``system_xxxx```confs. This can be done by swapping place and doing a ``return system_name, system_version, system_processor``.


The thing is that it would be weird to define ``tools.cmake.cmaketoolchain:system_name`` and that being completely rejected if you have ``tools.cmake.cmaketoolchain:user_toolchain`` defined?

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -574,20 +555,95 @@ class TryCompileBlock(Block):
...
+        os_host = self._conanfile.settings.get_safe("os")
+        if self._is_apple_cross_building():
+            # cross-build in Macos also for M1
+            return {'Macos': 'Darwin'}.get(os_host, os_host)
+        elif os_host != 'Android':
+            return self._get_generic_system_name()
+
+    def _system_version(self):
+        os_host = self._conanfile.settings.get_safe("os")
+        if self._is_apple_cross_building():
```
memsharded: This ``if self._is_apple_cross_building():`` is the possible factorization. If basically every method that does something starts with ``if self._is_apple_cross_building():``, then that means that it could be separated earlier?

memsharded (commented):
I agree that this is overall looking better.

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -726,6 +666,61 @@ def _get_compiler(self, generator):
...
+                    _system_version = self._conanfile.settings.get_safe("os.sdk_version")
+                    _system_processor = to_apple_arch(arch_host)
+                elif os_host != 'Android':
+                    system_name = self._get_generic_system_name()
+                    _system_version = self._conanfile.settings.get_safe("os.version")
+                    _system_processor = arch_host
+
+                if system_name is not None and system_version is None:
+                    system_version = _system_version
+                if system_name is not None and system_processor is None:
```
czoido, line 751: the logic changed a bit here, before we were only setting system_processor in the case that `if arch != arch_build:` but I think we should set this variable anyway if we activated the CMake cross-compiling setting CMAKE_SYSTEM_NAME manually.

lasote (approved):

franramirez688 (approved):

PR #10860, opened 2022-03-23
feature: Allow loading absolute paths in `.jinja` profiles

gabyx:
Changelog: Feature: Allow absolute and relative paths inside `.jinja` files.

This PR extends the `BaseLoader` sligthly by allowing to load full paths relative paths to `base_path`. Jinja2 does not allow this in its `FileSystemLoader` and its not aligned with the Conan text profile include statement `include(...)`.

Docs: https://github.com/conan-io/docs/pull/2466

- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request:
  `.jinja` templates can be located anywhere (lets say inside subfolder of root-dir (where `conanfile.py` lies. Syntax as `{% extends os.path.join(profile_dir, "../../default.jinja" %}` does not work since jinja does not allow ".." inside paths and also does not load full paths. 

### DIFF
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -129,7 +129,7 @@ def read_profile(profile_name, cwd, default_folder):
         class FullPathLoader(BaseLoader):
             """ Load absolute paths and relative paths to 'base_path'. """
             def __init__(self, base_path):
-                BaseLoader.__init__()
+                BaseLoader.__init__(self)
                 self.base_path = base_path
 
             def get_source(self, environment, template):
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -2,7 +2,7 @@ import os
 import platform
 from collections import OrderedDict, defaultdict
 
-from jinja2 import Environment, FileSystemLoader
+from jinja2 import Environment, BaseLoader
 
 from conan.tools.env.environment import ProfileEnvironment
 from conans.errors import ConanException, ConanV2Exception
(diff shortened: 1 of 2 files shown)


conans/client/profile_loader.py
```diff
@@ -125,7 +125,34 @@ def read_profile(profile_name, cwd, default_folder):
...
+
+                with open(template, "br") as f:
+                    contents = f.read().decode('utf-8')
+                    mtime = os.path.getmtime(template)
+
+                    def uptodate():
+                        try:
+                            return os.path.getmtime(template) == mtime
+                        except OSError:
+                            return False
```
lasote, line 150: This goes into more internal details of jinja that we would like (to maintain and manage).

lasote (commented):
Would this work avoid creating a custom subclass?

```
loader = jinja2.FileSystemLoader(base_path)
environment = Environment(loader=loader)
```

memsharded:
Hi, 

it has been a while with this PR not passing CI. In this time it doesn't seem that this feature has been requested by other users either. So I am closing this PR.

If there is a missing functionality in the profiles jinja templates, I'd recommend creating the unit-test that illustrate the need first, explaining a bit why that functionality is necessary. So far, it seems the relative ``imports`` jinja mechanism is working fine and would be good for using other jinja files, and having absolute paths seems very fragile, as it is a mechanism that would be easily broken when using those profiles in other machines (recommended ``conan config install``), so using relative paths seems always better.

Thanks!

PR #10862, opened 2022-03-23
[develop2] improvements in package_id

memsharded:
- New tests, a bit more exhaustive
- Fixed a bug in header->library
- Remove deducing package_id_mode=unknown(semver), for non-library, non-package-type known cases

### DIFF
--- a/conans/client/graph/compute_pid.py
+++ b/conans/client/graph/compute_pid.py
@@ -13,9 +13,9 @@ def compute_package_id(node, new_config):
     conanfile = node.conanfile
 
     unknown_mode = new_config.get("core.package_id:default_unknown_mode", default="semver_mode")
-    lib_mode = new_config.get("core.package_id:default_lib_mode", default="minor_mode")
+    non_embed_mode = new_config.get("core.package_id:default_non_embed_mode", default="minor_mode")
     # recipe_revision_mode already takes into account the package_id
-    bin_mode = new_config.get("core.package_id:default_bin_mode", default="recipe_revision_mode")
+    embed_mode = new_config.get("core.package_id:default_embed_mode", default="full_mode")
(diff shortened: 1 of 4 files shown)


conans/client/graph/compute_pid.py
```diff
@@ -14,7 +14,7 @@ def compute_package_id(node, new_config):
 
     unknown_mode = new_config.get("core.package_id:default_unknown_mode", default="semver_mode")
     lib_mode = new_config.get("core.package_id:default_lib_mode", default="minor_mode")
-    # TODO: Change it to "full_mode" including package_id
+    # recipe_revision_mode already takes into account the package_id
```
lasote, line 17: Should we rename it to `full_mode`?

conans/model/requires.py
```diff
@@ -317,6 +317,10 @@ def deduce_package_id_mode(self, pkg_type, dep_pkg_type, lib_mode, bin_mode, bui
                 self.package_id_mode = build_mode
             return  # At the moment no defaults
 
+        if pkg_type is PackageType.HEADER:
```
lasote, line 320: Do we know what is `dep_pkg_type` here? is it UNKNOWN? Could we assert it?

lasote (approved):

conans/client/graph/compute_pid.py
memsharded: I was trying to avoid it, as it would require yet another backport to 1.X, but probably makes sense, and it will only be an alias in 1.X

conans/model/requires.py
memsharded: It doesn't matter. As long as you are header-only, your transitive dependencies do not affect, they can be header, static-lib, shared-lib... they will still not affect the header-only artifacts.

PR #10872, opened 2022-03-24
simplifying copy_pattern.py

memsharded:
Changelog: Omit
Docs: Omit

Pure refactor, as recommended by TODOs in code


conan/tools/files/copy_pattern.py
```diff
@@ -66,147 +31,123 @@ def __call__(self, pattern, dst="", src="", keep_path=True, excludes=None, ignor
 
         return: list of copied files
         """
+    assert src != dst
+    assert not pattern.startswith("..")
```
franramirez688, line 35: We're changing the behavior, aren't we?

memsharded: This is a good point. I wanted to have the ``assert`` in place to raise if some test or something was relying on this behavior, and apparently nothing is. But it might happen that tests are ``xfailed``, so I'll try to investigate the test suite and check it.

memsharded: Just checked. It is ok, as this only affects the ``copy()`` new tool, and that shouldn't use ``../`` patterns at all. That was only a workaround when only ``export`` attribute existed. No longer necessary.

It seems that some test might break when merging this to Conan 2.0, but I think we want to block that functionality in 2.0, and force explicit ``copy()`` calls inside the ``export()`` method instead.

Looping-in @lasote for feedback.

franramirez688 (approved):

PR #10884, opened 2022-03-25
[MesonToolchain] Added option `wrap_mode=nofallback`

### LINKED ISSUE #10671: [feature] Meson : set wrap_mode = nofallback by default?
[wrap](https://mesonbuild.com/Wrap-dependency-system-manual.html) downloads some packages from the outside during the build, with I think is not desirable in many cases, for source immutability, and to control dependencies (avoid ODR violation and so on).
right now, I have to add
```
def generate(self):
    tc = MesonToolchain(self)
    tc.project_options["wrap_mode"] = "nofallback"
    ....
    tc.generate()
```

I'd like it if it could be automatic, or maybe have some built-in way to control it.

franramirez688:
Changelog: Feature: Added `wrap_mode=nofallback` project-option into `MesonToolchain` as default value.
Closes: https://github.com/conan-io/conan/issues/10671
Docs: omit

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -96,6 +96,8 @@ class MesonToolchain(object):
         else:
             self._b_vscrt = None
 
+        # default option
+        self.wrap_mode = "nofallback"
         self.project_options = {}
         self.preprocessor_definitions = {}
         self.pkg_config_path = self._conanfile.generators_folder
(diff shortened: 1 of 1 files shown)


conan/tools/meson/toolchain.py
```diff
@@ -189,6 +191,7 @@ def _add_apple_flags(self):
                 self.cpp_link_args += v
 
     def _context(self):
+        self.project_options.setdefault("wrap_mode", self.wrap_mode)
```
memsharded: why? Why not letting the interface be ``project_options``.

You can define in constructor ``self.project_options = {"wrap_mode": "nofallback"}`` and let users modify the dict if they want. This way we will not have to provide a explicit interface for each property

franramirez688: Changed.

memsharded (approved):

PR #10889, opened 2022-03-25
[develop2] Infinite value for requester

franramirez688:
Changelog: Feature: Defined a new `conan_requester.INFINITE_TIMEOUT = -1` variable to manage infinte timeouts via `[conf]`.


memsharded (approved):
Good for me, indeed -1 is sometimes used as infinite timeout. Just hope that ``requests`` will not change its default anytime soon, but I think that will not happen, so good

SSE4:
but won't it cause conan to hang indefinitely in some cases?

memsharded:
That is the default ``requests`` configuration, no reason to forbid it, if users want to configure an infinite timeout, that is their responsibility

PR #10898, opened 2022-03-27
Add support for ``clang`` to ``msvc_runtime_flag()``.

### LINKED ISSUE #10892: [bug] conan.tools.microsoft.msvc_runtime_flag doesn't work for clang-cl (returns an empty value)
see https://github.com/conan-io/conan-center-index/issues/9834

  * Operating System+version: Windows 10 Version 21H2 (OS Build 19044.1586)
  * Compiler+version: clang version 13.0.0
  * Conan version: Conan version 1.47.0-dev
  * Python version: Python 3.9.7

```
git clone https://github.com/conan-io/conan-center-index.git
conan create conan-center-index/recipes/openssl/1.x.x openssl/1.1.1m@ -pr ms_clang_cl -k
```

```
Configuration for profile ms_clang_cl:

[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.runtime=MD
compiler.version=13
os=Windows
[options]
[conf]
[build_requires]
*: ninja/1.10.2
[env]
PATH=['"C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\Llvm\\bin"']
CC=clang-cl
CXX=clang-cl
CONAN_CMAKE_GENERATOR=Ninja
CONAN_CMAKE_MAKE_PROGRRAM=ninja

```

memsharded:
Changelog: Fix: Add support for ``clang`` to ``msvc_runtime_flag()``. It requires defining ``compiler.runtime = static/dynamic`` definition, same as modern ``msvc`` compiler setting.
Docs: https://github.com/conan-io/docs/pull/2485

Close https://github.com/conan-io/conan/issues/10892

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -90,8 +90,13 @@ def msvc_runtime_flag(conanfile):
     if compiler == "Visual Studio":
         return runtime
     if runtime is not None:
+        if runtime == "static":
+            runtime = "MT"
+        elif runtime == "dynamic":
+            runtime = "MD"
+        else:
+            raise ConanException("compiler.runtime should be 'static' or 'dynamic'")
         runtime_type = settings.get_safe("compiler.runtime_type")
-        runtime = "MT" if runtime == "static" else "MD"
         if runtime_type == "Debug":
             runtime = "{}d".format(runtime)
         return runtime
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -89,7 +89,7 @@ def msvc_runtime_flag(conanfile):
     runtime = settings.get_safe("compiler.runtime")
     if compiler == "Visual Studio":
         return runtime
-    if compiler == "msvc" or compiler == "intel-cc":
+    if runtime is not None:
         runtime_type = settings.get_safe("compiler.runtime_type")
         runtime = "MT" if runtime == "static" else "MD"
         if runtime_type == "Debug":
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -7,7 +7,7 @@ from conan.tools.build.cross_building import cross_building, get_cross_building_
 from conan.tools.env import Environment
 from conan.tools.files.files import save_toolchain_args
 from conan.tools.gnu.get_gnu_triplet import _get_gnu_triplet
-from conan.tools.microsoft import VCVars, is_msvc
+from conan.tools.microsoft import VCVars, is_msvc, msvc_runtime_flag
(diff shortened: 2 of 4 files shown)


AndreyMlashkin:
@memsharded when is it going to be officially released?
looking forward for it

conan/tools/microsoft/visual.py
```diff
@@ -89,9 +89,14 @@ def msvc_runtime_flag(conanfile):
...
     if compiler == "Visual Studio":
         return runtime
-    if compiler == "msvc" or compiler == "intel-cc":
+    if runtime is not None:
+        if runtime == "static":
+            runtime = "MT"
+        elif runtime == "dynamic":
+            runtime = "MD"
+        else:
+            raise ConanException("compiler.runtime should be 'static' or 'dynamic'")
```
SpaceIm, line 98: So why do you still allow MD, MT, MTd, MDd in settings.yml for clang? It will break users still using these runtime values. A smooth transition would be appreciated.

SpaceIm: Something like this:

```python
def msvc_runtime_flag(conanfile):
    settings = conanfile.settings
    runtime = settings.get_safe("compiler.runtime")
    if runtime is None:
        return ""
    # TODO: old model, remove in conan v2?
    if runtime in ["MD", "MDd", "MT", "MTd"]:
        return runtime
    # new model
    if runtime == "static":
        runtime = "MT"
    elif runtime == "dynamic":
        runtime = "MD"
    else:
        raise ConanException("compiler.runtime should be 'static' or 'dynamic'")
    runtime_type = settings.get_safe("compiler.runtime_type")
    if runtime_type == "Debug":
        runtime = "{}d".format(runtime)
    return runtime
```

memsharded: It is protected earlier with 
```python
if compiler == "Visual Studio":
        return runtime
```

It will only reach this point for other compilers (including ``msvc``), which should be safe to assume they can only use static/dynamic runtime. But this method shouldn't' allow ``MD/MT...`` definitions to be forward compatible with 2.0

SpaceIm: It's not protected against `clang` which still allows MD, MDd, MTd, MT in setting.yml. It's the point of my comment.

AndreyMlashkin:
> @memsharded when is it going to be officially released? looking forward for it

Is there any roadmap with the dates?

conan/tools/microsoft/visual.py
memsharded: I was thinking forward to 2.0, but ok, lets support legacy settings in this new helper. https://github.com/conan-io/conan/pull/10920 should fix it, please check

Issue #10101, opened 2021-11-29
[feature] cmake presets generator


stefaneicher:
Would it make sense to create a cmake presets generator?
https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

We use the vitualenv generator and in vscode run the activate.sh to setup the environment. But in clion this does not work.
I would like to get a cmake idiomatic solution in which a cmake integration of any ide integrates seamless.
Cmake toolchaain is currently not a option since we have multiple packages in the repo and use workspaces during development that only work with the cmake generator.

memsharded:
Hi @stefaneicher 

The issue is that Conan by default should support cmake 3.15: https://github.com/conan-io/tribe/pull/4

While it could make sense to add later another generator for this functionality, at the moment we should focus on 1 canonical integration. This will be ``CMakeDeps`` and ``CMakeToolchain``, beware that the other generators will be removed in 2.0.
Workspaces are not production ready, we are working on improving the editables as a foundation for future workspaces, you should be checking the ``layout()`` functionality for it too. 

In any case, with the new functionalities it is now easier to create extensions. You can surely create a new block for ``CMakeToolchain`` with the new "blocks" functionality, or a new generator that creates such presets, it would be something like:
- Create a python_requires with a new CMakePresetsGenerator with a single ``generate()`` method
- Use that python_requires in your recipes, and add it to your ``generate()``

KerstinKeller:
@stefaneicher we implemented our own CMake Presets generator internally and it works great! You have a very transparent, identical workflow on all platforms, for all developers.
There is one catch tough: the last time I checked, the CMakePresets.json file needs to be located *next* to the main CMakeLists.txt file. Which means, either you have to generate it in your source directory (which is of course uncool, because generated files should be in a different location).
The other options (which we use), is to *also*  generate a "new" main CMakeLists.txt file, which includes the original top level CMake file.

```
cmake_minimum_required(VERSION 3.20)

project(libA)
# We also generate a few other things
add_subdirectory(D:/libA libA)
```

stefaneicher:
Thanks @KerstinKeller and @memsharded for your answers! In the meantime in the new clion release I can run a script and therefore call the generated [activate.sh](https://docs.conan.io/en/latest/reference/generators/virtualenv.html) script. 

https://www.jetbrains.com/clion/whatsnew/

![image](https://user-images.githubusercontent.com/1171777/149303581-04876025-83af-42ea-a609-987b5a4067f8.png)

[...]

PR #10903, opened 2022-03-28
CMakePresets instead of config file

lasote:
Changelog: Feature: Preliminar support for CMakePresets.json.
Docs: https://github.com/conan-io/docs/pull/2476

- There is a limitation, the `CMakePresets.json` has to be in the same folder as the `CMakeLists.txt` file to be located by `cmake --preset XXX` and the IDEs. But, for the end-user to leverage it,  it can be copied or symlinked and works fine (tested in Visual Studio Code).
- The limitation doesn't exist or interfere with Conan, because Conan is able to locate the file and read the contents previously stored at (conanbuild.conf). 
- This is a legit use of `CMakePresets.json`, it is not needed to call `cmake` with `--preset XXX`, this file is intended to be readable by IDE's (and Conan) to call "cmake" properly passing the right arguments. Because of that, there is no limitation to CMake 3.19.
- Notice that the entry `binaryDir` is indicating the IDEs to use that "build folder" so it could make every IDE follow the Conan build layout (and not the opposite).
- Pending: Look into multi config completing the same CMakePresets.json with different configurations?

Related https://github.com/conan-io/conan/issues/10101

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -87,9 +87,9 @@ class CMake(object):
 
         if not variables:
             variables = {}
-        variables.update(self._cacheVariables)
+        self._cacheVariables.update(variables)
 
-        arg_list.extend(["-D{}={}".format(k, v) for k, v in variables.items()])
+        arg_list.extend(['-D{}="{}"'.format(k, v) for k, v in self._cacheVariables.items()])
         arg_list.append('"{}"'.format(cmakelist_folder))
 
         command = " ".join(arg_list)
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -2,6 +2,7 @@ import os
 import platform
 
 from conan.tools.build import build_jobs
+from conan.tools.cmake.presets import load_cmake_presets
 from conan.tools.cmake.utils import is_multi_configuration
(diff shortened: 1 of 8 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -54,9 +55,10 @@ def __init__(self, conanfile, namespace=None):
         self._conanfile = conanfile
         self._namespace = namespace
 
-        toolchain_file_content = load_toolchain_args(self._conanfile.generators_folder, namespace=self._namespace)
-        self._generator = toolchain_file_content.get("cmake_generator")
-        self._toolchain_file = toolchain_file_content.get("cmake_toolchain_file")
+        cmake_presets = load_cmake_presets(conanfile.generators_folder)
+        self._generator = cmake_presets["configurePresets"][0]["generator"]
+        self._toolchain_file = cmake_presets["configurePresets"][0]["toolchainFile"]
+        self._cacheVariables = cmake_presets["configurePresets"][0]["cacheVariables"]
```
```suggestion
        self._cache_variables = cmake_presets["configurePresets"][0]["cacheVariables"]
```

Using camelCase because of copy-paste from json?

memsharded (approved):

franramirez688 (approved):

czoido (approved):

PR #10917, opened 2022-03-29
find_dependency following cmake_find_mode property

### LINKED ISSUE #10591: [bug] `cmake_find_mode` property is broken
**Abstract description of the problem:** In a conan-generated config/module, decision about `MODULE` or `NO_MODULE` in `find_dependency` call is made by `cmake_find_mode` of dependant package, but not by `cmake_find_mode` of dependency (as it should be). It can be seen in the code [here](https://github.com/conan-io/conan/blob/8c389f5add13f931e06a8335396cce23b65a1d40/conan/tools/cmake/cmakedeps/templates/config.py#L66-L70) (`is_module` is the property (based on `cmake_find_mode`) of the current package (not of the dependency)). 

**Example:** Let we have packages A -> B -> C, with the following `cmake_find_mode` A: "config", B: "config", C: "module". During the build of "A" conan generate `BConfig.cmake` with `find_dependency(C NO_MODULE)` (`NO_MODULE` because `B.cmake_find_mode == "config"`). But this `find_dependency` will fail, because `conan` generate `FindC.cmake` (not `CConfig.cmake`).

**Specific instance of this problem:** In the case of `boost/system` decribed in #8919 `conan` will fail to find `Boost` (from conan-generated `find_dependency(Boost NO_MODULE)`) on systems without system `BoostConfig.cmake` (e.g. [Fedora](https://src.fedoraproject.org/rpms/boost/blob/rawhide/f/boost.spec#_900))

**Suggestion:** It will be great if `conan` consider `cmake_find_mode` of _dependency_ in choosing between `MODULE` and `NO_MODULE` for`find_dependency` call. Or even consider not using `MODULE/NO_MODULE` options in `find_dependecy` at all.

lasote:
Changelog: BugFix: The `find_dependency` called internally in `CMakeDeps` generator to locate transitive dependencies now use the `MODULE/NO_MODULE` following the `cmake_find_mode` property when declared in the dependency.
Docs: 

Close #10591

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_data.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_data.py
@@ -41,6 +41,8 @@ class ConfigDataTemplate(CMakeDepsFileTemplate):
         # This is because in Conan 2.0 model, only the pure tools like CMake will be build_requires
         # for example a framework test won't be a build require but a "test/not public" require.
         dependency_filenames = self._get_dependency_filenames()
+        # Get the nodes that have the property cmake_find_mode=None (no files to generate)
+        dependency_find_modes = self._get_dependencies_find_modes()
         # package_folder might not be defined if Editable and layout()
         package_folder = self.conanfile.package_folder or ""
         package_folder = package_folder.replace('\\', '/').replace('$', '\\$').replace('"', '\\"')
@@ -52,7 +54,8 @@ class ConfigDataTemplate(CMakeDepsFileTemplate):
                 "config_suffix": self.config_suffix,
                 "components_names": components_names,
                 "components_cpp": components_cpp,
-                "dependency_filenames": " ".join(dependency_filenames)}
+                "dependency_filenames": " ".join(dependency_filenames),
+                "dependency_find_modes": dependency_find_modes}
 
     @property
     def template(self):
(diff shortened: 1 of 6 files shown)


memsharded (approved):

czoido (approved):

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -169,6 +175,22 @@ def _get_dependency_filenames(self):
 
         return ret
 
+    def _get_dependencies_find_modes(self):
+        ret = {}
+        if self.conanfile.is_build_context:
+            return ret
+        deps = self.conanfile.dependencies.filter({"build": False, "visible": True, "direct": True})
+        for dep in deps.values():
+            dep_file_name = get_file_name(dep, self.find_module_mode)
```
alex-700, line 184: Hi @lasote! Sorry, have seen PR only now (when it has already merged :)).

Here should be not `self.find_module_mode`, but something which depends on `values[find_mode.lower()]` from 191 line. Like `values[find_mode.values()] == "MODULE"` (or so) (doesn't matter what will be here for "none" find_mode). This is important for dependencies with different `cmake_file_name` and `cmake_module_file_name`.

UPD: Not only here, but in `_get_dependency_filenames` too.

PR #10928, opened 2022-03-30
[conf] Global flags changes

franramirez688:
Changelog: Feature: Renamed `tools.build:cppflags` to `tools.build:defines` and removed `tools.build:ldflags` (now it's the union between `tools.build:sharedlinkflags` and `tools.build:exelinkflags` in mostly all the cases).
Docs: https://github.com/conan-io/docs/pull/2484
Related to: https://github.com/conan-io/conan/pull/10800

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -563,6 +563,9 @@ class ExtraFlagsBlock(Block):
         {% if exelinkflags %}
         string(APPEND CONAN_EXE_LINKER_FLAGS "{% for exelinkflag in exelinkflags %} {{ exelinkflag }}{% endfor %}")
         {% endif %}
+        {% if defines %}
+        add_definitions({% for define in defines %} {{ define }}{% endfor %})
+        {% endif %}
     """)
 
     def context(self):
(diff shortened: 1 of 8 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -563,6 +563,9 @@ class ExtraFlagsBlock(Block):
         {% if exelinkflags %}
         string(APPEND CONAN_EXE_LINKER_FLAGS "{% for exelinkflag in exelinkflags %} {{ exelinkflag }}{% endfor %}")
         {% endif %}
+        {% if defines %}
+        add_definitions({% for define in defines %} {{ define }}{% endfor %})
```
czoido, line 567: Maybe we should use: [add_compile_definitions](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions) instead of [add_definitions](https://cmake.org/cmake/help/latest/command/add_definitions.html)

memsharded: Yes, this needs to be modernized if CMake 3.15 compliant

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
franramirez688: Fine for me but the thing is that we're using `add_definitons` on more sides (blocks/toolchain)... Should we consider changing all of them in another PR? Let me know your opinion.

memsharded: Yes, good for me, leave the ``add_definitions()`` by now, and then later (create an issue) lets modernize all occurences of add_definitions.

lasote (approved):

conan/tools/cmake/toolchain/blocks.py
franramirez688: For the record 😉  -> https://github.com/conan-io/conan/issues/10929

PR #10931, opened 2022-03-30
Fix/cmake find mode

### LINKED ISSUE #10930: Check find_mode potential issue before 1.47
Hi @lasote! Sorry, have seen PR only now (when it has already merged :)).

Here should be not `self.find_module_mode`, but something which depends on `values[find_mode.lower()]` from 191 line. Like `values[find_mode.values()] == "MODULE"` (or so) (doesn't matter what will be here for "none" find_mode). This is important for dependencies with different `cmake_file_name` and `cmake_module_file_name`.

UPD: Not only here, but in `_get_dependency_filenames` too.

_Originally posted by @alex-700 in https://github.com/conan-io/conan/pull/10917#discussion_r838721838_

lasote:
Changelog: omit
Docs: omit

Close #10930

@alex-700 How does it look?

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_data.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_data.py
@@ -1,6 +1,8 @@
 import os
 import textwrap
 
+from conan.tools.cmake.cmakedeps import FIND_MODE_NONE, FIND_MODE_CONFIG, FIND_MODE_MODULE, \
+    FIND_MODE_BOTH
 from conan.tools.cmake.cmakedeps.templates import CMakeDepsFileTemplate
 from conan.tools.cmake.utils import get_file_name, get_find_mode
 """
@@ -175,8 +177,6 @@ class ConfigDataTemplate(CMakeDepsFileTemplate):
         return ret
 
     def _get_dependencies_find_modes(self):
-        from conan.tools.cmake.cmakedeps.cmakedeps import FIND_MODE_NONE, FIND_MODE_CONFIG, \
-            FIND_MODE_MODULE, FIND_MODE_BOTH
         ret = {}
         if self.conanfile.is_build_context:
             return ret
--- a/conan/tools/cmake/cmakedeps/templates/target_data.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_data.py
@@ -2,7 +2,7 @@ import os
 import textwrap
 
 from conan.tools.cmake.cmakedeps.templates import CMakeDepsFileTemplate
-from conan.tools.cmake.utils import get_file_name
+from conan.tools.cmake.utils import get_file_name, get_find_mode
 
 """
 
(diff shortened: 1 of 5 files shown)


conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -181,13 +181,12 @@ def _get_dependencies_find_modes(self):
...
-            dep_file_name = get_file_name(dep, self.find_module_mode)
-            find_mode = dep.cpp_info.get_property("cmake_find_mode")
+            dep_file_name = get_file_name(dep)
+            find_mode = get_find_mode(dep)
             values = {"none": "",
                       "config": "NO_MODULE",
                       "module": "MODULE",
                       "both": "NO_MODULE" if not self.find_module_mode else "MODULE"}
-            find_mode = find_mode or "both"
             ret[dep_file_name] = values[find_mode.lower()]
```
alex-700: nit:
```suggestion
            ret[dep_file_name] = values[find_mode]
```

conan/tools/cmake/utils.py
```diff
@@ -4,14 +4,22 @@ def is_multi_configuration(generator):
...
     ret = conanfile.cpp_info.get_property("cmake_file_name")
     return ret or conanfile.ref.name
 
+
+def get_find_mode(conanfile):
+    """
+    :param conanfile: conanfile of the requirement
+    :return: "none" or "config" or "module" or "both"
+    """
+    tmp = conanfile.cpp_info.get_property("cmake_find_mode") or "both"
```
alex-700: Documentation [here](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html) says, that `cmake_find_mode` defaulted to "config". So should it be `or "config"`?

conan/tools/cmake/utils.py
```diff
@@ -4,14 +4,22 @@ def is_multi_configuration(generator):
     return "Visual" in generator or "Xcode" in generator or "Multi-Config" in generator
 
 
-def get_file_name(conanfile, find_module_mode=False):
+def get_file_name(conanfile, forced_module_mode=None):
     """Get the name of the file for the find_package(XXX)"""
     # This is used by the CMakeToolchain to adjust the XXX_DIR variables and the CMakeDeps. Both
     # to know the file name that will have the XXX-config.cmake files.
-    if find_module_mode:
+    if forced_module_mode or get_find_mode(conanfile) == "module":
```
alex-700: Should `FIND_MODE_MODULE` constant from [here](https://github.com/conan-io/conan/blob/develop/conan/tools/cmake/cmakedeps/cmakedeps.py#L14) be used? Should other hardcoded strings in this file be fixed too?

alex-700 (changes requested):

conan/tools/cmake/utils.py
lasote: Mhh not in this case, but yes, this is a bit messy so I will return None when not set.

conan/tools/cmake/utils.py
```diff
@@ -21,5 +24,7 @@ def get_find_mode(conanfile):
     :param conanfile: conanfile of the requirement
     :return: "none" or "config" or "module" or "both"
```
alex-700: doccomment should be updated for new return value None.

alex-700 (approved):
LGTM!
Together with the previous PR, this is a great fix!

conan/tools/cmake/utils.py
alex-700: Thanks for including comment in the "convertation to `{{ _DEPENDENCY }} _FIND_MODE`": now I understand this!

memsharded (approved):

PR #10948, opened 2022-04-03
[develop2] minor refactor to Settings, ANY

memsharded:
- The "ANY" needs to be a list element, not unique value
- Removed dead ``_headers_for_info``, that send settings header to server. We might want to recover this, but atm it is dead code.
- Remove ``apple_clang_compatible`` legacy from 1.X
- Remove unused ``load_from_package()`` method

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -40,12 +40,12 @@ os:
             catalyst:
                 ios_version: *ios_version
     Android:
-        api_level: ANY
+        api_level: [ANY]
     FreeBSD:
     SunOS:
     AIX:
     Arduino:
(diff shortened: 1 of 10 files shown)


conans/client/conf/__init__.py
```diff
@@ -40,12 +40,12 @@
...
                 ios_version: *ios_version
     Android:
-        api_level: ANY
+        api_level: [ANY]
     FreeBSD:
     SunOS:
     AIX:
     Arduino:
-        board: ANY
+        board: [ANY]
```
memsharded, line 48: TODO: reconsider what to do with Arduino, maybe move to inside "baremetal"

PR #10951, opened 2022-04-03
Use class Settings for ``ConanInfo`` instead of ``Values``. Fully dropping ``Values`` class

memsharded:
- ``Values`` class is fully removed
- Instead a ``Settings`` object is used everywhere.
- The ``ConanInfo`` reads a plain python list for serializing back to client conaninfo.txt for ``conan search binaries``. It might be replaced by the new serialization of the full conaninfo.txt text content
- ``compatibility()`` computations are done with a ``Settings`` object (that allows ANY) instead of a ``Values`` object too.


conans/model/settings.py
```diff
@@ -58,6 +57,18 @@ def copy(self):
             result._definition = {k: v.copy() for k, v in self._definition.items()}
         return result
 
+    def copy_conaninfo_settings(self):
+        """ deepcopy, recursive
+        """
+        result = SettingsItem({}, name=self._name)
+        result._value = self._value
+        if not isinstance(self._definition, dict):
+            result._definition = self._definition[:] + ["ANY"]
```
lasote, line 77: The `["ANY"]` is because you can manipulate a value in the `info` with anything, right? like `self.info.settings.compiler="potato"`

lasote (approved):
Much more simple!

conans/model/info.py
```diff
@@ -380,7 +379,17 @@ def loads(text):
...
+        def _loads_settings(settings_text):
+            settings_result = []
+            for line in settings_text.splitlines():
+                if not line.strip():
+                    continue
+                name, value = line.split("=", 1)
+                settings_result.append((name.strip(), value.strip()))
+            return settings_result
+
+        result.settings = _loads_settings(parser.settings)
```
franramirez688, line 395: Shouldn't it be better to have it on `Settings` side? I guess that you cannot use the current `Settings.loads()` but I don't get to understand why.

memsharded: That is a very good question. This is not an actual Settings object. This is a plain list of tuples (string elems) that needs to be loaded from conaninfo.txt to be serialized to json from the server to the client. It is not a full fledged Settings object, and it cannot be one, because for that, it should get the settings.yml file, and that only exists in the client side, whiles this operation happens in the server side too (when a client is doing a search or list of binaries in the server)

This is why I kept in this file. It is about reading conaninfo.txt and serializing it for client consumption (the client side also don't deserialize the settings at all to a ``Settings`` object, it keeps the plain python object.

conans/model/settings.py
memsharded: Yes, exactly. But with a caveat, you cannot manipulate to **random** strings things that contain subsettings. Because that would leave the thing in a undefined state, with some now inconsistent subsettings, that cannot be accessed anymore. So with this change the options are:

- If you need more "binary-compatible" descriptions of a compiler, lets say like "gcc_or_clang", then you need to add that string to ``settings.yml``. And add the subsettings that you want for it.
- Settings that are "final" (lists), like ``build_type``, or ``arch`` or ``compiler.version`` they can get any value without issues.

lasote: Better to document that at some point... 💊

memsharded: Totally agree, edited the above and added to the docstring.

conans/model/info.py
franramirez688: Ok, I got it. Thanks for the explanation! 👍

franramirez688 (approved):

PR #10952, opened 2022-04-03
[poc] try serializing the whole conaninfo.txt without parsing it

memsharded:
Changelog: omit
Docs: omit

This might help with the serialization issue of the ``conaninfo.txt`` that happens from server->client when doing a ``conan search <pkg-ref>`` to list binary configurations. In 2.0, the format changes, new fields might be added. The serialization is baked in servers like Artifactory and will break in 2.0 for this case.

Open questions:
- Robustness. Can something break? The payload of the Rest API is json.

### DIFF
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -1,4 +1,3 @@
-import base64
 import os
 import shutil
 import time
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -1,3 +1,4 @@
+import base64
 import os
 import shutil
 import time
@@ -9,6 +10,7 @@ from conans import DEFAULT_REVISION_V1
 from conans.client.cache.remote_registry import Remote
 from conans.errors import ConanConnectionError, ConanException, NotFoundException, \
     NoRestV2Available, PackageNotFoundException
+from conans.model.info import ConanInfo
 from conans.paths import EXPORT_SOURCES_TGZ_NAME, EXPORT_TGZ_NAME, PACKAGE_TGZ_NAME, rm_conandir
 from conans.search.search import filter_packages
 from conans.util import progress_bar
@@ -226,6 +228,10 @@ class RemoteManager(object):
 
     def search_packages(self, remote, ref, query):
         packages = self._call_remote(remote, "search_packages", ref, query)
+        # Avoid serializing conaninfo in server side
+        packages = {pid: ConanInfo.loads(base64.b64decode(data["content"].encode("utf-8")).decode("utf-8")).serialize_min()
+                    if "content" in data else data
+                    for pid, data in packages.items()}
         # Filter packages without recipe_hash, those are 1.X packages, the 2.0 are disregarded
         packages = {pid: data for pid, data in packages.items() if data.get("recipe_hash")}
         packages = filter_packages(query, packages)
--- a/conans/client/remote_manager.py
+++ b/conans/client/remote_manager.py
@@ -227,9 +227,9 @@ class RemoteManager(object):
         return self._call_remote(remote, "search", pattern, ignorecase)
 
     def search_packages(self, remote, ref, query):
-        packages = self._call_remote(remote, "search_packages", ref, query)
+        packages = self._call_remote(remote, "search_packages", ref)
         # Avoid serializing conaninfo in server side
-        packages = {pid: ConanInfo.loads(base64.b64decode(data["content"].encode("utf-8")).decode("utf-8")).serialize_min()
+        packages = {pid: ConanInfo.loads(data["content"]).serialize_min()
                     if "content" in data else data
                     for pid, data in packages.items()}
         # Filter packages without recipe_hash, those are 1.X packages, the 2.0 are disregarded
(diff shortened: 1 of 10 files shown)


memsharded:
Notes:
We will try doing a full structure serialization of the conaninfo file:

- It would be great if it maps directly to package_id, directly hashing the file. But not strictly necessary if not possible.
- Json order of dicts can be tricky to guarantee always same order (and same conaninfo serialization => same package_id). Maybe yaml?

lasote:
I'll investigate this and will suggest something. I would say with `yml` we can have the same issue of the order in dicts.

lasote:
Proposed PR: https://github.com/memsharded/conan/pull/44

conans/server/service/common/search.py
```diff
@@ -61,7 +60,8 @@ def search_packages(server_store, ref, query, look_in_all_rrevs):
     if not os.path.exists(server_store.conan_revisions_root(ref.copy_clear_rev())):
         raise RecipeNotFoundException(ref)
     infos = _get_local_infos_min(server_store, ref, look_in_all_rrevs)
-    return filter_packages(query, infos)
+    assert query is None, "The server is not filtering packages remotely anymore"
```
lasote, line 70:
```suggestion
    assert query is None, "The server is not filtering packages remotely anymore"
```
As suggested we can remove this in case there is an ancient conan client.

conans/client/remote_manager.py
```diff
@@ -1,3 +1,4 @@
+import base64
```
memsharded:
```suggestion
```

conans/server/service/common/search.py
```diff
@@ -1,10 +1,10 @@
+import base64
```
memsharded:
```suggestion
```

conans/server/service/common/search.py
memsharded: It is unlikely users will upgrade just the server without uploading the client, so it could be fine

memsharded (commented):
Probably it would be great to add a couple of tests with properties (options with values, for example, or settings), that contain a bunch of complicated characters that the json encoding should handle correctly.

lasote:
I added a test with ugly chars

conans/test/integration/remote/test_conaninfo_parsing.py
```diff
@@ -0,0 +1,32 @@
...
+
+    t = TestClient(default_server_user=True)
+    conanfile = textwrap.dedent("""
+    # coding=utf-8
+    from conan import ConanFile
+
+    class Recipe(ConanFile):
+        name = "weird_info"
+        version = "1.0"
+        options = {"ññ¨¨&是": ['是"是', "是是是"]}
```
memsharded: This is good, but maybe better to add more characters to be tested like {}, [], single and double quotes, multi-line values, etc.?

conans/server/service/common/search.py
```diff
@@ -33,10 +32,8 @@ def _get_local_infos_min(server_store, ref, look_in_all_rrevs):
                 info_path = os.path.join(server_store.package(pref), CONANINFO)
                 if not os.path.exists(info_path):
                     raise NotFoundException("")
-                conan_info_content = load(info_path)
-                info = ConanInfo.loads(conan_info_content)
-                conan_vars_info = info.serialize_min()
-                result[package_id] = conan_vars_info
+                content = load(info_path)
+                result[package_id] = {"content": content}
```
memsharded, line 39: I am rethinking this, maybe we need to keep the legacy "settings" and other fields if possible, it looks like that Artifactory will not be able to break this way, so maybe try to be consistent.

conans/test/integration/remote/test_conaninfo_parsing.py
lasote: feel free to modify your PR ;)

conans/server/service/common/search.py
lasote: I don't get what you mean

conans/test/integration/remote/test_conaninfo_parsing.py
memsharded: Oh, yes, that is my PR, I didn't realize 😅 . I'll try to add more tests then

conans/server/service/common/search.py
memsharded: Artifactory cannot return only the ``{"content": `` field immediately, because it will break all the older Conan clients, and migrations will take a while. So Artifactory will need to return both ``"content"`` key, plus the legacy "settings", "options", etc, to avoid breaking, even if in some cases, like when uploading a new modern 2.0 conaninfo.txt, it doesn't get those fields, and they are blank, but they should still be returned to the client side with ``{"settings": ..., "options": ...., "content": ....}`` to avoid old clients crashing.

conans/test/integration/remote/test_conaninfo_parsing.py
lasote: added

lasote:
Mostly as an example to be implemented in Artifactory, I kept (wth a FIXME) in the conan_server the old contents return.
We can remove it at some point in the future.

memsharded (commented):
LGTM

PR #10960, opened 2022-04-04
json_generator with layout() method accessing dependency version field

### LINKED ISSUE #10956: [bug] version is not set correctly when using layout
When layout is being used, recipe version is not set correctly somehow using json generator, it seems that version is not being fetched from package metadata when running conan install command!

### Environment Details
  * Operating System+version: macos
  * Compiler+version:  apple-clang 12.0
  * Conan version: Conan version 1.47.0
  * Python version: 3.9

### Steps to reproduce 
* create a conan demo project using `conan new demo/1.0.0 --template=cmake_lib` 
* create a local conan package `conan create .`
* generate deps using json generator `conan install demo/1.0.0@ -g json`
* inspect conanbuildinfo.json, version is set to null, however it should be 1.0.0

* remove the layout method from the conanfile.py and try again
* now version is set correctly 

btw, it seems to be the same issue for the description attribute, maybe other attributes as well

Screen Shot 2022-04-04 at 12 29 24 PM

lasote:
Changelog: Bugfix: The `json` generator was showing "None" in the `version` field of the dependencies when the `layout()` method was used.
Docs: omit

Close #10956 

To discuss if we consider this should be fixed given that the "layout()" feature is a 2.0 feature and json generator doesn't exist anymore.
Pure refactor plus adding these dirty lines:

```
if cpp_info.version is None:
    try:
        res["version"] = self.conanfile.dependencies.get(depname).ref.version
    except Exception:
        pass
```

### DIFF
--- a/conans/client/generators/json_generator.py
+++ b/conans/client/generators/json_generator.py
@@ -3,26 +3,6 @@ import json
 from conans.model import Generator
 
 
-def serialize_cpp_info(cpp_info):
-    keys = [
-        "version",
-        "description",
-        "rootpath",
-        "sysroot",
-        "include_paths", "lib_paths", "bin_paths", "build_paths", "res_paths",
-        "libs",
-        "system_libs",
-        "defines", "cflags", "cxxflags", "sharedlinkflags", "exelinkflags",
-        "frameworks", "framework_paths", "names", "filenames",
-        "build_modules", "build_modules_paths"
-    ]
-    res = {}
-    for key in keys:
-        res[key] = getattr(cpp_info, key)
-    res["cppflags"] = cpp_info.cxxflags  # Backwards compatibility
-    return res
-
-
 def serialize_user_info(user_info):
     res = {}
     for key, value in user_info.items():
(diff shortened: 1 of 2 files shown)


conans/test/integration/generators/json_test.py
```diff
@@ -26,7 +29,8 @@ def package_info(self):
         client.run("install Hello/0.1@lasote/testing -g json")
         conan_json = client.load("conanbuildinfo.json")
         data = json.loads(conan_json)
-
+        self.assertEqual(data["dependencies"][0]["version"], "0.1")
+        self.assertIsNone(data["dependencies"][0]["description"])
```
memsharded, line 33: Description keep being None even if defined?

conans/client/generators/json_generator.py
```diff
@@ -71,3 +51,31 @@ def get_options(self):
...
+            "build_modules", "build_modules_paths"
+        ]
+        res = {}
+        for key in keys:
+            res[key] = getattr(cpp_info, key)
+        res["cppflags"] = cpp_info.cxxflags  # Backwards compatibility
+        res["name"] = depname
+
+        # FIXME: trick for NewCppInfo objects when declared layout
+        if cpp_info.version is None:
```
memsharded: Maybe put the ``cpp_info.version`` inside the try-except, just in case ``version`` is not defined for NewCppInfo?

memsharded (approved):
Relatively low risk, I think it can be added, now the research and PR is done.
(Otherwise, this could have been left without fixing)

conans/test/integration/generators/json_test.py
lasote: Accessing the description is even more hacky, the ConanfileInterface doesn't contain that field (and I'm not sure why it should/and not other) that's why I left it (doesn't feel so critical as the version field)

boussaffawalid: I've tested this fix and it works for the version field, description still null but that is minor. Thanks for the quick fix

boussaffawalid (approved):

Issue #10963, opened 2022-04-05
[bug] Improve mingw support


AndreyMlashkin:
### Environment Details (include every applicable attribute)
  * Operating System+version: Wingows 10
  * Compiler+version: gcc version 11.2.0 (MinGW-W64 x86_64-posix-seh, built by Brecht Sanders)
  * Conan version: 1.47.0

### Steps to reproduce (Include if Applicable)
install mingw using choco:
choco install mingw -y

add new profile mingw:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.exception=seh
compiler.libcxx=libstdc++11
compiler.threads=win32
compiler.version=11.2
cppstd=17
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
CONAN_CMAKE_GENERATOR=Ninja
CC=gcc
CXX=g++

create new repository from examples:
conan new hello/0.1 -m=cmake_lib

try to build it:
conan create . --build missing -pr mingw

The build is ok, but the example executable can't be executed

AndreyMlashkin:
[mingw_log.txt](https://github.com/conan-io/conan/files/8416165/mingw_log.txt)

lasote:
Hi, looks like the error `3221225781` implies a missing DLL. Given that you are building a static one from the "hello" library, I would say there is something related to the MinGW installation. Would be great to know which DLL is not found so we can try to figure out if we can fix something in Conan. 
You can try to run `dumpbin /imports C:\Users\Administrator\sources\conan_test\test_package\cmake-build-release\conan\example.exe` or use any of this tools to try to locate it: https://www.raymond.cc/blog/check-what-dll-or-ocx-dependency-files-is-needed-for-a-software/

AndreyMlashkin:
As I said, I am using MinGW from choco.

And here is the output:
dumpbin /imports C:\Users\Administrator\sources\conan_test\test_package\cmake-build-release\conan\example.exe
Microsoft (R) COFF/PE Dumper Version 14.29.30141.0
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file C:\Users\Administrator\sources\conan_test\test_package\cmake-build-release\conan\example.exe
LINK : fatal error LNK1181: cannot open input file 'C:\Users\Administrator\sources\conan_test\test_package\cmake-build-release\conan\example.exe'

lasote:
[...]

PR #10968, opened 2022-04-05
Test mingw with cmake_lib

lasote:
Changelog: omit
Docs: omit

trying to debug  #10963
The issue was solved and not related to Conan but we could merge this as a simple example of cmake_lib and mingw.

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -22,24 +22,20 @@ def test_simple_cmake_mingw():
     client = TestClient()
     client.run("new hello/1.0 -m cmake_lib")
     client.save({"mingw": """
-
-    [settings]
-    os=Windows
-    arch=x86_64
-    build_type=Release
-    compiler=gcc
-    compiler.exception=seh
-    compiler.libcxx=libstdc++11
-    compiler.threads=win32
-    compiler.version=11.2
-    cppstd=17
-    [options]
-    [build_requires]
-    [env]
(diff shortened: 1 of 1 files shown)


memsharded (approved):

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -15,6 +15,33 @@
...
+    compiler.threads=win32
+    compiler.version=11.2
+    cppstd=17
+    [options]
+    [build_requires]
+    [env]
+    CC=gcc
+    CXX=g++
+    """})
+    client.run("create . --profile=mingw")
```
memsharded, line 36: The template include output of compiler version (which is now not well defined not checked, the compiler in CI might not be 11.2, and it won't be noticed). But checking som eof the binary output like ``__GNUC__`` and that it doesn't contain any trace of Visual Studio, would be good

memsharded (approved):

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -15,6 +15,33 @@
...
+    compiler=gcc
+    compiler.exception=seh
+    compiler.libcxx=libstdc++11
+    compiler.threads=win32
+    compiler.version=11.2
+    cppstd=17
+    [options]
+    [build_requires]
+    [env]
+    CC=gcc
```
memsharded: Also these env-vars are no longer needed, can be removed

lasote (commented):
good!

PR #10972, opened 2022-04-05
[MSBuildToolchain][conf] Adding extra flags to `MSBuildToolchain`

### LINKED ISSUE #10885: [feature] Add extra flags via `[conf]` into `MSBuildToolchain`
Related to https://github.com/conan-io/conan/pull/10800

Add the same mechanism into `MSBuildToolchain`

franramirez688:
Changelog: Feature: Added `cxxflags`, `cflags`, and `ldflags` attributes to `MSBuildToolchain`.
Changelog: Feature: Added mechanism to inject extra flags to `MSBuildToolchain` via `[conf]`.
Closes: https://github.com/conan-io/conan/issues/10885
Docs: https://github.com/conan-io/docs/pull/2507

### DIFF
--- a/conan/tools/microsoft/toolchain.py
+++ b/conan/tools/microsoft/toolchain.py
@@ -42,8 +42,6 @@ class MSBuildToolchain(object):
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
-        self.compiler_flags = []
-        self.linker_flags = []
         self.preprocessor_definitions = {}
         self.compile_options = {}
         self.configuration = conanfile.settings.build_type
@@ -65,6 +63,7 @@ class MSBuildToolchain(object):
     def generate(self):
         name, condition = self._name_condition(self._conanfile.settings)
         config_filename = "conantoolchain{}.props".format(name)
+        # Writing the props files
         self._write_config_toolchain(config_filename)
         self._write_main_toolchain(config_filename, condition)
         if self._conanfile.settings.get_safe("compiler") == "intel-cc":
@@ -127,12 +126,15 @@ class MSBuildToolchain(object):
         def format_macro(key, value):
             return '%s=%s' % (key, value) if value is not None else key
 
+        # FIXME: self.preprocessor_definitions is a dict but tools.build:defines is a list
         preprocessor_definitions = "".join(["%s;" % format_macro(k, v)
                                             for k, v in self.preprocessor_definitions.items()])
         extra_flags = self._get_extra_flags()
-        defines = preprocessor_definitions + "".join("%s;" % d for d in extra_flags["defines"])
-        compiler_flags = self.compiler_flags + extra_flags["compiler_flags"]
-        linker_flags = self.linker_flags + extra_flags["linker_flags"]
+        defines = preprocessor_definitions + "".join("%s;" % d for d in extra_flags["preprocessor_definitions"])
+        cxxflags = extra_flags["cxxflags"]
+        cflags = extra_flags["cxxflags"]
+        sharedlinkflags = extra_flags["sharedlinkflags"]
+        exelinkflags = extra_flags["exelinkflags"]
         cppstd = "stdcpp%s" % self.cppstd if self.cppstd else ""
         runtime_library = self.runtime_library
         toolset = self.toolset
@@ -149,8 +151,8 @@ class MSBuildToolchain(object):
                                   for k, v in self.compile_options.items())
         return {
             'defines': defines,
-            'compiler_flags': " ".join(compiler_flags),
-            'linker_flags': " ".join(linker_flags),
+            'compiler_flags': " ".join(cxxflags + cflags),
+            'linker_flags': " ".join(sharedlinkflags + exelinkflags),
             "cppstd": cppstd,
             "runtime_library": runtime_library,
             "toolset": toolset,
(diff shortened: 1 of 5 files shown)


conan/tools/microsoft/toolchain.py
```diff
@@ -187,3 +207,16 @@ def _write_main_toolchain(self, config_filename, condition):
...
+        # Now, it's time to get all the flags defined by the user
+        cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list)
+        cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list)
+        sharedlinkflags = self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)
+        exelinkflags = self._conanfile.conf.get("tools.build:exelinkflags", default=[], check_type=list)
+        defines = self._conanfile.conf.get("tools.build:defines", default=[], check_type=list)
+        return {
+            "compiler_flags": cflags + cxxflags,
+            "defines": defines,
+            "linker_flags": sharedlinkflags + exelinkflags,
```
SSE4: is it correct to mix shared and exe flags together?
I think the idea is to apply shared flags to .DLL targets only, and exe flags to .EXE targets only.

conan/tools/microsoft/toolchain.py
```diff
@@ -187,3 +207,16 @@ def _write_main_toolchain(self, config_filename, condition):
...
+
+    def _get_extra_flags(self):
+        # Now, it's time to get all the flags defined by the user
+        cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list)
+        cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list)
+        sharedlinkflags = self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)
+        exelinkflags = self._conanfile.conf.get("tools.build:exelinkflags", default=[], check_type=list)
+        defines = self._conanfile.conf.get("tools.build:defines", default=[], check_type=list)
+        return {
+            "compiler_flags": cflags + cxxflags,
```
SSE4: that also probably needs some sort of check/filter. it might be useless (if not breaking) to apply C++ specific flags to pure C targets.

franramirez688: `MSBuildDeps` is already doing this sum, I only copied and paste the same logic.

conan/tools/microsoft/toolchain.py
franramirez688: Yeah, I've to figure it out. `MSBuildDeps` was not even using the `exelinkflags`...

conan/tools/microsoft/toolchain.py
SSE4: well, it might be a serious bug, which we missed because `MSBuildDeps` is stiill not used so widely (in CCI at least).
if it's too much to fix in this PR, let's create an another issue then :)

memsharded (commented):
Keep flags separated. Both for linkflags, but also for cflags and cxxflags.

We will use some "language  = C or C++" property in recipes in the future to apply this kind of logic.

Keep cflags and cxxflags separated until the last minute, and aggregate them only in the template. User should be able to say ``toolchain.cflags = None`` (or something similar, not accurate), and that would get rid of the cflags, leaving the cxxflags.

franramirez688:
I have added:

```python
class MSBuildToolchain(object):
        # ....
        self.cxxflags = []
        self.cflags = []
        self.ldflags = []
```
I don't see the point of splitting `sharedlinkflags` and `exelinkflags` because there is no specific option in props to put each one. It's true that in `extra_flags` we have both of them separated.
Props file is being built in `generate()` method so users will be able to change whatever they want regarding the flags values before the file will be created.

conan/tools/microsoft/toolchain.py
franramirez688: I cannot see anything about this. They are read as `AdditionalOptions` in `<Link>` section and that's it.

conan/tools/microsoft/toolchain.py
franramirez688: Here is the same. I'm not able to get any clue about how to split both flags into different sections. Keeping it as it is.

conan/tools/microsoft/toolchain.py
```diff
@@ -14,10 +16,37 @@ class MSBuildToolchain(object):
 
     filename = "conantoolchain.props"
 
+    _config_toolchain_props = textwrap.dedent("""\
```
memsharded, line 19: I don't disagree with this refactor (making it a class attribute, instead of an internal method implementation detail), but it makes challenging to see what is changing in the template.

conan/tools/microsoft/toolchain.py
```diff
@@ -94,35 +124,19 @@ def _runtime_library(settings):
...
-                </ResourceCompile>
-              </ItemDefinitionGroup>
-              <PropertyGroup Label="Configuration">
-                <PlatformToolset>{}</PlatformToolset>
-              </PropertyGroup>
-            </Project>
-            """)
-        preprocessor_definitions = ";".join([format_macro(k, v)
-                                             for k, v in self.preprocessor_definitions.items()])
+        extra_flags = self._get_extra_flags()
```
memsharded: If you are not using the return thing ``_get_extra_flags()`` as an actual dictionary (to update or something, just getting the dict members), better make the method return a tuple of things.

memsharded (approved):
Looking good!

lasote (approved):
Add a functional test injecting some of these [conf] and see if everything builds (and respond to the flags if possible)

franramirez688:
@lasote I just added a new functional one 😉

conans/test/functional/toolchains/microsoft/test_msbuildtoolchain.py
```diff
@@ -0,0 +1,50 @@
...
+    tools.build:defines+=["DEF1", "DEF2"]
+    """)
+    client = TestClient(path_with_spaces=False)
+    client.run("new hello/0.1 --template=msbuild_exe")
+    client.save({
+        "myprofile": profile
+    })
+    # Let's import manually the created conantoolchain_release_x64.props
+    replace_in_file(MagicMock(), os.path.join(client.current_folder, "hello.vcxproj"),
+                    r'  <ImportGroup Label="PropertySheets">', toolchain_props)
```
franramirez688, line 45: I just opened an issue related to this replacement: https://github.com/conan-io/conan/issues/11024

lasote (approved):

PR #10978, opened 2022-04-05
Autotools templates for exe and lib

### LINKED ISSUE #10787: [feature] Create "new --template" for Autotools.
Like https://github.com/conan-io/conan/pull/10760/files  but for Autotools.
Tests using it, even with windows subsystems.
Better study if a predefined layout is convenient.

czoido:
Changelog: Feature: New templates for autotools exe and lib.
Changelog: Feature: Change `build_script_folder` argument from the configure to the `Autotools` build helper constructor.
Docs: https://github.com/conan-io/docs/pull/2511

Close #10787

TODO:

- [x] Local flow: this involves some changes on how the --prefix, etc.. configure args are set, still wip.
- [x] Install libtool Mac node
- [X] Static/shared


conan/tools/gnu/layout.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+def autotools_layout(conanfile):
```
memsharded: Isn't the ``basic_layout()`` enough?

conans/assets/templates/new_v2_autotools.py
```diff
@@ -0,0 +1,360 @@
...
+
+        def package(self):
+            autotools = Autotools(self)
+            autotools.install()
+
+        def package_info(self):
+            self.cpp_info.libs = ["{name}"]
+    """)
+
+source_h = textwrap.dedent("""
```
memsharded: This files can be reused from other template maybe? At least in 2.0 I just import them from cmake template

memsharded (approved):
Looking good

conans/assets/templates/new_v2_autotools.py
czoido: Yes, I'm making a second pass today and refactoring everything 👍

conans/assets/templates/new_v2_autotools.py
```diff
@@ -0,0 +1,241 @@
...
+        def layout(self):
+            basic_layout(self)
+
+        def generate(self):
+            at_toolchain = AutotoolsToolchain(self)
+            if self.options.shared:
+                at_toolchain.configure_args.extend(["--enable-shared", "--disable-static"])
+            else:
+                at_toolchain.configure_args.extend(["--disable-shared", "--enable-static",
+                                                    "--with-pic" if self.options.get_safe("fPIC", True) else "--without-pic"])
```
czoido: This is a common pattern in Conan Center recipes and probably we should add built-in support for this in the toolchain, leaving this external here for the moment

conan/tools/gnu/autotools.py
```diff
@@ -21,6 +22,24 @@ def configure(self, build_script_folder=None):
         http://jingfenghanmax.blogspot.com.es/2010/09/configure-with-host-target-and-build.html
         https://gcc.gnu.org/onlinedocs/gccint/Configure-Terms.html
         """
+        configure_args = []
```
czoido, line 29: I had to move setting the --prefix, etc... here so that the local flow works...
The build helper will only set the arguments related with the output folders... and the other arguments come from the toolchain. Just for testing this if this makes sense or not

conans/assets/templates/new_v2_autotools.py
czoido: I have moved this to the toolchain

conans/assets/templates/new_v2_autotools.py
```diff
@@ -0,0 +1,238 @@
...
+        def build(self):
+            with chdir(self, self.source_folder):
+                self.run("autoreconf --force --install")
+            autotools = Autotools(self)
+            autotools.configure(self.source_folder)
+            autotools.make()
+
+        def package(self):
+            autotools = Autotools(self)
+            autotools.configure(self.source_folder)
```
memsharded: This extra ``configure()`` shouldn't be necessary here, the project has been configured and built before (and it can be assumed that it has been configured and built)

conans/assets/templates/new_v2_autotools.py
```diff
@@ -0,0 +1,238 @@
...
+        def layout(self):
+            basic_layout(self)
+
+        def generate(self):
+            at_toolchain = AutotoolsToolchain(self)
+            at_toolchain.generate()
+
+        def build(self):
+            with chdir(self, self.source_folder):
+                self.run("autoreconf --force --install")
```
memsharded: Is this a common idiom? Do we want to make an ``autotools.autoreconf()`` method?

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -166,26 +167,25 @@ def generate(self, env=None, scope="build"):
         self.generate_args()
         VCVars(self._conanfile).generate(scope=scope)
 
+    def _shared_static_args(self):
```
czoido, line 169: recurrent pattern in Conan Center

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -166,26 +167,25 @@ def generate(self, env=None, scope="build"):
...
+        args = []
+        try:
+            shared = self._conanfile.options.shared
+            if shared:
+               args.extend(["--enable-shared", "--disable-static"])
+            else:
+               args.extend(["--disable-shared", "--enable-static"])
+
+            if shared:
+                args.extend(["--with-pic" if self._conanfile.options.fPIC else "--without-pic"])
```
memsharded: But fpic isn't mandatory for shared libs? Also, if this toolchain might be used by Windows/mingw builds, is it possible that this fail?

conan/tools/gnu/autotools.py
memsharded: Not sure about this, lets see what @lasote thinks

conans/assets/templates/new_v2_autotools.py
czoido: Yes, I agree. I had to put it there because of the local flow, I was still checking how we can do so that the autotools sets the package_folder in a previous steps...

conan/tools/gnu/autotools.py
lasote: I would need to know why the local flow is not working. If you want to call the build system directly, you have to look at the `conanbuild.conf` to see which arguments to apply according to the calculated by the toolchain, right? But the `conan build` would work in any case. I'm missing something.

memsharded: Lets talk about this, the rest of the PR is looking ready for merge

conan/tools/gnu/autotools.py
```diff
@@ -15,17 +17,39 @@ def __init__(self, conanfile, namespace=None):
                                                      namespace=namespace)
         self._configure_args = toolchain_file_content.get("configure_args")
         self._make_args = toolchain_file_content.get("make_args")
+        self.default_configure_install_args = True
+        self.build_script_folder = None
```
memsharded: This usage of ``build_script_folder`` as member is weird, it is passed as argument in the method, and only used inside that single method, it seems the member is unnecessary.

conan/tools/gnu/autotools.py
```diff
@@ -15,17 +17,39 @@ def __init__(self, conanfile, namespace=None):
...
 
-        configure_cmd = "{}/configure".format(source)
+        self.build_script_folder = self.build_script_folder or os.path.join(configure_folder)
+        configure_args = []
+        if self.default_configure_install_args and self._conanfile.package_folder:
+            def _get_cpp_info_value(name):
+                # Why not taking cpp.build? because this variables are used by the "cmake install"
+                # that correspond to the package folder (even if the root is the build directory)
+                elements = getattr(self._conanfile.cpp.package, name)
+                return elements[0] if elements else None
```
memsharded: Isn't this defining ``--bindir=${prefix}/None`` for some cases? If not, then it should be an assert?

conan/tools/gnu/autotools.py
czoido: Yes, it's weird, I did it because you need it for the configure inside the install and I wanted to avoid passing that in the constructor or in the install method, probably better to pass it in the constructor...

conan/tools/gnu/autotools.py
czoido: I can avoid that omitting the not defined ones, but in general they should be defined with the default value?

memsharded (approved):

lasote (approved):

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -166,26 +166,20 @@ def generate(self, env=None, scope="build"):
...
+            args.extend(["--enable-shared", "--disable-static"])
+        else:
+            args.extend(["--disable-shared", "--enable-static", "--with-pic"
+                        if self._conanfile.options.get_safe("fPIC", True)
+                        else "--without-pic"])
+        return args
+
     def generate_args(self):
         configure_args = []
+        configure_args.extend(self._shared_static_args())
```
mattyclarkson, line 181: Unconditionally adding these prevents using the `--enable-option-checking=fatal` option with projects that do not support these flags. For example, `make` and `m4` do not understand these flags.

conan/tools/gnu/autotools.py
```diff
@@ -4,28 +4,47 @@
...
-            source = os.path.join(self._conanfile.source_folder, build_script_folder)
+        configure_args = []
+        if self.default_configure_install_args and self._conanfile.package_folder:
+            def _get_argument(argument_name, cppinfo_name):
+                elements = getattr(self._conanfile.cpp.package, cppinfo_name)
+                return "--{}=${{prefix}}/{}".format(argument_name, elements[0]) if elements else ""
 
-        configure_cmd = "{}/configure".format(source)
+            # If someone want arguments but not the defaults can pass them in args manually
+            configure_args.extend(["--prefix=%s" % self._conanfile.package_folder.replace("\\", "/"),
```
mattyclarkson, line 36: Settng the `--prefix` to the package folder can result in the prefix getting baked into some open source projects and lead to unrelocatable packages.

An idiom for autotools projects that results in more commonly relocatable packages is to do:

```
./configure --prefix=/
make
make DESTDIR=<package-dir> install
```

That will result in the same layout as this but avoids _most_ baking of the prefix paths, there are still a few projects that use `DESTDIR` to bake in paths. Most projects use relative paths.

mattyclarkson (commented):
I'm late to the party on this PR, we've only recently started using Conan in anger.

I thought this might be the easiest way to give feedback to the relevant authors of the issues we've hit with `conan.tools.gnu`. I'm happy to spin out into issues and contribute patches.

memsharded:
@mattyclarkson 

> I thought this might be the easiest way to give feedback to the relevant authors of the issues we've hit with conan.tools.gnu. I'm happy to spin out into issues and contribute patches.

Yes, please, as this is already merged, I think it will be cleaner to report as new issues and handle them there. Thanks!

PR #10981, opened 2022-04-06
Rename CMakeToolchain variables to lowercase

### LINKED ISSUE #10866: Check CMakeToolchain template variable names homogeneity
Currently in the CMakeToolchain we have some variables that are passed to the templates that are uppercase, like `CMAKE_OSX_ARCHITECTURES`, others that are lowercase like `cmake_system_processor`...
We should decide a unique criteria for this. This is important because we have mechanism in CMakeToolchain that you can inject different values for the variables inside the blocks' templates like this:

```python
# modify one or more variables of the context
def generate(self):
    tc = CMakeToolchain(conanfile)
    # block.values is the context dictionary
    build_type = tc.blocks["generic_system"].values["build_type"]
    tc.blocks["generic_system"].values["build_type"] = "Super" + build_type

# modify the whole context values
def generate(self):
    tc = CMakeToolchain(conanfile)
    tc.blocks["generic_system"].values = {"build_type": "SuperRelease"}
```

So changing in the Conan codebase `build_type` to other name tomorrow will break users...

lasote:
Changelog: Fix: Renamed variables from the `CMakeToolchain` context in blocks to be all lowercase. e.g: `CMAKE_OSX_ARCHITECTURES` to `cmake_osx_architectures`.
Docs: omit

Close https://github.com/conan-io/conan/issues/10866

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -283,10 +283,10 @@ class AndroidSystemBlock(Block):
 
     template = textwrap.dedent("""
         # New toolchain things
-        set(ANDROID_PLATFORM {{ ANDROID_PLATFORM }})
-        set(ANDROID_STL {{ ANDROID_STL }})
-        set(ANDROID_ABI {{ ANDROID_ABI }})
-        include({{ ANDROID_NDK_PATH }}/build/cmake/android.toolchain.cmake)
+        set(ANDROID_PLATFORM {{ android_platform }})
+        set(ANDROID_STL {{ android_stl }})
+        set(ANDROID_ABI {{ android_abi }})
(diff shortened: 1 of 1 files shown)


lasote (commented):
A question, we don't have the context variables documented (at least not all of them) do we want to?

czoido (approved):

czoido:
> A question, we don't have the context variables documented (at least not all of them) do we want to?

Maybe we can wait a bit to document the context variables until these new tools are broadly used? At least the variables from CMake that are set in each block are documented. We recently had to modify the Apple block for example, and maybe as we evolve things like cross-compiling support we have to make more changes to the internals of this.

memsharded (approved):

PR #10983, opened 2022-04-06
Fix docstrings from Conan API and some tools

czoido:
Some fixes for docstrings in the Conan API and tools, now that we are using autodoc, the format should be like this: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

This will fix: https://github.com/conan-io/docs/pull/2492

### DIFF
--- a/conan/api/subapi/graph.py
+++ b/conan/api/subapi/graph.py
@@ -47,16 +47,16 @@ class GraphAPI:
     @api_method
     def load_root_test_conanfile(self, path, tested_reference, profile_host, profile_build,
                                  update=None, remotes=None, lockfile=None):
-        """
-        create and initialize a root node from a test_package/conanfile.py consumer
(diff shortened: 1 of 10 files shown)


lasote (approved):

memsharded (approved):
Lets make sure that the API docs have clear (manual) red note saying it is experimental and subject to change. We will not freeze the Python API in 2.0 GA, we will wait a couple of releases into 2.X to declare the Python API stable. It needs real usage to be stabilized.

czoido:
> Lets make sure that the API docs have clear (manual) red note saying it is experimental and subject to change. We will not freeze the Python API in 2.0 GA, we will wait a couple of releases into 2.X to declare the Python API stable. It needs real usage to be stabilized.

I have added a `api_warning.rst` to this PR: https://github.com/conan-io/docs/pull/2492/files

PR #10984, opened 2022-04-06
Adjust INTERFACE_LINK_DIRECTORIES in CMakeDeps

### LINKED ISSUE #10837: [bug] 'cmake' generator does not set INTERFACE_LINK_DIRECTORIES
  * Conan version: 1.46.0

The error surfaced when trying to consume  `boost` with option `magic_autolink=True`. This option causes the link library list to be empty. Instead the boost headers include `#pragma comment` statements to select the libraries to link against. For this approach to work the libraries need to be found by the compiler.

Here are the locations for both cmake generators ([cmake](https://github.com/conan-io/conan/blob/e98a7590b544e738fdb927678d34821ccb78ad78/conans/client/generators/cmake_common.py#L166) and [cmake_find_package](https://github.com/conan-io/conan/blob/8f7ac4306cdf7fc4536728b07457b20895be0023/conans/client/generators/cmake_find_package.py#L179)). The latter correctly sets the `INTERFACE_LINK_DIRECTORIES` variable while the former is missing that part.

lasote:
Changelog: Fix: The `CMakeDeps` generator now set `INTERFACE_LINK_DIRECTORIES` necessary when using auto link `'#pragma comment(lib, "foo")' ` when the required library sets the property `cmake_set_interface_link_directories`.
Docs: https://github.com/conan-io/docs/pull/2510

!!  For the documentation -> Property `cmake_set_interface_link_directories` only supported at the root cppinfo, components not supported (could be done later).

Close #10837

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -13,7 +13,6 @@ from conans.errors import ConanException
 from conans.util.files import save
 
 
-
 class CMakeDeps(object):
 
     def __init__(self, conanfile):
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -19,7 +19,6 @@ class CMakeDeps(object):
         self._conanfile = conanfile
         self.arch = self._conanfile.settings.get_safe("arch")
         self.configuration = str(self._conanfile.settings.build_type)
-        self.set_interface_link_directories = False
 
         # Activate the build config files for the specified libraries
         self.build_context_activated = []
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -19,6 +19,7 @@ class CMakeDeps(object):
         self._conanfile = conanfile
         self.arch = self._conanfile.settings.get_safe("arch")
         self.configuration = str(self._conanfile.settings.build_type)
+        self.set_interface_link_directories = False
 
         # Activate the build config files for the specified libraries
         self.build_context_activated = []
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -26,16 +26,15 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
         components_names = [(components_target_name.replace("::", "_"), components_target_name)
                             for components_target_name in components_targets_names]
 
-        is_win = hasattr(self.conanfile, "settings_build") and \
-            self.conanfile.settings_build.get_safe("os") == "Windows"
+        is_win = self.conanfile.settings.get_safe("os") == "Windows"
+        auto_link = self.conanfile.cpp_info.get_property("cmake_set_interface_link_directories")
         return {"pkg_name": self.pkg_name,
                 "root_target_name": self.root_target_name,
                 "config_suffix": self.config_suffix,
                 "deps_targets_names": ";".join(deps_targets_names),
                 "components_names": components_names,
                 "configuration": self.cmakedeps.configuration,
-                "set_interface_link_directories":
-                    self.cmakedeps.set_interface_link_directories and is_win}
+                "set_interface_link_directories": auto_link and is_win}
 
     @property
     def template(self):
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -115,6 +115,7 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
                      PROPERTY INTERFACE_LINK_LIBRARIES
                      $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}
                                                    ${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}}}> APPEND)
+
         set_property(TARGET {{root_target_name}}
                      PROPERTY INTERFACE_LINK_OPTIONS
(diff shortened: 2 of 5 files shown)


lasote (commented):
Should we do it opt-in? Only windows? Both? apparently only used for the use case of the `#pragma comment(lib, "foo")` for automatic linking (messy feature from MS IMHO)

conans/test/functional/toolchains/cmake/cmakedeps/test_link_order.py
```diff
@@ -226,7 +226,9 @@ def _get_link_order_from_cmake(content):
         if 'main.cpp.o -o example' in line:
             _, links = line.split("main.cpp.o -o example")
             for it_lib in links.split():
-                if it_lib.startswith("-l"):
+                if it_lib.startswith("-L") or it_lib.startswith("-Wl,"):
```
memsharded: Why ``-Wl,`` needs to be skipped here too? Seems a bit unexpected, I see why filtering -L, but not this one

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -128,6 +129,11 @@ def template(self):
                      PROPERTY INTERFACE_COMPILE_OPTIONS
                      $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_COMPILE_OPTIONS{{config_suffix}}}> APPEND)
 
+        # This is only used for '#pragma comment(lib, "foo")' (automatic link)
+        set_property(TARGET {{root_target_name}}
+                     PROPERTY INTERFACE_LINK_DIRECTORIES
```
memsharded, line 139: Now this is interesting. New from 3.13, maybe this can avoid us all the dirty stuff we are doing in our macro to locate the actual library files inside the ``lib_paths``? 

Apparently it is discouraged from the CMake docs (https://cmake.org/cmake/help/latest/command/target_link_directories.html#command:target_link_directories):
> Note This command is rarely necessary and should be avoided where there are other choices. Prefer to pass full absolute paths to libraries where possible, since this ensures the correct library will always be linked. The [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command provides the full path, which can generally be used directly in calls to [target_link_libraries()](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries). Situations where a library search path may be needed include:

So yeah, maybe we shouldn't do it.

Maybe a case for the traits? If the trait is ``libs=True``, but cpp_info.libs is empty, then add this?

memsharded (commented):
Yes, even if the risk doesn't seem large, there are other side effects (like command line length growing too much and overflowing cli) that could happen, so better do not add it by default.

It would be nice if to opt-in we don't need to invent a new utility or conf, and maybe the traits can be used for this.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
lasote: What if someone wants to link dynamically with boost even when the boost recipe declares the `libs`?
Should this be opt-in for the consumer at the `generate()` method adjusting something to the `CMakeDeps`?

conans/test/functional/toolchains/cmake/cmakedeps/test_link_order.py
lasote: Well this is part of the mess parsing the output in Macos for example (not windows) you receive this line (I guess because of the INTERFACE_LINK_DIRECTORIES):

`'-Wl,-rpath,/private/var/folders/5j/1gjq0lls4rd8mkcgs0jxt7z40000gn/T/tmp0ka4_4vvconans/pathwithoutspaces/data/libA/version/_/_/package/6bf13cf4373232118e57f68f27213bb66e030a17/lib'`

But we don't what that line to be processed because we are searching for the actual libs.
I will change it to `-Wl,-rpath` to be more clear.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
lasote: Anyway I would limit this to Windows until a legit use case exists in others. (and the messy changes for parsing the output won't be necessary)

memsharded: Even if this is only for Windows, I can see some risks adding this line in Windows (recall Windows is one of the OS that struggle with long compilation lines too). 

The thing is that this is borderline bad practice, linking with pragma is not clean, kind of a very messy feature. Better make it very opt-in. Possible options are:

- Explicit opt-in in the CMakeDeps customer side
- Explicit opt-in as a ``cpp_info.property`` in the package side
- Implicit opt-in with libs=empty + libdirs=not-empty combination defined in package side

I think I would add anything unless an opt-in enables this.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -26,12 +26,16 @@ def context(self):
         components_names = [(components_target_name.replace("::", "_"), components_target_name)
                             for components_target_name in components_targets_names]
 
+        is_win = hasattr(self.conanfile, "settings_build") and \
+            self.conanfile.settings_build.get_safe("os") == "Windows"
```
lasote: Change to host settings.

lasote (commented):
Change to property on the dependency and keep the windows check (host settings)

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -13,13 +13,13 @@
...
 
 
-
 class CMakeDeps(object):
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
         self.arch = self._conanfile.settings.get_safe("arch")
         self.configuration = str(self._conanfile.settings.build_type)
+        self.set_interface_link_directories = False
```
memsharded: No longer needed?

memsharded (approved):
Looking good

memsharded (approved):

PR #10985, opened 2022-04-06
[CMakeToolchain] Adding Xcode flags

### LINKED ISSUE #9448: [feature] Improve CMakeToolchain support for xcode flags
Coming from https://github.com/conan-io/conan-center-index/pull/5545

The `darwing-toolchain` generator recipe used to provide some useful configurable apple flags: https://github.com/ezored/conan-darwin-toolchain/blob/stable/1.3.0/conanfile.py#L87-L127

We should take a look at the cmake flags used there and see if we can improve the integration for cmake apple-clang projects

cc/ @theodelrieu @paulo-coutinho

franramirez688:
Changelog: Feature: Added new configuration fields: `tools.apple:enable_bitcode`, `tools.apple:enable_arc` and `tools.apple:enable_visibility`.
Changelog: Feature: Added new mechanism to inject common Xcode flags in `CMakeToolchain` generator if enabled Bitcode, ARC, or Visibility configurations.
Closes: https://github.com/conan-io/conan/issues/9448
Docs: https://github.com/conan-io/docs/pull/2506

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -4,13 +4,6 @@
 from conans.util.runners import check_output_runner
 
 
-def is_apple_cross_building(conanfile):
-    os_host = conanfile.settings.get_safe("os")
-    arch_host = conanfile.settings.get_safe("arch")
-    arch_build = conanfile.settings_build.get_safe("arch")
-    return os_host in ('iOS', 'watchOS', 'tvOS') or (os_host == 'Macos' and arch_host != arch_build)
-
-
 def is_apple_os(os_):
     """returns True if OS is Apple one (Macos, iOS, watchOS or tvOS"""
     return str(os_) in ['Macos', 'iOS', 'watchOS', 'tvOS']
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -1,8 +1,14 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import os
 
 from conans.util.runners import check_output_runner
+
+
+def is_apple_cross_building(conanfile):
+    os_host = conanfile.settings.get_safe("os")
+    arch_host = conanfile.settings.get_safe("arch")
+    arch_build = conanfile.settings_build.get_safe("arch")
+    return os_host in ('iOS', 'watchOS', 'tvOS') or (os_host == 'Macos' and arch_host != arch_build)
 
 
 def is_apple_os(os_):
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -6,7 +6,7 @@ from collections import OrderedDict
 from jinja2 import Template
 
 from conan.tools._compilers import architecture_flag
-from conan.tools.apple.apple import is_apple_os, to_apple_arch
+from conan.tools.apple.apple import is_apple_os, to_apple_arch, is_apple_cross_building
 from conan.tools.build import build_jobs
 from conan.tools.build.cross_building import cross_building
 from conan.tools.cmake.toolchain import CONAN_TOOLCHAIN_FILENAME
@@ -328,7 +328,63 @@ class AppleSystemBlock(Block):
         # Setting CMAKE_OSX_DEPLOYMENT_TARGET if "os.version" is defined by the used conan profile
         set(CMAKE_OSX_DEPLOYMENT_TARGET "{{ cmake_osx_deployment_target }}" CACHE STRING "")
         {% endif %}
+        {% if enable_bitcode is defined and enable_bitcode is sameas true %}
+        # Bitcode
+        set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE YES)
+        set(CMAKE_XCODE_ATTRIBUTE_BITCODE_GENERATION_MODE bitcode)
+        {% endif %}
+        {% if enable_arc is defined and enable_arc is sameas true %}
+        # ARC
+        set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES)
+        {% elif enable_arc is defined and enable_arc is sameas false %}
+        # ARC
+        set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC NO)
+        {% endif %}
+        {% if enable_visibility is defined and enable_visibility is sameas true %}
+        # Visibility
+        set(CMAKE_XCODE_ATTRIBUTE_GCC_SYMBOLS_PRIVATE_EXTERN YES)
+        {% elif enable_visibility is defined and enable_visibility is sameas false %}
+        # Visibility
+        set(CMAKE_XCODE_ATTRIBUTE_GCC_SYMBOLS_PRIVATE_EXTERN NO)
+        {% endif %}
+        {% if common_flags|length > 0 %}
+        string(APPEND CONAN_CXX_FLAGS "{% for flag in common_flags %} {{ flag }}{% endfor %}")
+        string(APPEND CONAN_C_FLAGS "{% for flag in common_flags %} {{ flag }}{% endfor %}")
+        string(APPEND CONAN_SHARED_LINKER_FLAGS "{% for flag in common_flags %} {{ flag }}{% endfor %}")
+        string(APPEND CONAN_EXE_LINKER_FLAGS "{% for flag in common_flags %} {{ flag }}{% endfor %}")
+        {% endif %}
         """)
+
+    def _get_xcode_common_flags(self):
+        """
+        Adding Xcode flags if enabled Bitcode, ARC or Visibility via conf.
+
+        :return: `dict` with common flags and boolean values if enabled bitcode, arc or visibility.
+        """
+        # Issue related: https://github.com/conan-io/conan/issues/9448
+        enable_bitcode = self._conanfile.conf.get("tools.apple:enable_bitcode",
+                                                  default=self._conanfile.settings.get_safe('os') in ["watchOS", "tvOS"],
+                                                  check_type=bool)
+        enable_arc = self._conanfile.conf.get("tools.apple:enable_arc", check_type=bool)
+        enable_visibility = self._conanfile.conf.get("tools.apple:enable_visibility", check_type=bool)
+        common_flags = []
+        if enable_bitcode:
+            if self._conanfile.settings.get_safe('build_type') == "Debug":
+                common_flags.append("-fembed-bitcode-marker")
+            else:
+                common_flags.append("-fembed-bitcode")
+        if enable_arc:
+            common_flags.append("-fobjc-arc")
+        elif enable_arc is False:
+            common_flags.append("-fno-objc-arc")
+        if enable_visibility is False:
+            common_flags.append("-fvisibility=hidden")
+        return {
+            "common_flags": common_flags,
+            "enable_bitcode": enable_bitcode,
+            "enable_arc": enable_arc,
+            "enable_visibility": enable_visibility
+        }
 
     def _apple_sdk_name(self):
         """
@@ -357,7 +413,7 @@ class AppleSystemBlock(Block):
 
     def context(self):
         os_ = self._conanfile.settings.get_safe("os")
-        if os_ not in ['Macos', 'iOS', 'watchOS', 'tvOS']:
+        if not is_apple_os(os_):
             return None
 
         arch = self._conanfile.settings.get_safe("arch")
@@ -378,6 +434,7 @@ class AppleSystemBlock(Block):
             # macOS like iOS, tvOS, or watchOS.
             ctxt_toolchain["cmake_osx_deployment_target"] = host_os_version
 
+        ctxt_toolchain.update(self._get_xcode_common_flags())
         return ctxt_toolchain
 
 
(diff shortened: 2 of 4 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,8 +328,64 @@ class AppleSystemBlock(Block):
...
+        Adding Xcode flags if enabled Bitcode, ARC or Visibility via conf.
+
+        :return: `dict` with common flags and boolean values if enabled bitcode, arc or visibility.
+        """
+        # Issue related: https://github.com/conan-io/conan/issues/9448
+        enable_bitcode = self._conanfile.conf.get("tools.apple:enable_bitcode",
+                                                  default=self._conanfile.settings.get_safe('os') in ["watchOS", "tvOS"],
+                                                  check_type=bool)
+        enable_arc = self._conanfile.conf.get("tools.apple:enable_arc", check_type=bool)
+        enable_visibility = self._conanfile.conf.get("tools.apple:enable_visibility", check_type=bool)
```
SSE4: I don't think this one is apple specific at all

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,8 +328,64 @@ class AppleSystemBlock(Block):
...
+            if self._conanfile.settings.get_safe('build_type') == "Debug":
+                common_flags.append("-fembed-bitcode-marker")
+            else:
+                common_flags.append("-fembed-bitcode")
+        if enable_arc:
+            common_flags.append("-fobjc-arc")
+        elif enable_arc is False:
+            common_flags.append("-fno-objc-arc")
+        if enable_visibility is False:
+            common_flags.append("-fvisibility=hidden")
```
SSE4: I don't like boolean option. maybe it's better to use an enum: `-fvisibility=[default|internal|hidden|protected]`

conan/tools/apple/apple.py
```diff
@@ -1,10 +1,16 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import os
 
 from conans.util.runners import check_output_runner
 
 
+def is_apple_cross_building(conanfile):
```
memsharded: Is this intended to be a public tool? I don't like it public, why the logic is not just embedded in the general ``cross_building()`` tool? It seems dirty that recipes would start to have custom code with ``is_apple_cross_building()``. Maybe let it be private to the toolchain at the moment?

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,8 +328,64 @@ class AppleSystemBlock(Block):
...
+        # Bitcode
+        set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE YES)
+        set(CMAKE_XCODE_ATTRIBUTE_BITCODE_GENERATION_MODE bitcode)
+        {% endif %}
+        {% if enable_arc is defined and enable_arc is sameas true %}
+        # ARC
+        set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES)
+        {% elif enable_arc is defined and enable_arc is sameas false %}
+        # ARC
+        set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC NO)
```
memsharded: Feels like 2 PR in one. One is the addition of bitcode and other flags, and the other the injection of general flags. But it is fine, it was just a bit confusing at first.

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,8 +328,64 @@ class AppleSystemBlock(Block):
...
+        # Issue related: https://github.com/conan-io/conan/issues/9448
+        enable_bitcode = self._conanfile.conf.get("tools.apple:enable_bitcode",
+                                                  default=self._conanfile.settings.get_safe('os') in ["watchOS", "tvOS"],
+                                                  check_type=bool)
+        enable_arc = self._conanfile.conf.get("tools.apple:enable_arc", check_type=bool)
+        enable_visibility = self._conanfile.conf.get("tools.apple:enable_visibility", check_type=bool)
+        common_flags = []
+        if enable_bitcode:
+            if self._conanfile.settings.get_safe('build_type') == "Debug":
+                common_flags.append("-fembed-bitcode-marker")
```
memsharded: I am a bit confused about this. Why adding these explicitly as flags? Aren't they implemented with ``CMAKE_XXXX`` variables above?

franramirez688: I've based the PR on this other recipe https://github.com/ezored/conan-darwin-toolchain/blob/stable/1.3.0/conanfile.py#L87-L127 I'm not saying that it's fine as it is and we shouldn't change anything but I started using exactly the same logic.

memsharded: Sure, make sense. Then yes, we need to understand a bit better the relationship between the CMAKE flags and those compiler flags (the flags might need to be added to other Conan toolchain integrations maybe too?)

franramirez688:
Basically, what I'm trying to do is to add some capabilities from the https://github.com/leetal/ios-cmake repository so we'll be able to add some common Xcode flags if those configurations are enabled/disabled.

conan/tools/cmake/toolchain/blocks.py
franramirez688: I don't really know how to put it. You're right, it's not specific to Apple but we're only using it for this one for now. You can always inject any flags thanks to `conf` so to convert it into an enum variable perhaps it's not needed as well...

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,6 +328,42 @@ class AppleSystemBlock(Block):
         # Setting CMAKE_OSX_DEPLOYMENT_TARGET if "os.version" is defined by the used conan profile
         set(CMAKE_OSX_DEPLOYMENT_TARGET "{{ CMAKE_OSX_DEPLOYMENT_TARGET }}" CACHE STRING "")
         {% endif %}
+        set(BITCODE "")
+        set(FOBJC_ARC "")
+        set(VISIBILITY "")
+        {% if enable_bitcode is defined and enable_bitcode is sameas true %}
```
lasote: Probably less fancy:
```
{% if enable_bitcode is not none %}  
{% if enable_bitcode %}  
    ---
{% else %}  
      ....
{% endif %}
{% endif %}
```

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,6 +328,42 @@ class AppleSystemBlock(Block):
         # Setting CMAKE_OSX_DEPLOYMENT_TARGET if "os.version" is defined by the used conan profile
         set(CMAKE_OSX_DEPLOYMENT_TARGET "{{ CMAKE_OSX_DEPLOYMENT_TARGET }}" CACHE STRING "")
         {% endif %}
+        set(BITCODE "")
```
lasote, line 331: Do we need to declare these variables with empty?

lasote (commented):
I have to check the toolchain you used as a reference to understand a little bit but at first sight, this looks good.

conan/tools/cmake/toolchain/blocks.py
franramirez688: I thought it was not needed too but one test failed because of that. I'll recheck it just in case

franramirez688: That was the reason: https://github.com/conan-io/conan/pull/10292/files#diff-a42d08ebe9e9d3240230bce3ce865eb2b03967fc334ab18ba057bda007990e8aR183

lasote: I don't get it.

conan/tools/cmake/toolchain/blocks.py
SSE4: I think it will introduce more confusion than needed. it would be better to do it right from the very beginning.
otherwise, users will ask, what will take more priority (`tools.apple:enable_visibility` vs flags injected in conf), and what happens if they conflict...

I would propose `tools.gnu:visiibility` with values `default|internal|hidden|protected`

conan/tools/cmake/toolchain/blocks.py
franramirez688: The PR https://github.com/conan-io/conan/pull/10292 was supposed to solve the issue when any project is configured with `-Werror=dev --warn-uninitialized` it was failing due to uninitialized variables, so CMakeToolchain got a test that it's checking if any variable is properly initialized, otherwise it fails:

**conans/test/functional/toolchains/cmake/test_cmake_toolchain.py**
```python
@pytest.mark.tool_cmake
def test_cmaketoolchain_no_warnings():
    """Make sure unitialized variables do not cause any warnings, passing -Werror=dev
    and --wanr-unitialized, calling "cmake" with conan_toolchain.cmake used to fail
    """
```

conans/model/conf.py
```diff
@@ -40,6 +40,9 @@
     "tools.system.package_manager:sudo": "Use 'sudo' when invoking the package manager tools in Linux (False by default)",
     "tools.system.package_manager:sudo_askpass": "Use the '-A' argument if using sudo in Linux to invoke the system package manager (False by default)",
     "tools.apple.xcodebuild:verbosity": "Verbosity level for xcodebuild: 'verbose' or 'quiet",
+    "tools.apple:enable_bitcode": "(boolean) Enable/Disable Bitcode Xcode flags",
```
memsharded: Are those Xcode exclusively? I think they are ``apple-clang`` compiler flags. Maybe change description of conf?

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -328,6 +328,42 @@ class AppleSystemBlock(Block):
...
+        # Bitcode OFF
+        set(CMAKE_XCODE_ATTRIBUTE_ENABLE_BITCODE "NO")
+        {% endif %}
+        {% if enable_arc %}
+        # ARC ON
+        set(FOBJC_ARC "-fobjc-arc")
+        set(CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC "YES")
+        {% elif enable_arc is not none %}
+        # ARC OFF
+        set(FOBJC_ARC "-fno-objc-arc")
```
memsharded, line 353: I am still a bit confused about the duplication, of setting the raw flags ``"-fno-objc-arc"``, and also the CMake ``CMAKE_XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC`` variable. Isn't that duplicating the flags?

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -357,15 +393,30 @@ def _apple_sdk_name(self):
...
         host_architecture = to_apple_arch(arch)
         host_os_version = self._conanfile.settings.get_safe("os.version")
         host_sdk_name = self._apple_sdk_name()
+        is_debug = self._conanfile.settings.get_safe('build_type') == "Debug"
 
-        ctxt_toolchain = {}
+        # Reading some configurations to enable or disable some Xcode toolchain flags and variables
+        # Issue related: https://github.com/conan-io/conan/issues/9448
+        # Based on https://github.com/leetal/ios-cmake repository
+        enable_bitcode = self._conanfile.conf.get("tools.apple:enable_bitcode",
```
memsharded: Are we sure the default for watchOS and tvOS should be ``True``? If that is the default, why the CMake or compiler is not setting them and Conan should set it?

conan/tools/cmake/toolchain/blocks.py
memsharded: Ok, I see it is because one thing is useful only for the Xcode generator.
The original iostoolchain did that, so better use the Cmake generator to only define the flags when needed (not Xcode), and only define the CMAKE_XCODE_ATTR... when needed.

czoido: Check this condition: https://github.com/leetal/ios-cmake/blob/14092d317b84a3037e71b870d2117dda126811a3/ios.toolchain.cmake#L779

memsharded (changes requested):
Please have a look at the comments

conan/tools/cmake/toolchain/blocks.py
SSE4: these are compiler flags, they aren't specific to Xcode somehow. could be used in autotools, CMake, Meson, etc.

conan/tools/cmake/toolchain/blocks.py
SSE4: no, it shouldn't

conan/tools/cmake/toolchain/blocks.py
lasote: @SSE4 yes, that's the point, but the original toolchain sets the `CMAKE_XCODE_ATTR...` when the generator is xcode OTHERWISE it sets the compiler flags. Here in this PR everything is set, so it will be duplicated somehow.

memsharded (approved):

lasote (approved):

czoido (approved):

MarkusWittlinger:
Hi @franramirez688 ,
I have a quick question (and I hope it is OK to comment here). 
Is there a reason why the VISIBILITY is not passed to the CONAN_C_FLAGS but only to the C++ ones?

I had a look at the ios-cmake repo and there it is set for both CMAKE_C_FLAGS and CMAKE_CXX_FLAGS (see https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake).

franramirez688:
> Hi @franramirez688 , I have a quick question (and I hope it is OK to comment here). Is there a reason why the VISIBILITY is not passed to the CONAN_C_FLAGS but only to the C++ ones?
> 
> I had a look at the ios-cmake repo and there it is set for both CMAKE_C_FLAGS and CMAKE_CXX_FLAGS (see https://github.com/leetal/ios-cmake/blob/master/ios.toolchain.cmake).

That's a very good question. It looks like a bug indeed 🤔 I don't remember why but it looks like it was missed...

PR #10997, opened 2022-04-08
POC: Using filesystem to create more specific CMakeDeps targets

### LINKED ISSUE #10181: [bug] `cmakedeps_macros.cmake` should use IMPORTED_IMPLIB on Windows

lasote:

Related to #10497

This is only a POC to explore this idea of preparing the library information from Python and passing CMake the library full path and library type (static, shared, unknown) to get a better-imported target without increasing the complexity of CMake code.

It simplifies a bit the target creation (no find_library call), on the other hand, bugs in library location and type could be introduced (previously managed by find_library) and the files need to be there (collect_libraries like).

We do like this? If yes, do we want this as the default (and only) operation mode or an opt/in/out?

Pending to verify what happen with editables when not built yet.

Close #10181


conan/tools/__init__.py
```diff
@@ -1,2 +1,23 @@
+import os
+
+from conans.cli.output import ConanOutput
+
 CONAN_TOOLCHAIN_ARGS_FILE = "conanbuild.conf"
 CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
+
+
+def get_lib_file_path(libdirs, lib):
```
lasote: This has been moved from the Bazel one. To be reused now by CMake

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -274,3 +290,32 @@ def join_paths_single_var(values):
...
+    def _get_lib_info(self, libdirs, lib):
+
+        path = get_lib_file_path(libdirs, lib)
+        if not path:
+            return None
+        libname = os.path.basename(path)
+        ret_path = path.replace('\\', '/').replace('$', '\\$').replace('"', '\\"')
+        ret_cmake_name = re.sub("[^0-9a-zA-Z]+", "_", libname)
+
+        ext = libname.split(".", 1)
```
gabyx, line 303: Use here os.path.split

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -274,3 +290,32 @@ def join_paths_single_var(values):
...
+            if ext == "a":
+                ret_type = "STATIC"
+            if ext == "lib":
+                if hasattr(self._conanfile, 'settings_build'):
+                    os_build = self._conanfile.settings_build.get_safe('os')
+                else:
+                    os_build = self._conanfile.settings.get_safe('os_build')
+                # FIXME: Find the DLL in the bindirs?
+                if os_build == "Windows" and self._conanfile.options.shared:
+                    ret_type = "SHARED"
```
gabyx, line 332: `STATIC` should be set otherwise, which is not after 318 is not the case...?

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -47,36 +47,22 @@ def template(self):
...
-                       set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_FOUND_LIBRARY})
-                       list(APPEND _CONAN_ACTUAL_TARGETS ${_LIB_NAME})
-                   else()
-                       conan_message(STATUS "Skipping already existing target: ${_LIB_NAME}")
-                   endif()
-                   list(APPEND _out_libraries_target ${_LIB_NAME})
-                   conan_message(DEBUG "Found: ${CONAN_FOUND_LIBRARY}")
+                   set(_LIB_TYPE ${${package_name}_LIB_${_LIBRARY_VAR_NAME}_TYPE${config_suffix}})
+                   set(_REAL_PATH ${${package_name}_LIB_${_LIBRARY_VAR_NAME}_REAL_PATH${config_suffix}})
+                   add_library(${_LIBRARY_VAR_NAME} ${_LIB_TYPE} IMPORTED)
```
gabyx, line 60: I think so far looks ok, but the CMake part still needs the stuff from : 
https://github.com/conan-io/conan/pull/10497/files#diff-c4faa1c61f8d99a6ae53496c9eea30c47d48491daa851281e034e60b83b5bb56R132-R135
to work correctly.
I would take the improvements in this PR and merge it with yours, removing the need to determine the library type because we have that now.

Otherwise this is really nice!

lasote: > I think so far looks ok, but the CMake part still needs the stuff from :

Which lines exactly?

> I would take the improvements in this PR and merge it with yours, removing the need to determine the library type because we have that now.

Please, take this as a POC, we have to discuss if we are adopting this it is too soon to merge it elsewhere.

gabyx: @lasote: No worries, I 've seen that it is a POC, just wanted to give some feedback: The link supposed to work, but it doesnt sorry: L132-L135

gabyx: https://github.com/conan-io/conan/pull/10497/files#r848234733

gabyx: I ssee this:

If `_REAL_PATH` is the `.dll` file on Windows I suppose, that will map to `IMPORTED_LOCATION`, so correct. The corresponding `.lib` file should map to `IMPORTED_IMPLIB` (on windows), maybe its not needed because the `.lib` is always searched next to the `.dll`. However if its possible and available I would set this too to be more correct.

franramirez688 (approved):
I think it's really interesting this POC.
Concerns about changing `find_library` to our own `libs_info` because it's a big change and, like @lasote is saying, it could bring a few/a lot of unexpected bugs.
I don't really know if it could be something opt-in to let a user chooses whatever they want...

Anyway, I think it's a good approach and we could figure out the result by applying it to different existing libraries in conan-center.

gabyx:
> I think it's really interesting this POC. Concerns about changing `find_library` to our own `libs_info` because it's a big change and, like @lasote is saying, it could bring a few/a lot of unexpected bugs. I don't really know if it could be something opt-in to let a user chooses whatever they want...
> 
> Anyway, I think it's a good approach and we could figure out the result by applying it to different existing libraries in conan-center.

I feel the same, however not soo concerned about replacing `find_library` because it anyway was only used because the information was not available. I think on linux/macos the POC (from a CMake perspective) is correct, but more work is needed on Windows to set the `.lib` part I guess to make it complete. Thanks a lot for taking this further.

lasote:
Hi, we talked in the team and we don't see this going forward, the risk of bugs and a step back in stability doesn't compensate for the value.

lasote:
We are going to study if we can simplify a bit by leveraging the "package_type" from conan 2.

memsharded:
We can closed this one as obsolete, targets will be defined by model, not checking actual files

PR #10999, opened 2022-04-08
POC "conan cache path" command

### LINKED ISSUE #10942: [feature] conan download to target directory
It would be nice if [conan download](https://docs.conan.io/en/latest/reference/commands/misc/download.html) could have an option to specify target directory for the download.

1. We want this for our legacy build system that will look for a library source by folder. This way we can support both `conan install` and legacy `make` files using same `conanfile.txt`
2. Makes it easier to peek into a package by making it end up in current directory
3. Makes it easier to compare against package already downloaded to `.conan` folder

lasote:
Missing tests. To validate if we like the syntax and the API

Close #10942

### DIFF
--- a/conan/api/conan_api.py
+++ b/conan/api/conan_api.py
@@ -1,5 +1,6 @@
 import sys
 
+from conan.api.subapi.cache import CacheAPI
 from conans import __version__ as client_version
 from conan.api.subapi.config import ConfigAPI
 from conan.api.subapi.download import DownloadAPI
@@ -52,6 +53,7 @@ class ConanAPIV2(object):
         self.new = NewAPI(self)
         self.upload = UploadAPI(self)
         self.download = DownloadAPI(self)
+        self.cache = CacheAPI(self)
 
 
 ConanAPI = ConanAPIV2
--- a/conan/api/subapi/cache.py
+++ b/conan/api/subapi/cache.py
@@ -1,0 +1,40 @@
+from conan.api.subapi import api_method
+from conans.cli.conan_app import ConanApp
+from conans.model.package_ref import PkgReference
+from conans.model.recipe_ref import RecipeReference
+
+
+class CacheAPI:
+
+    def __init__(self, conan_api):
+        self.conan_api = conan_api
+
+    @api_method
+    def exports_path(self, ref: RecipeReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref_layout = app.cache.ref_layout(ref)
+        return ref_layout.export()
+
+    @api_method
+    def exports_sources_path(self, ref: RecipeReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref_layout = app.cache.ref_layout(ref)
+        return ref_layout.export_sources()
+
+    @api_method
+    def sources_path(self, ref: RecipeReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref_layout = app.cache.ref_layout(ref)
+        return ref_layout.source()
+
+    @api_method
+    def build_path(self, pref: PkgReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref_layout = app.cache.pkg_layout(pref)
+        return ref_layout.build()
+
+    @api_method
+    def package_path(self, pref: PkgReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref_layout = app.cache.pkg_layout(pref)
+        return ref_layout.package()
--- a/conans/cli/commands/cache.py
+++ b/conans/cli/commands/cache.py
@@ -1,0 +1,87 @@
+import os
+
+from conan.api.conan_api import ConanAPIV2
+from conans.cli.command import conan_command, COMMAND_GROUPS, conan_subcommand
+from conans.cli.commands.list import json_formatter
+from conans.cli.output import ConanOutput
+from conans.errors import ConanException
+from conans.model.package_ref import PkgReference
+from conans.model.recipe_ref import RecipeReference
+
+
+@conan_command(group=COMMAND_GROUPS['consumer'], formatters={"json": json_formatter})
+def cache(conan_api: ConanAPIV2, parser, *args):
+    """Performs file operations in the local cache (of recipes and packages)"""
+    pass
+
+
+# FIXME: Formatters
+@conan_subcommand()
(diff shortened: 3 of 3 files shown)


memsharded (approved):
Looking good in general.

Find below a couple of tests I did while checking the UI.

Some things to consider:
- ``conan cache path <pkg-ref> source/recipe..`` with a full pkg-ref (not a recipe reference) fails, even if it would be completely possible to return it, as it is well defined. If we want it to fail, maybe a more explicit error msg
- The order of positional arguments not 100% natural, ``conan cache path recipe <ref>`` makes sense too. Maybe defaulting to recipe and package for <ref> and <pkg-ref> respectively and ``--source``, ``--exports-sources`` to alternatively obtain the others?

```python
from conans.test.assets.genconanfile import GenConanfile
from conans.test.utils.tools import TestClient


def test_cache_path_errors():
    c = TestClient()
    c.run("cache path x recipe", assert_error=True)
    assert "ERROR: Invalid recipe or package reference" in c.out
    c.run("cache path pkg/1.0#rev1 recipe", assert_error=True)
    assert "ERROR: No entry for recipe 'pkg/1.0#rev1'" in c.out


def test_cache_path():
    c = TestClient()
    c.save({"conanfile.py": GenConanfile("hello", "0.1")})
    c.run("create .")
    rev = c.exported_recipe_revision()
    pid = c.created_package_id("hello/0.1")
    prev = c.created_package_revision("hello/0.1")
    c.run(f"cache path hello/0.1#{rev} recipe")
    assert "conanfile.py" in c.out
    c.run(f"cache path hello/0.1#{rev} sources")
    # a bit difficult to validate the ouput, maybe check for some file inside?
    c.run(f"cache path hello/0.1#{rev}:{pid}#{prev} package")
    # a bit difficult to validate the ouput, maybe check for some file inside?
```

lasote:
> conan cache path <pkg-ref> source/recipe.. with a full pkg-ref (not a recipe reference) fails, even if it would be completely possible to return it, as it is well defined. If we want it to fail, maybe a more explicit error msg

True, I think we can run the command successfully in that case.

> The order of positional arguments not 100% natural, conan cache path recipe <ref> makes sense too. Maybe defaulting to recipe and package for and respectively and --source, --exports-sources to alternatively obtain the others?

I tried but I didn't like that `--build` was only valid for a package reference and `--recipe` , `--source` ... etc for a reference. If we allow a package reference for the latest it would be less messy.
We would need `--recipe` too if we finally accept a package reference to return "recipe" stuff, so not sure if default to the **conanfile's path** for a regular reference and to the package **folder** would be worth it. Explicit always better?

 @czoido @franramirez688  Feedback needed

franramirez688 (approved):

franramirez688:
Overall, it's looking good.

From my point of view, I'd like something like this:

```
conan cache RECIPE-REF  -> returns recipe cache folder by default
conan cache PKG-REF -> returns package cache folder by default
conan cache WHATEVER --folder [recipe, exports, exports_sources, sources, build, package]
```
The `path` subcommand, I guess it's not needed for now, is it? If we assume that `conan cache` is about local cache paths then I think it's explicit enough... Wdyt?

lasote:
> conan cache WHATEVER --folder [recipe, exports, exports_sources, sources, build, package]

The `recipe` is not a folder but the conanfile.py, the real folder is `exports`. I found it convenient to return the path to the conanfile.py but we can discuss it.

> The path subcommand, I guess it's not needed for now, is it? If we assume that conan cache is about local cache paths then I think it's explicit enough... Wdyt?

Sounds like the `cache` command would have more stuff, like Cleaning old directories? or whatever.

franramirez688:
> The recipe is not a folder but the conanfile.py, the real folder is exports. I found it convenient to return the path to the conanfile.py but we can discuss it.

Yeah, sorry, I meant the path to the conanfile.py.

> Sounds like the cache command would have more stuff, like Cleaning old directories? or whatever.

Oh, that's right. It could have more than one subcommand.

Then, my only suggestion is to avoid that `--recipe`, `--build`, etc., and use something like `--folder exports`. I think that's explicit enough:
```
conan cache path WHATEVER --folder [exports, exports_sources, ...]
```

lasote:
I meant that the argument cannot be named `--folder` if we are using `recipe` because `recipe` is not a folder.

franramirez688:
> I meant that the argument cannot be named `--folder` if we are using `recipe` because `recipe` is not a folder.

Ok, then, taking that into account, could this be a better approach to the previous one?

```
conan cache path WHATEVER  -> returns conanfile.py location by default
conan cache path WHATEVER --folder [exports, exports_sources, sources, build, package]
```
Where `WHATEVER` could be a package/recipe reference. It's true that `conan cache path PKG-REF` returns the conanfile.py location as well could be confusing but I guess it's clear enough for now.

lasote:
Wops, I pushed to my fork "develop2" by mistake. Self-reminder: Close this and open a branch when the discussion is over

memsharded:
I would prefer to have the default ``conan cache path <ref>`` to return the folder containing the ``conanfile.py`` by default. Because that is useful to open an IDE, or file explorer at that location, see everything that is there, conandata.yml and stuff. And it is very easy to open the ``conanfile.py`` there, just append ``conanfile.py`` explicitly. While getting the folder from the recipe conanfile.py is not possible. Also because it would be assymetric, ``conan cache path <pkg-ref>`` returns a folder and ``conan cache path <ref>`` returns a file? A bit unexpected.

lasote:
> Also because it would be assymetric, conan cache path <pkg-ref> returns a folder and conan cache path <ref> returns a file? A bit unexpected.

It is not asymmetric, because you specify explicitly `--folder` to get a folder, until then you have the conanfile.py path.

Said that I'm also ok with that approach too.
So:

```
conan cache path WHATEVER  -> returns the "exports" folder by default (like if `--folder` default is `exports`)
conan cache path WHATEVER --folder [exports, exports_sources, sources, build, package]
```

franramirez688:
I like it too! 👍  🔝

memsharded:
Yes, that is good too, go for it. I started my comment with "I would prefer", because it was just my personal preference, but I understand other approaches. Yours of always showing that folder, no matter if you pass a "ref" or a "pkg-ref" also sounds pretty good and consistent, I am totally fine with it if the rest of the team also likes it.

lasote:
Good! so I'm closing this (made by mistake in my develop2 branch) and open a new one updated with the feedback and tests.

PR #11004, opened 2022-04-10
Feature/develop2 settings preprocessor plugin

memsharded:
I am trying to get rid of the ``settings_preprocessor``, at the moment it is only used to define ``compiler.runtime_type=Release/Debug`` when it is not defined to follow the ``build_type``. It seems that this behavior shouldn't need to be defined so harcoded in Conan:

- It is used mostly for a UX reason, when users specify on cli ``--build_type=Debug``, they would need to remember to change too ``--compiler.runtime_type=Debug``, otherwise they would get the Release one and have link errors
- We wanted to maintain the flexibility for users to define the runtime type irrespective of the build_type (in case something custom is wanted for RelWithDebInfo and other possible custom configurations)

I am suggesting a plugin (I tried a hook, but it was just too dirty, architecture-wise).
Main issue is that I'd prefer not to create the plugin, but only use it if users install it, but that will still keep the same old poor UX issue?


lasote:
I don't see a great issue if we create the plugin if it doesn't exist, it is less coupled than the current preprocessor and indeed interesting for users.

memsharded:
> I don't see a great issue if we create the plugin if it doesn't exist, it is less coupled than the current preprocessor and indeed interesting for users.

Yes, honestly I was struggling initially because I tried to make it a hook, and it was certainly more uncomfortable to make the hook file generate automatically. For a plugins which is more decoupled, it is relatively easy, we already do the same for compatibility.py. I will add the generation.

memsharded:
Generation added, and tests fixed now, please check.

lasote (approved):

czoido (approved):

Issue #11009, opened 2022-04-11
[question] How to use self.run and VirtualBuildEnv to capture the output


gpxricky:
Hi,

when using the `VirtualBuildEnv` generator, is it still possible to capture the output of a `self.run` command properly without any additional steps? At present the output of the `run` method starts with

```
Capturing current environment in /some_path/deactivate_conanbuildenv-<arch>.sh': 'Configuring environment variables'
<output I am interested in>
```

Setting `env` to `None` doesn't work as it implicitly activates `conanbuild`. Or is the idea to use redirection in future - but then the question is if it would be better to remove the `output` argument?

Best regards,
Markus

- [X] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

gpxricky:
A simple workaround would be to use any non existing name for the `env` argument, e.g. `Dummy`. But this looks like a very bad design, isn't it?

lasote:
Maybe `env=""` but I think I didn't understand the underlying issue, Do you need to capture the output of a command to a variable or file? is something failing and not being printed to stdout? I don't know, how is that related to the VirtualBuildEnv generator.

EDIT: I tried and the `env=""` won't work. I'll open a PR to default to "conanbuild" instead of `None` to see if is there any other reason for that.

memsharded:
Yes, I don't think that disabling the environment is the solution. If the environment file was generated in the first place, it is very likely that it is still needed. And the ``self.run()`` command will run the command prefixed with the environment script, so yes, that will probably output that information too. I guess the only alternative is to let the capturer filter the first lines if they don't want them. Because even if 2.0 uses redirection instead of ``output`` (and it does, it gets an ``stdout`` argument instead), you will still get that in the stdout stream.

gpxricky:
Allowing `None` is fine for me at the moment, therefore I'll close this issue.

PR #11011, opened 2022-04-11
self.run(env="conanbuild") instead of "None"

lasote:
Changelog: omit

Related #11009
If we would like to skip the "conanbuild" for any reason to be applied at some `self.run` execution (sounds legit use case) we don't have a way to override the default to not apply `conanbuild`. Look at the test.
**FIXME**: I made this in develop2 but it should be done in "develop" I guess.

### DIFF
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -18,7 +18,7 @@ def run_in_windows_bash(conanfile, command, cwd=None, env=None):
     from conan.tools.env import Environment
     from conan.tools.env.environment import environment_wrap_command
     """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
-    if env:
+    if env and env != ["conanbuild"]:
         # Passing env invalidates the conanfile.environment_scripts
         env_win = [env] if not isinstance(env, list) else env
         env_shell = []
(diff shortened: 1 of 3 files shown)


conans/client/subsystems.py
```diff
@@ -18,7 +18,7 @@ def run_in_windows_bash(conanfile, command, cwd=None, env=None):
     from conan.tools.env import Environment
     from conan.tools.env.environment import environment_wrap_command
     """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
-    if env:
+    if env and env != ["conanbuild"]:
```
memsharded: This should be ``env != "conanbuild"``, specially in the default case?

conans/test/integration/environment/test_env.py
```diff
@@ -572,3 +572,44 @@ def package_info(self):
...
+           """)
+    client = TestClient()
+    client.save({"pkg.py": conanfile})
+    client.run("create pkg.py --name pkg --version 1.0")
+
+    # consumer
+    conanfile = textwrap.dedent(r"""
+               import os
+               from conan import ConanFile
+               from conan.tools.env import Environment
```
memsharded: Environment import not necessary

memsharded (approved):
Yes, it makese sense to change that default to allow the possibility of defining that no env-script file should be used at all.

lasote:
Opened in develop https://github.com/conan-io/conan/pull/11014

PR #11014, opened 2022-04-12
self.run(env="conanbuild") instead of "None"

lasote:
Changelog: omit
Docs: https://github.com/conan-io/docs/pull/2516

SAME AS https://github.com/conan-io/conan/pull/11011 BUT IN DEVELOP BRANCH

Related https://github.com/conan-io/conan/issues/11009
If we would like to skip the "conanbuild" for any reason to be applied at some self.run execution (sounds legit use case) we don't have a way to override the default to not apply conanbuild. Look at the test.

### DIFF
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -17,7 +17,7 @@ def run_in_windows_bash(conanfile, command, cwd=None, env=None):
     from conan.tools.env import Environment
     from conan.tools.env.environment import environment_wrap_command
     """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
-    if env:
+    if env and env != "conanbuild":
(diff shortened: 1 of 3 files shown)


conans/client/subsystems.py
```diff
@@ -17,7 +17,7 @@ def run_in_windows_bash(conanfile, command, cwd=None, env=None):
     from conan.tools.env import Environment
     from conan.tools.env.environment import environment_wrap_command
     """ Will run a unix command inside a bash terminal It requires to have MSYS2, CYGWIN, or WSL"""
-    if env:
+    if env and env != "conanbuild":
```
memsharded: I still don't fully understand this. If the user explicitly pass ``self.run(env=None)``, maybe ``conanbuild.xxx`` shouldnt be applied, even inside the subsystem?

memsharded (approved):

PR #11017, opened 2022-04-13
conan cache path command

lasote:
Changelog: Feature: New command "conan cache path" to obtain the path for the "exports", "exports_sources", "sources", "build" and "package".

Close https://github.com/conan-io/conan/issues/10942


conans/client/conanfile/package.py
```diff
@@ -47,6 +48,9 @@ def run_package_method(conanfile, package_id, hook_manager, conanfile_path, ref)
 
     prev = manifest.summary_hash
     scoped_output.info("Created package revision %s" % prev)
+    pref = PkgReference(ref, package_id)
+    pref.revision = prev
+    scoped_output.success("Full package reference: {}".format(pref.repr_notime()))
```
lasote, line 53: I really missed the full reference to copy-paste for obtaining the folders

conans/client/installer.py
```diff
@@ -359,7 +359,7 @@ def _handle_node_build(self, package, pkg_layout):
             # but better make sure here, and be able to report the actual folder in case
             # something fails)
             node.conanfile.folders.set_base_package(pkg_layout.package())
-            self._out.info("Package folder %s" % node.conanfile.package_folder)
+            node.conanfile.output.info("Package folder %s" % node.conanfile.package_folder)
```
lasote, line 362: Now it is scoped and looks a bit better.

conans/test/utils/tools.py
```diff
@@ -706,6 +706,11 @@ def created_package_revision(self, ref):
                                str(self.out)).group(1)
         return package_id
 
+    def created_package_reference(self, ref):
```
lasote, line 709: You can call this in the `TestClient` object to obtain the `PkgReference` just created.

memsharded (approved):

conans/test/integration/command_v2/test_cache_path.py
```diff
@@ -0,0 +1,91 @@
...
+    t.run("cache path patata/1.0#123123", assert_error=True)
+    assert "ERROR: No entry for recipe 'patata/1.0#123123'" in t.out
+
+    # JSON output
+    t.run("cache path {} --folder package --format json".format(pref.repr_notime()))
+    data = json.loads(t.stdout)
+    path = t.cache.pkg_layout(pref).package()
+    assert data["ref"] == pref.repr_notime()
+    assert data["folder"] == "package"
+    assert data["path"] == path
```
franramirez688, line 105: I think it could be fine to split this test a little bit:

```python
def test_cache_path_regular():
   # ...

def test_cache_path_regular_errors():
   # ...

def test_cache_path_regular_json_output_format():
    # ...
```
I think it's nice to see a more fine-grained test name when it fails.

franramirez688 (approved):

czoido:
I would prefer that doing something like: `conan cache path zlib/1.2.11#065588f5155f3d95d91233053068eca1` the command returns `/Users/user/.conan2/p/c95cc530f8fdb849` instead of `/Users/user/.conan2/p/c95cc530f8fdb849/e`
I don't see a reason to default to the export folder, I think that folder is the most expected result
or making the `--folder` argument mandatory could be other solution. But I'm fine merging this as is.

czoido (approved):

lasote:
I still think that returning `/Users/user/.conan2/p/c95cc530f8fdb849` is not a good idea because it means nothing in the conan model. These subfolders could be anywhere else and not even share a parent folder. Also, the subfolder names are kind of random for the user. The "export" folder IS a thing, the "source" folder IS a thing, a folder that arbitrarily contains both of them... is not a thing.

conans/test/integration/command_v2/test_cache_path.py
lasote: I changed it, now it is more fine-grained and less readable hahaha.

Issue #11013, opened 2022-04-12
[bug] Characters are not escaped in pkg config generators


StellaSmith:
https://github.com/conan-io/conan/blob/6739257c4e83943d736aa3eea9544281dce03c2e/conans/client/generators/pkg_config.py#L118-L122

see conan-io/conan-center-index#10257

lasote:
Hi, from [here](https://github.com/conan-io/conan-center-index/pull/10257) `self.cpp_info.defines.append("IMGUI_USER_CONFIG='\"imgui_user_config.h\"'")` you are escaping the value of the define, in the generator we cannot do that as the define is a whole string unless we probe that escaping everything also works, something like this?:


```
lines.append("Cflags: %s" % _concat_if_not_empty( 
     [include_dirs_flags, 
      cpp_info.cxxflags, 
      cpp_info.cflags, 
      ['-D"%s"' % d for d in cpp_info.defines]])) 
```

Another alternative could be replacing `"` with `\"` or stuff like that, but it is difficult to say what would work, we would need to investigate creating a package with a similar "define" and reusing with Meson and CMake.

StellaSmith:
I think that doing something like the following could work:
```py
lines.append("Cflags: %s" % _concat_if_not_empty( 
     [include_dirs_flags, 
      cpp_info.cxxflags, 
      cpp_info.cflags, 
      ["-D%s" % shlex.quote(d) for d in cpp_info.defines]])) 
```

lasote:
thanks, we will make sure with a couple of test with Meson and CMake.

lasote:
Let's try something like https://github.com/conan-io/conan/pull/11039

czoido:
Hi @StellaSmith,

Could you have a try to the changes in #11073 to check if that solves your problem.

You could try them installing conan from the PR branch using: 

`pip install git+https://github.com/czoido/conan.git@escape_pkg_config`

PR #11039, opened 2022-04-18
Escape quotes for XCodeDeps generator

czoido:
Changelog: Fix: Escape quotes in XCodeDeps generator.
Docs: omit

Testing the [imgui recipe](https://github.com/conan-io/conan-center-index/blob/6a06d9628bba096497e77d93065715a672cea382/recipes/imgui/all/conanfile.py#L73) from Conan Center looks like XCode won't accept the defines generated in the xcconfig files because quote character is not properly escaped.

Related to: https://github.com/conan-io/conan/issues/11013

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -146,11 +146,11 @@ class XcodeDeps(object):
             'system_libs': " ".join("-l{}".format(sys_lib) for sys_lib in cpp_info.system_libs),
             'frameworksdirs': " ".join('"{}"'.format(p) for p in cpp_info.frameworkdirs),
             'frameworks': " ".join("-framework {}".format(framework) for framework in cpp_info.frameworks),
-            'definitions': " ".join(cpp_info.defines),
-            'c_compiler_flags': " ".join(cpp_info.cflags),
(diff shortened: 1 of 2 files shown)


StellaSmith:
I would recommend using `shlex.quote`

memsharded:
> I would recommend using shlex.quote

But this is for an Xcode build file, not for terminal:

```
// Definition of Conan variables for {{name}}
        CONAN_{{name}}_BINARY_DIRECTORIES{{condition}} = {{bin_dirs}}
        CONAN_{{name}}_C_COMPILER_FLAGS{{condition}} = {{c_compiler_flags}}
```

Are the xcode build scripts variable values supposed to be escaped following the shell escaping?

StellaSmith:
I thought it worked the same way as a .pc file, just copy pasting the variable into a command line

memsharded:
> I thought it worked the same way as a .pc file, just copy pasting the variable into a command line

I think Xcode is a more advanced and complex build system and does many other things, beyond what a .pc /pkg-config does, that was the reason of my question. Maybe the best is to keep the current proposal of just scaping the quotes, and not the full shlex.quote alternative.

memsharded (approved):

PR #11055, opened 2022-04-20
self.export_source_folder (based on layout.py)

### LINKED ISSUE #10643: [feature] [2.0] exports_sources management
Currently, by default, in the cache, we are copying the `source folder` to the `build folder` unless the `no_copy_source=True`, but _always_, the `self.source_folder` and `self.build_folder` points to the right location.

It might be worth it to explore if the `exports_sources` could have a similar mechanism. Maybe, by default they could be copied to the `base source folder` but unless `self.no_copy_exports_sources=True`.  And of course, a `self.exports_sources_folder` will point to the right location. 

**Note**: Maybe we should start by declaring the `self.exports_sources_folder` variable pointing to the `folders.base_source` to have a "pointer" for the user that, for example, wants to know where are the patches exported and then implement the "optional" copy mechanism, because at the end of the day, typically if you export all the sources, you don't want to skip the copy and if you are just exporting some source files, the copy is cheap so the benefit might not worth it.

lasote:
Changelog: Fix: Renamed `self.base_source_folder` to `self.export_source_folder`. That variable was introduced to reference the folder where the `export_sources` are. Currently, they are copied to the `source` folder but might be changed in the future to avoid copying them, so `self.export_source_folder` will always point to the folder containing the `exports_sources`.
Docs: https://github.com/conan-io/docs/pull/2514

Similar than https://github.com/conan-io/conan/pull/11053
Close https://github.com/conan-io/conan/issues/10643

Warn: The `update_conandata` check is not possible anymore with this implementation.

### DIFF
--- a/conan/tools/files/conandata.py
+++ b/conan/tools/files/conandata.py
@@ -11,8 +11,6 @@ def update_conandata(conanfile, data):
     this only works for updating the conandata on the export() method, it seems it would
     be plain wrong to try to change it anywhere else
     """
-    if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
-        raise ConanException("The 'update_conandata()' can only be used in the 'export()' method")
     path = os.path.join(conanfile.export_folder, "conandata.yml")
     if os.path.exists(path):
         conandata = load(path)
--- a/conan/tools/files/conandata.py
+++ b/conan/tools/files/conandata.py
@@ -11,6 +11,8 @@ def update_conandata(conanfile, data):
     this only works for updating the conandata on the export() method, it seems it would
     be plain wrong to try to change it anywhere else
     """
+    if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
+        raise ConanException("The 'update_conandata()' can only be used in the 'export()' method")
     path = os.path.join(conanfile.export_folder, "conandata.yml")
     if os.path.exists(path):
         conandata = load(path)
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -57,7 +57,6 @@ class Folders(object):
         self._base_generators = output_folder or base_folder
         self._base_imports = output_folder or base_folder
 
-        self._base_export = output_folder or base_folder
         self._base_export_sources = output_folder or base_folder
 
     @property
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -137,8 +137,16 @@ class Folders(object):
     def set_base_imports(self, folder):
         self._base_imports = folder
 
+    @property
+    def base_export(self):
+        return self._base_export
+
     def set_base_export(self, folder):
         self._base_export = folder
+
+    @property
+    def base_export_sources(self):
+        return self._base_export_sources
 
     def set_base_export_sources(self, folder):
         self._base_export_sources = folder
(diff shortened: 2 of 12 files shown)


conans/test/integration/conanfile/conan_data_test.py
```diff
@@ -298,22 +298,6 @@ def source(self):
         assert "pkg/0.1: 0.1-url: myurl!!" in c.out
         assert "pkg/0.1: 0.2-url: new!!" in c.out
 
-    def test_conandata_update_error(self):
```
lasote: cannot be checked anymore

conan/tools/files/conandata.py
```diff
@@ -11,8 +11,6 @@ def update_conandata(conanfile, data):
     this only works for updating the conandata on the export() method, it seems it would
     be plain wrong to try to change it anywhere else
     """
-    if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
```
memsharded: This, we can still keep the ``if conanfile.export_folder is None``, if we invalidate it just after the ``export()`` method, can't we?

conans/model/layout.py
```diff
@@ -131,3 +136,9 @@ def base_imports(self):
 
     def set_base_imports(self, folder):
         self._base_imports = folder
+
+    def set_base_export(self, folder):
+        self._base_export = folder
+
+    def set_base_export_sources(self, folder):
+        self._base_export_sources = folder
```
memsharded, line 151: Seems a getter for this, so we avoid the above (in class ConanFile) ``return self.folders._base_export_sources``, accesing a private member.

Or maybe avoid the setters, getters, make them public attributes?

memsharded (commented):
I think your approach is cleaner than mine, because it is a full implementation (solving my open TODOs) following layout. I will close my PR.

conan/tools/files/conandata.py
lasote: I think it is not that easy, currently (in this PR), the `conanfile.export_folder` is available in many other places, because [this](https://github.com/conan-io/conan/pull/11055/files#diff-c2ca7643fc907db00b41a346805bc9926f9a77adeaefebb10f42d5c528544536R107-R108) and [this](https://github.com/conan-io/conan/pull/11055/files#diff-c15f4aacf65aa58b31a041bac63e4dd7337355d9ec44a8f32b2a5815d5eb5ea9R834-R835) because we want to read them pointing to the "base source", so we don't know if it is set because you are in the "export" method or not.

memsharded: I don't fully get it. ``conanfile.export_folder`` is defined just before calling ``conanfile.export()`` method, and removed after it. Not used anywhere else. The ``export_folder`` shouldnt be pointing to anywhere else in any other method, because it is only relevant to the recipe, and the recipe is no longer copied to the "source" folder in the cache.  What am I missing?

conans/model/layout.py
lasote: I added the getters. Could be done public too, doesn't look very useful but either critical.

conan/tools/files/conandata.py
lasote: Ok, if we don't want that variable to exist in any other place I guess it can be removed and the check restored.

memsharded (approved):
Clean and clear, looks ready to go!

PR #11057, opened 2022-04-20
Cmake complex strings

### LINKED ISSUE #11043: [bug] Support complex strings as definitions in CMakeToolchain

As per https://github.com/conan-io/conan/issues/10929, CMake now requires `add_compile_definitions()` instead of `add_definitions()` but Conan still does not support correct formatting of the definition construction when strings are complex and contain spaces and/or escaped string quotes `\"`.

The jinja template needs updating to check if the `value` of the item that is set via `CMakeToolchain.preprocessor_definitions[key] = value` is complex and adapt the resulting string that is sent to CMake's `add_compile_definitions()` accordingly.

I am currently working around this via an updated jinja template for CMakeToolchian to remove the wrapped string quotes and wrap the whole definition as a string to be able to have my value correctly interpreted.
_(note: example shown only for plain non-config/generator expressions versions of the preprep definitions)_

``` python
cmaketoolchain = CMakeToolchain(self)
# for Conan < 1.48.0
cmaketoolchain._template = cmaketoolchain._template.replace("add_definitions(-D{{ it }}={{ value }})", "add_compile_definitions(\"{{ it }}={{ value[1:-1] }}\")")
# for Conan >= 1.48.0
cmaketoolchain._template = cmaketoolchain._template.replace("add_definitions({{ it }}={{ value }})", "add_compile_definitions(\"{{ it }}={{ value[1:-1] }}\")")
```

To be able to use it as follows:
```
api_def_value = "extern \\\"C\\\""
cmaketoolchain.preprocessor_definitions["MY_API"] = api_def_value
```

That creates correct definition on (osx) clang compiler:
```
c++ -DMY_API="extern \"C\""  ....
```

czoido:
Changelog: Fix: Quote `add_compile_definitions` correctly in CMakeToolchain.
Changelog: Fix: Escape quotes in definitions in CMakeToochain.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/11043

This PR does two things, first quoting the contents of add_compile_definitions following [these recommendations](https://crascit.com/2022/01/25/quoting-in-cmake/).
And also escapes the quotes in the defines so that when generating the toolchain the escaped quotes are correctly preserved.

Looks like CMakeDeps already handles this, so `ConanFile.cpp_info.defines` should not be a problem but I'll add tests and try to homogenize.

TODO: As this is becoming a common pattern in all the generators we should try to homogenize this across all the Conan codebase

### DIFF
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -61,11 +61,11 @@ class CMakeToolchain(object):
             {% for it, values in var_config.items() %}
                 {% set genexpr = namespace(str='') %}
                 {% for conf, value in values -%}
-                set(CONAN_DEF_{{ it }} "{{ value }}")
+                set(CONAN_DEF_{{ conf }}{{ it }} "{{ value }}")
                 {% endfor %}
                 {% for conf, value in values -%}
                     {% set genexpr.str = genexpr.str +
-                                          '$<IF:$<CONFIG:' + conf + '>,${CONAN_DEF_' + it|string + '},' %}
+                                          '$<IF:$<CONFIG:' + conf + '>,${CONAN_DEF_' + conf|string + it|string + '},' %}
                     {% if loop.last %}{% set genexpr.str = genexpr.str + '""' -%}{%- endif -%}
                 {% endfor %}
                 {% for i in range(values|count) %}{% set genexpr.str = genexpr.str + '>' %}
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -49,7 +49,7 @@ class Variables(OrderedDict):
         for config, data in self._configuration_types.items():
             for key, var in data.items():
                 if isinstance(var, six.string_types):
-                    data[key] = '"{}"'.format(str(var).replace('"', '\\"'))
+                    data[key] = str(var).replace('"', '\\"')
 
 
 class CMakeToolchain(object):
@@ -61,8 +61,11 @@ class CMakeToolchain(object):
             {% for it, values in var_config.items() %}
                 {% set genexpr = namespace(str='') %}
                 {% for conf, value in values -%}
+                set(CONAN_DEF_{{ it }} "{{ value }}")
+                {% endfor %}
+                {% for conf, value in values -%}
                     {% set genexpr.str = genexpr.str +
-                                          '$<IF:$<CONFIG:' + conf + '>,' + value|string + ',' %}
+                                          '$<IF:$<CONFIG:' + conf + '>,${CONAN_DEF_' + it|string + '},' %}
                     {% if loop.last %}{% set genexpr.str = genexpr.str + '""' -%}{%- endif -%}
                 {% endfor %}
                 {% for i in range(values|count) %}{% set genexpr.str = genexpr.str + '>' %}
@@ -70,7 +73,7 @@ class CMakeToolchain(object):
                 {% if action=='set' %}
                 set({{ it }} {{ genexpr.str }} CACHE STRING
                     "Variable {{ it }} conan-toolchain defined")
-                {% elif action=='add_compile_definitions' %}
+                {% elif action=='add_compile_definitions' -%}
                 add_compile_definitions({{ it }}={{ genexpr.str }})
                 {% endif %}
             {% endfor %}
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -42,6 +42,14 @@ class Variables(OrderedDict):
                 ret.setdefault(k, []).append((conf, v))
         return ret
 
+    def quote_preprocessor_strings(self):
+        for key, var in self.items():
+            if isinstance(var, six.string_types):
+                self[key] = str(var).replace('"', '\\"')
+        for config, data in self._configuration_types.items():
+            for key, var in data.items():
+                if isinstance(var, six.string_types):
+                    data[key] = str(var).replace('"', '\\"')
 
 class CMakeToolchain(object):
 
@@ -62,7 +70,7 @@ class CMakeToolchain(object):
                 set({{ it }} {{ genexpr.str }} CACHE STRING
                     "Variable {{ it }} conan-toolchain defined")
                 {% elif action=='add_compile_definitions' %}
-                add_compile_definitions("{{ it }}={{ genexpr.str }}")
+                add_compile_definitions({{ it }}={{ genexpr.str }})
                 {% endif %}
             {% endfor %}
         {% endmacro %}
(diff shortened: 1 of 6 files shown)


conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -36,7 +36,7 @@ def generate(self):
                 tc.variables.release["MYVAR_CONFIG"] = "MYVAR_RELEASE"
                 tc.variables.debug["MYVAR2_CONFIG"] = "MYVAR2_DEBUG"
                 tc.variables.release["MYVAR2_CONFIG"] = "MYVAR2_RELEASE"
-                tc.preprocessor_definitions["MYDEFINE"] = "MYDEF_VALUE"
+                tc.preprocessor_definitions["MYDEFINE"] = "\"MYDEF_VALUE\""
                 tc.preprocessor_definitions["MYDEFINEINT"] = 42
                 tc.preprocessor_definitions.debug["MYDEFINE_CONFIG"] = "MYDEF_DEBUG"
```
SeanSnyders: Maybe add escaped string test here as well to test correct CMake generator expression building and interpretation with complex strings (spaces and escaped strings).

SeanSnyders (changes requested):
The changes looks workable overall (judging from the tests), but I do not have a knowledge in how to use this PR version of Conan specifically on my project.

Also, not sure where in this PR the code (tests or otherwise) is to ensure `ConanFile.cpp_info.defines` are tested accordingly for the test cases I provided.

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -60,17 +60,20 @@ class CMakeToolchain(object):
         {% macro iterate_configs(var_config, action) %}
             {% for it, values in var_config.items() %}
                 {% set genexpr = namespace(str='') %}
+                {% for conf, value in values -%}
+                set(CONAN_DEF_{{ it }} "{{ value }}")
```
czoido: I had to separate variable definitions for generator expressions to support escaping, it was not working otherwise. Check: https://crascit.com/2022/01/25/quoting-in-cmake/

SeanSnyders:
@czoido Can you point me to tests where these complex string situations are tested for `ConanFile.cpp_info.defines` ?
This is an important bit as part of fixing it for `CMakeToolchain`, some of these complex definitions need to be there when consumed by clients.

czoido:
> @czoido Can you point me to tests where these complex string situations are tested for `ConanFile.cpp_info.defines` ? This is an important bit as part of fixing it for `CMakeToolchain`, some of these complex definitions need to be there when consumed by clients.

Hi @SeanSnyders,
I have added a test: `test_cmakedeps_cppinfo_complex_strings` for that

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -36,10 +36,10 @@ def generate(self):
                 tc.variables.release["MYVAR_CONFIG"] = "MYVAR_RELEASE"
                 tc.variables.debug["MYVAR2_CONFIG"] = "MYVAR2_DEBUG"
                 tc.variables.release["MYVAR2_CONFIG"] = "MYVAR2_RELEASE"
-                tc.preprocessor_definitions["MYDEFINE"] = "MYDEF_VALUE"
+                tc.preprocessor_definitions["MYDEFINE"] = "\"MYDEF_VALUE\""
```
memsharded, line 64: I am not sure about this, that means that I, as a user, always need to double quote strings, or they will fail? That will surely bring some headaches to users that don't do that in the first place. Even if it is in the docs, it is very easy to omit the double quotes. I think that as long as the input is a string, the quotes should happen in the Conan internals as needed.

czoido: But how can we know that the value must be quoted? Can we distinguish between these two?

```
#define MYDEFINE MYDEF_VALUE
#define MYDEFINE "MYDEF_VALUE_STR"
```
I thin that you should care about what you want to define, and you could also use: 

```
tc.preprocessor_definitions["MYDEFINE"] = '"MYDEF_VALUE"'
```

memsharded: I see.
Then I guess you could do:
```python
tc.preprocessor_definitions["MYDEFINEINT"] = "42"
```
And the result will be identical, mapping to an integer in the C++ definition?

memsharded (commented):
I think it is worth adding some more characters in the tests, like $ and { and } and \ itself. If we want to define: ``#define MYVAR "Hello $ you { } \ bye"``, will it work? I think CMake might still struggle with the scaping of this.

czoido:
> I think it is worth adding some more characters in the tests, like $ and { and } and \ itself. If we want to define: `#define MYVAR "Hello $ you { } \ bye"`, will it work? I think CMake might still struggle with the scaping of this.

👍 I'll add support for more characters

conans/test/functional/toolchains/cmake/test_cmake.py
czoido: > I see. Then I guess you could do:
> 
> ```python
> tc.preprocessor_definitions["MYDEFINEINT"] = "42"
> ```
> 
> And the result will be identical, mapping to an integer in the C++ definition?

Yes, that should define an integer in that case

SeanSnyders: How do I then define a string with only numerical contents? Like `"12345"` ?
Escaping it `"\"12345\""` ?

So, in other words, no python type inference to "do the right thing"?
I have to admit, setting a def a a literal number like "42" (in python) and that becomes an integer, is a bit weird, no?

czoido: You could do 

`tc.preprocessor_definitions["MYDEFINEINT"] = 42`

that would be equivalent to:

`tc.preprocessor_definitions["MYDEFINEINT"] = "42"`

I don't think we should try from Conan to interpret if the definition is an int or not, because you could want to define something like this:

`tc.preprocessor_definitions["OTHERINT"] = "MYDEFINEINT"`

and you want that to be declared like `#define OTHERINT MYDEFINEINT` and not like `#define OTHERINT "MYDEFINEINT"`

memsharded (approved):

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -60,17 +60,20 @@ class CMakeToolchain(object):
...
+                                          '$<IF:$<CONFIG:' + conf + '>,${CONAN_DEF_' + conf|string + it|string + '},' %}
                     {% if loop.last %}{% set genexpr.str = genexpr.str + '""' -%}{%- endif -%}
                 {% endfor %}
                 {% for i in range(values|count) %}{% set genexpr.str = genexpr.str + '>' %}
                 {% endfor %}
                 {% if action=='set' %}
                 set({{ it }} {{ genexpr.str }} CACHE STRING
                     "Variable {{ it }} conan-toolchain defined")
-                {% elif action=='add_compile_definitions' %}
+                {% elif action=='add_compile_definitions' -%}
```
memsharded, line 76: The case below is not quoting the ``add_compile_definitions()`` as in the other cases, not sure if intentional, or it is missing something (is the part that sets values per Debug/Release config)

czoido: Yes, that was because with the generator expression the quoting is different, there are tests that cover that it works correctly, so it should be ok

memsharded (approved):

PR #11063, opened 2022-04-20
Add PKG_CONFIG_PATH to AutotoolsToolchain

### LINKED ISSUE #10639: [feature] Autotools + PkgConfigDeps - a canonical way to set PKG_CONFIG_PATH?
from https://github.com/conan-io/conan-center-index/pull/9488

if I use Autotools in conjunction with PkgConfigDeps, who is responsible to set PKG_CONFIG_PATH?
right now I had to add:
```
        tc = AutotoolsDeps(self)
        tc.environment.define("PKG_CONFIG_PATH", self.source_folder)  # is it really needed?
        tc.generate()
```
which is kinda ugly
if it doesn't make sense to fill it out of the box, it's probably worth to at least document, as in reality, almost every autotools-based recipe discover its dependencies via `pkg-config`.

czoido:
Changelog: Feature: Append the folder where the Conan generators were installed to `PKG_CONFIG_PATH` in AutotoolsToolchain.
Docs: https://github.com/conan-io/docs/pull/2509

Closes: https://github.com/conan-io/conan/issues/10639

### DIFF
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -153,7 +153,7 @@ class AutotoolsToolchain:
         env.append("CXXFLAGS", self._filter_list_empty_fields(self.cxxflags))
         env.append("CFLAGS", self._filter_list_empty_fields(self.cflags))
         env.append("LDFLAGS", self._filter_list_empty_fields(self.ldflags))
-        env.append_path("PKG_CONFIG_PATH", self._conanfile.generators_folder)
+        env.prepend_path("PKG_CONFIG_PATH", self._conanfile.generators_folder)
 
         return env
 
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -153,7 +153,7 @@ class AutotoolsToolchain:
         env.append("CXXFLAGS", self._filter_list_empty_fields(self.cxxflags))
         env.append("CFLAGS", self._filter_list_empty_fields(self.cflags))
         env.append("LDFLAGS", self._filter_list_empty_fields(self.ldflags))
-        env.define("PKG_CONFIG_PATH", self._conanfile.generators_folder)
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -153,6 +153,7 @@ def environment(self):
         env.append("CXXFLAGS", self._filter_list_empty_fields(self.cxxflags))
         env.append("CFLAGS", self._filter_list_empty_fields(self.cflags))
         env.append("LDFLAGS", self._filter_list_empty_fields(self.ldflags))
+        env.define("PKG_CONFIG_PATH", self._conanfile.generators_folder)
```
memsharded: I am not sure we want the ``autotoolstoolchain`` to completely re-define the PKG_CONFIG_PATH. In case there are some other .pc files in some other place and the user has PKG_CONFIG_PATH defined, we want to respect that. So it seems ``generators_folder`` should probably be pre-pended to make the Conan generated files a priority, but not remove the existing value.

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -153,6 +153,7 @@ def environment(self):
         env.append("CXXFLAGS", self._filter_list_empty_fields(self.cxxflags))
         env.append("CFLAGS", self._filter_list_empty_fields(self.cflags))
         env.append("LDFLAGS", self._filter_list_empty_fields(self.ldflags))
+        env.append_path("PKG_CONFIG_PATH", self._conanfile.generators_folder)
```
memsharded:
```suggestion
        env.prepend_path("PKG_CONFIG_PATH", self._conanfile.generators_folder)
```

Shouldn't this path go first, so it has priority?

czoido:
```suggestion
        env.append_path("PKG_CONFIG_PATH", self._conanfile.generators_folder)
```
Yes... 🤦

memsharded (approved):

lasote (approved):

PR #11066, opened 2022-04-20
Feature/running inside bash

### LINKED ISSUE #10900: [question/feature] conan.tools.env » VirtualRunEnv does not generate sh scripts on windows
Hello,

I'm trying to switch to the new [VirtualRunEnv](https://docs.conan.io/en/latest/reference/conanfile/tools/env/virtualrunenv.html) instead of [virtualrunenv](https://docs.conan.io/en/latest/reference/generators/virtualrunenv.html), but I've noticed that the new generator does not generate `sh` scripts for windows when conan command is being called from `bash` (git bash or MSYS2)

This was not the case before with the old virtualrunenv generator. 
Not sure if this is an intended behavior or a bug, but it would be really nice to support sh scripts for windows as well.

Thanks.

memsharded:
Changelog: Feature: New ``tools.microsoft.bash:active`` conf to indicate already running inside bash (but not targeting subsystem)

AFTER https://github.com/conan-io/conan/pull/11061

Close https://github.com/conan-io/conan/issues/10900

Open issues:
- We don't want to run inside "bash", so we need to make sure ``run_in_windows_bash()`` is not being triggered
- At the moment it requires defining the ``win_bash=True`` in the recipe, probably shouldn't be necessary. Actually the ``win_bash=True`` in recipes I find it a bit confusing every time I work on this. In this case, it will probably fir the ``run_in_windows_bash()``
- This is very difficult to test, as it would mean running the test from bash already, doesn't seem possible in the test suite.


memsharded:
Superseded by https://github.com/conan-io/conan/pull/12178/

Issue #11062, opened 2022-04-20
[feature] profile plugin check cppstd correctness


lasote:
The default profile plugin should validate the specified `compiler.cppstd` to check if the specified compiler (with the specified compiler version) support it. The default profile can be found at the `_load_profile_plugin` function at `profile_loader.py`.

PR #11077, opened 2022-04-21
profile plugin check cppstd 

lasote:
Changelog: Feature: Improved profile plugin to check if the specified compiler supports the specified cppstd.

Close https://github.com/conan-io/conan/issues/11062

### DIFF
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -74,8 +74,8 @@ def profile_plugin(profile):
 def _check_correct_cppstd(settings):
     from conan.tools.scm import Version
     def _error():
-        from conan.errors import ConanInvalidConfiguration
-        raise ConanInvalidConfiguration("The provided compiler.cppstd is not supported with the specified compiler")
+        from conan.errors import ConanException
+        raise ConanException("The provided compiler.cppstd is not supported with the specified compiler")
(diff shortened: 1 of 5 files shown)


conans/client/profile_loader.py
```diff
@@ -59,17 +59,57 @@ def get_default_build(self):
...
+                settings["compiler.runtime_type"] = runtime
+            except ConanException:
+                pass
+    _check_correct_cppstd(settings)
+
+def _check_correct_cppstd(settings):
+    from conan.tools.scm import Version
+    def _error():
+        from conan.errors import ConanInvalidConfiguration
+        raise ConanInvalidConfiguration("The provided compiler.cppstd is not supported with the specified compiler")
```
memsharded: I think this can be a ConanException. We don't want to check compatibility, if a profile is broken/badly defined, it should error, report it as a broken profile and stop?

memsharded (approved):
Looking good!

conans/client/profile_loader.py
lasote: Yes, I had the same doubt. Look like look are losing some potentially valuable information in the return code but we can improve later if needed. Let's raise ConanException

memsharded (approved):

czoido (approved):

PR #11083, opened 2022-04-23
Minor tweaks to markdown generator output

chausner:
Changelog: omit
Docs: omit

This suggests some minor tweaks to the generated Markdown output of the markdown generator.


conans/client/generators/markdown.py
```diff
@@ -152,9 +152,9 @@
     $ cmake . -DCMAKE_TOOLCHAIN_FILE=cmake-build-release/conan_toolchain.cmake
     $ cmake --build .
 
-    # for Windows and Visual Studio 2017
+    # for Windows and Visual Studio 2022
```
memsharded: As VS 2022 is not still active in ConanCenter, there are no binaries for it, not sure if we should keep a previous VS version?

memsharded (approved):

conans/client/generators/markdown.py
chausner: I see. I have reverted this change.

lasote (approved):

PR #11088, opened 2022-04-25
Fix binaryDir CMakePresets.json

### LINKED ISSUE #11084: [bug] CMakePresets.json contains binaryDir set to null
When using the new `CMakePresets.json` output generated by CMakeToolchain, the attribute `binaryDir` is set to `null` which is invalid according to the official [CMakePresets schema](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#schema). Also, when I try to load the CMakePresets.json with the CMake Tools extension of VS Code, it fails due to that line.

  * Compiler+version: MSVC 17.1.4
  * Conan version: 1.47
  * Python version: 3.9.6

Run:
```shell
conan install soxr/0.1.3@ --generator CMakeDeps --generator CMakeToolchain --build=missing
```

The following `CMakePresets.json` is generated:
```json
{
    "version": 3,
    "cmakeMinimumRequired": {
        "major": 3,
        "minor": 15,
        "patch": 0
    },
    "configurePresets": [
        {
            "name": "default",
            "displayName": "Default Config",
            "description": "Default configure using 'Visual Studio 17 2022' generator",
            "generator": "Visual Studio 17 2022",
            "cacheVariables": {},
            "toolchainFile": "[...]\\conan_toolchain.cmake",
            "binaryDir": null
        }
    ],
    "buildPresets": [
        {
            "name": "default",
            "configurePreset": "default"
        }
    ],
    "testPresets": [
        {
            "name": "default",
            "configurePreset": "default"
        }
    ]
}
```

/cc @lasote

lasote:
Changelog: Bugfix: The `binaryDir` field at `CMakePresets.json` is not set if the `conanfile.build_folder` is not available, avoiding a `null` value breaking the specification.
Docs: omit

Close https://github.com/conan-io/conan/issues/11084

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -26,6 +26,9 @@ def _contents(conanfile, toolchain_file, cache_variables, generator):
            }]
           }
     if conanfile.build_folder:
+        # If we are installing a ref: "conan install <ref>", we don't have build_folder, because
+        # we don't even have a conanfile with a `layout()` to determine the build folder.
+        # If we install a local conanfile: "conan install ." with a layout(), it will be available.
         ret["configurePresets"][0]["binaryDir"] = conanfile.build_folder
     return ret
 
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -6,26 +6,28 @@ from conans.util.files import save, load
 
 
 def _contents(conanfile, toolchain_file, cache_variables, generator):
-    return {"version": 3,
-            "cmakeMinimumRequired": {"major": 3, "minor": 15, "patch": 0},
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/presets.py
```diff
@@ -6,26 +6,28 @@
...
-            "testPresets": [{
+           }],
+           "testPresets": [{
               "name": "default",
               "configurePreset": "default"
-            }]
-            }
+           }]
+          }
+    if conanfile.build_folder:
```
memsharded, line 28: A clarifying note saying in which cases the ``build_folder`` is ``None``, but still this presets is being generated would help to understand this condition.

memsharded (approved):

Issue #10921, opened 2022-03-29
Continue compatibility.py features


memsharded:
Ok, lets move forward this for alpha.6. Still a couple of open items, but seems a good first step.

- Need to modernize ``Settings``, hopefully it can mimic a bit better what ``Options`` did, and not separate the ``values`` from it
- The current default proposed ``compatibility.py`` is just a POC, it only covers applications, and is not complete. It needs more work.
- Lets try to define ``self.settings`` and ``self.options`` in ``validate()`` and ``package_id()`` to be the ``info`` values. That would simplify recipes syntax and make it way more intuitive.
- Probably need helpers to query possible values in ``settings.yml`` for ``compatibility.py``
- It would be very useful if removed settings/options in ``configure()`` would also be filtered without needed to explicit the removal again in ``package_id()``.

I will also draft a proposal for the Tribe to discuss this.

_Originally posted by @memsharded in https://github.com/conan-io/conan/issues/10241#issuecomment-1080551991_

lasote:
I will continue with the `compatibility.py` improvements for apps and cppstd, trying to provide a way to check the values of the declared settings.

memsharded:
Closed by https://github.com/conan-io/conan/pull/11095

Issue #6157, opened 2019-11-28
Revise Cppstd compatibility model


memsharded:
There are recipes out there, even merge to Conan-center-index [like this](https://github.com/conan-io/conan-center-index/blob/master/recipes/absent/all/conandata.yml)

Absent is activating the c++17 standard in the [CmakeLists.txt](https://github.com/rvarago/absent/blob/master/CMakeLists.txt), but it is not defined in any input. 

The recipe is doing tricks as:

```python
def _supports_cpp17(self):
        supported_compilers = [("gcc", "7"), ("clang", "5"), ("apple-clang", "10"), ("Visual Studio", "15.7")]
        compiler = self.settings.compiler
        version = Version(self.settings.compiler.version)
        return any(compiler == e[0] and version >= e[1] for e in supported_compilers)
```
To check if the compiler provides capabilities for that C++ standard. But cppstd is not there at all.

It can be deduced that in reality different C++ standards for the same compiler version are binary compatible:
- https://stackoverflow.com/questions/46746878/is-it-safe-to-link-c17-c14-and-c11-objects
- https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2019

In that case, our model would be inaccurate. Maybe we should:
- Reconsider the compiler.cppstd setting? Deprecate it? Option?
- Provide ``package_info()`` helpers to define compatibility between them, or do erasure of the input setting.
- Consider what the default would be in Conan 2.0

Adnn:
We are revising our approach to Conan-ify our repository (and find out we have been doing several things wrong, see the #6176 you already linked ; )

In our opinion, the C++ standard should likely still introduce package-variability by defaut (i.e., change the package ID). Even though it was very instructive to read the linked SO answer regarding the major 3 compilers, it might not be future proof to remove `compiler.cppstd` completely:
* We all hope Conan will root deeply in the C++ community for decades to come, and even if today the flags from 11 to 17 don't break the ABI, this is just a "nice feature of those specific implementation", not something that the standard guarantees.  This would have to be revisited for each new standard-switch.
* More importantly, it really feels like Conan managed to be tool-chain agnostic. And I assume for some guys using more niche compilers, they cannot be sure about the ABI stability across C++ versions.

It feels like Conan should map its model on the standard, not on specific implementations, so we think your model is accurate here : )

[...]

PR #11095, opened 2022-04-25
Simplified app compatibility and defined default cppstd compatibility

lasote:
Close https://github.com/conan-io/conan/issues/6157
Close https://github.com/conan-io/conan/issues/10921

### DIFF
--- a/conans/client/graph/compatibility.py
+++ b/conans/client/graph/compatibility.py
@@ -23,6 +23,8 @@ def compatibility(conanfile):
 
 
 default_cppstd_compat = """
+from conan.tools.build import supported_cppstd
+
 
 def cppstd_compat(conanfile):
     # It will try to find packages with all the cppstd versions
@@ -33,17 +35,13 @@ def cppstd_compat(conanfile):
     if not compiler or not compiler_version or not cppstd:
         return []
     base = dict(conanfile.settings.values_list)
-
-    coordinates = "compiler.{}.cppstd".format(compiler)
-    conanfile.settings.get_definition()
-    cppstd_possible_values = conanfile.settings.compiler.cppstd.get_definition()
+    cppstd_possible_values = supported_cppstd(compiler, compiler_version)
     ret = []
     for _cppstd in cppstd_possible_values:
         if _cppstd is None or _cppstd == cppstd:
             continue
         configuration = base.copy()
         configuration["compiler.cppstd"] = _cppstd
-        # FIXME: Not checking max supported by the compiler
         ret.append({"settings": [(k, v) for k, v in configuration.items()]})
 
     return ret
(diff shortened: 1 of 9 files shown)


conans/client/graph/compatibility.py
```diff
@@ -10,72 +10,84 @@
...
+
+    compiler = conanfile.settings.get_safe("compiler")
+    compiler_version = conanfile.settings.get_safe("compiler.version")
+    cppstd = conanfile.settings.get_safe("compiler.cppstd")
+    if not compiler or not compiler_version or not cppstd:
+        return []
+    base = dict(conanfile.settings.values_list)
+
+    coordinates = "compiler.{}.cppstd".format(compiler)
+    cppstd_possible_values = conanfile.settings.get_definition_values(coordinates)
```
lasote: This is a bit ugly

conans/client/graph/compatibility.py
```diff
@@ -10,72 +10,84 @@
...
-    combinations = list(product(*factors))
-    for combination in combinations:
-        result.append({"settings": combination})
-    return result
+            configuration["compiler.runtime"] = "dynamic"
+            configuration["compiler.runtime_type"] = configuration["build_type"]
+
+
+    configuration["compiler.cppstd"] = conanfile.settings.get_safe("compiler.cppstd") or \
+                                       default_cppstd(compiler, compiler_version)
```
lasote, line 87: We need to get the default cppstd of the compiler after all, right?

conans/model/settings.py
```diff
@@ -298,3 +298,25 @@ def dumps(self):
         """
         return "\n".join(["%s=%s" % (field, value)
                           for (field, value) in self.values_list])
+
+    def get_definition_values(self, name):
```
lasote: Don't love this.

memsharded: Why you don't love this? The name of the method? the implementation?

conans/client/graph/compatibility.py
memsharded: I think we want here the information per compiler & compiler.version, so we know that gcc 5 will not support c++17, c++20, even if they are declared as possible values in ``settings.yml``.

memsharded (approved):
I think it is looking good, to talk a couple of minor issues about implementation and "uglyness"

conans/model/settings.py
lasote: I think both hahah

memsharded: Probably because we only want the internal definition?
I mean, instead of ``settings.get_definition_values("compiler.gcc.cppstd")``, which is the current approach, something more like: ``settings.get_definitions()`` return the barebones ``dict``, so: 
```
definitions = settings.get_definitions()
cppstd = definitions["gcc"]["cppstd"]  # maybe protect with ``try-except KeyError``
```
Maybe we also want to have a direct access to the ``settings_yaml`` definition, and not via de ``conanfile``? Just wondering, probably not.

lasote: > settings.get_definitions()

Yes, that would definitely be better.

memsharded (approved):
Excellent

PR #11098, opened 2022-04-25
Define CMP0091 CMake policy directly in Presets, and add a check in toolchain.

memsharded:
Changelog: Fix: using CMAKE_PROJECT_INCLUDE instead of presets to define variables that don't work in toolchains
Docs: Omit

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
@@ -5,6 +5,8 @@ import textwrap
 import pytest
 
 from conan.tools.cmake.presets import load_cmake_presets
+from conan.tools.microsoft.visual import vcvars_command
+from conans.client.tools import replace_in_file
 from conans.model.ref import ConanFileReference
 from conans.test.assets.cmake import gen_cmakelists
 from conans.test.assets.genconanfile import GenConanfile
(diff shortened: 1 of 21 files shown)


memsharded:
@czoido I had to revert moving the CMAKE_SH and CMAKE_MAKE_PROGRAM from Presets to CMAKE_PROJECT_INCLUDE, as it was failing, it seems the CMAKE_PROJECT_INCLUDE is already too late for CMAKE_SH

lasote (approved):
Solve conflicts.

czoido:
I think this is not still solving the issue, I'll try to add a failing test

memsharded:
> I think this is not still solving the issue, I'll try to add a failing test

I did some manual testing, it seemed to work, but yes, definitely better add some test. If you can't add it, let me know and I will do it.

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -436,3 +438,30 @@ def test_autolink_pragma_without_components(self):
...
+    client.run_command(dumpbind_cmd)
+    assert "LIBCMTD" in client.out
+
+    client.save({}, clean_first=True)
+    # Setting an older cmake_minimum_required in the CMakeLists fails, will link
+    # against the default debug runtime (MDd->MSVCRTD), not against MTd->LIBCMTD
+    client.run("new hello/0.1 --template=cmake_lib")
+    replace_in_file(os.path.join(client.current_folder, "CMakeLists.txt"),
+                    'cmake_minimum_required(VERSION 3.15)',
+                    'cmake_minimum_required(VERSION 3.1)', output=client.out)
```
memsharded: Isn't it possible that 3.1 is too old, and something else is broken? I think I tried with 3.10 or something like that.

czoido: It also fails for me with 3.14, I'll update the test to use a newer version

memsharded: Testing things. Can reproduce, it is not working, and it seems the reason is that CMAKE_PROJECT_INCLUDE needs to be _BEFORE ,and that is not working if ``cmake_minimum_required(VERSION 3.14)``. UDPATE: ACtually ``CMAKE_PROJECT_INCLUDE_BEFORE`` doesn't seem to work at all when defined in toolchain. I will keep investigating.

memsharded: I made it work by dropping CMAKE_PROJECT_INCLUDE, and defining the policy directly in the CMakePresets file.

Apparently, when interacting with cmake_minimum_required(), the policy would need to be set before ``project()`` and that means ``CMAKE_PROJECT_INCLUDE_BEFORE``. But it happens that the toolchain is not "activated" until after the ``project()``, so all attempts to define it in the toolchain are doomed. It is not possible.

czoido: Maybe we should check the policy in the toolchain and fail if not set to `NEW` so that we make sure that if someone calls the toolchain directly with `cmake -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake` the runtime type is properly set?

memsharded: I will try to add it, lets see if possible.

memsharded: https://github.com/conan-io/conan/pull/11098/commits/3719c2c5fa521b67f817bec0efb91bd149b1abdd seems to do the job

czoido (approved):

franramirez688 (approved):

PR #11103, opened 2022-04-26
CMakePresets.json multi build configurations

### LINKED ISSUE #11085: [bug] Generated CMakePresets.json does not set build configuration
The generated `CMakePresets.json` does not work out-of-the-box with the CMake Tools extension of VS Code.
It does not specify the build configuration which, according to the CMakePresets.json is optional, but VS Code seems to require it.
If it is not specified, it assumes the build configuration to be "Debug" which leads to build errors when `conan install` was previously called with `-s build_type=Release`.

I am not sure if the problem lies on the Conan side or the CMake Tools extension.
If it is something that needs to be improved or fixed in VS Code, I can also report it to them.

  * Compiler+version: MSVC 17.1.4
  * Conan version: 1.47
  * Python version: 3.9.6

1. Run:
   ```shell
   conan install soxr/0.1.3@ --generator CMakeDeps --generator CMakeToolchain --build=missing
   ```
2. Remove "binaryDir: null` from the generated `CMakePresets.json` to workaround #11084 
3. Add a dummy `CMakeLists.txt`:
   ```cmake
   cmake_minimum_required(VERSION 3.22)
   project(soxrtest CXX)

   find_package(soxr)
   add_executable(soxrtest soxrtest.cpp)
   target_link_libraries(soxrtest PRIVATE soxr::core)
   ```
4. Add a dummy `soxrtest.cpp`:
   ```cpp
   #include <soxr.h>

   int main() {
       return 0;
   }
   ```
5. Open folder in VS Code with the CMake Tools extension installed
6. Attempt to build the CMake target. It will fail (`soxr.h` not found) and a warning is displayed.

The CMake Tools extension fails to compile and prints the following warning:
```
[cpptools] The build configurations generated do not contain the active build configuration. Using "Debug" for CMAKE_BUILD_TYPE instead of "null" to ensure that IntelliSense configurations can be found
```

### Workarounds

I have found the following two workarounds:
- Running `conan install` with `-s build_type=Debug` makes the compilation succeed, the warning remains, though.
- Adding `"configuration": "Release"` to `CMakePresets.json` fixes compilation and makes the warning disappear:
  ```json
      "buildPresets": [
        {
            "name": "default",
            "configurePreset": "default",
            "configuration": "Release"
        }
    ],
  ```

Maybe the CMakeToolchain generator could detect which build types have been installed via `conan install` and generate appropriate build presets in `CMakePresets.json` with the `configuration` attribute set?

/cc @lasote

lasote:
Changelog: Feature: The `CMakePresets.json` (file generated when specified the `CMakeToolchain` generator) is appended with a new `buildPresets` after a `conan install` with a different `build_type` when using a multiconfiguration generator like `Visual Studio`.
Docs: https://github.com/conan-io/docs/pull/2512

Close https://github.com/conan-io/conan/issues/11085

PENDING TO TRY IT WITH VISUAL STUDIO CODE ON WINDOWS.

### DIFF
--- a/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
+++ b/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
@@ -1,3 +1,4 @@
+import json
 import os
 import platform
 import textwrap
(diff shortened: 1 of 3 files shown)


lasote:
@memsharded Could you try it on windows?

memsharded (approved):

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
```diff
@@ -354,3 +355,48 @@ def layout(self):
...
+    assert presets["buildPresets"][3]["configuration"] == "MinSizeRel"
+
+    # Repeat one
+    client.run("install mylib/1.0@ -g CMakeToolchain -s build_type=Debug --profile:h=profile")
+    client.run("install mylib/1.0@ -g CMakeToolchain -s build_type=Debug --profile:h=profile")
+    assert len(presets["buildPresets"]) == 4
+    assert presets["buildPresets"][0]["configuration"] == "Release"
+    assert presets["buildPresets"][1]["configuration"] == "Debug"
+    assert presets["buildPresets"][2]["configuration"] == "RelWithDebInfo"
+    assert presets["buildPresets"][3]["configuration"] == "MinSizeRel"
```
franramirez688, line 402: I don't know if it's needed but should we run a `client.run("create . mylib/1.0@ -s build_type=RelWithDebInfo --profile:h=profile")`, for example, and check that it's running ok?

franramirez688 (approved):

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
lasote: Well, this test is an empty conanfile that can run in any system, and actually, the `buildPresets` are not parsed/used by the cmake build helper, they are only for the consumers using the `CMakePresets.json`

franramirez688: Got it

PR #11111, opened 2022-04-27
New CMake imported target shared package_type based

lasote:
Following https://github.com/conan-io/conan/pull/10997 but using the develop2 package_type to deduce the "shared", otherwise or if the DLL is not located, it defaults to the UNKNOWN behavior.
TODO: Still, 100% based on `conanfile.package_type` no properties for changing that behavior.


conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -64,9 +64,37 @@ def template(self):
                    string(REGEX REPLACE "[^A-Za-z0-9.+_-]" "_" _LIBRARY_NAME ${_LIBRARY_NAME})
                    set(_LIB_NAME CONAN_LIB::${package_name}_${_LIBRARY_NAME}${config_suffix})
                    if(NOT TARGET ${_LIB_NAME})
+                       if(library_type STREQUAL "SHARED" AND is_host_windows)
+                         set(CMAKE_FIND_LIBRARY_SUFFIXES .dll ${CMAKE_FIND_LIBRARY_SUFFIXES})
```
lasote, line 68: Otherwise cmake only find static extensions.

czoido (approved):

conan/tools/cmake/cmakedeps/templates/target_data.py
```diff
@@ -58,6 +58,15 @@ def context(self):
...
                 "dependency_find_modes": dependency_find_modes}
 
+    @property
+    def cmake_package_type(self):
+        return {"shared-library": "SHARED",
+                "static-library": "STATIC"}.get(str(self.conanfile.package_type), "UNKNOWN")
+
+    @property
+    def is_host_windows(self):
+        return self.conanfile.settings.get_safe("os") == "Windows"
```
memsharded:
```suggestion
        # to account for all WindowsStore, WindowsCE and Windows OS in settings
        return "Windows" in self.conanfile.settings.get_safe("os", "")
```

memsharded (approved):

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -64,9 +64,37 @@ def template(self):
...
                        # Create a micro-target for each lib/a found
-                       add_library(${_LIB_NAME} UNKNOWN IMPORTED)
-                       set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_FOUND_LIBRARY})
+                       if (CONAN_SHARED_FOUND_LIBRARY)
+                           add_library(${_LIB_NAME} SHARED IMPORTED)
+                           set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_SHARED_FOUND_LIBRARY})
+                           set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_IMPLIB ${CONAN_FOUND_LIBRARY})
+                           conan_message(DEBUG "Found DLL and STATIC at ${CONAN_SHARED_FOUND_LIBRARY}, ${CONAN_FOUND_LIBRARY}")
+                        else()
+                           if(CONAN_DLL_NOT_FOUND)
```
franramirez688: I'm trying to understand the if-else structure...

```cmake
    if (CONAN_SHARED_FOUND_LIBRARY)
        # ....
    else()
       if(CONAN_DLL_NOT_FOUND)
           # If we haven't found the DLL, fallback to old behavior: UNKNOWN target linking the static
           add_library(${_LIB_NAME} UNKNOWN IMPORTED)
           conan_message(DEBUG "DLL library not found, creating UNKNOWN IMPORTED target")
       else()
           # library_type can be STATIC, still UNKNOWN (if no package type available in the recipe) or SHARED (no windows)
           add_library(${_LIB_NAME} ${library_type} IMPORTED)
           conan_message(DEBUG "Created target ${_LIB_NAME} ${library_type} IMPORTED")
       endif()
    endif()
```
Then if `library_type == "STATIC"` or not, it does not matter because `CONAN_DLL_NOT_FOUND` will be always `0`, won't it?

UPDATED: what I mean is that this code will never be executed:
```cmake
       else()
           # library_type can be STATIC, still UNKNOWN (if no package type available in the recipe) or SHARED (no windows)
           add_library(${_LIB_NAME} ${library_type} IMPORTED)
           conan_message(DEBUG "Created target ${_LIB_NAME} ${library_type} IMPORTED")
       endif()
```

lasote: I think, if `library_type != "SHARED"` or "not windows" it will enter there.

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -64,9 +64,37 @@ def template(self):
...
+                         if(NOT CONAN_SHARED_FOUND_LIBRARY)
+                           conan_message(STATUS "Cannot locate shared library: ${_LIBRARY_NAME}")
+                           set(CONAN_DLL_NOT_FOUND 1)
+                         endif()
+                       endif()
+
                        # Create a micro-target for each lib/a found
-                       add_library(${_LIB_NAME} UNKNOWN IMPORTED)
-                       set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_FOUND_LIBRARY})
+                       if (CONAN_SHARED_FOUND_LIBRARY)
```
memsharded: Isn't there a risk that ``CONAN_DLL_NOT_FOUND`` will take a value from a previous library (in the for-loop), then keep it to 1 all the time?

conan/tools/cmake/cmakedeps/templates/macros.py
memsharded: I agree that I also struggled a bit to follow the logic. Maybe it is worth to have 1 single level ``if-else`` structure starting from the ``if(library_type STREQUAL "SHARED" AND is_host_windows)``, even if there are more ``add_library()`` calls?

franramirez688: Yeah, I think so.

lasote: I think now is much clear, thanks for the concerns and suggestions.

franramirez688 (approved):
Great!

memsharded (approved):

PR #11113, opened 2022-04-27
[WIP] Relocatable libs for Macos

czoido:

Will fail in Linux because missing `libtool` -> updating the `ci-functional` image

This PR tries to address several problems with shared libraries in Macos.
- In develop2 the `autotools.install()` is made over a temporal folder, so the entry `LC_ID_DYLIB` in the dylib file is set to that temporal folder. That makes the configure fail when trying to build a consumer that requires that library because it first loads the dylib file and expects that its own location is in that temporal folder that no longer exists. For fixing this, after doing the `autotools.install()` all the found dylib files in the libdirs folder are patched with `install_name_tool` to set it to the value: `@rpath/<libname>.dylib`
- Added to the XcodeDeps that if one dependency is `shared=True` then the RPATH for those libraries are set in the consumer binary to the `libdirs` folders in the cache. This is necessary so that when you run `autotools.configure()` the library is found.

### DIFF
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -93,7 +93,7 @@ class Autotools(object):
         #  because there's no package_folder until the package step
         self.configure()
         self.make(target="install")
-        if self._conanfile.settings.get_safe("os") == "Macos" and self._conanfile.options.get_safe("shared", True):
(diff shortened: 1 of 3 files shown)


czoido:
Closing this as will be merged with develop branch

PR #11114, opened 2022-04-27
Relocatable libs for Macos

czoido:
Changelog: Fix: Make shared libraries build with Autotools relocatable in Macos by patching the install name (LC_ID_DYLIB) and setting to `@rpath/libname.dylib`.
Docs: https://github.com/conan-io/docs/pull/2518

Currently shared libraries build with Autotools in Macos are not relocatable. This is because they include the `LC_ID_DYLIB` field in the binary that hardcodes the folder where they were installed. Even if that library is found, at configure time the build will fail if they are not located in that folder. This means that just creating a package, uploading to a server and trying to consume that binary from other client will fail.

There could be two solutions for this:
- Setting the install_name linker flag like: `-Wl,-install_name,@rpath/foo.dylib`. The problem is that we can't do that in the toolchain because at the point the toolchain is generated we don't know the name of the libraries to be built.
- Patching the `LC_ID_DYLIB` field with `install_name_tool` in all the collected shared libraries that are installed. After doing the `autotools.install()` all the found dylib files in the libdirs folder are patched with `install_name_tool` to set it to the value: `@rpath/<libname>.dylib`. 

Also, to make this work we added to the XcodeDeps that if one dependency is `shared=True` then the RPATH for those libraries are set in the consumer binary to the `libdirs` folders in the cache. This is necessary so that when you run `autotools.configure()` the library is found.

### DIFF
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -93,7 +93,7 @@ class Autotools(object):
         #  because there's no package_folder until the package step
         self.configure()
         self.make(target="install")
-        if self._conanfile.settings.get_safe("os") == "Macos" and self._conanfile.options.get_safe("shared", True):
+        if self._conanfile.settings.get_safe("os") == "Macos" and self._conanfile.options.get_safe("shared", False):
             self._fix_macos_install_paths()
 
     def autoreconf(self, args=None):
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -6,6 +6,7 @@ from conans.client.subsystems import subsystem_path, deduce_subsystem
 from conans.client.build import join_arguments
 from conans.tools import args_to_string
 from conan.tools.files import chdir
+from conans.util.runners import check_output_runner
 
 
 class Autotools(object):
@@ -63,30 +64,33 @@ class Autotools(object):
         command = join_arguments([make_program, target, str_args, jobs])
         self._conanfile.run(command)
 
-    @staticmethod
-    def _collect_shared_lib_files(folder):
-        result = []
-        files = os.listdir(folder)
-        for f in files:
-            full_path = os.path.join(folder, f)
-            name, ext = os.path.splitext(f)
-            if ext == ".dylib" and not os.path.islink(full_path):
-                result.append(f)
-        return result
+    def _fix_osx_shared_install_name(self):
 
-    def _fix_macos_install_paths(self):
-        """
-        fix LC_ID_DYLIB in Macos
-        """
+        def _osx_collect_dylibs(lib_folder):
+            return [f for f in os.listdir(lib_folder) if f.endswith(".dylib")
+                    and not os.path.islink(os.path.join(lib_folder, f))]
+
+        def _fix_install_name(lib_name, lib_folder):
+            command = "install_name_tool -id @rpath/{} {}".format(lib_name, os.path.join(lib_folder,
+                                                                                         lib_name))
+            self._conanfile.run(command)
+
+        def _is_modified_install_name(lib_name, lib_folder):
+            """
+            Check that the user did not change the default install_name using the install_name
+            linker flag in that case we do not touch this field
+            """
+            command = "otool -D {}".format(os.path.join(lib_folder, lib_name))
+            out = check_output_runner(command).strip().split(":")[1]
+            return False if str(os.path.join(lib_folder, shared_lib)) in out else True
+
         libdirs = getattr(self._conanfile.cpp.package, "libdirs")
         for folder in libdirs:
             full_folder = os.path.join(self._conanfile.package_folder, folder)
-            shared_libs = self._collect_shared_lib_files(full_folder)
+            shared_libs = _osx_collect_dylibs(full_folder)
             for shared_lib in shared_libs:
-                command = "install_name_tool -id @rpath/{} {}".format(shared_lib,
-                                                                      os.path.join(full_folder,
-                                                                                   shared_lib))
-                self._conanfile.run(command)
+                if not _is_modified_install_name(shared_lib, full_folder):
+                    _fix_install_name(shared_lib, full_folder)
 
     def install(self):
         # FIXME: we have to run configure twice because the local flow won't work otherwise
@@ -94,7 +98,7 @@ class Autotools(object):
         self.configure()
         self.make(target="install")
         if self._conanfile.settings.get_safe("os") == "Macos" and self._conanfile.options.get_safe("shared", False):
-            self._fix_macos_install_paths()
+            self._fix_osx_shared_install_name()
 
     def autoreconf(self, args=None):
         command = ["autoreconf"]
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -78,14 +78,14 @@ class Autotools(object):
         """
         fix LC_ID_DYLIB in Macos
         """
-        shared_libs = []
-        package_folder = self._conanfile.package_folder
         libdirs = getattr(self._conanfile.cpp.package, "libdirs")
         for folder in libdirs:
-            full_folder = os.path.join(package_folder, folder)
-            shared_libs.extend(self._collect_shared_lib_files(full_folder))
+            full_folder = os.path.join(self._conanfile.package_folder, folder)
+            shared_libs = self._collect_shared_lib_files(full_folder)
             for shared_lib in shared_libs:
-                command = "install_name_tool -id @rpath/{} {}".format(shared_lib, os.path.join(full_folder, shared_lib))
+                command = "install_name_tool -id @rpath/{} {}".format(shared_lib,
+                                                                      os.path.join(full_folder,
+                                                                                   shared_lib))
                 self._conanfile.run(command)
 
     def install(self):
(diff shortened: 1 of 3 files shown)


conan/tools/gnu/autotoolsdeps.py
```diff
@@ -8,19 +8,33 @@ class AutotoolsDeps:
...
             # "targets" with autotools
             ret.merge(dep_cppinfo)
         return ret
 
+    def _rpaths_flags(self):
+        flags = []
+        for dep in self.ordered_deps:
+            if dep.options.get_safe("shared", False):
+                for libdir in dep.cpp_info.libdirs:
+                    flags.extend(["-Wl,-rpath -Wl,{}".format(libdir)])
```
memsharded: Why is this necessary now that it was not necessary before? 
Also this applies to all platforms, not only Macos, as the PR title suggest.

conan/tools/gnu/autotools.py
```diff
@@ -63,11 +63,38 @@ def make(self, target=None):
         command = join_arguments([make_program, target, str_args, jobs])
         self._conanfile.run(command)
 
+    @staticmethod
+    def _collect_shared_lib_files(folder):
```
memsharded: As it collects only for Macos, probably call it ``_osx_collect...`` or something like that?
As it can probably be shortened to:

```python
def _collect_lib_files(folder):
    files = [os.path.join(folder, f) for f in os.listdir(folder)]
    return [f for f in files if f.endswith(".dylib") and not os.path.islink(f)]
```
Maybe make it internal to the ``_fix_macos_install_paths``?

conan/tools/gnu/autotools.py
```diff
@@ -63,11 +63,38 @@ def make(self, target=None):
...
+    def _fix_macos_install_paths(self):
+        """
+        fix LC_ID_DYLIB in Macos
+        """
+        libdirs = getattr(self._conanfile.cpp.package, "libdirs")
+        for folder in libdirs:
+            full_folder = os.path.join(self._conanfile.package_folder, folder)
+            shared_libs = self._collect_shared_lib_files(full_folder)
+            for shared_lib in shared_libs:
+                command = "install_name_tool -id @rpath/{} {}".format(shared_lib,
```
memsharded: Is ``install_name_tool`` always installed? is it possible that this fail if it doesnt and there is some scenario that wouldn't require this?

memsharded (commented):
I think this might have some minor risks, hardcoding the ``install_name_tool`` rpath fix in the helper could break in some cases? Are there any legit use that would not require this to be done by Conan? Is it possible that some users implement the rpaths things inside their Makefiles directly?

czoido:
> I think this might have some minor risks, hardcoding the `install_name_tool` rpath fix in the helper could break in some cases? Are there any legit use that would not require this to be done by Conan? Is it possible that some users implement the rpaths things inside their Makefiles directly?

Yes, the user might change the `LC_ID_DYLIB` using the `install_name` linker flag. I was thinking that we may check the value of the id first with `otool` and change it just we case we are sure that it was not changed by the user's Makefile flags

conan/tools/gnu/autotools.py
czoido: It's installed with the XCode command line tools with more tools like `otool`, so I think it should not be a problem

conan/tools/gnu/autotoolsdeps.py
czoido: Yes, I think we can omit this in Linux, it won't have the same problem as in Macos.

conan/tools/gnu/autotools.py
```diff
@@ -63,11 +64,41 @@ def make(self, target=None):
...
+        def _osx_collect_dylibs(lib_folder):
+            return [f for f in os.listdir(lib_folder) if f.endswith(".dylib")
+                    and not os.path.islink(os.path.join(lib_folder, f))]
+
+        def _fix_install_name(lib_name, lib_folder):
+            command = "install_name_tool -id @rpath/{} {}".format(lib_name, os.path.join(lib_folder,
+                                                                                         lib_name))
+            self._conanfile.run(command)
+
+        def _is_modified_install_name(lib_name, lib_folder):
```
czoido, line 78: Maybe it would be better to provide a conf for this?

SSE4:
@SpaceIm what do you think about this one?
I personally don't like that `install` silently manipulates artifacts (as my expectation as a user that `install` just runs `make install`).
if it has to be done, maybe it needs its own method?

czoido:
> @SpaceIm what do you think about this one? I personally don't like that `install` silently manipulates artifacts (as my expectation as a user that `install` just runs `make install`). if it has to be done, maybe it needs its own method?

@SSE4 I don't know if there's other way to approach this problem. In develop branch the problem is not visible but it is in the develop2 branch because we run `make()` and `install()` in a `tmp` folder that is later removed and after that we move all the binaries to the final location. That makes that the **dylib** ends with an `install_name` pointing to the `tmp` folder. It would also happen if you upload the binaries compiled in one path and then you download them in other machine. The only other solution that I know (but is equivalent to this) is adding `install_name` to the linker flags but that should be done in the _AutotoolsToolchain_ and I thought that was a dirtier solution. Do you have any ideas? @SSE4 @SpaceIm ?

SSE4:
> > @SpaceIm what do you think about this one? I personally don't like that `install` silently manipulates artifacts (as my expectation as a user that `install` just runs `make install`). if it has to be done, maybe it needs its own method?
> 
> @SSE4 I don't know if there's other way to approach this problem. In develop branch the problem is not visible but it is in the develop2 branch because we run `make()` and `install()` in a `tmp` folder that is later removed and after that we move all the binaries to the final location. That makes that the **dylib** ends with an `install_name` pointing to the `tmp` folder. It would also happen if you upload the binaries compiled in one path and then you download them in other machine. The only other solution that I know (but is equivalent to this) is adding `install_name` to the linker flags but that should be done in the _AutotoolsToolchain_ and I thought that was a dirtier solution. Do you have any ideas? @SSE4 @SpaceIm ?

even if this folder isn't a temporary one, it doesn't necessarily exist for the consumer - as it will be under a user-specific cache folder. it basically makes such packages non-relocatable (and useless).

`-install_name` flag might be a more reliable (and proper) solution. why?
- current code only scans for `.dylib`, but extension isn't mandatory. it might be `.bundle`, `.framework`, `.kext`, `.plugin`
- `install_name_tool` might not be always available (e.g. in case of cross-compiling?)
- I am not entirely sure, but executables might need patching as well?

czoido:
> > > @SpaceIm what do you think about this one? I personally don't like that `install` silently manipulates artifacts (as my expectation as a user that `install` just runs `make install`). if it has to be done, maybe it needs its own method?
> > 
> > 
> > @SSE4 I don't know if there's other way to approach this problem. In develop branch the problem is not visible but it is in the develop2 branch because we run `make()` and `install()` in a `tmp` folder that is later removed and after that we move all the binaries to the final location. That makes that the **dylib** ends with an `install_name` pointing to the `tmp` folder. It would also happen if you upload the binaries compiled in one path and then you download them in other machine. The only other solution that I know (but is equivalent to this) is adding `install_name` to the linker flags but that should be done in the _AutotoolsToolchain_ and I thought that was a dirtier solution. Do you have any ideas? @SSE4 @SpaceIm ?
> 
> even if this folder isn't a temporary one, it doesn't necessarily exist for the consumer - as it will be under a user-specific cache folder. it basically makes such packages non-relocatable (and useless).
> 
> `-install_name` flag might be a more reliable (and proper) solution. why?
> 
> * current code only scans for `.dylib`, but extension isn't mandatory. it might be `.bundle`, `.framework`, `.kext`, `.plugin`
> * `install_name_tool` might not be always available (e.g. in case of cross-compiling?)
> * I am not entirely sure, but executables might need patching as well?

Ok, that makes sense, I think I can try again the install_name approach.

czoido:
@SSE4 I just recalled why I did the patching instead of using the linker flag, the problem is that you don't have access to the information of which libs you are building when generating the toolchain, you don't know their names until the `package_info` so I can't pass that linker flag, that's why I decided to patch the binaries as the only option. 

- `install_name_tool` I think should be always avaiable as it is part of the Xcode toolchain, as otool is. I think it will be there even if you cross-compile, because it's part of the same toolchain that is cross-compiling.
- I don't think that `LC_ID_DYLIB` is set for executables, so there should be no problem with that.

czoido:
I have added a test (`test_autotools_relocatable_libs_darwin_downloaded`) that failed in develop that a library built with autotools in Macos, then uploaded to a server and downloaded from other client fails when building.

memsharded (approved):

lasote (approved):

SpaceIm:
> Also, to make this work we added to the XcodeDeps that if one dependency is shared=True then the RPATH for those libraries are set in the consumer binary to the libdirs folders in the cache. This is necessary so that when you run autotools.configure() the library is found.

Completly missed this part.

@czoido but by doing this it pollutes `LC_RPATH` of shared libs & executables with hardcoded full paths (of build machine) of shared dependencies.
There is a workaround which avoids this pollution while allowing configure step to succeed on macOS: generate VirtualRunEnv in build scope in case of native build, it's even in the template of autotools based recipes of conan-center: https://github.com/conan-io/conan-center-index/blob/8b3c7ac87fccd44d30ae18890823f386d3e12a02/docs/package_templates/autotools_package/all/conanfile.py#L106-L110

EDIT: well forget what I said. I've tested and indeed even VirtualRunEnv in build scope doesn't help on macOS (it's useful on other *nix platforms only), SIP mess again.
I guess the only solution would be to improve `fix_apple_shared_install_name()` to cleanup these LC_RPATH which have absolute paths.

PR #11135, opened 2022-04-29
reviewing lockfiles for 2.0

memsharded:
Review from Tribe proposal https://github.com/conan-io/tribe/pull/34:

- Defined a default "conan.lock" filename
- Whenever the default "conan.lock" file is found in the current folder, it will be used as ``--lockfile=conan.lock``, automatically by commands that implement this argument
- ``conan lock create`` will also default to output "conan.lock", even without ``--lockfile-out`` argument
- Commands that implement a ``--lockfile-out`` will allow to save a lockfile as a result, even if a lockfile was not provided as input
- The logic for strict vs non-strict has changed its default. Now by default, commands are "strict", and the arg ``--lockfile-no-strict`` (arg name not great, we can try to think a better one) will allow non-locked ``requires`` to not fail
- The ``--clean`` argument that was existing only for ``conan lock create`` has been renamed to ``--lockfile-clean``, because it now can be used by the other commands that save a ``--lockfile-out``.

Second round of review:
- The ``conan.lock`` is assumed by default besides the ``conanfile``, not in the ``cwd``
- Renamed ``--lockfile-no-strict`` => ``--lockfile-partial``, it seems a bit more aligned with the actual behavior

### DIFF
--- a/conans/cli/commands/export.py
+++ b/conans/cli/commands/export.py
@@ -1,10 +1,9 @@
 import os
 
-from conans.cli.command import conan_command, COMMAND_GROUPS
-from conans.cli.commands import make_abs_path
+from conans.cli.command import conan_command, COMMAND_GROUPS, OnceArgument
 from conans.cli.commands.install import _get_conanfile_path
-from conans.cli.common import get_lockfile, add_lockfile_args, add_reference_args
-from conans.model.graph_lock import Lockfile
+from conans.cli.common import get_lockfile, add_reference_args
+from conans.cli.output import ConanOutput
 
 
 def common_args_export(parser):
@@ -13,12 +12,16 @@ def common_args_export(parser):
 
 
 @conan_command(group=COMMAND_GROUPS['creator'])
-def export(conan_api, parser, *args, **kwargs):
+def export(conan_api, parser, *args):
     """
     Export recipe to the Conan package cache
     """
     common_args_export(parser)
-    add_lockfile_args(parser)
+    parser.add_argument("-l", "--lockfile", action=OnceArgument,
+                        help="Path to a lockfile.")
+    parser.add_argument("--lockfile-no-strict", action="store_true",
+                        help="Raise an error if some dependency is not found in lockfile")
+
     args = parser.parse_args(*args)
 
     cwd = os.getcwd()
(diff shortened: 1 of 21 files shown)


SSE4:
what's the summary of the review? discussion is pretty long, it's hard to extract the final resolution

memsharded:
> what's the summary of the review? discussion is pretty long, it's hard to extract the final resolution

It is a draft, it will be posted on the Tribe repo when ready, and most likely submitted to the mailing list

conans/cli/commands/export.py
```diff
@@ -12,24 +12,23 @@ def common_args_export(parser):
...
-def export(conan_api, parser, *args, **kwargs):
+def export(conan_api, parser, *args):
     """
     Export recipe to the Conan package cache
     """
     common_args_export(parser)
-    add_lockfile_args(parser)
+    parser.add_argument("-l", "--lockfile", action=OnceArgument,
+                        help="Path to a lockfile.")
+    parser.add_argument("--lockfile-no-strict", action="store_true",
```
lasote: `--lockfile-partial` ?

memsharded: It could be a complete lockfile, but you changed one ``requires = xxx`` in your recipe and use against the lockfile, want to "lock whatever you can, but if there is something you cannot lock, then evaluate it as normal". I am not sure ``-lockfile-partial`` would be closer to that than ``no-strict``

lasote: >  "lock whatever you can, but if there is something you cannot lock, then evaluate it as normal"

So it applies the lockfile partially. Honestly, the `strict` might mean anything.

memsharded: Good for me. @czoido @franramirez688 ? ``--lockfile-partial`` better naming than ``--lockfile-no-strict``?

lasote: I have to say I don't love `--lockfile-partial` either... maybe we need more ideas.

memsharded: I have used ``--lockfile-partial`` I agree it is slightly better than ``--lockfile-strict`` in the meaning.
If anyone else got other ideas, they are welcome.

lasote (approved):

franramirez688 (approved):

czoido (approved):

Issue #11012, opened 2022-04-11
[feature] parse_conanfile at public ConanAPI


lasote:
We are going to need something similar to the old "conan inspect" command. It could be a custom command but we need a public API to call. There are methods in the `conan_api.graph` to load conanfiles but they process the settings and I guess when we inspect attributes we just want the pure python ones. (raw settings dict, for example).

This code might work but it is using internal implementation (`parse_conanfile`):

```python
from conans.cli.command import conan_command, COMMAND_GROUPS
from conans.cli.output import ConanOutput
from conans.client.loader import parse_conanfile


@conan_command(group=COMMAND_GROUPS['creator'])
def inspect(conan_api, parser, *args, **kwargs):
    """
    Return the specified attribute of an attribute
    """
    parser.add_argument("path", help="Path to a folder containing a recipe (conanfile.py)")
    parser.add_argument("attribute", help="Attribute of the conanfile to read")
    args = parser.parse_args(*args)

    _, conanfile = parse_conanfile(args.path)
    ret = getattr(conanfile, args.attribute)
    out = ConanOutput()
    out.write(str(ret))

```

Where it should live (at which subAPI) that function?

memsharded:
I think one of the most important things is to make this command a bit more limited than it used to be. In 1.X, if the ``inspected`` reference doesn't exist in the cache, it will downloaded to be able to read and parse it. We might want to reduce the old command so it never downloads.

It also worked on an ambiguous local path conanfile.py and cache recipe-reference. We probably want to disambiguate that too.

lasote:
Does that mean we are introducing "conan inspect" in the core again? This issue was created to implement something in the Conan API that could make a custom command work, your concerns are about the command.

memsharded:
> Does that mean we are introducing "conan inspect" in the core again? This issue was created to implement something in the Conan API that could make a custom command work, your concerns are about the command.

Yes, my concerns are about the command, or having a new API that mimics the previous command behavior. So I don't want any of those, not the command that can actually download from the server to inspect, nor the API that will download from the server to inspect.

memsharded:
Implemented in https://github.com/conan-io/conan/pull/11136

PR #11136, opened 2022-04-29
'conan inspect' command and 'graph.load_conanfile_class' API

lasote:
close https://github.com/conan-io/conan/issues/11012

`conan inspect path <path>` => With no attribute selection.

### DIFF
--- a/conans/cli/commands/inspect.py
+++ b/conans/cli/commands/inspect.py
@@ -1,0 +1,31 @@
+import json
+
+from conans.cli.command import conan_command, COMMAND_GROUPS
+from conans.cli.output import ConanOutput
+from conans.errors import ConanException
+
+
+def _inspect_json_formatter(data):
+    return json.dumps(data, indent=4)
+
+
+@conan_command(group=COMMAND_GROUPS['creator'], formatters={"json": _inspect_json_formatter})
+def inspect(conan_api, parser, *args, **kwargs):
+    """
+    Returns the specified attribute/s of a conanfile
+    """
+    parser.add_argument("path", help="Path to a folder containing a recipe (conanfile.py)")
+    parser.add_argument("attribute", help="Attribute of the conanfile to read", nargs="+")
+    args = parser.parse_args(*args)
+    out = ConanOutput()
+    conanfile = conan_api.graph.load_conanfile_class(args.path)
+    ret = {}
+    for attr in args.attribute:
+        try:
+            value = getattr(conanfile, attr)
+        except AttributeError:
+            raise ConanException("The conanfile doesn't have a '{}' attribute".format(attr))
+        ret[attr] = value
+        out.writeln("{}: {}".format(attr, value))
+
+    return {"attributes": ret}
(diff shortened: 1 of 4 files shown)


conans/cli/commands/inspect.py
```diff
@@ -0,0 +1,31 @@
...
+    return json.dumps(data, indent=4)
+
+
+@conan_command(group=COMMAND_GROUPS['creator'], formatters={"json": _inspect_json_formatter})
+def inspect(conan_api, parser, *args, **kwargs):
+    """
+    Returns the specified attribute/s of a conanfile
+    """
+    parser.add_argument("path", help="Path to a folder containing a recipe (conanfile.py)")
+    parser.add_argument("attribute", help="Attribute of the conanfile to read", nargs="+")
```
memsharded: Probably try to align this ``attribute`` with the ``--filter`` argument in ``conan graph info``

conans/cli/commands/inspect.py
```diff
@@ -0,0 +1,31 @@
...
+def _inspect_json_formatter(data):
+    return json.dumps(data, indent=4)
+
+
+@conan_command(group=COMMAND_GROUPS['creator'], formatters={"json": _inspect_json_formatter})
+def inspect(conan_api, parser, *args, **kwargs):
+    """
+    Returns the specified attribute/s of a conanfile
+    """
+    parser.add_argument("path", help="Path to a folder containing a recipe (conanfile.py)")
```
memsharded: Only for local conanfiles in user folder, so no possible to inspect one recipe in the cache? Will that need a custom command?

conans/cli/commands/inspect.py
```diff
@@ -0,0 +1,31 @@
...
+from conans.cli.command import conan_command, COMMAND_GROUPS
+from conans.cli.output import ConanOutput
+from conans.errors import ConanException
+
+
+def _inspect_json_formatter(data):
+    return json.dumps(data, indent=4)
+
+
+@conan_command(group=COMMAND_GROUPS['creator'], formatters={"json": _inspect_json_formatter})
```
memsharded: Maybe the inspect is useful as consumer, to know for example which options does a recipe have?

conans/cli/commands/inspect.py
lasote: `conan cache path` that's what we talked

conans/cli/commands/inspect.py
lasote: But a filter is to "discard all" and show only some fields. I'm not sure here we can/want to show a full set of fields from the recipe. Or do we? Which ones?

memsharded: The original ``conan inspect`` listed many fields from the recipe, by default. I think I'd try to align on ``conan graph info`` output?

conans/cli/commands/inspect.py
lasote: Alternatives?

`conan inspect [--path] [--reference] attr1 attr2`  => bit messy

`conan inspect path attr1 attr2` + `conan inspect reference attr1 attr2` => a bit more clean

Do we see `conan inspect` as a `name` that we want to reserve for more use cases like packages?

czoido (approved):

franramirez688 (approved):

memsharded (approved):

PR #11141, opened 2022-05-01
Fix nested Python Popen command

redradist:
Please see the stackoverflow issue:
https://stackoverflow.com/questions/72076014/python-popen-git-is-not-recognized-as-an-internal-or-external-command?noredirect=1#comment127352234_72076014

Also please check Popen documentation:
https://docs.python.org/3/library/subprocess.html#subprocess.Popen

### DIFF
--- a/conans/util/runners.py
+++ b/conans/util/runners.py
@@ -10,7 +10,7 @@ from conans.client.tools.files import load
 from conans.errors import CalledProcessErrorWithStderr
 from conans.util.files import rmdir
 from conans.util.log import logger
-
+import shutil
 
 if getattr(sys, 'frozen', False) and 'LD_LIBRARY_PATH' in os.environ:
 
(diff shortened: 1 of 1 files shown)


redradist:
@memsharded Please, check this PR, this issue happens on Windows with nested Popen command.
To fix it I added `shutil.which`

redradist:
Also I see the following error:
```console
'cmake' is not recognized as an internal or external command,
operable program or batch file.
```
But not found yet place where it happens ...
Also when invoke conan in Popen with create command

memsharded (changes requested):
In order to evaluate this, it is first necessary to:
- Provide a test, in this same PR, as minimal as possible that would fail without the proposed fix.
- Fix the broken tests, until the build is green

Also, the linked SO question as reference in https://stackoverflow.com/questions/72076014/python-popen-git-is-not-recognized-as-an-internal-or-external-command, doesn't seem to be greatly formulated, it is even being downvoted in SO, and it is not being acknowledged as an issue yet there either. So far, I don't think this is a Conan issue that needs to be fixed, there is not enough evidence for that. In order for this to be considered, solid evidence that Python ``Popen(cmd)`` really needs to do a ``which`` to locate ``cmd`` is necessary, but that sounds unlikely.

redradist:
@memsharded I've found that this issue even reproducible with cmake-conan, try the following config on Windows 10 machine:
```cmake
include(${CMAKE_BINARY_DIR}/conan.cmake)

set(CONAN_COMMAND conan)
conan_cmake_autodetect(settings)
conan_cmake_install(PATH_OR_REFERENCE ${CMAKE_CURRENT_LIST_DIR}
                    BUILD missing
                    REMOTE conancenter
                    SETTINGS ${settings})
```
and the following packages:
```console
[requires]
fmt/8.1.1
flatbuffers/2.0.0

[generators]
cmake
```

memsharded:
@redradist 

That is not a minimal reproducible example we can add in the test suite. Do not bring ``conan.cmake`` into this, it shouldn't be necessary to reproduce. 

Furthermore, I have tested your above snippets with minimum modifications (like adding missing ``project()`` and ``cmake_minimum_required()`` and, with `` cmake . -G "Visual Studio 15 2017 Win64" -DCMAKE_BUILD_TYPE=Release`` and it is working without issue, no failure there.

redradist:
> @redradist
> 
> That is not a minimal reproducible example we can add in the test suite. Do not bring `conan.cmake` into this, it shouldn't be necessary to reproduce.
> 
> Furthermore, I have tested your above snippets with minimum modifications (like adding missing `project()` and `cmake_minimum_required()` and, with ` cmake . -G "Visual Studio 15 2017 Win64" -DCMAKE_BUILD_TYPE=Release` and it is working without issue, no failure there.

@memsharded Then it is very strange ...
Probably something on my system ...
This issue reproducible for me only with conan.cmake

redradist:
@memsharded

Closing this PR.
Issue was related to Windows PATH environment variable longer than 8191 character.
Fixed in `conanex` removing duplicate paths, see https://github.com/redradist/conanex/blob/master/conanex/main.py#L23

PR #11148, opened 2022-05-02
Feature/develop2 compatibility always info

memsharded:
- Redefine ``conanfile.settings`` and ``conanfile.options`` just before calling ``package_id()`` and ``validate()``, so they can perform their tasks directly on ``self.settings`` and ``self.options``, avoiding the uglier ``self.info.xxxx``.
- This allows for ``validate()`` to raise exceptions over invalid configurations defined globally in ``compatibility.py`` in the cache
- Minor simplifications like avoiding checking ``callable()`` for every method.


lasote (approved):
Look good! Is anything pending to move if out of "draft"?

memsharded:
> Look good! Is anything pending to move if out of "draft"?

It has some minor migration 1.X -> 2.0 risk, I guess, I wanted some fresh look over it.

conans/client/graph/compatibility.py
```diff
@@ -91,21 +91,30 @@ def __init__(self, cache):
...
         for c in compat_infos:
             conanfile.info = c
+            conanfile.settings = c.settings
+            conanfile.options = c.options
             run_validate_package_id(conanfile)
             pid = c.package_id()
             if pid not in result and not c.invalid:
                 result[pid] = c
+        # Restore the original state
+        conanfile.info = original_info
```
SSE4, line 115: shouldn't it go under the `finally` block, so it always gets properly restored (e.g. in case of exception)?

memsharded: If there is an exception, the program will crash anyways, not necessary.

Issue #10964, opened 2022-04-05
[feature] [UX] Improve `conan new` command


franramirez688:
If you try out a simple `conan new -h`:
```
➜  tttt git:(develop2) conan new -h
usage: conan new [-h] [-d DEFINE] [-f] template

Create a new recipe (with conanfile.py and other files) from a predefined template

positional arguments:
  template              Template name, built-in predefined one or user one

optional arguments:
  -h, --help            show this help message and exit
  -d DEFINE, --define DEFINE
                        Define a template argument as key=value
  -f, --force           Overwrite file if exists
```

So, after that, you try to create  a CMake example:

```
➜  tttt git:(develop2) conan new cmake_lib
Traceback (most recent call last):
  File "/Users/franchuti/develop/conan/conans/cli/cli.py", line 163, in run
    command.run(self._conan_api, self._commands[command_argument].parser, args[0][1:])
  File "/Users/franchuti/develop/conan/conans/cli/command.py", line 157, in run
    info = self._method(conan_api, parser, *args)
  File "/Users/franchuti/develop/conan/conans/cli/commands/new.py", line 43, in new
    template_files = conan_api.new.render(template_files, definitions)
  File "/Users/franchuti/develop/conan/conan/api/subapi/new.py", line 87, in render
    v = Template(v, keep_trailing_newline=True).render(**definitions)
  File "/Users/franchuti/Library/Python/3.9/lib/python/site-packages/jinja2/environment.py", line 1090, in render
    self.environment.handle_exception()
  File "/Users/franchuti/Library/Python/3.9/lib/python/site-packages/jinja2/environment.py", line 832, in handle_exception
    reraise(*rewrite_traceback_stack(source=source))
  File "/Users/franchuti/Library/Python/3.9/lib/python/site-packages/jinja2/_compat.py", line 28, in reraise
    raise value.with_traceback(tb)
  File "<template>", line 5, in top-level template code
  File "/Users/franchuti/Library/Python/3.9/lib/python/site-packages/jinja2/environment.py", line 471, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'name' is undefined

ERROR: 'name' is undefined
```

Obviously, we're missing to specify the name and the version but `[-d DEFINE] ` appears like an optional thing so it's quite confusing:

```
➜  tttt git:(develop2) conan new cmake_lib -d name=hello -d version=0.1
File saved: CMakeLists.txt
File saved: conanfile.py
[...]

PR #11151, opened 2022-05-03
Improve conan new command

czoido:
Now you have to set all the context variables for the templates, If any of the variables is missing it will raise an error like:

```
ERROR: Missing definitions for the template. Required definitions are: 'name', 'version'
```

So I think that is enough and there's no need to add an aditional argument or use the help to output all the variables for the template. If you just call the command without arguments you will see the variables in the error output.

Closes: https://github.com/conan-io/conan/issues/10964

### DIFF
--- a/conan/api/subapi/new.py
+++ b/conan/api/subapi/new.py
@@ -2,7 +2,7 @@ import fnmatch
 import os
 import re
 
-from jinja2 import Template
+from jinja2 import Template, StrictUndefined
 
 from conan.api.subapi import api_method
 from conans.util.files import load
@@ -90,8 +90,8 @@ class NewAPI:
         definitions["conan_version"] = __version__
         definitions["package_name"] = name.replace("-", "_").replace("+", "_").replace(".", "_")
         for k, v in template_files.items():
-            k = Template(k, keep_trailing_newline=True).render(**definitions)
-            v = Template(v, keep_trailing_newline=True).render(**definitions)
+            k = Template(k, keep_trailing_newline=True, undefined=StrictUndefined).render(**definitions)
+            v = Template(v, keep_trailing_newline=True, undefined=StrictUndefined).render(**definitions)
             if v:
                 result[k] = v
         return result
--- a/conan/api/subapi/new.py
+++ b/conan/api/subapi/new.py
@@ -1,5 +1,6 @@
 import fnmatch
 import os
+import re
 
 from jinja2 import Template
 
@@ -85,7 +86,9 @@ class NewAPI:
     @staticmethod
     def render(template_files, definitions):
         result = {}
+        name = definitions.get("name", "Pkg")
         definitions["conan_version"] = __version__
+        definitions["package_name"] = name.replace("-", "_").replace("+", "_").replace(".", "_")
         for k, v in template_files.items():
             k = Template(k, keep_trailing_newline=True).render(**definitions)
             v = Template(v, keep_trailing_newline=True).render(**definitions)
--- a/conans/cli/commands/new.py
+++ b/conans/cli/commands/new.py
@@ -1,7 +1,7 @@
 import os
 import shutil
 
-from jinja2 import Environment, meta
+from jinja2 import Environment, meta, exceptions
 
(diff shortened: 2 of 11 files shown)


conan/api/subapi/new.py
```diff
@@ -85,7 +86,9 @@ def _read_files(self, template_folder):
     @staticmethod
     def render(template_files, definitions):
         result = {}
+        name = definitions.get("name", "Pkg")
```
memsharded, line 89: Do not default to a package with upper case, upper case names are no longer allowed in 2.0

conans/test/utils/tools.py
```diff
@@ -605,7 +605,7 @@ def alias(self, source, target):
         source = RecipeReference.loads(source)
         target = target.split("/", 1)[1]
         self.run(f"new alias -d name={source.name} -d version={source.version} "
-                 f"-d target={target} -f")
+                 f"-d target={target} -d revision_mode=hash -f")
```
memsharded: This should be able to be defaulted (maybe even removed from the ``alias`` template)

conans/cli/commands/new.py
```diff
@@ -40,6 +42,24 @@ def new(conan_api, parser, *args):
     if not template_files and not non_template_files:
         raise ConanException("Template doesn't exist or not a folder: {}".format(args.template))
 
+    template_vars = []
```
lasote: Not related to this line, but I think we talked about providing the list of predefined layouts (hardcoded) in the command help. (Line 15)

conan/api/subapi/new.py
czoido: It was not the name of the package but the python class name that we camel-cased in the 1.x branch, the name of the package is name, anyway I can default it to a lowercase

conans/test/utils/tools.py
czoido: This implementation won't accept default in templates, if we think we need them then I can make something different...

conans/test/integration/command/new_test.py
```diff
@@ -36,6 +36,18 @@ def test_new_cmake_exe(self):
         client.run("export . --user=myuser --channel=testing")
         assert "pkg/1.3@myuser/testing" in client.out
 
+    def test_new_missing_definitions(self):
+        client = TestClient()
+        client.run("new cmake_lib", assert_error=True)
+        assert "ERROR: Missing name, version definitions. " \
+               "Required definitions are: name, version" in client.out
+        client.run("new cmake_lib -d name=name", assert_error=True)
+        assert "ERROR: Missing version definitions. " \
```
lasote: Might be a bit confusing sentence. Probably better to single quote the attributes and use "definition" (singular).
`ERROR: Missing 'version' definition.`

conan/api/subapi/new.py
```diff
@@ -85,10 +86,12 @@ def _read_files(self, template_folder):
...
         definitions["conan_version"] = __version__
+        definitions["package_name"] = name.replace("-", "_").replace("+", "_").replace(".", "_")
         for k, v in template_files.items():
-            k = Template(k, keep_trailing_newline=True).render(**definitions)
-            v = Template(v, keep_trailing_newline=True).render(**definitions)
+            k = Template(k, keep_trailing_newline=True, undefined=StrictUndefined).render(**definitions)
```
czoido, line 93: I had to add `StrictUndefined` so that it raises an exception if a context variable is undefined.
That means that you have always to set all the definitions unless the variable has a default filter, that works as it used to work.

conans/test/integration/command/new_test.py
```diff
@@ -80,7 +89,7 @@ class Conan(ConanFile):
         save(os.path.join(path, "conanfile.py"), template1)
         save(os.path.join(path, "file.h"), "{{header}}")
 
-        client.run(f"new mytemplate -d name=hello -d version=0.1")
+        client.run(f"new mytemplate -d name=hello -d version=0.1 -d header=")
```
czoido, line 92: This change is a consequence of adding `StrictUndefined` mode to the templates.

memsharded (approved):

Issue #11161, opened 2022-05-04
Document available templates in conan new command


czoido:
List in the help all the pre-defined templates for conan new

memsharded:
Closed by https://github.com/conan-io/conan/pull/11170

PR #11170, opened 2022-05-04
conan new docstring add builtin templates

memsharded:
Close https://github.com/conan-io/conan/issues/11161

### DIFF
--- a/conans/cli/commands/new.py
+++ b/conans/cli/commands/new.py
@@ -14,7 +14,13 @@ def new(conan_api, parser, *args):
     """
     Create a new recipe (with conanfile.py and other files) from a predefined template
     """
-    parser.add_argument("template", help="Template name, built-in predefined one or user one")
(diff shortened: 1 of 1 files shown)


conans/cli/commands/new.py
```diff
@@ -14,7 +14,13 @@ def new(conan_api, parser, *args):
     """
     Create a new recipe (with conanfile.py and other files) from a predefined template
     """
-    parser.add_argument("template", help="Template name, built-in predefined one or user one")
+    parser.add_argument("template", help="Template name, built-in predefined one or user one. "
```
czoido, line 17: I guess we are intentionally not documenting the `conan new alias ...` ?

memsharded: I didn't do it on purpose, but I think yes, it is better not document that here yet.

czoido (approved):

PR #11178, opened 2022-05-05
Feature/docstrings cmake for the docs

lasote:
Some docstrings for cmake stuff for the docs.
Related https://github.com/conan-io/docs/pull/2524


czoido:
Aren't the constructor of CMakeToolchain and CMakeDeps missing?

conan/tools/cmake/cmake.py
```diff
@@ -137,6 +176,17 @@ def install(self, build_type=None):
         self._conanfile.run(command)
 
     def test(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
+        """
+        Equivalent to running cmake --build . --target=RUN_TESTS.
+
+        :param build_type: Use it only to override the value defined in the ``settings.build_type``.
+                           It can fail if the build is single configuration (e.g. Unix Makefiles), as
+                           in that case the build type must be specified at configure time, not build
+                           type.
```
```suggestion
                           time.
```

conan/tools/cmake/layout.py
```diff
@@ -2,6 +2,14 @@
 
 
 def cmake_layout(conanfile, generator=None, src_folder="."):
+    """
+
+    :param conanfile: The current recipe object. Always use ``self``.
+    :param generator: If specified, it will be used as the CMake generator if no declared
+                      ``tools.cmake.cmaketoolchain:generator``.
```
memsharded:
```suggestion
    :param generator: Allow defining the CMake generator. In most cases it doesn't need to be passed, as it will get the value from the configuration              ``tools.cmake.cmaketoolchain:generator``, or it will automatically deduce the generator from the ``settings``
```

conan/tools/cmake/cmake.py
```diff
@@ -32,13 +32,12 @@ def _cmake_cmd_line_args(conanfile, generator):
 
 
 class CMake(object):
-    """ CMake helper to use together with the toolchain feature. It implements a very simple
-    wrapper to call the cmake executable, but without passing compile flags, preprocessor
-    definitions... all that is set by the toolchain. Only the generator and the CMAKE_TOOLCHAIN_FILE
-    are passed to the command line, plus the ``--config Release`` for builds in multi-config
-    """
+    """ CMake helper to use together with the CMakeToolchain feature """
```
memsharded, line 35: Because the basic description of this class goes to the ``.rst`` file? Or who will explain that this class implements the command line wrapper over the cmake executable?

memsharded (approved):

czoido (approved):

conan/tools/cmake/cmake.py
lasote: It is explained on the introduction page, before the reference. This docstring indeed goes to the rst.

Issue #11176, opened 2022-05-05
[feature][settings] Improve Settings and `None` values


franramirez688:
`Settings` is not managing properly the `None` values. A clear example of that is:

```python
    def dumps(self):
        """ produces a text string with lines containing a flattened version:
        compiler.arch = XX
        compiler.arch.speed = YY
        """
        result = []
        for (name, value) in self.values_list:
            # It is important to discard None values, so migrations in settings can be done
            # without breaking all existing packages SHAs, by adding a first "None" option
            # that doesn't change the final sha
            if value != "None":
                result.append("%s=%s" % (name, value))
        return '\n'.join(result)
```
It shouldn't be needed to make that, or, at least, that `"None"` value should be `None` instead.

Related to: https://github.com/conan-io/conan/pull/11169

memsharded:
Closed by https://github.com/conan-io/conan/pull/11182

PR #11182, opened 2022-05-05
model settings None=not defined

memsharded:
Close https://github.com/conan-io/conan/issues/11176


czoido (approved):

conans/test/integration/settings/built_type_setting_test.py
```diff
@@ -48,15 +48,15 @@ def test(self):
...
 
-        # This won't fail, as it has a build_type=None, which is allowed
         client.run("export . --name=pkg --version=0.1 --user=lasote --channel=testing")
-        client.run("install --requires=pkg/0.1@lasote/testing -pr=myprofile --build='*'")
-        self.assertEqual(1, str(client.out).count("BUILD TYPE: Not defined"))
+        client.run("install --requires=pkg/0.1@lasote/testing -pr=myprofile --build='*'",
+                   assert_error=True)
+        assert "ERROR: Invalid setting 'None' is not a valid 'settings.build_type'" in client.out
+        assert "Possible values are [None, 'Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']" in \
+            client.out
```
franramirez688, line 61: A bit confusing... We're saying that `None` isn't a valid one but it's also a possible one:
```bash
ERROR: Invalid setting 'None' is not a valid 'settings.build_type'
Possible values are [None, 'Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
```
I think we should say instead:
```
ERROR: Invalid setting 'None' is not a valid 'settings.build_type'
Possible values are ['Debug', 'Release', 'RelWithDebInfo', 'MinSizeRel']
```
I guess that `null` is enough documented to know that it means you can keep empty that field

lasote (approved):

conans/test/integration/settings/built_type_setting_test.py
memsharded: Good point, I'll filter it in the output

memsharded: Done: https://github.com/conan-io/conan/pull/11182/commits/9ccce9f4af3820160c337df53c745973dbc480db

lasote (approved):

franramirez688 (approved):

System-Arch:
I believe this change introduced a bug that I am seeing while trying to use Conan 2.0.0-beta3. The issue relates to the following change in settings.py: validate()

ORIGINAL: 
  def validate(self):
        if self._value is None and "None" not in self._definition:

WITH THIS CHANGE:
    def validate(self):
        if self._value is None and None not in self._definition:

The error that I am seeing:
ERROR: conanfile.py (gmp/6.2.1): Invalid ID: Invalid: 'settings.compiler.threads' value not defined

Stepping through the code in the debugger reveals:
(Pdb) p self._value
None
(Pdb) p self._name
'settings.compiler.threads'
(Pdb) p self._definition
['None', 'posix', 'win32', 'ANY']

Thus, it seems the quotes should be restored around 'None' in the "if" statement cited above or the code that reads in the settings.yml file needs to treat None as a Python value rather than the string it appears to be read in as.

Thanks

memsharded:
Hi @System-Arch 

Could you please report this as a new issue? Better than the code internals, it would be much better if you provide a simple but full ``conanfile.py`` that we could reproduce the behavior you are describing. Thanks!

Issue #11129, opened 2022-04-28
[develop2] Test lockfiles ``conan lock add pkg/version#revision`` when existing revisions 


memsharded:
The manually added revision will in general not have a timestamp, but it is required for doing the right insertion and ordering.
Would we want to make it the "latest" by default when manually adding? Should it error if timestamp not present?

Note that it won't be an error if other revisions don't exist, as locking it is trivial.

Comes from the Tribe discussion https://github.com/conan-io/tribe/pull/34#discussion_r861111847 from @mec-irobot

memsharded:
Closed by https://github.com/conan-io/conan/pull/11184

PR #11184, opened 2022-05-06
``conan lock add`` revisions and timestamps

memsharded:
Close https://github.com/conan-io/conan/issues/11129


conans/test/integration/lockfile/test_user_overrides.py
```diff
@@ -102,3 +105,73 @@ def test_user_python_overrides():
...
+    c.run(f"lock add --requires=math/1.0#{rev1} --requires=unrelated/2.0")
+    c.run("graph info game --lockfile=conan.lock --lockfile-out=new.lock --lockfile-partial")
+    assert f"math/1.0#{rev1}" in c.out
+    assert f"math/1.0#{rev2}" not in c.out
+    assert f"math/1.0#{rev0}" not in c.out
+    # The resulting lockfile contains the full revision now
+    new_lock = c.load("new.lock")
+    assert f"math/1.0#{rev1}" in new_lock
+    assert f"math/1.0#{rev2}" not in new_lock
+    assert f"math/1.0#{rev0}" not in c.out
```
franramirez688, line 153: I don't get why this test is related to the change. The change is:

```python
        # When we are adding something manually to a lockfile, we want to prioritize it
        # so we make it the first element in the list
        self._requires.move_to_end(ref, last=False)
```
But, here, we're not checking any order of the requirements, right? This test is testing only that we're able to add manually revisions, isn't it?

franramirez688 (approved):

conans/test/integration/lockfile/test_user_overrides.py
memsharded: The other test is checking the actual order in the lockfile. This test is validating that adding multiple requirements to the lockfile, even if incomplete, it won't crash (it did crash before, because of the order was raising TypeError comparison between float <> None or str <> None

conans/model/graph_lock.py
```diff
@@ -146,15 +144,21 @@ def merge(self, other):
         self._python_requires.merge(other._python_requires)
 
     def add(self, requires=None, build_requires=None, python_requires=None):
+        """ adding new things manually to the lockfile puts them as the first element in the
+        list of locked requires, prioritizing them
+        """
```
franramirez688, line 151: It looks like obsolete. Remove it?

franramirez688 (approved):

PR #11188, opened 2022-05-06
ux: improve error message from CMakeDeps

jgsogo:
Changelog: omit
Docs: omit

In a big (or not so big) graph, we need to know where the error comes from. I don´t know if this is the best place to catch this and add the information to the error message, but... the idea is to get a better message as shown below:

---

The error goes from:

```
ERROR: Error in generator 'CMakeDeps': U is not absolute
```

to:

```
ERROR: Error in generator 'CMakeDeps': error generating context for 'abseil/20211102.0': U is not absolute
```


memsharded (approved):

SSE4:
@jgsogo what is `U` in your context?

PR #11190, opened 2022-05-06
[develop2] Internal info.validate() for complete definitions

memsharded:
- ``self.options.validate()`` was in a weird place, moved to its own method
- Added ``self.settings.validate()``
- Fixed computation of ``package_id()`` only for ``.py`` consumer recipes, not virtuals, not ``conanfile.txt``


franramirez688 (approved):
It's looking very good!

lasote (approved):
👍🏼

Issue #11193, opened 2022-05-07
[question] Is it possible in conan v2 to encode in a recipe that version of a specific tool_requires can be overridden?


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

When a library has a requirement like `protobuf` and uses `protoc` executable at build time to generate source code from `.proto` files, `protobuf` version in `requires` and `tool_requires` must be the same, and if version is overridden by downstream (even patch version), this recipe should generate a new package id and override `protobuf` version in tool_requires, not only requires.
AFAIK, there is no way to achieve such logic in conan v1. Is it possible in conan v2? Is it possible to encode this behavior in a recipe so that it can work out of the box for consumers?

For example:

```python
from conans import ConanFile

class FooConan(ConanFile):
    name = "foo"
    version = "1.0.0"

    def requirements(self):
        self.requires("protobuf/3.20.0")

    def build_requirements(self):
        self.tool_requires("protobuf/3.20.0") # here an option could tell that this version must always follow the one in requires

    def package_id(self):
        # some logic to always encode protobuf version in package id regardless of user policy
```

```python
from conans import ConanFile

class BarConan(ConanFile):
    name = "bar"
    version = "1.0.0"

    def requirements(self):
        self.requires("foo/1.0.0")
        self.requires("protobuf/3.19.0", override=True)
```

`conan install bar/1.0.0 -b missing` would trigger a new build of foo/1.0.0 package, with a different package id than the one relying on protobuf/3.20.0, where protobuf is 3.19.0 in tool_requires.

SpaceIm:
It seems quite similar than https://github.com/conan-io/conan/issues/11010 (but it's not exactly the same problem, it's closer to https://github.com/conan-io/conan/issues/11010#issuecomment-1108859548)

memsharded:
Please see the test in https://github.com/conan-io/conan/pull/11194. In summary:

Conan 2.0 allows overriding of upstream tool_requires, for that to happen:
- The package using that tool_require should declare it as ``visible=True``, so the downstream has visibility over them
- The downstream consumer can do a ``self.requires(..., override=True, build=True, headers=False, libs=False)`` (this UX could be improved allowing ``self.tool_requires(..., override=True)``, but that is just a easy syntax improvement)
- The evolution of host and build requires is still independent, if you want to override both, you need to override both explicitly.

memsharded:
[...]

PR #11194, opened 2022-05-07
[develop2] add test how to override tool-requires

memsharded:
Close https://github.com/conan-io/conan/issues/11193

Important:
the dependencies resolution algorithm has been changed to not return on first conflict/override point, but keep checking downstream to allow the downstream consumers to have higher priority while overriding/forcing. It also allows such overrides when the conflicting package base (the root node in the diamond causing the conflict), is above the overriding package.


conans/test/integration/graph/core/test_build_requires.py
```diff
@@ -556,6 +556,69 @@ def test_conflict_diamond(self):
...
+        # app -> libb -(br public)-> protobuf/0.1
+        #           \--------------> protobuf/0.2
+        #  \---(br, override)------> protobuf/0.2
+        self.recipe_conanfile("protobuf/0.1", GenConanfile())
+        self.recipe_conanfile("protobuf/0.2", GenConanfile())
+        self.recipe_conanfile("libb/0.1",
+                              GenConanfile().with_tool_requirement("protobuf/0.1", visible=True)
+                              .with_require("protobuf/0.2"))
+
+        deps_graph = self.build_graph(GenConanfile("app", "0.1").with_requires("libb/0.1")
```
memsharded, line 598: Add the ``self.tool_requries(...., override=True)`` to simplify UX

lasote (approved):
With your zoom explanation is now clear. Good job!
 Great tests too!!

czoido (approved):

jcar87 (approved):
LGTM!!

PR #11197, opened 2022-05-08
[develop2] Fixing BinaryInfo for search binary results

memsharded:
This implements the decoupling of ``load_binary_info()`` for the ``conan list packages`` functionality. It is only used for display in terminal for ``conan list packages`` (and the html table). The internal info is managed as a Python dict, json-like.

This PR does not cover yet, and it seems this should be another later PR:

- The generalization of the ``-q query`` parsing
- The generalization of the reading of ``conaninfo.txt``

### DIFF
--- a/conan/api/subapi/list.py
+++ b/conan/api/subapi/list.py
@@ -2,7 +2,6 @@ from typing import Dict
 
 from conan.api.model import Remote
 from conans.cli.conan_app import ConanApp
-from conans.model.info import BinaryInfo
 from conans.model.package_ref import PkgReference
 from conans.model.recipe_ref import RecipeReference
 from conans.search.search import get_cache_packages_binary_info, filter_packages
@@ -64,7 +63,7 @@ class ListAPI:
         return results
 
     def packages_configurations(self, ref: RecipeReference,
-                                remote=None) -> Dict[PkgReference, BinaryInfo]:
+                                remote=None) -> Dict[PkgReference, dict]:
         assert ref.revision is not None, "packages: ref should have a revision. " \
                                          "Check latest if needed."
         if not remote:
--- a/conan/api/subapi/list.py
+++ b/conan/api/subapi/list.py
@@ -1,11 +1,11 @@
 from typing import Dict
 
-from conan.api.model import Remote, PkgConfiguration
+from conan.api.model import Remote
 from conans.cli.conan_app import ConanApp
-from conans.errors import NotFoundException
+from conans.model.info import BinaryInfo
 from conans.model.package_ref import PkgReference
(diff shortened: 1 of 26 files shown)


conan/api/subapi/list.py
```diff
@@ -64,22 +64,20 @@ def package_revisions(self, pref: PkgReference, remote: Remote=None):
         return results
 
     def packages_configurations(self, ref: RecipeReference,
```
franramirez688, line 65: Perhaps, it's not the moment, but we should change this function name or add a docstring.

conans/model/info.py
```diff
@@ -392,13 +375,47 @@ def _loads_settings(settings_text):
...
+        self.requires = reqs_info
+        self.build_requires = build_requires_info
+        self.python_requires = python_requires
+
+    def clone(self):
+        """ Useful for build_id implementation and for compatibility()
+        """
+        result = ConanInfo()
+        result.invalid = self.invalid
+        result.settings = self.settings.copy()
```
franramirez688: Shouldn't we use `copy_conaninfo_settings` instead?

conans/model/info.py
```diff
@@ -443,25 +456,21 @@ def package_id(self):
...
+        # If the options are not fully defined, this is also an invalid case
+        try:
+            self.options.validate()
+        except ConanException as e:
+            self.invalid = BINARY_INVALID, str(e)
+
+        try:
+            self.settings.validate()
+        except ConanException as e:
+            self.invalid = BINARY_INVALID, str(e)
```
franramirez688: This `validate` method is fine to isolate this logic, but what I'd expect is to raise an exception too instead of only giving the value to the `self.invalid`.

franramirez688 (approved):
Good job! 👍

conans/model/info.py
memsharded: Not really, it is necessary to keep it without raising, because for the ``compatibility`` cases we want to be able to continue processing. It will only raise later, if you actually try to build or install such invalid configuration, but at this point, you can't raise yet.

conan/api/subapi/list.py
memsharded: Yes, more refactors and improvements are necessary in different places of the api interface, but lets focus now in the ``package_id`` goal, cleaning things around ``conaninfo``

conans/model/info.py
memsharded: No, that only needs to happen the first time, when the first ``ConanInfo`` is created

lasote (approved):

franramirez688 (approved):

PR #11199, opened 2022-05-09
[dev2][package_id] Part 3: Mechanism to add some `conf`s to `package_id` computation

### LINKED ISSUE #11163: [feature] [package_id] Improve  `ConanInfo` interface to add new `conf` values
Improve the way to add new configurations from `conf` to `ConanInfo.package_id()`. Some proposals:

* Via `[conf]`: having a new configuration to indicate all the configurations to be added like `core.package_id:add=["tools.build:cxxflags"]`
* Via `conanfile.py`: having something like this code:
```python
def package_id():
     self.info.conf.add("tools.build:cxxflags")
```
* ....

franramirez688:
Changelog: Feature: Added mechanism to include some configurations into `package_id` computation via `tools.info.package_id:confs` one.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/11163

### DIFF
--- a/conans/client/graph/compute_pid.py
+++ b/conans/client/graph/compute_pid.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 from conans.client.graph.graph import BINARY_INVALID, BINARY_ERROR
 from conans.errors import conanfile_exception_formatter, ConanInvalidConfiguration, \
     ConanErrorConfiguration, conanfile_remove_attr
+from conans.model.conf import Conf
 from conans.model.info import ConanInfo, RequirementsInfo, RequirementInfo, PythonRequiresInfo
 
 
@@ -18,6 +19,7 @@ def compute_package_id(node, new_config):
     embed_mode = new_config.get("core.package_id:default_embed_mode", default="full_mode")
     python_mode = new_config.get("core.package_id:default_python_mode", default="minor_mode")
     build_mode = new_config.get("core.package_id:default_build_mode", default=None)
+    conf_names = new_config.get("core.package_id:confs", default=[], check_type=list)
 
     python_requires = getattr(conanfile, "python_requires", None)
     if python_requires:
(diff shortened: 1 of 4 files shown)


conans/model/conf.py
```diff
@@ -8,14 +8,20 @@
...
     "core:non_interactive": "Disable interactive user input, raises error if input necessary",
-    "core.package_id:msvc_visual_incompatible": "Allows opting-out the fallback from the new msvc compiler to the Visual Studio compiler existing binaries",
     "core:default_profile": "Defines the default host profile ('default' by default)",
     "core:default_build_profile": "Defines the default build profile (None by default)",
     "core.upload:retry": "Number of retries in case of failure when uploading to Conan server",
     "core.upload:retry_wait": "Seconds to wait between upload attempts to Conan server",
     "core.download:parallel": "Number of concurrent threads to download packages",
     "core.download:retry": "Number of retries in case of failure when downloading from Conan server",
     "core.download:retry_wait": "Seconds to wait between download attempts from Conan server",
+    # Package ID related
```
memsharded: Recall that BUILT_IN_CONFS is no longer used in the whole project. The command ``conan config list`` was removed. Probably we should remove this?

conans/client/graph/compute_pid.py
```diff
@@ -18,6 +19,7 @@ def compute_package_id(node, new_config):
     embed_mode = new_config.get("core.package_id:default_embed_mode", default="full_mode")
     python_mode = new_config.get("core.package_id:default_python_mode", default="minor_mode")
     build_mode = new_config.get("core.package_id:default_build_mode", default=None)
+    conf_names = new_config.get("core.package_id:confs", default=[], check_type=list)
```
memsharded: Not sure if we want this to be ``core``, or it would be a ``profile`` one, something that users would prefer to define in the ``profiles``?

memsharded (commented):
Looking good so far

conans/client/graph/compute_pid.py
franramirez688: Yes, that was what I was trying to figure out. I opened the draft to see how it looked like, and, definitely, it should go through the user profile 😉

conans/model/conf.py
```diff
@@ -323,6 +322,40 @@ def filter_user_modules(self):
...
+        :return: a new `< Conf object >` with the configuration selected by `tools.info.package_id:confs`.
+        """
+        result = Conf()
+        # Reading the list of all the configurations selected by the user to use for the package_id
+        package_id_confs = self.get("tools.info.package_id:confs", default=[], check_type=list)
+        for conf_name in package_id_confs:
+            value = self.get(conf_name)
+            # Pruning None values, those should not affect package ID
+            if value is not None:
+                result.define(conf_name, value)
```
franramirez688, line 356: Take into account that `tools.build:cxxflags=[]` will be included into the `package_id` which is fine I guess.

memsharded: Uhm, I don't love it, because semantically, not adding any flag should result in the same ``package_id``. Probably check ``if value:`` for discarding empty strings and things too.

franramirez688: oh, yes, you're completely right.

memsharded (approved):

PR #11206, opened 2022-05-09
Docstrings for Docs for reference.tools.env

lasote:


conan/tools/env/environment.py
```diff
@@ -170,12 +174,20 @@ def __repr__(self):
...
-        values
+        """
+        :return: A string with a profile-like original definition, not the full environment
+                 values
         """
         return "\n".join([v.dumps() for v in reversed(self._values.values())])
 
     def define(self, name, value, separator=" "):
+        """
+        Overwrite previously existing `name` environment variable with new `value`
```
czoido: I think I would not use overwrite here, just define... like this?
```suggestion
        Define `name` environment variable with value `value`
```

conan/tools/env/environment.py
```diff
@@ -170,12 +174,20 @@ def __repr__(self):
...
+
+        :param name: Name of the variable to override
+        :param value: New value
+        :param separator: The character to separate appended or prepended values
```
czoido:
```suggestion
        :param name: Name of the variable
        :param value: Value that the environment variable will take
        :param separator: The character to separate appended or prepended values
```

conan/tools/env/environment.py
```diff
@@ -184,31 +196,66 @@ def define_path(self, name, value):
...
+
+        :param name: Name of the variable to append a new value
+        :param value: New value
+        :param separator: The character to separate the appended value with the previous value
+        """
         self._values.setdefault(name, _EnvValue(name)).append(value, separator)
 
     def append_path(self, name, value):
+        """
+        Similar to "append" method but indicating that the variable is a filesystem path.
```
czoido:
```suggestion
        Similar to "append" method but indicating that the variable is a filesystem path. It will automatically handle the path separators depending on the operating system.
```

conan/tools/env/environment.py
```diff
@@ -184,31 +196,66 @@ def define_path(self, name, value):
...
+
+        :param name: Name of the variable to prepend a new value
+        :param value: New value
+        :param separator: The character to separate the prepended value with the previous value
+        """
         self._values.setdefault(name, _EnvValue(name)).prepend(value, separator)
 
     def prepend_path(self, name, value):
+        """
+        Similar to "prepend" method but indicating that the variable is a filesystem path.
```
czoido:
```suggestion
        Similar to "prepend" method but indicating that the variable is a filesystem path. It will automatically handle the path separators depending on the operating system.
```

czoido (approved):
Just some minor comments

conan/tools/env/environment.py
memsharded: Define ``name`` environment variable ...... If the variable ``name`` already existed, this definition will overwrite the existing value.

Or something like that.

conan/tools/env/environment.py
```diff
@@ -184,31 +196,66 @@ def define_path(self, name, value):
...
         """
         self._values[name] = _EnvValue(name, None)
 
     def append(self, name, value, separator=None):
+        """
+        Append the `value` to an environment variable `name`
+
+        :param name: Name of the variable to append a new value
+        :param value: New value
+        :param separator: The character to separate the appended value with the previous value
```
memsharded: Clarify that if not specified it will be a blank space?

conan/tools/env/virtualrunenv.py
```diff
@@ -26,11 +26,15 @@ def _handle_paths(paths):
 
 
 class VirtualRunEnv:
-    """ captures the conanfile environment that is defined from its
-    dependencies, and also from profiles
+    """ Calculates the environment variables of the build time context and produces a conanbuildenv
```
memsharded:
```suggestion
    """ Calculates the environment variables of the runtime context and produces a conanrunenv
```

memsharded (approved):
Good!

memsharded (approved):

Issue #9460, opened 2021-08-21
[bug] tools.os_info.with_yum does not detect AlmaLinux (CentOS replacement)


daravi:
AlmaLinux is the CentOS successor. Similar to CentOS it uses `yum` or `dnf`. However, `with_yum` in `system_requirement` fails:

e.g.
```
opengl/system: WARN: Don't know how to install OpenGL for your distro.
opengl/system: ERROR: while executing system_requirements(): local variable 'packages' referenced before assignment
ERROR: Error in system requirements
```

I am not sure if this is a bug in Conan or AlmaLinux. I can do further debugging as requested.

### Environment Details (include every applicable attribute)
  * Operating System+version: AlmaLinux 8.4 (Electric Cheetah)
  * Compiler+version: LLVM 10.0.1
  * Conan version: 1.38.0
  * Python version: 3.9.6

memsharded:
Hi @daravi 

yes, please, can you verify you have the latest revision of ``opengl/system`` installed? I see:

```python
def system_requirements(self):
        packages = []
        if tools.os_info.is_linux and self.settings.os == "Linux":
            if tools.os_info.with_yum:
                if tools.os_info.linux_distro == "fedora" and tools.os_info.os_version >= "32":
                    packages = ["libglvnd-devel"]
                else:
                    packages = ["mesa-libGL-devel"]
            elif tools.os_info.with_apt:
                ubuntu_20_or_later = tools.os_info.linux_distro == "ubuntu" and tools.os_info.os_version >= "20"
                debian_11_or_later = tools.os_info.linux_distro == "debian" and tools.os_info.os_version >= "11"
                pop_os_20_or_later = tools.os_info.linux_distro == "pop" and tools.os_info.os_version >= "20"
                if ubuntu_20_or_later or debian_11_or_later or pop_os_20_or_later:
                    packages = ["libgl-dev"]
                else:
                    packages = ["libgl1-mesa-dev"]
            elif tools.os_info.with_pacman:
                packages = ["libglvnd"]
            elif tools.os_info.with_zypper:
                packages = ["Mesa-libGL-devel"]
            else:
                self.output.warn("Don't know how to install OpenGL for your distro.")
        elif tools.os_info.is_freebsd and self.settings.os == "FreeBSD":
            packages = ["libglvnd"]
        if packages:
            package_tool = tools.SystemPackageTool(conanfile=self, default_mode='verify')
            for p in packages:
                package_tool.install(update=True, packages=p)
```

[...]

PR #11212, opened 2022-05-10
Add Rocky Linux to with_yum

lbakman:
Changelog: Bugfix: Add Rocky Linux to with_yum.
Docs: omit

Similar to: #9460, only for Rocky Linux.

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -32,7 +32,7 @@ class _SystemPackageManagerTool(object):
             os_name = "msys2"
         manager_mapping = {"apt-get": ["Linux", "ubuntu", "debian"],
                            "yum": ["pidora", "scientific", "xenserver", "amazon", "oracle", "amzn",
-                                   "almalinux"],
+                                   "almalinux", "rocky"],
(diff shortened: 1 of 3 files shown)


memsharded (changes requested):
Thanks for contributing this @lbakman 

But this place you are doing the change is deprecated in favor of the new PackageManager implementation, so at least you should include changes in https://github.com/conan-io/conan/blob/develop/conan/tools/system/package_manager.py too. 

An entry check to an existing unit test, like ``test_package_manager_distro`` would be good too.

lbakman:
Good points.

I have included the changes to package_manager.py and added rocky/yum as a parameter to test_package_manager_distro.

One thing that caught my attention: I can see that both "rhel" and "centos" has been moved to "dnf". Even though I added "rocky" as a "yum" mapping, I would argue that both "almalinux" and "rocky" should be moved to "dnf" as well. This would be a breaking change compared to "conans/client/tools/oss.py", but I guess the same goes for "rhel" and "centos".

memsharded (approved):

memsharded:
Thanks!

Issue #11179, opened 2022-05-05
[question] how to specify `requires`/`tool_requires` for same dependency.


mattyclarkson:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

How does one add a requirement for both build and runtime?

I would like to build `autoconf` so have the following:

```py
    def requirements(self) -> None:
        self.requires(f"perl/5.34.1@{self._channel}")
        self.requires(f"m4/1.4.19@{self._channel}")

    def build_requirements(self) -> None:
        self.tool_requires(f"make/4.3@{self._channel}")
        self.tool_requires(f"m4/1.4.19@{self._channel}")
        self.tool_requires(f"perl/5.34.1@{self._channel}")
```

As `perl`/`m4` is needed for both building _and_ runtime of `autoconf`. However, doing `conan info $PWD/package/autoconf` results in the following:

```
Requires:
    perl/5.34.1@mychannel/testing
    m4/1.4.19@mychannel/testing
Build Requires:
    make/4.3@mychannel/testing
```

This means that the build fails beacause `perl` and the `m4` binaries are not exposed on the `PATH` for the build.

How do I get _both_: expose the requirement to the build _and_ use the requirement at runtime.

I'm using host/build contexts with 1.47 and trying to make my recipes as Conan v2 compatible as possible.

mattyclarkson:
Thinking about this more. It actually doesn't make sense to have it as `requires`: it's a tool with a dependency on another tool.

Hence, I would expect this:

```py
    def requirements(self) -> None:
        self.tool_requires(f"perl/5.34.1@{self._channel}")
        self.tool_requires(f"m4/1.4.19@{self._channel}")

    def build_requirements(self) -> None:
        self.tool_requires(f"make/4.3@{self._channel}")
```

That doesn't work though:

```
ERROR: conanfile.py (autoconf/None): Error in requirements() method, line 37
        self.tool_requires(f"perl/5.34.1@{self._channel}")
        AttributeError: 'Package' object has no attribute 'tool_requires'
```

memsharded:
Hi @mattyclarkson 

Because ``tool_requires()`` must be inside ``build_requirements()`` method, please try that and let me know.

mattyclarkson:
If I just put them in `build_requirements` the _build_ of `autoconf` works but when I go to _use_ `autoreconf` downstream it doesn't work because the `perl` interpreter isn't available.

[...]

PR #11213, opened 2022-05-10
add `test_tool_requires_and_requires`

mattyclarkson:
A test that verifies a dependency can be both a host *and* build requirement

Changelog: Bugfix: Allow dependencies to be both `host` *and* `build`

- [x] Refer to the issue that supports this Pull Request. (#11179)


memsharded (commented):
Hi @mattyclarkson 

I have fixed the test, it was missing the "build" profile, check the lines:
```python
# Note that I pass the -s:b os=Windows that enable the "build" profile and context
client.run("create . consumer/1.0@ -s:b os=Windows")
```

The graph resolution is different when not using the build and host contexts. You should use the "build" profile and context, because this is what Conan 2.0 will always use, and is opting-in into the recommended cross-building approach.

PR #11220, opened 2022-05-11
Add DCMAKE_BUILD_TYPE to markdown generator instructions

czoido:
Changelog: Fix: Add `-DCMAKE_BUILD_TYPE` to markdown generator instructions for CMake single config.
Docs: omit

After the changes introduced in 1.48 that the toolchain no longer contains the build type these instructions were not correct.

Should we port this to 1.48.1?


memsharded:
> Should we port this to 1.48.1?

I don't think it is worth a patch release only for this. If we have other fixes for a 1.48.1, then yes, I'd try to include this one too.

PR #11232, opened 2022-05-11
[develop2] Fix broken ``conan list packages --format=json`` output

memsharded:
It crashs now, it is not tested.

### DIFF
--- a/conans/cli/commands/__init__.py
+++ b/conans/cli/commands/__init__.py
@@ -26,7 +26,7 @@ class ConanJSONEncoder(JSONEncoder):
             return repr(o)
         elif isinstance(o, Remote):
             return o.name
-        raise TypeError("Don't know how to serialize a {} object".format(o.__class__))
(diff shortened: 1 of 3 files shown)


conans/cli/commands/__init__.py
```diff
@@ -26,7 +26,7 @@ def default(self, o):
             return repr(o)
         elif isinstance(o, Remote):
             return o.name
-        raise TypeError("Don't know how to serialize a {} object".format(o.__class__))
+        return JSONEncoder.default(self, o)
```
franramirez688, line 29: Should we manage any exceptions raised here?

franramirez688 (approved):

conans/cli/commands/__init__.py
memsharded: Not really, the ``JSONEncoder.defaul()`` implementation will do that ``raise`` internally

PR #11237, opened 2022-05-12
Docstrings for reference tools.meson

franramirez688:
Changelog: Added Sphinx docstrings to `conan.tools.meson` package.
Docs: https://github.com/conan-io/docs/pull/2533

**Important**: Merge this before `Docs` one.

### DIFF
--- a/conan/tools/meson/helpers.py
+++ b/conan/tools/meson/helpers.py
@@ -63,6 +63,12 @@ _cppstd_map = {
 
 
 def to_meson_machine(machine_os, machine_arch):
+    """Gets the OS system info as the Meson machine context.
+
+    :param machine_os: ``str`` OS name.
+    :param machine_arch: ``str`` OS arch.
+    :return: ``dict`` Meson machine context.
+    """
     system = _meson_system_map.get(machine_os, machine_os.lower())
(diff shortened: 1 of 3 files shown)


conan/tools/meson/helpers.py
```diff
@@ -63,6 +63,12 @@
 
 
 def to_meson_machine(machine_os, machine_arch):
+    """Gets the OS system info as the Meson machine context.
```
memsharded, line 66: These helpers are not public, so make sure they are not documented

conan/tools/meson/toolchain.py
```diff
@@ -272,11 +304,21 @@ def _context(self):
 
     @property
     def content(self):
+        """
+        Gets content of the file to be used by Meson as its context.
```
memsharded, line 308: Not public interface, the only one atm should be ``generate()``?

memsharded (approved):

conan/tools/meson/helpers.py
franramirez688: oh, ok. I'll remove them

conan/tools/meson/toolchain.py
franramirez688: Yeah, I think so. Let's make it private then.

lasote (approved):

PR #11217, opened 2022-05-10
Bugfix - Fix module config consistency

paulharris:
Changelog: (Feature | Fix | Bugfix): see lots of notes in issue #11201 

This PR is for discussion, and should be used with a profile with this added:
```
[conf]
tools.cmake.cmaketoolchain:find_package_prefer_config = False
```

I realise the detail in this PR is lacking, but please refer to the issue and I'll circle back as required.
This is mostly here so it is easier for the developers to know what I'm talking about in that issue.

Also refer to issue #11203 and check the other PRs and issues linked within.
This has been the source of much confusion, so it has become a bit of a sprawling issue.

I'll tick these later, I am new to this and don't have the time now.

- [x] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [ ] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one.

PR #11240, opened 2022-05-12
Assume only config files are generated in CMakeDeps generator when none is specified

### LINKED ISSUE #11201: [bug] CMakeDeps generates a MODULE require instead of CONFIG
See other bug report below,
a simple project requiring Freetype will fail as Freetype will call
find_dependency(brotli REQUIRE MODULE),
but brotli does not have a MODULE finder, only a CONFIG finder.

> RE my comment, "it isn't so simple as disabling that flag"
> 
> For my example, if you do a conan install, and then comment out that flag, it does find Freetype::Freetype, but it then fails to find Brotli (looking for Findbrotli.cmake)
> 
> ```
> CMake Error at /build/msrap2/devenv/share/cmake-3.23/Modules/CMakeFindDependencyMacro.cmake:47 (find_package):                                                                                                    
> No "Findbrotli.cmake" found in CMAKE_MODULE_PATH.
> Call Stack (most recent call first):
> build/FindFreetype.cmake:25 (find_dependency)
> CMakeLists.txt:5 (find_package)
> 
> FindFreetype.cmake:25 is:
>        find_dependency(${_DEPENDENCY} REQUIRED ${${_DEPENDENCY}_FIND_MODE}) 
> which ends up calling
> find_package(brotli REQUIRED;MODULE)
> ```

Seems to be a bug in `CMakeDeps`, and it should be reported to conan client. It doesn't make sense that `CMakeDeps` generate a module file for `freetype` recipe calling `find_dependency(brotli REQUIRED MODULE)`, since `brotli` recipe doesn't have `cmake_find_mode` set to `both`. It should call `find_dependency(brotli REQUIRED CONFIG)`.

_Originally posted by @SpaceIm in https://github.com/conan-io/conan-center-index/issues/10687#issuecomment-1120980081_

jcar87:
Changelog: Bugfix: Fix case where CMakeDeps assumes a module dependency when transitive dependencies do not define `cmake_find_mode` and fallback to a config one.

Docs: omit

Fixes https://github.com/conan-io/conan/issues/11201

The `CMakeDeps` generator explicitly specifies the search mode (config or module) for _transitive_ dependencies. It does so by inspecting the value of the `cmake_find_mode` property defined in `package_info()` in the individual packages, as well as whether or not it is generating a config or a module file. When a dependency specifies that both should be generated, the `CMakeDeps` generator will try to align modules to "depend" on modules, and configs to depend on configs.

The issue was that when a transitive dependency does _not_ define a `cmake_find_mode` explicitly, it should be assumed to be `config`, and thus, the transitive `find_dependency` calls should explicitly say `CONFIG` as that's the only one available to be searched for. The generator was working correctly in this case (generating the config when `cmake_find_mode` was not specified), but when encoding the dependency find mode in consumers, it was defaulting to `MODULE` instead of `CONFIG`. This was only causing issues when the direct dependency was consumed as a module instead of a config (as seen [here](https://github.com/conan-io/conan-center-index/pull/10698))

A test has been modified to cover the case when a consumer depends on package B which depends on package A, and package B is consumed from a module when package A only generates a config.

### DIFF
--- a/conan/tools/cmake/utils.py
+++ b/conan/tools/cmake/utils.py
@@ -22,9 +22,9 @@ def get_file_name(conanfile, forced_module_mode=None):
 def get_find_mode(conanfile):
     """
     :param conanfile: conanfile of the requirement
-    :return: "none" or "config" or "module" or "both" or None when not set
+    :return: "none" or "config" or "module" or "both" or "config" when not set
     """
     tmp = conanfile.cpp_info.get_property("cmake_find_mode")
     if tmp is None:
-        return None
+        return "config"
     return tmp.lower()
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
@@ -113,8 +113,17 @@ def test_reuse_with_modules_and_config(client):
     client.run("build . -if=install")
 
 
-@pytest.mark.parametrize("find_mode", ["both", "config", "module"])
-def test_transitive_modules_found(find_mode):
+find_modes = [
+    ("both", "both", ""),
+    ("config", "config", ""),
+    ("module", "module", ""),
+    ("both", None, ""),
+    ("both", None, "MODULE")
+]
+
+
+@pytest.mark.parametrize("find_mode_PKGA, find_mode_PKGB, find_mode_consumer", find_modes)
+def test_transitive_modules_found(find_mode_PKGA, find_mode_PKGB, find_mode_consumer):
     """
     related to https://github.com/conan-io/conan/issues/10224
     modules files variables were set with the pkg_name_FOUND or pkg_name_VERSION
@@ -127,7 +136,8 @@ def test_transitive_modules_found(find_mode):
         class Pkg(ConanFile):
             {requires}
             def package_info(self):
-                self.cpp_info.set_property("cmake_find_mode", "{mode}")
+                if "{mode}" != "None":
+                    self.cpp_info.set_property("cmake_find_mode", "{mode}")
                 self.cpp_info.set_property("cmake_file_name", "{filename}")
                 self.cpp_info.defines.append("DEFINE_{filename}")
             """)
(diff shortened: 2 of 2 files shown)


conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
```diff
@@ -113,8 +113,17 @@ def test_reuse_with_modules_and_config(client):
...
 
 
-@pytest.mark.parametrize("find_mode", ["both", "config", "module"])
-def test_transitive_modules_found(find_mode):
+find_modes = [
+    ("both", "both", ""),
+    ("config", "config", ""),
+    ("module", "module", ""),
+    ("both", None, ""),
+    ("both", None, "MODULE")
```
jcar87, line 121: This is the case that fails, that is fixed by the PR. It roughly replicates the issue from https://github.com/conan-io/conan-center-index/pull/10698.

When:

* Consumer has `find_package(MYPKGB)` and `MODULES` are searched first by default (this either by specifying it in the `find_package` call, or by setting `CMAKE_FIND_PACKAGE_PREFER_CONFIG` to `OFF` (otherwise the `CMakeToolchain` generator sets it to `ON`).

* `MYPKGB` depends on `MYPKGA `
* `MYPKGA` only provides a `config` because it does not specify `cmake_find_mode` _at all_ (when it does, it all works fine).

conan/tools/cmake/utils.py
```diff
@@ -22,9 +22,9 @@ def get_file_name(conanfile, forced_module_mode=None):
 def get_find_mode(conanfile):
     """
     :param conanfile: conanfile of the requirement
-    :return: "none" or "config" or "module" or "both" or None when not set
+    :return: "none" or "config" or "module" or "both" or "config" when not set
```
jcar87, line 25: This is in line with the documentation.

Note that if the packages explicitly use the python `None` value like this (instead of a string):
```self.cpp_info.set_property("cmake_find_mode", None)```

it would be mapped to `"config"` with this change - I wonder if we need to protect ourselves against that case, or if we should always assume that the property is always a string if defined.

File: conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py (hunk 1 above)
Comment (memsharded) [reply conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py:121]: Do you mean ``MYPKGB`` depends on ``MYPKGA``? Seems typo, reversed.

Comment (jcar87) [reply conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py:121]: Oops, you're right. I've updated the comment !

File: conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py [hunk 3]
```diff
@@ -113,8 +113,17 @@ def test_reuse_with_modules_and_config(client):
...
+]
+
+
+@pytest.mark.parametrize("find_mode_PKGA, find_mode_PKGB, find_mode_consumer", find_modes)
+def test_transitive_modules_found(find_mode_PKGA, find_mode_PKGB, find_mode_consumer):
     """
     related to https://github.com/conan-io/conan/issues/10224
     modules files variables were set with the pkg_name_FOUND or pkg_name_VERSION
```
franramirez688, line 129: I would update/review the docstring with the new issue solved by this PR.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
```diff
@@ -113,8 +113,17 @@ def test_reuse_with_modules_and_config(client):
...
+find_modes = [
+    ("both", "both", ""),
+    ("config", "config", ""),
+    ("module", "module", ""),
+    ("both", None, ""),
+    ("both", None, "MODULE")
+]
+
+
+@pytest.mark.parametrize("find_mode_PKGA, find_mode_PKGB, find_mode_consumer", find_modes)
```
franramirez688, line 125: In my opinion, I prefer to see all the inputs together instead of saving them in a variable (when the list of inputs is small enough), I think it's easier to read the test.

```suggestion
@pytest.mark.parametrize("find_mode_PKGA, find_mode_PKGB, find_mode_consumer", [
    ("both", "both", ""),
    ("config", "config", ""),
    ("module", "module", ""),
    ("both", None, ""),
    ("both", None, "MODULE")
])
```

conan/tools/cmake/utils.py
lasote: I don't think so. We try not to be overprotective, we end with a massive amount of checks that eventually make the things even worse 😓 So I wouldn't add a check here and maybe something to the documentation saying, please note that "none" has to be a string and not a `None` value that means "unset".

lasote (approved):

czoido (approved):

SSE4 (approved):

conan/tools/cmake/utils.py
paulharris: Is "unset" a valid setting?  If so then I'd like to suggest to not allow "None" or any non-strings, and have "unset" an explicit setting that someone can see in the code and documentation.

If the parameter should be a string, then surely that should be a pre-assert?
You have to do manual checks when using a language that isn't statically typed ... that is the penalty of the "convenience" of a language like Python.

Especially if the consequence of someone typing None instead of "None" is several hours/days of debugging and struggling (surprise = bad), instead of a simple error message.

paulharris:
Sorry, about all the noise, I've cleaned up the PRs and bugreports.

This merged PR only goes half way to fixing the problems related to module-config confusions.

Test case: conan-io/conan-center-index#10825
Discussion on the reasons behind the bug: #11201 
My proposed fix: #11217 

Shortest description of further problem:
https://github.com/conan-io/conan-center-index/pull/10825#issue-1237944885

PR #11244, opened 2022-05-13
Feature/develop2 build requires self

memsharded:
Support the case:

```python
class Pkg(ConanFile):
            name = "tool"
            version = "1.0"
            settings = "os"
            def build_requirements(self):
                if cross_building(self):
                    self.tool_requires("tool/1.0")
```

### DIFF
--- a/conans/client/graph/install_graph.py
+++ b/conans/client/graph/install_graph.py
@@ -25,10 +25,10 @@ class _InstallPackageReference:
         # be able to reproduce, typically host preferrably
         self.options = []  # to be able to fire a build, the options will be necessary
         self.filenames = []  # The build_order.json filenames e.g. "windows_build_order"
-        self.depends = []  # Possible dependencies to other binary, e.g. cross-build ICU
-
-    def __eq__(self, other):
-        return self.package_id == other.package_id
+        # If some package, like ICU, requires itself, built for the "build" context architecture
+        # to cross compile, there will be a dependency from the current "self" (host context)
+        # to that "build" package_id.
+        self.depends = []  # List of package_ids of dependencies to other binaries of the same ref
 
     @staticmethod
     def create(node):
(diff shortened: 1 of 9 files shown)


conans/client/graph/install_graph.py
```diff
@@ -24,6 +24,10 @@ def __init__(self):
         # be able to reproduce, typically host preferrably
         self.options = []  # to be able to fire a build, the options will be necessary
         self.filenames = []  # The build_order.json filenames e.g. "windows_build_order"
+        self.depends = []  # Possible dependencies to other binary, e.g. cross-build ICU
```
lasote: I would need some more explanations about this field and its usage.

memsharded: Added more comments

conans/client/graph/install_graph.py
```diff
@@ -25,6 +25,10 @@ def __init__(self):
         # be able to reproduce, typically host preferrably
         self.options = []  # to be able to fire a build, the options will be necessary
         self.filenames = []  # The build_order.json filenames e.g. "windows_build_order"
+        # If some package, like ICU, requires itself, built for the "build" context architecture
+        # to cross compile, there will be a dependency from the current "self" (host context)
+        # to that "build" package_id.
+        self.depends = []  # List of package_ids of dependencies to other binaries of the same ref
```
lasote, line 31: Thanks for the explanation. So, if that variable is used only for that, could the variable name be more explicit? something like `self.self_depend`?

lasote (approved):

conans/client/graph/install_graph.py
memsharded: The aim was that the order function:

```python
def _install_order(self):
        # TODO: Repeated, refactor
```

Could be made generic for both orders, and having the ``depends`` attribute be called the same for ``InstallRecipe`` and ``InstallPackage`` would help. But yes, I can rename it if you want and that could be addressed when that refactor happens. Or I can try to refactor now.

lasote: Ok, then. You can leave it.

czoido (approved):

franramirez688 (approved):

PR #11272, opened 2022-05-17
changing OBJECTS order in CMakeDeps

### LINKED ISSUE #11269: [bug] system_libs are linked before objects causing undefined references
I have a recipe which includes compiled objects which depend on system libraries so I produce the following package info
```
    def package_info(self):
        self.cpp_info.objects = [os.path.join("lib", o) for o in objects]
        self.cpp_info.system_libs = ["pthread", "dl"]
```

however, the CMake config files produced link the system libs before the objects, which, on some systems, makes the linker not find the required symbols:
```
(.text+0x191a89): undefined reference to `pthread_setaffinity_np'
(.text+0x191aba): undefined reference to `dlsym'
```

seems like the order should be reversed in https://github.com/conan-io/conan/blob/0eb405bde999d5b7b8e706b206940d2e24aa828e/conan/tools/cmake/cmakedeps/templates/target_configuration.py#L117-L120

  * Compiler+version: GCC 8
  * Conan version: 1.44.1

see above

memsharded:
Changelog: Bugfix: Link ``cpp_info.objects`` first in ``CMakeDeps`` generator, as they can have dependencies to ``system_libs`` that need to be after them in some platforms to correctly link.
Docs: Omit

Close https://github.com/conan-io/conan/issues/11269

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -117,7 +117,7 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
         set_property(TARGET {{root_target_name}}
                      PROPERTY INTERFACE_LINK_LIBRARIES
                      $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}}}
-                                                   ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
+                     ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
 
         set_property(TARGET {{root_target_name}}
                      PROPERTY INTERFACE_LINK_OPTIONS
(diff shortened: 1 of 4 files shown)


memsharded:
Draft, to do a first "look into" and check risks:

- Does the starting ``;`` really matter? It doesn't seem it breaks anything
- Can anything else break? Maybe the order of package libs (not system libs as reported by the user) with respect to objects? Or is it expected that objects in general will potentially link libraries, but not the other way round?

czoido (approved):
I can't think on anything that would break with this, seems like this is the correct order, for the leading `;` I suppose It's the same than it was before when we had it trailing? Maybe we could try to not add that just in case...

memsharded:
Yes, I thought so, but it seems it would mean to make conditional the whole expression, like:

```cmake
if(${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}})
                     $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}}}
                     ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
else()
                     ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
endif()
```
Which seems to be a bit ugly, I was trying to avoid it

czoido:
> Yes, I thought so, but it seems it would mean to make conditional the whole expression, like:
> 
> ```cmake
> if(${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}})
> set_property(TARGET {{root_target_name}}
>                      PROPERTY INTERFACE_LINK_LIBRARIES
>                      $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_OBJECTS{{config_suffix}}}
>                      ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
> else()
> set_property(TARGET {{root_target_name}}
>                      PROPERTY INTERFACE_LINK_LIBRARIES
>                      ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS{{config_suffix}}}> APPEND)
> endif()
> ```
> 
> Which seems to be a bit ugly, I was trying to avoid it

Yes, I agree. Looks good to me

PR #11280, opened 2022-05-17
[develop2] Upload refactors and better integrity

memsharded:
- Removed dead code in upload model
- Integrity improvements:
   - Checking recipe integrity too
   - Checking all packages and recipes not failing only at the first error
   - Better output of differences, printing folder and tabs
- Calling ``conan_api.upload.prepare(upload_bundle)`` directly from ``cmd``, not indirectly within the API
- Added some docstrings


lasote (approved):
Looks good!

PR #11284, opened 2022-05-17
More flexibility in Autotools tools to override arguments and avoid all default arguments

### LINKED ISSUE #11264: Use `DESTDIR={selfpackage_folder}` rather than `--prefix={self.package_folder}`
Settng the `--prefix` to the package folder can result in the prefix getting baked into some open source projects and lead to unrelocatable packages.

An idiom for autotools projects that results in more commonly relocatable packages is to do:

```
./configure --prefix=/
make
make DESTDIR=<package-dir> install
```

That will result in the same layout as this but avoids _most_ baking of the prefix paths, there are still a few projects that use `DESTDIR` to bake in paths. Most projects use relative paths.

_Originally posted by @mattyclarkson in https://github.com/conan-io/conan/pull/10978#discussion_r870134071_

### LINKED ISSUE #11265: Conditionally add `--enable-shared`/`--enable-static` configure flags.
Unconditionally adding these prevents using the `--enable-option-checking=fatal` option with projects that do not support these flags. For example, `make` and `m4` do not understand these flags.

_Originally posted by @mattyclarkson in https://github.com/conan-io/conan/pull/10978#discussion_r870131250_

czoido:
Changelog: Fix: Use `DESTDIR` argument in `make install` step instead of using the `--prefix` in configure.
Changelog: Feature: More flexibility in Autotools tools to override arguments and avoid all default arguments for `make`, `autoreconf` and `configure`.
Docs: https://github.com/conan-io/docs/pull/2562

Some notes:
- As suggested in [this issue](https://github.com/conan-io/conan/issues/11264), we are using now DESTDIR argument for make install.
- As a side effect, we can remove the second configure in install and put the `build_script_folder` argument back in the configure method instead of having it in the constructor.
- Also, now that we don't have to use the package_folder for the --prefix argument, I have removed the `default_configure_install_args` attribute and moved the default install arguments to the toolchain.

Closes: https://github.com/conan-io/conan/issues/11265
Closes: https://github.com/conan-io/conan/issues/11264


conans/assets/templates/new_v2_autotools.py
```diff
@@ -38,6 +37,7 @@ def layout(self):
 
         def generate(self):
             at_toolchain = AutotoolsToolchain(self)
+            at_toolchain.configure_args = at_toolchain.shared_configure_args()
```
memsharded: Don't love this UI, seems a bit spaguetti, as everything is inside ``AutotoolsToolchain``, it seems like something like ``at_toolchain.add_shared_flags()`` would be cleaner?

conan/tools/gnu/autotools.py
```diff
@@ -75,28 +76,26 @@ def _fix_install_name(lib_name, lib_folder):
...
-                if not _is_modified_install_name(shared_lib, full_folder):
+                if not _is_modified_install_name(shared_lib, full_folder, libdir):
                     _fix_install_name(shared_lib, full_folder)
 
     def install(self):
-        # FIXME: we have to run configure twice because the local flow won't work otherwise
-        #  because there's no package_folder until the package step
-        self.configure()
-        self.make(target="install")
+        self.make(target="install", args=["DESTDIR={}".format(self._conanfile.package_folder)])
```
mattyclarkson: If one sets `default_configure_install_args` to `False`, they may not want the `DESTDIR` either.

Having used the tools in anger for a bit, having extra arguments added under the hood seems to be an anti pattern and hides the magic.

Personally, I would much prefer if `autotoolstoolchain` auto populated the `autoreconf_args`/`configure_args`/`make_args` and I could modify them as I see fit before generation.

Usage would then be:

```py
    def generate(self) -> None:
        at = AutotoolsToolchain(self)
        print(at.autoreconf_args) # has the `--force` and `--install` arguments in
        print(at.configure_args) # has the `--prefix` and `--shared` arguments in
        print(at.make_args) # has `DESTDIR`  and `-j` in it
        # I know _exactly_ will be passed to `autoreconf`, `configure` and `make`
```

I'd even argue adding `--enable-checking=fatal` would be a good default that users can easily remove with:

```py
        at.configure_args = [a for a in at.configure_args if a.startswith("--enable-checking")]
```

Then, if I have a really wonky package I can customise the arguments entirely. There are so many odd autotools packages in the wild.

czoido: Thanks a lot for the review @mattyclarkson, I think something in that direction would make sense, I'll add some changes to see how it looks

conans/assets/templates/new_v2_autotools.py
```diff
@@ -38,6 +37,7 @@ def layout(self):
 
         def generate(self):
             at_toolchain = AutotoolsToolchain(self)
+            at_toolchain.configure_args = at_toolchain.shared_flags()
```
memsharded: So now the management of shared flag is not by default?

memsharded (approved):

conans/assets/templates/new_v2_autotools.py
czoido: I was thinking that maybe we can make it by default again as you have the possibility now to override all defaults...

memsharded: Yes, it would make sense, if other build system integrations as ``CMakeToolchain`` already handle the ``shared`` option, automatically, it seems that this could be the default here, and allow opt-out

memsharded (approved):

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -167,7 +168,7 @@ def generate(self, env=None, scope="build"):
         self.generate_args()
         VCVars(self._conanfile).generate(scope=scope)
 
-    def _shared_static_args(self):
+    def _default_configure_shared_flags(self):
```
memsharded, line 171: This is not great when we are creating an exe for example, so I would only activate this when the options ``shared`` and ``fPIC`` are actually defined in the recipe (Put a note for Conan 2.0 that the ``package_type`` can also be used for this)

memsharded (approved):

memsharded:
@mattyclarkson this is looking better for me now, but it would be great if you could have a look to the changes and see if they align with your expectations.

lasote (approved):

memsharded:
Merged, because we need to start branching 1.49, but please @mattyclarkson if you have any feedback, it would be great to have it. Thanks!

mattyclarkson:
> Merged, because we need to start branching 1.49, but please @mattyclarkson if you have any feedback, it would be great to have it. Thanks!

Yo, sorry, I was AFK for a while taking a break. This looks great :tada: Thanks for the changes. We will roll to the newer versions and try it out. If we have any extra feedback, I'll make sure to create an issue but this looks in great shape for 2.0.

conan/tools/gnu/autotools.py
```diff
@@ -75,36 +65,33 @@ def _fix_install_name(lib_name, lib_folder):
...
+        self.make(target="install", args=args)
         if self._conanfile.settings.get_safe("os") == "Macos" and self._conanfile.options.get_safe("shared", False):
             self._fix_osx_shared_install_name()
 
     def autoreconf(self, args=None):
-        command = ["autoreconf"]
-        args = args or ["--force", "--install"]
-        command.extend(args)
-        command = join_arguments(command)
+        args = args or []
```
mattyclarkson, line 93: Whilst getting familiar with the Conan code base, I noticed this Python anti-pattern.

The code base could simplify the code flow by moving towards immutable tuples rather than mutable lists.

```py
# Python3 with expandable iterables syntax
def autoreconf(self, args=("--force", "--install")):
    command = join_arguments(("autoreconf", self._autoreconf_args, *args))
    with chdir(self, self._conanfile.source_folder):
        self._conanfile.run(command)

# Python2/3 with extending from iterables
def autoreconf(self, args=("--force", "--install")):
    command = ["autoreconf"]
    command.extend(self._autoreconf_args)
    command.extend(args)
    with chdir(self, self._conanfile.source_folder):
        self._conanfile.run(command)
```

The above is totally valid because the static binding of the tuple to the `args` is immutable it is not possible to change the bound value within the function. This is a common problem in code bases that use mutable bindings in argument defaults which we are avoiding by using `None`. i.e. the following would break terribly:

```py
def autoreconf(self, args=["--force", "--install"]):
    args.append("foobar")  # everytime `autoreconf` is invoked `args` gets longer
```

In general, I personally use immutable values as much as possible to avoid unintended side effects. It's a shame that Python doesn't have records for immutable dictionaries.

conan/tools/gnu/autotools.py
```diff
@@ -75,36 +65,33 @@ def _fix_install_name(lib_name, lib_folder):
...
+        command = join_arguments(["autoreconf", self._autoreconf_args, args_to_string(args)])
         with chdir(self, self._conanfile.source_folder):
             self._conanfile.run(command)
```
mattyclarkson, line 96: It seems a common pattern in Conan that `self.run` needs to join arguments. It would be pretty neat if, run could accept an iterable and automatically join them:

```py
    def run(self, command, output=True, cwd=None, win_bash=False, subsystem=None, msys_mingw=True,
            ignore_errors=False, run_environment=False, with_login=True, env="conanbuild"):
        if not isinstance(co```py
    def run(self, command, output=True, cwd=None, win_bash=False, subsystem=None, msys_mingw=True,
            ignore_errors=False, run_environment=False, with_login=True, env="conanbuild"):
        if not isinstance(command, str) and isinstance(command, abc.Iterable):
            command = join_arguments(command)
```mmand, str) and isinstance(command, abc.Iterable):
            command = join_arguments(command)
```

For Conan v2.0 where we can shed Python 2, I'd change that API to force keywords and allow passing multiple arguments:

```py
    def run(self, *command, output=True, cwd=None, win_bash=False, subsystem=None, msys_mingw=True,
            ignore_errors=False, run_environment=False, with_login=True, env="conanbuild"):
        command = join_arguments(command)
```

Then it can be used more naturally:

```py
self.run("autoreconf", "--force", "--install", env="conanrun")
```

PR #11294, opened 2022-05-18
Add error if missing CMAKE_BUILD_TYPE when invoking CMake

### LINKED ISSUE #11168: Warn for CMakeDeps without ``-DCMAKE_BUILD_TYPE`` as it doesn't found headers
It results that not passing ``-DCMAKE_BUILD_TYPE=Release`` to CMake command line invocation, in single-config envs like Linux while using ``CMakeDeps`` results in headers not found, because it relies on generators expressions based on the build_type, which is not defined (it is not that it is Release or Debug by default).

So adding a warning for it (it doesn't seem easy to avoid this, except re-writing the whole generator to be single-config for those cases), could at least help to mitigate this.

Original issue: 

Thanks for the info, I can confirm that adding `-DCMAKE_BUILD_TYPE=Debug` fixes the issue; I will also try with the preset. Still, maybe the toolchain file could detect this and warn about a missing build type setting, instead of silently failing to append any flags to the targets?

_Originally posted by @MiSo1289 in https://github.com/conan-io/conan/issues/11167#issuecomment-1117329226_

czoido:
Changelog: Feature: Raise an error when running CMake if CMAKE_BUILD_TYPE is not defined and the generator is not multi-config.
Docs: https://github.com/conan-io/docs/pull/2557

I initially thought about making this a warning, but I think that throwing an error makes more sense, as having no build type will not work for unexpected reasons...

Closes: https://github.com/conan-io/conan/issues/11168

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -45,6 +45,13 @@ class ConfigTemplate(CMakeDepsFileTemplate):
             message(FATAL_ERROR "The 'CMakeDeps' generator only works with CMake >= 3.15")
         endif()
 
+        # Check that the -DCMAKE_BUILD_TYPE argument is always present
+        get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+        if(NOT isMultiConfig AND NOT CMAKE_BUILD_TYPE)
+            message(FATAL_ERROR "Please, set the CMAKE_BUILD_TYPE variable when calling to CMake "
+                                "adding the '-DCMAKE_BUILD_TYPE=<build_type>' argument.")
+        endif()
+
         {% if is_module %}
         include(FindPackageHandleStandardArgs)
         set({{ file_name }}_FOUND 1)
--- a/conan/tools/cmake/cmakedeps/templates/config.py
+++ b/conan/tools/cmake/cmakedeps/templates/config.py
@@ -45,13 +45,6 @@ class ConfigTemplate(CMakeDepsFileTemplate):
             message(FATAL_ERROR "The 'CMakeDeps' generator only works with CMake >= 3.15")
         endif()
 
-        # Check that the -DCMAKE_BUILD_TYPE argument is always present
-        get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
-        if(NOT isMultiConfig AND NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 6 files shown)


conan/tools/cmake/cmakedeps/templates/config.py
```diff
@@ -45,6 +45,13 @@ def template(self):
             message(FATAL_ERROR "The 'CMakeDeps' generator only works with CMake >= 3.15")
         endif()
 
+        # Check that the -DCMAKE_BUILD_TYPE argument is always present
+        get_property(isMultiConfig GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
```
memsharded: If not parameterized dependent on the package, maybe move it to a CMake macro in the ``macros.py``, so it is not repeated in every file?

memsharded (approved):
Looks good

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -240,7 +240,7 @@ class Conan(ConanFile):
 
     client.run("create dep")
     client.run("install .")
-    client.run_command("cmake . -DCMAKE_TOOLCHAIN_FILE=./conan_toolchain.cmake "
+    client.run_command("cmake . -DCMAKE_TOOLCHAIN_FILE=./conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Release"
```
memsharded: Maybe only add this for Linux and Mac, but not Windows? 
My concern is that we could be forcing to define it, even for multiconfig if there is some bug there, because in our testing we keep passing CMAKE_BUILD_TYPE also for multiconfig

conans/test/functional/toolchains/cmake/test_cmaketoolchain_paths.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, package=None, library=None, framework=None, include=None, pro
 
 
 def _cmake_command_toolchain(find_root_path_modes):
-    cmake_command = "cmake .. -DCMAKE_TOOLCHAIN_FILE=../conan_toolchain.cmake"
+    cmake_command = "cmake .. -DCMAKE_TOOLCHAIN_FILE=../conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Release"
```
memsharded: Same

memsharded (approved):

lasote (approved):

memsharded (approved):

memsharded (approved):

PR #11301, opened 2022-05-19
Try ANSI encoding before Windows-1252 on Windows

### LINKED ISSUE #11076: [bug] Running with win_bash=True handles encoding incorrectly
I made an program that outputs some non-ASCII characters and it works well on Windows or inside MSYS2, but when I try to run it inside Conan by invoking the `run` method with `win_bash=True`, the output contains garbage contents. This problem makes it difficult to read the Visual Studio (with localized output) error messages during `conan build`.

  * Compiler+version: Visual Studio 2019
  * Conan version: 1.47.0
  * Python version: built-in with Conan installer

Create a folder with the following 2 files:
- conanfile.py
```
from conans import ConanFile, tools
import os.path


class MsysTestConan(ConanFile):
    name = "msys-test"
    version = "1.0.0"

    def build_requirements(self):
        self.build_requires("msys2/cci.latest")

    def build(self):
        with tools.vcvars(self):
            self.run("cl {}".format(os.path.join(self.source_folder, "main.cpp")))
            self.run("./main.exe", win_bash=True)
            self.run("main.exe")
```
- main.cpp (must be saved in `UTF-8 with BOM`)
```
#include <iostream>

using namespace std;

int main() {
    cout << "你好\n";
    cout << "Olá\n";
}
```

Run this in PowerShell:
```
mkdir build
cd build
conan install ..
conan build ..
```

The final output is
```
ÄãºÃ
Ol¨¢
你好
Olá
```

Directly running the `main.exe` in PowerShell or MSYS2 bash outputs:
```
你好
Olá
```

which shows that `self.run("./main.exe", win_bash=True)` does not prints the correct output on the console.

```
C:\Users\taoyouh\source\playgrounds\conan-encoding\build>conan install ..
```

```
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=16
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

conanfile.py (msys-test/1.0.0): Installing package
Requirements
Packages
Build requirements
    msys2/cci.latest from 'conancenter' - Cache
Build requirements packages
    msys2/cci.latest:eee3fba89db6d777329de604625af8c30d46f080 - Cache

msys2/cci.latest: Already installed!
msys2/cci.latest: Creating MSYS_ROOT env var : C:\.conan\6696ea\1\bin\msys64
msys2/cci.latest: Creating MSYS_BIN env var : C:\.conan\6696ea\1\bin\msys64\usr\bin
msys2/cci.latest: Appending PATH env var with : C:\.conan\6696ea\1\bin\msys64\usr\bin
conanfile.py (msys-test/1.0.0): Applying build-requirement: msys2/cci.latest
conanfile.py (msys-test/1.0.0): Generator txt created conanbuildinfo.txt
conanfile.py (msys-test/1.0.0): Aggregating env generators
conanfile.py (msys-test/1.0.0): Generated conaninfo.txt
conanfile.py (msys-test/1.0.0): Generated graphinfo
```

```
C:\Users\taoyouh\source\playgrounds\conan-encoding\build>conan build ..
```
```
Using lockfile: 'C:\Users\taoyouh\source\playgrounds\conan-encoding\build/conan.lock'
Using cached profile from lockfile
conanfile.py (msys-test/1.0.0): Calling build()
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.11.11
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.29.30141 版
版权所有(C) Microsoft Corporation。保留所有权利。

main.cpp
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30133\include\ostream(746): warning C4530: 使用了 C++ 异常处
理程序，但未启用展开语义。请指定 /EHsc
C:\Users\taoyouh\source\playgrounds\conan-encoding\main.cpp(6): note: 查看对正在编译的函数 模板 实例化“std::basic_ostream<char,std::char_trait
s<char>> &std::operator <<<std::char_traits<char>>(std::basic_ostream<char,std::char_traits<char>> &,const char *)”的引用
Microsoft (R) Incremental Linker Version 14.29.30141.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:main.exe
main.obj
conanfile.py (msys-test/1.0.0): run_in_windows_bash: C:\.conan\6696ea\1\bin\msys64\usr\bin\bash.exe --login -c ^"cd \^"/c/users/taoyouh/source/
playgrounds/conan-encoding/build\^" ^&^& PATH=\^"/c/.conan/6696ea/1/bin/msys64/usr/bin:$PATH\^" ^&^& ./main.exe ^"
ÄãºÃ
Ol¨¢
你好
Olá

```

taoyouh:
Fix #11076

Changelog: Fix: outputs of subprocess using ANSI encoding are not decoded properly
Docs: no doc describes this part of Conan

### DIFF
--- a/conans/util/files.py
+++ b/conans/util/files.py
@@ -88,7 +88,10 @@ def _detect_encoding(text):
                 return encodings[bom], len(bom)
             except UnicodeDecodeError:
                 continue
-    decoders = ["utf-8", "Windows-1252"]
+    if sys.platform == "win32":
+        decoders = ["utf-8", "mbcs", "Windows-1252"]
+    else:
+        decoders = ["utf-8", "Windows-1252"]
     for decoder in decoders:
         try:
             text.decode(decoder)


conans/util/files.py
```diff
@@ -88,7 +88,10 @@ def _detect_encoding(text):
                 return encodings[bom], len(bom)
             except UnicodeDecodeError:
                 continue
-    decoders = ["utf-8", "Windows-1252"]
+    if sys.platform == "win32":
+        decoders = ["utf-8", "mbcs", "Windows-1252"]
```
memsharded, line 92: This is a bit risky. At the very least I would try to:
- Add the encoder later, after ``Windows-1252``
- Add some test that reproduce the error and is solved by this change

taoyouh: > This is a bit risky. At the very least I would try to:
> 
> * Add the encoder later, after `Windows-1252`
> * Add some test that reproduce the error and is solved by this change

I'm afarid that adding the encoder later may not work. `Windows-1252` is a single-byte code page and it covers almost every available value for a byte (with only 5 unused code points). For command line outputs, most of the short lines will succeed with `Windows-1252` decoder (which produces unreadable contents).

Testing this error requires altering the system locale settings. I'm still looking for a way to handle this.

SSE4: 1252 is just one of the possible encodings, popular in certain European countries I assume? e.g. In my location (Asia) 1251 is much more wide-spread. I assume to fix a problem "world-wide" we would need to try a bunch of different encodings?

taoyouh: > 1252 is just one of the possible encodings, popular in certain European countries I assume? e.g. In my location (Asia) 1251 is much more wide-spread. I assume to fix a problem "world-wide" we would need to try a bunch of different encodings?

I don't think it is possible to try those encodings. Most of mis-decoded contents are totally legal data, but only non-readable by human.

SSE4: yes, that's the problem. conan cannot know if an arbitrary file is cp1251, cp866 or koi8r. it can be decoded via any of the above, it will be just human-unreadable. unless conan aplies some heuristics (e.g. frequency of characters), there is no reliable way to deal with that code-page.
but if I read the issue correctly, it's child process launched by conan has a different text encoding from the original process, which caueses troubles. in that way I would prefer if child terminal just started with the "proper" (= same as parent) encoding.

taoyouh: The way that conan handles normal Windows subprocess is to send them directly to the console without decoding them. If we do the same for bash subprocess (using PR #11306), at least the display will be correct (but using custom `output` still has problem).

memsharded:
This is changing a lot in Conan 2.0, as this is a bit risky, it is better not to move it forward for Conan 1.X.

In Conan 2.0:
- All ``conanfile.py`` python files must be utf-8. No other encoding for conanfile.py will be supported.
- The only possible file that Conan could read that the encoding could vary would be the ``conanfile.txt`` files. These will still be checked for potential different encodings. Still for most cases, it would be recommended to use a utf-8 ``conanfile.py`` also for consuming-only conanfiles. If this is still an issue, I'd recommend opening a PR for ``develop2`` branch, there is a test that covers the different encodings use case: https://github.com/conan-io/conan/blob/develop2/conans/test/integration/conanfile/test_conanfile_txt_encodings.py

So closing this, thanks very much for contributing!

Issue #11261, opened 2022-05-16
[bug] Generated CMakeDeps modules may use wrong name for transitive dependencies


jcar87:
### Environment Details (include every applicable attribute)
  * Operating System+version: macOS 12.3
  * Compiler+version: AppleClang 13
  * Conan version: 1.48
  * Python version: 3.8.9

### Steps to reproduce (Include if Applicable)

Using the `CMakeDeps` generator in a consumer package that consumes `pkgB`:
* `pkgB` has a requirement on `pkgA`. 
* `pkgA` defines the `cmake_module_file_name` [property](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html#properties) in `package_info()` - to a value that is *not* the same as `cmake_file_name` (typically it will be the same name, but different use of upper/lower case)
* `pkgA` sets `cmake_find_mode` to `both` 

In the consumer package, when the `find_package(pkgB)` call searches for _modules_ first (either by explicitly setting the `MODULE` option in the `find_package()` call, or by overriding the [CMakeToolchain](https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmaketoolchain.html) behaviour of  `CMAKE_FIND_PACKAGE_PREFER_CONFIG` (setting it to `OFF`).

We may get three types of issues:

* On case *insensitive* filesystems (e.g. as default on windows and macos), CMake might actually find the module, and may issue a warning regarding a mismatch between the package name and the defined `<pkg_name>_FOUND` variable.
* In the case above, if the transitive dependency (`pkgA`) defines `cmake_module_target_name` to have a different name for modules, CMake will fail the generating stage if the target names it was expecting to exist are not defined (because they have a different name, and in this case, the target names are case sensitive across all platforms).
* On case sensitive filesystems the find_package call may fail altogether. 

The underlying issue is in the generated `module-xxx-data.cmake` file:
```
set(pkgB_COMPONENT_NAMES "")
list(APPEND pkgB_FIND_DEPENDENCY_NAMES pkgA)


set(pkgA_FIND_MODE "MODULE")
```

If the module name is `PKGA` (all upper case), it should be reflected in both the list and the `xxx_FIND_MODE` variable.

jcar87:
Closed via #11307

PR #11307, opened 2022-05-19
Bugfix/issue 11261 transitive module dependency wrong name

jcar87:
Changelog: Bugfix: Fix case where CMakeDeps generator may use the wrong dependency name for transitive dependencies.
Docs: Omit

This PR fixes issue #11261, where the generated _module_ files in the CMakeDeps generator _may_ use the wrong dependency name for transitive dependencies.  The issue was only happening when the following 3 conditions were true:
* we are generating a Find module
* we have a transitive dependency where `cmake_find_mode` is explicitly `both`
* the same transitive dependency defines a module name that is different from the config name (by setting the `cmake_module_file_name` property)

This PR fixes the issue and adds a test case that exercises exactly those 3 conditions. This should also fix a related issue where CMake may issue a _warning_ on case insensitive filesystems where the name of the `xxx_FOUND` variable does not match the name of the package passed to `find_package()`. 

Additionally:

* `get_file_name` has been renamed to `get_cmake_package_name` to try and reflect better how it's used - comments have been added to clarify that the function is not only used to determine the file name, but also the names of transitive dependencies and the defined variables. In CMake, for the same package all of the above need to be "the same" for the same package.

* In the CMakeDeps template, move the logic of the code that does `set({{ file_name }}_FOUND 1)` to the bottom of the file, _after_ all transitive dependencies have been found. Due to https://github.com/conan-io/conan/issues/11261 (which is still an outstanding issue), if the call to `find_dependency` is not successful, the top-level call to `find_package` may appear successful and then cause CMake failures further down the line. This would cause an early termination that should help troubleshooting.

* Add the `find_mode` flag when calling `get_cmake_package_name` in `conan/tools/cmake/cmakedeps/templates/target_data.py` - this is mostly what fixes the bug in this PR, and what was causing the erroneous behaviour of using the `config` name when we were generating a ` module`

* Add a test that covers the behaviour that we are trying to fix. This also sets the module name of the transitive dependency to something that is not a variation of lower/upper case, such that we can reproduce the same behaviour across filesystems (we were getting different results in case in/sensitive filesystems)

* Fix small issue with the test where the variable names were swapped (but were still testing the right thing)

### DIFF
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
@@ -65,52 +65,52 @@ def client():
     return t
 
 
-# @pytest.mark.tool_cmake
-# def test_reuse_with_modules_and_config(client):
-#     cpp = gen_function_cpp(name="main")
+@pytest.mark.tool_cmake
+def test_reuse_with_modules_and_config(client):
+    cpp = gen_function_cpp(name="main")
 
-#     cmake_exe_config = """
-#     add_executable(myapp main.cpp)
-#     find_package(MyDep) # This one will find the config
-#     target_link_libraries(myapp MyDepTarget::MyCrispinTarget)
-#     """
+    cmake_exe_config = """
+    add_executable(myapp main.cpp)
+    find_package(MyDep) # This one will find the config
+    target_link_libraries(myapp MyDepTarget::MyCrispinTarget)
+    """
 
-#     cmake_exe_module = """
-#     add_executable(myapp2 main.cpp)
-#     find_package(mi_dependencia) # This one will find the module
-#     target_link_libraries(myapp2 mi_dependencia_namespace::mi_crispin_target)
-#     """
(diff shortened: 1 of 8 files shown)


conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_find_module_and_config.py
```diff
@@ -115,10 +115,11 @@ def test_reuse_with_modules_and_config(client):
 
 find_modes = [
     ("both", "both", ""),
+    ("both", "both", "MODULE"),
```
jcar87, line 118: This is the new test case being added, where were would expect the generated `FindPKGB.cmake` to find PKGA using the module name specified by PKGA (which was not the case)

lasote (approved):
Nice and clear! 🎉  

I'm wondering if should we rename also the `self.find_module_mode` to something like `self.current_node_is_module` or `self.generating_module`. WDYT?

jcar87:
> self.generating_module

I like this one!! Will fix

lasote (approved):

PR #11308, opened 2022-05-19
Feature/cmake presets custom settings

### LINKED ISSUE #11189: [feature] Multiple parallel profiles support in `cmake_layout()`
Until now I used a simple workflow for each of the compilers using `Ninja Multi-Config` as the default CMake generator:

```
mkdir -p build/GCC-11 && cd build/GCC-11
conan install ../.. -pr gcc11
conan install ../.. -pr gcc11 -s build_type=Debug
cmake ../.. --toolchain conan_toolchain.cmake
```

I also set the following in VSCode which makes it work like a charm (with just 3 first steps from the above):

- `cmake.buildDirectory` to `${workspaceFolder}/build/${buildKitVendor}-${buildKitVersionMajor}`
- `cmake.configureArgs` to `--toolchain conan_toolchain.cmake`

However, the introduction of `cmake_layout()` as a recommended practice in modern Conan recipes will make the above a lot harder if not impossible so I am looking for the best alternative.

Right now while using `cmake_layout()`, no matter what the profile file or a current directory is, all the artifacts land out in:
- `<project_root>/build/generators`
- either `<project_root>/build` (muti-config) or `<project_root>/cmake-build-<build_type>` (single-config)

The above scenario accounts only for the differences in the `build_type` between several Conan runs. However, there are more variables there: compiler, compiler version, C++ version used, the standard library used, ... If we run Conan for such configurations the artifacts will collide as they will land out in the same subdirectories. We can use `-of <dir>` (i.e. `-of build/GCC-11`) to handle the differences manually but:
- `CMakeUserPresets.json` will not land in the correct place to be found by CMake invocations so one must copy it (and override) to the project source directory before every CMake build.
- There will be no aggregated `CMakeUserPresets.json` that will allow opening such a tree of configurations in the IDE. Even if there was one, all the preset names would collide as they do not use any configuration-specific prefix today.
- We might end up with some strange paths to work with like `build/GCC-11/build/Release` and `build/GCC-11/build/generators`

I strongly believe that is a good practice to check the source code on several different compilers before committing changes to the repo. Also such configurations are useful to easily reproduce issues raised by the customers or CI. It would be good if Conan could still facilitate such scenarios with the new `cmake_layout()` in the recipes.

Unfortunately, I do not have a production-ready solution at hand right now but I hope that we might find one together. The simplest one would be to use a profile name as a prefix/discriminator if it is different from the `default`. That, however, does not account for settings changes provided from the command line (like `conan install . -pr gcc11 -s compiler .version=10`) but I do not believe this is a common practice. Probably the most common is `conan install . -pr gcc11 -s build_type=Debug` which would be fine as those already land-out under the same file tree.

lasote:
Changelog: Feature: Improved `cmake_layout` and `CMakePresets.json` feature so you can manage different configurations using the same `CMakeUserPresets.json` not only for multi-config (Debug/Release) but for any set of settings specified in a new conf `tools.cmake.cmake_layout:build_folder_vars` that accepts a list of settings to use. e.g `tools.cmake.cmake_layout:build_folder_vars=["settings.compiler", "options.shared"]`
Docs: https://github.com/conan-io/docs/pull/2561

Close https://github.com/conan-io/conan/issues/11189

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -48,20 +48,19 @@ def get_build_folder_vars_suffix(conanfile):
                                     default=[], check_type=list)
     ret = []
     for s in build_vars:
+        tmp = None
         if s.startswith("settings."):
             _, var = s.split("settings.", 1)
             tmp = conanfile.settings.get_safe(var)
         elif s.startswith("options."):
             _, var = s.split("options.", 1)
-            tmp = "{}_{}".format(var, conanfile.options.get_safe(var))
-
+            value = conanfile.options.get_safe(var)
+            if value:
+                tmp = "{}_{}".format(var, value)
         else:
             raise ConanException("Invalid 'tools.cmake.cmake_layout.build_folder_vars' value, it has"
                                  " to start with 'settings.' or 'options.': {}".format(s))
         if tmp:
             ret.append(tmp.lower())
-
-    if not ret:
-        return ""
 
     return "-".join(ret)
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -19,13 +19,20 @@ def cmake_layout(conanfile, generator=None, src_folder="."):
         build_type = str(conanfile.settings.build_type)
     except ConanException:
         raise ConanException("'build_type' setting not defined, it is necessary for cmake_layout()")
+
+    suffix = get_custom_settings_suffix(conanfile)
     if multi:
         conanfile.folders.build = "build"
     else:
-        build_type = build_type.lower()
         conanfile.folders.build = "cmake-build-{}".format(build_type)
 
+    if suffix:
+        conanfile.folders.build += "-{}".format(suffix)
+
     conanfile.folders.generators = os.path.join("build", "generators")
+    if suffix:
+        conanfile.folders.generators += "-{}".format(suffix)
+
     conanfile.cpp.source.includedirs = ["include"]
 
     if multi:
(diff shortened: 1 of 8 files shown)


conans/model/conf.py
```diff
@@ -21,6 +21,7 @@
     "tools.cmake.cmaketoolchain:system_version": "Define CMAKE_SYSTEM_VERSION in CMakeToolchain",
     "tools.cmake.cmaketoolchain:system_processor": "Define CMAKE_SYSTEM_PROCESSOR in CMakeToolchain",
     "tools.env.virtualenv:auto_use": "Automatically activate virtualenv file generation",
+    "tools.cmake.cmake_layout.custom_settings": "Name of the settings that will produce a different build folder and different CMake presets names",
```
lasote: Please, I need a name for the conf, that is not a good one.

conan/tools/cmake/layout.py
```diff
@@ -34,3 +40,19 @@ def cmake_layout(conanfile, generator=None, src_folder="."):
...
+    build_settings = conanfile.conf.get("tools.cmake.cmake_layout.custom_settings",
+                                        default=[], check_type=list)
+    ret = []
+    for s in build_settings:
+        tmp = conanfile.settings.get_safe(s)
+        if tmp:
+            ret.append(tmp.lower())
+
+    if not ret:
+        return ""
```
franramirez688: I think this is not needed. `return "-".join(ret)` return `""` if ret is empty.

conans/model/conf.py
memsharded: ``tools.cmake.cmake_layout:build_folder_vars`` (could be ``params``, or something like that)

memsharded (approved):
This is really good.
The only thing I would consider is adding support for options too. Building something shared/static is sure one of the very popular variations, it would be good if it could be supported too.

That would mean that the configuration needs to be something like:
```
[conf]
tools.cmake.cmake_layout:build_folder_vars = ["settings.build_type", "settings.compiler", "setttings.compiler.version", "options.shared"]
```

franramirez688 (approved):

conans/model/conf.py
franramirez688: and `tools.cmake.cmake_layout:suffix_folder_vars`?

lasote: I think `tools.cmake.cmake_layout:build_folder_vars` is good enough.

memsharded: I think having the ``build`` in the name is good, as this maps to the ``self.build_folder`` (or ``self.folders.build`` inside the layout)

czoido (approved):

PR #11310, opened 2022-05-19
[develop2] Feature/develop2 package signing

### LINKED ISSUE #11235: [feature] Investigate the impact of adding new files to upload/download in the export and package folders
As a mechanism to be able to add metadata to packages and recipes, signatures, etc. Is it breaking to add extra files? Can it be fixed, optin or something?

memsharded:
Very early stage, to start discussion


conans/test/integration/test_pkg_signing.py
```diff
@@ -0,0 +1,43 @@
...
+            "export_sources/file2.txt": "file2!"})
+    signer = textwrap.dedent(r"""
+        import os
+
+        def sign(ref, files, folder):
+            print("Signing ref: ", ref)
+            print("Signing files: ", files)
+            signature = os.path.join(folder, "signature.asc")
+            open(signature, "w").write("\n".join(files))
+            return ["signature.asc"]
```
memsharded: This is a bit ugly, needing to return the name of the created files, but it is necessary to add the new files to the "upload bundle". We could try to auto-detect which files are new, but sounds trickier.

File paths are also absolute, I don't love it, but this is necessary because they can be in different folders, as ``conanfile.py`` and ``conanmanifest.txt`` for example might be in the ``export`` folder, while the ``conan_sources.tgz`` is created directly in the ``download`` folder, and the 3 files are the final package. It would be more elegant if just one folder of all files is passed, but for that we would need to copy ``conanfile.py`` and manifest from export->download. Is doable, but not what we are doing now.

conans/test/integration/test_pkg_signing.py
```diff
@@ -0,0 +1,43 @@
...
+
+        def verify(ref, files, folder):
```
memsharded: ``files`` is the list of files, excluding already the metadata, and they are absolute paths, while ``folder`` is the ``download/metatada`` folder

conans/test/integration/test_pkg_signing.py
lasote: I'm ok with the return, as far as it is documented sounds reasonable.
About the absolute paths, the same, is similar to the `cpp_info` fields, having the absolute path is very convenient.

conans/test/integration/test_pkg_signing.py
lasote: This interface looks indeed weird. I don't like the `folder` argument. I might prefer a `metadata_files` with the absolute paths, but I see that it would be confusing to introduce the `metadata` concept when the plugin is only about signing. But maybe this plugin could always write the returned files to `metadata/signing`, and then pass the `signed_files` instead of `folder`.

memsharded: It can be named ``metadata_folder`` instead of ``folder``? If passed a list of files, then users would need to iterate or find their wanted files in that list, instead of ``os.path.join(metadata_folder, "myfilename")``?

Other alternative is just passing the parent folder (this is the "download" one), which will contain:
- "donwnload" folder
   - conanfile.py
   - conanmanifest.txt
   - conan_sources.tgz
   - metadata
      - signature.asc
     
And then let the plugin itself should know the layout and relative ``metadata/signature.asc`` path?

lasote: >  If passed a list of files, then users would need to iterate or find their wanted files in that list

That is why I suggested a subfolder in the metadata for the signing, not only metadata. So everything in that folder "should" be of interest.

memsharded:
Did a few changes:
- Simplify the API, not passing files, just folders
- The folder passed is the "download" folder, the place where .tgz lives
- Signing does not have to return nothing
- Works by convention, plugins should save to ``metatada/sign/xxxx`` and read from there
- Forced the copy of files like conanfile.py, conaninfo.txt and conanmanifest.txt to the "download" folder

Please re-check

conans/client/pkg_sign.py
```diff
@@ -22,20 +22,21 @@ def sign(self, upload_data):
             return
 
         def _sign(ref, files, folder):
-            mkdir(folder)
-            signatures = self._sign(ref, list(files.values()), folder=folder)
-            for signature in signatures:
-                files[f"{METADATA}/{signature}"] = f"{folder}/{signature}"
+            metadata_sign = os.path.join(folder, METADATA, "sign")
```
lasote, line 25: Not sure if I like this "hardcoding" the `metadata/sign` but then letting the user plugin write in the correct place.

conans/test/integration/test_pkg_signing.py
```diff
@@ -15,21 +15,26 @@ def test_pkg_sign():
...
+        def sign(ref, folder):
             print("Signing ref: ", ref)
-            print("Signing files: ", files)
-            signature = os.path.join(folder, "signature.asc")
+            print("Signing folder: ", folder)
+            files = []
+            for f in sorted(os.listdir(folder)):
+                if os.path.isfile(os.path.join(folder, f)):
+                    files.append(f)
+            signature = os.path.join(folder, "metadata", "sign", "signature.asc")
```
lasote: Related to my previous comment, here the user needs to compose the correct path because it HAS to be that path, so, it looks weird.

conans/client/cmd/uploader.py
```diff
@@ -187,8 +190,19 @@ def _compress_recipe_files(self, layout, ref):
         src_files, src_symlinked_folders = gather_files(export_src_folder)
         src_files.update(src_symlinked_folders)
 
-        result = {CONANFILE: files.pop(CONANFILE),
-                  CONAN_MANIFEST: files.pop(CONAN_MANIFEST)}
+        # We do a copy of conanfile and conanmanifest to the download_export_folder
```
lasote, line 193: I guess this fixes the need of passing absolute paths (previous review). 
This might be confusing, not sure it is worth it. For me, the abs paths were kind of ok.

lasote (commented):
I'm not convinced. At some points looks like this plugin might be more general, like "metadata" oriented more than "signing" oriented but in other parts is restricted. I would say we still need to iterate this.

conans/client/pkg_sign.py
memsharded: Updated the interface to be explicit: ``def sign(ref, artifacts_folder, signature_folder):``

conans/test/integration/test_pkg_signing.py
memsharded: Updated the interface to be explicit: ``def sign(ref, artifacts_folder, signature_folder):``

conans/client/cmd/uploader.py
memsharded: It is not about the absolute paths. The paths are still absolute, check:
```python
result = {CONANFILE: os.path.join(download_export_folder, CONANFILE),
                  CONAN_MANIFEST: os.path.join(download_export_folder, CONAN_MANIFEST)}
```
It is doing a copy of the files to the "download" location, and using that location for the upload, in the same way it is used in the download. I think this can simplify later in the future if creating tarballs of packages in the cache from the "upload_bundle" is allowed, and then restoring that tarball directly in another cache.

conans/client/pkg_sign.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class PkgSignaturesPlugin:
+    def __init__(self, cache):
+        self._cache = cache
+        signer = os.path.join(cache.plugins_path, "sign", "sign.py")
+        if os.path.isfile(signer):
+            mod, _ = load_python_file(signer)
+            # TODO: At the moment it requires both methods sign and verify, but that might be relaxed
+            self._sign = mod.sign
```
lasote: Could we rename it to  "mod_sign"? it is a bit confusing to review with the `def _sign`

conans/test/integration/test_pkg_signing.py
```diff
@@ -0,0 +1,47 @@
...
+            "export/file1.txt": "file1!",
+            "export_sources/file2.txt": "file2!"})
+    signer = textwrap.dedent(r"""
+        import os
+
+        def sign(ref, artifacts_folder, signature_folder):
+            print("Signing ref: ", ref)
+            print("Signing folder: ", artifacts_folder)
+            files = []
+            for f in sorted(os.listdir(artifacts_folder)):
```
lasote, line 22: Might be possible that the `signature_folder` contains something that would be signed again?

lasote (approved):
Is better now. More clear for the user, and less error prone.

conans/client/pkg_sign.py
memsharded: Sure, done!

conans/test/integration/test_pkg_signing.py
memsharded: Not sure what you mean, the code is getting just the files with:
```python
for f in sorted(os.listdir(artifacts_folder)):
                if os.path.isfile(os.path.join(artifacts_folder, f)):
                    files.append(f)
```
It is true that users could recurse into the ``download/metadata`` folder, and that is unexpected, that folder should explicitly be left out of signing, by definition.

lasote: > It is true that users could recurse into the download/metadata folder, and that is unexpected, that folder should explicitly be left out of signing, by definition.

That is what I meant.

If that is `by definiton` and you trust the users, then ok.

memsharded: Of course, I'd prefer this not be this way, but that would mean having the ``metadata`` folder in a different location, for the upload it might be possible, but still in the download they will find the ``metadata`` subfolder there, because while downloading from the server everything will be in this folder. So maybe it is better that users are aware of this fact from the very beginning, in both upload and download flows?

lasote: ok

franramirez688 (approved):

czoido (approved):

jcar87 (approved):

PR #11317, opened 2022-05-23
Support CMAKE_SYSROOT in CMakeToolchain

czoido:
Changelog: Feature: Add `CMAKE_SYSROOT` support for `CMakeToolchain`.
Changelog: Feature: Add `--sysroot` support for `AutotoolsToolchain` and remove support for `cpp_info.sysroot` in `AutotoolsDeps`.
Changelog: Feature: Add `tools.build:sysroot` conf.

Docs: https://github.com/conan-io/docs/pull/2564

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -683,6 +683,9 @@ class GenericSystemBlock(Block):
         set(CMAKE_RC_COMPILER {{ compiler_rc }})
         endif()
         {% endif %}
+        {% if cmake_sysroot %}
+        set(CMAKE_SYSROOT {{ cmake_sysroot }})
+        {% endif %}
         """)
 
     def _get_toolset(self, generator):
@@ -824,6 +827,8 @@ class GenericSystemBlock(Block):
 
         system_name, system_version, system_processor = self._get_cross_build()
 
+        cmake_sysroot = self._conanfile.conf.get("tools.cmake.cmaketoolchain:cmake_sysroot")
+
         return {"compiler": compiler,
                 "compiler_rc": compiler_rc,
                 "compiler_cpp": compiler_cpp,
@@ -831,7 +836,8 @@ class GenericSystemBlock(Block):
                 "generator_platform": generator_platform,
                 "cmake_system_name": system_name,
                 "cmake_system_version": system_version,
-                "cmake_system_processor": system_processor}
+                "cmake_system_processor": system_processor,
+                "cmake_sysroot": cmake_sysroot}
 
 
 class OutputDirsBlock(Block):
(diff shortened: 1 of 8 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -824,14 +828,18 @@ def context(self):
 
         system_name, system_version, system_processor = self._get_cross_build()
 
+        cmake_sysroot = self._conanfile.conf.get("tools.cmake.cmaketoolchain:cmake_sysroot")
```
memsharded: We have ``cpp_info.sysroot``, aren't we processing it anymore?

memsharded: Ok, yes, lets use the ``conf`` and forget about ``cpp_info.sysroot``, sounds good @lasote?

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
memsharded: The only thing, maybe better generalize the name to something like ``tools.buid:sysroot``, so it makes sense for autotools and other build systems.

memsharded (changes requested):

czoido:
> The only thing, maybe better generalize the name to something like tools.buid:sysroot, so it makes sense for autotools and other build systems.

Yes, I'll check the autotools case as well and make something generic if needed 👍

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -824,14 +828,21 @@ def context(self):
 
         system_name, system_version, system_processor = self._get_cross_build()
 
+        # This is handled by the tools.apple:sdk_path and CMAKE_OSX_SYSROOT in Apple
+        cmake_sysroot = None
+        if not is_apple_os(self._conanfile.settings.get_safe("os")):
```
memsharded: why? As this is a defined conf, is it really necessary to not apply it the OS is an apple one?
Is it possible that there there are some scenario (like cross-building or the like) that would want to use the ``sysroot``? Wouldn't it be enough that the user don't define ``tools.build:sysroot`` in their profile (or inside the package, the package recipe would know)

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -64,7 +66,12 @@ def __init__(self, conanfile, namespace=None):
                 # https://man.archlinux.org/man/clang.1.en#Target_Selection_Options
                 self.apple_arch_flag = "-arch {}".format(apple_arch) if apple_arch else None
                 # -isysroot makes all includes for your library relative to the build directory
-                self.apple_isysroot_flag = "-isysroot {}".format(sdk_path) if sdk_path else None
+                self.sysroot_flag = "-isysroot {}".format(sdk_path) if sdk_path else None
+
+        if not is_apple_os(self._conanfile.settings.get_safe("os")):
```
memsharded: Same as above?

memsharded (approved):
Looking good

conan/tools/cmake/toolchain/blocks.py
czoido: Yes, maybe it's better not to do this, I'll remove the check 👍

PR #11321, opened 2022-05-24
[feature] Add tools can_run

### LINKED ISSUE #11035: [feature] Ability to identify if binaries can be executed on a particular platform
Typically there is made use of `tools.cross_building(self.settings)` to identify that for example test binaries can't be executed on a particular platform. Though this particular solution is not really sufficient to handle for example the case of running on a macOS with Apple Silicon (M1) chipset, as in that particular case you can execute both x86_64 and armv8 binaries without any issue (even without Rosetta enabled). 

Extending `tools.cross_building` with yet another parameter similar to `skip_x64_x86` seems not a good option. Therefore it would be good to have a dedicated method for this usecase like for example `tools.can_run()`.

uilianries:
- Added under cross_building namespace
- TODO: Elaborate tests 

Changelog: Feature: Tools `can_run` validates if it is possible to run a and application build for a non-native architecture.
Docs: https://github.com/conan-io/docs/pull/2547
closes #11035

### DIFF
--- a/conan/tools/build/cross_building.py
+++ b/conan/tools/build/cross_building.py
@@ -36,6 +36,6 @@ def can_run(conanfile):
     See https://github.com/conan-io/conan/issues/11035
     """
     allowed = conanfile.conf.get("tools.build.cross_building:can_run", check_type=bool)
-    if not allowed:
+    if allowed is None:
         return not cross_building(conanfile)
-    return True
+    return allowed
--- a/conan/tools/build/cross_building.py
+++ b/conan/tools/build/cross_building.py
@@ -28,3 +28,14 @@ def get_cross_building_settings(conanfile):
                 os_host, arch_host)
     else:
         return os_host, arch_host, os_host, arch_host
+
+
+def can_run(conanfile):
+    """
+    Validates if the current build platform can run a file which is not for same arch
+    See https://github.com/conan-io/conan/issues/11035
+    """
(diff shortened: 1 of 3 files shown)


conan/tools/build/cross_building.py
```diff
@@ -28,3 +28,14 @@ def get_cross_building_settings(conanfile):
...
         return os_host, arch_host, os_host, arch_host
+
+
+def can_run(conanfile):
+    """
+    Validates if the current build platform can run a file which is not for same arch
+    See https://github.com/conan-io/conan/issues/11035
+    """
+    allowed = conanfile.conf.get("tools.build.cross_building:can_run", check_type=bool)
+    if not allowed:
```
memsharded: I would take both ``True`` and ``False`` as definitive values, and return them.

And then, only fallback to ``not cross_building()`` for the case the ``can_run`` is ``None`` (not defined by user)

memsharded (changes requested):

conan/tools/build/cross_building.py
uilianries: Yes, totally makes sense! I gonna update it!

uilianries: Done.

memsharded (approved):

memsharded:
Thanks!

PR #11325, opened 2022-05-25
Feature/docs conanfile attributes docstrings

lasote:
Docstrings for https://github.com/conan-io/docs/pull/2546


conans/model/conan_file.py
```diff
@@ -10,44 +10,205 @@
...
+
+    #: The home web page of the library being packaged.
     homepage = None
 
+    #: Controls when the current package is built during a ``conan install``.
+    #: The allowed values are:
+    #:
+    #: - ``"missing"``: Conan builds it from source if there is no binary available.
+    #: - ``"always"``: Conan always builds it from source.
+    #: - ``None`` (default value): This package won't be build unless the policy is specified
```
memsharded, line 64:
```suggestion
    #: - ``never``: This package cannot be built from sources, it is always created with ``conan export-pkg``
    #: - ``None`` (default value): This package won't be build unless the policy is specified
```

conans/model/conan_file.py
```diff
@@ -10,44 +10,205 @@
...
+    #:    headers, libraries, runtime information...
+    #:
+    #: The valid values are:
+    #:
+    #:     - **application**: The package is an application.
+    #:     - **library**: The package is a generic library. It will try to determine
+    #:       the type of library (from ``shared-library``, ``static-library``, ``header-library``)
+    #:       reading the ``self.options.shared`` (if declared) and the ``self.options.header_only``:
+    #:
+    #:          - **shared-library**: The package is a shared library.
```
memsharded: Not sure if this needs to be inside ``library`` type, might be confusing, can be at the same level

conans/model/conan_file.py
```diff
@@ -10,44 +10,205 @@
...
+    cpp = None
+
+    #: For the dependant recipes, the declared environment variables will be present during the
+    #: build process. Should be only filled in the ``package_info()`` method.
+    #: It will be initialized in the ``__init__()`` with an ``Environment()`` object.
+    buildenv_info = None
+
+    #: For the dependant recipes, the declared environment variables will be present at runtime.
+    #: Should be only filled in the ``package_info()`` method.
+    #: It will be initialized in the ``__init__()`` with an ``Environment()`` object.
```
memsharded: This reads a bit too internal implementation, but not sure how to explain it better.

memsharded (approved):

czoido (approved):

PR #11328, opened 2022-05-25
refactor run_in_windows_bash

### LINKED ISSUE #4680: [Conan 2.0] Study removal of `--login` by default at `run_in_windows_bash`
Valuable feedback here:
https://github.com/conan-io/conan/issues/4304

Study if remove by default the ``--login`` in Conan 2.0

memsharded:
Changelog: Omit
Docs: Omit

To prepare better for https://github.com/conan-io/conan/pull/11066

Close https://github.com/conan-io/conan/issues/4680

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -122,7 +122,7 @@ class CMake(object):
         if cmd_line_args:
             args += ['--'] + cmd_line_args
 
-        arg_list = [args_to_string([bf]), build_config, args_to_string(args)]
+        arg_list = ['"{}"'.format(bf), build_config, args_to_string(args)]
         arg_list = " ".join(filter(None, arg_list))
         command = "%s --build %s" % (self._cmake_program, arg_list)
         self._conanfile.output.info("CMake command: %s" % command)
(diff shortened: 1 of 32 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -122,7 +122,7 @@ def _build(self, build_type=None, target=None, cli_args=None, build_tool_args=No
         if cmd_line_args:
             args += ['--'] + cmd_line_args
 
-        arg_list = [args_to_string([bf]), build_config, args_to_string(args)]
```
lasote, line 125: Risk of 🍔 ?

conan/tools/env/environment.py
```diff
@@ -13,34 +13,35 @@ class _EnvVarPlaceHolder:
     pass
 
 
-def environment_wrap_command(env_filenames, cmd, subsystem=None, cwd=None):
-    assert env_filenames
+def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None, accept=None):
```
lasote:
```suggestion
def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None, accepted_extensions=None):
```

conans/client/subsystems.py
```diff
@@ -89,6 +90,14 @@ def escape_windows_cmd(command):
 
 
 def deduce_subsystem(conanfile, scope):
+    """ used by:
+    - EnvVars: to decide is using :  ; as path separator, translate paths to subsystem
```
lasote:
```suggestion
    - EnvVars: to decide if using :  ; as path separator, translate paths to subsystem
```

lasote (approved):

conan/tools/cmake/cmake.py
memsharded: Actually, that ``args_to_string`` calls ``subprocess.list2cmdline`` which is an undocumented, apparently internal ``subprocess`` function. I don't think we should be calling that at all. I prefer to risk the 🍔 than depending on that.

Anyone can provide some evidence about ``subprocess.list2cmdline`` being ok to be used?

memsharded: https://github.com/conan-io/conan/issues/11358

PR #11330, opened 2022-05-25
Feature/issue 11150

### LINKED ISSUE #11150: [feature] Remove or change `copy_symlink_folders` in `conan.tools.files.copy`
Hi,

when updating our Conan recipes to be prepared for Conan 2.X me and my collegues were struggling with the implementation of the new copy method (`conan.tools.files.copy`).

Our build infrastructure might be a little bit special. Anyway, we often have the situation that the toolchain creates symlink folders and files in the `build folder`. When using the new `copy` method to copy files from the `build folder` to the `package folder` we were really surprised at the beginning about some strange folder structure in the `package folder`.
It took us some time to analyze what's going on. At the end we realized that it is the parameter `copy_symlink_folders` in the `copy` method that has caused the issue.

I'm pretty sure there was a very good reason to introduce this parameter, however I would like to give you some experience from practical usage:

1.) From my point of view, the `copy` method is intended to copy files instead of folders. Therefore I think it might be a little bit unusual to combine both functionalities.

2.) When copying files, e.g. via `copy(self, "*.h", self.build_folder, self.package_folder)`, I don't see any reason why other files shall be copied to the package folder too. It would maybe different if there would be a `.h` file in a symlink folder, but the implementation of `copy_symlink_folders` just copies everything, no matter if a `.h` file is in a symlink folder or not.

3.) The same behavior applies to the copying of subfolder files, e.g. via `copy(self, "inc/*.h", self.build_folder, self.package_folder)`. To our surprise it will copy symlink folders outside of the `inc` subfolder too (as the `build_folder` is used as base) and will entirely ignore the pattern - this fact is somehow indicated in the documentation of the method, however I can tell you that it is still a big surprise for developers ;).

4.) Although the parameter `copy_symlink_folders` does not make sense in our situation, even worser (for us) is that the default is set to `True`. It should be set to `False` in my opinion.

5.) Even better I would like to ask if it makes (more) sense to introduce an operation like `copy_folder` which may handle symlink folders too. This may also be used to solve other issues (like #7789 which lacks about empty folders in the `package folder`). A clear separation between files and folders would make it easier, much clearer and less surprising - at least for us.

What do you think?

Below you'll find a quick and dirty Conan recipe to demonstrate some of the surprises (of course, setting the `copy_symlink_folders` explicitly to `False` would solve the issues):

```
import os
from conan import ConanFile
from pathlib import Path

class TestConan(ConanFile):
    name = "test"
    version = "1.0.0"

    def build(self):
        test = Path(self.build_folder, "test.h")
        test.touch()
        inc_folder = Path(self.build_folder, "inc")
        inc_folder.mkdir(parents=True, exist_ok=True)
        test2 = inc_folder.joinpath("test2.h")
        test2.touch()
        gen_folder = Path(self.build_folder, "gen")
        gen_folder.mkdir(parents=True, exist_ok=True)
        bin = gen_folder.joinpath("test.bin")
        bin.touch()
        sym_folder = Path(self.build_folder, "sym")
        sym_folder.symlink_to(gen_folder, target_is_directory=True)

    def package(self):
        if os.getenv("SURPRISE_1"):
            # Surprise no. 1
            # Copies the symlink folder although no *.h file is in there
            copy(self, "*.h", self.build_folder, self.package_folder)

        if os.getenv("SURPRISE_2"):
            # Surprise no. 2
            # Copies symlink folder althoug pattern searches in subfolder
            copy(self, "inc/*.h", self.build_folder, self.package_folder)
```

lasote:
Changelog: Feature: Remove the parameter ``copy_symlink_folders`` of the ``conan.tool.files.copy``  function and now, any symlink file pointing to a folder will be treated as a regular file.
Docs: https://github.com/conan-io/docs/pull/2558

Now, any symlink file pointing to a folder will be treated as a regular file, so they will only be copied if the specified ``pattern`` matches with the file. At the destination folder, it will be created pointing to the exact same folder (absolute or relative) being the user's responsibility (if necessary) to make the symlink a relative path before copying it.

Close #11150


memsharded (approved):
Looking good. The test is broken in 2.7: ``from pathlib import Path``

Issue #11186, opened 2022-05-06
[feature][develop2] Migration strategy for generated files


memsharded:
- settings.yml
- compatibility.py & compatibility files
- other possible extensions

I'd suggest something trivial, like adding a comment line at the beginning of the file ``#This is a Conan generated file``, and using that mark as a green light to replace the content when newer Conan versions are released that would update such files.

memsharded:
Closed by https://github.com/conan-io/conan/pull/11333

PR #11333, opened 2022-05-26
[dev2] Proposing migrations mechanism

franramirez688:
Changelog: Feature: Migrations mechanism based on Conan comment `This file was generated by Conan`.
Closes: https://github.com/conan-io/conan/issues/11186


conans/client/migrations.py
```diff
@@ -1,8 +1,42 @@
...
+
+        if not self.can_be_overwritten(settings_path):
+            out.warning(f"Migration: {settings_path} does not contain the Conan"
+                        f" comment: '{CONAN_GENERATED_COMMENT}'. Ignoring it")
+            return
+        else:
+            out.warning("Migration: Updating settings.yml")
+            new_settings = get_default_settings_yml()
+            if new_settings == load(settings_path):
+                out.info("Migration: settings.yml is already up to date")
```
memsharded: Can omit this message

conans/client/migrations.py
```diff
@@ -1,8 +1,42 @@
...
+            out.warning("Migration: This conan installation doesn't have settings yet")
+            out.warning("Nothing to migrate here, settings will be generated automatically")
+            return
+
+        if not self.can_be_overwritten(settings_path):
+            out.warning(f"Migration: {settings_path} does not contain the Conan"
+                        f" comment: '{CONAN_GENERATED_COMMENT}'. Ignoring it")
+            return
+        else:
+            out.warning("Migration: Updating settings.yml")
```
memsharded: Omit this message

conans/migrations.py
```diff
@@ -39,3 +43,14 @@ def _load_old_version(self):
         except Exception:
             old_version = None
         return old_version
+
+    def can_be_overwritten(self, file_path):
+        with open(file_path) as f:
+            for line in f.readlines():
```
memsharded: Force it to be the first line.

conans/migrations.py
```diff
@@ -39,3 +43,14 @@ def _load_old_version(self):
...
+                if self.file_comment_pattern.match(line):
```
memsharded: Do a ``if mystring in line``, keep it simple

franramirez688: Ok, I only wanted it to be safer. Changing it.

conans/client/graph/compatibility.py
```diff
@@ -88,17 +92,38 @@ def app_compat(conanfile):
 """
 
 
+def get_default_compatibility():
```
memsharded: Remove this, it is pure boilerplate without a purpose.

conans/client/migrations.py
```diff
@@ -1,8 +1,69 @@
...
+        self._update_file(settings_path, get_default_settings_yml())
+
+    def _update_binary_compatibilty_files(self, cache):
+        """
+        Update compatibility.py, app_compat.py, and cppstd_compat.py.
+        """
+        from conans.client.graph.compatibility import BinaryCompatibility, get_default_cppstd_compat,\
+            get_default_app_compat, get_default_compatibility
+
+        compatibility_file, app_compat_file, cppstd_compat_file = \
```
memsharded: Too specific knowledge here, I think I would put these migration lines with the internal details inside ``BinaryCompatibility``

conans/migrations.py
```diff
@@ -39,3 +41,26 @@ def _load_old_version(self):
...
+
+        :param file_path: ``str`` file path
+        :return: ``bool`` ``True``  whether the file can be overwritten, otherwise, ``False``
+        """
+        with open(file_path) as f:
+            # Forcing it to have the comment in the first line
+            if CONAN_GENERATED_COMMENT in f.readline():
+                return True
+            else:
+                return False
```
memsharded:
```suggestion
        # The first line of the file should contain the comment
        return CONAN_GENERATED_COMMENT in open(file_path).readline()
```

conans/migrations.py
```diff
@@ -39,3 +41,26 @@ def _load_old_version(self):
         except Exception:
             old_version = None
         return old_version
+
+    @staticmethod
+    def can_be_overwritten(file_path):
```
memsharded: Do not make it an independent method, embed it in the ClientMigrator ``_update_file``

conans/client/migrations.py
```diff
@@ -1,8 +1,69 @@
...
+        if not self.can_be_overwritten(file_path):
+            out.warning(f"Migration: {file_path} does not contain the Conan"
+                        f" comment: '{CONAN_GENERATED_COMMENT}'. Ignoring it")
+            return
+        else:
+            if new_content == load(file_path):
+                return
+            else:
+                save(file_path, new_content)
+                out.success(f"Migration: Successfully updated {file_name}")
```
memsharded:
```suggestion
        content = load(file_path)
        if  CONAN_GENERATED_COMMENT not in content.split("\n", 1)[0]:
            out.warning(f"Migration: {file_path} does not contain the Conan"
                        f" comment: '{CONAN_GENERATED_COMMENT}'. Ignoring it")
        elif content != new_content:
            save(file_path, new_content)
            out.success(f"Migration: Successfully updated {file_name}")
```

Then this method could be static or even a free function that can be applied anywhere (like directly inside ``BinaryCompatibility``)

conans/client/graph/compatibility.py
franramirez688: Then, I'll use only functions instead.

conans/migrations.py
franramirez688: 🤦

conans/client/migrations.py
franramirez688: I don't think so. I think it's the same knowledge as having another function migrating the `settings.yml` or whatever. I don't really like the idea of having some classes having their own `migrate()` function and others without it.

memsharded (approved):

conans/client/conf/__init__.py
```diff
@@ -119,3 +120,8 @@
 
 def get_default_settings_yml():
     return default_settings_yml
+
+
+def migrate_settings_file(cache, updater):
+    settings_path = cache.settings_path
+    updater(settings_path, get_default_settings_yml())
```
memsharded: Where is the ``updater()`` function?
I think the rename broke it, and it is not covered by tests?

In that case, it would be necessary to add some tests, use as inspirations the ones in conan 1.X, how the version.txt is forced to trigger an actual migration.

memsharded (changes requested):
I missed that it was broken, but not covered by tests. Please add them, and probably we will request some extra review after that. Thanks!

conans/client/conf/__init__.py
franramirez688: Why is it broken? The `updater()` function is passed as parameter. I wanted all the migration logic to be at the same place so the `updater` in this case is the `update_file()` one that is coming from `client.migrations`.

Having a look at those tests.

memsharded (approved):

PR #11334, opened 2022-05-26
Select components for XcodeDeps direct dependencies

czoido:
Changelog: Feature: Select components to use for `XcodeDeps` direct dependencies setting `self.cpp.package.requires` inside the `layout()` method.
Docs: TODO

This is useful in the case an application depends on a package with components and you just want to add an specific component, so for example, you require boost but you just want boost::random then:

```python
class MyApplicationConan(ConanFile):
    name = "myapplication"
    version = "1.0"
    requires = "boost/1.79.0"
    generators = "XcodeToolchain"

    exports_sources = "app.xcodeproj/*", "app/*"

    def layout(self):
         self.cpp.package.requires = ["boost::random"]

    def generate(self):
        deps = XcodeDeps(self)
        deps.generate()

    def build(self):
        xcode = XcodeBuild(self)
        xcode.build("app.xcodeproj", target="app")
```

This will **raise** if you require a component that is **not valid** or that is **not from a direct dependency**.


conan/tools/apple/xcodedeps.py
```diff
@@ -129,8 +129,24 @@ def __init__(self, conanfile):
...
 
+    @property
+    def use_components(self):
+        return self._use_components
+
+    @use_components.setter
+    def use_components(self, components):
+        # we can only specify to use components from direct dependencies
+        valid_component_pkg_names = [dep.ref.name for dep in self._conanfile.dependencies.direct_host.values()]
+        for comp in components:
```
memsharded, line 143: I feel this check doesn't belong here, but elsewhere.

memsharded (commented):
This looks like this would be a general feature, for all build systems, and there is still some concerning overlap with ``cpp_info.requires``, that I agree it is not good for the wanted behavior, but I don't know a better solution.

I think we should discuss and think about this a bit deeper.

memsharded:
The idea is to use

```python
def layout(self):
     self.cpp.package.requires = ["pkg::component"]
```

Because that is also useful for consumers, and it only has to be declared once, and it is generic for all generators.

conan/tools/apple/xcodedeps.py
```diff
@@ -239,6 +255,11 @@ def get_content_for_component(self, pkg_name, component_name, cpp_info, reqs):
         result[file_dep_name] = dep_content
         return result
 
+    def _add_component(self, pkg_name, component_name):
```
franramirez688, line 258: I think the function name sounds weird because we're not adding any component at all... Could we rename it?

conan/tools/apple/xcodedeps.py
franramirez688: I'd simplify it. IMO, I think it could help:

```python

        # ....
        self.use_components = []  # removing the property and the setter

    # .....

    def _validate_use_components(self, deps):
        """Validating the `self.use_components` property"""
        if not self.use_components:
             return
        # we can only specify to use components from direct dependencies
        valid_component_pkg_names = [dep.ref.name for dep in deps]
        for comp in self.use_components:
            if comp.split("::")[0] not in valid_component_pkg_names:
                raise ConanException(
                    "Component '{}' is not a direct dependency for this package. "
                    "You can only specify components for direct dependencies "
                    "with XcodeDeps.use_components".format(comp))

    def _content(self):
        result = {}

        host_reqs = self._conanfile.dependencies.host.values()

        # At first, let's validate the use_components if exist
        self._validate_use_components(host_reqs)

        # Generate the config files for each component with name conan_pkgname_compname.xcconfig
        # If a package has no components the name is conan_pkgname_pkgname.xcconfig
        # Then all components are included in the conan_pkgname.xcconfig file
        for dep in host_reqs:
            # .....
```

WDYT?

czoido:
Finally, we will wait a bit to implement this feature when we have a clearer idea on how to implement this. A couple alternatives:

 ```
   # option 1
    def requirements(self):
        self.requires("boost/1.79.0", components=["boost::random"])

    # option 2
    def layout(self):
        self.cpp.package.requires = ["boost::random"]

    # option 3
    def generate(self):
        deps = XcodeDeps(self)
        deps.use_components = ["boost::random", "poco"]
        deps.generate()

    def build(self):
        xcode = XcodeBuild(self)
        xcode.build("app.xcodeproj", target="app")
```

There's an easy workaround if you just want to consume certain components that is modifying the generated xcconfig files after the generate() call.

PR #11343, opened 2022-05-27
Bazeldeps: get lib file path by basename

### LINKED ISSUE #11331: [bug] bazeldeps skip lib incorrectly if filename both starts and ends with lib
  * Compiler+version:
  * Conan version: 1.48.1
  * Python version: 3.10.1

If a lib filename both starts and ends with lib, when use conan install with bazel generator, the lib will be incorrectly skipped.
For example, if I have a lib named `libxxx.lib`, the generated bazel BUILD file won't contains this lib.

The bug if caused in `_get_lib_file_paths` in `conan/tools/google/bazeldeps.py`:

```python
    def _get_lib_file_paths(self, libdirs, lib):
        for libdir in libdirs:
            if not os.path.exists(libdir):
                self._conanfile.output.warning("The library folder doesn't exist: {}".format(libdir))
                continue
            files = os.listdir(libdir)
            for f in files:
                full_path = os.path.join(libdir, f)
                if not os.path.isfile(full_path):  # Make sure that directories are excluded
                    continue
                name, ext = os.path.splitext(f)
                if ext in (".so", ".lib", ".a", ".dylib", ".bc"):
                    if ext != ".lib" and name.startswith("lib"):
                        name = name[3:]
                if lib == name:
                    return full_path
        self._conanfile.output.warning("The library {} cannot be found in the "
                                       "dependency".format(lib))
```

In our case, the lib is named `libacsdk.lib`, the `name` is `acsdk`, so the `if lib == name` will never be true.

One possible fix is test if the `lib` is equal to `f`(filename) before normalize the lib name. Like this:

```python
    def _get_lib_file_paths(self, libdirs, lib):
        for libdir in libdirs:
            if not os.path.exists(libdir):
                self._conanfile.output.warning("The library folder doesn't exist: {}".format(libdir))
                continue
            files = os.listdir(libdir)
            for f in files:
                full_path = os.path.join(libdir, f)
                if not os.path.isfile(full_path):  # Make sure that directories are excluded
                    continue
               # if filename is equal to lib, I think we should return it.
               if lib == f:
                    return full_path
                name, ext = os.path.splitext(f)
                if ext in (".so", ".lib", ".a", ".dylib", ".bc"):
                    if ext != ".lib" and name.startswith("lib"):
                        name = name[3:]
                if lib == name:
                    return full_path
        self._conanfile.output.warning("The library {} cannot be found in the "
                                       "dependency".format(lib))
```

If this fix is OK I can create a pull request.

kkpattern:
Changelog: Fix: BazelDeps generator cannot find a lib when it's named with the basename of the lib file.
Docs: omit

Users may not name their libraries in a conventional way.
Sometimes directly use the basename of the lib file as lib name.
We need to be able to find the lib in this case.

Close: #11331 

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -168,6 +168,10 @@ class BazelDeps(object):
                 full_path = os.path.join(libdir, f)
                 if not os.path.isfile(full_path):  # Make sure that directories are excluded
                     continue
+                # Users may not name their libraries in a conventional way. For example, directly
+                # use the basename of the lib file as lib name.
+                if f == lib:
+                    return full_path
                 name, ext = os.path.splitext(f)
                 if ext in (".so", ".lib", ".a", ".dylib", ".bc"):
                     if ext != ".lib" and name.startswith("lib"):
--- a/conans/test/unittests/tools/google/test_bazeldeps.py
+++ b/conans/test/unittests/tools/google/test_bazeldeps.py
@@ -50,6 +50,41 @@ def test_bazeldeps_dependency_buildfiles():
             assert 'deps = [\n    \n    ":lib1_precompiled",' in dependency_content
 
 
+def test_bazeldeps_get_lib_file_path_by_basename():
+    conanfile = ConanFile(Mock(), None)
+
+    cpp_info = CppInfo("mypkg", "dummy_root_folder1")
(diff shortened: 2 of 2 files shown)


memsharded (approved):
Thanks very much for contributing this!

PR #11350, opened 2022-05-27
raise pyjwt version

### LINKED ISSUE #11324: [bug] [security] conan depends on vulnerable pyjwt (Key confusion through non-blocklisted public key formats)
See https://github.com/advisories/GHSA-ffqj-6fqr-9h24. This recommends update to pyjwt>=2.4.0, but conan (from PyPI) depends on pyjwt<2.0.0.

Recommended read: https://iscinumpy.dev/post/bound-version-constraints/.

memsharded:
Changelog: Bugfix: Force ``conan_server`` to use ``pyjwt>=2.4.0`` to solve a known vulnerability. 
Docs: Omit

Close https://github.com/conan-io/conan/issues/11324


lasote (approved):

memsharded:
@lasote I have had to do some more changes to pass the tests in py27 and "fake" the ``jwt`` for testing, please re-review

conans/server/conanjwt.py
```diff
@@ -0,0 +1,18 @@
+from collections import OrderedDict
+
+try:
+    from jwt import encode, decode, DecodeError, ExpiredSignatureError
+except ImportError:  # Just for testing purposes
```
lasote: Time to drop python 2, latest supported 1.48. This is messy.

lasote (changes requested):

conans/server/conanjwt.py
memsharded: Agree, done

lasote (approved):
Change the changelog, probably worth it to mention this is only python 3.
About that, Should we introduce this in the `setup.py`? we don't have it in the develop2 either, but probably we should?

(https://packaging.python.org/en/latest/guides/dropping-older-python-versions/#specify-the-version-ranges-for-supported-python-distributions)

I'm proposing these changes in a different PR so I'm merging this. But change the changelog!!!!

PR #11355, opened 2022-05-28
Fix: interface_library should be used with shared_library on Windows

### LINKED ISSUE #11347: [bug] Bazeldeps: build file is not correct for shared libary on Windows platform
  * Compiler+version:
  * Conan version: 1.48.1
  * Python version: 3.10.1

For shared libraries on Windows platform, the correct way to import them is using `interface_library` and `shared_library` together, like this:

```python
cc_import(
  name = "mylib",
  hdrs = ["mylib.h"],
  # mylib.lib is a import library for mylib.dll which will be passed to linker
  interface_library = "mylib.lib",
  # mylib.dll will be available for runtime
  shared_library = "mylib.dll",
)
```

Here is the [doc](https://docs.bazel.build/versions/main/be/c-cpp.html#cc_import).

Currently, bazel generator will generate BUILD file with shared libraries on Windows in this way:

```python
cc_import(
    name = "XXXprecompiled",
    shared_library = "lib/XXX.lib"
)
```

According to the conan [doc](https://docs.conan.io/en/latest/howtos/manage_shared_libraries/env_vars.html#building-an-application-using-the-shared-library-from-tool-a), DLL files are put in the bin directory. So we need to get corresponding DLL paths from bin directory.

Another problem we have is that DLL file of a library may depends on another DLL file. The publisher may put them all in the bin dir. But those DLL file won't be mention in the `libs` so bazel generator will skip those DLL libs.

We tried a super hacky way to solve this problem: We iterator the bin dir, if DLL file is encountered we list it as a shared library. If the corresponding interface lib is in `libs`, we will generate `cc_import` with both `interface_library` and `shared_library`. If only DLL file is present, we generate `cc_import` with `shared_library` only. Like this:

```python
cc_import(
    name = "XXXprecompiled",
    interface_library = "lib/XXX.lib",
    shared_library = "bin/XXX.dll"
)


cc_import(
    name = "YYYprecompiled",
    shared_library = "bin/YYY.dll"
)
```

This also solves another problem where the publisher may forget to the `shared` attribute to `True`. Bazel generator will assume the lib is static.

This solution actually works well internally in our project. But I feel it's too hacky for a general solution. If anyone is interested in our solution, I can create a branch and link it here.

kkpattern:
Changelog: Fix: Use `interface_library` with `shared_library` on Windows in BazelDeps.
Docs: omit
Close: #11347 

When importing shared libraries, check if there is a corresponding DLL file in the bin directory. If the DLL file exists, we should use `cc_import` of the form:

```python
cc_import(
    name = "libname",
    interface_library = "lib/libname.lib",
    shared_library = "bin/libname.dll",
)
```

Bazel doc [here](https://bazel.build/reference/be/c-cpp#cc_import).

I also added a new `MockConanFileDeps` to avoid the infinite loop in unittest(No need to patch the `conans.ConanFile.dependecies` for all).

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -66,6 +66,14 @@ class BazelDeps(object):
             )
             {% endfor %}
 
+            {% for libname, (lib_path, dll_path) in shared_with_interface_libs.items() %}
+            cc_import(
+                name = "{{ libname }}_precompiled",
+                interface_library = "{{ lib_path }}",
+                shared_library = "{{ dll_path }}",
+            )
+            {% endfor %}
+
             cc_library(
                 name = "{{ name }}",
                 {% if headers %}
@@ -81,9 +89,12 @@ class BazelDeps(object):
                 linkopts = [{{ linkopts }}],
                 {% endif %}
                 visibility = ["//visibility:public"],
-                {% if libs %}
+                {% if libs or shared_with_interface_libs %}
                 deps = [
                 {% for lib in libs %}
+                ":{{ lib }}_precompiled",
+                {% endfor %}
+                {% for lib in shared_with_interface_libs %}
                 ":{{ lib }}_precompiled",
                 {% endfor %}
                 {% for dep in dependencies %}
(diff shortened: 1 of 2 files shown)


memsharded (approved):
This is looking nice and clean (despite the problem is intrinsically a bit complex, given the context information), so thanks very much for contributing it.

lasote (approved):

PR #11365, opened 2022-05-31
Add fix_apple_shared_install_name tool

### LINKED ISSUE #10253: [feature] Provide macOs packages to build with @rpath in install names
The rather old answer describes the problem quite well: https://github.com/conan-io/conan/issues/1238#issuecomment-313683525

On macOS it would help a lot if all packages in ConanCenter have an option (if not defaulted…) to have install names prepended with @rpath. This is already available for conan builds but probably in a lot of packages not activated etc? This would align it better to Linux and would make debugging much easier. As executables can use the RPATH facility to load the libs from Conan. Lots of packages come with naked install names posing quite some headache and not so good behavior (see answer above).

czoido:
Changelog: Feature: Do not automatically fix the shared libraries to add the rpath in Apple and add an external tool `tools.apple.fix_apple_shared_install_name` to do it optionally in recipes for packages that do not set the correct `LC_ID_DYLIB`.
Docs: https://github.com/conan-io/docs/pull/2562

Closes: https://github.com/conan-io/conan/issues/10253


conans/assets/templates/new_v2_autotools.py
```diff
@@ -48,6 +49,8 @@ def build(self):
         def package(self):
             autotools = Autotools(self)
             autotools.install()
+            if self.settings.get_safe("os") == "Macos" and self.options.get_safe("shared", False):
```
memsharded: I'd say that we want to put these conditionals inside the tool itself. This way, it keeps working for example in 2.0 when we use ``package_type`` to check for "shared", and any evolution like being enabled for iOS & family.

conan/tools/files/files.py
```diff
@@ -520,3 +521,22 @@ def swap_child_folder(parent_folder, child_folder):
...
+    def _osx_collect_dylibs(lib_folder):
+        return [f for f in os.listdir(lib_folder) if f.endswith(".dylib")
+                and not os.path.islink(os.path.join(lib_folder, f))]
+
+    def _fix_install_name(lib_name, lib_folder):
+        command = "install_name_tool -id @rpath/{} {}".format(lib_name, os.path.join(lib_folder,
+                                                                                     lib_name))
+        conanfile.run(command)
+
+    if conanfile.settings.get_safe("os") == "Macos" and conanfile.options.get_safe("shared", False):
```
franramirez688: Shouldn't it be for any Apple OS?

```suggestion
    if is_apple_os(conanfile.settings.get_safe("os")) and conanfile.options.get_safe("shared", False):
```

memsharded (approved):

franramirez688 (approved):

conan/tools/files/files.py
```diff
@@ -520,3 +522,22 @@ def swap_child_folder(parent_folder, child_folder):
     child = os.path.join(parent_folder, child_folder)
     for f in os.listdir(child):
         shutil.move(os.path.join(child, f), os.path.join(parent_folder, f))
+
+
+def fix_apple_shared_install_name(conanfile):
+    def _osx_collect_dylibs(lib_folder):
+        return [f for f in os.listdir(lib_folder) if f.endswith(".dylib")
```
jcar87: I can't remember 100%, but I believe on apple platforms frameworks are also affected. For example, a framework called `Foo.framework`, may have the actual "shared library" file inside the bundle like this:

`Foo.framework/Versions/A/Foo`, and I suspect that file has an install name (`LC_ID_DYLIB`) that may need patching. I could be wrong, but this is worth checking.

conan/tools/files/files.py
```diff
@@ -520,3 +522,22 @@ def swap_child_folder(parent_folder, child_folder):
...
+                and not os.path.islink(os.path.join(lib_folder, f))]
+
+    def _fix_install_name(lib_name, lib_folder):
+        command = "install_name_tool -id @rpath/{} {}".format(lib_name, os.path.join(lib_folder,
+                                                                                     lib_name))
+        conanfile.run(command)
+
+    if is_apple_os(conanfile.settings.get_safe("os")) and conanfile.options.get_safe("shared", False):
```
jcar87: I think there could be situations (perhaps in private recipes) where the `shared` option is not defined (maybe the library can "only" work as a shared library and the maintainers didn't bother to add shared True/False toggle - can be the case with libraries that define global symbols were it can be dangerous if they end up in multiple shared libraries loaded by the same process..) 

I would probably err on the side of caution and not require that the `shared` option exists or is set to true? / Or perhaps, change the default in `get_safe` to true: assume that if it is not defined but the user is calling it, that they want this to be executed?

conan/tools/files/files.py
```diff
@@ -520,3 +522,22 @@ def swap_child_folder(parent_folder, child_folder):
...
         shutil.move(os.path.join(child, f), os.path.join(parent_folder, f))
+
+
+def fix_apple_shared_install_name(conanfile):
+    def _osx_collect_dylibs(lib_folder):
+        return [f for f in os.listdir(lib_folder) if f.endswith(".dylib")
+                and not os.path.islink(os.path.join(lib_folder, f))]
+
+    def _fix_install_name(lib_name, lib_folder):
+        command = "install_name_tool -id @rpath/{} {}".format(lib_name, os.path.join(lib_folder,
```
jcar87: This is good!!

Two comments - one is about the actual install name:
We may have these 3 files:
```
libopencv_core.3.4.17.dylib
libopencv_core.3.4.dylib -> libopencv_core.3.4.17.dylib
libopencv_core.dylib -> libopencv_core.3.4.dylib
```

we need  to run the command on the file that is _not_ a symlink, but the install name is actually `/libopencv_core.3.4.dylib` - so neither the leaf name without a version, nor the one with the full version :/ What I mean is that the `lib_name` here may not coincide with the filename of the file that is not a symlink

The second one is about packages that contain multiple shared libraries, that may depend on each other. If we change the install name of a library that is depended by others, those _others_ also need to be patched to refer to the new install name. For example with opencv, for simplicity:

* `libopencv_core.dylib`
* `libopencv_imgproc.dylib`

If we inspect the properties of the second one:
```
 % otool -L libopencv_imgproc.dylib 
libopencv_imgproc.dylib:
	/path/without/patching/libopencv_imgproc.3.4.dylib (compatibility version 3.4.0, current version 3.4.17)
	/path/without/patching/libopencv_core.3.4.dylib (compatibility version 3.4.0, current version 3.4.17)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```

calling `install_name_tool -id @rpath/libopencv_core.dylib libopencv_core.dylib` will change `opencv_core` itself, but and we can also fix the install name of `opencv_imgproc`, but we still need `opencv_imgproc` to depend on `@rpath/libopencv_core.dylib` as well, which can be achieved by:

```
install_name_tool -change /path/without/patching/libopencv_core.3.4.dylib @rpath/libopencv_core.3.4.dylib libopencv_imgproc.dylib
```

Otherwise we may end up with dependencies that will break at link time for consumers

jcar87 (commented):
I like that this is a separate tool!!

Added some comments with a few corner cases !

conan/tools/files/files.py
memsharded: This might be partially addressed in Conan 2.0 ``package_type`` which is the way to specify something is a shared library, even if no ``shared`` option is defined (because the package is always shared)

conan/tools/files/files.py
czoido: Thanks! Those are interesting cases that we should have in mind. I'll try to add some tests. 
Changing the install names of the dependencies with `install_name_tool -change` is something we may achieve. For the symlinks and names problem maybe it's more complicated and if some library has that problem then it's something that should be fixed in that library and not via our tool... WDYT?

conan/tools/files/files.py
czoido: Thanks! I'll check that

conan/tools/files/files.py
jcar87: I think we can ignore the symlinks (as long as we only patch "one" file :P, so the current logic of patching the file that is not a symlink is fine.

Perhaps we could have a "logic" that "replaces" the path of the install name without affecting the leaf name?
e.g. if  `LC_ID_DYLIB` is `/path/to/library.1.2.dylib` we extract that and replace the path leading to the leaf name. I think `os.path.dirname()` can give us the bit that needs replacing -  The tricky bit is extracting the `LD_ID_DYLIB` which may require parsing the output of `otool -l` :(

czoido: Yes, I think that's we can do 👍

conan/tools/files/files.py
```diff
@@ -520,3 +522,42 @@ def swap_child_folder(parent_folder, child_folder):
     child = os.path.join(parent_folder, child_folder)
     for f in os.listdir(child):
         shutil.move(os.path.join(child, f), os.path.join(parent_folder, f))
+
+
+def fix_apple_shared_install_name(conanfile):
+
+    def _get_install_name(path_to_dylib):
+        command = "otool -D {}".format(path_to_dylib)
```
jcar87: !!! `otool -D`, didn't know about this one, I love it!

conans/test/functional/toolchains/gnu/test_v2_autotools_template.py
```diff
@@ -167,3 +167,142 @@ def build(self):
     client2.run_command("build-release/greet")
     assert "Hello World Release!" in client2.out
 
+
+@pytest.mark.skipif(platform.system() not in ["Darwin"], reason="Only affects apple platforms")
+@pytest.mark.tool_autotools()
+def test_autotools_fix_shared_libs():
```
jcar87, line 173: This looks good!!! It's testing the right things :D 

it makes me wonder whether we want to couple the autotools toolchain _with_ the  `fix_apple_shared_install_name` function? We are now testing both things together. In the future there could be case where a new change breaks some behaviour of the autotools toolchain, but the `fix_apple_shared_install_name` function still works fine?

I think we could simplify the generation of the files maybe by doing it directly, e.g. something along the lines of:
```
echo "extern int hello;" | clang++ -x c++ -dynamiclib -o libhello.dylib  -current_version 1.2.3 -compatibility_version 1.2  -install_name /path/to/libhello.dylib - 

echo "extern int goodbye;" | clang++ -x c++ -dynamiclib -o libgoodbye.3.5.dylib -install_name /path/to/libgoodbye.3.dylib -L $(pwd) -lhello
```
In case this helps simplify the logic / reduce lines of code, while still fully testing the function :D

jcar87 (approved):
LGTM! only a small comment but we can address in the future without impacting functionality

PR #11367, opened 2022-05-31
System package manager install set of equivalent packages

czoido:
Changelog: Feature: Add `install_substitutes` to system package manager tools to be able to install sets of packages that are equivalent with different names for different distros.
Docs: https://github.com/conan-io/docs/pull/2563

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -61,6 +61,9 @@ class _SystemPackageManagerTool(object):
         if self._active_tool == self.__class__.tool_name:
             return method(*args, **kwargs)
 
+    def install_substitutes(self, *args, **kwargs):
+        return self.run(self._install_substitutes, *args, **kwargs)
+
     def install(self, *args, **kwargs):
         return self.run(self._install, *args, **kwargs)
 
@@ -69,6 +72,20 @@ class _SystemPackageManagerTool(object):
 
     def check(self, *args, **kwargs):
         return self.run(self._check, *args, **kwargs)
+
+    def _install_substitutes(self, *packages_substitutes, update=False, check=True, **kwargs):
+        errors = []
+        for packages in packages_substitutes:
+            try:
+                self.install(packages, update, check, **kwargs)
+                break
+            except ConanException as e:
+                errors.append(e)
+
+        if errors:
+            for error in errors:
+                self._conanfile.output.warn(str(error))
+            raise ConanException("None of the installs for the package substitutes succeeded.")
 
     def _install(self, packages, update=False, check=True, **kwargs):
         if update:
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -77,8 +77,7 @@ class _SystemPackageManagerTool(object):
         errors = []
         for packages in packages_substitutes:
             try:
-                self.install(packages, update, check, **kwargs)
-                break
+                return self.install(packages, update, check, **kwargs)
             except ConanException as e:
                 errors.append(e)
 
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -81,10 +81,9 @@ class _SystemPackageManagerTool(object):
             except ConanException as e:
                 errors.append(e)
 
-        if errors:
-            for error in errors:
-                self._conanfile.output.warn(str(error))
-            raise ConanException("None of the installs for the package substitutes succeeded.")
+        for error in errors:
+            self._conanfile.output.warn(str(error))
+        raise ConanException("None of the installs for the package substitutes succeeded.")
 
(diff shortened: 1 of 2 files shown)


memsharded (approved):

memsharded (approved):

conans/test/functional/tools/system/package_manager_test.py
```diff
@@ -30,11 +29,51 @@ def system_requirements(self):
...
+                # FIXME this is needed because the ci-functional apt-get update fails
+                try:
+                    self.run("sudo apt-get update")
+                except Exception:
+                    pass
+                apt = Apt(self)
+                {}
+        """)
+
+    client.run("remove test/1.0 -f")
```
franramirez688: Why do we have to remove that package? Does it even exist?

conans/test/functional/tools/system/package_manager_test.py
```diff
@@ -30,11 +29,51 @@ def system_requirements(self):
...
+    assert "ERROR: while executing system_requirements(): " \
+           "None of the installs for the package substitutes succeeded." in client.out
+
+    client.run_command("sudo apt remove nano -yy")
+    installs = 'apt.install_substitutes(["non-existing1", "non-existing2"], ["nano"], ["non-existing3"])'
+    client.save({"conanfile.py": conanfile_py.format(installs)})
+    client.run("create . test/1.0@ -c tools.system.package_manager:mode=install "
+               "-c tools.system.package_manager:sudo=True")
+    assert "1 newly installed" in client.out
+    print(client.out)
```
franramirez688:
```suggestion
```

conans/test/functional/tools/system/package_manager_test.py
```diff
@@ -30,11 +29,51 @@ def system_requirements(self):
...
+               "-c tools.system.package_manager:sudo=True", assert_error=True)
+    assert "dpkg-query: no packages found matching non-existing1:amd64" in client.out
+    assert "dpkg-query: no packages found matching non-existing2:amd64" in client.out
+    assert "dpkg-query: no packages found matching non-existing3:amd64" in client.out
+    assert "dpkg-query: no packages found matching non-existing4:amd64" in client.out
+    assert "ERROR: while executing system_requirements(): " \
+           "None of the installs for the package substitutes succeeded." in client.out
+
+    client.run_command("sudo apt remove nano -yy")
+    installs = 'apt.install_substitutes(["non-existing1", "non-existing2"], ["nano"], ["non-existing3"])'
```
franramirez688, line 62: I think it could be better to have one-failing and one-succeeding in the same list:

```suggestion
    installs = 'apt.install_substitutes(["non-existing1", "non-existing2"], ["non-existing3", "nano"])'
```

conan/tools/system/package_manager.py
```diff
@@ -70,6 +73,18 @@ def update(self, *args, **kwargs):
...
+        errors = []
+        for packages in packages_substitutes:
+            try:
+                return self.install(packages, update, check, **kwargs)
+            except ConanException as e:
+                errors.append(e)
+
+        for error in errors:
+            self._conanfile.output.warn(str(error))
+        raise ConanException("None of the installs for the package substitutes succeeded.")
```
franramirez688, line 86: Would it be possible to simplify it?

```python
    def _install_substitutes(self, *packages_substitutes, update=False, check=True, **kwargs):
        for packages in packages_substitutes:
            try:
                return self.install(packages, update, check, **kwargs)
            except ConanException as e:
                self._conanfile.output.warn(str(e))

        raise ConanException("None of the installs for the package substitutes succeeded.")
```

franramirez688 (approved):

conans/test/functional/tools/system/package_manager_test.py
czoido: It's because the system_requirements method only works in the first export of the package, I'll add a note, it's a bit confusing

conans/test/functional/tools/system/package_manager_test.py
czoido: That would fail, you need at least one list of all valid packages

conan/tools/system/package_manager.py
czoido: The thing is that I did not want to output warning messages unless one of them is failing... if there's one valid combination I would prefer not to confuse users throwing warnings that do not apply to them...

conans/test/functional/tools/system/package_manager_test.py
czoido: And yes, the remove was in the wrong place, thank you, I'll fix that 👍
In fact is not necessary. at all as the first export does not succeed, I'll remove it.

conan/tools/system/package_manager.py
franramirez688: Sure!

PR #11381, opened 2022-06-02
Port autotools changes from 1.49 to 1.48.2

czoido:
Changelog: Feature: Port https://github.com/conan-io/conan/pull/11365 and https://github.com/conan-io/conan/pull/11284 to 1.48.2
Docs: omit

⚠️ Important! Do not merge with squash

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -13,9 +13,11 @@ from fnmatch import fnmatch
 import six
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
+from conan.tools.apple.apple import is_apple_os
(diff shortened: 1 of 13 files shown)


conan/tools/files/files.py
```diff
@@ -13,9 +13,11 @@
 import six
 
 from conan.tools import CONAN_TOOLCHAIN_ARGS_FILE, CONAN_TOOLCHAIN_ARGS_SECTION
+from conan.tools.apple.apple import is_apple_os
```
memsharded: This import here doesn't apply?

memsharded (approved):

conan/tools/files/files.py
czoido: yes, sorry, that slipped through, I'll fix that and take it to develop after it

Issue #11293, opened 2022-05-18
[develop2][feature] Allow defining different ``--build`` for build and host, or for different configurations


memsharded:
Following up from https://github.com/conan-io/conan/pull/11244, the problem is:

```python
class Pkg(ConanFile):
            name = "tool"
            version = "1.0"
            settings = "os"
            def build_requirements(self):
                if cross_building(self):
                    self.tool_requires("tool/1.0")
```

- When doing a ``conan create .`` that implies ``--build=tool``, so both the "build" and "host" packages are always re-built from sources, when only the "host" should be rebuild in general if the "build" binary already exists
- The ``conan create . --build=missing`` can work, but it has the inconvenient that it could start building from source unexpected missing dependencies binaries, without noticing.

Some possibilities:
- ``--build:host`` and ``--build:build`` contexts
- ``--build=only_current_if_missing``. This could have other advantages, like ``conan create`` flows that do not unnecessarily re-build from sources from the exact same source code (same recipe revision).

memsharded:
Closed by https://github.com/conan-io/conan/pull/11424

PR #11382, opened 2022-06-02
Add new ``--build=missing:pattern`` and apply it to ``conan create``

memsharded:
Proof of concept for https://github.com/conan-io/conan/issues/11293

This approach has some potential downsides:

- ``conan create`` even for a package without dependencies will hit the servers, trying to find a precompiled binary for the current package we are commanding to create (to avoid re-creating something that was already created). This might not be super intuitive, for example if one remote is unavailable, it will error (even if the package has no dependencies)
- When the package binary depends on variable, non modeled environment, subsequents ``conan create`` will not rebuild new binaries unless ``conan create --build=*`` or something like that is commanded.

To discuss the tradeoffs.


memsharded:
Update after review:

- The default keeps the old one ``conan create`` always build from source, unless different ``--build`` specified
- Changed syntax to ``--build missing:pattern`` (Note, @lasote for some reason it doesn't need ``--build=missing:pkg*``, but a ``--build=missing:pkg`` also seems to work
- Added new test to cover the behavior.

lasote (approved):

lasote:
> Changed syntax to --build missing:pattern (Note, @lasote for some reason it doesn't need --build=missing:pkg*, but a --build=missing:pkg also seems to work

🤔 will check

lasote (changes requested):
I don't know what is going on but something is not working at all.
This code is never called for the "pkg" reference:

```
    def should_build_missing(self, conanfile):
        for pattern in self.build_missing_patterns:
            if ref_matches(conanfile.ref, pattern, is_consumer=False):
                return True
```
So it really doesn' t matter whether you use the "pkg*" or "pkg".

memsharded:
Test updated in https://github.com/conan-io/conan/pull/11382/commits/052e10c1742264cebd9fd59169f32200b9b0e432, please @lasote re-check, I hope now it is more clear the behavior.

franramirez688 (approved):

lasote (approved):
👍🏼

czoido (approved):

Issue #11228, opened 2022-05-11
[question][develop2] Review encodings auto detection with BOM


memsharded:
Something that I think is worth reviewing is the following:

```python
def _detect_encoding(text):
    import codecs
    encodings = {codecs.BOM_UTF8: "utf_8_sig",
                 codecs.BOM_UTF16_BE: "utf_16_be",
                 codecs.BOM_UTF16_LE: "utf_16_le",
                 codecs.BOM_UTF32_BE: "utf_32_be",
                 codecs.BOM_UTF32_LE: "utf_32_le",
                 b'\x2b\x2f\x76\x38': "utf_7",
                 b'\x2b\x2f\x76\x39': "utf_7",
                 b'\x2b\x2f\x76\x2b': "utf_7",
                 b'\x2b\x2f\x76\x2f': "utf_7",
                 b'\x2b\x2f\x76\x38\x2d': "utf_7"}
```

We are using the BOM in ``decode_text()`` to detect the encoding. It feels a bit too magic for me.

SSE4:
there is no magic, everything is straighforward. just read BOM to determine the text file encoding, then use it to properly decode the file.

memsharded:
I disagree that such a code have to be used in the ``decode_text()`` helper that is used mainly in the Conan HTTP Api, because the encoding shouldn't be detected from BOMs, but it should be already defined by the protocol (utf-8). Also when loading Conan files (conaninfo.txt, conanmanifest.txt, etc...) which would be the majority of cases, same, we should define the encoding of those files (utf-8) and respect it.

So for most usages, this shouldn't be detected. In which cases this auto-detection is necessary?

SSE4:
that's to be defined where to apply on case by case basis in the codebase.
if it's a file downloaded from HTTP (`conanmanifest.txt`?), then encoding should be obtained from HTTP headers (`Content-Type`, `charset`). it could be also defined by the protocol, but specifying `Accept` and/or `Accept-Charset` might be needed to enforce it to be compatible with all HTTP servers.
if it's a file created/edited by user (`conanfile.txt`, `profile`, `settings.yml`, `conan.conf`, etc.), then it could be in an arbitrary encoding, depends on user's system and text editor (e.g. on Windows systems certain editors, like `notepad.exe`, tend to default to UTF16LE with BOM). removing a detection here may result in a massive breakage and complaints about conan misreading a configuration.
[...]

PR #11384, opened 2022-06-02
remove BOM magic handling, everything is utf-8 unless stated explicitly

memsharded:
Close https://github.com/conan-io/conan/issues/11228


conans/test/unittests/client/util/files/load_test.py
```diff
@@ -1,20 +0,0 @@
-# -*- coding: utf-8 -*-
```
SSE4, line 1: but you removed the test, how do we know does it still load conanfile.txt with UTF16 LE/BE?

memsharded: That is the thing, I don't think that having conanfiles, or any other file to be read and processed by Conan in any random encoding is good. All Conan related files should be utf-8. 

If we think that there is some exceptional case like conanfile.txt, we might apply some logic to that specific case. What we cannot be doing is applying such BOM detection to every string and file that is read by Conan.

I will try to restore the text that throws a more clear error message if a conanfile.txt is not utf-8.

SSE4: I am okay to removing the detection for network data, and files conan creates on its own.
but for files created by user (such as conanfile.txt) it should be kept IMO
we already know it happens, and we already know editors store files in different encodings. we had valid report from real users.
the test removed was testing exactly `conanfile.txt` case, I think it should be kept, and it should be kept green.
it's fine removing `decode_text` everywhere in codebase, except these places handling user-created files (`conanfile.xt`, `profile`, `settings.yml`, etc.)

memsharded: Ok, I have maintained the check and allowing those encodings, just for ``conanfile.txt`` and ``profile``. Settings.yml, while allowing user changes, is kind of a Conan generated file too, lets keep it utf-8 at the moment.

lasote (approved):
Looks good, still a draft?

memsharded:
> Looks good, still a draft?

I think with the latest changes keeping the detection for user files, it should be ready.

SSE4:
> I think with the latest changes keeping the detection for user files, it should be ready.

I think it's good as well. are we missing any user-created files (maybe layout, editable, workspace, etc.?)

SSE4 (approved):

memsharded:
> I think it's good as well. are we missing any user-created files (maybe layout, editable, workspace, etc.?)

Layout, editable or workspace external files do not exist anymore in 2.0, so shouldn't be an issue.

PR #11391, opened 2022-06-03
More friendly cmake_layout 

### LINKED ISSUE #11189: [feature] Multiple parallel profiles support in `cmake_layout()`

lasote:
Changelog: Feature: Changed `cmake_layout` with a more natural and intuitive `folders.build` and `folders.generators`. Also changed the `CMakePresets.json` preset names.
Docs: https://github.com/conan-io/docs/pull/2618

Closes https://github.com/conan-io/conan/issues/11189

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -2,7 +2,7 @@ import json
 import os
 import platform
 
-from conan.tools.cmake.layout import get_build_folder_vars_suffix
+from conan.tools.cmake.layout import get_build_folder_custom_vars
 from conan.tools.cmake.utils import is_multi_configuration
 from conans.errors import ConanException
 from conans.util.files import save, load
@@ -20,22 +20,22 @@ def _add_build_preset(conanfile, multiconfig):
 
 def _build_preset_name(conanfile):
     build_type = conanfile.settings.get_safe("build_type")
-    suffix = get_build_folder_vars_suffix(conanfile)
-    if suffix:
+    custom_conf = get_build_folder_custom_vars(conanfile)
+    if custom_conf:
         if build_type:
-            return "{}-{}".format(build_type, suffix)
+            return "{}-{}".format(custom_conf, build_type.lower())
         else:
-            return suffix
+            return custom_conf
     return build_type or "default"
 
 
 def _configure_preset_name(conanfile, multiconfig):
     build_type = conanfile.settings.get_safe("build_type")
-    suffix = get_build_folder_vars_suffix(conanfile)
-    base = "default" if multiconfig or not build_type else build_type
-    if suffix:
-        return "{}-{}".format(base, suffix)
-    return base
+    custom_conf = get_build_folder_custom_vars(conanfile)
+    suffix = "default" if multiconfig or not build_type else str(build_type).lower()
+    if custom_conf:
+        return "{}-{}".format(custom_conf, suffix)
+    return suffix
 
 
 def _add_configure_preset(conanfile, generator, cache_variables, toolchain_file, multiconfig):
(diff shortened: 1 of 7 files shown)


memsharded (approved):
Love it, great job!

conan/tools/cmake/presets.py
```diff
@@ -20,22 +20,22 @@ def _add_build_preset(conanfile, multiconfig):
...
     build_type = conanfile.settings.get_safe("build_type")
-    suffix = get_build_folder_vars_suffix(conanfile)
-    base = "default" if multiconfig or not build_type else build_type
-    if suffix:
-        return "{}-{}".format(base, suffix)
-    return base
+    custom_conf = get_build_folder_custom_vars(conanfile)
+    suffix = "default" if multiconfig or not build_type else str(build_type).lower()
+    if custom_conf:
+        return "{}-{}".format(custom_conf, suffix)
```
mpusz: I would put `default` only if there is no `custom_conf`. It looks much better to have:
- config preset `gcc-11`
- build preset `gcc-11-release`

rather than:

- config preset `default-gcc-11`
- build preset `gcc-11-release`

In this case, what is the `default` needed for? I think that only `gcc-11` is much better.
Only when there is no `custom_conf` than we need to provide some name and `default` is as good as any other here in such case ;-)

lasote: Good point. Done!

franramirez688 (approved):

czoido (approved):

PR #11407, opened 2022-06-06
Add relative path support to Git SCM tool

### LINKED ISSUE #11408: [feature] Add relative path support to Git SCM tool
When trying to setup Conan packaging with the target source in a git submodule relative to the recipe, I was receiving
empty commit ids. Upon further investigation, it seems that the current assumption is that the folder path to `Git()`
will be absolute. A minor adjustment allows relative paths to be used as well: since the git command always `chdir`s to the
target git folder, we can perform the rev-parse relative to the current directory and receive the expected behavior.

Note that the current default value of "." is a relative path, which suggests that other relative paths should work.

nmasseyKM:
Changelog: Feature: Add relative path support to Git SCM tool.
Docs: omit

Closes: #11408

When trying to setup Conan packaging with the target source in a submodule relative to the recipe, I was receiving
empty commit ids. Upon further investigation, it seems that the current assumption is that the folder path to Git()
will be absolute. A minor adjustment allows relative paths to be used as well: since the git command always `chdir`s to the
target git folder, we can perform the `rev-parse` relative to the current directory and receive the expected behavior.

- Test cases have been added.
- Documentation does not seem to warrant updates since this behavior is implied by the currently document default value for `folder`

### DIFF
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -22,7 +22,7 @@ class Git(object):
             # --full-history is needed to not avoid wrong commits:
             # https://github.com/conan-io/conan/issues/10971
             # https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt-Defaultmode
-            commit = self._run('rev-list HEAD -n 1 --full-history -- "{}"'.format(self.folder))
+            commit = self._run('rev-list HEAD -n 1 --full-history -- "."')
             return commit
         except Exception as e:
             raise ConanException("Unable to get git commit in '%s': %s" % (self.folder, str(e)))
--- a/conans/test/unittests/tools/scm/test_git_get_commit.py
+++ b/conans/test/unittests/tools/scm/test_git_get_commit.py
@@ -1,6 +1,7 @@
 import os
 
 from conan.tools.scm import Git
+from conans.tools import chdir
 from conans.test.utils.mocks import MockConanfile
 from conans.test.utils.tools import TestClient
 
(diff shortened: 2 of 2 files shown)


conan/tools/scm/git.py
```diff
@@ -22,7 +22,7 @@ def get_commit(self):
             # --full-history is needed to not avoid wrong commits:
             # https://github.com/conan-io/conan/issues/10971
             # https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt-Defaultmode
-            commit = self._run('rev-list HEAD -n 1 --full-history -- "{}"'.format(self.folder))
+            commit = self._run('rev-list HEAD -n 1 --full-history -- "."')
```
memsharded, line 25: But this change will not guarantee that this is run in the specified ``self.folder``, but in the ``cwd``, which is not guarantee that it will be ``self.folder``, if users specify a different value in the constructor, which they can.

conans/test/unittests/tools/scm/test_git_get_commit.py
```diff
@@ -35,11 +35,17 @@ def test_change_branch_in_root_commit():
 def test_multi_folder_repo():
     c = TestClient()
     conanfile = MockConanfile({})
-    c.save({"lib_a/conanfile.py": ""})
+    c.save({"conanfile.py": ""})
```
memsharded: If this test was not flawed by design, it is much better to leave it as-is, and then create a new test, even if it is very similar to the existing one, with small variations.

memsharded (commented):
Thanks very much for contributing to Conan.
Please have a look to the comments, and to the failing tests in CI: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-11407/2/pipeline

conan/tools/scm/git.py
nmasseyKM: The definition of `_run` above executes with `chdir(self.folder)`
The double use of self.folder means it ends up trying to resolve a relative path relative to the folder itself, and usually fails to find any commits touching that subtree.

Given the other uses of `_run`, its behavior seems correct and the adjustment to `get_commit` seemed more appropriate.

conans/test/unittests/tools/scm/test_git_get_commit.py
nmasseyKM: Sure. I started that way and noticed I was duplicating a lot of setup.

conan/tools/scm/git.py
memsharded: Oh, yes, you are right, I missed that part. Looks good to me then.

memsharded (approved):

conans/test/unittests/tools/scm/test_git_get_commit.py
memsharded: Yes, I know, but in those cases we prefer to do it in 2 steps:
- First add new tests, so it is evident that existing tests are not altered. Get approval and merge the PR
- Do a second PR which is a pure refactor of tests, extracting the common parts.

Just a detail, but as "not breaking" is important for us, we like to do it this way if possible. Thanks!

conans/test/unittests/tools/scm/test_git_get_commit.py
```diff
@@ -97,3 +97,121 @@ def test_multi_folder_repo():
     c.run_command("git rev-parse HEAD")
     commit_real = str(c.out).splitlines()[0]
     assert new_commit_root == commit_real
+
+def test_relative_folder_repo():
+    def _chdir(new_path):
```
nmasseyKM: I couldn't find a helper on TestClient to do this, and I wanted to minimize use of conan source in unit tests.
Otherwise, I'd just import `chdir` from `conans.tools.files`

memsharded: Yes, ``TestClient`` has a ``chdir()`` ``@contextmanager`` that you can use as ``with myclient.chdir("myfolder"):``

nmasseyKM: nevermind, I see the tools being used in other tests.

PR #11414, opened 2022-06-07
Fixed two bugs with CMakePresets

### LINKED ISSUE #11409: Prevent repeating the same preset in `CMakePresets.json`
Is it possible to prevent repeating the same preset in `CMakePresets.json`? If I run `conan install` several times I end up with:

```json
   ...,
    "buildPresets": [
        {
            "name": "Release-gcc-11",
            "configurePreset": "default-gcc-11",
            "configuration": "Release"
        },
        {
            "name": "Release-gcc-11",
            "configurePreset": "default-gcc-11",
            "configuration": "Release"
        }
    ],
   ...
```

_Originally posted by @mpusz in https://github.com/conan-io/conan/issues/11189#issuecomment-1147720498_

### LINKED ISSUE #11413: [bug] CMakeUserPresets.json including missing directory
Calling "conan install" for different configurations, especially using the `tools.cmake.cmake_layout:build_folder_vars`,  if you remove one of the folders generated e.g: `build/gcc-5` and then you perform new `conan install` commands, the entry for the inclusion of `build/gcc-5/generators/CMakePresets.json` is not cleaned, causing `cmake` failing with a hard to understand error.

lasote:
Changelog: Bugfix: The `CMakePresets.json` file, when performing several `conan install` commands, ended-up with several entries of duplicated "buildPresets" if used with a `multi-configuration` cmake generator.
Changelog: Fix:  The "include" paths list in a `CMakeUserPresets.json`, when performing a new `conan install`, are now checked, and in case some of them is a missing path (deleted directory), it is cleaned from the list. 
Docs: omit

Closes https://github.com/conan-io/conan/issues/11413
Closes https://github.com/conan-io/conan/issues/11409

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -123,21 +123,13 @@ def write_cmake_presets(conanfile, toolchain_file, generator):
                 data = {"version": 4, "include": [preset_path]}
             else:
                 data = json.loads(load(user_presets_path))
-                data = _clean_deleted_presets(data)
+                # Clear the folders that have been deleted
+                data["include"] = [i for i in data["include"] if os.path.exists(i)]
                 if preset_path not in data["include"]:
                     data["include"].append(preset_path)
 
             data = json.dumps(data, indent=4)
             save(user_presets_path, data)
-
-
-def _clean_deleted_presets(data):
-    includes = []
-    for include in data["include"]:
-        if os.path.exists(include):
-            includes.append(include)
-    data["include"] = includes
-    return data
 
 
 def load_cmake_presets(folder):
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/presets.py
```diff
@@ -126,13 +123,23 @@ def write_cmake_presets(conanfile, toolchain_file, generator):
...
             save(user_presets_path, data)
 
 
+def _clean_deleted_presets(data):
+    includes = []
+    for include in data["include"]:
+        if os.path.exists(include):
+            includes.append(include)
+    data["include"] = includes
+    return data
```
memsharded:
```suggestion
def _clean_deleted_presets(data):
    data["include"] = [i for i in data["include"] if os.path.exists(i)]
```

And as it is called in a single place, maybe just use this single line instead

memsharded (approved):

PR #11416, opened 2022-06-07
Fix: unittests should not assume "master" is git's default branch

nmasseyKM:
Changelog: Omit
Docs: Omit

This patchset repairs five tests which fail if git's default branch is changed
from `master`.

`git` allows for changing the default branch name during `init` by setting
`init.defaultbranch`. If a user has this set, some of the unit tests start failing.
We should minimize the number of assumptions made when a developer
is running `python -m pytest` from within a virtualenv.

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -245,7 +245,9 @@ class XcodeDeps(object):
         # Generate the config files for each component with name conan_pkgname_compname.xcconfig
         # If a package has no components the name is conan_pkgname_pkgname.xcconfig
         # Then all components are included in the conan_pkgname.xcconfig file
-        for dep in self._conanfile.dependencies.host.values():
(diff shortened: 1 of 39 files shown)


memsharded (approved):
Looking good, thanks!

memsharded (commented):
Having a second thought, wouldn't it be more robust to force the definition of the default branch inside ``client.init_git_repo()``? that way, it will doable in just 1 place, and it will be future proof, if new tests are added that don't do the try-getdefaultbranch-except, they will still be ok.

What do you think?

memsharded:
I have done the changes, the idea is that those are the minimal changes to avoid the problem, while being robust against the same error repeating again in the future for new tests.

After this is merged, the default ``master`` can be changed to ``main`` (and tests fixed accordingly), but that should be better done in a separate PR.

Issue #11327, opened 2022-05-25
[develop2] [feature] Add ``prev``, ``settings``, ``options`` to ``conan graph info`` output


memsharded:
Now the serialization of the information is missing some fields.

Verify that subsettings are also dumped.

SeanSnyders:
As per feature request #11362, this command should also return relevant information that would fully describe the state for consumption of the package. This should include details/info from `conanfile.cpp_info` (amongst others) as is constructed via the `package_info()` method.
This would enable the consumption of Conan packages in toolchains that do not depend on Conan itself.

memsharded:
Fixed by https://github.com/conan-io/conan/pull/11417

The serialization of the full ``cpp_info`` remains open. Moving to next iteration.

franramirez688:
Since https://github.com/conan-io/conan/pull/11434 was merged, full `cpp_info` is now serialized so we can close the issue. It will be released in the next _Conan 2.0.0-beta1_ version.

PR #11417, opened 2022-06-07
[dev2] Adding new fields to `Node` serialization

franramirez688:
Changelog: Feature: Adding `prev` to `Node.serialize()` method.
Changelog: Feature: Adding subsettings to `Settings.serialize()` method.
Closes: https://github.com/conan-io/conan/issues/11327
Docs: omit


conans/model/settings.py
```diff
@@ -172,7 +172,8 @@ def __init__(self, definition=None, name="settings", parent_value="settings"):
                       for k, v in definition.items()}
 
     def serialize(self):
-        return {k: v.value for k, v in self._data.items()}
+        # Returning all the information available, including the sub-settings
+        return self.items()
```
memsharded: But this change is not returning a "string" object, but a ``SettingsItem`` one.
The idea is that ``serialize()`` methods return hierarchy of python primitives, strings, dicts, lists, numbers...

conans/model/settings.py
```diff
@@ -172,7 +172,8 @@ def __init__(self, definition=None, name="settings", parent_value="settings"):
                       for k, v in definition.items()}
 
     def serialize(self):
-        return {k: v.value for k, v in self._data.items()}
+        # Returning all the information available, including the sub-settings
+        return {k: v for k, v in self.items()}
```
memsharded: This was done on purpose to not keep using the legacy bit of mess that is behind ``self.items()``, that calls ``self.fields()``,  ``self.values_list``.

memsharded (commented):
Looks good

lasote (approved):

conans/model/settings.py
franramirez688: Ok, let me have a look at it.

franramirez688: Refactored a little bit

memsharded (approved):

PR #11421, opened 2022-06-08
Fix colors in output windows cmd

lasote:
Fixed the colors in any windows terminal (and the rest).

Close https://github.com/conan-io/conan/pull/11421

### DIFF
--- a/conans/client/userio.py
+++ b/conans/client/userio.py
@@ -12,31 +12,22 @@ def is_terminal(stream):
 
 def color_enabled(stream):
     """
-    NO_COLOR: No colors and no colorama
-    CLICOLOR: When 1, colors if is_terminal
-    CLICOLOR_FORCE: When 1, always colors (except if NO_COLOR)
+    NO_COLOR: No colors
+
+    https://no-color.org/
+
+    Command-line software which adds ANSI color to its output by default should check for the
+    presence of a NO_COLOR environment variable that, when present (**regardless of its value**),
+    prevents the addition of ANSI color.
     """
-    if os.getenv("NO_COLOR") == "1":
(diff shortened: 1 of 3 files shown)


conans/client/userio.py
```diff
@@ -26,14 +27,17 @@ def color_enabled(stream):
 
 def init_colorama(stream):
     import colorama
-
     if not color_enabled(stream):
-        if os.getenv("NO_COLOR") != "1":
-            colorama.init(strip=True)
-        return False
+        if os.getenv("NO_COLOR") == "1":
+            # Do not initialize colorama
```
lasote: Before anyone says this can be simplified. I know, but honestly I prefer it like this to show that the "NO_COLOR" means no initialize colorama, otherwise initialize colorama but striping. This is complex enough to make it even more.

conans/client/userio.py
```diff
@@ -26,14 +27,17 @@ def color_enabled(stream):
 
 def init_colorama(stream):
     import colorama
-
     if not color_enabled(stream):
-        if os.getenv("NO_COLOR") != "1":
-            colorama.init(strip=True)
-        return False
+        if os.getenv("NO_COLOR") == "1":
```
czoido: I see here: https://no-color.org/ that: _"when present (**regardless of its value**), prevents the addition of ANSI color._"
So maybe checking that is not `None` is ok.

SSE4:
cmd.exe
image

powershell
image

windows terminal
image

cygwin
image

msys2
image

far manager
image

SSE4 (approved):

SSE4:
also works from inside the PyCharm terminal:
image

czoido (approved):

PR #11427, opened 2022-06-09
[dev2] Json formatter for `conan create`

franramirez688:
Changelog: Feature: Added JSON formatter to `conan create` command.
Changelog: Feature: Added `serialize` method to `Profile` class.
Depends on: https://github.com/conan-io/conan/pull/11417 (this must be merged first)


conans/cli/commands/create.py
```diff
@@ -14,7 +15,17 @@
...
 
-@conan_command(group=COMMAND_GROUPS['creator'])
+def json_create(info):
+    deps_graph, profile_build, profile_host = info
+    result = {
+        "profile_build": profile_build.serialize(),
+        "profile_host": profile_host.serialize(),
+        "deps_graph": deps_graph.serialize(),
+    }
+    return json.dumps(result, indent=4)
```
SSE4: can it output a package revision created?

franramirez688: The `deps_graph` has that information.

SSE4: thanks. it wasn't obvious, I was thinking it's just an information about dependencies.
is it something like `deps_graph[<package name>].prev`?

franramirez688: Yeah, I'm working on that. I only opened the Draft to see all the tests are running ok. Now, let me think of the final format. I hope it will be clearer later.

conans/cli/commands/create.py
```diff
@@ -14,7 +15,17 @@
 from conans.util.files import chdir, mkdir
 
 
-@conan_command(group=COMMAND_GROUPS['creator'])
+def json_create(info):
+    deps_graph, profile_build, profile_host = info
+    result = {
+        "profile_build": profile_build.serialize(),
+        "profile_host": profile_host.serialize(),
+        "deps_graph": deps_graph.serialize(),
```
memsharded: lets use just ``graph`` for naming
```suggestion
        "graph": deps_graph.serialize(),
```

conans/model/conf.py
```diff
@@ -114,6 +114,17 @@ def dumps(self):
         else:
             return "{}={}".format(self._name, self._value)
 
+    def serialize(self):
+        if self._value is None:
+            _value = "!"  # unset
+        elif self._value_type is list and _ConfVarPlaceHolder in self._value:
+            v = self._value[:]
+            v.remove(_ConfVarPlaceHolder)
```
memsharded, line 122: This is indeed a challenge to represent in a serialized format.

conans/model/settings.py
```diff
@@ -172,7 +172,8 @@ def __init__(self, definition=None, name="settings", parent_value="settings"):
                       for k, v in definition.items()}
 
     def serialize(self):
-        return {k: v.value for k, v in self._data.items()}
+        # Returning all the information available, including the sub-settings
+        return {k: v for k, v in self.items()}
```
memsharded: This needs to be updated from the other PR

conans/model/conf.py
memsharded: I mean, this serialization is losing information, cannot reconstruct profile from it.

SSE4: the profile looks like an input of the `conan create`, and you already know it prior calling the command.
you don't have to obtain it as a result of the `conan create` - I believe profile is immutable and never modified by the command, right?
in this case, I think JSON resut of the conan should describe the output of the command - e.g. package ID and package revision created.

conans/test/integration/command/create_test.py
```diff
@@ -409,3 +409,140 @@ def test_create_build_missing():
...
+                                 'compiler.libcxx': 'libstdc++',
+                                 'compiler.version': '12',
+                                 'os': 'Linux'},
+                    'tool_requires': {}}
+    assert profile_host == info["profile_host"]
+    graph_info = info["graph"]
+    hello_pkg_ref = 'hello/0.1#18d5440ae45afc4c36139a160ac071c7'
+    pkg_pkg_ref = 'pkg/0.2#db78b8d06a78af5c3ac56706f133098d'
+    assert graph_info["conanfile"]["recipe"] == "Virtual"
+    assert graph_info["conanfile"]["package_id"] is None
```
SSE4: why None? I don't get why `package_Id` of conanfile cannot be computed. same for PREV.

franramirez688: Because this is the consumer information, purely, your ` conanfile.py` one. Then, that `package_id` has to be `None`, and, if you read below, you'll see `assert graph_info[hello_pkg_ref]["package_id"] == "8eba237c0fb239fcb7daa47979ab99258eaaa7d1"` which is the created package for that conanfile.

conans/model/conf.py
franramirez688: I put them there because I considered that they could be useful, but if you all think they're not giving any extra information, I can remove them.

SSE4: yeah, that's the information you already should know prior to calling `conan create`, so might be useless

franramirez688: @memsharded wdyt?

memsharded: Yes, at this point in time, I think it is better to leave it out from this command. We were commenting that this was a possibility to add in the future, but it is not really necessary to add the profiles to the json output at this moment, that will have lower risk.

franramirez688: Done. I've kept the `serialize()` function just in case.

conans/test/integration/command/create_test.py
```diff
@@ -409,3 +409,114 @@ def test_create_build_missing():
...
+def test_create_format_json():
+    """
+    Tests the ``conan create . -f json`` result
+
+    The result should be something like:
+
+    {
+        'graph': {
+            'conanfile': {...},  # Consumer
+            'hello/0.1#18d5440ae45afc4c36139a160ac071c7': {...},
```
lasote: Is this recursive or something? Is the graph represented somehow?

conans/cli/commands/create.py
```diff
@@ -14,7 +15,15 @@
 from conans.util.files import chdir, mkdir
 
 
-@conan_command(group=COMMAND_GROUPS['creator'])
+def json_create(info):
+    deps_graph, profile_build, profile_host = info
+    graph_info = {}
+    for n in deps_graph.serialize()["nodes"]:
```
memsharded: This is weird, that the ``nodes`` field is being removed just for this json. This will create inconsistencies in other commands. So if we want to remove the ``nodes`` it should be done in ``graph.serialize()``, but not here.

conans/model/conf.py
memsharded: Ok, but please mark this as #FIXME to indicate this is not a definitive thing, and require some thinking.

conans/test/integration/command/create_test.py
```diff
@@ -409,3 +409,114 @@ def test_create_build_missing():
...
+            'pkg/0.2#44a1a27ac2ea1fbcf434a05c4d57388d': {'...},
+        }
+        'profile_build': {'build_env': '',
```
memsharded: Remove this, not in the json atm

memsharded (changes requested):

conans/cli/commands/create.py
memsharded: I am not saying that it should be removed in ``graph.serialize()``, it depends on the model there.

franramirez688: Then, let me get it back. I guess it's not worth changing it in `graph.serialize()` for now.

czoido: I agree with getting the `nodes` field back, besides that, LGTM

czoido (approved):

conans/test/integration/command/create_test.py
franramirez688: Yes, the graph is represented as it is. I got back the original serialization so it's using the whole information from `DepsGraph.serialize()` without pruning anything.

memsharded (approved):

lasote (approved):

PR #11441, opened 2022-06-12
[develop2] Let dependency define package_id modes

memsharded:


lasote (commented):
Ok to the code & implementation but I would wait until the feedback of the Tribe. 
It could be quite messy to debug in environments like conan-center.

memsharded:
Agree, lets move it to beta.2 then, and wait for the feedback.

lasote (approved):

conans/model/requires.py
```diff
@@ -321,6 +324,10 @@ def deduce_package_id_mode(self, pkg_type, dep_pkg_type, non_embed_mode, embed_m
             self.package_id_mode = "unrelated_mode"
             return
 
+        # If the dependency defines the mode, that has priority over default
+        embed_mode = getattr(dep_conanfile, "embed_mode", embed_mode)
```
memsharded: Maybe the naming should be more explicit? Like ``package_id_embed_mode``?

memsharded: Done

lasote (approved):

czoido (approved):

Issue #11245, opened 2022-05-13
[question] deterministic build and source dir 


boussaffawalid:
I finally manage to achieve 100% deterministic builds. One of the issue was randomness in conan build and source directories when creating conan packages. 

The solution I used with the [old cmake build helper  ](https://docs.conan.io/en/1.43/reference/build_helpers/cmake.html) was to create directory junction and pass them to cmake instead of the conan source and build folders: 

```
mklink /J x:\ci\source\ self.source_folder
mklink /J x:\ci\build\ self.build_folder
```

```
cmake.configure(source_dir="x:\ci\source\", build_dir="x:\ci\build\")
```

Now I switched to the new conan helpers https://docs.conan.io/en/latest/reference/conanfile/tools/cmake.html and it is not possible anymore to override source_dir and build_dir. 

I tried using the layout method

```
def layout(self):
    self.folders.source = "x:\ci\source\"
    self.folders.build = "x:\ci\build\"
```
but I got blocked by source_dir not available in the layout method.

Any advices on how to achieve deterministic build and source dirs in conan.

memsharded:
Using an absolute path for all packages seems a bit fragile, IMHO.

Then I'd like to understand that "randomness", because Conan cache folders are deterministic in 1.X (for the same package, they build in the same folder always). What am I missing?

boussaffawalid:
Build and source folder information are propagated to binaries because of many reasons (usage of `__FILE__` macro for example).  
When rebuilding the same source code but in different location, you get different binaries. And because conan use different paths for different versions, final binaries are different even if source code did not change.

memsharded:
> Build and source folder information are propagated to binaries because of many reasons (usage of __FILE__ macro for example).
When rebuilding the same source code but in different location, you get different binaries. 

Sure, we know about this :) https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html

The question is why do you need to do that. If you are using Conan 1.X, as long as you put the Conan cache in the same place (lets say, ``C:/myconan``, you will also get exactly the same path, like ``.../data/pkg/version/user/channel/package_id/build``, this path is fully deterministic in Conan 1.X.


The second question is regarding: 

> but I got blocked by source_dir not available in the layout method.

What do you mean? No idea what ``source_dir`` in the ``layout()`` method means, yes, in that method you get ``self.folders.xxx`` definitions, but you can define that there.

boussaffawalid:
[...]

PR #11447, opened 2022-06-13
[develop2] use always same paths in the conan cache for temporary folders too

memsharded:
Let the cache always use the same paths for builds

related: https://github.com/conan-io/conan/issues/11245

### DIFF
--- a/conan/cache/cache.py
+++ b/conan/cache/cache.py
@@ -1,7 +1,6 @@
 import hashlib
 import os
 import shutil
-import uuid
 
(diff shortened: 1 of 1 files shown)


jcar87 (approved):
LGTM!

If conan wipes the contents of the build folder at the beginning (if it existed previously), and it has a separate location per package ID, it makes sense that this isn't randomised. Especially if there's "less" risk by the removal of the `--keep-build` option in conan 2

lasote (approved):

czoido (approved):

PR #11455, opened 2022-06-14
Remove default Frameworks

lasote:
Changelog: Fix: When using the new layout() feature of Conan 2.0, change the default `cpp_info.frameworkdirs = ["Frameworks"]` to `cpp_info.frameworkdirs = []`, because it is more common to not have packaged Apple frameworks and declaring a missing folder can cause issues with the new toolchains.
Docs: https://github.com/conan-io/docs/pull/2613


memsharded (approved):
Seems this is fine, as it is protected by the ``layout()`` existence opt-in

lasote:
I'll double-check that the default `["Frameworks"]` is there without the `layout()`

memsharded (approved):

PR #11458, opened 2022-06-14
[develop2][wip] check if package_id(), validate() do not define settings and options

memsharded:


memsharded:
Continues in https://github.com/conan-io/conan/pull/11462

PR #11461, opened 2022-06-15
Test weird behavior header-only

lasote:


czoido:
Closed by https://github.com/conan-io/conan/pull/11462

PR #11454, opened 2022-06-14
Reproduce fail with compatible packages in develop2

czoido:
Changelog: (Feature | Fix | Bugfix): Describe here your pull request
Docs: https://github.com/conan-io/docs/pull/XXXX

Reproducing this: https://ci.conan.io/blue/organizations/jenkins/Examples2.0/detail/PR-23/7/pipeline/24/

It can be avoided accessing the option with get_safe, but it may be worth checking

PR #11462, opened 2022-06-15
Compatible packages fail fixes

memsharded:
Close https://github.com/conan-io/conan/pull/11461
Close https://github.com/conan-io/conan/pull/11454

- Removes the assignment of ``settings`` and ``options`` to use the ``info.settings`` and ``info.options`` automagically inside ``validate()`` and ``package_id()`` methods
- Important: ``validate()`` and ``package_id()`` methods should operate always now with ``self.info`` fields, otherwise ``compatibility will be broken``
- Relaxed the assignment of ``settings`` and ``options`` of compatible packages to ``self.settings`` and ``self.options`` so the values are updated within ``package_info()``. Now, only an ``update`` is computed, but removed settings and options in ``package_id()`` will still have a value in ``self.settings`` and ``self.options`` in ``package_info()``.
- If some ``package_info()`` needed more accurate information of what has been removed, they should check ``self.info.settings`` instead.

### DIFF
--- a/conans/client/graph/compute_pid.py
+++ b/conans/client/graph/compute_pid.py
@@ -59,7 +59,7 @@ def run_validate_package_id(conanfile):
     # IMPORTANT: This validation code must run before calling info.package_id(), to mark "invalid"
     if hasattr(conanfile, "validate"):
         with conanfile_exception_formatter(conanfile, "validate"):
-            with conanfile_remove_attr(conanfile, ['cpp_info'], "validate"):
+            with conanfile_remove_attr(conanfile, ['cpp_info', 'settings', 'options'], "validate"):
                 try:
                     conanfile.validate()
                 except ConanInvalidConfiguration as e:
@@ -70,7 +70,7 @@ def run_validate_package_id(conanfile):
     # Once we are done, call package_id() to narrow and change possible values
     if hasattr(conanfile, "package_id"):
         with conanfile_exception_formatter(conanfile, "package_id"):
-            with conanfile_remove_attr(conanfile, ['cpp_info'], "package_id"):
+            with conanfile_remove_attr(conanfile, ['cpp_info', 'settings', 'options'], "package_id"):
                 conanfile.package_id()
 
     conanfile.info.validate()
--- a/conans/client/graph/compute_pid.py
+++ b/conans/client/graph/compute_pid.py
@@ -49,8 +49,6 @@ def compute_package_id(node, new_config):
                                conf=conanfile.conf.copy_conaninfo_conf())
     conanfile.original_info = conanfile.info.clone()
 
-    conanfile.settings = conanfile.info.settings
-    conanfile.options = conanfile.info.options
     run_validate_package_id(conanfile)
 
     info = conanfile.info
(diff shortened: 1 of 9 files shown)


czoido (approved):

franramirez688 (approved):

conans/test/functional/toolchains/microsoft/test_msbuild.py
```diff
@@ -607,21 +607,3 @@ def test_toolchain_win_multi(self):
             else:
                 self.assertNotIn("hello.dll", client.out)
             self.assertIn("KERNEL32.dll", client.out)
-
-
-def test_msvc_runtime_flag_common_usage():
```
memsharded, line 612: This test was removed, because it was a weird usage of the ``msvc_runtime_flag()`` for validation. This would raise as it is internally accessing ``self.settings``.

If validation of the runtime is desired:
- ``if self.info.settings.get_safe("compiler.runtime") == "static": raise ConanInvalid....`` (only works with ``msvc`` compiler)
- If we need backwards compatibility to Conan 1.X, we need a helper that helps abstracting ``Visual Studio`` and ``msvc`` compilers. ConanCenter has stopped creating ``shared=True`` + VS MT builds, shouldn't be an issue, the 2 recipes that implement this ``validate()`` can remove it.

conans/test/integration/package_id/test_cache_compatibles.py
```diff
@@ -236,3 +236,131 @@ class Pkg(ConanFile):
...
+        conanfile = textwrap.dedent("""
+            from conan import ConanFile
+            from conan.errors import ConanErrorConfiguration
+            from conan.tools.build import check_min_cppstd, valid_min_cppstd
+            class Pkg(ConanFile):
+                name = "pkg"
+                version = "0.1"
+                settings = "os", "arch", "compiler", "build_type"
+                def validate(self):
+                    if int(str(self.info.settings.compiler.cppstd).replace("gnu", "")) <= 14:
```
memsharded, line 347: Building this validation atm (for hard ConfigurationError, not fallback to other compatible binary) is ugly, but feasible. We can improve ``check_min_cppstd()`` helper later.

conans/client/graph/compute_pid.py
```diff
@@ -59,7 +59,7 @@ def run_validate_package_id(conanfile):
     # IMPORTANT: This validation code must run before calling info.package_id(), to mark "invalid"
     if hasattr(conanfile, "validate"):
         with conanfile_exception_formatter(conanfile, "validate"):
-            with conanfile_remove_attr(conanfile, ['cpp_info'], "validate"):
+            with conanfile_remove_attr(conanfile, ['cpp_info', 'settings', 'options'], "validate"):
```
lasote, line 62: Missing a test for this?

memsharded: The options are covered by test change: 
```
assert "OPTIONS: 'self.options' access in 'package_id()' method is forbidden" in client.out
```
In https://github.com/conan-io/conan/pull/11462/files#diff-c37f80c54d0b5b5a86d89e74a9e8b7babd561bdd1d99956812971941f2ec11a0R120

Do you think a settings one is also needed?

lasote: This is `validate` method, not `package_id`

memsharded: Added tests

lasote (approved):

franramirez688 (approved):
Cool!

czoido (approved):

PR #11471, opened 2022-06-16
Feature/components default new cppinfo

lasote:
Changelog:  Fix: When using the `layout()` (as opt-in of some Conan 2.0 features) the default `includedirs` of the declared components is now `["include"]` the default of the `libdirs` is `["lib"]` and the default for the `bindirs` is `["bin"]`. Previously, the components always had empty fields. Also the default of `cpp_info.resdirs` is now empty.
Docs: https://github.com/conan-io/docs/pull/2613

### DIFF
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -203,7 +203,7 @@ class AutotoolsToolchain:
                                        _get_argument("includedir", "includedirs"),
                                        _get_argument("oldincludedir", "includedirs"),
                                        _get_argument("datarootdir", "resdirs")])
-        return configure_install_flags
+        return [el for el in configure_install_flags if el]
 
     def _default_autoreconf_flags(self):
         return ["--force", "--install"]
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -675,7 +675,7 @@ class BinaryInstaller(object):
                         conanfile.cpp_info = CppInfo(conanfile.name, package_folder,
                                                      default_values=CppInfoDefaultValues())
                         if not is_editable:
-                            conanfile.cpp.package.set_relative_base_folder(conanfile.package_folder)
+                            # conanfile.cpp.package.set_relative_base_folder(conanfile.package_folder)
                             # Copy the infos.package into the old cppinfo
                             fill_old_cppinfo(conanfile.cpp.package, conanfile.cpp_info)
                         else:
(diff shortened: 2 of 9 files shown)


conans/client/installer.py
```diff
@@ -675,7 +675,6 @@ def _call_package_info(self, conanfile, package_folder, ref, is_editable):
                         conanfile.cpp_info = CppInfo(conanfile.name, package_folder,
                                                      default_values=CppInfoDefaultValues())
                         if not is_editable:
-                            conanfile.cpp.package.set_relative_base_folder(conanfile.package_folder)
```
lasote, line 678: Apparently, this was unnecessary all the paths that are declared later while `conanfile.package_info()`, are also converted to abs, so no need to do it here for the "defaults" or the "cpp.package".

conans/model/new_build_info.py
```diff
@@ -40,6 +40,10 @@ def __init__(self):
         self.sysroot = None
         self.requires = None
 
+        if with_defaults:
+            self.includedirs = ["include"]
```
memsharded, line 44: And not ``bin``?

conans/model/new_build_info.py
```diff
@@ -70,10 +74,10 @@ def get_init(self, attribute, default):
 
 class NewCppInfo(object):
 
-    def __init__(self):
-        self.components = DefaultOrderedDict(lambda: _NewComponent())
+    def __init__(self, with_defaults=False):
```
memsharded, line 78: It is not clear who is calling ``NewCppInfo(with_defaults=True)`` to change this default behavior.

lasote: https://github.com/conan-io/conan/pull/11471/files#diff-f33d0f25da77d5b8727e9c2e931247512829f64f3ac1e688ae7aa35e0a06e39cR11

conans/model/new_build_info.py
lasote: Yeah, we can add it too.

lasote:
@memsharded Pending to define if we want default `resdirs`=["res"]. I've removed it in my last commit.

memsharded:
> @memsharded Pending to define if we want default resdirs=["res"]. I've removed it in my last commit.

yes, I think we can leave ``res`` out, because that is actually very uncommon, if it is not common better do not define it as default.

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -203,7 +203,7 @@ def _get_argument(argument_name, cppinfo_name):
                                        _get_argument("includedir", "includedirs"),
                                        _get_argument("oldincludedir", "includedirs"),
                                        _get_argument("datarootdir", "resdirs")])
-        return configure_install_flags
+        return [el for el in configure_install_flags if el]
```
lasote, line 206: When empty "cpp.package.resdirs" this caused (harmful but weird) `''` configure argument.

memsharded (approved):

paulharris:
@lasote I think you need to look at this again.
This change broke my vtk recipe (expectedly) as it was not setting resdirs, and I had assumed the files would end up in "res" eg "res/licences"

However, the files now end up packagedir/None/licences !
What is the "None" doing there?
What is an empty resdirs supposed to mean exactly?  Should the files end up in packagedir/licences going forward?

paulharris:
Also, side note, there doesn't seem to be any docs about adjusting dirs (in the layout(self) method) AND using the cmake_layout(self) call.

eg: 
https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html?highlight=resdirs
does not include any mention of cmake_layout(self)

Nor this page: https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmaketoolchain.html?highlight=resdirs

It is not clear if further adjustments should be done before, or after cmake_layout(self) ... ie it is not clear if cmake_layout will overwrite any previous settings, or will incorporate them...

paulharris:
Last problem, the migration docs say `You can adjust the cpp_info in the layout method too, not only for a package in the cache, that was typically done in the package_info() method using the self.cpp_info,` 

but I have found setting `self.cpp_info.components[comp].resdirs = ["res"]` has no effect, yet that pattern is symmetrical with the old setting for libdirs.

PR #11505, opened 2022-06-22
MSBuildDeps: map armv8 to ARM64 platform (#11504)

### LINKED ISSUE #11504: [feature] Add armv8 support to MSBuildDeps generator
Map `armv8` arch to MSBuild `ARM64` platform

PR to follow.

Passing  `-g MSBuildDeps -s arch=armv8` to `conan install` results in errors of the form:

```log
conanfile.txt: ERROR: Traceback (most recent call last):
  File "xxx\venv\lib\site-packages\conans\client\generators\__init__.py", line 177, in write_generators
    generator.generate()
  File "xxx\venv\lib\site-packages\conan\tools\microsoft\msbuilddeps.py", line 128, in generate
    raise ConanException("MSBuildDeps.platform is None, it should have a value")
conans.errors.ConanException: MSBuildDeps.platform is None, it should have a value

ERROR: Error in generator 'MSBuildDeps': MSBuildDeps.platform is None, it should have a value
```

markferry:
Changelog: Feature: map `armv8` arch to `ARM64` MSBuild platform in MSBuildDeps generator.

Docs: Omit

### DIFF
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -10,7 +10,7 @@ def msbuild_verbosity_cmd_line_arg(conanfile):
 
 
 def msbuild_arch(arch):
-    return {'x86': 'Win32',
+    return {'x86': 'x86',
             'x86_64': 'x64',
             'armv7': 'ARM',
             'armv8': 'ARM64'}.get(str(arch))
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -10,7 +10,7 @@ def msbuild_verbosity_cmd_line_arg(conanfile):
 
 
 def msbuild_arch(arch):
-    return {'x86': 'x86',
+    return {'x86': 'Win32',
             'x86_64': 'x64',
             'armv7': 'ARM',
             'armv8': 'ARM64'}.get(str(arch))
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -6,6 +6,7 @@ from xml.dom import minidom
 from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
+from conan.tools.microsoft.msbuild import msbuild_arch
 from conans.errors import ConanException
 from conans.util.files import load, save
 
@@ -105,8 +106,7 @@ class MSBuildDeps(object):
     def __init__(self, conanfile):
         self._conanfile = conanfile
         self.configuration = conanfile.settings.build_type
-        self.platform = {'x86': 'Win32',
-                         'x86_64': 'x64'}.get(str(conanfile.settings.arch))
+        self.platform = msbuild_arch(conanfile.settings.arch)
         # ca_exclude section
         # TODO: Accept single strings, not lists
         self.exclude_code_analysis = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis",
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -6,7 +6,6 @@ from xml.dom import minidom
 from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
-from conan.tools.microsoft.msbuild import msbuild_arch
 from conans.errors import ConanException
 from conans.util.files import load, save
 
@@ -106,7 +105,12 @@ class MSBuildDeps(object):
     def __init__(self, conanfile):
         self._conanfile = conanfile
         self.configuration = conanfile.settings.build_type
-        self.platform = msbuild_arch(conanfile.settings.arch)
+        # TODO: This platform is not exactly the same as ``msbuild_arch``, because it differs
+        # in x86=>Win32
+        self.platform = {'x86': 'Win32',
+                         'x86_64': 'x64',
+                         'armv7': 'ARM',
+                         'armv8': 'ARM64'}.get(str(conanfile.settings.arch))
         # ca_exclude section
         # TODO: Accept single strings, not lists
         self.exclude_code_analysis = self._conanfile.conf.get("tools.microsoft.msbuilddeps:exclude_code_analysis",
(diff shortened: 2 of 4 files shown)


markferry:
Needs a testcase...

memsharded (commented):
Thanks for contributing. 

Better please add the full one:

```python
{"x86": "Win32",
                    "x86_64": "x64",
                    "armv7": "ARM",
                    "armv8": "ARM64"}.get(...)
```
There is an instance of this in ``CMakeToolchain``, maybe we should factorize it (but that can be done later, just focus on getting this right).

For tests, it is enough to add an ``integration`` test, don't hesitate to ask for guidance if you need it.

markferry:
> There is an instance of this in `CMakeToolchain`, maybe we should factorize it (but that can be done later, just focus on getting this right).

Yup, seems beyond a reasonable refactoring threshold:

```
conan/tools/cmake/toolchain/blocks.py :747 col 31| "armv8": "ARM64"}.get(arch)
conan/tools/microsoft/msbuild.py:16 col 23| 'armv8': 'ARM64'}.get(str(arch))
conans/client/build/cmake_flags.py:104 col 27| "armv8": "ARM64"}.get(arch)
conans/client/build/msbuild.py:61 col 79| msvc_arch = {'x86': 'x86', 'x86_64': 'x64', 'armv7': 'ARM', 'armv8': 'ARM64'}
conans/client/build/msbuild.py:136 col 32| 'armv8': 'ARM64'}
conans/test/functional/layout/test_build_system_layout_helpers.py:38 col 46| subfolders_arch = {"armv7": "ARM", "armv8": "ARM64", "x86": None, "x86_64": "x64"}
```

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -12,6 +12,13 @@
 VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
 
 
+def msbuild_arch(arch):
+    return {'x86': 'x86',
```
memsharded:
```suggestion
    return {'x86': 'Win32',
```

I think this is the right one, and the reason the unit test is failing

markferry: :facepalm:

memsharded: You also need to check the CLA signing above, please

conan/tools/microsoft/msbuild.py
```diff
@@ -10,7 +10,7 @@ def msbuild_verbosity_cmd_line_arg(conanfile):
 
 
 def msbuild_arch(arch):
-    return {'x86': 'x86',
+    return {'x86': 'Win32',
```
cognomen, line 13: Bug not manifested, I think, because the only place where this would have been used was defaulted:

[conan/tools/microsoft/layout.py](/conan-io/conan/blob/b6bcdc5f24bb9ab5e4ca8143a42637b1f010d471/conan/tools/microsoft/layout.py#L13):

```python
    if conanfile.settings.arch != "x86":
        arch = msbuild_arch(conanfile.settings.arch)
```

markferry: Bug not manifested, I think, because the only place where this would have been used was defaulted:

[conan/tools/microsoft/layout.py](/conan-io/conan/blob/b6bcdc5f24bb9ab5e4ca8143a42637b1f010d471/conan/tools/microsoft/layout.py#L13):

```python
    if conanfile.settings.arch != "x86":
        arch = msbuild_arch(conanfile.settings.arch)
```

memsharded: I can have a look later to those broken tests, thanks!

markferry: I have access to VS2019 but not VS2017 or VS2015.

Creating an empty C++ project in VS2019 creates `x86` and `x64` configurations by default.

I'm not (normally) a Windows developer so not sure if these defaults changed recently. If they did `msbuild_arch` behaviour would depend on the msbuild version :/

markferry: https://github.com/MicrosoftDocs/visualstudio-docs/issues/5419#issuecomment-637858985:

> "Win32" is used for C++ projects and it means "x86." I think what's confusing here is that you really have both "project-level platforms" and "solution-level platforms" and the project platforms come from the language-specific project systems. C++ projects use "Win32" and "x64", but the solution platforms use 'x86' and 'x64', NOT Win32. But, when you choose 'x86' as the solution configuration, VS selects the Win32 project configuration, so it works. This could be explained in the topic.

And [later](https://github.com/MicrosoftDocs/visualstudio-docs/issues/5419#issuecomment-885815357):

> As best as I can tell, whether a solution has "Win32" or "x86" as a platform depends on what version of Visual Studio it was originally created with. My oldest solutions have "Win32" and newer ones have "x86".

So I'm not sure what's best to do here.

memsharded: Yes, x86 is the configuration created, but internally, for some things, VS is expecting the ``Win32``.
Some tests were breaking, because for the command line it expects the ``x86``, while in the .props file it expects the ``Win32``. 

I have merged develop and added the architectures, without trying to use the ``msbuild_arch`` as common, please check if it works for your case.

markferry: Thanks! Works for me.

Sorry about the kerfuffle over reusing `msbuild_arch`...

markferry:
Testcase to follow.

conan/tools/microsoft/msbuild.py
memsharded: > Thanks! Works for me.

Great! 

> Sorry about the kerfuffle over reusing msbuild_arch...

No prob at all :)

markferry:
> Testcase to follow.

Will pick this up again on Monday

markferry:
> For tests, it is enough to add an `integration` test, don't hesitate to ask for guidance if you need it.

Something like this fd4899246e70d5e3c791fa18da7df343d6d5b9b6?

memsharded:
> > For tests, it is enough to add an `integration` test, don't hesitate to ask for guidance if you need it.
> 
> Something like this [fd48992](https://github.com/conan-io/conan/commit/fd4899246e70d5e3c791fa18da7df343d6d5b9b6)?

Yes, looks good!

memsharded (approved):

markferry:
Looks like it died: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-11505/8/pipeline/37

memsharded:
Yes, don't worry, already merged, it will be in 1.50. 
Thanks very much for contributing to Conan!

PR #11527, opened 2022-06-25
Keep only one flavor of the same libs in `collect_libs()`

### LINKED ISSUE #11282: [bug] tool.collect_libs behaves differently on mac and linux
Mac and Linux both creaete symlinks without package version and libs with package version.

Linux: libmylib.so -> llibmylib.so.X.Y.Z
Mac: libmylib.dylib -> llibmylib.X.Y.Z.dylib

tools.collect_libs with make a os.path.splitext and check if the extension is a valid lib extension. Both symlinks are fine:
 libmylib.so  -> ext = .so (valid and listed)
 libmylib.dylib -> ext = .dylib (valid and listed)
 
The actualy libs however behave defferntly:
 libmylib.so.X.Y.Z -> ext = .Z (not valid is discarded
 libmylib.X.Y.Z.dylib -> ext = .dylib (valid and listed)

Not sure what the corect behaviour should be probaly both should be listed in both cases. But behaving like this is clearly a bug

### LINKED ISSUE #11526: [bug] `collect_libs()` collects all flavors (version, soversion) of the same dylib
  * Operating System+version: macOS Montery
  * Compiler+version: apple-clang 13
  * Conan version: 1.49.0
  * Python version: 3.9.13

- call conan install with some generator on a recipe relying on `tools.collect_libs()` in which version and soversion are installed. For example `conan install proj/9.0.1@ -g CMakeDeps`
- there are multiple references of the same real lib due to various symlinks in lib folder, which are not reduced to the canonical lib name. Indeed, on macOS, version & soversion are added before `.dylib` extension, so `collect_libs()` also lists them.

In lib folder of proj:

```bash
$ ls -la
total 10704
drwxr-xr-x  5 spaceim  staff      160 16 jui 10:34 .
drwxr-xr-x  9 spaceim  staff      288 16 jui 10:34 ..
-rwxr-xr-x  1 spaceim  staff  5480152 16 jui 10:34 libproj.25.9.0.1.dylib
lrwxr-xr-x  1 spaceim  staff       22 16 jui 10:34 libproj.25.dylib -> libproj.25.9.0.1.dylib
lrwxr-xr-x  1 spaceim  staff       16 16 jui 10:34 libproj.dylib -> libproj.25.dylib
```

`tools.collect_libs()` returns `proj`, `proj.25` & `proj.25.9.0.1`, while it could be smart and only returns `proj` by grouping symlinks and real file together, and keeping only the shortest name in each group.

SpaceIm:
closes https://github.com/conan-io/conan/issues/11282
closes https://github.com/conan-io/conan/issues/11526

Changelog: Fix: In `conans.tools.collect_libs` and `conan.tools.files.collect_libs`, avoids to list other flavors (version, soversion) of the same `dylib` on macOS, so that behavior is consistent on Linux & macOS.
Docs: https://github.com/conan-io/docs/pull/2610

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -515,11 +515,7 @@ def collect_libs(conanfile, folder=None):
         name, ext = os.path.splitext(f)
         if ext != ".lib" and name.startswith("lib"):
             name = name[3:]
-        if name in result:
-            conanfile.output.warn("Library '%s' was either already found in a previous "
-                                  "'conanfile.cpp_info.libdirs' folder or appears several "
-                                  "times with a different file extension" % name)
-        else:
+        if name not in result:
             result.append(name)
     result.sort()
     return result
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -494,24 +494,33 @@ def collect_libs(conanfile, folder=None):
     else:
         lib_folders = [os.path.join(conanfile.package_folder, folder)
                        for folder in conanfile.cpp_info.libdirs]
-    result = []
+
+    ref_libs = {}
     for lib_folder in lib_folders:
         if not os.path.exists(lib_folder):
             conanfile.output.warn("Lib folder doesn't exist, can't collect libraries: "
                                   "{0}".format(lib_folder))
             continue
+        # In case of symlinks, only keep shortest file name in the same "group"
         files = os.listdir(lib_folder)
         for f in files:
             name, ext = os.path.splitext(f)
             if ext in (".so", ".lib", ".a", ".dylib", ".bc"):
-                if ext != ".lib" and name.startswith("lib"):
-                    name = name[3:]
-                if name in result:
-                    conanfile.output.warn("Library '%s' was either already found in a previous "
-                                          "'conanfile.cpp_info.libdirs' folder or appears several "
-                                          "times with a different file extension" % name)
-                else:
-                    result.append(name)
+                real_lib = os.path.basename(os.path.realpath(os.path.join(lib_folder, f)))
+                if real_lib not in ref_libs or len(f) < len(ref_libs[real_lib]):
+                    ref_libs[real_lib] = f
+
+    result = []
+    for f in ref_libs.values():
+        name, ext = os.path.splitext(f)
+        if ext != ".lib" and name.startswith("lib"):
+            name = name[3:]
+        if name in result:
+            conanfile.output.warn("Library '%s' was either already found in a previous "
+                                  "'conanfile.cpp_info.libdirs' folder or appears several "
+                                  "times with a different file extension" % name)
+        else:
+            result.append(name)
     result.sort()
     return result
 
--- a/conans/test/unittests/tools/files/collect_lib_test.py
+++ b/conans/test/unittests/tools/files/collect_lib_test.py
@@ -55,6 +55,24 @@ def test_collect_libs():
     result = collect_libs(conanfile)
     assert ["mylib"] == result
 
+    # Keep only the shortest lib name per group of symlinks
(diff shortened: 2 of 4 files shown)


conan/tools/files/files.py
```diff
@@ -494,24 +494,29 @@ def collect_libs(conanfile, folder=None):
...
             if ext in (".so", ".lib", ".a", ".dylib", ".bc"):
-                if ext != ".lib" and name.startswith("lib"):
-                    name = name[3:]
-                if name in result:
-                    conanfile.output.warn("Library '%s' was either already found in a previous "
-                                          "'conanfile.cpp_info.libdirs' folder or appears several "
-                                          "times with a different file extension" % name)
-                else:
-                    result.append(name)
+                real_lib = os.path.basename(os.path.realpath(os.path.join(lib_folder, f)))
```
memsharded, line 509: This is a bit tricky way to follow symlinks and obtain the name, otherwise it seems very redudant. Better put a comment directly about this line to explain it.

conans/test/unittests/tools/files/collect_lib_test.py
```diff
@@ -43,21 +43,6 @@ def test_collect_libs():
...
-    conanfile.folders.set_base_package(temp_folder())
-    conanfile.cpp_info = CppInfo(conanfile.name, "")
-    custom_mylib_path = os.path.join(conanfile.package_folder, "custom_folder", "mylib.lib")
-    lib_mylib_path = os.path.join(conanfile.package_folder, "lib", "mylib.lib")
-    save(custom_mylib_path, "")
-    save(lib_mylib_path, "")
-    conanfile.cpp_info.libdirs = ["lib", "custom_folder"]
-    result = collect_libs(conanfile)
-    assert ["mylib"] == result
-    assert "Library 'mylib' was either already found in a previous "\
```
memsharded, line 57: Why this test is being removed? Maybe it is fine to remove the warning check, but this test should still work and ``result == ["mylib"]``?
Also, it is also that necessary to remove the warning? If ``collect_libs()`` is decided to use one specific name in a group of related ones, isn't it good to show some information about this?

memsharded (changes requested):
I would be less concerned of something breaking if this didn't remove existing tests.
Also, it would be great if it added one or more tests that actually cover the intended use cases with symlinks.

If there is some risk of breaking, then I'd recommend to do it only in the new ``conan.tools`` space, but no in ``conans.client.tools``

conans/test/unittests/tools/files/collect_lib_test.py
SpaceIm: Yes, the first assert is still valid. The second assert fails if I keep the warning, because now there is a first pass grouping the same libs in all folders, either symlinks or real file, and it loses the capability to track duplicated real files (but it can still detect libs with different extensions). Displaying a warning for symlinks is unnecessary noise. I can try to track duplicated real files, but I'm not sure it's worth the effort and overhead.

SpaceIm: I'll add tests for symlinks

conan/tools/files/files.py
SpaceIm: Isn't `os.path.realpath` the most straightforward way to find real file?

conans/test/unittests/tools/files/collect_lib_test.py
memsharded: Good, lets keep the first asset, and I agree, not worth the effort to keep the warning, it can be removed.

SpaceIm:
This test should be skipped on windows.

memsharded (approved):

czoido (approved):

PR #11530, opened 2022-06-26
further testing subsystems

memsharded:
Changelog: Feature: Adding new ``msys2_ucrt64``, ``msys2_mingw64``, ``msys2_clang64`` subsystems in ``os.subsystem`` for Windows.
Docs: Omit

Testing:
- Msys2 subsystems, including ucrt64, clang64, mingw
- Extending runtime checks
- Extending subsystem definitions checks
- Adding Native LLVM/Clang tests


conans/client/conf/__init__.py
```diff
@@ -27,7 +27,7 @@
     # - For building cross compilation tools: Where the cross compiler will run.
     os:
         Windows:
-            subsystem: [None, cygwin, msys, msys2, wsl]
+            subsystem: [None, cygwin, msys, msys2, msys2_ucrt64, msys2_ming32, msys2_ming64, msys2_clang64, wsl]
         WindowsStore:
```
czoido:
```suggestion
            subsystem: [None, cygwin, msys, msys2, msys2_ucrt64, msys2_mingw32, msys2_mingw64, msys2_clang64, wsl]
        WindowsStore:
```

conans/client/conf/__init__.py
```diff
@@ -27,7 +27,7 @@
     # - For building cross compilation tools: Where the cross compiler will run.
     os:
         Windows:
-            subsystem: [None, cygwin, msys, msys2, wsl]
+            subsystem: [None, cygwin, msys, msys2, msys2_ucrt64, msys2_ming32, msys2_ming64, msys2_clang64, wsl]
```
czoido: Not a big deal, but maybe having both msys2 and msys could be a bit confusing? Because looks like the "base" subsystem for msys2 [is called msys](https://www.msys2.org/docs/environments/).
I really don't know if there's people using msys instead of msys2

conans/client/migrations_settings.py
```diff
@@ -3712,4 +3712,155 @@
...
+# which the tools generate code
+os_target: [Windows, Linux, Macos, Android, iOS, watchOS, tvOS, FreeBSD, SunOS, AIX, Arduino, Neutrino]
+arch_target: [x86, x86_64, ppc32be, ppc32, ppc64le, ppc64, armv5el, armv5hf, armv6, armv7, armv7hf, armv7s, armv7k, armv8, armv8_32, armv8.3, sparc, sparcv9, mips, mips64, avr, s390, s390x, asm.js, wasm, sh4le, e2k-v2, e2k-v3, e2k-v4, e2k-v5, e2k-v6, e2k-v7, xtensalx6, xtensalx106, xtensalx7]
+
+# Rest of the settings are "host" settings:
+# - For native building/cross building: Where the library/program will run.
+# - For building cross compilation tools: Where the cross compiler will run.
+os:
+    Windows:
+        subsystem: [None, cygwin, msys, msys2, msys2_ucrt64, msys2_ming32, msys2_ming64, msys2_clang64, wsl]
```
czoido:
```suggestion
        subsystem: [None, cygwin, msys, msys2, msys2_ucrt64, msys2_mingw32, msys2_mingw64, msys2_clang64, wsl]
```

conans/client/conf/__init__.py
lasote: I think `msys` is mostly dead, but we shouldn't get rid of the setting. Maybe in 2.0

memsharded: Yeah, tat ``msys`` intended to be to the previous msys2 version, which is mostly dead, but we probably shouldn't remove it, as we could break. Lets get rid of it in Conan 2.0

czoido (approved):

lasote (approved):

SSE4 (approved):
it's good to see these sub-systems are tested and pass tests.
how about recording a runtime used as a sub-setting? or is it supposed to be a follow-up PR?

franramirez688 (approved):

memsharded:
> how about recording a runtime used as a sub-setting? or is it supposed to be a follow-up PR?

It was removed from this PR, it will be a follow-up PR where those subsettings are actually used.

PR #11553, opened 2022-06-30
Limit usage of list2cmdline in check_output_runner to Windows

lasote:
I have not removed the `list2cmdline` (well, the command is removed but not the escaping with other new function) following the strategy of "I prefer to see it fail and fix later" because the need is now clear and tested and the usage is more limited and verified. In my opinion, it makes sense to keep a function to prepare any call to the command line, because it is indeed very necessary to quote special characters or the system calls will fail.

- More controlled usage
- Replaced `list2cmdline` with a basic tool that should cover 99% of weird command line parsing
- Introduced a test with a real executable reading the parameters. When issues appear, we can add a new assert to the test easily.

Changelog: omit
Docs: omit

### DIFF
--- a/conan/tools/build/__init__.py
+++ b/conan/tools/build/__init__.py
@@ -1,10 +1,10 @@
-import subprocess
 import sys
+from shlex import quote
 
-from conan.tools.build.cpu import build_jobs
-from conan.tools.build.cross_building import cross_building, can_run
 from conan.tools.build.cppstd import check_min_cppstd, valid_min_cppstd, default_cppstd, \
     supported_cppstd
+from conan.tools.build.cpu import build_jobs
+from conan.tools.build.cross_building import cross_building, can_run
 
 
 def use_win_mingw(conanfile):
@@ -19,11 +19,30 @@ def use_win_mingw(conanfile):
     return False
 
 
-def args_to_string(args):
+def cmd_args_to_string(args):
     if not args:
         return ""
-    # FIXME: This is ugly, hardcoding condition how to parse args
     if sys.platform == 'win32':
-        return subprocess.list2cmdline(args)
+        return _windows_cmd_args_to_string(args)
     else:
-        return " ".join("'" + arg.replace("'", r"'\''") + "'" for arg in args)
+        return _unix_cmd_args_to_string(args)
+
+
+def _unix_cmd_args_to_string(args):
+    """Return a shell-escaped string from *split_command*."""
+    return ' '.join(quote(arg) for arg in args)
+
+
+def _windows_cmd_args_to_string(args):
+    # FIXME: This is not managing all the parsing from list2cmdline, but covering simplified cases
+    ret = []
+    for arg in args:
+        # escaped quotes have to be double escaped
+        arg = arg.replace(r'\"', r'\\"')
+        # quotes have to be escaped
+        arg = arg.replace(r'"', r'\"')
+        # if argument have spaces, quote it
+        if ' ' or '\t' in arg:
+            ret.append('"{}"'.format(arg))
+    return " ".join(ret)
+
--- a/conan/tools/build/__init__.py
+++ b/conan/tools/build/__init__.py
@@ -42,7 +42,9 @@ def _windows_cmd_args_to_string(args):
         # quotes have to be escaped
         arg = arg.replace(r'"', r'\"')
         # if argument have spaces, quote it
-        if ' ' or '\t' in arg:
+        if ' ' in arg or '\t' in arg:
             ret.append('"{}"'.format(arg))
+        else:
+            ret.append(arg)
     return " ".join(ret)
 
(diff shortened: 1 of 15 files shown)


conans/test/functional/toolchains/gnu/autotools/test_ios.py
```diff
@@ -74,9 +74,9 @@ def build(self):
     configure_args = conanbuild["configure_args"]
     make_args = conanbuild["make_args"]
     autoreconf_args = conanbuild["autoreconf_args"]
-    assert configure_args == "'--prefix=/' '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' " \
+    assert configure_args == "--prefix=/ '--bindir=${prefix}/bin' '--sbindir=${prefix}/bin' " \
                              "'--libdir=${prefix}/lib' '--includedir=${prefix}/include' " \
                              "'--oldincludedir=${prefix}/include' '--datarootdir=${prefix}/res' " \
-                             "'--host=aarch64-apple-ios' '--build=x86_64-apple-darwin'"
+                             "--host=aarch64-apple-ios --build=x86_64-apple-darwin"
```
lasote, line 80: It is not necessary to put quotes if there is no unsafe characters like the `$` so these are not quoted

conans/test/functional/toolchains/gnu/autotools/test_ios.py
```diff
@@ -74,9 +74,9 @@ def build(self):
...
     assert make_args == ""
-    assert autoreconf_args == "'--force' '--install'"
+    assert autoreconf_args == "--force --install"
```
lasote, line 82: Idem

conan/tools/build/__init__.py
```diff
@@ -19,11 +19,36 @@ def use_win_mingw(conanfile):
...
-        return " ".join("'" + arg.replace("'", r"'\''") + "'" for arg in args)
+        return _unix_cmd_args_to_string(args)
+
+
+def _unix_cmd_args_to_string(args):
+    """Return a shell-escaped string from *split_command*."""
+    return ' '.join(quote(arg) for arg in args)
+
+
+def _windows_cmd_args_to_string(args):
```
lasote, line 36: NOTE: This is doing more or less the same than `list2cmdline` but the original implementation is a little bit more complex (not too much) and covers more cases with "escapes of escapes of escaped quotes". I would say it would be convenient to just copy-paste the python original code.

conans/client/conf/detect_vs.py
```diff
@@ -149,7 +150,8 @@ def vswhere(all_=False, prerelease=False, products=None, requires=None, version=
 
     try:
         from conans.util.runners import check_output_runner
-        output = check_output_runner(arguments).strip()
+        cmd = cmd_args_to_string(arguments)
```
memsharded, line 153: This used to work without it, it started to fail now?

conans/test/functional/util/test_cmd_args_to_string.py
```diff
@@ -0,0 +1,96 @@
...
+import pytest
+
+from conan.tools.build import cmd_args_to_string
+from conans.test.utils.tools import TestClient
+from conans.util.files import chdir
+from conans.util.runners import check_output_runner
+
+
+@pytest.fixture(scope="module")
+def application_folder():
```
memsharded, line 14: Wouldn't it be simpler to feed it into Python ``argparse``?

conan/tools/build/__init__.py
```diff
@@ -19,11 +19,36 @@ def use_win_mingw(conanfile):
...
-        return subprocess.list2cmdline(args)
+        return _windows_cmd_args_to_string(args)
     else:
-        return " ".join("'" + arg.replace("'", r"'\''") + "'" for arg in args)
+        return _unix_cmd_args_to_string(args)
+
+
+def _unix_cmd_args_to_string(args):
+    """Return a shell-escaped string from *split_command*."""
+    return ' '.join(quote(arg) for arg in args)
```
memsharded, line 33: Not ``shlex.join()``?

memsharded (commented):
Looking good

conan/tools/build/__init__.py
lasote: ``New in version 3.8``

conans/client/conf/detect_vs.py
lasote: Look at the removed line at check_output_runner, it was the only point it was receiving a list of arguments, so now I'm doing the conversion (like in the rest of the calls) before calling it: https://github.com/conan-io/conan/pull/11553/files#diff-644cfd55dac44955193dc4beefbfcfa730a68c207e0f2b2e43f60a622fe9c7e0R87-L89

conans/test/functional/util/test_cmd_args_to_string.py
lasote: Simpler but not 100% sure if more accurate.

memsharded (approved):

PR #11556, opened 2022-06-30
Add new layout() test for projects sharing common code in parent folders

memsharded:
Changelog: Feature: Added a new ``self.folders.subproject`` for ``layout()`` to be able to define layouts that goes up to the parent or siblings folders, together with the ``self.folders.root = ".."``.
Docs: https://github.com/conan-io/docs/pull/2662

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -3,7 +3,7 @@ import os
 from conans.errors import ConanException
 
 
-def cmake_layout(conanfile, generator=None, src_folder=".", subfolder=None):
+def cmake_layout(conanfile, generator=None, src_folder="."):
     gen = conanfile.conf.get("tools.cmake.cmaketoolchain:generator", default=generator)
     if gen:
         multi = "Visual" in gen or "Xcode" in gen or "Multi-Config" in gen
@@ -14,13 +14,14 @@ def cmake_layout(conanfile, generator=None, src_folder=".", subfolder=None):
         else:
             multi = False
 
-    conanfile.folders.source = src_folder if not subfolder else os.path.join(subfolder, src_folder)
+    subproject = conanfile.folders.subproject
+    conanfile.folders.source = src_folder if not subproject else os.path.join(subproject, src_folder)
     try:
         build_type = str(conanfile.settings.build_type)
     except ConanException:
         raise ConanException("'build_type' setting not defined, it is necessary for cmake_layout()")
 
-    build_folder = "build" if not subfolder else os.path.join(subfolder, "build")
+    build_folder = "build" if not subproject else os.path.join(subproject, "build")
     custom_conf = get_build_folder_custom_vars(conanfile)
     if custom_conf:
         build_folder = "{}/{}".format(build_folder, custom_conf)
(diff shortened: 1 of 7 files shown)


prince-chrismc (approved):
This works 🚀 

The downside to the custom layout is losing the generator folder being set by `cmake_layout` and it's [handling around the conf and multi](https://github.com/conan-io/conan/commit/005fc53485557b0a570bb71670f2ca9c66082165#diff-dbf5d536275be2e3c8db49e46f4b65b830abcb38becc44a077c7d3271c2e0d82R28-R31)

conan/tools/cmake/layout.py
```diff
@@ -14,13 +14,13 @@ def cmake_layout(conanfile, generator=None, src_folder="."):
         else:
             multi = False
 
-    conanfile.folders.source = src_folder
+    conanfile.folders.source = src_folder if not subfolder else os.path.join(subfolder, src_folder)
```
prince-chrismc: sub folder is above source dir? 😕 Is that from the perspective of the root folder?

Maybe "nested" or "project" not sure whats better, maybe some is enough?

memsharded: Yes, the idea is that ``subfolder`` is from the "root" folder, to get you back to the subproject folder where the conanfile.py is. And then, the ``src_folder`` defines the source folder within that subproject (typically, where the CMakeLists.txt of that subproject is)

prince-chrismc (approved):

lasote (changes requested):
I think understanding why going to `..` to later come back to `pkg` is too much. There is no way a "subfolder" argument will be interpreted and used in the correct way.

conans/test/functional/layout/test_in_subfolder.py
```diff
@@ -54,3 +58,112 @@ def build(self):
...
+        from conan.tools.files import load, copy, save
+        class Pkg(ConanFile):
+            name = "pkg"
+            version = "0.1"
+            settings = "os", "compiler", "build_type", "arch"
+            generators = "CMakeToolchain"
+
+            def layout(self):
+                self.folders.root = ".."
+                cmake_layout(self, subfolder="pkg")
```
memsharded:
```suggestion
            def layout(self):
                self.folders.root = ".."
                self.folders.subproject = "pkg"
                cmake_layout(self)
```

memsharded: Done

memsharded:
I have changed the approach, with a ``self.folders.subproject`` definition, and extended its application to other generators as well. Please re-check.

conans/model/layout.py
```diff
@@ -31,6 +31,10 @@ def __init__(self):
         # Relative location of the project root, if the conanfile is not in that project root, but
         # in a subfolder: e.g: If the conanfile is in a subfolder then self.root = ".."
         self.root = None
+        # The relative location wrt to the project root of the subproject containing the
```
lasote:
```suggestion
        # The relative location with respect to the project root of the subproject containing the
```

We hate acronyms

lasote (approved):

franramirez688 (approved):

prince-chrismc (approved):
👍 subproject is a more clear name

PR #11557, opened 2022-06-30
[POC] MesonToolchain and GNU flags coming from dependencies

### LINKED ISSUE #10683: [question] MesonToolchain + AutotoolsDeps - how to propagate an environment?
in Meson world, most of the dependencies are found via pkg-config (*.pc files).
however, not all the libraries provide .pc files, and in these cases, dependencies found by other means (e.g. environment variables like `SDL2_CONFIG`/`FREETYPE_CONFIG`, or just library name is well known so it's directly linked via -`l`)
for these dependencies, it might use [find_library](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerfind_library) over [dependency](https://mesonbuild.com/Reference-manual_functions.html#dependency) : https://gitlab.gnome.org/GNOME/glib/-/blob/2.71.2/meson.build#L2078
this obviously requires compiler flags for this dependency to be set (`-I`, `-L` and -`l`, most commonly).
`AutotoolsDeps` already can set and generate required environment variables (`CFLAGS`/`LDFLAGS`) that meson could theoretically pick, but it's not fully clear how to propagate them to the `MesonToolchain`, so it saves them to the cross/native file.
studying how `AutotoolsDeps` work in conjunction with `Autotools`:
- `AutotoolsDeps` generates `conanautotoolsdeps.sh` with environment variables
- `conanbuild.sh` somehow sources `conanautotoolsdeps.sh` (for me it's not fully clear how does it know what it needs to source the file)
- `conanfile.run` uses `conanbuild.sh` internally

for `Autotools` it works, as it uses `self.run("configure ...")`, while for `MesonToolchain` it doesn't - as it doesn't run anything during the writing of the cross/native file.

trying the following:
```
        tc = AutotoolsDeps(self)
        tc.generate()
        with tools.environment_append(tc.vars()):
            tc = MesonToolchain(self)
            ...
            tc.generate()
```
doesn't help. it's due to the fact `MesonToolchain` creates its own `VirtualBuildEnv` instance from scratch, and I don't see any possibility of how to inject env there.

### LINKED ISSUE #11397: [bug] AutoToolsDeps sets includes to CPPFLAGS, but MesonToolchain reads only CFLAGS/CXXFLAGS?
related to #10683, coming from https://github.com/conan-io/conan-center-index/pull/11053

in an attempt to migrate `Glib` to `MesonToolchain` I had to put the following workaround:
```
        cppflags = tc.vars().get("CPPFLAGS")
        tc.environment.append('CFLAGS', cppflags)
        tc.environment.append('CXXFLAGS', cppflags)
```
to propagate `CPPFLAGS` into `CFLAGS`/`CXXFLAGS`.
`AutoToolsDeps` sets includedirs to `CPPFLAGS` only: 
https://github.com/conan-io/conan/blob/1.48.2/conan/tools/gnu/autotoolsdeps.py#L46
https://github.com/conan-io/conan/blob/1.48.2/conan/tools/gnu/autotoolsdeps.py#L74
while `MesonToolchain` currently only reads from `CFLAGS`/`CXXFLAGS`:
https://github.com/conan-io/conan/blob/1.48.2/conan/tools/meson/toolchain.py#L154-L157

so they kinda cannot connect to each other as is, which is disappointing. I don't know where shoud be the right fix, in `MesonToolchain` or `AutoToolsDeps`.

  * Operating System+version:
  * Compiler+version:
  * Conan version:
  * Python version:

### LINKED ISSUE #11398: [bug] `MesonToolchain` -LIBPATH is ignored?
to be honest, I currently don't understand it conan client issue or even meson issue (I mean meson-build project iself).

migrating `Glib` to `MesonToolchain` in https://github.com/conan-io/conan-center-index/pull/11053, I had to put the following workaround:
```
        ldflags = tc.vars().get("LDFLAGS")
        ldflags = ldflags.replace("-LIBPATH", "/LIBPATH")
        tc.environment.define('LDFLAGS', ldflags)
```
which appears to be working normally.

so, if I try without a workaround, I get `Library gnuintl found: NO` (while with old `Meson` build helper it's found).
digging into the logs I can see (in `meson-log.txt`):
```
Command line:  cl /LIBPATH:C:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib /LIBPATH:C:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib /LIBPATH:C:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\lib /LIBPATH:C:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\lib /LIBPATH:C:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\lib /LIBPATH:C:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib -IC:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include -IC:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include -IC:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\include -IC:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\include -IC:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\include -IC:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include C:\.conan\96ccea\1\build_subfolder\meson-private\tmpx47m1x8r\testfile.c /FeC:\.conan\96ccea\1\build_subfolder\meson-private\tmpx47m1x8r\output.exe /nologo /showIncludes /utf-8 /MDd -DFFI_BUILDING -DPCRE_STATIC=1 /nologo /showIncludes /utf-8 /Od /Oi- gnuintl.lib /link /nologo /release /nologo 

Code:
 int main(void) { return 0; }

Compiler stdout:
 testfile.c
LINK : fatal error LNK1181: cannot open input file 'gnuintl.lib'

Compiler stderr:
 cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib'
cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib'
cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\lib'
cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\lib'
cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\lib'
cl : Command line warning D9002 : ignoring unknown option '/LIBPATH:C:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib'
```

we can see `-LIBPATH` arguments are passed into command line, but in a wrong order (they have to be passed at the end, after the [/link](https://docs.microsoft.com/en-us/cpp/build/reference/link-pass-options-to-linker?view=msvc-170)):
```
The /link option and its linker options must appear after any file names and CL options.
```
that's probably due to how meson inserts `link_args` into the command line.
in the toolchain (`conan_meson_native.ini`) I see linking flags are set:
```
c_link_args = ['-LIBPATH:C:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib', '-LIBPATH:C:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\lib', '-LIBPATH:C:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib']
cpp_link_args = ['-LIBPATH:C:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib', '-LIBPATH:C:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\lib', '-LIBPATH:C:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\lib', '-LIBPATH:C:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib']
```
I've tried to manually insert the `/link` at the beginning of `c_link_args`/`cpp_link_args`, but it didn't help.
the only thing that reliably fixed it was replacing `-LIBPATH` with `/LIBPATH`, now it gets the correct command line and libintl is found:
```
Command line:  cl -IC:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include -IC:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include -IC:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\include -IC:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\include -IC:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\include -IC:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\include C:\.conan\96ccea\1\build_subfolder\meson-private\tmpclicrq26\testfile.c /FeC:\.conan\96ccea\1\build_subfolder\meson-private\tmpclicrq26\output.exe /nologo /showIncludes /utf-8 /MDd -DFFI_BUILDING -DPCRE_STATIC=1 /nologo /showIncludes /utf-8 /Od /Oi- gnuintl.lib /link /nologo /release /nologo /LIBPATH:C:\Users\sse4\.conan\data\zlib\1.2.12\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib /LIBPATH:C:\Users\sse4\.conan\data\libffi\3.4.2\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib /LIBPATH:C:\Users\sse4\.conan\data\pcre\8.45\_\_\package\7a162084f5550c625a8950e9b1175d1474d2fbab\lib /LIBPATH:C:\Users\sse4\.conan\data\libgettext\0.21\_\_\package\e0c22822cdf05b624135e1696ae5cb784a23aeb3\lib /LIBPATH:C:\Users\sse4\.conan\data\bzip2\1.0.8\_\_\package\589a23dff5fdb23a7fb851223eb766480ead0a9a\lib /LIBPATH:C:\Users\sse4\.conan\data\libiconv\1.16\_\_\package\d057732059ea44a47760900cb5e4855d2bea8714\lib 

Code:
 int main(void) { return 0; }

Compiler stdout:
 testfile.c

Compiler stderr:
 
Library gnuintl found: YES
```

/cc @ericLemanissier 

  * Operating System+version:
  * Compiler+version:
  * Conan version:
  * Python version:

franramirez688:
Changelog: Feature: Added `MesonDeps` as a new generator to create an extra file with all the GNU flags from all the dependencies. `Meson` build-helper will add that one (if exists) to complement the created one by `MesonToolchain`.
Docs: https://github.com/conan-io/docs/pull/2654
Closes: https://github.com/conan-io/conan/issues/10683 https://github.com/conan-io/conan/issues/11397 https://github.com/conan-io/conan/issues/11398

Motivation: Dependencies which are defining only GNU flags, MesonToolchain will never get anything from their build environments because it needs those flags to initialize its own instance, so right now, it's possible to work-around it but it's a bit hard. This PR is proposing that alternative but, for sure, it's open to being discussed.

### DIFF
--- a/conan/tools/gnu/autotoolsdeps.py
+++ b/conan/tools/gnu/autotoolsdeps.py
@@ -4,79 +4,66 @@ from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
 from conans.model.new_build_info import NewCppInfo
 
 
-def _get_cpp_info(deps):
-    ret = NewCppInfo()
-    for dep in deps:
-        dep_cppinfo = dep.cpp_info.aggregated_components()
-        # In case we have components, aggregate them, we do not support isolated
-        # "targets" with autotools
-        ret.merge(dep_cppinfo)
-    return ret
-
-
-def _rpaths_flags(deps):
-    flags = []
-    for dep in deps:
-        flags.extend(["-Wl,-rpath -Wl,{}".format(libdir) for libdir in dep.cpp_info.libdirs
-                      if dep.options.get_safe("shared", False)])
-    return flags
-
-
-def ordered_deps(conanfile):
-    deps = conanfile.dependencies.host.topological_sort
-    return[dep for dep in reversed(deps.values())]
-
-
-def get_gnu_deps_flags(conanfile):
-    """
-    Given a ConanFile object, this function returns all the GNU flags from all the
-    dependencies.
-
-    :param conanfile: The current recipe object. Always use ``self``.
-    :return: ``tuple`` of all the GNU flags.
-    """
-    deps = ordered_deps(conanfile)
-    flags = GnuDepsFlags(conanfile, _get_cpp_info(deps))
-
-    # cpp_flags
-    cpp_flags = []
-    cpp_flags.extend(flags.include_paths)
-    cpp_flags.extend(flags.defines)
-
-    # Ldflags
-    ldflags = flags.sharedlinkflags
-    ldflags.extend(flags.exelinkflags)
-    ldflags.extend(flags.frameworks)
-    ldflags.extend(flags.framework_paths)
-    ldflags.extend(flags.lib_paths)
-
-    # set the rpath in Macos so that the library are found in the configure step
-    if conanfile.settings.get_safe("os") == "Macos":
-        ldflags.extend(_rpaths_flags(deps))
-
-    # libs
-    libs = flags.libs
-    libs.extend(flags.system_libs)
-
-    # cflags
-    cflags = flags.cflags
-    cxxflags = flags.cxxflags
-    return cflags, cxxflags, cpp_flags, libs, ldflags
-
-
 class AutotoolsDeps:
     def __init__(self, conanfile):
         self._conanfile = conanfile
         self._environment = None
+        self._ordered_deps = []
         check_using_build_profile(self._conanfile)
+
+    @property
+    def ordered_deps(self):
+        if not self._ordered_deps:
+            deps = self._conanfile.dependencies.host.topological_sort
+            self._ordered_deps = [dep for dep in reversed(deps.values())]
+        return self._ordered_deps
+
+    def _get_cpp_info(self):
+        ret = NewCppInfo()
+        for dep in self.ordered_deps:
+            dep_cppinfo = dep.cpp_info.aggregated_components()
+            # In case we have components, aggregate them, we do not support isolated
+            # "targets" with autotools
+            ret.merge(dep_cppinfo)
+        return ret
+
+    def _rpaths_flags(self):
+        flags = []
+        for dep in self.ordered_deps:
+            flags.extend(["-Wl,-rpath -Wl,{}".format(libdir) for libdir in dep.cpp_info.libdirs
+                          if dep.options.get_safe("shared", False)])
+        return flags
 
     @property
     def environment(self):
         # TODO: Seems we want to make this uniform, equal to other generators
         if self._environment is None:
-            # Get all the GNU flags from all the dependencies
-            cflags, cxxflags, cpp_flags, libs, ldflags = get_gnu_deps_flags(self._conanfile)
-            # Create the environment
+            flags = GnuDepsFlags(self._conanfile, self._get_cpp_info())
+
+            # cpp_flags
+            cpp_flags = []
+            cpp_flags.extend(flags.include_paths)
+            cpp_flags.extend(flags.defines)
+
+            # Ldflags
+            ldflags = flags.sharedlinkflags
+            ldflags.extend(flags.exelinkflags)
+            ldflags.extend(flags.frameworks)
+            ldflags.extend(flags.framework_paths)
+            ldflags.extend(flags.lib_paths)
+
+            ## set the rpath in Macos so that the library are found in the configure step
+            if self._conanfile.settings.get_safe("os") == "Macos":
+                ldflags.extend(self._rpaths_flags())
+
+            # libs
+            libs = flags.libs
+            libs.extend(flags.system_libs)
+
+            # cflags
+            cflags = flags.cflags
+            cxxflags = flags.cxxflags
+
             env = Environment()
             env.append("CPPFLAGS", cpp_flags)
             env.append("LIBS", libs)
@@ -89,5 +76,5 @@ class AutotoolsDeps:
     def vars(self, scope="build"):
         return self.environment.vars(self._conanfile, scope=scope)
 
-    def generate(self, scope="build"):
+    def generate(self,  scope="build"):
         self.vars(scope).save_script("conanautotoolsdeps")
--- a/conan/tools/gnu/autotoolsdeps.py
+++ b/conan/tools/gnu/autotoolsdeps.py
@@ -4,66 +4,79 @@ from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
 from conans.model.new_build_info import NewCppInfo
 
 
+def _get_cpp_info(deps):
+    ret = NewCppInfo()
+    for dep in deps:
+        dep_cppinfo = dep.cpp_info.aggregated_components()
+        # In case we have components, aggregate them, we do not support isolated
+        # "targets" with autotools
+        ret.merge(dep_cppinfo)
+    return ret
+
+
+def _rpaths_flags(deps):
+    flags = []
+    for dep in deps:
+        flags.extend(["-Wl,-rpath -Wl,{}".format(libdir) for libdir in dep.cpp_info.libdirs
+                      if dep.options.get_safe("shared", False)])
+    return flags
+
+
+def ordered_deps(conanfile):
+    deps = conanfile.dependencies.host.topological_sort
+    return[dep for dep in reversed(deps.values())]
+
+
+def get_gnu_deps_flags(conanfile):
+    """
+    Given a ConanFile object, this function returns all the GNU flags from all the
+    dependencies.
+
+    :param conanfile: The current recipe object. Always use ``self``.
+    :return: ``tuple`` of all the GNU flags.
+    """
+    deps = ordered_deps(conanfile)
+    flags = GnuDepsFlags(conanfile, _get_cpp_info(deps))
+
+    # cpp_flags
+    cpp_flags = []
+    cpp_flags.extend(flags.include_paths)
+    cpp_flags.extend(flags.defines)
+
+    # Ldflags
+    ldflags = flags.sharedlinkflags
+    ldflags.extend(flags.exelinkflags)
+    ldflags.extend(flags.frameworks)
+    ldflags.extend(flags.framework_paths)
+    ldflags.extend(flags.lib_paths)
+
+    # set the rpath in Macos so that the library are found in the configure step
+    if conanfile.settings.get_safe("os") == "Macos":
+        ldflags.extend(_rpaths_flags(deps))
+
+    # libs
+    libs = flags.libs
+    libs.extend(flags.system_libs)
+
+    # cflags
+    cflags = flags.cflags
+    cxxflags = flags.cxxflags
+    return cflags, cxxflags, cpp_flags, libs, ldflags
+
+
 class AutotoolsDeps:
     def __init__(self, conanfile):
         self._conanfile = conanfile
         self._environment = None
-        self._ordered_deps = []
         check_using_build_profile(self._conanfile)
-
-    @property
-    def ordered_deps(self):
-        if not self._ordered_deps:
-            deps = self._conanfile.dependencies.host.topological_sort
-            self._ordered_deps = [dep for dep in reversed(deps.values())]
-        return self._ordered_deps
-
-    def _get_cpp_info(self):
-        ret = NewCppInfo()
-        for dep in self.ordered_deps:
-            dep_cppinfo = dep.cpp_info.aggregated_components()
-            # In case we have components, aggregate them, we do not support isolated
-            # "targets" with autotools
-            ret.merge(dep_cppinfo)
-        return ret
-
-    def _rpaths_flags(self):
-        flags = []
-        for dep in self.ordered_deps:
-            flags.extend(["-Wl,-rpath -Wl,{}".format(libdir) for libdir in dep.cpp_info.libdirs
-                          if dep.options.get_safe("shared", False)])
-        return flags
 
     @property
     def environment(self):
         # TODO: Seems we want to make this uniform, equal to other generators
         if self._environment is None:
-            flags = GnuDepsFlags(self._conanfile, self._get_cpp_info())
-
-            # cpp_flags
-            cpp_flags = []
-            cpp_flags.extend(flags.include_paths)
-            cpp_flags.extend(flags.defines)
-
-            # Ldflags
-            ldflags = flags.sharedlinkflags
-            ldflags.extend(flags.exelinkflags)
-            ldflags.extend(flags.frameworks)
-            ldflags.extend(flags.framework_paths)
-            ldflags.extend(flags.lib_paths)
-
-            ## set the rpath in Macos so that the library are found in the configure step
-            if self._conanfile.settings.get_safe("os") == "Macos":
-                ldflags.extend(self._rpaths_flags())
-
-            # libs
-            libs = flags.libs
-            libs.extend(flags.system_libs)
-
-            # cflags
-            cflags = flags.cflags
-            cxxflags = flags.cxxflags
-
+            # Get all the GNU flags from all the dependencies
+            cflags, cxxflags, cpp_flags, libs, ldflags = get_gnu_deps_flags(self._conanfile)
+            # Create the environment
             env = Environment()
             env.append("CPPFLAGS", cpp_flags)
             env.append("LIBS", libs)
@@ -76,5 +89,5 @@ class AutotoolsDeps:
     def vars(self, scope="build"):
         return self.environment.vars(self._conanfile, scope=scope)
 
-    def generate(self,  scope="build"):
+    def generate(self, scope="build"):
         self.vars(scope).save_script("conanautotoolsdeps")
--- a/conan/tools/gnu/gnudeps_flags.py
+++ b/conan/tools/gnu/gnudeps_flags.py
@@ -48,7 +48,6 @@ class GnuDepsFlags(object):
         os_ = self._conanfile.settings.get_safe("os")
         if not frameworks or not is_apple_os(os_):
             return []
-        # FIXME: Missing support for subsystems
         compiler = self._conanfile.settings.get_safe("compiler")
         if str(compiler) not in self._GCC_LIKE:
             return []
@@ -60,15 +59,14 @@ class GnuDepsFlags(object):
     def _format_include_paths(self, include_paths):
         if not include_paths:
             return []
-        # FIXME: Missing support for subsystems
-        return ["-I%s" % (self._adjust_path(include_path))
+        pattern = "/I%s" if is_msvc(self._conanfile) else "-I%s"
+        return [pattern % (self._adjust_path(include_path))
                 for include_path in include_paths if include_path]
 
     def _format_library_paths(self, library_paths):
         if not library_paths:
             return []
-        # FIXME: Missing support for subsystems
-        pattern = "-LIBPATH:%s" if is_msvc(self._conanfile) else "-L%s"
+        pattern = "/LIBPATH:%s" if is_msvc(self._conanfile) else "-L%s"
         return [pattern % self._adjust_path(library_path)
                 for library_path in library_paths if library_path]
 
--- a/conan/tools/meson/meson.py
+++ b/conan/tools/meson/meson.py
@@ -1,7 +1,8 @@
 import os
 
 from conan.tools.build import build_jobs
-from conan.tools.meson import MesonToolchain
+from conan.tools.meson.toolchain import MesonToolchain
+from conan.tools.meson.mesondeps import MesonDeps
 
 
 class Meson(object):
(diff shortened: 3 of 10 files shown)


conans/test/functional/toolchains/meson/test_meson_and_gnu_deps_flags.py
```diff
@@ -0,0 +1,65 @@
...
+import pytest
+
+from conans.test.assets.sources import gen_function_cpp
+from conans.test.functional.toolchains.meson._base import TestMesonBase
+from conans.test.utils.tools import TestClient
+
+
+class TestMesonToolchainAndGnuFlags(TestMesonBase):
+
+    @pytest.mark.skipif(platform.system() == "Windows", reason="Unix only")
```
SSE4: why not Windows? if so, we cannot be sure it fixes LIBPATH problem

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -4,66 +4,79 @@
 from conans.model.new_build_info import NewCppInfo
 
 
+def _get_cpp_info(deps):
+    ret = NewCppInfo()
+    for dep in deps:
+        dep_cppinfo = dep.cpp_info.aggregated_components()
+        # In case we have components, aggregate them, we do not support isolated
```
SSE4: what exactly does "isolated" target mean?

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -4,66 +4,79 @@
...
+                      if dep.options.get_safe("shared", False)])
+    return flags
+
+
+def ordered_deps(conanfile):
+    deps = conanfile.dependencies.host.topological_sort
+    return[dep for dep in reversed(deps.values())]
+
+
+def get_gnu_deps_flags(conanfile):
```
SSE4: a free function probably doen't belong to `AutoToolsDeps` generator, it should be somewhere else in `conan.tools`?

conan/tools/gnu/autotoolsdeps.py
franramirez688: I've only used the functions that were already there. There's nothing new here.

conan/tools/gnu/autotoolsdeps.py
franramirez688: The logic was already here, so I've kept it here as well. I don't think that it has to be in Conan tools but I'm not saying that this is the better place or even if we want it. By the record, the final import is `from conan.tools.gnu import get_gnu_deps_flags` so I guess it makes sense... Let's discuss! 😄

conans/test/functional/toolchains/meson/test_meson_and_gnu_deps_flags.py
```diff
@@ -0,0 +1,61 @@
...
+        class App(ConanFile):
+            settings = "os", "arch", "compiler", "build_type"
+            requires = "hello/0.1"
+
+            def layout(self):
+                self.folders.build = "build"
+
+            def generate(self):
+                # Get GNU flags from all the dependencies
+                cflags, cxxflags, cpp_flags, libs, ldflags = get_gnu_deps_flags(self)
```
SSE4: perhaps, we need some universal interface (object) that could be returned from or accepted by various tools (build helpers).
how about some `NamedTuple` or struct?

conan/tools/gnu/autotoolsdeps.py
```diff
@@ -4,66 +4,79 @@
...
+
+    # Ldflags
+    ldflags = flags.sharedlinkflags
+    ldflags.extend(flags.exelinkflags)
+    ldflags.extend(flags.frameworks)
+    ldflags.extend(flags.framework_paths)
+    ldflags.extend(flags.lib_paths)
+
+    # set the rpath in Macos so that the library are found in the configure step
+    if conanfile.settings.get_safe("os") == "Macos":
```
SSE4: it's weird why is it only Macos?

lasote (commented):
Not sure about the toolchain being populated with the dependencies.
I need some clarification because I've read the issues associated. Still, I'm unsure about what would happen if the workarounds that were introduced [here](https://github.com/conan-io/conan/issues/11398#issuecomment-1147381047) are moved to a `MesonDeps` generator that is based on the `AutotoolsDeps`? I think it should work right? Why introduce in the toolchain the information about the deps?  The `Meson` tool, when run, will load first the launcher for the environment. am I missing something?

franramirez688:
> moved to a MesonDeps generator that is based on the AutotoolsDeps?

Yeah, I think I'm going to try a `MesonDeps` instead of this. I think it's going to be much cleaner.
Thanks for the feedback!

SSE4:
> Yeah, I think I'm going to try a `MesonDeps` instead of this. I think it's going to be much cleaner. Thanks for the feedback!

I can't find anything about `MesonDeps`, neither in the codebase, nor in the documentation. do you have a link?

franramirez688:
> I can't find anything about `MesonDeps`, neither in the codebase, nor in the documentation. do you have a link?

Sure you cannot because that's a new generator. It's not created yet. I'm going to study if it's feasible or not and then, update the PR.

conan/tools/gnu/gnudeps_flags.py
```diff
@@ -60,15 +59,14 @@ def _format_frameworks(self, frameworks, is_path=False):
...
+        pattern = "/I%s" if is_msvc(self._conanfile) else "-I%s"
+        return [pattern % (self._adjust_path(include_path))
                 for include_path in include_paths if include_path]
 
     def _format_library_paths(self, library_paths):
         if not library_paths:
             return []
-        # FIXME: Missing support for subsystems
-        pattern = "-LIBPATH:%s" if is_msvc(self._conanfile) else "-L%s"
+        pattern = "/LIBPATH:%s" if is_msvc(self._conanfile) else "-L%s"
```
franramirez688, line 69: It's true that these ones were a bug if the compiler was MSVC:

* https://docs.microsoft.com/en-us/cpp/build/reference/i-additional-include-directories?view=msvc-170
* https://docs.microsoft.com/en-us/cpp/build/reference/libpath-additional-libpath?view=msvc-170

conan/tools/meson/mesondeps.py
```diff
@@ -0,0 +1,105 @@
...
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._ordered_deps = []
+        # constants
+        self.c_args = []
+        self.c_link_args = []
+        self.cpp_args = []
+        self.cpp_link_args = []
+
+    # TODO: Add all this logic to GnuDepsFlags? Distinguish between GnuFlags and GnuDepsFlags?
```
franramirez688, line 33: If this POC is accepted, then I'll propose a refactor of the actual `GnuDepsFlags`.

* GnuDepsFlags -> GnuFlags
* (_new_) GnuDepsFlags getting the GNU flags for all the dependencies

conan/tools/meson/meson.py
```diff
@@ -15,10 +16,18 @@ def configure(self, reconfigure=False):
...
+        has_deps_flags = os.path.exists(deps_flags)
+
         if os.path.exists(cross):
             cmd += ' --cross-file "{}"'.format(cross)
+            if has_deps_flags:
+                cmd += ' --cross-file "{}"'.format(deps_flags)
         else:
             cmd += ' --native-file "{}"'.format(native)
+            if has_deps_flags:
+                cmd += ' --native-file "{}"'.format(deps_flags)
```
franramirez688, line 29: https://mesonbuild.com/Machine-files.html#loading-multiple-machine-files

SSE4: but it overrides, isn't it?

conan/tools/gnu/gnudeps_flags.py
SSE4: what exactly was a bug?

franramirez688: @SSE4 Now, the Meson test is passing because of this change. Anyway, as you already know, the Meson bug is still there.

franramirez688: It was the `-LIBPATH:` as valid compile options, wasn't it?

conan/tools/meson/mesondeps.py
SSE4: it might be too mind blowing (for users), let's not mulltiply entities and use Occam's razor
e.g. we already have too many disputes on what `cmake` generator to use everywhere, and it's hard enough to explain

conan/tools/gnu/gnudeps_flags.py
SSE4: yes. just in case, for `link.exe` both shoud be legal:
```
On the command line, an option consists of an option specifier, either a dash (-) or a forward slash (/), followed by the name of the option. 
```

franramirez688: Anyway, if they're valid both of them, then, let's use the forward slash. What do you think?

SSE4: yes, I have no objections. I'd prefer the way it works with meson.

conan/tools/meson/meson.py
franramirez688: No, it's not. They are being added as an extended part of the MesonToolchain:

```
    c_args = {{c_args}} + preprocessor_definitions + c_args
    c_link_args = {{c_link_args}} + c_link_args
    cpp_args = {{cpp_args}} + preprocessor_definitions + cpp_args
    cpp_link_args = {{cpp_link_args}} + cpp_link_args
```

So, the only part that is being overridden, it's the first definition of the constants:
```
    [constants]
    ........
    c_args = []
    c_link_args = []
    cpp_args = []
    cpp_link_args = []
```

conan/tools/meson/mesondeps.py
franramirez688: I mean. I'm not thinking of a new generator. GnuDepsFlags is mostly used as an internal helper class and, apart from that, its name is saying `Deps` but it's not managing any deps, so, what I'm saying is to do this refactor for internal purpose. Users should not be affected at all.

SSE4 (approved):

conan/tools/meson/mesondeps.py
memsharded: GnuDepsFlags will not be exposed to users, it is internal

conan/tools/meson/meson.py
memsharded: I am not sure this won't be breaking. The values from multiple machine files override each other, only last file wins. If MesonDeps adds some flags, and MesonToolchain too, they will remove the previously defined one. I don't think the idiom ``c_args = {{c_args}} + preprocessor_definitions + c_args`` will work. Am I misunderstanding something?
Most likely a test that covers this is necessary.

memsharded (changes requested):

conan/tools/meson/meson.py
franramirez688: Yes, this is not working so. AutotoolDeps, as MesonDeps, is getting the flags from all the dependencies altogether so I'm creating only one file, the `conan_meson_deps_flags.ini`. It's not created a file per dependency.

franramirez688: And, the test is already there for sure. That test is not passing without those deps flags. I can try to debug something from the log or try to print anything else.

memsharded: It is not a problem of one file per dependency. It will be the ``conan_meson_deps_flags.ini``, being overriden by the ``conan_meston_toolchain.ini`` file values for those cppflags, etc.

SSE4: IMO this complicates local flow significantly. now, if one wants to replicate conan buid exactly, will have to type something like `--native-file conan_meson_deps_flags1.ini --native-file conan_meson_deps_fags2.ini ... --native-file conan_meson_native.ini`, and it has to be in the correct order (besides the fact it's a long command line). I would just write everything into `conan_meson_native.ini` to simplify the workflow and reduce a chance of human error.

franramirez688: No, you'll only have a single file for all the deps: `--native-file conan_meson_deps_flags.ini`, and that's it. Have a look at the last test that I've added.

conan/tools/gnu/gnudeps_flags.py
```diff
@@ -60,15 +59,14 @@ def _format_frameworks(self, frameworks, is_path=False):
     def _format_include_paths(self, include_paths):
         if not include_paths:
             return []
-        # FIXME: Missing support for subsystems
-        return ["-I%s" % (self._adjust_path(include_path))
+        pattern = "/I%s" if is_msvc(self._conanfile) else "-I%s"
```
memsharded, line 62: Why? In theory VS also admits ``-I`` without problems. Maybe it is a bit more compatible if this is used by some other build tool, like autotools + cl.exe?

memsharded (approved):

conan/tools/gnu/gnudeps_flags.py
franramirez688: Yes, it admits both but Meson is failing because of an internal bug on its side. So this change is making life easier for Meson users and works with VS.

conan/tools/gnu/autotoolsdeps.py
SSE4: it doesn't answer my question...

conan/tools/gnu/gnudeps_flags.py
SSE4: if it's more compatible with tools (e.g. meson), let's switch

lasote (approved):

PR #11560, opened 2022-06-30
Add alwayslink attribute to bazel static library import

### LINKED ISSUE #11559: [bug] bazel generator not usable for packages with multiple libs, like openssl
### Environment Details 
  * Operating System+version: All
  * Compiler+version: Not relevant
  * Conan version: All
  * Python version: All

### Steps to reproduce 

Use the bazel generator and use libcurl on Linux in your project,  so it depends on openssl.
libcurl will not be usable in your bazel project, since libcrypto.a and libssl.a appear in the wrong order for the linker

## Solution

To make the order work, the `alwayslink = True` attribute should be set on the cc_import libs

So this:
```
cc_import(
    name = "ssl_precompiled",
    static_library = "lib/libssl.a",
)

cc_import(
    name = "crypto_precompiled",
    static_library = "lib/libcrypto.a",
)
```
should look like that 

```
cc_import(
    name = "ssl_precompiled",
    static_library = "lib/libssl.a",
    alwayslink = True,
)

cc_import(
    name = "crypto_precompiled",
    static_library = "lib/libcrypto.a",
    alwayslink = True,
)
```

in the generated BUILD file for openssl

we can also discuss if it would not make sense to set `visibility = ["//visibility:public"],` on the cc_import part, since there are use cases where users use only libcrypto , but not libssl from openssl, but that might be a different issue

a4z:
Fixes #11559 , if this solution is taken.

There is no test currently for the static library only version
Test exists only for the case when shared_with_interface_libs contains
elements.

Changelog: omit
Docs: omit

Before adding something to the doc, we should discuss in #11559 if this is a viable solution

### DIFF
--- a/conans/test/unittests/tools/google/test_bazeldeps.py
+++ b/conans/test/unittests/tools/google/test_bazeldeps.py
@@ -220,7 +220,7 @@ cc_library(
     includes=["include"],
     visibility=["//visibility:public"],
     deps = [
-        #donotsort
+        # do not sort
         ":lib1_precompiled",
     ],
 )
--- a/conans/test/unittests/tools/google/test_bazeldeps.py
+++ b/conans/test/unittests/tools/google/test_bazeldeps.py
@@ -58,7 +58,7 @@ def test_bazeldeps_dependency_buildfiles():
                 assert 'linkopts = ["/DEFAULTLIB:system_lib1"]' in dependency_content
             else:
                 assert 'linkopts = ["-lsystem_lib1"],' in dependency_content
-            assert 'deps = [\n    \n    ":lib1_precompiled",' in dependency_content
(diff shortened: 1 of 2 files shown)


conans/test/unittests/tools/google/test_bazeldeps.py
```diff
@@ -220,6 +220,7 @@ def test_bazeldeps_shared_library_interface_buildfiles():
     includes=["include"],
     visibility=["//visibility:public"],
     deps = [
+        #donotsort
```
a4z: Is this on purpose missing all spaces in the comment ?

czoido: Oh, thanks, no, it's a mistake. I'll fix that

czoido:
```suggestion
        # do not sort
```

czoido (approved):

PR #11564, opened 2022-07-02
Bugfix: Follow SONAME to generate BazelDeps correctly

tonyseek:
Changelog: (Feature | Fix | Bugfix):

- Bugfix: `BazelDeps` generates incorrect `cc_import` when the shared library is versioned by `SONAME`.
        **Description:** If A and B are both shared libraries, and A links to B, the linker assigns the `NEEDED` tag of A according to the `SONAME` tag of B. The file name of B was chosen if and only if the `SONAME` tag of B does not exist. Because of this rule, the `BazelDeps` should generate `cc_import` directive of shared libraries from their `SONAME` tags instead of thier file names. Without that, we meet library loading errors in runtime because the Bazel don't know how to put the file (perhaps a symlink) named by its `SONAME` tag into the cache directory.</sup>

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -3,6 +3,8 @@ import platform
 import textwrap
 
 from jinja2 import Template
+from elftools.elf.elffile import ELFFile
+from elftools.common.exceptions import ELFError
 
 from conan.tools._check_build_profile import check_using_build_profile
 from conans.errors import ConanException
@@ -162,6 +164,7 @@ class BazelDeps(object):
                     _relativize_path(lib_path, package_folder),
                     _relativize_path(dll_path, package_folder))
             elif lib_path:
+                lib_path = self._follow_soname_of_shared_lib_if_possible(lib_path)
                 libs[lib] = _relativize_path(lib_path, package_folder)
 
         context = {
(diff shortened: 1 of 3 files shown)


memsharded (changes requested):
I am sorry, but adding ``pyelftools`` as a new dependency for solving this issue seems too much.
The maintenance burden of each transitive dependency is important, so we will only add third parties when strictly necessary. Conan 2.0 has even removed some dependencies.

Conan 2.0 has added some ``property`` to let package creator specify SONAME, and consumers can read that property. The general approach for Conan is declarative, packages should define what they have, instead of relying on inspecting binaries.

tonyseek:
Hi @memsharded, Thank you for your review.

> The maintenance burden of each transitive dependency is important

I can realize and understand this point. There are ways more lightweight to inspect ELF files.

> The general approach for Conan is declarative, packages should define what they have, instead of relying on inspecting binaries.

I can agree with this approach truly. But there are most of recipes they have no `SONAME` property during this progress. It fails users of `BazelDeps` until all recipes have done their revision.

I'm thinking it may be possible to inspect binaries as a transitional way, before all SONAME-agnostic recipes become SONAME-aware, to make `BazelDeps` work on Linux as soon as possible. When a recipe declare its SONAME property explicitly, it stops to guess what is the SONAME and respect the declarted SONAME. Once all recipes declare their SONAME property (when they need of couse), the transitional way could be removed safely.

If you would consider to accept this direction, I could revise the implementation of ELF inspecting, to avoid from introducing third-party dependencies.

PR #11569, opened 2022-07-04
Add ability to reference files in the local filesystem with file:// s…

### LINKED ISSUE #8846: [feature] Support local paths in conans.tools.get
Is it possible to allow using file:// URLs as input parameter for tools.get? This might be useful for testing with local archives by changing only conandata.yml

jcar87:
Changelog: Feature: Add ability to download files in the local filesystem using `file://` URIs.
Docs: https://github.com/conan-io/docs/pull/2635
Close https://github.com/conan-io/conan/issues/8846

## Description
It has been requested be able to download local files with `tools.download()` and `tools.get()` (https://github.com/conan-io/conan/issues/8846 and https://github.com/conan-io/conan-center-index/issues/10623).

This draft PR gives `conan.tools.files.download()` (which is used by `.get()`) the ability to resolve URLs that start with `file://` to files in the local filesystem. In that scenario, the URL provided will be converted to a filesystem path, and `shutil.copy()` will be used to copy the file into the destination.

The underling implementation uses `urllib` functionality to perform this conversion, rather than any internal logic. It should be noted that while the `file://` URI scheme is specified by [RFC 8089](https://datatracker.ietf.org/doc/html/rfc8089), different platforms, applications and implementations may behave slightly differently (see discussion [here](https://discuss.python.org/t/file-uris-in-python/15600)).

Note that it is possible for a `file://` uri to point to resources on a remote system - this implementation makes an assumption that the path returned by the `urllib` library can be accessed without any additional handling.  
On windows this may not cover all cases, see: https://docs.microsoft.com/en-us/dotnet/api/system.uri.isunc?view=net-6.0 

Note that with this proposed implementation the downloads cache is skipped - under the assumption that the files to retrieve are already in the local filesystem. This may not always be true, since it's possible to locally mount network resources - happy to  try and look into making this interact with the downloads cache if necessary!

`pyfakefs` allows mocking a filesystem for python functions


conan/tools/files/files.py
```diff
@@ -164,13 +165,21 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
...
 
     def _download_file(file_url):
-        # The download cache is only used if a checksum is provided, otherwise, a normal download
-        run_downloader(requester=requester, output=out, verify=verify, download_cache=download_cache,
-                       user_download=True, url=file_url,
-                       file_path=filename, retry=retry, retry_wait=retry_wait, overwrite=overwrite,
-                       auth=auth, headers=headers, md5=md5, sha1=sha1, sha256=sha256)
+        if file_url.startswith('file:'):
+            filepath = _path_from_file_uri(file_url)
+            shutil.copyfile(filepath, filename)
```
jcar87: I still wonder if this is the best place to do it - we'd have to ignore `retry`, `retry_wait`, `auth` and `headers`, but still have to check the checksums and overwrite! (working on it!)

The best alternative would be: `conans/client/downloaders/file_downloader.py`

conans/requirements_dev.txt
```diff
@@ -1,6 +1,7 @@
 pytest>=6.1.1, <7.0.0; python_version > '3.0'
 pytest>=4.6.11; python_version < '3.0'
 pytest-xdist  # To launch in N cores with pytest -n
+pyfakefs>=4.5.6
```
memsharded: I'd say definitely not worth the complexities and problems for using more third parties, for this very specific use case.

memsharded: Sorry, I didn't realize it was ``requirements_dev.txt``. No problem with this one, if it really helps for the test (if there is an easy way to do the test without it, better, of course)

jcar87: Yup, just used for testing!

I exposes an in-memory filesystem to tests such that python functions that operate on the filesystem read from this instead. 

For the purposes of this test, writing a temp file on the actual file system and reading back is also an option. I wanted to a be a little more explicit making sure that paths like `file:///C:/path/to/file.txt` are resolved properly on Windows - but I think if we used the existing `temp_folder()` functionality in the test, the paths would be implicitly platform-appropriate anyway, so we would achieve the same testing. Let me know if this is preferred!

memsharded: I'd say yes, we already do a ton of ``temp_folder()`` in our tests suite, and I think this could perfectly work, please give it a try.

conans/client/downloaders/local_file_downloader.py
```diff
@@ -0,0 +1,29 @@
...
+
+class LocalFileDownloader(object):
+
+    def __init__(self, output):
+        self._output = output
+
+    def download(self, url, file_path, md5=None, sha1=None, sha256=None, **kwargs):
+
+        file_origin = self._path_from_file_uri(url)
+        copyfile(file_origin, file_path)
```
jcar87: Note here: `shutil.copyfile()` will overwrite the file if it already exists - I've not added any logic to handle that case. From what I can see, in the newer implementation we are defaulting to `overwrite=True` always, and the `overwrite` argument is no longer exposed.
See here: https://github.com/conan-io/conan/blob/develop/conan/tools/files/files.py#L151

But I can see the file downloader does interpret this flag:
https://github.com/conan-io/conan/blob/develop/conans/client/downloaders/file_downloader.py#L46-L53

If there are plans to make overwrites configurable (via `conf` like with the other options), then it's worth implementing now

conans/requirements_dev.txt
jcar87: Done! Still does the job :D

conans/test/unittests/tools/files/test_downloads.py
```diff
@@ -157,6 +158,33 @@ def test_download_no_retries_errors(self, bottle_server):
         assert "Waiting" not in str(conanfile.output)
         assert "retry" not in str(conanfile.output)
 
+    def test_download_localfile(self):
+        conanfile = ConanFileMock()
+        conanfile._conan_requester = requests
+
+        file_location =  os.path.join(temp_folder(), "file.txt")
+        with open(file_location, 'w') as textfile:
+            textfile.write('this is some content\n')
```
memsharded:
```suggestion
        save(file_location, "this is some content\n")
```

We have one line helpers for this.

conans/test/unittests/tools/files/test_downloads.py
```diff
@@ -157,6 +158,33 @@ def test_download_no_retries_errors(self, bottle_server):
...
+
+        file_url = f"file:///{file_location}"
+        file_md5 = "a0b156435474e688206c68e5c66a3327"
+        
+        dest = os.path.join(temp_folder(), "downloaded_file.txt")
+        download(conanfile, file_url, dest, md5=file_md5)
+        content = load(dest)
+        assert "this is some content" == content.rstrip()
```
memsharded:
```suggestion
        assert "this is some content" == content
```

The rstrip() is probably unnecessary, and risks hiding a bug somewhere. The file must be identical

memsharded (approved):
Looking good

conans/test/unittests/tools/files/test_downloads.py
jcar87: Fixed it by changing the md5 that is checked above (as it stands, given the md5 check, this is probably redundant).

Issue was that most command line text editors and the shell itself will append a newline character, so all of my attempts to get the md5 sum of a string will give me the md5 sum of a string ending with the newline character. Given that a few lines above we have control of the exact byte-wise contents of the file, I think it's safe to remove the `rstrip()`

czoido (approved):

Issue #11275, opened 2022-05-17
[question] How should replacement for `scm` work if recipe not in root Git dir


gabyx:
I am working on replacing the `scm` attribute in https://docs.conan.io/en/latest/reference/conanfile/tools/scm/git.html#example-implementing-the-scm-feature
I stumbled upon the fact that when the `conanfile.py` is in a subfolder `/workspace/repo/tests/project` (could also be a submodule, but ignore that for now) of the main repository `/workspace/repo` , then the example does not work, or I don't know how to work around this:

```python
  def export(self):
    git = self.mymodule.GitRepo(self, self.recipe_folder)
    url, commit = git.get_url_and_commit()
    # We store the current url and commit in conandata.yml
    update_conandata(self, {"sources": {"commit": commit, "url": url}})

  def source(self):
    print("SRC::",os.getcwd()) # Bug: THIS is the self.folder.sources, also it should not be??

    sources = self.conan_data["sources"]
    
    self.mymodule.clone_and_checkout(self,
        url=sources["url"], dest="../../repo", ref=sources["commit"])

  def layout(self):
    self.folders.source = "repo/tests/project"
```

My problem is that: `sources["url"]`  in `sources` points to a local folder on `conan create` which is good, but its the folder where the recipe lies `/workspace/repo/tests/project`, and not the root Git dir `/workspace/repo`.

How can we fix this? Do I need to check if `self.conan_data["sources"]["url"]` is a absolute path instead of an URL? 

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Hi @gabyx 

Sorry for the delay in responding.

I am putting this test, to validate that the recipe can be inside a subfolder and not the root:
```python
class TestConanFileSubfolder:
    conanfile = textwrap.dedent("""
        import os
        from conan import ConanFile
        from conan.tools.scm import Git
        from conan.tools.files import update_conandata, load

        class Pkg(ConanFile):
            name = "pkg"
            version = "0.1"

            def export(self):
                git = Git(self, os.path.dirname(self.recipe_folder))
                url, commit = git.get_url_and_commit()
                # We store the current url and commit in conandata.yml
                update_conandata(self, {"sources": {"commit": commit, "url": url}})
                self.output.info("URL: {}".format(url))
                self.output.info("COMMIT: {}".format(commit))

            def layout(self):
                pass # self.folders.source = "source"

[...]

PR #11573, opened 2022-07-05
fix: Correct test

gabyx:
Changelog: (Bugfix): Fix the test `TestConanFileSubfolder` which should actually fail.

Issue: #11275

This test should be written like this, and will fail because the bug is not yet fixed.
(if its a bug)

### DIFF
--- a/conans/test/functional/tools/scm/test_git.py
+++ b/conans/test/functional/tools/scm/test_git.py
@@ -474,7 +474,7 @@ class TestConanFileSubfolder:
             version = "0.1"
 
             def export(self):
-                git = Git(self, os.path.dirname(self.recipe_folder))
+                git = Git(self, self.recipe_folder)
                 url, commit = git.get_url_and_commit()
                 # We store the current url and commit in conandata.yml
                 update_conandata(self, {"sources": {"commit": commit, "url": url}})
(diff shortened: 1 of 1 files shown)


conans/test/functional/tools/scm/test_git.py
```diff
@@ -474,40 +474,41 @@ class Pkg(ConanFile):
             version = "0.1"
 
             def export(self):
-                git = Git(self, os.path.dirname(self.recipe_folder))
+                git = Git(self, self.recipe_folder)
```
memsharded, line 477: But your git repo is not the folder in which the ``conanfile.py`` lives, it is the parent folder. So the ``os.path.dirname()`` is necessary to use the Git repo folder.

conans/test/functional/tools/scm/test_git.py
```diff
@@ -474,40 +474,41 @@ class Pkg(ConanFile):
...
                 url, commit = git.get_url_and_commit()
                 # We store the current url and commit in conandata.yml
                 update_conandata(self, {"sources": {"commit": commit, "url": url}})
                 self.output.info("URL: {}".format(url))
                 self.output.info("COMMIT: {}".format(commit))
 
             def layout(self):
-                pass # self.folders.source = "source"
+                self.folders.source = "project"
```
memsharded, line 485: Or if not the ``source`` folder would be ``.``, not ``project`` (as ``project`` only happens if the repo folder is the parent one)

memsharded (commented):
Please add a new failing test (it could start from a copy & paste of this one, then doing the changes), instead of modifying the existing one. But in general we don't change green passing tests, unless there were bugs in the test themselves, are result of other minor unrelated changes (changes in the stdout), or things like that.

Then also please have a look to the comments, I'd say there are some inconsistencies in the definition of the folder locations.

memsharded:
This PR seems related too: https://github.com/conan-io/conan/pull/11556

memsharded:
Hi @gabyx 

This was assigned for 1.51, but we didn't get any response, so I am moving it to next iteration 1.52. Please check the above comments and let us know.

gabyx:
@memsharded: Thanks, jeah I did not have time yet but I hope I can move the test.

memsharded:
Hi @gabyx 

This PR has been stale for a few months now, we have been moving it from release to release, I hope it is not an issue anymore, so I am closing it, many thanks for contributing!
If you would like to re-open please do, and please have a look to the comments, specially the one to not change existing passing tests, but add new tests that capture the desired behavior.

Issue #11525, opened 2022-06-25
[feature] Direct dependencies on components


Ahajha:
Motivation:
It is likely not too uncommon for users (myself included) to want to depend on only some components of a package, that's what they're there for after all. However, this has a few minor inconveniences:

1. By declaring that you depend on the entire package, any catch-all linking grabs all components, with no way to change this. (see the `cmake` generator)
1. Even if you are using a target-based approach (like CMakeDeps + CMakeToolchain), you still need to specify you are only using the given component in two places, once at the link site (e.g. CMake) and once in the `package_info()` method.
1. Also with a targets based approach, generators like CMakeToolchain still allow you to link with any other components, potentially allowing for desync between the build system and `package_info()`, and also if it prints the available components it can clutter the output (for example, pulling one component from Boost).
1. Users of the package can still use the other components.

Desired functionality:
I would like to be able to replace the following:
```python
def requirements(self):
    self.requires("boost/1.76.0")

def package_info(self):
    self.cpp_info.requires = ['boost::iostreams']
```
with something like:
```python
def requirements(self):
    self.requires("boost::iostreams/1.76.0")
```
For multiple components, perhaps something like:
```python
def requirements(self):
    self.requires("boost::{iostreams, system}/1.76.0")
```

This would do a few things, addressing the points above directly:
1. Any catch-all linking will only pull the specified component(s).
1. Automatically specify that only the requested component(s) will be propagated to consumers as if done the current way in `package_info()`.
1. Only the component(s) requested (and any components they depend on) should be forwarded to the generators.
1. Users of the package would have to specify additional components (perhaps in the same way) if they wanted to use non-propagated components.

I would be okay with this being split up, even just point 2 would be a nice QOL increase to start.

I would love feedback/input, maybe there's something I missed that would make this unnecessary, but if not I think this would be a nice way of more directly expressing intent.

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

czoido:
Hi @Ahajha,
Thanks a lot for your question. I think it makes sense to be able to just select the components you want to depend on and not propagate everything. But yes, this could be made in two steps, one for simplifying the syntax and not having to specify it twice in the `package_info` and the second to restrict the information passed downstream.
[...]

PR #11574, opened 2022-07-05
Test that consumers of packages that use cpp_info requires on components don't propagate other components

czoido:

top/1.0 --> two components cmp1, cmp2
middle/1.0 --> uses `self.cpp_info.requires = ["top::cmp1"]` 
consumer --> requires middle/1.0 and can use the target `top::cmp2` without failing.

Related with: https://github.com/conan-io/conan/issues/11525


memsharded (commented):
I would really try to test this as ``integration``, to try to keep test times as low as possible.
It shouldn't be necessary to build anything, components can be header only, and checks can be done on the generated files (if there is a ``comp2-config.cmake``, then it was not excluded and can be used)

czoido:
> I would really try to test this as `integration`, to try to keep test times as low as possible. It shouldn't be necessary to build anything, components can be header only, and checks can be done on the generated files (if there is a `comp2-config.cmake`, then it was not excluded and can be used)

Ok, made it header only and just checking that after conan install some files define the top::cmp2 component

memsharded (approved):

PR #11575, opened 2022-07-05
[dev2][conf] Improvements

### LINKED ISSUE #11346: [dev2][feature] `conf` review and improvements
Several points here:
* Fix the wrong behavior about writing `-c core.xxx="whatever"` via CLI. This should raise an error because those configurations are not used by the profiles. Could we add another option like `-c:global core.xxxxx="whatever"`?
* Centralize all the existing configurations (`tools.xxx` and `core.xxx`) in one place (a python list, dict, whatever), and raise an error if `conf.get("xxxxx")` is not declared there (except for `user.xxxx` configurations).
* Perhaps, we should bring the `conan config list` command back to show all the available configurations. 

_Related to https://github.com/conan-io/conan/issues/11205_

franramirez688:
Changelog: Bugfix: Fixed the wrong behavior about writing `-c core.xxx=value` via CLI.
Changelog: Feature: Raising an exception if the internal configuration is used and it does not exist in the configuration list. It only affects `tools*` and `core*` ones.
Changelog: Feature: Brought `conan config list` command back.
Closes: https://github.com/conan-io/conan/issues/11346

### DIFF
--- a/conans/cli/commands/config.py
+++ b/conans/cli/commands/config.py
@@ -1,5 +1,8 @@
+import json
+
 from conans.cli.command import conan_command, COMMAND_GROUPS, conan_subcommand
-from conans.cli.output import ConanOutput
+from conans.cli.output import ConanOutput, Color
+from conans.model.conf import BUILT_IN_CONFS
 from conans.util.config_parser import get_bool_from_text
 
 
@@ -46,3 +49,17 @@ def config_home(conan_api, parser, subparser, *args):
     home = conan_api.config.home()
     ConanOutput().info(f"Current Conan home: {home}")
     return home
+
+
+@conan_subcommand(formatters={"json": lambda x: json.dumps(x, indent=4)})
+def config_list(conan_api, parser, subparser, *args):
+    """
+    Prints all the Conan available configurations: core and tools.
+    """
+    out = ConanOutput()
+    confs = BUILT_IN_CONFS
+    out.writeln(f"Supported Conan global.conf and [conf] properties:",
+                fg=Color.BRIGHT_CYAN)
+    for k, v in confs.items():
+        out.writeln(f"{k}: {v}")
+    return confs
--- a/conans/cli/commands/config.py
+++ b/conans/cli/commands/config.py
@@ -1,6 +1,5 @@
-import json
-
 from conans.cli.command import conan_command, COMMAND_GROUPS, conan_subcommand
+from conans.cli.commands import json_formatter
 from conans.cli.output import ConanOutput, Color
 from conans.model.conf import BUILT_IN_CONFS
 from conans.util.config_parser import get_bool_from_text
@@ -51,7 +50,7 @@ def config_home(conan_api, parser, subparser, *args):
     return home
 
 
-@conan_subcommand(formatters={"json": lambda x: json.dumps(x, indent=4)})
+@conan_subcommand(formatters={"json": json_formatter})
 def config_list(conan_api, parser, subparser, *args):
     """
     Prints all the Conan available configurations: core and tools.
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -38,6 +38,7 @@ BUILT_IN_CONFS = {
     "tools.build:skip_test": "Do not execute CMake.test() and Meson.test() when enabled",
     "tools.build:jobs": "Default compile jobs number -jX Ninja, Make, /MP VS (default: max CPUs)",
     "tools.build:sysroot": "Pass the --sysroot=<tools.build:sysroot> flag if available. (None by default)",
+    "tools.build.cross_building:can_run": "",
     "tools.cmake.cmaketoolchain:generator": "User defined CMake generator to use instead of default",
     "tools.cmake.cmaketoolchain:find_package_prefer_config": "Argument for the CMAKE_FIND_PACKAGE_PREFER_CONFIG",
     "tools.cmake.cmaketoolchain:toolchain_file": "Use other existing file rather than conan_toolchain.cmake one",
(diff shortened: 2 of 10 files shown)


conans/model/conf.py
```diff
@@ -62,19 +70,27 @@
...
     "tools.build:cflags": "List of extra C flags used by different toolchains like CMakeToolchain, AutotoolsToolchain and MesonToolchain",
     "tools.build:defines": "List of extra definition flags used by different toolchains like CMakeToolchain and AutotoolsToolchain",
     "tools.build:sharedlinkflags": "List of extra flags used by CMakeToolchain for CMAKE_SHARED_LINKER_FLAGS_INIT variable",
     "tools.build:exelinkflags": "List of extra flags used by CMakeToolchain for CMAKE_EXE_LINKER_FLAGS_INIT variable",
+    # Package ID composition
+    "tools.info.package_id:confs": "List of existing configuration to be part of the package ID",
 }
 
 
+CORE_CONF_PATTERN = re.compile(r"^core[\.:]")
```
memsharded, line 91: Maybe just check ``if any(conf.startswith(s) for s in ("core.", "tools.")`` would be better than a regex?

conans/model/conf.py
```diff
@@ -231,14 +247,18 @@ def _get_boolean_value(value):
...
-
         Get all the values of the given configuration name.
 
         :param conf_name: Name of the configuration.
         :param default: Default value in case of conf does not have the conf_name key.
         :param check_type: Check the conf type(value) is the same as the given by this param.
                            There are two default smart conversions for bool and str types.
         """
+        # Skipping this check only the user.* configurations
+        if USER_CONF_PATTERN.match(conf_name) is None and conf_name not in BUILT_IN_CONFS:
```
memsharded, line 264: Maybe just checking that it doesn't start with ``user.``, then it MUST exist in the dict.

memsharded (approved):
Looking good!

conans/model/conf.py
franramirez688: I've seen a couple of core confs which are not covered by that `startswith` like:

```
"core:required_conan_version": "Raise if current version does not match the defined range.",
"core:non_interactive": "Disable interactive user input, raises error if input necessary",
```
So, I think any user could define `user:whatever=xxxxx` and it'd be certainly fine, wouldn't it? That's the reason why I've added these simple patterns.

memsharded: Yes, that is right, I forgot about those.

conans/test/unittests/cli/common_test.py
```diff
@@ -0,0 +1,51 @@
...
+from conans.client.cache.cache import ClientCache
+from conans.errors import ConanException
+from conans.test.utils.test_files import temp_folder
+
+
+@pytest.fixture()
+def conan_api():
+    tmp_folder = temp_folder()
+    cache = ClientCache(tmp_folder)
+    save(None, cache.default_profile_path, textwrap.dedent("""\
```
memsharded: A blank file is probably good too

conans/test/unittests/cli/common_test.py
```diff
@@ -0,0 +1,51 @@
...
+        conf_build=None,
+        conf_host=None
+    )
+
+
+@pytest.mark.parametrize("conf_name", [
+    "core.doesnotexist:never",
+    "core:doesnotexist"
+])
+def test_core_confs_not_allowed_via_cli(conan_api, argparse_args, conf_name):
```
memsharded, line 40: This test reads complex, too many mocks, etc. If it is a unit test of profile loader, why unittesting the ``get_profiles_from_args`` which is a cli helper, instead of testing the profile loader (that can be directly called by the Python API ProfileAPI without calling that cli helper?

franramirez688: Because I wanted to test the function that is used by several commands at the same time. If I test the ProfileLoader I could not ensure that the commands are getting that function so I thought it was better covered so. That was the reason.
Talking about mocks, I don't think a mock for argparse is too much but I can change the test and call only the  ProfileLoader function if you think is good enough.

conans/model/conf.py
franramirez688: Like that must be a pattern, could we mark this comment as resolved? Because I think it's doing exactly what you're saying

conans/test/unittests/cli/common_test.py
franramirez688: I just added another unittest for ProfileLoader. If you want, I can get rid of this one.

memsharded: No need to remove it now it is done, but in general, the definition of "unittest" is testing the unit that is getting the code changes, in this case the "ProfileLoader". It is not about ensuring that certain commands behave some way, that is an "integration" test. The test that you added is much simpler (less mocking), and it guarantees that no matter what changes in the command layer, new commands, refactors, etc, profiles will keep raising if bad conf is defined. For example the ``ProfileAPI`` sub-api was not being covered by the first test, but now it is. Thanks!

memsharded (approved):

conans/cli/commands/config.py
```diff
@@ -46,3 +49,17 @@ def config_home(conan_api, parser, subparser, *args):
     home = conan_api.config.home()
     ConanOutput().info(f"Current Conan home: {home}")
     return home
+
+
+@conan_subcommand(formatters={"json": lambda x: json.dumps(x, indent=4)})
```
czoido: Maybe do like other commands and use `json_formatter` imported from `conans.cli.commands`?

czoido (approved):

PR #11578, opened 2022-07-05
[develop2] remove old profile [env]

memsharded:
- Remove some legacy unused [env] in profiles
- Simplify the default profile output, to only show used fields


jcar87 (approved):
LGTM!

PR #11580, opened 2022-07-06
validate_build over settings

lasote:
Changelog: Feature: Introduced a `validate_build()` method in the recipes to validate if the package can be built according to `self.settings` and `self.options` instead of `self.info.settings` and `self.info.options` that it is used to validate the `binary`.
Docs: https://github.com/conan-io/docs/pull/2667

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -52,7 +52,7 @@ class GraphBinariesAnalyzer(object):
                 node.binary = BINARY_INVALID
             else:
                 node.binary = BINARY_BUILD
-                node.prev = None
+            node.prev = None
             return True
 
     def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
@@ -308,7 +308,7 @@ class GraphBinariesAnalyzer(object):
                         node.binary = BINARY_INVALID
                     else:
                         node.binary = BINARY_BUILD
-                        node.prev = None
+                    node.prev = None
                 else:
                     conanfile.output.info("Package is up to date")
 
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -50,6 +50,7 @@ class GraphBinariesAnalyzer(object):
             conanfile.output.info('Forced build from source')
             node.binary = BINARY_BUILD
             node.prev = None
+            _call_validate_build(node)
             return True
 
     def _evaluate_clean_pkg_folder_dirty(self, node, package_layout, pref):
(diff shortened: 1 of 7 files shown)


conans/client/graph/graph_binaries.py
```diff
@@ -451,3 +455,14 @@ def reevaluate_node(self, node, remotes, build_mode, update):
...
+
+
+def _call_validate_build(node):
+    conanfile = node.conanfile
+    if hasattr(conanfile, "validate_build") and callable(conanfile.validate_build):
+        with conanfile_exception_formatter(str(conanfile), "validate_build"):
+            try:
+                conanfile.validate_build()
+            except ConanInvalidConfiguration as e:
+                conanfile.info.invalid = str(e)
```
memsharded: This is weird. If it cannot be built, it doesn't depend on the ``info`` for ``package_id``. It is a property of the ``conanfile.cant_build = str(e)``. Overriding ``info.invalid`` can destroy information of the potential valid/invalid of the ``package_id``

conans/client/graph/graph_binaries.py
```diff
@@ -295,6 +298,7 @@ def _process_node(self, node, pref, build_mode, update, remotes):
                     conanfile.output.info("Outdated package!")
                     node.binary = BINARY_BUILD
                     node.prev = None
+                    _call_validate_build(conanfile)
```
memsharded: I would try to call this earlier in the process, just once, as it doesn't have much effect besides setting an attribute, it should be possible.

memsharded (commented):
Too many tests are broken, but I fail to see how these changes can cause that breakage

conans/client/graph/graph_binaries.py
lasote: ok, this is a copy paste I didn't know anything about the `cant_build`

conans/client/graph/graph_builder.py
```diff
@@ -471,4 +471,12 @@ def _create_new_node(self, current_node, dep_graph, requirement, check_updates,
 
         dep_graph.add_node(new_node)
         dep_graph.add_edge(current_node, new_node, requirement)
+
+        if hasattr(dep_conanfile, "validate_build") and callable(dep_conanfile.validate_build):
```
lasote: Probably do the `validate_build()` here is too soon to have the final options.

memsharded: It is, it will not have the final options values.

conans/client/graph/graph_manager.py
```diff
@@ -9,7 +9,7 @@
 from conans.client.graph.graph_binaries import RECIPE_CONSUMER, RECIPE_VIRTUAL, BINARY_EDITABLE, \
     BINARY_UNKNOWN
 from conans.client.graph.graph_builder import DepsGraphBuilder
-from conans.errors import ConanException, conanfile_exception_formatter
+from conans.errors import ConanException, conanfile_exception_formatter, ConanInvalidConfiguration
```
memsharded: unused?

conans/client/graph/graph_builder.py
lasote: Moved, please review. I don't really know the best place to do it.

memsharded (approved):
Looks good.

Don't forget to mark it as experimental in the docs

czoido (approved):

Issue #11571, opened 2022-07-04
[bug] Errors for incompatible option values for dependencies only shown for specific requires order


foundry-markf:
I was investigating a Conan recipe setup that I expected to result in an error, because a dependency was being requested (directly and indirectly) with different option values. Instead, I saw no error, and the dependency was configured not how the top-level recipe asked for it, but from how the indirect dependency configured it.

The only difference is that the `self.requires` order is changed.

I was expecting that the order of requirements would not be significant, so can't explain this.

### Environment Details (include every applicable attribute)
  * Operating System+version: Observed on Mac (10.14) and Windows (10)
  * Compiler+version: Independent of compiler
  * Conan version: 1.48-1.50
  * Python version: 3.10

### Steps to reproduce (Include if Applicable)
I've created a gist as a significantly cut down reproducible of what I saw today: https://gist.github.com/foundry-markf/c2b459405b8c3aab78645171cedf6ef4

Run `./build.sh`

### Logs (Executed commands with output) (Include/Attach if Applicable)
My log from Mac is included in the gist: https://gist.github.com/foundry-markf/c2b459405b8c3aab78645171cedf6ef4#file-mf_log-txt

Output from `conan search` to determine which configuration of the dependent library had been chosen in the unexpectedly succeeding case: https://gist.github.com/foundry-markf/c2b459405b8c3aab78645171cedf6ef4#file-mf_search_log-txt

memsharded:
Hi @foundry-markf 

We have been investigating this, and trying different approaches:

- Indeed in Conan 1.X the conflict is order-dependent. The way and order the dependencies are evaluated affects the possible conflicts
- The problem still persisted in Conan 2.0, in different ways, and it was a bit more manageable, but still very confusing.
- We have tried a few ways to improve the potential options conflicts detection both in 1.X and 2.0, and we have concluded that it is not worth the fragility and complexity.

So we are going with the following approach for 2.0:
- Definitions of options for dependencies in conanfile.py recipes are a prioritized but not conflicting definition. The order of evaluation of the graph might produce different results in some situations.
- Thus, options never conflict, they will get different values based on the priorities:
   - Consumers always have higher priority than dependencies
   - Profile and command line definitions always have top priority. No matter what recipes say, if a profile define some values, they will be assigned, even if they are incompatible with values defined by recipes ``configure()`` or ``default_options``
   - In case of same level dependencies (diamonds), the first one to expend (the first one in ``requires``) will have precedence
[...]

PR #11583, opened 2022-07-06
[develop2] managing options conflicts with profiles priority

memsharded:
Investigate https://github.com/conan-io/conan/issues/11571

After some discussions, this PR is proposing:

- Removing options conflict in Conan 2.0. They are simply too problematic, complicated, fragile and confusing to users
- Recipes defining options for their dependencies propagate according to priorities. First one to define wins. Downstream recipes defining things have precedence. Profiles and command line defined options has highest precedence, they should always be respected, even if recipes says a different thing
- Recipes that will definitely not work with some of their dependencies having some invalid option values, should raise ``ConanErrorConfiguration`` in the ``validate()`` method

### DIFF
--- a/conans/test/integration/graph/core/test_options.py
+++ b/conans/test/integration/graph/core/test_options.py
@@ -55,25 +55,30 @@ class TestOptions(GraphManagerTest):
                                 (liba, True, True, False, False)])
         _check_transitive(libb, [(liba, True, True, False, False)])
 
-    def test_diamond_conflict(self):
+    def test_diamond_no_conflict(self):
         # app -> libb0.1 ---------------> liba0.1
         #    \-> libc0.1 (liba shared) -> liba0.1
+        # Just the first expended one will prevail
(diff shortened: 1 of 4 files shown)


conans/test/integration/graph/core/test_options.py
```diff
@@ -55,20 +55,86 @@ def test_app_override(self):
                                 (liba, True, True, False, False)])
         _check_transitive(libb, [(liba, True, True, False, False)])
 
-    def test_diamond_conflict(self):
+    def test_diamond_no_conflict(self):
         # app -> libb0.1 ---------------> liba0.1
         #    \-> libc0.1 (liba shared) -> liba0.1
+        # Just the first expended one will prevail
```
lasote:
```suggestion
        # Just the first expanded one will prevail
```

lasote (approved):
🆗

czoido (approved):

franramirez688 (approved):

PR #11585, opened 2022-07-07
Feature : provide Path accesssors

### LINKED ISSUE #11304: [feature] add `self.*_path` properties that return `pathlib.Path`

We have `recipe_folder`, `source_folder`, etc that return `Optional[str]`.

It would be nice to work with filesystem paths using [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html).

As the `*_folder` can be `None` in certain stages, it is usually a programming error accessing them when they _are_ `None`.

It would be nice to have the following accessors on the `ConanFile`:

```py
    @property
    def package_path(self) -> Path:
        """
        :returns: The location of the final packaging location              
        """
        assert self.package_folder is not None, "`package_folder` is `None`"
        return Path(self.package_folder)
```

Then we can use it like so:

```py
    def build(self) -> None:
        exe = self.package_path / "bin" / "hello-world"
        exe.parent.mkdir()
        with exe.open(encoding="utf8") as stream:
            stream.write("#! /bin/sh\necho 'Hello, world!'')
        exe.chmod(0o755)
```

Which is much nicer that the equivalent code with `os.path`.

SSE4:
closes: #11304 

Changelog: Feature: Provide Path accessors in Conanfile.
Docs: omit

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -1,6 +1,7 @@
 import os
 import platform
 from contextlib import contextmanager
+from pathlib import Path
 
 import six
 from six import string_types
@@ -259,6 +260,11 @@ class ConanFile(object):
         return self.folders.source_folder
 
     @property
+    def source_path(self) -> Path:
+        assert self.source_folder is not None, "`source_folder` is `None`"
+        return Path(self.source_folder)
+
+    @property
     def export_sources_folder(self):
         """points to the base source folder when calling source() and to the cache export sources
         folder while calling the exports_sources() method. Prepared in case we want to introduce a
@@ -266,16 +272,36 @@ class ConanFile(object):
         return self.folders.base_export_sources
 
     @property
+    def export_sources_path(self) -> Path:
+        assert self.export_sources_folder is not None, "`export_sources_folder` is `None`"
+        return Path(self.export_sources_folder)
+
+    @property
     def export_folder(self):
         return self.folders.base_export
+
+    @property
+    def export_path(self) -> Path:
+        assert self.export_folder is not None, "`export_folder` is `None`"
+        return Path(self.export_folder)
 
     @property
     def build_folder(self):
         return self.folders.build_folder
 
     @property
+    def build_path(self) -> Path:
+        assert self.build_folder is not None, "`build_folder` is `None`"
+        return Path(self.build_folder)
+
+    @property
     def package_folder(self):
         return self.folders.base_package
+
+    @property
+    def package_path(self) -> Path:
+        assert self.package_folder is not None, "`package_folder` is `None`"
+        return Path(self.package_folder)
 
     @property
     def install_folder(self):
@@ -286,6 +312,11 @@ class ConanFile(object):
     def generators_folder(self):
         # FIXME: Remove in 2.0, no self.install_folder
         return self.folders.generators_folder if self.folders.generators else self.install_folder
+
+    @property
+    def generators_path(self) -> Path:
+        assert self.generators_folder is not None, "`generators_folder` is `None`"
+        return Path(self.generators_folder)
 
     @property
     def imports_folder(self):
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -8,6 +8,8 @@ import uuid
 
 import yaml
 
+from pathlib import Path
+
 from conan.tools.cmake import cmake_layout
 from conan.tools.google import bazel_layout
 from conan.tools.microsoft import vs_layout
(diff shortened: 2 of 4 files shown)


conans/model/conan_file.py
```diff
@@ -258,39 +259,79 @@ def new_cpp_info(self):
...
+        assert self.package_folder is not None, "`package_folder` is `None`"
+        return Path(self.package_folder)
+
     @property
     def install_folder(self):
         # FIXME: Remove in 2.0, no self.install_folder
         return self.folders.base_install
 
+    @property
+    def install_path(self) -> Path:
```
memsharded: This one disappears in 2.0, better remove it.

conans/model/conan_file.py
```diff
@@ -258,39 +259,79 @@ def new_cpp_info(self):
...
+    def generators_path(self) -> Path:
+        assert self.generators_folder is not None, "`generators_folder` is `None`"
+        return Path(self.generators_folder)
+
     @property
     def imports_folder(self):
         return self.folders.imports_folder
 
+    @property
+    def imports_path(self) -> Path:
```
memsharded: This one disappears in 2.0, better remove it.

conans/model/layout.py
```diff
@@ -129,6 +149,11 @@ def imports_folder(self):
 
         return os.path.join(self._base_imports, self.imports)
 
+    @property
+    def imports_path(self) -> Path:
```
memsharded: This one disappears in 2.0, better remove it.

conans/model/layout.py
```diff
@@ -1,7 +1,7 @@
 import os
 
 from conans.model.new_build_info import NewCppInfo
-
+from pathlib import Path
```
memsharded: I'd say that the ``Folders`` ``xxxx_folder`` are not intended nor documented for users yet. 
If anything we want to replace the ``xxx_folder`` for the path alternatives. But this is not necessary in this PR, as this is intended only for user recipes which is always ``self.xxxx_path``, and not ``self.folders.xxxx_path``.

So maybe leave this file untouched.

memsharded (commented):
Please add some basic ``integration`` test that checks these added methods.

conans/model/layout.py
SSE4: okay, I left it untouched

conans/model/layout.py
SSE4: removed

conans/model/conan_file.py
SSE4: removed

conans/model/conan_file.py
SSE4: removed

SSE4:
> Please add some basic `integration` test that checks these added methods.

added (updated an existing test with more checks)

ytimenkov:
Out of curiosity: why not to change existing fields but add new ones? What are the potential issues with changing the type?

SSE4:
> Out of curiosity: why not to change existing fields but add new ones? What are the potential issues with changing the type?

it may break existing recipes. e.g. the one doing something like `isinstance(self.package_folder, str)` or similar.

ytimenkov:
> it may break existing recipes. e.g. the one doing something like `isinstance(self.package_folder, str)` or similar.

Thank you for quick reply. That part I have figured out. Was more interested if there are recipes which do this and why.

In my limited python experience calling `isinstance` is discouraged. The most common check for type I know is `str` vs `list` of strings (where `str` is converted to a single-element `list`, but the check is for `list`, because it can be something else) for caller convenience. Other examples `str` vs `bytes` - just call `encode`, and there is `is None` for optionals.

SSE4:
I don't know any recipe doing it out of my mind. still, it's possible, some recipe does `some_python_func(self.package_folder)` where `some_python_func` calls `is_instance` under the hood.

nethertheless, there could be other breaking cases, e.g.:
```
var = self.package_folder.replace("foo", "bar")
```
will crash if `self.package_folder` is `pathlib.Path`, but it will work if it's just a regular string. and `replace` for sure is done in many recipes.
an example: https://github.com/conan-io/conan-center-index/blob/c102ed11805c88f636198d1feec892eea8c20579/recipes/bison/all/conanfile.py#L156

for sure it could be changed in 2.0, but for 1.x I would say it's too risky.

memsharded:
I am merging this because it is done, but after talking to the team, please let me clarify:

- We really, really need to stop adding new things to Conan 1.X. Only add things that are strictly necessary for the migration to 2.0. This was clearly not necessary.
- We will not document it yet, now things added to 1.X basically need to be documented twice, and in 2.0 is complicated because we are still discussing many things around the docs.
- This will be there as experimental and not supported yet. If it becomes a maintenance or support burden, we will remove it, but we will not invest more time in things that could derive from this feature, until Conan 2.X

Thanks for understanding

PR #11609, opened 2022-07-11
Test that cpp_info.requires works correctly for XcodeDeps

czoido:
Changelog: omit
Docs: Maybe Document a bit about selecting components for direct dependencies

Related to: https://github.com/conan-io/conan/issues/11525

### DIFF
--- a/conans/test/integration/toolchains/apple/test_xcodedeps.py
+++ b/conans/test/integration/toolchains/apple/test_xcodedeps.py
@@ -195,7 +195,6 @@ def test_xcodedeps_cppinfo_requires():
             name = "lib_a"
             version = "1.0"
             settings = "os", "compiler", "build_type", "arch"
-            exports_sources = "include/*"
 
             def package_info(self):
                 self.cpp_info.components["cmp1"].includedirs = ["include"]
@@ -214,7 +213,6 @@ def test_xcodedeps_cppinfo_requires():
             name = "lib_{name}"
             version = "1.0"
             settings = "os", "compiler", "build_type", "arch"
-            exports_sources = "include/*"
 
             def requirements(self):
                 self.requires("lib_a/1.0")
--- a/conans/test/integration/toolchains/apple/test_xcodedeps.py
+++ b/conans/test/integration/toolchains/apple/test_xcodedeps.py
@@ -267,6 +267,8 @@ def test_xcodedeps_cppinfo_requires():
 
     lib_b = client.load("conan_lib_b_lib_b.xcconfig")
 
+    # check that nothing from other components than the specified in the cpp_info.requires
+    # from lib_b and lib_c exist in the xcconfig that adds the includes from components
     assert "cmp1" in lib_b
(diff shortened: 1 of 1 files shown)


conans/test/integration/toolchains/apple/test_xcodedeps.py
```diff
@@ -196,3 +196,85 @@ def package_info(self):
...
+    """
+    client = TestClient()
+    lib_a = textwrap.dedent("""
+        from conan import ConanFile
+        class lib_aConan(ConanFile):
+            name = "lib_a"
+            version = "1.0"
+            settings = "os", "compiler", "build_type", "arch"
+            def package_info(self):
+                self.cpp_info.components["cmp1"].includedirs = ["include"]
```
lasote, line 218: Would it be worth using a different `include` folder to make the final assert more meaningful?

lasote (approved):

franramirez688 (approved):

conans/test/integration/toolchains/apple/test_xcodedeps.py
czoido: That is just used to create the components, the test does not test that include directory exists, because that is added in the own component file: `conan_lib_a_cmp1_release_x86_64.xcconfig` it tests that `conan_lib_b_lib_b.xcconfig` does not include the xcconfig of the component, I'll add some comments to clarify the test

lasote: I know, but eventually, the "include" string should appear in the file I guess?

czoido: Yes, it appears in the file from the four components:

- `conan_lib_a_cmp1_release_x86_64.xcconfig`
- `conan_lib_a_cmp2_release_x86_64.xcconfig`
- ...

like:

```
HEADER_SEARCH_PATHS_lib_a_cmp1[config=Release][arch=x86_64][sdk=*] = ".../package/4b92821f1af51ada4198018b78c8b197b32c6c18/include"
```
Then, that variable is inherited in the `conan_lib_a_cmp1.xcconfig` like:

```
#include "conan_lib_a_cmp1_release_x86_64.xcconfig"

HEADER_SEARCH_PATHS = $(inherited) $(HEADER_SEARCH_PATHS_lib_a_cmp1)
```

and then, the lib_b xcconfig includes `#include "conan_lib_a_cmp1.xcconfig"`

```
// Conan XcodeDeps generated file for lib_b::lib_b
// Includes all configurations for each dependency

// Includes for lib_a::cmp1 dependency
#include "conan_lib_a_cmp1.xcconfig"

#include "conan_lib_b_lib_b_release_x86_64.xcconfig"
```

But in this file we have no trace of include directories, just variables that inherit from that value in `conan_lib_a_cmp1_release_x86_64.xcconfig` so I can't assert that value

lasote: ok. nevermind.

Issue #11592, opened 2022-07-08
[feature] Please provide all methods in base ConanFile


ytimenkov:
Currently, `ConanFile` base class doesn't provide methods which need to be overridden by user (for example `package()`), this makes writing mixins which add different steps to "common" flow harder.

Consider example:
```python

class MixinHeaders(ConanFile):
    def package():
        super().package() # < See below
        self.copy("*.h")

class MixinLibraries(ConanFile):
    def package():
        super().package() # < See below
        self.copy("*.so*");

# Consumer:
class MyConan(ConanFIle):
    python_requires_extend = ["shared.MixinHeaders", "shared.MixinLibraries"]
```

As I understand, in order for this to work with arbitrary set of mixins all of them should call `super().package()` (something called MRO involved). The problem is that for one of them `super()` will be `ConanFile` which doesn't have `package()` and recipe fails with `AttributeError`.

To work this around I had to use something like in every mixin:
```python
try:
    super_package = getattr(super(), "package")
except AttributeError:
    pass
else:
    super_package()
```
which is bit ugly.

Maybe I get something wrong.

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Not related, but I think it is important, feedback: Please try to avoid complex inheritance schemes, like transitive inheritance or multiple inheritance. The ``python_requires_extend`` mechanism might be limited and not be able to support some of the Python standard inheritance. 

My general recommendation for all readers would be:
- Prefer composition over inheritance (that is a known design pattern, not only for this). In this case, it means explicitly calling the ``python_required`` modules, instead of extending them.
- Avoid multiple inheritance, including transitive inheritance. Try to reduce as much as possible the number of different python_requires packages to maintain


For this specific feature, I am not sure. I have been lately considering going the other way, as it has some advantages to not have the methods declared in the base ``ConanFile`` class, as it is more evident and explicit when the user is providing a given method or not, and that can be easily used to code behavior.

ytimenkov:
I do agree that aggregation is to prefer. I just thought that since this mechanism is provided it should be used for these purposes. Was much skeptical myself about sharing python code as base classes when they were first introduced.

> For this specific feature, I am not sure. I have been lately considering going the other way

Yes, I noticed that conan prints / warns if there is no user-provided method. Was not sure about how important that feature was or mere convenience.

Let's close this one. Thanks for sharing your vision.

[...]

PR #11612, opened 2022-07-11
[develop2] remove conanfile base methods

memsharded:
Close https://github.com/conan-io/conan/issues/11592

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -1,9 +1,6 @@
 import os
 
-import platform
-from contextlib import contextmanager
 from pathlib import Path
-
 
 from conans.cli.output import ConanOutput
 from conans.client.subsystems import command_env_wrapper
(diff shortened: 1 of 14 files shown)


conans/model/conan_file.py
```diff
@@ -403,48 +400,6 @@ def generators_path(self) -> Path:
...
-
-    def package(self):
-        """ package the needed files from source and build folders.
-        E.g. copy(self, "\*.h", os.path.join(self.source_folder, "src/includes"), os.path.join(self.package_folder, "includes"))
-        """
-        self.output.warning("This conanfile has no package step")
-
-    def package_info(self):
-        """ define cpp_build_info, flags, etc
-        """
```
franramirez688, line 446: Should we add as ConanFile's docstring the list of methods? Only to be able to see the list of all the methods available.

franramirez688 (approved):

conans/model/conan_file.py
memsharded: To the general docstring of the class? If you are reading the docstring in the conan docs, you will already have there the list of methods available, won't you?

PR #11613, opened 2022-07-12
fix cmake generator typo

abouvier:
Changelog: Omit
Docs: Omit

https://cmake.org/cmake/help/latest/command/find_library.html

The correct keyword is `NAMES`, not `NAME`. Currently `find_library()` try to find two libraries: one literally named `NAME` and one named with the content of `${_LIBRARY_NAME}`.

### DIFF
--- a/conans/client/generators/cmake_common.py
+++ b/conans/client/generators/cmake_common.py
@@ -257,7 +257,7 @@ class CMakeCommonMacros:
     conan_find_libraries_abs_path = textwrap.dedent("""
         function(conan_find_libraries_abs_path libraries package_libdir libraries_abs_path)
             foreach(_LIBRARY_NAME ${libraries})
-                find_library(CONAN_FOUND_LIBRARY NAME ${_LIBRARY_NAME} PATHS ${package_libdir}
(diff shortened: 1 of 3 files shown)


memsharded (commented):
Thanks for fixing this. Can you please have a look to the broken test?

In any case, an important observation: this generator is legacy. Conan 2.0 has already removed it, and it is necessary to move to CMakeDeps and CMakeToolchain to be 2.0 ready.

memsharded:
> In any case, an important observation: this generator is legacy. Conan 2.0 has already removed it, and it is necessary to move to CMakeDeps and CMakeToolchain to be 2.0 ready.

In that line, it should be important to check if this is happening too in CMakeDeps

memsharded:
Yes, this seems it is not affecting CMakeDeps, it was already fixed there with ``NAMES``

abouvier:
> Thanks for fixing this. Can you please have a look to the broken test?

The test apparently failed because an assertion was trying to find a string with the old faulty keyword.

franramirez688 (approved):

franramirez688:
Thanks for contributing @abouvier 
The fix is going to be released in the next Conan 1.51 version.

Issue #11597, opened 2022-07-09
[bug] resdirs not working how user expects


paulharris:
### Environment Details (include every applicable attribute)
  * Operating System+version: Linux
  * Compiler+version: gcc
  * Conan version: 1.50.0
  * Python version: 3.good.one

See #11471 
copying the comment stream in here, as it isn't a simple obvious fix for me.

@lasote I think you need to look at this again.
This change broke my vtk recipe (expectedly) as it was not setting resdirs, and I had assumed the files would end up in "res" eg "res/licences"

However, the files now end up packagedir/None/licences !
What is the "None" doing there?
What is an empty resdirs supposed to mean exactly? Should the files end up in packagedir/licences going forward?

----

Also, side note, there doesn't seem to be any docs about adjusting dirs (in the layout(self) method) AND using the cmake_layout(self) call.

eg:
https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html?highlight=resdirs
does not include any mention of cmake_layout(self)

Nor this page: https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmaketoolchain.html?highlight=resdirs

It is not clear if further adjustments should be done before, or after cmake_layout(self) ... ie it is not clear if cmake_layout will overwrite any previous settings, or will incorporate them...

----

Last problem, the migration docs say You can adjust the cpp_info in the layout method too, not only for a package in the cache, that was typically done in the package_info() method using the self.cpp_info,

but I have found setting self.cpp_info.components[comp].resdirs = ["res"] has no effect, yet that pattern is symmetrical with the old setting for libdirs.

----

I changed my recipe to:
```
def layout(self):
   cmake_layout(self)
   self.cpp.package.resdirs = ["res"]
```
And I removed these calls in package(self):
```
self.cpp_info.components[comp].resdirs = ["res"]
```
ie "res" is only defined in one place.

This resulted in an empty folder `packagedir/res/licences` - right location, wrong content.
The licence files were put in `packagedir/licences`

ie, no more "None" in the folder path.

I put back the cpp_info...resdirs calls, that didn't help either.

I can't figure out how to do this ...

The recipe (with resdirs lines commented out) is here for reference:
https://github.com/conan-io/conan-center-index/pull/10776
Still doesn't build in the CI as the cgns recipe has unexpected errors in the CI.

paulharris:
I think it is because cmake_flags.py and blocks.py are setting
[...]

PR #11614, opened 2022-07-12
Do not set default dirs if there is no value

lasote:
Changelog: Bugfix: `CMakeToolchain` avoid setting `CMAKE_INSTALL_XXX` variables if the values are not defined.
Docs: omit

Related https://github.com/conan-io/conan/issues/11597 pending to verify if closes

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -854,13 +854,21 @@ class OutputDirsBlock(Block):
 
         return textwrap.dedent("""
            set(CMAKE_INSTALL_PREFIX "{{package_folder}}")
+           {% if default_bin %}
            set(CMAKE_INSTALL_BINDIR "{{default_bin}}")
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -854,13 +854,21 @@ def template(self):
 
         return textwrap.dedent("""
            set(CMAKE_INSTALL_PREFIX "{{package_folder}}")
+           {% if default_bin %}
```
franramirez688, line 857: Could any of these directories be `""`? I mean, to specify the current directory we should write `"."`, right?

franramirez688 (approved):

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
memsharded: Yes, if defined, at least a "." is necessary, the empty directory semantics is not very clear.

Issue #11555, opened 2022-06-30
XcodeDeps Generator - Traits Support


maikelvdh:
Similar as #11520, but then for XcodeDeps generator.

memsharded:
Closed by https://github.com/conan-io/conan/pull/11615

PR #11615, opened 2022-07-12
Add traits support for XcodeDeps

czoido:
Changelog: Feature: Add traits support for XcodeDeps.
Changelog: Bugfix: Fix frameworkdirs not taken into account in XcodeDeps generator.
Docs: TODO

Closes: https://github.com/conan-io/conan/issues/11555


conan/tools/apple/xcodedeps.py
```diff
@@ -122,22 +122,33 @@ def generate(self):
...
+
         fields = {
             'pkg_name': pkg_name,
             'comp_name': comp_name,
             'include_dirs': " ".join('"{}"'.format(p) for p in _merged_vars("includedirs")),
             'lib_dirs': " ".join('"{}"'.format(p) for p in _merged_vars("libdirs")),
             'libs': " ".join("-l{}".format(lib) for lib in _merged_vars("libs")),
             'system_libs': " ".join("-l{}".format(sys_lib) for sys_lib in _merged_vars("system_libs")),
-            'frameworksdirs': " ".join('"{}"'.format(p) for p in _merged_vars("frameworkdirs")),
+            'frameworkdirs': " ".join('"{}"'.format(p) for p in _merged_vars("frameworkdirs")),
```
czoido: This was a bug, `frameworkdirs` were never added to xcconfig files

czoido: PR to develop to fix: https://github.com/conan-io/conan/pull/11617

conan/tools/apple/xcodedeps.py
```diff
@@ -147,6 +147,16 @@ def _merged_vars(name):
             'condition': _xcconfig_conditional(self._conanfile.settings)
         }
 
+        if not require.headers:
```
memsharded, line 150: Please not that CMakeDeps is doing:
```python
 if require and not require.libs and not require.headers:
            self.defines = ""
            self.compile_definitions = ""
            self.cxxflags_list = ""
            self.cflags_list = ""
            self.sharedlinkflags_list = ""
            self.exelinkflags_list = ""
            self.objects_list = ""
```
So if there are no headers and libs at all, it is also dropping all definitions and flags.
Maybe we want to discuss this, not sure what is the best approach, but it should be consistent.

memsharded (commented):
Looks good, but there is a point we should discuss to align.

conan/tools/apple/xcodedeps.py
czoido: Yes, I had that code before, but then I realized that when the condition `if not require.libs and not require.headers` is met, no files are generated at all because the dependency is not passed in the loop that iterates over the dependencies, so I thought it made not sense... I'm currently testing that `if not require.libs and not require.headers` happens, no _xcconfig_ files are generated at all.

memsharded: Maybe still necessary to discuss a bit, to consider other cases. The full logic for skipping binaries is:
```python
if require.headers or require.libs or require.run or require.build:
    # do not skip it
```

So build_requires or something that has the ``run=True`` trait, can contribute things. For XcodeDeps, ``build=True`` might not be an issue, but ``run=True`` maybe is? (same for other generators)

czoido: Yes, then I'll put that code back better and consider the case

czoido: Done

memsharded (approved):

PR #11618, opened 2022-07-12
[MesonToolchain] Adding default dirs

### LINKED ISSUE #9713: [feature] do we need MesonLayout?

### LINKED ISSUE #11596: [bug] MesonToolchain: Doesn't configure the default install destination directories as expected
  * Operating System+version: Fedora Linux 36
  * Compiler+version: GCC 12
  * Conan version: 1.50.0
  * Python version: 3.10.5

Using the MesonToolchain, create any package on a RHEL system which uses `lib64` instead of `lib` or which packages anything in the default `datadir`, `mandir`, `localedir`, or `infodir`.

Many Conan recipes expect recipes to install libraries under the `lib` directory and under the `res` directory during the `package` step. MesonToolchain and/or Meson could configure these defaults with how Conan expects these directories to be organized, but it doesn't do that at the moment.

Things may be installed under `lib64` on RHEL systems and many things end up in the `share` directory.

Right now, the workaround I use is to configure `MesonToolchain` with `project_options` to set `libdir`, `datadir`, `localedir`, `mandir`, and `infodir` as appropriate.

I had to open the PR conan-io/conan-center-index#11347 to fix man pages being installed to an incorrect location in the package method.
See conan-io/conan-center-index#11610 and conan-io/conan-center-index#11580 for instances where this has been necessary.

franramirez688:
Changelog: Feature: Adding default directories to MesonToolchain.
Docs: https://github.com/conan-io/docs/pull/2652

Closes: https://github.com/conan-io/conan/issues/9713 
Closes: https://github.com/conan-io/conan/issues/11596

**Note**: Using the `basic_layout` seems to be enough for now.

### DIFF
--- a/conan/tools/meson/meson.py
+++ b/conan/tools/meson/meson.py
@@ -54,6 +54,8 @@ class Meson(object):
         self._conanfile.run(cmd)
 
     def test(self):
+        if self._conanfile.conf.get("tools.build:skip_test"):
+            return
         meson_build_folder = self._conanfile.build_folder
         cmd = 'meson test -v -C "{}"'.format(meson_build_folder)
         # TODO: Do we need vcvars for test?
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -126,6 +126,9 @@ class MesonToolchain(object):
         self.project_options = {
             "wrap_mode": "nofallback"  # https://github.com/conan-io/conan/issues/10671
         }
+        # Add all the default dirs
+        self.project_options.update(self._get_default_dirs())
+
         self.preprocessor_definitions = {}
         self.pkg_config_path = self._conanfile.generators_folder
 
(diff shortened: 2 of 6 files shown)


conan/tools/meson/meson.py
```diff
@@ -45,6 +45,8 @@ def install(self):
         self._conanfile.run(cmd)
 
     def test(self):
+        if self._conanfile.conf.get("tools.build:skip_test"):
+            return
```
franramirez688, line 58: Missing these lines from new Meson build-helper

SSE4:
it would be helpful to add some test/example on how to use `basic_layout` together with `MesonToolchain`. AFAIK existing meson tests don't create an instance of `basic_layout`...

conan/tools/meson/toolchain.py
```diff
@@ -164,6 +167,45 @@ def __init__(self, conanfile, backend=None):
...
+            ret.update({
+                'datadir': datadir,
+                'localedir': datadir,
+                'mandir': datadir,
+                'infodir': datadir
+            })
+        if includedir:
+            ret["includedir"] = includedir
+        if libdir:
+            ret["libdir"] = libdir
```
SSE4, line 230: at least one test (`conans\test\functional\toolchains\meson\test_install.py`) uses hard-coded `tc.project_options["libdir"] = "lib"`. it would be nice to check this test works without hard-coded value now.

franramirez688: yes, sure. I put it as a draft like I did not add any tests yet

franramirez688:
> AFAIK existing meson tests don't create an instance of basic_layout...

The `meson_lib` template is using that layout and we already have this functional test `test_meson_lib_template` using it. Should it be enough? Perhaps I could add another one if you think that is really needed.

SSE4:
> The `meson_lib` template is using that layout and we already have this functional test `test_meson_lib_template` using it. Should it be enough? Perhaps I could add another one if you want is really needed.

yeah, as I have figured it out from https://docs.conan.io/en/latest/reference/conanfile/tools/layout.html?highlight=basic_layout#basic-layout:
```
from conan.tools.layout import basic_layout

def layout(self):
    basic_layout(self)
```
and `conanfile_sources_v2` does exactly that.

SSE4 (approved):

conans/test/functional/toolchains/meson/test_meson.py
```diff
@@ -94,3 +98,45 @@ def test_build(self):
...
+        project('tutorial', 'cpp')
+        # Creating specific library
+        hello = library('hello', 'src/hello.cpp', install: true)
+        # Creating specific executable
+        executable('demo', 'src/main.cpp', link_with: hello, install: true)
+        # Creating specific data in src/ (they're going to be exported)
+        install_data(['src/file1.txt', 'src/file2.txt'])
+        """)
+        # Replace meson.build
+        self.t.save({"meson.build": _meson_build,
```
memsharded: I think if so much replace of the initial template is going to happen, then maybe it is worth to define things here?

I am also happy if we add the ``install`` bit to the ``meson`` templates, for example, adding a License, and making sure that the install is packaging the license too?

memsharded (approved):
Looking good!

memsharded (approved):

PR #11624, opened 2022-07-12
fixing missing parameter in api

### LINKED ISSUE #11621: [bug] conan user authentication accidentally skips SSL certificate verification
Note: worst case scenario of this bug is that user authentication data is sent in clear text to a man-in-the-middle.

  * Operating System+version: doesnt matter
  * Compiler+version: doesnt matter
  * Conan version: 1.44.1 (but current code in github seems to have the same issue)
  * Python version: doesnt matter

There is a bug in the instantiation of [RestV1Methods](https://github.com/conan-io/conan/blob/fae0b53bbf729e66ca8b2c3a594dd767ab32325f/conans/client/rest/rest_client_v1.py#L33) which again inherits the constructor from [RestCommonMethods](https://github.com/conan-io/conan/blob/fae0b53bbf729e66ca8b2c3a594dd767ab32325f/conans/client/rest/rest_client_common.py#L63).

The constructor expects
`(...  output, requester, config, verify_ssl ...)`
(note the **`config`** parameter)
parameters, whereas [the calling code in RestApiClient.authenticate](https://github.com/conan-io/conan/blob/fae0b53bbf729e66ca8b2c3a594dd767ab32325f/conans/client/rest/rest_client.py#L113) uses 
`(... self._output, self._requester, self._verify_ssl ...)`

### Logs
I discovered this while debugging some certificate issues ... no log available.

memsharded:
Changelog: Omit
Docs: Omit

Fix https://github.com/conan-io/conan/issues/11621

### DIFF
--- a/conans/client/rest/rest_client.py
+++ b/conans/client/rest/rest_client.py
@@ -111,7 +111,8 @@ class RestApiClient(object):
 
     def authenticate(self, user, password):
         api_v1 = RestV1Methods(self._remote_url, self._token, self._custom_headers, self._output,
-                               self._requester, self._verify_ssl, self._artifacts_properties)
+                               self._requester, self._config, self._verify_ssl,
+                               self._artifacts_properties)
 
         if self._refresh_token and self._token:
             token, refresh_token = api_v1.refresh_token(self._token, self._refresh_token)


lasote (approved):
About porting it to 1.50, it is worthy to mention that this is broken since 2/1/2020
https://github.com/conan-io/conan/pull/6271

Issue #10857, opened 2022-03-23
[feature] Support `QUIET` option when `find_package` config files generated by `CMakeDeps`.


hwhsu1231:
According to the [CMake](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) documentation, we can use `QUIET` option in `find_package`. 

> The `QUIET` option disables informational messages, including those indicating that the package cannot be found if it is not `REQUIRED`.

Therefore, the output messages won't be too much. However, it seems that those messages **still exist** even though I add `QUIET` option in `find_package`.

According to the [CMake](https://cmake.org/cmake/help/latest/command/find_package.html#package-file-interface-variables) documentation, it seems that the solution is to add `<PackageName>_FIND_QUIETLY` variable in Config files.

> `<PackageName>_FIND_QUIETLY`
> &emsp;&emsp;True if `QUIET` option was given

## Screenshots

![image](https://user-images.githubusercontent.com/71438617/159626216-486e88d3-540c-40c5-9749-e003bc3cbc5c.png)

## Output Messages

<details><summary>Click to expand output message of find_package</summary>

```
[cmake] -- Conan: Component target declared 'Qt6::Core'
[cmake] -- Conan: Component target declared 'Qt6::Platform'
[cmake] -- Conan: Component target declared 'Qt6::Gui'
[cmake] -- Conan: Component target declared 'Qt6::QWindowsIntegrationPlugin'
[cmake] -- Conan: Component target declared 'Qt6::QSQLiteDriverPlugin'
[cmake] -- Conan: Component target declared 'Qt6::QPSQLDriverPlugin'
[cmake] -- Conan: Component target declared 'Qt6::Network'
[cmake] -- Conan: Component target declared 'Qt6::Sql'
[cmake] -- Conan: Component target declared 'Qt6::Test'
[cmake] -- Conan: Component target declared 'Qt6::Widgets'
[cmake] -- Conan: Component target declared 'Qt6::PrintSupport'
[cmake] -- Conan: Component target declared 'Qt6::OpenGL'
[cmake] -- Conan: Component target declared 'Qt6::OpenGLWidgets'
[cmake] -- Conan: Component target declared 'Qt6::Concurrent'
[cmake] -- Conan: Component target declared 'Qt6::Xml'
[cmake] -- Conan: Component target declared 'Qt6::LinguistTools'
[cmake] -- Conan: Component target declared 'Qt6::UiPlugin'
[...]

Issue #9959, opened 2021-11-05
[bug] CMake Generator QUIET behaviour and message types.


n00b42:
1. According to [CMake](https://cmake.org/cmake/help/latest/command/find_package.html) `find_package(<pkg> QUIET)` should not print (informational) message (unless it could not be found...). However both cmake_find_package as well as CMakeDeps generators still print output even on success.  
E.g. `find_package(libsystemd REQUIRED QUIET)` results in:
~~~
-- Conan: Target declared 'libsystemd::libsystemd'
-- Conan: Component target declared 'libcap::cap'
-- Conan: Target declared 'libcap::libcap'
-- Conan: Target declared 'libmount::libmount'
[...]
~~~

I found CONAN_CMAKE_SILENT_OUTPUT which silences the output, but my question is if "QUIET" should not already do that.

2. Looking into the corresponding generated cmake files, I found something else: it looks like the message type is dropped, is this intentional?
~~~
 function(conan_message MESSAGE_OUTPUT)
     if(NOT CONAN_CMAKE_SILENT_OUTPUT)
         message(${ARGV${0}})
~~~
I.e. `conan_message(FATAL_ERROR [...]` will result in a `message([...]`

### Environment Details (include every applicable attribute)
  * Operating System+version: debian/testing (Debian 5.14.9-2 (2021-10-03) x86_64 GNU/Linux)
  * Compiler+version: gcc 10.3.0
  * Conan version: 1.42.0
  * Python version: 3.9.7

PR #11625, opened 2022-07-12
remove conan_message

### LINKED ISSUE #11369: [bug] CMakeDeps: CONAN_CMAKE_SILENT_OUTPUT ignores FATAL_ERROR messages
When using CMakeDeps we can set `CONAN_CMAKE_SILENT_OUTPUT` to silence the output from all calls to `conan_message`.  As of Conan 1.45.0, this will also mean that those messages with side effects (ie. `MESSAGE_TYPE` of `FATAL_ERROR` or `SEND_ERROR`) will not execute either.

It should probably be something like:
```cmake
function(conan_message MESSAGE_TYPE MESSAGE_CONTENT)
	if(NOT CONAN_CMAKE_SILENT_OUTPUT OR MESSAGE_TYPE STREQUAL "FATAL_ERROR" OR MESSAGE_TYPE STREQUAL "SEND_ERROR")
		message(${MESSAGE_TYPE} "${MESSAGE_CONTENT}")
	endif()
endfunction()
```

memsharded:
Changelog: Fix: Remove ``conan_message()`` in ``CMakeDeps``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/11369

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -66,7 +66,7 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         foreach(_COMPONENT {{ '${' + pkg_name + '_COMPONENT_NAMES' + '}' }} )
             if(NOT TARGET ${_COMPONENT})
                 add_library(${_COMPONENT} INTERFACE IMPORTED)
-                message(VERBOSE "Conan: Component target declared '${_COMPONENT}'")
+                message(STATUS "Conan: Component target declared '${_COMPONENT}'")
             endif()
         endforeach()
 
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -66,13 +66,13 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         foreach(_COMPONENT {{ '${' + pkg_name + '_COMPONENT_NAMES' + '}' }} )
             if(NOT TARGET ${_COMPONENT})
                 add_library(${_COMPONENT} INTERFACE IMPORTED)
-                conan_message(STATUS "Conan: Component target declared '${_COMPONENT}'")
+                message(VERBOSE "Conan: Component target declared '${_COMPONENT}'")
             endif()
         endforeach()
 
         if(NOT TARGET {{ root_target_name }})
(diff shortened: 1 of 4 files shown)


lasote (approved):

conan/tools/cmake/cmakedeps/templates/targets.py
```diff
@@ -66,13 +66,13 @@ def template(self):
         foreach(_COMPONENT {{ '${' + pkg_name + '_COMPONENT_NAMES' + '}' }} )
             if(NOT TARGET ${_COMPONENT})
                 add_library(${_COMPONENT} INTERFACE IMPORTED)
-                conan_message(STATUS "Conan: Component target declared '${_COMPONENT}'")
+                message(STATUS "Conan: Component target declared '${_COMPONENT}'")
```
praetorian20, line 69: Would you be opposed to changing this and line 75 to `VERBOSE` instead of `STATUS`?

I have a project where several submodules import conan packages (the super project is only setting up CMake paths to all the conan packages, not importing them) and these messages get printed several times now. I was previously silencing them by setting `CONAN_CMAKE_SILENT_OUTPUT`.

memsharded: The thing is that this information is generally super helpful for many users, that will get the targets listed by default. Because not listing them makes challenging to know what would be the available target names, and changing the default verbosity of CMake for this is not that straightforward (or users won't really run in verbose, because they don't know that Conan will show the targets if they do that). So I am concerned about the experience for many Conan users, specially learners.

praetorian20: I was worried you might say that :)

How about checking `<PackageName>_FIND_QUIETLY` within the generated code and if it's set either disable the message completely or change it to `VERBOSE`? That would give users control over the printing through the `find_package` call. I've never tried using this variable before, so unsure if it will actually work, but I could give it a shot and upload a PR.

I found related bug/feature requests too while Googling this, #10857 and #9959

memsharded: Yes, the current implementation do not implement QUIET handling, and it would make sense that it is listened. If you want to try it, these would be the guidelines:
- CMakeDeps only. It is not worth the effort of modifying the other legacy ones, Conan 2.0 is next month
- Don't overengineer with too much abstraction. A bit of repetition is not that bad if makes things cleaner.

Issue #11521, opened 2022-06-24
[develop2] Think new codebase layout


memsharded:
We're maintaining `develop` and `develop2` branches and the conflicts are huge, so we'll have to keep it till we freeze 1.x and merge `develop2` -> `develop`. After that, we'll update all the imports from docs.

_Originally posted by @franramirez688 in https://github.com/conan-io/docs/pull/2607#discussion_r905832887_

memsharded:
We need to move the public parts, like ``conans.cli`` to final ``conan.xxx`` namespace. So we need to outline the final structure. Not change it yet completely (conflicts develop->develop2), but maybe we can move ``conans.cli`` for a start.

SSE4:
how about re-importing via a namespace alias, e.g. `conan.cli` -> `conans.cli`?

lasote:
Tasks:

- Move `cli` to `conan`
- Remove `output` from cli
- Asset from `api/helpers/` to `_internals`
- conan.py and setup.py try to move to cli

memsharded:
Closed by #11630

PR #11630, opened 2022-07-13
2.0 project codebase layout: Moved cli, extracted output, moved assets

lasote:
Close https://github.com/conan-io/conan/issues/11521


czoido (approved):

franramirez688 (approved):
Should we keep more or less the same structure as the public one for `_internal` when the modules are related? What I mean is `from conan._internal.api.new.[template]` -> `from conan._internal.api.subapi.new.[template]`

conan/_internal/api/new/autoools_exe.py
```diff
@@ -1,4 +1,4 @@
-from conan.api.helpers.new.autotools_lib import configure_ac, makefile_am
+from conan._internal.api.new.autotools_lib import configure_ac, makefile_am
```
memsharded: I am not loving that this is highlighted by linters and IDE as a warning of importing something internal.
Is there any workaround for this warning? Maybe relative imports? but those are also not recommended?

conan/api/subapi/download.py
```diff
@@ -1,7 +1,7 @@
 from conan.api.model import Remote
 from conan.api.subapi import api_method
-from conans.cli.conan_app import ConanApp
-from conans.cli.output import ConanOutput
+from conan.cli.command import ConanOutput
```
memsharded: But then the ``ConanOutput`` is still inside ``cli``. Isn't this ``from conan.cli`` import from the ``api`` that we want to avoid? It is true that the ``ConanApp``, maybe it belong below ``api``?

lasote:
> Should we keep more or less the same structure as the public one for _internal when the modules are related? What I mean is from conan._internal.api.new.[template] -> from conan._internal.api.subapi.new.[template]

@franramirez688 🤷🏼 I don't think the internal module will follow exactly any structure, I bet much of the "conans" workspace will end up there so for me it is not about having a 1->1 with the api.

conan/_internal/api/new/autoools_exe.py
lasote: I can't find anything but double underscore `__internal` that causes no issues with the linters. I can't find any documentation about that double underscore in module names so apparently, it is meaningless.

memsharded (changes requested):
Lets change to ``from conan.internal.xxx`` it is clean for IDEs and linter, and enough warning for users that might be tempted to import from it.

memsharded (approved):

PR #11632, opened 2022-07-13
[MesonToolchain] Added Objective-C/C++ arguments

### LINKED ISSUE #11628: [MesonToolchain] Add Objective-C/C++ arguments
@franramirez688 this seems to be a line in `GLib` triggering the failure:
```
add_languages('objc')
```
using [add_languages](https://mesonbuild.com/Reference-manual_functions.html#add_languages) (I believe `objcpp` is another valid value here).
we can start by adding a test using it and ensuring it passes for M1 cross-build.

_Originally posted by @SSE4 in https://github.com/conan-io/conan-center-index/issues/11613#issuecomment-1182967678_

franramirez688:
Changelog: Feature: Added `objc_args`/`objc_link_args`/`objcpp_args`/`objcpp_link_args` to MesonToolchain for Objective-C/C++ language.
Closes: https://github.com/conan-io/conan/issues/11628
Docs: https://github.com/conan-io/docs/pull/2652

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -40,6 +40,8 @@ class MesonToolchain(object):
     [binaries]
     {% if c %}c = '{{c}}'{% endif %}
     {% if cpp %}cpp = '{{cpp}}'{% endif %}
+    {% if objc %}objc = '{{objc}}'{% endif %}
+    {% if objcpp %}objcpp = '{{objcpp}}'{% endif %}
     {% if c_ld %}c_ld = '{{c_ld}}'{% endif %}
     {% if cpp_ld %}cpp_ld = '{{cpp_ld}}'{% endif %}
     {% if ar %}ar = '{{ar}}'{% endif %}
@@ -57,11 +59,17 @@ class MesonToolchain(object):
     {% if b_staticpic %}b_staticpic = {{b_staticpic}}{% endif %}
     {% if cpp_std %}cpp_std = '{{cpp_std}}' {% endif %}
     {% if backend %}backend = '{{backend}}' {% endif %}
+    {% if pkg_config_path %}pkg_config_path = '{{pkg_config_path}}'{% endif %}
+    # C/C++ arguments
     c_args = {{c_args}} + preprocessor_definitions + deps_c_args
     c_link_args = {{c_link_args}} + deps_c_link_args
     cpp_args = {{cpp_args}} + preprocessor_definitions + deps_cpp_args
     cpp_link_args = {{cpp_link_args}} + deps_cpp_link_args
-    {% if pkg_config_path %}pkg_config_path = '{{pkg_config_path}}'{% endif %}
+    # Objective-C/C++ arguments
+    objc_args = {{objc_args}} + preprocessor_definitions + deps_c_args
+    objc_link_args = {{objc_link_args}} + deps_c_link_args
+    objcpp_args = {{objcpp_args}} + preprocessor_definitions + deps_cpp_args
+    objcpp_link_args = {{objcpp_link_args}} + deps_cpp_link_args
 
     {% for context, values in cross_build.items() %}
     [{{context}}_machine]
@@ -160,6 +168,13 @@ class MesonToolchain(object):
         self.c_link_args = self._get_env_list(build_env.get("LDFLAGS", []))
         self.cpp_args = self._get_env_list(build_env.get("CXXFLAGS", []))
         self.cpp_link_args = self._get_env_list(build_env.get("LDFLAGS", []))
+        # Objective C/C++ ones
+        self.objc = "clang"
+        self.objcpp = "clang++"
+        self.objc_args = self._get_env_list(build_env.get('OBJCFLAGS', []))
+        self.objc_link_args = self._get_env_list(build_env.get('LDFLAGS', []))
+        self.objcpp_args = self._get_env_list(build_env.get('OBJCXXFLAGS', []))
+        self.objcpp_link_args = self._get_env_list(build_env.get('LDFLAGS', []))
 
         # Apple flags
         self.apple_arch_flag = []
@@ -238,6 +253,12 @@ class MesonToolchain(object):
         self.cpp_args.extend(apple_flags + extra_flags["cxxflags"])
         self.c_link_args.extend(apple_flags + extra_flags["ldflags"])
         self.cpp_link_args.extend(apple_flags + extra_flags["ldflags"])
+        # Objective C/C++
+        self.objc_args.extend(self.c_args)
+        self.objcpp_args.extend(self.cpp_args)
+        # These link_args have already the LDFLAGS env value so let's add only the new possible ones
+        self.objc_link_args.extend(apple_flags + extra_flags["ldflags"])
+        self.objcpp_link_args.extend(apple_flags + extra_flags["ldflags"])
 
         return {
             # https://mesonbuild.com/Machine-files.html#properties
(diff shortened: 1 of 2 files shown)


franramirez688:
I want to add another test using Objective-C++ (typical hello.lib and demo app linking with it)

SSE4 (approved):
this is outstanding PR, it looks fantastic!

conan/tools/meson/toolchain.py
```diff
@@ -160,6 +168,13 @@ def __init__(self, conanfile, backend=None):
         self.c_link_args = self._get_env_list(build_env.get("LDFLAGS", []))
         self.cpp_args = self._get_env_list(build_env.get("CXXFLAGS", []))
         self.cpp_link_args = self._get_env_list(build_env.get("LDFLAGS", []))
+        # Objective C/C++ ones
+        self.objc = "clang"
```
memsharded: Is this hardcoded? Is objectiveC compiler always clang? 
Maybe this should only be enabled for Apple systems?

conans/test/functional/toolchains/meson/test_meson_and_objc.py
```diff
@@ -0,0 +1,110 @@
...
+    os = {os}
+    os.version = {os_version}
+    os.sdk = {os_sdk}
+    arch = {arch}
+    compiler = apple-clang
+    compiler.version = 12.0
+    compiler.libcxx = libc++
+
+    [buildenv]
+    LDFLAGS=-framework Foundation
```
memsharded: This sounds it belongs to the build script, and shouldn't be injected?

conans/test/functional/toolchains/meson/test_meson_and_objc.py
```diff
@@ -0,0 +1,110 @@
...
+    t = TestClient()
+    t.save({"conanfile.py": _conanfile_py,
+            "meson.build": _meson_build_objc,
+            "main.m": app,
+            "profile_host": profile})
+
+    t.run("install . --profile:build=default --profile:host=profile_host")
+    t.run("build .")
+    assert "Objective-C compiler for the host machine: clang" in t.out
+    assert "Objective-C compiler for the build machine: cc" in t.out
```
memsharded: What does this mean? Isnt the compiler always clang as defined above?

conans/test/functional/toolchains/meson/test_meson_and_objc.py
```diff
@@ -0,0 +1,110 @@
...
+
+    demo = os.path.join(t.current_folder, "build", "demo")
+    assert os.path.isfile(demo) is True
```
memsharded, line 137: Isn't it possible to run the application? That would be nice.

memsharded (commented):
This is looking good, though I wonder a bit about objectiveC support in general:
- Shouldn't it be its own compiler? Are there packages mixed with C/C++ and objC? Only ObjC? 
- What happens with the binary model for objC?
- What happens with other build systems like CMake for ObjC?

conans/test/functional/toolchains/meson/test_meson_and_objc.py
SSE4: I doubt as it's a cross-building use-case

conan/tools/meson/toolchain.py
SSE4: I think nowadays it's pretty much limited to Apple ecosytem, and almost never used outside of it
the only wide-spread and production ready Objective-C implementation is Clang
there was a GNUStep project, but I would say it's exotic, and problematic (it doesn't implement all the frameworks, so it's incomplete and useless)

conans/test/functional/toolchains/meson/test_meson_and_objc.py
SSE4: here we probably care only about host compiler from the host profile
if we need to build something for build profile, it will be build requires, and information will be read from another profile (build)

SSE4:
> * Shouldn't it be its own compiler? 

in which context?
if you mean compiler executable, there are just `clang` and `clang++`. there are no dedicated executables to compile Objective-C[++] code. it's basically controlled by flags (e.g. `-x objective-c`)
if you mean compiler as a suite, it's part of LLVM/Clang package.
if you mean compiler as a conan setting, it is just `apple-clang` entry.

> * Are there packages mixed with C/C++ and objC? Only ObjC?

there are MANY packages mixing C/C++ with Objective-C/Objective-C++.
Objective-C[++] basically acts as a glue for al the modern Apple APIs - there is just no way to invoke these APIs from C/C++.

> * What happens with the binary model for objC?

it's basically follows pure C model more or less. we didn't need to complicate it in the past.

> * What happens with other build systems like CMake for ObjC?

CMake supports Objective-C[++] out of the box. other build systems (e.g. autotools) as well. probably some rare used might not support it, but it has nothing to do with conan.

conans/test/functional/toolchains/meson/test_meson_and_objc.py
memsharded: Isn't it possible to build objC for native OSX too?

conan/tools/meson/toolchain.py
memsharded: So you agree that this should be limited to Apple systems only?

Are we talking about apple-clang or clang compiler? What is the connection to the ``settings.compiler``?

conans/test/functional/toolchains/meson/test_meson_and_objc.py
SSE4: ofcourse it is possible. but this test and issue was about cross-building.

conan/tools/meson/toolchain.py
SSE4: I am talking about `apple-clang`
I agree it might be limited for apple systems

franramirez688: I agree with that.

memsharded:
Nice, thanks for the explanations, @SSE4, I think now it is a bit more clear, and can be moved forward.
Maybe just conditioning on the "apple" platform is enough @franramirez688, and it can be marked as ready for review

franramirez688:
> I want to add another test using Objective-C++ (typical hello.lib and demo app linking with it)

For now, I'm not adding it because of the complexity of the code. It seems that a simple Objective-C++ (Objective-C and C++) project must have a kind of wrapper in the middle of C++ and Objective-C (see more information [here](https://stackoverflow.com/questions/19229574/calling-c-from-objective-c)). I guess it's enough with those tests to verify it's working.

conan/tools/meson/toolchain.py
```diff
@@ -187,6 +206,13 @@ def _resolve_apple_flags(self):
         self.apple_arch_flag = ["-arch", arch] if arch else []
         self.apple_isysroot_flag = ["-isysroot", sdk_path] if sdk_path else []
         self.apple_min_version_flag = [apple_min_version_flag(self._conanfile)]
+        # Objective C/C++ ones
+        self.objc = "clang"
+        self.objcpp = "clang++"
+        self.objc_args = self._get_env_list(build_env.get('OBJCFLAGS', []))
```
memsharded, line 212: I still feel this is problematic, having 2 mechanisms to define CXX_FLAGS (env and conf). Not related to this PR, I know it is already this way for C/C++ flags, but something we want to clarify for the future.

memsharded (approved):

conan/tools/meson/toolchain.py
franramirez688: I absolutely agree. I can open a "look-into" feature to have this into account for future releases.

PR #11643, opened 2022-07-14
Available folders in post_export hook

lasote:
Changelog: Fix: The `conanfile.export_folder` and `conanfile.export_sources_folder` are now available in the `post_hook()`.
Docs: omit

### DIFF
--- a/conans/test/integration/hooks/hook_test.py
+++ b/conans/test/integration/hooks/hook_test.py
@@ -11,33 +11,51 @@ import os
 
 def pre_export(conanfile):
     conanfile.output.info("Hello")
+    # TODO: To have the export_folder here needs a bit more deep refactoring
+    assert conanfile.export_folder is None
 
 def post_export(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.export_folder, "export_folder not defined"
+    assert conanfile.export_sources_folder, "export_sources_folder not defined"
 
 def pre_source(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.source_folder, "source_folder not defined"
 
 def post_source(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.source_folder, "source_folder not defined"
 
 def pre_generate(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.generators_folder, "generators_folder not defined"
 
 def post_generate(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.generators_folder, "generators_folder not defined"
 
 def pre_build(conanfile):
     conanfile.output.info("Hello")
+    assert conanfile.source_folder, "source_folder not defined"
+    assert conanfile.build_folder, "build_folder not defined"
 
 def post_build(conanfile):
     conanfile.output.info("Hello")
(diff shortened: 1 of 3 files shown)


conans/test/integration/layout/export_folder_variable_test.py
```diff
@@ -57,7 +57,8 @@ def generate(self):
                 assert os.path.exists(self.export_sources_folder)
 
             def export_sources(self):
-                assert self.export_folder is None
+                # We need it available for the post_export hook so it is available
+                assert os.path.exists(self.export_folder)
```
memsharded, line 61: Agree that this was just a defensive feature, I am fine with changing this, no big problem (still shouldn't be documented or used that ``self.export_folder`` can be used in ``exports_sources()`` methods.

memsharded (approved):

conans/test/integration/layout/export_folder_variable_test.py
```diff
@@ -57,7 +57,8 @@ def generate(self):
                 assert os.path.exists(self.export_sources_folder)
 
             def export_sources(self):
-                assert self.export_folder is None
+                # We need it available for the post_export hook so it is available
```
lasote, line 60: This is not so good, it can be limited with a bit more code invalidating temporarily the export_folder during the export_source() call. WDYT?

conans/test/integration/layout/export_folder_variable_test.py
```diff
@@ -97,7 +98,8 @@ def export(self):
                 assert os.path.exists(self.export_folder)
 
             def export_sources(self):
-                assert self.export_folder is None
+                # We need it available for the post_export hook so it is available
```
lasote, line 101: Idem

conans/test/integration/hooks/hook_test.py
```diff
@@ -11,33 +11,53 @@
 
 def pre_export(conanfile):
     conanfile.output.info("Hello")
+    # TODO: To have the export_folder here needs a bit more deep refactoring
```
lasote, line 14: How bad is this? how is the pre_export supposed to work? it might need also the export_folder, right?

conans/test/integration/layout/export_folder_variable_test.py
lasote: ok, you already answered.

conans/test/integration/layout/export_folder_variable_test.py
lasote: ok

conans/test/integration/hooks/hook_test.py
memsharded: ``pre_export`` hook probably shouldn't have access to the export folders? If it happens before the actual export, those folders are not even created. Maybe ``pre_export`` is intended to operate mostly in the user space?

SSE4: why not? I might need to export some files from hook before conan exports from recipe

memsharded: some ConanCenter example? And some example that cannot be done in the post_hook?

memsharded (approved):

PR #11655, opened 2022-07-15
CMakePresets compatible with schema 2

### LINKED ISSUE #11648: [feature] Try to aggregate CMakePresets instead of including them
The`CMakeUserPresets` include different `CMakePreset` files from the generators folder, but the `include` functionality requires CMake >=3.23.
We can try to aggregate the contents directly maybe following a `conf`.

Related https://github.com/conan-io/conan/issues/11623

### LINKED ISSUE #11649: [feature] CMakePresets toolchain inclusion
The `toolchainFile` element requires CMake >=3.21, a cacheVariables with CMAKE_TOOLCHAIN_FILE could be used instead.

Might be opt-in with the same conf used for https://github.com/conan-io/conan/issues/11648

Related https://github.com/conan-io/conan/issues/11623

lasote:
Changelog: Feature: Introduced a new conf `tools.cmake.cmaketoolchain.presets:max_schema_version` to generate `CMakePresets.json` and `CMakeUserPresets.json` compatible with the specified json schema version. The minimum value supported is `>=2`.
Docs: https://github.com/conan-io/docs/pull/2666

To discuss: Is this an opt-in? or maybe an opt-out to use the "include" schema?

Close #11649
Close #11648

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -69,7 +69,8 @@ def _add_configure_preset(conanfile, generator, cache_variables, toolchain_file,
 
 
 def _forced_schema_2(conanfile):
-    version = conanfile.conf.get("tools.cmake.cmaketoolchain.presets:max_schema_version", check_type=int)
+    version = conanfile.conf.get("tools.cmake.cmaketoolchain.presets:max_schema_version",
+                                 check_type=int, default=4)
     if version < 2:
         raise ConanException("The minimun value for 'tools.cmake.cmaketoolchain.presets:"
                              "schema_version' is 2")
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -70,7 +70,10 @@ def _add_configure_preset(conanfile, generator, cache_variables, toolchain_file,
 
 def _forced_schema_2(conanfile):
     version = conanfile.conf.get("tools.cmake.cmaketoolchain.presets:max_schema_version",
-                                 check_type=int, default=4)
+                                 check_type=int)
+    if not version:
+        return False
+
     if version < 2:
         raise ConanException("The minimun value for 'tools.cmake.cmaketoolchain.presets:"
                              "schema_version' is 2")
@@ -88,6 +91,10 @@ def _schema_version(conanfile, default):
 
 
 def _contents(conanfile, toolchain_file, cache_variables, generator):
+    """
+    Contents for the CMakePresets.json
+    It uses schema version 3 unless it is forced to 2
+    """
     ret = {"version": _schema_version(conanfile, default=3),
            "cmakeMinimumRequired": {"major": 3, "minor": 15, "patch": 0},
            "configurePresets": [],
@@ -148,6 +155,10 @@ def save_cmake_user_presets(conanfile, preset_path):
     # Try to save the CMakeUserPresets.json if layout declared and CMakeLists.txt found
     if conanfile.source_folder and conanfile.source_folder != conanfile.generators_folder:
         if os.path.exists(os.path.join(conanfile.source_folder, "CMakeLists.txt")):
+            """
+                Contents for the CMakeUserPresets.json
+                It uses schema version 4 unless it is forced to 2
+            """
             user_presets_path = os.path.join(conanfile.source_folder, "CMakeUserPresets.json")
             if not os.path.exists(user_presets_path):
                 data = {"version": _schema_version(conanfile, default=4),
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/presets.py
```diff
@@ -63,8 +68,27 @@ def _add_configure_preset(conanfile, generator, cache_variables, toolchain_file,
     return ret
 
 
+def _forced_schema_2(conanfile):
+    version = conanfile.conf.get("tools.cmake.cmaketoolchain.presets:max_schema_version",
+                                 check_type=int, default=4)
+    if version < 2:
+        raise ConanException("The minimun value for 'tools.cmake.cmaketoolchain.presets:"
```
memsharded:
```suggestion
        raise ConanException("The minimum value for 'tools.cmake.cmaketoolchain.presets:"
```

conan/tools/cmake/presets.py
```diff
@@ -117,25 +141,48 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables):
...
 
+
+def save_cmake_user_presets(conanfile, preset_path):
     # Try to save the CMakeUserPresets.json if layout declared and CMakeLists.txt found
     if conanfile.source_folder and conanfile.source_folder != conanfile.generators_folder:
         if os.path.exists(os.path.join(conanfile.source_folder, "CMakeLists.txt")):
             user_presets_path = os.path.join(conanfile.source_folder, "CMakeUserPresets.json")
             if not os.path.exists(user_presets_path):
-                data = {"version": 4, "include": [preset_path], "vendor": {"conan": dict()}}
+                data = {"version": _schema_version(conanfile, default=4),
```
memsharded, line 164: This is confusing, the above uses a default of =3, and now a default=4. Not sure what it means.

conan/tools/cmake/presets.py
```diff
@@ -53,8 +53,13 @@ def _add_configure_preset(conanfile, generator, cache_variables, toolchain_file,
             "description": "'{}' configure using '{}' generator".format(name, generator),
             "generator": generator,
             "cacheVariables": cache_variables,
-            "toolchainFile": toolchain_file,
+
            }
+    if not _forced_schema_2(conanfile):
+        ret["toolchainFile"] = toolchain_file
+    else:
+        ret["cacheVariables"]["CMAKE_TOOLCHAIN_FILE"] = toolchain_file
```
memsharded, line 61: I'd say this is not being tested? Conan will extract for ``CMake`` call the toolchain file from the presets file, but this has changed, and no change in ``CMake`` class?

memsharded (commented):
This is looking good, though the ``_schema_version`` is a bit confusing.

conan/tools/cmake/presets.py
lasote: This is because the schema version of the `CMakePreset` is 3 but the version of the `CMakeUserPreset` is 4. 
Let me do it better.

conan/tools/cmake/presets.py
lasote: What do you mean? The `cacheVariables` were already being read by the `CMake` build helper.

conan/tools/cmake/presets.py
lasote: Bit better now?

conan/tools/cmake/presets.py
memsharded: Oh, yes, I see, this is managed automatically too, you are right.

conan/tools/cmake/presets.py
memsharded: It still seems a bit confusing, because if user provides ``schema=3`` in the conf, that will be ignored, and generate a ``schema=2`` in the json, right?

lasote: Well, the user provides a `tools.cmake.cmaketoolchain.presets:max_schema_version=3`, that won't be ignored and generates a `schema=2`.

lasote: I can convert the conf to `tools.cmake.cmaketoolchain.presets:v2_schema=True` if you prefer, but I thought that we might end with more similar configs in the future.

memsharded: Oh, I see, I overlooked it is ``max_schema_version``, not ``schema_version``. Then it makes a bit more sense, yeah.

memsharded (approved):

franramirez688 (approved):

czoido (approved):

PR #11662, opened 2022-07-18
[dev2] Components requires loop

### LINKED ISSUE #11549: [bug] Components cycles not detected
Conan 1.50 introduced a new message for loop detection in components in https://github.com/conan-io/conan/pull/11519

This had been removed in 2.0, so the test fails, it is:
```python
@pytest.mark.xfail(reason="The detection of cycles have been removed in 2.0, this test hangs")
def test_components_cycles():
```

The test hangs forever, we need to change the algorithm and raise the error message again

franramirez688:
Changelog: Bugfix: Bringing back the components requires detection loop code.
Closes: https://github.com/conan-io/conan/issues/11549
Docs: omit

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -440,7 +440,7 @@ class CppInfo(object):
                 req_processed = [n for n in c.required_component_names if n not in processed]
                 if not req_processed and name not in processed:
                     processed.append(name)
-            else:
+            if not processed:
                 # Detected cycle components requirements!
                 self._raise_circle_components_requires_error()
 
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -434,14 +434,15 @@ class CppInfo(object):
         processed = []  # Names of the components ordered
         # FIXME: Cache the sort
         while (len(self.components) - 1) > len(processed):
+            cached_processed = processed[:]
             for name, c in self.components.items():
                 if name is None:
                     continue
                 req_processed = [n for n in c.required_component_names if n not in processed]
                 if not req_processed and name not in processed:
                     processed.append(name)
-            if not processed:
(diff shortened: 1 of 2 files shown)


conans/model/build_info.py
```diff
@@ -421,6 +440,9 @@ def get_sorted_components(self):
                 req_processed = [n for n in c.required_component_names if n not in processed]
                 if not req_processed and name not in processed:
                     processed.append(name)
+            if not processed:
```
memsharded: I think this might not cover all loops. This might need an aggregated variable in the internal ``for`` loop, that checks that processed is actually growing in every iteration.

memsharded (changes requested):

conans/model/build_info.py
franramirez688: I have been checking the loop and, from my point of view, there is no evidence that we need to have another aggregated variable. Do you have any idea about a possible situation that it could fail?

franramirez688: Added one more test to check another kind of cycle

conans/test/integration/test_components.py
```diff
@@ -34,9 +34,52 @@ def test(self):
             """)
     c.save({"conanfile.py": conanfile,
             "test_package/conanfile.py": test_conanfile})
-    c.run("create .", assert_error=True)
-    print(c.out)
-    assert "ERROR: There is a dependency loop in 'self.cpp_info.components' requires:" in c.out
-    assert "a requires c"
-    assert "b requires a"
-    assert "c rquires b"
```
franramirez688, line 42: with the `c.run("******", assert_error=True)` those asserts are never reached. Apart from that, the last asserts were testing nothing.

memsharded (approved):

PR #11667, opened 2022-07-18
[bug] get_cpus is not supporting cgroup2 #11635

### LINKED ISSUE #11635: [bug] get_cpus is not supporting cgroup2
The current detection of https://github.com/conan-io/conan/blob/1.46.2/conans/client/tools/oss.py#L36-L39 seems no longer applicable in case of cgroup2 as the `/sys/fs/cgroup/cpu/cpu.cfs_quota_us` and `/sys/fs/cgroup/cpu/cpu.cfs_period_us` are replaced by `/sys/fs/cgroup/cpu.max` (in case of docker executed with `--cpus`).

There could be checked if cgroup2 is active for example by checking for existence of `/sys/fs/cgroup/cgroup.controllers` file and based on that the values for `quota` and `period` could be taken from `cpu.max` file.

MariaMozgunova:
Changelog: Fix: Get the cpu count for cgroup2 from `/sys/fs/cgroup/cpu.max`.
Docs: https://github.com/conan-io/docs/pull/2658

Fixes #11635

Gets the cpu count for cgroup2 from `/sys/fs/cgroup/cpu.max` file. From [docs](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html), `/sys/fs/cgroup/cpu.max` file has the following format: `$MAX $PERIOD`.
In case `$MAX` is "max", return 1.
If there is no `$PERIOD`, use only `$MAX` as the return value.
When both values are integers, use them as before.

### DIFF
--- a/conan/tools/build/cpu.py
+++ b/conan/tools/build/cpu.py
@@ -1,5 +1,6 @@
 import math
 import multiprocessing
+import os
 
 from conans.util.files import load
 
@@ -15,8 +16,18 @@ def _cpu_count():
     try:
         try:
             # This is necessary to deduce docker cpu_count
-            cfs_quota_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_quota_us"))
-            cfs_period_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_period_us"))
+            # cgroup2
+            if os.path.exists("/sys/fs/cgroup/cgroup.controllers"):
+                cpu_max = load("/sys/fs/cgroup/cpu.max").split()
+                if cpu_max[0] == "max":
+                    return 1
+                elif len(cpu_max) == 1:
+                    cfs_quota_us, cfs_period_us = int(cpu_max[0]), 1
+                else:
+                    cfs_quota_us, cfs_period_us = map(int, cpu_max)
+            else:  # cgroup1
+                cfs_quota_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_quota_us"))
+                cfs_period_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_period_us"))
             if cfs_quota_us > 0 and cfs_period_us > 0:
                 return int(math.ceil(cfs_quota_us / cfs_period_us))
         except (EnvironmentError, TypeError):
--- a/conan/tools/build/cpu.py
+++ b/conan/tools/build/cpu.py
@@ -16,15 +16,15 @@ def _cpu_count():
     try:
         try:
             # This is necessary to deduce docker cpu_count
+            cfs_quota_us = cfs_period_us = 0
             # cgroup2
             if os.path.exists("/sys/fs/cgroup/cgroup.controllers"):
                 cpu_max = load("/sys/fs/cgroup/cpu.max").split()
-                if cpu_max[0] == "max":
-                    return 1
-                elif len(cpu_max) == 1:
-                    cfs_quota_us, cfs_period_us = int(cpu_max[0]), 1
-                else:
-                    cfs_quota_us, cfs_period_us = map(int, cpu_max)
+                if cpu_max[0] != "max":
+                    if len(cpu_max) == 1:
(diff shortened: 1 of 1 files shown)


MariaMozgunova:
Should I document the changes to the Conan docs?

SSE4 (approved):

memsharded:
> Should I document the changes to the Conan docs?

Better wait until it is merged, and then yes, a PR to the docs, if necessary, would be nice, just a couple of lines if it makes sense

MariaMozgunova:
Do I need to do something for this PR to be merged?

memsharded (approved):

memsharded:
> Do I need to do something for this PR to be merged?

No, it is good, it just takes some time to review, too many ongoing things. It doesn't have tests, but honestly all tests are going to be pure mocks, so probably it is good.

Maybe @maikelvdh want to comment something more, if not, this will be merged and released in 1.51 (still a few days, Conan 2.0-beta.2 is in the next release)

conan/tools/build/cpu.py
```diff
@@ -15,8 +16,18 @@ def _cpu_count():
     try:
         try:
             # This is necessary to deduce docker cpu_count
-            cfs_quota_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_quota_us"))
-            cfs_period_us = int(load("/sys/fs/cgroup/cpu/cpu.cfs_period_us"))
+            # cgroup2
+            if os.path.exists("/sys/fs/cgroup/cgroup.controllers"):
+                cpu_max = load("/sys/fs/cgroup/cpu.max").split()
+                if cpu_max[0] == "max":
+                    return 1
```
maikelvdh: In case of `max` it means basically no limits are established and therefore we should be able to rely on the `multiprocessing.cpu_count()`. In the current logic you would otherwise effectively in case of docker when cgroup V2 present and without any limiting of CPU cores applied just run with 1 jobs always.

conan/tools/build/cpu.py
```diff
@@ -15,8 +16,18 @@ def _cpu_count():
...
+                elif len(cpu_max) == 1:
+                    cfs_quota_us, cfs_period_us = int(cpu_max[0]), 1
```
maikelvdh: To keep it in line with the actual official documentation and the microsecond notation
```suggestion
                    cfs_quota_us, cfs_period_us = int(cpu_max[0]), 100000
```

maikelvdh (changes requested):

conan/tools/build/cpu.py
memsharded: But is it not an issue that ``multiprocessing.cpu_count()`` is used? because it will not match the docker assigned cores, and it can also overflow the container capacity?

maikelvdh: That is exactly what you want in case of docker if you are not using the runtime options to limit the CPU (https://docs.docker.com/config/containers/resource_constraints/#cpu). In case you are using this options to limit you will get not `max` as value.

memsharded: Great, thanks very much for the clarification. @MariaMozgunova could you please apply this in your PR? Thanks!

maikelvdh:
Thanks a lot @MariaMozgunova for addressing this problem 🚀 !

memsharded:
Merged, thanks for your contribution! It will be part of 1.51

If you now want to do a PR to the docs, something very simple about the behavior for cgroups2, a couple of lines would be enough, that would be great.

MariaMozgunova:
@memsharded Can you please hint to me what file precisely should be modified?

uilianries:
@MariaMozgunova The documentation already describes cgroup usage, so with your hotfix, the follow part should updated:

https://github.com/conan-io/docs/blob/79feb32d0be4d531a363b8948b6d8a2289e059ab/reference/tools.rst#toolscpu_count

The website will look like this: https://docs.conan.io/en/latest/reference/tools.html#tools-cpu-count

Thank you for your contribution!

PR #11668, opened 2022-07-18
Fix Dnf and Yum to accept 100 as a return code

### LINKED ISSUE #11661: [bug] system package failure on dnf and yum
[alma linux](https://github.com/bincrafters/system-packages-checks/runs/7368984232?check_suite_focus=true#step:6:214), [centos](https://github.com/bincrafters/system-packages-checks/runs/7368984251?check_suite_focus=true#step:6:184) and [fedora](https://github.com/bincrafters/system-packages-checks/runs/7368984210?check_suite_focus=true#step:6:206) fail with `ERROR: while executing system_requirements(): Error 100 while executing yum check-update -y`. This error code signals that there are some update available, so it should be ignored (it was ignored in the previous system package handler)
cf https://github.com/conan-io/conan/blob/f9215513368e576a82f071ab5d7b4cf990d156ca/conans/client/tools/system_pm.py#L348 and https://github.com/conan-io/conan/blob/f9215513368e576a82f071ab5d7b4cf990d156ca/conans/client/tools/system_pm.py#L317

  * Operating System+version: linux
  * Compiler+version:
  * Conan version: 1.50
  * Python version:

czoido:
Changelog: Fix: Fix `Dnf` and `Yum` to accept 100 as a return code.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/11661

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -60,7 +60,7 @@ class _SystemPackageManagerTool(object):
         askpass = "-A " if self._sudo and self._sudo_askpass else ""
         return "{}{}".format(sudo, askpass)
 
-    def _run_method(self, method, *args, **kwargs):
+    def run(self, method, *args, **kwargs):
         if self._active_tool == self.__class__.tool_name:
             return method(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


memsharded (approved):

franramirez688 (approved):
All the tests from `conans/test/integration/tools/system/package_manager_test.py` are unit tests instead of integration ones. Even the tests marked as `@pytest.mark.skipif` are totally useless because they seem to be passing in all the OSs. I only mention it to keep it in mind.

Issue #11542, opened 2022-06-28
[feature] .conanrc to specify repo-local workflow


blackliner:
Coming from a discussion with @memsharded:

Conan currently does not support **implicit** configuration of the `conan` cache (templates, hooks, packages etc), but only explicitly by setting a env variable [CONAN_USER_HOME](https://docs.conan.io/en/latest/reference/env_vars.html?highlight=conan_user_home#conan-user-home).

This request is about introducing a `.conanrc` called file that will be placed at the root of a repository, and contains useful metainformation, like `local_user_home=True` etc. Any `conan` CLI call will end up traversing the `PWD` until it finds such an `.conanrc` file and use it.

This will allow us to make the repository the real source of truth, instead of relying on a central cache. This can be an issue for example if you are working on different projects that use different, conflicting hooks.

This idea is heavily inspired by bazel's [.bazelrc](https://bazel.build/docs/bazelrc), and can be expanded to other use-cases:
- define custom path to `.conan/config`, so it utilizes the repos filesystem instead of creating a new one, making `conan config install` calls obsolete. This would cover pofiles, hooks, remotes, templates...

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Lets do it in 2 steps:

- Recover the ``storage.path`` configuration in the home global.conf
- Define a .conanrc that contains env-var CONAN_HOME to define the location of the home (can be relative path to the .conanrc file, to define a local home per project)

memsharded:
Both features, the necessary storage-path configuration and the ``.conanrc`` file to define the home location, have been merged, and will be available in beta.2

PR #11675, opened 2022-07-20
Add conan.conanrc file to setup the conan user home

czoido:
Closes: https://github.com/conan-io/conan/issues/11542

You can create a _.conanrc_ file in the folder where you are running conan (or any parent folder), it can have this content:

Set the conan home to an absolute folder
```
# accepts comments
conan_home=/absolute/folder
```

Set the conan home to a relative folder inside the current folder
```
# accepts comments
conan_home=./relative folder/inside current folder
```

Set the conan home to a relative folder outside the current folder
```
# accepts comments
conan_home=../relative folder/outside current folder
```

Set the conan home to a path containing the `~` that will be expanded to the system's user home
```
# accepts comments
conan_home=~/use the user home to expand it
```
The _.conanrc_ file is searched for in all parent folders so, if for example, in this structure:

````
.
.conanrc
├── project1
└── project2
````

And you are running from folder `project1` the parent folders are traversed recursively until a _.conanrc_ is found in case it exists.

### DIFF
--- a/conans/paths/__init__.py
+++ b/conans/paths/__init__.py
@@ -14,7 +14,7 @@ DEFAULT_CONAN_HOME = ".conan2"
 def get_conan_user_home():
     def _read_user_home_from_rc():
         try:
-            conanrc_path = os.path.join(os.getcwd(), "conan.conanrc")
+            conanrc_path = os.path.join(os.getcwd(), ".conanrc")
             values = {k: str(v) for k, v in
                       (line.split('=') for line in open(conanrc_path).read().splitlines() if
                        not line.startswith("#"))}
--- a/conans/paths/__init__.py
+++ b/conans/paths/__init__.py
@@ -25,9 +25,12 @@ def get_conan_user_home():
     def _user_home_from_conanrc_file():
         try:
             conanrc_path = _find_conanrc_file()
-            values = {k: str(v) for k, v in
-                      (line.split('=') for line in open(conanrc_path).read().splitlines() if
-                       not line.startswith("#"))}
+
+            with open(conanrc_path) as conanrc_file:
+                values = {k: str(v) for k, v in
+                          (line.split('=') for line in conanrc_file.read().splitlines() if
+                           not line.startswith("#"))}
+
             conan_home = values["conan_home"]
             # check if it's a local folder
             if conan_home[:2] in ("./", ".\\") or conan_home.startswith(".."):
--- a/conans/paths/__init__.py
+++ b/conans/paths/__init__.py
@@ -1,5 +1,4 @@
 # coding=utf-8
-from configparser import ConfigParser
 import os
 import platform
 
@@ -13,22 +12,21 @@ DEFAULT_CONAN_HOME = ".conan2"
 
 
 def get_conan_user_home():
-    conanrc_home = None
-
-    conanrc_config = ConfigParser()
-    conanrc_file = conanrc_config.read(os.path.join(os.getcwd(), "conan.conanrc"))
-
-    if conanrc_file:
+    def _read_user_home_from_rc():
         try:
-            init_section = conanrc_config["init"]
-            conanrc_home = init_section["conan_home"]
-        except KeyError:
-            pass
+            conanrc_path = os.path.join(os.getcwd(), "conan.conanrc")
+            values = {k: str(v) for k, v in
+                      (line.split('=') for line in open(conanrc_path).read().splitlines() if
+                       not line.startswith("#"))}
+            conan_home = values["conan_home"]
+            # check if it's a local folder
+            if conan_home and (conan_home[:2] in ("./", ".\\") or conan_home.startswith("..")):
+                conan_home = os.path.abspath(os.path.join(os.getcwd(), conan_home))
+            return conan_home
+        except (IOError, KeyError):
+            return None
 
-    if conanrc_home and (conanrc_home[:2] in ("./", ".\\") or conanrc_home.startswith("..")):  # local
-        conanrc_home = os.path.abspath(os.path.join(os.getcwd(), conanrc_home))
-
-    user_home = conanrc_home or os.getenv("CONAN_HOME")
+    user_home = _read_user_home_from_rc() or os.getenv("CONAN_HOME")
     if user_home is None:
         # the default, in the user home
         user_home = os.path.join(conan_expand_user("~"), DEFAULT_CONAN_HOME)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: conans/paths/__init__.py [hunk 1]
```diff
@@ -12,7 +13,22 @@
 
 
 def get_conan_user_home():
-    user_home = os.getenv("CONAN_HOME")
+    conanrc_home = None
+
+    conanrc_config = ConfigParser()
```
lasote: Why not a "key=value" file without sections? Do we expect to have multiple?

czoido: Probably not, I can simplify it 👍

conans/paths/__init__.py
```diff
@@ -12,7 +12,21 @@
...
-    user_home = os.getenv("CONAN_HOME")
+    def _read_user_home_from_rc():
+        try:
+            conanrc_path = os.path.join(os.getcwd(), "conan.conanrc")
+            values = {k: str(v) for k, v in
+                      (line.split('=') for line in open(conanrc_path).read().splitlines() if
+                       not line.startswith("#"))}
+            conan_home = values["conan_home"]
+            # check if it's a local folder
+            if conan_home and (conan_home[:2] in ("./", ".\\") or conan_home.startswith("..")):
```
czoido:
```suggestion
            if conan_home[:2] in ("./", ".\\") or conan_home.startswith(".."):
```

conans/paths/__init__.py
```diff
@@ -12,7 +12,21 @@
 
 
 def get_conan_user_home():
-    user_home = os.getenv("CONAN_HOME")
+    def _read_user_home_from_rc():
+        try:
+            conanrc_path = os.path.join(os.getcwd(), "conan.conanrc")
```
memsharded: I think the file can be called just ``.conanrc``
```suggestion
            conanrc_path = os.path.join(os.getcwd(), ".conanrc")
```

conans/paths/__init__.py
```diff
@@ -12,7 +12,21 @@
...
+            values = {k: str(v) for k, v in
+                      (line.split('=') for line in open(conanrc_path).read().splitlines() if
+                       not line.startswith("#"))}
+            conan_home = values["conan_home"]
+            # check if it's a local folder
+            if conan_home[:2] in ("./", ".\\") or conan_home.startswith(".."):
+                conan_home = os.path.abspath(os.path.join(os.getcwd(), conan_home))
```
memsharded: The home should be computed relative to the folder containing the ``.conanrc``, not ``os.getcwd()``, otherwise it will be confusing.

conans/test/unittests/paths/user_home_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+from conans.paths import get_conan_user_home, DEFAULT_CONAN_HOME
+from conans.test.utils.test_files import temp_folder
+from conans.util.files import chdir
+
+
+def test_conanrc_abs_path_get_conan_user_home():
+    _temp_folder = temp_folder(path_with_spaces=True)
+    with chdir(_temp_folder):
+        with open(os.path.join(_temp_folder, "conan.conanrc"), 'w+') as file:
```
memsharded:
```suggestion
        with open(os.path.join(_temp_folder, ".conanrc"), 'w+') as file:
```

conans/paths/__init__.py
czoido: I search for the `.conanrc` file in the `os.getcwd()` so it will be the same folder...
I assumed this file is in the root folder of a repo and that you call conan from that folder. Should we traverse the directory structure to find for this files in some way?

memsharded: Yes, that was the original idea. Too often you are not in the root, but in a ``build`` subfolder or something.

memsharded (approved):

conans/paths/__init__.py
```diff
@@ -12,21 +12,31 @@
...
+                return conanrc_file
+            else:
+                path = path.parent
+
+    def _user_home_from_conanrc_file():
         try:
-            conanrc_path = os.path.join(os.getcwd(), ".conanrc")
+            conanrc_path = _find_conanrc_file()
             values = {k: str(v) for k, v in
                       (line.split('=') for line in open(conanrc_path).read().splitlines() if
```
lasote: I was wondering if this file descriptor keeps open until the process finish but apparently is doing well, it gets closed after the `for` loop. Just for my own knowledge, is that a feature of python3? it was always like that? I cannot find documentation about the automatic close. If it was accidental, should we read the contents first?

lasote (changes requested):
The issue to be closed described also that we wanted to recover the "storage" path (different from the cache). Right?

conans/paths/__init__.py
czoido: I found some info on Stack Overflow and apparently the garbage collection will handle this in different ways depending on the Python implementation, but I think that if we want to make sure that this won't be problematic in the future we should probably change it. I think I'll better remove the file from the comprehension.

memsharded:
@lasote the storage  path was already merged in https://github.com/conan-io/conan/pull/11672

lasote (approved):

prince-chrismc:
This made playing with Conan 2.0 so much easier! ❤️ 

Now I just keep `.conanrc` in my root project folder with `conan_home=~/.conan2` and magic 🧙 I get

```sh
$ conan config home
Current Conan home: /Users/christopherm/.conan2
```

👏

memsharded:
@prince-chrismc To make things clear, ``.conan2`` is already the default name in Conan 2.0 beta to avoid colliding with Conan 1.X. But the feature should work for other names too, please check

prince-chrismc:
It does, `conan_home=/tmp/meetup/.conan_two`

```
Current Conan home: /Users/christopherm/.conan_two
 christopherm@christopherm-mac  ~/meetup   conan2 ±  conan config home
Initialized file: '/tmp/meetup/.conan_two/settings.yml'
Initialized file: '/tmp/meetup/.conan_two/extensions/plugins/compatibility/compatibility.py'
Initialized file: '/tmp/meetup/.conan_two/extensions/plugins/compatibility/app_compat.py'
Initialized file: '/tmp/meetup/.conan_two/extensions/plugins/compatibility/cppstd_compat.py'
Initialized file: '/tmp/meetup/.conan_two/extensions/plugins/profile.py'
Current Conan home: /tmp/meetup/.conan_two
```

blackliner:
Amazing, thanks for this! This will make working in multiple repositories with different, conflicting hooks a dream. Also, monorepo ❤️

PR #11678, opened 2022-07-20
AutotoolsToolchain: empty None values from self fields + Refactor

lasote:
Changelog: Bugfix: The `AutotoolsToolchain` now clears `None` values from the attributes `.cxxflags`, `.cflags`, `.ldflags` and `.defines`.
Changelog: Feature: The `AutotoolsToolchain` attributes `.cxxflags`, `.cflags`, `.ldflags` and `.defines` can be read at any moment, now is not needed to call `.environment()` to get them calculated. In the other hand, if you want to add additional flags the following attributes have to be used: `.extra_cxxflags`, `.extra_cflags`, `.extra_ldflags` and `.extra_defines`
Docs: https://github.com/conan-io/docs/pull/2660

### DIFF
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -152,14 +152,19 @@ class AutotoolsToolchain:
                             + apple_flags + extra_flags["ldflags"])
         self.defines.extend([self.ndebug, self.gcc_cxx11_abi] + extra_flags["defines"])
 
+        self.cxxflags = self._filter_list_empty_fields(self.cxxflags)
+        self.cflags = self._filter_list_empty_fields(self.cflags)
+        self.ldflags = self._filter_list_empty_fields(self.ldflags)
+        self.defines = self._filter_list_empty_fields(self.defines)
+
         if is_msvc(self._conanfile):
             env.define("CXX", "cl")
             env.define("CC", "cl")
 
-        env.append("CPPFLAGS", ["-D{}".format(d) for d in self._filter_list_empty_fields(self.defines)])
-        env.append("CXXFLAGS", self._filter_list_empty_fields(self.cxxflags))
-        env.append("CFLAGS", self._filter_list_empty_fields(self.cflags))
-        env.append("LDFLAGS", self._filter_list_empty_fields(self.ldflags))
+        env.append("CPPFLAGS", ["-D{}".format(d) for d in self.defines])
+        env.append("CXXFLAGS", self.cxxflags)
+        env.append("CFLAGS", self.cflags)
+        env.append("LDFLAGS", self.ldflags)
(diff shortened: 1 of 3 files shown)


conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -152,14 +152,19 @@ def environment(self):
                             + apple_flags + extra_flags["ldflags"])
         self.defines.extend([self.ndebug, self.gcc_cxx11_abi] + extra_flags["defines"])
 
+        self.cxxflags = self._filter_list_empty_fields(self.cxxflags)
```
memsharded: I think the problem is that the code is not well formed. Probably the ``cxxflags`` should be ready and filtered None values, before even calling ``environment()``, and that function should only convert to an ``Environment`` object. So something like this is possible:

```python
at = AutotoolsToolchain(self)
at.cxxflags = .... # do something
env = at.environmet()
```

lasote: agree.

lasote: I will provide the refactor.

lasote: Now I see why it is done like this. The `cxxflags` are composed by public values that also could be adjusted after instancing the `AutotoolsToolchain`:

```python

at = AutotoolsToolchain(self)
at.libcxx = "whatever" # This affects cxxflags so cannot be computed at constructor.
at.cxxflags = .... # do something
at.environment()
# Now cxxflags is calculated with libcxx value
```

lasote: Might we prefer to do setters for attributes like `libcxx` to recalculate the `cxxflags`? soundsa bit ugly

memsharded: Then we probably need to split, have some ``self.extra_cxxflags`` or something, that will be just added to the ``libcxx`` ones, and then not making available the aggregated ``self.cxxflags``? Or leave the ``self.cxxflags`` public, for users, but internally use them only for aggregated to the others?

lasote: ~or maybe setter/getter for `cxxflags`... I will give it a try.~ No, that would be black magic.

franramirez688 (commented):
It looks great but I'm wondering if we really need those `extra_***` attributes. The configurations `self._conanfile.conf.get("tools.build:*****")` are meant to add those extra flags so I think it could be a bit of a boilerplate mechanism. What do you think?

lasote:
The mechanism was already there, I suppose that some recipes (only for the recipe) might need to introduce something in some configuration that Conan is not providing. I would love to remove the 80% of the customization power but I'm afraid we can't.

franramirez688 (approved):

memsharded (approved):

Issue #11520, opened 2022-06-24
MSBuildDeps Generator - Traits Support


santhonisz:
Hi there,

I've been experimenting with the latest Conan 2.0.0-beta1 release - it's looking great! 

I was wondering if you could clarify if/when requirement traits support is coming the `MSBuildDeps` generator; currently is appears that transitive dependencies are always still included in the generated `.props` files and the various requirement traits don't impact the generated files e.g. in the `AdditionalIncludeDirectories` property etc.

Thanks!

memsharded:
Hi @santhonisz 

Yes, you are right, some of the traits behavior are only applied to the ``CMake`` generators at the moment.
It totally makes sense to apply them to ``MSBuildDeps``, lets do it, thanks for the feedback.

memsharded:
Implemented in https://github.com/conan-io/conan/pull/11680, will be in next beta.2

It also includes support for components.

PR #11680, opened 2022-07-20
MSBuildDeps traits

memsharded:
Close https://github.com/conan-io/conan/issues/11520

As I needed some dependencies, I reused the data from the CMake tests for traits, making it a ``session`` fixture. At least in Windows that reduces test time +50%, from 30 secs in my computer to 12secs (only for the previous tests, savings will be bigger if I counted the new tests). 

Seems a good approach to save test time, but the team should be aware of this, and use it accordingly, discussing when necessary what common fixtures make sense.


lasote (approved):
Nice.

czoido (approved):

PR #11712, opened 2022-07-27
Update just in mode install and don't fail if mode check for tools.system.package_manager

### LINKED ISSUE #11706: [feature] make system_requirements error logger more informative
currently, error log looks like:
```
xorg/system: ERROR: while executing system_requirements(): Can't update because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
```
If should contain a list, which packages are missing.

czoido:
Changelog: Feature: Do the package manager update just in mode `install` and ignore if mode is `check`.
Docs: https://github.com/conan-io/docs/pull/2663

Just checking if this may make more sense than raising an exception if `mode=check` and `update=True` because unless you use `mode=install` you will never check if the packages are installed.

Closes: https://github.com/conan-io/conan/issues/11706

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -137,10 +137,6 @@ class _SystemPackageManagerTool(object):
         if self._mode == self.mode_install:
             command = self.update_command.format(sudo=self.sudo_str, tool=self.tool_name)
             return self._conanfile_run(command, self.accepted_update_codes)
-        else:
-            self._conanfile.output.info(f"Did not update '{str(self.__class__.__name__)}'. "
-                                        f"To run the update operation set the 'tools.system.package_manager:mode' "
-                                        f"configuration to 'install'")
 
     def _check(self, packages):
         missing = [pkg for pkg in packages if self.check_package(self.get_package_name(pkg)) != 0]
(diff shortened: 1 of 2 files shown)


conan/tools/system/package_manager.py
```diff
@@ -132,16 +132,15 @@ def _install(self, packages, update=False, check=True, **kwargs):
...
-                                                                                     self.mode_install))
-        command = self.update_command.format(sudo=self.sudo_str, tool=self.tool_name)
-        return self._conanfile_run(command, self.accepted_update_codes)
+        # we just update the package manager database in case we are in 'install mode'
+        # in case we are in check mode warn about that but don't fail
+        if self._mode == self.mode_install:
+            command = self.update_command.format(sudo=self.sudo_str, tool=self.tool_name)
+            return self._conanfile_run(command, self.accepted_update_codes)
+        else:
+            self._conanfile.output.info(f"Did not update '{str(self.__class__.__name__)}'. "
```
lasote: I might skip even the message. If the mode is `check` it is very likely you don't expect the `update` to be done.

lasote (approved):

conan/tools/system/package_manager.py
memsharded: Agree

memsharded (approved):

Issue #11703, opened 2022-07-26
[bug] LIBS variable not correct for linking static libraries when building with autoconf tools


jngrb:
### Environment Details (include every applicable attribute)
  * Operating System+version: Ubuntu 20.04
  * Compiler+version: gcc-9
  * Conan version: conan 1.47.0
  * Python version: Python 3.8.10

### Steps to reproduce (Include if Applicable)
1. create a dummy conanfile.py for wrapping a static library, e.g. to wrap `/usr/lib/x86_64-linux-gnu/libcrypto.a`, use this package_info method:

```
  def package_info(self):
    self.cpp_info.libdirs = ["/usr/lib/x86_64-linux-gnu"]
    self.cpp_info.libs = ["libcrypto.a"]
```

2. have a second recipe that consumes the dummy recipe and builds a executable or library with autotools. Excerpts from such a conanfile.py:

```
from conans import ConanFile, AutoToolsBuildEnvironment
[...]
  def build(self)
    [...]
    autotools = AutoToolsBuildEnvironment(self, win_bash=tools.os_info.is_windows)
    configure_args = [...]
    autotools.configure(vars=self._autotools_vars, args=configure_args)
    [...]
```

### Actual behaviour

autotools / configure is called with environment variable

```
LIBS=-llibcrypto.a
```

### Expected behaviour

autotools / configure is called with environment variable

```
LIBS=-l:libcrypto.a
```

References:
* see <https://stackoverflow.com/questions/6578484/telling-gcc-directly-to-link-a-library-statically>
* and <https://sourceware.org/binutils/docs/ld/Options.html>

> -l _namespec_: [...] If _namespec_ is of the form _:filename_, ld will search the library path for a file called _filename_, otherwise it will search the library path for a file called _libnamespec.a_ [or _libnamespec.so].

### Logs (Executed commands with output) (Include/Attach if Applicable)

I found the relevant python methods to be these:

* [conans/client/build/compiler_flags.py#L244](https://github.com/conan-io/conan/blob/release/1.50/conans/client/build/compiler_flags.py#L244)
* [conan/tools/gnu/gnudeps_flags.py#L73](https://github.com/conan-io/conan/blob/release/1.50/conan/tools/gnu/gnudeps_flags.py#L73)

memsharded:
Hi @jngrb 

[...]

PR #11713, opened 2022-07-27
WIP: [autotools / gnu tools] use "-l:" for static libraries

jngrb:
attempt to solve issue #11703 - tested with gcc and ld on Ubuntu and a Yocto variant

This PR is WIP:

* TODO: detect that the linker 'ld' is the one in use

### DIFF
--- a/conan/tools/gnu/gnudeps_flags.py
+++ b/conan/tools/gnu/gnudeps_flags.py
@@ -80,6 +80,11 @@ class GnuDepsFlags(object):
                 if not library.endswith(".lib"):
                     library += ".lib"
                 result.append(library)
+            elif library.startswith("lib") and library.endswith(".a"):
+                # link by filename
+                # TODO: at this point, we should only use -l:
+                # if we are sure that we use ld as linker
+                result.append("-l:%s" % library)
             else:
                 result.append("-l%s" % library)
         return result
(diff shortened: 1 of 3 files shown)


conan/tools/gnu/gnudeps_flags.py
```diff
@@ -80,6 +80,11 @@ def _format_libraries(self, libraries):
                 if not library.endswith(".lib"):
                     library += ".lib"
                 result.append(library)
+            elif library.startswith("lib") and library.endswith(".a"):
+                # link by filename
+                # TODO: at this point, we should only use -l:
+                # if we are sure that we use ld as linker
```
memsharded, line 86: Probably the way to go, aligned with Conan 2.0 design, is not try to detect it, but use a ``conf`` that explicitly defines it (opt-in or opt-out depending on the default we chose). Better than tricky, and not 100% robust auto-detections.

conans/client/build/compiler_flags.py
```diff
@@ -250,6 +250,11 @@ def format_libraries(libraries, settings):
             if not library.endswith(".lib"):
                 library += ".lib"
             result.append(library)
+        elif library.startswith("lib") and library.endswith(".a"):
```
memsharded, line 253: This is the legacy build systems integrations support, that are already removed in 2.0. Better not touch them.

memsharded (commented):
We will try to have some discussion around this for next iteration (1.52, as 1.51 is being shipped right now)

PR #11716, opened 2022-07-27
Update system package manager db right before install, instead before check

### LINKED ISSUE #11715: [bug] System Manager updates always when a package is installed already
It's listed as a bug, because the old `SystemPackageTool` has a different behavior, which, had an advantage:

When a system package is installed on your machine, and you call `SystemPackageTool.install(..., update=True)`, Conan checks first if that package is already installed, then will run update + install after that, only when is not installed. The only exception is when the parameter `force=True` is configured (False by default).

More information: https://github.com/conan-io/conan/blob/cb25aa8d8391cfb50edbdd59c74aa286f7f2be02/conans/client/tools/system_pm.py#L123

The new designed of System Manager, runs `update` first, then it checks if it's installed or not:

https://github.com/conan-io/conan/blob/develop/conan/tools/system/package_manager.py#L106

Thus, does not make sense forcing a new update, when all package are already installed. It would be better checking first, then compare if some package is missing, then install them.

Related to https://github.com/conan-io/conan-center-index/issues/11936

/cc @emzeat

  * Operating System+version: Linux, BTW I use Arch
  * Compiler+version: GCC 11
  * Conan version: 1.50.0
  * Python version: 3.10.4

Running on conanio/gcc10 docker image

```
sudo apt-get -qq update
sudo apt-get -qq install -y --no-install-recommends --no-install-suggests vim
```

Now, create a dummy conanfile.py with the follow section:

```
from conan import ConanFile
from conan.tools.system.package_manager import Apt

class Bugged(ConanFile):

        apt = Apt(self)
        apt.install(["vim"], update=True, check=True)
```

And don't forget to update the configuration mode:

`tools.system.package_manager:mode=install`

czoido:
Changelog: Feature: Update system package manager db right before install, instead before check
Docs: omit

We can also consider if we want to manager this via a conf or we want to leave in the recipes...

Closes: https://github.com/conan-io/conan/issues/11715

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -103,9 +103,6 @@ class _SystemPackageManagerTool(object):
         raise ConanException("None of the installs for the package substitutes succeeded.")
 
     def _install(self, packages, update=False, check=True, **kwargs):
-        if update:
-            self.update()
-
         if check:
             packages = self.check(packages)
 
@@ -120,6 +117,10 @@ class _SystemPackageManagerTool(object):
                                                                                      self.mode_check,
                                                                                      self.mode_install))
         elif packages:
+
+            if update:
+                self.update()
+
             packages_arch = [self.get_package_name(package) for package in packages]
             if packages_arch:
                 command = self.install_command.format(sudo=self.sudo_str,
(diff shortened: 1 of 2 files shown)


memsharded (approved):
Looks good, I'll wait for @lasote review too

lasote (approved):

lizzyd710:
Was this included in the release 1.51.0?

czoido:
Hi @lizzyd710 ,
Yes, it is part of the 1.51 release

Issue #11702, opened 2022-07-25
[develop2][bug] A bad custom command, with a wrong import breaks Conan


memsharded:
As it fails at import time, it no longer can execute anything, and the traces are confusing.

I think it should be necessary to wrap the error and include a clear error message: "this is broken, remove this command to continue"

memsharded:
Fixed in https://github.com/conan-io/conan/pull/11720

PR #11720, opened 2022-07-28
Capture error loading custom command

lasote:
Changelog: Fix: If Conan fails to load a custom command now it fails with a useful error message.
Docs: omit

Closes https://github.com/conan-io/conan/issues/11702

### DIFF
--- a/conan/cli/cli.py
+++ b/conan/cli/cli.py
@@ -32,12 +32,14 @@ class Cli:
         self._conan_api = conan_api
         self._groups = defaultdict(list)
         self._commands = {}
+
+    def _add_commands(self):
         conan_commands_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "commands")
         for module in pkgutil.iter_modules([conan_commands_path]):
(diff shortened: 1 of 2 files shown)


conan/cli/cli.py
```diff
@@ -32,20 +32,26 @@ def __init__(self, conan_api):
...
 
         sys.path.append(custom_commands_path)
         for module in pkgutil.iter_modules([custom_commands_path]):
             module_name = module[1]
             if module_name.startswith("cmd_"):
-                self._add_command(module_name, module_name.replace("cmd_", ""))
+                try:
+                    self._add_command(module_name, module_name.replace("cmd_", ""))
+                except Exception as e:
+                    raise ConanException("Error loading custom command "
```
memsharded: I think it would be better to write an error, but continue. The problem with this is that a defective command won't allow you to get a fix for it with ``conan config install`` for example. If the skipped command is used later, then it will fail, no prob.

memsharded (approved):

Issue #11722, opened 2022-07-28
[feature] Allowing a per-local-cache user declaration of compiler compatibility for all dependents


foundry-markf:
Hi,

This proposal is to satisfy two points of view, discussed between myself and @FnGyula in our products usage of Conan, against a large set of dependent packages (that we maintain).

We follow guidelines in our industry to maximise compatibility in the domain, which often requires a compiler change (across 3 desktop platforms) every year. We are aiming at reducing the initial up-front-cost of trying out new compiler versions in product, without needing a whole library stack rebuild at the same time. Currently, we set the `compiler.version` back in profiles for any dependent that has not been rebuilt, but it is an effort that is then undone when rebuilds eventually occur.

The short of it is to allow users to declare compiler compatibility, per local cache, for dependent packages, without needing to modify any recipes.

## First pov
The main idea here is to avoid the catch22 with compatible packages:
1. You have a recipe
2. Have a new compiler
3. See that it is compatible
4. If I update the recipe to add it as a compatible compiler, change the recipe for the package_id
5. Now I have no binary for that new revision and might as well build it against the new compiler all the same

## Second pov
A generalisation of the existing Intel compiler compatibility with GCC versions, in the global user-space to decide on what they're comfortable with, in order to provide fallbacks for dependent packages while upgrading a product's choice of compiler version.

I have a proof of concept proposal, including an integration test, that I will link to this in a PR shortly.

We have this in mind for Conan 1.x+

- [x ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

AbrilRBS:
This PR was left open by mistake after closing the linked PR - As @memsharded mentioned in the PR, this is already solved in Conan 2 with the `compatibility.py` plugin - [Docs here](https://docs.conan.io/2/reference/extensions/binary_compatibility.html) for anyone interested :)

PR #11723, opened 2022-07-28
Opt-in compatible compilers per local cache rather than per recipe

foundry-markf:
Changelog: (Feature): Fixes issue #11722. Adding the ability to specify compatible compilers per local cache for all dependent packages, to avoid the need to modify their recipes. Use case is to ease the testing/migration of product builds to new compiler versions that produces ABI compatible code with existing packages in Artifactory.

cc @FnGyula

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -159,6 +159,16 @@ _t_default_settings_yml = Template(textwrap.dedent("""
                     threads: [None]
                     exceptions: [None]
 
+    # opt-in for stating which versions of compilers are compatible with other versions
+    # do not specify this as part of your recipe settings
+    #compiler_compatibility:
+    #    apple-clang:
+    #        "13.0": ["12.0", "11.0", "10.0"]
+    #    gcc:
+    #        "9": ["6", "4.8"]
+    #    Visual Studio:
+    #        "16": ["15", "14"]
+
     build_type: [None, Debug, Release, RelWithDebInfo, MinSizeRel]
 
 
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -353,7 +353,7 @@ class GraphBinariesAnalyzer(object):
         node.id_indirect_prefs.difference_update(node.id_direct_prefs)
         return node.id_direct_prefs, node.id_indirect_prefs
 
-    def _compute_package_id(self, node, default_package_id_mode, default_python_requires_id_mode):
+    def _compute_package_id(self, node, default_package_id_mode, default_python_requires_id_mode, profile_host):
         """
         Compute the binary package ID of this node
         :param node: the node to compute the package-ID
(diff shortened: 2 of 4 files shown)


conans/client/conf/__init__.py
```diff
@@ -159,6 +159,16 @@
                     threads: [None]
                     exceptions: [None]
 
+    # opt-in for stating which versions of compilers are compatible with other versions
```
memsharded, line 162: This is to be implemented in the ``compatibility.py`` plugin in 2.0.

conans/client/graph/graph_binaries.py
```diff
@@ -399,6 +399,17 @@ def _compute_package_id(self, node, default_package_id_mode, default_python_requ
         if apple_clang_compatible:
             conanfile.compatible_packages.append(apple_clang_compatible)
 
+        # opt-in compiler compatibility
+        try:
+            compiler_name = conanfile.settings.compiler.value
+            compiler_version = conanfile.settings.compiler.version.value
+            for compat_version in getattr(profile_host.processed_settings.compiler_compatibility[compiler_name], compiler_version).values_range:
```
memsharded, line 406: What is this ``compiler_compatibility``?

memsharded (commented):
This is very unlikely that it will fly for Conan 1.X scope.

I'd recommend having a look and trying the ``compatibility.py`` plugin in 2.0, and reporting against it.

conans/client/graph/graph_binaries.py
progician: This is a dictionary in the settings.yml where the user can add compatible compilers ... (perhaps `compatible_compilers` would be indeed a better name for it).

conans/client/conf/__init__.py
progician: > This is very unlikely that it will fly for Conan 1.X scope.

Is there any particular reason for it? This feature works pretty well with the current 1.X set ups and does not break compatibility with anything in 1.X.

memsharded: Yes, every single new feature requires way more effort than it seems in the PR. The later derivatives, questions, support tickets, bug reports, fixes, documentation, ux issues, discussions of approaches with the team, etc, are like 95% of the work to do, the PR with the initial feature is just the tip of the iceberg. As we are already in beta.2, we don't have time or resources to do this in 1.X scope, all of our efforts goes to 2.0, and just backport to 1.X the necessary things for the upgrade.

memsharded:
This functionality is provided in 2.0 by the ``compatibility.py`` plugin. If such a plugin wants to load a yaml file from the cache to get its inputs to define compatibility, that is perfectly doable too, but it is better to keep the responsibilities de-coupled, ``settings.yml`` defines the space, ``compatibility.py`` defines the binary compatibility.

So I think we can close this PR, thanks very much for your contribution!

Issue #11727, opened 2022-07-29
[bug] '.' is not recognized as an internal or external command on windows in 2.0.0-beta2


ssrobins:
### Environment Details (include every applicable attribute)
  * Operating System+version: Windows 10 21H1 19043.1826
  * Compiler+version: msvc 193
  * Conan version: 2.0.0-beta2
  * Python version: 3.10.5

### Steps to reproduce (Include if Applicable)
1. git clone https://github.com/ssrobins/conan-recipes.git
1. cd conan-recipes\recipes\box2d
1. Then run any of these variations of the conan create command (tried to 'black-box' a workaround :P)
    - The path . after --user:
`conan create --update --user ssrobins . -pr:h=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x86 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x64 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/default -s build_type=Debug`

    - The path . at the end:
`conan create --update --user ssrobins -pr:h=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x86 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x64 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/default -s build_type=Debug .`

    - Full path at the end:
`conan create --update --user ssrobins -pr:h=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x86 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/windows_x64 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts/../profiles/default -s build_type=Debug C:\Users\Steve\Code\conan-recipes\recipes\box2d`

    - All backslashes:
`conan create --update --user ssrobins -pr:h=C:\Users\Steve\Code\conan-recipes\scripts\..\profiles\windows_x86 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts\..\profiles\windows_x64 -pr:b=C:\Users\Steve\Code\conan-recipes\scripts\..\profiles\default -s build_type=Debug C:\Users\Steve\Code\conan-recipes\recipes\box2d`

    - All forward slashes:
[...]

PR #11728, opened 2022-07-29
BUGFIX: Fail if no 'os' is specified in build profile and want to run an exte…

lasote:
Changelog: Bugfix: If the 'os' is not specified in the build profile and a recipe, in Windows, wanted to run a command, it failed to prepend the build environment launcher causing a hard-to-debug command line error. Now Conan raises if the "os" is not declared in that case.
Docs: omit

Close https://github.com/conan-io/conan/issues/11727

### DIFF
--- a/conans/test/integration/command/upload/syncronize_test.py
+++ b/conans/test/integration/command/upload/syncronize_test.py
@@ -15,7 +15,6 @@ class SynchronizeTest(unittest.TestCase):
 
     def test_upload(self):
         client = TestClient(servers={"default": TestServer()}, inputs=["admin", "password"])
-        save(client.cache.default_profile_path, "")
         ref = RecipeReference.loads("hello/0.1@lasote/stable")
         files = {"conanfile.py": GenConanfile("hello", "0.1").with_exports("*"),
                  "to_be_deleted.txt": "delete me",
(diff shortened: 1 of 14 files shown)


memsharded (approved):
Looking good

lasote:
A lot of tests fail now because many of them use an incomplete `build profile`.  I want to make sure first if they should be fixed (I think it makes sense) before doing it so I tagged as "look into".

conans/test/integration/command/upload/syncronize_test.py
```diff
@@ -15,7 +15,6 @@ class SynchronizeTest(unittest.TestCase):
 
     def test_upload(self):
         client = TestClient(servers={"default": TestServer()}, inputs=["admin", "password"])
-        save(client.cache.default_profile_path, "")
```
memsharded, line 18: Why removing this is necessary? It seems only export + upload are used here, which should never need the profile at all?

memsharded (approved):

conans/test/integration/command/upload/syncronize_test.py
lasote: The line 73 is doing `client.run("install --requires=%s --build missing" % str(ref))` so building the package.

memsharded (approved):

PR #11744, opened 2022-08-01
Expose default profile in test client and set current architecture in conftest

jcar87:
Changelog: omit
Docs: omit

* Add method that exposes the default host profile in the Test Client, for convenience.
* When writing the default profiles used for tests in `conftest.py`, set the architecture base on the current native architecture (as per the Python interpreter).

This is to enable tests that make calls to a `TestClient` object and make assumptions about the output (e.g. filenames that encode the architecture setting, or assert output in compiled code or tool logs).

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -163,6 +163,8 @@ tools_locations = {
 
 
 # TODO: Make this match the default tools (compilers) above automatically
+arch = platform.machine()
+arch_setting = "armv8" if arch in ["arm64", "aarch64"] else arch
 default_profiles = {
     "Windows": textwrap.dedent("""\
         [settings]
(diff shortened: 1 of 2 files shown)


conans/test/conftest.py
```diff
@@ -163,6 +163,8 @@
 
 
 # TODO: Make this match the default tools (compilers) above automatically
+arch = platform.machine()
+arch_setting = "armv8" if arch in ["arm64", "aarch64"] else arch
```
jcar87, line 167: On Windows/ARM, I believe the _native_ Python interpreter reports `ARM64` (all capitals). Maybe for a future update, since this is not currently commonplace and we have no way of testing it.

Also apparently a native arm64 Python interpreter is a very recent thing :D

lasote (approved):

czoido (approved):

PR #11746, opened 2022-08-01
Replace hardcoded arch assumptions in tests

jcar87:
Changelog: Omit
Docs: Omit

This PR replaces instances in tests where the `x86_64` architecture is hardcoded in the test after a call to the `TestClient()` - typically because we are asserting a specific output or file existing (that includes the `x86_64` string), or because a file that contains the `x86_64` substring needs to be subsequently loaded or read.

This relies on the logic introduced in https://github.com/conan-io/conan/pull/11744 that exposes a method in the `TestClient` object to query the default architecture, which is set in `conftest.py`.

### DIFF
--- a/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
+++ b/conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
@@ -37,7 +37,8 @@ def test_package_from_system():
     assert os.path.exists(os.path.join(client.current_folder, "dep1-config.cmake"))
     assert not os.path.exists(os.path.join(client.current_folder, "dep2-config.cmake"))
     assert not os.path.exists(os.path.join(client.current_folder, "custom_dep2-config.cmake"))
-    dep1_contents = client.load("dep1-release-x86_64-data.cmake")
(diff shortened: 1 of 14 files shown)


conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -62,7 +63,7 @@ class Pkg(ConanFile):
         """)
     client.save({"conanfile.py": consumer})
     client.run("install . -s:b os=Windows -s:h os=Linux -s:h compiler=gcc -s:h compiler.version=7 "
-               "-s:h compiler.libcxx=libstdc++11 --build=missing")
+               "-s:h compiler.libcxx=libstdc++11 -s:h arch=x86_64 --build=missing")
```
jcar87, line 66: Here I figured that since we are being explicit about the other settings in the command line, we can also pass the architecture to match the `client.load()` call one line below. I'm not sure there's much to gain by letting it fallback to the default profile in this particular instance

lasote (approved):

czoido (approved):

PR #11747, opened 2022-08-01
Fix test to reflect that intel executable can run on Apple Silicon

jcar87:
Changelog: Omit
Docs: Omit

This PR fixes a single test to reflect that the produced executable _can_ run on Apple Silicon Macs - provided that Rosetta is installed.

(Draft)

### DIFF
--- a/conans/test/functional/toolchains/apple/test_xcodetoolchain.py
+++ b/conans/test/functional/toolchains/apple/test_xcodetoolchain.py
@@ -9,13 +9,13 @@ from conans.test.utils.tools import TestClient
 test = textwrap.dedent("""
     import os
     from conan import ConanFile, tools
-    from conan.tools.build import cross_building
+    from conan.tools.build import can_run
     class TestApp(ConanFile):
         settings = "os", "compiler", "build_type", "arch"
         def requirements(self):
             self.requires(self.tested_reference_str)
         def test(self):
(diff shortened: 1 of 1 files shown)


conans/test/functional/toolchains/apple/test_xcodetoolchain.py
```diff
@@ -83,6 +83,7 @@ def package_info(self):
     sdk_version = "11.3"
     settings = "-s arch=x86_64 -s os.sdk_version={} -s compiler.cppstd={} " \
```
jcar87, line 84: Unlike te tests in this PR  https://github.com/conan-io/conan/pull/11746 - this one explicitly sets `-s arch=x86_64`, so it's not falling back on the default profile.

This PR is one of two possible choices: we can reflect the fact that macOS on Apple Silicon can run `x86_64` binaries _just fine_ (provided Rosetta is installed) - and then fix the test such that the newr `conan.tols.build.can_run` is used...

**OR** we could dynamically set `-s arch=` according to the current architecture, and amend the `asserts` further down accordingly. 


I think either capture the intent of the test, since the arch setting controls how the xcode integration files are generated, then it just so happens that xcode takes care of the rest.

lasote (approved):

czoido (approved):

PR #11748, opened 2022-08-01
Fix logic when arm64 executables run on macOS Apple Silicon during tests

jcar87:
Changelog: Omit
Docs: Omit

These tests create executables for *both* `iOS`/`armv8` and `macOS/armv8`, and try to _execute_ the program and assert that it results in a failure and that the shell prints the "Bad CPU type in executable" message.

This PR replaces the logic such that no attempts are made at "running" the executable, relying solely on on the output of `lipo` to ascertain the architecture of the produced binary. I believe this is still a sufficient check that preserves the intention of the test.

### DIFF
--- a/conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py
+++ b/conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py
@@ -503,7 +503,7 @@ target_link_libraries(${PROJECT_NAME} hello::libhello)
 
 @pytest.mark.tool("cmake")
 @pytest.mark.skipif(platform.system() != "Darwin", reason="Only OSX")
-def test_m1():
+def test_iphoneos_crossbuild():
     profile = textwrap.dedent("""
         include(default)
         [settings]
@@ -514,9 +514,9 @@ def test_m1():
     """).format()
 
     client = TestClient(path_with_spaces=False)
-    client.save({"m1": profile}, clean_first=True)
+    client.save({"ios-armv8": profile}, clean_first=True)
     client.run("new cmake_lib -d name=hello -d version=0.1")
-    client.run("create . --profile:build=default --profile:host=m1 -tf None")
+    client.run("create . --profile:build=default --profile:host=ios-armv8 -tf None")
 
     main = gen_function_cpp(name="main", includes=["hello"], calls=["hello"])
     # FIXME: The crossbuild for iOS etc is failing with find_package because cmake ignore the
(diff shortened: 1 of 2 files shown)


conans/test/functional/toolchains/cmake/cmakedeps/test_apple_frameworks.py
```diff
@@ -503,7 +503,7 @@ def test(self):
 
 @pytest.mark.tool("cmake")
 @pytest.mark.skipif(platform.system() != "Darwin", reason="Only OSX")
-def test_m1():
+def test_iphoneos_crossbuild():
```
jcar87, line 506: Renamed because this particular test does _not_ test any logic that appears directly related to M1 or Apple Silicon, but rather a simple case of cross-compiling for iOS.

conans/test/functional/toolchains/cmake/test_cmake_toolchain_m1.py
```diff
@@ -64,8 +64,6 @@ def build(self):
     assert "CMAKE_SYSTEM_PROCESSOR: arm64" in client.out
     main_path = "./build/Release/main.app/main" if op_system == "iOS" \
         else "./build/Release/main"
-    client.run_command(main_path, assert_error=True)
-    assert "Bad CPU type in executable" in client.out
```
jcar87, line 68: A note on this test: 
This test is parameterised and produces two `armv8` executables, one for `iOS` and another one for `macOS`. When running on an Intel Mac, both are cases of cross-building. But when running on an Apple Silicon Mac, the `macOS` one is a straightforward case (no cross-building).

I do wonder if we are losing any code paths being tested by this small distinction.

conans/test/functional/toolchains/cmake/test_cmake_toolchain_m1.py
```diff
@@ -64,8 +64,6 @@ def build(self):
     assert "CMAKE_SYSTEM_PROCESSOR: arm64" in client.out
     main_path = "./build/Release/main.app/main" if op_system == "iOS" \
         else "./build/Release/main"
-    client.run_command(main_path, assert_error=True)
```
lasote, line 67: Can that assert be kept checking the architecture?

lasote (approved):

conans/test/functional/toolchains/cmake/test_cmake_toolchain_m1.py
jcar87: (adding this to keep it on the records, since we discussed it in person)

The architecture of the built file is already checked below:
```
    client.run_command("lipo -info {}".format(main_path))
    assert "Non-fat file" in client.out
    assert "is architecture: arm64" in client.out
```

So running it to confirm that the "bad CPU type in executable" is printed, as well as checking the architecture by querying the output of the `lipo` tool would appear as two different ways of checking the same post-condition. 

However we could indeed keep the assert conditional.

On a separate now, since this test is checking that we can successfully build a binary for iOS with sdk 12.0 as the base target, I've added a couple of asserts checking that (since it's still possible to accidentally build it for macos)

Issue #11755, opened 2022-08-02
[bug] Variable XXX_LIBRARIES not populated in Conan >=1.51


lasote:
From comment here: https://github.com/conan-io/conan/pull/11691#discussion_r934454174

Looks like the variable that previously contained the library names stopped being populated.

czoido:
Closed by https://github.com/conan-io/conan/pull/11756 to be released soon in 1.51.1

PR #11756, opened 2022-08-02
Bugfixes 11743 and 11755

lasote:
Changelog: Bugfix: The CMakeDeps generator failed for consumers with CMake projects doing an `add_subdirectory` with two different `find_package` calls to the same package, one in the main `CMakeLists.txt` and the other in the subdirectory.
Changelog: Bugfix: The CMakeDeps generator stopped to fill the `XXX_LIBRARIES` CMake variable, especially when using components.
Docs: omit

Close https://github.com/conan-io/conan/issues/11755
Close https://github.com/conan-io/conan/issues/11743

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -168,7 +168,9 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
 
         ########## AGGREGATED GLOBAL TARGET WITH THE COMPONENTS #####################
         {%- for comp_variable_name, comp_target_name in components_names %}
+
         set_property(TARGET {{root_target_name}} PROPERTY INTERFACE_LINK_LIBRARIES {{ comp_target_name }} APPEND)
+
         {%- endfor %}
 
         {%- endif %}
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -65,25 +65,13 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
         set(_{{ pkg_name }}_DEPENDENCIES{{ config_suffix }} "{{ deps_targets_names }}")
 
         set({{ pkg_name }}_LIBRARIES_TARGETS{{ config_suffix }} "") # Will be filled later
+        set({{ pkg_name }}_LIBRARIES{{ config_suffix }} "") # Will be filled later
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -121,6 +109,9 @@ def template(self):
                      $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_LIB_DIRS{{config_suffix}}}> APPEND)
         {%- endif %}
 
+        # For the modules (FindXXX)
+        list(APPEND {{ pkg_name }}_LIBRARIES{{ config_suffix }} {{root_target_name}})
```
lasote, line 113: Note, here I'm attaching the target instead of the path to the libraries, like in previous versions. That is because the macro `conan_package_library_targets` was simplified and only returns the targets. 

When someone uses the `Boost_LIBRARIES` is typically to link a target with them so it should be ok and it is wide easier in code, and also, the transitive dependencies of the package were always added as targets to the list.

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -174,9 +168,7 @@ def template(self):
 
         ########## AGGREGATED GLOBAL TARGET WITH THE COMPONENTS #####################
         {%- for comp_variable_name, comp_target_name in components_names %}
-
-        target_link_libraries({{root_target_name}} INTERFACE {{ comp_target_name }})
```
lasote, line 178: This is what it was causing the https://github.com/conan-io/conan/issues/11755 error. When using an `add_subdirectory` the target exists but fails to use `target_link_libraries`. This is apparently related to the scope of the target. Creating the targets as INTERFACE GLOBAL worked too but looked more dangerous.

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_components.py
```diff
@@ -322,3 +322,63 @@ def build(self):
...
+
+            message("AGGREGATED LIBS: ${Boost_LIBRARIES}")
+    """)
+
+    t.save({"conanfile.py": conanfile,
+            "CMakeLists.txt": cmakelists, "src/CMakeLists.txt": sub_cmakelists})
+    t.run("install .")
+    # only doing the configure failed before #11743 fix
+    t.run("build .")
+    assert "AGGREGATED LIBS: boost::B;boost::A" in t.out
```
lasote, line 384: These are the targets, in previous versions of Conan in that variable appeared the path to the libraries files, except for the dependencies that were also targets and are kept as targets

czoido (approved):

jcar87 (approved):
LGTM!! This was a tricky one!!

PR #11772, opened 2022-08-03
xcodedeps: Fix dependency resolution if transitive dependencies have …

the-nic:
Changelog: Bugfix: Fix XcodeDeps component handling in transitive dependencies
Docs: omit

…components

When a dependency of a dependency would have components, only the default
name `conan_dep_dep.xconfig` would be included. However, this file was never
generated, as they are in the form `conan_dep_component.xconfig`.

Fix this by adding all components instead.

Also allow the components to have dashes and use _format_name() to
avoid issues when parsing the xcconfig file.

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -230,6 +230,7 @@ class XcodeDeps(object):
 
                 sorted_components = dep.cpp_info.get_sorted_components().items()
                 for comp_name, comp_cpp_info in sorted_components:
+                    comp_name = _format_name(comp_name)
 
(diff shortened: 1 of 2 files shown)


czoido:
Hi @the-nic,
Thanks a lot for the contribution.
You are right, in that case it was not working correctly. I'm adding a test to confirm that these changes make it work as expected.

the-nic:
> Thanks a lot for the contribution.
You are right, in that case it was not working correctly. I'm adding a test to confirm that these changes make it work as expected.

Thanks for the review and adding the test!

czoido (approved):

memsharded (approved):

Issue #11773, opened 2022-08-03
[bug] system lib (dl) ordering change on link


dfleury2:
### Environment Details (include every applicable attribute)
  * Operating System+version: Debian bullseye
  * Compiler+version: gcc 11
  * Conan version: 1.51
  * Python version: 3.9.2

### Steps to reproduce (Include if Applicable)
Not easy, but lot of dependencies and openssl with dl usage (using crypto)

### Logs (Executed commands with output) (Include/Attach if Applicable)
```
/usr/bin/c++  -pipe -Wall -fstack-protector-all -fstack-check -O3 -DNDEBUG  -Wl,-z,relro -Wl,--as-needed CMakeFiles/bin_beauty_application_singleton.dir/beauty_application_singleton.cpp.o -o ../../../bin/beauty_application_singleton  ../../../lib/libbeauty.a /home/dfleury/.conan/data/openssl/1.1.1q/_/_/package/fae585d7c68637b721c0c4c0d1c52174d067eda8/lib/libssl.a -ldl /home/dfleury/.conan/data/openssl/1.1.1q/_/_/package/fae585d7c68637b721c0c4c0d1c52174d067eda8/lib/libcrypto.a -lrt -lpthread -lstdc++fs 
```

Here -ldl is placed before crypto.a, so I get 
```
usr/bin/ld: /home/dfleury/.conan/data/openssl/1.1.1q/_/_/package/fae585d7c68637b721c0c4c0d1c52174d067eda8/lib/libcrypto.a(dso_dlfcn.o): in function `dlfcn_globallookup':
dso_dlfcn.c:(.text+0x13): undefined reference to `dlopen'
```

If I use conan 1.49 or conan 1.50, it works fine.
```
[...]

PR #11788, opened 2022-08-05
Refactored CMakeDep using always targets instead of lists

lasote:
Changelog: Bugfix: Refactored CMakeDep using always targets instead of lists.
Docs:  omit

The bug appeared when components require system libraries, and those were not linked to the `IMPORTED` but to the component target. CMake was not able to order the system libraries correctly causing linker errors. So it has been fixed using always targets instead of lists, the dependencies of a package/component (such as the system libraries, apple frameworks, and transitive dependencies) are linked to an `INTERFACE` target that later is linked to the `IMPORTED` targets when `cpp_info.libs` are declared, otherwise to the global or component target. 

Close https://github.com/conan-io/conan/issues/11773

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -96,7 +96,7 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
             set_property(TARGET {{root_target_name}}
                          PROPERTY INTERFACE_LINK_LIBRARIES
                          $<$<CONFIG:{{configuration}}>:{{ '${'+pkg_name+'_OBJECTS'+config_suffix+'}' }}>
-                         ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS}x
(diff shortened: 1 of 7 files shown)


conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -78,103 +89,132 @@ def template(self):
         set(CMAKE_PREFIX_PATH {{ '${' }}{{ pkg_name }}_BUILD_DIRS{{ config_suffix }}} {{ '${' }}CMAKE_PREFIX_PATH})
 
         {% if not components_names %}
-        ########## GLOBAL TARGET PROPERTIES {{ configuration }} ########################################
```
lasote, line 81: Sorry for this diff, it is mostly tabulation of the CMake code, I think now it is much more readable and the generated files don't look so bad, actually improves readability too.

jcar87 (approved):
LGTM!

czoido (approved):

PR #11803, opened 2022-08-08
Enable get_safe on self.info.settings and self.info.options

### LINKED ISSUE #11802: [feature] Implement `get_safe` in develop for self.info objects
In Conan 2.X there is `get_safe` over `self.info.settings` and `self.info.options` but there is not in Conan 1.X.
It is needed when, for example, there is a `header_only` option and you need to check in the validate some stuff.

Related https://github.com/conan-io/conan/issues/11786

lasote:
Changelog: Feature: The `self.info.settings` and `self.info.options` now allow the `get_safe("foo")` as it is available in `self.settings` and `self.options`.
Docs: omit

Close #11802

### DIFF
--- a/conans/test/integration/package_id/test_validate.py
+++ b/conans/test/integration/package_id/test_validate.py
@@ -1,4 +1,5 @@
 import json
+import re
 import textwrap
 import unittest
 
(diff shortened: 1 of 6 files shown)


conans/test/integration/package_id/test_validate.py
```diff
@@ -39,6 +40,59 @@ def validate(self):
...
+                def validate(self):
+                    if self.info.options.get_safe("header_only") == "False":
+                        if self.info.settings.get_safe("os") == "Linux":
+                          raise ConanInvalidConfiguration("This package cannot exist in Linux")
+                        check_min_cppstd(self, 11)
+                        # These configurations are impossible
+                        if self.info.settings.os != "Windows" and self.info.options.shared:
+                            raise ConanInvalidConfiguration("shared is only supported under windows")
+
+                    # HOW CAN WE VALIDATE CPPSTD > 11 WHEN HEADER ONLY?
```
lasote, line 67: This is an old question and I don't know if we have a new answer for it.

franramirez688 (approved):

czoido (approved):

conans/test/integration/package_id/test_validate.py
memsharded: Isn't it possible to check the ``settings`` directly here? Is it the ``check_min_cppstd()`` that checks ``self.info`` or is there other issue?

memsharded (approved):

conans/test/integration/package_id/test_validate.py
lasote: `check_min_cppstd()` is using setting in develop but info in develop2.
About using settings here, is it viable/convenient with the compatible.py packages mechanism?

Issue #11794, opened 2022-08-05
[bug] Can't find conan.tools.system


lizzyd710:
Since [this PR](https://github.com/conan-io/conan-center-index/pull/11698) was merged updating xorg, the CI checks running on Ubuntu for [Cytopia ](https://github.com/CytopiaTeam/Cytopia/) have been failing when trying to install dependencies. The logs say
`ModuleNotFoundError: No module named 'conan.tools.system'`--but it can find  `conan.tools.gnu` perfectly fine. #10260 is similar to what is happening here, but with a different module, and this is happening on CI not my local machine.

### Environment Details (include every applicable attribute)
  * Operating System+version: Ubuntu 20.04 (on GitHub runner)
  * Compiler+version: gcc 9
  * Conan version: 1.51.0
  * Python version:

### Logs (Executed commands with output) (Include/Attach if Applicable)
[Build log](https://github.com/CytopiaTeam/Cytopia/runs/7696989519?check_suite_focus=true) and [sonarcloud scan log](https://github.com/CytopiaTeam/Cytopia/runs/7601870549?check_suite_focus=true).

czoido:
Hi @lizzyd710,

Thanks a lot for reporting.

Yes, this is similar to this one: https://github.com/conan-io/conan/issues/11676
Adding that import as hidden-import to _pyinstaller.py_, so it is bundled with the installer should fix the issue. I'll open a PR and point it to the 1.51.1 that we will release this week.

By the way, in case that's an option for you, if you install conan with pip (that is the [recommended way](https://docs.conan.io/en/latest/installation.html)) this should not happen.

czoido:
Closed by https://github.com/conan-io/conan/pull/11805 to be released in Conan 1.51.1 very soon.

PR #11805, opened 2022-08-08
Add conan.tools.system as hidden-import to pyinstaller.py

czoido:
Changelog: Fix: Add `conan.tools.system` as hidden-import to `pyinstaller.py`, so it is bundled with the installer.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/11794

Note: extended a test in the ci job that creates the windows installer and debian package to test this works.


jcar87 (approved):
LGTM!

PR #11818, opened 2022-08-09
xcodedeps: provide package root dir

the-nic:
Changelog: Feature: Defines the `PACKAGE_ROOT_<package>` variable in XcodeDeps generated files.
Docs: https://github.com/conan-io/docs/pull/2717

If one wants to access other folders than the provided ones (like
resources), the provided variables dont contain enough information
to access these folders.

Provide a new variable, `PACKAGE_ROOT_<package>_<component>` pointing
to the root directory of the package allowing to use it within xcode.

We have to pass the package folder as another parameter, because the
`transitive_cpp_info` variable does not contain the `rootpath`
attribute and we also cant access the package_folder attribute from
`self._conanfile.dependencies.package_folder` directly because
the package name is already formatted via `_format_name`.

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -61,6 +61,7 @@ class XcodeDeps(object):
     general_name = "conandeps.xcconfig"
 
     _conf_xconfig = textwrap.dedent("""\
+        PACKAGE_ROOT_{{pkg_name}}{{condition}} = {{root}}
         // Compiler options for {{pkg_name}}::{{comp_name}}
         HEADER_SEARCH_PATHS_{{pkg_name}}_{{comp_name}}{{condition}} = {{include_dirs}}
         GCC_PREPROCESSOR_DEFINITIONS_{{pkg_name}}_{{comp_name}}{{condition}} = {{definitions}}
@@ -122,7 +123,7 @@ class XcodeDeps(object):
         for generator_file, content in generator_files.items():
             save(generator_file, content)
 
-    def _conf_xconfig_file(self, pkg_name, comp_name, transitive_cpp_infos):
+    def _conf_xconfig_file(self, pkg_name, comp_name, package_folder, transitive_cpp_infos):
         """
         content for conan_poco_x86_release.xcconfig, containing the activation
         """
@@ -133,6 +134,7 @@ class XcodeDeps(object):
         fields = {
             'pkg_name': pkg_name,
             'comp_name': comp_name,
+            'root': package_folder,
             'include_dirs': " ".join('"{}"'.format(p) for p in _merged_vars("includedirs")),
             'lib_dirs': " ".join('"{}"'.format(p) for p in _merged_vars("libdirs")),
             'libs': " ".join("-l{}".format(lib) for lib in _merged_vars("libs")),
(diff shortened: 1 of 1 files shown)


conan/tools/apple/xcodedeps.py
```diff
@@ -223,6 +225,7 @@ def _content(self):
         all_deps = list(host_req.values()) + list(test_req.values())
         for dep in all_deps:
 
+            dep_cpp_info = self._conanfile.deps_cpp_info[dep.ref.name]
```
memsharded: ``deps_cpp_info`` cannot be used anymore in new generators. Access is via ``self._conanfile.ref.name`` for conanfile and ``self.dependencies`` for dependencies.

the-nic: But it seems that `self.dependencies` does not contain the root path. How can I get that attribute?

memsharded: ``self.dependencies["pkg"].package_folder``. The interface is described in https://docs.conan.io/en/latest/reference/conanfile/dependencies.html

the-nic: Thanks, changed the commit accordingly.

czoido:
Hi @the-nic,

Thanks for the contribution.

I think that maybe it would be better to use the resdirs information from the cppinfo so that if a package have a directory for resources it is available. And a variable something like `RESOURCE_PATHS_libname_libname[...][...][...] = ...`

Is there a case where you need to access directly the package folder that is not covered by the case that you add the resdir to the cppinfo and have access to it?

the-nic:
> I think that maybe it would be better to use the resdirs information from the cppinfo so that if a package have a directory for resources it is available. And a variable something like `RESOURCE_PATHS_libname_libname[...][...][...] = ...`

Certainly, this would be a good addition, too.

> Is there a case where you need to access directly the package folder that is not covered by the case that you add the resdir to the cppinfo and have access to it?

The problem is, these are often multiple paths and if I have to reference a specific file from the package, it its impossible to do without the single root path. Additionally, the cmake generator also provides a package root path, so they become a bit more on par.

czoido (approved):

memsharded (approved):

czoido (changes requested):
Hi @the-nic,
I was thinking that probably it would be better to define this variable in the file that includes all components for the package instead here, and not with the component suffix but just as: `PACKAGE_ROOT_<package>` because that information will be the same for all the components. WDYT?

the-nic:
> Hi @the-nic, I was thinking that probably it would be better to define this variable in the file that includes all components for the package instead here, and not with the component suffix but just as: `PACKAGE_ROOT_<package>` because that information will be the same for all the components. WDYT?

Good point, I'll update the PR

the-nic:
Hi @czoido I tried your suggestion right now and the problem is that the even though the information is the same for all components, its not the same for all configurations. And only the component files seem to be configuration-aware. Otherwise, we'd have to update the package file on each time a new configuration is added...

czoido:
Hi @the-nic,
Sorry, you are right, I did not take that into account. Then maybe it's better to just leave them in the components files, but remove the `_{{comp_name}}` and let them be defined in multiple times (I suppose it won't be a problem, but I had to test that first).

the-nic:
@czoido  I have removed the component name from the root dir variable.

czoido:
Thanks a lot @the-nic,
I'll do some checks and then I think it's good to merge.
Maybe in the future if we add more variables at package level it would be worth to make a separate file that defines them all and that is included from the file that includes the components, but if the multiple definition is not a problem I think this is good enough for the moment.
Thanks!

czoido (approved):

PR #11820, opened 2022-08-09
When the layout, use different folder for the 'conan infos' files

### LINKED ISSUE #11779: [question] Investigate generated files when layout() method is specified (Conan v1 compat)
Screenshot 2022-08-04 at 15 08 56

* Conan 1.51.0
* A recipe contains a `test_package/conanfile.py` that uses the `layout()` method and the new generators
* `conan create` will cause multiple files to be generated _next to_ the test package's `conanfile.py`.

This behaviour is different than:
* Conan 1.x without the layout() method -> everything is generated in the `build` subfolder
* Conan 2.x _with_ the layout method -> everything is generated in a `test_output` folder

lasote:
Changelog: Fix: When the `layout()` is declared in the recipe, the files such as `conainfo.txt`, `conanbuildinfo.txt` were written in the `test_package` folder, now there are written in the `generators_folder`.
Docs: omit

Close https://github.com/conan-io/conan/issues/11779


conans/client/cmd/test.py
```diff
@@ -24,18 +24,22 @@ def install_build_and_test(app, conanfile_abs_path, reference, graph_info,
         # Don't use "test_package/build/HASH/" as the build_f
         delete_after_build = False
         test_build_folder = base_folder
+        # Don't mess the test_package folder with the conaninfo.txt and others
+        # We don't have the conanfile loaded yet so we cannot know the generators_folder
+        install_folder = os.path.join(base_folder, "install")
```
memsharded: But this is changing the default install location for all cases, unconditionally, so this might break something? Sounds a bit risky

conans/test/functional/toolchains/gnu/test_v2_autotools_template.py
```diff
@@ -93,7 +93,7 @@ def test_autotools_relocatable_libs_darwin():
     shutil.move(os.path.join(package_folder, "lib"), os.path.join(client.current_folder, "tempfolder"))
     # will fail because rpath does not exist
     client.run_command("test_package/build-release/main", assert_error=True)
-    assert "Library not loaded: @rpath/libhello.0.dylib" in client.out
+    assert "Library not loaded: @rpath/libhello.0.dylib" in str(client.out).replace("'", "")
```
lasote, line 96: This is not related to this PR but fails with my apple-clang that output with quotes the rpath

conans/client/cmd/test.py
lasote: Well, only when the layout is declared, but, please, the title said: `WIP BROKEN` and I removed the reviewers... :P

memsharded (approved):
I think there is some risk of breaking some users, but as it is connected to the experimental layout() feature, it should be ok, and the proposed new behavior seems to make sense

PR #11823, opened 2022-08-09
tools.scm.Version in develop as is in develop2

### LINKED ISSUE #11821: [feature] Conan tools.scm.Version became very difficult to use
With the introduction of the new [tools.scm.Version](https://docs.conan.io/en/latest/reference/conanfile/tools/scm/other.html#version), it no longer parses `semver` as described on the documentation, but also makes users life harder. Comparing both `Version` behavior, we have:

```python
from conans import tools as tools_legacy
from conan.tools import scm

VERSION = "1.2.3"

if __name__ == "__main__":
    legacy_version = tools_legacy.Version(VERSION)
    print(f"Legacy Version: {legacy_version}")
    print(f"Major: {legacy_version.major} - Minor: {legacy_version.minor} - Patch: {legacy_version.patch}")

    version = scm.Version(VERSION)
    print(f"New version: {version}")
    print(f"Major: {version.major(fill=False)} - Minor: {version.minor(fill=False)} - Patch: {version.patch()}")
    print(f"Major: {version[0]} - Minor: {version[1]} - Patch: {version[2]}")
```

Output:

```
Legacy Version: 1.2.3
Major: 1 - Minor: 2 - Patch: 3
New version: 1.2.3
Major: 1 - Minor: 1.2 - Patch: 1.2.3
Major: 1 - Minor: . - Patch: 2
```

So far, we needed to dig an index to parse and use what we want: https://github.com/conan-io/conan-center-index/pull/12083#pullrequestreview-1064201491

My suggestion is adding a pattern support to `Version` where we could extend it:

    version = Version('1.2.3', pattern='semver')
    version.major # Yes, it's a property, not a method.

Or, we could provide new classes, with Version as the base:

```python
from conan.tools.scm import SemVersion

version = SemVersion('1.2.3')
version.major # Yes, it's a property, not an method.
```

- [ X ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

lasote:
Changelog: Bugfix: The `tool.scm.Version` model has been ported from 2.X to keep the same behavior in Conan 1.X.
Docs: omit

Close https://github.com/conan-io/conan/issues/11821

### DIFF
--- a/conan/tools/scm/__init__.py
+++ b/conan/tools/scm/__init__.py
@@ -1,2 +1,200 @@
 from conan.tools.scm.git import Git
-from conans.model.version import Version
+# develop2 uncomment the next line
+# from conans.model.version import Version
+
+
+# ######## DEVELOP2: Delete all the contents from here, this is copy-paste from the Version model
+# ######## at conans.model.version (for an easier migration)
+from functools import total_ordering
+from conans.errors import ConanException
+
+
+@total_ordering
+class _VersionItem:
+    """ a single "digit" in a version, like X.Y.Z all X and Y and Z are VersionItems
+    They can be int or strings
+    """
+    def __init__(self, item):
+        try:
+            self._v = int(item)
+        except ValueError:
+            self._v = item
+
+    @property
+    def value(self):
+        return self._v
+
+    def __str__(self):
+        return str(self._v)
+
+    def __add__(self, other):
+        # necessary for the "bump()" functionality. Other aritmetic operations are missing
+        return self._v + other
+
+    def __eq__(self, other):
+        if not isinstance(other, _VersionItem):
+            other = _VersionItem(other)
+        return self._v == other._v
+
+    def __hash__(self):
+        return hash(self._v)
+
+    def __lt__(self, other):
+        """
+        @type other: _VersionItem
+        """
+        if not isinstance(other, _VersionItem):
+            other = _VersionItem(other)
+        try:
+            return self._v < other._v
+        except TypeError:
+            return str(self._v) < str(other._v)
+
+
+@total_ordering
+class Version:
+    """
+    This is NOT an implementation of semver, as users may use any pattern in their versions.
+    It is just a helper to parse "." or "-" and compare taking into account integers when possible
+    """
+    def __init__(self, value):
+        value = str(value)
+        self._value = value
+
+        items = value.rsplit("+", 1)  # split for build
+        if len(items) == 2:
+            value, build = items
+            self._build = Version(build)  # This is a nested version by itself
+        else:
+            value = items[0]
+            self._build = None
+
+        items = value.rsplit("-", 1)  # split for pre-release
+        if len(items) == 2:
+            value, pre = items
+            self._pre = Version(pre)  # This is a nested version by itself
+        else:
+            value = items[0]
+            self._pre = None
+        items = value.split(".")
+        items = [_VersionItem(item) for item in items]
+        self._items = tuple(items)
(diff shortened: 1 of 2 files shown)


memsharded (approved):
I guess it is ok, and necessary for the migration.

conan/tools/scm/__init__.py
```diff
@@ -1,2 +1,200 @@
...
+        if not isinstance(other, _VersionItem):
+            other = _VersionItem(other)
+        try:
+            return self._v < other._v
+        except TypeError:
+            return str(self._v) < str(other._v)
+
+
+@total_ordering
+class Version:
```
uilianries, line 55: Could you please add `as_list`, so it could be minimal compatible with what we have in Conan Center Index now.

memsharded: I wouldn't like that, I think that ``as_list`` is an ugly interface that if we do that it will "infect" Conan 2.0.
If necessary, please let us know the use cases of such method, and we can think about what is the right interface for it.

uilianries: I still prefer the legacy, as described in this PR. The only thing is re-editing recipes again in CCI. Also, we will need to create some `if conan_version else` because the different behavior from 1.49 (i guess) til 1.52

memsharded: I haven't found occurrences of ``as_list`` in conan-center-index (only one method in openssl recipe), am I missing something?

uilianries: Boost recipe: https://github.com/conan-io/conan-center-index/blob/master/recipes/boost/all/conanfile.py#L1461

If you are running GH search, forget, it's useless most of time.

memsharded: No, I was running my local clone. But it wasn't updated. It seems it is only the boost recipe, and it was very recently added.

It seems what the boost recipe could be done more idiomatically with ``version.major`` and ``version.minor`` isn't it?

memsharded: Furthermore, ``as_list`` is not a documented/public method at all, shouldn't be used in any of the ``Version`` implementations? While ``major`` and ``minor`` are. Not sure why they used that in the boost recipe.

uilianries: The problem was, Boost recipe need to compare some major and minor versions, but the new scm.Version does not support semver, so it return "1.2" for Version.minor(fill=False). Of course, the previous behavior, with semver, was much better. I gonna revert those changes and keep aligned with this new PR. Thanks.

memsharded: Not sure what you mean. This new ``Version`` in this PR supports semver like comparison between versions, no problem with that. How is that related to the ``as_list``  usage, which is done to just extract the major, minor and patch to compose some strings (suffixes) in the boost recipe?

uilianries: - The new Version changed major, minor and patch from property to method
- When you call `conan.tools.scm.Version('1.2.3').minor()` it returns `1.2.Z`
- When you call `conan.tools.scm.Version('1.2.3').minor(fill=False)` it returns `1.2`
- But still, I only want the minor, the `2`.
- As alternative, I can consume from `conan.tools.scm.Version('1.2.3')[1]`

Anyway, this new PR fixes the case, so we can use Version.major as a property again.

uilianries: When I said `The new Version`, I meant which is production right now and running on CCI, not this PR. This PR is correct, it follows the previous and expected behavior.

PR #11826, opened 2022-08-10
[MSBuildDeps] Bad XML format

### LINKED ISSUE #11822: [bug] .\site-packages\conan\tools\microsoft\msbuilddeps.py
conanfile.py (curaengine/None): ERROR: Traceback (most recent call last):
  File "c:\program files\python38\lib\site-packages\conans\client\generators\__init__.py", line 181, in write_generators
    generator.generate()
  File "c:\program files\python38\lib\site-packages\conan\tools\microsoft\msbuilddeps.py", line 110, in generate
    generator_files = self._content()
  File "c:\program files\python38\lib\site-packages\conan\tools\microsoft\msbuilddeps.py", line 338, in _content
    result.update(self._package_props_files(dep, build=False))
  File "c:\program files\python38\lib\site-packages\conan\tools\microsoft\msbuilddeps.py", line 308, in _package_props_files
    result[comp_filename] = self._dep_props_file(full_comp_name, comp_filename,
  File "c:\program files\python38\lib\site-packages\conan\tools\microsoft\msbuilddeps.py", line 229, in _dep_props_file
    dom = minidom.parseString(content_multi)
  File "c:\program files\python38\lib\xml\dom\minidom.py", line 1969, in parseString
    return expatbuilder.parseString(string)
  File "c:\program files\python38\lib\xml\dom\expatbuilder.py", line 927, in parseString
    return builder.parseString(string)
  File "c:\program files\python38\lib\xml\dom\expatbuilder.py", line 225, in parseString
    parser.Parse(string, True)
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 6, column 33

ERROR: Error in generator 'MSBuildDeps': not well-formed (invalid token): line 6, column 33

---

problem in generated xml node used (libmpdecimal++)
symbols ++ wrong with xml; format nodenames

```
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  </ImportGroup>
  <PropertyGroup>
    <conan_mpdecimal_libmpdecimal++_props_imported>True</conan_mpdecimal_libmpdecimal++_props_imported>
  </PropertyGroup>
</Project>
```

---

temporary solution (.\site-packages\conan\tools\microsoft\msbuilddeps.py)
line: 285+

---
            for comp_name, comp_info in dep.cpp_info.components.items():
                if comp_name is None:
                    continue

                # <SOLUTION>
                comp_name = comp_name.replace("+", "_")
                # </SOLUTION>
                    
                full_comp_name = "{}_{}".format(dep_name, comp_name)
                vars_filename = "conan_%s_vars%s.props" % (full_comp_name, conf_name)
                activate_filename = "conan_%s%s.props" % (full_comp_name, conf_name)
---

franramirez688:
Changelog: Bugfix: Added `+` and `-` symbols to be replaced by `_` in XML element names.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/11822

### DIFF
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -1,5 +1,6 @@
 import fnmatch
 import os
+import re
 import textwrap
 from xml.dom import minidom
 
@@ -10,6 +11,7 @@ from conans.errors import ConanException
 from conans.util.files import load, save
 
 VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
+NON_DIGIT_PATTERN = re.compile(r"\D")
 
 
 class MSBuildDeps(object):
@@ -125,10 +127,13 @@ class MSBuildDeps(object):
     @staticmethod
     def _dep_name(dep, build):
         dep_name = dep.ref.name
-        dep_name = dep_name.replace(".", "_")
         if build:  # dep.context == CONTEXT_BUILD:
             dep_name += "_build"
-        return dep_name
+        return MSBuildDeps._get_valid_xml_format(dep_name)
+
+    @staticmethod
+    def _get_valid_xml_format(name):
+        return NON_DIGIT_PATTERN.sub("_", name)
 
     def _vars_props_file(self, dep, name, cpp_info, deps, build):
         """
@@ -223,7 +228,6 @@ class MSBuildDeps(object):
             </Project>
             """)
             content_multi = Template(content_multi).render({"name": dep_name})
-
         # parse the multi_file and add new import statement if needed
         dom = minidom.parseString(content_multi)
         import_vars = dom.getElementsByTagName('ImportGroup')[0]
@@ -282,7 +286,7 @@ class MSBuildDeps(object):
             for comp_name, comp_info in dep.cpp_info.components.items():
                 if comp_name is None:
                     continue
-                full_comp_name = "{}_{}".format(dep_name, comp_name)
+                full_comp_name = "{}_{}".format(dep_name, self._get_valid_xml_format(comp_name))
                 vars_filename = "conan_%s_vars%s.props" % (full_comp_name, conf_name)
                 activate_filename = "conan_%s%s.props" % (full_comp_name, conf_name)
                 comp_filename = "conan_%s.props" % full_comp_name
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -6,7 +6,6 @@ from xml.dom import minidom
 from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
-from conans.client.graph.graph import CONTEXT_BUILD
 from conans.errors import ConanException
 from conans.util.files import load, save
 
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -11,7 +11,7 @@ from conans.errors import ConanException
 from conans.util.files import load, save
 
 VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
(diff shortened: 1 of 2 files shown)


conan/tools/microsoft/msbuilddeps.py
```diff
@@ -1,16 +1,17 @@
...
 
 from jinja2 import Template
 
 from conan.tools._check_build_profile import check_using_build_profile
-from conans.client.graph.graph import CONTEXT_BUILD
 from conans.errors import ConanException
 from conans.util.files import load, save
 
 VALID_LIB_EXTENSIONS = (".so", ".lib", ".a", ".dylib", ".bc")
+EXCLUDED_SYMBOLS_PATTERN = re.compile(r"[\.+]")
```
franramirez688: Do you think that we should add more symbols? More strict pattern? Or is it even too much?

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -126,10 +127,13 @@ def _condition(self):
...
         dep_name = dep.ref.name
-        dep_name = dep_name.replace(".", "_")
         if build:  # dep.context == CONTEXT_BUILD:
             dep_name += "_build"
-        return dep_name
+        return MSBuildDeps._get_valid_xml_format(dep_name)
+
+    @staticmethod
+    def _get_valid_xml_format(name):
+        return EXCLUDED_SYMBOLS_PATTERN.sub("_", name)
```
memsharded, line 136: Maybe a ``x`` looks better in the generated files? That would leave the question is how to replace the ``-`` and the ``.`` too, so I guess it is good as it is.

conan/tools/microsoft/msbuilddeps.py
memsharded: Maybe the ``-`` which is allowed too

conan/tools/microsoft/msbuilddeps.py
```diff
@@ -126,10 +127,13 @@ def _condition(self):
     @staticmethod
     def _dep_name(dep, build):
         dep_name = dep.ref.name
-        dep_name = dep_name.replace(".", "_")
         if build:  # dep.context == CONTEXT_BUILD:
             dep_name += "_build"
-        return dep_name
+        return MSBuildDeps._get_valid_xml_format(dep_name)
```
memsharded, line 132:
```python
        return re.sub(dep_name, ....)
```
As a general rule I try to follow, I try to keep things as local as possible, no need to add new methods if they are only called from one place, and they are single line. Also I wouldn't add a file scope "constant" to be used just in one function, I would put it inside the function first. 

No need to change anything, it is good as-is, but I think in the long term is better to work that way.

memsharded (approved):
I think that as a workaround it might be ok, but the general policy should be to avoid special symbols completely. Because there is a failure mode, if one package is called ``hello+``, and another is ``hello-``, they will conflict and overwrite each other.

Question to discuss with the team, should we further restrict the Conan 2.0 current ``validation_pattern = re.compile(r"^[a-z0-9_][a-z0-9_+.-]{1,100}$")`` validation pattern for the package names?

conan/tools/microsoft/msbuilddeps.py
franramirez688: Yes, but the thing is that we have to change the component names (a loop for each dep) as it's used below (line https://github.com/conan-io/conan/pull/11826/files#diff-2424e36bff4aa01d6e99c02874db657d1a135ce15dd457f48ea57c69cb470a48R289), so I proposed that compiled pattern to be more efficient. Anyway, yes, I agree with you.

conan/tools/microsoft/msbuilddeps.py
franramirez688: Yeah, I think it's good as it is, but it could be ok to make different replacements for each one. For instance:

* `.` -> `_`
* `+` or `-` -> `x`

czoido (approved):

PR #11828, opened 2022-08-10
do not skip __pycache__ folder if CONAN_KEEP_PYTHON_FILES

### LINKED ISSUE #11825: [bug] __pycache__ folders are ignored even when keep_python_files is set
`__pycache__` folders are ignored during the packaging even when the parameter **keep_python_files** is set to **true**.

```python
def gather_files(folder):
    file_dict = {}
    symlinks = {}
    keep_python = get_env("CONAN_KEEP_PYTHON_FILES", False)
    for root, dirs, files in walk(folder):
        dirs[:] = [d for d in dirs if d != "__pycache__"]  # Avoid recursing pycache
        for d in dirs:
```

memsharded:
Changelog: Bugfix: Do not skip __pycache__ folder if `CONAN_KEEP_PYTHON_FILES`.
Docs: Omit

Close https://github.com/conan-io/conan/issues/11825

This is a draft, to evaluate what could be broken

### DIFF
--- a/conans/model/manifest.py
+++ b/conans/model/manifest.py
@@ -25,7 +25,8 @@ def gather_files(folder):
     symlinks = {}
     keep_python = get_env("CONAN_KEEP_PYTHON_FILES", False)
     for root, dirs, files in walk(folder):
-        dirs[:] = [d for d in dirs if d != "__pycache__"]  # Avoid recursing pycache
+        if not keep_python:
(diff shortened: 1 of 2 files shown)


jcar87:
Looks like this would fix the issue! Are there no tests that should be updated to check the correct behaviour?

memsharded:
Test added @jcar87

jcar87 (approved):
LGTM!

Issue #11763, opened 2022-08-02
[question] 2.0 Tool Requires


beenjohn:
When generating a lockfile in 2.0, the tool requires don't end up in the lock file. With cmake listed as a tool requires, "conan create ... --lockfile=..." return an error that cmake isn't listed in the lockfile. Is this expected behavior?

- [X] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

czoido:
Hi @beenjohn,

Can you provide more details, maybe a reproducible example?

For me, with a dummy recipe like this:

```python
from conan import ConanFile
from conan.tools.cmake import cmake_layout


class mylibRecipe(ConanFile):
    name = "mylib"
    version = "1.0"
    # Binary configuration
    settings = "os", "compiler", "build_type", "arch"

    def layout(self):
        cmake_layout(self)

    def requirements(self):
        self.requires("zlib/1.2.11")
        self.tool_requires("cmake/3.19.8")

```

And doing:

```
conan lock create .
conan create . --lockfile=conan.lock
```

Works, creating a lockfile like this one:

```
{
    "version": "0.5",
    "requires": [
        "zlib/1.2.11#4524fcdd41f33e8df88ece6e755a5dcc%1650538915.154"
    ],
    "build_requires": [
        "cmake/3.19.8#7c53e8476ea7485f06e2b463d88f6787%1652284298.374"
    ],
    "python_requires": []
}
```

beenjohn:
@czoido,

Apologies for the delayed response. To clarify, the issue I'm seeing in 2.0.0-beta2 is that the tool_requires in the test_package is not ending up in the lockfile, but conan create fails due to the missing requirement. See below:

libsodium (rough draft) 2.X conanfile:

```
from conan import ConanFile
from conan.tools import files as conan_files
from conan.tools.microsoft import msvc_runtime_flag
from conan.tools.apple.apple import is_apple_os
from conans.util.env import get_env
from conan.tools.gnu import Autotools
from conan.tools.gnu import AutotoolsToolchain
from conans.errors import ConanInvalidConfiguration
import os

required_conan_version = "2.0.0-beta2"

class LibsodiumConan(ConanFile):
    name = "libsodium"
    version = "1.0.18"

    description = "A modern and easy-to-use crypto library."
    license = "ISC"
    url = "https://github.com/conan-io/conan-center-index"
[...]

PR #11841, opened 2022-08-11
Add failing test for #11763 

czoido:
Testing: https://github.com/conan-io/conan/issues/11763


memsharded:
Alternatives I can think of:

- For ``lock create .``, construct the graph from the ``test_package/conanfile.py``, exactly the same as ``conan create`` does. This is a bit assymetric with respect to other "local" commands as ``conan graph``, ``conan build``, ``conan install`` that will take the package ``conanfile.py`` as root of the graph, not the test_package one. This approach will effectively lock the test_package "build_requires" in the lockfile, something that I have seen also requested and users struggling in 1.X
- Capture the lockfile as it is doing now, without locking the ``test_package/conanfile.py`` and relax the "strict" attribute for the ``test_package/conanfile.py``, so it doesn't fail when it cannot lock its ``tool_requires`` because they are not in the lockfile.

memsharded:
Update: The 1st approach has a challenge: to load the ``test_package/conanfile.py`` it is necessary to know the ``ref`` of the package. This is known at ``conan create`` time because it is exported first into the cache, and its reference is computed. Without exporting it, it becomes very complicated to obtain the reference. It might also have some other complexities, like locking the self version.

memsharded:
I have submitted a potential solution: make explicit the locking of test_package. Basically the line:
```python
c.run("lock create test_package --lockfile=conan.lock --lockfile-out=conan.lock")
```
Note this is done after capturing the main lockfile, and it will append its ``tool_requires`` to the initial lockfile. The ``self.tested_reference_str`` is None in this case, the graph is not fully expanded, only the new requires declared by ``test_package/conanfile.py``.

Wdyt?

czoido:
> I have submitted a potential solution: make explicit the locking of test_package. Basically the line:
> 
> ```python
> c.run("lock create test_package --lockfile=conan.lock --lockfile-out=conan.lock")
> ```
> 
> Note this is done after capturing the main lockfile, and it will append its `tool_requires` to the initial lockfile. The `self.tested_reference_str` is None in this case, the graph is not fully expanded, only the new requires declared by `test_package/conanfile.py`.
> 
> Wdyt?

For me explicitly locking the test_package would be good enough 👍

memsharded:
The workaround explicitly locking ``test_package`` seems to be good atm.
We probably want to revisit this after https://github.com/conan-io/conan/pull/12046, moving this to beta.4

Issue #11852, opened 2022-08-11
[bug] Targets defined by CMakeDeps generator have incorrect target libraries


andranik3949:
After updating to conan 1.51.2, I started getting link errors when trying to build my project in Visual Studio. This affected all configurations except for Debug. 

After some investigation I noticed that all configurations of some of the dependencies were present in additional dependencies list. (e.g. both `spdlog.lib` and `spdlogd.lib` were present in the project dependencies for Release configuration)

Looking further into it I found that this happens when dependencies have different lib names based on the configuration (e.g. `spdlog.lib` for Release and `spdlogd.lib` for Debug)

The bug is caused by not using generator expressions for `<pkg_name>_LIBRARIES_TARGETS` when defining `INTERFACE_LINK_LIBRARIES` target property in generated CMake files.

This has also been fixed by [this](https://github.com/conan-io/conan/commit/307bb66ba5bb816cdd5768288ea4e277ba7a6388) commit, and then immediately broken by [this](https://github.com/conan-io/conan/commit/91933762620e55e54bb9a9576983a162ae645f13) refactoring commit

### Environment Details (include every applicable attribute)
  * Operating System+version: Windows 10
  * Compiler+version: MSVC 19.33.31629.0
  * Conan version: 1.51.2
  * Python version: 3.8.5

andranik3949:
I'm also willing to contribute a fix for this issue :)

lasote:
Hi, I've reproduced it in a test, I think (further investigation needed) the bug is not in the lines you mention because the `_LIBRARIES_TARGETS` is indeed multi config but there is a bug there generating the IMPORTED libraries. Thanks for reporting, I'll report ASAP and will open a new release process with the fix.

lasote:
The fix has been merged and it will be released at 1.51.3

PR #11859, opened 2022-08-12
Macro `conan_package_library_targets` reverted to single config

lasote:
Changelog: Bugfix: The CMakeDeps generator was not managing correctly the IMPORTED LOCATION of the libraries for different `build_type`.
Docs: omit

Close https://github.com/conan-io/conan/issues/11852

After investigating (thanks @jcar87) looks like the `IMPORTED_LOCATION_<build_type>` doesn't make the targets of the library files multi-config, CMake only uses it as a priority based on other things like the config mappings. More (cryptic) info can be found here: https://gitlab.kitware.com/cmake/cmake/-/issues/16280
So I decided to revert that part of the macro keeping the improvements introduced in the last release, that **so far** have worked ok.

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -80,6 +80,12 @@ class MacrosTemplate(CMakeDepsFileTemplate):
 
            # ONLY FOR DEBUGGING PURPOSES
            foreach(_CONAN_ACTUAL_TARGET ${_CONAN_ACTUAL_TARGETS})
+              get_target_property(imported_location ${_CONAN_ACTUAL_TARGET} IMPORTED_LOCATION_DEBUG)
+              message(VERBOSE "Target Properties: ${_CONAN_ACTUAL_TARGET} IMPORTED_LOCATION_DEBUG ='${imported_location}'")
+
+              get_target_property(imported_location ${_CONAN_ACTUAL_TARGET} IMPORTED_LOCATION_RELEASE)
+              message(VERBOSE "Target Properties: ${_CONAN_ACTUAL_TARGET} IMPORTED_LOCATION_RELEASE ='${imported_location}'")
+
               get_target_property(linked_libs ${_CONAN_ACTUAL_TARGET} INTERFACE_LINK_LIBRARIES)
               message(VERBOSE "Target Properties: ${_CONAN_ACTUAL_TARGET} INTERFACE_LINK_LIBRARIES ='${linked_libs}'")
            endforeach()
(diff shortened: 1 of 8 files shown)


conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -41,10 +41,8 @@ def template(self):
            endif()
        endmacro()
 
-       function(conan_package_library_targets libraries package_libdir deps_target out_libraries_target config package_name)
-           set(_out_libraries "")
+       function(conan_package_library_targets libraries package_libdir deps_target out_libraries_target config_suffix package_name)
            set(_out_libraries_target "")
-           set(_CONAN_ACTUAL_TARGETS "")
```
lasote, line 47: it was not needed, not related with the bug

conan/tools/cmake/cmakedeps/templates/macros.py
```diff
@@ -74,14 +69,8 @@ def template(self):
...
            # Add the dependencies target for all the imported libraries
-           foreach(_CONAN_ACTUAL_TARGET ${_CONAN_ACTUAL_TARGETS})
-               set_property(TARGET ${_CONAN_ACTUAL_TARGET} PROPERTY INTERFACE_LINK_LIBRARIES ${deps_target} APPEND)
-           endforeach()
-
-           # ONLY FOR DEBUGGING PURPOSES
-           foreach(_CONAN_ACTUAL_TARGET ${_CONAN_ACTUAL_TARGETS})
-              get_target_property(linked_libs ${_CONAN_ACTUAL_TARGET} INTERFACE_LINK_LIBRARIES)
-              message(VERBOSE "Target Properties: ${_CONAN_ACTUAL_TARGET} INTERFACE_LINK_LIBRARIES ='${linked_libs}'")
+           foreach(_T ${_out_libraries_target})
```
lasote, line 72: pure refactor to not use `_CONAN_ACTUAL_TARGETS` but `_out_libraries_target`

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -96,7 +96,7 @@ def template(self):
             set_property(TARGET {{root_target_name}}
                          PROPERTY INTERFACE_LINK_LIBRARIES
                          $<$<CONFIG:{{configuration}}>:{{ '${'+pkg_name+'_OBJECTS'+config_suffix+'}' }}>
-                         ${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS}
+                         $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_LIBRARIES_TARGETS}>
```
lasote, line 99: So now we have to "protect" these library targets with the configuration

andranik3949: Nice, this is exactly what I meant! Doing this by locally modifying generated cmake files is exactly how I worked around the issue ^^

memsharded (approved):

jcar87 (approved):
LGTM!!!

PR #11868, opened 2022-08-15
add new members to ConanFileInterface

### LINKED ISSUE #11836: [feature] Access to license and homepage for dependencies in conanfile.py
For some automated license report we would need to have access to the attributes `description`, `license` and `homepage` which can be set in the conan recipe.
As self.deps_cpp_info, self.deps_env_info or self.deps_user_info will be removed in conan 2.0, this should probably be added to [self.dependencies](https://docs.conan.io/en/latest/reference/conanfile/dependencies.html#dependencies-interface).
> Do not use self.deps_cpp_info, self.deps_env_info or self.deps_user_info, these have been removed in 2.0.

During this change it would be great to just have access to all attributes for the dependencies

memsharded:
Changelog: Feature: Add new members to ``ConanFileInterface`` for ``self.dependencies`` access.
Docs: https://github.com/conan-io/docs/pull/2716

Close https://github.com/conan-io/conan/issues/11836

### DIFF
--- a/conans/model/conanfile_interface.py
+++ b/conans/model/conanfile_interface.py
@@ -27,6 +27,10 @@ class ConanFileInterface:
         return not self.__eq__(other)
 
     @property
+    def recipe_folder(self):
+        return self._conanfile.recipe_folder
+
+    @property
     def package_folder(self):
         return self._conanfile.package_folder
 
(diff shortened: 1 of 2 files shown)


jcar87 (approved):
LGTM!

Issue #11862, opened 2022-08-12
[bug] CMakeDeps <package>_INCLUDE_DIR not set in 1.51


jsallay:
### Environment Details (include every applicable attribute)
  * Operating System+version:Ubuntu 20.04 (docker)
  * Compiler+version: gcc9
  * Conan version: 1.51.2
  * Python version: 3.8.10

### Steps to reproduce (Include if Applicable)
It looks like a bug was introduced starting in conan 1.51.0.  I work with some projects that don't always use CMake targets.  They will
sometimes refer to variables like `<package>_INCLUDE_DIR`.  This hasn't caused any problems up until this point.  Today, I upgraded from 1.49.0 to 1.51.2 and my builds stopped working.  Digging through the generated CMake files (using the 2.0 CMakeDeps/ToolChain stuff), and using gmp as an example, gmp-config.cmake contains:
```
set(gmp_INCLUDE_DIRS ${gmp_INCLUDE_DIRS_RELEASE} )
```
In conan version 1.50.2, gmp-release-x86-64-data.cmake contains
```
set(gmp_INCLUDE_DIRS_RELEASE "${gmp_PACKAGE_FOLDER_RELEASE}/include")
```
Starting in version 1.51.0, the `gmp_INCLUDE_DIRS_RELEASE` is not set (in fact many variables are no longer set at the top level in the *-data.cmake files).

Note also that this is not a gmp recipe specific issue.  I am seeing the same behaviour in all the recipes I using (most of which are straight from conan center index).  I am guessing that this is a result of https://github.com/conan-io/conan/pull/11673 but I haven't looked at the code close enough to be sure.

jcar87:
Hi @jsallay  - thank you for reporting this issue! We will investigate this as a possible regression and aim to have a fix in version `1.51.3`.

> I work with some projects that don't always use CMake targets

Out of curiosity, we'd be interested to learn about technical limitations that prevent migration to using imported targets. Targets are more versatile and cover cases that the variable-based approach are not always able to satisfy, and are currently favoured by the CMake maintainers as well. The variables are kept for compatibility purposes for older projects, while targets are the preferred approach.

SpaceIm:
> Out of curiosity, we'd be interested to learn about technical limitations that prevent migration to using imported targets. Targets are more versatile and cover cases that the variable-based approach are not always able to satisfy, and are currently favoured by the CMake maintainers as well. The variables are kept for compatibility purposes for older projects, while targets are the preferred approach.

[...]

PR #11874, opened 2022-08-16
CMakeDeps: Variables for merged global target, even when only components

lasote:
Changelog: Fix: CMakeDeps now generates CMake variables as `XXX_INCLUDE_DIR` in the `XXXX-data.cmake` for the global cpp_info even if components are declared in the recipe. It is obtained by merging the components `cpp_info` into a global one.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/11862


memsharded (approved):
It is really a pity that we cannot move forward to better cmake practices to support that legacy usage, but I guess it is necessary...

jcar87 (approved):

czoido (approved):

Issue #11881, opened 2022-08-16
[question/bug?] create profile for clang and _GLIBCXX_USE_CXX11_ABI


ibis-hdl:
Following [[question] Is libstdc++11 used in Linux clang? #7002](https://github.com/conan-io/conan/issues/7002) Clang is also affected from C++11 ABI library 'problem'.

I my project, I got in (ubuntu 22) CLI
```
$ export CONAN_V2_MODE=1
$ CXX=clang++ conan profile new --detect --force clang
$ CXX=g++ conan profile new --detect --force gcc
$ grep -H libcxx ~/.conan/profiles/[cg]*
/home/vscode/.conan/profiles/clang:compiler.libcxx=libstdc++
/home/vscode/.conan/profiles/gcc:compiler.libcxx=libstdc++11
```
Obviously, this works in Conan v2 mode correct for GCC, but not for Clang, isn't it? Not shure if this is a bug or I do miss something ...

Conan is version 1.51.2

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

jcar87:
Hi @ibis-hdl  - thank you for reporting this. Indeed, on most recent Linux distros that also provide `clang`, it is typical that `clang` uses the GNU C++ Library (libstdc++), which is the one that sets the `_GLIBCXX_USE_CXX11_ABI` macro to `1` by default (from what I have observed, the compiler itself doesn't).

In general (regardless of this), we do recommend to use profiles explicitly rather than relying on auto-detection. If you need to share Conan configuration across multiple users, we recommend using the `conan config` command, you can read about it here: https://docs.conan.io/en/latest/reference/commands/consumer/config.html?highlight=config#conan-config-install 

We have been very cautious to change the defaults in Conan 1.x to prevent unintended breakages for users with existing Conan installations. However, we will issue a fix in Conan 2.0, thanks so much for reporting this!

ibis-hdl:
Thanks for your effort. I assume I have to cut off conan-cmake here, since it relies on auto-detected profiles. The sharm of this is, that I've only one CLI call of the cmake exe ...

memsharded:
Hi @ibis-hdl 

I have checked this, and CONAN_V2_MODE is no longer being maintained in 1.X.
[...]

PR #11886, opened 2022-08-16
[Discussion] Detect libstdc++ ABI mode when using Clang on Linux

jcar87:
For discussion only - 

Clang on Linux (in particular the clang toolchain as provided directly by the distros) will typically fall back to using the GNU C++ library (`libstdc++`), rather than `libc++`.  From what I can see, it is the standard library (in its headers), and _not_ the compiler that defines the a default value for the `_GLIBCXX_USE_CXX11_ABI` macro and determines which ABI tag is used.

However, the Conan autodetection on Linux/clang++ is not taking this into consideration, falling back to the "old" ABI even when it is available, it is the default, and `CONAN_V2_MODE` is explicitly requested.

From https://github.com/conan-io/conan/issues/11881:
```
CONAN_V2_MODE=1 CXX=clang++ conan profile new --detect --force clangprofile
```
this currently detects `libstdc++`. This PR adds the necessary changes so that we run the same detection as we do with `gcc` and report the same result. 

Note, in Conan 2 the detection is correct for g++, but not for clang, this:
```
CXX=clang++ conan profile detect --name foobar
```
will also set it to `compiler.libcxx=libstdc++`.

### DIFF
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -168,17 +168,17 @@ def _get_profile_compiler_version(compiler, version, output):
     return version
 
 
-def _detect_gcc_libcxx(executable, version, output, profile_name, profile_path):
-    # Assumes a working g++ executable
+def _detect_glibcxx_cxx11_abi(executable, version, output, profile_name, profile_path):
+    # Assumes a working g++/clang++ executable
     new_abi_available = Version(version) >= Version("5.1")
     if not new_abi_available:
         return "libstdc++"
 
     if not get_env(CONAN_V2_MODE_ENVVAR, False):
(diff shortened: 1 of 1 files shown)


memsharded (commented):
I think at this point in time, I'd prefer not to target this kind of changes to 1.X releases anymore, but to 2.0 only. To 1.X should only be targeted critical fixes and backport tools necessary for migrating to 2.0. As this issue would be nicely solved with profiles, and it is the recommended approach already, I'd suggest targeting this PR to 2.0 directly.

jcar87:
👍  Will revisit this fix in the 2.0 branch!

Issue #11224, opened 2022-05-11
[feature] Issues with the CMakeToolchain generator


fredizzimo:
We are building custom workspace functionality for Conan 1 based on the existing code, but using CMakeToolchain and the new layout method instead of of the old layout configuration files. It's not that generic at the moment, since it's generating everything dynamically based on our project structure, so it does not need a definition yml file. It's mostly working as it should now, but there are some issues with the toolchain generator.

1. Toolchains are not really suitable for this use case. We want to use the same model that was used in the original workspaces model of using `add_subdirectory`. But that is problematic, since you can only specify the toolchain for the root level cmake file. Our current workaround for this is to just include the toolchain file, which seems to work, except for problem 2. But this is not how cmake toolchains should be used, and it's probably not guaranteed to work in the future, if more stuff is added to the toolchain. Additionally I'm not so sure about the whole idea of using toolchains for setting normal variables and project configuration, IMO toolchains should be reserved for what they actually are meant for, compiler toolchains and stuff. I think that we need a new additional generator that generates just plain cmake files with the project configuration, that can be then included manually. Then  we could include a toolchain file at the top level, if needed, and this generated cmake file in each subproject.  And when the package is actually exported, it would use both files.
1. The variables are set to the cache and not force overwritten. In our case this means that you can't use the same variable in two different projects, which means that they can't be cached. Other use cases need the cache see, https://github.com/conan-io/conan/issues/7832, but even in those cases I believe that the variables should be force set, otherwise you have to delete the cache in order to be able to change the variable from the recipe. So you should probably provide an interface for setting different variable types.
1. You can't provide a template file to the generator, only a static toolchain file. If a template file was allowed, then we would at least have a quite elegant workaround for the issue, by making a custom template that just sets normal variables instead of cache ones. Custom blocks could also be an alternative, but sadly an interface for setting that does not seem to be available either. So I think my current workaround would be to just make a custom class that overrides the internal `_template`, that's not elegant, nor future proof, but at least I think it will work.

I was also considering using cmake external projects instead of `add_subdirectory`, but sadly, the IDE integration for that is not great, so I'm not sure if that's a path I'm willing to take. I have read most of the issues related to workspaces here, and some people are suggesting that, but if you go that way with the official conan 2 integration, you should definitely make sure that the ide integration works properly. 

[...]

PR #11902, opened 2022-08-17
Fix require_overrides for virtual packages

fredizzimo:
Changelog: Bugfix: Workspaces with multiple root nodes did not respect require_overrides, but now they are passed correctly

When testing my custom workspace functionality, I noticed that the `require_overrides` were not passed along, which seems like a mistake, since it's passed with all other types of root packages. So this fixes that.

### DIFF
--- a/conans/client/graph/graph_manager.py
+++ b/conans/client/graph/graph_manager.py
@@ -151,7 +151,8 @@ class GraphManager(object):
         profile_host.dev_reference = create_reference  # Make sure the created one has develop=True
 
         if isinstance(reference, list):  # Install workspace with multiple root nodes
-            conanfile = self._loader.load_virtual(reference, profile_host, scope_options=False)
+            conanfile = self._loader.load_virtual(reference, profile_host, scope_options=False,
+                                                  require_overrides=require_overrides)
             # Locking in workspaces not implemented yet
             return Node(ref=None, context=CONTEXT_HOST, conanfile=conanfile, recipe=RECIPE_VIRTUAL)
 


memsharded (commented):
Thanks for your contribution.

Could you please try to provide a test that is fixed by this? Much better if it doesn't use the workspace feature, as it has been removed from Conan 2.0 (to be replaced by user scripts until we resume work on workspaces in 2.X). Ask for guidance if you need it, thanks!

fredizzimo:
Yes, of course. I will try to find a way to do that.

fredizzimo:
Hi @memsharded,

I think I need some help with the testing. I found similar tests to what I need in `install_test.py`. The problem is that they all test the public cli interface, and as far as I can see there's no way to create a "virtual package" through that currently. 

I'm able to to call the api `install_reference` (passing an array of references) to test the functionality of this pull request. However, the TestClient does not have any functionality for capturing the output when doing that. I also don't see any other tests that are testing the API directly for an example. 

I also don't see any similar tests at the unit test, or functional level, and I'm not even sure if testing this is possible at that level, at least not without heavy mocking.

So, I see a couple of alternatives,
1. Add a `run_api` function to the `TestClient`, which works exactly as testing the command line. 
2. Add the possibility of specifying multiple references to the install command. But that's a bit bigger change, which at least need additional tests, and additionally might not be something you want to expose.

What do you think? I'm not familiar with the code so you probably have some better suggestions.

memsharded:
I see what you mean. I was possibly recalling the Conan 2.0 codebase, where you can do ``conan install --requires=pkga/1.0 --requires=pkgb/2.0``, so it is easier to fire this behavior without workspaces. You are probably right, and it is not possible to test this without workspaces, the only public cli using this api is the workspaces.

Maybe we can just merge it as-is. If you have tested the fix locally, and nothing else breaks, we might be good, workspaces are removed in 2.0 anyway (so I'd recommend to try to move out of workspaces to a custom script and then feed that feedback when we resume work on 2.0 new workspaces)

fredizzimo:
Yes, this fix works for me locally, and the existing tests passes here in this pull request, so it should be fairly safe to merge. But the call is up to you. 

Note, that we have some completely custom workspace functionality locally. It uses cmake `add_subdirectory`, and has a custom CMakeDeps generator, like discussed here https://github.com/conan-io/conan/issues/11224. This works very well for our use cases.  It also uses a custom API function by deriving from ConanAPI, so that I'm able to use temporary editable packages, and I'm able to use the public released conan version.

Unfortunately I'm not able to share the code right now, but at some point, when it has been proven a bit more on our side, I would want to work on enabling that functionality entirely through the public conan API. I think that an install command that can take a list of packages to put into editable mode (but don't actually install or compile those), and a list of root package references would be enough, in addition to the toolchain changes of course. The root CMakeLists.txt file with all the add_subdirectory calls can then be done from the outside, without using conan.

If this kind of functionality was provided then we could also add tests for these virtual packages.

memsharded:
Hi @fredizzimo 

I am very sorry, it seems this PR fell through the cracks (It sometimes happen due to high load, and also during the summer that many of the team take some vacation...)

I guess that you found a workaround around this, so maybe this is no longer needed? Also, this was for 1.X, and we are avoiding doing changes to 1.X unless very strictly necessary. Please let me know if we can close this PR, or you'd still want to move it forward.

fredizzimo:
@memsharded

Yes, indeed we have a workaround. We used a very dirty hack of patching the function directly from our code.

Now when we are in progress of updating to Conan 2.0, that hack does not seem to be necessary, `api.graph.load_graph_requires` works as it should.

Closing this.

memsharded:
For workspace feature in 2.0 it would be great to count with your feedback, this is the ticket that you should follow to track progress when we resume this for 2.X roadmap: https://github.com/conan-io/conan/issues/12466. Thanks again!

PR #11908, opened 2022-08-18
Read required_conan_version in text mode before loading the recipe

### LINKED ISSUE #11239: [bug] `required_conan_version` is not evaluated before python imports
  * Compiler+version: apple-clang 13.1
  * Conan version: 1.48.0
  * Python version: 3.9.12

see https://github.com/conan-io/conan-center-index/pull/10633#issuecomment-1124783149

- use conan < 1.45.0, for example 1.44.0
- `conan install yaml-cpp/0.6.3`
- Instead of the meaningful message `ERROR: Error loading conanfile at (...): Current Conan version (1.44.0) does not satisfy the defined one (>=1.45.0).` we could expect due to `required_conan_version = ">=1.45.0"`, there are python imports error.

log from https://github.com/conan-io/conan-center-index/pull/10633#issuecomment-1124783149:

```
00:00:22  yaml-cpp/0.6.3: Retrieving from server 'conancenter'
00:00:22  yaml-cpp/0.6.3: Trying with 'conancenter'...
00:00:22  Downloading conanmanifest.txt
00:00:22  Downloading conanfile.py
00:00:22  Downloading conan_export.tgz
00:00:22  yaml-cpp/0.6.3: Downloaded recipe revision 0
00:00:22  ERROR: yaml-cpp/0.6.3: Cannot load recipe.
00:00:22  Error loading conanfile at '***/CONAN_CACHE_arm64/.conan/data/yaml-cpp/0.6.3/_/_/export/conanfile.py': Unable to load conanfile in ***/CONAN_CACHE_arm64/.conan/data/yaml-cpp/0.6.3/_/_/export/conanfile.py
00:00:22    File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/imp.py", line 171, in load_source
00:00:22      module = _load(spec)
00:00:22    File "<frozen importlib._bootstrap>", line 711, in _load
00:00:22    File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
00:00:22    File "<frozen importlib._bootstrap_external>", line 850, in exec_module
00:00:22    File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
00:00:22    File "***/CONAN_CACHE_arm64/.conan/data/yaml-cpp/0.6.3/_/_/export/conanfile.py", line 1, in <module>
00:00:22      from conan.tools.microsoft import is_msvc, is_msvc_static_runtime
00:00:22  ImportError: cannot import name 'is_msvc' from 'conan.tools.microsoft' (***/lib/python3.9/site-packages/conan/tools/microsoft/__init__.py)
```

lasote:
Changelog: Feature: Fail sooner and with a meaningful error if the specified required version is not satisfied.
Docs: omit

Conan will try to read the `required_conan_version` from the recipes without loading the recipe (python interpreter) to be able to check it even if there are imports broken/missing for the user conan current installed version. So Conan can fail sooner and with a meaningful error if the specified required version is not satisfied.

Close https://github.com/conan-io/conan/issues/11239

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -434,20 +434,23 @@ def _parse_conanfile(conan_file_path):
     current_dir = os.path.dirname(conan_file_path)
     sys.path.insert(0, current_dir)
     try:
-        version_txt = get_required_conan_version_without_loading(conan_file_path)
-        if version_txt:
-            validate_conan_version(version_txt)
-
         old_modules = list(sys.modules.keys())
         with chdir(current_dir):
             old_dont_write_bytecode = sys.dont_write_bytecode
             sys.dont_write_bytecode = True
-            loaded = imp.load_source(module_id, conan_file_path)
-            sys.dont_write_bytecode = old_dont_write_bytecode
+            try:
+                # FIXME: imp is deprecated in favour of implib
+                loaded = imp.load_source(module_id, conan_file_path)
+            except ImportError:
+                version_txt = get_required_conan_version_without_loading(conan_file_path)
+                if version_txt:
+                    validate_conan_version(version_txt)
+                raise
 
-        required_conan_version = getattr(loaded, "required_conan_version", None)
-        if required_conan_version:
-            validate_conan_version(required_conan_version)
+            required_conan_version = getattr(loaded, "required_conan_version", None)
+            if required_conan_version:
+                validate_conan_version(required_conan_version)
+            sys.dont_write_bytecode = old_dont_write_bytecode
 
         # These lines are necessary, otherwise local conanfile imports with same name
         # collide, but no error, and overwrite other packages imports!!
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -2,6 +2,7 @@ import fnmatch
 import imp
 import inspect
 import os
+import re
 import sys
 import types
 import uuid
@@ -433,6 +434,10 @@ def _parse_conanfile(conan_file_path):
     current_dir = os.path.dirname(conan_file_path)
     sys.path.insert(0, current_dir)
     try:
+        version_txt = get_required_conan_version_without_loading(conan_file_path)
+        if version_txt:
+            validate_conan_version(version_txt)
+
         old_modules = list(sys.modules.keys())
         with chdir(current_dir):
             old_dont_write_bytecode = sys.dont_write_bytecode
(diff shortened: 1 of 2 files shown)


czoido (approved):

conans/client/loader.py
```diff
@@ -433,6 +434,10 @@ def _parse_conanfile(conan_file_path):
     current_dir = os.path.dirname(conan_file_path)
     sys.path.insert(0, current_dir)
     try:
+        version_txt = get_required_conan_version_without_loading(conan_file_path)
```
memsharded: I don't love it that we are running a regex against potentially long conanfiles just for this corner case. This will also validate 2 times the same thing, later it is evaluated again.

When we first discussed this issue, you suggested that this would be only checked on ``ImportError`` exceptions, not always, that seemed a bit better.

In any case, if we add this, I would expect it to be only 1.X and drop it in 2.0, is that the plan?

lasote: But in 2.0 there will be the same issue, right? Why introduce it to 1.X then? Why not just drop this? Take into account that in 2.2 we could introduce a new tool and importing it would cause an Import Error too in 2.0 clients.

> When we first discussed this issue, you suggested that this would be only checked on ImportError exceptions, not always, that seemed a bit better.

Yeah, that's true.

memsharded: Yes, that is true, but way more unlikely. We have hit this issue because of the massive migration of imports to new ``conan.tools`` ones, but didn't have this issue in the last 4 years, so it seems that we wouldn't have added it if it wasn't for the migration. In any case, if we are doing it only on import error, I am fine keeping it in 2.0.

conans/client/loader.py
```diff
@@ -437,13 +438,20 @@ def _parse_conanfile(conan_file_path):
         with chdir(current_dir):
             old_dont_write_bytecode = sys.dont_write_bytecode
             sys.dont_write_bytecode = True
-            loaded = imp.load_source(module_id, conan_file_path)
+            try:
+                # FIXME: imp is deprecated in favour of implib
+                loaded = imp.load_source(module_id, conan_file_path)
```
memsharded, line 443: I think I would leave the ``sys.dont_write_bytecode`` around the ``load_source()``, otherwise it might be left in inconsistent state if there is a validate version exception raised.

conans/client/loader.py
```diff
@@ -475,3 +483,20 @@ def _parse_conanfile(conan_file_path):
         sys.path.pop(0)
 
     return loaded, module_id
+
+
+def get_required_conan_version_without_loading(conan_file_path):
```
memsharded:
```suggestion
def _get_required_conan_version_without_loading(conan_file_path):
```

conans/client/loader.py
```diff
@@ -437,13 +438,20 @@ def _parse_conanfile(conan_file_path):
         with chdir(current_dir):
             old_dont_write_bytecode = sys.dont_write_bytecode
             sys.dont_write_bytecode = True
-            loaded = imp.load_source(module_id, conan_file_path)
+            try:
+                # FIXME: imp is deprecated in favour of implib
+                loaded = imp.load_source(module_id, conan_file_path)
+            except ImportError:
+                version_txt = get_required_conan_version_without_loading(conan_file_path)
```
memsharded:
```suggestion
                version_txt = _get_required_conan_version_without_loading(conan_file_path)
```

memsharded (approved):
Much better only at import error time.

PR #11911, opened 2022-08-18
Add export_conandata_patches utility

### LINKED ISSUE #11770: [feature] Add a function `conan.tools.files.export_conandata_patches`

In conan-center-index, several versions share the same `conanfile.py` but may have different patches depending on the version. If we export patches with `exports_sources = "patches/**"`, each version ends up with patches of unrelated versions, and adding or modifying a patch of a version can change RREV of other versions.

So in conan-center-index, we have introduced this kind of pattern:

```python
    def export_sources(self):
        for p in self.conan_data.get("patches", {}).get(self.version, []):
            copy(self, p["patch_file"], self.recipe_folder, self.export_sources_folder)
```

It would be nice to provide a function `conan.tools.files.export_conandata_patches` similar to `conan.tools.files.apply_conandata_patches`, to avoid this boilerplate.

### LINKED ISSUE #11923: [question] apply_conandata_patches handling some types
Current implementation is:
```python
    if isinstance(patches, dict):
        assert conanfile.version, "Can only be applied if conanfile.version is already defined"
        entries = patches.get(str(conanfile.version), [])
    elif isinstance(patches, Iterable):
        entries = patches
    else:
        return
```
So it will silently skip ``patches`` field in conandata.yml with unexpected types? If the type is a string (possible user mistake), it will fail uglyly as considered ``Iterable``?

MartinDelille:
Changelog: Feature: Add `export_conandata_patches` tool.
Docs: https://github.com/conan-io/docs/pull/2720

Closes: https://github.com/conan-io/conan/issues/11770
Closes: https://github.com/conan-io/conan/issues/11923

This issue has no test because I didn't managed to make it work but you can see this as a quick proposal before going further in the implementation.

### DIFF
--- a/conan/tools/files/patches.py
+++ b/conan/tools/files/patches.py
@@ -118,3 +118,27 @@ def apply_conandata_patches(conanfile):
         else:
             raise ConanException("The 'conandata.yml' file needs a 'patch_file' or 'patch_string'"
                                  " entry for every patch to be applied")
+
+
+def export_conandata_patches(conanfile):
+    """
+    Exports patches stored in 'conanfile.conan_data' (read from 'conandata.yml' file). It will export
+    all the patches under 'patches' entry that matches the given 'conanfile.version'. If versions are
+    not defined in 'conandata.yml' it will export all the patches directly under 'patches' keyword.
+    """
+
+    patches = conanfile.conan_data.get('patches')
+
+    if isinstance(patches, dict):
+        assert conanfile.version, "Can only be exported if conanfile.version is already defined"
+        entries = patches.get(conanfile.version, [])
+    elif isinstance(patches, Iterable):
+        entries = patches
+    else:
+        return
+    for it in entries:
+        if "patch_file" in it:
+            # The patch files are located in the root src
+            entry = it.copy()
+            patch_file = os.path.join(conanfile.folders.base_source, entry.pop("patch_file"))
+            conanfile.copy(patch_file)
--- a/conan/tools/files/patches.py
+++ b/conan/tools/files/patches.py
@@ -1,14 +1,11 @@
 import logging
 import os
+import shutil
 
-import patch_ng, copy
+import patch_ng
 
 from conans.errors import ConanException
-
-try:
-    from collections.abc import Iterable
-except ImportError:
-    from collections import Iterable
+from conans.util.files import mkdir
 
 
 class PatchLogHandler(logging.Handler):
@@ -97,16 +94,20 @@ def apply_conandata_patches(conanfile):
         - patch_file: "patches/0001-buildflatbuffers-cmake.patch"
     ```
     """
+    if conanfile.conan_data is None:
+        raise ConanException("conandata.yml not defined")
 
     patches = conanfile.conan_data.get('patches')
+    if not patches:
+        raise ConanException("No patches defined in conandata")
 
     if isinstance(patches, dict):
         assert conanfile.version, "Can only be applied if conanfile.version is already defined"
         entries = patches.get(conanfile.version, [])
-    elif isinstance(patches, Iterable):
+    elif isinstance(patches, list):
         entries = patches
     else:
-        return
+        raise ConanException("conandata.yml 'patches' should be a list or a dict {version: list}")
     for it in entries:
         if "patch_file" in it:
             # The patch files are located in the root src
@@ -126,19 +127,24 @@ def export_conandata_patches(conanfile):
     all the patches under 'patches' entry that matches the given 'conanfile.version'. If versions are
     not defined in 'conandata.yml' it will export all the patches directly under 'patches' keyword.
     """
+    if conanfile.conan_data is None:
+        raise ConanException("conandata.yml not defined")
 
     patches = conanfile.conan_data.get('patches')
+    if not patches:
+        raise ConanException("No patches defined in conandata")
 
     if isinstance(patches, dict):
         assert conanfile.version, "Can only be exported if conanfile.version is already defined"
         entries = patches.get(conanfile.version, [])
-    elif isinstance(patches, Iterable):
+    elif isinstance(patches, list):
         entries = patches
     else:
-        return
+        raise ConanException("conandata.yml 'patches' should be a list or a dict {version: list}")
     for it in entries:
-        if "patch_file" in it:
-            # The patch files are located in the root src
-            entry = it.copy()
-            patch_file = os.path.join(conanfile.folders.base_source, entry.pop("patch_file"))
-            copy(conanfile, patch_file, conanfile.recipe_folder, conanfile.export_sources_folder)
+        patch_file = it.get("patch_file")
+        if patch_file:
+            src = os.path.join(conanfile.recipe_folder, patch_file)
+            dst = os.path.join(conanfile.export_sources_folder, patch_file)
+            mkdir(os.path.dirname(dst))
+            shutil.copy2(src, dst)
--- a/conan/tools/files/patches.py
+++ b/conan/tools/files/patches.py
@@ -99,7 +99,8 @@ def apply_conandata_patches(conanfile):
 
     patches = conanfile.conan_data.get('patches')
     if patches is None:
-        raise ConanException("No patches defined in conandata")
+        conanfile.output.info("apply_conandata_patches(): No patches defined in conandata")
+        return
 
     if isinstance(patches, dict):
         assert conanfile.version, "Can only be applied if conanfile.version is already defined"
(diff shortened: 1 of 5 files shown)


MartinDelille:
@jgsogo I'll be happy to have your feedback on this feature that would replace this two lines by a simpler `export_conandata_patches(self)`. Feel free to ping anyone who would be interested in this.

jgsogo:
I suggested it too on my side... just to discover that @SpaceIm already suggested some days ago. IIRC it is assigned to 1.53

jgsogo:
Anyway, I don't know if there's already a PR for it. When I'm back on my computer I'll have a look to the implementation

SpaceIm:
related issue: https://github.com/conan-io/conan/issues/11770

conan/tools/files/patches.py
```diff
@@ -118,3 +118,27 @@ def apply_conandata_patches(conanfile):
...
+    elif isinstance(patches, Iterable):
+        entries = patches
+    else:
+        return
+    for it in entries:
+        if "patch_file" in it:
+            # The patch files are located in the root src
+            entry = it.copy()
+            patch_file = os.path.join(conanfile.folders.base_source, entry.pop("patch_file"))
+            conanfile.copy(patch_file)
```
SpaceIm:
```suggestion
            conanfile.copy(patch_file)
```

I guess `conan.tools.files.copy` is more appropriate.

conan/tools/files/patches.py
```diff
@@ -118,3 +118,27 @@ def apply_conandata_patches(conanfile):
...
+
+    patches = conanfile.conan_data.get('patches')
+
+    if isinstance(patches, dict):
+        assert conanfile.version, "Can only be exported if conanfile.version is already defined"
+        entries = patches.get(conanfile.version, [])
+    elif isinstance(patches, Iterable):
+        entries = patches
+    else:
+        return
```
memsharded: No silent return if it is a string for example. If it is None when not defined, return earlier.

memsharded (commented):
This would need some integration tests

conan/tools/files/patches.py
MartinDelille: > No silent return if it is a string for example.

What do you mean by that ? I know patch can be string then not exported then.

> If it is None when not defined, return earlier.

Why the behaviour should be different from `apply_conandata_patches` ?

conans/test/functional/tools/test_files.py
```diff
@@ -340,3 +340,28 @@ def build(self):
...
+        """)
+    conandata_yml = textwrap.dedent("""
+        patches:
+          - patch_file: "patches/0001-buildflatbuffers-cmake.patch"
+            base_path: "relative_dir"
+    """)
+
+    client = TestClient()
+    client.save({"conanfile.py": conanfile})
+    client.run("create .")
```
MartinDelille, line 399: Here is a first integration test failing locally:

```
E           Exception:
E           create .
E           [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'url'. It is recommended to add it as attribute
E           [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'license'. It is recommended to add it as attribute
E           [HOOK - attribute_checker.py] pre_export(): WARN: Conanfile doesn't have 'description'. It is recommended to add it as attribute
E           Exporting package recipe
E           mypkg/1.0: Calling export_sources()
E           ERROR: mypkg/1.0: Error in export_sources() method, line 13
E           	export_conandata_patches(self)
E           	AttributeError: 'NoneType' object has no attribute 'get'
E
E           --------------------------------------------------------------------------------
```

conan/tools/files/patches.py
memsharded: Oh, I wasn't aware that the ``apply_conandata_patches`` had this behavior. Then it can be left the same in this PR, and I will create an issue to check this.

memsharded:
We would like this to be in 1.52 soon, so I have contributed a review, including completing the test and more error checking.

MartinDelille:
Thanks!

conan/tools/files/patches.py
```diff
@@ -97,16 +94,20 @@ def apply_conandata_patches(conanfile):
         - patch_file: "patches/0001-buildflatbuffers-cmake.patch"
     ```
     """
+    if conanfile.conan_data is None:
+        raise ConanException("conandata.yml not defined")
 
     patches = conanfile.conan_data.get('patches')
+    if patches is None:
+        raise ConanException("No patches defined in conandata")
```
SpaceIm: Why raising now? It should be no-op.
A recipe may want to always call `export_conandata_patches`, then `apply_conandata_patches`, regardless of conandata content, so that the recipe is generic and ready to new contributions adding patches.
Moreover the reducer removes `patches` of the conandata.yml specific version in local cache if there is no patch, so I guess it would break many recipes.

memsharded: > A recipe may want to always call export_conandata_patches, then apply_conandata_patches, regardless of conandata content, so that the recipe is generic and ready to new contributions adding patches.

I don't agree with this, this would end with all recipes in ConanCenter adding these lines, just in case it is necessary to apply a patch in the future? That adds unnecessary clutter to recipes. If we are not adding these lines to all recipes, then it will only be added to recipes that have patches, and accordingly have a ``patches`` field defined in the ``conandata.yml``. I think using a method ``export_conandata_patches()`` in a recipe implies a pre-condition that there are ``patches`` in ``conandata`` to export, otherwise, why are you calling that method?

In some exceptional case, if you want to be future-ready, you can still define the ``patches: {}`` in the conadata.yml

SpaceIm: But it raises now in `apply_conandata_patches` as well, and as I said the reducer removes `patches` field when there is no patch for a specific version (there might be patches for a version and not for others, this is why we fallback to empty dic/list in each `get` in the loop we do in CCI recipes).

SpaceIm: > In some exceptional case, if you want to be future-ready, you can still define the patches: {} in the conadata.yml

The advantage to be future-ready is to avoid to modify `conanfile.py`, and thus RREV of versions not affected by a new patch file.

memsharded: But the code is only requesting that there is a field ``patches`` in the ``conandata.yml``. Not that it is filled with patches. It is fine if some version doesn't have patches. But if I am using a ``xxx_conandata_patches``, the expectation is that the ``conandata.yml`` will contain a ``patches`` section, and not a ``patchs`` or a ``patch`` or a ``pathces`` one (or no one).

SpaceIm: There is no `patches` field in the `conandata.yml` in local cache after a `conan create`/`conan export` if the specific exported version has no patch.

Please test with cjson in CCI for example.
See the conandata.yml: https://github.com/conan-io/conan-center-index/blob/624d6cfa702eecd4f0af9d32c1d33e4e409f7631/recipes/cjson/all/conandata.yml
There is a `patches` field, but only a patch for 1.7.12. Try to export 1.7.15, check conandata.yml in local cache: no `patches` field.

memsharded: Ok, I see the problem. The ConanCenter "reducer" hook, is completely dropping the keys when the version is not defined. This is unfortunate, I didn't expect this, I think the "reducer" should reduce things to the current version information, but not completely drop a full section. 

I'll relax that assumption.

memsharded: Changed, now it will only print one informative message telling that there are no patches to apply (that might help to debug some cases)

memsharded: Looking good now @SpaceIm?

conan/tools/files/patches.py
```diff
@@ -118,3 +120,33 @@ def apply_conandata_patches(conanfile):
...
+        entries = patches
+    else:
+        raise ConanException("conandata.yml 'patches' should be a list or a dict {version: list}")
+    for it in entries:
+        patch_file = it.get("patch_file")
+        if patch_file:
+            src = os.path.join(conanfile.recipe_folder, patch_file)
+            dst = os.path.join(conanfile.export_sources_folder, patch_file)
+            mkdir(os.path.dirname(dst))
+            shutil.copy2(src, dst)
```
SpaceIm, line 152: if an element of the list doesn't contain a `patch_file` field, should it raise or display a warning?

memsharded: According to the logic, elements in the list can be string patches, if they define ``patch_string``, so it seems valid to skip it without warning or error

SpaceIm: Ok I see, so LGTM

conan/tools/files/patches.py
SpaceIm: yes

jcar87 (approved):
LGTM!

PR #11915, opened 2022-08-19
Promote to_apple_arch in conan.tools.apple

### LINKED ISSUE #11865: [feature] Expose `is_apple_os` & `to_apple_arch` in `conan.tools.apple`

`is_apple_os` is widely used in conan-center-index (almost 10% of recipes), it prevents their full migration to conan v2.

jcar87:
Changelog: Feature: Promote `to_apple_arch` in the new `conan.tools.apple` module. 
Docs: https://github.com/conan-io/docs/pull/2722

Closes https://github.com/conan-io/conan/issues/11865

Notes:
* Where possible, have replaced uses of `conan.tools.apple.apple.to_apple_arch` to the new one taking a `conanfile` as an argument
* Elsewhere with this wasn't possible (because there's no `conanfile` object in the context), have made those call the private implementation.
* Have added a barebones test to cover this

### DIFF
--- a/conan/tools/apple/__init__.py
+++ b/conan/tools/apple/__init__.py
@@ -3,7 +3,6 @@
 # from conan.tools.apple.apple import apple_dot_clean
 # from conan.tools.apple.apple import apple_sdk_name
 # from conan.tools.apple.apple import apple_deployment_target_flag
-# from conan.tools.apple.apple import to_apple_arch
 from conan.tools.apple.apple import fix_apple_shared_install_name, is_apple_os, to_apple_arch
 from conan.tools.apple.xcodedeps import XcodeDeps
 from conan.tools.apple.xcodebuild import XcodeBuild
--- a/conan/tools/apple/__init__.py
+++ b/conan/tools/apple/__init__.py
@@ -4,7 +4,7 @@
 # from conan.tools.apple.apple import apple_sdk_name
 # from conan.tools.apple.apple import apple_deployment_target_flag
 # from conan.tools.apple.apple import to_apple_arch
-from conan.tools.apple.apple import fix_apple_shared_install_name, is_apple_os
+from conan.tools.apple.apple import fix_apple_shared_install_name, is_apple_os, to_apple_arch
 from conan.tools.apple.xcodedeps import XcodeDeps
 from conan.tools.apple.xcodebuild import XcodeBuild
 from conan.tools.apple.xcodetoolchain import XcodeToolchain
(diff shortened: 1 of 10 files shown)


conan/tools/apple/__init__.py
```diff
@@ -4,7 +4,7 @@
 # from conan.tools.apple.apple import apple_sdk_name
 # from conan.tools.apple.apple import apple_deployment_target_flag
 # from conan.tools.apple.apple import to_apple_arch
```
lasote:
```suggestion

```

lasote (approved):

conan/tools/apple/apple.py
```diff
@@ -11,16 +12,10 @@ def is_apple_os(conanfile):
...
     """converts conan-style architecture into Apple-style arch"""
-    return {'x86': 'i386',
-            'x86_64': 'x86_64',
-            'armv7': 'armv7',
-            'armv8': 'arm64',
-            'armv8_32': 'arm64_32',
-            'armv8.3': 'arm64e',
-            'armv7s': 'armv7s',
-            'armv7k': 'armv7k'}.get(str(arch))
+    arch_ = conanfile.settings.get_safe("arch")
```
memsharded, line 17: I am slightly concerned about some recent tools similar to this, as I am wondering what will happen if at some point users want to use it for the build context. They will not be able. So the options would be:
- To expose the underlying dict, something like ``apple_archs(arch)``
- To create a new ``to_apple_arch_build(conanfile)`` (a bit ugly?)

I am fine with this change at this moment, but something to have in mind and maybe discuss.

jcar87: That's a good point! If there's a risk that we may end up with too many small tool functions that have a similar issue, I'd be more inclined to try and sort this out sooner rather than later. `to_apple_arch` has very few uses in Conan Center at the moment, so we may have an opportunity to get it right to cover all cases (although we also want to avoid a repeat of the situations where recipes are using it from the "private" module `apple.apple`)

SSE4: there are some other alternatives, like:
- change context in scope via RAII, e.g. `with context_switch(ctx="build"): do something`
- provide a nice moniker like `if is_apple_arch(conanfile.build)` where `conanfile.build` is `ConanFIle` instance with context switched to build

memsharded: Related https://github.com/conan-io/conan/issues/11926

memsharded (approved):
We will move forward with this approach, and add a ``context_build=False`` argument in the future if need to apply to the build context.

PR #11917, opened 2022-08-19
CMakeUserPresets custom locations

### LINKED ISSUE #11707: [feature] location of the generated CMakeUserPresets.json
The Conan documentation states:

> If you declare a layout() in the recipe and your CMakeLists.txt file is found in the conanfile.source_folder folder, Conan will place a CMakeUserPresets.json to include the CMakePresets.json generated with the above specification.

However, there is no way to set the source folder from the `conan install` command. Besides the obvious use case to override the defaults from the command line, such an option could also be useful in big projects to select which directory with CMakeLists.txt should be used as an entry point for CMake build. For example, if I have several standalone subprojects in the tree and only one conanfile.py to define the dependencies with `conan install --source-folder` I could control where the `CMakeUserPresets.json` is being generated and then build such a subproject in isolation using CMake presets.

Alternatively, `CMakeUserPresets.json` could be generated in the path of `conan install` invocation, but probably this is a worse solution.

### LINKED ISSUE #12277: [bug] CMakeToolChain.generate + Conan 1.53 changed functionality
Hi,
after upgrading Conan 1.53 CMakeToolChain.generate functionality appears to be changed.
Suddenly a "CMakeUserPresets.json" file started to appear in the package's source folder, which practically breaks Conan lifecycle, as the .git repo becames dirty, "conan export" doesn't update the "scm" "auto" attributes, and surely it is not possible to do "conan upload" anymore. 

Any suggestion?
Thanks,

lasote:
Changelog: Feature: Add `.user_presets_path` attribute to `CMakeToolchain` to customize the location of `CMakeUserPresets.json` or to skip generating it.
Docs: https://github.com/conan-io/docs/pull/2825

The location of the generated `CMakeUserPresets.json` can be customized in the `CMakeToolchain` using the attribute `.user_presets_path` to an absolute path or a relative to `self.source_folder`. If `.user_presets_path` is set to `False` it will skip the generation of the file.

Close https://github.com/conan-io/conan/issues/11707
Close: https://github.com/conan-io/conan/issues/12277

### DIFF
--- a/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
+++ b/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
@@ -820,3 +820,28 @@ def test_user_presets_custom_location(path):
     else:
         assert not os.path.exists(os.path.join(client.current_folder, "CMakeUserPresets.json"))
         assert not os.path.exists(os.path.join(client.current_folder, "False", "CMakeUserPresets.json"))
+
+
+def test_set_cmake_lang_compilers_and_launchers():
+    profile = textwrap.dedent(r"""
+    [settings]
+    os=Windows
+    arch=x86_64
+    compiler=clang
+    compiler.version=15
+    compiler.libcxx=libstdc++11
+    [conf]
+    tools.build:compiler_executables={"c": "/my/local/gcc", "cpp": "g++", "rc": "C:\\local\\rc.exe"}
+    """)
+    client = TestClient(path_with_spaces=False)
+    conanfile = GenConanfile().with_settings("os", "arch", "compiler")\
+        .with_generator("CMakeToolchain")
+    client.save({"conanfile.py": conanfile,
+                "profile": profile})
+    client.run("install . -pr:b profile -pr:h profile")
+    toolchain = client.load("conan_toolchain.cmake")
+    assert 'set(CMAKE_C_COMPILER clang)' not in toolchain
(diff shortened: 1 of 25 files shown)


conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
```diff
@@ -719,3 +720,42 @@ def layout(self):
...
+                import os
+                from conan import ConanFile
+                from conan.tools.cmake import cmake_layout, CMakeToolchain
+
+                class Conan(ConanFile):
+                    settings = "os", "arch", "compiler", "build_type"
+
+                    def generate(self):
+                        t = CMakeToolchain(self)
+                        t.user_presets_path = os.path.join(self.source_folder, "subproject")
```
memsharded: It would be nice to allow relative path here too ``t.user_presets_path = "subproject"``

memsharded (commented):
As the location of the presets files is kind of fixed, I don't see much the purpose of the ``conf``, as it seems the location should be defined by the CMakeToolchain.

Maybe once the ``user_presets_path`` is defined it should force the creation of the file, without checking the other conditions? And maybe the opposite, a ``=False`` value guarantees that it will never be created?

czoido (approved):

franramirez688 (approved):

jcar87 (approved):
LGTM! It addresses the original issue, but also gives a way out of generating the file altogether, as the presence of the file may cause some IDEs to default to using it

memsharded (approved):

PR #11920, opened 2022-08-19
[Meson] Removing reconfigure from Meson.install()

### LINKED ISSUE #11910: [bug] `install` method of `conan.tools.meson.Meson` calls `meson setup --reconfigure` before installing
  * Compiler+version: AppleClang 13
  * Conan version: 1.51.2
  * Python version: 3.10.6

- checkout a recipe based on `conan.tools.meson.MesonToolchain` & `conan.tools.meson.Meson` (for example in https://github.com/conan-io/conan-center-index/pull/12300)
- run `conan create`

During install step, an unnecessary reconfiguration is executed (it can be quite long depending on the project). So the behavior differs from CMake helper where no reconfiguration is done when install method is called.

https://c3i.jfrog.io/c3i/misc/logs/pr/12300/1-linux-gcc/dav1d/1.0.0//79ab1d7c838f7fd4d71114687478abc26e7bc7ef-build.txt

franramirez688:
Changelog: Feature: Removed opt-out `meson.configure(reconfigure=True)` from `Meson.install()` function. Now, it only configures the `prefix` path without reconfiguring.
Changelog: Bugfix: Added the `pkg-config` paths for MacOS and Linux in `conftest.py`.
Closes: https://github.com/conan-io/conan/issues/11910

### DIFF
--- a/conan/tools/meson/meson.py
+++ b/conan/tools/meson/meson.py
@@ -48,7 +48,6 @@ class Meson(object):
         self._conanfile.run(cmd)
 
     def install(self):
-        self.configure(reconfigure=True)  # To re-do the destination package-folder
         meson_build_folder = self._conanfile.build_folder
         cmd = 'meson install -C "{}"'.format(meson_build_folder)
         self._conanfile.run(cmd)
--- a/conan/tools/meson/meson.py
+++ b/conan/tools/meson/meson.py
@@ -20,43 +20,46 @@ class Meson(object):
         has_deps_flags = os.path.exists(deps_flags)
 
         if os.path.exists(cross):
-            cmd += ' --cross-file "{}"'.format(cross)
+            cmd += f' --cross-file "{cross}"'
             if has_deps_flags:
-                cmd += ' --cross-file "{}"'.format(deps_flags)
+                cmd += f' --cross-file "{deps_flags}"'
         else:
-            cmd += ' --native-file "{}"'.format(native)
+            cmd += f' --native-file "{native}"'
             if has_deps_flags:
-                cmd += ' --native-file "{}"'.format(deps_flags)
+                cmd += f' --native-file "{deps_flags}"'
 
-        cmd += ' "{}" "{}"'.format(build_folder, source_folder)
+        cmd += f' "{build_folder}" "{source_folder}"'
         if self._conanfile.package_folder:
-            cmd += ' -Dprefix="{}"'.format(self._conanfile.package_folder)
+            cmd += f' -Dprefix="{self._conanfile.package_folder}"'
         if reconfigure:
-            cmd += ' --reconfigure'
-        self._conanfile.output.info("Meson configure cmd: {}".format(cmd))
+            cmd += f' --reconfigure'
+        self._conanfile.output.info(f"Meson configure cmd: {cmd}")
         self._conanfile.run(cmd)
 
     def build(self, target=None):
         meson_build_folder = self._conanfile.build_folder
-        cmd = 'meson compile -C "{}"'.format(meson_build_folder)
+        cmd = f'meson compile -C "{meson_build_folder}"'
         njobs = build_jobs(self._conanfile)
         if njobs:
-            cmd += " -j{}".format(njobs)
+            cmd += f" -j{njobs}"
         if target:
-            cmd += " {}".format(target)
-        self._conanfile.output.info("Meson build cmd: {}".format(cmd))
+            cmd += f" {target}"
+        self._conanfile.output.info(f"Meson build cmd: {cmd}")
         self._conanfile.run(cmd)
 
     def install(self):
+        # To re-do the destination package-folder
+        if self._conanfile.package_folder:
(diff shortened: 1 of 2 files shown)


memsharded:
``test_meson_lib_template`` test failed in Windows, seems the ``configure()`` call is necessary?

memsharded (changes requested):
Seems this was necessary

conan/tools/meson/meson.py
```diff
@@ -20,44 +20,46 @@ def configure(self, reconfigure=False):
...
-        self._conanfile.output.info("Meson build cmd: {}".format(cmd))
+            cmd += f" {target}"
+        self._conanfile.output.info(f"Meson build cmd: {cmd}")
         self._conanfile.run(cmd)
 
     def install(self):
-        self.configure(reconfigure=True)  # To re-do the destination package-folder
+        # To re-do the destination package-folder
+        if self._conanfile.package_folder:
+            self._conanfile.run(f"meson configure --prefix {self._conanfile.package_folder}")
```
eli-schwartz: `meson configure` is going to update the coredata, causing `meson install` to internally re-run `meson setup --reconfigure`

So I don't think there is any point in changing this, other than moving some work from one subprocess to another. It is NOT fixing the linked issue.

eli-schwartz (commented):
So the issue is that --prefix is modified after the build, during install? Why isn't this correct at first? :eyes:

conan/tools/meson/meson.py
```diff
@@ -20,44 +20,48 @@ def configure(self, reconfigure=False):
...
+        self._conanfile.output.info(f"Meson build cmd: {cmd}")
         self._conanfile.run(cmd)
 
     def install(self):
-        self.configure(reconfigure=True)  # To re-do the destination package-folder
+        # To fix the destination package-folder
+        if self._conanfile.package_folder:
+            cmd = f'meson configure --prefix "{self._conanfile.package_folder}"'
+            self._conanfile.run(cmd)
+            self._conanfile.output.info(f"Meson configure (prefix) cmd: {cmd}")
```
franramirez688, line 55: It's the only real change. Sorry for the long diff! I should have changed it in another PR 😅

conans/test/conftest.py
```diff
@@ -47,7 +48,9 @@
         "0.28": {
             "path": {
                 # Using chocolatey in Windows -> choco install pkgconfiglite --version 0.28
-                'Windows': "C:/ProgramData/chocolatey/lib/pkgconfiglite/tools/pkg-config-lite-0.28-1/bin"
+                'Windows': "C:/ProgramData/chocolatey/lib/pkgconfiglite/tools/pkg-config-lite-0.28-1/bin",
+                'Darwin': "/opt/homebrew/bin/pkg-config" if MacOS_M1 else "/usr/local/bin/pkg-config",
+                'Linux': "/usr/bin/pkg-config"
```
franramirez688, line 53: It was a bug in our test suite. All the tests specifying that tool were skipping MacOS and Linux.

conan/tools/meson/meson.py
franramirez688: As far as I can see in the Meson documentation (https://mesonbuild.com/Commands.html#examples), `meson configure` is used to change option values and not to reconfigure everything, isn't it?

Now, I'm running: 

```python
cmd = f'meson configure "{self._conanfile.build_folder}" -Dprefix="{self._conanfile.package_folder}"'
```
So, it should mean that I'm not doing more than changing the `prefix` value. If it does not work like that, is there a way to avoid that full reconfiguration and change only the `prefix` value?

eli-schwartz: Have you tried running it?

...

`meson configure` is used to change option values. It can be run multiple times in a row, does not perform reconfiguration. If the change isn't a no-op, ninja automatically runs `meson setup --reconfigure` to apply the changes.

`meson setup --reconfigure` is used:
- to apply previous updated options
- to force the reconfiguration, even when nothing changed, for example if you've installed new dependencies and suchlike.

> So, it should mean that I'm not doing more than changing the `prefix` value. If it does not work like that, is there a way to avoid that full reconfiguration and change only the `prefix` value?

No.

The full reconfiguration is what processes the changed `prefix` value, optionally enables/disables part of the build or passes `-DPACKAGE_PREFIX=/the/prefix/value` or `configure_file(input: 'constants.py.in', output: 'constants.py', configuration: confdata_with_directory_paths)`, or any of various other things that software might do.

The last step of the full reconfiguration is to write out install.dat, which contains the paths that files will be installed to.

For software that isn't relocatable, changing the `prefix` value might force ***all*** source files to be recompiled. Meson cannot know if that is the case until after it does a full reconfiguration and re-creates the ninja file, which may or may not result in changed compilation commands.

Meson does cache what it can (e.g. found dependencies, compiler checks) so the full reconfiguration is often a lot faster than the first time, though it isn't instant.

franramirez688:
Given the comment https://github.com/conan-io/conan/pull/11920#discussion_r951531831, it does not make sense to make that change.
I'll open another PR to fix the bug related to the Conan test suite.

Issue #11797, opened 2022-08-06
[question] conan v2: Shallow clone with conan.tools.scm.Git


Corristo:
With the conan v1 `scm` feature it was possible to perform a shallow clone, which can save quite a lot of time when cloning a large repository such as llvm or chromium. Is there a way to do the same with the new `Git` tool?

- [ x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

czoido:
Hi @Corristo,

Thanks a lot for the feedback, I agree this is a valid feature for the git tool. I'm marking this as a feature and assign to 1.52 milestone.

memsharded:
Implemented in https://github.com/conan-io/conan/pull/11921, will be in 1.52.

Corristo:
Wow that was quick. Thank you very much.

PR #11921, opened 2022-08-19
Expose additional args when calling Git.clone()

jcar87:
Changelog: Feature: Add ability to pass additional arguments to `conan.tools.scm.Git.clone()`.
Docs: https://github.com/conan-io/docs/pull/2721

As a way of addressing: https://github.com/conan-io/conan/issues/11797 

This PR exposes an `args` parameter when calling `Git.clone()`, so that we can forward additional parameters to the git command - thus enabling more advanced behaviours than cloning the default branch.

This can satisfy the issue in #11797 (by passing e.g. by passing `['--depth', '1']`). I believe the most common case when building recipes that clone from a git repository is to want to have it set to a specific commit (branch or tag). Some repositories take considerable time to clone, so the ability to tweak the `git clone` wall to make it as quick as possible seems like a good feature to have.

I'm unsure as to whether to expose a specific parameter for a shallow clone, what should it be called? `tools.Git()` had a `shallow=True/False`, but other git wrappers for Python have a specific `depth` argument, which may match more closely. And if we were to do that, why not others that also control likely toggles, like the submodule recursive (or not) clone behaviour. Exposing an args argument could be a good compromise to cover these cases.

### DIFF
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -76,7 +76,8 @@ class Git(object):
         folder = self._run("rev-parse --show-toplevel")
         return folder.replace("\\", "/")
 
-    def clone(self, url, target="", args=[]):
+    def clone(self, url, target="", args=None):
+        args = args or []
         if os.path.exists(url):
             url = url.replace("\\", "/")  # Windows local directory
         mkdir(self.folder)
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -76,12 +76,12 @@ class Git(object):
         folder = self._run("rev-parse --show-toplevel")
         return folder.replace("\\", "/")
 
-    def clone(self, url, target=""):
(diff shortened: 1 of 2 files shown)


conan/tools/scm/git.py
```diff
@@ -76,12 +76,12 @@ def get_repo_root(self):
         folder = self._run("rev-parse --show-toplevel")
         return folder.replace("\\", "/")
 
-    def clone(self, url, target=""):
+    def clone(self, url, target="", args=[]):
```
memsharded: Recall that a list is not good for default args in python! :)
```suggestion
    def clone(self, url, target="", args=None):
        args = args or []
```

memsharded (approved):
I agree a generic args is fine, no need to do specific python interface for shallow or others.
I think this can go to 1.52, very low risk, and useful to have.

czoido (approved):

PR #11928, opened 2022-08-22
Make conan_version public under conan as a Version object

### LINKED ISSUE #11760: [feature] Support conan __version__ under conan namespace
When users need to show, or even compare which Conan client is running, from the recipe, they can usually import `__version__`:

```python
from conans import __version__ as conan_version

print(f"Conan version: {conan_version}")
```

However, it's not possible to do same with the new namespace `conan`. It would be great having the same feature available, since some methods from Conan 2.0 are not compatible with Conan 1.0 and to keep a recipe working for both during the transition, it's important to detect which Conan version is running.

The real case comes from `self.requires()` which is incompatible and breaks according to the version:

https://docs.conan.io/en/1.51/reference/conanfile/methods.html#requirements
vs
https://docs.conan.io/en/2.0/reference/conanfile/methods.html#requires

The version 2.0 offers `transitive_headers`, which is not optional on 1.x

As workaround, needed to create:

```python
requires = lambda ref: self.requires(ref, transitive_headers=True) if Version(conan_version) >= "2.0.0-beta" else self.requires(ref)
```

czoido:
Changelog: Feature: Make the version of the Conan client available under `conan` and make it a `Version` object so it can be compared.
Docs: https://github.com/conan-io/docs/pull/2719

Closes: https://github.com/conan-io/conan/issues/11760

### DIFF
--- a/conan/__init__.py
+++ b/conan/__init__.py
@@ -1,1 +1,6 @@
 from conans.model.conan_file import ConanFile
+from conan.tools.scm import Version
+from conans import __version__
+
+
+conan_version = Version(__version__)
--- a/conans/test/integration/tools/conan_version_test.py
+++ b/conans/test/integration/tools/conan_version_test.py
@@ -1,0 +1,26 @@
+import textwrap
+
+from conans.test.utils.tools import TestClient
+from conans import __version__
+
+
+def test_conan_version():
+
+    conanfile = textwrap.dedent("""
+        from conan import ConanFile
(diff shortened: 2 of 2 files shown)


memsharded (approved):

uilianries (approved):
LGTM

franramirez688 (approved):

jcar87 (approved):

Issue #3168, opened 2018-07-04
Link versioned .so with CMake generator


db4:
I have a package that provides a versioned .so library (say, `libtest.so.1.0.0`). How to set `cpp_info.libs` so that CMake generator would be able to link it? Direct
```python
    self.cpp_info.libs = ['libtest.so.1.0.0']
```
doesn't work; if one passes`libtest.so.1.0.0` as a link library to cmake it strips `lib`, adds `-l` and then `ld` linker will fail:
```
/usr/bin/ld: cannot find -ltest.so.1.0.0
```
The only workaround I have found so far is to prepend a colon to a library file:
```python
    self.cpp_info.libs = [':libtest.so.1.0.0']
```
But is it OK for other generators? Maybe it should be done by a generator itself?

memsharded:
This is interesting. The typical approach when libraries generate such ``libtest.so.1.0.0`` version is that they also provide a ``libtest.so`` file which is a symlink to the 1.0.0 version. Doesn't such package have this symlink? You could try to pass the ``symlinks=True`` to ``self.copy()``?

For sure it won't work for other OS, so at least it needs to be conditioned to some ``if self.settings.os == "Linux"``, and I have no idea if the ``:libtest`` notation would affect other generators (they usually just pass that name as the ``-lxxxx`` flag in unix like systems. 

If I had to do it, I think I will aim to have a ``libtest`` file (either symlinking or renaming the final file), it looks to me a bit incorrect having to use the full version there. But please tell me about your use case. Thanks!

db4:
OK, I'm trying to fetch `cpp_info` from existing CMake project (I still hope to package OpenCV this way if you are interested). My previous approach with printing CMake values and parsing the output was too complex and did not work correctly under Unix-like OSes (library order was not preserved that is critical for `ld`, there were some problems with generator expressions etc.), so I implemented cmake server-based approach: https://github.com/db4/conan-cmake-config-tools/tree/cmake-server. It works and I almost happy with the results, but OpenCVConfig.cmake directly links versioned .so files. I could rewrite .so names by some heuristic rules, use symlinks etc but why? Looks like prepending `:` to a library file is enough for `ld` linker.

memsharded:
This ticket is about legacy and removed ``cmake`` generator, closing it as outdated.

PR #11929, opened 2022-08-22
[AutotoolsDeps] Allow passing filename to the linker if cpp_info spec…

### LINKED ISSUE #11703: [bug] LIBS variable not correct for linking static libraries when building with autoconf tools

jcar87:
Changelog: Feature: Allow passing filename to the linker on Unix systems for libraries in `cpp_info` where a file extension (`.a`, `.so`) is specified.
Docs: TODO

Closes https://github.com/conan-io/conan/issues/11703

On Unix-like systems where the `-l` linker flags assumes a name without the `lib` prefix or a file extension, the linker will follow its own rules to locate the library in the library search directories. On Linux (and according to the documentation, systems that support ELF shared libraries), shared libraries will be searched first. 

For libraries defined in `self.cpp_info.libs` we assume the name passed is without a prefix or extension, and pass that directly as an `-l` flag. 

The current behaviour should cover most cases, especially if libraries follow the practice of _not_ including both shared and static variants of the library (which is the case in Conan Center).

However, in the issue described in https://github.com/conan-io/conan/issues/11703, one may be specifying a "system library", for which the two variants may exist (`.a` and `.so`). In case we prefer the static one, we have a few options:
1) Passing the full path at link time (no flags)
2) Pass `-Bstatic` to the linker right before the `-l` flag, and revert it back to its default (`-Bdynamic`) afterwards. 
3) Use the `-l:` notation, documented [here](https://linux.die.net/man/1/ld):
```
           Add the archive or object file specified by namespec to the
           list of files to link.  This option may be used any number of
           times.  If namespec is of the form :filename, ld will search
           the library path for a file called filename, otherwise it
           will search the library path for a file called libnamespec.a.
```

I have verified that GNU linker (bfd), gold, and mold all accept this notation.

Notably, the default linker on macOS does *not* accept this notation - it always favours shared libraries and does not allow specifying the filename with the `:` notation, nor the `-Bstatic`/`-Bdynamic` flags.

### DIFF
--- a/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
+++ b/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
@@ -59,7 +59,7 @@ def test_link_cppinfo_libs_with_filename():
     client.run("create libhello.py -ohello:my_option=archive")
     client.run("create libhello.py -ohello:my_option=just_lib")
 
-    client.run("install conanfile.txt -g AutotoolsDeps -ohello:my_option=shared_object")
+    client.run("install conanfile.txt -g AutotoolsDeps -ohello*:my_option=shared_object")
     deps_file = client.load("conanautotoolsdeps.sh")
     assert "-l:libhello.so -l:libz.so" in deps_file
 
--- a/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
+++ b/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
@@ -55,7 +55,7 @@ def test_link_cppinfo_libs_with_filename():
 
     client = TestClient()
     client.save({"libhello.py": libhello, "conanfile.txt": conanfile_consumer})
-    client.run("create libhello.py -ohello:my_option=shared_object")
+    client.run("create libhello.py -ohello*:my_option=shared_object")
     client.run("create libhello.py -ohello:my_option=archive")
     client.run("create libhello.py -ohello:my_option=just_lib")
 
--- a/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
+++ b/conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
@@ -39,10 +39,13 @@ def test_link_cppinfo_libs_with_filename():
         def package_info(self):
             if self.options.my_option == "shared_object":
                 self.cpp_info.libs = ["libhello.so"]
+                self.cpp_info.system_libs = ["libz.so"]
             elif self.options.my_option == "archive":
                 self.cpp_info.libs = ["libhello.a"]
+                self.cpp_info.system_libs = ["libz.a"]
             else:
                 self.cpp_info.libs = ["hello"]
+                self.cpp_info.system_libs = ["z"]
     """)
 
     conanfile_consumer = textwrap.dedent("""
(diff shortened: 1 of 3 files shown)


conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
```diff
@@ -24,6 +24,53 @@ def test_link_lib_correct_order():
...
+    hello/0.1
+    """)
+
+    client = TestClient()
+    client.save({"libhello.py": libhello, "conanfile.txt": conanfile_consumer})
+    client.run("create libhello.py -ohello:my_option=shared_object")
+    client.run("create libhello.py -ohello:my_option=archive")
+    client.run("create libhello.py -ohello:my_option=just_lib")
+
+    client.run("install conanfile.txt -g AutotoolsDeps -ohello:my_option=shared_object")
```
memsharded: To make it 2.0 ready:
```suggestion
    client.run("install conanfile.txt -g AutotoolsDeps -ohello*:my_option=shared_object")
```

conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
```diff
@@ -24,6 +24,53 @@ def test_link_lib_correct_order():
...
+    """)
+
+    conanfile_consumer = textwrap.dedent("""
+    [requires]
+    hello/0.1
+    """)
+
+    client = TestClient()
+    client.save({"libhello.py": libhello, "conanfile.txt": conanfile_consumer})
+    client.run("create libhello.py -ohello:my_option=shared_object")
```
memsharded: To make it 2.0 ready
```suggestion
    client.run("create libhello.py -ohello*:my_option=shared_object")
```

conans/test/integration/toolchains/gnu/test_autotoolsdeps.py
```diff
@@ -24,6 +24,53 @@ def test_link_lib_correct_order():
...
+    client.run("create libhello.py -ohello:my_option=archive")
+    client.run("create libhello.py -ohello:my_option=just_lib")
+
+    client.run("install conanfile.txt -g AutotoolsDeps -ohello:my_option=shared_object")
+    deps_file = client.load("conanautotoolsdeps.sh")
+    assert "-l:libhello.so -l:libz.so" in deps_file
```
memsharded, line 64: Isn't it worth to have a functional test that will actually compile (and that would fail to compile if this fix was not merged) to guarantee this fully works?

memsharded:
Seems https://github.com/conan-io/conan/issues/3168 is related?

jcar87:
Closing to revisit in time for 1.55.

PR #11931, opened 2022-08-22
Added pkg-config paths for Linux and Macos

franramirez688:
Changelog: omit
Docs: omit

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -34,7 +34,9 @@ tools_locations = {
 }
 """
 
-MacOS_M1 = all([platform.system() == "Darwin", platform.machine() == "arm64"])
+MacOS_arm = all([platform.system() == "Darwin", platform.machine() == "arm64"])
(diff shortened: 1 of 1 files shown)


memsharded (approved):
This is concerning. There are other tools, Ninja, Bazel (not defined in Linux), that would have a similar problem.
I think it was not the original intention to skip the test if the path was not defined, just to use the defined path in the cases it was defined.

We definitely need to review this.

conans/test/conftest.py
```diff
@@ -34,6 +34,7 @@
 }
 """
 
+MacOS_M1 = all([platform.system() == "Darwin", platform.machine() == "arm64"])
```
jcar87: mmm if one of my two branches for fixing the tests on Apple Silicon (I had a few variations...) I had something along the lines of:

```
homebrew_root = "/opt/homebrew" if MacOS_arm else "/usr/local"
```

then you can reference this on the macos paths and avoid more conditionals.

jcar87 (approved):

PR #11935, opened 2022-08-23
add BUILD_TESTING=OFF to presets if tools.build:skip_test

### LINKED ISSUE #11905: [question] Why does test_requires is installed when not testing?
The enigmatic feature `test_requires` is mentioned in Docs in some points:

https://docs.conan.io/en/latest/reference/commands/creator/create.html?highlight=test_requires
https://docs.conan.io/en/latest/reference/conanfile/dependencies.html?highlight=test_requires#iterating-dependencies
https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html?highlight=test_requires#requirements
https://docs.conan.io/en/latest/devtools/build_requires.html?highlight=test_requires#build-and-host-contexts

There is no specif section about that feature, but is clear that's a shortcut to `self.build_requires(<ref>, force_host_context=True)`.

However, as it should be combined when testing a package with unit tests or similar, it would be nice installing test requirements only when indeed running tests. For instance:

```python
def build_requirements(self):
    self.test_requires("gtest/1.12.1")

def build(self):
    cmake = CMake(self)
    cmake.configure()
    cmake.build()
    cmake.test()
```

`conan create . foobar/0.1.0 -pr:h=default -pr:b=default -c tools.build:skip_test=True`

This command tells to Conan to skip the test step, which means, we don't need to install GTest neither run our unit tests. However, Conan always install gtest as it's a regular build requirement.

So, you can tell, but, it built all tests, only didn't execute them ... But again, we can avoid building them by:

```python
def generate(self):
    tc = CMakeToolchain(self)
    # Maybe Conan manages it automatically but I didn't check
    tc.variables["BUILD_TESTING"] = self.conf_info.get("tools.build:skip_test", check_type=bool)
    tc.generate()
```

Of course, users can apply the same rule to avoid installing gtest, but it seems be a dubious as we only want to install a requirement for unit tests.  

memsharded:
Changelog: Feature: Add `BUILD_TESTING=OFF` to CMakeToolchain presets if `tools.build:skip_test`.
Docs: https://github.com/conan-io/docs/pull/2713

Close https://github.com/conan-io/conan/issues/11905

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -133,13 +133,18 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables):
         if "CMAKE_SH" not in cache_variables:
             cache_variables["CMAKE_SH"] = "CMAKE_SH-NOTFOUND"
 
-        cmake_make_program = conanfile.conf.get("tools.gnu:make_program", default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
+        cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
+                                                default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
         if cmake_make_program:
             cmake_make_program = cmake_make_program.replace("\\", "/")
             cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
 
     if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
         cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
+
+    if "BUILD_TESTING" not in cache_variables:
+        if conanfile.conf.get("tools.build:skip_test", check_type=bool):
+            cache_variables["BUILD_TESTING"] = "OFF"
 
     preset_path = os.path.join(conanfile.generators_folder, "CMakePresets.json")
     multiconfig = is_multi_configuration(generator)
--- a/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
+++ b/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
@@ -466,7 +466,7 @@ def test_toolchain_cache_variables():
     client.save({"conanfile.py": conanfile})
     with mock.patch("platform.system", mock.MagicMock(return_value="Windows")):
         client.run("install . mylib/1.0@ -c tools.cmake.cmaketoolchain:generator='MinGW Makefiles' "
-                   "-c tools.gnu:make_program='MyMake'")
+                   "-c tools.gnu:make_program='MyMake' -c tools.build:skip_test=True")
     presets = json.loads(client.load("CMakePresets.json"))
     cache_variables = presets["configurePresets"][0]["cacheVariables"]
     assert cache_variables["foo"] == 'ON'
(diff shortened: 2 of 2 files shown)


jcar87:
This looks good but I have a question!
`tools.build:skip_test` is documented as:

> Do not execute CMake.test() and Meson.test() when enabled

In this PR, we would be a going a step beyond that - and potentially not building tests at all (depending on whether or not the projects use `BUILD_TESTING` to control the building of the tests. 

This can be convenient as on some projects this can save considerable compilation time, but on the other hand in some projects the tests are the first attempt at linking things that consume the library and it can serve as an early way of catching linking issues. Although it could be argued that is the purpose of the `test_package` anyway.

memsharded:
Yes, I would say that the expected behavior of the Conan integrations, regarding ``skip_test`` is to avoid building them too, if they are not going to be ran anyway. I think the docs can be updated to reflect this. As always, this is only the default, and projects listening to BUILD_TESTING should behave correctly for both True and False values, and can define themselves in the recipe a different logic to ``tc.cache_variables["BUILD_TESTING"] = xxxx`` if they want to, and that will be respected.

franramirez688 (approved):
I think that this is the expected behavior, and the user will save time whenever he does not want to run all the tests.

conan/tools/cmake/presets.py
```diff
@@ -133,14 +133,19 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables):
...
+        cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
+                                                default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
         if cmake_make_program:
             cmake_make_program = cmake_make_program.replace("\\", "/")
             cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
 
     if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
         cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
 
+    if "BUILD_TESTING" not in cache_variables:
```
uilianries, line 145: What about `CMakeToolchain.variables`? Is it a possible scenario, right?

memsharded: It sounds like ``CMakeToolchain.variables`` should be for user variables, while everything that is a CMake built-in, should be to ``cache_variables`` otherwise there is a risk that the ``variables`` will not have enough priority. @jcar87 wdyt?

jcar87: I'd say that's a good approximation!

I can find widespread evidence on GitHub that `BUILD_TESTING` is treated as a cache variable by projects:
1) For example [here](https://github.com/dduan/Pathos/blob/5efb5fdfd6d89cd394c0f9e3094336628d8d9903/CMakeLists.txt#L8-L9) by calling `option()` 

2) It will also be a cache variable if  the user provides it via `-DBUILD_TESTING=ON/OFF` - in this case, the user's preference will take precedence and be respected regardless of whether we define it as a `cache_variable` or the project sets an `option()` - 

So I think the way it currently is in this PR is the safest and least likely to have side effects - especially if there's any chance projects define or reference `BUILD_TESTING` before the first call to `project()` (have found  examples [here](https://github.com/johannes51/AerospaceFramework/blob/1ab876ce80ea67a305ba0fe03eb9ab172d1ff24c/dependencies/CMakeLists.txt#L3) and [here](https://github.com/EagleDad/cmake-playground/blob/c80eebec3c667e53e47f96eb89da30dc748a4980/CMakeLists.txt#L22))

memsharded: Ok, lets recommend that in the docs then 👍

uilianries: I'm not in the opposite of using `cached_variables`, but still, users are free to use both and it will have a different behavior, depending the usage.

memsharded: Yes, but if we explicitly say it in the docs, they should use each one for the documented cases, so they will not be "free" to use them invariantly for all cases.

PR #11941, opened 2022-08-23
Try not to skip tests on CI if the tool path is not defined for a Version

### LINKED ISSUE #11933: Review Ninja, Bazel tests running in CI or not
This is concerning. There are other tools, Ninja, Bazel (not defined in Linux), that would have a similar problem.
I think it was not the original intention to skip the test if the path was not defined, just to use the defined path in the cases it was defined.

We definitely need to review this.

_Originally posted by @memsharded in https://github.com/conan-io/conan/pull/11931#pullrequestreview-1081236885_

czoido:
Changelog: omit
Docs: omit

Closes: https://github.com/conan-io/conan/issues/11933

Just to check, let's see if this fixes the related issue.

#TAGS: svn, slow

### DIFF
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -54,7 +54,6 @@ tools_locations = {
             "path": {
                 # Using chocolatey in Windows -> choco install pkgconfiglite --version 0.28
                 'Windows': "C:/ProgramData/chocolatey/lib/pkgconfiglite/tools/pkg-config-lite-0.28-1/bin",
-                'Linux': "/usr/bin/",
             }
         }},
     'autotools': {"exe": "autoconf"},
--- a/conans/test/conftest.py
+++ b/conans/test/conftest.py
@@ -23,6 +23,12 @@ tools_locations = {
         "3.16": {"disabled": True},
         "3.17": {"disabled": True},
         "3.19": {"path": {"Windows": "C:/ws/cmake/cmake-3.19.7-win64-x64/bin"}},
+        # To explicitly skip one tool for one version, define the path as None
+        # if you don't define the path for one platform it will run the test with the
+        # tool in the path. For example here it will skip the test with CMake in Darwin but
+        # in Linux it will run with the version found in the path if it's not specified
+        "3.23": {"path": {"Windows": "C:/ws/cmake/cmake-3.19.7-win64-x64/bin",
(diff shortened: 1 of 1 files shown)


conans/test/conftest.py
```diff
@@ -23,6 +23,12 @@
         "3.16": {"disabled": True},
         "3.17": {"disabled": True},
         "3.19": {"path": {"Windows": "C:/ws/cmake/cmake-3.19.7-win64-x64/bin"}},
+        # To explicitly skip one tool for one version, define the path as None
```
memsharded: I think I would reverse the logic, and make it explicit: ``"Darwin": "skip-tests"``, because the semantics are not well defined, what happens with Linux in this case, it is None too, so it will be skipped? Doing this, the ``undefined`` in the code below is not necessary, ``None`` becomes undefined (pythonic), and ``skip-tests`` means skip those tests. Wdyt?

czoido: Yes, it's cleaner to reverse the logic and use the `skip-tests`, I saw that we had this previously and thought that maybe we were using this to explicitly skip: https://github.com/conan-io/conan/blob/260aedbaa49dea43711b38bbdc0fdbf6d2c65dad/conans/test/conftest.py#L85-L86 so I maintained that in case it was used in the confuser, but I will change that for a `skip-tests` there and do the changes 👍

memsharded (approved):

conans/test/conftest.py
```diff
@@ -50,45 +57,45 @@
...
+                'Darwin': f"{homebrew_root}/bin",
+                'Linux': "/usr/bin"
             }
         }},
     'autotools': {"exe": "autoconf"},
     'cmake': {
         "default": "3.15",
         "3.15": {
-            "path": {'Windows': 'C:/cmake/cmake-3.15.7-win64-x64/bin',
+            "path": {'Windows': 'C:\\cmake\\cmake-3.15.7-win64-x64\\bin',
```
memsharded: Isn't it a bad UX that we Windows users will have to define our ``conftest_user.py`` with double backslash?

SpaceIm:
Since this commit I can't run tests anymore locally on my mac, it fails with `Failed with Error: [undefined]failed on setup with "Failed: Required 'cmake' tool version 'Any' is not available"`

PR #11968, opened 2022-08-26
add targets to MSBuild

### LINKED ISSUE #11966: [feature] MSBuild add targets parameter again
The new [MSBuild](https://docs.conan.io/en/latest/reference/conanfile/tools/microsoft.html#msbuild) didn't port `targets` attribute from old [MSBuild](https://docs.conan.io/en/latest/reference/build_helpers/visual_studio.html#build).

Motivation: I started to port [xz_utils](https://github.com/conan-io/conan-center-index/pull/12308) recipe and it comes with .sln files, and when need to build static or shared, each configuration follows an unique target.

I think it should be a `build()` parameter, as other variables, like platform and build type will not change:

    msbuild = MSBuild(self)
    msbuild.build(targets=["static"])

memsharded:
Changelog:  Feature: Add ``MSBuild().build(.., targets=["target1"])`` argument to new ``MSBuild``.
Docs: https://github.com/conan-io/docs/pull/2724

Close https://github.com/conan-io/conan/issues/11966

### DIFF
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -28,7 +28,7 @@ class MSBuild(object):
             msvc_arch = conanfile.settings.get_safe("os.platform")
         self.platform = msvc_arch
 
-    def command(self, sln):
+    def command(self, sln, targets=None):
         cmd = ('msbuild "%s" /p:Configuration=%s /p:Platform=%s'
                % (sln, self.build_type, self.platform))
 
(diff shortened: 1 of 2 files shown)


conan/tools/microsoft/msbuild.py
```diff
@@ -41,10 +41,15 @@ def command(self, sln):
         if maxcpucount:
             cmd += " /m:{}".format(maxcpucount)
 
+        if targets:
+            if not isinstance(targets, list):
```
SSE4, line 45: why? I would allow any `Iterable`, so generator/co-routine could be used here

memsharded: I prefer interfaces to be as defined, limited, and constrained as possible, and open them up when real use cases happen. If you have a generator just call the method ``msbuild.build(...targets=list(yourstuff))``

Example: If we allow an ``Iterable``, it happens that a string is ``Iterable``, so users passing a string ``target="mytarget"``, will get a ``/target=m;y;t;a;r;g;e;t`` that will produce a more difficult to interpret later at build output. You can say, now you have to handle the ``string`` case differently in the method, and that starts to add code to the method, that needs to be tested, documented, and maintained, and this is exactly what I want to avoid

franramirez688 (approved):

PR #11971, opened 2022-08-27
Added endeavouros to the distro list in `with_pacman` in oss.py  #11970

### LINKED ISSUE #11970: [bug] EndeavourOS not among list of distros listed for pacman support
  * Operating System+version: EndeavourOS 2022.06.23
  * Compiler+version: Clang 14.0.6
  * Conan version: 1.51.3
  * Python version: 3.10.6

In a python interpreter
```bash
>>> from conans import tools
>>> tools.os_info.with_pacman
False
>>>tools.os_info.is_linux
True
>>> tools.os_info.linux_distro
'endeavouros'
```

We need to add `endeavouros` to this [line](https://github.com/conan-io/conan/blob/develop/conans/client/tools/oss.py#L220)

As a result of the above, OpenGL recipe doesn't know how to install itself, xorg/system also fails to install some dependencies as well

```log
opengl/system: Package installed 5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
opengl/system: Downloaded package revision 0
xorg/system: Retrieving package 5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9 from remote 'conancenter' 
Downloading conanmanifest.txt completed [0.06k]                                          
Downloading conaninfo.txt completed [0.16k]                                              
Downloading conan_package.tgz completed [0.06k]                                          
Decompressing conan_package.tgz completed [0.00k]                                        
xorg/system: Package installed 5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
xorg/system: Downloaded package revision 0
opengl/system: WARN: Don't know how to install OpenGL for your distro.
ERROR: xorg/system: Error in package_info() method, line 79
        pkg_config.fill_cpp_info(
        ConanException: pkg-config command ['pkg-config', '--print-provides', 'xvmc', '--print-errors'] failed with error: Command 'pkg-config --print-provides xvmc --print-errors' returned non-zero exit status 1.
Package xvmc was not found in the pkg-config search path.
Perhaps you should add the directory containing `xvmc.pc'
to the PKG_CONFIG_PATH environment variable
Package 'xvmc', required by 'virtual:world', not found
```

cchulo:
Changelog: Bugfix: Add "endeavouros" to the list of distros with pacman.
Docs: Omit

Close #11970

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -38,7 +38,7 @@ class _SystemPackageManagerTool(object):
                                    "almalinux", "rocky"],
                            "dnf": ["fedora", "rhel", "centos", "mageia"],
                            "brew": ["Darwin"],
-                           "pacman": ["arch", "manjaro", "msys2"],
+                           "pacman": ["arch", "manjaro", "msys2", "endeavouros"],
                            "choco": ["Windows"],
                            "zypper": ["opensuse", "sles"],
                            "pkg": ["freebsd"],
--- a/conans/client/tools/oss.py
+++ b/conans/client/tools/oss.py
@@ -217,7 +217,7 @@ class OSInfo(object):
     @property
     def with_pacman(self):
         if self.is_linux:
-            return self.linux_distro in ["arch", "manjaro"]
+            return self.linux_distro in ["arch", "manjaro", "endeavouros"]
         elif self.is_windows and which('uname.exe'):
             uname = check_output_runner(['uname.exe', '-s'])
             return uname.startswith('MSYS_NT') and which('pacman.exe')
--- a/conans/client/tools/oss.py
+++ b/conans/client/tools/oss.py
@@ -217,7 +217,7 @@ class OSInfo(object):
     @property
     def with_pacman(self):
         if self.is_linux:
-            return self.linux_distro in ["arch", "manjaro", "endeavouros"]
+            return self.linux_distro in ["arch", "manjaro"]
         elif self.is_windows and which('uname.exe'):
             uname = check_output_runner(['uname.exe', '-s'])
             return uname.startswith('MSYS_NT') and which('pacman.exe')


memsharded (changes requested):
Thanks for contributing

The integration that you are modifying is legacy, and shouldn't be used anymore. 
The recommended one is in ``from conan.tools.system ...``

Please have a look to https://docs.conan.io/en/latest/reference/conanfile/tools.html, and the place to do the modifications is in https://github.com/conan-io/conan/blob/develop/conan/tools/system/package_manager.py

cchulo:
@memsharded done

memsharded (approved):

memsharded (changes requested):
Have you tried this manually and it is working?
I thought it was ok, but I did a routine check on https://github.com/python-distro/distro/blob/master/src/distro/distro.py, and I don't think that ``distro.id()`` will be returning the ``endeavouros`` string. Am I missing something? Is ``distro`` capable of returning this even if not documented explicitly in the project?

cchulo:
well this is why I had the previous change, I traced the code back to oss.py, though you said it was legacy, that's what led me to that file in the first place

cchulo:
Actually I should be more specific, the initial commit to oss.py was traced back from OpenGL's recipe, tools which is imported from `conans` and then just reading that code led me to `oss.py`, I decided not to change that file at your advise that it is legacy and instead implemented the change at the other location you suggested, not exactly sure how to test this change to ensure that it is working. Still with your earlier advise to setup `[conf]` flags in my profile I was able to work around this anyway.

Give me a couple of more days to test this. If you have any ideas as to how to proceed please let me know.

memsharded:
No, the code is ok, the change you did put the fix in the right place. I think the problem would be the same in the other legacy place as well.

The underlying issue is that Conan uses the return of ``distro.id()`` to feed it into those dictionaries. But it seems that ``distro.id()`` doesn't return ``endeavouros`` string, at least my inspection of the ``distro`` code. The way to check it would be just ``import distro + distro.id()`` in your ``python`` in your system.

cchulo:
> No, the code is ok, the change you did put the fix in the right place. I think the problem would be the same in the other legacy place as well.
> 
> The underlying issue is that Conan uses the return of `distro.id()` to feed it into those dictionaries. But it seems that `distro.id()` doesn't return `endeavouros` string, at least my inspection of the `distro` code. The way to check it would be just `import distro + distro.id()` in your `python` in your system.

I had my hard drive fail, its going to take me a bit to get back to you, I will ping you as soon as I am able to, would be back up and running, but I have work, I may not be able to get back to this until this upcoming weekend

cchulo:
@memsharded sorry for that week-long delay, got around to it, and yes, distro.id() on my OS  the method returns correctly:

```shell
>>> import distro
>>> distro.id()
'endeavouros'
>>> 
```

Is it safe to merge this in now?

memsharded (approved):

PR #11979, opened 2022-08-30
[PkgConfigDeps] Tool requires

### LINKED ISSUE #11710: [question] How to generate PkgConfig files for tool_requires dependencies and consume them with MesonToolchain?
Several Meson-based packages in Conan Center use PkgConfig files to find build tools.
For instance, the [wayland](https://github.com/conan-io/conan-center-index/blob/master/recipes/wayland/all/conanfile.py) package contains a tool that generates source files called `wayland-scanner`.
To handle the complexities of cross-compilation, the `wayland` package requires that `wayland-scanner` be installed on the build machine.
Unlike typical build tools, however, it uses `wayland-scanner`'s PkgConfig file to locate the executable instead of searching for it in `PATH`.
Suffice to say this requires that PkgConfig files for `tool_requires` dependencies be available.

The workaround I implemented in the `wayland` package is not correct because it relies on a private method in Conan to generate the PkgConfig files for the `wayland` package.
Yes, the `wayland` package requires itself as a `tool_requires`, but only when it is being cross-compiled.

```python
# Generate PC files for the tool_requires wayland package to ensure wayland-scanner is found for build machine.
if cross_building(self):
    native_generators_folder = os.path.join(self.generators_folder, "native")
    mkdir(self, native_generators_folder)
    for pc_name, pc_content in get_pc_files_and_content(self, self.dependencies.build["wayland"]).items():
        save(self, os.path.join(native_generators_folder, pc_name), pc_content)
    tc.project_options["build.pkg_config_path"] = native_generators_folder
```

Another instance of this kind of usage is the `xkbcommon` package, which also requires the `wayland` package's PkgConfig files be available in order to find build tools during Meson's generation phase.
See conan-io/conan-center-index#11636 for details.

Note that `MesonToolchain` also requires a special option in order for it to find the PkgConfig files on the _build_ machine: `build.pkg_config_path`.
Relying on the normal `pkg_config_path` variable or corresponding environment variable will __not__ work in this situation.
Ideally, `MesonToolchain`  would also integrate with this use case by automatically defining `build.pkg_config_path` to include the necessary directory for all `tool_requires` dependencies.

franramirez688:
Changelog: Feature: Added mechanism to create .pc files for build requirements.
Docs: https://github.com/conan-io/docs/pull/2725
Closes: https://github.com/conan-io/conan/issues/11710

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -184,11 +184,17 @@ class PCContentGenerator:
 
 class PCGenerator:
 
-    def __init__(self, conanfile, dep, dep_suffix=""):
+    def __init__(self, conanfile, dep, build_context_suffix=None):
         self._conanfile = conanfile
         self._dep = dep
-        self._dep_suffix = dep_suffix
+        self._build_context_suffix = build_context_suffix or {}
+        self._dep_suffix = self._build_require_suffix(dep)
         self._content_generator = PCContentGenerator(self._conanfile, self._dep)
+
+    def _build_require_suffix(self, req):
+        if not req.is_build_context:
+            return ""
+        return self._build_context_suffix.get(req.ref.name, "")
 
     def _get_cpp_info_requires_names(self, cpp_info):
         """
@@ -226,7 +232,7 @@ class PCGenerator:
                 req_conanfile = self._dep
             comp_name = _get_component_name(req_conanfile, comp_ref_name)
             if not comp_name:
-                pkg_name = _get_package_name(req_conanfile, suffix=self._dep_suffix)
+                pkg_name = _get_package_name(req_conanfile, suffix=self._build_require_suffix(req_conanfile))
                 # Creating a component name with namespace, e.g., dep-comp1
                 comp_name = _get_name_with_namespace(pkg_name, comp_ref_name)
             ret.append(comp_name)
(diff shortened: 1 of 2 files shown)


conans/test/integration/toolchains/gnu/test_pkgconfigdeps.py
```diff
@@ -573,3 +579,152 @@ def package_info(self):
...
+                self.build_requires("tool/1.0")
+
+            def generate(self):
+                tc = PkgConfigDeps(self)
+                tc.build_context_activated = ["tool"]
+                tc.generate()
+        """)
+    client.save({"conanfile.py": conanfile}, clean_first=True)
+    with pytest.raises(Exception) as e:
+        client.run("install . -pr:h default -pr:b default")
```
memsharded: use the ``assert_error``
```suggestion
    client.run("install . -pr:h default -pr:b default", assert_error=True)
```

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -184,11 +184,18 @@ def shortened_content(self, info):
...
 
-    def __init__(self, conanfile, dep):
+    def __init__(self, conanfile, dep, build_context_suffix=None):
         self._conanfile = conanfile
+        self._build_context_suffix = build_context_suffix or {}
         self._dep = dep
+        self._dep_suffix = self._build_require_suffix(dep)
         self._content_generator = PCContentGenerator(self._conanfile, self._dep)
 
+    def _build_require_suffix(self, req):
```
memsharded: This looks a bit messy, difficult to understand, with ``build_context_suffix``, ``build_require_suffix`` and ``dep_suffix``. But if it works, probably can be left now as-is, and improved later

memsharded (approved):

memsharded (approved):

jcar87 (approved):

PR #11359, opened 2022-05-30
Modern winbash improvements

memsharded:
Changelog: (Feature | Fix | Bugfix): Describe here your pull request
Docs: https://github.com/conan-io/docs/pull/XXXX

AFTER  https://github.com/conan-io/conan/pull/11328

PR #11981, opened 2022-08-30
[bugfix] Environment shell script path is now formatted based on the subsystem

### LINKED ISSUE #11980: [bug] paths for environment files are not formatted for subsystem when run in win_bash on Windows
The `wrapped_user_cmd` in the `subsystems.py` doesn't take the subsystem into account, for the environment `.sh` files after the bash statement

  * Compiler+version: N/A
  * Conan version: 1.50.1 (and dev)
  * Python version: 3.10.4

Run the `create` command on a recipe which uses msys2 on Windows.
I'm currently working on my attempt to modernize the `autoconf` recipe https://github.com/Ultimaker/conan-center-index/blob/modernize_autoconf/recipes/autoconf/all/conanfile.py

```powershell
----Running------
> "C:\.conan\376fd79\1\build-release\conan\conanbuild.bat" && "C:\.conan\376fd79\1\build-release\conan\msys2_mode.bat" && C:\.conan\7c3916\1\bin\msys64\usr\bin\bash.exe -c ^". \^"C:\.conan\376fd79\1\build-release\conan\conanbuild.sh\^" ^&^& \^"/c/.conan/376fd79/1/source/configure\^" --prefix=/ --bindir=${prefix}/bin --sbindir=${prefix}/bin --libdir=${prefix}/lib --includedir=${prefix}/include --oldincludedir=${prefix}/include ^"
```

notice the formatting of the `conanbuild.sh` path: `C:\.conan\376fd79\1\build-release\conan\conanbuild.sh`

jellespijker:
The shell script for the environments should be formatted based on the subsystem now.

Fixes #11980

Changelog: (Bugfix): Environment shell script path is now formatted based on the subsystem

### DIFF
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -72,6 +72,7 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
     # Wrapping the inside_command enable to prioritize our environment, otherwise /usr/bin go
     # first and there could be commands that we want to skip
     wrapped_user_cmd = environment_wrap_command(env, envfiles_folder, command,
+                                                subsystem=subsystem,
                                                 accepted_extensions=("sh", ))
     wrapped_user_cmd = _escape_windows_cmd(wrapped_user_cmd)
 


memsharded (commented):
This would need a test that would be failing without this fix. I am concerned that this apparently innocent fix could be breaking other users, unfortunately the subsystems are not sufficiently tested.

We need to have a look to the whole subsystems thing, as it is still not a good fit. We have some ongoing experimentation efforts in https://github.com/conan-io/conan/pull/11359 and https://github.com/conan-io/conan/pull/11066, but nothing that looks good enough so far...

jellespijker:
Yes I agree with the risk there, I will investigate this in more detail. Also try out how it behaves in more user cases.

The new win-bash approach looks promising btw

jellespijker:
@memsharded the failing test occurred in #11987 which turned out to be a non issue. I do think this PR is save to merge and actually fixes the bug in described in #11980

jellespijker:
@memsharded / @lasote  has this fix been considered to make it to the next release. The error you were concerned about happened in another PR (which is now closed since it turned out to be a non-issue) This fix is relatively safe and contained imo.

And I think I will run into issues with some open PR's:
- https://github.com/conan-io/conan-center-index/pull/12892
- https://github.com/conan-io/conan-center-index/pull/12898#issuecomment-1246196467

jellespijker:
As my colleague @Joeydelarago pointed out, the hint is in the comment:
> This would need a test that would be failing without this fix.

I will write a testcase for this PR

memsharded:
Hi @jellespijker 

It has been some time, maybe this was no longer a priority? Failed to write the testcase? 
It is true that it is weird that ``subsystem`` is not passed there, but on the other hand, there have been no reports about this, so it seems subsystems are fine with the native path? It would be good to know a bit better, have failing tests for this.

jellespijker:
I'm fine with closing this PR. Our initial work, which required this fix, didn't end up in our production env. So this isn't PR isn't need by us anymore.

Issue #11892, opened 2022-08-17
Re-introduce the CLICOLOR_FORCE environment variable (or equivalent) in 2.0


santhonisz:
Hi there,

I just noticed that in Conan 2.0 the `CLICOLOR_FORCE` environment variable was removed as part of https://github.com/conan-io/conan/pull/11421. While I completely understand and respect the decision to adopt the standardised `NO_COLOR` approach to disabling colored output, the problem is that our (Windows) CI environments such as TeamCity won't default to using color as they have no TTY attached when running builds e.g. the stdout/stderr streams are captured for reporting in the CI system. In Conan 1.x we could elect to force the colored output, but now we will seemingly no longer have that ability?

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

memsharded:
Closed by https://github.com/conan-io/conan/pull/12028, will be in beta.3

PR #12028, opened 2022-09-05
CLICOLOR_FORCE support

lasote:
Changelog: Feature: Added support for `CLICOLOR_FORCE` env var, that will activate the colors in the output if the value is declared and different to `0`.

Close https://github.com/conan-io/conan/issues/11892

### DIFF
--- a/conans/client/userio.py
+++ b/conans/client/userio.py
@@ -18,7 +18,15 @@ def color_enabled(stream):
     Command-line software which adds ANSI color to its output by default should check for the
     presence of a NO_COLOR environment variable that, when present (**regardless of its value**),
     prevents the addition of ANSI color.
+
+    CLICOLOR_FORCE: Force color
+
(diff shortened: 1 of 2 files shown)


conans/client/userio.py
```diff
@@ -18,7 +18,15 @@ def color_enabled(stream):
     Command-line software which adds ANSI color to its output by default should check for the
     presence of a NO_COLOR environment variable that, when present (**regardless of its value**),
     prevents the addition of ANSI color.
+
+    CLICOLOR_FORCE: Force color
+
+    https://bixense.com/clicolors/
     """
+    if os.getenv("CLICOLOR_FORCE", "0") != "0":
```
memsharded: I think ``conan_output_logger_format`` takes more priority than this, should be before this.

memsharded (approved):

Issue #11786, opened 2022-08-05
[question] check_min_cppstd context


uilianries:
The documentation is clear about header-only packages [here](https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html#the-package-id-method), but it's not helpful when a recipe may be header-only or not (e.g. fmt, spdlog). So, based on the real case https://github.com/conan-io/conan-center-index/pull/12021, I have the follow statement:

```python
def package_id(self):
    if self.info.options.header_only:
         self.info.clear()

def validate(self):
    if self.info.settings.compiler.cppstd:
         check_min_cppstd(self, 11)
    if self.info.settings.os != "Windows" and (self.info.options.wchar_support or self.info.options.wchar_filenames):
        raise ConanInvalidConfiguration("wchar is only supported under windows")
    if not self.info.options.header_only and self.info.options.shared and is_msvc_static_runtime(self):
        raise ConanInvalidConfiguration("Visual Studio build for shared library with MT runtime is not supported")
```

When the option `header_only=True` is defined, the package ID is cleared, then `self.info.settings` and `self.info.options` are sanitized, resulting on the follow error:

```
ERROR: Traceback (most recent call last):
  File ".../conans/errors.py", line 34, in conanfile_exception_formatter
    yield
  File ".../conans/client/graph/graph_binaries.py", line 411, in _compute_package_id
    conanfile.validate()
  File "${HOME}/.conan/data/spdlog/1.10.0/_/_/export/conanfile.py", line 73, in validate
    if self.info.settings.compiler.cppstd:
AttributeError: 'NoneType' object has no attribute 'cppstd'
```

This error **only** occurs with Conan 1.x (I'm running 1.51.0), but with Conan 2.0-beta2 it works fine.

I see few options:

- Update Conan 1.x to have same behavior as 2.0 for `self.info`
- Add `get_safe` to `self.info.settings` and `self.info.options`
- Need to mix `self.options.get_safe('header_only')` with `self.info.`

I personally prefer the first option, so we don't need to update the recipe again.

#### Environment

Conan Version: 1.51.0
OS: Linux Ubuntu 16.04
Architecture: x86_64
Docker image: conanio/gcc10-ubuntu16.04:1.51.0

#### Steps to reproduce

```
pip install conan==1.51.0
git clone https://github.com/uilianries/conan-center-index.git
cd conan-center-index/recipes/spdlog/all
git checkout 72df1ad8336c00ed427f864523b7037a08e7ff13
[...]

PR #12031, opened 2022-09-05
Test check_min_cppstd and validate_build

czoido:
Add test for two cases:
- header only packages that don't have self.info but can check the min cppstd in the validate() method without failing
- packages that need a certain cppstd to build but can be consumed without setting the cppstd because they could have a pure C interface for example 

Related to: https://github.com/conan-io/conan/issues/11786

### DIFF
--- a/conans/test/integration/tools/cppstd_minimum_version_test.py
+++ b/conans/test/integration/tools/cppstd_minimum_version_test.py
@@ -49,3 +49,28 @@ class CppStdMinimumVersionTests(unittest.TestCase):
         self.client.run("create . --user=user --channel=channel -pr myprofile", assert_error=True)
         self.assertIn("Invalid: Current cppstd (%s) is lower than the required C++ standard (17)."
                       % cppstd, self.client.out)
+
+
+def test_header_only_check_min_cppstd():
+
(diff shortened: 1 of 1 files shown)


lasote (approved):
I see this is pointing to develop2, Does this works in develop?

czoido:
> I see this is pointing to develop2, Does this works in develop?

It does not work well in develop, I opened this PR to test it: https://github.com/conan-io/conan/pull/12039

conans/test/integration/tools/cppstd_minimum_version_test.py
```diff
@@ -49,3 +49,63 @@ def test_invalid_cppstd_from_settings(self, cppstd):
...
+            settings = "compiler"
+            def package_id(self):
+                self.info.clear()
+            def validate(self):
+                check_min_cppstd(self, "11")
+        """)
+    client = TestClient()
+    client.save({"conanfile.py": conanfile})
+    client.run("create . -s compiler.cppstd=14")
+    client.run("install --require=fake/0.1@ -s compiler.cppstd=14")
```
memsharded, line 76: I don't think this is working as expected, I'll submit a new check

PR #12034, opened 2022-09-05
[POC] Initial cache for CMake

franramirez688:
Changelog: Feature: Added a `conan_intial_cache[_BUILD_TYPE].cmake` file to let `CMake` build helper run `cmake -C conan_intial_cache[_BUILD_TYPE].cmake ...` and load all the cache variables (`-Dxxxx` ones passed CLI).


conan/tools/cmake/cmake.py
```diff
@@ -67,28 +84,16 @@ def configure(self, variables=None, build_script_folder=None):
             cmakelist_folder = os.path.join(self._conanfile.source_folder, build_script_folder)
 
         build_folder = self._conanfile.build_folder
-        generator_folder = self._conanfile.generators_folder
-
         mkdir(self._conanfile, build_folder)
 
         arg_list = [self._cmake_program]
+        # Saving the initial cache cmake file
+        cache_path = self._save_cmake_initial_cache(extra_cache_variables=variables)
```
memsharded: I know this is a PoC, but just in case, the idea is that the file with the cache variables is saved by the toolchain, as an alternative or complement to the presets and toolchain. And ``CMake`` would just use it for ``-C file`` if existing.

franramirez688: Yeah, I know. I wanted to see how it looked there, but I agree, the toolchain should save that file, and if CMake receives more variables, it'll have to update the file before running the `-C file` command.

conan/tools/cmake/cmake.py
```diff
@@ -67,28 +65,17 @@ def configure(self, variables=None, build_script_folder=None):
...
 
         build_folder = self._conanfile.build_folder
-        generator_folder = self._conanfile.generators_folder
-
         mkdir(self._conanfile, build_folder)
 
         arg_list = [self._cmake_program]
+        # Running the initial cache file
+        if variables:
+            update_cmake_initial_cache(self._conanfile, variables)
```
memsharded: Maybe this is not necessary. If ``CMake`` helper wants to pass things in the command line, let it do it. At this point it is to late for the user to be able to use the cache file generated by the toolchain, so this can probably be simplified.

conan/tools/cmake/cmake_initial_cache.py
```diff
@@ -0,0 +1,74 @@
...
+            v = value
+        else:
+            type_ = "STRING"
+            v = f'"{value}"'
+        ret.append(f'set({name} {v} CACHE {type_} "" FORCE)')
+    return "\n".join(ret)
+
+
+def get_initial_cache_path(conanfile):
+    return os.path.join(conanfile.generators_folder, "CMakeInitialCache.cmake")
```
memsharded: Use a ``conan_initial_cache.cmake``, more aligned with the ``conan_toolchain.cmake`` naming

conan/tools/cmake/cmake_initial_cache.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Note: it needs the cmake presets file to load all the existing cache variables
+    already defined.
+
+    :param conanfile: ``ConanFile`` instance
+    """
+    cmake_presets = load_cmake_presets(conanfile.generators_folder)
+    configure_preset = get_configure_preset(cmake_presets, conanfile)
+    toolchain_file = configure_preset.get("toolchainFile")
+    cache_variables = configure_preset["cacheVariables"]
```
memsharded: If we go this way, the initial cache doesn't take the values from the Presets, but instead of? Can the presets also define a cache file?

conan/tools/cmake/cmake.py
franramirez688: Maybe, but I only wanted to keep all the possible flags in one file because if the user passes more than one, it'll be helpful to have all of them saved. Anyway, I can change this behavior if you think so

conan/tools/cmake/cmake_initial_cache.py
franramirez688: Well, I don't think so. Presets has all the information about the cache variables that's correct, but it couldn't call the initial cache file directly (as far as I know), so they are different ways to load the same cache variables. Then, for the developer flow, you could run only the presets one (it already has all the cache variables) or call the initial cache file at first.

lasote (changes requested):
I think the `CMake` build helper should still read the `CMakePresets` instead of the new cache file.  Some reasons: You still read the `self._generator ` from there, all the information in the Presets is all you need, and maybe in the future we add some configuration to inherit some presets from the user. So, unless we keep upgrading the CMake build helper reading the CMakePresets it won't work.

In addition to that, the purpose of the CMakePresets regarding tools, IDEs, etc is exactly that, parse the json and call `cmake` accordingly. 

Nevertheless, I consider the new file very useful and a simpler way for the consumer in the command line, with CMake version restrictions, to call it easier, but as a "fail safe" feature in case it cannot use Presets.

conan/tools/cmake/cmake.py
```diff
@@ -57,38 +58,48 @@ def __init__(self, conanfile):
...
+        """
+        Get all the variables that should be passed the CLI as ``cmake -Dxxxxx -Dyyyyy ....``
+        """
+        cache_variables = {}
+        generator_folder = self._conanfile.generators_folder
+        if self._toolchain_file:
+            if os.path.isabs(self._toolchain_file):
+                toolchain_file = self._toolchain_file
+            else:
+                toolchain_file = os.path.join(generator_folder, self._toolchain_file)
```
lasote: I know this is not from your PR, but I think this might be unnecessary, the toolchain file should be always there or we probably have to raise an error, and about the absolute path, we always write the absolute path unless I'm not seeing something. 
I say this because I would like to simplify so the toolchain can be directly extracted from the `CMakePresets.json` file.
And why? well, I still see the "initial cache" isolated from this. I would mind if the logic from there is reading also the toolchain from the file and not reusing code from the build helper.

conan/tools/cmake/cmake.py
```diff
@@ -57,38 +58,48 @@ def __init__(self, conanfile):
...
+            cache_variables["CMAKE_TOOLCHAIN_FILE"] = toolchain_file.replace("\\", "/")
+
+        if self._conanfile.package_folder:
+            pkg_folder = self._conanfile.package_folder.replace("\\", "/")
+            cache_variables["CMAKE_INSTALL_PREFIX"] = pkg_folder
```
lasote: This is the only variable I don't know what to do with it for the "initial cache", but I would take into account that the "initial cache" idea is to simplify the user calling `cmake` manually, so the INSTALL_PREFIX should be anyway specified because it is a local folder where you are going to install the contents.

conan/tools/cmake/cmake.py
```diff
@@ -57,38 +58,48 @@ def __init__(self, conanfile):
...
         build_folder = self._conanfile.build_folder
-        generator_folder = self._conanfile.generators_folder
-
         mkdir(self._conanfile, build_folder)
 
         arg_list = [self._cmake_program]
+        # Running the initial cache file
+        cache_variables = self._cache_variables
+        cache_variables.update(variables or {})
+        update_cmake_initial_cache(self._conanfile, cache_variables)
```
lasote: I don't like to update the initial cache from here, I would keep calling cmake with the "-D" like before.

conan/tools/cmake/cmake_initial_cache.py
```diff
@@ -0,0 +1,54 @@
...
+
+    :param conanfile: ``ConanFile`` instance
+    :param cache_variables: ``dict`` with variables to be saved as CACHE ones
+    """
+    cache_cmake_path = get_initial_cache_path(conanfile)
+    content = _format_cache_variables(cache_variables)
+    save(cache_cmake_path, "\n".join(content))
+
+
+def update_cmake_initial_cache(conanfile, cache_variables):
```
lasote: I don't really see a use case for this update. Is it valid to run `cmake` with `-C` and `-D` together? The install prefix can be typed manually if needed.

lasote (changes requested):
I still see the new "cache" too coupled to the "CMake" tool and I think they should be isolated even if they share a bit of code for extracting the information from the "CMakePresets".

lasote (commented):
Looking better!

czoido:
Finally, we have decided that this should be postponed until we get real feedback of users. The intention of this PR was to cover the case of people using older CMake versions (<3.19) that do not support CMakePresets to let them define the same toolchain cache_variables that are defined in the CMakePresets passing `-c initial_cache_vars_file` when configuring.
Instead of that, we will output the line with all the -D variables so users can do that manually until we get more feedback about this.

PR #12049, opened 2022-09-07
CMakeDeps managing CMAKE_MAP_IMPORTED_CONFIG_XXX

### LINKED ISSUE #12041: [bug] MAP_IMPORTED_CONFIG_* does not work in CMakeDeps
Hello, sorry for unformatted issue!

In #11788 @lasote changed the way how dependencies handled by introducing new targets ([here](https://github.com/conan-io/conan/blame/develop/conan/tools/cmake/cmakedeps/templates/target_configuration.py#L70) and [here](https://github.com/conan-io/conan/blame/develop/conan/tools/cmake/cmakedeps/templates/target_configuration.py#L150)). It seems that these targets were forgotten to mark `IMPORTED`.

The specific problem: this change breaks our build, where we are using dependencies with `build_type=Release`, while building consumer with any other `build_type` (e.g. `RelWithDebInfo`). We are doing this by carefully maintain the `CMAKE_MAP_IMPORTED_CONFIG_*` in consumer.

lasote:
Changelog: Bugfix: The `CMakeDeps` generator was not working properly when the consumer maps configurations of the dependencies using `CMAKE_MAP_IMPORTED_CONFIG_XXX`.
Docs: https://github.com/conan-io/docs/pull/2826

Close https://github.com/conan-io/conan/issues/12041


memsharded (approved):
Looking good, a bit risky, but IMPORTED is what should be used, and config suffix too

jcar87 (approved):
This looks good to me with two caveats:

* ~**Without this fix**, a situation where a consumer project as "Debug" as the build type, and the dependencies where installed as "Release" - CMake gives you a warning (or an error dependency on policy). The warning is indicative of the mismatch and the potential problems. **With this fix** in the same situation, there's no longer a CMake warning, but compilation will fail. - this may make the user experience confusing.~ 
If the consumer project does not explicitly do mapping, the observed behaviour is the same with and without this PR.


* The `CMAKE_MAP_IMPORTED_CONFIG_` in the consumers side accept, as value, a list, not just a single config. If we have a situation where a consumer runs `conan install` for multiple build types, and the CMake side has a configuration that can be satisfied by "any" of those - they could express something like `set(CMAKE_MAP_IMPORTED_CONFIG_RELEASE Release RelWithDebInfo MinSizeRel Debug "")`. I'd say this is very unusual, but as explained in this CMake [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/23660), that would cause _multiple_ generator expressions to be matched, likely resulting in libraries from different build times to be simultaneously linked...

PR #12050, opened 2022-09-07
Test source called once for all configurations

czoido:
Looks like it the source() method is called everytime for each configuration. Testing this does not work properly... and adding a fix

### DIFF
--- a/conans/test/integration/conanfile/folders_access_test.py
+++ b/conans/test/integration/conanfile/folders_access_test.py
@@ -98,40 +98,35 @@ class TestFoldersAccess(unittest.TestCase):
         self.client.run("export . --user=conan --channel=stable")
 
     def test_source_local_command(self):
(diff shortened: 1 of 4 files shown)


conans/test/integration/conanfile/folders_access_test.py
```diff
@@ -98,40 +98,35 @@ def setUp(self):
         self.client.run("export . --user=conan --channel=stable")
 
     def test_source_local_command(self):
-        c1 = conanfile % {"no_copy_source": False, "source_with_infos": False,
```
czoido, line 101: `source_with_infos` was not used any more, as there were two configurations that produced the same package revision but for the second time the source method did not run now the tests failed as a side effect of the changes in the PR

memsharded (approved):

lasote (approved):

Issue #11857, opened 2022-08-12
[develop2] Discuss special chars +-. in package names


memsharded:
I think that as a workaround it might be ok, but the general policy should be to avoid special symbols completely. Because there is a failure mode, if one package is called ``hello+``, and another is ``hello-``, they will conflict and overwrite each other.

Question to discuss with the team, should we further restrict the Conan 2.0 current ``validation_pattern = re.compile(r"^[a-z0-9_][a-z0-9_+.-]{1,100}$")`` validation pattern for the package names?

_Originally posted by @memsharded in https://github.com/conan-io/conan/pull/11826#pullrequestreview-1067886976_

memsharded:
Lets add a warning at export time, that warns when ``.-+`` is used in other fields rather than version field.

memsharded:
Closed by https://github.com/conan-io/conan/pull/12053, will be in beta.3

PR #12053, opened 2022-09-07
warn if special chars in exported refs

memsharded:
Close https://github.com/conan-io/conan/issues/11857

### DIFF
--- a/conans/model/recipe_ref.py
+++ b/conans/model/recipe_ref.py
@@ -3,6 +3,7 @@ import fnmatch
 import re
 from functools import total_ordering
 
+from conan.api.output import ConanOutput
 from conans.errors import ConanException
 from conans.model.version import Version
 from conans.util.dates import timestamp_to_str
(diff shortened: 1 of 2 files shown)


conans/model/recipe_ref.py
```diff
@@ -135,6 +136,16 @@ def validate_ref(self):
...
 
+        # Warn if they use .+- in the name/user/channel, as it can be problematic for generators
+        pattern = re.compile(r'[.+-]')
+        if pattern.search(self.name):
+            ConanOutput().warning(f"Name containing special chars is discouraged '{self.name}'")
+        if self.user and pattern.search(self.user):
+            ConanOutput().warning(f"User containing special chars is discouraged '{self.user}'")
+        if self.channel and pattern.search(self.channel):
+            ConanOutput().warning(f"Channel containing special chars is discouraged "
+                                  f"'{self.channel}'")
```
franramirez688, line 147: Could it be a bit noisy? I mean, how many times are those warnings appearing if the name contains any of those symbols?

franramirez688 (approved):

conans/model/recipe_ref.py
memsharded: This appears only when a package is being exported, that is, ``conan export`` or ``conan create``. But not in every other usage, requirement, or anywhere else

franramirez688: That's great, then.

PR #12060, opened 2022-09-07
[develop2] minor refactor for new test_package

memsharded:


conan/cli/commands/create.py
```diff
@@ -81,13 +81,11 @@ def create(conan_api, parser, *args):
                                                                 profile_host=profile_host)
 
     out.title("Computing dependency graph")
-    check_updates = args.check_updates if "check_updates" in args else False
```
memsharded, line 84: This was unused, this doesn't exist in a create command

conan/cli/commands/test.py
```diff
@@ -38,30 +38,36 @@ def test(conan_api, parser, *args):
     out.info("Profile build:")
     out.info(profile_build.dumps())
 
+    deps_graph = run_test(conan_api, path, ref, profile_host, profile_build, remotes, lockfile,
+                          args.update, build_modes=None)
+    save_lockfile_out(args, deps_graph, lockfile, cwd)
```
memsharded, line 43: This is a small change, not a pure refactor: the lockfile is saved only after the test is successful. If the test fails, no lockfile is created. I think it is better than having a failing test, yet had created a lockfile, wdyt?

czoido (approved):

PR #12075, opened 2022-09-09
define source_folder in test_package when layout()

### LINKED ISSUE #12073: [bug] Append `self.source_folder` into `Environment()`: TypeError: sequence item 0: expected str instance, NoneType found
  * Compiler+version: AppleClang 13
  * Conan version: 1.52.0
  * Python version: 3.10.6

run conan create on a recipe with this logic in `generate()`

```python
from conan.tools.cmake import cmake_layout

[...]

def layout(self):
    cmake_layout(self)

def generate(self):
    env = Environment()
    env.prepend_path("PKG_CONFIG_PATH", self.source_folder)
    env.vars(self).save_script("conanbuildenv_pkg_config_path")
```

Actually I've tested this block in a test_package (for conan v2 migration of `pkgconf` recipe in CCI)

```
ERROR: pkgconf/1.7.4 (test package): Error in generate() method, line 43
	envvars.save_script("conanbuildenv_pkg_config_path")
	TypeError: sequence item 0: expected str instance, NoneType found
```

memsharded:
Changelog: Feature: Define ``source_folder`` in test_package when ``layout()`` is defined.
Docs: Omit

This is a test, to see if something would be broken

Close https://github.com/conan-io/conan/issues/12073

### DIFF
--- a/conans/client/manager.py
+++ b/conans/client/manager.py
@@ -83,6 +83,7 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
         conanfile.folders.set_base_install(install_folder)
         conanfile.folders.set_base_imports(install_folder)
         conanfile.folders.set_base_generators(base_folder)
+        conanfile.folders.set_base_source(base_folder)
 
     if hasattr(conanfile, "layout") and test:
         install_folder = conanfile.generators_folder
--- a/conans/test/integration/layout/test_layout_generate.py
+++ b/conans/test/integration/layout/test_layout_generate.py
@@ -52,9 +52,9 @@ def test_layout_generate():
 def test_generate_source_folder():
     c = TestClient()
     conanfile = textwrap.dedent("""
+        import os
         from conan import ConanFile
         from conan.tools.cmake import cmake_layout
-        from conan.tools.env import Environment
 
         class Pkg(ConanFile):
             settings = "os", "compiler", "build_type", "arch"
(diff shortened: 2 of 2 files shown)


conans/client/manager.py
```diff
@@ -83,6 +83,7 @@ def deps_install(app, ref_or_path, install_folder, base_folder, graph_info, remo
         conanfile.folders.set_base_install(install_folder)
         conanfile.folders.set_base_imports(install_folder)
         conanfile.folders.set_base_generators(base_folder)
+        conanfile.folders.set_base_source(base_folder)
```
lasote, line 86: This would also be assigned when `no layout defined`. Is that ok?

conans/test/integration/layout/test_layout_generate.py
```diff
@@ -48,3 +49,50 @@ def build(self):
...
+        from conan.tools.cmake import cmake_layout
+
+        class Pkg(ConanFile):
+            settings = "os", "compiler", "build_type", "arch"
+
+            def layout(self):
+                cmake_layout(self)
+
+            def generate(self):
+                self.output.info("PKG_CONFIG_PATH {}!".format(os.path.exists(self.source_folder)))
```
lasote, line 66: how is related the PKG_CONFIG_PATH with this change in the code base?

conans/client/manager.py
memsharded: I think so, some ``test_package`` like the reported one has:
```python
def generate(self):
    env = Environment()
    env.prepend_path("PKG_CONFIG_PATH", self.source_folder)
```
Will still want the ``self.source_folder`` as argument whether the ``layout()`` is defined or not. I was exactly trying to see if this could break something because of this.

conans/test/integration/layout/test_layout_generate.py
memsharded: It is just an example, the original reported issue wanted to set-up the PKG_CONFIG_PATH, but it could be any other message.

czoido (approved):

PR #12081, opened 2022-09-12
add also test_requires to CMakeToolchain directories

memsharded:
Changelog: Fix: Add the ``test_requires`` to ``CMakeToolchain`` directories/folders to find cmake files inside them.
Docs: Omit

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -523,8 +523,9 @@ class FindFiles(Block):
         os_ = self._conanfile.settings.get_safe("os")
         is_apple_ = is_apple_os(self._conanfile)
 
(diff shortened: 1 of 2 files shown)


czoido (approved):

SSE4:
I see `Docs: omit` here, but can we document it better? right now it's a bit obscure...
e.g. there are `tool_requires`, `build_requires`, `test_requires` in addition to regular `requirements`.
I see that they may have an additional arguments (e.g. `transitive_headers` and `force_host_context` are briefly mentioned in [Migrating the recipes](https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html) article).
however, I cannot find a reference for their full signatures anywhere, e.g. what is the list of possible parameters and what are their types?

PR #12097, opened 2022-09-13
Fix XcodeDeps using full package reference in case of package with components

### LINKED ISSUE #12089: [bug] XcodeDeps using full package reference in case of package with components leads to invalid xcconfig setup
XcodeDeps generated setup is incorrect in case you are referencing to a full package in the requires section of cpp_info object, whilst the referenced package might be sliced into further subcomponents.

Either the `conan_<dependency>_<dependency>.xcconfig` should be produced additionally, or it should be referenced in the `include` reference as `conan_<dependency>.xcconfig`. 

Unless it might be expected that consumer will have to switch to utilization of components, but then an error should be thrown when full package reference is utilised within cpp_info requirement context.

  * Operating System+version: n/a
  * Compiler+version: n/a
  * Conan version: 1.52
  * Python version: 3.9

```
diff --git a/conans/test/functional/toolchains/apple/test_xcodedeps_components.py b/conans/test/functional/toolchains/apple/test_xcodedeps_components.py
index 2a50ab1e0..9b0d13c2a 100644
--- a/conans/test/functional/toolchains/apple/test_xcodedeps_components.py
+++ b/conans/test/functional/toolchains/apple/test_xcodedeps_components.py
@@ -1,3 +1,4 @@
+import glob
 import os
 import platform
 import textwrap
@@ -136,7 +137,7 @@ def test_xcodedeps_components():
         self.cpp_info.libs = ["chat"]
         self.cpp_info.includedirs.append("include")
         self.cpp_info.libdirs.append("lib")
-        self.cpp_info.requires.append("network::client")
+        self.cpp_info.requires.append("network::network")
     """

     cmakelists_chat = textwrap.dedent("""
@@ -146,7 +147,7 @@ def test_xcodedeps_components():
         add_library(chat src/chat.cpp include/chat.h)
         target_include_directories(chat PUBLIC include)
         set_target_properties(chat PROPERTIES PUBLIC_HEADER "include/chat.h")
-        target_link_libraries(chat network::client)
+        target_link_libraries(chat network::network)
         install(TARGETS chat)
         """)

@@ -184,9 +185,9 @@ def test_xcodedeps_components():
     client.run("install chat/1.0@ -g XcodeDeps --install-folder=conan -s build_type=Debug "
                "--build=missing")
     chat_xcconfig = client.load(os.path.join("conan", "conan_chat_chat.xcconfig"))
-    assert '#include "conan_network_client.xcconfig"' in chat_xcconfig
-    assert '#include "conan_network_server.xcconfig"' not in chat_xcconfig
-    assert '#include "conan_network_network.xcconfig"' not in chat_xcconfig
+    assert '#include "conan_network_network.xcconfig"' in chat_xcconfig
+    assert os.path.isfile(os.path.join(client.current_folder, "conan", "conan_network.xcconfig"))
+    assert os.path.isfile(os.path.join(client.current_folder, "conan", "conan_network_network.xcconfig"))
     host_arch = client.get_default_host_profile().settings['arch']
     arch = "arm64" if host_arch == "armv8" else host_arch
     client.run_command("xcodegen generate")
```

czoido:
Changelog: Fix: Fix XcodeDeps using full package reference in case of package with components.
Docs: omit

Testing a fix when the syntax `self.cpp_info.requires = ["dep::dep"]` or `self.cpp_info.components["comp"].requires = ["dep::dep"]` is used for a package with components, it should include all the components

Closes: https://github.com/conan-io/conan/issues/12089

### DIFF
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -77,9 +77,8 @@ class XcodeDeps(object):
     _dep_xconfig = textwrap.dedent("""\
         // Conan XcodeDeps generated file for {{pkg_name}}::{{comp_name}}
         // Includes all configurations for each dependency
-        {% for dep in deps %}
-        // Includes for {{dep[0]}}::{{dep[1]}} dependency
-        #include "conan_{{format_name(dep[0])}}_{{format_name(dep[1])}}.xcconfig"
+        {% for include in deps_includes %}
+        #include "{{include}}"
         {% endfor %}
         #include "{{dep_xconfig_filename}}"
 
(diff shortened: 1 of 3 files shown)


czoido:
@maikelvdh could you please check if this would fix your use case?

maikelvdh:
@czoido this seems to address the issue reported! thanks a lot! 

With regards of the invalid component name reference that will be handled separately I suppose?

czoido:
> @czoido this seems to address the issue reported! thanks a lot!
> 

Thanks for testing!

> With regards of the invalid component name reference that will be handled separately I suppose?

Yes, I think that check should be addressed elsewhere, we already removed those checks in the cpp_info for 2.0 so we will check if we should recover some of them.

PR #12117, opened 2022-09-14
[Meson] Meson appending extra machine files

### LINKED ISSUE #11961: [feature] MesonToolchain custom binaries
Meson can use the binaries section in the meson ini file for finding programs. See https://mesonbuild.com/Machine-files.html#binaries

Currently only a specific list of binaries are added `c, cpp, objc, objcpp, c_ld, cpp_ld, ar, strip, as, windres, pkgconfig`. It would be great if these were dictionary that allows any binary.

franramirez688:
Changelog: Feature: Add `tools.meson.mesontoolchain:extra_machine_files=["FILENAMES"]` to `Meson` build helper to append machine files to the the ones created by Conan.
Docs: https://github.com/conan-io/docs/pull/2831
Closes: https://github.com/conan-io/conan/issues/11961

### DIFF
--- a/conans/test/functional/toolchains/meson/_base.py
+++ b/conans/test/functional/toolchains/meson/_base.py
@@ -1,5 +1,4 @@
 import platform
-import sys
 import unittest
 
 import pytest
@@ -8,7 +7,6 @@ from conans.test.utils.tools import TestClient
 
 
 @pytest.mark.tool_meson
-@pytest.mark.skipif(sys.version_info.major == 2, reason="Meson not supported in Py2")
 @pytest.mark.skipif(platform.system() not in ("Darwin", "Windows", "Linux"),
                     reason="Not tested for not mainstream boring operating systems")
 class TestMesonBase(unittest.TestCase):
--- a/conans/test/functional/toolchains/meson/test_meson.py
+++ b/conans/test/functional/toolchains/meson/test_meson.py
@@ -220,7 +220,7 @@ def test_meson_and_user_filenames_composition():
     client.run("install . -pr=profile -if build")
     client.run("build . -bf build", assert_error=True)  # fails because of the fake compiler
     # Checking the order of the appended user file (the order matters)
-    assert f'meson setup --native-file "{client.current_folder}/build/conan_meson_native.ini" ' \
-           f'--native-file "myfilename.ini"' in client.out
+    conan_meson_native = os.path.join(client.current_folder, "build", "conan_meson_native.ini")
+    assert f'meson setup --native-file "{conan_meson_native}" --native-file "myfilename.ini"' in client.out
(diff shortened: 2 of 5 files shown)


conans/test/functional/toolchains/meson/_base.py
```diff
@@ -8,7 +7,6 @@
 
 
 @pytest.mark.tool_meson
-@pytest.mark.skipif(sys.version_info.major == 2, reason="Meson not supported in Py2")
```
franramirez688, line 11: No python2x anymore

conans/test/functional/toolchains/meson/test_meson.py
```diff
@@ -167,3 +170,57 @@ def test(self):
...
+        compiler.cppstd=17
+        compiler.libcxx=libstdc++11
+        build_type=Release
+
+        [conf]
+        tools.meson.mesontoolchain:user_filenames=["myfilename.ini"]
+   """)
+    myfilename = textwrap.dedent("""
+    [binaries]
+    c = '/usr/fake/path/gcc'
```
memsharded: The definition of compilers might be "standardized" by Conan in its own configuration (or handle env-var CC CXX for everyone), so maybe not a good example

memsharded (commented):
I am not convinced that the user story is injecting some meson files to all meson projects, or the users are missing a "local" way to inject those files inside package recipes, but like a different file for each project. I don't know enough about Meson usage and patterns, but we need to think about it.

eli-schwartz:
From a documentation perspective, what is a meson "user filename"? Shouldn't this use a name more evocative of the fact that Meson calls them native machine files?

lasote (approved):

lasote:
> From a documentation perspective, what is a meson "user filename"? Shouldn't this use a name more evocative of the fact that Meson calls them native machine files?

But there are `native` and `cross` files, right? Maybe something like  `tools.meson.mesontoolchain:additional_machine_files` ?

franramirez688:
> I am not convinced that the user story is injecting some meson files to all meson projects, or the users are missing a "local" way to inject those files inside package recipes, but like a different file for each project. I don't know enough about Meson usage and patterns, but we need to think about it.

I've changed the configuration name to `tools.meson.mesontoolchain:extra_machine_files` so I think it's more evident than before. In my opinion, the user story is if you want to add your own variables from any section to the existing Conan ones, so we don't need to add more public attributes to the `MesonToolchain` class interface to manage all the machine file sections.

franramirez688:
> From a documentation perspective, what is a meson "user filename"? Shouldn't this use a name more evocative of the fact that Meson calls them native machine files?

Changed to `tools.meson.mesontoolchain:extra_machine_files`. I think it's clear enough, doesn't it?

memsharded (approved):

PR #12124, opened 2022-09-15
add get_version to git helper

AndreyMlashkin:

### DIFF
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -4,6 +4,7 @@ from conan.tools.files import chdir
 from conans.errors import ConanException
 from conans.util.files import mkdir
 from conans.util.runners import check_output_runner
+from conan.tools.scm import Version
 
 
 class Git(object):
@@ -26,6 +27,12 @@ class Git(object):
             return commit
         except Exception as e:
             raise ConanException("Unable to get git commit in '%s': %s" % (self.folder, str(e)))
+
+    def get_version(self):
+        try:
+            return Version(self._run('describe --always --tags --abbrev=9'))
+        except Exception as e:
+            raise ConanException("Unable to get git version in '%s': %s" % (self.folder, str(e)))
 
     def get_remote_url(self, remote="origin"):
         remotes = self._run("remote -v")


AndreyMlashkin:
Calling `describe --always --tags --abbrev=9` is a super convinient way of getting a current version of a repository. If you are in a tagged version, you will get a string like 12.34.56, if you are not, you will get a string like:
12.34.56-**8**-g169116910
Where:
12.34.56 is the latest tag
**8** - is a number of commits on top of the tag 
g169116910 is a hash of the latest commit

AndreyMlashkin:
About a CI error: why it can't import Version?
```
from conan.tools.scm import Version
ImportError: cannot import name 'Version'
```

conan/tools/scm/git.py
```diff
@@ -27,6 +28,12 @@ def get_commit(self):
         except Exception as e:
             raise ConanException("Unable to get git commit in '%s': %s" % (self.folder, str(e)))
 
+    def get_version(self):
```
memsharded, line 31: This might be a bit confusing, in the past this interface was used to get the Git version, not the repo version.

conan/tools/scm/git.py
```diff
@@ -4,6 +4,7 @@
 from conans.errors import ConanException
 from conans.util.files import mkdir
 from conans.util.runners import check_output_runner
+from conan.tools.scm import Version
```
memsharded, line 7: If the import is failing, could be some circular imports, try to put it local to the method.

conan/tools/scm/git.py
```diff
@@ -27,6 +28,12 @@ def get_commit(self):
         except Exception as e:
             raise ConanException("Unable to get git commit in '%s': %s" % (self.folder, str(e)))
 
+    def get_version(self):
+        try:
+            return Version(self._run('describe --always --tags --abbrev=9'))
```
memsharded, line 33: The 9 magic number seems a bit arbitray, isn't it? Maybe it is a good common practice for git? Users might want another lenght?

memsharded (commented):
I am not sure about this, seems a bit too user specific. 
Maybe we can just make the ``self._run()`` method public, so users can easily do ``git.run(...)`` in their recipes, to adjust to their needs?

Also, a unit test to cover it would be necessary, please ask if you need help with it (but lets get first a bit of extra feedback about this, I'll assign other reviewers)

conan/tools/scm/git.py
AndreyMlashkin: Yes, it's arbitray, you're right. Github UI, for example, shows 7 symbols of the last cache

czoido:
Thanks a lot for the contribution but I would say that this looks very specific of a use case and is not such a popular pattern so that many users would benefit from this. 
IMHO I don't think we should add this to the client as it can be easily used in recipes just calling `self.run("git describe --always --tags --abbrev=9"` or adding the feature to use `git.run("...")` as @memsharded said

AndreyMlashkin:
Agree, as I am using it this way with conan v1

AndreyMlashkin:
You still may consider adding a method

get_the_last_tag, calling
```git describe --always --tags```

AndreyMlashkin:
Feel free to close this task, when you create a PR for making _run method public :)
I would leave this opened as reminder

memsharded:
Closed by https://github.com/conan-io/conan/pull/12165, will be in next 1.53 release

PR #12131, opened 2022-09-15
Add check during package download to prevent race condition

nanaceba:
Changelog: (Bugfix): In some cases we see race condition when 2 conan instances running at the same time and downloading same package. While locks prevent two conans from downloading and overwriting files in parallel, they currently do not prevent second conan from purging the directory prepared by the first one and re-unzipping files there.

### DIFF
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -26,6 +26,7 @@ from conans.model.env_info import EnvInfo
 from conans.model.graph_info import GraphInfo
 from conans.model.graph_lock import GraphLockFile
 from conans.model.info import PACKAGE_ID_UNKNOWN
+from conans.model.manifest import FileTreeManifest
 from conans.model.new_build_info import NewCppInfo, fill_old_cppinfo
 from conans.model.ref import PackageReference
 from conans.model.user_info import DepsUserInfo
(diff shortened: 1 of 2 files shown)


memsharded:
Hi @nanaceba 

Thanks for your contribution.

I'd like to understand better:

>  they currently do not prevent second conan from purging the directory prepared by the first one and re-unzipping files there.

Why is that the current lock is not protecting against this?

As you can see in the broken tests in our CI https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-12131/4/pipeline, this kind of change is risky and can break things. 

Also, the priority right now is 2.0. Concurrency in the 2.0 cache is still not there, but we will eventually work on it, after 2.0. Changes in 1.X branches only happen for regressions and necessary backports and features for the 1.X -> 2.0 migration.

memsharded:
It seems this PR got stale, as the changes didn't pass CI tests, and the changes seemed a bit too risky.

Also, Conan 1.X is not getting this kind of changes anymore, but any significant new changes like this one should target Conan 2.0. In this regard Conan 2.0 has completely removed concurrency for the cache at this moment, and it will be a 2.X roadmap feature to implement it again.

Issue #12128, opened 2022-09-15
conan-server version 2.0.0b3 does not start


fourbft:
This is a conan 2.0.0-beta issue. I hope, this issue tracker is for 2.0 development issues, too. (Haven't found any information about special report channels for 2.0)

### Environment Details (include every applicable attribute)
  * Ubuntu-22.04, but same behavior in a clean alpine-python docker image (with python 3.9)
  * Python 3.10.4
  * Conan version 2.0.0b3
  
### Steps to reproduce (Include if Applicable)
Install conan_server:

$ sudo pip3 install conan_server

Try to start conan_server:

$ conan_server --help

### Logs (Executed commands with output) (Include/Attach if Applicable)
```
$ conan_server --help
Traceback (most recent call last):
  File "/usr/local/bin/conan_server", line 5, in <module>
    from conans.conan_server import run
  File "/usr/local/lib/python3.10/dist-packages/conans/conan_server.py", line 3, in <module>
    from conans.server.launcher import ServerLauncher
  File "/usr/local/lib/python3.10/dist-packages/conans/server/launcher.py", line 8, in <module>
    from conans.server.conf import get_server_store
  File "/usr/local/lib/python3.10/dist-packages/conans/server/conf/__init__.py", line 15, in <module>
    from conans.server.store.server_store import ServerStore
  File "/usr/local/lib/python3.10/dist-packages/conans/server/store/server_store.py", line 5, in <module>
    from conans.model.package_ref import PkgReference
  File "/usr/local/lib/python3.10/dist-packages/conans/model/package_ref.py", line 2, in <module>
    from conans.model.recipe_ref import RecipeReference
  File "/usr/local/lib/python3.10/dist-packages/conans/model/recipe_ref.py", line 6, in <module>
    from conan.api.output import ConanOutput
  File "/usr/local/lib/python3.10/dist-packages/conan/__init__.py", line 1, in <module>
    from conans.model.conan_file import ConanFile
  File "/usr/local/lib/python3.10/dist-packages/conans/model/conan_file.py", line 7, in <module>
    from conans.model.conf import Conf
  File "/usr/local/lib/python3.10/dist-packages/conans/model/conf.py", line 6, in <module>
    from conans.model.recipe_ref import ref_matches
ImportError: cannot import name 'ref_matches' from partially initialized module 'conans.model.recipe_ref' (most likely due to a circular import) (/usr/local/lib/python3.10/dist-packages/conans/model/recipe_ref.py)
```

[...]

PR #12134, opened 2022-09-16
Test conan_server runs in conan ci

czoido:
Test the installation of conan_server after running functional tests.

Closes: https://github.com/conan-io/conan/issues/12128


.ci/jenkins/testsv2.jenkins
```diff
@@ -50,6 +50,15 @@ String getDockerImage(String moduleName) {
     return ""
 }
 
+String shouldTestServer(String moduleName) {
```
memsharded, line 53: String return type reads weird

memsharded (approved):
I think in the long term I'd make this kind of test a "e2e" or "installer" test, and put it in a separate independent folder and job, that runs after functional, only pre-release or nightly.

czoido:
> I think in the long term I'd make this kind of test a "e2e" or "installer" test, and put it in a separate independent folder and job, that runs after functional, only pre-release or nightly.

Yes, I agree, this was a temporary fix. Let's try to improve testing in the future

Issue #11658, opened 2022-07-17
[develop2] Consider re-enabling of file download cache


memsharded:
We are removing the "file download cache" in 2.0, because of some complexities, missing development in the 2.0 cache concurrency and lack of consistent feedback of the performance gains of the "file download cache".

In order to consider re-enabling the file download cache we need reports for different users about:

- What performance gains are being observed by enabling it
- What is the most important perceived value
- Is it being enabled for developers machines, CI, or both? For all OS? is the performance similar in all OS?
- Is it necessary to decouple caching of internet files (sources downloads) and caching of files coming from a conan server (like a conan_package.tgz from a Conan package binaries)

As we are no longer implementing this kind of new features in Conan 1.X, and the feature has been removed in 2.0, I am closing this issue, and opening a new one for 2.X scope from this comment.

_Originally posted by @memsharded in https://github.com/conan-io/conan/issues/9428#issuecomment-1186622504_

madebr:
I heavily use this caching whilst creating and/ormodifying recipes for 3rd party dependencies, e.g. for those on conan-center-index.
It is a huge time saver + is also useful to keep working on a recipe when the internet connection is flaky.

memsharded:
> I heavily use this caching whilst creating and/ormodifying recipes for 3rd party dependencies, e.g. for those on conan-center-index.

So that means, that you are only interested in the caching of downloaded sources from the internet, and not the rest of the features.

This use case might be also partially addressed by the recent ``file://`` support in get/download. Not exactly the same, it still requires some manual work, but would be something to consider until this is decided.

MTomBosch:
Hello,
for us the performance gain is not the only argument why it is required.
For us as important as the performance is the reduction of traffic from our JFrog SaaS instance towards the CI/CD build machines.
Without the downloac cache we will need to pay a 5 digit number $ per year more than right now with the download cache.
If there is no download cache or something similar to reduce the traffic from Artifactory towards the CI/CD build machine we will not be able to move to Conan 2.0 at all. So please get it working for 2.0 again or provide something similar.

jellespijker:
We also rely heavily on caching downloads between pipeline runs on our GitHub Actions.
https://github.com/Ultimaker/Cura/blob/d69819a24e89af2b35266b25c5bc91561048bb78/.github/workflows/conan-package-create.yml#L80

[...]

PR #12138, opened 2022-09-18
Re-enabling file download cache in 2.0

memsharded:
Close https://github.com/conan-io/conan/issues/11658


conans/model/conf.py
```diff
@@ -49,6 +50,7 @@
     "tools.cmake.cmaketoolchain:toolset_arch": "Toolset architecture to be used as part of CMAKE_GENERATOR_TOOLSET in CMakeToolchain",
     "tools.cmake.cmaketoolchain.presets:max_schema_version": "Generate CMakeUserPreset.json compatible with the supplied schema version",
     "tools.cmake.cmake_layout:build_folder_vars": "Settings and Options that will produce a different build folder and different CMake presets names",
+    "tools.files.download:download_cache": "Define the cache folder to store downloads from files.download()/get()",
```
czoido, line 53: Why do we need two separate confs? Why not using only one configuration and store all the downloaded files in the same cache?

memsharded: There are 2 reasons:
- One technical: the ``conanfile.conf`` received by recipes do not have visibility over the ``core.*`` configuration that is suppossed to be internal only, but not visible or usable by recipes. But having a ``tools.*`` configuration for defining how the internal Conan artifacts downloads are cached seems wrong, as it has nothing to do with the tools, it is something "core"
- The other is functional: there might be users that want to cache only the third-party things downloaded from the internet, but not their own Artifactory files. Or maybe the cleaning policy is different, keeping like forever the files downloaded from ``source()`` from the internet, while cleaning from time to time the cache from internal conan artifacts.

memsharded: I am not saying this is immutable, it is true that it is not super-intuitive at first sight. Just trying to explain why it is the way it is. In Conan 1.X is already similar, ``conan.conf`` is the core one, while profiles ``tools.*`` is the downloaded-from-internet one.

czoido: I see, I was not aware of that, I guess that users demanded and use that functionality 👍

czoido (approved):

czoido:
While testing this locally I noticed that if you specify a folder that is not absolute it will put the file download cache inside the folder of the package in the local cache, like this:

image


It's not that is critical, but maybe if it does not make sense to use a folder that is not absolute for this, we should at least warn the user.

memsharded:
> It's not that is critical, but maybe if it does not make sense to use a folder that is not absolute for this, we should at least warn the user.

No, this is not good, I'll fix it

memsharded:
Fixed

jcar87 (approved):

PR #12145, opened 2022-09-19
add support for windows on ARM64 native platform

### LINKED ISSUE #12144: [bug] ConanException: unsupported architecture armv8 on Windows on ARM
Attempting to use `conans.CMake.configure` on WIndows on ARM, fails with the error:

```
ConanException: unsupported architecture armv8
```

I tracked the error to be [here](https://github.com/conan-io/conan/blob/a620e5b7ba6b0d2128edd2056b7391b5fabcd77d/conans/client/tools/win.py#L447).

Apparently, Conan still isn't aware of the native ARM support that came with Visual Studio 2022 17.3 (there is now `vcvarsarm64.bat` which configures for native ARM development on the host ARM machine).

  * Operating System+version: Windows 11 on ARM64
  * Compiler+version: Visual Studio 2022 17.3 Developer Preview for ARM64
  * Conan version: 1.52.0
  * Python version: 3.9.7

DoDoENT:
Changelog: Fix: Add support for windows on ARM64 native platform.
Docs: omit

Closes: #12144

### DIFF
--- a/conans/client/tools/win.py
+++ b/conans/client/tools/win.py
@@ -442,6 +442,11 @@ def vcvars_command(conanfile=None, arch=None, compiler_version=None, force=False
                        'x86_64': 'x86_amd64',
                        'armv7': 'x86_arm',
                        'armv8': 'x86_arm64'}.get(arch_setting)
+    elif arch_build == 'armv8':
+        vcvars_arch = {'x86': 'arm64_x86',
+                       'x86_64': 'arm64_x64',
+                       'armv7': 'arm64_arm',
+                       'armv8': 'arm64'}.get(arch_setting)
 
     if not vcvars_arch:
         raise ConanException('unsupported architecture %s' % arch_setting)
--- a/conans/test/functional/tools/old/vcvars/vcvars_arch_test.py
+++ b/conans/test/functional/tools/old/vcvars/vcvars_arch_test.py
@@ -53,6 +53,17 @@ class VCVarsArchTest(unittest.TestCase):
         settings.arch = 'x86'
         self.assert_vcvars_command(settings, "amd64_x86")
 
+        settings.arch_build = 'armv8'
+        settings.arch = 'armv8'
+
+        self.assert_vcvars_command(settings, "arm64")
+
+        self.arch = 'x86'
+        self.assert_vcvars_command(settings, "arm64_x86")
+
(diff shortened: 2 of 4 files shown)


conans/client/tools/win.py
```diff
@@ -442,6 +442,11 @@ def vcvars_command(conanfile=None, arch=None, compiler_version=None, force=False
                        'x86_64': 'x86_amd64',
                        'armv7': 'x86_arm',
                        'armv8': 'x86_arm64'}.get(arch_setting)
+    elif arch_build == 'armv8':
```
memsharded, line 445: This is modifying only the legacy tool ``from conans``
This should be done as a secondary thing, as a backport, but this should be addressed first in ``from conan`` space.

Also some unittest would be necessary, please ask for guidance if you need it.
Thanks for contributing to Conan!

DoDoENT: Thank you for the pointers. I've added the same code also for non-legacy `from conan`.

I've also updated the only unit test I found that deals with testing the vcvars, but it also works with the legacy `from conans`. Also, how to run just this single test? I've run `python3 -m tox -e full` as described in the documentation and more than 1400 tests have failed, most of them completely unrelated to my change. Are they supposed to fail?

conans/test/functional/tools/old/vcvars/vcvars_arch_test.py
```diff
@@ -53,6 +53,17 @@ def test_arch(self):
         settings.arch = 'x86'
         self.assert_vcvars_command(settings, "amd64_x86")
 
+        settings.arch_build = 'armv8'
+        settings.arch = 'armv8'
+
+        self.assert_vcvars_command(settings, "arm64")
+
+        self.arch = 'x86'
```
memsharded:
```suggestion
        settings.arch = 'x86'
```

conans/test/functional/tools/old/vcvars/vcvars_arch_test.py
```diff
@@ -53,6 +53,17 @@ def test_arch(self):
...
+        self.assert_vcvars_command(settings, "arm64_x86")
+
+        self.arch = 'x86_64'
```
memsharded:
```suggestion
        settings.arch = 'x86_64'
```

PR #12156, opened 2022-09-20
Adding libcxx support to MesonToolchain

### LINKED ISSUE #12009: [bug] MesonToolchain does not respect libcxx setting
  * Operating System+version: Ubuntu 22.04 LTS
  * Compiler+version: Clang 13
  * Conan version: 1.52.0
  * Python version: 3.10.3

When running a verbose build, the `-stdlib=libc++` option does not appear to be passed to Clang, causing the test package to fail with linking errors. As there are no other CCI packages using the new MesonToolchain to build C++ libraries, I'm not sure if this is a bug or if I am missing some generation step.

See libsigcpp recipe in https://github.com/conan-io/conan-center-index/pull/12775

memsharded:
Changelog: Bugfix: Implement correct ``libcxx`` support in ``MesonToolchain``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/12009

### DIFF
--- a/conan/tools/_compilers.py
+++ b/conan/tools/_compilers.py
@@ -79,7 +79,7 @@ def libcxx_flags(conanfile):
                "libstdc++": "-library=stdcpp"
                }.get(libcxx)
     elif compiler == "qcc":
-        return "-Y _{}".format(libcxx)
+        lib = "-Y _{}".format(libcxx)
 
     if compiler in ['clang', 'apple-clang', 'gcc']:
         if libcxx == "libstdc++":
--- a/conan/tools/_compilers.py
+++ b/conan/tools/_compilers.py
@@ -57,6 +57,39 @@ def architecture_flag(settings):
     return ""
 
 
+def libcxx_flags(conanfile):
+    libcxx = conanfile.settings.get_safe("compiler.libcxx")
+    if not libcxx:
+        return None, None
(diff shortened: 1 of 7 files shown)


conan/tools/_compilers.py
```diff
@@ -57,6 +57,39 @@ def architecture_flag(settings):
...
+               "libstlport": "-library=stlport4",
+               "libstdc++": "-library=stdcpp"
+               }.get(libcxx)
+    elif compiler == "qcc":
+        lib = "-Y _{}".format(libcxx)
+
+    if compiler in ['clang', 'apple-clang', 'gcc']:
+        if libcxx == "libstdc++":
+            stdlib11 = "_GLIBCXX_USE_CXX11_ABI=0"
+        elif libcxx == "libstdc++11" and conanfile.conf.get("tools.gnu:define_libcxx11_abi",
```
czoido, line 87: Shouldn't this be an `or` ?

memsharded: No. Compilers default is setting it to 1, so it is not necessary to explicitly do it, and some compilers might even emit a warning or error if the flag is unexpected, this is why the conf ``tools.gnu:define_libcxx11_abi`` was added, as an opt-in to force the definition of _GLIBCXX_USE_CXX11_ABI=1, which is not necessary in most cases

PR #12169, opened 2022-09-21
Add message to invoke cmake manually as an alternative to call the presets

czoido:
Changelog: Feature: Add message to help users that have old CMake versions invoking CMake manually with the same information stored in the CMakePresets.
Docs: omit

Closes: https://github.com/conan-io/conan/pull/12034

Defining this:

```python
    def generate(self):
        tc = CMakeToolchain(self)
        tc.cache_variables["MYNUMBER"] = 1
        tc.cache_variables["MYMESSAGE"] = "MyValue"
        tc.cache_variables["MESSAGE_SPACES"] = "My    Value"
        tc.generate()
```

Something like:

```
conanfile.py (hello/0.1): Preset 'release' added to CMakePresets.json. Invoke it manually using 'cmake --preset release' or if your CMake version is not compatible with the presets feature (<3.19) call cmake configure like: 'cmake <path> -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/Users/carlosz/Documents/developer/conan/sandbox/toolchain_vars/build/generators/conan_toolchain.cmake -DMYNUMBER=1 -DMYMESSAGE=MyValue -DMESSAGE_SPACES="My    Value" -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Release'
```

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -94,8 +94,9 @@ def _configure_preset(conanfile, generator, cache_variables, toolchain_file, mul
         if "CMAKE_TOOLCHAIN_FILE" not in cache_variables_info else ""
 
     conanfile.output.info(f"Preset '{name}' added to CMakePresets.json. Invoke it manually using "
-                          f"'cmake --preset {name}' or if your CMake version is not compatible with "
-                          f"the presets feature (<3.19) call cmake configure like: 'cmake <path> "
+                          f"'cmake --preset {name}'")
+    conanfile.output.info(f"If your CMake version is not compatible with "
+                          f"CMakePresets (<3.19) call cmake like: 'cmake <path> "
                           f"-G {_format_val(generator)} {add_toolchain_cache}"
                           f"{cache_variables_info}'")
     return ret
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/presets.py
```diff
@@ -84,6 +84,20 @@ def _configure_preset(conanfile, generator, cache_variables, toolchain_file, mul
...
+
+    def _format_val(val):
+        return f'"{val}"' if type(val) == str and " " in val else f"{val}"
+
+    # https://github.com/conan-io/conan/pull/12034#issuecomment-1253776285
+    cache_variables_info = " ".join([f"-D{var}={_format_val(value)}" for var, value in cache_variables.items()])
+    add_toolchain_cache = f"-DCMAKE_TOOLCHAIN_FILE={toolchain_file} " \
+        if "CMAKE_TOOLCHAIN_FILE" not in cache_variables_info else ""
+
+    conanfile.output.info(f"Preset '{name}' added to CMakePresets.json. Invoke it manually using "
```
memsharded, line 96: Maybe splitting in different lines will help readability

memsharded (approved):

conan/tools/cmake/presets.py
czoido: I can split it in two lines but the line is going to be superlong anyway because of the arguments for the cmake invokation... :/

memsharded (approved):

PR #12174, opened 2022-09-22
Default to source_folder in conan.tools.files.patch if exports_sources is not defined and using local flow

### LINKED ISSUE #11781: [bug] self.export_sources_folder is empty in local build
### Environment Details
  * Operating System+version: Linux Mint 20.3
  * Compiler+version: gcc 9
  * Conan version: 1.51
  * Python version: 3.8.10

### Steps to reproduce
```
conan install . -pr:h host -if build
conan source . -sf build
conan build . -bf build -sf build
```

In both `source()` and `build()` methods `self.export_sources_folder` is empty. If I use `conan create`, it's the same as `self.source_folder`.

This breaks the `patch` tool which uses `export_sources_folder` internally.

czoido:
Changelog: Fix: Default to source_folder in conan.tools.files.patch if exports_sources is not defined
Docs: TODO

Closes: https://github.com/conan-io/conan/issues/11781

### DIFF
--- a/conan/tools/files/patches.py
+++ b/conan/tools/files/patches.py
@@ -49,7 +49,7 @@ def patch(conanfile, base_path=None, patch_file=None, patch_string=None, strip=0
     if patch_file:
         # trick *1: patch_file path could be absolute (e.g. conanfile.build_folder), in that case
         # the join does nothing and works.
-        patch_path = os.path.join(conanfile.export_sources_folder, patch_file)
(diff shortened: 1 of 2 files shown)


memsharded:
This works without the ``conan source . -sf=build`` argument. All the new tools assume usage of ``layout()`` and deprecation of the local flow arguments (except ``--output-folder``). So this probably not worth the risk of breaking other things.

PR #12178, opened 2022-09-22
initial new proposal for win_bash_run

### LINKED ISSUE #10900: [question/feature] conan.tools.env » VirtualRunEnv does not generate sh scripts on windows

### LINKED ISSUE #12192: [feature] Don't trick users in `conan.tools.microsoft.unix_path` with internal details like win_bash or subsystem

Please, do not try to be too smart and guess what users want: https://github.com/conan-io/conan/blob/0fe6058771c7d9585f9fed1935a4e0ec078d3314/conan/tools/microsoft/subsystems.py#L5-L6

When someone uses `unix_path`, he wants a unix_path regardless of the value of `self.win_bash`.

Because maybe he may want to enable win_bash in very specific parts of a conanfile (for example only in build() & package() when we may really rely on a shell, but not in generate() where we prepare files to be injected and don't call any bash script), and for such function as unix_path, it's really unexpected to not get... a unix path.

Maybe also we want to have a unix_path in package_info(), and win_bash doesn't really make sense, it can take any value, it doesn't matter.

I expect many mistakes with current logic in conan-center recipes.

According to the current logic of `unix_path`, it should be called `native_system_path` or something like this, it's misleading.

memsharded:
Changelog: Feature: New ``conanfile.win_bash_run`` and ``tools.microsoft.bash:active`` config for better Windows bash management.
Docs: https://github.com/conan-io/docs/pull/2803

Close https://github.com/conan-io/conan/issues/10900
Close https://github.com/conan-io/conan/issues/12192

### DIFF
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -36,6 +36,12 @@ SFU = 'sfu'  # Windows Services for UNIX
 
 def command_env_wrapper(conanfile, command, envfiles, envfiles_folder, scope="build"):
     from conan.tools.env.environment import environment_wrap_command
+    if getattr(conanfile, "conf", None) is None:
+        # TODO: No conf, no profile defined!! This happens at ``export()`` time
+        #  Is it possible to run a self.run() in export() in bash?
+        #  Is it necessary? Shouldn't be
+        return command
+
     active = conanfile.conf.get("tools.microsoft.bash:active", check_type=bool)
     subsystem = conanfile.conf.get("tools.microsoft.bash:subsystem")
 
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -34,10 +34,21 @@ WSL = 'wsl'  # Windows Subsystem for Linux
 SFU = 'sfu'  # Windows Services for UNIX
 
 
-def command_env_wrapper(conanfile, command, envfiles, envfiles_folder):
+def command_env_wrapper(conanfile, command, envfiles, envfiles_folder, scope="build"):
     from conan.tools.env.environment import environment_wrap_command
-    if platform.system() == "Windows" and conanfile.win_bash:  # New, Conan 2.0
-        wrapped_cmd = _windows_bash_wrapper(conanfile, command, envfiles, envfiles_folder)
+    active = conanfile.conf.get("tools.microsoft.bash:active", check_type=bool)
+    subsystem = conanfile.conf.get("tools.microsoft.bash:subsystem")
+
+    if platform.system() == "Windows" and (
+            (conanfile.win_bash and scope == "build") or
+            (conanfile.win_bash_run and scope == "run")):
+        if subsystem is None:
+            raise ConanException("win_bash/win_bash_run defined but no "
+                                 "tools.microsoft.bash:subsystem")
+        if active:
+            wrapped_cmd = environment_wrap_command(envfiles, envfiles_folder, command)
+        else:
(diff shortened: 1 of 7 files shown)


conans/client/subsystems.py
```diff
@@ -50,9 +67,8 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
 
     subsystem = conanfile.conf.get("tools.microsoft.bash:subsystem")
     shell_path = conanfile.conf.get("tools.microsoft.bash:path")
-    if not subsystem or not shell_path:
-        raise ConanException("The config 'tools.microsoft.bash:subsystem' and "
-                             "'tools.microsoft.bash:path' are "
+    if not shell_path:
+        raise ConanException("The vconfig 'tools.microsoft.bash:path' is "
```
czoido:
```suggestion
        raise ConanException("The config 'tools.microsoft.bash:path' is "
```

czoido (approved):

uilianries (approved):

franramirez688 (approved):

jcar87 (approved):

schwaerz:
I am very confused about how `VirtualRunEnv` is supposed to work.

When using conan 1.60.1 and `VirtualRunEnv`, running Conan from within my git bash (in the **test package's** `test` method), I will always run into:
```
    '.' is not recognized as an internal or external command,
    operable program or batch file.
```

I think to remember that I got it running once for one conanfile by passing the following parameters when building on Windows....
```
--conf:host tools.microsoft.bash:active=True --conf:host tools.microsoft.bash:subsystem=msys2
```
however, it does not seem to work now. Most probably I did something on top which I cannot remember... Still getting the same error. Conan still tries to run `conanrun.sh` using Windows cmd.

Please note: I do not use any profile for the build. Sticking to the default profile - which does not contain any subsystem setting:
```
$ conan profile show default
Configuration for profile default:

[settings]
os=Windows
os_build=Windows
arch=x86
arch_build=x86_64
compiler=Visual Studio
compiler.version=15
build_type=Release
[options]
libdwarf:shared=True
libelf:shared=True
qt:shared=True
xerces-c:shared=True
[conf]
[build_requires]
[env]
```

The only stable workaround I found is the following:

```python
        if tools.detected_os() == "Windows":
            self.run("bash -c \". conanrun.sh && <cmd to run>\"")
        else:
            self.run("<cmd to run>", env="conanrun")
```

Obviously, I'd expect that I can use the same code on Windows and Linux. Also, I'd like to be able to use the same bash script on Windows and Linux. This used to be possible using 'virtualrunenv' (while this one did not set any custom environment variables), but is no longer possible with 'VirtualRunEnv' or 'VirtualBuildEnv'.

As the ticktes I found so far on this topic seem to be all closed: Is there still any improvement planned? Is this possibly one available with Conan 2.x and not yet merged back to 1.x?

Should I open a new ticket for this?

PR #12194, opened 2022-09-24
AutotoolsToolchain: curate paths coming from `tools.build:compiler_executables` on Windows before defining `CC`, `CXX` etc

SpaceIm:
Changelog: (Fix): In `AutotoolsToolchain`, if build machine is windows, curate paths coming from `tools.build:compiler_executables` before defining `CC`, `CXX`, `CCAS`, `NVCC` & `FC` env vars
Docs: https://github.com/conan-io/docs/pull/2845

Scope of this PR has been simplified: https://github.com/conan-io/conan/pull/12194#issuecomment-1334528040

==========
Old logic description here but not valid anymore:

This PR tries to honor `CC`/`CXX`/`LD`/`AR`/`NM`/`OBJDUMP`/`RANLIB`/`STRIP` if they are set from profile -specially if a path was given-, fall back to good defaults for msvc if they are not set in profile, and convert them to a value compatible with autotools scripts & current subsysem (specially on Windows):

- Users may set a full path to their compiler tools in these variables (specially when they have several compiler versions), and on Windows they likely set in their profile a "windows like" full path, so this path is converted to a path suitable for the subsystem used for autotools build.
- If one of these variables is set with a path, but this path contains spaces, it falls back to a default value, because scripts generated by autotools can't handle spaces in paths of these variables (even if they are protected, path is quoted etc). A warning is displayed in this case.
  As a workaround on Windows, users may use short file names to refer to their compiler path if the long name contains spaces (`PROGRA~1` instead of `Program Files` for example). I guess it might be possible for conan, if build machine is windows, to find the short file name (if it exists) of directories containing spaces, and do this conversion on the fly, but it's not implemented here (`win32api.GetShortPathName` from `pywin32`?).
- If one of these variables is set but it's not a path (quite common also, like `CC=cl`), nothing special is done (except adding extra options like `-nologo` or eventually wrap with `compile_wrapper` & `ar_wrapper` as explained below).
- If one of these variables is not set, it falls back to a good default if msvc, and nothing special is done for other compilers.
- `-nologo` is added to `CC`, `CXX` & `LD` vars if msvc to avoid noisy messages during build.

All these values are stored in new `AutotoolsToolchain` attributes: `cc`, `cxx`, `ld`, `ar`, `nm`, `objdump`, `ranlib`, `strip`. If needed, these attributes can be overridden before calling `generate()` (where these values are injected in conanautotoolstoolchain script injected in build scope if they are different than the ones of `VirtualBuildEnv`).

This PR also adds two new arguments to `AutotoolsToolchain` constructor: `compile_wrapper` & `ar_wrapper`. They are None by default, but a recipe can fill them with path to a wrapper script like `compile` or `ar-lib` from `automake` (or sometimes source code of the library) to provide robust autotools compatibility to compilers like msvc.
If set, `compile_wrapper` wraps compiler calls (`CC` & `CXX`).
If set, `ar_wrapper` wraps archiver calls (`AR`).

A typical usage in an autotools based recipe would be:

```python
    @property
    def _settings_build(self):
        return getattr(self, "settings_build", self.settings)

    @property
    def _user_info_build(self):
        return getattr(self, "user_info_build", self.deps_user_info)

    def build_requirements(self):
        if self._settings_build.os == "Windows":
            if is_msvc(self):
                # to get compile & ar-lib wrappers scripts if not provided in lib source code
                self.tool_requires("automake/1.16.5")
            if not self.conf.get("tools.microsoft.bash:path", default=False, check_type=bool):
                # to get a bash
                self.tool_requires("msys2/cci.latest")
            self.win_bash = True

    def generate(self):
        tc = AutotoolsToolchain(
            self,
            compile_wrapper = self._user_info_build["automake"].compile if is_msvc(self) else None,
            ar_wrapper = self._user_info_build["automake"].ar_lib if is_msvc(self) else None,
        )
        tc.generate()
```

Since `AutotoolsToolchain` may override CC, CXX etc, recipes using it must not call `VirtualBuildEnv` after `AutotoolsToolchain`, always before if they have to call it.

=======


conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -123,8 +123,9 @@ def defines(self):
     def environment(self):
         env = Environment()
         if is_msvc(self._conanfile):
-            env.define("CXX", "cl")
-            env.define("CC", "cl")
+            env.define("CXX", "cl -nologo")
+            env.define("CC", "cl -nologo")
+            env.define("LD", "link -nologo")
```
SpaceIm, line 128: Actually, I'm wondering whether conan should not first check whether CC, CXX and LD are set in profile (for example a user may have several Visual Studio installation, and point to a specific one).
For each of them, if it's set, transform the path to unix_path and add `-nologo`, if not set to `cl -nologo` etc.

SpaceIm:
@madebr @ericLemanissier if you have some time, I would appreciate feedbacks.

SpaceIm:
I'm wondering whether AutotoolsToolchain could expose arguments to conditionally prefix these env vars with custom values in a recipe, like it's done in https://github.com/conan-io/conan-center-index/blob/1d04372ae1bead2e86a3f7183561361e9fcd9761/recipes/coin-cgl/all/conanfile.py#L90-L93

SpaceIm:
Why not providing these scripts in conan (but GPL-2.0-or-later license)?

<details><summary>compile wrapper</summary>

```
#! /bin/sh
# Wrapper for compilers which do not understand '-c -o'.

scriptversion=2018-03-07.03; # UTC

# Copyright (C) 1999-2021 Free Software Foundation, Inc.
# Written by Tom Tromey <tromey@cygnus.com>.
#
# This program is free software; you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation; either version 2, or (at your option)
# any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program.  If not, see <https://www.gnu.org/licenses/>.

# As a special exception to the GNU General Public License, if you
# distribute this file as part of a program that contains a
# configuration script generated by Autoconf, you may include it under
# the same distribution terms that you use for the rest of that program.

# This file is maintained in Automake, please report
# bugs to <bug-automake@gnu.org> or send patches to
# <automake-patches@gnu.org>.

nl='
'

# We need space, tab and new line, in precisely that order.  Quoting is
# there to prevent tools from complaining about whitespace usage.
IFS=" ""	$nl"

file_conv=

# func_file_conv build_file lazy
# Convert a $build file to $host form and store it in $file
# Currently only supports Windows hosts. If the determined conversion
# type is listed in (the comma separated) LAZY, no conversion will
# take place.
func_file_conv ()
{
  file=$1
  case $file in
    / | /[!/]*) # absolute file, and not a UNC file
      if test -z "$file_conv"; then
	# lazily determine how to convert abs files
	case `uname -s` in
	  MINGW*)
	    file_conv=mingw
	    ;;
	  CYGWIN* | MSYS*)
	    file_conv=cygwin
	    ;;
	  *)
	    file_conv=wine
	    ;;
	esac
      fi
      case $file_conv/,$2, in
	*,$file_conv,*)
	  ;;
	mingw/*)
	  file=`cmd //C echo "$file " | sed -e 's/"\(.*\) " *$/\1/'`
	  ;;
	cygwin/* | msys/*)
	  file=`cygpath -m "$file" || echo "$file"`
	  ;;
	wine/*)
	  file=`winepath -w "$file" || echo "$file"`
	  ;;
      esac
      ;;
  esac
}

# func_cl_dashL linkdir
# Make cl look for libraries in LINKDIR
func_cl_dashL ()
{
  func_file_conv "$1"
  if test -z "$lib_path"; then
    lib_path=$file
  else
    lib_path="$lib_path;$file"
  fi
  linker_opts="$linker_opts -LIBPATH:$file"
}

# func_cl_dashl library
# Do a library search-path lookup for cl
func_cl_dashl ()
{
  lib=$1
  found=no
  save_IFS=$IFS
  IFS=';'
  for dir in $lib_path $LIB
  do
    IFS=$save_IFS
    if $shared && test -f "$dir/$lib.dll.lib"; then
      found=yes
      lib=$dir/$lib.dll.lib
      break
    fi
    if test -f "$dir/$lib.lib"; then
      found=yes
      lib=$dir/$lib.lib
      break
    fi
    if test -f "$dir/lib$lib.a"; then
      found=yes
      lib=$dir/lib$lib.a
      break
    fi
  done
  IFS=$save_IFS

  if test "$found" != yes; then
    lib=$lib.lib
  fi
}

# func_cl_wrapper cl arg...
# Adjust compile command to suit cl
func_cl_wrapper ()
{
  # Assume a capable shell
  lib_path=
  shared=:
  linker_opts=
  for arg
  do
    if test -n "$eat"; then
      eat=
    else
      case $1 in
	-o)
	  # configure might choose to run compile as 'compile cc -o foo foo.c'.
	  eat=1
	  case $2 in
	    *.o | *.[oO][bB][jJ])
	      func_file_conv "$2"
	      set x "$@" -Fo"$file"
	      shift
	      ;;
	    *)
	      func_file_conv "$2"
	      set x "$@" -Fe"$file"
	      shift
	      ;;
	  esac
	  ;;
	-I)
	  eat=1
	  func_file_conv "$2" mingw
	  set x "$@" -I"$file"
	  shift
	  ;;
	-I*)
	  func_file_conv "${1#-I}" mingw
	  set x "$@" -I"$file"
	  shift
	  ;;
	-l)
	  eat=1
	  func_cl_dashl "$2"
	  set x "$@" "$lib"
	  shift
	  ;;
	-l*)
	  func_cl_dashl "${1#-l}"
	  set x "$@" "$lib"
	  shift
	  ;;
	-L)
	  eat=1
	  func_cl_dashL "$2"
	  ;;
	-L*)
	  func_cl_dashL "${1#-L}"
	  ;;
	-static)
	  shared=false
	  ;;
	-Wl,*)
	  arg=${1#-Wl,}
	  save_ifs="$IFS"; IFS=','
	  for flag in $arg; do
	    IFS="$save_ifs"
	    linker_opts="$linker_opts $flag"
	  done
	  IFS="$save_ifs"
	  ;;
	-Xlinker)
	  eat=1
	  linker_opts="$linker_opts $2"
	  ;;
	-*)
	  set x "$@" "$1"
	  shift
	  ;;
	*.cc | *.CC | *.cxx | *.CXX | *.[cC]++)
	  func_file_conv "$1"
	  set x "$@" -Tp"$file"
	  shift
	  ;;
	*.c | *.cpp | *.CPP | *.lib | *.LIB | *.Lib | *.OBJ | *.obj | *.[oO])
	  func_file_conv "$1" mingw
	  set x "$@" "$file"
	  shift
	  ;;
	*)
	  set x "$@" "$1"
	  shift
	  ;;
      esac
    fi
    shift
  done
  if test -n "$linker_opts"; then
    linker_opts="-link$linker_opts"
  fi
  exec "$@" $linker_opts
  exit 1
}

eat=

case $1 in
  '')
     echo "$0: No command.  Try '$0 --help' for more information." 1>&2
     exit 1;
     ;;
  -h | --h*)
    cat <<\EOF
Usage: compile [--help] [--version] PROGRAM [ARGS]

Wrapper for compilers which do not understand '-c -o'.
Remove '-o dest.o' from ARGS, run PROGRAM with the remaining
arguments, and rename the output as expected.

If you are trying to build a whole package this is not the
right script to run: please start by reading the file 'INSTALL'.

Report bugs to <bug-automake@gnu.org>.
EOF
    exit $?
    ;;
  -v | --v*)
    echo "compile $scriptversion"
    exit $?
    ;;
  cl | *[/\\]cl | cl.exe | *[/\\]cl.exe | \
  icl | *[/\\]icl | icl.exe | *[/\\]icl.exe )
    func_cl_wrapper "$@"      # Doesn't return...
    ;;
esac

ofile=
cfile=

for arg
do
  if test -n "$eat"; then
    eat=
  else
    case $1 in
      -o)
	# configure might choose to run compile as 'compile cc -o foo foo.c'.
	# So we strip '-o arg' only if arg is an object.
	eat=1
	case $2 in
	  *.o | *.obj)
	    ofile=$2
	    ;;
	  *)
	    set x "$@" -o "$2"
	    shift
	    ;;
	esac
	;;
      *.c)
	cfile=$1
	set x "$@" "$1"
	shift
	;;
      *)
	set x "$@" "$1"
	shift
	;;
    esac
  fi
  shift
done

if test -z "$ofile" || test -z "$cfile"; then
  # If no '-o' option was seen then we might have been invoked from a
  # pattern rule where we don't need one.  That is ok -- this is a
  # normal compilation that the losing compiler can handle.  If no
  # '.c' file was seen then we are probably linking.  That is also
  # ok.
  exec "$@"
fi

# Name of file we expect compiler to create.
cofile=`echo "$cfile" | sed 's|^.*[\\/]||; s|^[a-zA-Z]:||; s/\.c$/.o/'`

# Create the lock directory.
# Note: use '[/\\:.-]' here to ensure that we don't use the same name
# that we are using for the .o file.  Also, base the name on the expected
# object file name, since that is what matters with a parallel build.
lockdir=`echo "$cofile" | sed -e 's|[/\\:.-]|_|g'`.d
while true; do
  if mkdir "$lockdir" >/dev/null 2>&1; then
    break
  fi
  sleep 1
done
# FIXME: race condition here if user kills between mkdir and trap.
trap "rmdir '$lockdir'; exit 1" 1 2 15

# Run the compile.
"$@"
ret=$?

if test -f "$cofile"; then
  test "$cofile" = "$ofile" || mv "$cofile" "$ofile"
elif test -f "${cofile}bj"; then
  test "${cofile}bj" = "$ofile" || mv "${cofile}bj" "$ofile"
fi

rmdir "$lockdir"
exit $ret

# Local Variables:
# mode: shell-script
# sh-indentation: 2
# eval: (add-hook 'before-save-hook 'time-stamp)
# time-stamp-start: "scriptversion="
# time-stamp-format: "%:y-%02m-%02d.%02H"
# time-stamp-time-zone: "UTC0"
# time-stamp-end: "; # UTC"
# End:

```

</details>

<details><summary>ar-lib wrapper</summary>

```
#! /bin/sh
# Wrapper for Microsoft lib.exe

me=ar-lib
scriptversion=2019-07-04.01; # UTC

# Copyright (C) 2010-2021 Free Software Foundation, Inc.
# Written by Peter Rosin <peda@lysator.liu.se>.
#
# This program is free software; you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation; either version 2, or (at your option)
# any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program.  If not, see <https://www.gnu.org/licenses/>.

# As a special exception to the GNU General Public License, if you
# distribute this file as part of a program that contains a
# configuration script generated by Autoconf, you may include it under
# the same distribution terms that you use for the rest of that program.

# This file is maintained in Automake, please report
# bugs to <bug-automake@gnu.org> or send patches to
# <automake-patches@gnu.org>.


# func_error message
func_error ()
{
  echo "$me: $1" 1>&2
  exit 1
}

file_conv=

# func_file_conv build_file
# Convert a $build file to $host form and store it in $file
# Currently only supports Windows hosts.
func_file_conv ()
{
  file=$1
  case $file in
    / | /[!/]*) # absolute file, and not a UNC file
      if test -z "$file_conv"; then
	# lazily determine how to convert abs files
	case `uname -s` in
	  MINGW*)
	    file_conv=mingw
	    ;;
	  CYGWIN* | MSYS*)
	    file_conv=cygwin
	    ;;
	  *)
	    file_conv=wine
	    ;;
	esac
      fi
      case $file_conv in
	mingw)
	  file=`cmd //C echo "$file " | sed -e 's/"\(.*\) " *$/\1/'`
	  ;;
	cygwin | msys)
	  file=`cygpath -m "$file" || echo "$file"`
	  ;;
	wine)
	  file=`winepath -w "$file" || echo "$file"`
	  ;;
      esac
      ;;
  esac
}

# func_at_file at_file operation archive
# Iterate over all members in AT_FILE performing OPERATION on ARCHIVE
# for each of them.
# When interpreting the content of the @FILE, do NOT use func_file_conv,
# since the user would need to supply preconverted file names to
# binutils ar, at least for MinGW.
func_at_file ()
{
  operation=$2
  archive=$3
  at_file_contents=`cat "$1"`
  eval set x "$at_file_contents"
  shift

  for member
  do
    $AR -NOLOGO $operation:"$member" "$archive" || exit $?
  done
}

case $1 in
  '')
     func_error "no command.  Try '$0 --help' for more information."
     ;;
  -h | --h*)
    cat <<EOF
Usage: $me [--help] [--version] PROGRAM ACTION ARCHIVE [MEMBER...]

Members may be specified in a file named with @FILE.
EOF
    exit $?
    ;;
  -v | --v*)
    echo "$me, version $scriptversion"
    exit $?
    ;;
esac

if test $# -lt 3; then
  func_error "you must specify a program, an action and an archive"
fi

AR=$1
shift
while :
do
  if test $# -lt 2; then
    func_error "you must specify a program, an action and an archive"
  fi
  case $1 in
    -lib | -LIB \
    | -ltcg | -LTCG \
    | -machine* | -MACHINE* \
    | -subsystem* | -SUBSYSTEM* \
    | -verbose | -VERBOSE \
    | -wx* | -WX* )
      AR="$AR $1"
      shift
      ;;
    *)
      action=$1
      shift
      break
      ;;
  esac
done
orig_archive=$1
shift
func_file_conv "$orig_archive"
archive=$file

# strip leading dash in $action
action=${action#-}

delete=
extract=
list=
quick=
replace=
index=
create=

while test -n "$action"
do
  case $action in
    d*) delete=yes  ;;
    x*) extract=yes ;;
    t*) list=yes    ;;
    q*) quick=yes   ;;
    r*) replace=yes ;;
    s*) index=yes   ;;
    S*)             ;; # the index is always updated implicitly
    c*) create=yes  ;;
    u*)             ;; # TODO: don't ignore the update modifier
    v*)             ;; # TODO: don't ignore the verbose modifier
    *)
      func_error "unknown action specified"
      ;;
  esac
  action=${action#?}
done

case $delete$extract$list$quick$replace,$index in
  yes,* | ,yes)
    ;;
  yesyes*)
    func_error "more than one action specified"
    ;;
  *)
    func_error "no action specified"
    ;;
esac

if test -n "$delete"; then
  if test ! -f "$orig_archive"; then
    func_error "archive not found"
  fi
  for member
  do
    case $1 in
      @*)
        func_at_file "${1#@}" -REMOVE "$archive"
        ;;
      *)
        func_file_conv "$1"
        $AR -NOLOGO -REMOVE:"$file" "$archive" || exit $?
        ;;
    esac
  done

elif test -n "$extract"; then
  if test ! -f "$orig_archive"; then
    func_error "archive not found"
  fi
  if test $# -gt 0; then
    for member
    do
      case $1 in
        @*)
          func_at_file "${1#@}" -EXTRACT "$archive"
          ;;
        *)
          func_file_conv "$1"
          $AR -NOLOGO -EXTRACT:"$file" "$archive" || exit $?
          ;;
      esac
    done
  else
    $AR -NOLOGO -LIST "$archive" | tr -d '\r' | sed -e 's/\\/\\\\/g' \
      | while read member
        do
          $AR -NOLOGO -EXTRACT:"$member" "$archive" || exit $?
        done
  fi

elif test -n "$quick$replace"; then
  if test ! -f "$orig_archive"; then
    if test -z "$create"; then
      echo "$me: creating $orig_archive"
    fi
    orig_archive=
  else
    orig_archive=$archive
  fi

  for member
  do
    case $1 in
    @*)
      func_file_conv "${1#@}"
      set x "$@" "@$file"
      ;;
    *)
      func_file_conv "$1"
      set x "$@" "$file"
      ;;
    esac
    shift
    shift
  done

  if test -n "$orig_archive"; then
    $AR -NOLOGO -OUT:"$archive" "$orig_archive" "$@" || exit $?
  else
    $AR -NOLOGO -OUT:"$archive" "$@" || exit $?
  fi

elif test -n "$list"; then
  if test ! -f "$orig_archive"; then
    func_error "archive not found"
  fi
  $AR -NOLOGO -LIST "$archive" || exit $?
fi

```

</details>

madebr:
I would do this another way and add `cc/cxx/cpp/ld/ar/ranlib/nm` attributes to the `AutotoolsToolchain` class.
When constructing a `AutotoolsToolchain` object, assign values to them.

This approach allows a recipe to override any variable without conan trying to be too smart (e.g. unwanted conversion of a path to unix path).

So allow this code:
```python
def generate(self):
    tc = AutotoolsToolchain(self,
        compile_wrapper=self._user_info_build["automake"].compile if is_msvc(self) else None,
        ar_wrapper=self._user_info_build["automake"].ar_lib if is_msvc(self) else None,
    )
    assert tc.cc == "/path/to/compile cl"
    assert tc.cxx == "/path/to/compile cl"
    assert tc.ar == "/path/to/ar-lib lib"
    tc.cc = "my_awesome_compiler"
    assert tc.cc == "my_awesome_compiler"
    assert tc.cxx == "/path/to/compile cl"
    tc.generate()
```

SpaceIm:
The main problem I see with this approach is to delegate to recipes the responsibility to honor CC, CXX etc set as absolute paths in profiles, it might mean lot of boilerplate in each recipe while it would be better to delegate all of this to conan. If conan or recipe overrides CC with just cl while users have set this env var as an absolute path in their profile, it might not compile with expected msvc version.

madebr:
If set, I would let the constructor initialize `cc` with the value of `CC` of the build host profile (after going through `tools.unix_path`).
This behavior also applies for a clang profile on any platform.

SpaceIm:
But you lose compile or ar-lib wrapper injection, it shouldn't be handled by final users, it's an internal build detail of the recipe.

SpaceIm:
Oh, I think I see what you mean, you just want to populate all these attributes during initialialization, expose them so that they may eventually be overridden by a recipe before tc.generate(), and after that they are injected in build scope during tc.generate()?

Just want to say that all these variables would be just None when related env var is not set in profile (except for msvc).

SpaceIm:
something like https://github.com/conan-io/conan/pull/12194/commits/fc3c72c6abda1da24951483bf251cf444b28d237?

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -43,6 +44,20 @@ def __init__(self, conanfile, namespace=None):
...
+        self.ld = self._ld
+        self.ar = self._ar
+        self.nm = self._nm
+        self.objdump = self._objdump
+        self.ranlib = self._ranlib
+        self.strip = self._strip
+
+        # Wrappers paths for tools not compatible with autotools, like msvc
+        self.compile_wrapper = None
+        self.arlib_wrapper = None
```
madebr: You need to move these forwards, before `self.cc` etc are assigned.

madebr:
> something like [fc3c72c](https://github.com/conan-io/conan/commit/fc3c72c6abda1da24951483bf251cf444b28d237)?

Yes, that's what I meant.
I think `AutotoolsEnvironment` is the correct location for this because other build systems don't need this behavior.

SpaceIm:
> I think `AutotoolsEnvironment` is the correct location for this because other build systems don't need this behavior.

There is no such class. You mean `AutotoolsToolchain`?

conan/tools/gnu/autotoolstoolchain.py
SpaceIm: moved to constructor arguments: https://github.com/conan-io/conan/commit/40fdb2000fd727a4f2f1309c3fb848910b2cb0d8

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -120,11 +133,97 @@ def defines(self):
...
+            ":" if is_msvc(self._conanfile) else None,
+        )
+
+    def _get_strip(self):
+        return self._exe_env_var_to_unix_path(
+            "STRIP",
+            ":" if is_msvc(self._conanfile) else None,
+        )
+
+    def _exe_env_var_to_unix_path(self, env_var, default=None, extra_options=[], wrapper=None):
```
madebr: Using a list as default keyword value in Python is dangerous.
The list is not modified in the body, but might be in later iterations.
It's safer to set the default value to `None` and then set the parameter to an empty list in the body.

Many Python linters warn about this.

madebr:
> > I think `AutotoolsEnvironment` is the correct location for this because other build systems don't need this behavior.
> 
> There is no such class. You mean `AutotoolsToolchain`?

Yes, I haven't used the conan 2.0 classes much yet, so I don't know the new terminology that good yet.

conan/tools/gnu/autotoolstoolchain.py
SpaceIm: https://github.com/conan-io/conan/pull/12194/commits/f03d98dbdf0e399f0555416f5de7eece3d52fcd4

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -120,11 +133,99 @@ def defines(self):
...
+            "STRIP",
+            ":" if is_msvc(self._conanfile) else None,
+        )
+
+    def _exe_env_var_to_unix_path(self, env_var, default=None, extra_options=None, wrapper=None):
+        """
+            Convenient method to convert env vars like CC, CXX or LD to values compatible with autotools.
+            If env var doesn't exist, returns default.
+        """
+        exe = VirtualBuildEnv(self._conanfile).vars().get(env_var)
```
SpaceIm: I've tested and it's always `None`, regardless of env var I put under `[env]` in my profile (tested with 1 & 2 profiles).

I've tried with `VirtualBuildEnv` and `conans.tools.get_env`. What's wrong? I see that `MesonToolchain` also tries to get `CC`, `CXX` etc from profile with `VirtualBuildEnv`, does it work actually?

madebr: I don't have much experience with conan 2.0 (yet), but did you use a `[buildenv]` section in your host profile? See [this link](https://docs.conan.io/en/2.0/tutorial/consuming_packages/cross_building_with_conan.html?highlight=profile#conan-two-profiles-model-build-and-host-profiles).

SpaceIm: Indeed thanks, it honors env from profile now.

But now I see two issues:

- if `VirtualBuildEnv` is called after `AutotoolsToolchain` in `generate()` of a recipe (I'm testing with https://github.com/conan-io/conan-center-index/pull/13135), it overrides all the CC etc logic of `AutotoolsToolchain`, so it's very important to call `VirtualBuildEnv` before any other generator also relying on `Environment` to set env vars.
- if there is a space in these paths (which is likely the case for cl, link in Program Files), even `unix_path` doesn't help (nor protecting spaces with `\`, or protecting full path with quotes, and I've tested with and without env vars wrapped by compile & ar-lib, errors either happen in those wrappers or later on in configure).
  For example for `CC`, `C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.33.31629\bin\Hostx64\x64\cl.exe` in profile is converted to `/c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/program files/microsoft visual studio/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo` in https://github.com/conan-io/conan-center-index/pull/13135.

SpaceIm: For the second point, the best I can do is display a warning and suggest to set CC etc with short file names.

For example on my computer I can set in profile:

```ini
[buildenv]
CC=C:\PROGRA~1\MIB055~1\2022\Community\VC\Tools\MSVC\14.33.31629\bin\Hostx64\x64\cl.exe
```

then it works and full path is honored

```
checking for gcc... /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo
checking whether the C compiler works... yes
checking for C compiler default output file name... conftest.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... obj
checking whether the compiler supports GNU C... no
checking whether /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo accepts -g... no
checking for /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo option to enable C11 features... unsupported
checking for /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo option to enable C99 features... unsupported
checking for /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo option to enable C89 features... unsupported
checking whether /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo understands -c and -o together... yes
checking whether the compiler is clang... no
checking for compiler option needed when checking for declarations... none
checking whether make supports the include directive... yes (GNU style)
checking dependency style of /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo... msvc7
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... no
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... no
checking for wchar.h... yes
checking for minix/config.h... no
checking for pthread.h... no
checking for sys/param.h... no
checking for sys/socket.h... no
checking for dirent.h... no
checking for stdio_ext.h... no
checking for getopt.h... no
checking for sys/cdefs.h... no
checking for threads.h... no
checking for iconv.h... no
checking for limits.h... yes
checking for crtdefs.h... yes
checking for wctype.h... yes
checking for langinfo.h... no
checking for xlocale.h... no
checking for math.h... yes
checking for sys/mman.h... no
checking for malloc.h... yes
checking for spawn.h... no
checking for sys/time.h... no
checking for sys/random.h... no
checking for sys/wait.h... no
checking for features.h... no
checking for arpa/inet.h... no
checking for semaphore.h... no
checking for netdb.h... no
checking for netinet/in.h... no
checking for sys/select.h... no
checking for sys/ioctl.h... no
checking for sys/uio.h... no
checking whether it is safe to define __EXTENSIONS__... yes
checking whether _XOPEN_SOURCE should be defined... no
checking build system type... x86_64-pc-mingw64
checking host system type... x86_64-pc-mingw64
checking how to run the C preprocessor... /c/users/spaceim/.conan/data/m4/1.4.19/_/_/build/0a420ff5c47119e668867cdb51baff0eca1fdb68/src/build-aux/compile /c/progra~1/mib055~1/2022/community/vc/tools/msvc/14.33.31629/bin/hostx64/x64/cl.exe -nologo -E
```

SpaceIm:
```
self = <[AttributeError("'ConanFileMock' object has no attribute 'display_name'") raised in repr()] ConanFileMock object at 0x1067fc520>

    @property
    def buildenv(self):
        # Lazy computation of the package buildenv based on the profileone
>       if not isinstance(self._conan_buildenv, Environment):
E       AttributeError: 'ConanFileMock' object has no attribute '_conan_buildenv'

/Users/charpentiera/Documents/Personnel/projects/conan/conans/model/conan_file.py:206: AttributeError
```

There are these errors since https://github.com/conan-io/conan/pull/12194/commits/6a415fd41fc08279b929a6bf2498eddfc376ca89, so due to `VirtualBuildEnv` invocation, but I don't know why. Seems to come from usage of a legacy `MockConanfile` in these conan v2 tests.

SpaceIm:
There is this logic currently:

```python
    def generate(self, env=None, scope="build"):
        env = env or self.environment()
        env = env.vars(self._conanfile, scope=scope)
        env.save_script("conanautotoolstoolchain")
        self.generate_args()
```

It's fragile. If someone calls VirtualBuildEnv.generate() after AutotoolsToolchain.generate(), it could defeat the logic of this PR I guess. How to ensure that `conanautotoolstoolchain` script is always called last in this build scope?

Moreover, what happen if someone mix-up something like AutotoolsToolchain & CMakeToolchain in build scope because it has to call autotools then cmake in build()?

@memsharded Wouldn't it be better to append conanautotoolstoolchain & vcvars env scripts in a custom scope, and this custom scope would also call all scripts of build scope first? Then Autotools build helper would always call self.run with this custom scope instead of build scope. It allows a better encapsulation, and to avoid side effects.

conan/tools/microsoft/win32api/win32api.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+if os.name == "nt":
+    import ctypes
+    from ctypes import wintypes
+
+
+    def get_short_path_name(path):
```
memsharded, line 8: This needs explanation why it is needed, I am not sure how this is related to improving the AutotoolsToolchain

memsharded (commented):
All the main idea of env-vars is that they are not necessary to handle them, they go from environment to tools.

> This PR tries to honor CC/CXX/LD/AR/NM/OBJDUMP/RANLIB/STRIP if they are set from profile -specially if a path was given-, fall back to good defaults for msvc if they are not set in profile, and convert them to a value compatible with autotools scripts & current subsysem (specially on Windows):

It would be much better if we define the things as real use cases, like: "I want to compile with autotools in Windows, with MinGW compiler". Then we can define the functional tests to make sure things work and compile, and starting from the tests we can look for the best approaches. If it means managing some env-vars, we can think about it. But maybe it is defining ``conf``, we don't know, and need to understand better the full user story.

conan/tools/microsoft/win32api/win32api.py
SpaceIm: It's explained in https://github.com/conan-io/conan/pull/12194#issue-1384813935:

> If one of these variables is set with a path, but this path contains spaces, it falls back to a default value, because scripts generated by autotools can't handle spaces in paths of these variables (even if they are protected, path is quoted etc). A warning is displayed in this case.
> 
> As a workaround on Windows, users may use short file names to refer to their compiler path if the long name contains spaces (PROGRA~1 instead of Program Files for example). **I guess it might be possible for conan, if build machine is windows, to find the short file name (if it exists) of directories containing spaces, and do this conversion on the fly, but it's not implemented here (win32api.GetShortPathName from pywin32?).**

memsharded: It is too complicated, difficult to maintain. I prefer to limit the functionality to paths without spaces, honestly, we cannot add this complexity to satisfy every corner case in the world. If autotools does not support paths with spaces, it does not support paths with spaces, so things should go in paths without spaces.

SpaceIm: But why not implementing this workaround if it works? It's not so complicated and allows to support absolute path to Visual Studio. For me underlying build system is an internal detail.
It shouldn't fail to build if someone specificy some usual profile (CC set to a specific compiler, CC not set so it takes default compiler).
It should honor CC, CXX, LD, etc from profile (overridding CC to just `cl` is the very last solution because you may not honor compiler version from profile when there are several Visual Studio installations).

SpaceIm:
You mean define `tools.build:cc`, `tools.build:cxx`, `tools.build:ld` & `tools.build:ar` conf, then inject their value in each conan toolchain?
And assume that users don't define these values in `[env]` (or `[buildenv]`) section anymore (it was the canonical way to disambiguate compilers in conan v1), but in `conf`?

But I'm surprised because CMakeToolchain, MesonToolchain & QbsToolchain have already very strong assumption about CC, CXX env vars, so why this PR would be different?

SpaceIm:
@memsharded I'll split this PR in two PRs and open issues before:

- a first one with a proposal for new `conf` parameters: something like `tools.c:compiler`, `tools.cxx:compiler`, `tools:archiver`, `tools:linker`, `tools.python:interpreter`, `tools.fortran:compiler`, `tools.perl:interpreter` etc
- a second one with a proposal to handle msvc in a robust way in AutotoolsToolchain, it might be either a companion generator or some integration in AutotoolsToolchain. Il will also honor conf like `tools.c:compiler` etc in a robust way.

SpaceIm:
@memsharded I've refactored this PR to take into account new `tools.build:compiler_executables`, and only focus on curation of paths coming from this conf if build machine is Windows (it's too tedious to delegate this task to recipes).

So an autotools based recipe could look like:

```python
def generate(self):
    tc = AutotoolsToolchain(self)
    env = tc.environment()
    if is_msvc(self):
        compile_wrapper = unix_path(self, self._user_info_build["automake"].compile)
        ar_wrapper = unix_path(self, self._user_info_build["automake"].ar_lib)
        env_vars = env.vars(self)
        env.define("CC", f"{compile_wrapper} {env_vars.get('CC', 'cl')} -nologo")
        env.define("CXX", f"{compile_wrapper} {env_vars.get('CXX', 'cl')} -nologo")
        env.define("LD", "link -nologo")
        env.define("AR", f"{ar_wrapper} \"lib -nologo\"")
        env.define("NM", "dumpbin -symbols")
        env.define("OBJDUMP", ":")
        env.define("RANLIB", ":")
        env.define("STRIP", ":")
    tc.generate(env)
```

Therefore it would be able to honor `c` & `cpp` values of `tools.build:compiler_executables` if msvc, instead of hardcoding `cl` in `CC` & `CXX`.
For MinGW it would work out of the box even if consumer has set `tools.build:compiler_executables` with paths containing back slashes or spaces, so it provides a robust & consistent behavior to consumers who don't have to be concerned by internal details of recipes (for example CMakeToolchain also curates these paths by replacing back slashes by forward slashes in `CMAKE_<LANG>_COMPILER`).

SpaceIm:
@memsharded any chance to consider this PR? I don't know what to do, should I close it?

memsharded:
The main use case for this would be the ``msvc`` compiler in "Program Files" folders, as explained above.
But this would be dangerous as ``AutotoolsToolchain`` for ``msvc`` compiler is already creating and activating a ``vcvars`` environment. I am not sure that defining a different compiler location other than the one currently active in the current msvc environment is safe, but doesn't look so.

Together with the complexity of having to deal with win32 specific APIs, I am afraid that we cannot move this PR forward at this moment. Thanks very much for your contribution anyway.

Regarding some of the other discussions above, we definitely have in mind to develop some built-in helper that will automatically manage the definition of the environment, injection of wrappers, etc in ``AutotoolsToolchain``, just we didn't have time with the 2.0, but it will be addressed later, and we think that could help to simplify many ConanCenter recipes.

SpaceIm:
Could you at least convert paths from tools.build:compiler_executables with `unix_path` please? It's totally broken currently when build machine is Windows (and msvc is not the only compiler, you can use MinGW, or cross-build to Android with Android NDK, cross build to webassembly with emsdk etc, I have the feeling that conan is somewhat msvc centric when it comes to Windows, but there are other compilers). See https://github.com/conan-io/conan-center-index/issues/17143

SSE4:
why was it rejected?
if it's too risky or hard, it might be put into opt-in.

yowidin:
@memsharded @madebr any feedback would be appreciated.

memsharded:
The use case, need, and solved problem is not evident enough to introduce the proposed complexity.

Quoting has worked well for lots of similar cases, are we fully sure that this is not just a matter of adding some quote somewhere, and avoid the added complexity?

For this to be considered, it would be necessary more evidence, well described detailed and real scenarios with reproducible cases (something that could also be a test in the test suite).

SpaceIm:
@memsharded I think that curating paths from `tools.build:compiler_executables` with `unix_path()` is not complex. Just drop the `get_short_path_name()` call which was the logic you were scary about.

It would become a one line change in AutotoolsToolchain:

```patch
-env.define(env_var, compilers_by_conf[comp])
+env.define(env_var, unix_path(self._conanfile, compilers_by_conf[comp]))
```

memsharded:
That change might be fair, but that is unrelated to this current PR.
For this kind of change to be moved forward it is better to create a new issue, not comments on PRs, specially when they are closed.

SpaceIm:
This PR was exactly about that: curating CC, CXX etc on Windows. Why do you say it's unrelated?

memsharded:
The PR was about Windows shortening syntax with ``~``, not about the Windows subsystems paths, it is a very different problem, for different scope, that works in different cases.

SpaceIm:
It was about curation of env vars in general. There is a call to unix_path in this PR. Nevermind.

memsharded:
This was not about env-vars in general but about the ``conf`` of ``tools.build:compiler_executables``. 

You are totally right about ``unix_path`` was there. The problem is that 99% of the code in this PR was the ``ctypes`` Windows path shortening and the possibility of dropping it and leaving the``unix_path`` was done after the PR was closed, and that is very hard to track correctly. In the eyes of the reviewers, this PRs was not about ``unix_paths`` fixes for Windows subsystems, but something very different. Because all that ``ctypes`` code would be running always, not only for Windows subsystems. 
So either the PR is re-opened, removing the rejected parts, or a new issue is created, or a new PR with just the ``unix_path`` is done (and hopefully with some tests, in general PRs without accompanying tests will not be merged, and the absence of tests make the PRs less likely to be moved forward, specially for something that is supposedly fixing a bug)

yowidin:
@memsharded sorry for not reading the PR, I just assumed it was related since it was linked to my [issue](https://github.com/conan-io/conan-center-index/issues/17143) (can you also take a look at it for more context?).

My question is whether this kind of handling (namely wrapping compiler paths in `unix_path`) is something that could/should be implemented in the conan itself.

In my understanding with conan2, the conan is no longer "passive", in the sense that it just tagged the binaries before, but also tries to take care of the right flags for different platforms and settings, using different build tools. If that's correct, then this kind of fix shouldn't be out of place (in another PR and with all the tests of course).

memsharded:
Sure, @yowidin, thanks for the clarifications.

Yes, the purpose of the build system integrations like ``AutotoolsToolchain`` is to automate the most common cases, and Windows subsystems support is intended, it is definitely possible that there is a gap/bug in it and that it might require a fix.

I am going to spin off a new ticket for it.

SpaceIm:
> This was not about env-vars in general but about the `conf` of `tools.build:compiler_executables`.

This PR has been submitted before availability of `tools.build:compiler_executables` in conan client. In its original form it was catching CC, CXX etc env vars from profile and curate them so that it was robust to any path flavor set from profile (windows style, or spaces in paths) so that users don't have to care about underlying build system.

> (and hopefully with some tests, in general PRs without accompanying tests will not be merged, and the absence of tests make the PRs less likely to be moved forward, specially for something that is supposedly fixing a bug)

I've found only 1 test of AutotoolsToolchain on Windows where autotools are called then configure then make, but it's disabled, so it's hard to write anything for this PR if the basic test without any CC/CXX set is not even covered :s

memsharded:
https://github.com/conan-io/conan/pull/13867 adding ``unix_paths`` for AutotoolsToolchain ``tools.build:compiler_executables``, for Conan 2.0.5

PR #12230, opened 2022-10-02
new profile [runenv] section

memsharded:
Changelog: Feature: Implement a new ``[runenv]`` section in the Profile, to define the runtime environment.
Docs: https://github.com/conan-io/docs/pull/2771

I hit this limitation while preparing docs for Windows-CLang integration:

- When building for Msys2-Clang, I can add ``[buildenv]...<path/to/clang/bin>`` so the clang compiler is found
- But when I am executing a ``test_package`` ``test()`` or something in the "run" scope, like ``self.run("myexe", env="conanrun")`` it will fail, because it cannot find the ``libc++.dll`` and ``libunwind.dll`` shared libs in the msys2 clang folder.

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -220,7 +220,7 @@ class ConanFile(object):
             self._conan_runenv = self._conan_runenv.get_profile_env(ref_str)
         return self._conan_runenv
 
-    def initialize(self, settings, env, buildenv=None, runenv=True):
+    def initialize(self, settings, env, buildenv=None, runenv=None):
         self._conan_buildenv = buildenv
         self._conan_runenv = runenv
         if isinstance(self.generators, str):
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -212,7 +212,7 @@ class ConanFile(object):
 
     @property
     def runenv(self):
-        # Lazy computation of the package buildenv based on the profileone
+        # Lazy computation of the package runenv based on the profile one
         from conan.tools.env import Environment
         if not isinstance(self._conan_runenv, Environment):
             # TODO: missing user/channel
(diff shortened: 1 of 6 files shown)


conans/model/conan_file.py
```diff
@@ -209,8 +210,19 @@ def buildenv(self):
             self._conan_buildenv = self._conan_buildenv.get_profile_env(ref_str)
         return self._conan_buildenv
 
-    def initialize(self, settings, env, buildenv=None):
+    @property
+    def runenv(self):
+        # Lazy computation of the package buildenv based on the profileone
```
jcar87: The comment mentions `buildenv`, should it be `runenv`?

jcar87:
```suggestion
        # Lazy computation of the package runenv based on the profile one
```
?

jcar87 (approved):
Just a very minor comment.

> But when I am executing a test_package test() or something in the "run" scope, like self.run("myexe", env="conanrun") it will fail, because it cannot find the libc++.dll and libunwind.dll shared libs in the msys2 clang folder.

Is this with a `clang-cl.exe` or `clang++.exe` that are not part of a "system" package or similar? Otherwise I'm surprised `libc++.dll` is not already found in the "default" search path in this environment (although I'm not very familiar with msys2)

conans/model/conan_file.py
memsharded: yes, comment was not updated,thanks!

memsharded:
> Is this with a clang-cl.exe or clang++.exe that are not part of a "system" package or similar? Otherwise I'm surprised libc++.dll is not already found in the "default" search path in this environment (although I'm not very familiar with msys2)

Yes, the msys2 clang++.exe can be installed in a custom location, like ``C:/msys2/clang64/bin/clang++.exe``. This ``C:/msys2/clang64/bin`` folder can be indeed added to the system PATH and everything will work. But if this folder is not added, and it is defined in the Conan profile (which makes sense if we are dealing with multiple compilers and versions), then, the ``libc++.dll`` will not be in the path either.

This behavior is hidden because in test we add the ``pytest.mark.tool("msys2")`` to the test, which injects the path to msys2 during the whole test.

czoido (approved):

PR #12235, opened 2022-10-04
Follow conventions for Conan 2.0 CLI output

czoido:
Still working on it ...

Following the convention for Conan 2.0 that the results for the commands are dumped to the stdout. Rest of operations while running the command are dumped to stderr.

Several fixes in the output for the CLI:
- All commands must define a `text` formatter if they have to output something to the stdout. 
- The `text` formatter is not shown in the help and is used by default in case the command defined it.

I think that the formatters were not correctly used in the code because they only formatted the output and not write anything and maybe that was a bit confusing, so now the formatters are changed to also do the actual output to stdout using `cli_out_write()`

### DIFF
--- a/conan/cli/command.py
+++ b/conan/cli/command.py
@@ -225,14 +225,3 @@ def conan_subcommand(formatters=None):
         return cmd
 
     return decorator
-
-
-def cli_out_write(data, fg=None, bg=None, endline="\n", indentation=0):
-    fg_ = fg or ''
-    bg_ = bg or ''
-    if color_enabled(sys.stdout):
-        data = f"{' ' * indentation}{fg_}{bg_}{data}{Style.RESET_ALL}{endline}"
-    else:
-        data = f"{' ' * indentation}{data}{endline}"
-
-    sys.stdout.write(data)
--- a/conan/cli/command.py
+++ b/conan/cli/command.py
@@ -75,7 +75,7 @@ class SmartFormatter(argparse.HelpFormatter):
 
 class BaseConanCommand(object):
     def __init__(self, method, formatters=None):
-        self._formatters = {}
+        self._formatters = {"text": lambda x: None}
         self._method = method
         self._name = None
         self._parser = None
@@ -96,9 +96,13 @@ class BaseConanCommand(object):
     def _init_log_levels(self):
         add_log_level_args(self._parser)
 
+    @property
+    def _help_formatters(self):
+        return [formatter for formatter in list(self._formatters) if formatter != "text"]
+
     def _init_formatters(self):
-        if self._formatters:
-            help_message = "Select the output format: {}".format(", ".join(list(self._formatters)))
+        if self._help_formatters:
+            help_message = "Select the output format: {}".format(", ".join(list(self._help_formatters)))
             self._parser.add_argument('-f', '--format', action=OnceArgument, help=help_message)
 
     @property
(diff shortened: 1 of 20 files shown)


conan/cli/commands/inspect.py
```diff
@@ -1,30 +1,33 @@
...
+        if isinstance(value, dict):
+            out.writeln(f"{name}:")
+            for k, v in value.items():
+                out.writeln(f"    {k}: {v}")
+        else:
+            out.writeln("{}: {}".format(name, value))
 
 
-@conan_command(group=COMMAND_GROUPS['consumer'])
+@conan_command(group=COMMAND_GROUPS['consumer'], formatters={"text": inspect_text_formatter, "json": default_json_formatter})
```
czoido: The changes in this file are not all related to where the output goes. I removed the `path` subcommand from `conan inspect` because I think it was probably a mistake when porting the command? You had to invoke conan inspect like `conan inspect path conanfile.py` instead of `conan inspect conanfile.py`

memsharded: It was not a mistake, we were discussing this ``path`` as an approach to leave the possibility open for the future to inspect references in the cache and in remotes

czoido: Maybe we should consider using a syntax more aligned with other commands, something like:
```
conan inspect conanfile.py
conan inspect --reference=zlib/1.2.11
```
I would leave the inspect on remote case for users to implement their own using custom commands.
I'll revert the changes and maybe we can discuss a change of syntax in other PR or issue.

conan/cli/commands/__init__.py
```diff
@@ -30,9 +32,13 @@ def default(self, o):
         return JSONEncoder.default(self, o)
 
 
-def json_formatter(data):
+def default_json_formatter(data):
     myjson = json.dumps(data, indent=4, cls=ConanJSONEncoder)
-    return myjson
+    ConanOutput(stream=sys.stdout).writeln(myjson)
```
memsharded: I am not sure about this. We were discussing about how ``ConanOutput`` was always for our messages to stderr, while we would use other things for stdout. Becaye also there is extra configurations, like loggers or the logger-as-json that can affect how the ``ConanOutput`` behaves, and that would be bad for the ``stdout`` output, wouldn't it?

conan/cli/command.py
```diff
@@ -75,7 +75,7 @@ def _fill_text(self, text, width, indent):
 
 class BaseConanCommand(object):
     def __init__(self, method, formatters=None):
-        self._formatters = {}
+        self._formatters = {"text": lambda x: None}
```
memsharded, line 78: Maybe it is better to leave it empty, and if something returns something but ``text`` is not implemented, then fail?

conan/cli/commands/cache.py
```diff
@@ -1,18 +1,18 @@
...
 
-@conan_command(group=COMMAND_GROUPS['consumer'], formatters={"json": json_formatter})
+@conan_command(group=COMMAND_GROUPS['consumer'])
 def cache(conan_api: ConanAPIV2, parser, *args):
     """Performs file operations in the local cache (of recipes and packages)"""
     pass
 
 
-@conan_subcommand()
+@conan_subcommand(formatters={"text": default_text_formatter})
```
memsharded, line 15: This is related, this ``default_text_formatter`` is internally using ``ConanOutput.info()``. Which we might want to later add some ``INFO:`` tag, for example, or have something that adds timestamps to our output.

memsharded (commented):
Need to discuss some approaches

conan/cli/commands/cache.py
czoido: That was a mistake, I'll change that

conan/cli/commands/__init__.py
czoido: Ok, I see, then probably what needs to be used is `cli_out_write` (we can change the name later perhaps), I'll make the changes and we see from there

conan/cli/command.py
czoido: I made it that way because there are cases that you don't have a text formatter but you do have a json formatter, so you have to return data but the text formatter will just do nothing

memsharded: Ok, makes sense 👍

memsharded (approved):
Great job

PR #11492, opened 2022-06-20
Better clang Windows support

memsharded:
Changelog: Feature: Improve Clang support in Windows.
Docs: TBD

Tested, all with CMake:
- Windows LLVM/Clang with mingw generator
- Windows LLVM/Clang with Ninja and NMake generators
- Windows LLVM/Clang Using different versions of the VS runtime
- Windows Visual Studio integrated ClangCL
- Windows Msys2 LLVM/Clang targeting native Windows, Mingw Clang toolchain

Windows Visual Studio ClangCL with externally installed LLVM/CLang is **impossible** with ``CMake``, it seems possible with native Visual Studio.

Close https://github.com/conan-io/conan/issues/10955
Close https://github.com/conan-io/conan/issues/10835
Close https://github.com/conan-io/conan/issues/1839
Close https://github.com/conan-io/conan/issues/6927
Close https://github.com/conan-io/conan/issues/7868
Close https://github.com/conan-io/conan/issues/9505

PR #12241, opened 2022-10-04
[feature] Detect if running clang compiler on Windows

### LINKED ISSUE #12175: [feature] Recognize Clang CL on Windows as tool method
Just like `is_msvc`, usually, in CCI, we need to use the follow idiom to check when running clang-cl:

```python
@property
def _is_clang_cl(self)
    return (self.settings.os == "Windows" and self.settings.compiler == "clang") or \
            self.settings.get_safe("compiler.toolset") == "ClangCL"
```

It would be great internalizing it as part of `conan.tools.microsoft` so we no longer need to copy/past that method.

You can find its usage by the follow search: https://github.com/conan-io/conan-center-index/search?q=_is_clang_cl

uilianries:
Changelog: Feature: `conan.tools.microsoft.is_clang_cl` detects if running Clang on Windows
Docs: https://github.com/conan-io/docs/pull/2774

close #12175

### DIFF
--- a/conan/tools/microsoft/__init__.py
+++ b/conan/tools/microsoft/__init__.py
@@ -5,3 +5,4 @@ from conan.tools.microsoft.subsystems import unix_path
 from conan.tools.microsoft.toolchain import MSBuildToolchain
 from conan.tools.microsoft.visual import msvc_runtime_flag, VCVars, is_msvc, \
     is_msvc_static_runtime, check_min_vs
+from conan.tools.microsoft.clang import is_clang_cl
--- a/conan/tools/microsoft/clang.py
+++ b/conan/tools/microsoft/clang.py
@@ -1,0 +1,14 @@
+
+def is_clang_cl(conanfile, build_context=False):
+    """ Validate if current compiler in settings is 'clang' and running on Windows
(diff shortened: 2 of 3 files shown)


memsharded (changes requested):
I am afraid this is no longer valid. The ``clangCL`` is nothing specific anymore, and is not part of the new ``msvc`` compiler either, the right compiler is ``clang``, with generator ``Visual Studio``.

Check: https://github.com/conan-io/conan/pull/11492

uilianries:
Closing this one. With #11492 is clear that a generic check for CLang on Windows will not work for many cases. So far, 3% of CCI recipes need to validate clang on Windows, but we don't have a clear situation about which clang and its version. It will be better analyzing each case there and consider the new proposal from #11492.

SpaceIm:
Is it more clear now that https://github.com/conan-io/conan/pull/11492 has been merged? 3% of CCI recipes try to manage clang CL or a mingw clang flavor, but many more are broken and would need better detection of clang CL.

It would also be nice to have a `is_mingw()` (because it can be gcc targeting Windows, or a clang flavor targeting windows).

PR #12133, opened 2022-09-16
WIP: Fix rpath for executables in `fix_apple_shared_install_name` tool

lasote:
Changelog: Feature: The tool `fix_apple_shared_install_name` now also patches the `<package_id>/bin` executables found to point to the libraries at `../lib/xxx.dylib`.
Docs: https://github.com/conan-io/docs/pull/XXXX

Close https://github.com/conan-io/conan/issues/12107

PR #12249, opened 2022-10-05
Fix library install names in packaged executables

### LINKED ISSUE #12107: [feature] fix_apple_shared_install_name: also fix install_name in LC_LOAD_DYLIB of executables in cpp_info.bindirs

For all dylib located in cpp_info.libdirs, `conan.tools.apple.fix_apple_shared_install_name` changes `install_name` to `@rpath` in their `LC_ID_DYLIB` section and also `install_name` of their dependencies in `LC_LOAD_DYLIB`.

But executables located in cpp_info.bindirs are not fixed, so they may not be relocatable if they have been linked to dylib of the same recipe fixed afterwards by `conan.tools.apple.fix_apple_shared_install_name` (because these executables have been created before the call of this method, so they also need this fix).

For example I'm migrating `capnproto` recipe to conan v2, here is the result of `otool -L` on one of its executables after `package()`, even with `fix_apple_shared_install_name(self)`:

```
/Users/spaceim/.conan/data/capnproto/0.10.1/_/_/package/86b3b8f1d2506eef508c4383f35d09efc02ebc0b/bin/capnp:
	/lib/libcapnpc-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	/lib/libcapnp-json-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	/lib/libcapnp-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	/lib/libkj-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.12)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

(`libcapnpc`, `libcapnp-json`, `libcapnp` & `libkj` are libraries created & packaged in the same recipe, and properly fixed by `fix_apple_shared_install_name()`)

I would have expected:

```
/Users/spaceim/.conan/data/capnproto/0.10.1/_/_/package/86b3b8f1d2506eef508c4383f35d09efc02ebc0b/bin/capnp:
	@rpath/libcapnpc-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libcapnp-json-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libcapnp-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libkj-0.10.1.dylib (compatibility version 0.0.0, current version 0.0.0)
	@rpath/libz.1.dylib (compatibility version 1.0.0, current version 1.2.12)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1300.23.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

jcar87:
Changelog: Feature: The tool `fix_apple_shared_install_name` now also patches the `<package_id>/bin` executables found to point to the libraries at `../lib/xxx.dylib`.
Docs: https://github.com/conan-io/docs/pull/2806

This PR is built upon: https://github.com/conan-io/conan/pull/12133

Main differences from the earlier PR:
* To locate executables inside the binfolders, call `otool -hv` which prints a summary of the Mach-O header:
```
Mach header
      magic  cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64    ARM64        ALL  0x00     EXECUTE    20       1120   NOUNDEFS DYLDLINK TWOLEVEL PIE
```
Where `EXECUTE` is what we're after. This simplifies things as we no longer need to check for the executable flag or make other assumptions.

* After retrieving the library dependencies from the executable (with `otool -L`), work out which entries need patching by comparing the entries with the `substitutions` dictionary - this dictionary already includes a list of which libraries in the package were packaged, so we know exactly which ones to patch. This fixes logic to not rely on other reasonings such as checking if install names start with `/lib`

* If the executable references a dependency on `/lib/libfoo.dylib`, and this library is inside the package and we have patched the install name to `@rpath/libfoo.dylib`, then the executable has the reference replaced to exactly that (rather than `@executable_path/../lib/libfoo.dylib` - this is because one typically expects the reference of the dependency to match the install name of the dependency. 

* To support the two points above, and to overcome some inconsistency of libtool - had to add some logic to handle a reference to `/lib/libfoo.1.dylib` even if the install name did not include this originally. 

* Add explicit rpath entries to the executables that are patched. In the typical case, the executable will contain an entry to `@rpath/../lib` - relative paths are computed for this. 

Co-authored with @lasote 

Close https://github.com/conan-io/conan/issues/12107

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -204,8 +204,7 @@ def fix_apple_shared_install_name(conanfile):
         for libdir in libdirs:
             full_folder = os.path.join(conanfile.package_folder, libdir)
             if not os.path.exists(full_folder):
-                # as method package is running before package_info, the cpp.package might be
-                # wrong
+                # Should this be an error?
                 continue
             shared_libs = _darwin_collect_dylibs(full_folder)
             # fix LC_ID_DYLIB in first pass
@@ -231,8 +230,8 @@ def fix_apple_shared_install_name(conanfile):
         for bindir in bindirs:
             full_folder = os.path.join(conanfile.package_folder, bindir)
             if not os.path.exists(full_folder):
-                # as method package is running before package_info, the cpp.package might be
-                # wrong
+                # Skip if the folder does not exist inside the package
+                # (e.g. package does not contain executables but bindirs is defined)
                 continue
             executables = _darwin_collect_executables(full_folder)
             for executable in executables:
@@ -258,4 +257,8 @@ def fix_apple_shared_install_name(conanfile):
 
     if is_apple_os(conanfile) and conanfile.options.get_safe("shared", False):
         substitutions = _fix_dylib_files(conanfile)
+
+        # Only "fix" executables if dylib files were patched, otherwise 
+        # there is nothing to do.
+        if substitutions:
         _fix_executables(conanfile, substitutions)
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -181,6 +181,17 @@ def fix_apple_shared_install_name(conanfile):
                 ret.append(full_path)
         return ret
 
+    def _get_rpath_entries(binary_file):
+        entries = []
+        command = "otool -l {}".format(binary_file)
+        otool_output = check_output_runner(command).splitlines()
+        for count, text in enumerate(otool_output):
+            pass
+            if "LC_RPATH" in text:
+                rpath_entry = otool_output[count+2].split("path ")[1].split(" ")[0]
+                entries.append(rpath_entry)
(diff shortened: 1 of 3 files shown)


czoido (approved):
really nice! :raised_hands:

czoido (approved):

PR #12319, opened 2022-10-17
Use settings_target in xcrun instead of settings

### LINKED ISSUE #12260: [bug] Conan v2 think that macOS SDK is an iOS SDK
  * Operating System+version: macos latest
  * Compiler+version: clang 14
  * Conan version: 1.53
  * Python version: 3.9

```
git clone https://github.com/nativium/nativium.git -b fix-conan-1-53
cd nativium
python3 nativium.py conan setup
python3 nativium.py target ios setup --arch=arm64 --build=debug --group=ios
python3 nativium.py target ios build --arch=arm64 --build=debug --group=ios
```

```
CMake Error at /opt/homebrew/Cellar/cmake/3.24.2/share/cmake/Modules/Platform/iOS-Initialize.cmake:4 (message):
  
  /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk
  is not an iOS SDK
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.24.2/share/cmake/Modules/CMakeSystemSpecificInitialize.cmake:21 (include)
  CMakeLists.txt:8 (project)
```

The problem is here:
https://github.com/nativium/nativium/blob/fix-conan-1-53/conan/darwin-toolchain/conanfile.py#L102

The `XCRun` object from conan don't check correctly the settings. It get the settings from conanfile, but need be something like i do, that is try to get `settings_target`. Or make XCRun receive `settings` and not `conanfile`:

```
settings_target = None

if hasattr(self, "settings_target"):
    settings_target = self.settings_target
    self.output.info("Settings: Target")
else:
    settings_target = self.settings
    self.output.info("Settings: Default")
```

The conan XCRun lines is here:
https://github.com/conan-io/conan/blob/develop/conan/tools/apple/apple.py#L82-L87

czoido:
Changelog: Fix: Use `settings_target` in XCRun instead of settings as default and provide argument `use_settings_target ` to select using the active context settings.
Docs: https://github.com/conan-io/docs/pull/2797

`use_settings_target` is `False` by default because because of the use of this tool in ConanCenter, for example _openssl_ builds with _AutoTools_ so needs the **sysroot** (uses _XCRun.sdk_) and is a require by _CMake_ so in the **build context** it needs the `settings_build`, not the `settings_target`

To discuss: `use_settings_target ` default and name.

TODO: Check other tools that may need this fix too

Closes: https://github.com/conan-io/conan/issues/12260

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -79,14 +79,14 @@ def apple_sdk_path(conanfile):
 
 class XCRun(object):
 
-    def __init__(self, conanfile, sdk=None, use_target_settings=False):
+    def __init__(self, conanfile, sdk=None, use_settings_target=False):
         """sdk=False will skip the flag
            sdk=None will try to adjust it automatically
            target_settings=True try to use settings_target in case they exist"""
 
         # FIXME: 2.0: remove "hasattr()" condition
         settings = conanfile.settings
-        if use_target_settings and hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
+        if use_settings_target and hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
             settings = conanfile.settings_target
 
         if sdk is None and settings:
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -83,13 +83,10 @@ class XCRun(object):
         """sdk=False will skip the flag
            sdk=None will try to adjust it automatically"""
 
-        if hasattr(conanfile, "settings_target"):
-            settings_target = conanfile.settings_target
-        else:
-            settings_target = conanfile.settings
+        settings = conanfile.settings_target if hasattr(conanfile, "settings_target") else conanfile.settings
 
-        if sdk is None and settings_target:
-            sdk = apple_sdk_name(settings_target)
+        if sdk is None and settings:
+            sdk = apple_sdk_name(settings)
         self.sdk = sdk
 
     def _invoke(self, args):
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -79,14 +79,15 @@ def apple_sdk_path(conanfile):
 
 class XCRun(object):
 
-    def __init__(self, conanfile, sdk=None):
+    def __init__(self, conanfile, sdk=None, use_target_settings=True):
         """sdk=False will skip the flag
-           sdk=None will try to adjust it automatically"""
+           sdk=None will try to adjust it automatically
+           target_settings=True try to use settings_target in case they exist"""
 
-        if hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
+        # FIXME: 2.0: remove "hasattr()" condition
+        settings = conanfile.settings
+        if use_target_settings and hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
             settings = conanfile.settings_target
-        else:
-            settings = conanfile.settings
 
         if sdk is None and settings:
             sdk = apple_sdk_name(settings)
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -83,10 +83,13 @@ class XCRun(object):
         """sdk=False will skip the flag
            sdk=None will try to adjust it automatically"""
 
-        settings = conanfile.settings_target if hasattr(conanfile, "settings_target") else conanfile.settings
+        if hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
+            settings_target = conanfile.settings_target
+        else:
+            settings_target = conanfile.settings
 
-        if sdk is None and settings:
-            sdk = apple_sdk_name(settings)
+        if sdk is None and settings_target:
+            sdk = apple_sdk_name(settings_target)
         self.sdk = sdk
 
     def _invoke(self, args):
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -79,7 +79,7 @@ def apple_sdk_path(conanfile):
 
 class XCRun(object):
 
-    def __init__(self, conanfile, sdk=None, use_target_settings=True):
+    def __init__(self, conanfile, sdk=None, use_target_settings=False):
         """sdk=False will skip the flag
            sdk=None will try to adjust it automatically
            target_settings=True try to use settings_target in case they exist"""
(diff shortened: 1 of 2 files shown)


conan/tools/apple/apple.py
```diff
@@ -82,8 +82,14 @@ class XCRun(object):
     def __init__(self, conanfile, sdk=None):
         """sdk=False will skip the flag
            sdk=None will try to adjust it automatically"""
-        if sdk is None and conanfile.settings:
-            sdk = apple_sdk_name(conanfile.settings)
+
+        if hasattr(conanfile, "settings_target") and conanfile.settings_target is not None:
```
memsharded: So we can assume that if ``settings_target`` exists, this means that we are already in a ``tool_require`` and we always want to refer our ``XCRun()`` to the target context? Aren't there any cases, where it will want just the ``XCRun()`` of the current "build" context? Maybe it should be an optional argument?

czoido: Yes, that also may happen, I'll add an argument so you can get the info based on the build settings.

jcar87: I've looked into this a bit deeper and I believe that relying on `settings_target` by default should be correct. The information that `xcrun` outputs is "on the build machine, _for_ the target platform" - it's only used in the context of building something from source.

The cases I've seen in Conan Center, call `xcrun` to set up arguments to pass to build tools (so it would impact the  `generate()` and `build()` methods).

The case in https://github.com/conan-io/conan/issues/12260 is actually invoking it in `package_info()`, for a package that is a `tool_require` (likely in a profile), and `xcrun` is called in the `package_info`. So "only used in the context of building something from source" still applies, but now with a level of indirection.

If a `tool_require` A has a dependency on another `tool_require` B such that A is built from source, I believe the `settings_target` should cover this case, by taking the values from the `build` profile (docs [here](https://docs.conan.io/en/latest/systems_cross_building/cross_building.html?highlight=settings_target)).

The only case where I can see where one may need to call `xcrun` from the build context settings, is those recipes that need to invoke two compilers, for example those that first need to bootstrap a tool with the "build" context compiler, and then invoke the compiler that is set up for the host context. There aren't many of these, and in those cases because the "build" contest is always macOS to run on macOS, they may be able to rely on calling the compiler with its default without going through `xcrun` -

TL;DR: always use `settings_target` by default for the overwhelmingly typical case, leave an option to force the build context "just in case"

memsharded (approved):

conan/tools/apple/apple.py
```diff
@@ -79,11 +79,18 @@ def apple_sdk_path(conanfile):
 
 class XCRun(object):
 
-    def __init__(self, conanfile, sdk=None):
+    def __init__(self, conanfile, sdk=None, use_target_settings=False):
```
franramirez688: Would it be better to call it `use_settings_target` to be more intuitive?

conans/test/functional/tools/test_apple_tools.py
```diff
@@ -24,3 +25,96 @@ def build(self):
...
     client.run("create .")
     assert "Xcode.app/Contents/Developer/Platforms/MacOSX.platform" in client.out
+
+
+@pytest.mark.skipif(platform.system() != "Darwin", reason="Requires Xcode")
+def test_xcrun_in_tool_requires():
+    # https://github.com/conan-io/conan/issues/12260
+    client = TestClient()
+    tool = textwrap.dedent("""
+        from conans import ConanFile
```
franramirez688:
```suggestion
        from conan import ConanFile
```

conans/test/functional/tools/test_apple_tools.py
```diff
@@ -24,3 +25,96 @@ def build(self):
...
+
+            def package_info(self):
+                xcrun = XCRun(self{})
+                self.output.info("sdk: %s" % xcrun.sdk)
+        """)
+    client.save({"br.py": tool.format(", use_target_settings=True")})
+    client.run("export br.py br/0.1@")
+
+    conanfile = textwrap.dedent("""
+        from conans import ConanFile
```
franramirez688:
```suggestion
        from conan import ConanFile
```

conans/test/functional/tools/test_apple_tools.py
```diff
@@ -24,3 +25,96 @@ def build(self):
...
+def test_xcrun_in_required_by_tool_requires():
+    """
+    ConanCenter case, most typical, openssl builds with autotools so needs the sysroot
+    and is a require by cmake so in the build context it needs the settings_build, not
+    the settings_target, that's why the use_target_settings default is False
+    """
+    client = TestClient()
+
+    openssl = textwrap.dedent("""
+        from conans import ConanFile
```
franramirez688:
```suggestion
        from conan import ConanFile
```

conans/test/functional/tools/test_apple_tools.py
```diff
@@ -24,3 +25,96 @@ def build(self):
...
+        from conan.tools.apple import XCRun
+        class Pkg(ConanFile):
+            settings = "os", "compiler", "build_type", "arch"
+            def build(self):
+                xcrun = XCRun(self)
+                self.output.info("sdk for building openssl: %s" % xcrun.sdk)
+        """)
+
+    consumer = textwrap.dedent("""
+        from conans import ConanFile
```
franramirez688:
```suggestion
        from conan import ConanFile
```

conans/test/functional/tools/test_apple_tools.py
```diff
@@ -24,3 +25,96 @@ def build(self):
...
+    """)
+
+    client.save({"cmake.py": GenConanfile("cmake", "1.0").with_requires("openssl/1.0"),
+                 "openssl.py": openssl,
+                 "consumer.py": consumer,
+                 "profile_ios": profile_ios})
+
+    client.run("export openssl.py openssl/1.0@")
+    client.run("export cmake.py")
+    client.run("create consumer.py consumer/1.0@ --build -pr:h=./profile_ios -pr:b=default --build")
```
czoido:
```suggestion
    client.run("create consumer.py consumer/1.0@ -pr:h=./profile_ios -pr:b=default --build")
```

franramirez688 (approved):

conan/tools/apple/apple.py
czoido: yes, that's probably better, I'll change that 👍

jcar87 (approved):

Issue #12327, opened 2022-10-18
[bug] Conan 1.X: prefer build_requires from direct/indirect dependencies over build_requires from profiles


thorsten-klein:
### Environment Details (include every applicable attribute)
  * Operating System+version: Ubuntu 20.04
  * Compiler+version: clang-13
  * Conan version: 1.44.1
  * Python version: 3.8.10

### Steps to reproduce (Include if Applicable)
==> mybin.py <==
```
import os

from conans import ConanFile

class ConanRecipe(ConanFile):
    name = "mybin"
    version = "1.0.0"

    def build(self):
        pass
        
    def package(self):
        os.makedirs(os.path.join(self.package_folder, "bin"), exist_ok=True)
        self.run("cd %s/bin && echo 'echo mybin' > mybin; chmod +x mybin" % self.package_folder)
```

==> myproject.py <==
```
import os

from conans import ConanFile

class ConanRecipe(ConanFile):
    name = "myproject"
    version = "1.0.0"

    def build_requirements(self):
        self.build_requires("mybin/1.0.0")
    
    def build(self):
        print("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
        self.run("mybin")
        print("xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx")
        self.run("which -a mybin")
 ```
       
==> sysroot.py <==
```
import os

from conans import ConanFile

class ConanRecipe(ConanFile):
    name = "sysroot"
    version = "1.0.0"

    def build(self):
        pass
        
    def package(self):
        os.makedirs(os.path.join(self.package_folder, "bin"), exist_ok=True)
        self.run("cd %s/bin && echo 'echo sysroot' > mybin; chmod +x mybin" % self.package_folder)
```

==> sysroot.profile <==
```
[build_requires]
sysroot/1.0.0
```

==> run.sh <==
```
#!/bin/bash

set -ex

conan export sysroot.py
conan export mybin.py
conan export myproject.py test/test

rm -rf build && mkdir build
cd build

conan install myproject/1.0.0@test/test -pr:h=../sysroot.profile -pr:b=../sysroot.profile --build=missing --build=myproject
```

### Logs (Executed commands with output) (Include/Attach if Applicable)
```
myproject/1.0.0@test/test: Calling build()
xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
[...]

PR #12328, opened 2022-10-18
prefer direct + indirect build_requires over build_requires from profiles

thorsten-klein:
proposed solution for #12327


thorsten-klein:
no PR required for conan 1.X
Issue can be workarounded as described in #12327

PR #12349, opened 2022-10-20
Do not add architecture to system packages by default, only when cross-compiling

### LINKED ISSUE #12320: [feature] install system packages without architecture
Currently, recipes with arch settings trying to install system packages always install the system package with architecture matching the profile. This makes xorg/system fail on some distros (alma, Fedora and Centos), because some of the system packages do not have an architecture:
the failures:
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369817368#step:6:191
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369817556#step:6:180
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369816469#step:6:263
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369819520#step:6:238
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369819717#step:6:198
- https://github.com/bincrafters/system-packages-checks/actions/runs/3266259888/jobs/5369817145#step:6:244

these two system packages have no arch, cf!
- https://almalinux.pkgs.org/9/almalinux-appstream-x86_64/xkeyboard-config-devel-2.33-2.el9.noarch.rpm.html
- https://almalinux.pkgs.org/9/almalinux-crb-x86_64/xorg-x11-xtrans-devel-1.4.0-8.el9.noarch.rpm.html

We could consider adding a method like `_SystemPackageManagerTool.install_archless(...)` ? Or an automatic fallback to archless if there is no existing package for the setting's arch ?

czoido:
Changelog: Feature: Do not add architecture to system packages by default, only when cross-compiling.
Docs: omit

The most common situation when installing system packages is that you don't use the architecture in the package name unless you are cross-compiling, so this defaults to that behaviour.
Also it's testing how to install archless packages even with cross-compilation, but that's something we probably want to change as the code would look a bit ugly. You can workaround that with an empty dictionary as arch_names in classes that accept the argument:

```
Apt(self).install(["libudev-dev"], update=True, check=True)
Apt(self, arch_names={}).install(["archless-package"], update=True, check=True)
```

Closes: https://github.com/conan-io/conan/issues/12320

### DIFF
--- a/conans/test/integration/tools/system/package_manager_test.py
+++ b/conans/test/integration/tools/system/package_manager_test.py
@@ -10,7 +10,7 @@ from conan.tools.system.package_manager import Apt, Dnf, Yum, Brew, Pkg, PkgUtil
 from conans import Settings
 from conans.errors import ConanException
 from conans.model.conf import Conf
-from conans.test.utils.mocks import ConanFileMock
+from conans.test.utils.mocks import ConanFileMock, MockSettings
 
 
 @pytest.mark.parametrize("platform, tool", [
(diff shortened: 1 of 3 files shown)


conans/test/integration/tools/system/package_manager_test.py
```diff
@@ -244,3 +286,5 @@ def test_tools_check(tool_class, result):
         context_mock.return_value = "host"
         tool = tool_class(conanfile)
         tool.check(["package"])
+
+    assert tool._conanfile.command == result
```
czoido, line 290: this test was missing the final check, so it was not testing anything... 🙄

ericLemanissier:
I don't think this closes https://github.com/conan-io/conan/issues/12320, because the situation stays the same during cross compilation: for a given recipe either:
- all system packages have an arch suffix (if the recipe has an arch setting) or
- no system package has an arch suffix (if the recipe does not have an arch setting)

To fix the issue, we would need to have a method which installs the system packages it receives as argument, with `noarch`/`all`/`any` architecture suffix, depending on the package manager.
Just to be clear, there is no need for this feature currently in CCI. The offending case in CCI (`xorg/system`) was solved by removing all the arch-less system packages from `xorg/system`, and putting them in a dedicated recipe without arch setting

czoido:
> I don't think this closes #12320, because the situation stays the same during cross compilation: for a given recipe either:
> 
> * all system packages have an arch suffix (if the recipe has an arch setting) or
> * no system package has an arch suffix (if the recipe does not have an arch setting)
> 
> To fix the issue, we would need to have a method which installs the system packages it receives as argument, with `noarch`/`all`/`any` architecture suffix, depending on the package manager. Just to be clear, there is no need for this feature currently in CCI. The offending case in CCI (`xorg/system`) was solved by removing all the arch-less system packages from `xorg/system`, and putting them in a dedicated recipe without arch setting

Hi @ericLemanissier, thanks for the comments. The intention of this PR is to:
- Change the default when invoking system package managers and don't add the architecture to the package names for the most common case that is not cross-compiling. As you say this would not close the issue.
- Provide a way of forcing installing packages without the architecture appended to the package name even if you are cross-compiling. This is not yet covered in the PR with a solution, I commented above that this can be workarounded with something like this:

```
Apt(self, arch_names={}).install(["archless-package"], update=True, check=True)
```

Passing `arch_names` as an empty dictionary in the constructor will make that the name resolution skips adding the architecture for all cases (cross-compiling and not). But that is ugly and probably we want to pass a new argument to cover this case. That workaround I think would close https://github.com/conan-io/conan/issues/12320 as then you also could mix archless packages with packages with arch in the same recipe without having to write a new recipe for the archless ones which I think would be more convenient. Am I missing something and maybe this would not be enough to cover the case?

Thanks a lot.

memsharded (approved):

jcar87 (approved):
Looks like a good default with the possibility of adding fine-grained control in the future :+1:

Issue #12334, opened 2022-10-18
[question] conan 2.0 migration path for `env_info` and `user_info`


prince-chrismc:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

This came up in https://github.com/conan-io/conan-center-index/issues/13503, but it's also in a few other recipes

Currently in Conan 1.x we needed to use the `env_info` and `user_info` these are removed in Conan 2.0 and result is hard errors when they are present. https://docs.conan.io/en/latest/migrating_to_2.0/recipes.html#the-package-info-method.

In ConanCenter, this means we need to have both the old and new while down stream recipes are not yet updated. so we are adding a check for the `conan_version` to ensure the problematic code is not called in the new version.

https://github.com/conan-io/conan-center-index/blob/412a59575357fb352b14caa29312263c4081897e/recipes/strawberryperl/all/conanfile.py

```py
from conan import ConanFile, conan_version

    def package_info(self):
        # TODO remove once conan v2 is the only support and recipes have been migrated
        if Version(conan_version) < "2.0.0-beta":
            bin_path = os.path.join(self.package_folder, "bin")
            self.env_info.PATH.append(bin_path)
            self.user_info.perl = perl_path
```

czoido:
Thanks @prince-chrismc for reporting this. I think we could:

- Implement `env_info` and `user_info` as fake attributes in Conan 2.0 so the error is not thrown
- [Output a warning message](https://github.com/conan-io/conan/issues/12345) in 2.0 to inform users that those attributes are deprecated (we should also include other things like .names that are still there to provide compatibility). This is in order to reduce the risk of breaking users when we pass from Conan 2.0 to 2.X

I'm marking this for the next 2.0-beta5

uilianries:
Okay, so we no longer need to check Conan version + env_info, correct?

Asking because PRs in CCI started to use it and I would like to avoid asap.

czoido:
@prince-chrismc I'm closing this as https://github.com/conan-io/conan/pull/12351 was merged (will be released in beta5) and now Conan 2.0 recipes won't fail if they have `env_info` and `user_info` although it will have no effect at all

prince-chrismc:
🚀

Issue #12345, opened 2022-10-20
[feature] Add warning about things that will dissapear in 2.X from 2.0


czoido:
We currently provide some "fake" attributes in the ConanFile for 2.0 in order to not break compatibility with 1.X recipes. For example `.names` and `.filenames`. But these attributes will dissapear at some point in Conan 2.X and we want users to be aware that these things will throw errors before they get them. The idea is start to output warning messages when those attributes are used and make them errors if `required_conan_version>=2.0`

Related to: https://github.com/conan-io/conan/issues/12334

franramirez688:
Closes by https://github.com/conan-io/conan/pull/12351

PR #12351, opened 2022-10-20
Implement `env_info` and `user_info` as fake attributes in Conan 2.0 so the error is not thrown

czoido:
Changelog: Implement `env_info` and `user_info` as fake attributes in Conan 2.0 so the error is not thrown
Docs: add note to migration guide

Closes: https://github.com/conan-io/conan/issues/12334

TODO: [Output a warning message](https://github.com/conan-io/conan/issues/12345) in 2.0 to inform users that those attributes are deprecated (we should also include other things like .names that are still there to provide compatibility). This is in order to reduce the risk of breaking users when we pass from Conan 2.0 to 2.X

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -1,9 +1,11 @@
 import os
+from collections import defaultdict
 from pathlib import Path
 
 from conan.api.output import ConanOutput
 from conans.client.subsystems import command_env_wrapper
 from conans.errors import ConanException
+from conans.model.build_info import MockInfoProperty
 from conans.model.conf import Conf
 from conans.model.dependencies import ConanFileDependencies
 from conans.model.layout import Folders, Infos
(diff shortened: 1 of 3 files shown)


conans/model/conan_file.py
```diff
@@ -255,6 +258,12 @@ def __init__(self, display_name=""):
         self.folders = Folders()
         self.cpp = Infos()
 
+        # LEGACY 1.X fields, can be removed in 2.X
+        self.deps_cpp_info = defaultdict(lambda: MockInfoProperty("ConanFile.deps_cpp_info"))
```
memsharded: I am not sure about these, these are not write-attributes, they are read-aggregations. I think this needs to disappear in 1.X. One thing is not wanting to break external consumers that are in 1.X, but this is being a consumer, and if it has to work in 2.0, this shouldn't be part of it.

memsharded (approved):

Issue #12372, opened 2022-10-21
[bug] 2.0 check_component_requires and ConanFileDependencies


beenjohn:
In relation to https://github.com/conan-io/conan/issues/12027 and https://github.com/conan-io/conan/pull/12280:

Similar to the dependency graph in issue #12027 I have a transitive dependency identical to a direct dependency. It looks to me like new function conans.model.build_info.check_component_requires is checking for direct dependencies conflicts with cpp_info via the ConanFileDependencies interface. 

The issue i'm seeing is a I have:

```
...
self.requires("packageA/0.0.1")    # This package also requires boost/1.76.0
self.requires("boost/1.76.0")
...
self.cpp_info.component["component"].requires(["boost::boost"])
```
Output: `ERROR: packageB/0.0.1: required component package 'boost::' not in dependencies`

I'm a little unclear of exactly where the traits for the dependencies are set, but the transitive dependency appears to be overwriting the list entry for the direct entry. From debugging, conanfile.dependencies["boost/1.76.0"].direct = False which is getting thrown out by the filter call in check_component_requires.

I'm not sure if this is expected behavior or not, but it seems like the entries conanfile.dependencies should |= some of the traits (run, libs, headers, direct, etc.) if there is a duplicate dependency.

### Environment Details (include every applicable attribute)
`conan profile show`
```
Host profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=9
os=Linux
os.distro=Ubuntu20.04
[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True
```
`python3 --version`
`Python 3.8.10`

beenjohn:
From further inspection, this change to requires.aggregate(...) fixes the issue for me:
```
--- requires.py	2022-10-24 17:59:55.657001602 +0000
+++ requires.py 2022-10-24 18:00:53.529600419 +0000
@@ -227,6 +227,7 @@
         self.run = self.run or other.run
         self.visible |= other.visible
         self.force |= other.force
+        self.direct |= other.direct
         if not other.test:
             self.test = False  # it it was previously a test, but also required by non-test
         # TODO: self.package_id_mode => Choose more restrictive?

```
Is this something I should submit a PR for or is the requires.direct trait intentionally left out of the aggregate function?

memsharded:
Hi @beenjohn 

[...]

PR #12388, opened 2022-10-24
Bug/requires aggregation with transitive deps

beenjohn:
Changelog: Bugfix: Fix an issue with requires.direct being overwritten based on requirement ordering in recipe.
Docs: https://github.com/conan-io/docs/pull/XXXX (TBD)

See: https://github.com/conan-io/conan/issues/12372

### DIFF
--- a/conans/model/requires.py
+++ b/conans/model/requires.py
@@ -232,7 +232,7 @@ class Requirement:
         self.run = self.run or other.run
         self.visible |= other.visible
         self.force |= other.force
-        # self.direct |= other.direct
+        self.direct |= other.direct
         if not other.test:
             self.test = False  # it it was previously a test, but also required by non-test
         # TODO: self.package_id_mode => Choose more restrictive?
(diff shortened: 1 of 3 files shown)


memsharded (commented):
I was reviewing this, modifying the test, so it didn't modify a previous test (to keep its checks), but instead adding a new, clean one.
But I tried to remove the fix, to see if the test was failing, but apparently it is passing. It would be great to have a test that fails without the fix. I can also try to have a look later

memsharded:
> But I tried to remove the fix, to see if the test was failing, but apparently it is passing. It would be great to have a test that fails without the fix. I can also try to have a look later

I managed to make it fail in one of our "core graph tests", pushed changes, I think this will be enough, so I will merge it when green. Thanks for raising and fixing this bug!

beenjohn:
@memsharded Thanks for improving the test. I've reviewed your changes to improve future PRs. I'm glad to be able to help where I can. We're getting close to switching over to 2.X in our build system, and it's already improving our development processes greatly.

PR #12440, opened 2022-11-01
Swap package_id()-validate() order and fix validate_build()

memsharded:
Changelog: Omit
Docs: Omit

Several internal changes:
- Swap ``package_id()`` and ``validate()`` order, so it behaves like in 2.0, and allows ``validate()`` of a header-only to work, because ``package_id()`` has not removed settings yet.
- Copying the ``ConanInfo`` information shouldn't copy the ``invalid`` flag, otherwise all package_id can become invalid

The risk of this is medium

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@ Decentralized, open-source (MIT), C/C++ package manager.
 - Homepage: https://conan.io/
 - Github: https://github.com/conan-io/conan
 - Docs: https://docs.conan.io/en/latest/
-- Slack: https://cpplang-inviter.cppalliance.org/ (#conan channel)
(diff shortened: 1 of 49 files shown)


memsharded:
I am extracting https://github.com/conan-io/conan/pull/12536 for the ``should_build`` which seems to make sense irrespetive of the order of package_id/validate

memsharded:
At the moment this no longer is necessary. Latest changes, going back to ``self.settings`` in ``validate()`` instead of ``self.info`` is good. Validating a header-only is still possible, because ``self.info.clear()`` does not affect the ``validate()`` with ``check_min_cpsstd()`` that internally checks ``self.settings.compiler.cppstd``

Issue #12236, opened 2022-10-04
[bug] Bazel generator add tools and build dependencies as bazel deps 


a4z:
### Environment Details (include every applicable attribute)
  * Operating System+version: Any
  * Compiler+version: Not relevant
  * Conan version: 1.51.2 and 1.52.0 , but maybe all since the bazel generator exists
  * Python version: 3.10.6 , but not relevant

### Steps to reproduce (Include if Applicable)

Install / use a package like `mongo-c-driver` and use the bazel generator to get a bazel file

```
    deps = [
        # do not sort
        ":mongoc-static-1.0_precompiled",
        ":bson-static-1.0_precompiled",
        "@openssl",
        "@snappy",
        "@zlib",
        "@zstd",
        "@icu",
        "@pkgconf",
    ],
```

`"@pkgconf"` is a tool and build requirement. 

Please have a look at the recipe:
https://github.com/conan-io/conan-center-index/blob/5225f30ee00d5f295419dc632cd29d4491a9d394/recipes/mongo-c-driver/all/conanfile.py#L94

Such build or tool requirements should not be on the list of dependencies.  They might not even exist in case of cross-compiling.

czoido:
Hi @a4z,

Thanks for reporting, I'll mark as an issue and we will have a look for the next release.

a4z:
I just rebuilt mongo-c-driver and still got `@pkgconf` in the final BUILD file
Seems the complexity of this issue is not that low ...  

Huge thanks for keeping up on the issue!  ❤️ Looking forward to 1.55. release!

PR #12445, opened 2022-11-02
Do not include build-context dependencies in Bazel BUILD file

### LINKED ISSUE #12444: [bug] Bazel BUILD file should not include build-context dependencies
This relates to my [previous change](https://github.com/conan-io/conan/pull/10686) which added transitive dependencies to Bazel `BUILD` files. When a dependency which has build-time dependencies is added, it adds the build-context dependencies as well, even though these are only needed for building the dependency, not for using it.

### Environment Details
  * Operating System+version: Ubuntu 21.04
  * Compiler+version: GCC 10.4
  * Conan version: 1.53.0
  * Python version: Python 3.10.6

Add a dependency on libCurl and run the Bazel generator. The first time it runs, it generates dependencies like this:

```
    deps = [
        # do not sort
    ":curl_precompiled",
    "@openssl",
    "@zlib",
    "@c-ares",
    "@libtool",
    "@pkgconf",
    "@automake",
    "@autoconf",
    "@m4",
    ],
```

The dependencies such as `autoconf` and `m4` are things used for building the libCurl library but are not dependencies which are needed to link against it, so should not be included here.

Running a second time without any other changes will, now the libCurl library has already been built, then generate the appropriate deps:

```
    deps = [
        # do not sort
    ":curl_precompiled",
    "@openssl",
    "@zlib",
    "@c-ares",
    ],
```

PR that fixes this at #12445

sburton84:
Changelog: Bugfix: Do not include build-context dependencies in Bazel BUILD file
Docs: omit

Fixes #12444

### DIFF
--- a/conan/tools/google/bazeldeps.py
+++ b/conan/tools/google/bazeldeps.py
@@ -147,7 +147,8 @@ class BazelDeps(object):
 
         dependencies = []
         for req, dep in dependency.dependencies.items():
-            dependencies.append(dep.ref.name)
+            if not dep.is_build_context:
+                dependencies.append(dep.ref.name)
 
         shared_library = dependency.options.get_safe("shared") if dependency.options else False
 
--- a/conans/test/unittests/tools/google/test_bazeldeps.py
+++ b/conans/test/unittests/tools/google/test_bazeldeps.py
@@ -14,6 +14,7 @@ from conans.model.ref import ConanFileReference
 from conans.test.utils.mocks import MockOptions
 from conans.test.utils.test_files import temp_folder
 from conans.util.files import save
+from conans.client.graph.graph import CONTEXT_BUILD
 
 
 class MockConanFileDeps(ConanFile):
@@ -95,8 +96,7 @@ def test_bazeldeps_get_lib_file_path_by_basename():
                 assert 'linkopts = ["-lsystem_lib1"],' in dependency_content
             assert 'deps = [\n        # do not sort\n    \n    ":liblib1.a_precompiled",' in dependency_content
 
-
-def test_bazeldeps_dependency_transitive():
+def create_bazeldeps_with_transitive_dependency(build = False):
     # Create main ConanFile
     conanfile = ConanFile(Mock(), None)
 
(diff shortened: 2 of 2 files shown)


franramirez688 (commented):
Thanks for contributing @sburton84 

I think it's fine the change (as far as I understand your issue) but I believe that could be great to have a dedicated integration test checking the dependencies. Could you add one here `conans/test/integration/toolchains/google/test_bazel.py`? If you have any kind of problem, I can help you for sure.

czoido:
Related to: https://github.com/conan-io/conan/issues/12236

sburton84:
I've added a test for this now.

I'm also not sure whether it's better for the check to be `if not dep.is_build_context` or `if not req.build`? Are these equivalent or is there any subtle difference that could cause these to return different values?

franramirez688:
> I'm also not sure whether it's better for the check to be if not dep.is_build_context or if not req.build? Are these equivalent or is there any subtle difference that could cause these to return different values?

No, they are not. `dep.is_build_context` returns true or false if that dependency (`dep`) belongs to the build context, i.e., if it's a `tool_require`, and `req.build` returns all the build requirements belonging to that requirement (`req`). So that's fine as it is I think. Please @memsharded correct me if I'm wrong.

memsharded:
> No, they are not. dep.is_build_context returns true or false if that dependency (dep) belongs to the build context, i.e., if it's a tool_require, and req.build returns all the build requirements belonging to that requirement (req). So that's fine as it is I think. Please @memsharded correct me if I'm wrong.

``req.build`` would be just a boolean trait of the requirement ``req``, being it True or False. It can also be used to filter the requirements in the build context, as from the point of view of the consumer, the ``build`` trait is also transitive. But using ``dep.is_build_context`` also seems good and explicit.

My only concern is that this might break some users, that relied (incorrectly) on this. But as this is experimental, this is totally fine, lets wait for their feedback and inform them. This generator should be as the others:
- by default, do not generate information for ``tool_requires``
- Allow, if necessary a method to generate information for ``tool_requires``, like ``BazelDeps.build_context_deps`` or something similar to what ``CMakeDeps`` does
- Check what information and how can ``tool_requires`` provide to Bazel

memsharded:
Superseded by https://github.com/conan-io/conan/pull/12543, many thanks for contributing!

PR #12457, opened 2022-11-03
automatically add msys2/usr/bin path

### LINKED ISSUE #12110: [bug] self.win_bash and msys2, internal commands not found
While testing this PR: https://github.com/conan-io/conan/pull/12095

I saw that I couldn't run any command belonging to msys2 (installed packages there) located in `/usr/bin`, but there were tests that work because they use the `@pytest.mark.tool_msys2` that is putting the `C:/msys64/usr/bin` in the path in the `conftest.py`.

I think that is very confusing because a recipe trying to run a command like `autoconf` in `msys2` doesn't find it, nevertheless, we have 2 conf to "set up" the environment: `tools.microsoft.bash:subsystem` and `tools.microsoft.bash:path`. So we could:

0) Ignore this and expect the user to have the `c:/mysys2/../usr/bin` in the path.
1) Somehow declare the PATH to the folder "/usr/bin" before running the command
2) Have a conf for that path? Looks like too much.

PD: I haven't tested it this happens in other subsystems.
PD: This is failing in develop2 too.

memsharded:
Changelog: Feature: Automatically add the msys2 ``usr/bin`` folder where ``bash.exe`` is to the PATH.
Docs: Omit

Close https://github.com/conan-io/conan/issues/12110

### DIFF
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -75,6 +75,9 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
         # Configure MSYS2 to inherith the PATH
         msys2_mode_env = Environment()
         _msystem = {"x86": "MINGW32"}.get(conanfile.settings.get_safe("arch"), "MINGW64")
+        # https://www.msys2.org/wiki/Launchers/ dictates that the shell should be launched with
+        # - MSYSTEM defined
+        # - CHERE_INVOKING is necessary to keep the CWD and not change automatically to the user home
         msys2_mode_env.define("MSYSTEM", _msystem)
         msys2_mode_env.define("MSYS2_PATH_TYPE", "inherit")
         # So --login do not change automatically to the user home
@@ -96,6 +99,8 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
                                                 accepted_extensions=("sh", ))
     wrapped_user_cmd = _escape_windows_cmd(wrapped_user_cmd)
 
+    # according to https://www.msys2.org/wiki/Launchers/, it is necessary to use --login shell
+    # running without it is discouraged
     final_command = '{} --login -c {}'.format(wrapped_shell, wrapped_user_cmd)
     return final_command
 
--- a/conans/client/subsystems.py
+++ b/conans/client/subsystems.py
@@ -77,6 +77,8 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
         _msystem = {"x86": "MINGW32"}.get(conanfile.settings.get_safe("arch"), "MINGW64")
         msys2_mode_env.define("MSYSTEM", _msystem)
         msys2_mode_env.define("MSYS2_PATH_TYPE", "inherit")
+        # So --login do not change automatically to the user home
+        msys2_mode_env.define("CHERE_INVOKING", "1")
         path = os.path.join(conanfile.generators_folder, "msys2_mode.bat")
         # Make sure we save pure .bat files, without sh stuff
         wb, conanfile.win_bash = conanfile.win_bash, None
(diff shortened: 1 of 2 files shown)


czoido (approved):

jcar87 (approved):

conans/client/subsystems.py
```diff
@@ -77,8 +77,14 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
         _msystem = {"x86": "MINGW32"}.get(conanfile.settings.get_safe("arch"), "MINGW64")
         msys2_mode_env.define("MSYSTEM", _msystem)
         msys2_mode_env.define("MSYS2_PATH_TYPE", "inherit")
+        # Automatically add the current bash.exe path to the PATH
+        if os.path.isabs(shell_path):
+            msys2_mode_env.prepend_path("PATH", os.path.dirname(shell_path))
         path = os.path.join(conanfile.generators_folder, "msys2_mode.bat")
+        # Make sure we save pure .bat files, without sh stuff
+        wb, conanfile.win_bash = conanfile.win_bash, None
```
memsharded, line 87: I have to say that I don't love this approach, I tried different alternatives: passing an argument, forcing the .bat extension to override the configuration... I didn't love them either, and all had other potential downsides too, so this is the best I could come up with

memsharded:
Relevant: https://www.msys2.org/wiki/Launchers/

conans/client/subsystems.py
```diff
@@ -91,7 +94,7 @@ def _windows_bash_wrapper(conanfile, command, env, envfiles_folder):
                                                 accepted_extensions=("sh", ))
     wrapped_user_cmd = _escape_windows_cmd(wrapped_user_cmd)
 
-    final_command = '{} -c {}'.format(wrapped_shell, wrapped_user_cmd)
+    final_command = '{} --login -c {}'.format(wrapped_shell, wrapped_user_cmd)
```
memsharded, line 104: Trying with --login, lets see what happens

memsharded: It is weird, the ``test_autotools_bash_complete`` fails with 
```
aclocal-1.16: error: 'configure.ac' is required
```
when using --login, I don't know how it is connected

jcar87: mm it could be related - that error would come from `autoconf`, which is being called in that test - however, `configure.ac` is also provided by that test. I suspect `--login` is changing the current working directory, so it's probably defaulting to the "user home", and thus `autoconf` no longer finds `configure.ac` in the cwd

memsharded: Ok. The ``--login`` is effectively doing a ``cd ~`` to my current user home, so it doesn't find the project files expected to be. This is the fragility of the ``--login``, that it depends on the user running the command and what they might have configured in their ``~/.bash_profile, ~/.bash_login, and ~/.profile``

jcar87: I think the setting to cwd to the current working directory is more of a shell behaviour rather than a user configuration files issue.

I can understand the concerns of loading `~/.bash_profile` and similar, but I have a suspicion that in _practical_ terms, in most other scenarios if Conan is being invoked from a shell that already loaded those files anyways and the environment state is propagated to all the subprocesses. So on Linux, macOS, on Windows if Conan is invoked from one of the subsystem shells - those files (~/.bash_profile, ~/.bash_login, and ~/.profile) are already loaded and exposed.

I think the problem here is unique in the sense that Conan (the Python interpreter running `conan`) may be invoked say from a windows command prompt or windows powershell, but it is running some commands by invoking a `bash.exe` directly, and not the command prompt wrappers they recommend, so the shell is misconfigured (and thus `/usr/bin` is missing). My concern by automatically adding `msys2/usr/bin` is that there may be other things in the future that we realise are missing or not set up, and users may have a difficult time reproducing or troubleshooting the issues because when they try to run commands inside their msys2 shells, things will work for them.

bash will load `/etc/profile`, followed by the user profile files. It does have a flag `--no-profile` to inhibit the behaviour, but I'm not sure whether that applies to all files (`/etc` and the user ones), or just the user ones.

memsharded: It seems that using ``--login`` and defining ``CHERE_INVOKING`` at least the test suite is green.

PR #12464, opened 2022-11-04
Stop mutating CMAKE_PREFIX_PATH in config file generated by CMakeDeps

### LINKED ISSUE #12237: [bug] self.cpp_info.builddirs default value causes CMAKE_PREFIX_PATH to be mutated during find_package

jcar87:
Changelog: Bugfix: Fix issue where find_program does not work for a tool requirement in the build context, when the dependency is also a regular requirement in the host context. 
Docs: Omit

Close https://github.com/conan-io/conan/issues/12237

### Notes
This fixes a bug where the following two invocations were dependent on the order of the calls. As investigated in https://github.com/conan-io/conan/issues/12237, there does not appear to be a case where the generated config files by `CMakeDeps` need to alter the value of `CMAKE_PREFIX_PATH` - where needed, that is handled by `CMakeToolchain`.

For the following, when cross building, in both instances the following two invocations should find the binary executable from the package in the build context. Before this fix, calling `find_program` after `find_package` would result in the binary executable being found in the host context (which may not run at all).

```
        find_package(foobar CONFIG REQUIRED)
        find_program(FOOBIN_EXECUTABLE foobin)
```
and this
```
        find_program(FOOBIN_EXECUTABLE foobin)
        find_package(foobar CONFIG REQUIRED)
```

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -88,7 +88,6 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
 
         # FIXME: What is the result of this for multi-config? All configs adding themselves to path?
         set(CMAKE_MODULE_PATH {{ '${' }}{{ pkg_name }}_BUILD_DIRS{{ config_suffix }}} {{ '${' }}CMAKE_MODULE_PATH})
-        set(CMAKE_PREFIX_PATH {{ '${' }}{{ pkg_name }}_BUILD_DIRS{{ config_suffix }}} {{ '${' }}CMAKE_PREFIX_PATH})
 
         {% if not components_names %}
 
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
@@ -1141,7 +1141,10 @@ def test_find_program_for_tool_requires():
     client = TestClient()
 
     conanfile = textwrap.dedent("""
+        import os
+
         from conan import ConanFile
+        from conan.tools.files import copy
         class TestConan(ConanFile):
             name = "foobar"
             version = "1.0"
@@ -1150,8 +1153,8 @@ def test_find_program_for_tool_requires():
             def layout(self):
                 pass
             def package(self):
-                self.copy(pattern="lib*", dst="lib")
-                self.copy(pattern="*bin", dst="bin")
+                copy(self, pattern="lib*", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"))
+                copy(self, pattern="*bin", src=self.build_folder, dst=os.path.join(self.package_folder, "bin"))
     """)
 
     host_profile = textwrap.dedent("""
@@ -1181,6 +1184,8 @@ def test_find_program_for_tool_requires():
                 "build_profile": build_profile
                 })
 
+    xxx = client.get_default_build_profile()
+
     client.run("create . -pr:b build_profile -pr:h build_profile")
     client.run("create . -pr:b build_profile -pr:h host_profile")
 
@@ -1201,9 +1206,8 @@ def test_find_program_for_tool_requires():
     """)
 
     cmakelists_consumer = textwrap.dedent("""
-        set(CMAKE_CXX_COMPILER_WORKS 1)
         cmake_minimum_required(VERSION 3.15)
-        project(Hello LANGUAGES CXX)
+        project(Hello LANGUAGES NONE)
         find_package(foobar CONFIG REQUIRED)
         find_program(FOOBIN_EXECUTABLE foobin)
         message("foobin executable: ${FOOBIN_EXECUTABLE}")
--- a/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
@@ -1185,7 +1185,7 @@ def test_find_program_for_tool_requires():
     client.run("create . -pr:b default -pr:h host_profile")
 
     conanfile_consumer = textwrap.dedent("""
-        from conans import ConanFile
+        from conan import ConanFile
         from conan.tools.cmake import cmake_layout
         class PkgConan(ConanFile):
             settings = "os", "arch", "compiler", "build_type"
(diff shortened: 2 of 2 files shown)


conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1129,3 +1129,100 @@ def test_cmake_toolchain_vars_when_option_declared():
...
+            def layout(self):
+                pass
+            def package(self):
+                self.copy(pattern="lib*", dst="lib")
+                self.copy(pattern="*bin", dst="bin")
+    """)
+
+    host_profile = textwrap.dedent("""
+    [settings]
+        os=Linux
```
memsharded, line 1162: If this is a functional test, this will probably be broken in Win/OSX? need a ``pytest.skip``?

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1129,3 +1129,100 @@ def test_cmake_toolchain_vars_when_option_declared():
...
+        from conans import ConanFile
+        class TestConan(ConanFile):
+            name = "foobar"
+            version = "1.0"
+            settings = "os", "arch", "compiler", "build_type"
+            exports_sources = "*"
+            def layout(self):
+                pass
+            def package(self):
+                self.copy(pattern="lib*", dst="lib")
```
memsharded: Use modern ``copy()`` for merging to 2.0?

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1129,3 +1129,100 @@ def test_cmake_toolchain_vars_when_option_declared():
...
+def test_find_program_for_tool_requires():
+    """Test that the same reference can be both a tool_requires and a regular requires,
+    and that find_program (executables) and find_package (libraries) find the correct ones
+    when cross building.
+    """
+
+    client = TestClient()
+
+    conanfile = textwrap.dedent("""
+        from conans import ConanFile
```
memsharded:
```suggestion
        from conan import ConanFile
```
So it merges to 2.0

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1129,3 +1129,100 @@ def test_cmake_toolchain_vars_when_option_declared():
...
+                "foobin": "",
+                "host_profile": host_profile,
+                "default": build_profile
+                })
+
+    client.run("create . -pr:b default -pr:h default")
+    client.run("create . -pr:b default -pr:h host_profile")
+
+    conanfile_consumer = textwrap.dedent("""
+        from conans import ConanFile
```
memsharded:
```suggestion
        from conan import ConanFile
```

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
jcar87: This doesn't really test functionality that is platform specific - it's "cmake only" - figured that by fixing the profiles and setting `CMAKE_CXX_COMPILER_WORKS` in CMake (I've seen this done in other tests, this would enable us to runt this test on all platforms, or at least to write it in a way that is more platform agnostic

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
jcar87: Will do! Will update shortly. The test as it is needs other changes to work on 2.0

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
memsharded: I thought that providing a profile like that for Windows would fail at some point, like the armv8 architecture, or something like that. Probably can be simplified and left mostly with the platform specific ones for testing, but I am ok with it as-is

jcar87: oh, the host profile could be listing a random architecture here that is not x86_64 - this test is not exercising anything that is platform specific, hence I feel it's safe to "fix" the profiles like many other tests do. A different approach could be tried if we want the profiles to match something that makes more sense on the machine the test is running on.

memsharded (approved):

SpaceIm:
Should I close https://github.com/conan-io/conan/pull/12229?

conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -88,7 +88,6 @@ def template(self):
 
         # FIXME: What is the result of this for multi-config? All configs adding themselves to path?
         set(CMAKE_MODULE_PATH {{ '${' }}{{ pkg_name }}_BUILD_DIRS{{ config_suffix }}} {{ '${' }}CMAKE_MODULE_PATH})
```
SpaceIm, line 90: this one is harmless but honestly a config file manipulating CMAKE_MODULE_PATH is unusual, I think it's too intrusive.

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1129,3 +1129,104 @@ def test_cmake_toolchain_vars_when_option_declared():
     t.run_command("cmake -S . -B build/ -DBUILD_SHARED_LIBS=ON")
     assert "mylib target type: SHARED_LIBRARY" in t.out
     assert "mylib position independent code: ON" in t.out
+
+
+#@pytest.mark.tool_cmake()
```
czoido:
```suggestion
@pytest.mark.tool_cmake
```

czoido (approved):

franramirez688 (approved):

PR #12505, opened 2022-11-10
Fix passing build jobs number (-j) to Autotools build helper

### LINKED ISSUE #12296: [bug] `conan.tools.gnu.Autotools`: number of processors can't be customized through `make` & `install` methods
  * Compiler+version: AppleClang 14.0
  * Conan version: 1.53.0
  * Python version: 3.10.7

- checkout https://github.com/conan-io/conan-center-index/pull/13447/commits/0f24d242d38d0674c5fd5bd5073cb10c3831c919, where `-j1` is passed to args of Autotools `install` method (this library can use several processors during build, but there might be race conditions during installation, so `-j1` is injected during installation only).
- see resulting install command, it's of the form: `make install -j1 DESTDIR=/Users/jenkins/w/prod/BuildSingleReference/.conan/data/coin-utils/2.11.4/_/_/package/e967b6f7ee847ba2cdb4c6a8f494f465b893beb3 -j4`

```
ERROR: coin-utils/2.11.4: Error in package() method, line 137
	autotools.install(args=["-j1", f"DESTDIR={unix_path(self, self.package_folder)}"])
	ConanException: Error 2 while executing make install -j1 DESTDIR=/Users/jenkins/w/prod/BuildSingleReference/.conan/data/coin-utils/2.11.4/_/_/package/e967b6f7ee847ba2cdb4c6a8f494f465b893beb3 -j4
```

Issue here: https://github.com/conan-io/conan/blob/3bfb9ae191ded2e4fc4a31be8910883e23dc0f84/conan/tools/gnu/autotools.py#L44-L55

conan tests whether `-j` was passed to `make_args` of `AutotoolsToolchain`, but args passed by user through `install` or `make` methods are not checked.

### LINKED ISSUE #12494: [feature] AutotoolsToolchain should support install_args analogous to make_args for consistency.
Hi @czoido, Hi @memsharded 

Thank you for quickly addressing the primary issue in #12470.  This feature request captures the second suggestion proffered there:
 

> Some Makefiles don't do well with parallel execution (-j > 1) during the install phase, and thus it would be nice if AutotoolsToolchain supported an additional attribute called "install_args" analogous to "make_args" as the only place one can disable parallel execution today is via make_args and that affects both the "make" and "install" Autotools methods.

If one tries to specify -j1 via autotools.install(args=["-j1"]), they suffer two consequences:

1. The -j1 has no effect because the resulting make invocation has both -j1 and -jN and the latter wins.
2.  The default DESTDIR value is lost unless it is explicitly added to the args value.

Thanks

czoido:
Changelog: Fix: Give priority to `-j` argument set by user in recipe over the default set by conan in `Autotools` build helper.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/12494
Closes: https://github.com/conan-io/conan/issues/12296

### DIFF
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -49,7 +49,7 @@ class Autotools(object):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        jobs_already_passed = re.search(r"-j\d+", join_arguments([str_args, str_extra_args]))
+        jobs_already_passed = re.search(r"-j(\d+)", join_arguments([str_args, str_extra_args]))
         if not jobs_already_passed and "nmake" not in make_program.lower():
             njobs = build_jobs(self._conanfile)
             if njobs:
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -1,4 +1,5 @@
 import os
+import re
 
 from conan.tools.build import build_jobs
 from conan.tools.files.files import load_toolchain_args
@@ -48,7 +49,8 @@ class Autotools(object):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        if "-j" not in (str_args + str_extra_args) and "nmake" not in make_program.lower():
+        jobs_already_passed = re.search(r"-j\d+", join_arguments([str_args, str_extra_args]))
+        if not jobs_already_passed and "nmake" not in make_program.lower():
             njobs = build_jobs(self._conanfile)
             if njobs:
                 jobs = "-j{}".format(njobs)
@@ -58,7 +60,7 @@ class Autotools(object):
     def install(self, args=None, target="install"):
         args = args if args else []
         str_args = " ".join(args)
-        if "DESTDIR" not in str_args:
+        if "DESTDIR=" not in str_args:
             args.insert(0, "DESTDIR={}".format(unix_path(self._conanfile, self._conanfile.package_folder)))
         self.make(target=target, args=args)
 
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -49,7 +49,7 @@ class Autotools(object):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        jobs_already_passed = re.search(r"(^-j\d+\s+)|(\W-j\d+\s*)", join_arguments([str_args, str_extra_args]))
+        jobs_already_passed = re.search(r"(^-j\d+)|(\W-j\d+\s*)", join_arguments([str_args, str_extra_args]))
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/autotools.py
```diff
@@ -48,16 +48,18 @@ def make(self, target=None, args=None):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        if "-j" not in str_args and "nmake" not in make_program.lower():
+        if "-j" not in (str_args + str_extra_args) and "nmake" not in make_program.lower():
```
SpaceIm: This PR may be a good opportunity to provide a more robust test of `-j<number>` in args with regex. Currently if args contains something like `DESTDIR=/user/smith-john/.conan/...`, conan doesn't inject `-j<njobs>`.

conan/tools/gnu/autotools.py
```diff
@@ -48,16 +49,19 @@ def make(self, target=None, args=None):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        if "-j" not in str_args and "nmake" not in make_program.lower():
+        jobs_already_passed = re.search(r"-j(\d+)", join_arguments([str_args, str_extra_args]))
```
memsharded: I am not sure if this should be expected. The moment Conan has a standard, defined mechanism to customize the paralelism, this shouldn't be taken into account:
- The profile ``njobs`` should have priority to what the recipe says. In general, anything defined in the profiles should have priorities
- Why would a recipe hardcoded the number of cores it is using in a custom ``arg`` in the helper?

SpaceIm: Install target doesn't always support multiprocessing in Makefiles (there are examples in conan-center), so recipe must ensure robust installation regardless of njobs in profile.

SpaceIm: it's still a little bit fragile, a pattern like `/users/foo-j47/...` may be detected.

a regex like this one `(^-j\d+$)|(^-j\d+\s+)|(\s+-j\d+$)|(\s+-j\d+\s+)` for example?

conan/tools/gnu/autotools.py
```diff
@@ -48,16 +49,19 @@ def make(self, target=None, args=None):
         str_args = self._make_args
         str_extra_args = " ".join(args) if args is not None else ""
         jobs = ""
-        if "-j" not in str_args and "nmake" not in make_program.lower():
+        jobs_already_passed = re.search(r"(^-j\d+)|(\W-j\d+\s*)", join_arguments([str_args, str_extra_args]))
```
SpaceIm, line 52: it can still find unusal patterns like `/users/-j1/` or `-j8bdfsdb`

czoido: Thanks for the suggestion, I know this does not cover every hypothetical case but I wanted to leave the regex as simple as possible...

memsharded (approved):

franramirez688 (approved):

PR #12507, opened 2022-11-10
second approach to correct require equality

memsharded:
Complement to https://github.com/conan-io/conan/pull/12506, implementing the approach of changing the equality of ``Requirement``, but not ``RecipeReference``


czoido:
Closed in favor of https://github.com/conan-io/conan/pull/12506

PR #12355, opened 2022-10-20
Honor conf `tools.gnu:pkg_config` in `CMakeToolchain`

SpaceIm:
Changelog: (Feature): Ensure that `CMakeToolchain` will enforce usage of pkgconf executable set in `tools.gnu:pkg_config` config.
Docs: https://github.com/conan-io/docs/pull/XXXX

partial implementation of https://github.com/conan-io/conan/issues/12354

- [x] Refer to the issue that supports this Pull Request.
- [ ] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [ ] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 

<sup>**Note:** By default this PR will skip the slower tests and will use a limited set of python versions. Check [here](https://github.com/conan-io/conan/blob/develop/.github/PR_INCREASE_TESTING.md) how to increase the testing level by writing some tags in the current PR body text.</sup>

Issue #12354, opened 2022-10-20
[feature] Ensure that `tools.gnu:pkg_config` is discoverable in all conan v2 build helpers


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Currently `tools.gnu:pkg_config` is only used in `PkgConfig` class (and it's explicitly described like this in `conan config list`). It would be a good idea to extend capabilities of this conf so that when it's set, it ensures that this specific executable is found by standard usage of CMake, Autotools, Meson etc...

For example in CMake, standard usage is to call `find_package(PkgConfig)`. There is a documented CMake variable `PKG_CONFIG_EXECUTABLE` which can be overridden to force this module file to use executable pointed by this variable: https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#variables-affecting-behavior

eli-schwartz:
> There is a documented CMake variable `PKG_CONFIG_EXECUTABLE` which can be overridden to force this module file to use executable pointed by this variable

Note that all 3 of CMake, Meson, and autotools will respect the environment variable `$PKG_CONFIG` -- cmake will override that with the cmake variable ~~prefixed by `CMAKE_`~~ suffixed by `_EXECUTABLE` because I can't read and also because why ever not specify that it is an executable, and Meson will override that if you have a machine file whose `[binaries]` section maps a value for pkg-config.

But just using the environment variable is simple and consistent.

SpaceIm:
Yes that's probably the easiest solution. The problem is that in conan it would likely be something handled by `VirtualBuildEnv`, but for some reason I don't understand it's disabled by default in conan v1 client.

note: I'm not sure there is a `CMAKE_PKG_CONFIG` cmake variable, it's not documented

memsharded:
> Yes that's probably the easiest solution. The problem is that in conan it would likely be something handled by VirtualBuildEnv, but for some reason I don't understand it's disabled by default in conan v1 client.

It is disabled by default in Conan 1.X, because it is a new thing that could break users. It is opt-in in two ways:
- Declaring ``generators = "VirtualBuildEnv"`` in the recipe
- Using the ``tools.env.virtualenv:auto_use=True`` conf for enabling it for all packages

SpaceIm:
[...]

PR #12513, opened 2022-11-11
[CMakeToolchain][env] Added PKG_CONFIG_PATH env variable and PKG_CONFIG_EXECUTABLE one

### LINKED ISSUE #11962: add the `PKG_CONFIG_PATH` via CMake build helper or even CMakeToolchain
In https://github.com/conan-io/conan-center-index/pull/12241 there was some wonky code to get the pkg_config support in CMake to work with the new generators 

Part of the conversation (and linking to the solutions) seems like it could be continued here

***

> For some reason, pkg_check_modules tries to find .pc files not in the generator_folder, but in the generator_folder/lib/pkgconfig.

After digging deeper into it, I figured out that CMake is adding the suffix `lib/pkgconfig` to the `CMAKE_PREFIX_PATH` (pointing to the `package_folder` where the `*.pc` files are created). You can have a look at the issue https://gitlab.kitware.com/cmake/cmake/-/issues/18150

> I thought the PkgConfigDeps generator should resolve that (via env vars maybe), but it doesn't

Yes, you're right, but it should not be necessary as far as CMake should use that `CMAKE_PREFIX_PATH` by default. As I said above, CMake is adding that suffix, so it's getting failed because there are no PC files there.

> is this the right way to use CMakeToolchain and PkgConfigDeps generators together?

Given that information, I tend to say that yes, it could be one way to solve that issue, and I think there could be another workaround like:

```python
    def build(self):
        pkg_lib_folder = os.path.join(self.generators_folder, "lib", "pkgconfig")
        copy(self, "*.pc", self.generators_folder, pkg_lib_folder)
        cmake = CMake(self)
        cmake.configure()
        cmake.build()
```
Anyway, we could discuss if it could be a chance to create a mechanism to add the `PKG_CONFIG_PATH` via CMake build helper or even CMakeToolchain to avoid those workarounds.

_Originally posted by @franramirez688 in https://github.com/conan-io/conan-center-index/pull/12241#discussion_r955000431_

franramirez688:
Changelog: Feature: Added generators folder to `PKG_CONFIG_PATH` environment variable in `CMakeToolchain`.
Changelog: Feature: Ensure that `CMakeToolchain` will enforce usage of `pkg-config` executable set in `tools.gnu:pkg_config` config.
Closes: https://github.com/conan-io/conan/issues/11962
Partial implementation of https://github.com/conan-io/conan/issues/12354
Docs: https://github.com/conan-io/docs/pull/2832

**Note**: Originally started by https://github.com/conan-io/conan/pull/12355

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -559,6 +559,20 @@ class FindFiles(Block):
         }
 
 
+class PkgConfigBlock(Block):
+    template = textwrap.dedent("""
+        {% if pkg_config %}
+        set(PKG_CONFIG_EXECUTABLE {{ pkg_config }} CACHE FILEPATH "pkg-config executable")
+        {% endif %}
+        """)
+
+    def context(self):
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", default=False, check_type=str)
+        if pkg_config:
+            pkg_config = pkg_config.replace("\\", "/")
+        return pkg_config
+
+
 class UserToolchain(Block):
     template = textwrap.dedent("""
         {% for user_toolchain in paths %}
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -575,7 +575,11 @@ class PkgConfigBlock(Block):
             pkg_config = pkg_config.replace("\\", "/")
         pkg_config_path = self._conanfile.generators_folder
         if pkg_config_path:
-            pkg_config_path += os.pathsep
+            # Path sep based on "os" host settings (as it's done
+            # for "fpic" calculation for instance)
+            os_ = self._conanfile.settings.get_safe("os")
+            pathsep = ":" if os_ != "Windows" else ";"
+            pkg_config_path = pkg_config_path.replace("\\", "/") + pathsep
         return {"pkg_config": pkg_config,
                 "pkg_config_path": pkg_config_path}
 
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -565,7 +565,11 @@ class PkgConfigBlock(Block):
         set(PKG_CONFIG_EXECUTABLE {{ pkg_config }} CACHE FILEPATH "pkg-config executable")
         {% endif %}
         {% if pkg_config_path %}
+        if (DEFINED ENV{PKG_CONFIG_PATH})
         set(ENV{PKG_CONFIG_PATH} "{{ pkg_config_path }}$ENV{PKG_CONFIG_PATH}")
+        else()
+        set(ENV{PKG_CONFIG_PATH} "{{ pkg_config_path }}")
+        endif()
         {% endif %}
         """)
 
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -12,6 +12,7 @@ from conan.tools.build.cross_building import cross_building
 from conan.tools.cmake.toolchain import CONAN_TOOLCHAIN_FILENAME
 from conan.tools.intel import IntelCC
 from conan.tools.microsoft.visual import is_msvc, msvc_version_to_toolset_version
+from conans.client.subsystems import deduce_subsystem, WINDOWS
 from conans.errors import ConanException
 from conans.util.files import load
 
@@ -579,10 +580,9 @@ class PkgConfigBlock(Block):
             pkg_config = pkg_config.replace("\\", "/")
         pkg_config_path = self._conanfile.generators_folder
         if pkg_config_path:
-            # Path sep based on "os" host settings (as it's done
-            # for "fpic" calculation for instance)
-            os_ = self._conanfile.settings.get_safe("os")
-            pathsep = ":" if os_ != "Windows" else ";"
+            # hardcoding scope as "build"
+            subsystem = deduce_subsystem(self._conanfile, "build")
+            pathsep = ":" if subsystem != WINDOWS else ";"
             pkg_config_path = pkg_config_path.replace("\\", "/") + pathsep
         return {"pkg_config": pkg_config,
                 "pkg_config_path": pkg_config_path}
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -570,7 +570,7 @@ class PkgConfigBlock(Block):
         pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", default=False, check_type=str)
         if pkg_config:
             pkg_config = pkg_config.replace("\\", "/")
-        return pkg_config
+        return {"pkg_config": pkg_config}
 
 
 class UserToolchain(Block):
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -567,7 +567,7 @@ class PkgConfigBlock(Block):
         """)
 
     def context(self):
-        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", default=False, check_type=str)
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", check_type=str)
         if pkg_config:
             pkg_config = pkg_config.replace("\\", "/")
         return {"pkg_config": pkg_config}
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -153,7 +153,7 @@ class FPicBlock(Block):
     template = textwrap.dedent("""
         {% if fpic %}
         message(STATUS "Conan toolchain: Setting CMAKE_POSITION_INDEPENDENT_CODE={{ fpic }} (options.fPIC)")
-        set(CMAKE_POSITION_INDEPENDENT_CODE {{ fpic }} CACHE BOOL "Position independent code") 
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -559,6 +559,31 @@ def context(self):
...
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", check_type=str)
+        if pkg_config:
+            pkg_config = pkg_config.replace("\\", "/")
+        pkg_config_path = self._conanfile.generators_folder
+        if pkg_config_path:
+            # Check if user has already defined their own PKG_CONFIG_PATH variable
+            user_pkg_config_path = os.getenv("PKG_CONFIG_PATH")
+            if user_pkg_config_path:
+                # Prepending the Conan one
+                pkg_config_path = pkg_config_path + os.pathsep + user_pkg_config_path
```
franramirez688: Should I use this approach from `EnvVars`?

```python
        self._subsystem = deduce_subsystem(conanfile, scope)

    @property
    def _pathsep(self):
        return ":" if self._subsystem != WINDOWS else ";"
```

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -559,6 +559,31 @@ def context(self):
...
+        """)
+
+    def context(self):
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", check_type=str)
+        if pkg_config:
+            pkg_config = pkg_config.replace("\\", "/")
+        pkg_config_path = self._conanfile.generators_folder
+        if pkg_config_path:
+            # Check if user has already defined their own PKG_CONFIG_PATH variable
+            user_pkg_config_path = os.getenv("PKG_CONFIG_PATH")
```
czoido: I think you should use the `PKG_CONFIG_PATH` environment variable directly in CMake, not getting the value here.

conan/tools/cmake/toolchain/blocks.py
czoido: Yes, I think the correct way to do this is to deduce the subsystem and then aplying the path separator according to that

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1227,7 +1227,63 @@ def build_requirements(self):
...
+    pkg_check_modules(DEP REQUIRED IMPORTED_TARGET dep)
+    """)
+    client.save({
+        "dep/conanfile.py": dep,
+        "pkg/conanfile.py": pkg,
+        "pkg/CMakeLists.txt": cmakelists
+    })
+    client.run("create dep/conanfile.py")
+    client.run("create pkg/conanfile.py")
+    assert "Found dep, version 1.0" in client.out
```
czoido, line 1289: I think it also would be nice to test the contents of the toolchain file and check that the `PKG_CONFIG_PATH` that already existed is there and the one from Conan is correctly prepended.

franramirez688: Added to `test_pkg_config_block`

conan/tools/cmake/toolchain/blocks.py
franramirez688: The question is: should I assume that `scope == "build"`? Here, we don't have any scope defined.

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1227,7 +1227,63 @@ def build_requirements(self):
...
+        # Verify binary executable is found from build context package,
         # and library comes from host context package
         assert f"package/{build_context_package_id}/bin/foobin" in client.out
         assert f"package/{host_context_package_id}/include" in client.out
+
+
+@pytest.mark.tool_pkg_config
+def test_cmaketoolchain_and_pkg_config_path():
+    """
+    Lightweight test which is loading a dependency as a *.pc file through
```
memsharded, line 1239: Nice job with this test!

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -559,6 +559,35 @@ def context(self):
...
+    def context(self):
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", check_type=str)
+        if pkg_config:
+            pkg_config = pkg_config.replace("\\", "/")
+        pkg_config_path = self._conanfile.generators_folder
+        if pkg_config_path:
+            # Path sep based on "os" host settings (as it's done
+            # for "fpic" calculation for instance)
+            os_ = self._conanfile.settings.get_safe("os")
+            pathsep = ":" if os_ != "Windows" else ";"
```
czoido: Shouldn't we have to take the Windows subsystems into account for the path separators?

memsharded: Also if the ``conan_toolchain.cmake`` will run in the "build" machine, it should be using the ``settings_build``. 
You can check the code in ``deduce_subsystem()`` (maybe use the function too):
```python
 if scope.startswith("build"):
        the_os = conanfile.settings_build.get_safe("os")
```

franramirez688: I agree with you but what about all the pieces of code checking only if `self._conanfile.settings.get_safe("os") == "Windows"` not only CMakeToolchain, I'm talking about the rest of `conan.tools`. Should we review all of them? Only to check that we're not messing anything.

Examples:
```python
class VCVars:
    # .....

    def generate(self, scope="build"):
        conanfile = self._conanfile
        os_ = conanfile.settings.get_safe("os")
        if os_ != "Windows":
            return
```
```python
class TargetConfigurationTemplate(CMakeDepsFileTemplate):
     # ....

    @property
    def context(self):
        # ....
        is_win = self.conanfile.settings.get_safe("os") == "Windows"
```
Those are some examples.

memsharded: The ``VCVars`` is most likely not an issue, as VC cannot easily be a cross-compiler to another OS, so the ``settings`` should be ok.

The ``is_win`` for CMakeDeps seems related to ``set_interface_link_directories``, which is for "host" Windows systems too, so it is probably ok too.

The ``CMakeToolchain`` has probably some larger likelihood to be used for cross-build

czoido (approved):

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -559,6 +560,34 @@ def context(self):
...
+    def context(self):
+        pkg_config = self._conanfile.conf.get("tools.gnu:pkg_config", check_type=str)
+        if pkg_config:
+            pkg_config = pkg_config.replace("\\", "/")
+        pkg_config_path = self._conanfile.generators_folder
+        if pkg_config_path:
+            # hardcoding scope as "build"
+            subsystem = deduce_subsystem(self._conanfile, "build")
+            pathsep = ":" if subsystem != WINDOWS else ";"
+            pkg_config_path = pkg_config_path.replace("\\", "/") + pathsep
```
jcar87, line 586: I like this! :D Matches the documentation - and still allows (a few lines above in the template block) to fall back on anything the user may have previously specified

```
PKG_CONFIG_PATH

A colon-separated (on Windows, semicolon-separated) list of directories to search for .pc files. The default directory will always be searched after searching the path; the default is libdir/pkgconfig:datadir/pkgconfig where libdir is the libdir where pkg-config and datadir is the datadir where pkg-config was installed.
```

hoyhoy:
I'm running conan version 1.60.1 and I'm seeing `self.conf.get("tools.gnu:pkg_config", check_type=str) == None` when my `conan.conf` is set with the following.

```
[tools]
gnu:pkg_config = /usr/local/bin/pkgconf
```

Also, if I set pkg_config with the conan config command, it corrupts `conan.conf`.

`conan config set tools.gnu:pkg_config=/usr/local/bin/pkgconf`

`gnu = pkg_config = /usr/local/bin/pkgconf`

It only seems to work if I put a `[conf]` section in my conan profile.

memsharded:
Hi @hoyhoy 

This is a closed issue from 8 months ago, it would be better to report this as a new ticket.
There are some known issues, if the ``pkgconf`` is called at ``package_info()`` method it will not work, because environment or conf is not injected that way at that point.

Also ``[tools]`` is not a valid section, and ``conan.conf`` is not the right place to put it (in any case, it would be ``global.conf``. Likewise ``gnu:pkg_config`` is not a conf, but it must be ``tools.gnu:pkg_config=value``, the full string, and in ``global.conf``

hoyhoy:
Hmmm, `conan config list` doesn't refer to `conan.conf`?  `conan config set` absolutely changes that file.   Also, `self.conf.get("tools.gnu:pkg_config", check_type=str)` doesn't refer to `conan.conf`?  If it's `global.conf`, why does `conan config` write to `conan.conf`?  I'm not sure how to write an issue.  It definitely seems like something is wrong.

memsharded:
> Hmmm, conan config list doesn't refer to conan.conf? conan config set absolutely changes that file. Also, self.conf.get("tools.gnu:pkg_config", check_type=str) doesn't refer to conan.conf? If it's global.conf, why does conan config write to conan.conf?

No, you are mixing things. In Conan 1.X the two different confs coexist:
- ``conan.conf`` is the old, legacy one. ``conan config set/update`` writes to it, but those commands are also removed in Conan 2.0. It contains the legacy configuration items
- ``global.conf`` or ``profile`` files are the new ones. ``conan config list`` list the possible values, but they should go to ``global.conf`` or ``profiles``. ``tools.gnu:pkg_config`` is a new configuration listed here, it should be defined in ``global.conf`` or profile.

PR #12517, opened 2022-11-11
allowing placeholders in EnvVars

### LINKED ISSUE #12427: [feature] Extend EnVars API for custom generators
We are building a customizable generator for CMake presets. The generator uses the `VirtualBuildEnv` to define the environment variables in the presets. However, the public API of `EnvVars` does not support custom placeholders.

Our current workaround:
```python
vars = VirtualBuildEnv(conanfile).vars()
env = {
    # Double escape "{}"
    name: value.get_str(f'$penv{{{{{name}}}}}', vars._subsystem, vars._pathsep) for name, value in vars._values.items()
}
```

Possible API:
```python
env = VirtualBuildEnv(conanfile).vars().items(placeholder='$penv{{{name}}}')
```

Note: The included presets generator is not suitable for us. We need a presets file where developers can switch between presets (e.g. debug, release, tests, ...) without rerunning `conan install`.

memsharded:
Changelog: Feature: Allow placeholders in `EnvVars`.
Docs: https://github.com/conan-io/docs/pull/2835

Close https://github.com/conan-io/conan/issues/12427

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -251,16 +251,30 @@ class EnvVars:
     def keys(self):
         return self._values.keys()
 
-    def get(self, name, default=None):
+    def get(self, name, default=None, placeholder=None):
+        """
+        if placeholder is specified it means that we want to get a placeholder instead of the
+        real value of the env-var
+        """
         v = self._values.get(name)
         if v is None:
             return default
-        return v.get_value(self._subsystem, self._pathsep)
+        if placeholder:
+            return v.get_str(placeholder, self._subsystem, self._pathsep)
+        else:
+            return v.get_value(self._subsystem, self._pathsep)
 
-    def items(self):
-        """returns {str: str} (varname: value)"""
-        return {k: v.get_value(self._subsystem, self._pathsep)
-                for k, v in self._values.items()}.items()
+    def items(self, placeholder=None):
+        """returns {str: str} (varname: value)
+        if placeholder is specified it means that we want to get a placeholder instead of the
+        real values of the env-vars
+        """
+        if placeholder:
+            return {k: v.get_str(placeholder, self._subsystem, self._pathsep)
+                    for k, v in self._values.items()}.items()
+        else:
+            return {k: v.get_value(self._subsystem, self._pathsep)
+                    for k, v in self._values.items()}.items()
 
     @contextmanager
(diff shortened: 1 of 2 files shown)


conan/tools/env/environment.py
```diff
@@ -251,16 +251,30 @@ def __getitem__(self, name):
     def keys(self):
         return self._values.keys()
 
-    def get(self, name, default=None):
+    def get(self, name, default=None, placeholder=None):
+        """
+        if placeholder is specified it means that we want to get a placeholder instead of the
+        real value of the env-var
```
jcar87: The name `placeholder` and this comment confused me a little bit. Am I understand this correctly, in that `placeholder` only makes sense for variables that have had values set with `append` or `prepend` rather than with `define`, and that the placeholder applies to the pre-existing value? So rather than resolving that value, a placeholder is referenced instead, where `{name}` in the placeholder will resolve to the variable name. 

I'm not sure if `placeholder` as a name captures the functionality, and from this comment I understood something different - maybe something like "if placeholder is specified, it will be used in place of the pre-existing value of the env-var, where {name} will resolve to the variable name"?

memsharded: Yes, only ``append`` and ``prepend`` will contain an internal list like ``["userdata1", CURRENT_ENV_VAR_VALUE, "userdata2]``, because ``define()`` will overwrite existing env-var value.

The idea is that when ``placeholder`` is defined, instead of replacing with the current env-var value, it is replaced with a typical placeholder, like ``%envvar%``, ``$envvar`` as string literal.

I see what you mean, but failed to bring a better name, any suggestion? Or just improving the docstring?

jcar87: I like the word "reference", since we're dealing with how whether a variable is resolved to a reference - 
e.g. something along the lines of:

`variable_reference`: if specified, use a variable reference instead of the pre-existing value of environment variable, where `{name}` can be used to refer to the name of the variable. 

I think this naming would be consistent with https://www.gnu.org/software/make/manual/html_node/Reference.html

memsharded: Sounds good, I like it more, I'll change it

memsharded: Done

jcar87 (approved):

PR #12543, opened 2022-11-15
[BazelDeps][bugfix] Reading `dependencies.direct_host`

### LINKED ISSUE #12236: [bug] Bazel generator add tools and build dependencies as bazel deps

### LINKED ISSUE #12444: [bug] Bazel BUILD file should not include build-context dependencies

franramirez688:
Changelog: Bugfix: Do not include build-context dependencies in Bazel BUILD file.
Docs: omit

Closes: https://github.com/conan-io/conan/issues/12444
Closes: https://github.com/conan-io/conan/issues/12236

Note: this PR should supersede https://github.com/conan-io/conan/pull/12445

### DIFF
--- a/conans/test/integration/toolchains/google/test_bazel.py
+++ b/conans/test/integration/toolchains/google/test_bazel.py
@@ -76,3 +76,48 @@ def test_bazel_exclude_folders():
     c.run("install dep/0.1@ -g BazelDeps")
     build_file = c.load("dep/BUILD")
     assert 'static_library = "lib/libmymath.a"' in build_file
+
+
+def test_bazeldeps_and_build_requires():
+    """
+    Testing that direct build requires are not included in dependencies BUILD files
+
+    Issues related:
+        * https://github.com/conan-io/conan/issues/12444
+        * https://github.com/conan-io/conan/issues/12236
+    """
+    c = TestClient()
+    tool = textwrap.dedent("""
+        import os
+        from conan import ConanFile
+
+        class ToolConanfile(ConanFile):
+            name = "tool"
+            version = "0.1"
+
+            def package_info(self):
+                self.cpp_info.libs = ["mymath"]
+        """)
+    c.save({"tool/conanfile.py": tool})
+    c.run("create tool")
+    dep = textwrap.dedent("""
+        import os
+        from conan import ConanFile
+        from conan.tools.files import save
+        class ExampleConanIntegration(ConanFile):
+            name = "dep"
+            version = "0.1"
+
+            def build_requirements(self):
+                self.build_requires("tool/0.1")
+            def package(self):
+                save(self, os.path.join(self.package_folder, "lib", "mymath", "otherfile.a"), "")
+                save(self, os.path.join(self.package_folder, "lib", "libmymath.a"), "")
+            def package_info(self):
+                self.cpp_info.libs = ["mymath"]
+        """)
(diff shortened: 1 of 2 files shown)


conans/test/integration/toolchains/google/test_bazel.py
```diff
@@ -76,3 +76,48 @@ def package_info(self):
...
+                self.build_requires("tool/0.1")
+            def package(self):
+                save(self, os.path.join(self.package_folder, "lib", "mymath", "otherfile.a"), "")
+                save(self, os.path.join(self.package_folder, "lib", "libmymath.a"), "")
+            def package_info(self):
+                self.cpp_info.libs = ["mymath"]
+        """)
+    c.save({"dep/conanfile.py": dep})
+    c.run("create dep")
+    c.run("install dep/0.1@ -g BazelDeps")
```
czoido: I think this will not test the changes correctly, because you do not build when installing, so the build requires will never be there anyway...
```suggestion
    c.run("export dep")
    c.run("install dep/0.1@ -g BazelDeps --build=missing")
```

franramirez688: You're right! Indeed, I checked the `mongo-c-driver` dependency so, but this one did not put the same 🤦

czoido (approved):

memsharded (approved):

PR #12551, opened 2022-11-15
WIP: CMake: escape cmake executable path in command lines

Novakov:
Changelog: Fix: Excape path to CMake executable specified as `conan_cmake_program`

`conan.CMake` helper used path to CMake executable (as specified in `conan_cmake_program` configuration option) directly in command line constructed using string formatting. On Windows, CMake by default installs to `C:/Program files` directory which results in command line: `C:/Program files/CMake/bin/cmake.exe -D...`. According to command line parsing rules, `C:/Program` is interpreted as executable name which is invalid.

Use `args_to_string` to properly escape path to `cmake` even if it contains spaces.

Update: I'm having some trouble with getting unit tests to pass, nothing serious, just tedious

### DIFF
--- a/conans/client/build/cmake.py
+++ b/conans/client/build/cmake.py
@@ -278,8 +278,11 @@ class CMake(object):
             pkg_env = {"PKG_CONFIG_PATH": self._conanfile.install_folder} if set_env else None
 
         with environment_append(pkg_env):
-            command = "cd %s && %s %s" % (args_to_string([self.build_dir]), self._cmake_program,
-                                          arg_list)
+            command = "cd %s && %s %s" % (
+                args_to_string([self.build_dir]),
+                args_to_string([self._cmake_program]),
+                arg_list
+            )
             if platform.system() == "Windows" and self.generator == "MinGW Makefiles":
                 with tools.remove_from_path("sh"):
                     self._run(command)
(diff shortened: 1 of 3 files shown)


memsharded (commented):
Hi @Novakov 

This ``CMake`` integration is now considered legacy and superseded by the one from ``from conan.tools.cmake import CMake``, as it has been removed already from 2.0

Any change, including fixes, should be targeted to the new tools first, and only in exceptional cases doing changes in legacy integrations would be accepted.

Thanks very much for contributing!

Novakov:
@memsharded It that case there is no need for this pull request. I should note here that it is pretty hard to find `from conan.tools.cmake import CMake` in conan docs.

memsharded:
Thanks for the feedback, yes we are aware that the 1.X docs are a bit messy at the moment, due to this migration process to 2.0. In case of doubt, there would be the go-to pages of the docs:
- The 2.0 migration guide: https://docs.conan.io/en/latest/conan_v2.html. It is actually collection of things that should be used in 1.X to be ready for the migration
- The reference of the new tools: https://docs.conan.io/en/latest/reference/conanfile/tools.html
- The new Conan 2.0 docs: https://docs.conan.io/en/2.0/index.html

Issue #12549, opened 2022-11-15
[question] Review ``conan cache path``


memsharded:
I have been writing the docs for it, trying it manually, and found it quite confusing, I have been struggling a bit to use it, eg:

```
conan cache path zlib/1.2.12#6758146baf425dc62ecc5246a1e955e4:2401fa1d188d289bb25c37cfa3317e13e377a351#latest
C:\Users\memsharded\.conan2\p\8a18fc6015f5b185\e
```
Even if you provide the full path, it will still return the path to the ``export`` folder of the conanfile.py, not the package folder, and  it was surprising
```
conan cache path zlib/1.2.12#6758146baf425dc62ecc5246a1e955e4:2401fa1d188d289bb25c37cfa3317e13e377a351 --folder=package
ERROR: '--folder package' requires a valid package reference
```
The problem here is that it is missing ``#latest`` as ``package_revision``, but it is challenging to get that.

czoido:
Closed by: https://github.com/conan-io/conan/pull/12554

PR #12554, opened 2022-11-16
[develop2] Improve cache path

memsharded:
Close https://github.com/conan-io/conan/issues/12549


conan/cli/commands/cache.py
```diff
@@ -18,55 +18,37 @@ def cache_path(conan_api: ConanAPIV2, parser, subparser, *args):
         Shows the path af a given reference
     """
     subparser.add_argument("reference", help="Recipe reference or Package reference")
-    subparser.add_argument("--folder", choices=['exports', 'exports_sources', 'sources', 'build',
-                                                'package'], default="exports",
```
czoido, line 22: Why removing the `--folder=package`?

czoido:
Maybe this should still work when doing `conan cache path zlib/1.2.12#latest` and `zlib/1.2.12#pkgid#latest` so that's aligned with the other commands?

conan/cli/commands/cache.py
memsharded: Because it is redundant, not needed. By default it is giving you the ``package`` folder (95% of cases for a full ``pref``). Why adding another argument that will give further incompatibilities when the arg is a ``ref``. 
Same with ``export`` folder, that is the default for a ``ref``

memsharded:
> Maybe this should still work when doing conan cache path zlib/1.2.12#latest and zlib/1.2.12#pkgid#latest so that's aligned with the other commands?

Which other commands? IMO, the ``#latest`` doesn't provide any value here, probably needs to be revised in the other commands?

czoido:
> > Maybe this should still work when doing conan cache path zlib/1.2.12#latest and zlib/1.2.12#pkgid#latest so that's aligned with the other commands?
> 
> Which other commands? IMO, the `#latest` doesn't provide any value here, probably needs to be revised in the other commands?

Yes, I think that if we are removing the` #latest` here we should check if we should remove that in every other places like `conan list`, `conan download` and `conan upload` and default to the latest if not specified.

czoido (approved):

Issue #12474, opened 2022-11-06
[feature] Add more config in `conf` in order to provide a canonical way to set path to compiler, linker etc


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Currently, there is no canonical way to ensure that a specific compiler will be used when different versions of the same compiler or even different compilers are installed on build machine.
`compiler` values in profile are just a promise of conan user to conan client/recipe, an implicit contract that this compiler and this version will be found by underlying build system. `CC`, `CXX`, `LD`, `AR` etc env vars are just GNU conventions and they are not understood by all build systems.

Therefore I advice to add config like these ones (not necessarily these names):

- `tools.c.compiler:path`
- `tools.cxx.compiler:path`
- `tools.linker:path`
- `tools.archiver:path`

and also these ones for example:
- `tools.fortran.compiler:path`
- `tools.python.interpreter:path`
- `tools.perl.interpreter:path`
- some conf for CUDA maybe

To ensure than these paths are honored, conan build helpers would intercept these values and inject whatever env var or specific variable understood by underlying build system.

For example:
- `AR` env var is not understood by CMake, so if `tools.archiver:path` is set, it would be translated to `CMAKE_AR` CMake variable in `CMakeToolchain`).
- `tools.c.compiler:path` would be translated to `CC` env var in `AutotoolsToolchain`, and on Windows this value would be curated: converted to unix path and eventually each folder translated to their short name to avoid spaces in paths (it's what CMake does sometimes). Why curated? Because underlying build system should be an internal detail, users should be able to specify the native path format to their tool and not worry about its underlying usage.

Moreover, when using a custom build system in a recipe, it would provide a consistent way to honor paths eventually specified by users.
A recipe could even conditionally avoid to add a `tool_requires` if path of related tool is already filled by user in config (`cpython` may not be needed in tool_requires if `tools.python.interpreter:path` is set).

The idea would be to listen these config at top priority, then eventually their GNU equivalent (because it was more or less expected by conan users that CC/CXX/LD/AR etc in profile should be honored, even though there is absolutely no guarantee that a random build system can honor this wish out of the box), then eventually some default assumption based on compiler settings in profile, then let underlying build system uses its default detection mechanism.

memsharded:
Lets considered this closed by https://github.com/conan-io/conan/pull/12556

[...]

PR #12556, opened 2022-11-16
[CMakeToolchain] Setting `CMAKE_<LANG>_COMPILER`

### LINKED ISSUE #9962: [feature] Setting `CMAKE_<LANG>_COMPILER` in a toolchain
## Problem statement

Right now it is common to use a different compiler for `conan install` and CMake configure steps which result in an error during project configuration.

## Potential solution

From my observation, it is typical to set `CXX` and `CC` environment variables in the Conan profile file. In case a user provides them and uses such a profile during `conan install` step it would be great if `CMakeToolchain` would set `CMAKE_<LANG>_COMPILER` cache variables based on those.

Thanks to that we will always end up with consistent builds and users will just have to call:

```
conan install .. -pr my_profile
cmake .. -toolchain conan_toolchain.cmake
```

franramirez688:
Changelog: Feature: Add `tools.build:compiler_executables` configuration to set compilers variables in CMakeToolchain, MesonToolchain, and AutoToolsToolchain, e.g., `CMAKE_<LANG>_COMPILER` in `CMakeToolchain`.
Changelog: Fix: Remove hardcoded definitions of `CMAKE_CXX_COMPILER` in `CMakeToolchain`.
Docs: https://github.com/conan-io/docs/pull/2833
Closes: https://github.com/conan-io/conan/issues/9962

See also https://github.com/conan-io/conan/pull/12635

 -- Coming from https://github.com/conan-io/conan/pull/12577 (CLOSED)

The hardcoded definition of CMAKE_XXX_COMPILER in CMakeToolchain can be removed:

* In Ninja it is good if VCVars activate the environment and put cl.exe there
* For Windows-Clang, it is necessary to add CC/CXX in [buildenv] in the profile. This is not great, because it forces users to have to activate the environment too. It would be better if there is a built-in mechanism like in https://github.com/conan-io/conan/pull/12556 that puts it directly in conan_toolchain.cmake


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -843,16 +848,22 @@ def context(self):
...
         toolset = self._get_toolset(generator)
-
-        compiler, compiler_cpp, compiler_rc = self._get_compiler(generator)
+        compilers_by_conf = self._conanfile.conf.get("tools.cmake.cmaketoolchain:compilers",
+                                                     default=[], check_type=list)
+        # FIXME: do we want to keep this legacy part?
+        compiler, compiler_cpp, compiler_rc = None, None, None
+        if not compilers_by_conf:
+            # Configuration has preference for setting compilers
+            compiler, compiler_cpp, compiler_rc = self._get_windows_compilers(generator)
```
franramirez688: Should I keep this if the new approach is good?

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
```diff
@@ -784,3 +784,27 @@ def test_pkg_config_block():
...
+def test_set_cmake_lang_compilers():
+    profile = textwrap.dedent(r"""
+    [settings]
+    os=Windows
+    arch=x86_64
+    compiler=clang
+    compiler.version=15
+    compiler.libcxx=libstdc++11
+    [conf]
+    tools.cmake.cmaketoolchain:compilers = [("C", "/my/local/gcc"), ("cxx", "C:\\local\\g++"),]
```
memsharded: It would be better to have a ``c_compiler`` and ``cxx_compiler`` confs, no need for this tuple trick, and C, C++ compilers deserve their own named conf.

My main doubt is if this would be a ``tools.build:cxx_compiler`` generic one instead of a CMake specific one, so other generators could also leverage them (probably defining directly the CC/CXX env-var?)

franramirez688: I see your point, but I think they should be 5:

```
C
CXX
OBJC (Objective C)
OBJCXX (Objective C++)
RC (Windows Resource Compiler)
```

conan/tools/cmake/toolchain/blocks.py
SSE4: why is it named `_get_windows_compilers`? shouldn't it be platform-agnostic here?

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
memsharded: I am fine with having 5 properties. Maybe it would be more idiomatic to have a dict instead? Maybe it is the list of tuples that is hurting a bit the UI?

conan/tools/cmake/toolchain/blocks.py
franramirez688: Sure, but the implementation looks like this:

```python
    def _get_windows_compilers(self, generator):
        compiler = self._conanfile.settings.get_safe("compiler")

        compiler_c = compiler_cpp = compiler_rc = None
        # TODO: Check if really necessary now that conanvcvars is used
        if "Ninja" in str(generator) and is_msvc(self._conanfile):
            compiler_c = compiler_cpp = "cl"
        elif os_ == "Windows" and compiler == "clang" and "Visual" not in generator:
            # definition of compiler only if not using the VS Clang toolset
            compiler_rc = "clang"
            compiler_c = "clang"
            compiler_cpp = "clang++"

        return compiler_c, compiler_cpp, compiler_rc
```
So, I thought it could be a better name. I can keep the previous name for sure.

conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
czoido: I like the idea of it being a dictionary instead of adding the 5 configs, but if we add them I would probably also add `CUDA`.

franramirez688: In my opinion, I prefer having five properties if it's alright for you all.

franramirez688: Finally, dict-like is good for me 😅

conan/tools/cmake/toolchain/blocks.py
SSE4: well, there is FIXME above, maybe drop it, as it's legacy part?
otherwise, I think we previously agreed to use `blocks` approach, so it might have sense to isolate this kind of code into blocks like `MSVCNinja` and `ClangWindows` (if it's really needed).
we know, the name of compiler might be different, such as `clang-13.exe` or `i686-w64-mingw32-clang++.exe`

memsharded: I think putting this in a dedicated ``CompilerBlock`` or something like that (just 1 block for the definition of this, not one block per use case), makes sense.

franramirez688: It sounds good to me!

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -667,8 +667,63 @@ class TryCompileBlock(Block):
...
-class GenericSystemBlock(Block):
+class CompilersBlock(Block):
     template = textwrap.dedent(r"""
+        {% if compiler %}
+        if(NOT DEFINED ENV{CC})
+        set(CMAKE_C_COMPILER {{ compiler }})
+        endif()
+        {% endif %}
+        {% if compiler_cpp %}
+        if(NOT DEFINED ENV{CXX})
```
memsharded: I would like the opinion of @czoido and @jcar87 about this (I know it was already there ,not this PR).
If Conan is controlling the CMAKE_CXX_COMPILER, maybe that should still be prioritized against a general environment CXX env-var? Not sure about this.

memsharded (approved):

conan/tools/cmake/toolchain/blocks.py
czoido: But if I'm not wrong the current priority as this PR is would be:
- `tools.build:compilers `configuration
- environment variables
- the result from `_get_windows_compilers`

So it should be already prioritized.

I was also wondering if now that we have the conf we still need to set to set these values using the  `_get_windows_compilers` functions?

memsharded: I am not sure. If ``ENV{CXX}`` is defined, it will not generate the ``CMAKE_CXX_COMPILER`` entry in the toolchain at all.

> I was also wondering if now that we have the conf we still need to set to set these values using the _get_windows_compilers functions?

I think it is nice to have those, because those are the default needed to work with those setups, and it can be extra burden on users to specify the ``cl.exe`` compiler when they are targeting the ``compiler=msvc`` setting, the VS in installed correctly, just because the setup is via ``Ninja`` instead of other CMake generator.

czoido: > I am not sure. If `ENV{CXX}` is defined, it will not generate the `CMAKE_CXX_COMPILER` entry in the toolchain at all.
> 

But the `CMAKE_<LANG>_COMPILER` from the conf is set below in this code:

```
{% for lang, compiler_path in compilers_by_conf.items() %}
set(CMAKE_{{ lang|upper }}_COMPILER "{{ compiler_path|replace('\\', '/') }}")
{% endfor %}
```

I think probably the logic should be changed, because the conf will override the other values even if you are not setting them in the conf.

> > I was also wondering if now that we have the conf we still need to set to set these values using the _get_windows_compilers functions?
> 
> I think it is nice to have those, because those are the default needed to work with those setups, and it can be extra burden on users to specify the `cl.exe` compiler when they are targeting the `compiler=msvc` setting, the VS in installed correctly, just because the setup is via `Ninja` instead of other CMake generator.

👍

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -667,6 +667,61 @@ class TryCompileBlock(Block):
...
+        set(CMAKE_CXX_COMPILER {{ compiler_cpp }})
+        endif()
+        {% endif %}
+        {% if compiler_rc %}
+        if(NOT DEFINED ENV{RC})
+        set(CMAKE_RC_COMPILER {{ compiler_rc }})
+        endif()
+        {% endif %}
+        {% for lang, compiler_path in compilers_by_conf.items() %}
+        set(CMAKE_{{ lang|upper }}_COMPILER "{{ compiler_path|replace('\\', '/') }}")
```
memsharded: > But the CMAKE_<LANG>_COMPILER from the conf is set below in this code:

@czoido I see, I didn't realize it. Maybe it sounds a bit risky/not clean to potentially have 2 definitions of CMAKE_CXX_COMPILER in the same toolchain file? Maybe Conan should explicitly decide which one prevails and generate only 1 entry of CMAKE_XXX_COMPILER?

conan/tools/cmake/toolchain/blocks.py
franramirez688: Sorry, I've added an extra if-else statement to keep a clear structure in the jinja template

memsharded (approved):

memsharded (approved):

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -123,7 +123,12 @@ def defines(self):
 
     def environment(self):
         env = Environment()
-        if is_msvc(self._conanfile):
+        compilers = self._conanfile.conf.get("tools.build:compilers", default={}, check_type=dict)
+        if compilers:
+            for compiler, compiler_path in compilers.items():
+                # CC, CXX
+                env.define(compiler.upper(), compiler_path)
```
czoido: Do we want to define all the "tools.build:compilers" values to environment variables directly?
for example the dict could be:
``
 {"Fortran": "path/to/fortran"}
``
because the var you want to set for CMake is `CMAKE_Fortran_COMPILER`
but for Autotools would be something like `FC` or `F77` ?

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -123,7 +123,12 @@ def defines(self):
...
+        elif is_msvc(self._conanfile):
             env.define("CXX", "cl")
             env.define("CC", "cl")
```
czoido: Isn't this the same default we removed for CMake?

jcar87 (commented):
LGTM! My two cents based on some recent discussions. 


Regarding the name of the conf:
Users might intuitively expect that given a compiler name as a setting:

```
[settings]
compiler=xxx
```
That the following may "look" redundant or repetitive and users may wonder "am I not defining the same thing??"
```
[conf]
tools.build:compilers
```

When in fact they are two different things. I wonder if perhaps `tools.build:compiler_executables` may be a more descriptive name? So that the setting refers to the name of the compiler in general, while the `compiler_executables` is more about the actual program / command to run.

Regarding the documentation / description:

> "Defines a Python dict-like with the compilers path to be used, e.g., {'CXX': '/usr/bin/clang++'}"

It may be worth mentioning that mentioning that this accepts a full path, a "name" (as done in one of the tests), in which case the underlying tools which works in the path. 

Regarding more advanced usage:
Some tools support `CC` and `CXX` to have values that also have spaces - for example, I believe `CXX=ccache clang++-13` would be a valid value when building with Autotools (leaving aside the fact that ccache may need further env vars to be set up correctly). Is this something we expect to work? What about build systems that don't support this because they have separate looking for compiler launchers? e.g. `CMAKE_COMPILER_LAUNCHER=ccache` + `CMAKE_CXX_COMPILER=clang++-13` would work, but `CMAKE_CXX_COMPILER="ccache clang++-13"` might not work?

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -667,6 +667,22 @@ class TryCompileBlock(Block):
...
+        {% endfor %}
+    """)
+
+    def context(self):
+        # Reading configuration from "tools.build:compilers" -> {"C": "/usr/bin/gcc"}
+        compilers = self._conanfile.conf.get("tools.build:compilers", default={}, check_type=dict)
+        # Users used to define CC as env variable. Translating it just in case.
+        if "CC" in compilers:
+            compilers["C"] = compilers.pop("CC")
+        return {"compilers": compilers}
```
czoido: I'm not sure about this... if we are directly translating the dictionary _key_ to the CMAKE_KEY_COMPILER CMake variable, then we know that are a more or less restricted set of accepted keys, something like: `CXX`, `C`, `CSharp`, `CUDA`, `Fortran`, `HIP`, `ISPC`, `ASM`... that are possible values for the LANG in CMake. Maybe we should fail in this case? Maybe we should define the set of generic keys we accept?

memsharded: That would make sense: to have some Conan convention to name compilers, and have a mapping from those Conan names to every build system specific names.

franramirez688:
Pointing to your comments and the original purpose of this PR, I've proposed another point of view exclusively for CMake. Let's discuss this new approach.

memsharded (commented):
This is something that I struggle to see. The hardcoded definition of ``clang`` have been removed from the toolchain, and now it will be necessary to specify it in the profile. If we want to build with clang some dependencies from ConanCenter we will need something like:

```
[conf]
tools.cmake.cmaketoolchain:set_compilers={"CXX": "C:/path/to/my/clang++", "C": ....}
tools.meson.mesontoolchain:set_compilers={"cpp": "C:/path/to/my/clang++", "c": ....}
tools.microsoft.msbuildtoolchain:set_compiler=...
tools.autotools.autotoolstoochain:set_compilers=={"CXX": "C:/path/to/my/clang++", "CC": ....}

# And maybe for some others Boost, OpenSSL, we also need the environment
[buildenv]
CC=(path)C:/path/to/my/clang
CXX=(path)C:/path/to/my/clang++
```

Of course there are some tricks that might not make necessary all of the above, like ``MesonToolchain`` is now extracting things from the environment to define the ``meson_toolchain.ini`` values of ``compiler=xxxxxx``, which is not great IMO, and also autotools might leverage the env-var directly, no strictly needed that this value is defined in the ``autotoolstoolchain.sh`` (but it would be great for symmetry with other tools).

I am not convinced this is the best approach, and I still think that an abstract definition of the current compilers executables should be good enough for all build systems. Or just rely on the environment CC/CXX, and allow CMakeToolchain to capture it, so it is not strictly necessary for CMake users (the majority) to activate an environment script always, just to operate.

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -667,6 +667,37 @@ class TryCompileBlock(Block):
...
+        compilers = {}
+        compiler_launchers = {}
+        # Allowed <LANG> variables (and <LANG>_LAUNCHER)
+        compilers_mapping = {"cc": "C", "cuda": "CUDA", "cxx": "CXX",
+                             "objc": "OBJC", "objcxx": "OBJCXX", "rc": "RC"}
+        for comp, lang in compilers_mapping.items():
+            # To set CMAKE_<LANG>_COMPILER
+            if comp in compilers_by_conf:
+                compilers[lang] = compilers_by_conf[comp]
+            # To set CMAKE_<LANG>_COMPILER_LAUNCHER
```
memsharded: I am afraid that ``COMPILER_LAUNCHER`` vars are CMake 3.17: https://cmake.org/cmake/help/latest/envvar/CMAKE_LANG_COMPILER_LAUNCHER.html, so this wouldn't be valid, we have the commitment to support 3.15: https://github.com/conan-io/tribe/pull/4

memsharded (changes requested):

conan/tools/cmake/toolchain/blocks.py
franramirez688: Oh, you're completely right. Changing it

conans/model/conf.py
```diff
@@ -52,6 +52,7 @@
     "tools.build:defines": "List of extra definition flags used by different toolchains like CMakeToolchain and AutotoolsToolchain",
     "tools.build:sharedlinkflags": "List of extra flags used by CMakeToolchain for CMAKE_SHARED_LINKER_FLAGS_INIT variable",
     "tools.build:exelinkflags": "List of extra flags used by CMakeToolchain for CMAKE_EXE_LINKER_FLAGS_INIT variable",
+    "tools.build:compiler_executables": "Defines a Python dict-like with the compilers path to be used. Allowed keys {'cc', 'cuda', 'cxx', 'objc', 'objcxx', 'rc'}",
```
franramirez688: I think it would be great to have a simple mechanism to check possible values for any configuration if necessary.

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -667,6 +667,30 @@ class TryCompileBlock(Block):
...
+
+    def context(self):
+        # Reading configuration from "tools.build:compiler_executables" -> {"C": "/usr/bin/gcc"}
+        compilers_by_conf = self._conanfile.conf.get("tools.build:compiler_executables", default={},
+                                                     check_type=dict)
+        # Map the possible languages
+        compilers = {}
+        compiler_launchers = {}
+        # Allowed <LANG> variables (and <LANG>_LAUNCHER)
+        compilers_mapping = {"cc": "C", "cuda": "CUDA", "cxx": "CXX",
```
czoido: Maybe I would use ``c`` instead of ``cc`` and ``cpp`` instead of ``cxx``. I think it's also worth adding something like:
`fortran` -> `Fortran`
`asm` -> `ASM`

czoido:
Just a couple of comments. I think the last approach is much better, I really like it! On my side it's ready once they are solved

czoido (approved):

memsharded (approved):
Definitely much better, I like it.
Just recall to document it as ``experimental``, just in case.

SpaceIm:
I guess this PR also addresses https://github.com/conan-io/conan/issues/12474 (but maybe not all cases like archiver path, linker path or fortran path).

SSE4:
@franramirez688 seems like it is breaking change, so it should be reverted (see https://github.com/conan-io/conan-center-index/issues/14508)

SpaceIm:
Maybe not revert, just set default `CC`/`CXX` env vars to `cl` in `AutotoolsToolchain` if `tools.build:compiler_executables` doesn't have `c`/`cpp` keys. It broke m4 recipe in conan-center (other recipes are ok, they didn't trust CC/CXX set by AutotoolsToolchain if msvc).

PR #12557, opened 2022-11-16
[develop2] fixing conan remote remove pattern

memsharded:
``conan remote remove *`` was not working


conans/client/cache/remote_registry.py
```diff
@@ -1,3 +1,4 @@
+import fnmatch
```
franramirez688: are we using it?

franramirez688 (approved):

PR #12572, opened 2022-11-20
[develop2] Improve ``#latest`` and pattern selection from remove/upload/download

memsharded:


franramirez688 (approved):
Great job! 👏 My only suggestion is to change the `conan/internal/api/select_pattern.py` (it seems to me it's related to the database) to something more descriptive as `conan/internal/api/references_filter.py`.

memsharded:
> Great job! 👏 My only suggestion is to change the conan/internal/api/select_pattern.py (it seems to me it's related to the database) to something more descriptive as conan/internal/api/references_filter.py.

Yeah, I didn't love the name, if the rest of the team also likes ``ReferenceFilter`` as name, I'll change it.

czoido:
> > Great job! 👏 My only suggestion is to change the conan/internal/api/select_pattern.py (it seems to me it's related to the database) to something more descriptive as conan/internal/api/references_filter.py.
> 
> Yeah, I didn't love the name, if the rest of the team also likes `ReferenceFilter` as name, I'll change it.

Maybe the name is not the optimal but I think it represents the purpose of the class good enough, let's merge and change the name later if we still don't like it....

PR #12580, opened 2022-11-22
fixing runenv for cmake.test()

### LINKED ISSUE #12462: [bug] buildenv vs runenv for cmake.test() helper.
Follow-up to #12407

I can see now why conan applies `[buildenv]` to this its `cmake.test` helper (since it's really `cmake --build <build_folder> --target test`), but it feels like many cases with shared libraries would also want the `[runenv]`. For example, they might need to find .dll files from libraries that were in requires. I would expect most test content to be pretty indifferent to the buildenv (unless doing something unusual, for example testing a header-only library by calling `$(CC) -fsyntax-only`).

But tests that call executable you just built (which linked to stuff in requires) don't seem that exotic. So I tested this, and it's indeed not working. I'm surprised I never hit that for real, but I guess I don't have *all* that many shared-library dependencies, and the one that is common (zlib1.dll) is kind of accidentally working because git for windows puts one in PATH, and I mostly live in git-bash.

  * Operating System+version: Windows 10 200H2
  * Conan version: 1.53.0
  * Python version: 3.10.7

`conan create conanfile.py -pr conanprofile.txt`

<details>
<summary>conanfile.py</summary>

```py
from conan import ConanFile
from conan.tools.cmake import CMake

class CTestConan(ConanFile):
    name = "ctest"
    version = "1.0"
    default_options = {"zlib:shared": True}
    generators = "CMakeToolchain", "CMakeDeps"
    requires = "zlib/1.2.12"

    exports_sources = "CMakeLists.txt", "main.cpp"

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()
        cmake.test()
        # self.run("ctest", cwd=self.build_folder, run_environment=True)
        #self.run("ctest", cwd=self.build_folder, env=["conanbuild","conanrun"])

    def package_info(self):
        self.cpp_info.libs = ["hello"]
```
</details>

<details>
<summary>CMakeLists.txt</summary>

```cmake
cmake_minimum_required(VERSION 3.23)
project(test LANGUAGES CXX)

find_package(ZLIB)

add_executable(main main.cpp)
target_link_libraries(main ZLIB::ZLIB)

enable_testing()
add_test(NAME main COMMAND main)
add_test(NAME zlib1 COMMAND where zlib1.dll)
add_test(NAME env COMMAND ${CMAKE_COMMAND} -E environment)
```
</details>

<details>
<summary>main.cpp</summary>

```c++
#include <zlib.h>
int main() {
	z_stream zs = {};
	inflateInit(&zs);
	inflateEnd(&zs);
}
```
</details>

<details>
<summary>conanprofile.txt</summary>

```ini
[buildenv]
BUILDENV=1
[runenv]
RUNENV=1
[conf]
tools.env.virtualenv:auto_use=True
```
</details>

If run from cmd (with no other zlib1.dll already in PATH from some random place), main fails with Exit code 0xc0000135 (STATUS_DLL_NOT_FOUND) and <build_folder>/Testing/Temporary/LastTest.log shows 

> 1/3 Test: main Exit code 0xc0000135

> 2/3 Test: zlib1
> Command: "C:/Windows/System32/where.exe" "zlib1.dll"
> Output:
> INFO: Could not find files for the given pattern(s).

> 3/3 Test: env
> Command: "C:/Program Files/CMake/bin/cmake.exe" "-E" "environment"
> BUILDENV=1

but no `RUNENV=1` -- so `[buildenv]` was applied, but not `[runenv]`. This is what I expected from #12407, but it seems like we would want `[runenv]` too when running tests.

---

If run from git-bash instead of cmd, all 3 tests "pass", but the _build_folder_/Testing/Temporary/LastTest.log shows that this is because we found zlib1 from mingw64:

> 3/3 Test: zlib1
> Command: "C:/Windows/System32/where.exe" "zlib1.dll"
> Output:
>C:\Program Files\Git\mingw64\bin\zlib1.dll

And that's why main.exe succeeded, that was enough to make a simple executable load and close enough to not crash. But we've mismatched versions/settings/options pretty much freely, since conan had no part in picking this one.

---

If I replace the `cmake.test()` with a `self.run("ctest", cwd=self.build_folder, run_environment=True)`, then PATH gets `%CONAN_USER_HOME%\.conan\data\zlib\1.2.12\_\_\package\ad5261bf6074807e7189c351b0f79b113bf2f6c0\bin`, and so main.exe gets the correct zlib1.dll. But it still doesn't get the `[runenv]RUNENV=1`, so apparently `run_environment=True` still does not pick up the profile's `[runenv]`?

If I try to get both by passing a list, i.e.`self.run("ctest", cwd=self.build_folder, env=["conanbuild","conanrun"])`, I unexpected get neither `[buildenv]` nor `[runenv]`.

The documentation at https://docs.conan.io/en/latest/reference/conanfile/other.html#output-and-running says this should work
> the environment or list of environment activations scripts to pre-pend to the given `command`. 

but the code to force a single `str` into into a list:

https://github.com/conan-io/conan/blob/5bf9aacedf9cc71c52ef548b079285715b51ba10/conans/model/conan_file.py#L441

will just end up in its `else []` case if the caller actually anything other than `str`

### Suggested fix

So it seems like Conanfile.run(env=...) should be fixed to recognize `isinstance(list)` (as is already documented), and then CMake.test -> CMake._build should pass in `env=["conanbuild","conanrun"]`, rather than just "conanbuild".

https://github.com/conan-io/conan/blob/5bf9aacedf9cc71c52ef548b079285715b51ba10/conan/tools/cmake/cmake.py#L157-L158
https://github.com/conan-io/conan/blob/5bf9aacedf9cc71c52ef548b079285715b51ba10/conan/tools/cmake/cmake.py#L127

memsharded:
Changelog: Fix: Fix `runenv` for `cmake.test()`.
Docs: Omit

Close https://github.com/conan-io/conan/issues/12462

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -99,7 +99,7 @@ class CMake(object):
         with chdir(self, build_folder):
             self._conanfile.run(command)
 
-    def _build(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
+    def _build(self, build_type=None, target=None, cli_args=None, build_tool_args=None, env=""):
         bf = self._conanfile.build_folder
         is_multi = is_multi_configuration(self._generator)
         if build_type and not is_multi:
@@ -127,7 +127,7 @@ class CMake(object):
         arg_list = " ".join(filter(None, arg_list))
         command = "%s --build %s" % (self._cmake_program, arg_list)
         self._conanfile.output.info("CMake command: %s" % command)
-        self._conanfile.run(command)
+        self._conanfile.run(command, env=env)
 
     def build(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
         self._build(build_type, target, cli_args, build_tool_args)
@@ -149,7 +149,7 @@ class CMake(object):
         self._conanfile.output.info("CMake command: %s" % command)
         self._conanfile.run(command)
 
-    def test(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
+    def test(self, build_type=None, target=None, cli_args=None, build_tool_args=None, env=""):
         if self._conanfile.conf.get("tools.build:skip_test", check_type=bool):
             return
         if not target:
@@ -157,5 +157,7 @@ class CMake(object):
             is_ninja = "Ninja" in self._generator
             target = "RUN_TESTS" if is_multi and not is_ninja else "test"
 
+        # The default for ``test()`` is both the buildenv and the runenv
+        env = ["conanbuild", "conanrun"] if env is "" else env
         self._build(build_type=build_type, target=target, cli_args=cli_args,
-                    build_tool_args=build_tool_args)
+                    build_tool_args=build_tool_args, env=env)
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -158,6 +158,6 @@ class CMake(object):
             target = "RUN_TESTS" if is_multi and not is_ninja else "test"
 
         # The default for ``test()`` is both the buildenv and the runenv
-        env = ["conanbuild", "conanrun"] if env is "" else env
+        env = ["conanbuild", "conanrun"] if env == "" else env
         self._build(build_type=build_type, target=target, cli_args=cli_args,
                     build_tool_args=build_tool_args, env=env)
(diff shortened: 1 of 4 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -149,13 +149,15 @@ def install(self, build_type=None):
...
+    def test(self, build_type=None, target=None, cli_args=None, build_tool_args=None, env=""):
         if self._conanfile.conf.get("tools.build:skip_test", check_type=bool):
             return
         if not target:
             is_multi = is_multi_configuration(self._generator)
             is_ninja = "Ninja" in self._generator
             target = "RUN_TESTS" if is_multi and not is_ninja else "test"
 
+        # The default for ``test()`` is both the buildenv and the runenv
+        env = ["conanbuild", "conanrun"] if env is "" else env
```
czoido:
```suggestion
        env = ["conanbuild", "conanrun"] if env == "" else env
```

czoido (approved):

PR #12598, opened 2022-11-24
editables env-vars and confs

### LINKED ISSUE #12499: Define abstract env-vars in package_info() for editable and unix_path
Comes from https://github.com/conan-io/conan/issues/12192#issuecomment-1308589581_

Some packages want to define env-vars with paths, and they often use ``self.package_folder`` for it, so this is broken for ``editable`` packages, for example, in which ``package_folder=None``. It also doesn't work for other abstractions over paths as ``unix_paths``, which depend on the consumer, not the package creator.

Let's investigate some mechanism that allow to define abstract env-vars for PATHs that are resolved better in the consumer side.

memsharded:
Changelog: Feature: Implement ``editable`` env-vars via ``layouts.xxx.buildenv_info`` and ``layouts.xxx.runenv_info`` and conf via ``layouts.xxx.conf_info``.
Docs: https://github.com/conan-io/docs/pull/2834

Close https://github.com/conan-io/conan/issues/12499

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -154,6 +154,12 @@ class _EnvValue:
         previous_value = os.getenv(self._name)
         return self.get_str(previous_value, subsystem, pathsep)
 
+    def set_relative_base_folder(self, folder):
+        if not self._path:
+            return
+        self._values = [os.path.join(folder, v) if v != _EnvVarPlaceHolder else v
+                        for v in self._values]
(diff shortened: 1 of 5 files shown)


franramirez688 (approved):
It looks so simple and harmless! 👏

czoido (approved):

jcar87 (approved):

PR #12609, opened 2022-11-25
Set some CMakeDeps properties from consumer

### LINKED ISSUE #7118: [question] Is it possible to select different generators for different packages?
We would like to use Conan with a CMake project.

For dependencies that provide package-config file or a find-module file (`<Package>Config.cmake`, `Find<Package>.cmake`) we would like to use `cmake_paths` generator.

In other cases we would like to generate find-module files  with `cmake_find_package` generator.

In a mixed situation [the solution is to use both](https://jfreeman.dev/blog/2019/05/22/trying-conan-with-modern-cmake:-dependencies/#both) `cmake_paths` and `cmake_find_package` generators and explicitly tell CMake which mode to use for `find_package`:
```CMake
find_package(pcf_package CONFIG REQUIRED)
find_package(fm_package MODULE REQUIRED)
```
Another option is to use `CMAKE_FIND_PACKAGE_PREFER_CONFIG`.

But wouldn't it be great instead to tell Conan which generator to use for which package?
Something like:
```bash
[generators]
cmake_find_package # default
packageA:cmake_paths # prefer package A's package-config file
```

**Question:**
Is something like this possible? 
Is there a better way to handle such mixed situations? 
If not, would that be something worthy of a feature-request?

Thank you!

### LINKED ISSUE #12534: [question] CMake specifics polluting cpp_info

It seems to me that most package recipes are now setting many cmake generator specific entries on the self.cpp_info. This seems wrong from a conceptual perspective to me. Shouldn't the cpp_info only hold information about the libraries, directories and flags in a generic way?

The package creator cannot know which build system the package consumer will be using, so setting cmake specifics on the cpp_info for the case when the package consumer does use cmake feels very weird to me.

For example: self.cpp_info.set_property("cmake_find_mode", "both")

Why should the package creator care which cmake_find_mode is used by the consumer? Shouldn't the package consumer decide which cmake_find_mode is used? And if the package consumer uses meson or something else, this setting is useless anyway?

czoido:
Changelog: Feature: Provide the ability to set CMakeDeps properties from consumer side.
Docs: https://github.com/conan-io/docs/pull/2827

Gives you the possibility to overwrite some properties set in dependencies in CMakeDeps: `cmake_file_name`, `cmake_target_name`, `cmake_find_mode`, `cmake_module_file_name` and `cmake_module_target_name`

The interface is:

```
def generate(self):
    deps = CMakeDeps(self)
    deps.set_property("<dep>", "cmake_find_mode", "module")
    deps.set_property("<dep>::<component>",  "cmake_target_name", "mycompname")
    deps.generate()
```
for the cases with `build_context_activated`:

```
def generate(self):
    ...
    deps = CMakeDeps(self)
    deps.set_property("<dep>", "cmake_find_mode", "module", build_context=True)
    deps.set_property("<dep>::<component>",  "cmake_target_name", "mycompname", build_context=True)
    deps.generate()
    ...
```

You can also invalidate from consumer the properties set upstream assigning a `None`:

```
def generate(self):
    ...
    deps.set_property("<dep>", "cmake_module_target_name",  None)
```

Closes: https://github.com/conan-io/conan/issues/12534
Close https://github.com/conan-io/conan/issues/7118

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -119,7 +119,7 @@ class CMakeDeps(object):
 
     def get_property(self, prop, dep):
         try:
-            return self._properties[dep].get(prop)
+            return self._properties[dep].get(prop) or dep.get_property(prop)
         except KeyError:
             return dep.get_property(prop)
 
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -115,12 +115,12 @@ class CMakeDeps(object):
             ret[config.filename] = config.render()
 
     def set_property(self, dep, prop, value, build_context=False):
-        build_suffix = "_build" if build_context else ""
+        build_suffix = "&build" if build_context else ""
         self._properties.setdefault(f"{dep}{build_suffix}", {}).update({prop: value})
 
     def get_property(self, prop, dep, comp_name=None):
         dep_name = dep.ref.name
-        build_suffix = "_build" if str(
+        build_suffix = "&build" if str(
             dep_name) in self.build_context_activated and dep.context == "build" else ""
         dep_comp = f"{str(dep_name)}::{comp_name}" if comp_name else f"{str(dep_name)}"
         try:
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -118,31 +118,31 @@ class CMakeDeps(object):
     def set_property(self, dependency, prop, value):
         self._properties[dependency].update({prop: value})
 
-    def get_property(self, prop, dependency, component=None):
-        dep_property = dependency.cpp_info.get_property(prop) if not component else \
-        dependency.cpp_info.components[component].get_property(prop)
-        dep_and_comp = dependency.ref.name if not component else f"{dependency.ref.name}::{component}"
+    def get_property(self, prop, dep, component=None):
+        dep_property = dep.cpp_info.get_property(prop) if not component else \
+        dep.cpp_info.components[component].get_property(prop)
+        dep_and_comp = dep.ref.name if not component else f"{dep.ref.name}::{component}"
         return self._properties[dep_and_comp].get(prop) or dep_property
 
-    def get_cmake_package_name(self, req, module_mode=None):
+    def get_cmake_package_name(self, dep, module_mode=None):
         """Get the name of the file for the find_package(XXX)"""
         # This is used by CMakeDeps to determine:
         # - The filename to generate (XXX-config.cmake or FindXXX.cmake)
         # - The name of the defined XXX_DIR variables
         # - The name of transitive dependencies for calls to find_dependency
-        if module_mode and self.get_find_mode(req) in [FIND_MODE_MODULE, FIND_MODE_BOTH]:
-            ret = self.get_property("cmake_module_file_name", req)
+        if module_mode and self.get_find_mode(dep) in [FIND_MODE_MODULE, FIND_MODE_BOTH]:
+            ret = self.get_property("cmake_module_file_name", dep)
             if ret:
                 return ret
-        ret = self.get_property("cmake_file_name", req)
-        return ret or req.ref.name
+        ret = self.get_property("cmake_file_name", dep)
+        return ret or dep.ref.name
 
-    def get_find_mode(self, req):
+    def get_find_mode(self, dep):
         """
-        :param req: requirement
+        :param dep: requirement
         :return: "none" or "config" or "module" or "both" or "config" when not set
         """
-        tmp = self.get_property("cmake_find_mode", req)
+        tmp = self.get_property("cmake_find_mode", dep)
         if tmp is None:
             return "config"
         return tmp.lower()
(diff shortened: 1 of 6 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, conanfile):
 
         # Enable/Disable checking if a component target exists or not
         self.check_components_exist = False
+        self._properties = defaultdict(dict)
```
memsharded: I don't love the defaultdict, because it can easily create entries for non-existing dependencies, which is valid python, but non logical model. I typically prefer to resort to regular dict and the idiom ``mydict.setdefault(key, default_value).update...``

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -112,3 +114,35 @@ def _generate_files(self, require, dep, ret, find_module_mode):
...
+    def set_property(self, dep, prop, value):
+        self._properties[dep].update({prop: value})
+
+    def get_property(self, prop, dep, component=None):
+        dep_property = dep.cpp_info.get_property(prop) if not component else \
+        dep.cpp_info.components[component].get_property(prop)
+        dep_and_comp = dep.ref.name if not component else f"{dep.ref.name}::{component}"
+        return self._properties[dep_and_comp].get(prop) or dep_property
+
+    def get_cmake_package_name(self, dep, module_mode=None):
```
memsharded, line 132: Nice refactor 👍

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -112,3 +114,35 @@ def _generate_files(self, require, dep, ret, find_module_mode):
         # file is common for the different configurations.
         if not os.path.exists(config.filename):
             ret[config.filename] = config.render()
+
+    def set_property(self, dep, prop, value):
+        self._properties[dep].update({prop: value})
```
memsharded: Using the bare dependency name might fail if there are 2 dependencies with same name as dependencies in the build and host context (``build_context_activated``) with same reference?

conans/test/integration/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -187,3 +187,70 @@ def package_info(self):
...
+        from conan import ConanFile
+        from conan.tools.cmake import CMakeDeps, cmake_layout
+        class FooConan(ConanFile):
+            settings = "os", "compiler", "build_type", "arch"
+            requires = "bar/1.0"
+            def layout(self):
+                cmake_layout(self)
+            def generate(self):
+                deps = CMakeDeps(self)
+                deps.set_property("bar", "cmake_find_mode", "{find_mode}")
```
memsharded:
```suggestion
                deps.set_property(self.dependencies["bar"], "cmake_find_mode", "{find_mode}")
```
This might dissambiguate, I know it is not as clean and nice, just a rough idea, but to consider how to resolve name collisions. Other alternatives:
- ``deps.set_property(...., build_context=True)``
- Allow both ``str`` and ``ConanFileInterface`` from ``self.dependencies`` as argument.
- ?

memsharded (commented):
Looks good, clean and promising to deliver something useful.
A couple minor things to discuss, good job.

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -112,3 +113,35 @@ def _generate_files(self, require, dep, ret, find_module_mode):
         # file is common for the different configurations.
         if not os.path.exists(config.filename):
             ret[config.filename] = config.render()
+
+    def set_property(self, dep, prop, value):
+        self._properties.setdefault(dep, {}).update({prop: value})
+
+    def get_property(self, prop, dep):
+        try:
+            return self._properties[dep].get(prop) or dep.get_property(prop)
```
memsharded: ``cmakedeps.set_property(..., property_name, None)`` or ``value=""`` too) will still fallback to the dependency value? I think there should be a way to invalidate/clear the upstream value too. So as long as it doesn't ``KeyError``, I'd respect the local value?

SpaceIm:
Looks very promising for conan-center as well, to avoid breakage in recipes when their dependencies break these properties (for example when suddenly a new version of a library implements a CMake config file, we have to update these names in this recipe, but it can break downstream recipes which were relying on default values `<recipeName>-config.cmake` & target `<recipeName>::<recipeName>`).

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -112,3 +113,41 @@ def _generate_files(self, require, dep, ret, find_module_mode):
...
         if not os.path.exists(config.filename):
             ret[config.filename] = config.render()
+
+    def set_property(self, dep, prop, value, build_context=False):
+        build_suffix = "_build" if build_context else ""
+        self._properties.setdefault(f"{dep}{build_suffix}", {}).update({prop: value})
+
+    def get_property(self, prop, dep, comp_name=None):
+        dep_name = dep.ref.name
+        build_suffix = "_build" if str(
```
memsharded:
```suggestion
        build_suffix = "&build" if str(
```

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -112,3 +113,41 @@ def _generate_files(self, require, dep, ret, find_module_mode):
         # file is common for the different configurations.
         if not os.path.exists(config.filename):
             ret[config.filename] = config.render()
+
+    def set_property(self, dep, prop, value, build_context=False):
+        build_suffix = "_build" if build_context else ""
```
memsharded:
```suggestion
        build_suffix = "&build" if build_context else ""
```

Better use something that is an invalid package name, because if some user has a "host" package called ``mypkg_build`` and a tool called ``mypkg``, this will error.

memsharded (approved):

franramirez688 (approved):
It looks great! 👏

jcar87 (approved):
This is amazing!!!

Issue #9751, opened 2021-10-06
Unhelpful error message (ERROR: 'NoneType' object is not iterable.) when working with faulty (gitlab) remote and range requirement


a-kryuk:
When I try to 'conan install' version-range-based requirements for a package from a GitLab remote package repository, I get an unhelpful Python message "ERROR: 'NoneType' object is not iterable." I believe it's triggered by some faulty implementation on the GitLab repository side, but it would be helpful if Conan provided a more specific indication of the error cause. 
My example conanfile.py:
```python
from conans import ConanFile

class BadConan(ConanFile):
  name = "badconan"
  version = "1.0.0"
  description = "Demonstrate problem"
  requires = [
    "pkgA/[>=1.0]@user/stream"
  ]
```
And here's the (edited) output:
```
$ conan --version
Conan version 1.40.3

$ export CONAN_LOGGING_LEVEL=10

$ conan install .
DEBUG :conan_api.py   [176]: INIT: Using config '/home/user/.conan/conan.conf' [2021-10-06 11:23:34,400]
DEBUG :tracer.py      [156]: CONAN_API: install(path=.,name=None,version=None,user=None,channel=None,settings=None,options=None,env=None,profile_names=None,conf=None,profile_build=ProfileData(profiles=None, settings=None, options=None, env=None, conf=None),remote_name=None,verify=None,manifests=None,manifests_interactive=None,build=None,update=False,generators=None,no_imports=False,install_folder=None,lockfile=None,lockfile_out=None,require_overrides=None) [2021-10-06 11:23:34,402]
DEBUG :profile_loader.py[120]: PROFILE LOAD: /home/user/.conan/profiles/default [2021-10-06 11:23:34,403]
DEBUG :profile_loader.py[120]: PROFILE LOAD: /home/user/.conan/profiles/default [2021-10-06 11:23:34,440]
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=9
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

DEBUG :rest_client_common.py[160]: REST: ping: https://center.conan.io/v1/ping [2021-10-06 11:23:34,446]
DEBUG :rest_client.py [58]: REST: Cached capabilities for the remote: ['complex_search', 'checksum_deploy', 'revisions', 'matrix_params'] [2021-10-06 11:23:34,538]
[...]

PR #12633, opened 2022-11-30
Add context for error; fix for Issue #9751

adam248:
Changelog: omit
Docs: omit

### DIFF
--- a/conans/client/rest/rest_client_common.py
+++ b/conans/client/rest/rest_client_common.py
@@ -242,7 +242,10 @@ class RestCommonMethods(object):
         """
         url = self.router.search(pattern, ignorecase)
         response = self.get_json(url)["results"]
-        return [ConanFileReference.loads(reference) for reference in response]
+        try:
+            return [ConanFileReference.loads(reference) for reference in response]
+        except TypeError as te:
+            raise Exception("Unexpected response from server url: `%s`." % url)
 
     def search_packages(self, ref):
         """Client is filtering by the query"""
--- a/conans/client/rest/rest_client_common.py
+++ b/conans/client/rest/rest_client_common.py
@@ -245,7 +245,9 @@ class RestCommonMethods(object):
         try:
             return [ConanFileReference.loads(reference) for reference in response]
         except TypeError as te:
-            raise Exception("Unexpected response from server url: `%s`." % url)
+            raise ConanException("Unexpected response from server.\n"
+                                 "URL: `{}`\n"
+                                 "Expected an iterable, but got {}.".format(url, type(response)))
 
     def search_packages(self, ref):
         """Client is filtering by the query"""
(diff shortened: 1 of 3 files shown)


adam248:
Fixes Issue #9751

conans/client/rest/rest_client_common.py
```diff
@@ -242,7 +242,10 @@ def search(self, pattern=None, ignorecase=True):
         """
         url = self.router.search(pattern, ignorecase)
         response = self.get_json(url)["results"]
-        return [ConanFileReference.loads(reference) for reference in response]
+        try:
+            return [ConanFileReference.loads(reference) for reference in response]
+        except TypeError as te:
+            raise Exception("Unexpected response from server url: `%s`." % url)
```
memsharded: It would probably be useful to display the ``response`` too, to give some idea of what could be happening.

memsharded (commented):
It would be nice to add some test to the suite that covers this, there are some tests mocking the server responses, probably they can be extended. Have a look if possible, and if not, let us know, we might try to contribute it.

conans/client/rest/rest_client_common.py
adam248: I agree. However, I believe that the `TypeError` raised alongside the custom `Exception` would show the `response` type. (It is my understanding that both Exceptions will be included in the Traceback.) In this issue's case; None, or any other type that is not iterable in a list-comp. I felt that the `response` type, along with the `url`, would be sufficient information to assist with a debug and I was trying to avoid producing redundant information.
Happy to reconsider though... 👍

adam248:
> It would be nice to add some test to the suite that covers this, there are some tests mocking the server responses, probably they can be extended. Have a look if possible, and if not, let us know, we might try to contribute it.

Will look into writing a test once you are happy with my Exception handling methodology, just let us know and I will try to add it to this pull request.

conans/client/rest/rest_client_common.py
memsharded: > I agree. However, I believe that the TypeError raised alongside the custom Exception would show the response type. (It is my understanding that both Exceptions will be included in the Traceback.)

Many exceptions do not display a stack trace. As Conan is more an app than a library, we try to hide internal implementation details as much as possible, and most of the exceptions are captured into a ``ConanException`` that only displays the error message, but not the stack trace (I am not sure in this case if that Exception might be captured somewhere into a ConanException). The idea is that in most cases, if Conan is raising an exception that it knows what is happening, it should raise a ConanException. So it is always better try to provide an error message that is as user-oriented and self-contained as possible. In this case showing the response could help, because there are sometimes that the URL is incorrect, and return html, or there is a proxy returning something in the response too. IN those cases, printing that to the user might be useful information.

adam248: Ah, I got it. User-friendly error messages, not dev friendly. Will refactor, and look into using a `ConanException` instead. Thanks.

adam248:
I believe all issues have been addressed with my final commit. 
Good to go❓

memsharded (approved):

memsharded:
Thanks for your contribution! Merged, this will be in conan 1.56

PR #12635, opened 2022-11-30
[conf] objcpp instead of objcxx

franramirez688:
Changelog: omit
Docs: omit

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -682,7 +682,7 @@ class CompilersBlock(Block):
         compilers = {}
         # Allowed <LANG> variables (and <LANG>_LAUNCHER)
         compilers_mapping = {"c": "C", "cuda": "CUDA", "cpp": "CXX", "objc": "OBJC",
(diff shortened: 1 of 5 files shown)


conan/tools/cmake/toolchain/blocks.py
```diff
@@ -682,7 +682,7 @@ def context(self):
         compilers = {}
         # Allowed <LANG> variables (and <LANG>_LAUNCHER)
         compilers_mapping = {"c": "C", "cuda": "CUDA", "cpp": "CXX", "objc": "OBJC",
-                             "objcxx": "OBJCXX", "rc": "RC", 'fortran': "Fortran", 'asm': "ASM",
+                             "objcpp": "OBJCXX", "rc": "RC", 'fortran': "Fortran", 'asm': "ASM",
```
franramirez688, line 685: it was weird to have `cpp` and `objcxx`

czoido: Maybe we should have a small test that covers that the mapping is good ?

memsharded (approved):

czoido (approved):

PR #12639, opened 2022-11-30
[develop2] test that it is possible to overwrite builtin commands

memsharded:


timviola-cmi:
@memsharded Looks good. Let me know if I'm understanding this correctly. Looks like the user adds the  `@conan_command()` decorator to a function with the proper signature and stores that in the commands/<customdir> folder of the local conan cache. Also, for invocation the user enters `conan <customdir>:install ...` instead of the traditional `conan install ...`. What we've proved here is the user can create a custom command with the same name as a builtin command and it works.

memsharded:
> [memsharded](https://github.com/memsharded) Looks good. Let me know if I'm understanding this correctly. Looks like the user adds the @conan_command() decorator to a function with the proper signature and stores that in the commands/ folder of the local conan cache. Also, for invocation the user enters conan <customdir>:install ... instead of the traditional conan install .... What we've proved here is the user can create a custom command with the same name as a builtin command and it works.

Actually both are possible:

- If you put the command inside a subfolder "myteam", the command will be ``conan myteam:install`` and will not overwrite/shadow the builtin ``conan install``
- If you  put it in the commands root folder, then it will overwrite the ``conan install`` builtin one.

I'd tend to favor the first one, leaving the built-in there, and making it explicit the command is our specific custom one.

franramirez688 (approved):

jcar87 (approved):

PR #12665, opened 2022-12-06
[NMakeToolchain] Refactoring to expose similar interface than other toolchains & honor build config

SpaceIm:
Changelog: Fix: Refactoring of `NMakeToolchain` to expose similar attributes than other toolchains, and honor build config like cflags, cxxflags, sharedlinkflags, exelinkflags, defines & compiler_executables.
Docs: https://github.com/conan-io/docs/pull/2848

### DIFF
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -1,5 +1,6 @@
 from conan.tools._compilers import cppstd_flag, build_type_flags, build_type_link_flags
 from conan.tools.env import Environment
+from conan.tools.microsoft.visual import msvc_runtime_flag, VCVars
 
 
 class NMakeToolchain(object):
@@ -35,7 +36,6 @@ class NMakeToolchain(object):
         self.msvc_runtime_flag = self._get_msvc_runtime_flag()
 
     def _get_msvc_runtime_flag(self):
-        from conan.tools.microsoft import msvc_runtime_flag
         flag = msvc_runtime_flag(self._conanfile)
         if flag:
             flag = f"/{flag}"
@@ -109,5 +109,4 @@ class NMakeToolchain(object):
         env = env or self.environment()
         env = env.vars(self._conanfile, scope=scope)
         env.save_script("conannmaketoolchain")
-        from conan.tools.microsoft import VCVars
         VCVars(self._conanfile).generate(scope=scope)
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -81,14 +81,19 @@ class NMakeToolchain(object):
         defines = [f"/D{d.replace('=', '#')}" for d in self.defines]
         return nologo + self.cxxflags + self._curate_options(conf_cflags) + defines
 
+    @property
+    def _link(self):
+        nologo = ["/NOLOGO"]
+        return nologo + self.ldflags
+
     def environment(self):
         env = Environment()
         # Injection of compile flags in CL env-var:
         # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
         env.append("CL", self._cl)
-        # Injection of link flags in LINK env-var:
+        # Injection of link flags in _LINK_ env-var:
         # https://learn.microsoft.com/en-us/cpp/build/reference/linking
-        env.append("LINK", self.ldflags)
+        env.append("_LINK_", self._link)
         # Also define some special env-vars which can override special NMake macros:
         # https://learn.microsoft.com/en-us/cpp/build/reference/special-nmake-macros
         conf_compilers = self._conanfile.conf.get("tools.build:compiler_executables", default={}, check_type=dict)
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -44,6 +44,19 @@ class NMakeToolchain(object):
     def _curate_options(self, options):
         return [f"{opt[0].replace('-', '/')}{opt[1:]}" for opt in options if len(opt) > 1]
 
+    def _defines_for_cl(self, defines):
+        curated_defines = []
+        for define in defines:
+            if "=" in define:
+                macro, value = define.split("=", 1)
+                if value.isnumeric():
+                    curated_defines.append(f"/D{macro}#{value}")
+                else:
+                    curated_defines.append(f"/D{macro}#\\\"{value}\\\"")
+            else:
+                curated_defines.append(f"/D{define}")
+        return curated_defines
+
     @property
     def cflags(self):
         bt_flags = self.build_type_flags if self.build_type_flags else []
@@ -78,8 +91,7 @@ class NMakeToolchain(object):
         conf_cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list)
         # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
         # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
-        defines = [f"/D{d.replace('=', '#')}" for d in self.defines]
-        return nologo + self.cxxflags + self._curate_options(conf_cflags) + defines
+        return nologo + self.cxxflags + self._curate_options(conf_cflags) + self._defines_for_cl(self.defines)
 
     @property
     def _link(self):
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -35,7 +35,16 @@ class NMakeToolchain(object):
         flag = msvc_runtime_flag(self._conanfile)
         if flag:
             cppflags.append("-{}".format(flag))
+        cppflags.extend(self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list))
+        cppflags.extend(self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list))
         return " ".join(cppflags).replace("-", "/")
+
+    @property
+    def link_flags(self):
+        linkflags = []
+        linkflags.extend(self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list))
+        linkflags.extend(self._conanfile.conf.get("tools.build:exelinkflags", default=[], check_type=list))
+        return " ".join(linkflags).replace("-", "/")
 
     @property
     def environment(self):
@@ -44,6 +53,7 @@ class NMakeToolchain(object):
             env = Environment()
             # The whole injection of toolchain happens in CL env-var, the others LIBS, _LINK_
             env.append("CL", self.cl_flags)
+            env.append("_LINK_", self.link_flags)
             self._environment = env
         return self._environment
 
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -99,7 +99,7 @@ class NMakeToolchain(object):
     @property
     def _link(self):
         nologo = ["/nologo"]
-        return nologo + self.ldflags
+        return nologo + self._ldflags
 
     def environment(self):
         env = Environment()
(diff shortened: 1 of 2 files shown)


conan/tools/microsoft/nmaketoolchain.py
```diff
@@ -15,42 +16,122 @@ def __init__(self, conanfile):
...
+            compilers_mapping = {
+                "AS": "asm",
+                "CC": "c",
+                "CPP": "cpp",
+                "CXX": "cpp",
+                "RC": "rc",
+            }
+            for env_var, comp in compilers_mapping.items():
+                if comp in conf_compilers:
+                    env.define(env_var, conf_compilers[comp])
```
SpaceIm, line 106: Out of scope of this PR, but I believe this toolchain should set all these env vars by default for clang-cl if their corresponding tools.build:compiler_executables key is not set.

It would be nice in order to avoid this kind of fragile logic (and broken for clang mingw):
https://github.com/conan-io/conan-center-index/blob/2ecc63a88002b1294f0ea0650f27c6012933aafa/recipes/libjpeg/all/conanfile.py#L33-L35
https://github.com/conan-io/conan-center-index/blob/2ecc63a88002b1294f0ea0650f27c6012933aafa/recipes/libjpeg/all/conanfile.py#L106-L114

SpaceIm:
@memsharded could you review this PR please? It would be nice to guarantee that we can use NMakeToolchain in conan-center and hope honoring all conf from profile to avoid discrepancies in recipes behavior.

conan/tools/microsoft/nmaketoolchain.py
```diff
@@ -15,42 +16,122 @@ def __init__(self, conanfile):
...
+        env.append("CFLAGS", self._cflags)
+        env.append("CPPFLAGS", self._cxxflags)
+        env.append("CXXFLAGS", self._cxxflags)
```
memsharded: Why defining these variables? If they are used, it will be a duplication of flags (they are already defined in ``CL``). If they are not used, why defining them?

conans/test/functional/toolchains/test_nmake_toolchain.py
```diff
@@ -21,11 +29,32 @@ def test_toolchain_nmake(compiler, version, runtime, cppstd, build_type):
...
+    conf_preprocessors = {}
+    for define in defines:
+        if "=" in define:
+            key, value = define.split("=", 1)
+            if not value:
+                conf_preprocessors[key] = "1"
+            else:
+                conf_preprocessors[key] = value
+        else:
+            conf_preprocessors[define] = "1"
```
memsharded: instead of re-arranging, provide the defines already in a dict, and compose it into list when necessary

conans/test/functional/toolchains/test_nmake_toolchain.py
```diff
@@ -41,17 +70,20 @@ def build(self):
...
                  "makefile": makefile,
-                 "simple.cpp": gen_function_cpp(name="main", includes=["dep"], calls=["dep"])},
+                 "simple.cpp": gen_function_cpp(name="main", includes=["dep"], calls=["dep"], preprocessor=conf_preprocessors.keys())},
                 clean_first=True)
-    client.run("install . {}".format(settings))
+    client.run(f"install . {settings} {conf}")
     client.run("build .")
     client.run_command("simple.exe")
     assert "dep/1.0" in client.out
+    assert f"main: {'Debug' if build_type == 'Debug' else 'Release'}!" in client.out
```
memsharded: Debug/Release message already checked in ``check_exe_run()``

conans/test/functional/toolchains/test_nmake_toolchain.py
```diff
@@ -41,17 +70,20 @@ def build(self):
...
+    for preprocessor_name, preprocessor_value in conf_preprocessors.items():
```
memsharded: ``check_exe_run(..., definitions={dict})`` will already do this inside the function.

conan/tools/microsoft/nmaketoolchain.py
memsharded, line 106: NMake not listening to these CC/CXX, etc env-vars, only to CL

conan/tools/microsoft/nmaketoolchain.py
```diff
@@ -15,42 +16,122 @@ def __init__(self, conanfile):
...
-        if bt:
-            cppflags.extend(bt)
+        self._cppstd = cppstd_flag(self._conanfile.settings)
+        self._msvc_runtime_flag = self._get_msvc_runtime_flag()
 
-        cppstd = cppstd_flag(self._conanfile.settings)
-        if cppstd:
-            cppflags.append(cppstd)
-        from conan.tools.microsoft import msvc_runtime_flag
+    def _get_msvc_runtime_flag(self):
```
memsharded: Seems unnecessary, the formatting of flag is doing nothing, the ``msvc_runtime_flag()`` can be used directly instead of this method.

conan/tools/microsoft/nmaketoolchain.py
SpaceIm: It's not what microsoft documentation says: https://learn.microsoft.com/en-us/cpp/build/reference/special-nmake-macros

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: `@pytest.mark.parametrize` cannot take dict values AFAIK. Do you have an example?

conan/tools/microsoft/nmaketoolchain.py
SpaceIm: It does something, it prepends `/`. It's just a copy paste of an autotoolstoolchain private method.

conan/tools/microsoft/nmaketoolchain.py
memsharded: I tried to make it work while developing the NMakeToolchain and for some reason it didn't work. If it is expected to work then the way to move this forward is:

- remove the injection of flags into ``CL`` and use the other options flags as their own env-vars, 
- modify ``NMakeDeps`` to be consistent with this
- check that tests correctly work, this test is not validating anywhere that flags ``["/GL"], ["/GL"], ["/LTCG"], ["/LTCG"]`` are actually being honored

conans/test/functional/toolchains/test_nmake_toolchain.py
memsharded:
```python
[{"TEST_DEFINITION1": None}, {"TEST_DEFINITION2": "0"}, .........],
```

conans/test/functional/toolchains/test_nmake_toolchain.py
```diff
@@ -8,11 +8,19 @@
...
-@pytest.mark.parametrize("compiler, version, runtime, cppstd, build_type",
-                         [("msvc", "190", "dynamic", "14", "Release"),
-                          ("msvc", "191", "static", "17", "Debug")])
+@pytest.mark.parametrize(
+    "compiler, version, runtime, cppstd, build_type, defines, cflags, cxxflags, sharedlinkflags, exelinkflags",
+    [
+        ("msvc", "190", "dynamic", "14", "Release", [], [], [], [], []),
+        ("msvc", "190", "dynamic", "14", "Release",
+         ["TEST_DEFINITION1", "TEST_DEFINITION2=0", "TEST_DEFINITION3=", "TEST_DEFINITION4=TestPpdValue4"],
+         ["/GL"], ["/GL"], ["/LTCG"], ["/LTCG"]),
```
memsharded, line 17: Who checks and validates that these ["/GL"], ["/GL"], ["/LTCG"], ["/LTCG"] flags are being honored?

conan/tools/microsoft/nmaketoolchain.py
SpaceIm: I guess they can be removed yes, as you wish. They are official documented macro of NMake: https://learn.microsoft.com/en-us/cpp/build/reference/special-nmake-macros?view=msvc-170

But defining them may improve robustness, I'm not sure. I'm pretty sure it doesn't hurt at least. It deserves extra experimentations since microsoft documentation is poor. Maybe remove them and left a comment?

By the way, from what I understand from microsoft documentation, CL and LINK are not specific to NMake but to cl & link.

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: I don't know how to test that in produced binaries. I'm not sure that it's really tested in any other functional test of conan client actually. The best I could do is a unittest to check whether these flags are passed to CL.

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: I think I've tried and it didn't work. I can try again.

conans/test/functional/toolchains/test_nmake_toolchain.py
memsharded: Yes, that is true, it is difficult. I have checked other tests for other build systems, and they do check just reading the generated files or the command line. Still worth it, at least guarantee they are there.

SpaceIm:
`NMakeDeps` must also be fixed, it doesn't inject `cpp_info.defines` and `cpp_info.system_libs` of dependencies. I'm pretty sure it's the reason of these errors in https://github.com/conan-io/conan-center-index/pull/15386 (since `cpp_info.defines` are not propagated and GDAL depends on libjpeg static by default in c3i builds, `LIBJPEG_STATIC` is not injected, therefore GDAL compiles with a wrong mangling of libjpeg interface induced by `_delcspec(dllimport)`, then it fails afterwards at link time in test package).
I've opened another PR: https://github.com/conan-io/conan/pull/12944

conan/tools/microsoft/nmaketoolchain.py
```diff
@@ -15,42 +16,122 @@ def __init__(self, conanfile):
...
+
+    def _sanitized_defines(self, defines):
+        sanitized_defines = []
+        for define in defines:
+            if "=" in define:
+                # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
+                # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
+                macro, value = define.split("=", 1)
+                if not value:
+                    sanitized_defines.append(f"/D{macro}")
```
SpaceIm: I've found an issue here while consuming `proj` recipe (it defines `PROJ_DLL=` if static) in a NMake build, `FOO=` definition must be translated to `/DFOO#` actually.

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: Ok, I'll see what I can do

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: fixed

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: fixed

conans/test/functional/toolchains/test_nmake_toolchain.py
SpaceIm: Actually a change like this wouldn't be robust enough.
If I try to inject `/DTEST_DEFINITION1` for `{"TEST_DEFINITION1": None}` as an input, it won't be understood like this by `check_exe_run()` which would expect `{"TEST_DEFINITION1": "1"}`. `/DTEST_DEFINITION1` is equivalent to `/DTEST_DEFINITION1=1`, but I also want to test that `/DTEST_DEFINITION1` AND `/DTEST_DEFINITION1=1` work in `NMakeToolchain`. I even want to test that `/DTEST_DEFINITION1=` works. How to disambiguate when I serialize this dictionary for `tools.build:defines`, and even after when I pass this dictionary to `check_exe_run`?
Moreover source code generated by `gen_function_cpp()` is not robust to a true empty definition like `/DTEST_DEFINITION1=`.

conan/tools/microsoft/nmaketoolchain.py
SpaceIm: I've removed CFLAGS, CXXFLAGS and CPPFLAGS, they have no effect by default indeed. I've kept CC & CXX & CPP, I'm pretty sure these ones are listened. By default CPP is `cl` and if compiler is clang-cl, CPP must be changed.

memsharded (approved):

PR #12666, opened 2022-12-06
[develop2] moving global graph_compute to GraphAPI

memsharded:
It is a bad example for readers of our code that we have a global, free function for something that looks better as part of the PythonAPI


conan/cli/commands/install.py
```diff
@@ -109,21 +42,40 @@ def install(conan_api, parser, *args):
...
+                                               cwd=cwd,
+                                               partial=args.lockfile_partial)
+    profile_host, profile_build = conan_api.profiles.get_profiles_from_args(args)
+    if path:
+        deps_graph = conan_api.graph.load_graph_consumer(path, args.name, args.version,
+                                                         args.user, args.channel,
+                                                         profile_host, profile_build, lockfile,
+                                                         remotes, args.update)
+    else:
+        deps_graph = conan_api.graph.load_graph_requires(args.requires, args.tool_requires,
```
prince-chrismc, line 73: 👍  This is much better then and giant function above. I going to open an issue for this since the custom commands I was working on but this address the suggestions I had

czoido (approved):

conan/api/subapi/graph.py
```diff
@@ -87,7 +89,7 @@ def load_root_test_conanfile(self, path, tested_reference, profile_host, profile
         return root_node
 
     @api_method
-    def load_root_virtual_conanfile(self, profile_host, requires=None, tool_requires=None):
+    def _load_root_virtual_conanfile(self, profile_host, requires=None, tool_requires=None):
```
prince-chrismc, line 92: Should this be marked as an API method and have an underscore? I thought underscore meant private function? 🤔

memsharded: Probably not, the "virtual" conanfile seems a bit of implementation detail, I don't see a clear use case of users having to do that, now that we exposed a ``load_graph_consumer(path, ...)`` and ``load_graph_requires(requires, tool_requires...)`` methods. We don't need to do it now, lets keep the API minimal to the construction of our own built-in commands, and open it as necessary when requested by users. So this might be good as private at the moment, it will not be part of the public api.

prince-chrismc: I agree it should be private, thats why I am confused why it's `@api_method` :)

conan/cli/commands/install.py
prince-chrismc: https://github.com/prince-chrismc/conan-center-index/commit/1909a3e875fc9b35ffb5c5b00080ce9de087f7be

This is much better for UX 
image

conan/api/subapi/graph.py
memsharded: Oh, I see, you are right, that ``api_method`` should be removed.

PR #12673, opened 2022-12-07
Fix MockInfoProperty to mimic list interface

jcar87:
Usage of this in recipes:
```
self.env_info.PATH.append("/path/to/folder")
```

is failing in 2.0.0b6 with the following error:
```
'NoneType' object has no attribute 'append'
```

because the `env_info` accessor was not returning an object with a list-like interface. The 1.x implementation would have returned an empty list, so this does the same - still with no consequence as usage of `env_info` in 2.0 is a no-op.

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -50,7 +50,7 @@ class MockInfoProperty(object):
     def __getattr__(self, attr):
         if attr != "_message":
             ConanOutput().warning(self._message)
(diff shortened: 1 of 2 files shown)


memsharded (approved):
Probably is good enough atm, if there are some more failing cases, we can try to provide a more powerful Mock

prince-chrismc:
🚀

Issue #12336, opened 2022-10-18
[feature] Add `is_cl_like()` or improve `is_msvc()` method to perfectly mimic behavior of `MSVC` variable in CMake


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Currently conan provides `is_msvc()` method which returns True if `compiler=Visual Studio` or `compiler=msvc` in settings. In conan-center recipes, this method is widely used for different logics, and one of them is for some branching in package() and package_info() depending on `MSVC` variable in upstream CMakeLists (conditional lib name, renaming of files, injection of definitions etc).
But `MSVC` variable in CMake is not only True for Visual Studio, but also for any cl like compiler like clang-cl: https://cmake.org/cmake/help/latest/variable/MSVC.html
The problem is that depending on clang flavor on Windows, it may or may not be a cl like compiler for CMake. It leads to wrong assumption in several recipes, for example zlib where `cpp_info.libs` may be wrong depending on clang flavor: https://github.com/conan-io/conan-center-index/issues/13474

For example a CMakeLists may have this:

```cmake
add_library(mylib ...)
if(MSVC AND NOT BUILD_SHARED_LIBS)
    set_target_properties(mylib PROPERTIES OUTPUT_NAME "mylib-static")
endif()
```

so that in conan recipe we usually write:

```python
    package_info(self):
        suffix = "-static" if is_msvc(self) and not self.options.shared else ""
        self.cpp_info.libs = [f"mylib{suffix}"]
```

It's very common, and it's broken for clang-cl.

So some folks using clang-cl were disappointed and began to put this kind of snippet in CCI recipes (and this time it's broken for non cl-like clang flavors on windows, so other folks complain that their clang flavor on windows doesn't work anymore):

```python
    @property
    def _is_clang_cl(self):
        return self.settings.os == "Windows" and self.settings.compiler == "clang"

    package_info(self):
        suffix = "-static" if (is_msvc(self) or self._is_clang_cl) and not self.options.shared else ""
        self.cpp_info.libs = [f"mylib{suffix}"]
```

So it would be nice if conan could provide a method to perfectly mimic https://cmake.org/cmake/help/latest/variable/MSVC.html logic. It may be another method than `is_msvc`, or an additional argument to `is_msvc` in order to switch to CMake `MSVC` behavior.

memsharded:
[...]

PR #12678, opened 2022-12-08
Add `conan.tools.gnu.is_mingw()` in conan API

SpaceIm:
Changelog: (Feature): Add conan.tools.gnu.is_mingw() function
Docs: https://github.com/conan-io/docs/pull/2849

Partially implement features requested in https://github.com/conan-io/conan/issues/12336#issuecomment-1342464853. Such function is useful in recipes to know whether a user is using a canonical conan profile describing a MinGW toolchain, since there might be specific branching in a recipe based on this toolchain (using Autotools instead of NMake or MSBuild, deduce information for package info depending on logic in underlying build files etc).

### DIFF
--- a/conan/tools/gnu/mingw.py
+++ b/conan/tools/gnu/mingw.py
@@ -6,9 +6,10 @@ def is_mingw(conanfile):
     """
     host_os = conanfile.settings.get_safe("os")
     host_subsystem = conanfile.settings.get_safe("os.subsystem")
-    is_windows_native = host_os == "Windows" and host_subsystem not in ["cygwin", "wsl"]
+    is_wsl = host_os == "Windows" and host_subsystem == "wsl"
+    is_cygwin = host_os == "Windows" and host_subsystem == "cygwin"
     host_compiler = conanfile.settings.get_safe("compiler")
-    is_mingw_gcc = is_windows_native and host_compiler == "gcc"
-    is_mingw_clang = is_windows_native and host_compiler == "clang" and \
+    is_mingw_gcc = host_os == "Windows" and not (is_wsl or is_cygwin) and host_compiler == "gcc"
+    is_mingw_clang = host_os == "Windows" and not (is_wsl or is_cygwin) and host_compiler == "clang" and \
                      not conanfile.settings.get_safe("compiler.runtime")
     return is_mingw_gcc or is_mingw_clang
--- a/conan/tools/gnu/mingw.py
+++ b/conan/tools/gnu/mingw.py
@@ -5,8 +5,10 @@ def is_mingw(conanfile):
     :return: True, if the host compiler is related to MinGW, otherwise False.
     """
     host_os = conanfile.settings.get_safe("os")
+    host_subsystem = conanfile.settings.get_safe("os.subsystem")
+    is_windows_native = host_os == "Windows" and host_subsystem not in ["cygwin", "wsl"]
     host_compiler = conanfile.settings.get_safe("compiler")
-    is_mingw_gcc = host_os == "Windows" and host_compiler == "gcc"
-    is_mingw_clang = host_os == "Windows" and host_compiler == "clang" and \
+    is_mingw_gcc = is_windows_native and host_compiler == "gcc"
+    is_mingw_clang = is_windows_native and host_compiler == "clang" and \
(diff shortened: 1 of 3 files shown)


conan/tools/gnu/mingw.py
```diff
@@ -0,0 +1,12 @@
...
+    """
+    Validate if current compiler in host setttings is related to MinGW
+    :param conanfile: ConanFile instance
+    :return: True, if the host compiler is related to MinGW, otherwise False.
+    """
+    host_os = conanfile.settings.get_safe("os")
+    host_compiler = conanfile.settings.get_safe("compiler")
+    is_mingw_gcc = host_os == "Windows" and host_compiler == "gcc"
+    is_mingw_clang = host_os == "Windows" and host_compiler == "clang" and \
+                     not conanfile.settings.get_safe("compiler.runtime")
```
SpaceIm, line 12: it's worth noting that I've prefered this condition over `conanfile.settings.get_safe("compiler.libcxx")` because libcxx is removed in `configure()` of pure C recipes, so it's not a reliable compiler attribute in order to disambiguate between clang mingw and llvm-clang/clang-cl.

(but honestly if there was an extra attribute in conan settings of clang compiler in order to set windows flavor, it would be far easier)

conan/tools/gnu/mingw.py
```diff
@@ -0,0 +1,13 @@
+def is_mingw(conanfile):
+    """
+    Validate if current compiler in host setttings is related to MinGW
+    :param conanfile: ConanFile instance
+    :return: True, if the host compiler is related to MinGW, otherwise False.
+    """
+    host_os = conanfile.settings.get_safe("os")
```
CJCombrink, line 7: Also adding my comment here[[1]](https://github.com/conan-io/conan-center-index/pull/18283#discussion_r1250611884):

> This feels like a lot of "work" that will result in false in probably most cases (I don't have numbers but I assume mingw has quite a bit lower market share compared to gcc and msvc")
I guess it would still be better to get the short circuit behaviour ~compared to the old code~ where it returns false as soon as some condition is reached that makes it false.

I would propose something like early return:
```
def is_mingw(conanfile):
    if conanfile.settings.get_safe("os") != "Windows":
        return False
    if conanfile.settings.get_safe("os.subsystem") == "cygwin":
        return False
    host_compiler = conanfile.settings.get_safe("compiler")
    if host_compiler == "gcc":
        return True
    if host_compiler ==  "clang" and not conanfile.settings.get_safe("compiler.runtime"):
        return True
    return False
```

PS: From trying to map the current code to my snippet I can only conclude that the current code is very hard to read and understand (if my mapping is not correct it proves that the proposed code is perhaps overly complex).

memsharded: I agree with this, part this PR never got prioritized is because the logic reads too compact, and it also introduces new aspects that are not common practice in ConanCenter recipes, in which the majority just check ``Windows && gcc => MinGW``

memsharded (commented):
One of the unknowns of this PR is that it introduced non-established practice in ConanCenter, in which all recipes just check for Windows&gcc, but things like cygwin or clang are not even considering.

As Conan 2.0 is already 1 year old, I am closing this PR for 1.X releases, the current workaround in recipes is enough, and I am moving it as a new ticket for 2.X releases, in order to be able to further improve recipes in the future

conan/tools/gnu/mingw.py
mhthies: > and it also introduces new aspects that are not common practice in ConanCenter recipes, in which the majority just check Windows && gcc => MinGW

This might be a new aspect, but it's an important one. According to https://blog.conan.io/2022/10/13/Different-flavors-Clang-compiler-Windows.html the compiler.runtime setting is the official way to distinguish between MinGW clang and VS clang-cl. Some recipes in Conan Center already take this into account, but most do not. This is one of the main reasons why we are struggling a lot with introducing Conan at our company (we use MinGW clang for compatibility and performance reasons).

This PR could help a lot with easier integration and broader adoption of this Windows clang distinction through Conan Center.

memsharded: Hi @mhthies. There is an open ticket for this helper here: https://github.com/conan-io/conan/issues/15648, lets try to follow up there.

PR #12695, opened 2022-12-12
Add better error UX when spaces are found in required_conan_version

### LINKED ISSUE #12692: [bug] behaviour when there are spaces in required_conan_version
`required_conan_version = ">= 1.36.0"`

This currently fails in conan 2 due to the space. What should the behaviour be? If this is an error, we need better UX to inform the user.

AbrilRBS:
Changelog: (Fix): Display a better error when `required_conan_version` has spaces between the operator and the version.

This used to never allow the version check, even if the error was in the syntax, so now we just give a better error message

Closes #12692 

### DIFF
--- a/conans/model/version_range.py
+++ b/conans/model/version_range.py
@@ -1,5 +1,6 @@
 from collections import namedtuple
 
+from conans.errors import ConanException
 from conans.model.recipe_ref import Version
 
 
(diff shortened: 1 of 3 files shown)


memsharded (approved):
Fantastic!

PR #12696, opened 2022-12-12
Allow `required_conan_version` to be commented out

AbrilRBS:
Changelog: (Fix): Allows `required_conan_verison` to be commented out.

@jcar87 found this today while trying to migrate some recipe and was trying to ignore the min version


conans/test/integration/conanfile/required_conan_version_test.py
```diff
@@ -77,3 +77,17 @@ class Lib(ConanFile):
                             """)
         client.save({"conanfile.py": conanfile})
         client.run("export . --name=pkg --version=1.0")
+
+    def test_commented_out_required_conan_version(self):
+        """Used to not be able to comment out required_conan_version"""
+        client = TestClient()
+        conanfile = textwrap.dedent("""
+                            from conan import ConanFile
+                            # required_conan_version = ">=100.0"
```
memsharded: I am not sure how this failed before. Probably the ``import`` need to fail too? and ``assert_error=True``, but version not validated?

AbrilRBS: You were right, thanks for the headsup! I've now fixed it

PR #12704, opened 2022-12-14
Use YAML issue templates

AbrilRBS:
Changelog: omit
Docs: omit

This streamlines the process of filling out issues.
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

Although it's classified as a "Beta" feature, GitHub [is already recommending migrating to this where possible](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository):
> This is the legacy workflow to create an issue template. We recommend using the upgraded multiple issue template builder or issue forms to create issue templates. For more information, see "[About issue and pull request templates](https://docs.github.com/en/articles/about-issue-and-pull-request-templates)."

As for the contents, I've mostly kept the same text than in the previous version for each issue type, but added some extras here and there. 
Note that even though the checkmarks are set as `required: true` this does not seem to have any effect at the moment.

With this change, the issue templates would look like this now. Note how we have more changes to include extra info and show some placeholder examples:
| Pic | Type |
|-----|------|
|imagen| Bug |
|imagen| Feature request |


memsharded (approved):
This is looking good, and I think it will indeed help to file better issues. 
I am already excited and thinking if we could add some other things to improve and encourage users to report good issues, like a space to submit a URL with a full repro case. But we can definitely think about that later, I'd say move to the new format.

AbrilRBS:
Yes, my idea was to talk with someone about what changes would be good to have, will discuss later :)

czoido (approved):

PR #12715, opened 2022-12-15
Respect `force_host_context` when resolving dependencies

### LINKED ISSUE #12710: [question] `conf_info` and `test_requires`
### What is your question?

I have a system root package that does the following:

```py
    def package_info(self) -> None:
        self.conf_info.define("tools.build:sysroot", f"{self.package_path}")
```

I want to test that this is working, so use the following in the corresponding `test_package/conanfile.py`:

```py
    def build_requirements(self) -> None:
        assert self.tested_reference_str is not None
        self.test_requires(self.tested_reference_str)
```

However, the `tools.build:sysroot` is not applied to the configuration for my test package.

I cannot therefore test the package in the way it will be used.

mattyclarkson:
When a depedency is added with `test_requires` within `build_requirements` we set `force_host_context` on the requirement.

However, when we build the graph and resolve the recursive dependencies, we fail to respect that and set the profile/context to the target node context.

If the original target node has a `build` context, we start to resolve the nested node as `build` rather than as `host` context.

This patch adds a new branch to the context decision to respect the `force_host_context` on the requirement and set the context appropriately.

Closes #12710

Changelog: (Fix): Resolve `test_require` dependencies in the `host` context.

- [ ] ~~I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one.~~

### DIFF
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -437,6 +437,9 @@ class DepsGraphBuilder(object):
         if context_switch:
             profile = profile_build
             context = CONTEXT_BUILD
+        elif requirement.force_host_context:
+            profile = profile_host
+            context = CONTEXT_HOST
         else:
             profile = profile_host if current_node.context == CONTEXT_HOST else profile_build
             context = current_node.context


conans/client/graph/graph_builder.py
```diff
@@ -437,6 +437,9 @@ def _create_new_node(self, current_node, dep_graph, requirement, check_updates,
         if context_switch:
             profile = profile_build
             context = CONTEXT_BUILD
+        elif requirement.force_host_context:
```
memsharded, line 440: Hi @mattyclarkson 

Thanks very much for your contribution. However, I don't think this holds. A ``test_requires`` with ``force_host_context=True`` of a ``tool_requires``, is also in the "build" context, not in the "host" context. That is ``cmake`` recipe can have a ``test_requires = "gtest/1.2.3"``, and when ``cmake`` is used as ``tool_requires``, "gtest" package must end in the "build" context with the "build" profile.

memsharded: It seems it breaks some tests covering this case: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-12715/1/pipeline/41

PR #12722, opened 2022-12-16
Raise error when a generator is defined in generators and generate()

AbrilRBS:
This used to be a mistake, but no error was ever given to the user

Changelog: (Feature): Raise an error when a generator is both defined in generators attribute and instantiated in generate() method
Docs: Omit

### DIFF
--- a/conan/tools/__init__.py
+++ b/conan/tools/__init__.py
@@ -7,5 +7,5 @@ CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
 def _check_duplicated_generator(generator, conanfile):
     if generator.__class__.__name__ in conanfile.generators:
         raise ConanException(f"{generator.__class__.__name__} is declared in the generators "
-                             "attribute, but was instantiated in the generate() method toos. "
+                             "attribute, but was instantiated in the generate() method too. "
                              "It should only be present in one of them.")
--- a/conan/tools/__init__.py
+++ b/conan/tools/__init__.py
@@ -7,5 +7,5 @@ CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
 def _check_duplicated_generator(generator, conanfile):
     if generator.__class__.__name__ in conanfile.generators:
         raise ConanException(f"{generator.__class__.__name__} is declared in the generators "
-                             "attribute, but was also instantiated in the generate() method. "
+                             "attribute, but was instantiated in the generate() method toos. "
                              "It should only be present in one of them.")
--- a/conan/tools/__init__.py
+++ b/conan/tools/__init__.py
@@ -1,11 +1,4 @@
-from conan.errors import ConanException
-
 CONAN_TOOLCHAIN_ARGS_FILE = "conanbuild.conf"
 CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
 
 
-def _check_duplicated_generator(generator, conanfile):
-    if generator.__class__.__name__ in conanfile.generators:
-        raise ConanException(f"{generator.__class__.__name__} is declared in the generators "
-                             "attribute, but was instantiated in the generate() method too. "
-                             "It should only be present in one of them.")
--- a/conan/tools/__init__.py
+++ b/conan/tools/__init__.py
@@ -1,2 +1,11 @@
+from conans.errors import ConanException
+
 CONAN_TOOLCHAIN_ARGS_FILE = "conanbuild.conf"
 CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
+
+
+def _check_duplicated_generator(generator, conanfile):
+    if generator.__class__.__name__ in conanfile.generators:
+        raise ConanException(f"{generator.__class__.__name__} is declared in the generators "
+                             "attribute, but was also instantiated in the generate() method. "
+                             "It should only be present in one of them.")
--- a/conan/tools/__init__.py
+++ b/conan/tools/__init__.py
@@ -1,4 +1,4 @@
-from conans.errors import ConanException
+from conan.errors import ConanException
 
 CONAN_TOOLCHAIN_ARGS_FILE = "conanbuild.conf"
 CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
--- a/conan/tools/apple/xcodedeps.py
+++ b/conan/tools/apple/xcodedeps.py
@@ -106,7 +106,7 @@ class XcodeDeps(object):
 
     def __init__(self, conanfile):
         if self.__class__.__name__ in conanfile.generators:
-            raise ConanException(f"{self.__class__.__name__} is declared in the generators"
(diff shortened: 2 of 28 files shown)


memsharded (commented):
Beware, it happens that the ``VirtualBuildEnv`` and ``VirtualRunEnv`` are implicit, even if they are not declared, they are called after ``generate()``.

I think that is the thing breaking the tests.

AbrilRBS:
Yeah, I had some errors locally with integration tests that looked like remote issues, and this slipped thru the cracks when checking for real errors.

Most of the were something like this, let's discuss it tomorrow:
<details><summary>Example of one of the test errors I'm getting, click to expand</summary>           

```raise Exception(exc_message)
E           Exception: 
E           upload * --confirm -r default
E           Checking which revisions exist in the remote server
E           Checking which revisions exist in the remote server
E           Preparing artifacts to upload
E           Compressing recipe...
E           Compressing package...
E           Please log in to "default" to perform this action. Execute "conan user" command.
E           Remote 'default' username: Please enter a password for "admin" account: ERROR: b"module 'jwt' has no attribute 'encode'"
E           
E           Invalid server response, check remote URL and try again. [Remote: default]```
</details>

File: conan/tools/apple/xcodedeps.py [hunk 1]
```diff
@@ -105,6 +105,10 @@ class XcodeDeps(object):
         """)
 
     def __init__(self, conanfile):
+        if self.__class__.__name__ in conanfile.generators:
```
SSE4: it's duplicated multiple times, maybe it makes sense to extract code to its own function?

conan/tools/apple/xcodedeps.py
```diff
@@ -105,6 +105,10 @@ class XcodeDeps(object):
...
+            raise ConanException(f"{self.__class__.__name__} is declared in the generators "
```
SSE4: maybe it's time to start defining some more specific exception classes, instead of just using vague/generic `ConanException` for everything. IMO it should greatly improve debugging and troubleshooting.
maybe something like `GeneratorException` fits here.

conan/tools/apple/xcodedeps.py
memsharded: I thought that this would take just 2 lines, the ``if`` and then 1 line for the error message, not 3.
So @RubenRBS do whatever you think it is best, I am happy with either way.

conan/tools/apple/xcodedeps.py
memsharded: Only if at some place the new exception types are captured with a specific purpose and differently to other ``ConanException`` exceptions.

SSE4: yes, that's one of the reasons. right now, looking at PR itself, it's pretty hard to tell where exception is captured, if it's captured at all.

conan/tools/apple/xcodedeps.py
AbrilRBS: My only issue with this is the thing we mentioned on friday that it's not obvious where to place such helper function. I've done so in tool's init, but happy to change. Thanks @SSE4 for your suggestion :)

conan/tools/apple/xcodedeps.py
AbrilRBS: The point of this exception is not to be captured other than to be displayed to the user as an error, right? So it would make little sense to customize

memsharded: Yes, no need to be captured, no special handling rather than showing an error message, this is why a ``ConanException`` is good enough.

conans/client/generators/__init__.py
```diff
@@ -93,7 +93,12 @@ def write_generators(conanfile, hook_manager):
 
     if conanfile.generators:
         conanfile.output.info(f"Writing generators to {new_gen_folder}")
-    for generator_name in set(conanfile.generators):
+    # generators check that they are not present in the generators field,
+    # to avoid duplicates between the generators attribute and the generate() method
+    # They would raise an exception here if we don't invalidate the field while we call them
+    old_generators = set(conanfile.generators)
+    conanfile.generators = []
```
SSE4, line 100: maybe it needs a sort of context manager here?
in case exception happens on line 114, `conanfile.generators` will be never restored to an original state and it will remain empty forever

memsharded: A ``try-finally`` would be good, yes.

AbrilRBS: Nice catch! Thanks again for your reviews :)

AbrilRBS:
I've noticed that VirtualRunEnv and VirtualBuildEnv are instantiated in lots of places just to handle some envs, even though the generator is defined in the attribute.
Maybe it would be better for the check to run in the generate() method of at least those generators, and possible the rest too?

conan/tools/__init__.py
```diff
@@ -1,2 +1,11 @@
+from conans.errors import ConanException
```
memsharded:
```suggestion
from conan.errors import ConanException
```

conan/tools/__init__.py
```diff
@@ -1,2 +1,11 @@
...
+
 CONAN_TOOLCHAIN_ARGS_FILE = "conanbuild.conf"
 CONAN_TOOLCHAIN_ARGS_SECTION = "toolchain"
+
+
+def _check_duplicated_generator(generator, conanfile):
+    if generator.__class__.__name__ in conanfile.generators:
+        raise ConanException(f"{generator.__class__.__name__} is declared in the generators "
+                             "attribute, but was instantiated in the generate() method toos. "
```
memsharded:
```suggestion
                             "attribute, but was instantiated in the generate() method too. "
```

conans/test/integration/conanfile/generators_list_test.py
```diff
@@ -61,3 +61,25 @@ def init(self):
...
+
+            def generate(self):
+                tc = CMakeToolchain(self)
+                tc.generate()
+        """)
+
+        t = TestClient()
+        t.save({'conanfile.py': conanfile})
+        # This used to not throw any errors
+        t.run("install .", assert_error=True)
```
memsharded, line 85:
```suggestion
        t.run("install .", assert_error=True)
        assert "was instantiated in the generate() method too" in t.out
```

memsharded (approved):
Importing a private function from another module can be a linting or style error, we might need to rename it (later, not now), and if there are concerns of being accessible for users, it should be moved to ``internal`` namespace.

SSE4:
> Importing a private function from another module can be a linting or style error, we might need to rename it (later, not now), and if there are concerns of being accessible for users, it should be moved to `internal` namespace.

yes, I think `conan.tools` is a public namespace for recipe authors. if function isn't supposed to be used in recipes, it should go to some other place. also we usually use more specific nested namespaces, like `conan.tools.xxx`.
leading underscore implies private function, so it's okay.

AbrilRBS:
Thanks for your suggestion @SSE4. I ended up adding it to `tools/internal` as suggested

PR #12732, opened 2022-12-19
Skip subdirectories in fix_apple_shared_install_name

### LINKED ISSUE #12727: [bug] `fix_apple_shared_install_name()` fails if there is a subfolder in bin folder
### Environment details

* Compiler+version: apple-clang 14
* Conan version: 1.55.0
* Python version: 3.10.9

### Steps to reproduce

build a recipe creating a subfolder in bin folder & calling fix_apple_shared_install_name() in package()

### Logs

```
ERROR: xapian-core/1.4.19: Error in package() method, line 134
	CalledProcessErrorWithStderr: Command 'otool -hv /Users/spaceim/.conan/data/xapian-core/1.4.19/_/_/package/fecdf3b9eb7fa1cc93253eb65f840989ddd93cd3/bin/share' returned non-zero exit status 1.
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/otool-classic: can't map file: /Users/spaceim/.conan/data/xapian-core/1.4.19/_/_/package/fecdf3b9eb7fa1cc93253eb65f840989ddd93cd3/bin/share (Invalid argument)
```

jcar87:
Changelog: Bugfix: Fix issue in `fix_apple_shared_install_name` when libdirs or bindirs have subfolders.
Docs: Omit

Close: https://github.com/conan-io/conan/issues/12727

Co-authored-by: Rubén Rincón Blanco

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -171,7 +171,7 @@ def fix_apple_shared_install_name(conanfile):
         if binary_type not in ("DYLIB", "EXECUTE") or os.path.islink(file) or os.path.isdir(file):
             return False
         check_file = f"otool -hv {file}"
-        return True if binary_type in check_output_runner(check_file) else False
+        return binary_type in check_output_runner(check_file)
 
     def _darwin_collect_binaries(folder, binary_type):
         return [os.path.join(folder, f) for f in os.listdir(folder) if _darwin_is_binary(os.path.join(folder, f), binary_type)]
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -167,22 +167,14 @@ def fix_apple_shared_install_name(conanfile):
         install_name = check_output_runner(command).strip().split(":")[1].strip()
         return install_name
 
+    def _darwin_is_binary(file, binary_type):
+        if binary_type not in ("DYLIB", "EXECUTE") or os.path.islink(file) or os.path.isdir(file):
+            return False
+        check_file = f"otool -hv {file}"
+        return True if binary_type in check_output_runner(check_file) else False
+
     def _darwin_collect_binaries(folder, binary_type):
-        ret = []
-        if binary_type not in ("DYLIB", "EXECUTE"):
-            return ret
-
-        for root, _, files in os.walk(folder):
-            for f in files:
-                full_path = os.path.join(root, f)
-                if os.path.islink(full_path):
-                    continue
-
-                check_lib = f"otool -hv {full_path}"
-                if binary_type in check_output_runner(check_lib):
-                    ret.append(full_path)
-
-        return ret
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
👍

jcar87 (commented):
On second thought, chances are this function shouldn't recursively find files  as we would expect `libdirs` and `bindirs` to list all directories in the `package_info()` instead, in case there's actually relevant libraries in those subfolders

conan/tools/apple/apple.py
```diff
@@ -167,9 +167,14 @@ def _get_install_name(path_to_dylib):
...
         return install_name
 
-    def _darwin_collect_dylibs(lib_folder):
-        return [os.path.join(lib_folder, f) for f in os.listdir(lib_folder) if f.endswith(".dylib")
-                and not os.path.islink(os.path.join(lib_folder, f))]
+    def _darwin_is_binary(file, binary_type):
+        if binary_type not in ("DYLIB", "EXECUTE") or os.path.islink(file) or os.path.isdir(file):
+            return False
+        check_file = f"otool -hv {file}"
+        return True if binary_type in check_output_runner(check_file) else False
```
franramirez688:
```suggestion
        return binary_type in check_output_runner(check_file)
```

memsharded:
> On second thought, chances are this function shouldn't recursively find files as we would expect libdirs and bindirs to list all directories in the package_info() instead, in case there's actually relevant libraries in those subfolders

It is also important to consider that consumers via ``VirtualXXXEnv`` will add the ``bindirs`` path to the PATH env-var in order to find executables and DYLD_LIBRARY_PATH for shared libraries. If libraries are in nested subfolders, they will not be found, so it seems those folders should be listed in ``bindirs/libdirs``?

franramirez688 (approved):

conan/tools/apple/apple.py
```diff
@@ -167,9 +167,14 @@ def _get_install_name(path_to_dylib):
...
         return install_name
 
-    def _darwin_collect_dylibs(lib_folder):
-        return [os.path.join(lib_folder, f) for f in os.listdir(lib_folder) if f.endswith(".dylib")
-                and not os.path.islink(os.path.join(lib_folder, f))]
+    def _darwin_is_binary(file, binary_type):
+        if binary_type not in ("DYLIB", "EXECUTE") or os.path.islink(file) or os.path.isdir(file):
+            return False
+        check_file = f"otool -hv {file}"
+        return binary_type in check_output_runner(check_file)
```
SSE4, line 174: I suppose `check_output_runner` may raise, therefore any error from `otool` invocation (like, it failed to parse a particular file, or just missing), will result in failure to create a package.
should it have `try/catch` block here to return `False` in case of any errors?

czoido (approved):

memsharded (approved):

PR #12742, opened 2022-12-20
Check for comment only in line start for required_conan_version

AbrilRBS:
Changelog: (Bugfix) Allow comments for `required_conan_version`
Docs: Omit

This ensures that both `# required_conan_version = ">=xxx"` and `required_conan_version = ">=xxx" # This is a comment` work when a bad import didn't allow us to load the file normally.

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -364,9 +364,9 @@ def _get_required_conan_version_without_loading(conan_file_path):
     txt_version = None
 
     try:
-        found = re.search(r"(.*)required_conan_version\s*=\s*(.*)", contents)
+        found = re.search(r"(.*)required_conan_version\s*=\s*[\"'](.*)[\"']", contents)
         if found and "#" not in found.group(1):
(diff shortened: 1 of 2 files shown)


memsharded (approved):

prince-chrismc:
I was curious what this was doing but it's not clear...
Is there test code to show what can or could not be in front? I'd love to see examples for reference 🙏

memsharded:
It is a bit of an edge case:

- If you have a wrong import in the recipe
- And you comment out the ``required_conan_version``
- Conan was still raising an error of an invalid Conan version because import was raising, even if the statement is commented out, because the regex was not prepared to filter out commented statement.

AbrilRBS:
I found another bug with `required_conan_version = ">=xxx" # This is a comment` was trying to compare against the `>=xxx # This is a comment` version.
As suggested by @prince-chrismc and @czoido I've also added more tests and context to make sure everything looks good :)

memsharded (approved):

czoido (approved):

PR #12744, opened 2022-12-20
[develop2] CMakeDeps fix when a transitive dependency is skipped and there are components.requires

memsharded:


jcar87 (approved):
LGTM, this fixes the issue we saw earlier!

PR #12746, opened 2022-12-20
[develop2] Improve a bit the output msgs

memsharded:
Changelog: Feature: Improved global Conan output messages (create, install, export, etc.)
Docs: omit


conans/client/generators/__init__.py
```diff
@@ -31,10 +31,11 @@ def _get_generator_class(generator_name):
     try:
         generator_class = _generators[generator_name]
         # This is identical to import ... form ... in terms of cacheing
-        return getattr(importlib.import_module(generator_class), generator_name)
     except KeyError as e:
         raise ConanException(f"Invalid generator '{generator_name}'. "
                              f"Available types: {', '.join(_generators)}") from e
+    try:
+        return getattr(importlib.import_module(generator_class), generator_name)
```
AbrilRBS, line 38: 👍

conan/tools/cmake/presets.py
```diff
@@ -200,6 +208,7 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables,
 
     preset_content = json.dumps(data, indent=4)
     save(preset_path, preset_content)
+    ConanOutput(str(conanfile)).info("CMakeToolchain generated: CMakePresets.json")
```
AbrilRBS, line 211: I'm guessing these are `ConanFileInterface`s and that's why they can't just call `conanfile.output...`, right?

AbrilRBS (approved):
This looks so much nicer now!

czoido (approved):

conans/client/generators/__init__.py
memsharded: yes, my IDE linter was warning against possible undefinition of variables, which might happen when a multi-line ``try`` clause can raise different exceptions in different lines. I try to remove these linter warnings when possible, as I have learned that often they can be a bug waiting for a corner case to become an issue.

conan/tools/cmake/presets.py
memsharded: No, good catch, I think this might be a full conanfile, this might be replaced by ``conanfile.output.info``

franramirez688 (approved):

Issue #5429, opened 2019-07-02
Possibly undocumented special semantic for a "test" sub-directory in a package


jegabe:
When using conan 1.16.1 under Windows, the following "magic" happens:

When having a sub-directory named "test" with another conanfile.py and CMakeLists.txt contained (along with the usual conanfile.py and CMakeLists.txt in the root folder of my working copy), conan tries to build something inside the "test" folder as well. When the folder is renamed to "tests" (adding an 's'), this behavior changes.

I'm not sure if it is a leftover of the "test_package" development, wanted (probably undocumented) behavior or a bug.

Here is the folder structure of my project

root folder
 |
+ - conanfile.py
+ - CMakeLists.txt
+ - test
      |
     + - conanfile.py
     + - CMakeLists.txt

memsharded:
Hi @jegabe 

You are totally right, the behavior changes. This is some legacy, as originally the ``test_package`` folder was called ``test`` folder. We introduced the new name and un-documented the previous one, but didn't remove the behavior to not unnecessarily break users.

I think it deserves a note in the docs by now, and that it should be completely removed in conan 2.0

Thanks for reporting!

memsharded:
https://github.com/conan-io/conan/pull/12752 is dropping the ``test`` name as a Conan ``test_package`` folder, for 2.0

czoido:
Closed by https://github.com/conan-io/conan/pull/12752 will be in beta7

PR #12752, opened 2022-12-21
[develop2] drop legacy test folder name for test_package

memsharded:
Close https://github.com/conan-io/conan/issues/5429

### DIFF
--- a/conan/cli/commands/create.py
+++ b/conan/cli/commands/create.py
@@ -153,20 +153,16 @@ def test_package(conan_api, deps_graph, test_conanfile_path, tested_python_requi
 
 
 def _get_test_conanfile_path(tf, conanfile_path):
-    """Searches in the declared test_folder or in the standard locations"""
+    """Searches in the declared test_folder or in the standard "test_package"
+    """
 
     if tf is False:
         # Look up for testing conanfile can be disabled if tf (test folder) is False
         return None
 
-    test_folders = [tf] if tf else ["test_package", "test"]
     base_folder = os.path.dirname(conanfile_path)
-    for test_folder_name in test_folders:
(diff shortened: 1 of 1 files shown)


AbrilRBS (approved):

czoido (approved):

czoido:
I can't see any tests in develop2 testing this, maybe we should add one

memsharded:
> I can't see any tests in develop2 testing this, maybe we should add one

Because nowhere else in our codebase we were using ``test`` as a ``test_package`` folder, this is many years old legacy.
Is it possible to add some test testing that local folders ``test``, ``tests``, ``testing``, etc are not taken as a ``test_package``, but seems a bit overkill to test this no-feature?

czoido:
> > I can't see any tests in develop2 testing this, maybe we should add one
> 
> Because nowhere else in our codebase we were using `test` as a `test_package` folder, this is many years old legacy. Is it possible to add some test testing that local folders `test`, `tests`, `testing`, etc are not taken as a `test_package`, but seems a bit overkill to test this no-feature?

I meant that we are not testing anywhere (or I could not find at least) the error message that appears if you do a `conan create . -tf=nonexistent_or_without_conanfile_folder` I thought that it would be tested, but it's not so related with this PR, so no problem

memsharded:
> I meant that we are not testing anywhere (or I could not find at least) the error message that appears if you do a conan create . -tf=nonexistent_or_without_conanfile_folder I thought that it would be tested, but it's not so related with this PR, so no problem

Ok, that makes sense. It is working fine:
```
(conan2_36) λ conan create . -tf=kk
ERROR: test folder 'kk' not available, or it doesn't have a conanfile.py
```
but it could be covered by one test too, I'll try to add it later.

Issue #5634, opened 2019-08-15
Add warning against non-string option values


solvingj:
Here are some of the options for the bincrafters wxwidgets package.  

```               
               "zlib": [None, "system", "zlib"],
               "png": [None, "system", "libpng"],
               "jpeg": [None, "system", "libjpeg", "libjpeg-turbo", "mozjpeg"],
               "tiff": [None, "system", "libtiff"],
               "expat": [None, "system", "expat"],
```
Here are the defaults:
```
               "zlib": "zlib",
               "png": "libpng",
               "jpeg": "libjpeg",
               "tiff": "libtiff",
               "expat": "expat",
```
In summary, the use of the literal python `None` as a non-default option here actually renders the `None` case/option/branch unreachable.  The python literal `None` can only ever work as a default option, making it special, awkward, and this ticket suggests adding a warning against it. 

The fundamental problem is that the command-line interface of Conan means users can only ever pass the string `None`.  Of course, strings can be converted to other types behind the scenes, but such implicit magic is generally hard for users to predict without reading documentation, and both unexpected and undesirable in this kind of case. The unfortunate consequence in the case of the wxwidgets recipe was that the following function to act on these options has been subtly and silently broken since it's inception: 
```
        def option_value(option):
            return 'OFF' if option is None else 'sys'
```
In this case, option always had a value (even when users pass the string `None`), so the condition was never `True` and this function always returned `sys`. 

There are at least 5 different ways Conan team could choose to approach this problem.  This includes: 
* supporting an empty-value syntax for None:  `-o mypackage:myoption=`
* supporting a "special" syntax for None:  `-o mypackage:myoption=_None_`
* adding automatic coercion behavior from `None` string to literal `None` 
* adding an environment variables to let users opt into one of the above behaviors
* doing nothing at all, because "authors should just know this is how python works"
* adding a warning, and avoiding the pattern in future conan-center packages

I've thought about them for a bit, and am choosing to suggest inserting a lexer warning and potentially mentioning as a best practice in the documentation: "Always make all possible option values strings, because that's the only thing that users can pass at the CLI". 

One might point out that the literal `None` works just fine as a default value, and this strategy is used in a number of packages.  The problem is that `None` is just different than strings, and simply by being different it leads to unexpected behavior in corner cases. 

[...]

PR #12761, opened 2022-12-21
fix comparison of option None

memsharded:
Close https://github.com/conan-io/conan/issues/5634

### DIFF
--- a/conans/test/unittests/model/options_test.py
+++ b/conans/test/unittests/model/options_test.py
@@ -206,6 +206,21 @@ class TestOptionsNone:
         text = self.sut.dumps()
         assert text == ""
 
+    def test_boolean_none(self):
+        options = Options({"static": [None, "None", 1, 2]})
+        assert options.static != 1
+        assert not (options.static == 1)
+        assert options.static != "None"
+        assert not (options.static == "None")
+        assert options.static == None
+        assert not (options.static != None)
+
+        options.static = "None"
+        assert options.static == "None"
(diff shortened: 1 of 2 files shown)


conans/test/unittests/model/options_test.py
```diff
@@ -206,6 +206,21 @@ def test_dumps(self):
         text = self.sut.dumps()
         assert text == ""
 
+    def test_boolean_none(self):
+        options = Options({"static": [None, "None", 1, 2]})
```
AbrilRBS, line 210: Looks good and your comment in the linked issue nicely explains how this works now.

Would maybe adding some asserts like `assert not options.static` be useful?

This might just be me being clueless yet as to how things work (even after reading some of the code) but I had to double check that `assert not options.static` would work as expected locally, and that for `state = None` the assert would pass (which it did and I expected it it), and that `static = "None"` would not pass the assert, which it did pass so I was not expecting that!

AbrilRBS (approved):
Thanks for the easy review :)

I do have one question regarding the behaviour of checks like `if not options.shared` when `options.shared = "None"` or similar

conans/test/unittests/model/options_test.py
memsharded: The case ``assert not options.static`` is already sufficiently covered in the test above, I was only covering here the pending bug of confusing Python ``None`` with a string ``"None"``, which was really more confusing.

czoido (approved):

PR #12762, opened 2022-12-21
Add cmake install component support #10359

### LINKED ISSUE #10359: [feature] Allow specifying an install component for the CMake helper
One is capable of passing `--component` to CMake's `--install` command, however the same functionality is missing from the CMake helper's `install()` method: https://docs.conan.io/en/latest/reference/build_helpers/cmake.html#install

Since Conan does not package CMake configs generated by recipes, in many cases the recipe can just define the runtime component and there won't be a need to `rmdir` additional files.

I suggest adding a `component=None` argument to this method and restructuring it in terms of the cmake generated `cmake_install.cmake` script. The code would look something like this:

```py
def install(... component=None):
    if component is not None:
        if isinstance(component, str):
            component_args = [f"-DCOMPONENT={component}"]
        else:
            component_args = [f"-DCOMPONENT={c}" for c in component]
    else:
        component_args = [""]
    for comp in component_args:
        self.run(f"cmake -DBUILD_TYPE={build_type} {comp} -P cmake_install.cmake", cwd=bindir)
```

or maybe a new method like `install_component()` with a required positional argument, if such a break for v2 is not desirable.  
`find -name conanfile.py -exec grep -nHEe '\.install\([^)]' {} \;` shows no usage for this method being used with any arguments in CCI though.

neuj21:
Changelog: Feature: Add the ability to provide a `--component` argument with the `cmake.install` function. 
Docs: https://github.com/conan-io/docs/pull/2893

Related issue #10359

### DIFF
--- a/conans/test/unittests/tools/cmake/test_cmake_install.py
+++ b/conans/test/unittests/tools/cmake/test_cmake_install.py
@@ -42,8 +42,7 @@ def test_run_install_component(generator):
     cmake = CMake(conanfile)
     cmake.install(component="foo")
 
-    search_pattern = "--component foo" if platform.system() == "Windows" else "'--component' 'foo'"
-    assert search_pattern in conanfile.command
+    assert "--component foo" in conanfile.command
 
 
 @pytest.mark.parametrize("generator", [
--- a/conans/test/unittests/tools/cmake/test_cmake_install.py
+++ b/conans/test/unittests/tools/cmake/test_cmake_install.py
@@ -1,0 +1,77 @@
+import platform
+import pytest
+
+from conan.tools.cmake import CMake
+from conan.tools.cmake.presets import write_cmake_presets
+from conans.client.conf import get_default_settings_yml
+from conans.model.conf import Conf
+from conans.model.settings import Settings
+from conans.test.utils.mocks import ConanFileMock
+from conans.test.utils.test_files import temp_folder
+
+
+@pytest.mark.parametrize("generator", [
+    ("NMake Makefiles"),
+    ("Ninja Makefiles"),
+    ("Ninja Multi-Config"),
+    ("Unix Makefiles"),
+    ("Visual Studio 14 2015"),
+    ("Xcode"),
+])
+def test_run_install_component(generator):
+    """
+    Testing that the proper component is installed.
+    Issue related: https://github.com/conan-io/conan/issues/10359
+    """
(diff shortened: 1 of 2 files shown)


conans/test/unittests/tools/cmake/test_cmake_install.py
```diff
@@ -0,0 +1,76 @@
...
+from conan.tools.cmake import CMake
+from conan.tools.cmake.presets import write_cmake_presets
+from conans.client.conf import get_default_settings_yml
+from conans.model.conf import Conf
+from conans.model.settings import Settings
+from conans.test.utils.mocks import ConanFileMock
+from conans.test.utils.test_files import temp_folder
+
+
+@pytest.mark.parametrize("generator", [
```
memsharded: I appreciate the thorough testing, but it is really necessary to test with all CMake generators?
The logic for ``--component`` doesn't seem tied to the generator at all, so maybe this can be simplified a bit.

conans/test/unittests/tools/cmake/test_cmake_install.py
```diff
@@ -0,0 +1,76 @@
...
+    conanfile.folders.generators = "."
+    conanfile.folders.set_base_generators(temp_folder())
+    conanfile.settings = settings
+    conanfile.folders.set_base_package(temp_folder())
+
+    write_cmake_presets(conanfile, "toolchain", generator, {})
+    cmake = CMake(conanfile)
+    cmake.install()
+
+    assert "--component" not in conanfile.command
```
memsharded: I think the "negative" case is sufficiently covered by the rest of the test suite. Hundreds of tests would fail if this was happening.

memsharded (commented):
Thanks very much for the contribution, looking good!

conans/test/unittests/tools/cmake/test_cmake_install.py
neuj21: Probably not needed, I originally put it in because I was unsure if the implementation would need to change based on the generator but later determined it would not. I will simplify when I can.

conans/test/unittests/tools/cmake/test_cmake_install.py
neuj21: Sounds good. Will remove in upcoming commit.

Thank you!

memsharded (approved):

franramirez688 (approved):

PR #12765, opened 2022-12-22
[develop2] update new templates

memsharded:
From https://github.com/conan-io/docs/pull/2853#discussion_r1055331184


conan/internal/api/new/autoools_exe.py
```diff
@@ -66,7 +66,7 @@ def layout(self):
         basic_layout(self)
 
     def test(self):
-        if not cross_building(self):
+        if not can_run(self):
```
czoido:
```suggestion
        if can_run(self):
```

AbrilRBS (approved):
👍 other than what Carlos found

Issue #3584, opened 2018-09-20
[Commands] conan remote update


lasote:
Review command for 2.0.
Especially the `conan remote update` drives me crazy when I just need to update the URL or only the name or only the SSL and I need to paste all again. Maybe `conan remote update name [-name xxx] [-url xxx] [-verify_ssl xxx]?

memsharded:
Closed by https://github.com/conan-io/conan/pull/12770

PR #12770, opened 2022-12-22
[develop2] minor improvements remote-update

memsharded:
- Remove ``conan remote move``, as ``conan remote update`` already implements it
- Remove unnecessary check of argument as int
 
Fix https://github.com/conan-io/conan/issues/3584

### DIFF
--- a/conan/cli/commands/remote.py
+++ b/conan/cli/commands/remote.py
@@ -66,28 +66,16 @@ def remote_add(conan_api, parser, subparser, *args):
     """
     subparser.add_argument("name", help="Name of the remote to add")
     subparser.add_argument("url", help="Url of the remote")
-    subparser.add_argument("--secure", dest="secure", action='store_true',
-                           help="Don't allow insecure server connections when using SSL")
     subparser.add_argument("--insecure", dest="secure", action='store_false',
                            help="Allow insecure server connections when using SSL")
     subparser.add_argument("--index", action=OnceArgument, type=int,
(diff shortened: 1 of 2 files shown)


conan/cli/commands/remote.py
```diff
@@ -116,14 +104,16 @@ def remote_update(conan_api, parser, subparser, *args):
                            help="Insert the remote at a specific position in the remote list")
     subparser.set_defaults(secure=None)
     args = parser.parse_args(*args)
-    if not (args.url is not None or args.secure is not None):
+    if args.url is None and args.secure is None and args.index is None:
```
AbrilRBS, line 107: Quick quiestion. You deleted --secure from the add command, but then what's different in the update command to still needed it too?

AbrilRBS (approved):
Quick noobie question

conan/cli/commands/remote.py
memsharded: You might need to update a "insecure" remote and make it "secure", and this update must be explicit, it is not good that it will update it to make it secure by default (because you are changing the URL or the index). While the default for ``conan remote add`` is being secure, always, unless you explicitly opt-out, and thus ``--secure`` in ``conan remote add`` is completely unnecessary.

AbrilRBS: That makes lots of sense, thank you!

czoido (approved):

PR #12777, opened 2022-12-23
[develop2] remove unnecessary xz checks

memsharded:
The check of other possible compression formats is dead, we don't plan to change the format in 2.0, and it is not possible to do it in 2.X without breaking and opt-in, so the check doesn't provide any migration path to that potential new feature anyway.

### DIFF
--- a/conans/test/unittests/util/xz_test.py
+++ b/conans/test/unittests/util/xz_test.py
@@ -1,86 +1,21 @@
 import os
 import tarfile
-from unittest import TestCase
 
 from conan.tools.files import unzip
-from conans.model.package_ref import PkgReference
-from conans.model.recipe_ref import RecipeReference
-from conans.test.assets.genconanfile import GenConanfile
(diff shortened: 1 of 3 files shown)


conans/test/unittests/util/xz_test.py
```diff
@@ -1,86 +1,21 @@
...
-from conans.test.assets.genconanfile import GenConanfile
 from conans.test.utils.mocks import ConanFileMock
 from conans.test.utils.test_files import temp_folder
-from conans.test.utils.tools import NO_SETTINGS_PACKAGE_ID, TestClient, TestServer
-from conans.util.files import load, save_files, save
+from conans.util.files import load, save
 
 
-class XZTest(TestCase):
+def test_unzip_can_xz():
```
AbrilRBS, line 10: Nice catch keeping this test

AbrilRBS (approved):

PR #12780, opened 2022-12-23
Make ConanOutput write methods chainable

AbrilRBS:
Changelog: Feature: Allow chaining of `self.output` write methods
Docs: omit

Allows the ConanOutput write methods to be chained, which simplifies some instances where it's used a lot, by removing the need to save it to a variable or initialize the object for every line

### DIFF
--- a/conan/api/output.py
+++ b/conan/api/output.py
@@ -82,11 +82,11 @@ class ConanOutput:
         return hasattr(self.stream, "isatty") and self.stream.isatty()
 
     def writeln(self, data, fg=None, bg=None):
-        self.write(data, fg, bg, newline=True)
+        return self.write(data, fg, bg, newline=True)
 
     def write(self, data, fg=None, bg=None, newline=False):
         if conan_output_level > LEVEL_NOTICE:
-            return
+            return self
         if self._color and (fg or bg):
             data = "%s%s%s%s" % (fg or '', bg or '', data, Style.RESET_ALL)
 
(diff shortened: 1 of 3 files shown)


conan/api/output.py
```diff
@@ -82,24 +82,25 @@ def is_terminal(self):
...
         self.stream.write(data)
         self.stream.flush()
+        return self
 
     def rewrite_line(self, line):
         tmp_color = self._color
         self._color = False
         TOTAL_SIZE = 70
-        LIMIT_SIZE = 32  # Hard coded instead of TOTAL_SIZE/2-3 that fails in Py3 float division
+        LIMIT_SIZE = TOTAL_SIZE // 2 - 3
```
AbrilRBS, line 103: The old comment complained about Py3 division being float by default. Using the built-in int division fixes the issue (`//`)

memsharded (approved):

conans/test/unittests/client/conan_output_test.py
```diff
@@ -51,3 +51,13 @@ def test_output_forced_but_conan_logger():
...
+
+
+def test_output_chainable():
+    try:
+        ConanOutput(scope="My package")\
+            .title("My title")\
+            .highlight("Worked")\
+            .info("But there was more that needed to be said")
+    except Exception as e:
+        assert False, f"Chained ConanOutput write methods raised an exception {e}"
```
franramirez688, line 63: In my opinion, it could be easier to check the output itself from any existing test. I'd recommend changing some existing integration ones that use the `self.output` attribute several times.

Let me suggest this one for instance:

*conans/test/integration/py_requires/python_requires_test.py:test_multiple_requires_error*

```python
                def build(self):
                    self.output.info("PKG1 N: %s" % self.python_requires["pkg1"].conanfile.name)\
                               .info("PKG1 V: %s" % self.python_requires["pkg1"].conanfile.version)\
                               .info("PKG1 U: %s" % self.python_requires["pkg1"].conanfile.user)\
                               .info("PKG1 C: %s" % self.python_requires["pkg1"].conanfile.channel)\
                               .info("PKG1 : %s" % self.python_requires["pkg1"].module.myvar)\
                               .info("PKG2 : %s" % self.python_requires["pkg2"].module.myvar)\
                               .info("PKG1F : %s" % self.python_requires["pkg1"].module.myfunct())\
                               .info("PKG2F : %s" % self.python_requires["pkg2"].module.myfunct())
```
What do you think? I believe that changing that one should be good enough.

franramirez688 (approved):

conans/test/unittests/client/conan_output_test.py
AbrilRBS: The more tests the merrier, yes! I'll add the one you suggested, thanks :)

AbrilRBS: Done, thanks for your review!

PR #12793, opened 2022-12-27
Feature/develop2 conf deprecations

memsharded:


conans/test/unittests/model/test_conf.py
```diff
@@ -217,7 +217,6 @@ def test_conf_get_check_type_and_default():
     c = ConfDefinition()
     c.loads(text)
     assert c.get("user.company.cpu:jobs", check_type=int) == 5
-    assert c.get("user.company.cpu:jobs", check_type=str) == "5"  # smart conversion
```
memsharded, line 220: I am proposing removing the smart conversions, wouldn't they be potentially hiding errors? why if we have explicit typing, why do we want to bypass it? Lets talk about it.

franramirez688: In my opinion, I don't think that we're hiding any errors at all, indeed, I think we're helping users to avoid annoying quotes or double quotes at any time.

Users were used to putting:

```
user.company.cpu:jobs=5
tools.system.package_manager:tool=pacman
tools.apple.xcodebuild:verbosity=quiet
```
Why are not we failing because of that? The lack of `'` or `"` should be an error as well, shouldn't it?
But I think Conan can be smart enough in some cases and perform those conversions.

memsharded: Yes, I agree with that. I realized it when I had to change some tests to add quotes. I am probably dropping this PR, or reducing it to just the ``[]`` access. Checking it.

PR #12795, opened 2022-12-28
do not allow mixing legacy helpers with new generators

memsharded:
Changelog: Omit
Docs: Omit

There are too many user errors because of this, making it a hard error.

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -17,6 +17,7 @@ class CMakeDeps(object):
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
+        self._conanfile.must_use_new_helpers = True  # TODO: Remove 2.0
         self.arch = self._conanfile.settings.get_safe("arch")
         self.configuration = str(self._conanfile.settings.build_type)
 
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -118,6 +118,7 @@ class CMakeToolchain(object):
 
     def __init__(self, conanfile, generator=None):
         self._conanfile = conanfile
+        self._conanfile.must_use_new_helpers = True  # TODO: Remove 2.0
         self.generator = self._get_generator(generator)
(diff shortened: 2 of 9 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -17,6 +17,7 @@ class CMakeDeps(object):
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
+        self._conanfile.must_use_new_helpers = True  # TODO: Remove 2.0
```
AbrilRBS, line 20: If this is a private internal variable, shouldn't it be prefixed by an underscore?

memsharded: It is not internal, it is a conan one. In the docs the rules are:
- All the public ``ConanFile`` namespace belongs to Conan. We can declare as many attrs and methods in Conanfiles for Conan usage
- Private ``_xxxx`` are for users, except ``._conan_xxx``

So this could be named as-is, or ``_conan_must_use_new_helpers``, but that would violate accessing it from the outside of ConanFile

AbrilRBS: Ohh thanks for the explanation, I had come to the conclusion that it was not like that (Now people using `_get_gnu_triplet` makes more sense too)

Can you point me to the docs that explain this? I was clueless that it was written down somewhere, and I can't find it by searching in the obvious places :)

AbrilRBS (approved):

conan/tools/cmake/cmakedeps/cmakedeps.py
memsharded: https://docs.conan.io/en/latest/reference/conanfile.html

AbrilRBS: Might be worth to add it somewhere else too, I expected that file to just be a recopilation of the inner sections and not provide new information

memsharded: I agree I don't love it in the current location, but I am not sure in which other place it would be better. Suggestions welcome.

PR #12799, opened 2022-12-28
Add Fish output for VirtualEnv

AbrilRBS:
Changelog: (Feature): Add `tools.env.virtualenv:fish` conf to enable creation of Fish-compatible launchers in VirtualEnv
Docs: https://github.com/conan-io/docs/pull/2880

I've been bitten a few times now where I wanted to execute `conanrun.sh` but had to first switch from my fish terminal back to bash, as their syntax is not complatible.

This is a 1-to-1 translation of the sh generation targeted at fish (So specially the is_defined variable could be rewritten more fish-y)

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -18,9 +18,9 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
     if not env_filenames:
         return cmd
     filenames = [env_filenames] if not isinstance(env_filenames, list) else env_filenames
-    bats, shs, ps1s = [], [], []
+    bats, shs, ps1s, fishes = [], [], [], []
 
-    accept = accepted_extensions or ("ps1", "bat", "sh")
+    accept = accepted_extensions or ("ps1", "bat", "fish", "sh")
     # TODO: This implemantation is dirty, improve it
     for f in filenames:
         f = f if os.path.isabs(f) else os.path.join(env_folder, f)
@@ -28,6 +28,9 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
             if os.path.isfile(f) and "sh" in accept:
                 f = subsystem_path(subsystem, f)
                 shs.append(f)
+        elif f.lower().endswith(".fish"):
+            if os.path.isfile(f) and "fish" in accept:
+                fishes.append(f)
         elif f.lower().endswith(".bat"):
             if os.path.isfile(f) and "bat" in accept:
                 bats.append(f)
@@ -35,9 +38,10 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
             if os.path.isfile(f):
                 ps1s.append(f)
         else:  # Simple name like "conanrunenv"
-            path_bat = "{}.bat".format(f)
-            path_sh = "{}.sh".format(f)
-            path_ps1 = "{}.ps1".format(f)
+            path_bat = f"{f}.bat"
+            path_sh = f"{f}.sh"
+            path_ps1 = f"{f}.ps1"
+            path_fish = f"{f}.fish"
             if os.path.isfile(path_bat) and "bat" in accept:
                 bats.append(path_bat)
             if os.path.isfile(path_ps1) and "ps1" in accept:
@@ -45,10 +49,12 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
             if os.path.isfile(path_sh) and "sh" in accept:
                 path_sh = subsystem_path(subsystem, path_sh)
                 shs.append(path_sh)
+            if os.path.isfile(path_fish) and "fish" in accept:
+                fishes.append(path_fish)
 
-    if bool(bats) + bool(shs) + bool(ps1s) > 1:
+    if bool(bats) + bool(shs) + bool(ps1s) + bool(fishes) > 1:
         raise ConanException("Cannot wrap command with different envs,"
-                             " {} - {} - {}".format(bats, shs, ps1s))
+                             " {} - {} - {} - {}".format(bats, shs, ps1s, fishes))
 
     if bats:
         launchers = " && ".join('"{}"'.format(b) for b in bats)
@@ -60,6 +66,9 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
         # TODO: at the moment it only works with path without spaces
         launchers = " ; ".join('"&\'{}\'"'.format(f) for f in ps1s)
         return 'powershell.exe {} ; cmd /c {}'.format(launchers, cmd)
+    elif fishes:
+        launchers = " && ".join('. "{}"'.format(f) for f in shs)
+        return "{} && {}".format(launchers, cmd)
     else:
         return cmd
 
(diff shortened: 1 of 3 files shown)


conans/model/conf.py
```diff
@@ -85,6 +85,7 @@
     "tools.apple:enable_arc": "(boolean) Enable/Disable ARC Apple Clang flags",
     "tools.apple:enable_visibility": "(boolean) Enable/Disable Visibility Apple Clang flags",
     "tools.env.virtualenv:powershell": "If it is set to True it will generate powershell launchers if os=Windows",
+    "tools.env.virtualenv:fish": "If True, generate fish-compatible launchers",
```
AbrilRBS, line 88: Are the conf definitions in the docs generated automatically from this file? Or should I add this line in the sister docs PR?

conan/tools/env/environment.py
```diff
@@ -486,11 +495,43 @@ def save_sh(self, file_location, generate_deactivate=True):
         content = "\n".join(result)
         save(file_location, content)
 
+    def save_fish(self, file_location, generate_deactivate=True):
```
AbrilRBS, line 498: Note sure how to add tests for this. Locally I have one that expects to have fish available, but that's not true ion most cases. I know there's a config for these kinds of stuff, but I'm guessing if adding a new entry there would be worth it?

AbrilRBS (commented):
Some general questions for whoever reviews this

conans/model/conf.py
memsharded: No, the docs for the confs are not automatically generated from this yet. We are already generating some docs from docstrings, not sure how to do this, but I guess it should be doable. At the moment, yes, manual addition is necessary.

conan/tools/env/environment.py
memsharded: It is difficult to have fully functional tests for this, we would need to install fish in the CI servers. I guess we could start with some basic integration tests, and checking the generated file important parts could be enough.

conans/client/generators/__init__.py
```diff
@@ -130,12 +130,15 @@ def deactivates(filenames):
...
         ps1s = []
+        fishes = []
         for env_script in env_scripts:
             path = os.path.join(conanfile.generators_folder, env_script)
             if env_script.endswith(".bat"):
                 bats.append(path)
             elif env_script.endswith(".sh"):
                 shs.append(subsystem_path(subsystem, path))
+            elif env_script.endswith(".fish"):
+                fishes.append(subsystem_path(subsystem, path))
```
memsharded, line 141: fish exists/work in msys2 and other subsystems?

conan/tools/env/environment.py
```diff
@@ -18,37 +18,43 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
     if not env_filenames:
         return cmd
     filenames = [env_filenames] if not isinstance(env_filenames, list) else env_filenames
-    bats, shs, ps1s = [], [], []
+    bats, shs, ps1s, fishes = [], [], [], []
 
-    accept = accepted_extensions or ("ps1", "bat", "sh")
+    accept = accepted_extensions or ("ps1", "bat", "fish", "sh")
```
memsharded, line 23: grep in the codebase, who is calling ``accepted_extensions=("sh", ))`` and I guess you want to add ``fish`` there too.

memsharded (commented):
I think it would be good adding 2 tests:
- One basic integration test, that can run in all platforms, that just check the generated file
- One functional test with ``@pytest.mark.tool("fish")``, adding ``fish`` to ``conftest.py`` that you can run locally, and that initially is ``disabled`` in the default ``conftest.py``, and enabled in yours ``conftest_user.py``

conans/client/generators/__init__.py
AbrilRBS: Yep! I also didn't think it would, but I was pleasently supprised that a package for msys2 exists and so I added the subsystem path call here once I found out :)

conan/tools/env/environment.py
```diff
@@ -18,37 +18,43 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
...
     # TODO: This implemantation is dirty, improve it
     for f in filenames:
         f = f if os.path.isabs(f) else os.path.join(env_folder, f)
         if f.lower().endswith(".sh"):
             if os.path.isfile(f) and "sh" in accept:
                 f = subsystem_path(subsystem, f)
                 shs.append(f)
+        elif f.lower().endswith(".fish"):
```
SSE4, line 31: it might be fragile to use extension-based detection.
maybe, it should parse shebang/interpreter as a first try, and only fallback to the extension check if there is no shebang?
that might be worth it, as `.fish` extension is not a standard one like `.sh` and used for few other things (so not every .fish file is a fish script), and vice versa, some shell scripts might have different extension (or might have no extension at all), and it's completely normal in *nix world, where executable files are identified by the interpreter/shebang string instead of extension.

conan/tools/env/environment.py
```diff
@@ -18,37 +18,43 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
...
             if os.path.isfile(path_ps1) and "ps1" in accept:
                 ps1s.append(path_ps1)
             if os.path.isfile(path_sh) and "sh" in accept:
                 path_sh = subsystem_path(subsystem, path_sh)
                 shs.append(path_sh)
+            if os.path.isfile(path_fish) and "fish" in accept:
+                fishes.append(path_fish)
 
-    if bool(bats) + bool(shs) + bool(ps1s) > 1:
+    if bool(bats) + bool(shs) + bool(ps1s) + bool(fishes) > 1:
```
SSE4, line 55: adding bool as a number looks weird here...
I suppose any `bool` has only two values, `true` (1) and `false` (0).
therefore, a condition like `bool_var > 1` has no sense, as both valid values `true`/`false` are never greater than one.
I don't think we define `plus` operator for boolean values, so this code doesn't look good.
even if it works as expected, I'd suggest to re-write it to avoid computing sums of boolean values.

conan/tools/env/environment.py
```diff
@@ -486,11 +495,43 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+        deactivate_file = os.path.join(filepath, "deactivate_{}".format(filename))
+        deactivate = textwrap.dedent("""\
+                   echo "echo Restoring environment" > "{deactivate_file}"
+                   for v in {vars};
+                       set is_defined true
+                       set value (printenv $v) || set is_defined "" || true
+                       if [ -n "$value" ]
+                           or [ -n "$is_defined" ];
+                           echo "set --export $v '$value'" >> "{deactivate_file}";
+                       else; echo set --erase $v >> "{deactivate_file}";
```
SSE4, line 509: I am not an expert in `fish` shell, but it appears as a subject for an injection.
if I set `v = something && rm -rf /` it will evaluate into `echo set --erase something && rm -rf /`, which is dangerous.
I don't know for sure, but perhaps it needs some sort of quoting or sanitation to avoid interpreting values as commands.

conan/tools/env/environment.py
```diff
@@ -500,20 +541,28 @@ def save_script(self, filename):
...
             else:
-                filename = filename + (".bat" if is_bat else ".sh")
+                filename = filename + ".sh"
 
         path = os.path.join(self._conanfile.generators_folder, filename)
         if is_bat:
             self.save_bat(path)
         elif is_ps1:
             self.save_ps1(path)
+        elif is_fish:
```
SSE4, line 564: as number of values grows here, it keeps adding more and more chained conditions.
some sort of generalization would be nice to apply (at least map `format` -> `save_func`, or maybe as a new parameterized function).

conan/tools/env/environment.py
AbrilRBS: You're right that this looks a bit iffy if you're not used to it, but in actuality, in Python Bool is a subtype of Int, so operations such as addition are well-defined https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex

https://godbolt.org/z/PdYvYhfco

AbrilRBS: But you're right that there might be a more expressive way for this, thanks for the suggestion!

conan/tools/env/environment.py
AbrilRBS: From Fish' help:
> The names given to variables and functions (so called “identifiers”) have to follow certain rules:
 A variable name cannot be empty. It can contain only letters, digits, and underscores. It may begin and end with any of those characters.

It will break sooner than this if any bad char is found, but I guess it does not hurt to be triple-sure :)

conan/tools/env/environment.py
AbrilRBS: I'm all for clean code, and if any more are added in the future, that'd be probably the right call :)

But right now I think that `save_script` is the perfect place to have this logic and an if with 3 branches is the easiest solution

AbrilRBS:
Closing this until I think more about this and have a bit more time to get into it. Currently there's a problem where the build() method tries to `self.run` code with a conanbuild environment, and this fails at `Popen` as it uses `sh` by default, and the generated `cananbuild.fish` is not parseable. This works great for VirtualRunEnv so I'm thinking that maybe reducing the scope of the PR to only generate fish files for this generators could be an ok tradeoff

PR #12802, opened 2022-12-28
Add barebones template to `conan new`

AbrilRBS:
Changelog: Feature: Add barebones template option to conan new.
Docs: https://github.com/conan-io/docs/pull/2881

While testing some functionality, I had the need to create a very basic recipe to test some commands on. As I thought that "there probably is a template for it in conan new" and was disappointed when it wasn't.

I had to implement the ability to add multiple values to the same definition, so something like `-d requires=math/1.0 -d requires=ai/1.3` would result in
```py
def requirements(self):
  self.requires("math/1.0")
  self.requires("ai/1.3")
```
without breaking

- [ ] (Will do) I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one. 


conan/internal/api/new/barebones.py
```diff
@@ -0,0 +1,11 @@
+_conanfile = '''\
+from conan import ConanFile
+
+class BarebonesConanfile(ConanFile):
+    name = "barebones"
```
memsharded: Isn't the literal "barebones" name a bit unexpected? The ``cmake_lib`` is not called ``cmake_lib/1.0``. Maybe something neutral like ``pkg``?

conan/internal/api/new/barebones.py
```diff
@@ -0,0 +1,11 @@
...
+    version = "0.0"
+    description = "A bare-bones recipe"
```
memsharded: I like the idea of a barebones recipe that does nothing, but maybe we can use it as a empty skeleton that guides users to some simple use cases? Like adding a ``requirements()`` method with a commented out ``self.requires()``, a ``build()``, ``package()`` methods, with commented out examples and explanations?

Does this make sense? probably not worth the effort?

Maybe it is easy to allow customizations? for example, adding ``-d requires=xxx`` and let the template add requires, allow custom (but not mandatory) pkg name and version, etc.

conan/cli/commands/new.py
```diff
@@ -15,7 +15,7 @@ def new(conan_api, parser, *args):
     Create a new recipe (with conanfile.py and other files) from a predefined template
     """
     parser.add_argument("template", help="Template name, built-in predefined one or user one. "
-                        "You can use built-in templates: cmake_lib, cmake_exe, "
+                        "You can use built-in templates: basic, cmake_lib, cmake_exe, "
```
memsharded: The ``basic`` name doesn't match the ``barebones`` name above.
I don't love the ``barebones`` as a ``conan new barebones``, and I think I'd prefer ``conan new basic``, but naming is hard...

memsharded (commented):
I like the idea of having some other convenient templates, they are useful for teaching, testing, examples, etc. I think we can also provide other improvements for the templates, like some templates with ``requires`` (not in this PR, just an idea).

Please check the comments, add some simple test to cover it, and it should be good.

conan/cli/commands/new.py
SSE4: IMO `basic` is better than `barebones` for sure (what bones? sacrums? and why are they bare? xD)
I would avoid slang terms as much as possible, as they might confuse non-native speakers

conan/internal/api/new/barebones.py
AbrilRBS: Sounds great, and I think that brings enough value to new users that might just want to mess around to be worth the effort :) I'll get on it

conan/api/subapi/new.py
```diff
@@ -82,12 +84,20 @@ def _read_files(self, template_folder):
 
         return template_files, non_template_files
 
+    @staticmethod
+    def render_requires(requires):
+        if isinstance(requires, str):
+            requires = [requires]
+
+        return "\n".join([f'self.requires("{require}")' for require in requires])
+
```
AbrilRBS: So that it can be reused in the future cmake templates PR

conan/internal/api/new/basic.py
```diff
@@ -0,0 +1,29 @@
+_conanfile = '''\
+from conan import ConanFile
+
+class BasicConanfile(ConanFile):
+    name = "{% if name is defined %}{{ name }}{% else %}pkg{% endif %}"
+    version = "{% if version is defined %}{{ version }}{% else %}1.0{% endif %}"
+    description = "{% if description is defined %}{{ description }}{% else %}A bare-bones recipe{% endif %}"
+
+    # Text about the requirements() method
```
AbrilRBS: Todo text

conan/api/subapi/new.py
```diff
@@ -84,12 +84,20 @@ def _read_files(self, template_folder):
 
         return template_files, non_template_files
 
+    @staticmethod
+    def render_requires(requires):
```
memsharded: Why not using a ``for`` loop in the template directly? This seems a bit hybrid, doing some rendering here and some in the template?

conans/test/utils/tools.py
```diff
@@ -718,6 +719,15 @@ def assert_listed_binary(self, requires, build=False, test=False):
             else:
                 raise AssertionError(f"Cant find {r}-{kind} in {reqs}")
 
+    def assert_in_file_contents(self, filename, contents: List[str], path=None):
```
AbrilRBS: I looked around (admittedly not _too_ hard) and saw no utility that implemented something like this, and it feels like moving this common check here would be ok

AbrilRBS (commented):
Thanks for the reviews, I've implemented most of the requested changes, with only the explanation comments left to do

memsharded (approved):
Looking good

conan/api/subapi/new.py
AbrilRBS: You're right. I've moved the rendering to jinja and kept only the logic here, good suggestion!

RubenRBS:
Wrong button, not ready for review yet! I need to get the explanation comments done still and fix the test

conan/internal/api/new/basic.py
```diff
@@ -1,13 +1,24 @@
+def inject_get_or_else(variable, default):
```
memsharded, line 1: There is a way to create functions in jinja templates directly, but I am fine with this too.

memsharded (approved):

conans/test/integration/command/new_test.py
```diff
@@ -45,6 +45,21 @@ def test_new_missing_definitions(self):
...
+            tc.assert_in_file_contents("conanfile.py", ['# self.requires("zlib/1.2.13")'])
+        except Exception as e:
+            assert False, f"conan new basic should work without any extra arguments: {e}"
+
+        tc.run("new basic -d name=mygame -d requires=math/1.0 -d requires=ai/1.0 -f")
+        tc.assert_in_file_contents("conanfile.py", [
+            'self.requires("math/1.0")',
+            'self.requires("ai/1.0")',
+            'name = "mygame"'
+        ])
```
memsharded:
```suggestion
        conanfile = tc.load("conanfile.py")
        assert 'self.requires("math/1.0")' in conanfile
        assert 'self.requires("ai/1.0")' in conanfile
        assert 'name = "mygame"' in conanfile
```

There is an evil side in moving asserts to helpers (even though I also do it often), and is that those asserts do not report the same, and it is worse experience for developers when they fail.
It doesn't seem the helper really saves many lines, and the result is equally readable?

conans/test/utils/tools.py
```diff
@@ -718,6 +719,15 @@ def assert_listed_binary(self, requires, build=False, test=False):
...
+        path = path or self.current_folder
+        file = os.path.join(path, filename)
+        assert os.path.exists(file)
+        with open(file) as f:
+            file_contents = "\n".join(f.readlines())
```
memsharded: It is possible to do ``file_contents = f.read()`` to read the whole file, no need to readlines and join them

conans/test/integration/command/new_test.py
AbrilRBS: Oh I missed the fact that TestClient.load did exactly what I wanted to! Thanks for the headsup :)

conans/test/integration/command/new_test.py
```diff
@@ -45,6 +45,20 @@ def test_new_missing_definitions(self):
         client.run("new cmake_lib -d version=myversion", assert_error=True)
         assert "Missing definitions for the template. Required definitions are: 'name', 'version'" in client.out
 
+    def test_new_basic_template(self):
+        tc = TestClient()
+        try:
+            tc.run("new basic")
+            assert '# self.requires("zlib/1.2.13")' in tc.load("conanfile.py")
+        except Exception as e:
+            assert False, f"conan new basic should work without any extra arguments: {e}"
```
memsharded:
```suggestion
    tc.run("new basic")
    assert '# self.requires("zlib/1.2.13")' in tc.load("conanfile.py")
```
It is not necessary to capture unexpected exceptions. If that happens the test would fail anyway. Otherwise we would need to capture absolutely every ``tc.run()`` with try-except just for an extra message? 
In other words, line 54 is never executed, which doesn't make much sense.

conans/test/utils/tools.py
```diff
@@ -14,6 +14,7 @@
 import zipfile
 from collections import OrderedDict
 from contextlib import contextmanager
+from typing import List
```
memsharded: Not necessary now?

conans/test/utils/tools.py
```diff
@@ -816,6 +817,7 @@ def search(self, pattern, remote=None, assert_error=False, args=None):
         data = json.loads(self.load(".tmp.json"))
         return data
 
+
```
memsharded: why extra blank line?

AbrilRBS: Huh, weird, no idea how it got there, good catch

memsharded (approved):

Issue #12155, opened 2022-09-20
[question] What is the equivalent of `conans.tools.MSBuild.build()` `upgrade_project` in `conan.tools.microsoft.MSBuid`?


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

To build `xz_utils/5.2.5` with Visual Studio 2022 in https://github.com/conan-io/conan-center-index/pull/13038, vcxproj file must be updated with a different Platform Toolset, how to achieve that in conan v2?

(otherwise it fails with this kind of error since we try to build against a vs2017 solution:

```
C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppBuild.targets(460,5): error MSB8020: The build tools for Visual Studio 2017 (Platform Toolset = 'v141') cannot be found.
To build using the v141 build tools, please install Visual Studio 2017 build tools.
Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Retarget solution". [C:\Users\spaceim\.conan\data\xz_utils\5.2.5\_\_\build\5a61a86bb3e07ce4262c80e1510f9c05e9b6d48b\src\windows\vs2017\liblzma.vcxproj]
```
)

SpaceIm:
I had to write something like this in `xz_utils`:

```python
    def _fix_msvc_platform_toolset(self, vcxproj_file, old_toolset):
        platform_toolset = {
            "Visual Studio": {
                "8": "v80",
                "9": "v90",
                "10": "v100",
                "11": "v110",
                "12": "v120",
                "14": "v140",
                "15": "v141",
                "16": "v142",
                "17": "v143",
            },
            "msvc": {
                "170": "v110",
                "180": "v120",
                "190": "v140",
                "191": "v141",
                "192": "v142",
                "193": "v143",
            }
        }.get(str(self.settings.compiler)).get(str(self.settings.compiler.version))
        replace_in_file(
            self,
            vcxproj_file,
            f"<PlatformToolset>{old_toolset}</PlatformToolset>",
            f"<PlatformToolset>{platform_toolset}</PlatformToolset>",
        )
```

then:

```python
        if (self.settings.compiler == "Visual Studio" and Version(self.settings.compiler) >= "15") or \
[...]

Issue #3859, opened 2018-10-26
[build system][MSBuild] Investigate Microsoft.Cpp.Default.props


memsharded:
https://msdn.microsoft.com/en-us/library/669zx6zc.aspx

This file is automatically loaded by the IDE, so might be a good mechanism to define conan dependencies properties.

solvingj:
The other option is `Directory.Build.props`  and/or `Directory.Build.targets`. 

https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2017

memsharded:
I was investigating this, and the value seems marginal:
It only works in VS >15. The default can't be changed without breaking, so generating ``conanbuildinfo.props`` is still a must. Writing also a Directory.Build.props could overwrite existing users file, not easy to differentiate if it is a file from user or a file from Conan (yes we could check some patterns, but still seems a bit fragile).

We will have a look again when revisiting the build helpers and generators.

SSE4:
about MSBuild targets vs Add-on (plug-in):
the main disadvantage of MSBuild targets is inability to subscribe on solution load event. as result, it has the following consequences (as GUI and IntelliSense load before any MSBuild tasks run):
- IntelliSense doesn't work
- developers have to reload solution
- non-intuitive workflow (run build -> it generates conanbuildinfo.props -> reload -> run build again)
some other flaws of MSBuild tasks:
- hard to distribute, deploy and provide versioning
- they are not designed to interact with the project structure or anything visible in the VS GUI
- they can't trigger other events in the GUI like refresh
- seems like Microsoft isn't going to provide resources for props, instead, they are focusing on CMake/OpenFolder

santhonisz:
Hi there,

As part of our preparation for the upcoming release, I have been looking at the latest Conan 2.0 MSBuild generators and came to the same thought as was raised many years ago in this issue. Namely, I was wondering of the value in the creation/modification of a `Directory.Build.props` file as part of the new `MSBuildToolchain` generator. This would provide a very convenient approach to automatically including the new `conantoolchain.props` file without having to manually add it the relevant `.vcxproj` file. In our company's case we have literally hundreds of individual projects and Conan recipes, so anything that can steamline the process of updating to the new generators would be beneficial.

I have done some experimentation and by using the `ForceImportAfterCppDefaultProps` property the `conantoolchain.props` can be imported as desired, ensuring the platform toolset is applied e.g.

```xml
<?xml version="1.0" ?>
[...]

PR #12817, opened 2022-12-31
MSBuild: add argument to build() to opt-in autoconsumption of props files generated by MSBuildToolchain & MSBuildDeps

SpaceIm:
related to https://github.com/conan-io/conan/issues/12155#issuecomment-1252832378

Changelog: (Fix): In `MSBuild`, add an argument to `build()` method to opt-in automatic consumption of props files generated by `MSBuildToolchain` & `MSBuildDeps`, without requiring intrusive modification of vcxproj files.
Docs: https://github.com/conan-io/docs/pull/2885

### DIFF
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -1,3 +1,5 @@
+import os
+
 from conans.errors import ConanException
 
 
@@ -45,6 +47,15 @@ class MSBuild(object):
             if not isinstance(targets, list):
                 raise ConanException("targets argument should be a list")
             cmd += " /target:{}".format(";".join(targets))
+
+        props_paths = []
+        for props_file in ("conantoolchain.props", "conandeps.props"):
+            props_path = os.path.join(self._conanfile.generators_folder, props_file)
+            if os.path.exists(props_path):
+                props_paths.append(props_path)
+        if props_paths:
+            props_paths = ";".join(props_paths)
+            cmd += f" /p:ForceImportBeforeCppTargets=\"{props_paths}\""
 
         return cmd
 
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -70,7 +70,7 @@ class MSBuild(object):
                     break
 
         if concrete_props_file:
-            concrete_props_file = os.path.join(self._conanfile.generators_folders, concrete_props_file)
+            concrete_props_file = os.path.join(self._conanfile.generators_folder, concrete_props_file)
 
         if not concrete_props_file or not os.path.exists(concrete_props_file):
             raise ConanException(
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -1,5 +1,7 @@
 import os
 
+from conan.tools.microsoft.msbuilddeps import MSBuildDeps
+from conan.tools.microsoft.toolchain import MSBuildToolchain
 from conans.errors import ConanException
 
 
@@ -49,7 +51,7 @@ class MSBuild(object):
             cmd += " /target:{}".format(";".join(targets))
 
         props_paths = []
-        for props_file in ("conantoolchain.props", "conandeps.props"):
+        for props_file in (MSBuildToolchain.filename, MSBuildDeps.filename):
             props_path = os.path.join(self._conanfile.generators_folder, props_file)
             if os.path.exists(props_path):
                 props_paths.append(props_path)
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -85,13 +85,16 @@ class MSBuild(object):
         namespace = namespace.group(0) if namespace else ""
         importgroup_element = root.find(f"{namespace}ImportGroup")
         if importgroup_element:
-            expected_condition = (f"'$(Configuration)' == '{self.configuration}' And "
-                                  f"'$(Platform)' == '{msbuild_arch_to_conf_arch(self.platform)}'")
-            for import_element in importgroup_element.iter(f"{namespace}Import"):
-                condition = import_element.attrib.get("Condition")
-                if expected_condition == condition:
-                    concrete_props_file = import_element.attrib.get("Project")
-                    break
+            import_elements = importgroup_element.findall(f"{namespace}Import")
+            if len(import_elements) == 1:
+                concrete_props_file = import_elements[0].attrib.get("Project")
+            else:
+                expected_condition = (f"'$(Configuration)' == '{self.configuration}' And "
+                                      f"'$(Platform)' == '{msbuild_arch_to_conf_arch(self.platform)}'")
+                for import_element in import_elements:
+                    if expected_condition == import_element.attrib.get("Condition"):
+                        concrete_props_file = import_element.attrib.get("Project")
+                        break
 
         if concrete_props_file:
             concrete_props_file = os.path.join(self._conanfile.generators_folder, concrete_props_file)
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -4,7 +4,6 @@ import xml.etree.ElementTree as ET
 
 from conan.tools.microsoft.msbuilddeps import MSBuildDeps
 from conan.tools.microsoft.toolchain import MSBuildToolchain
-from conan.tools.microsoft.visual import msbuild_arch
 from conans.errors import ConanException
 
 
@@ -16,6 +15,20 @@ def msbuild_verbosity_cmd_line_arg(conanfile):
         return '/verbosity:{}'.format(verbosity)
 
 
+def msbuild_arch(arch):
+    return {'x86': 'x86',
+            'x86_64': 'x64',
+            'armv7': 'ARM',
+            'armv8': 'ARM64'}.get(str(arch))
+
+
+def msbuild_arch_to_conf_arch(arch):
+    return {'x86': 'Win32',
+            'x64': 'x64',
+            'ARM': 'ARM',
+            'ARM64': 'ARM64'}.get(str(arch))
+
+
 class MSBuild(object):
     def __init__(self, conanfile):
         self._conanfile = conanfile
@@ -23,7 +36,8 @@ class MSBuild(object):
         self.configuration = conanfile.settings.get_safe("build_type")
         # if platforms:
         #    msvc_arch.update(platforms)
-        msvc_arch = msbuild_arch(conanfile)
+        arch = conanfile.settings.get_safe("arch")
+        msvc_arch = msbuild_arch(arch)
         if conanfile.settings.get_safe("os") == "WindowsCE":
             msvc_arch = conanfile.settings.get_safe("os.platform")
         self.platform = msvc_arch
@@ -68,7 +82,8 @@ class MSBuild(object):
         namespace = namespace.group(0) if namespace else ""
         importgroup_element = root.find(f"{namespace}ImportGroup")
         if importgroup_element:
-            expected_condition = f"'$(Configuration)' == '{self.configuration}' And '$(Platform)' == '{self.platform}'"
+            expected_condition = (f"'$(Configuration)' == '{self.configuration}' And "
+                                  f"$(Platform)' == '{msbuild_arch_to_conf_arch(self.platform)}'")
             for import_element in importgroup_element.iter(f"{namespace}Import"):
                 condition = import_element.attrib.get("Condition")
                 if expected_condition == condition:
@@ -81,8 +96,7 @@ class MSBuild(object):
         if not concrete_props_file or not os.path.exists(concrete_props_file):
             raise ConanException(
                 f"MSBuildToolchain props file is missing for configuration={self.configuration} and "
-                f"platform={self.platform}. 'configuration' and 'platform' attributes must be the same "
-                "respectively in MSBuildToolchain and MSBuild."
+                f"platform={msbuild_arch_to_conf_arch(self.platform)}."
             )
 
         return concrete_props_file
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -34,16 +34,12 @@ class MSBuild(object):
         self.platform = msvc_arch
         self.toolset = msvs_toolset(conanfile)
 
-    def command(self, sln, targets=None, auto_inject_deps_props=True):
+    def command(self, sln, targets=None):
         cmd = (f'msbuild "{sln}" /p:Configuration={self.build_type} '
                f"/p:Platform={self.platform} /p:PlatformToolset={self.toolset}")
 
-        # Autoconsume toolchain props, but opt-out dependencies props
         props_paths = []
-        props_candidates = [MSBuildToolchain.filename]
-        if auto_inject_deps_props:
-            props_candidates.append(MSBuildDeps.filename)
-        for props_file in props_candidates:
+        for props_file in (MSBuildToolchain.filename, MSBuildDeps.filename):
             props_path = os.path.join(self._conanfile.generators_folder, props_file)
             if os.path.exists(props_path):
                 props_paths.append(props_path)
@@ -67,8 +63,8 @@ class MSBuild(object):
 
         return cmd
 
-    def build(self, sln, targets=None, auto_inject_deps_props=True):
-        cmd = self.command(sln, targets=targets, auto_inject_deps_props=auto_inject_deps_props)
+    def build(self, sln, targets=None):
+        cmd = self.command(sln, targets=targets)
         self._conanfile.run(cmd)
 
     @staticmethod
--- a/conan/tools/microsoft/msbuild.py
+++ b/conan/tools/microsoft/msbuild.py
@@ -1,7 +1,9 @@
 import os
+import xml.etree.ElementTree as ET
 
 from conan.tools.microsoft.msbuilddeps import MSBuildDeps
 from conan.tools.microsoft.toolchain import MSBuildToolchain
+from conan.tools.microsoft.visual import msbuild_arch
 from conans.errors import ConanException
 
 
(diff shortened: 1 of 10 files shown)


SpaceIm:
I've tested with libsodium recipe (without this ugly workaround https://github.com/conan-io/conan-center-index/blob/4e0dc05121ed9999dcd3ed8841844757f6495ce0/recipes/libsodium/all/conanfile.py#L137-L171 -mandatory without the fix in this PR-, and a fix in custom `configuration` of its `MSBuildToolchain`) and it works like a charm, now vc runtime can be honored (as well as multiprocessing build).

SpaceIm:
see these conan-center PR for current workarounds:
https://github.com/conan-io/conan-center-index/pull/15041
https://github.com/conan-io/conan-center-index/pull/15043
https://github.com/conan-io/conan-center-index/pull/15046
https://github.com/conan-io/conan-center-index/pull/15050

It's more or less always the same boilerplate (boilerplate we could remove with this PR):
https://github.com/conan-io/conan-center-index/blob/4a2cbbba7fce7da91a656890744f3463c7429846/recipes/libsodium/all/conanfile.py#L130-L145
https://github.com/conan-io/conan-center-index/blob/22d52120bb27d81d551ada9255e012755b023276/recipes/xz_utils/all/conanfile.py#L104-L128
https://github.com/conan-io/conan-center-index/blob/110eb267a9776031a5189abd06254b07c88abf25/recipes/libmicrohttpd/all/conanfile.py#L137-L151
https://github.com/conan-io/conan-center-index/blob/a4ba448dc6c474198d669242ea5e1e626349581e/recipes/zimg/all/conanfile.py#L93-L108

conan/tools/microsoft/msbuild.py
```diff
@@ -27,10 +32,20 @@ def __init__(self, conanfile):
...
-               % (sln, self.build_type, self.platform))
+        cmd = (f'msbuild "{sln}" /p:Configuration={self.build_type} '
+               f"/p:Platform={self.platform} /p:PlatformToolset={self.toolset}")
+
+        props_paths = []
+        for props_file in (MSBuildToolchain.filename, MSBuildDeps.filename):
+            props_path = os.path.join(self._conanfile.generators_folder, props_file)
+            if os.path.exists(props_path):
+                props_paths.append(props_path)
+        if props_paths:
```
memsharded: This cannot be made default, because that will be forcing the injection of whole dependencies into the current project, and there are many users that don't want that, but manual addition of individual dependencies into different subprojects.

At least, it should have an opt-out to allow users defining it.

conan/tools/microsoft/msbuild.py
```diff
@@ -27,10 +32,20 @@ def __init__(self, conanfile):
         if conanfile.settings.get_safe("os") == "WindowsCE":
             msvc_arch = conanfile.settings.get_safe("os.platform")
         self.platform = msvc_arch
+        self.toolset = msvs_toolset(conanfile)
 
     def command(self, sln, targets=None):
-        cmd = ('msbuild "%s" /p:Configuration=%s /p:Platform=%s'
-               % (sln, self.build_type, self.platform))
+        cmd = (f'msbuild "{sln}" /p:Configuration={self.build_type} '
+               f"/p:Platform={self.platform} /p:PlatformToolset={self.toolset}")
```
memsharded: This adds back the PlatformToolset as argument only, and that doesn't work for user consumption, that will only be loading the toolchain. That configuration needs to be in the toolchain file.

conans/test/functional/toolchains/microsoft/test_msbuilddeps.py
```diff
@@ -115,9 +115,6 @@
   </ImportGroup>
   <ImportGroup Label="Shared">
   </ImportGroup>
-  <ImportGroup Label="PropertySheets">
```
memsharded: This test is testing exactly that different subprojects can depend on individual props, without needing to link all dependencies, please leave this test not modified.

memsharded (changes requested):
There are different things proposed in this PR, and they probably require separate discussions:
- Using /p:ForceImportBeforeCppTargets to automatically inject property files: At least requires an opt-out, but probably an opt-in is better, we need to discuss this.
- Passing the PlatformToolset in the command line is hiding if the PlatformToolset definition from the toolchain works.
- Making visible for users the ``vs_toolset()`` function

It seems that the best would be to make the ``p:ForceImportBeforeCppTargets`` available as opt-in, and cover it with new test, instead of modifying existing tests.

conan/tools/microsoft/msbuild.py
SpaceIm: I've not removed PlatformToolset from toolchain (actually I've not modified content of files generated by toolchains). It's added as an argument to msbuild because when `conantoolchain.props` is injected with `/p:ForceImportBeforeCppTargets`, I've seen that PlatformToolset from this toolchain was ignored.

conan/tools/microsoft/msbuild.py
SpaceIm: So keep this logic for `conantoolchain.props`, but allow to not autoconsume `conandeps.props`?

Regarding what users want, I'm wondering: users who manually edit their MSBuild project files, are they really using MSBuild conan helper? Or just MSBuildToolchain & MSBuildDeps?
This PR is really about providing the same high level logic in this toolchain/helper family than all other families in conan v2. It's the only family where the build helper does not automatically consume its toolchains. There is no opt-in or opt-out in Autotools helper to not globally inject AutotoolsDeps, so why a different behavior in MSBuild helper?

conan/tools/microsoft/msbuild.py
SpaceIm: Actually, from my experimentation when `conantoolchain.props` is injected with `/p:ForceImportBeforeCppTargets`:
- Everything under `ItemDefinitionGroup` is honored.
- Everything under `PropertyGroup Label="Configuration"` (`PlatformToolset` & custom properties) is ignored by MSBuild.

This is why I inject `/p:PlatformToolset` in msbuild command line (and if I could, I would provide a properties argument to `msbuild` build helper to enforce properties in command line).

SpaceIm: https://learn.microsoft.com/en-us/cpp/build/project-property-inheritance

> When you set a property with property pages, it's also important to pay attention to where you set it. If you set a property to "X" in a .props file, but the property is set to "Y" in the project file, then the project will build with the property set to "Y".

https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-properties

> MSBuild lets you set properties on the command line by using the -property (or -p) switch. These global property values override property values that are set in the project file. This includes environment properties, but does not include reserved properties, which cannot be changed.

conan/tools/microsoft/msbuild.py
memsharded: From the feedback from users using this integration, it is very common that they have different subprojects and they want to inject different dependencies to different subprojects. If MSBuild pass ``conandeps.props`` by default, this cannot be done.

There are definitely other build systems that do the same, CMake uses the same pattern, users need to define ``find_package()`` and ``target_link_libraries()`` to the things that they want to use, it is not automagically applied to the whole project. For sure that is a built-in construct, but this level of control is what can be achieved with the MSBuild integration. The fact that it is not possiblel to do it in Autotools, for example, shouldn't be used as the design goal for it.

conan/tools/microsoft/msbuild.py
memsharded: The problem with ``/p:ForceImportBeforeCppTargets`` is the developer experience. Most developers with VS will just double click the project or solution Whatever is being passed should be in the ``vcxproj`` by default, it cannot rely on cli args to work. This is probably the most massive use case of this integration, there are companies out there using it, and hundreds of developers do this daily. So this cannot be changed if this doesn't keep working as-is (this is why I suggest an opt-in for ConanCenter recipes, which are typically not opened by developers to touch the source code)

SpaceIm: There is no breaking change, I've not changed toolchain files.

Here is another proposal:

- add something like `props_autoinjection` argument to `build()` method of `MSBuild`, `False` by default.
- add properties dict attribute to `MSBuild`
- when `props_autoinjection` is False, current behavior is kept, so no breaking change at all.
- when `props_autoinjection` is True:
  - it collects root props of `MSBuildToolchain` & `MSBuildDeps` and inject them to `/p:ForceImportBeforeCppTargets`.
  - it also injects `/p:<property>=<value>` to msbuild command line
- keep previous tests as is, an write other for the `props_autoinjection=True` flavor.

SpaceIm: Regarding properties, to avoid duplication between toolchain and MSBuild, there is a more complex solution: gather all properties by parsing all props of a specific configuration programmatically, and inject them in command line if props_autoinjection is True.

memsharded: Yes, sorry, I meant the ``/p:ForceImportBeforeCppTargets`` is a breaking change when applied to ``conandeps.props``. For the toolchain, the main issue is that it can easily be hiding errors for developers opening the IDE

Yes, I think this approach would be better. Just a couple of suggestions:
- Make explicit the intention, ``props_autoinjection`` is a bit generic, something like ``force_imports`` or ``force_import_generated_files``
- Generic property injection is a different feature and ``MSBuildToolchain`` already provides injection of properties via toolchain (so they work for developers opening the IDE, without needing the Conan ``build()`` injection), so this should be a separate discussion if there is a use case for it.

SpaceIm: > Generic property injection is a different feature and MSBuildToolchain already provides injection of properties via toolchain (so they work for developers opening the IDE, without needing the Conan build() injection), so this should be a separate discussion if there is a use case for it.

As I said in https://github.com/conan-io/conan/pull/12817#discussion_r1059935272, properties defined in toolchain files are not honored when injected through `/p:ForceImportBeforeCppTargets`, so for me it's related. `/p:ForceImportBeforeCppTargets` would be an internal detail of `force_import_generated_files =True`, the goal of this argument is to honor everything defined in these props, by any means.

conan/tools/microsoft/msbuild.py
SpaceIm: I mostly agree for `*Deps` generators (even thought it's transparent in case of CMake), but `MSBuild` is the only build helper not consuming automatically & transparently its "glue toolchain".

conan/tools/microsoft/visual.py
```diff
@@ -268,3 +268,14 @@ def is_msvc_static_runtime(conanfile):
     :return: True, if msvc + runtime MT. Otherwise, False
     """
     return is_msvc(conanfile) and "MT" in msvc_runtime_flag(conanfile)
+
+
+def msbuild_arch(conanfile):
+    if conanfile.settings.get_safe("os") == "WindowsCE":
+        return conanfile.settings.get_safe("os.platform")
```
memsharded: Those 2 lines are new logic not previously in the code, and those are untested. Better leave them as-is, and if necessary, treat the problem of WindowsCE in a different PR.

conan/tools/microsoft/msbuild.py
```diff
@@ -48,11 +47,75 @@ def command(self, sln, targets=None):
...
+    def build(self, sln, targets=None, force_import_generated_files=False):
+        cmd = self.command(sln, targets=targets, force_import_generated_files=force_import_generated_files)
         self._conanfile.run(cmd)
 
     @staticmethod
     def get_version(_):
         return NotImplementedError("get_version() method is not supported in MSBuild "
                                    "toolchain helper")
+
+    def _get_concrete_props_file(self, root_props_file):
```
memsharded, line 80: I am not sure why the specific file needs to be passed, if the general toolchain file already contains conditionals to pick the right configuration one.

conan/tools/microsoft/msbuild.py
```diff
@@ -48,11 +47,75 @@ def command(self, sln, targets=None):
...
+                "respectively in MSBuildToolchain and MSBuild."
+            )
+
+        return concrete_props_file
+
+    def _get_msbuildtoolchain_properties(self, root_props_file):
+        properties = {}
+
+        # Get properties from props file of configuration and platform
+        concrete_props_file = self._get_concrete_props_file(root_props_file)
```
memsharded, line 114: Maybe this can be avoided with ``ForceImportAfterCppTargets``?

memsharded (commented):
It seems it is getting better, but I am still a bit concerned about ``MSBuild`` doing too much, parsing XML files to extract the properties...

Also, this new functionality should be covered by new tests, please add them when this converges to something that we agree on. Many thanks!

conan/tools/microsoft/msbuild.py
SpaceIm: > if the general toolchain file already contains conditionals to pick the right configuration one.

Because it doesn't work.
I'm testing with several CCI recipes for the moment, and properties are ignored when they come only from props file (msbuild complains about PlatformToolset since it's not overridden, and when there is a hardcoded WholeProgramOptimization I can't disable it through a custom property in MSBuildToolchain). When I extract these values from toolchain file and inject them in command line, it works.

conan/tools/microsoft/msbuild.py
SpaceIm: Yes maybe, I can try

SpaceIm: Nop, it doesn't help :(

conan/tools/microsoft/msbuild.py
SpaceIm: (obviously, I would be more than happy if a smart solution not involving to parse xml files could be found)

conan/tools/microsoft/msbuild.py
```diff
@@ -16,10 +22,18 @@ def msbuild_arch(arch):
             'armv8': 'ARM64'}.get(str(arch))
 
 
+def msbuild_arch_to_conf_arch(arch):
+    return {'x86': 'Win32',
```
SpaceIm: I'm a little bit surprised of this difference. In conan-center, recipes using legacy MSBuild were always overriding x86 to pass Win32. Is it not a mistake?

conan/tools/microsoft/visual.py
SpaceIm: reverted

SpaceIm:
```
E                   TypeError: test_toolchain_win_vs2017() missing 5 required positional arguments: 'compiler', 'version', 'runtime', 'cppstd', and 'force_import_generated_files'
```

I don't understand. What am I missing?

EDIT: ok there is a mixup of Unittest, parameterized and pytest.

SpaceIm:
@memsharded I've updated a functional test_msbuild test to test both flavors of force_import_generated_files.

conan/tools/microsoft/toolchain.py
```diff
@@ -52,27 +61,23 @@ def __init__(self, conanfile):
         self.cflags = []
         self.ldflags = []
         self.configuration = conanfile.settings.build_type
+        self.platform = arch_to_vcproj_platform(str(conanfile.settings.arch))
```
SpaceIm, line 64: I've exposed platform attribute in MSBuildToolchain because it was exposed in MSBuildDeps and they share the same underlying logic, but I'm not sure it makes sense. Is it really customizable in MSBuild files?

conans/test/functional/toolchains/microsoft/test_msbuild.py
```diff
@@ -1,13 +1,10 @@
...
 import platform
-import shutil
 import textwrap
-import unittest
 
 import pytest
-from parameterized import parameterized, parameterized_class
 
-from conan.tools.microsoft.visual import vcvars_command
+from conan.tools.microsoft.visual import msvc_version_to_vs_ide_version, vcvars_command
```
memsharded, line 7: Please do not modify existing tests. There is a risk of breaking some functionality. The feature that you are proposing is a opt-in, so the previous tests shouldn't be modified, and adding a new test that uses the new functionality is the way to go.

conan/tools/microsoft/toolchain.py
memsharded: I am ok with it, I think it is customizable, but in the worst case, it is not critical that it is exposed, just don't change it if you don't need to change it.

conan/tools/microsoft/msbuild.py
```diff
@@ -48,11 +68,83 @@ def command(self, sln, targets=None):
...
+                    properties[propert_name] = child.text
+        return properties
+
+    def _force_import_generated_files_cmd_line_arg(self):
+        cmd_args = []
+        props_paths = []
+
+        # MSBuildToolchan must be in generators for this MSBuild mode
+        msbuildtoolchain_file = os.path.join(self._conanfile.generators_folder, MSBuildToolchain.filename)
+        if not os.path.exists(msbuildtoolchain_file):
```
memsharded, line 131: It is possible that some users want to inject only dependencies information, but they don't want to use the toolchain at all? Sounds very possible.

conan/tools/microsoft/msbuild.py
```diff
@@ -16,10 +22,21 @@ def msbuild_arch(arch):
             'armv8': 'ARM64'}.get(str(arch))
 
 
+def msbuild_arch_to_conf_arch(arch):
```
memsharded, line 25: If used only once, then it is better to keep it local initially, until repetition pattern clearly arises (ideally at least 3 occurrences)

conan/tools/microsoft/msbuild.py
memsharded: I still see a lot of complexity in this functionality.
Wouldn't it be way more straightforward to just inject the ``<import>`` section in the ``vcxproj`` directly?

memsharded (commented):
I am still seeing quite a complexity in this PR that I feel it is not great for future issues and maintenance.
For all company usage, adding the .props files to the projects is necessary anyway.
For ConanCenter usage, I think there are way simpler alternatives:
- Patching the ``vcxproj`` to inject the .props files, in the recipe
- Implementing the injection of .props files in the ``vcxproj`` with a helper.
In all cases, the functionality should be oriented to get to the desired recommended state: a ``vcxproj`` that imports the ``.props`` files in the right location. Having to parse the xml of the projects to extract import files, to extract properties, to pass them in command line is too complex IMO.

Asking for further reviews.

conans/test/functional/toolchains/microsoft/test_msbuild.py
SpaceIm: It's deeply modified because this test was not easily extendable, but it tests the same things (and even more, properties injection was not tested actually, and it allows to test other Visual Studio versions than 14, it's an issue for local tests). It seems weird to add a separate test with a risk to have discrepancies between both flavors.
But yeah if you want I can move this to another file, it's just a lot of duplicated code (specially this big vcxproj file).

conan/tools/microsoft/msbuild.py
SpaceIm: Il would make little sense. It would be like allowing `CMake` build helper to not consume `conan_toolchain.cmake`.

conan/tools/microsoft/msbuild.py
SpaceIm: It's not exposed in `__init__.py` of `microsoft` folder, so it's not public according to conan documentation. What should I do? There are other non-public functions in this file actually.

SpaceIm:
>I am still seeing quite a complexity in this PR that I feel it is not great for future issues and maintenance.
> For all company usage, adding the .props files to the projects is necessary anyway.

It's necessary for company usage when they work in Visual Studio IDE, but it can be an issue if they want to package an external dependency with build files not under their control, like conan-center.

> For ConanCenter usage, I think there are way simpler alternatives:
> 
> Patching the vcxproj to inject the .props files, in the recipe

It's what I've done in several recipes to fix them (and it was not obvious to find where this file should be injected in vcxproj files, it's very important actually and not documented in conan doc). And before I did, all packages were broken (even the template written by conan-center maintainers) due to the confusion introduced by this discrepancy behavior between MSBuildToolchain/MSBuild and other build helpers families. It's a good indication that it's not intuitive and there is something wrong.

santhonisz:
Just wondering if an alternate/more viable solution to this could be something along the lines of https://github.com/conan-io/conan/issues/3859#issuecomment-1280544191? 

This would be also be a similar approach to what is suggested for `vcpkg`: https://vcpkg.io/en/docs/users/buildsystems/msbuild-integration.html#import-props-and-targets

memsharded:
Following up on some older PRs.
I am going to close this as not planned, it has too much complexity, doing too many different things in the same PR, modifying tests that shouldn't be modified, so the risk is high. Thanks very much for your contribution anyway, it is still interesting and can be useful for future improvements.

If we want to pursue improvement in this direction, it should be done in smaller and more focused steps.
A first possible small step would be doing just the ``cmd_args.append(f"/p:ForceImportBeforeCppTargets=\"{';'.join(props_paths)}\"")`` argument appending, with some minimal new test and not modifying existing ones. If some things do not work because the project hardcodes toolsets or something, it is not a problem, that could be a different thing, better handle it separately.

PR #5502, opened 2019-07-17
Bugfix: Parallel download of files

SSE4:
Changelog: Bugfix: Parallel download of files
Docs: omit
@TAGS: slow, svn
@PYVERS: Macos@py27, Windows@py36, Linux@py27, py34
related: https://github.com/conan-io/conan/issues/5477

benchmark is here: https://github.com/SSE4/conan-speed-test
parallelised function: `_download_files_to_folder` 

basically, it's getting about 2 times faster for me. `tqdm` progress bars work for me for parallel downloads (on OSX).

sequential (overral):
```
conan create . bincrafters/testing -k  8.30s user 3.77s system 5% cpu 4:00.04 total
conan create . bincrafters/testing -k  8.26s user 3.74s system 4% cpu 4:11.55 total
```
parallel (overral):
```
conan create . bincrafters/testing -k  9.13s user 4.71s system 10% cpu 2:12.07 total
conan create . bincrafters/testing -k  9.34s user 4.96s system 11% cpu 2:00.64 total
```

sequential (detailed):
```
boost_random/1.69.0@bincrafters/stable: Not found in local cache, looking in remotes...
boost_random/1.69.0@bincrafters/stable: Trying with 'conan-center'...
boost_random/1.69.0@bincrafters/stable: Trying with 'bincrafters'...
Downloading conanmanifest.txt: 100%|##########| 103/103 [00:00<00:00, 97.5kB/s]
Downloading conanfile.py: 100%|##########| 496/496 [00:00<00:00, 382kB/s]
Downloading conan_export.tgz: 100%|##########| 758/758 [00:00<00:00, 571kB/s]
AAA download files took: 0.755559921265
Decompressing conan_export.tgz: 766B [00:00, 195kB/s]
boost_random/1.69.0@bincrafters/stable: Downloaded recipe revision 0
boost_base/1.69.0@bincrafters/stable: Not found in local cache, looking in remotes...
boost_base/1.69.0@bincrafters/stable: Trying with 'conan-center'...
boost_base/1.69.0@bincrafters/stable: Trying with 'bincrafters'...
Downloading conanmanifest.txt: 100%|##########| 103/103 [00:00<00:00, 80.8kB/s]
Downloading conanfile.py: 100%|##########| 10.1k/10.1k [00:00<00:00, 9.81MB/s]
Downloading conan_export.tgz: 100%|##########| 758/758 [00:00<00:00, 589kB/s]
AAA download files took: 0.62668299675
[...]

PR #12828, opened 2023-01-02
Feature/develop2 restapi parallel download

memsharded:
Changelog: omit
Docs: omit

- Follow-up from https://github.com/conan-io/conan/pull/5502, trying to speed up installation of dependencies with threads (measured in my machine speed-up of 20-30%)
- Minor cleanings of some traces


AbrilRBS (approved):
Today you mentioned that you changed this to only log downloaded files that exceed 1MB in size, but I don't see that change here. Did I misunderstand and it's just "future work"?

memsharded:
> Today you mentioned that you changed this to only log downloaded files that exceed 1MB in size, but I don't see that change here. Did I misunderstand and it's just "future work"?

Nop, I had to revert that change, too many unrelated tests breaking. I will remove the description.

czoido (approved):

PR #12829, opened 2023-01-03
Fixing cmake_layout single config

### LINKED ISSUE #12827: [bug] conan_toolchain.cmake is shared by both build_types
### Environment details

* Operating System+version: Linux 6.1.1-1-MANJARO
* Compiler+version: N/A
* Conan version: 1.56
* Python version: 3.10.8
* CMake 3.24.2

### Steps to reproduce

1. create a CMakeLists.txt with an option
2. enable that option based on build_type in the conanfile
3. run the conan install for both buildtypes `conan install ... && conan install ... -s build_type=Debug`
4. Spend half an hour wondering why the cmake expression generator which is assigned to that option keeps on having the same value.
5. Observe the CMakePresets.json and notice that the conan_toolchain.cmake is used by both ConfigurePresets.
6. Observe the value of the option specified in the CMakeLists.txt haveing that same value.

### Logs

_No response_

memsharded:
Changelog: Bugfix: Make ``cmake_layout`` to use single-config folder for generated files.
Docs: https://github.com/conan-io/docs/pull/2891

Close https://github.com/conan-io/conan/issues/12827

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -22,20 +22,21 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
         raise ConanException("'build_type' setting not defined, it is necessary for cmake_layout()")
 
     build_folder = build_folder if not subproject else os.path.join(subproject, build_folder)
-    custom_conf = get_build_folder_custom_vars(conanfile)
-    if custom_conf:
-        build_folder = os.path.join(build_folder, custom_conf)
+    config_build_folder, user_defined_build = get_build_folder_custom_vars(conanfile)
 
-    if multi:
-        conanfile.folders.build = build_folder
+    if config_build_folder:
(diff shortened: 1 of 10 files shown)


conan/tools/cmake/layout.py
```diff
@@ -49,17 +50,11 @@ def get_build_folder_custom_vars(conanfile):
                                     default=[], check_type=list)
     ret = []
     for s in build_vars:
+        group, var = s.split(".", 1)
```
SSE4, line 53: what if it doesn't have a dot, will it raise with a weird exception here?

memsharded: Yes, but that was already happening before, nothing new, and I didn't want to add extra noise and tests, as this PR is already quite potentially breaking.

czoido (approved):

jcar87 (approved):
LGTM!

PR #12830, opened 2023-01-03
Add optional requirements fields to all templates

AbrilRBS:
Changelog: (Feature): Add optional requirements fields to all templates
Docs: https://github.com/conan-io/docs/pull/2881

Just as for the basic template, this now lets you add multiple requirements to your templates if needed, such as:
`conan new cmake_exe -d name=game -d version=1.0 -d requires=maths/1.3 -d requires=ai/2.0`

### DIFF
--- a/conan/internal/api/new/autoools_exe.py
+++ b/conan/internal/api/new/autoools_exe.py
@@ -43,6 +43,13 @@ class {{package_name}}Conan(ConanFile):
     def package(self):
         autotools = Autotools(self)
         autotools.install()
+
+    {% if requires is defined -%}
+    def requirements(self):
+        {% for require in as_iterable(requires) -%}
+        self.requires("{{ require }}")
+        {% endfor %}
+    {%- endif %}
 """
 makefile_am_exe = """
 bin_PROGRAMS = {{name}}
--- a/conan/cli/commands/new.py
+++ b/conan/cli/commands/new.py
@@ -67,8 +67,9 @@ def new(conan_api, parser, *args):
                 ast = env.parse(templ_str)
                 template_vars.extend(meta.find_undeclared_variables(ast))
 
-            injected_vars = {"conan_version", "package_name"}
-            template_vars = list(set(template_vars) - injected_vars)
+            injected_vars = {"conan_version", "package_name", "as_iterable"}
+            optional_vars = {"requires"}
+            template_vars = list(set(template_vars) - injected_vars - optional_vars)
             template_vars.sort()
             return template_vars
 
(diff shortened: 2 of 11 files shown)


conan/internal/api/new/autoools_exe.py
```diff
@@ -43,6 +43,13 @@ def build(self):
     def package(self):
         autotools = Autotools(self)
         autotools.install()
+
+    {% if requires is defined -%}
```
memsharded, line 47: I'll try to go beyond this:
- Add ``find_package()`` or equivalent to build scripts
- Add a ``#include "<name>.h"`` to code
- Add a call ``<name>();`` to code

That is, make it a fully functional compiling and linking requires (that cannot be made automatic for libs like zlib, but it is a template, you should replace the call with ``zlib()`` to something that makes sense. 

Maybe not for all build systems, but maybe for CMake?

memsharded (commented):
Some test might be failing for changed hashes of contents (you might avoid it by avoiding extra blank lines in the template if requires not defined)

AbrilRBS:
Oops, I thought I had submitted this as a draft. I knew about the failing tests and wanted to fix it before un-drafting it, but I guess it's too late now for it now that you've reviewed it @memsharded 😄

conan/internal/api/new/autoools_exe.py
AbrilRBS: Nice idea

SSE4, line 47: the same code added to 10 different files... it doesn't scale and violate DRY - if you need to change it for some reason, you will need to change 10 files and keep all them in mind.

AbrilRBS: After looking into this just now, it's a bit less straightforward than I originally thought, let's do this in a different PR :)

AbrilRBS: You're right that repetition is always a smell, but in this case:
 -  Files are not modified nearly enough to warrant spending more time than the necessary on them
 - There's no good place for it :( Moving templating code to Python seems counterintuitive and Jinja has no easy "define a function to be used elsewhere" concept other than template inheritance, which seems excesive to save a copy-paste in these files

So I'd leave this here as-is. I'll be revisiting this as I still need to implement some more helpful stuff for Cmake in a later PR. Either way, thanks a lot for your review, I really appreciate you taking the time to doublecheck my code and suggest improvements :)

memsharded (approved):

memsharded (commented):
I am fine with merging this, and postponing the implementation of real include and linking and executing dependencies for a later PR, and maybe not necessary to do it at once for all templates, but one at a time also good (maybe it would have been better to introduce the new ``requires`` in templates one at a time while implementing the whole functionality, but I am fine with it too this way)

AbrilRBS:
After talking with @czoido and @memsharded, let's rethink how we approach this and merge something minimal, and then create a cmake template that works, and we can then add more interesting stuff such as the toy example suggested by Carlos

Issue #12533, opened 2022-11-14
[bug] conan info --package-filter ignored for json output?


foundry-markf:
When using `conan info --package-filter` on a package with dependencies, the filtering is ignore when written out to JSON, but not to the console.

I'm unsure of whether this is intentional, in that JSON you can do your own filtering, but there's no message from the Conan command to say this is the case, and the [docs](https://docs.conan.io/en/latest/reference/commands/consumer/info.html) don't hint at it. So this makes me think it's a bug.

### Environment Details (include every applicable attribute)
  * Operating System+version: mac Catalina (but have seen it across the desktop platforms)
  * Compiler+version: N/A
  * Conan version: 1.54.0
  * Python version: 3.10.8

### Steps to reproduce (Include if Applicable)
I used Conan Center recipes for easiest to reproduce.

#### Output to console working as expected
First, see the full output from a package with multiple dependencies:
```
(venv) FWMBP064:test mark.final$ conan info openexr/3.1.5@
```

<details>
  <summary>Command output</summary>
  
  ```
imath/3.1.5
    ID: 853c4b61e2571e98cd7b854c1cda6bc111b8b32c
    BuildID: None
    Context: host
    Remote: conancenter=https://center.conan.io
    URL: https://github.com/conan-io/conan-center-index
    Homepage: https://github.com/AcademySoftwareFoundation/Imath
    License: BSD-3-Clause
    Description: Imath is a C++ and python library of 2D and 3D vector, matrix, and math operations for computer graphics.
    Topics: computer-graphics, matrix, openexr, 3d-vector
    Provides: imath
    Recipe: Cache
    Binary: Missing
    Binary remote: conancenter
    Creation date: 2022-08-22 22:00:44 UTC
    Required by:
        openexr/3.1.5
openexr/3.1.5
    ID: 2962f428ff4f8584b8b012a084e242a3fab3919f
    BuildID: None
    Context: host
    Remote: conancenter=https://center.conan.io
    URL: https://github.com/conan-io/conan-center-index
    Homepage: https://github.com/AcademySoftwareFoundation/openexr
    License: BSD-3-Clause
    Description: OpenEXR is a high dynamic-range (HDR) image file format developed by Industrial Light & Magic for use in computer imaging applications.
    Topics: openexr, hdr, image, picture
    Provides: openexr
    Recipe: Cache
    Binary: Missing
    Binary remote: conancenter
    Creation date: 2022-10-31 21:45:36 UTC
    Requires:
[...]

PR #12836, opened 2023-01-04
Feature/develop2 info package filters

memsharded:
Changelog: Fix: Make ``graph info`` filters to work on json output too
Docs: Omit

Close https://github.com/conan-io/conan/issues/12533
There is still some larger work pending in this area, but too much for this PR:
- Improve the serialization of the graph, serializating better the depenencies (requires)
- Use the serialized graph instead of the native Conan object as input to .dot and .html graph renderings

### DIFF
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -5,6 +5,7 @@ from jinja2 import Template, select_autoescape
 
 
 from conan.api.output import cli_out_write
+from conan.cli.formatters.graph.graph_info_text import filter_graph
 from conan.cli.formatters.graph.info_graph_dot import graph_info_dot
 from conan.cli.formatters.graph.info_graph_html import graph_info_html
 from conans.client.graph.graph import BINARY_CACHE, \
@@ -90,7 +91,6 @@ class _Grapher(object):
 
 
 def _render_graph(graph, template, template_folder):
-    graph = _Grapher(graph)
     from conans import __version__ as client_version
     template = Template(template, autoescape=select_autoescape(['html', 'xml']))
     return template.render(graph=graph, base_template_path=template_folder, version=client_version)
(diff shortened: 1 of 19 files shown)


conan/cli/formatters/graph/graph.py
```diff
@@ -99,6 +100,14 @@ def _render_graph(graph, template, template_folder):
...
     graph = result["graph"]
     conan_api = result["conan_api"]
+    field_filter = result["field_filter"]
+    if field_filter and "requires" not in field_filter:
+        field_filter.append("requires")
+    package_filter = result["package_filter"]
+    serial = graph.serialize()
+    # TODO: This is not used, it is necessary to update the renderings to use the serialized graph
+    #  instead of the native graph
+    serial = filter_graph(serial, package_filter, field_filter)
```
AbrilRBS: Do we need to save the output of the function if it's not used elsewhere?

AbrilRBS (approved):
In the original issue (Or elsewhere) you mentioned that it made little sense to add filtering to outputs meant for further machine processing because you can always filter the resulting json by yourself. I'm guessing that by implementing this, we're walking back on that, right?

conan/cli/formatters/graph/graph.py
memsharded: It is temporary, with the ``TODO``. The idea is that this ``serial`` is the thing that should be fed to the rendering templates, and not the graph object.

memsharded:
> In the original issue (Or elsewhere) you mentioned that it made little sense to add filtering to outputs meant for further machine processing because you can always filter the resulting json by yourself. I'm guessing that by implementing this, we're walking back on that, right?

Yes, we were reconsidering this, and apparently it would match user expectations, if they provide both the json and a filter is because they want it filtered. It also makes sense to be able to trim the output html/dot graph to the packages that you want, so the filter seems also useful for the other formatters. That seems to mean that all the formatters should be consistent with the user input, filtering out whatever the user wants to filter out. Wdyt?

conan/cli/formatters/graph/graph.py
AbrilRBS: Ohh now the PR comment makes sense, I didn't quite understand what it was in reference to, ty for the clarification!

AbrilRBS:
Yes, that's how I would have guessed all formatters behaved had I not seen the issues related to this, so I think that this change is very useful for those cases :)

conan/cli/formatters/graph/graph.py
```diff
@@ -99,6 +100,14 @@ def _render_graph(graph, template, template_folder):
 def format_graph_html(result):
     graph = result["graph"]
     conan_api = result["conan_api"]
+    field_filter = result["field_filter"]
+    if field_filter and "requires" not in field_filter:
```
SSE4: should it check for `field_filter is not None` instead (as empty list may evaluate to False in python)?

czoido (approved):

conan/cli/formatters/graph/graph.py
memsharded: Removing this, it doesn't really apply

Issue #12816, opened 2022-12-31
[feature] Filter out `.DS_Store` on macOS in command `conan profile list`


SpaceIm:
### What is your suggestion?

If there is a `.DS_Store` file in profiles folder, it's listed as a profile by `conan profile list` command. It should not be displayed.

### Have you read the CONTRIBUTING guide?

- [x] I've read the CONTRIBUTING guide

AbrilRBS:
Thanks for your suggestion, you're right that having clear outputs and not lying to the user is a good idea fi we know something is not a profile. 

For anyone interested in contributing this feature, I'll lay down the necessary steps to implement it, but please feel free to ask any questions if you find any problems during the process:


1. Setup the developing environment. When forking the project, make sure to uncheck "Clone only main branch", as we'll need to checkout the `develop2` branch, and this is by far the easiest way. Once you've done so, docs for setting up the env can be found in https://github.com/conan-io/conan#install-a-development-version. Open the project in your IDE of preference, everything _should_ be automatically picked up.
2. The command we want to modify is `profile`. You'll find its CLI in `conan/cli/commands/profile.py`. Its implementation is super simple, as it relies in the public conan API implementation for profiles found in `conan/api/subapi/profiles.py`
3. The interesting `list` function just walks the path of the profiles and is where we'll want to implement the changes. Feel free to do so as you see fit, but think about longterm maintenance and how we might want to in the future add some more folders/files (Such as `.git` to give you an example) The next step gives you info about creating the tests if you want to first do so before implementing the functionality
4. Tests in conan are currently found in `conans/test/`, and for this you'll want to create a new integration test in the corresponding `conans/test/integration/commands/test_profile.py` file, in which you:
 - Instantiate a TestClient()
 - Add a profile to be filtered out to its cache folder. Some useful attributes and methods for this will be `TestClient::cache.profile_path` for the folder of this TestClient profile folder path in which to add the empty profile, and the `TestClient::run` method to invoke the `conan profile` command. Note that this command stores its output in `TestClient::out`, so you can assert for any kind of output over that attribute
5. Send us a PR, we'll be happy to review it and help you where needed

dead-tech:
I would like to give this a go 😄

memsharded:
Fixed by @dead-tech in https://github.com/conan-io/conan/pull/12845, for beta.8

PR #12845, opened 2023-01-04
Allow specifying files/directories to ignore when listing cached profiles.

dead-tech:
Implements feature requested in #12816.

This is currently set to ignore just '.DS_Store' which was requested in the issue. To extend this behaviour it should be enough to modify `paths_to_ignore` in conan/api/subapi/profiles.py, as well as implementing the tests accordingly.

Changelog: (Feature): Allow specifying files/directories to ignore when listing cached profiles.
Docs: Omit

### DIFF
--- a/conan/api/subapi/profiles.py
+++ b/conan/api/subapi/profiles.py
@@ -68,9 +68,7 @@ class ProfilesAPI:
         profiles = []
         profiles_path = self._cache.profiles_path
         if os.path.exists(profiles_path):
-            for current_directory, directories, files in os.walk(profiles_path, followlinks=True):
-                directories[:] = [dir for dir in directories if os.path.relpath(
-                    os.path.join(current_directory, dir)) not in paths_to_ignore]
+            for current_directory, _, files in os.walk(profiles_path, followlinks=True):
                 files = filter(lambda file: os.path.relpath(
                     os.path.join(current_directory, file), profiles_path) not in paths_to_ignore, files)
 
--- a/conan/api/subapi/profiles.py
+++ b/conan/api/subapi/profiles.py
@@ -62,10 +62,18 @@ class ProfilesAPI:
         List all the profiles file sin the cache
         :return: an alphabetically ordered list of profile files in the default cache location
         """
+        # List is to be extended (directories should not have a trailing slash)
+        paths_to_ignore = ['.DS_Store']
+
         profiles = []
         profiles_path = self._cache.profiles_path
         if os.path.exists(profiles_path):
-            for current_directory, _, files in os.walk(profiles_path, followlinks=True):
+            for current_directory, directories, files in os.walk(profiles_path, followlinks=True):
+                directories[:] = [dir for dir in directories if os.path.relpath(
+                    os.path.join(current_directory, dir)) not in paths_to_ignore]
+                files = list(filter(lambda file: os.path.relpath(
+                    os.path.join(current_directory, file), profiles_path) not in paths_to_ignore, files))
+
                 for filename in files:
                     rel_path = os.path.relpath(os.path.join(current_directory, filename),
                                                profiles_path)
--- a/conan/api/subapi/profiles.py
+++ b/conan/api/subapi/profiles.py
@@ -71,8 +71,8 @@ class ProfilesAPI:
             for current_directory, directories, files in os.walk(profiles_path, followlinks=True):
                 directories[:] = [dir for dir in directories if os.path.relpath(
                     os.path.join(current_directory, dir)) not in paths_to_ignore]
-                files = list(filter(lambda file: os.path.relpath(
-                    os.path.join(current_directory, file), profiles_path) not in paths_to_ignore, files))
+                files = filter(lambda file: os.path.relpath(
+                    os.path.join(current_directory, file), profiles_path) not in paths_to_ignore, files)
 
                 for filename in files:
                     rel_path = os.path.relpath(os.path.join(current_directory, filename),
--- a/conans/test/integration/command/test_profile.py
+++ b/conans/test/integration/command/test_profile.py
@@ -11,21 +11,14 @@ def test_profile_path():
 
 def test_ignore_paths_when_listing_profiles():
     c = TestClient()
-    ignore_paths = ['.DS_Store']
+    ignore_path = '.DS_Store'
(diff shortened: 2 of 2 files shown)


conans/test/integration/command/test_profile.py
```diff
@@ -1,7 +1,31 @@
...
+            os.makedirs(base_dir, exist_ok=True)
+            open(os.path.join(base_dir, os.path.basename(path)), 'w').close()
+        else:
+            open(rel_file_path, 'w').close()
+
+    c.run("profile list")
+
+    for path in ignore_paths:
+        rel_file_path = os.path.relpath(os.path.join(c.cache.profiles_path, path))
+        assert rel_file_path not in c.out
```
memsharded:
```suggestion
    for path in ignore_paths:
        assert path not in c.out

```
It is easier and also more robust. Negative tests like ``not in`` can easily be a false negative, for example if you make some mistake in the path composition, or the path have different backslash, or something, it will not be found in the output, while you really don't want to see ``.DS_Store`` at all in the output.

conans/test/integration/command/test_profile.py
```diff
@@ -1,7 +1,31 @@
...
 
 def test_profile_path():
     c = TestClient()
     c.run("profile path default")
     assert "default" in c.out
+
+
+def test_ignore_paths_when_listing_profiles():
+    c = TestClient()
+    ignore_paths = ['.DS_Store']
```
memsharded: It is perfectly fine to start with test as simple as possible, like managing only a single ``.DS_Store``, and not needing a ``for``

conans/test/integration/command/test_profile.py
```diff
@@ -1,7 +1,31 @@
...
+    # Create files, dirs if necessary, at the ignore paths
+    for path in ignore_paths:
+        rel_file_path = os.path.join(c.cache.profiles_path, path)
+
+        base_dir = os.path.join(c.cache.profiles_path, os.path.dirname(path))
+        if base_dir:
+            os.makedirs(base_dir, exist_ok=True)
+            open(os.path.join(base_dir, os.path.basename(path)), 'w').close()
+        else:
+            open(rel_file_path, 'w').close()
```
memsharded: I'd say this is a bit overly complex. Something like:
```python
os.makedirs(c.cache.profiles_path, exist_ok=True)  # just in case
os.utime(os.path.join(c.cache.profiles_path, ".DS_Store"))  # This is a "touch"
```
should be good enough

conan/api/subapi/profiles.py
```diff
@@ -62,10 +62,18 @@ def list(self):
...
+        # List is to be extended (directories should not have a trailing slash)
+        paths_to_ignore = ['.DS_Store']
+
         profiles = []
         profiles_path = self._cache.profiles_path
         if os.path.exists(profiles_path):
-            for current_directory, _, files in os.walk(profiles_path, followlinks=True):
+            for current_directory, directories, files in os.walk(profiles_path, followlinks=True):
+                directories[:] = [dir for dir in directories if os.path.relpath(
+                    os.path.join(current_directory, dir)) not in paths_to_ignore]
```
memsharded: I think this might not be necessary, a bit speculative of something that could happen in the future, but is not happening yet, as the ``.DS_store`` is always a file, never a folder. It is better in general to do the minimal changes to the code, and specially not add things that are actually not covered by tests.

memsharded (commented):
Thanks very much @dead-tech for your contribution!
I think it is looking good in general. I have added a few comments with feedback about some minor design and test issues, but nothing critical there.

conans/test/integration/command/test_profile.py
dead-tech: This still needs `open(os.path.join(c.cache.profiles_path, '.DS_Store')).close()` to work I think. As far as I am aware `os.utime` just updates the last modified time of the file.

conan/api/subapi/profiles.py
dead-tech: My bad I think I might overengineered it by stressing too much on the long-term scalability @RubenRBS was mentioning in the issue.

conans/test/integration/command/test_profile.py
memsharded: Yes, you are right, I thought that ``utime`` was equivalent to ``touch`` and create the file if not existing.
Yes, the ``open()`` might be necessary then. There is some ``save(path, content)`` helper function too, but the open() is fine.

dead-tech:
Thank you so much for the quick review 😊 @memsharded

conan/api/subapi/profiles.py
AbrilRBS: No worries :)

memsharded: Yeah, don't worry, I was giving some feedback based on my own experience, but those things are completely subject to opinions and discussion. This is why I was saying that the PR is good, and my feedback was minor issues. Thanks!

conan/api/subapi/profiles.py
```diff
@@ -62,10 +62,16 @@ def list(self):
...
         :return: an alphabetically ordered list of profile files in the default cache location
         """
+        # List is to be extended (directories should not have a trailing slash)
+        paths_to_ignore = ['.DS_Store']
+
         profiles = []
         profiles_path = self._cache.profiles_path
         if os.path.exists(profiles_path):
             for current_directory, _, files in os.walk(profiles_path, followlinks=True):
+                files = filter(lambda file: os.path.relpath(
```
AbrilRBS, line 72: Nice idea using a relpath here! I wouldn't have thought of that :)

AbrilRBS (approved):
Looks great, thank you very much for taking the time to implement this feature. It should be released in 1.57 in the near future :)

conan/api/subapi/profiles.py
memsharded: this is still a bit especulative, until ``paths_to_ignore`` really contains paths, it is possible to keep this even simpler with ``files = [f for f in files if f not in paths_to_ignore]`` or even doing a ``if filename in paths_to_ignore: continue`` below.

In any case, this is a very minor issue, so everything is looking good, thanks very much for your contribution, lets merge it for next release.

memsharded (approved):

PR #12846, opened 2023-01-04
Also write CMake Test Presets

gjasny:
Changelog: Feature: CMakeToolchain also write CMake Test Presets.
Docs: https://github.com/conan-io/docs/pull/2889

Right now the `CMakeToolchain` generator only writes configure and build presets. That requires some guesswork to find the build directory in case the `cmake_layout` is used.

My testcase:
```python
from conan import ConanFile
from conan.tools.cmake import cmake_layout

class Pkg(ConanFile):
    generators = "CMakeToolchain"

    def layout(self):
    	cmake_layout(self)
```

```cmake
project(testcase NONE)
enable_testing()
add_test(NAME some_test COMMAND ${CMAKE_COMMAND} -E true)
```

```shell
conan install . 
cmake --preset release
cmake --build --preset release
ctest --preset release # this fails right now but succeeds with this PR
```

If you welcome adding test presets I'll extend add the necessary docs repo PR.

Thanks,
Gregor

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -10,16 +10,24 @@ from conans.util.files import save, load
 
 
 def _build_preset(conanfile, multiconfig):
+    return _common_build_and_test_preset_fields(conanfile, multiconfig)
+
+
+def _test_preset(conanfile, multiconfig):
+    return _common_build_and_test_preset_fields(conanfile, multiconfig)
+
+
+def _common_build_and_test_preset_fields(conanfile, multiconfig):
     build_type = conanfile.settings.get_safe("build_type")
     configure_preset_name = _configure_preset_name(conanfile, multiconfig)
-    ret = {"name": _build_preset_name(conanfile),
+    ret = {"name": _build_and_test_preset_name(conanfile),
            "configurePreset": configure_preset_name}
     if multiconfig:
         ret["configuration"] = build_type
     return ret
 
 
-def _build_preset_name(conanfile):
+def _build_and_test_preset_name(conanfile):
     build_type = conanfile.settings.get_safe("build_type")
     custom_conf = get_build_folder_custom_vars(conanfile)
     if custom_conf:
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/presets.py
```diff
@@ -30,6 +30,10 @@ def _build_preset_name(conanfile):
     return build_type.lower() if build_type else "default"
 
 
+def _test_preset(conanfile, multiconfig):
```
memsharded: So the test preset is basically identical to the build preset, just different name? Maybe would it be possible to do an extension/inclusion of the build one, instead of a repetition (talking from ignorance, I know little about test presets)

memsharded (commented):
Looking good in general, I think it might be possible to add it, I don't see any important blocker or risk. Thanks very much for contributing to Conan!

conan/tools/cmake/presets.py
gjasny: The [build and test presets](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#test-preset) share some common fields like `name` and `configurePreset`. But both presets also allow some distinct features. I'll try to do a sensible renaming of the helper functions.

memsharded (approved):
Looking good, thanks very much!

gjasny:
I assume the changelog in the docs repo gets automatically populated from the Changelog line. I did not add any functionality that's worth updating or adding the docs repo.

memsharded:
> I assume the changelog in the docs repo gets automatically populated from the Changelog line. I did not add any functionality that's worth updating or adding the docs repo.

The changelog is automatic from the first line. 
Probably this change does deserve another PR to the docs repo to introduce the test presets. It could be just some lines summarizing the feature, basically the lines that you added to the tests calling ``ctest --preset ....`` and some snippet of the generated presets file. Don't worry if you cannot do it, we might be able to contribute it ourselves.

czoido (approved):

gjasny:
Thanks for reviewing. I'll add the documentation by the end of today.

PR #12873, opened 2023-01-09
add `conan.tools.android.android_abi()`

### LINKED ISSUE #12814: [feature] Port conans.tools.to_android_abi to conan v2 tools
### What is your suggestion?

This function is used in opencv 4.x recipe of conan-center. It's the only legacy function remaining in https://github.com/conan-io/conan-center-index/pull/14947

SpaceIm:
Changelog: Feature: Add `conan.tools.android.android_abi()` function to return the Android standard ABI name based on Conan.
Docs: https://github.com/conan-io/docs/pull/2975

closes https://github.com/conan-io/conan/issues/12814

### DIFF
--- a/conan/tools/android/utils.py
+++ b/conan/tools/android/utils.py
@@ -1,0 +1,25 @@
+from conan.errors import ConanException
+
+
+def android_abi(conanfile, context="host"):
+    """
+    Returns Android-NDK ABI
+    :param conanfile: ConanFile instance
+    :param context: either "host", "build" or "target"
+    :return: Android-NDK ABI
+    """
+    if context not in ["host", "build", "target"]:
+        raise ConanException("context argument must be either 'host', 'build' or 'target'")
+
+    settings = getattr(conanfile, f"settings_{context}", conanfile.settings)
+    arch = settings.get_safe("arch")
+    # https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html
+    return {
+        "armv5el": "armeabi",
+        "armv5hf": "armeabi",
+        "armv5": "armeabi",
+        "armv6": "armeabi-v6",
+        "armv7": "armeabi-v7a",
+        "armv7hf": "armeabi-v7a",
+        "armv8": "arm64-v8a",
+    }.get(arch, arch)
--- a/conan/tools/android/utils.py
+++ b/conan/tools/android/utils.py
@@ -9,9 +9,15 @@ def android_abi(conanfile, context="host"):
     :return: Android-NDK ABI
     """
     if context not in ["host", "build", "target"]:
-        raise ConanException("context argument must be either 'host', 'build' or 'target'")
+        raise ConanException(f"context argument must be either 'host', 'build' or 'target', was {context}")
 
-    settings = getattr(conanfile, f"settings_{context}", conanfile.settings)
+    try:
+        settings = getattr(conanfile, f"settings_{context}")
+    except AttributeError:
+        if context == "host":
+            settings = conanfile.settings
+        else:
+            raise ConanException(f"settings_{context} not declared in recipe")
     arch = settings.get_safe("arch")
     # https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html
     return {
--- a/conan/tools/android/utils.py
+++ b/conan/tools/android/utils.py
@@ -8,8 +8,8 @@ def android_abi(conanfile, context="host"):
     :param context: either "host", "build" or "target"
     :return: Android-NDK ABI
     """
-    if context not in ["host", "build", "target"]:
-        raise ConanException(f"context argument must be either 'host', 'build' or 'target', was {context}")
+    if context not in ("host", "build", "target"):
+        raise ConanException(f"context argument must be either 'host', 'build' or 'target', was '{context}'")
 
     try:
         settings = getattr(conanfile, f"settings_{context}")
--- a/conans/test/unittests/tools/android/test_android_tools.py
+++ b/conans/test/unittests/tools/android/test_android_tools.py
@@ -1,0 +1,36 @@
+from conans.test.utils.mocks import ConanFileMock, MockSettings
+from conan.tools.android import android_abi
+
+
+def test_tools_android_abi():
+    settings_linux = MockSettings({"os": "Linux", "arch": "x86_64"})
+
+    for (arch, expected) in [
+        ("armv5el", "armeabi"),
+        ("armv5hf", "armeabi"),
+        ("armv5", "armeabi"),
+        ("armv6", "armeabi-v6"),
+        ("armv7", "armeabi-v7a"),
+        ("armv7hf", "armeabi-v7a"),
+        ("armv8", "arm64-v8a"),
+        ("x86", "x86"),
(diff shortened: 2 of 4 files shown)


conan/tools/android/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    arch = settings.get_safe("arch")
+    # https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html
+    return {
+        "armv5el": "armeabi",
+        "armv5hf": "armeabi",
+        "armv5": "armeabi",
+        "armv6": "armeabi-v6",
+        "armv7": "armeabi-v7a",
+        "armv7hf": "armeabi-v7a",
+        "armv8": "arm64-v8a",
```
SSE4, line 30: we're not handling x86 and MIPS variants here. at least x86 is popular for running Simulator builds (plus, there are real x86 boards with Android). MIPS was used by previous (old) NDK releases, but I would add for completeness.
(also I've heard Android is adding RISC-V now, but I don't know ABIs right now)

SpaceIm: arch in profile and ndk abi are identical for x86 and mips.

SpaceIm: it should be tested in unit test

SSE4: oh, good catch! sorry for the noise :)

SpaceIm: No it's fine, I've extended archs tested in the unit test.

SSE4 (approved):

AbrilRBS:
I think this would be cool to have added in the docs :)

SpaceIm:
Sure, but when I know it has a chance to be merged. I don't want to work on documentation if this feature is rejected.

conan/tools/android/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Returns Android-NDK ABI
+    :param conanfile: ConanFile instance
+    :param context: either "host", "build" or "target"
+    :return: Android-NDK ABI
+    """
+    if context not in ["host", "build", "target"]:
+        raise ConanException("context argument must be either 'host', 'build' or 'target'")
+
+    settings = getattr(conanfile, f"settings_{context}", conanfile.settings)
```
memsharded: If we are providing the context, then falling back to the host settings doesn't look great, it should be explicit.
I ask you for the "build" context, you return the build context or fail. If this is being used in a cross-build scenario with 1 profile only, then it is allowed (should) fail.

memsharded (commented):
I think this is looking good and could be merged, but I'd prefer to avoid a silent fallback to another context, better raise if profile build or target not correctly defined.

conans/test/unittests/tools/android/test_android_tools.py
```diff
@@ -0,0 +1,68 @@
+from conans.test.utils.mocks import ConanFileMock, MockSettings
+from conan.tools.android import android_abi
+
+
+def test_tools_android_abi():
+    settings_linux = MockSettings({"os": "Linux", "arch": "foo"})
+
+    for (arch, expected) in [
```
AbrilRBS, line 10: This is usually better to be done with a `@parametrized` decorator in the function itself, so failures are easier to debug, but I'm ok with this either way :)

conan/tools/android/utils.py
AbrilRBS: I agree with @memsharded, this is something that should not fall back if missing. What usecase did you have in mind for allowing the fallback?

conan/tools/android/utils.py
```diff
@@ -0,0 +1,25 @@
...
+    # https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html
+    return {
+        "armv5el": "armeabi",
+        "armv5hf": "armeabi",
+        "armv5": "armeabi",
+        "armv6": "armeabi-v6",
+        "armv7": "armeabi-v7a",
+        "armv7hf": "armeabi-v7a",
+        "armv8": "arm64-v8a",
+    }.get(arch, arch)
```
AbrilRBS, line 31: In the old code, `android_abi` would have been None for missing archs. The second parameter here changes that bahaviour, which might be more correct (No idea), but we should make sure it does not have any unintended side-effects

AbrilRBS (commented):
Thanks for taking the time to implement this. I only have a few minor comments before we can merge :)

conans/test/unittests/tools/android/test_android_tools.py
SpaceIm: I remember to have written tests with @pytest.mark.parametrize, then move to this implementation but I don't remember why. I can change back to parametrize.

conan/tools/android/utils.py
SpaceIm: Well there are 2 old codes, inconsistent:
- legacy `toos.to_android_abi()` => it fallbacks to `str(conanfile.settings.arch)`
- CMakeToolchain => fallbacks to None

So a decision has to be made for this new one, as you wish, I've not a strong opinion on this.

conan/tools/android/utils.py
SpaceIm: Yes I had 1 profile in mind, to not raise even if it makes little sense to call this function with 1 profile. Also I didn't want to be too strict with settings_target since it depends on context in which a recipe is called, I'm not sure of the consequences in recipes.

So what would you want? Raise `ConanException` if `settings_{context}` is not an attribute of conanfile? Even for settings_host?

conans/test/unittests/tools/android/test_android_tools.py
memsharded: No big deal, if it works with parametrize easily, then do it, but don't waste time if you hit some issue.

conan/tools/android/utils.py
memsharded: I think falling back to the arch input instead of ``None`` might be better, for cases where the users extend ``settings.yml`` with their own architectures, if they map directly to android ones, then this will work. So I am fine with this as-is.

conan/tools/android/utils.py
memsharded: ``settings_host`` obviously doesn't exist, so it would be necessary to do the right thing, and do an ``if context == "host"`` or a dict mapping or something like that. But yes, it is better to raise if not defined, and this function should only be used with 2 profiles. It is typically better to be initially restrictive and raise, then letting some behavior slip, and then users depending on this behavior (for 1 profile) will complain later if this breaks for some reason that we need to force exclusively for 2 profiles (already happened a few times)

AbrilRBS (approved):
Thanks! (Sorry about the delay!)
Please don't forget about adding docs for this, will merge once linked so it's easier to write the changelog!

memsharded:
We are releasing 1.58 asap, this would be missing ``settings`` raising when not defined, moving to 1.59

memsharded (commented):
Apparently this is not that needed in ConanCenter, but I am fine with moving this forward for 1.59, if the settings fallback to host is removed.

memsharded (approved):

SpaceIm:
Thanks for the final polishing. It will be useful in android-ndk recipe (but not anymore in opencv recipe, I have a PR removing its usage).

PR #12879, opened 2023-01-10
Add throw flag for `check_min_vs` tool function

AbrilRBS:
Changelog: Feature: Adds new `raise_invalid` argument for `check_min_vs` to avoid raising if check fails.
Docs: It will automatically be added to the docs

Simplifies instances where a `try/catch` would be needed for cases where custom logic is wanted  (in CCI, this is usually adding `-FS` where necessary)


franramirez688 (approved):

AbrilRBS:
Closing in favour of #12880

PR #12880, opened 2023-01-10
Add throw flag for `check_min_vs` tool function

AbrilRBS:
Changelog: Feature: Adds new `raise_invalid` argument for `check_min_vs` to avoid raising if the check fails.
Docs: https://github.com/conan-io/docs/pull/2890

Simplifies instances where a `try/catch` would be needed for cases where custom logic is wanted  (in CCI, this is usually adding `-FS` where necessary)


memsharded (approved):

jcar87 (approved):
LGTM!

franramirez688 (approved):

PR #12882, opened 2023-01-10
allow skip the generation of conanvcvars.bat

### LINKED ISSUE #12855: [question] CMakeToolchain requires VCVars
### What is your question?

We distribute build tools (including msvc, and the SDKs),  internally in our company via `build_requires` in our build profiles

For example: 

```
[settings]
os=Windows
arch=x86_64
compiler=msvc
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=192
compiler.update=8
compiler.cppstd=14
build_type=Release

[options]
[build_requires]
cmake/3.24.2
ninja/1.11.1
vc/19.28@internal/dk
ms_sdk/21322@internal/dk
ms_wdk/21322@internal/dk

[env]
CMAKE_BUILD_TYPE=Release
CMAKE_GENERATOR=Ninja

```
the `vc`, `ms_sdk`, and `ms_wdk` recipes were created by us.  They package up the build tool assets (binaries, headers, resource files, etc...), sets environment variables similar to what `vcvarsall.bat` does.  

This allows us to quickly change compilers with a simple build profile change.

With `CMakeToolChain`,  it is expecting to run vcvars for the `msvc` compiler,  which typically doesn't exist in our environments.

Before I go too deep into describing our situation,  is there a proper way to support msvc build tools via build requires?

memsharded:
Changelog: Feature: Allow skip the generation of conanvcvars.bat with empty ``tools.microsoft.msbuild:installation_path``
Docs: https://github.com/conan-io/docs/pull/2892

I am proposing the use of ``tools.microsoft.msbuild:installation_path='`` to skip the generation of the conanvcvars.bat. I am reusing the ``tools.microsoft.msbuild:installation_path`` conf, because it is used exclusively already for the ``VCVars`` functionality, and to avoid the proliferation of tons of ``confs`` for minor functionalities like this one.

Close https://github.com/conan-io/conan/issues/12855

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -73,7 +73,7 @@ class VCVars:
             return
 
         vs_install_path = conanfile.conf.get("tools.microsoft.msbuild:installation_path")
-        if vs_install_path == "disabled":
+        if vs_install_path == "":  # Empty string means "disable"
             return
 
         if compiler == "clang":
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -72,6 +72,10 @@ class VCVars:
         if compiler not in ("Visual Studio", "msvc", "clang"):
             return
 
+        vs_install_path = conanfile.conf.get("tools.microsoft.msbuild:installation_path")
+        if vs_install_path == "disabled":
+            return
+
         if compiler == "clang":
             # The vcvars only needed for LLVM/Clang and VS ClangCL, who define runtime
             if not conanfile.settings.get_safe("compiler.runtime"):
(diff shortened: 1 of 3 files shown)


AbrilRBS (approved):
Thanks for the explanation on ehy you're reusing `tools.microsoft.msbuild:installation_path` Is there prior art to special-casing to `disabled` in order to turn off some functionality in the codebase?
Also, did you consider disabling this by setting the conf to an empty string? (Just as #12868 proves it to be possible) And why did you decide on the `disabled` way? Less explicit?

franramirez688 (approved):

czoido:
I think it's good to reuse the conf for this but I was wondering if we could use an empty path value as the indication to skip the file generation, like:

```
tools.microsoft.msbuild:installation_path=""
```

I'm saying that because if we reuse more confs to do a similar thing it's easier to know that you have to set an empty value than using a keyword like "disabled"

WDYT?

memsharded:
> Also, did you consider disabling this by setting the conf to an empty string? (Just as https://github.com/conan-io/conan/pull/12868 proves it to be possible) And why did you decide on the disabled way? Less explicit?

yes, I was thinking of making it more explicit, the empty string meaning seems a bit less clear to me.
Some similar "disabling" syntax in Conan:
``--test-folder=None`` reads from cli ``"None"`` as a string, and disables ``test_package``. It has always read dirty to me, ``--test-folder=disabled`` would be a bit nicer (the likelihood of having a subfolder called ``"disabled"`` is a bit higher than ``"None"``, but should be relatively low. In theory, the more correct approach is a new arg ``--disable-test-folder``, I guess this was done to avoid the explosion of command cli args, that are already too many.

Issue #12804, opened 2022-12-29
[bug] [develop2] ``conan list`` output different


memsharded:
### Environment details

Latest ``develop2`` branch

### Steps to reproduce

The result from ``conan list`` is different if the command is local cache or remote

```
conan list name/version@#ae217408f5df09637e4b1bdbe2795f63:95e963b230e58ec1da5084c3e7261e46329279ca#* -f json --cache

{
    "Local Cache": {
        "app/version#4c07c7e7d383293da58cec0b5a48299d": {
            "app/version#4c07c7e7d383293da58cec0b5a48299d:4738c78949301e2e09ba11870074f07541f0eb06#771292fe732e4ef1ae9dccd8099f1be4": {}
        }
    }
}

------------------------------------------------------------------

conan list name/version@#ae217408f5df09637e4b1bdbe2795f63:95e963b230e58ec1da5084c3e7261e46329279ca#* -f json --remote conanRemote

{
    "conanRemote": {
        "revisions": [
            "name/version#rrev:pkgid#prev1%0.0",
            "name/version#rrev:pkgid#prev2%0.0"
        ]
    }
}
```

czoido:
Closed by https://github.com/conan-io/conan/pull/12883

PR #12883, opened 2023-01-10
[develop2] New serialization json and printing for ``conan list``

memsharded:
Changelog: Feature: New serialization json and printing for ``conan list``.

Close https://github.com/conan-io/conan/issues/12804


conans/test/integration/command_v2/list_test.py
```diff
@@ -2,600 +2,547 @@
...
+        expected = textwrap.dedent("""\
+          conf
+            conf/1.0
+              revisions
+                e4e1703f72ed07c15d73a555ec3a2fa1 (10-11-2023 10:13:13)
+                  packages
+                    78c6fa29e8164ce399087ad6067c8f9e2f1c4ad0
+                      info
+                        conf
+                          tools.build:cxxflags: ['--flag1']
```
franramirez688, line 313: IMO, I know that this output will have colors, but without `:` or even `[]` to distinguish what are the sections it's a bit difficult for me to read all the information.

conan/api/model.py
```diff
@@ -81,26 +83,31 @@ def prefs(self):
...
+                rev_dict = revisions_dict.setdefault(ref.revision, {})
+                if ref.timestamp:
+                    rev_dict["timestamp"] = timestamp_to_str(ref.timestamp)
+                if prefs:
+                    packages_dict = rev_dict.setdefault("packages", {})
+                    for pref_info in prefs:
+                        pref, info = pref_info
+                        pid_dict = packages_dict.setdefault(pref.package_id, {})
+                        if info is not None:
+                            pid_dict["info"] = info
```
franramirez688, line 105: Why could `info` not be an empty `dict` by default?

I saw this line above:

```python
binary_info = configurations.get(pref) if configurations is not None else None
```
So, I don't know if I'm missing anything else about any caveats.

franramirez688 (approved):
Overall, I think it looks really good! Great job!

conan/api/model.py
memsharded: I am trying to distinguish the case where the ``info`` is retrieved, even when it is empty, from the cases where the ``info`` is not retrieved at all (like only listing ``prevs`` of a known package_id). ``info=None`` means it is not retrieved at all, while ``info={}`` happens for example for header-only packages which information has been retrieved.

conans/test/integration/command_v2/list_test.py
memsharded: Yes, I agree. We can format this with whatever syntax we want, this is a proposal for the "structure", I think if we agree on it, we can polish the output format details very easily, lets talk about what we want and I'll do the changes in a following PR.

conan/api/model.py
franramirez688: That's what I wanted to know. Thanks for the clarification! 😉

AbrilRBS (approved):

jcar87 (approved):

czoido (approved):

PR #12889, opened 2023-01-11
[AutotoolsToolchain] Improve `update_xxxxx_args` behavior

franramirez688:
Changelog: Feature: AutotoolsToolchain helper functions: `update_configure_args`, `update_make_args`, and `update_autoreconf_args` can also add new values
Docs: https://github.com/conan-io/docs/pull/2895

Summary: before this change, it was only possible to update/remove existing values. Now, you could add new ones. Trying to avoid weird situations like:

```python
        tc = AutotoolsToolchain(self)
        tc.configure_args.extend([
            "--datarootdir=${prefix}/lib", # do not use share
            "--disable-layoutex",
            "--disable-layout"])
        tc.update_configure_args({"--force": None})
```
After my change, then it could be reduced to:
```python
        tc = AutotoolsToolchain(self)
        tc.update_configure_args({
            "--force": None,
            "--datarootdir": "${prefix}/lib", # do not use share
            "--disable-layoutex": "",
            "--disable-layout": ""})
```

Both are working, but I think the latest one could be the expected behavior. WDYT?

### DIFF
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -237,15 +237,7 @@ class AutotoolsToolchain:
             return ret
 
         def _dict_to_list(flags):
-            ret = []
-            for k, v in flags.items():
-                # None values are pruned from the global result
-                if v is None:
-                    continue
-                # If value, it's assumed the flag will be flag=value, else keeping flag only
-                # For instance: {"--opt1": "whatever", "-abc": ""} --> ["--opt1=whatever", "-abc"]
-                ret.append(f"{k}={v}" if v else k)
-            return ret
+            return [f"{k}={v}" if v else k for k, v in flags.items() if v is not None]
 
         self_args = getattr(self, attr_name)
         # FIXME: if xxxxx_args -> dict-type at some point, all these lines could be removed
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -224,23 +224,36 @@ class AutotoolsToolchain:
 
     # FIXME: Remove all these update_xxxx whenever xxxx_args are dicts or new ones replace them
     def _update_flags(self, attr_name, updated_flags):
-        _new_flags = []
-        self_args = getattr(self, attr_name)
-        for index, flag in enumerate(self_args):
-            flag_name = flag.split("=")[0]
-            if flag_name in updated_flags:
-                new_flag_value = updated_flags[flag_name]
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):

czoido (approved):

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -224,23 +224,36 @@ def update_autoreconf_args(self, updated_flags):
...
+        def _dict_to_list(flags):
+            ret = []
+            for k, v in flags.items():
+                # None values are pruned from the global result
+                if v is None:
+                    continue
+                # If value, it's assumed the flag will be flag=value, else keeping flag only
+                # For instance: {"--opt1": "whatever", "-abc": ""} --> ["--opt1=whatever", "-abc"]
+                ret.append(f"{k}={v}" if v else k)
+            return ret
```
memsharded:
```suggestion
            return [f"{k}={v}" if v else k for k, v in flags.items() if v is not None]
```
And as it is only used once, just do not add a function, but have it inline (the less jumps, the easier to read)

conan/tools/gnu/autotoolstoolchain.py
```diff
@@ -224,23 +224,36 @@ def update_autoreconf_args(self, updated_flags):
...
-                if new_flag_value is None:
-                    continue
-                elif not new_flag_value:
-                    _new_flags.append(flag_name)
+
+        def _list_to_dict(flags):
+            ret = {}
+            for flag in flags:
+                # Only splitting if "=" is there
+                option = flag.split("=")
```
memsharded:
```suggestion
                option = flag.split("=", 1)
```
Avoid potential strings with the "=" inside (unusual, but who knows)

memsharded (approved):

conan/tools/gnu/autotoolstoolchain.py
franramirez688: Sometimes, I prefer to read several lines instead a quite-compressed one, but, yes, in this case, I believe it's easy to read.

memsharded: Depends on the complexity, the semantics, the number of re-uses...
Typical things I check:
- If it is a oneliner or two liner
- If is using typical list comprehension patterns
- If the intention can be summarized in one line comment.
In this case, something like this reads better than having to jump to a function and read many lines:
```python
... # flags computed here
# Flatten flags dict to a list of ``k=v`` or ``k`` removing None values
my_list = [f"{k}={v}" if v else k for k, v in flags.items() if v is not None]
# use my list
```

Issue #6370, opened 2020-01-16
Use 'PAX_FORMAT' in TarFile.taropen


jgsogo:
Changes in Python 3.8 forced us to set explicitly the tar format in https://github.com/conan-io/conan/pull/6355, for Conan v2.0 it would be nice to change the format (**it needs to be explicit so it behaves the same for all python versions**) to the recommended one (`PAX_FORMAT`) for new Python versions (https://github.com/python/cpython/pull/12355)

memsharded:
Closed by https://github.com/conan-io/conan/pull/12899 for 2.0-beta.9

PR #12899, opened 2023-01-13
[develop2] change default tar format

memsharded:
Changelog: Fix: Change tar format to PAX, which is the Python3.8 default.
Docs: Omit

I have been reading about this decision in Python threads, the rational is that PAX is more modern, powerful and less vendor specific than GNU previous format. PAX_FORMAT is already available in Python 3.6, so it shouldn't be an issue. 
This also means that we are changing how we compress all our tgzs, we should carefully consider possible implications: Could the new tgz break Conan 1.X clients using the same ``conan_export_sources.tgz`` created with 2.0?

Close https://github.com/conan-io/conan/issues/6370


AbrilRBS (approved):

franramirez688 (approved):

AbrilRBS:
As for your question, this seems to work ok
Screenshot 2023-01-16 at 22 33 06

czoido (approved):

Issue #8443, opened 2021-02-04
[bug] Cross compiling with dual profile breaks with fPIC


anton-danielsson:
I'm unable to use almost any package when cross compiling from Windows to Android.

It seems like the
```py 
   def config_options(self):
        if self.settings.os == "Windows":
            del self.options.fPIC
```
pattern that is used in almost all recipes breaks when I try use the new `--profile:host/--profile:build` logic to cross compile on Windows...

As a work around I tried to replace the `config_options` logic with:
```py
    def package_id(self):
        if self.settings.os == "Windows":
            del self.info.options.fPIC
```
and it seems to work.
It feels like a hack though.

I first create an issue in CCI (https://github.com/conan-io/conan-center-index/issues/4483) but I think it actually belongs here.

### Environment Details (include every applicable attribute)
  * Operating System+version: **Windows 10**
  * Compiler+version: **Visual Studio 16/Android Clang 11**
  * Conan version: **conan 1.33.1**
  * Python version: **Python 3.7.8**

### Conan profile
```
Configuration (profile_host):
[settings]
arch=armv8
build_type=Release
compiler=clang
compiler.libcxx=libc++
compiler.version=11
os=Android
os.api_level=22
[options]
[build_requires]
*: android-ndk/r22
[env]

Configuration (profile_build):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=16
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
```

### Steps to reproduce (Include if Applicable)

Minimal example:
```py
from conans import ConanFile

class fooConan(ConanFile):
    name = "foo"
    generators = "cmake", "cmake_find_package_multi"

    settings = "os", "arch", "compiler", "build_type"

    requires = (
        "protobuf/3.13.0", # to get Android libprotobuf
    )

    build_requires = (
        "protobuf/3.13.0", # to get Windows proto.exe
    )
```

run:
```bash
conan create . version@dirac/testing \
    --profile:host android-ndk-r20-api-21-arm64-v8a-clang-libcxx.txt \
    --profile:build default \
    --build=outdated
```

You get the error:
```
ERROR: protobuf/3.13.0: option 'fPIC' doesn't exist
Possible options are ['shared', 'with_zlib', 'lite']
```

Having either or of `requires` and `build_requires` works.

### Logs (Include/Attach if Applicable)

<details><summary>Click to expand log</summary>

```
[...]

PR #12905, opened 2023-01-16
fix failure of removing options for build_requires in cross-build

memsharded:
Changelog: Bugfix: fix failure of removing options for build_requires in cross-build.
Docs: Omit

Very exploratory, potentially high risk fix for https://github.com/conan-io/conan/issues/8443.


memsharded:
@czoido this is relatively high risk, but I have protected it with the ``"build"`` profile existence (which is still "under development" if I recall correctly?), so it requires the 2 profiles as opt-in, and it might unblock a good number of users hit by this.

Issue #5959, opened 2019-10-22
[question] Is it possible to list the system requirements for a package?


peter-moran:
Is it possible to list the system requirements for a package?

For example, I would like to find out what system packages `conan install qt/5.13.1@bincrafters/stable` would install.

The reason I need this is so I can build my binaries on one machine and then copy them to another machine for use, and I'd like a clean list of all the system packages needed on that other machine (I'm open to other ways of doing this if there are).

nicememory:
I think conan is just the tool for it. Think about pip for Python. Conan is the 'pip' counterpart for C/C++.
Build the binaries with conan and upload the binaries to a conan server. Then download the binaries on the other machine from conan server.

memsharded:
Hi @peter-moran 

Unfortunately, right now there is no way to get this list of system requirements.
At least, it is something that is in our roadmap. We don't like the fact that conan recipes are installing system packages by default (and suddenly a sudo permission is prompted to the user), so we are changing this. The new approach will list the things that are required and raise if they are not present. Maybe install, but probably that will be done as opt-in. 

We need to discuss and think about it. In any case, it won't be possible until Conan 2.0, because that would break too much.

Related issue: https://github.com/conan-io/conan/issues/4226 suggests to store the system requirements in metadata, for example to help uninstalling

peter-moran:
Thanks for the response, completely understandable.

Is it currently possible to have conan run just the `system_requirements()` part of an `conan install`, or is just all or none? This could help me, say, set up a docker image with the system requirements automatically pre-installed.

uilianries:
> Is it currently possible to have conan run just the system_requirements() part of an conan install, or is just all or none

Unfortunately, there is not way to run only system_requirements. If you are using docker or VM, you can try Ansible or something similar.

memsharded:
I have been checking this that was pending for 2.0 milestone.
Unfortunately, the complexity is still very large for the value, and we don't have enough bandwidth to consider this for the final 2.0 release next month. So lets move it to 2.X roadmap, these would be more or less the necessary steps as I envision:

AlternativeA: declarative
- Define a known, structured format for the packages to install, most likely in ``conandata.yml``. This is a bit challenging, need to figure out the structure to account for all variability
- Define a high level wrapper of all "SystemPackageTools" that will call the right one according to the ``conandata.yml`` structure
- Have a way (generic) to collect data from ``conandata.yml``
This  would have the advantage to be platform independent, can collect all requirements for all platforms at once

[...]

PR #12912, opened 2023-01-17
[develop2] Improving system_requirements

memsharded:
Changelog: Fix: Improving system_requirements.

Allow collecting the ``system_requires`` of the graph:
- In ``create`` and ``install`` commands
- In ``graph info`` commands, without installing anything
- Added new ``collect`` mode for system-requires
- Decouple the installation of system_requires (doing it first) from the installation of the graph packages.

Close https://github.com/conan-io/conan/issues/5959

### DIFF
--- a/conan/cli/commands/graph.py
+++ b/conan/cli/commands/graph.py
@@ -156,7 +156,7 @@ def graph_info(conan_api, parser, subparser, *args):
         conan_api.graph.analyze_binaries(deps_graph, args.build, remotes=remotes, update=args.update,
                                          lockfile=lockfile)
         print_graph_packages(deps_graph)
-        if profile_host.conf.get("tools.system.package_manager:mode") == "collect":
+        if profile_host.conf.get("tools.system.package_manager:mode") in ("check", "report"):
             conan_api.install.install_system_requires(deps_graph)
 
         lockfile = conan_api.lockfile.update_lockfile(lockfile, deps_graph, args.lockfile_packages,
--- a/conan/cli/commands/graph.py
+++ b/conan/cli/commands/graph.py
@@ -156,7 +156,9 @@ def graph_info(conan_api, parser, subparser, *args):
         conan_api.graph.analyze_binaries(deps_graph, args.build, remotes=remotes, update=args.update,
                                          lockfile=lockfile)
         print_graph_packages(deps_graph)
-        if profile_host.conf.get("tools.system.package_manager:mode") in ("check", "report"):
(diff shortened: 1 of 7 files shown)


conan/tools/system/package_manager.py
```diff
@@ -147,7 +148,12 @@ def _install_substitutes(self, *packages_substitutes, update=False, check=True,
         raise ConanException("None of the installs for the package substitutes succeeded.")
 
     def _install(self, packages, update=False, check=True, **kwargs):
-        if check:
+        pkgs = self._conanfile.system_requires.setdefault(self._active_tool, [])
+        pkgs.extend(p for p in packages if p not in pkgs)
+        if self._mode == self.mode_collect:
+            return
+
+        if check or self._mode == self.mode_check:
```
memsharded: @czoido I am not sure if this was a bug or on-purpose, but ``apt`` was not running checks for mode=="check"?

conan/cli/commands/graph.py
```diff
@@ -156,6 +156,8 @@ def graph_info(conan_api, parser, subparser, *args):
         conan_api.graph.analyze_binaries(deps_graph, args.build, remotes=remotes, update=args.update,
                                          lockfile=lockfile)
         print_graph_packages(deps_graph)
+        if profile_host.conf.get("tools.system.package_manager:mode") == "collect":
```
memsharded: This is the first usage of such a conf in our code, getting it directly from profile instead of ``conanfile``, but I don't see a strong reason to not allow it in our codebase.

conan/tools/system/package_manager.py
czoido: I would say it is on-purpouse as it was conceived, `mode==check` is the default mode and means that it won't install anything but it will check packages at most if you set that as an argument in the `install` method or call directly to the `check` method, but does not mean that it will force the check even if you set the `check` argument to `False`.

memsharded: But it has a problem, ``apt-get`` is not doing even a check, but always returning the ``install()`` packages list, because the override has different default: `` def install(self, packages, update=False, check=False, recommends=False):``

memsharded: Finally did:
- ``mode=check`` always does a check, raise if packages not found. Can be done in ``graph info`` too
- ``mode=report`` do nothing except collect and report the system_requires. Do not check, do not raise.

czoido (approved):

Issue #12909, opened 2023-01-16
[feature] Authenticate against disabled remote


slietzau:
### What is your suggestion?

We have several remotes we use for different maturity levels. The most volatile one should be disabled by default.
We use `conan config install` to distribute the configuartion internally. Because we have a number of remotes we also automate the authentication to the remotes with a script so the user has to input his password only once.
Trying to authenticate with a disabled remote however fails with an error that the remote is disabled.

This combination makes it awkward to implement our workflow because we have to first authenticate against all remotes and then call `conan remote disable` for all that should be disabled. It would be better to define remotes as disabled in `remotes.json` but still allow `conan user` to authenticate against them.

I tried both `1.X` and `2.0.0-beta8`.
On a side note: is it expected, that `2.0` doesn't list disabled remotes with `conan remote list`? The only way I found was to open `remotes.json` in a text editor.

### Have you read the CONTRIBUTING guide?

- [x] I've read the CONTRIBUTING guide

memsharded:
Hi @slietzau 

Thanks very much for your suggestion.
I think it makes sense, both things:
- ``conan remote list`` should probably list all remotes, or at the very least have options to display disabled ones
- I don't see a strong reason to not allowing authentication against a disabled remote, as long as it is not for resolution, I am fine with it.

Lets discuss with the team, and we might be able to have it in next 2.0-beta.9

memsharded:
Closed by https://github.com/conan-io/conan/pull/12937 and #12913, for next beta.9

PR #12913, opened 2023-01-17
Enable authentication against a disabled remote

AbrilRBS:
Changelog: Fix: Enable authentication against disabled remotes.
Docs: Omit

There's no need to force a remote to be enabled for it to be able to be authenticated against, as disabling a remote should only affect if they are used for dependency resolving
Part 1 of 2 for #12909

### DIFF
--- a/conan/cli/commands/remote.py
+++ b/conan/cli/commands/remote.py
@@ -177,7 +177,7 @@ def remote_login(conan_api, parser, subparser, *args):
                                 'requested interactively (not exposed)')
 
     args = parser.parse_args(*args)
-    remotes = conan_api.remotes.list(pattern=args.remote)
+    remotes = conan_api.remotes.list(pattern=args.remote, only_enabled=False)
     if not remotes:
         raise ConanException("There are no remotes matching the '{}' pattern".format(args.remote))
 
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -458,21 +458,17 @@ class EnvVars:
 
     def save_sh(self, file_location, generate_deactivate=True):
         filepath, filename = os.path.split(file_location)
-        deactivate_file = os.path.join(filepath, "deactivate_{}".format(filename))
         deactivate = textwrap.dedent("""\
-           echo "echo Restoring environment" > "{deactivate_file}"
-           for v in {vars}
-           do
-               is_defined="true"
-               value=$(printenv $v) || is_defined="" || true
-               if [ -n "$value" ] || [ -n "$is_defined" ]
-               then
-                   echo export "$v='$value'" >> "{deactivate_file}"
-               else
-                   echo unset $v >> "{deactivate_file}"
-               fi
-           done
-           """.format(deactivate_file=deactivate_file, vars=" ".join(self._values.keys())))
+            deactivate_conanrun () {{
+                echo "Restoring environment"
+                for v in {vars}
+                do
+                    export $v="$_CONAN_OLD_${{v}}"
+                    unset "_CONAN_OLD_${{v}}"
+                done
+                unset -f deactivate
+            }}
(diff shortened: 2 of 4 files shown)


prince-chrismc:
Test code please :) 🙏

AbrilRBS:
Yes! Sorry @prince-chrismc I didn't leave a message in the PR. Got into a rabbit-hole while writing tests for this PR which turned into #12915 and it delayed them a bit, but I've got them now, thanks for noticing :)

prince-chrismc:
Usage makes sense 👍

memsharded (approved):

memsharded:
Just a quick note, this doesn't fully close https://github.com/conan-io/conan/issues/12909 (but it will be automatically closed when this is merged, because of the comment above). The missing bit is the ``conan remote list`` displaying all remotes.

conan/cli/commands/remote.py
```diff
@@ -177,7 +177,7 @@ def remote_login(conan_api, parser, subparser, *args):
                                 'requested interactively (not exposed)')
 
     args = parser.parse_args(*args)
-    remotes = conan_api.remotes.list(pattern=args.remote)
+    remotes = conan_api.remotes.list(pattern=args.remote, only_enabled=False)
```
franramirez688, line 180: Could you please add a test checking that we're listing both enabled and disabled remotes?

franramirez688 (commented):
It looks good! For me, it's only missing a new test to check that we're really listing both remotes now. Good job! 👍

SSE4:
that looks suspicious. if remote is disabled, then no traffic should be sent to/from it. why does conan need to authenticate against explicitly disabled remote in the first place? seems to be a security issue IMO.

conan/cli/commands/remote.py
memsharded: Listing is a different fix, in another place, and needs another test.

memsharded:
> that looks suspicious. if remote is disabled, then no traffic should be sent to/from it. why does conan need to authenticate against explicitly disabled remote in the first place? seems to be a security issue IMO.

This works only for the explicit ``conan remote login`` command. If the user is actively issuing a command, why would be a security issue? Why requesting the user to ``conan remote enable`` + ``conan remote login`` + ``conan remote disable``? It looks good.

franramirez688 (approved):

conan/cli/commands/remote.py
franramirez688: Sorry, I did not realize this command was the remote login. Good then!

SSE4:
> This works only for the explicit `conan remote login` command. If the user is actively issuing a command, why would be a security issue? Why requesting the user to `conan remote enable` + `conan remote login` + `conan remote disable`? It looks good.

that might be exploited via similarly looking remote names, like similar characters or just minor typos in words.
e.g. good luck distinguishing `conan` from `сonan` or `conаn`.  and in this case, you can easily send your credentials to the wrong remote, so they will be stolen or leaked. very easy to exploit via simple social engineering by sending a bit differently looking commands.
if user decided to disable remote, it should remain to be disabled IMO, it was disabled for reason, and it should be respected.

memsharded:
> that might be exploited via similarly looking remote names, like similar characters or just minor typos in words.
e.g. good look distinguishing conan from сonan or conаn. and in this case, you can easily send your credentials to the wrong remote, so they will be stolen or leaked. very easy to exploit via simple social engineering by sending a bit differently looking commands.

All of that can already happen with enabled remotes, if you add some remote to your existing remotes and that remote is malicious, then good luck, I don't think not allowing the authenticate for the reason it is disabled will save you from anything. ``conan remote disable`` cannot be a security measurement, it was always decided to avoid fetching packages from it, not as a security thing, as it doesn't make sense from the security point of view.

Issue #10979, opened 2022-04-05
[bug] The new VirtualRunEnv never clean the deactivate script (it should use a temp deactivate file)


CPickens42:
### Environment Details
  * Operating System+version: Linux (archlinux)
  * Conan version: 1.47.0
  * Python version: 3.9

### Steps to reproduce (Include if Applicable)
`conanfile.txt`:
```
[requires]
fmt/8.0.0
# testing only deps below
doctest/2.4.6

[generators]
CMakeToolchain
CMakeDeps
VirtualRunEnv
```

```sh
$ conan install . --install-folder=.conanvirtualenv --build=missing --profile:build=default
$ source .conanvirtualenv/conanrun.sh
Capturing current environment in .conanvirtualenv/deactivate_conanrunenv-release-x86_64.sh
Configuring environment variables
$ source .conanvirtualenv/deactivate_conanrun.sh
Restoring environment
$ source .conanvirtualenv/conanrun.sh
Capturing current environment in .conanvirtualenv/deactivate_conanrunenv-release-x86_64.sh
Configuring environment variables
$ source .conanvirtualenv/deactivate_conanrun.sh
Restoring environment
Restoring environment
```
### Explanation of the issue

The issue here is, the more we do activate/deactivate the more the string "Restoring environment" will be displayed (and after a while it can be very annoying on my shell) and the more the script `deactivate_conanrunenv-release-x86_64.sh` will be full. When using the VirtualRunEnv I found that very annoying after a while. I would recommend that:
* the activate script create a temp deactivate file with `mktemp`
* stores the place of the deactivate file in the env variable `CONAN_DEACTIVATE_FILE`
* source $CONAN_DEACTIVATE_FILE in the non temp deactivate script.

I could do a PR for linux with that, but since I don't know how to do that on Windows (I don't have Windows installed at all) I think my PR won't be accepted.

@memsharded @lasote @czoido: I think you guys are the one working on [this recently](https://github.com/conan-io/conan/commits/develop/conan/tools/env) (so I hope it is okay to highlight you), what do you think?

memsharded:
I have tried this in Windows, apparently it is not happening there.
Also inspected the code in .sh, at least first inspection it is not evident why this is happening, but I will like to understand better the root cause. 

Also, quick question, where does the ``conanvirtualenv`` folder comes from? It is not evident from your code why the scripts would be located there.

CPickens42:
[...]

PR #12914, opened 2023-01-17
wip: Make deactivate_conanx a function instead of an extra file

### LINKED ISSUE #5719: Collision when jointly using deactivate.sh and deactivate_run.sh
To help us debug your issue please explain:

- [x ] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [ x] I've specified the Conan version, operating system version and any tool that can be relevant.
- [ x] I've explained the steps to reproduce the error or the motivation/use case of the question/suggestion.

The scripts generated with virtualenv and virtualrunenv generators have a problem:
They both use the same environment variable OLD_PS1 and hence collide with each other during deactivation.

I suggest do define a variable OLD_PS1_RUN in the virtualrunenv generator in order to avoid the collision.

AbrilRBS:
As a draft so I can share the idea around, and unfinished as of now.
This is something I already talked with @memsharded (That having a deactivate could be cool), but this does not do half what Python's virtualenv does to ensure correctness.

After a `source conanrun.sh`, now you'll get a `deactivate_conanrun` etc for each env. Note that this function does not get propagated to spawned childs of the shell in cases of VirtualRunEnv, this is in line with Python's virtualenv behaviour
For #10979

Close https://github.com/conan-io/conan/issues/5719


Bearwolves:
@RubenRBS : We tried out your unfinished work to see whether that would solve #12971.

We faced multiple issues, I report them already to hopefully ease your work on that topic :)

- shell function can not have a `dot` e.g. `.` in their name (a property would be nice as in https://github.com/conan-io/conan/blob/5420fc6c182a7862228c778b6299bc5a16aa3530/conan/tools/env/virtualbuildenv.py#L23-L30)
- `argument list too long` => `The issue was that printenv was being called with "$<VARNAME>" instead of just "<VARNAME>", leading to the value of the variable was being polled from printenv instead of telling printenv what variable's value to print.`
- conan code uses at some places the deactivate script like [here](https://github.com/conan-io/conan/blob/develop/conans/client/generators/__init__.py#L307-L310)
  - we created a deactivate file which calls the function
- Another issue we faced (in a very specific case which I can understand may not be supported), nested virtual environment. It happens as by default we are always sourcing the `conanbuild.sh` in our build scripts. In one specific case where we want to test the generated `deploy` output of package we overwriting the `PATH` to empty `PATH=` while deactivating a second time, if I remember correctly.

Herewith the patch a colleague used to move forward on that topic. I kept on purpose all the printf, which may also help you during your development phase!

Hope it helps :)

```
--- environment.py	2023-01-27 14:30:58.000000000 +0100
+++ environment.py	2023-01-30 16:05:45.769436416 +0100
@@ -378,21 +378,29 @@

     def save_sh(self, file_location, generate_deactivate=True):
         filepath, filename = os.path.split(file_location)
-        deactivate_file = os.path.join(filepath, "deactivate_{}".format(filename))
+        print(f"filename = {filename}")
+        filename_without_dots = filename.replace('.', '_')
+        print(f"filename_without_dots = {filename_without_dots}")
         deactivate = textwrap.dedent("""\
-           echo "echo Restoring environment" > "{deactivate_file}"
-           for v in {vars}
-           do
-               is_defined="true"
-               value=$(printenv $v) || is_defined="" || true
-               if [ -n "$value" ] || [ -n "$is_defined" ]
-               then
-                   echo export "$v='$value'" >> "{deactivate_file}"
-               else
-                   echo unset $v >> "{deactivate_file}"
-               fi
-           done
-           """.format(deactivate_file=deactivate_file, vars=" ".join(self._values.keys())))
+            deactivate_{filename_without_dots} () {{
+                echo "Restoring environment"
+                for v in {vars}
+                do
+                    export ${{v}}=$(printenv _CONAN_OLD_${{v}})
+                    printf "%s = %s\\n" "${{v}}" "$(printenv ${{v}})"
+                    unset "_CONAN_OLD_${{v}}"
+                done
+                unset -f deactivate_{filename_without_dots}
+            }}
+
+            printf "I'm in '{filename}'...\\n\\n"
+
+            printf "Over-Writing content 'deactivate_{filename_without_dots}' to file 'deactivate_{filename}'\\n"
+            printf "See https://stackoverflow.com/questions/8818119/how-can-i-run-a-function-from-a-script-in-command-line\\n\\n"
+
+            echo "deactivate_{filename_without_dots}" > {filepath}/deactivate_{filename}
+
+            """.format(vars=" ".join(self._values.keys()), filename_without_dots=filename_without_dots, filename=filename, filepath=filepath))
         capture = textwrap.dedent("""\
               {deactivate}
               """).format(deactivate=deactivate if generate_deactivate else "")
@@ -400,10 +408,15 @@
         for varname, varvalues in self._values.items():
             value = varvalues.get_str("${name}", self._subsystem, pathsep=self._pathsep)
             value = value.replace('"', '\\"')
+            result.append(f'export _CONAN_OLD_{varname}=$(printenv {varname})')
+            result.append(f'echo _CONAN_OLD_{varname}=${varname}')
             if value:
-                result.append('export {}="{}"'.format(varname, value))
+                result.append(f'export {varname}="{value}"')
+                result.append(f'echo {varname}="${varname}"')
             else:
                 result.append('unset {}'.format(varname))
+                result.append('echo "{}" was empty!'.format(varname))
+            result.append('echo\n')

         content = "\n".join(result)
         save(file_location, content)
```

FYI @memsharded

AbrilRBS:
Hi @Bearwolves! Thank you so much for taking the time to post this, it's extremely appreciated. Sorry that you found so many issues with the draft, we are focused on releasing 2.0 some time this month so work in this got sidelined.

I plan to resume work on this after the release, and I'll be sure to use your patch as a starting point :)

Bearwolves:
@RubenRBS : Any news on that feature? :)

AbrilRBS:
To be carried out in the future, I know @uilianries wanted to do something in these same lines too at some point, maybe we can pick it up in the future with him, but the current approach is not something I'm comfortable with so let's start over

PR #12915, opened 2023-01-17
Adds tracebacks for failing tests

AbrilRBS:
Changelog: omit
Docs: omit

Locally I was not finding a way to get traceback for failing tests other than the code in test utils itself. 

As this change is a bit risky as it touches the api of the cli entrypoint, and maybe there's a better way to do what this PR sets out to do, I've set it as a draft until we can discuss it further.

This clearly breaks any test that relies on the exit code of any command that raises any exception, so maybe a better way to rely this information is needed (26 tests in the whole integration suitcase)

This changes outputs such as:
image

to
image
(In fact the new screenshot is out-of-date, and the first 2 entries are now skipped as they'll always be `TestClient` code)

### DIFF
--- a/conan/cli/cli.py
+++ b/conan/cli/cli.py
@@ -135,7 +135,7 @@ class Cli:
         cli_out_write("")
         cli_out_write('Conan commands. Type "conan <command> -h" for help', Color.BRIGHT_YELLOW)
 
-    def run(self, *args):
+    def run(self, *args, **kwargs):
         """ Entry point for executing commands, dispatcher to class
         methods
         """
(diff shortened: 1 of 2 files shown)


memsharded (commented):
I like the idea of getting the traces in tests output, but I'd prefer not to have to touch the prod code ``conan/cli/cli.py`` to inject artifacts that are exclusively for testing. I think it should be doable  within the ``test`` scope, I'd try a bit more there.

AbrilRBS:
Yes that's why I wanted to bring this PR up, I also didn't like to modify cli.py for this, but I didn't see how to do it any other way. I'l wait for you to come up with something :)

AbrilRBS:
Closing in favour of #12928

PR #12932, opened 2023-01-20
Feature/cmake presets prefix

memsharded:
Changelog: Feature: Define ``CMakeToolchain.presets_prefix`` to allow prefixing generated ``CMakePresets.json`` for inclusion

- Defines ``presets_prefix`` 
- Allows custom ``user_presets_path``, like ``ConanPresets.json`` to be included


czoido (approved):

franramirez688 (approved):
Overall, it looks fine, but it brings me some doubts about if we should start thinking if the path-to-presets should be managed purely by the `cmake_layout` instead of having that logic here. Perhaps, it's not the moment now, but the code is getting more and more complex because of that.

AbrilRBS (approved):
This has given me a chance to look into how the presets work so thanks! I can't speak for its functionality (Nor the usability concerns you had), but the code itself looks good :)

memsharded:
Outdated, closing

PR #12946, opened 2023-01-21
[develop2] unify selects

memsharded:
Changelog: omit

AFTER: https://github.com/conan-io/conan/pull/12942

There are 2 ``select()`` methods, that are very similar with same goals, unify them into a single one.
Requires some minor changes in behavior (and tests) for upload/download/remove


conan/api/model.py
```diff
@@ -32,107 +30,60 @@ def __repr__(self):
...
-                        if info is not None:
-                            pid_dict["info"] = info
-                        if pref.revision:
-                            prevs_dict = pid_dict.setdefault("revisions", {})
-                            prev_dict = prevs_dict.setdefault(pref.revision, {})
-                            if pref.timestamp:
-                                prev_dict["timestamp"] = timestamp_to_str(pref.timestamp)
-        return ret
+            revs_dict = self.recipes[str(pref.ref)]["revisions"]
+            rev_dict = revs_dict[pref.ref.revision]
```
SSE4: why not `setdefault` here as well? may it raise `KeyError` in some circumstances?

conan/api/model.py
```diff
@@ -32,107 +30,60 @@ def __repr__(self):
...
+                prevs_dict = package_dict.setdefault("revisions", {})
+                prev_dict = prevs_dict.setdefault(pref.revision, {})
+                if pref.timestamp:
+                    prev_dict["timestamp"] = pref.timestamp
 
+    def add_configurations(self, confs):
+        for pref, conf in confs.items():
+            rev_dict = self.recipes[str(pref.ref)]["revisions"][pref.ref.revision]
+            try:
+                rev_dict["packages"][pref.package_id]["info"] = conf
```
SSE4, line 68: maybe move `try` one line above, as `self.recipes` (or its siblings) may potentially raise `KeyError` as well?

conan/api/subapi/list.py
```diff
@@ -127,29 +129,31 @@ def select(self, pattern, package_query=None, remote=None):
...
                 if "*" not in pattern.package_id and pattern.prev is not None:
                     prefs.append(PkgReference(rrev, package_id=pattern.package_id))
+                    packages = {}
                 else:
-                    packages = self.conan_api.list.packages_configurations(rrev, remote)
+                    packages = self.packages_configurations(rrev, remote)
                     if package_query is not None:
-                        packages = self.conan_api.list.filter_packages_configurations(packages,
-                                                                                      package_query)
+                        packages = self.filter_packages_configurations(packages, package_query)
```
SSE4, line 136: might it be sub-optimal? as `self.packages_configurations` may generate a list of billions items, and then apply a filter `self.filter_packages_configurations` over it just to shrink it to the couple of items?

conan/api/subapi/list.py
```diff
@@ -100,6 +100,7 @@ def select(self, pattern, package_query=None, remote=None):
         if "*" in pattern.ref or not pattern.version or \
                 (pattern.package_id is None and pattern.rrev is None):
             refs = self.conan_api.search.recipes(pattern.ref, remote=remote)
+            refs = sorted(refs)  # Order alphabetical and older versions first
```
SSE4: can above function accept sorting, so list returned already in the right order from the very beginning? same for line 122

conan/api/subapi/list.py
```diff
@@ -82,15 +82,15 @@ def packages_configurations(self, ref: RecipeReference,
...
+    def filter_packages_configurations(pkg_configurations, query):
         """
         :param pkg_configurations: Dict[PkgReference, PkgConfiguration]
         :param query: str like "os=Windows AND (arch=x86 OR compiler=gcc)"
         :return: Dict[PkgReference, PkgConfiguration]
         """
         return filter_packages(query, pkg_configurations)
 
-    # TODO: could it be possible to merge this with subapi.search.select?
     def select(self, pattern, package_query=None, remote=None):
```
SSE4, line 94: it's not currently `@staticmethod`, so it's a bit unclear, does it actually modify the state of `self` object? from the name, I suppose, it shouldn't. it throws some exceptions in the middle of the work (e.g. `NotFoundException`), is it possible that function leaves an object in partially modified state?

conan/api/model.py
memsharded: The revision must exist at this point, I don't want to add it under any circumstance if it doesn't exist. If it raises KeyError it would be a bug, it should be reported and fixed.

conan/api/subapi/list.py
memsharded: There are no packages with a billion binaries. Largest use case will be ConanCenter with +100.
In any case, the server is not capable of filtering packages, so it should be done this way anyway.

conan/api/model.py
```diff
@@ -30,7 +33,7 @@ def __repr__(self):
         return str(self)
 
 
-class SelectBundle:
+class PackagesList:
```
memsharded, line 36: Renamed to a better name

conan/api/model.py
```diff
@@ -44,12 +47,13 @@ def add_refs(self, refs):
                 if ref.timestamp:
                     rev_dict["timestamp"] = ref.timestamp
 
-    def add_prefs(self, prefs):
+    def add_prefs(self, rrev, prefs):
         # Prevs already come in ASCENDING order, so upload does older revisions first
+        revs_dict = self.recipes[str(rrev)]["revisions"]
```
memsharded, line 52: This guarantees that a "package" section is there, even empty, when package_id patterns are used.

conan/api/model.py
```diff
@@ -85,3 +91,66 @@ def prefs(ref, recipe_bundle):
 
     def serialize(self):
         return self.recipes
+
+
+class ListPattern:
```
memsharded, line 96: This was an ``conan.internal`` import, so I had to move it and make it public ,as it was being used by commands, and it is an API public model. I also renamed it from SelectPattern->ListPattern

conan/api/model.py
```diff
@@ -85,3 +91,66 @@ def prefs(ref, recipe_bundle):
...
+        recipe, package = split(expression, ":")
+        self.raw = expression
+        self.ref, rrev = split(recipe, "#", rrev)
+        ref, user_channel = split(self.ref, "@")
+        self.name, self.version = split(ref, "/")
+        self.user, self.channel = split(user_channel, "/")
+        self.rrev, _ = split(rrev, "%")
+        self.package_id, prev = split(package, "#", prev)
+        self.prev, _ = split(prev, "%")
+        if only_recipe:
```
memsharded, line 116: I have embedded some duplicated logic code in commands into the pattern itself

conan/api/subapi/cache.py
```diff
@@ -45,6 +46,13 @@ def package_path(self, pref: PkgReference):
         ref_layout = app.cache.pkg_layout(pref)
         return ref_layout.package()
 
+    @api_method
+    def check_integrity(self, package_list):
```
memsharded, line 50: This has been moved from the ``UploadAPI`` to here, because checking package integrity in the cache is not exclusively a "upload" operation. This will enable the ``conan cache check-integrity`` command with just a couple of lines.

conan/api/subapi/upload.py
```diff
@@ -13,13 +11,11 @@ def __init__(self, conan_api):
         self.conan_api = conan_api
 
     @api_method
-    def get_bundle(self, expression, package_query=None, only_recipe=False):
-        ref_pattern = SelectPattern(expression)
```
memsharded, line 17: This has been extracted to the command, to align with all other List related commands.

conan/cli/commands/list.py
```diff
@@ -71,7 +72,7 @@ def format_timestamps(item):
             for k, v in item.items():
                 if isinstance(v, dict) and v.get("timestamp"):
                     timestamp = v.pop("timestamp")
-                    k = f"{k} ({timestamp})"
+                    k = f"{k} ({timestamp_to_str(timestamp)})"
```
memsharded, line 75: This was a bug, printing only timestamps

conan/internal/integrity_check.py
```diff
@@ -0,0 +1,56 @@
+from conan.api.output import ConanOutput
+from conans.errors import ConanException
+from conans.model.package_ref import PkgReference
+from conans.model.recipe_ref import RecipeReference
+
+
+class IntegrityChecker:
```
memsharded, line 7: Just moved from other location, no changes in IntegrityChecker

franramirez688 (approved):
@memsharded This is really great!! 👏  Great job!

czoido (approved):

PR #12956, opened 2023-01-23
MSBuildDeps: static libraries use the Lib section instead of Link

### LINKED ISSUE #12954: [bug] Visual Studio properties missing for statically linked libraries
### Environment details

* Compiler+version: Visual Studio 2022 (17.4.0)
* Conan version: 1.57.0
* Python version: 3.10.2

### Steps to reproduce

The `MSBuildDeps` generator doesn't create the necessary properties for statically linked libraries. It creates a `Link` block which is suitable for dynamically linked projects, but is missing the `Lib` block needed for static libs. I've created a small project to demonstrate and opened PR #12956 with my proposed solution.

1. Extract [MissingLibDemo.zip](https://github.com/conan-io/conan/files/10481747/MissingLibDemo.zip)
2. In the `StaticLib` folder, create a `conan` directory
3. From `StaticLib/conan`, run `conan install .. -g=MSBuildDeps --build=missing`
4. Open `MissingLibDemo.sln` in Visual Studio and build `Release` / `x64`.
5. Note that `ConsoleApp` fails to link with `error LNK2001: unresolved external symbol deflateInit_`
6. Right click `StaticLib` / `Properties` / `General` / `Configuration Type` - change to `Dynamic Library`
7. Build again
8. Notice the successful build
9. Switch `StaticLib` back to `Static library`
10. Close the solution
11. Edit `StaticLib/conan/conan_zlib_release_x64.props`. Copy and paste the `Link` block and rename `Link` to `Lib`
12. Reopen the solution and build
13. Notice that the build is successful

### Logs

_No response_

illusionistpm:
Changelog: (Bugfix): MSBuildDeps needs to create a Lib section to support static libraries

Fixes #12954

### DIFF
--- a/conan/tools/microsoft/msbuilddeps.py
+++ b/conan/tools/microsoft/msbuilddeps.py
@@ -71,6 +71,12 @@ class MSBuildDeps(object):
               <AdditionalDependencies>$(Conan{{name}}SystemLibs)%(AdditionalDependencies)</AdditionalDependencies>
               <AdditionalOptions>$(Conan{{name}}LinkerFlags) %(AdditionalOptions)</AdditionalOptions>
             </Link>
+            <Lib>
+              <AdditionalLibraryDirectories>$(Conan{{name}}LibraryDirectories)%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+              <AdditionalDependencies>$(Conan{{name}}Libraries)%(AdditionalDependencies)</AdditionalDependencies>
+              <AdditionalDependencies>$(Conan{{name}}SystemLibs)%(AdditionalDependencies)</AdditionalDependencies>
+              <AdditionalOptions>$(Conan{{name}}LinkerFlags) %(AdditionalOptions)</AdditionalOptions>
+            </Lib>
             <Midl>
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Thanks very much for your PR.
I'd like to understand better why this has not been failing before. We have some tests in our suite that try to cover it, with real linking of static libraries. Even the ``conan new hello/0.1 -m=msbuild_lib`` creates a template that works linking a static library from the package in the ``test_package`` via ``MSBuildDeps``, and it is working.

It would be great to add some test to the suite that would be failing without this proposed patch to understand better, and also to capture the use case and make sure it keeps working in the future.

memsharded:
Sorry, I didn't realize about https://github.com/conan-io/conan/issues/12954, I will check it too.

memsharded:
I have been reading your code, and initially I suspected that it could be due to transitive dependencies. But this test https://github.com/conan-io/conan/blob/953f2adb78c190f82f45e758560a97bac6147ec0/conans/test/functional/toolchains/microsoft/test_msbuilddeps.py#L414 validates too that this works for transitive dependencies of static libraries too, the consumer application executable builds and runs.

memsharded:
Thanks very much for your contribution, but as commented in https://github.com/conan-io/conan/issues/12954#issuecomment-1400847213, this would not be the best approach to the default expected behavior of ``MSBuildDeps``, so closing this PR.

PR #12970, opened 2023-01-25
Add "application-with-library" as package_type

### LINKED ISSUE #12969: [feature] Conan 2.0 needs a package_type for applications with libraries
### What is your suggestion?

Various applications such as GCC and Flex include headers and runtime libraries that are used by compiled or generated code. If one specifies the package_type as "application" for these package recipes, clients will not see their headers or libraries due to the following code in conans/model/requires.py:

>         if pkg_type is PackageType.APP:
>             # Change the default requires headers&libs to False for APPS
>             set_if_none("_headers", False)
>             set_if_none("_libs", False)
>             set_if_none("_run", True)

Thus Conan 2.0 would benefit from having an additional package_type "application-with-library" to support this use case. It could be implemented with the following one-line addition:

>     class PackageType(Enum):
>         LIBRARY = "library"  # abstract type, should contain shared option to define
>         STATIC = "static-library"
>         SHARED = "shared-library"
>         HEADER = "header-library"
>         BUILD_SCRIPTS = "build-scripts"
>         APP = "application"
>         APP_LIB = "application-with-library"    # Support applications with libraries (e.g., GCC, Flex, etc.)
>         PYTHON = "python-require"
>         UNKNOWN = "unknown"

I realize this is equivalent to not specifying any package_type but it is inconsistent to not be able to specify the package_type explicitly for every recipe. Doing so permits things like programmatically querying for all applications in a given code base, which becomes impossible when some of the applications are not labelled thusly. Thanks,

System-Arch:
Changelog: (Feature): Add "application-with-library" as additional package_type (Closes #12969)

### DIFF
--- a/conans/model/pkg_type.py
+++ b/conans/model/pkg_type.py
@@ -49,6 +49,9 @@ class PackageType(Enum):
         if conanfile_type is not None:  # Explicit definition in recipe
             try:
                 conanfile_type = PackageType(conanfile_type)
+                    if conanfile_type == PackageType.APP_LIB:
+                        # Treat 'application-with-library' as an alias for 'library'
+                        conanfile_type = PackageType.LIBRARY
             except ValueError:
                 raise ConanException(f"{conanfile}: Invalid package type '{conanfile_type}'. "
                                      f"Valid types: {[i.value for i in PackageType]}")
--- a/conans/model/pkg_type.py
+++ b/conans/model/pkg_type.py
@@ -49,9 +49,9 @@ class PackageType(Enum):
         if conanfile_type is not None:  # Explicit definition in recipe
             try:
                 conanfile_type = PackageType(conanfile_type)
-                    if conanfile_type == PackageType.APP_LIB:
-                        # Treat 'application-with-library' as an alias for 'library'
-                        conanfile_type = PackageType.LIBRARY
+                if conanfile_type == PackageType.APP_LIB:
+                    # Treat 'application-with-library' as an alias for 'library'
+                    conanfile_type = PackageType.LIBRARY
             except ValueError:
                 raise ConanException(f"{conanfile}: Invalid package type '{conanfile_type}'. "
                                      f"Valid types: {[i.value for i in PackageType]}")
--- a/conans/model/pkg_type.py
+++ b/conans/model/pkg_type.py
@@ -10,6 +10,7 @@ class PackageType(Enum):
     HEADER = "header-library"
     BUILD_SCRIPTS = "build-scripts"
     APP = "application"
+    APP_LIB = "application-with-library"  # Support applications with libraries (e.g., GCC, Flex, etc.)
     PYTHON = "python-require"
     UNKNOWN = "unknown"
 


memsharded (changes requested):
Hi @System-Arch 

Thanks for your contribution.
However, I am afraid it is not as simple as adding a new type. The package-types are abstractions over a set of default requirement traits and over the propagation rules of traits in transitive dependencies.

You might want to watch https://www.youtube.com/watch?v=kKGglzm5ous if you haven't yet.

In this case, it might be even a bit more complicated. The ``library`` package-type is also an abstraction over package ``options.shared`` and it actually converts dynamically to ``shared-library`` or ``static-library`` depending on the option value.

We have already considered this type, but the thing is that it doesn't have any useful characteristics to be managed in the graph propagation. The best approach for this is to separate the application and the libraries in different packages in some cases, and in the other select the type that is more used, either ``library`` or ``application``.

Also, for most cases, labeling things as "library" is good enough, because when an application is involved in a dependency graph, it is via ``tool_requires()``, which overrides the package type traits with its own traits, using the applications inside the package, not the library. So ``package_type = "library"`` would be good for most of these cases like flex and protobuf.

System-Arch:
Hi @memsharded,

Thanks for your comments.

> Thanks for your contribution. However, I am afraid it is not as simple as adding a new type. The package-types are abstractions over a set of default requirement traits and over the propagation rules of traits in transitive dependencies.

Yes, I get that, but my impression was the default requirement traits are headers=True; libs=True, and Run=true, which is  what I believe we need for this hybrid use case.

> You might want to watch https://www.youtube.com/watch?v=kKGglzm5ous if you haven't yet.

Yes, I watched it about four months ago and it was my inspiration for shifting all efforts to Conan 2.0  because the traits model is much closer to our own internal build system.

> In this case, it might be even a bit more complicated. The `library` package-type is also an abstraction over package `options.shared` and it actually converts dynamically to `shared-library` or `static-library` depending on the option value.

Yes, but as I understand it, they only differ in that shared-library has run=True while static-library has run=False. Since we need run=True for the app itself, doesn't that cover the bases here?

> We have already considered this type, but the thing is that it doesn't have any useful characteristics to be managed in the graph propagation. The best approach for this is to separate the application and the libraries in different packages in some cases, and in the other select the type that is more used, either `library` or `application`.

While splitting code into two packages may be practical for internally developed code, it doesn't seem realistic for well-established 3p entities like GCC, Flex, protobuf, etc.

> Also, for most cases, labeling things as "library" is good enough, because when an application is involved in a dependency graph, it is via `tool_requires()`, which overrides the package type traits with its own traits, using the applications inside the package, not the library. So `package_type = "library"` would be good for most of these cases like flex and protobuf.

Again, this defeats the original objective, which is to be able to tell which packages contain applications and are therefore runnable. Having this ability is useful for things like bootstrapping or security management. While, I acknowledge that the proposed setting doesn't change the internal behavior of Conan 2, it provides this tracking and auditing ability that is missing today as well as with the proposed alternative (of using "library"). Thanks for reconsidering.

memsharded:
> Yes, I get that, but my impression was the default requirement traits are headers=True; libs=True, and Run=true, which is what I believe we need for this hybrid use case.

No, the default is ``run=False``. It will only be set to True if the trait is explicitly defined or if ``package_type`` is "shared-library" or "application".

> Yes, but as I understand it, they only differ in that shared-library has run=True while static-library has run=False. Since we need run=True for the app itself, doesn't that cover the bases here?

It would be necessary to implement 2 extra package types (and the built-in automatic deduction based on the ``options.shared`` value), the ``application-with-library-static`` would still be necessary, to define the consumers who are linking against the static library inside the package that they do not need to propagate the "run" requirement. This is very important to be able to isolate and skip static dependencies, otherwise, you will always depend on the flex binary, just because the flex application was used at build time of a given package, even if that package is building a shared library that is embedding the flex library statically. So such ``application-with-library-static`` would need the ``run=False`` trait when used in the "host" context, and then deactivate ``libs=False``, ``headers=False`` and activate ``run=True`` in the "build" context., and then it will be the same as ``static-library``, no difference in behavior (but still a lot of work to define and process that "alias", and also work to document and explain users that in practice there is no difference in the graph, and it would be just for informational purposes). If that is the case, maybe we want to investigate the possibility of making it a hard alias earlier in the process to the "library" type, but not sure if it is worth the effort.

System-Arch:
Hi @memsharded,

> If that is the case, maybe we want to investigate the possibility of making it a hard alias earlier in the process to the "library" type, but not sure if it is worth the effort.

Thanks for the additional insights. Based on your comments above, I have updated my PR with what I think you were suggesting. It appears to work in my test environment with Flex. Thus putting it out there for consideration. Thanks

SSE4:
IMO it doesn't scale, as we will run into Cartesian product pretty soon: `application_with_library`, `application_with_build_script`, `library_with_build_scripts` and so on.
instead, I suppose it could be a list, like:
```
package_type = "application", "library", "build_scripts"
```
so it combines all the properties for the package types listed.

eric-riff-simplisafe:
Hi Team

I've just encountered this limitation myself.

We have a couple of internal packages that provide libraries + applications (that run on host, not on build like protobuf). There are also similar packages on conan index, like openssl. It provides libraries and also some apps
```
p/opens3cb14df694a85/p/
├── bin
│   ├── c_rehash
│   └── openssl
├── conaninfo.txt
├── conanmanifest.txt
├── include
│   └── openssl
│       ├── aes.h
│       ├── asn1err.h
<...>
├── lib
│   ├── libcrypto.so -> libcrypto.so.3
│   ├── libcrypto.so.3
│   ├── libssl.so -> libssl.so.3
│   ├── libssl.so.3
│   └── ossl-modules
│       ├── fips.so
│       └── legacy.so
<...>
```
So I'm interested on this topic. I _think_ that what I need is the traits to resolve to whatever `library` would do + `run=True` for the app? 

Is it possible to define the `run=True` on the _source_ recipe itself? I know I can do it from the consumer side with
```py
self.requires('openssl/version', run=True)
``` 
But I have to remember to do that everywhere we consume these packages.

Thanks!

memsharded:
> We have a couple of internal packages that provide libraries + applications (that run on host, not on build like protobuf).

What is the relationship of a regular "library" package to a package that is an ``application``? How does a static library for example "depends" or "consumes" an application in the "host" context?

PR #12981, opened 2023-01-26
[cmake_layout] Add mechanism to create different test build folders

### LINKED ISSUE #10201: [question] layout and --test-build-folder
Hi, 

TLTR:
If a `test_package` conanfile uses `layout()` is `--test-bulid-folder` ignored?

-----

I'm trying to write a `test_package` parametrized by generator (to test `CMakeDeps` & `cmake_find_package`).

```
test_package/
├── CMakeDeps
│   └── conanfile.py
├── cmake_find_package
│   └── conanfile.py
├── CMakeLists.txt
└── test_package.cpp
```

`test_package/cmake_find_package/conanfile.py` is a pretty standard test package with the particularity that `source_folder` is declared as one level up to reuse the same `CMakeLists.txt`/`test_package.cpp` between both tests:
```
from conans import ConanFile, CMake, tools
import os

class LoggingTestConan(ConanFile):
    generators = "cmake_find_package"

    def build(self):
        cmake = CMake(self)
        cmake.configure(source_folder="..")
        cmake.build()

    def test(self):
        if not tools.cross_building(self.settings):
            bin_path = os.path.join("bin", "test_package")
            self.run(bin_path, run_environment=True)
```
`conan test test_package/cmake_find_package foo/1.2.3@ -tbf build_old` generates a `build_old` folder.

In the `CMakeDeps` case:
```
from conan import ConanFile
from conan.tools.cmake import CMake
from conan.tools.layout import cmake_layout
from conans import tools
import os

class LoggingTestConan(ConanFile):
    generators = "CMakeToolchain", "CMakeDeps"

    def layout(self):
        cmake_layout(self)
        self.folders.source = ".."

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()

    def test(self):
        if not tools.cross_building(self.settings):
            bin_path = os.path.join("bin", "test_package")
            self.run(bin_path, run_environment=True)

```
`conan test test_package/CMakeDeps foo/1.2.3@ -tbf build_new` does not generate any `build_new` folder. Instead, it follows what the `cmake_layout` imposes:
```
test_package/CMakeDeps/
├── cmake-build-release
│   └── ...
├── conanbuildinfo.txt
├── conanfile.py
├── conaninfo.txt
├── conan.lock
└── graph_info.json
```
I this expected behavior? If it is, could `conan test` detect if you're using `layout` and warn you about this?

Also, is there any way to reduce file cluttering (`conanbuildinfo.txt`, `conaninfo.txt`, `conan.lock`, `graph_info.json`)  in  `test_package` folder? (they're being generated outside the build folder)
With `conan install` we can control this via `--install-folder` but I didn't find a way with `conan test`. 

Thanks,

AbrilRBS:
Changelog: Feature: Add new configuration `tools.cmake.cmake_layout:test_build_folder` to indicate a different root test build folder.
Changelog: Feature: Adding to `cmake_layout()` a new mechanism to create different test build folders based (mainly) on the settings.
Docs: WIP
Closes: https://github.com/conan-io/conan/issues/10201

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -1,6 +1,4 @@
 import os
-import shutil
-from pathlib import Path
 
 from conans.errors import ConanException
 from conans.model.ref import ConanFileReference
@@ -71,7 +69,7 @@ def get_build_folder_custom_vars(conanfile):
 
 def get_root_test_build_folder(conanfile):
     """
-    Get a different root test build folder path (or the test_package/ one) if
+    Gets a different root test build folder path (or the test_package/ one) if
     tools.cmake.cmake_layout:test_build_folder is specified.
 
     If specified the test output build folder could look like this:
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -1,7 +1,10 @@
-import datetime
 import os
+import shutil
+from pathlib import Path
 
 from conans.errors import ConanException
+from conans.model.ref import ConanFileReference
+from conans.util.sha import sha1
 
 
 def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"):
@@ -23,15 +26,16 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
         raise ConanException("'build_type' setting not defined, it is necessary for cmake_layout()")
 
     build_folder = build_folder if not subproject else os.path.join(subproject, build_folder)
+    test_root_build_folder = get_root_test_build_folder(conanfile)  # only for test_package's
     config_build_folder, user_defined_build = get_build_folder_custom_vars(conanfile)
+
     if config_build_folder:
-        build_folder = os.path.join(build_folder, config_build_folder)
+        build_folder = os.path.join(test_root_build_folder, build_folder, config_build_folder)
     if not multi and not user_defined_build:
-        build_folder = os.path.join(build_folder, build_type)
+        build_folder = os.path.join(test_root_build_folder, build_folder, build_type)
+
     conanfile.folders.build = build_folder
-
     conanfile.folders.generators = os.path.join(conanfile.folders.build, "generators")
-
     conanfile.cpp.source.includedirs = ["include"]
 
     if multi and not user_defined_build:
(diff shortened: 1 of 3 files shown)


conans/test/functional/command/test_command_test.py
```diff
@@ -30,3 +33,10 @@ def test_conan_test(self):
         subfolder = "Release" if platform.system() != "Windows" else ""
         assert os.path.exists(os.path.join(client.current_folder, "test_package",
                                            "build", subfolder, "generators", "conaninfo.txt"))
+
+    def test_whateverrr(self):
+        tc = TestClient()
+        tc.run("new hello/1.0 -m=cmake_lib")
```
memsharded: Do not use a full blown test, that compiles, and it is very slow, just to test folders.
Should be moved to ``integration``, and being tested with some simple recipe that do not build anything.

conans/test/functional/command/test_command_test.py
```diff
@@ -30,3 +33,10 @@ def test_conan_test(self):
...
+        test_root_folder_output = temp_folder()
+        tc.run(f"create . hello/1.0@ -c user.test_build_folder='{test_root_folder_output}'")
```
memsharded: Not enough testing. The functionality should already create new different ``test_package/build/gcc_10_sha1`` folders for different configs, even if no ``user.conf`` is provided.

Then, for the other functionality of relocating the whole would be probably a  ``core.xxxx`` one.

conan/tools/cmake/layout.py
```diff
@@ -41,25 +42,44 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
...
+        test_build_root_folder = conanfile.conf.get("user.test_build_folder", check_type=str,
+                                                    default=".")
+        # Start the folder path with something common, so it can get ignored programmatically
+        test_build_folder = os.path.join(test_build_root_folder, "test_output")
+        # Get each config and create a name based on their values
+
+        test_build_settings = [f"settings.{setting}" for setting in conanfile.settings.fields]
+        configuration_discriminant = _calculate_build_vars(conanfile, test_build_settings)
+        # Append a timestamp/counter
+        configuration_discriminant.append(datetime.datetime.utcnow().strftime("%Y%m%d-%H%M%S%f"))
```
memsharded: We learnt that timestamps are not good for this, need to be immutable?

conan/tools/cmake/layout.py
```diff
@@ -22,15 +24,16 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
         raise ConanException("'build_type' setting not defined, it is necessary for cmake_layout()")
 
     build_folder = build_folder if not subproject else os.path.join(subproject, build_folder)
+    test_root_build_folder = get_root_test_build_folder(conanfile)  # only for test_package's
     config_build_folder, user_defined_build = get_build_folder_custom_vars(conanfile)
+
     if config_build_folder:
-        build_folder = os.path.join(build_folder, config_build_folder)
+        build_folder = os.path.join(test_root_build_folder, build_folder, config_build_folder)
```
memsharded, line 31: There is something strange here. This is not only for test, but it also applies to normal developer layouts, where the "build" folder is. So there should be a conditional: "if we are in a test_package, use this ``test_root_build_folder``, or use ``build_folder`` otherwise"

conan/tools/cmake/layout.py
```diff
@@ -63,3 +65,49 @@ def get_build_folder_custom_vars(conanfile):
...
+          name the [REF_NAME]_[REF_VERSION] as prefix, so it could look like:
+          >> [TEST_ROOT_PATH]/[REF_NAME]_[REF_VERSION]-test_output-[BUILD_TYPE]_[COMPILER]_[COMPILER_VERSION]-[SHA1]
+        * If tools.cmake.cmake_layout:test_build_folder is not specified, it will work as usual.
+    """
+    # Get root build folder from global.conf or profile
+    test_build_root_folder = conanfile.conf.get("tools.cmake.cmake_layout:test_build_folder",
+                                                check_type=str)
+    tested_reference = conanfile.tested_reference_str
+    # FIXME: Should we have a better way to check that?
+    if tested_reference is None or test_build_root_folder is None:
```
memsharded, line 93: I am afraid this is not the expected behavior that we were discussing in the issue. The strategy should always return a different folder than it is doing right now, which is broken. Not only when ``tools.cmake.cmake_layout:test_build_folder`` is defined, but always, and ``tools.cmake.cmake_layout:test_build_folder`` is used to be able to use a different location, like ``C:/mytemp`` for the test_package build directories. But this PR should be creating always the layout:
```
conanfile.py
test_package
   | - conanfile.py
   | - build
          | - gcc10x64debug_af2
          | - msvc191x86release_b21
```
Every folder gets deleted just before being re-created
Then, if ``tools.cmake.cmake_layout:test_build_folder=C:/mytemp`` the layout would be something like:
```
conanfile.py
test_package
   | - conanfile.py

C:/mytemp
   | - mypkg/myversion/recipe_revision_firstNchars/test
                                                       | - build
                                                              | - gcc10x64debug_af2
                                                              | - msvc191x86release_b21
```

conan/tools/cmake/layout.py
franramirez688: If it's not a test folder, it will be an empty string (the conditional is inside the function), so there is no need to put more conditionals there.

memsharded: Yes, but I am afraid the expected order is the reversed one, the test-folder is closer (or substitute for the ``config_build_folder``.

PR #12991, opened 2023-01-27
fix test_package output folder

memsharded:
Changelog: Fix: Fixed the ``layout()`` of ``test_package`` that caused concurrency issues and other CMake misconfigurations

### DIFF
--- a/conans/client/cmd/test.py
+++ b/conans/client/cmd/test.py
@@ -20,15 +20,10 @@ def install_build_and_test(app, conanfile_abs_path, reference, graph_info,
 
     # FIXME: Remove this check in 2.0, always use the base_folder
     conanfile = app.loader.load_basic(conanfile_abs_path)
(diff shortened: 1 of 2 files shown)


memsharded:
This will not work, it is impossible to make it work with modern ``layout()`` for ``test_package``.
The alternative for concurrent and better usage is:
- Keep the ``layout()`` and use ``git clean -xdf`` or similar cleaning strategy to remove the previous ``test_package/build`` folders. This is probably good for most developers.
- To drop the ``test_package/conanfile.py`` ``layout()``, if the issue is a big pain to do concurrent test in CI, this will use the complete unique hash per configuration.
- If we want to make it compatible and ready for 2.0, this is a problem. The best approach is to use different ``test_package`` and ``test_package_v2`` folders, and launch the second with ``conan create ... -tf=test_package_v2`` in Conan 2.0. This is the approach we are using in ConanCenter.

PR #12992, opened 2023-01-27
[develop2] Fixing issues with test_package output folder

memsharded:
Changelog: Fix: Fixing issues with test_package output folder.

New proposal to deal with the ``test_package`` issues:

- It is almost impossible to fix it in 1.X, so giving up, will propose another alternatives below
- in 2.0, remove the artificial ``test_output`` folder
- ``test_package`` can remove the ``build_folder`` if it is different than the ``source_folder``, and it will remove it before reusing
- ``test_package`` defines a new, more complete folder name like ``gcc10x86release14static`` to avoid collisions. 

TODOs:
- Add a more complete hash to the test_package build_folder name to completely avoid collisions
- Allow defining more parameters for ``build_folder_vars`` in ``cmake_layout()`` directly

### DIFF
--- a/conan/tools/cmake/layout.py
+++ b/conan/tools/cmake/layout.py
@@ -41,7 +41,7 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
 
     conanfile.cpp.source.includedirs = ["include"]
 
-    if multi and not user_defined_build:
+    if multi:
         conanfile.cpp.build.libdirs = ["{}".format(build_type)]
         conanfile.cpp.build.bindirs = ["{}".format(build_type)]
     else:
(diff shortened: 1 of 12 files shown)


conan/tools/cmake/layout.py
```diff
@@ -51,8 +51,12 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
 
 def get_build_folder_custom_vars(conanfile):
 
-    build_vars = conanfile.conf.get("tools.cmake.cmake_layout:build_folder_vars",
-                                    default=[], check_type=list)
+    if conanfile.tested_reference_str:
```
AbrilRBS, line 54: After your explanation I quite like this. My only gripe would be that as far as I can see, we have now lost the `test_output` bit in the path, so it'd be harder to gitignore these folders now.
May be worth re-adding them before merging (Or if it's still there, please point me to the code that does that as I must have missed it!)

conan/tools/cmake/layout.py
```diff
@@ -41,7 +41,7 @@ def cmake_layout(conanfile, generator=None, src_folder=".", build_folder="build"
 
     conanfile.cpp.source.includedirs = ["include"]
 
-    if multi and not user_defined_build:
+    if multi:
```
AbrilRBS, line 44: Nice catch :)

AbrilRBS (approved):
👍 your explanation answered most of my questions for this

franramirez688 (approved):

conan/tools/cmake/layout.py
memsharded: This is a ``cmake_layout``, the base folder name is already an argument to the function (default ``"build"``), so that will still be present, is the same default name that must be ignored for local builds (``conan install .``) too.

Issue #13000, opened 2023-01-29
[bug] conan 2.x: linking against a shared lib with private shared dependencies fails on Linux


SpaceIm:
### Environment details

* Operating System+version: Linux agent in c3i
* Compiler+version: gcc any version
* Conan version: conan 2.0.0-beta8 I guess
* Python version: I don't know

### Steps to reproduce

On Linux, run conan create *:shared=True against any recipe with private dependencies, and see link errors in test package:

`conan create . libpng/1.6.39@ -o *:shared=True`

### Logs

```
********************************************************************************
conan test conan-center-index/recipes/libpng/all/test_package/conanfile.py libpng/1.6.38@#00bf8a010936677ddb13cabdaa82dcce -pr:h /home/conan/w/prod-v2/BuildSingleReference/17875/5d252ca9-86d9-4830-a9f0-ecb330766f27/profile_linux_11_libstdcpp11_gcc_release_64.-shared-True.txt -c:h tools.system.package_manager:mode=install -c:h tools.system.package_manager:sudo=True -pr:b /home/conan/w/prod-v2/BuildSingleReference/17875/5d252ca9-86d9-4830-a9f0-ecb330766f27/profile_linux_11_libstdcpp11_gcc_release_64.-shared-True.txt -c:b tools.system.package_manager:mode=install -c:b tools.system.package_manager:sudo=True
********************************************************************************

-------- Input profiles --------
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux
[options]
*/*:shared=True
[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux
[options]
*/*:shared=True
[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True


-------- test_package: Computing dependency graph --------
Graph root
    libpng/1.6.38 (test package): /home/conan/w/prod-v2/BuildSingleReference/conan-center-index/recipes/libpng/all/test_package/conanfile.py
Requirements
    libpng/1.6.38#00bf8a010936677ddb13cabdaa82dcce - Cache
    zlib/1.2.13#13c96f538b52e1600c40b88994de240f - Cache

-------- test_package: Computing necessary packages --------

-------- Computing necessary packages --------
[...]

PR #13010, opened 2023-01-30
[develop2] fixing transitive linux libraries cmake

memsharded:
Changelog: Fix: Fixing transitive shared linux libraries in CMakeDeps.

Close https://github.com/conan-io/conan/issues/13000

There were an issue where building against transitive shared dependencies in Linux (several transitive levels), was failing when this happens in different machines (need to upload & install in a different machine to reproduce), because the linker wants information about the libs, even if not strictly necessary to the final consumer application.

The approach that seems to work is:

- Add INTERFACE_LINK_DIRECTORIES to each target creation
- Make sure that the traits in CMakeDeps do not invalidate ``self.lib_paths``, only ``libs`` to still avoiding direct linking.

This might need to extend to components too.

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/target_configuration.py
+++ b/conan/tools/cmake/cmakedeps/templates/target_configuration.py
@@ -108,6 +108,10 @@ class TargetConfigurationTemplate(CMakeDepsFileTemplate):
             set_property(TARGET {{root_target_name}}
                          PROPERTY INTERFACE_INCLUDE_DIRECTORIES
                          $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_INCLUDE_DIRS{{config_suffix}}}> APPEND)
+            # Necessary to find LINK shared libraries in Linux
+            set_property(TARGET {{root_target_name}}
+                         PROPERTY INTERFACE_LINK_DIRECTORIES
+                         $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_LIB_DIRS{{config_suffix}}}> APPEND)
             set_property(TARGET {{root_target_name}}
                          PROPERTY INTERFACE_COMPILE_DEFINITIONS
                          $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_COMPILE_DEFINITIONS{{config_suffix}}}> APPEND)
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/cmakedeps/templates/target_configuration.py
```diff
@@ -108,6 +108,10 @@ def template(self):
             set_property(TARGET {{root_target_name}}
                          PROPERTY INTERFACE_INCLUDE_DIRECTORIES
                          $<$<CONFIG:{{configuration}}>:${{'{'}}{{pkg_name}}_INCLUDE_DIRS{{config_suffix}}}> APPEND)
+            # Necessary to find LINK shared libraries in Linux
+            set_property(TARGET {{root_target_name}}
+                         PROPERTY INTERFACE_LINK_DIRECTORIES
```
jcar87, line 113: I'm a bit confused - I think this would only solve the problem in some cases, but not all.

If the `INTERFACE_LINK_DIRECTORIES` property eventually maps to cmake passing `-L` entries to the. linker, the linker documentation (https://linux.die.net/man/1/ld) for `-L` states that these only apply to libraries linked via `-l`. This is relevant since the linker _will not_ search in `-L` directories for "indirect" transitive runtime dependencies that are also not expressed in the linker flags.

That's what `-rpath` or `-rpath-link` achieves that `-L` doesn't (`-rpath` works ok when not cross building, `-rpath-link` is needed when a sysroot is also  passed).

memsharded: The command for creating ``app`` in my test is:

```
/usr/bin/cmake -E cmake_link_script CMakeFiles/app.dir/link.txt --verbose=1
/usr/bin/g++ -m64 -O3 -DNDEBUG -m64 CMakeFiles/app.dir/src/app.cpp.o -o app   -L"/tmp/tmpbiy1ou19conans/path with spaces/.conan2/p/31071b4017b74cd1/p/lib"  -L"/tmp/tmpbiy1ou19conans/path with spaces/.conan2/p/bbd4a1e27c1d25cc/p/lib" 

 -Wl,-rpath,"/tmp/tmpbiy1ou19conans/path with spaces/.conan2/p/31071b4017b74cd1/p/lib:/tmp/tmpbiy1ou19conans/path with spaces/.conan2/p/bbd4a1e27c1d25cc/p/lib" 

"/tmp/tmpbiy1ou19conans/path with spaces/.conan2/p/31071b4017b74cd1/p/lib/libchat.so"
```

I have isolated the ``-Wl, -rpath`` line, that contains 2 entries, for chat and hello.

czoido (approved):

PR #13013, opened 2023-01-31
[develop2] transitivity fixes in PkgConfigDeps for components

memsharded:
Changelog: Fix: PkgConfiDeps is using the wrong ``dependencies.host`` from dependencies instead of ``get_transitive_requires()`` computation.

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -7,6 +7,7 @@ from jinja2 import Template, StrictUndefined
 from conan.internal import check_duplicated_generator
 from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
 from conans.errors import ConanException
+from conans.model.dependencies import get_transitive_requires
 from conans.util.files import save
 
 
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -226,13 +228,8 @@ class PkgConfigConan(ConanFile):
...
             def package_info(self):
                 self.cpp_info.requires = ["other::cmp1"]
-        ```
-        Or:
 
-        ```python
-        from conan import ConanFile
-        class PkgConfigConan(ConanFile):
-            requires = "other/1.0"
+            # Or:
```
franramirez688, line 232: Why has it been removed?

franramirez688 (approved):
Looks good! FYI, `XcodeDeps` and `BazelDeps` are also using this `dependencies.direct_host` so I guess that we will have to change them as well.

czoido (approved):

conan/tools/gnu/pkgconfigdeps.py
franramirez688: Oh, ok, I saw it 😅  I preferred to duplicate the example instead of this, but I think it's OK as well

memsharded: Not removed, just compacted, reusing the same part above, and putting the OR inside the ``package_info()``, so it is easier to see what is the difference.

PR #13020, opened 2023-02-01
[develop2] fix test_folder arg

memsharded:
Changelog: omit 

Pending change of ``--test-folder=None`` => ``--test_folder=""`` to disable test_package


uilianries:
As `None` would be converted to a string, because is an argument, makes more sense passing an empty argument indeed. 👍

AbrilRBS (approved):

franramirez688 (approved):
I think it's an improvement, but I'd still prefer to see something like `--skip-tf` (perhaps not the best name for sure 😅). Anyway, as discussed, we should think if the way to disable configurations by putting that empty string is the best option. For now, they're unified under the same criteria, so that's great 👍

czoido (approved):

PR #13031, opened 2023-02-02
Expose conanfile.output to ConanFileInterface

AbrilRBS:
Changelog: Omit
Docs: Omit

For https://github.com/conan-io/examples2/pull/61, the `copy` method in `tools.files` expects a ConanFile as its first argument, but there's not much stopping us form also using a `ConanFileInterface`, so that for example copy can be called on files from `self.depenencies`. The only issue was a missing exposed property, so this adds that to the interface.

This might not be a good solution as it expands the public surface of this file, and there's a possible workaround in passing None as the argument which suppresses all output, so I'm happy to close if desired :)

### DIFF
--- a/conans/model/conanfile_interface.py
+++ b/conans/model/conanfile_interface.py
@@ -118,3 +118,7 @@ class ConanFileInterface:
     @property
     def url(self):
         return self._conanfile.url
+
+    @property
+    def output(self):
+        return self._conanfile.output


memsharded (commented):
Hi @RubenRBS 

The idea is that the output of the dependency shouldn't be used from the consumers (``self.dependencies``) side. All it does is to encapsulate the package prefix (like ``pkg/version: ``) to the output, but this isn't what should be shown in output, but the one of the consumer side calling it. The intention of that prefix is that messages outputed from the recipe have that prefix, but messages from the consumers have the prefix of the consumer. In this case, the one doing the ``copy()`` is the one to be passed, not the dependency.

AbrilRBS:
Oh, that's completely reasonable, thanks for your input, much appreciated :)

PR #13032, opened 2023-02-02
--lockfile=None literal string removed

memsharded:
Changelog: omit 

To align with other inputs logic

### DIFF
--- a/conan/api/subapi/lockfile.py
+++ b/conan/api/subapi/lockfile.py
@@ -28,7 +28,7 @@ class LockfileAPI:
         :param lockfile: the name of the lockfile file
         """
         if lockfile == "":
-            # Allow a way with ``--lockfile=None`` to optout automatic usage of conan.lock
+            # Allow a way with ``--lockfile=""`` to optout automatic usage of conan.lock
             return
 
         cwd = cwd or os.getcwd()
--- a/conan/api/subapi/lockfile.py
+++ b/conan/api/subapi/lockfile.py
@@ -27,7 +27,7 @@ class LockfileAPI:
         :param conanfile_path: The full path to the conanfile, if existing
         :param lockfile: the name of the lockfile file
         """
-        if lockfile == "None":
+        if lockfile == "":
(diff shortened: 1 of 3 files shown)


conan/api/subapi/lockfile.py
```diff
@@ -27,7 +27,7 @@ def get_lockfile(self, lockfile=None, conanfile_path=None, cwd=None, partial=Fal
         :param conanfile_path: The full path to the conanfile, if existing
         :param lockfile: the name of the lockfile file
         """
-        if lockfile == "None":
+        if lockfile == "":
             # Allow a way with ``--lockfile=None`` to optout automatic usage of conan.lock
```
franramirez688: Apart of this comment:

```suggestion
            # Allow a way with ``--lockfile=""`` to optout automatic usage of conan.lock
```
Please, review the `get_lockfile()` docstring because it refers to the `None` value several times.

Another comment about it. I'm seeing the other API functions use as the first argument `lockfile`, and they distinguish the value `None` too, so I'm wondering if they should take into account this new empty value.

franramirez688 (commented):
I'll approve, but you know, I'd like to see `--skip-lockfile` or something similar... 😅 
Only one concern/question about the PR.

conan/api/subapi/lockfile.py
```diff
@@ -27,8 +27,8 @@ def get_lockfile(self, lockfile=None, conanfile_path=None, cwd=None, partial=Fal
         :param conanfile_path: The full path to the conanfile, if existing
         :param lockfile: the name of the lockfile file
         """
```
franramirez688, line 30: Please, review this docstring because it's saying:
```
If lockfile is explicitly defined, it would be either absolute or relative to cwd
the lockfile file must exist. If lockfile="None" (as string, no lockfile will be used)
```

memsharded: Sorry, my bad, not committing changes. Fixed.

franramirez688 (approved):

PR #13034, opened 2023-02-02
Fix issue extracting install name from universal dylib

### LINKED ISSUE #13014: [bug] fix_apple_shared_install_name fails for universal dylib binaries
### Environment details

* Operating System+version: 13.1
* Compiler+version: apple-clang 14
* Conan version: 1.58
* Python version: 3.7.12

### Steps to reproduce

1. clone https://github.com/dbschenker/conan-gurobi
2. run `conan create conan-gurobi/gurobi/all/conanfile.py 10.0.0@`

### Logs

```
$ conan create conan-gurobi/gurobi/all/conanfile.py 10.0.0@

ERROR: gurobi/10.0.0: Error in package_info() method, line 79
    fix_apple_shared_install_name(self)  # pre-built C lib contains an absolute path
    ConanException: Error 2 while executing install_name_tool /Users/lwienkam/.conan/data/gurobi/10.0.0/_/_/package/daebbf7064402b094badcd78dbe07eac9f97be77/lib/libgurobi100.dylib -id @rpath/libgurobi100.dylib (architecture arm64)
```

it looks like the stuff in brackets `(architecture arm64)` is added to the command but it should not, and thus `install_name_tool` fails

jcar87:
Changelog: Bugfix: Fix issue in `fix_apple_shared_install_name` handling dylib files with multiple architectures (universal).
Docs: Omit

Close: https://github.com/conan-io/conan/issues/13014

The `otool -D` command will report the install name multiple times for `dylib` files that have multiple architectures (universal binaries). In practice, the values should be the same: this PR causes it to pick the first one encountered. When subsequently calling `install_name_tool`, it will be correctly replaced in all architectures anyway.

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -159,13 +159,17 @@ class XCRun(object):
         """path to libtool"""
         return self.find('libtool')
 
+def _get_dylib_install_name(path_to_dylib):
+    command = "otool -D {}".format(path_to_dylib)
+    output =  iter(check_output_runner(command).splitlines())
+    # Note: if otool return multiple entries for different architectures
+    # assume they are the same and pick the first one.
+    for line in output:
+        if ":" in line:
+            return next(output)
+    raise ConanException(f"Unable to extract install_name for {path_to_dylib}")
 
 def fix_apple_shared_install_name(conanfile):
-
-    def _get_install_name(path_to_dylib):
-        command = "otool -D {}".format(path_to_dylib)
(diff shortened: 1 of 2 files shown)


czoido (approved):

SSE4:
I don't think it's guaranteed values are always the same.
especially, for the two .dylibs built separately and then joined via `lipo`.
I suppose returning a dict `{"arch": "install_name"}` and then iterating over it is more reliable and fail proof.

PR #13039, opened 2023-02-03
Add CONAN_DISABLE_STRICT_MODE to be able to build old packages (#13037)

### LINKED ISSUE #13037: [bug] Conan v1.57 breaks recipes which use CMakeDeps together with old CMake tool
### Environment details

* Operating System+version:
  * Windows
* Compiler+version:
  * VS2019
* Conan version:
  * 1.57
* Python version:
  * 3.10.5

### Steps to reproduce

1. Create recipe with CMakeDeps to generate CMake config files for the dependencies.
2. Use the old CMake helper tool to build the project.
3. You get a nice exception:

> Using the wrong 'CMake' helper. To use CMakeDeps, CMakeToolchain you should use 'from conan.tools.cmake import CMake'

Since cmake_find_package is deprecated, broken and won't be fixed, projects were forced to use the CMakeDeps generator.

But some projects kept using the old CMake tool, which worked OK. CMakeDeps just generates some CMake config files for the dependencies after all. But now this is broken.

Migration to CMake toolchains is not instantaneous, it takes time to get adjusted to, especially when managing many projects.

The statements in this "Stable" section sound like this change should be considered a bug:

https://docs.conan.io/en/latest/introduction.html#stable

> From Conan 1.0 and onwards, there is a commitment to stability, with the goal of not breaking user space while evolving the tool and the platform. This means:
> Moving forward to following minor versions 1.1, 1.2, …, 1.X should never break existing recipes, packages or command line flows
> If something is breaking, it will be considered a bug and reverted

On the other hand this statement sounds like this change should be considered acceptable (not a bug):

https://docs.conan.io/en/latest/reference/conanfile/tools/cmake/cmakedeps.html

> Important
> Some of the features used in this section are still under development, while they are recommended and usable and we will try not to break them in future releases, some breaking changes might still happen if necessary to prepare for the Conan 2.0 release.

However, breaking old packages with client updates somehow is not very acceptable to me though.

### Logs

```

ERROR: my_pkg/1.5.0: Error in build() method, line 31
        cmake = CMake(self)
        ConanException: Using the wrong 'CMake' helper. To use CMakeDeps, CMakeToolchain you should use 'from conan.tools.cmake import CMake'

```

stevn:
Changelog: Feature: Add a `CONAN_DISABLE_STRICT_MODE`  environment variable as a workaround to be able to build old packages which used CMakeDeps or MSBuildDeps without switching to the new CMakeToolchain or MSBuildToolchain generator as well. 
Docs: https://github.com/conan-io/docs/pull/2950

Closes: https://github.com/conan-io/conan/issues/13037

### DIFF
--- a/conans/client/build/cmake.py
+++ b/conans/client/build/cmake.py
@@ -47,8 +47,12 @@ class CMake(object):
         :param generator_platform: Generator platform name or none to autodetect (-A cmake option)
         """
         if getattr(conanfile, "must_use_new_helpers", None):
-            raise ConanException("Using the wrong 'CMake' helper. To use CMakeDeps, CMakeToolchain "
-                                 "you should use 'from conan.tools.cmake import CMake'")
+            wrong_helper_msg = "Using the wrong 'CMake' helper. To use CMakeDeps, CMakeToolchain "\
+                               "you should use 'from conan.tools.cmake import CMake'"
+            if get_env("CONAN_DISABLE_STRICT_MODE", False):
+                conanfile.output.warning(wrong_helper_msg)
+            else:
+                raise ConanException(wrong_helper_msg)
         self._append_vcvars = append_vcvars
         self._conanfile = conanfile
         self._settings = conanfile.settings
--- a/conans/client/build/cmake.py
+++ b/conans/client/build/cmake.py
@@ -48,7 +48,9 @@ class CMake(object):
         """
         if getattr(conanfile, "must_use_new_helpers", None):
             wrong_helper_msg = "Using the wrong 'CMake' helper. To use CMakeDeps, CMakeToolchain "\
-                               "you should use 'from conan.tools.cmake import CMake'"
+                               "you should use 'from conan.tools.cmake import CMake'. "\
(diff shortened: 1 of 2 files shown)


SSE4 (approved):
I agree this is useful to keep old recipes working.
it's kinda annoying to constantly ask to migrate and update recipes, as it always requires a huge effort. you might have another priorities, but still want to receive new fixes.

stevn:
Yeah, we are in the middle of migrating our packages to Conan 2.0. But only for new versions. We won't be touching the old versions.

stevn:
See https://github.com/conan-io/docs/pull/2950

memsharded:
I am merging this, but please be aware this is declared a bug, and even if the bug apparently is not affecting you, it is still a bug and this is a "I still want to rely in this broken approach for a while" workaround, and it is possible that it is not working in other ways which will not be fixed. The legacy ``CMake`` helper will not get fixes or adaptations, and it might even raise for other reasons in future releases, so better fix the recipes to use the modern ``from conan.tools.cmake import CMake`` helper as soon as possible.

Thanks for the contribution!

PR #13040, opened 2023-02-03
Add the ability of changing cmake path

uilianries:
Changelog: Feature: Default CMake binary path can be configured by tools.cmake:path config
Docs: https://github.com/conan-io/docs/pull/2949

As checking CMake version is becoming more popular on CCI, we could provide the option of choosing which CMake should be used by the user, which means, once configured, people could skip installing CMake as tool requirement:

Plus, CMake 3.15 is expected by Conan 2.0, but some projects are using newer versions, so they really need an update version of CMake.

```python
def build_requirements(self):
    if not self.config.get('tools.cmake:path', check_type=str):
        tool_requires('cmake/3.21.5')
```

Also, in case user has more than one CMake installed (cross-building case), it will be much to choose which version should be used. 

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -60,7 +60,7 @@ class CMake(object):
         self._toolchain_file = configure_preset.get("toolchainFile")
         self._cache_variables = configure_preset["cacheVariables"]
 
-        self._cmake_program = "cmake"  # Path to CMake should be handled by environment
+        self._cmake_program = conanfile.conf.get("tools.cmake:path", check_type=str, default="cmake")
 
     def configure(self, variables=None, build_script_folder=None, cli_args=None):
         cmakelist_folder = self._conanfile.source_folder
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, conanfile):
         self._toolchain_file = configure_preset.get("toolchainFile")
         self._cache_variables = configure_preset["cacheVariables"]
 
-        self._cmake_program = "cmake"  # Path to CMake should be handled by environment
+        self._cmake_program = conanfile.conf.get("tools.cmake:path", check_type=str, default="cmake")
```
memsharded, line 63: This feature would be in contradiction with the many existing ``tool_requires`` to ``cmake``, without any validation of the version, which is a bit dangerous.

Probably it is also not aligned with conan 2.0 principle that ``tool_requires`` can affect the ``package_id``.
I think a version-checking approach, with explicit modeling of the ``tool_requires`` and valid for all ``tool_requires`` would be something like: https://github.com/conan-io/conan/pull/10166 (still the desired version has to be in the Path, but I think that is the responsibility of the users).

Let's discuss the different approaches with the team too.

uilianries:
Superseded by https://github.com/conan-io/conan/pull/10166

PR #13041, opened 2023-02-03
Export msvs_toolset to conan.tools.microsoft.visual

### LINKED ISSUE #12044: [feature] Provide conan.tools.microsoft.msvs_toolset equivalent of conans.tools.msvs_toolset
Would it be possible to provide the [msvs_toolset](https://docs.conan.io/en/latest/reference/tools.html#tools-msvs-toolset) function in the `conan` namespace and add support for the `msvc` compiler in it?
I had to workaround the use of the original `msvs_toolset` function in my recent update to the Boost recipe in conan-io/conan-center-index#12819.

uilianries:
Changelog: Feature: Add `msvs_toolset` method to `conan.tools.microsoft.visual` to identify which toolset version is associated to the compiler version.
Docs: https://github.com/conan-io/docs/pull/2951

It's useful in few recipes in CCI, the build system is not well planned by the project and we need from the scratch, like, passing all env vars and compiler arguments.

closes #12044

/cc @jwillikers

### DIFF
--- a/conan/tools/microsoft/toolchain.py
+++ b/conan/tools/microsoft/toolchain.py
@@ -7,7 +7,7 @@ from jinja2 import Template
 from conan.tools._check_build_profile import check_using_build_profile
 from conan.tools.build import build_jobs
 from conan.tools.intel.intel_cc import IntelCC
-from conan.tools.microsoft.visual import VCVars, msvc_version_to_toolset_version
+from conan.tools.microsoft.visual import VCVars, msvs_toolset
 from conans.errors import ConanException
 from conans.util.files import save, load
 
@@ -54,7 +54,7 @@ class MSBuildToolchain(object):
         self.configuration = conanfile.settings.build_type
         self.runtime_library = self._runtime_library(conanfile.settings)
         self.cppstd = conanfile.settings.get_safe("compiler.cppstd")
-        self.toolset = self._msvs_toolset(conanfile)
+        self.toolset = msvs_toolset(conanfile)
         self.properties = {}
         check_using_build_profile(self._conanfile)
 
@@ -81,37 +81,6 @@ class MSBuildToolchain(object):
             IntelCC(self._conanfile).generate()
         else:
             VCVars(self._conanfile).generate()
-
-    @staticmethod
-    def _msvs_toolset(conanfile):
-        settings = conanfile.settings
-        compiler = settings.get_safe("compiler")
-        compiler_version = settings.get_safe("compiler.version")
-        if compiler == "msvc":
-            subs_toolset = settings.get_safe("compiler.toolset")
-            if subs_toolset:
-                return subs_toolset
-            return msvc_version_to_toolset_version(compiler_version)
-        if compiler == "intel":
-            compiler_version = compiler_version if "." in compiler_version else \
-                "%s.0" % compiler_version
-            return "Intel C++ Compiler " + compiler_version
-        if compiler == "intel-cc":
-            return IntelCC(conanfile).ms_toolset
-        if compiler == "Visual Studio":
-            toolset = settings.get_safe("compiler.toolset")
-            if not toolset:
-                toolsets = {"17": "v143",
-                            "16": "v142",
-                            "15": "v141",
-                            "14": "v140",
-                            "12": "v120",
-                            "11": "v110",
-                            "10": "v100",
-                            "9": "v90",
-                            "8": "v80"}
-                toolset = toolsets.get(compiler_version)
-            return toolset or ""
 
     @staticmethod
     def _runtime_library(settings):
--- a/conan/tools/microsoft/toolchain.py
+++ b/conan/tools/microsoft/toolchain.py
@@ -117,7 +117,7 @@ class MSBuildToolchain(object):
 
         cppstd = "stdcpp%s" % self.cppstd if self.cppstd else ""
         runtime_library = self.runtime_library
-        toolset = self.toolset
+        toolset = self.toolset or ""
         compile_options = self._conanfile.conf.get("tools.microsoft.msbuildtoolchain:compile_options",
                                                    default={}, check_type=dict)
         self.compile_options.update(compile_options)
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -60,19 +60,6 @@ def msvc_version_to_toolset_version(version):
     return toolsets.get(str(version))
 
 
-def visual_version_to_toolset_version(version):
-    toolsets = {"17": "v143",
-                "16": "v142",
-                "15": "v141",
-                "14": "v140",
-                "12": "v120",
-                "11": "v110",
-                "10": "v100",
-                "9": "v90",
-                "8": "v80"}
-    return toolsets.get(str(version))
-
-
 class VCVars:
     def __init__(self, conanfile):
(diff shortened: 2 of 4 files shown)


jwillikers (approved):

conan/tools/microsoft/visual.py
```diff
@@ -275,3 +276,39 @@ def is_msvc_static_runtime(conanfile):
...
+            toolsets = {"17": "v143",
+                        "16": "v142",
+                        "15": "v141",
+                        "14": "v140",
+                        "12": "v120",
+                        "11": "v110",
+                        "10": "v100",
+                        "9": "v90",
+                        "8": "v80"}
+            toolset = toolsets.get(compiler_version)
```
SSE4: seems like it may return `None` here? not documented right now, should be `toolset.get(compiler_version, "")` instead?

conan/tools/microsoft/visual.py
```diff
@@ -275,3 +276,39 @@ def is_msvc_static_runtime(conanfile):
...
+    settings = conanfile.settings
+    compiler = settings.get_safe("compiler")
+    compiler_version = settings.get_safe("compiler.version")
+    if compiler == "msvc":
+        subs_toolset = settings.get_safe("compiler.toolset")
+        if subs_toolset:
+            return subs_toolset
+        return msvc_version_to_toolset_version(compiler_version)
+    if compiler == "intel":
+        compiler_version = compiler_version if "." in compiler_version else \
```
SSE4, line 299: may raise if `compiler_version` is `None`

conan/tools/microsoft/visual.py
```diff
@@ -275,3 +276,39 @@ def is_msvc_static_runtime(conanfile):
...
+    :return: Microsoft toolset when compiler is valid. Otherwise, an empty string.
+    """
+    settings = conanfile.settings
+    compiler = settings.get_safe("compiler")
+    compiler_version = settings.get_safe("compiler.version")
+    if compiler == "msvc":
+        subs_toolset = settings.get_safe("compiler.toolset")
+        if subs_toolset:
+            return subs_toolset
+        return msvc_version_to_toolset_version(compiler_version)
```
SSE4, line 297: this one may raise `KeyError`, should be protected here

conan/tools/microsoft/visual.py
```diff
@@ -275,3 +276,39 @@ def is_msvc_static_runtime(conanfile):
...
+    if compiler == "intel":
+        compiler_version = compiler_version if "." in compiler_version else \
+            "%s.0" % compiler_version
+        return "Intel C++ Compiler " + compiler_version
+    if compiler == "intel-cc":
+        return IntelCC(conanfile).ms_toolset
```
SSE4, line 303: seems like also raises in some cases https://github.com/conan-io/conan/blob/bcbaf3958818e7f1d8da87c57a3c8fb793852388/conan/tools/intel/intel_cc.py#L45

conan/tools/microsoft/visual.py
uilianries: fixed, using None. good catch

conan/tools/microsoft/visual.py
uilianries: good catch, instead, I'll check `compiler == intel and compiler_version`

conan/tools/microsoft/visual.py
uilianries: it's accurate as you are using intel-cc compiler on Darwin. It's a case which I barely can think we will have in production.

SSE4: there are more places where it raises (https://github.com/conan-io/conan/blob/bcbaf3958818e7f1d8da87c57a3c8fb793852388/conan/tools/intel/intel_cc.py#L51)
but my point is, we choose one of two:
- we make an effort that tool doesn't raise, ever, and only returns `None` if it cannot find toolset
- or we document a behavior that function may raise in certain scenarios

conan/tools/microsoft/visual.py
uilianries: added safe return mechanism to msvc_version_to_toolset_version

conan/tools/microsoft/toolchain.py
```diff
@@ -148,7 +117,7 @@ def format_macro(key, value):
 
         cppstd = "stdcpp%s" % self.cppstd if self.cppstd else ""
         runtime_library = self.runtime_library
-        toolset = self.toolset
+        toolset = self.toolset or ""
```
uilianries, line 120: toolset uses string on next steps and None would be stringfyied to "None", which is not acceptable.

conan/tools/microsoft/visual.py
uilianries: > or we document a behavior that function may raise in certain scenarios

Sounds reasonable. I'll update the PR https://github.com/conan-io/docs/pull/2951

conan/tools/microsoft/visual.py
```diff
@@ -56,7 +57,20 @@ def msvc_version_to_toolset_version(version):
                 '191': 'v141',
                 '192': 'v142',
                 "193": 'v143'}
-    return toolsets[str(version)]
+    return toolsets.get(str(version))
+
+
+def visual_version_to_toolset_version(version):
```
memsharded: Leave this an internal implementation detail of ``msvs_toolset``, not even as a separate function.

conans/test/unittests/tools/microsoft/test_msvs_toolset.py
```diff
@@ -0,0 +1,68 @@
...
+    ("10", "v100"),
+    ("9", "v90"),
+    ("8", "v80")])
+def test_visual_studio_default(compiler_version, expected_toolset):
+    conanfile = ConanFileMock()
+    conanfile.settings = MockSettings({"compiler": "Visual Studio", "compiler.version": compiler_version})
+    assert expected_toolset == msvs_toolset(conanfile)
+
+
+@pytest.mark.parametrize("compiler_version,expected_toolset", [
```
memsharded: Better if 3 arguments, ``compiler_version, toolset, expected_toolset``, to understand the inputs and outputs and visually see the expected equals the input toolset (even a 1 line comment saying it would be nice)

memsharded (approved):
Good job, please have a look to the minor comments and can be merged after it for next release

conan/tools/microsoft/visual.py
uilianries: Done.

conans/test/unittests/tools/microsoft/test_msvs_toolset.py
uilianries: Done.

uilianries:
I've added better docu string on tests and better description on the method. Now I guess is more about the rules:

- msvs_toolset can return a toolset
- When settings.compiler.toolset is configured, it will be returned as the same when using msvc or Visual Studio as compiler
- When there is no settings.compiler.version configured, but the compiler is msvc or Visual studio, a default toolset will be returned
- When there is not compiler version, neither compiler toolset configured, it returns None
- When using intel as compiler, it should return a standard string with the compiler version as part of the toolset
- For intel-cc there is a rule related to its version, it should follow date format, otherwise, will raise a ConanException
- When running intel-cc, the compiler.mode influences the toolset output directly

memsharded (approved):

Issue #11562, opened 2022-07-01
[bug] conan remove can't cleanup local cache (2.0)


sobczyk:
missing equvalent of `conan remove -b ref`

Conan 1.x has a feature to remove source folders and build folders
Conan 2.0 lacks that feature.

The feature would be needed for CI, where building large number of packages can exhaust disk space

czoido:
Hi @sobczyk,

As you say this feature is not available in the `conan remove` command for Conan 2.0 but we have just documented one of the new features in Conan 2.0: Conan custom commands.

- https://docs.conan.io/en/2.0/examples/extensions/commands/clean/custom_command_clean_revisions.html
- https://docs.conan.io/en/2.0/reference/extensions/custom_commands.html#reference-commands-custom-commands

With this new feature you can create your own custom commands, on top of the new PythonAPI. So if one feature is not implemented with the released Conan version you can always implement one command that covers your needs for ci or other environments. 
Please note that the Conan PythonAPI is still in development and may change, but the idea is to document it and that it will be stable in the near future. 
One of the examples in the docs is to create a `conan clean` command that removes all package and recipe revisions but the latest, that could serve as a starting point to your command removing build and source folders. 
Hope this helps!

fschoenm:
> As you say this feature is not available in the `conan remove` command for Conan 2.0 but we have just documented one of the new features in Conan 2.0: Conan custom commands.

Sorry, but is the idea of Conan 2.0 that basic functionality is not going to be implemented and everybody has to implement their own custom commands to get back to the same feature set that Conan 1.0 already had?

That leaves a very bad impression.

czoido:
Hi @fschoenm,

Thanks for the feedback. With my mention of the custom commands, I just wanted to point out that we can now provide users with a powerful mechanism to develop their customized commands and unblock them in certain situations. 

We are not saying that this functionality or others will not be built-in in Conan because you can do it yourself, if most users consider that this or other feature must be there we will implement them. 

In this specific case we thought that as build folders do not exist in the remotes it would make more sense to have another separate command to clean the cache (that's why we added the [conan clean](https://docs.conan.io/en/2.0/examples/extensions/commands/clean/custom_command_clean_revisions.html) example to the 2.0 docs) but we also wanted to know if users demanded that funcionality before adding new commands. Probably I should have explained this more in detail in my comment above, sorry for that.

db4:
[...]

PR #13050, opened 2023-02-03
[develop2] Cache clean

memsharded:
Changelog: Feature: Add `conan cache clean` subcommand to clean build and source folders.
Close https://github.com/conan-io/conan/issues/11562

### DIFF
--- a/conan/cli/commands/cache.py
+++ b/conan/cli/commands/cache.py
@@ -51,7 +51,7 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
     return path
 
 
-@conan_subcommand(formatters={"text": default_text_formatter})
+@conan_subcommand()
 def cache_clean(conan_api: ConanAPI, parser, subparser, *args):
     """
         Shows the path in the Conan cache af a given reference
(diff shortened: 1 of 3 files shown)


SSE4 (commented):
is it safe? what happens if it raises in the middle of removal, will it leave some incomplete or inconsistent cache state? (or user just hits Ctrl-C/Ctrl-Break in the middle of the command).

conan/cli/commands/cache.py
```diff
@@ -48,3 +49,26 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
         else:
             raise ConanException(f"'--folder {args.folder}' requires a recipe reference")
     return path
+
+
+@conan_subcommand(formatters={"text": default_text_formatter})
```
czoido: If the command has no output to console we don't need to have the text formatter.
```suggestion
@conan_subcommand()
```

czoido (approved):
Just the comment about the formatter and that the command required at least one argument saying what you want to clean.

czoido:
> is it safe? what happens if it raises in the middle of removal, will it leave some incomplete or inconsistent cache state? (or user just hits Ctrl-C/Ctrl-Break in the middle of the command).

@SSE4 I think this should be safe as we are not touching the database, if something goes wrong, a second execution would fix the  clean that did not finish.

conan/cli/commands/cache.py
```diff
@@ -48,3 +49,29 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
...
+                           help="Remove all packages (empty) or provide a query: "
+                                "os=Windows AND (arch=x86 OR compiler=gcc)")
+    args = parser.parse_args(*args)
+
+    if not args.source and not args.build and not args.download:
+        raise ConanException("Define at least one argument from --source, --build, --download")
+
+    ref_pattern = ListPattern(args.pattern, rrev="*", package_id="*", prev="*")
+    package_list = conan_api.list.select(ref_pattern, package_query=args.package_query)
+    conan_api.cache.clean(package_list, source=args.source, build=args.build,
```
AbrilRBS, line 76: Love how clean this looks

AbrilRBS (approved):
👍

PR #13066, opened 2023-02-07
remove option.fPIC for shared in conan new templates

memsharded:
Changelog: Feature: Remove option.fPIC for shared in `conan new` templates.

This is already common practice in ConanCenter, in practically all recipes:

- Update our templates to reflect it
- Update some tests with same pattern to further test it


AbrilRBS (approved):

AbrilRBS:
Note that cci already suggests this behaviour https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py, so this will get both templates in-sync

czoido (approved):

PR #13087, opened 2023-02-09
[develop2] fix package_type=header-library deduction

memsharded:
Changelog: omit 

It was broken when recipes removed the ``shared`` option in ``configure()`` for 2.0


jcar87 (approved):
LGTM - this fixes the issue :D

PR #13093, opened 2023-02-10
[develop2] editable ui

memsharded:
Changelog: Feature: Improve `conan editable` ui.
To align better with the rest of the new 2.0 CLI


conan/cli/commands/editable.py
```diff
@@ -21,55 +20,56 @@ def editable_add(conan_api, parser, subparser, *args):
     package is required, it is used from this <path> location instead of from the cache
     """
     subparser.add_argument('path', help='Path to the package folder in the user workspace')
-    subparser.add_argument('reference', help='Package reference e.g.: mylib/1.X@user/channel')
+    add_reference_args(subparser)
     subparser.add_argument("-of", "--output-folder",
                            help='The root output folder for generated and build files')
```
czoido, line 25: Is this `--output-folder` still relevant?

conan/cli/commands/editable.py
```diff
@@ -21,55 +20,56 @@ def editable_add(conan_api, parser, subparser, *args):
     package is required, it is used from this <path> location instead of from the cache
     """
     subparser.add_argument('path', help='Path to the package folder in the user workspace')
-    subparser.add_argument('reference', help='Package reference e.g.: mylib/1.X@user/channel')
+    add_reference_args(subparser)
```
czoido, line 23: Maybe we could also add the `--requires` argument here?

conan/cli/commands/editable.py
memsharded: Yes, it is passed below as ``args.output_folder`` and managed by editables.

conan/cli/commands/editable.py
memsharded: I am not sure what you mean. The ``name``, ``version``, ``user``, ``channel`` can be read now from the recipe, but if not defined in recipe, they need to be defined in command line. Same as ``conan create . ``, ``conan install <path>``, ``conan graph info <path>``, etc.

The ``--requires`` argument doesn't fit here?

AbrilRBS (approved):

Issue #1430, opened 2017-06-23
Cannot have two different remotes with identical URL


cleeland:
I tried adding a new remote using the same URL as an existing remote and received an error that the first remote already exists with the same URL.

Is there a design reason why I can't have multiple remotes with the same URL?

lasote:
Because a remote with the same URL is the same remote. Right? Why do you want to add it twice?

memsharded:
It will unnecessarily duplicate the time to look for dependencies when installing and the package not in that repo. Also, the tracking of the origin of the packages in the registry.txt file is by name. It makes easier managing the file.

cleeland:
> Why do you want to add it twice?
Because I have two conceptual remotes, though sometimes they are hosted at the same URL depending on deployment context.

> tracking of the origin of the packages in the registry.txt file is by name. It makes easier managing the file.
by "name" you mean "short name", right, not the URL?

> It will unnecessarily duplicate the time to look for dependencies when installing and the package not in that repo.

That just sounds like an unfortunate side-effect rather than a reason not to permit it.

memsharded:
> Because I have two conceptual remotes, though sometimes they are hosted at the same URL depending on deployment context.

Could you please elaborate it further. AFAIK, with both conan_server and Artifactory, it is not possible to have 2 different repos in the same URL, so it will be just one, and then makes sense that it is only addressable by one URL and one remote name. What do you mean with "two conceptual remotes"? And how do you merge them when you put them in the same URL?

> That just sounds like an unfortunate side-effect rather than a reason not to permit it.
It is not a side effect, it might be the cause that this check is in place. It is more common to inadvertently/incorrectly add a remote twice because not realizing that you already have it declared, than wanting to use 2 different names for the same remote. It is very, very, very common that users request protection against this type of mistakes (which later confuses and annoys them), so it is likely that this feature comes from users feedback.

lasote:
> Because I have two conceptual remotes, though sometimes they are hosted at the same URL depending on deployment context.

That sounds like a bad network configuration

cleeland:
Suppose I am a company and I have multiple products. For various reasons, it makes sense to have a conan remote for each product because conceptually they are distinct.  Those products are created and managed by distinct groups within my organization, so I want each group to manage and tend to their own server, so each remote have different URLs.  A group should only have write/upload permission to its repository.

[...]

PR #13101, opened 2023-02-11
[develop2] warning for duplicated URL remote

memsharded:
Changelog: omit

Close https://github.com/conan-io/conan/issues/1430

### DIFF
--- a/conans/test/integration/command/remote_test.py
+++ b/conans/test/integration/command/remote_test.py
@@ -247,7 +247,7 @@ class RemoteTest(unittest.TestCase):
         self.assertIn("ERROR: Remote 'origin' can't be found or is disabled", self.client.out)
 
     def test_duplicated_error(self):
-        """ check remote name and URL are not duplicated
+        """ check remote name are not duplicated
         """
         self.client.run("remote add remote1 http://otherurl", assert_error=True)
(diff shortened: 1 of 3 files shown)


conans/test/integration/command/remote_test.py
```diff
@@ -278,3 +278,16 @@ def test_invalid_url(self):
                       self.client.out)
         self.client.run("remote list")
         self.assertIn("pepe.org", self.client.out)
+
+
+def test_duplicated_url():
+    """ allow duplicated URL with --force
```
AbrilRBS, line 284: @memsharded this comment implies `--force` would be needed for this to work, but I do not see such either in the commands below or the related code. Is this a bug or a typo?

memsharded: You are right, the comment is obsolete. No --force is necessary, it was replaced with a warning. Do you think that we should change it to ``--force``?

AbrilRBS: That would certainly be my expectation for this, yes :)

memsharded: Sure, yes, I agree, lets change it

memsharded: https://github.com/conan-io/conan/pull/13112

PR #13107, opened 2023-02-12
refactor: Replace deprecated imp.load_source with importlib

### LINKED ISSUE #13106: [bug] DeprecationWarning for `imp`
### Environment details

* Operating System+version: macOS Ventura 13.2
* Compiler+version: Apple clang version 14.0.0 (clang-1400.0.29.202)
* Conan version: develop/1.58
* Python version: 3.11.1

### Steps to reproduce

1. Set up the "normal" developer environment with a recent Python, like 3.11.1
2. Run the model tests:  `python -m tox -- conans/test/unittests/model`

### Logs

The log of the tests will expose the following warning:

```
conans/client/loader.py:2
  /Users/gyula.gubacsi/Contrib/conan/conans/client/loader.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
    import imp
```

progician:
When you running the model unit tests, I see the following deprecation warning:

```
conans/client/loader.py:2
  /Users/gyula.gubacsi/Contrib/conan/conans/client/loader.py:2: DeprecationWarning: the imp module is deprecated in favour of importlib and slated for removal in Python 3.12; see the module's documentation for alternative uses
    import imp
```

In Python 3.12 the imp module will be officially removed as it was deprecated since 3.3. The test run already exposes a warning about this and this can break as soon as we switch to Python 3.12 which is already in alpha phase.

The code is more complex because the imp.load_source function is now decomposed in a more flexible, powerful import customisation system but that comes with a more verbose usage too.

The key here is to set up a loader that brings in the arbitrary file as a python module code and then we need to create the module from the spec of the loaded code. After that's done, we also have to execute the module in order to have all the contents of the module available.

Running the unit tests after the patch I no longer see the DeprecationWarning and the unit tests pass all the same.

Fixes #13106

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -1,5 +1,6 @@
 import fnmatch
-import imp
+from importlib.machinery import SourceFileLoader
+from importlib.util import spec_from_loader, module_from_spec
 import inspect
 import os
 import re
@@ -439,8 +440,10 @@ def _parse_conanfile(conan_file_path):
             old_dont_write_bytecode = sys.dont_write_bytecode
             try:
                 sys.dont_write_bytecode = True
-                # FIXME: imp is deprecated in favour of implib
-                loaded = imp.load_source(module_id, conan_file_path)
+                loader = SourceFileLoader(module_id, conan_file_path)
+                spec = spec_from_loader(loader.name, loader)
+                loaded = module_from_spec(spec)
+                loader.exec_module(loaded)
                 sys.dont_write_bytecode = old_dont_write_bytecode
             except ImportError:
                 version_txt = _get_required_conan_version_without_loading(conan_file_path)


memsharded (commented):
Hi @progician 

However, this has already been fixed in Conan 2.0, and we are no longer doing this kind of changes in 1.X codebase, we prefer to avoid potential risks associated to this kind of changes, and focusing all of our efforts in 2.0.
Changes to 1.X codebase should only be done for fixes and other possible features that would be strictly needed for the 1.X ->2.0 migraiton

progician:
Not sure if I understand this well. Are you saying that Conan 1.x will be left unusable with the upcoming Python version upgrade? This change is not something you need to maintain much, it is pretty much a straightforward removal of the use of a soon-to-be-removed deprecated module in Python 3.12.

I see no downside to the change unless you want to _force_ people into Conan 2.0.

memsharded:
Hi @progician 

Updating the major Python versions is not always that easy and there is always a risk. Fixing this ``imp`` removal doesn't mean that everything else will be working correctly. We have had at least a couple of broken things by Python major versions (3.8 and 3.9 if I recall correctly). This removal of this ``imp`` in 3.12 might be only one thing, but the upgrade might cause other side issues as well, the list of removed things is large (https://docs.python.org/3.12/whatsnew/3.12.html#removed), and I see at least a couple of other issues that could impact us (we have some distutils usage in our codebase, for example)

> This change is not something you need to maintain much, it is pretty much a straightforward removal of the use of a soon-to-be-removed deprecated module in Python 3.12.

Changing this code might make Conan work in Python 3.12, but also can cause unexpected problems with previous Python versions, like for example for users in a self-compiled Python 3.6 in an older Linux distro for different reasons. We have been there a few times, causing us tons of painful hours of debugging and support.

 What I am saying is that we prefer to avoid the risk of breaking existing Conan users with their current setup than allowing very latest Python 3.12. Yes, it is better to leave it completely broken in 3.12 than the false promise of "it works in 3.12", and then being broken because we need to revert this PR because it is breaking other existing users, which will for sure not be able to update their Python version.

>  I see no downside to the change unless you want to force people into Conan 2.0.

I don't want to force people into 2.0 (even if this will eventually happen, because 1.X is not going to be supported forever). But I don't want to be forced into Python 3.12 either, especially if Python 3.12 release is expected in October (it is now alpha5, https://peps.python.org/pep-0693/). The risk of breaking existing 1.X users for something that is going to be released in 8 months is simply not worth the risk.

progician:
Hi @memsharded !

Thanks for your detailed answer! I understand your concerns, even if I don't agree with your conclusion. Given that Conan is a Python project that is supposed to support a range of Python interpreter versions, to address the anxiety of about these different versions, the solution is generally to use a range of Python versions in the CI.

My PR is not forcing anyone to use 3.12. The method used in the current code is deprecated since Python 3.x, and it is wonder that it works at all because even the Python 3.6 docs no longer mention the imp.load_source function.

It is true that I use Python 3.11.1 as default on my computer but to make sure that all is fine, I've run the unit&integration tests on my patch on Python 3.6.15. Both of the environment fared well. I checked the CI and it seems that you are testing with 3.6.12 and it my patch is passing all the (fast) tests. So I'm very confused as to what is the source of your Python version concern here? The Python API I'm using in the patch was introduced long before Python 3.6 and this version is not supported for over a year now. As far as I can tell, this API I'm using was valid in Python 3.5.

I don't try to make Conan 1.x fully 3.12 compatible in one go but this is a reasonable step toward it.

memsharded:
> Thanks for your detailed answer! I understand your concerns, even if I don't agree with your conclusion.

No prob, I also understand your intention and approach. I guess it also depends a lot on the "scars" that everyone has, and from different points of views there are different priorities.

> It is true that I use Python 3.11.1 as default on my computer but to make sure that all is fine, I've run the unit&integration tests on my patch on Python 3.6.15. Both of the environment fared well. I checked the CI and it seems that you are testing with 3.6.12 and it my patch is passing all the (fast) tests. So I'm very confused as to what is the source of your Python version concern here?

I know it doesn't looks very probable, but one of the things that I have learned these years is about the "Hyrums Law", so this is definitely very possible. And with the very large amount of users, we have experienced it already in very different ways. Mandatory https://xkcd.com/1172/ 😂 but believe me, it is not that far from reality: We know that there are huge enterprises running Conan out there with their own compilation of Python. I have met personally 2 of them. Yes, a custom Python one, built with arcane constraints and very unexpected limitations. This is also the typical user that will get back to us quite upset if Conan breaks that way. And you will not know, because it will not be a Github ticket, but directly to my email.

As Python 3.12 is expected in October, I propose to avoid this risk at this moment, and re-evaluate the weeks before the final release, when we will have more information. Thanks very much for contributing!

PR #13117, opened 2023-02-13
[develop2] test_package --build=missing

memsharded:
Changelog: Feature: Allow ``test_package`` to process ``--build`` arguments (computing --build=never for the main, non test_package graph).

### DIFF
--- a/conans/test/integration/command/test_package_test.py
+++ b/conans/test/integration/command/test_package_test.py
@@ -2,8 +2,6 @@ import os
 import textwrap
 import unittest
 
-import pytest
-
 from conans.model.recipe_ref import RecipeReference
 from conans.paths import CONANFILE
 from conans.test.utils.tools import NO_SETTINGS_PACKAGE_ID, TestClient, GenConanfile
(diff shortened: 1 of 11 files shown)


SSE4:
I wonder, if we can avoid special treatments of test_package in graph code altogether? IIRC it was one of the things that caused sufferings in v1. in the end, test_package is just a regular consumer package, and from the graph's perspective it should be viewed as a normal thing :thinking:

conans/test/integration/command/test_package_test.py
```diff
@@ -59,6 +57,7 @@ def test_other_requirements(self):
         client.save({CONANFILE: GenConanfile().with_name("hello").with_version("0.1"),
                      "test_package/conanfile.py": test_conanfile})
         client.run("create . --user=user --channel=channel")
+        print(client.out)
```
AbrilRBS:
```suggestion
```

AbrilRBS (approved):

czoido (approved):

PR #13132, opened 2023-02-15
[MesonToolchain] Removed `package_folder` check

franramirez688:
Changelog: Bugfix: Local build was not adding default dirs in MesonToolchain (only in Linux).
Docs: omit

Related to https://github.com/conan-io/conan/pull/13118


franramirez688:
@memsharded it works in develop2 as well

Issue #5059, opened 2019-04-30
Relocatable toolchain package


jasal82:
I want to create a package encapsulating a toolchain that I can use as `build_requires` in other packages. The toolchain in question is generated by Yocto as SDK installation script (a shell script). During the installation process of the toolchain the installer relocates the executables to whatever directory I chose to install them in, rewriting all of the shared library RPATHs to fixed absolute paths. It doesn't make sense to package the filesystem snapshot of the toolchain after installation because Conan will install the toolchain dependency to some random directory, breaking the RPATH links and thus making all executables unexecutable ;)

How can I package such a toolchain without breaking it? I see two options:
* Package the installer itself instead of the installed directory tree and install it during the `build_requires` resolving phase of the consumer; is that possible somehow with Conan?
* Do not relocate executables and make Yocto save a relocation script that can be executed later when the packaged directory tree is unpacked to adapt the RPATHs to the actual installation directory; there are options in the SDK installer for that but they are marked as advanced debugging options so I'd rather not touch them. I still need a way to hook into the `build_requires` installation phase so that I can call the relocation script.

memsharded:
Hi @jasal82 

Some of the team have been recently using yocto too (ping @danimtb and @lasote). We plan to write down it in docs.

Indeed, RPATHs and absolute paths are a problem, because of relocation of files in other machines. Conan typically relies on not using or discarding RPATHs in libraries and later using runtime LD_LIBRARY_PATH.

For running a custom relocation script in the installation phase, the truth is that Conan doesn't have a specific hook, cause in the design is that Conan packages cannot be changed after installation. Every package contains a manifest with the files and checksum, and one package that is installed can in theory be uploaded to a server. Doing something like changing the files will break the package, the manifest will be outdated, and the package revision (versioning) of the package becomes a nightmare.

(We might need to revise this concept, and analyse the issue again, I am marking this issue for discussion.)

If the package is consumed as ``build_require``, it is not enough to consume its env vars (like LD_LIBRARY_PATH) explicitly, or even implicitly (with ``virtualrunenv`` generator)? Does the script do other things besides changing RPATHs?

memsharded:
[...]

PR #13169, opened 2023-02-20
[develop2] test for relocatable toolchains

memsharded:
Changelog: omit
Docs: omit

Close https://github.com/conan-io/conan/issues/5059

This test allows:
- To create relocatable toolchain packages
- A relocatable toolchain package requires some customization in every machine, that would break immutability
- We an use a base immutable package with the heavy bits, that take long to build
- Another package ``tool_requires`` this one, and uses ``build_policy="missing"`` and ``upload_policy="skip"`` to avoid uploading, and to build it when necessary, in a transparent way
- This other package implements in its ``package()`` or ``build()`` or both the necessary customizations to the current machine, copying from the other package what it needs


conan/cli/printers/graph.py
```diff
@@ -99,7 +99,7 @@ def _format_requires(title, reqs_to_print):
             return
         output.info(title, Color.BRIGHT_YELLOW)
         for pref, (status, remote) in sorted(reqs_to_print.items(), key=repr):
-            if remote is not None:
+            if remote is not None and status != BINARY_SKIP:
```
franramirez688, line 99: Is this a fix related to the PR?

memsharded: yes, otherwise the checks ``c.assert_listed_binary({"base/1.0": ("62e589af96a19807968167026d906e63ed4de1f5", "Skip"),`` failed as it was printing ``Skip (default)`` which doesn't make much sense

conans/test/integration/build_requires/test_relocatable_toolchain.py
```diff
@@ -0,0 +1,84 @@
+import textwrap
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+def test_relocatable_toolchain():
```
franramirez688, line 7: A docstring explaining the test and a link to the issue that closes would be great.

franramirez688 (approved):

czoido (approved):

AbrilRBS (approved):
Also TIL that you can specify `upload_policy`!

conans/test/integration/build_requires/test_relocatable_toolchain.py
memsharded: Done

PR #13185, opened 2023-02-21
[develop2] change default profile apple-clang to gnu17

memsharded:
Changelog: Feature: Change default profile cppstd for apple-clang to gnu17.
Docs: https://github.com/conan-io/docs/pull/3002

The gnu98 default for apple-clang in OSX is clearly not sufficient for many cases, not enough for ConanCenter development, many modern open sources libraries will require a modern c++17.

Other modern compilers like MSVC > 2015 default to c++14, and gcc defaults to gnu14 or gnu17 depending on the compiler version (gcc 11 defaults to gnu17), so better if apple-clang is aligned.

### DIFF
--- a/conan/cli/cli.py
+++ b/conan/cli/cli.py
@@ -179,8 +179,7 @@ class Cli:
             error = "*********************************************************\n" \
                     f"Recipe '{pkg}' seems broken.\n" \
                     f"It is possible that this recipe is not Conan 2.0 ready\n"\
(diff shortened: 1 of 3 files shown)


jcar87 (approved):
worth noting that new C+ xcode projects do indeed have a higher C++ standard level than the default for the command line compiler, so I think this is sensible

czoido (approved):

PR #13218, opened 2023-02-23
conanfile.txt => .py converter

### LINKED ISSUE #13203: [feature] Make a easy path for converting conanfile.txt to conanfile.py
### What is your suggestion?

When only consuming straightforward packages to build a simple project, [conanfile.txt](https://docs.conan.io/2/reference/conanfile_txt.html) is more than enough, but in case your project grows and need to add new logic, like using different package versions according to a specific OS, then [conanfile.py](https://docs.conan.io/2/reference/conanfile.html) is a better solution. However, converting `conanfile.txt` to a `conanfile.py` can be a monotonous task if you need to update many recipes in your enterprise code base.

It would be great if was possible to use a tool which read a `conanfile.txt` and generate an equivalent `conanfile.py`. With the new Conan 2.x feature [custom commands](https://docs.conan.io/2/reference/extensions/custom_commands.html), it would be possible to add such feature, but still it would be better if it could be centralized and maintained by Conan organization.

 

memsharded:
NOT TO MERGE, just to prove that this is very straightforward.

Better move it to its own repo of custom commands, lets discuss exactly where (or if we want this or not).

Close https://github.com/conan-io/conan/issues/13203

### DIFF
--- a/conans/test/integration/command_v2/test_convert_txt.py
+++ b/conans/test/integration/command_v2/test_convert_txt.py
@@ -23,6 +23,12 @@ convert = textwrap.dedent("""\
        txt = ConanFileTextLoader(open(path, "r").read())
        template = textwrap.dedent('''\\
             from conan import ConanFile
+            {% if layout == "cmake_layout" %}
+            from conan.tools.cmake import cmake_layout
+            {% endif %}
+            {% if layout == "vs_layout" %}
+            from conan.tools.microsoft import vs_layout
+            {% endif %}
 
             class Pkg(ConanFile):
                 {% if generators %}
@@ -101,6 +107,7 @@ def test_convert_txt():
     c.run("convert-txt .", redirect_stdout="conanfile.py")
     expected = textwrap.dedent("""\
         from conan import ConanFile
+        from conan.tools.cmake import cmake_layout
 
         class Pkg(ConanFile):
             generators = "CMakeToolchain", "CMakeDeps",
--- a/conans/test/integration/command_v2/test_convert_txt.py
+++ b/conans/test/integration/command_v2/test_convert_txt.py
@@ -1,0 +1,143 @@
+import textwrap
+
+from conans.test.utils.tools import TestClient
+
+
+convert = textwrap.dedent("""\
+    import os, textwrap
+    from jinja2 import Template
+
+    from conan.cli.command import conan_command
+    from conan.api.output import cli_out_write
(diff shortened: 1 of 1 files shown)


uilianries:
@memsharded I converted this PR in draft mode to avoid any possible merge action.

conans/test/integration/command_v2/test_convert_txt.py
```diff
@@ -0,0 +1,143 @@
...
+    def convert_txt(conan_api, parser, *args, **kwargs):
+       \"""
+       convert a conanfile.txt to a conanfile.py
+       \"""
+       parser.add_argument("path", help="Path to a folder containing a conanfile.txt")
+       args = parser.parse_args(*args)
+
+       path = conan_api.local.get_conanfile_path(args.path, os.getcwd(), py=False)
+       txt = ConanFileTextLoader(open(path, "r").read())
+       template = textwrap.dedent('''\\
```
uilianries, line 24:
```suggestion
       template = textwrap.dedent('''
```
Otherwise, will print `\` in middle of the recipe content.

conans/test/integration/command_v2/test_convert_txt.py
```diff
@@ -0,0 +1,143 @@
...
+            from conan import ConanFile
```
uilianries, line 25:
```suggestion
            from conan import ConanFile
```
Need to consider the `cmake_layout`

uilianries:
The idea is pretty solid, I tested locally, by installing the script as custom command on Conan 2.0.0 and it generates a conanfile.py. It still needs some improvements, but the current path is solid.

```
mkdir -p ${CONAN_HOME}/extensions/commands/convert
cp convert_txt.py ${CONAN_HOME}/extensions/commands/convert/cmd_convert_txt.py
conan convert:convert_txt conanfile.txt > conanfile.py
```

The `conanfile.py` content:

```python
\
from conan import ConanFile

class Pkg(ConanFile):
    generators = "CMakeToolchain", "CMakeDeps",
    default_options = {'hello*:shared': 'True'}

    def requirements(self):
        self.requires("hello/0.1")

    def build_requirements(self):
        self.test_requires("gtest/1.0")
        self.tool_requires("cmake/3.15")
        self.tool_requires("ninja/1.0")

    def layout(self):
        cmake_layout(self)

```

Still need to add some logic to choose which layout should be imported, but the idea is good enough.

AbrilRBS:
I think having a repository for custom commands that users can install would be a great idea

PR #13230, opened 2023-02-24
Fix macOS <-> Linux cross cflags resolution

### LINKED ISSUE #13219: [bug] AutotoolsToolchain uses the build compiler rather than host compiler
### Environment details

* Operating System+version: M1 macOS 13.1
* Compiler+version: N/A
* Conan version: 2.0.0 and 1.59.0
* Python version: 3.10.10

### Steps to reproduce

1. Define a profile that targets Linux (an actual cross toolchain is not needed):

```
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=11

[buildenv]
CC=x86_64-linux-gnu-gcc
CXX=x86_64-linux-gnu-g++
AR=x86_64-linux-gnu-ar
LD=x86_64-linux-gnu-ld
```

2. Create a simple test conanfile.py that uses the AutotoolsToolchain tool and dumps out its generated cflags:

```python
from conan import ConanFile
from conan.tools.gnu import AutotoolsToolchain

class TestConan(ConanFile):
    settings = { "os", "arch" }
    def generate(self):
        tc = AutotoolsToolchain(self)
        tc.generate(tc.environment())

        raise Exception(tc.cflags)
```

3. Run a build using the default build profile and cross host profile:

```
$ conan install --profile:build default --profile:host ./linux-cross --build=missing .
```

### Logs

```
Profile host:
[settings]
arch=x86_64
compiler=gcc
compiler.version=11
os=Linux
[buildenv]
CC=x86_64-linux-gnu-gcc
CXX=x86_64-linux-gnu-g++
AR=x86_64-linux-gnu-ar
LD=x86_64-linux-gnu-ld

Profile build:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=14
os=Macos


Graph root
    conanfile.py: /Users/sfackler/test/conanfile.py

======== Finalizing install (deploy, generators) ========
conanfile.py: Calling generate()
conanfile.py: Generators folder: /Users/sfackler/test
ERROR: conanfile.py: Error in generate() method, line 10
        raise Exception(tc.cflags)
        Exception: ['-isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk', '-arch x86_64']
```

Note that the resolved cflags contain flags meant for clang while the host profile indicates the compiler is gcc. It is picking up some of the host profile though since it's setting the architecture to x86_64 rather than the build profile's armv8.

This is a stripped down reproduction of a failure to cross compile the openssl recipe on macOS that fails due to the same introduction of clang cflags to the GCC compiler.

sfackler:
Changelog: Fix: Fix flags passed by AutotoolsToolchain when cross compiling from macOS to a non-Apple OS.
Docs: Omit

Fixes #13219 

### DIFF
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -1,5 +1,5 @@
 from conan.internal import check_duplicated_generator
-from conan.tools.apple.apple import apple_min_version_flag, to_apple_arch, apple_sdk_path
+from conan.tools.apple.apple import apple_min_version_flag, is_apple_os, to_apple_arch, apple_sdk_path
 from conan.tools.apple.apple import get_apple_sdk_fullname
 from conan.tools.build import cmd_args_to_string, save_toolchain_args
 from conan.tools.build.cross_building import cross_building
@@ -76,7 +76,7 @@ class AutotoolsToolchain:
             # Build triplet
             self._build = _get_gnu_triplet(os_build, arch_build, compiler=compiler)
             # Apple Stuff
-            if os_host in ["Macos", "iOS", "watchOS", "tvOS"]:
+            if os_build == "Macos" and is_apple_os(conanfile):
                 # SDK path is mandatory for cross-building
                 sdk_path = apple_sdk_path(self._conanfile)
                 if not sdk_path:
--- a/conan/tools/gnu/autotoolstoolchain.py
+++ b/conan/tools/gnu/autotoolstoolchain.py
@@ -76,7 +76,7 @@ class AutotoolsToolchain:
             # Build triplet
             self._build = _get_gnu_triplet(os_build, arch_build, compiler=compiler)
             # Apple Stuff
-            if os_build == "Macos":
+            if os_host == "Macos":
                 # SDK path is mandatory for cross-building
                 sdk_path = apple_sdk_path(self._conanfile)
                 if not sdk_path:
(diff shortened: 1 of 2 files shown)


sfackler:
ping

jcar87:
Hi @sfackler, thanks so much for your detailed report and for opening this PR.

Indeed I believe the logic was not robust enough to account for the fact that you can target Linux from macOS. 

I have made a small fix to your PR - I believe the conditional should cover the case where `os_build` is MacOS, _and_ the target OS is one of the Apple operating systems. 
* The call to `apple_sdk_path()` has a call to `xcrun`, which is a utility only available on macOS - thus, the conditional should cover that case.
* The flags `-isysroot` (specifically pointing to an Apple-provided SDK) and `-arch` carry two implications: that the sysroot is an Apple provided SDK (and thus, the `os_host` is an Apple one), and that the compiler accepts the `-arch` flag (clang or AppleClang, although we are not specifically checking for this condition).

Note for @czoido: there's a chance this doesn't fully cover the case of targetting iOS from Linux (see https://github.com/tpoechtrager/cctools-port) - but this PR at least doesn't prevent that :D

jcar87 (approved):

memsharded (approved):
It would be great to add a test for this.

AbrilRBS (approved):

czoido (approved):

Issue #13209, opened 2023-02-23
[bug] CMakeDeps.generate() fails when `build_context_activated` is used


petermbauer:
### Environment details

* Windows 10:
* Compiler: MSVC 2022
* Conan version: 1.58.0
* Python version: 3.7.5

### Steps to reproduce

1. have this in the `conanfile.py`
```
    tool_requires = ["clang-style-tools-integration/2.3.1@us/release"]
    
    def generate(self):
        deps = CMakeDeps(self)
        deps.build_context_activated = ["clang-style-tools-integration"]
        deps.generate()
```
2. the potentially relevant part of the `conanfile.py` of the `clang-style-tools-integration` package:
```
    def package_info(self):
        # to be used with the cmake_find_package and cmake_find_package_multi Conan generators
        self.cpp_info.build_modules.append("FindClangStyleToolsIntegration.cmake")
        # to be used with the newer CMakeDeps generator
        self.cpp_info.set_property('cmake_build_modules', ["FindClangStyleToolsIntegration.cmake"])
        self.cpp_info.set_property("cmake_find_mode", "config")
```
3. run `conan install ..`

When using the same package as "private" requires instead of as `tool_requires` and without the `deps.build_context_activated =` it works without any other changes so i assume the recipe is basically ok.

### Logs

```
conanfile.py (fusiongui/3b6cca5-dirty): Calling generate()
conanfile.py (fusiongui/3b6cca5-dirty): Preset 'relwithdebinfo' added to CMakePresets.json. Invoke it manually using 'cmake --preset relwithdebinfo'
conanfile.py (fusiongui/3b6cca5-dirty): If your CMake version is not compatible with CMakePresets (<3.19) call cmake like: 'cmake <path> -G Ninja -DCMAKE_TOOLCHAIN_FILE=C:\Users\bauepete\workspaces\fusiongui_cloned\fusion\build_conantest\conan_toolchain.cmake -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=RelWithDebInfo'
ERROR: conanfile.py (fusiongui/3b6cca5-dirty): Error in generate() method, line 44
        deps.generate()
        AttributeError: 'NoneType' object has no attribute 'lower'
```

memsharded:
Hi @petermbauer 

Thanks for your detailed report.

It seems you are missing ``deps.build_context_build_modules = ["xxxx"]``
[...]

PR #13260, opened 2023-02-27
fix CMakeDeps for tool_requires without build_type

memsharded:
Changelog: Bugfix: Fix ``CMakeDeps`` for ``tool_requires`` when ``build_type`` doesn't exist in the "build" profile
Docs: Omit

This is exploratory, not sure what other things it could break, to be investigated.
Close https://github.com/conan-io/conan/issues/13209


memsharded:
Closing this, too risky, I will do a new one for 2.0

Issue #13233, opened 2023-02-25
[bug] 2.0 Woes


shakfu:
### Environment details

* Operating System+version: macOS, linux, windows 11
* Compiler+version: gcc 12+
* Conan version: 2.0.0
* Python version: 3.10.x

### Steps to reproduce

My conanfile.txt:

```text
[requires]
bdwgc/8.2.2
pcre/8.45

[generators]
cmake_find_package
cmake_paths

[options]
bdwgc:cplusplus=True
bdwgc:cord=False
bdwgc:gcj_support=False
bdwgc:java_finalization=False
bdwgc:shared=False
pcre:build_pcrecpp=True
pcre:build_pcregrep=False
pcre:shared=False
pcre:with_bzip2=False
pcre:with_zlib=False
```

```bash
% cd build
% conan install .. --build=missing

ERROR: The default build profile '/Users/sa/.conan2/profiles/default' doesn't exist.
You need to create a default profile (type 'conan profile detect' command)
or specify your own profile with '--profile:build=<myprofile>'

 % conan profile detect
Found apple-clang 14.0
apple-clang>=13, using the major as version
Detected profile:
[settings]
arch=x86_64
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=14
os=Macos

% conan install .. --build=missing

ERROR: Error while parsing [options] in conanfile
Options should be specified as 'pkg:option=value'
```

Clearly my options are well specified and working in conan 1.59.0 so why am I getting an error here?

FaerHack:
conan v1 uses ~/.conan as home directory and conan v2 uses ~/.conan2 as home directory, so you need to copy your existing profile to ~/.conan2/profiles directory

memsharded:
Hi  @shakfu 

There are a few things that Conan 2.0 has change wrt Conan 1.X. 
Please have a look to https://docs.conan.io/1/conan_v2.html

You need to do ``-o pkg*:option=value``, all are patterns now, need "*"
Also cmake_find_package, cmake_paths have been removed, you need to use the new ones (already available in 1.X)

shakfu:
Ok, but I'm not sure why it was decided that it was somehow ok for a package manager to break everyone's working `conanfile.txt`. I guess the lessons of the python2 to python3 transition were somehow forgotten.

```cmake_
[requires]
bdwgc/8.2.2
pcre/8.45

[generators]
CMakeDeps
CMakeToolchain

[...]

PR #13264, opened 2023-02-28
Improve error message for unified patterns in options

oxygene:
Changelog: Fix: Improve error message for unified patterns in options.
Docs: omit

Include the option in the proposed solution to make it easier for the user to understand the new unified pattern syntax.

Also, avoid advocating `pkg*:option` to prevent ambiguities with libraries with the same prefix, e.g. `pkg-ng`. Only propose using the more explicit
 `pkg/*:option` notation.

See https://github.com/conan-io/conan/issues/13233#issuecomment-1447664410


memsharded (approved):
Thanks very much for improving this!

PR #13269, opened 2023-02-28
Add support for `.conanignore` in `conan config install`

AbrilRBS:
Changelog: Feature: Add `.conanignore` support to `conan config install`.
Docs: https://github.com/conan-io/docs/pull/3036

This allows adding a `.conanignore` file to the _root_ of the installed folder (Whether it comes from a git repository, a zip file or a local folder) that will allow skipping copying matching paths to the Conan HOME. This will allow to have extra files in the installed folder that do not pollute the user folder.

Discussed in https://github.com/conan-io/command-extensions/pull/1


conans/client/conanignore.py
```diff
@@ -0,0 +1,22 @@
+import os
+import fnmatch
+
+
+class ConanIgnoreMatcher:
```
AbrilRBS: I do not know where this file should live. This is a temporary location until someone points me at the right place :)

memsharded: In general, it should follow the same rule as any other function, class, module: minimum necessary scope. In this case, if this is only used by ``config_install``, it should be inside ``conans/client/conf``. Even as a private class inside ``config_installer.py``

conans/client/conf/config_installer.py
```diff
@@ -99,10 +100,17 @@ def _process_folder(config, folder, cache):
         raise ConanException("No such directory: '%s'" % str(folder))
     if config.source_folder:
         folder = os.path.join(folder, config.source_folder)
+    conanignore_path = os.path.join(folder, '.conanignore')
+    conanignore = None
+    if os.path.exists(conanignore_path):
+        conanignore = ConanIgnoreMatcher(conanignore_path)
```
memsharded, line 126: probably use it to conanignore ``dirs`` too.

memsharded (approved):

czoido (approved):

Issue #13272, opened 2023-02-28
[bug] Conan 1.x: imported modules remain in sys.modules cache


thorsten-klein:
### Environment details

* Operating System+version: Ubuntu 20.04
* Compiler+version: -
* Conan version: 1.59.0
* Python version: 3.8.10

### Steps to reproduce

Tree:
```
├── myproject
│   └── conanfile.py
└── mystuff
    └── MyConanFileBaseClass.py

```

myproject/conanfile.py:
```
from mystuff import MyConanFileBaseClass

class ConanRecipe(MyConanFileBaseClass.Base):
    name = "hello-world"

    def requirements(self):
        super().requirements()
        self.requires("mylib/1.0.0")
```
**Note:** When exporting the conan recipe, in `export()` method we copy the `mystuff` directory in the `export_folder`. The `sys.path` contains at index `0` this export_folder (correct).

mylib (in conan cache):

Tree:
```
export/
├── conanfile.py
└── mystuff
    └── MyConanFileBaseClass.py

```

conanfile.py
```
from mystuff import MyConanFileBaseClass

class ConanRecipe(MyConanFileBaseClass.Base):
    name = "mylib"

    def requirements(self):
        super().requirements()
```

Commands: 

Run `conan install` on local conanfile:
```
export PYTHONPATH=$PWD
conan install myproject
```

Result:
`from mystuff import MyConanFileBaseClass` from mylib imports the wrong module (NOT the one in its export folder)

thorsten-klein:
This is default behavior of python modules. Issue can  be closed.

PR #13273, opened 2023-02-28
Remove all modules (which are imported in conanfile.py) from modules cache

thorsten-klein:
Ref: #13272

Changelog: (Fix): Remove all modules (which are imported in conanfile.py) from modules cache

### DIFF
--- a/conans/test/integration/conanfile/conanfile_export_test.py
+++ b/conans/test/integration/conanfile/conanfile_export_test.py
@@ -1,0 +1,114 @@
+import unittest
+import sys
+import os
+import textwrap
+from conans.test.utils.tools import TestClient
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.mocks import RedirectedTestOutput
+
+conanfile = textwrap.dedent("""
+    from conan import ConanFile
+    import myfile
+    from myfile import MyClass
+    from myfile import number as number
+    
+    class Pkg(ConanFile):
+        name = "{name}"
+        version = "0.1"
+        exports = "myfile.py"
+
+        def package_info(self):
+            self.output.info(f"NUMBER1: {{myfile.number}}")
+            self.output.info(f"NUMBER2: {{MyClass.number}}")
+            self.output.info(f"NUMBER3: {{number}}")
+        """)
+myfile = textwrap.dedent("""\
+    number = {number}
+
+    class MyClass:
+        number = {number}
+    """)
+
+conanfile_main = GenConanfile().with_requires("pkg1/0.1","pkg2/0.1")
+conanfile_main._imports.append("import myfile")
+conanfile_main._imports.append("from myfile import MyClass")
+conanfile_main._imports.append("from myfile import number as number")
+
+class ExportTest(unittest.TestCase):
+    def test_local_imports(self):
+        c = TestClient()
+        c.save({"pkg1/conanfile.py": conanfile.format(name="pkg1"),
+                "pkg1/myfile.py": myfile.format(number="42"),
+                "pkg2/conanfile.py": conanfile.format(name="pkg2"),
+                "pkg2/myfile.py": myfile.format(number="123"),
+                "app/conanfile.py": GenConanfile().with_requires("pkg1/0.1", "pkg2/0.1")})
+    
+        c.run("create pkg1")
+        c.run("create pkg2")
+        c.run("install app")
+        
+        # do not take any cached module
+        assert "pkg1/0.1: NUMBER1: 42" in c.out
+        assert "pkg1/0.1: NUMBER2: 42" in c.out
+        assert "pkg1/0.1: NUMBER3: 42" in c.out
+        assert "pkg2/0.1: NUMBER1: 123" in c.out
+        assert "pkg2/0.1: NUMBER2: 123" in c.out
+        assert "pkg2/0.1: NUMBER3: 123" in c.out
+        
+    def test_imports_sys_path(self):
+        c = TestClient()
+        
+        c.save({"pkg1/conanfile.py": conanfile.format(name="pkg1"),
+                "pkg1/myfile.py": myfile.format(number="42"),
+                "pkg2/conanfile.py": conanfile.format(name="pkg2"),
+                "pkg2/myfile.py": myfile.format(number="123"),
+                "base/myfile.py": myfile.format(number="666"),
+                "app/conanfile.py": conanfile_main})
+    
+        c.run("create pkg1")
+        c.run("create pkg2")
+
+        c.out = RedirectedTestOutput()
+        base_path = os.path.join(c.current_folder, "base")
+        try:
+            sys.path.insert(0, base_path)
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Thanks for your contribution.

Note the test suite is failing, the check of the current dir folder is necessary for some cases as the one capture in the test. It would also be great to provide first a failing test that captures the failure we are trying to fix, that would help to come out with a fix for it.

memsharded:
I was concerned about this, I have tried to reproduce with a test. 
So far this test seems to be working fine, in case it is useful:
```python
def test_repeated_imports_same_name():
    c = TestClient()
        from conan import ConanFile
        import myfile
        from myfile import MyClass
        from myfile import number as number
        class Pkg(ConanFile):
            name = "{name}"
            version = "0.1"
            exports = "myfile.py"

            def package_info(self):
                self.output.info(f"NUMBER1: {{myfile.number}}")
                self.output.info(f"NUMBER2: {{MyClass.number}}")
                self.output.info(f"NUMBER3: {{number}}")
            """)
    myfile = textwrap.dedent("""\
        number = {number}

        class MyClass:
            number = {number}
        """)
    c.save({"pkg1/conanfile.py": conanfile.format(name="pkg1"),
            "pkg1/myfile.py": myfile.format(number="42"),
            "pkg2/conanfile.py": conanfile.format(name="pkg2"),
            "pkg2/myfile.py": myfile.format(number="123"),
            "app/conanfile.py": GenConanfile().with_requires("pkg1/0.1", "pkg2/0.1")})

    c.run("create pkg1")
    c.run("create pkg2")
    c.run("install app")
    assert "pkg1/0.1: NUMBER1: 42" in c.out
    assert "pkg1/0.1: NUMBER2: 42" in c.out
    assert "pkg1/0.1: NUMBER3: 42" in c.out
    assert "pkg2/0.1: NUMBER1: 123" in c.out
    assert "pkg2/0.1: NUMBER2: 123" in c.out
    assert "pkg2/0.1: NUMBER3: 123" in c.out
```

thorsten-klein:
I have added a test to reproduce the issue. 
It only occurs when you export python modules which are in any parent or sibling folder of the conanfile.py
This is the case if you have some base classes stored in your repository which are used and exported in multiple conanfiles.

conans/test/unittests/client/conanfile_loader_test.py
```diff
@@ -373,9 +373,43 @@ def append(data):
 
         try:
             sys.path.append(temp)
+            # myconanlogger module must not be cached between multiple conanfiles
```
memsharded: Everything that is not a local import, local to the current conanfile.py is a global import, and shouldn't be a different instance for every conanfile. Otherwise it means that every conanfile.py in the graph would have its own set of private imports of all Python standard library, and of every other Python modules and dependencies that are being imported.

The way it is designed is that the relative, "private" imports of the current conanfile.py are indeed private. But if something is imported from the global Python scope (because it is in ``sys.path``), then it should work the same way as Python, it should be the same instance, not separate. ``myconanlogger`` should be a unique instance, and this is exactly what the test was testing. Because otherwise, every possible Python import that relies on a global state, like a logger, would be broken.

So I am afraid this test cannot be modified, this change could cause quite a headache to a few users.

conans/test/integration/conanfile/conanfile_export_test.py
```diff
@@ -0,0 +1,94 @@
...
+                number = {number}
+            """)
+        
+        conanfile_main = GenConanfile().with_requires("pkg1/0.1")
+        conanfile_main._imports.append("import myfile")
+        conanfile_main._imports.append("from myfile import MyClass")
+        conanfile_main._imports.append("from myfile import number as number")
+        c.save({"pkg1/conanfile.py": conanfile.format(name="pkg1"),
+                "pkg1/myfile.py": myfile.format(number="42"),
+                "base/myfile.py": myfile.format(number="123"),
```
memsharded: This is not a sibling, this is being added to the Python global import ``sys.path``.

memsharded (commented):
It seems the main issue is mixing the global Python import system, finding things in ``sys.path``, and the local dynamic Conan import system.

I think it is not possible to mix them, whatever a ``conanfile.py`` is importing should be either local/dynamic or global, but having both at the same time will be problematic.

The test below cannot be modified, it was capturing a necessary behavior that is that Python imports found in sys.path must be global and shared, because otherwise certain behaviors of imports and modules that depends on some global state would be broken. We have hit this in the past with users using logging libraries, database libraries or remote/http libraries.

conans/test/unittests/client/conanfile_loader_test.py
thorsten-klein: So what would be your suggestion here if we really want to have the local import of pkg1 (instead of reusing the global import)? 
We have a monorepo with many conanfile.py and in a separate folder the python modules which should be exported together when doing `conan export`. 
How can we import the modules without a `sys.path.append`?

thorsten-klein:
Let me go one step back. The issue only occurs when doing a `conan install` with a local file. Then the export is not correctly done. So what about reacting on this? What is the easiest way to find out if it is a local conanfile? Would this make sense?

conans/test/unittests/client/conanfile_loader_test.py
memsharded: I am afraid that basically you can't. This is one of the major reasons ``python_requires`` was introduced long time ago.
The problem is the 
```
 from ...common.myfile import MyClass
E           ImportError: attempted relative import with no known parent package
```
Python doesn't let you to import another file in a sibling folder if you are not in a known package. But ``conanfile.py`` are not regular Python packages, so this doesn't apply and doesn't work.

So the options are:
- You have a copy of what you want to have together with the ``conanfile.py`` as a local import and sibling of ``conanfile.py``, and that file is ``exports`` together with the ``conanfile.py``.
- There is some global Python import that is in the Python path all the time (it could be a ``pip`` installed package, for example). Those files are not local to the recipe ``conanfile.py`` and they are not exported a copy with the ``conanfile.py``
- Use ``python_requires`` to have a common shared python code among several packages

thorsten-klein: > You have a copy of what you want to have together with the conanfile.py as a local import and sibling of conanfile.py, and that file is exports together with the conanfile.py.

This is no solution for us since we have stored our modules at a separate place and we cannot move them.

> Use python_requires to have a common shared python code among several packages

This is not possible for us since we work in a monorepo. We want to have reproducible builds, so our `python_requires` must be a full RREV. So `python_requires` is not very useful here since we would have to adapt each conanfile.py every time we make a change at the base class in order to apply the changes. So we will need something to take a base class from local file - and this is exactly what we do here.

What also comes into my mind: We could rename the classes during the export, so that there is no name-clash anymore with the `sys.path` modules. What do you think about this approach?
Unfortunately there is also some bug (Ref: #13362)

thorsten-klein:
This PR is obsolete.

PR #13287, opened 2023-03-01
remove --logger and traces

### LINKED ISSUE #13182: Review json logger in 2.0
### What is your suggestion?

There is a json logger in 2.0, which is not documented, and the interface not fully clear.
Let's review it for 2.0.1

memsharded:
Changelog: Fix: Remove ``--logger`` json logging and legacy traces.
Docs: https://github.com/conan-io/docs/pull/3050

Close https://github.com/conan-io/conan/issues/13182


AbrilRBS:
If anyone relies on this output we'd love to know your what your usecase was!

This should have been removed before 2.0 but it fell thru the cracks, and the reason has mainly to do with us thinking that it does not fall under Conan's umbrella to emit this kind of output, which seems like the responsibility of log aggregators, but we're open to discussing bringing this back in some form or another if anyone is impacted by this change

AbrilRBS (approved):
We'll need to regenerate most of the command reference docs, as they show `--logger`, but that's not too big of a priority

PR #13293, opened 2023-03-01
remove vcvars telemetry

### LINKED ISSUE #13196: [question] Possible Conan 2.0 issue w/ VCVars generator
Hi there,

Firstly, congratulations on the long-awaited Conan 2.0 release!

During some local experimenting, I have come across an intermittent issue where the `package` step seems to fail when running `conan create` for one of our recipes. It appears to not be able to copy the temporary build folder contents in the Conan cache to the final package folder e.g.

<details><summary><strong>Log Output</strong></summary>

```log
...
sdo-field-builder/0.1.0@shared: Exiting [SdoFieldBuilderConan].[build]
sdo-field-builder/0.1.0@shared: Package '723257509aee8a72faf021920c2874abc738e029' built
sdo-field-builder/0.1.0@shared: Build folder C:\Users\santhonisz\.conan2\p\t\sdo-f82806b5cba72d\b\.
sdo-field-builder/0.1.0@shared: Generating the package
sdo-field-builder/0.1.0@shared: Temporary package folder C:\Users\santhonisz\.conan2\p\t\sdo-f82806b5cba72d\p
sdo-field-builder/0.1.0@shared: Calling package()
sdo-field-builder/0.1.0@shared: package(): Packaged 1 '.exe' file: SdoFieldBuilder.exe
sdo-field-builder/0.1.0@shared: package(): Packaged 1 '.config' file: SdoFieldBuilder.exe.config
sdo-field-builder/0.1.0@shared: package(): Packaged 1 '.pdb' file: SdoFieldBuilder.pdb
sdo-field-builder/0.1.0@shared: Created package revision 7d7feccfc264dcd58e2f6cc82a123153
sdo-field-builder/0.1.0@shared: Package '723257509aee8a72faf021920c2874abc738e029' created
sdo-field-builder/0.1.0@shared: Full package reference: sdo-field-builder/0.1.0@shared#e8405177dfc3a5f5140beacf9169c826:723257509aee8a72faf021920c2874abc738e029#7d7feccfc264dcd58e2f6cc82a123153
ERROR: Couldn't move folder: C:\Users\santhonisz\.conan2\p\t\sdo-f82806b5cba72d->C:\Users\santhonisz\.conan2\p\sdo-f019a2dea9907b
Destination path 'C:\Users\santhonisz\.conan2\p\sdo-f019a2dea9907b\sdo-f82806b5cba72d' already exists
Folder might be busy or open. Close any app using it and retry.
```
</details>

<details><summary><strong>Host Profile</strong></summary>

```
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=192
os=Windows
```
</details>

I wont bore you with the details of how I determined it (long and complex), but the root cause appears to be down to a race condition caused by the deault behaviour of the `vsdevcmd.bat` file in Visual Studio 2019+ to trigger the sending of telemetry data to Microsoft. That (PowerShell) process is locking the tempory build folder for a brief amount of time and leads to the intermittent failure in the `package` method execution.

As you would know the VCVars generator creates the call to `vcvarsall.bat` in the generated `conanvcvars.bat` file, which ultimately executes the `vsdevcmd.bat` file. 

Investigation revealed that it's possible to turn off the sending of telemetry data by setting the `VSCMD_SKIP_SENDTELEMETRY` environment variable. I tried it out and that seems to completely resolve the problem, and make the initial startup of `vcvarsall.bat` much quicker as well.

I did have the thought to set the environment variable in the `[buildenv]` section of the profile, however that unfortunatlely doesn't work as the `conanbuildenv` file generated to set the environment variable isn't called in the `conanbuild.bat` until after the call to `conanvcvars.bat` e.g.

```
@echo off
call "%~dp0/conanvcvars.bat"
call "%~dp0/conanbuildenv-release-x86_64.bat"
```

So I guess the question becomes is this something that Conan should set automatically e.g. in the `conanvcvars.bat` file, or rather is it something that end users should be made aware of and it is their responsibility to set in their development environments etc?  

It would appear the same issue came up for `vcpkg`, albeit for performance reasons, and they agreed to make it part of their CLI tools:

https://github.com/microsoft/vcpkg/issues/10468
https://github.com/microsoft/vcpkg/issues/8756

Finally I should also add that I've only experienced this issue when developing on my local PC and not in our CI environments, which leads me to beleive there may be a performance element associated with when the problem occurs.

I look forward to your feedback!

memsharded:
Changelog: Fix: Remove ``vcvars.bat`` VS telemetry env-var, to avoid Conan hanging.
Docs: Omit

Lets see if this doesn't break anything.
Close https://github.com/conan-io/conan/issues/13196

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -142,6 +142,8 @@ class VCVars:
 
         content = textwrap.dedent("""\
             @echo off
+            set __VSCMD_ARG_NO_LOGO=1
+            set VSCMD_SKIP_SENDTELEMETRY=1
             {}
             """.format(vcvars))
         from conan.tools.env.environment import create_env_script
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -144,8 +144,9 @@ class VCVars:
             @echo off
             set __VSCMD_ARG_NO_LOGO=1
             set VSCMD_SKIP_SENDTELEMETRY=1
+            echo conanvcvars.bat: Activating environment Visual Studio {} - {} - vcvars_ver={}
             {}
-            """.format(vcvars))
+            """.format(vs_version, vcvarsarch, vcvars_ver, vcvars))
         from conan.tools.env.environment import create_env_script
         create_env_script(conanfile, content, CONAN_VCVARS_FILE, scope)
 
--- a/conans/test/functional/toolchains/microsoft/test_msbuild.py
+++ b/conans/test/functional/toolchains/microsoft/test_msbuild.py
@@ -342,10 +342,10 @@ def check_msvc_runtime_flag(msvc_version):
     assert "MSVC FLAG=MD!!" in client.out
 
 
-vs_versions = [{"vs_version": "15", "msvc_version": "191", "ide_year": "2017", "toolset": "v141"}]
+vs_versions = [{"vs_version": "15", "msvc_version": "191", "toolset": "v141"}]
 
 if "17" in tools_locations['visual_studio'] and not tools_locations['visual_studio']['17'].get('disabled', False):
-    vs_versions.append({"vs_version": "17", "msvc_version": "193", "ide_year": "2022", "toolset": "v143"})
+    vs_versions.append({"vs_version": "17", "msvc_version": "193", "toolset": "v143"})
 
 
 @parameterized_class(vs_versions)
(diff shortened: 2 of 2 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -142,8 +142,11 @@ def generate(self, scope="build"):
 
         content = textwrap.dedent("""\
             @echo off
+            set __VSCMD_ARG_NO_LOGO=1
```
AbrilRBS, line 145: This one has 6 hits on Google, besides the linked issues in the original report, I guess it disables a vs from outputting its name at some point, but I wonder why such a variable is so hidden for such a basic functionality 🤔

AbrilRBS (approved):
Looks like a good idea, and I don't see how disabling telemetry would break anyone (Let's hope noone is relying on the logo output tho)

PR #13297, opened 2023-03-02
Do not copy conan_server to conan folder when generating binaries with pyinstaller

czoido:
Changelog: omit
Docs: omit

When generating binaries with pyinstaller do not copy conan_server to conan folder.

### DIFF
--- a/pyinstaller.py
+++ b/pyinstaller.py
@@ -102,7 +102,6 @@ def pyinstall(source_folder):
         print("Unable to remove old folder", e)
 
     conan_path = os.path.join(source_folder, 'conans', 'conan.py')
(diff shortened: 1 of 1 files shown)


pyinstaller.py
```diff
@@ -139,10 +139,6 @@ def pyinstall(source_folder):
                     % (command, source_folder, conan_server_path, win_ver),
                     cwd=pyinstaller_path, shell=True)
 
-    conan_bin = os.path.join(pyinstaller_path, 'dist', 'conan')
```
memsharded, line 142: I think we want to remove the full ``subprocess.call`` above and the ``_run_bin()`` below?

czoido: let's remove them!

AbrilRBS (approved):

PR #13298, opened 2023-03-02
Ensure topics and provides are always serialized as lists

AbrilRBS:
Changelog: Bugfix: Ensure that `topics` are always serialized as lists.
Docs: omit
Changelog: Bugfix: Ensure that `provides` are always serialized as lists.
Docs: omit

This came about with cci's CI breaking due to a broken logic that assumed topics were always a list, which was not the case for some of the recipes. After hotfixing its code, this PR now aims to ensure that both `topics` and `provides` are always serialized as lists, even if the recipe defines only a string for those attributes.

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -115,11 +115,18 @@ class ConanFile:
 
     def serialize(self):
         result = {}
+
+        # Some fields are requested in the docs to be tuples, but we accept single strings too,
+        # ensure those fields are serialized as lists so things like CIs don't break unexpectedly,
+        # as it did for cci
+        def _ensure_iterable_if_needed(k, value):
+            return [value] if k in ("topics", "provides") and isinstance(value, str) else value
+
         for a in ("url", "license", "author", "description", "topics", "homepage", "build_policy",
                   "revision_mode", "provides", "deprecated", "win_bash"):
             v = getattr(self, a)
             if v is not None:
-                result[a] = v
+                result[a] = _ensure_iterable_if_needed(a, v)
         result["package_type"] = str(self.package_type)
         result["settings"] = self.settings.serialize()
         if hasattr(self, "python_requires"):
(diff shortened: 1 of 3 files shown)


conans/test/integration/command/info/info_test.py
```diff
@@ -55,6 +55,49 @@ class MyTest(ConanFile):
...
+                            class MyTest(ConanFile):
+                                name = "pkg"
+                                version = "0.2"
+                                provides = ("bar",)
+                                topics = ("foo",)
+                            """)
+        client.save({"conanfile.py": conanfile})
+        client.run("graph info . --format=json")
+        recipe = json.loads(client.stdout)["nodes"][0]
+        assert type(recipe["topics"]) == list
```
memsharded, line 74: type checks are redudant if the line below you are explicitly comparing against a list

memsharded (approved):

conans/test/integration/command/info/info_test.py
AbrilRBS: Yes, I know :) The assert is there just in case it ever fails, the diff error message is clearer

conans/model/conan_file.py
```diff
@@ -115,11 +115,18 @@ def __init__(self, display_name=""):
 
     def serialize(self):
         result = {}
+
+        # Some fields are requested in the docs to be tuples, but we accept single strings too,
+        # ensure those fields are serialized as lists so things like CIs don't break unexpectedly,
+        # as it did for cci
+        def _ensure_iterable_if_needed(k, value):
```
memsharded: On a second thought, lets change them to list at origin. ``provides`` is already converted to tuple just for checking in ``check_graph_provides()`` function. 
It will be cleaner if it is accessed at some point by deployers, users navigating ``self.dependencies``, etc

memsharded (changes requested):

memsharded (approved):

czoido (approved):

PR #13299, opened 2023-03-02
fix ux messages

### LINKED ISSUE #13295: [bug] A couple of UX improvements
### Environment details

* Operating System+version: any
* Compiler+version: any
* Conan version: 2.0.0
* Python version: any

### Steps to reproduce

1. `conan install` now apparently takes *only* the path to the dir containing conanfile. However, `conan install` without any arguments gives the following error message
```
ERROR: Please specify at least a path to a conanfile or a valid reference.
```
I was told that the reference still can be added via `conan install --requires=<reference>`, but that wasn't by any means obvious. I suggest having something like this
```
ERROR: Please specify a path to a conanfile. To install a reference, use --requires=<reference>.
```

2.  When dealing with a recipe that is not 2.0-ready, there’s an error line
```
If it is your recipe, check it is updated to 2.0
```
I am not a native English speaker, but it sounds to me that it should be “check *if* it is updated” or “check *that* it is updated”

3. ~I have a perfectly valid `conanfile.txt` from 1.X~ already fixed
```
❯ cat conanfile.txt
[requires]
sdl/[^2.26.1]

[options]
sdl:x11=False
```
On 2.0, `conan install` fails with
```
ERROR: Error while parsing [options] in conanfile
Options should be specified as 'pkg:option=value'
```
It took me a couple of minutes to understand that in Conan 2.0 the syntax should be `sdl/*:x11=False`.
While I totally understand that for Conan devs this message looks totally logical, I wish it was clearer for people who migrated from 1.X

### Logs

_No response_

memsharded:
Changelog: Fix: Minor improvements to ``conan install`` and 2.0-readiness error messages.
Docs: Omit

Close https://github.com/conan-io/conan/issues/13295

### DIFF
--- a/conan/cli/args.py
+++ b/conan/cli/args.py
@@ -1,5 +1,5 @@
 from conan.cli.command import OnceArgument
-
+from conans.errors import ConanException
 
 _help_build_policies = '''Optional, specify which packages to build from source. Combining multiple
     '--build' options on one command line is allowed.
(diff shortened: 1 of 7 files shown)


AbrilRBS (approved):
Nice!

PR #13321, opened 2023-03-03
Add  `core.version_ranges:resolve_prereleases` conf

AbrilRBS:
Changelog: Omit
Docs: Omit

Future changelog: Feature: Add `core.version_ranges:resolve_prereleases` conf to control whether version ranges can resolve to prerelease versions

I still have some questions regarding the implementation of the feature itself,
but this is now at a point where it makes sense to open the PR,
even though there are still places that need updating (And so tests are failing)

Specially I'd need to discuss if changing the VersionRange api is ok,
as this removes the `__contains__` function in favour of a `contains` method,
to be able to pass it extra arguments, which can't be done with the `in` operator.

The logic for the conf is:
 - `True`: Always allows to resolve prerelease versions, whatever the expression says
 - `False`: Never allows prerelease versions to be resolved
 - `None`: Listens to whatever the the range expression says

The open questions are:
 - Is `None` the best default for an unspecified conf? My guess it that yes, but I better make sure
 - Is the current implementation even ok? I had a good look into how this could be done, and this seemed to me like the only viable alternative, but happy to hear otherwise.
 - Is there a better name for the conf?: A: Yes. `core.version_ranges:resolve_prereleases` (Was  `core.ranges_resolve_prereleases`)

### DIFF
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -118,7 +118,7 @@ class DepsGraphBuilder(object):
 
         elif prev_version_range is not None:
             # TODO: Check user/channel conflicts first
-            if not prev_version_range.contains(require.ref.version):
+            if not prev_version_range.contains(require.ref.version, resolve_prereleases):
                 raise GraphError.conflict(node, require, prev_node, prev_require, base_previous)
         else:
             def _conflicting_refs(ref1, ref2):
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -39,11 +39,12 @@ class DepsGraphBuilder(object):
         root_node.conanfile.settings_target = None
 
         self._prepare_node(root_node, profile_host, profile_build, Options())
-        self._initialize_requires(root_node, dep_graph, graph_lock)
+        resolve_prereleases = self._cache.new_config.get('core.ranges_resolve_prereleases',
+                                                         default=None, check_type=bool)
+        self._initialize_requires(root_node, dep_graph, graph_lock, resolve_prereleases)
         dep_graph.add_node(root_node)
 
         open_requires = deque((r, root_node) for r in root_node.conanfile.requires.values())
-        resolve_prereleases = self._cache.new_config.get('core.ranges_resolve_prereleases', default=None, check_type=bool)
         try:
             while open_requires:
                 # Fetch the first waiting to be expanded (depth-first)
@@ -53,7 +54,7 @@ class DepsGraphBuilder(object):
                 new_node = self._expand_require(require, node, dep_graph, profile_host,
                                                 profile_build, graph_lock, resolve_prereleases)
                 if new_node:
-                    self._initialize_requires(new_node, dep_graph, graph_lock)
+                    self._initialize_requires(new_node, dep_graph, graph_lock, resolve_prereleases)
                     open_requires.extendleft((r, new_node)
                                              for r in reversed(new_node.conanfile.requires.values()))
             self._remove_overrides(dep_graph)
@@ -159,12 +160,12 @@ class DepsGraphBuilder(object):
                     node.conanfile.requires.tool_require(str(tool_require),
                                                          raise_if_duplicated=False)
 
-    def _initialize_requires(self, node, graph, graph_lock):
+    def _initialize_requires(self, node, graph, graph_lock, resolve_prereleases):
         # Introduce the current requires to define overrides
         # This is the first pass over one recipe requires
         if graph_lock is not None:
             for require in node.conanfile.requires.values():
-                graph_lock.resolve_locked(node, require)
+                graph_lock.resolve_locked(node, require, resolve_prereleases)
 
         for require in node.conanfile.requires.values():
             self._resolve_alias(node, require, graph)
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -39,7 +39,7 @@ class DepsGraphBuilder(object):
         root_node.conanfile.settings_target = None
 
         self._prepare_node(root_node, profile_host, profile_build, Options())
-        resolve_prereleases = self._cache.new_config.get('core.ranges_resolve_prereleases',
+        resolve_prereleases = self._cache.new_config.get('core.version_ranges:resolve_prereleases',
                                                          default=None, check_type=bool)
         self._initialize_requires(root_node, dep_graph, graph_lock, resolve_prereleases)
         dep_graph.add_node(root_node)
--- a/conans/client/graph/graph_builder.py
+++ b/conans/client/graph/graph_builder.py
@@ -25,6 +25,9 @@ class DepsGraphBuilder(object):
         self._remotes = remotes  # TODO: pass as arg to load_graph()
         self._update = update
         self._check_update = check_update
+        self._resolve_prereleases = self._cache.new_config.get('core.version_ranges:'
+                                                                'resolve_prereleases',
+                                                                default=None, check_type=bool)
 
     def load_graph(self, root_node, profile_host, profile_build, graph_lock=None):
         assert profile_host is not None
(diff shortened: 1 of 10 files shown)


conans/client/graph/range_resolver.py
```diff
@@ -25,12 +25,16 @@ def resolve(self, require, base_conanref, remotes, update):
             require.ref = previous_ref
             return
 
+        # If the conf is None, do whatever the range specifies
+        # True: Force prerelease resolving, False: Force no prerelease resolving
+        resolve_prereleases = self._cache.new_config.get('core.ranges_resolve_prereleases', default=None, check_type=bool)
```
memsharded: confs should be ``xxx.yyy:conf=value``

memsharded (commented):
I know it is not as clean, but a global class attribute ``VersionRange.accept_prereleases`` would simplify a lot the implementation and avoid having to propagate the parameter everywhere? Most likely better to do it right as you did, but seemed a bit painful to see so many changes to deliver 1 bit of information to the right place.

conans/client/graph/range_resolver.py
AbrilRBS: `core.version_ranges:resolve_prereleases` it is then

memsharded: Still not xxx.yyy:conf=value, just ``core.ranges_resolve_prereleases``, missing ":"

memsharded: I think we will definitely need some command line ``-global-conf`` argument to be able to define core values like this one

AbrilRBS: Hahaha, I changed the PR title, but forgot about the code itself 🤦

AbrilRBS:
Would be great to have an example in the docu telling the story for this feature (Or at the very least tests for CI!)

conans/client/graph/range_resolver.py
```diff
@@ -25,12 +25,16 @@ def resolve(self, require, base_conanref, remotes, update):
             require.ref = previous_ref
             return
 
+        # If the conf is None, do whatever the range specifies
+        # True: Force prerelease resolving, False: Force no prerelease resolving
+        resolve_prereleases = self._cache.new_config.get('core.version_ranges:resolve_prereleases', default=None, check_type=bool)
```
memsharded: Lets get the ``self._resolve_prereleases`` as attribute of the ``RangeResolver`` in the constructor too, in the same way as the ``DepsGraphBuilder``, so it is done only once, and not once per resolved range.

conans/model/version_range.py
```diff
@@ -96,9 +101,10 @@ def __init__(self, expression):
     def __str__(self):
         return self._expression
 
-    def __contains__(self, version):
+    def contains(self, version, conf_resolve_prerelease=None):
```
memsharded: I'd say this default is not the safest, we might skip some ``version in range`` occurrence that needs the conf passed, and we are not passing it. I'd suggest dropping the ``=None`` default and having this forced, even if it needs changing more tests.

conans/model/version_range.py
```diff
@@ -56,9 +56,14 @@ def first_non_zero(main):
         else:
             return [_Condition(operator, Version(version))]
 
-    def valid(self, version):
+    def valid(self, version, conf_resolve_prepreleases=None):
```
memsharded: ``def valid`` happens to be used only internally in ``VersionRange``, it can be made private (I was concerned about the ``=None`` default too)

conans/model/graph_lock.py
```diff
@@ -201,12 +201,12 @@ def serialize(self):
             result["alias"] = {repr(k): repr(v) for k, v in self._alias.items()}
         return result
 
-    def resolve_locked(self, node, require):
+    def resolve_locked(self, node, require, resolve_prereleases=None):
```
memsharded: Better drop the ``=None`` default?

conans/client/graph/graph_builder.py
```diff
@@ -157,12 +160,12 @@ def _prepare_node(node, profile_host, profile_build, down_options):
                     node.conanfile.requires.tool_require(str(tool_require),
                                                          raise_if_duplicated=False)
 
-    def _initialize_requires(self, node, graph, graph_lock):
+    def _initialize_requires(self, node, graph, graph_lock, resolve_prereleases):
```
memsharded: I think we might have another approach here: 
if we define here for all ``requires`` we define that such a ``require`` allows pre-releases, and that is automatically passed to the ``VersionRange`` when requested via property, then the number of modifications necessary might be way less?

Just an idea, I am still not super happy with this PR, it is likely that if we remove the ``=None`` defaults, it would require even more changes.

conans/model/version_range.py
AbrilRBS: This already breaks the `version in range` syntax, so it would be failing because it wouldn't be able to find the `__contains__` operator, but I understand your point, having so many defaults smells a bit :)

memsharded: Yeah, I meant that we had some ``if range.contains(version)`` without really providing the necessary conf there, which seems mandatory.

conans/client/graph/graph_builder.py
```diff
@@ -99,22 +103,22 @@ def _expand_require(self, require, node, graph, profile_host, profile_build, gra
...
+                if version_range.contains(prev_ref.version, resolve_prereleases):
                     require.ref = prev_ref
                 else:
                     raise GraphError.conflict(node, require, prev_node, prev_require, base_previous)
 
         elif prev_version_range is not None:
-            # TODO: CHeck user/channel conflicts first
-            if require.ref.version not in prev_version_range:
+            # TODO: Check user/channel conflicts first
+            if not prev_version_range.contains(require.ref.version):
```
memsharded: why this doesn't crash if not provided the arg? the intention is that it is mandatory to always pass the "resolve_prereleases" arg.

conans/test/unittests/model/version/test_version_range.py
```diff
@@ -47,10 +47,45 @@ def test_range(version_range, conditions, versions_in, versions_out):
...
     for v in versions_in:
-        assert Version(v) in r
+        assert r.contains(Version(v), None)
 
     for v in versions_out:
-        assert Version(v) not in r
+        assert not r.contains(Version(v), None)
+
+
+@pytest.mark.parametrize("version_range, resolve_prereleases, versions_in, versions_out", [
```
memsharded, line 56: Do you know that multiple ``pytest.mark.parametrize`` one on top of the other do the cartesian product? you can reduce these lines 3x with it.
```python
@pytest.mark.parametrize("resolve_prereleases", [True, False, None])
@pytest.mark.parametrize("version_range, versions_in, versions_out", [
```

memsharded: Furthermore, you can also factorize the ``version_range`` and ``versions_in/out``, and it will be even shorter

conans/test/integration/graph/version_ranges/version_range_conf.py
```diff
@@ -0,0 +1,108 @@
...
+    tc.run("create base/conanfile.py --version=2.5.0-pre")
+
+    conanfilev1 = textwrap.dedent("""
+    from conan import ConanFile
+
+    class Package(ConanFile):
+        name = "pkg"
+        version = "1.0"
+        requires = "base/[>1 <2]"
+    """)
```
memsharded:
```suggestion
    conanfilev1 =GenConanfile("pkg", "1.0").with_requires("base/[>1 <2]")
```

conans/test/integration/graph/version_ranges/version_range_conf.py
```diff
@@ -0,0 +1,108 @@
...
+
+    tc.save({"base/conanfile.py": base})
+    tc.run("create base/conanfile.py --version=1.5.1")
+    tc.run("create base/conanfile.py --version=2.5.0-pre")
+
+    conanfilev1 = textwrap.dedent("""
+    from conan import ConanFile
+
+    class Package(ConanFile):
```
memsharded: same as above with ``GenConanfile``

conans/client/graph/graph_builder.py
```diff
@@ -215,7 +219,7 @@ def _resolve_recipe(self, ref, graph_lock):
         return new_ref, dep_conanfile, recipe_status, remote
 
     @staticmethod
-    def _resolved_system_tool(node, require, profile_build, profile_host):
+    def _resolved_system_tool(node, require, profile_build, profile_host, resolve_prereleases):
```
memsharded, line 220: Maybe it would be more idiomatic to remove the ``staticmethod`` and use internally directly ``self._resolve_prereleases``, but I am fine with keeping it static and making evident what is passed as argument.

memsharded (approved):

Issue #13325, opened 2023-03-04
[bug] qcc cannot calculate dep graph because it fails to produce a list of compatible cppstd per compiler verison


tannerbitz:
### Environment details

* Operating System+version:   Ubuntu 20.04
* Compiler+version:  qcc 8.3.0
* Conan version:  2.0.1
* Python version:   3.8.10

### command
conan build -pr:b linux64 -pr:b qnx64 -s build_type=Release --version <package-ver> <conanfile.py location>

### qnx64 profile
[settings]
os=Neutrino
os.version=7.1
arch=x86_64
compiler=qcc
compiler.version=8.3
compiler.libcxx=cxx
compiler.cppstd=17

[conf]
tools.build:compiler_executables={'c':'qcc', 'cpp': 'q++'}
tools.build:cflags=["-Vgcc_ntox86_64","-fpic", "-D__USESRCVERSION", "-D_QNX_SOURCE", "-DQNX"]
tools.build:cxxflags=["-Vgcc_ntox86_64", "-Y _cxx", "-std=c++17", "-Wc,-isysroot,{{os.getenv('QNX_TARGET')}}", "-fpic", "-D__USESRCVERSION", "-D_QNX_SOURCE", "-DQNX", "-Wno-deprecated-declarations", "-Wno-unused-parameter", "-Wno-unused-variable", "-Wno-ignored-attributes"]
tools.build:sharedlinkflags=["-Wl,--build-id=md5"]
tools.cmake.cmaketoolchain:generator=Eclipse CDT4 - Unix Makefiles
tools.cmake.cmaketoolchain:system_name=QNX
tools.cmake.cmaketoolchain:system_processor=x86_64
tools.cmake.cmaketoolchain:system_version=7.1.0
tools.gnu:host_triplet=x86_64-pc-nto-qnx7.1.0
tools.cmake.cmaketoolchain:user_toolchain=["{{ os.path.join(profile_dir, "..", "toolchains", "qnx_toolchain_preamble.cmake") }}"]

### linux64 profile
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu14
compiler.libcxx=libstdc++11
compiler.version=9
os=Linux

[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True

### Logs

```
======== Computing necessary packages ========
Traceback (most recent call last):
[...]

PR #13328, opened 2023-03-05
Nicer error in the compatibility plugin with custom compilers

AbrilRBS:
Changelog: Bugfix: Nicer error in the compatibility plugin with custom compilers.
Docs: Omit

After seeing https://github.com/conan-io/conan/issues/13325, figured I'd improve the UX of the raised error. My first idea was to make the function return an empty list if the compiler was not known, but that would change current behaviour and so I decided to at least improve the generated exception, happy to discuss if going that way would work or not (Mostly so I get to learn about it a bit more :))

Also, I noticed that the plugin code itself is _not_ currently directly tested and we might want to look into improving that, but I tried and got hit a wall because of how the function is defined in the code, it makes it hard to test. We could get it to run by running the same code that the `test_migrations` code does, but I was not able to come up with a test that exercised the compatibility feature, so any help is welcomed!


memsharded:
> After seeing https://github.com/conan-io/conan/issues/13325, figured I'd improve the UX of the raised error. My first idea was to make the function return an empty list if the compiler was not known, but that would change current behaviour and so I decided to at least improve the generated exception, happy to discuss if going that way would work or not (Mostly so I get to learn about it a bit more :))

Yes, definitely a better error until we discuss what would be a better default (taking into account that there will be users adding their own compilers to ``settings.yml``, that would hit this error).

> Also, I noticed that the plugin code itself is not currently directly tested and we might want to look into improving that, but I tried and got hit a wall because of how the function is defined in the code, it makes it hard to test. We could get it to run by running the same code that the test_migrations code does, but I was not able to come up with a test that exercised the compatibility feature, so any help is welcomed!

I think there are tests like ``TestCacheCompatibles`` that validate behavior of ``compatibility.py``, isn't that enough?

conans/client/graph/compatibility.py
```diff
@@ -36,6 +37,8 @@ def cppstd_compat(conanfile):
         return []
     base = dict(conanfile.settings.values_list)
     cppstd_possible_values = supported_cppstd(conanfile)
+    if cppstd_possible_values is None:
```
memsharded, line 40: Lets change it to return [] and warning

memsharded (changes requested):

memsharded (approved):
I think a unit test would be great.

AbrilRBS:
Done, I have added a case where it all works just to make sure

PR #13330, opened 2023-03-05
initial metadata proposal

memsharded:
Changelog: Omit
Docs: Omit

This would be the 2nd part of the proposal, isolated from many other things. Just the creation of metadata from both recipes and hooks

The full proposal would be:
Very preliminary proposal for metadata, leveraging the infra that package signing introduced:

- files that do not belong to the package/recipe, are not listed in conanmanifest.txt
- Files are uploaded if existing, but not downloaded by default
- They are not located in the export, source or package folder

Can be used for:
- Build logs
- Storing test executables
- Storing test_package


conans/test/integration/editable/test_editable_python_requires.py
```diff
@@ -0,0 +1,41 @@
...
+        number = 42
+
+        class Dep(ConanFile):
+            name = "dep"
+            version = "1.0"
+        """)
+    pkg = textwrap.dedent("""
+        from conan import ConanFile
+
+        number = 42
```
AbrilRBS: Whi is this number = 42 needed here for this test?

AbrilRBS (approved):

conans/test/integration/editable/test_editable_python_requires.py
memsharded: sorry, this test slipped here, not belongs here, I will remove it.

czoido (approved):

AbrilRBS (approved):

maksim-petukhou:
Will this feature help with https://github.com/conan-io/conan-center-index/issues/1982?

memsharded:
This feature will definitely allow to store .pdbs for Windows debug builds.
However, this doesn't mean that ConanCenter will do it and store the pdbs. .pdbs are massive, and in practice they are not used by the vast majority of ConanCenter users, so it would be a huge cost for little value. For this use case, the recommended approach would be to build from sources and store the pdbs in your own server, if you want to leverage this new feature, it could be a good fit.

PR #13354, opened 2023-03-07
Add `conan config show <conf>` command

AbrilRBS:
Changelog: Feature: Add `conan config show <conf>` command.
Docs: https://github.com/conan-io/docs/pull/3091

Let's you get the value for any given configuration. This is the simplest implementation, which does not allow you to input additional profiles nor conf values in the cli, so for now just works from the global.conf, but can be extended in the future:

 - Allow to take into account profiles and conf values supplied by the CLI
 - Pass a pattern to match the confs against and show the values of every matching ones

### DIFF
--- a/conan/cli/commands/config.py
+++ b/conan/cli/commands/config.py
@@ -66,3 +66,13 @@ def config_list(conan_api, parser, subparser, *args):
     """
     parser.parse_args(*args)
     return BUILT_IN_CONFS
+
+
+@conan_subcommand(formatters={"text": list_text_formatter, "json": default_json_formatter})
+def config_get(conan_api, parser, subparser, *args):
+    """
+    Get the value of the specified conf
+    """
+    subparser.add_argument('conf', help='Conf item for which to query its value')
+    args = parser.parse_args(*args)
+    return {args.conf: conan_api.config.get(args.conf)}
--- a/conan/cli/commands/config.py
+++ b/conan/cli/commands/config.py
@@ -69,10 +69,11 @@ def config_list(conan_api, parser, subparser, *args):
 
 
 @conan_subcommand(formatters={"text": list_text_formatter, "json": default_json_formatter})
-def config_get(conan_api, parser, subparser, *args):
+def config_show(conan_api, parser, subparser, *args):
     """
     Get the value of the specified conf
     """
-    subparser.add_argument('conf', help='Conf item for which to query its value')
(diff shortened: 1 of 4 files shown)


conan/cli/commands/config.py
```diff
@@ -66,3 +66,13 @@ def config_list(conan_api, parser, subparser, *args):
...
     parser.parse_args(*args)
     return BUILT_IN_CONFS
+
+
+@conan_subcommand(formatters={"text": list_text_formatter, "json": default_json_formatter})
+def config_get(conan_api, parser, subparser, *args):
+    """
+    Get the value of the specified conf
+    """
+    subparser.add_argument('conf', help='Conf item for which to query its value')
```
memsharded: Won't you make it a pattern, so:
```bash
conan config get *cmake*  # list all confs related to cmake
conan config get * # list the whole result of rendering global.conf
```
I think that might be quite useful to see with a single command your whole evaluated global.conf?

AbrilRBS: Only 1 UX question regarding this remains. Do you think confs for which the user has not set any value should be reported?

memsharded: Nop, only defined confs in ``global.conf``

conans/model/conf.py
```diff
@@ -475,12 +479,20 @@ def __bool__(self):
...
+        Get the value of the conf name requested and convert it to the [type]-like passed.
         """
         pattern, name = self._split_pattern_name(conf_name)
         return self._pattern_confs.get(pattern, Conf()).get(name, default=default,
                                                             check_type=check_type)
 
+    def show(self, pattern):
+        result = {}
+
+        for p in self._pattern_confs.values():
```
memsharded: This is dropping the patterns if they exist, I think they should be shown, they are shown for options, for example, like ``*:shared=True`` is shown in the ``conan profile show`` command.

conans/test/integration/command/config_test.py
```diff
@@ -108,3 +108,28 @@ def _assert_config_not_exists(path):
     _assert_config_exists("foo")
 
     os.listdir(tc.current_folder)
+
+
+def test_config_show():
+    globalconf = textwrap.dedent("""
+    tools.build:jobs=42
```
memsharded, line 115: Add some pattern to the test too

memsharded (approved):

PR #13365, opened 2023-03-07
Avoid raising fatal exception if custom command under a folder breaks

AbrilRBS:
Changelog: Bugfix: Avoid raising fatal exceptions for malformed custom commands.
Docs: Omit

Commands under a prefix were not being guarded against fatal exceptions, so an error in those meant you needed to remove the file manually to be able to work with Conan again


memsharded (approved):

memsharded (commented):
Maybe add a unit test?

AbrilRBS:
That's the idea yes :)

AbrilRBS:
Done

memsharded (approved):

PR #13385, opened 2023-03-08
[test] Adjust info folders test to be compatible with Conan 2.0

uilianries:
The commando `info --path` no longer exists and should be replaced by `graph info` instead.

Plus, source folder, build folder and package folder are not showed by `graph info` command.

Renamed to test_ suffix to follow pytest pattern.


conans/test/integration/command/info/test_info_folders.py
```diff
@@ -0,0 +1,150 @@
...
+deps_txt_file = """
+[requires]
+package2/0.2.0@user/testing
+"""
+
+user = "user"
+channel = "testing"
+user_channel = f"{user}/{channel}"
+reference1 = f"package/0.1.0@{user_channel}"
+reference2 = f"package2/0.2.0@{user_channel}"
```
memsharded: But why have globals? They can be safely defined inside the tests, it is more clear

uilianries: Done, removed

PR #13386, opened 2023-03-08
Create summary of cached content

AbrilRBS:
Changelog: Feature: Create summary of cached content.
Docs: Omit

Co-Authored-By: @memsharded 

Initial work towards the sources backup feature, this PR aims to implement a simple "summary.json" for sources downloaded by a conanfile, which stores info about which reference downloaded which URL, relying in the sha256 checksum as the key.

It also changes the cache layout a bit by moving the mentioned downloaded files by the conanfile to their own `s/` subfolder, and the rest to `c/`

Omitted docs because it only makes sense to document this once the whole feature lands

### DIFF
--- a/conans/client/downloaders/caching_file_downloader.py
+++ b/conans/client/downloaders/caching_file_downloader.py
@@ -55,7 +55,8 @@ class CachingFileDownloader:
                 h = sha256
                 sources_cache = True
             else:
-                ConanOutput().warning("Use sha256 as file checksums for downloaded sources")
+                ConanOutput()\
+                    .warning("Expected sha256 to be used as file checksums for downloaded sources")
         if h is None:
             h = self._get_hash(url, md5, sha1, sha256)
 
--- a/conans/client/downloaders/caching_file_downloader.py
+++ b/conans/client/downloaders/caching_file_downloader.py
@@ -1,3 +1,4 @@
+import json
 import os
 import shutil
 from contextlib import contextmanager
@@ -5,7 +6,8 @@ from threading import Lock
 
 from conan.api.output import ConanOutput
 from conans.client.downloaders.file_downloader import FileDownloader
-from conans.util.files import mkdir, set_dirty_context_manager, remove_if_dirty
+from conans.model.conan_file import ConanFile
+from conans.util.files import mkdir, set_dirty_context_manager, remove_if_dirty, load, save
 from conans.util.locks import SimpleLock
 from conans.util.sha import sha256 as compute_sha256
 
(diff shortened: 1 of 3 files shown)


conans/client/downloaders/caching_file_downloader.py
```diff
@@ -43,17 +47,49 @@ def _lock(self, lock_id):
...
+                if os.path.exists(summary_path):
+                    summary = json.loads(load(summary_path))
+                else:
+                    summary = {}
+
+                try:
+                    summary_key = conanfile.ref.repr_notime()
+                except AttributeError:
+                    if conanfile.recipe_folder is not None:
+                        summary_key = conanfile.recipe_folder
```
memsharded: Recipe folder might be also variable, for different machines, probably always "unknown"?

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -43,17 +47,49 @@ def _lock(self, lock_id):
...
-        h = self._get_hash(url, md5, sha1, sha256)
+    def _caching_download(self, url, file_path, md5, sha1, sha256, conanfile: ConanFile, **kwargs):
+        sources_cache = False
+        h = None
+        if conanfile is not None:
+            if sha256:
+                h = sha256
+                sources_cache = True
+            else:
+                ConanOutput()\
```
memsharded, line 57: Note for ourselves: we need modes of behavior for the caching, like being able to raise if the thing is not cached, to raise if it cannot be cached because of no sha256, etc. For future work.

memsharded (commented):
Looking good. Not sure what would be the best flow, if we should merge, or we should target "develop2" for 2.1, or how to approach it. Ideally, this part of the cache could start to be used asap I guess

conans/test/integration/cache/download_cache_test.py
```diff
@@ -164,3 +167,58 @@ def test_download_relative_error(self):
...
+            assert len(content["unknown"]) == 2
+
+            # Ensure the cache is working and we didn't break anything by modifying the summary
+            client.run("source .")
+            assert "Downloading file" not in client.out
+
+            client.run("create . --format=json")
+            content = json.loads(load(os.path.join(tmp_folder, "s", sha256 + ".json")))
+            out = json.loads(client.stdout)
+            assert content[out["graph"]["nodes"][1]["ref"]] == ["http://localhost.mirror:5000/myfile.txt"]
```
AbrilRBS: Just checking if this is indeed the correct check using the json output or if there's another way @memsharded?

AbrilRBS (commented):
Changed to always use "unknown" as an origin if the ref is None, and added a test path for when the ref is actually there, but have a question about the usage of the `--format=json` output

conans/test/integration/cache/download_cache_test.py
memsharded: No, better remove the ``out`` part, not necessary.
If you want the exported ref, ``c.exported_recipe_revision()`` will get it

AbrilRBS: Got it, thanks!

PR #13403, opened 2023-03-09
ISSUE-13392: Add build scripts folder to 'autoreconf' method of 'Auto…

### LINKED ISSUE #13392: [bug] Can't specify Autotools.autoreconf() location to the build directory
### Environment details

* Operating System+version: Linux
* Compiler+version: gcc 8.5
* Conan version: 2.0.1
* Python version: 3.6.8

### Steps to reproduce

I tried to run autoreconf on my configure.ac file in the build folder. However, even doing things like the following:

```
from conan.tools.gnu import Autotools

class MyPkg(ConanFile):

   def build(self):
       with chdir(self, self.build_folder):
            autotools = Autotools(self)
            autotools.autoreconf()
            ...
```
still results in `Autotools.autoreconf` looking for the `configure.ac` file elsewhere. At least `Autotools.configure` allows you to specify the `build_script_folder` location, but `Autotools.autoreconf` does not. Looking at the source code, it appears to be hardcoded to look for it in the recipe's source folder...

Seems like the user should be able to decide without having to change something as fundamental as the recipe's source folder location...

### Logs

_No response_

gabeblack:
Changelog: Feature: Adds an optional "build_script_folder" argument to the `autoreconf` method of the `Autotools` class. It mirrors the same argument and behavior of the `configure` method of the same class. That is, it allows one to override where the tool is run (by default it runs in the `source_folder`.

Docs: Omit

Close #13392
No PR -- the docs are automatically updated from the mark-up in the method documentation.

### DIFF
--- a/conan/tools/gnu/autotools.py
+++ b/conan/tools/gnu/autotools.py
@@ -98,16 +98,20 @@ class Autotools(object):
             args.insert(0, "DESTDIR={}".format(unix_path(self._conanfile, self._conanfile.package_folder)))
         self.make(target=target, args=args)
 
-    def autoreconf(self, args=None):
+    def autoreconf(self, build_script_folder=None, args=None):
         """
         Call ``autoreconf``
 
         :param args: (Optional, Defaulted to ``None``): List of arguments to use for the
                      ``autoreconf`` call.
+        :param build_script_folder: Subfolder where the `configure` script is located. If not specified
+                                    conanfile.source_folder is used.
         """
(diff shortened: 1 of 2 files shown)


memsharded (approved):
Excellent many thanks!

memsharded:
And merged, you are now a contributor to Conan, thank you :) 

It has been merged to ``develop2`` branch, but I will merge it to ``release/2.0``, so it is released in next Conan 2.0.2

PR #13411, opened 2023-03-10
Restore 2.0.1 version in release/2.0 branch

czoido:
Changelog: omit
Docs: omit

Somehow the -dev version finished in the release branch. Normally the version in release branch is updated right before releasing, so the next change in this branch should be to 2.0.2 before release. We could also use 2.0.2-dev from the moment we know that there could be a 2.0.2.


memsharded (approved):

memsharded:
It was probably me doing some of the merges from develop2, sorry

Issue #12239, opened 2022-10-04
[feature] Allow users to enable verbose Makefile in CMakeToolchain


SpaceIm:
- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

Same feature request than https://github.com/conan-io/conan/issues/9772, but for `CMakeToolchain`.

The idea would be to inject `set(CMAKE_VERBOSE_MAKEFILE ON)` in CMakeToolchain if a specific conf is enabled.

Maybe a common conf could be shared to enable verbosity to all build systems supporting this kind of feature.

memsharded:
Existing ones:

"tools.microsoft.msbuild:verbosity": "Verbosity level for MSBuild: 'Quiet', 'Minimal', 'Normal', 'Detailed', 'Diagnostic'",
"tools.apple.xcodebuild:verbosity": "Verbosity level for xcodebuild: 'verbose' or 'quiet",

PR #13428, opened 2023-03-13
Add global verbosity conf `tools.build:verbosity`

AbrilRBS:
Changelog: Bugfix: Add global verbosity conf `tools.build:verbosity` instead of individual ones.
Docs: https://github.com/conan-io/docs/pull/3107

Adds `tools.build:verbosity` which tries to unify the verbosity levels of the build systems.

Right now XCode and MSBuild

### DIFF
--- a/conan/tools/apple/xcodebuild.py
+++ b/conan/tools/apple/xcodebuild.py
@@ -14,14 +14,22 @@ class XcodeBuild(object):
     def _verbosity(self):
         verbosity = self._conanfile.conf.get("tools.build:verbosity")
         if verbosity:
-            if verbosity not in ("Quiet", "Minimal", "Normal", "Detailed", "Diagnostic"):
+            if verbosity not in ("quiet", "error", "warning", "notice", "status", "verbose",
+                                 "normal", "debug", "v", "trace", "vv"):
                 raise ConanException(f"Value '{verbosity}' for 'tools.build:verbosity' is not valid")
             else:
-                verbosity = {"Quiet": "quiet",
-                             "Minimal": "quiet",
-                             "Normal": None,
-                             "Detailed": "verbose",
-                             "Diagnostic": "verbose"}.get(verbosity)
+                # quiet, nothing, verbose
+                verbosity = {"quiet": "quiet",
+                             "error": "quiet",
+                             "warning": "quiet",
+                             "notice": "quiet",
+                             "status": None,
+                             "verbose": None,
+                             "normal": None,
+                             "debug": "verbose",
+                             "v": "verbose",
+                             "trace": "verbose",
+                             "vv": "verbose"}.get(verbosity)
         if verbosity is not None:
             return "-{}".format(verbosity)
         return ""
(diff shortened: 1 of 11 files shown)


memsharded:
> As for CMake, there's CMAKE_VERBOSE_MAKEFILE (Requested in https://github.com/conan-io/conan/issues/12239) to control the build verbosity itself, but is set to FALSE by default by the project() so I don't think adding it to the toolchain would actually work? I still need to test locally

It can be added to the ``presets`` file too via CMakeToolchain, and it will be passed in command line arg

conan/tools/apple/xcodebuild.py
```diff
@@ -12,7 +12,16 @@ def __init__(self, conanfile):
 
     @property
     def _verbosity(self):
-        verbosity = self._conanfile.conf.get("tools.apple.xcodebuild:verbosity", default="", check_type=str)
+        verbosity = self._conanfile.conf.get("tools.build.verbosity")
```
memsharded:
```suggestion
        verbosity = self._conanfile.conf.get("tools.build:verbosity")
```

conan/tools/apple/xcodebuild.py
```diff
@@ -12,7 +12,16 @@ def __init__(self, conanfile):
...
+        if verbosity:
+            if verbosity not in ("Quiet", "Minimal", "Normal", "Detailed", "Diagnostic"):
+                raise ConanException(f"Unknown value '{verbosity}' for 'tools.build.verbosity'")
+            else:
+                verbosity = {"Quiet": "quiet",
```
memsharded: Probably this is not the best for multi-build-system, because if you provide for example "Minimal" it won't do anything. I think it would be necessary to map all possible input values.

AbrilRBS: I see what you mean, I've centered the mapping to only have normal not affect the output

conan/tools/cmake/cmake.py
```diff
@@ -111,6 +111,20 @@ def configure(self, variables=None, build_script_folder=None, cli_args=None):
         with chdir(self, build_folder):
             self._conanfile.run(command)
 
+    @property
+    def _verbosity(self):
+        verbosity = self._conanfile.conf.get("tools.build.verbosity")
```
```suggestion
        verbosity = self._conanfile.conf.get("tools.build:verbosity")
```

memsharded (commented):
Looks good. I'll try to add the CMAKE_MAKE_VERBOSE thing in it, and if there would be a way that the CMake + VS could also print verbose compile command lines, that would be great.

conan/tools/cmake/cmake.py
AbrilRBS: Omg 🤦 I ctrl-f to replace all the occurrences, and still this fell thru the cracks

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -190,6 +191,10 @@ def generate(self):
             VCVars(self._conanfile).generate()
         toolchain = os.path.abspath(os.path.join(self._conanfile.generators_folder,
                                                  toolchain_file or self.filename))
+
+        if "CMAKE_MAKE_VERBOSE" not in self.cache_variables:
+            self.cache_variables["CMAKE_MAKE_VERBOSE"] = self._conanfile.conf.get("tools.build:verbosity", check_type=bool)
```
memsharded: fix this bool mapping to ON/OFF

memsharded (approved):

memsharded (approved):

memsharded (approved):

czoido (approved):

PR #13461, opened 2023-03-16
Proposal for sources backup feature

memsharded:
Changelog: Feature: Automatic server side source backups for third parties.
Docs: Omit


conan/api/subapi/upload.py
```diff
@@ -40,3 +45,36 @@ def upload(self, package_list, remote):
         app.remote_manager.check_credentials(remote)
         executor = UploadExecutor(app)
         executor.upload(package_list, remote)
+
+    def upload_backup_sources(self, package_list):
+        app = ConanApp(self.conan_api.cache_folder)
+        config = app.cache.new_config
+        # TODO: Rethink this conf, does this live in core?
```
AbrilRBS:
```suggestion
```

The re-thinking was done :)

conans/client/rest/file_uploader.py
```diff
@@ -47,6 +47,10 @@ def _dedup(self, url, headers, auth):
         if response.status_code == 201:  # Artifactory returns 201 if the file is there
             return response
 
+    def exist(self, url, auth):
```
AbrilRBS:
```suggestion
    def exists(self, url, auth):
```

conan/api/subapi/upload.py
```diff
@@ -40,3 +45,36 @@ def upload(self, package_list, remote):
...
+        files = DownloadCache(download_cache_path).get_files_to_upload(package_list)
+        uploader = FileUploader(app.requester, verify=False, config=config)
+        # TODO: Dedup and list files
+        for file in files:
+            # Check Artifactory's HEAD
+            basename = os.path.basename(file)
+            full_url = url + basename
+            try:
+                upload = file.endswith(".json")
+                upload = upload or not uploader.exist(full_url, auth=None)
```
AbrilRBS:
```suggestion
                upload = upload or not uploader.exists(full_url, auth=None)
```

conan/api/subapi/upload.py
```diff
@@ -40,3 +45,36 @@ def upload(self, package_list, remote):
...
+                if upload:
+                    ConanOutput().info(f"Uploading file '{basename}' to server")
+                    uploader.upload(full_url, file, dedup=False, auth=None)
+                else:
+                    ConanOutput().info(f"File '{basename}' already in server, skipping upload")
+            except (AuthenticationException, ForbiddenException) as e:
+                raise ConanException(f"The source backup server '{url}' need authentication"
```
AbrilRBS:
```suggestion
                raise ConanException(f"The source backup server '{url}' needs authentication"
```

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -46,61 +47,68 @@ def _lock(self, lock_id):
...
+                self._file_downloader.download(download_url + h + ".json",
+                                               cached_path + ".json", **kwargs)
+                found_in_backup = True
+                conanfile.output.info(f"Sources from {url} found in remote backup {download_url}")
+                break
+            except NotFoundException:
+                # TODO: What happens if sha256 missmatch?
+                pass
+            except (AuthenticationException, ForbiddenException) as e:
+                raise ConanException(f"The source backup server '{download_url}' need authentication"
```
AbrilRBS:
```suggestion
                raise ConanException(f"The source backup server '{download_url}' needs authentication"
```

conan/api/subapi/upload.py
```diff
@@ -40,3 +45,36 @@ def upload(self, package_list, remote):
...
+            return
+        url = url if url.endswith("/") else url + "/"
+        download_cache_path = config.get("tools.files.download:download_cache")
+        download_cache_path = download_cache_path or config.get("core.download:download_cache")
+        if download_cache_path is None:
+            raise ConanException("Need to define 'core.download:download_cache'")
+
+        files = DownloadCache(download_cache_path).get_files_to_upload(package_list)
+        uploader = FileUploader(app.requester, verify=False, config=config)
+        # TODO: Dedup and list files
```
AbrilRBS:
```suggestion
        # TODO: List files
```

AbrilRBS (approved):
I'll push a fix for the tests failing now

conan/api/subapi/upload.py
memsharded: Actually de-dup was finally not active. I was testing it and it could be problematic for other servers. It needs to be a new conf, but I think this optimization can wait a bit.

conans/test/integration/cache/download_cache_test.py
```diff
@@ -199,8 +199,8 @@ def source(self):
 
             assert 2 == len(os.listdir(os.path.join(tmp_folder, "s")))
             content = json.loads(load(os.path.join(tmp_folder, "s", sha256 + ".json")))
-            assert "http://localhost:5000/myfile.txt" in content["unknown"]
-            assert len(content["unknown"]) == 1
+            assert "http://localhost:5000/myfile.txt" in content["references"]["unknown"]
```
memsharded: Maybe we should add a test that using name and version in the command line or in the recipe changes this unknown to an actual reference.

I am also changing my mind, to upload it anyway even if unknown. There is the URL and the sha256, not really relevant if using only from a consumer without ref?

memsharded (commented):
Good!

conans/test/integration/cache/download_cache_test.py
AbrilRBS: For source it does not change, the conanfile's `_conan_node` will still be empty, so even if `name` and `version` are set, calling `ref` will still throw an `AttributeError` exception and so we'll still identify it as "unknown" :/

I have some concerns around uploading unkowns, will let you know next week :)

AbrilRBS: After discussing it, let's not upload unkowns, the workflow does not make much sense

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -47,44 +47,48 @@ def _lock(self, lock_id):
...
-            else:
-                raise ConanException("Backup sources cache missed but "
-                                     "'core.backup_sources:cache_miss_policy' "
-                                     f"has an unknown value of '{policy}'")
-        return found_in_backup
+                raise ConanException(
+                    f"The source backup server '{download_url}' needs authentication"
+                    f"/permissions, please provide 'source_credentials.json': {e}")
+            except ConanException as e:
+                # TODO: Rethink how we detect a broken origin
```
AbrilRBS: This is NOT how we should do it, but pushed with this to gather feedback on how to properly handle this. A 500 error is raised as a ConanException, but we don't want to catch _every_ ConanException as a pass if trying with internet 🤔 
Maybe a different Exception type such as NotFound and Authentication do? I'd have done like so from the begining, but I have my doubts that this could be a breaking behaviour in the tools.download

AbrilRBS (commented):
* Implemented an ordered list to allow the user to define the order of source lookup.
* Ensured that the original URL can be broken in the list and that we still look for it.
* Moved download/upload tests to their own file

AbrilRBS (commented):
Added test when the checksum of origin is different, the flow continues with the next in the list

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -1,106 +1,165 @@
...
 from conans.util.files import mkdir, set_dirty_context_manager, remove_if_dirty, load, save
-from conans.util.locks import SimpleLock
-from conans.util.sha import sha256 as compute_sha256
 
 
-class CachingFileDownloader:
+def sources_caching_download(conanfile, urls, file_path,
+                             retry, retry_wait, verify_ssl, auth, headers,
+                             md5, sha1, sha256):
+    helpers = getattr(conanfile, "_conan_helpers")
```
AbrilRBS: I like this way of silencing the protected method access warnings 😄

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -1,106 +1,165 @@
...
+        self._file_downloader = FileDownloader(requester)
+
+    def download(self, url, file_path, auth, verify_ssl, retry, retry_wait):
+        if not self._download_cache:
+            self._file_downloader.download(url, file_path, retry=retry, retry_wait=retry_wait,
+                                           verify_ssl=verify_ssl, auth=auth, overwrite=False)
+            return
+
+        h = compute_sha256(url.encode())
+        cached_path = os.path.join(self._download_cache, "c", h)
```
AbrilRBS: I don't like how to create the class, you now need something that should be specific to the class. But passing it `_download_cache` would also not work (I _think_), so I think will stay as-is :/

conans/client/downloaders/caching_file_downloader.py
```diff
@@ -1,106 +1,165 @@
...
+        return
+
+    # We are going to use the download_urls definition for backups
+    download_cache = download_cache or cache.default_sources_backup_folder
+    # regular local shared download cache, not using Conan backup sources servers
+    backups_urls = backups_urls or ["origin"]
+    if download_cache and not os.path.isabs(download_cache):
+        raise ConanException("core.download:download_cache must be an absolute path")
+
+    cached_path = os.path.join(download_cache, "s", sha256)
```
AbrilRBS: Same thing here, this class initialization seems weird now

AbrilRBS (approved):
I worry about DownloadCache's state right now, but is in no way a blocker for me

conans/client/downloaders/caching_file_downloader.py
memsharded: Changed, with a large refactor to organize it a bit

conans/client/downloaders/download_cache.py
```diff
@@ -0,0 +1,86 @@
...
+from contextlib import contextmanager
+from threading import Lock
+
+from conans.util.dates import timestamp_now
+from conans.util.files import load, save
+from conans.util.locks import SimpleLock
+from conans.util.sha import sha256 as compute_sha256
+
+
+class DownloadCache:
```
AbrilRBS, line 12: ❤️

AbrilRBS (approved):
This is the one. Congratulations on getting this thru the finish line :)

I think that the tests might need some love in the future, but we can live with some repetition in there for now. Let's see how this works for cci

czoido (approved):

PR #13470, opened 2023-03-17
Add basic package filter for conan list html output

czoido:
Changelog: Feature: Add filtering for options and settings in conan list html output.
Docs: omit

This is also passing the arguments from the cli to show the input in the top navbar in the html

image

### DIFF
--- a/conan/cli/command.py
+++ b/conan/cli/command.py
@@ -93,9 +93,6 @@ class BaseConanCommand:
             raise ConanException("{} is not a known format. Supported formatters are: {}".format(
                 formatarg, ", ".join(self._help_formatters)))
 
-        if type(info) is dict:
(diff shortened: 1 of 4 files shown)


AbrilRBS (approved):

conan/cli/command.py
```diff
@@ -93,6 +93,9 @@ def _format(self, parser, info, *args):
             raise ConanException("{} is not a known format. Supported formatters are: {}".format(
                 formatarg, ", ".join(self._help_formatters)))
 
+        if type(info) is dict:
```
memsharded: This sounds a bit hack, I think I'd move it to the ``conan list command``, even if it requires gathering together a couple of arguments, not that many (pattern mostly)

memsharded (commented):
html and functionality looks good, but I'd move the ``cli_args`` to the ``conan list`` command exclusively

PR #13516, opened 2023-03-23
Fix subcommands from commands with underscores

czoido:
Changelog: Bugfix: Fix subcommands names when the parent command has underscores.
Docs: omit

When implementing a custom command that has something like:

```
@conan_command(group="Custom commands")
def build_info(conan_api: ConanAPI, parser, *args):
...

@conan_subcommand()
def build_info_create(conan_api: ConanAPI, parser, subparser, *args):
...
```
I noticed that the subcommand name was not properly formed, it appeared as `info-create` instead of `create`.

Maybe we want to get rid of the restriction of being the main command the start of the subcommand names and just use:

```
@conan_command(group="Custom commands")
def build_info(conan_api: ConanAPI, parser, *args):
...

@conan_subcommand()
def create(conan_api: ConanAPI, parser, subparser, *args):
...
```

### DIFF
--- a/conan/cli/command.py
+++ b/conan/cli/command.py
@@ -149,6 +149,7 @@ class ConanCommand(BaseConanCommand):
             self._subcommand_parser = self._parser.add_subparsers(dest='subcommand',
                                                                   help='sub-command help')
             self._subcommand_parser.required = True
+        subcommand.set_name(self.name)
         subcommand.set_parser(self._parser, self._subcommand_parser)
         self._subcommands[subcommand.name] = subcommand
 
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
Nice!

> Maybe we want to get rid of the restriction of being the main command the start of the subcommand names and just use

In my head it makes more sense how it is currently, but I see how this can be prettier to write too :)

memsharded:
> In my head it makes more sense how it is currently, but I see how this can be prettier to write too :)

I think I also prefer the consistency of having every command prefixed with the parent one.

memsharded (approved):

PR #13526, opened 2023-03-24
relocatable full_deploy CMake

### LINKED ISSUE #13511: Deploy thirdparties in conan 2.X
Hello,

In our project I have to ship our thirdparties together with our libraries (we develop an SDK like product, and we have to avoid thirdparty library version missmatches).

Our conan thirdparties comes from our conanfile.py with the running of conan install. In the conan file we use CMakeDeps generator. We use cmake to build our project after conan install ran. So I plan to ship the whole "conan install directory" to customers.

Since conan install directory contains only links to the local cache I used the two built in deployer of conan (direct and full deploys). For me (with 2.0 and 2.0.1) any of them worked perfect. In both cases the xxx_PACKAGE_FOLDER_RELEASE directory points to the deployed folder, but all paths in the xxx_BUILD_MODULES_PATHS_RELEASE variable still points to the local cache.

But independently the wrong BUILD_MODULES_PATHS variable the generated cmake variables are not relocatable, since absolute paths are used.

I tried to write a custom deployer and tried to change the deploy folder (dep.set_deploy_folder(new_folder)) to ${CMAKE_CURRENT_LIST_DIR} but that caused error in install since (this is logical) that is not a real directory.

Is there any way to make the cmake config  files relocatable in the deploy folder? I think if someone call deploy on its dependencies than probably want a hierarchy without absolute paths.

I also attach a sample project too. I called it with command:
`conan install . --output-folder install --profile:build ./gcc2004 --profile:host ./gcc2004 --build=missing --settings=build_type=Release -r conancenter --deploy=direct_deploy`
[conan_test.zip](https://github.com/conan-io/conan/files/11049294/conan_test.zip)

thanks

memsharded:
Changelog: Feature: The ``full_deploy`` deployer together with ``CMakeDeps`` generator learned to create relative paths deploys, so they are relocatable.
Docs: Omit

Close https://github.com/conan-io/conan/issues/13511

### DIFF
--- a/conan/tools/env/environment.py
+++ b/conan/tools/env/environment.py
@@ -414,8 +414,11 @@ class EnvVars:
             {deactivate}
             """).format(deactivate=deactivate if generate_deactivate else "")
         result = [capture]
+        location = os.path.dirname(file_location)
         for varname, varvalues in self._values.items():
             value = varvalues.get_str("%{name}%", subsystem=self._subsystem, pathsep=self._pathsep)
+            # To make the script relocatable
+            value = value.replace(location, "%~dp0")
             result.append('set "{}={}"'.format(varname, value))
 
         content = "\n".join(result)
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -447,28 +447,30 @@ class CppInfo(object):
                 origin = getattr(component, varname)
                 if origin is not None:
                     origin[:] = [os.path.join(folder, el) for el in origin]
-            if component._generator_properties is not None:
-                updates = {}
-                for prop_name, value in component._generator_properties.items():
-                    if prop_name == "cmake_build_modules":
-                        if isinstance(value, list):
-                            updates[prop_name] = [os.path.join(folder, v) for v in value]
-                        else:
-                            updates[prop_name] = os.path.join(folder, value)
-                component._generator_properties.update(updates)
+            properties = component._generator_properties
+            if properties is not None:
+                modules = properties.get("cmake_build_modules")  # Only this prop at this moment
+                if modules is not None:
+                    assert isinstance(modules, list), "cmake_build_modules must be a list"
+                    properties["cmake_build_modules"] = [os.path.join(folder, v) for v in modules]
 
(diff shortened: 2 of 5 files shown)


conan/tools/env/environment.py
```diff
@@ -414,8 +414,11 @@ def save_bat(self, file_location, generate_deactivate=True):
             {deactivate}
             """).format(deactivate=deactivate if generate_deactivate else "")
         result = [capture]
+        location = os.path.dirname(file_location)
         for varname, varvalues in self._values.items():
             value = varvalues.get_str("%{name}%", subsystem=self._subsystem, pathsep=self._pathsep)
+            # To make the script relocatable
+            value = value.replace(location, "%~dp0")
```
memsharded, line 421: This was easy to make Env generators to be relocatable for Windows/.bat, but it is challenging to do for .sh in Linux, due to not existing a generic pure shell (no bash only) solution

conans/model/build_info.py
```diff
@@ -447,28 +447,30 @@ def set_relative_base_folder(self, folder):
...
-            if component._generator_properties is not None:
-                updates = {}
-                for prop_name, value in component._generator_properties.items():
-                    if prop_name == "cmake_build_modules":
-                        if isinstance(value, list):
-                            updates[prop_name] = [os.path.join(folder, v) for v in value]
-                        else:
-                            updates[prop_name] = os.path.join(folder, value)
-                component._generator_properties.update(updates)
+            properties = component._generator_properties
```
memsharded, line 450: This was a minor refactor, no need to iterate all properties, if only 1 used ( I had to do the same below)

conans/test/functional/command/test_install_deploy.py
```diff
@@ -15,8 +17,30 @@
 def test_install_deploy():
     c = TestClient()
     c.run("new cmake_lib -d name=hello -d version=0.1")
+    c.run("create . -o *:shared=True -tf=")
+    conanfile = textwrap.dedent("""
```
memsharded, line 21: Adding an actual executable script to the test, to proof the final deploy is relocatable

czoido (approved):

PR #13529, opened 2023-03-24
tool_requires test with multiple versions of same package

### LINKED ISSUE #13521: [feature] require the same thing twice with different options
### What is your suggestion?

I would like to require the same package twice, or more with different options.

I have a package called `gcc/9.1.0`, which takes the options `os_target` and `arch_target`.

I have a recipe called `wine` and what I really want to do with that one is

```py
 def build_requirements(self):
    self.build_requries("gcc/11.3.0"). # provides x86_64-linux-gnu-gcc, options set automatically from settins_target
    self.build_requires("gcc/9.1.0", options={os_target="Windows", arch_target="x86") # provides i686-w64-mingw32-gcc
    self.build_requires("gcc/9.1.0", options={os_target="Windows", arch_target="x86_64") # provides x86_64-w64-mingw32-gcc
```

I tried wrapper recipes that require gcc, with every option of `direct`, `visible`, `run`, and `build` that i can think of, but nothing does what i want.

Whats blocking this from working is two things.   First is the requires have the same package name and that's not allowed.
possible solution...

```py
 def build_requirements(self):
    self.build_requries("gcc/11.3.0").
    self.build_requires("gcc/9.1.0", options={os_target="Windows", arch_target="x86", as="gcc-win32")
    self.build_requires("gcc/9.1.0", options={os_target="Windows", arch_target="x86_64", as="gcc-win64") 
```

then the windows compiler could appear in `self.dependencies["gcc-win32"]`

The second issue is that even if I create wrapper packages to require gcc, I can't get them to do what i wnat.
If the wrapper requires gcc with `visible=Fasle`, then the graph looks right, but the `gcc` packages  have `binary=Skip` set, so they can't be used

If `visible=True`, then the graph only gets one node for gcc, which isn't what we want.

memsharded:
Changelog: Feature: Demonstrate that it is possible to ``tool_requires`` different versions of the same package.
Docs: https://github.com/conan-io/docs/pull/3219

Close https://github.com/conan-io/conan/issues/13521


AbrilRBS (approved):
Nice!

PR #13542, opened 2023-03-27
some refactors and improvements in ConanProxy

memsharded:
Changelog: Omit
Docs: Omit

Refactors in ConanProxy, to prepare for some potential optimizations while resolving the dependency graph
- Cleaning the ugly initialization of ``self.update`` and ``self.check_update`` outside of ``__init__()``, passing as args.
- Reduced the level of indentations by early returns, reducing complexity
- Avoided repeated slow call to ``find_in_remotes`` for the ``--update`` case. Updates should be a bit faster now

### DIFF
--- a/conans/client/graph/proxy.py
+++ b/conans/client/graph/proxy.py
@@ -114,7 +114,7 @@ class ConanProxy(object):
         return found_rrev.get("remote"), found_rrev.get("ref")
 
     def _download_recipe(self, ref, remotes, scoped_output, update, check_update):
-        # When a recipe doesn't existin local cache, it is retrieved from sservers
+        # When a recipe doesn't existin local cache, it is retrieved from servers
         scoped_output.info("Not found in local cache, looking in remotes...")
         if not remotes:
             raise ConanException("No remote defined")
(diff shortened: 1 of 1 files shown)


conans/client/graph/proxy.py
```diff
@@ -118,26 +113,23 @@ def _find_newest_recipe_in_remotes(self, reference, remotes):
...
-            # If incomplete, resolve the latest in server
-            if not reference.revision:
-                reference = self._remote_manager.get_latest_recipe_reference(ref, remote)
-            self._remote_manager.get_recipe(reference, the_remote)
-            self._cache.update_recipe_timestamp(reference)
-            scoped_output.info("Downloaded recipe revision %s" % reference.revision)
-            return reference
-
+    def _download_recipe(self, ref, remotes, scoped_output, update, check_update):
+        # When a recipe doesn't existin local cache, it is retrieved from sservers
```
AbrilRBS:
```suggestion
        # When a recipe doesn't existin local cache, it is retrieved from servers
```

AbrilRBS (approved):

czoido (approved):

PR #13562, opened 2023-03-29
New ``revision_mode = scm_folder`` for monorepos

memsharded:
Changelog: Feature: New ``revision_mode = "scm_folder"`` for mono-repo projects that want to use ``scm`` revisions.
Docs: https://github.com/conan-io/docs/pull/3218

This is a preliminary experiment to discuss about https://github.com/conan-io/docs/issues/3135

The ``revision_mode = "scm"`` doesn't work in mono-repo because it takes the root folder commit. 
The current PR approach **is broken**, as it will take the folder commit when the ``conanfile.py`` is inside a subfolder (this layout is allowed)


conans/test/functional/revisions_test.py
```diff
@@ -796,35 +796,6 @@ def test_upload_no_overwrite_packages(self):
                          pref2.revision)
 
 
-class SCMRevisions(unittest.TestCase):
```
memsharded, line 799: These tests were redundant, consolidated them in the above tests

czoido (approved):

franramirez688 (approved):

PR #13581, opened 2023-03-31
clean temp folders in cache

### LINKED ISSUE #13576: [feature] Add an option in `conan cache clean` to remove files in `<conan_home>/p/t` folder
### What is your suggestion?

When a build fails, temporary build directories are kept in p/t folder, and they are never removed. I don't see any command to remove these directories unless I've missed something in documentation. Of course a simple rm does the job, but since conan cache is now an internal detail, a dedicated command or option in existing command would make sense.

memsharded:
Changelog: Feature: ``conan cache clean`` learned the ``--all`` and ``--temp`` to clean everything (sources, builds) and also the temporary folders.
Docs: https://github.com/conan-io/docs/pull/3145

Close https://github.com/conan-io/conan/issues/13576

- Minor refactor to improve ProfilesAPI interface, avoiding possible caching bugs

### DIFF
--- a/conan/api/subapi/cache.py
+++ b/conan/api/subapi/cache.py
@@ -1,3 +1,5 @@
+import shutil
+
 from conan.internal.conan_app import ConanApp
 from conan.internal.integrity_check import IntegrityChecker
 from conans.errors import ConanException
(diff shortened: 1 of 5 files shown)


conan/api/subapi/cache.py
```diff
@@ -47,17 +49,21 @@ def check_integrity(self, package_list):
         checker = IntegrityChecker(app)
         checker.check(package_list)
 
-    def clean(self, package_list, source=None, build=None, download=None):
+    def clean(self, package_list, source=True, build=True, download=True, temp=True):
```
AbrilRBS, line 52: What's the reasoning behind this specific change? This effectively makes the default API behaviour be "Clean all except if I tell you otherwise", while the CLI's is "Don't clean anything I don't tell you to", which seems weird

AbrilRBS (commented):
A small question, otherwise looks good :)

conan/api/subapi/cache.py
memsharded: I was thinking of the programmatic view:
```
conan_api.cache.clean()
```
Not doing absolutely nothing (it doesn't have arg validation)
I also thought that the expected outcome of such line is that the cache is cleaned, and not requiring ``clean(True, True, True, True)`` that sounds dirtier

The CLI at least has arg validation and will raise if no arg is provided.
And I agree, it is a bit weird to have to add all, a bit dirty, this is the reason I added ``--all`` argument. I know we discussed this UI, but it felt dirtier when I added a 4th argument

AbrilRBS: 👍 sounds good, thanks for the clarification!

AbrilRBS (approved):

PR #13592, opened 2023-04-01
Add B2Deps generator.

grafikrobot:
Changelog: Feature: Adds a B2Deps generator for the B2 build system.
Docs: https://github.com/conan-io/docs/pull/3142

- [x] I've opened another PR in the Conan docs repo to the ``release/2.0`` branch, documenting this one.

### DIFF
--- a/conan/tools/b2/b2deps.py
+++ b/conan/tools/b2/b2deps.py
@@ -82,7 +82,12 @@ class B2Deps(object):
                 self._content[dep_variant_jam] = ""
             # Declare/define the local project for the dependency.
             cbiv = [
-                f'# {dependency.pref}',
+                f'#| {dependency.pref}',
+                '[settings]',
+                dep_settings.dumps(),
+                '[options]',
+                dep_options.dumps(),
+                '|#',
                 f'pkg-project {dep_name_b2} ;']
             # Declare any system libs that we refer to (in usage requirements).
             system_libs = set(dependency.cpp_info.system_libs)
--- a/conan/tools/b2/b2deps.py
+++ b/conan/tools/b2/b2deps.py
@@ -26,6 +26,7 @@ class B2Deps(object):
             check_duplicated_generator(self, self._conanfile)
             generator_files = self.content
             for generator_file, content in generator_files.items():
+                self._conanfile.output.info(f"Saved B2Deps file {generator_file}")
                 save(generator_file, content)
 
     @property
@@ -74,7 +75,8 @@ class B2Deps(object):
             # The settings the dependency requires, i.e. finds relevant.
             dep_settings = dependency.settings
             # The variant specific file to add this dependency to.
-            dep_variant_jam = B2Deps._conanbuildinfo_variation_jam(dep_settings)
+            dep_variant_jam = B2Deps._conanbuildinfo_variation_jam(
+                dep_name_b2, dep_settings)
             if not dep_variant_jam in self._content:
                 self._content[dep_variant_jam] = ""
             # Declare/define the local project for the dependency.
@@ -173,8 +175,9 @@ class B2Deps(object):
         return cbi_target
 
     @staticmethod
-    def _conanbuildinfo_variation_jam(settings):
-        return 'conanbuildinfo-%s.jam' % B2Deps._b2_variation_key(settings)
+    def _conanbuildinfo_variation_jam(name, settings):
+        return 'conanbuildinfo-{}-{}.jam'.format(
+            name, B2Deps._b2_variation_key(settings))
 
     @staticmethod
     def _b2_variation_key(settings):
--- a/conan/tools/b2/b2deps.py
+++ b/conan/tools/b2/b2deps.py
@@ -7,6 +7,10 @@ from hashlib import md5
 
 
 class B2Deps(object):
+    """
+    B2Deps generates files that are automatically loaded by the B2 build system.
+    The files define localized subprojects and targets for dependencies.
+    """
 
     def __init__(self, conanfile):
         self._conanfile = conanfile
@@ -14,7 +18,7 @@ class B2Deps(object):
 
     def generate(self):
         """
-        This method will save the generated files to the conanfile.generators_folder
+        This method will save the generated files to the conanfile.source_folder
         """
         source_folder = self._conanfile.source_folder
         self._conanfile.output.highlight(f"Writing B2Deps to {source_folder}")
@@ -28,9 +32,11 @@ class B2Deps(object):
     def content(self):
         """
         Generates two content files: conanbuildinfo.jam and
-        conanbuildinfo-ID.jam which the former defines sub-projects for each
-        package and loads the other files and the latter define variables and
-        targets for the packages.
+        conanbuildinfo-ID.jam. The former defines common package definition
+        function and include the latter conanbuildinfo-ID.jam files. The
+        conanbuildinfo-IS.jam files define sub-projects and targets for each
+        settings variation. The generated files have stable names and content.
+        Hence they can be added to source control.
         """
         self._content = {}
         self._content_conanbuildinfo_jam()
--- a/conan/tools/b2/b2deps.py
+++ b/conan/tools/b2/b2deps.py
@@ -72,11 +72,12 @@ class B2Deps(object):
             dep_cpp_info = dependency.cpp_info
             if '_'+dep_name in dep_cpp_info.components:
                 dep_cpp_info = dep_cpp_info.components['_'+dep_name]
-            # The settings the dependency requires, i.e. finds relevant.
+            # The settings and options the dependency requires, i.e. finds relevant.
             dep_settings = dependency.settings
+            dep_options = dependency.options
             # The variant specific file to add this dependency to.
             dep_variant_jam = B2Deps._conanbuildinfo_variation_jam(
-                dep_name_b2, dep_settings)
+                dep_name_b2, dep_settings, dep_options)
             if not dep_variant_jam in self._content:
                 self._content[dep_variant_jam] = ""
             # Declare/define the local project for the dependency.
@@ -88,11 +89,11 @@ class B2Deps(object):
             for name, component in dependency.cpp_info.get_sorted_components().items():
                 system_libs |= set(component.system_libs)
             cbiv += self._content_conanbuildinfo_variation_declare_syslibs(
-                dep_name_b2, system_libs, settings=dep_settings)
+                dep_name_b2, system_libs, settings=dep_settings, options=dep_options)
             # Declare any package libs for usage requirements. The first one is
             # the main/global dependency.
             cbiv += self._content_conanbuildinfo_variation_declare_libs(
-                dep_name_b2, dep_cpp_info, settings=dep_settings)
+                dep_name_b2, dep_cpp_info, settings=dep_settings, options=dep_options)
             # Followed by any components of the dependency. But skipping the
             # special _depname component. As that is already declare as the
             # main/global lib.
@@ -100,14 +101,14 @@ class B2Deps(object):
                 if name.lower() == '_'+dep_name_b2:
                     continue
                 cbiv += self._content_conanbuildinfo_variation_declare_libs(
-                    dep_name_b2, component, settings=dep_settings)
+                    dep_name_b2, component, settings=dep_settings, options=dep_options)
             # Declare the main target of the dependency. This is an alias that
             # refers to all the previous targets and adds all the defines,
             # flags, etc for consumers.
             cbiv += self._content_conanbuildinfo_variation_declare_target(
                 dep_name_b2, dep_name_b2,
                 dep_cpp_info,
-                settings=dep_settings)
+                settings=dep_settings, options=dep_options)
             # Similarly declare the component targets, if any.
             for name, component in dependency.cpp_info.get_sorted_components().items():
                 # Again, always skipping the kludge component as it's already
@@ -115,14 +116,14 @@ class B2Deps(object):
                 if "_"+dep_name_b2 == name.lower():
                     continue
                 cbiv += self._content_conanbuildinfo_variation_declare_target(
-                    dep_name_b2, name.lower(), component, settings=dep_settings)
+                    dep_name_b2, name.lower(), component, settings=dep_settings, options=dep_options)
             # Add the combined text.
             self._content[dep_variant_jam] += "\n".join(cbiv)+"\n"
 
-    def _content_conanbuildinfo_variation_declare_libs(self, name, cpp_info, settings=None):
+    def _content_conanbuildinfo_variation_declare_libs(self, name, cpp_info, settings=None, options=None):
         name = name.lower()
         cbi_libs = []
-        variation = ' '.join(b2_features(B2Deps._b2_variation(settings)))
+        variation = ' '.join(b2_features(B2Deps._b2_variation(settings, options)))
         for lib in cpp_info.libs:
             search = ' '.join(
                 [f'<search>"{b2_path(d.replace(self._conanhome, "$(CONAN_HOME)"))}"' for d in cpp_info.libdirs+cpp_info.bindirs])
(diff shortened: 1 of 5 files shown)


conan/tools/b2/b2deps.py
```diff
@@ -0,0 +1,335 @@
...
+            for generator_file, content in generator_files.items():
+                save(generator_file, content)
+
+    @property
+    def content(self):
+        """
+        Generates two content files: conanbuildinfo.jam and
+        conanbuildinfo-ID.jam. The former defines common package definition
+        function and include the latter conanbuildinfo-ID.jam files. The
+        conanbuildinfo-IS.jam files define sub-projects and targets for each
```
SSE4: ID or IS?

grisumbras:
Do I understand it correctly that when this generator, for some dependencies A and B that have variation id Y, it creates a file `conanbuildinfo-Y.jam`? And if I then use a different profile which results in requirement of A but not B with the same variation id Y, `conanbuildinfo-Y.jam` will be overwritten and priviously declared alternatives for B will be lost?

If that's true, I think, the better approach would be to generate a `conandep-A-Y.jam` file for every dep A. This is how many CMake generators work.

Also, variation should take into account some common options, not just seetings. E.g. here's my implementation https://github.com/conan-io/conan/pull/13594/files#diff-869685af1d66e4b7f1a45a6609d7454943c8371bcbe02cda3ddc0e830cce1a23R5. Some of it was copied from the generator in Conan v1, some from Boost's Conan recipe.

grisumbras:
> Good idea. Although the only option I see in your code is `shared`. I do see the `compiler.threads` and `compiler.runtime` that I'm not accounting for also.
To be honest, the only other option I see other generators handling is `fPIC`.

memsharded:
@grisumbras submitted a PR a few hours later,  please read https://github.com/conan-io/conan/pull/13594#issuecomment-1493436725

conan/tools/b2/b2deps.py
```diff
@@ -0,0 +1,284 @@
...
+
+class B2Deps(object):
+    """
+    B2Deps generates files that are automatically loaded by the B2 build system.
+    The files define localized subprojects and targets for dependencies.
+    """
+
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
+        self._conanhome = get_conan_user_home()
```
memsharded, line 17: Access to the ``get_conan_user_home()`` sounds like a not valid solution. A generator shouldn't need the conan home location, that should be an abstraction for the generators.

conan/tools/b2/b2deps.py
```diff
@@ -0,0 +1,284 @@
...
+            dep_name_b2 = dep_name.lower()
+            # The dependency cpp_info. We need to consider that there's a
+            # "_depname" component. Such components are a kludge to appease
+            # cmake generators and will eventually go away. This special
+            # component holds the real root definitions of the dependency.
+            dep_cpp_info = dependency.cpp_info
+            if '_'+dep_name in dep_cpp_info.components:
+                dep_cpp_info = dep_cpp_info.components['_'+dep_name]
+            # The settings and options the dependency requires, i.e. finds relevant.
+            dep_settings = dependency.settings
```
memsharded: All the other generators don't use the settings of the dependency, but instead the settings of the consumer. When something like ``conan install -s &:build_type=Debug -s build_type=Release`` the generated files have a "_debug" in their name, it means: for the "debug" build of the consumer, use this (and it can be a release artifact, it doesn't matter).

conan/tools/b2/b2deps.py
```diff
@@ -0,0 +1,284 @@
...
+        # Requirements:
+        cbi_target += [
+            f'  {" ".join(b2_features(b2_variation(settings, options)))}']
+        cbi_target += [
+            f'  <source>lib.{l}' for l in cpp_info.libs+cpp_info.system_libs]
+        # No default-build:
+        cbi_target += ["  : :"]
+        # Usage-requirements:
+        cbi_target += [
+            f'  <include>"{b2_path(d.replace(self._conanhome, "$(CONAN_HOME)"))}"' for d in cpp_info.includedirs]
```
memsharded, line 174: Is CONAN_HOME an environment variable? Because it will not be defined by Conan in Conan 2.0. Who is defining CONAN_HOME?

conans/test/unittests/tools/b2/test_b2deps.py
```diff
@@ -0,0 +1,57 @@
...
+        "arch": ["x86"]})
+    conanfile.settings.build_type = "Release"
+    conanfile.settings.arch = "x86"
+
+    cpp_info = CppInfo(set_defaults=True)
+
+    conanfile_dep = ConanFile(None)
+    conanfile_dep.cpp_info = cpp_info
+    conanfile_dep._conan_node = Mock()
+    conanfile_dep._conan_node.ref = RecipeReference.loads("OriginalDepName/1.0")
```
memsharded, line 32: uppercase refs are not allowed in 2.0 anymore

memsharded (commented):
I wonder if given the lack of expertise in b2 by the maintainer team, if it would be better to have the generator distributed in some other way, like in ConanCenter (to be used as python_require, might require some change in ConanCenter CI to allow it). That would also allow to make the generator available simultaneously to 1.X and 2.0, and move it in parallel, taking less maintenance efforts and allowing faster fixes (not having to wait until the next Conan release)

conan/tools/b2/b2deps.py
grafikrobot: It's used to figure out host independent paths instead of the absolute paths that normally end up in generator files. That part of the path in the generated files is replaced with a Jam variable, `$(CONAN_HOME)`, which is dynamically obtained when the consumer does a build. I admit what I'm doing is novel. But seems perfectly doable/reasonable within the context of the generator and install IMO.

conan/tools/b2/b2deps.py
grafikrobot: For this generator the dual build types would still generate two files. It's just that the Debug vs. Release is encoded in the hash value of `conanbuilinfo-<depname>-<hash>.jam`. By using the settings of the dependencies it's possible to accumulate definitions across multiple conan install invocations. For example your `conan install -s &:build_type=Debug -s build_type=Release` would also produce the same results as:

```
conan install -s build_type=Debug
conan install -s build_type=Release
```

And you could keep doing such invocations for other variations. Accumulating all the installed variations. Which B2 would automatically match when invoked without any extra work/changes to the user Jamfiles.

conan/tools/b2/b2deps.py
grafikrobot: It's not an env variable. It's a Jam variable defined by the generated `conanbuildinfo.jam` file <https://github.com/conan-io/conan/pull/13592/files#diff-2f8866e89e91704d46e92b61ac2e9da1e16eac946031b507359e48edd9b3b4b4R269>. The value is computed when the user/consumer invokes their b2 build command to match whichever conan home directly they have at the time.

conans/test/unittests/tools/b2/test_b2deps.py
grafikrobot: B2Deps and the generated Jamfiles lowercase all such references. And that particular code was copy/pasted from the CmakeDeps test case.

SSE4:
> That would also allow to make the generator available simultaneously to 1.X and 2.0, and move it in parallel, taking less maintenance efforts and allowing faster fixes (not having to wait until the next Conan release)

I don't think we should care about 1.X compatibility here, as I suppose we already agreed 1.X will not gain any new features (such as new generators), given the fact 2.0 is already released and 1.X user-base is shrinking rapidly.

SSE4 (approved):

SSE4:
is there anything else preventing PR from merging? I would love to try it to simplify boost build.

SSE4:
@memsharded how can we move this one forward?

conan/tools/b2/b2deps.py
```diff
@@ -0,0 +1,284 @@
...
+import path ;
+import project ;
+import modules ;
+import feature ;
+import os ;
+
+rule pkg-project ( id )
+{
+    local id-mod = [ project.find $(id:L) : . ] ;
+    if ! $(id-mode)
```
grisumbras: This seems to be a typo

grafikrobot: It certainly is !

SSE4:
any update? if something prevents merging it, please tell, so we can address.
otherwise, if there are no objections, please merge it.

SSE4:
@conan-io/barbarians @jcar87 @RubenRBS @memsharded @franramirez688 is it possible to move this one forward? it's already opened for two months, I think PR is "brewed" already :) I have many people (enterprise) waiting on this one. as there were no objections for two months, I think it could be merged safely.

conan/tools/b2/b2deps.py
memsharded: I am still not sure about this. Generators also must work the same when a package is in ``editable`` mode for example, so the package is not in the cache at all, not in the CONAN_HOME. The same should happen for "deployed" artifacts, like https://blog.conan.io/2023/05/23/Conan-agnostic-deploy-dependencies.html, where the generated files will be used without Conan, and should be relocatable.

> It's used to figure out host independent paths instead of the absolute paths that normally end up in generator files.

This is another concern that should be a bit more orthogonal. I think it is better to have first the "simple" approach, in which paths are absolute paths to the cache, and work from there, based on the needs, than adding this that might have other unexpected rough edges

conan/tools/b2/b2deps.py
memsharded: This still doesn't hold. Take into account:
```python
name = "mypkg"
def requires(self):
      if self.settings.build_type == "Release":
               self.requires("myalloc/2.0")
      else:
               self.requires("myalloc/1.0")
```
(I have seen this in production code, more than once)

The settings are the settings of the consumer, and should match just 1 config. It is better not to try to implement a ``multi-config`` approach within a single ``conan install``, or it will have other unexpected consequences.

grisumbras: Why wouldn't this work, though? You would have 2 completely different targets in `/myalloc` project with these two installs. With

```
exe myexe : main.cpp /myalloc//myalloc ;
```
and  building with `b2 variant=release` the build sustem would choose one, and when building with `b2 variant=debug`, it would choose another.

memsharded: The only way that the dependency graph will point to ``myalloc/1.0`` is executing a ``conan install -s build_type=Debug``. It is completely impossible get the 2 different dependency graphs in one single ``conan install`` command, like ``myalloc/1.0`` will not even be in the computer if not executed. So:

> For example your ``conan install -s &:build_type=Debug -s build_type=Release would also produce the same results as
> conan install -s build_type=Debug
conan install -s build_type=Release

Cannot hold, it is not physically possible.

grisumbras: Oh, sure, that's true. But I don't think that it means this multi-config setup doesn't work. You just have to run `conan install` several times.

memsharded: No, it will not work. If for example,``myalloc`` is header-only, then it will not have variability based on the settings, and it will produce the same exact file if you use the dependency settings. So the Debug installation and the Release installation will overwrite each other, making the above scenario fail. This is why it is important that the setting used is the consumer setting, not the dependency setting.

conan/tools/b2/b2deps.py
grafikrobot: Good points on editable and deploy. I'll have to think about those, and do some testing.

grafikrobot: > > It's used to figure out host independent paths instead of the absolute paths that normally end up in generator files.
> 
> This is another concern that should be a bit more orthogonal. I think it is better to have first the "simple" approach, in which paths are absolute paths to the cache, and work from there, based on the needs, than adding this that might have other unexpected rough edges

Thing is where b2 deps puts the generated files it's expected for them to be stable and likely included in source control. As it's how the b2 conan integration works. Without that the automatic declaration of package targets just doesn't happen. Or, like the previous version of this, it means users needing to rerun the b2deps for each environment.

grafikrobot: > I am still not sure about this. Generators also must work the same when a package is in `editable` mode for example, so the package is not in the cache at all, not in the CONAN_HOME. The same should happen for "deployed" artifacts, like https://blog.conan.io/2023/05/23/Conan-agnostic-deploy-dependencies.html, where the generated files will be used without Conan, and should be relocatable.

First.. There are multiple references to `--deploy=full_deploy` in that blog post which is incorrect. They should be `--deployer=full_deploy` (it confused me for a few minutes when I copy pasted commands and they errored.).

Second.. Both editable and full_deploy work just fine as is with this b2deps generator. The reason being that the use of conanhome only happens for packages that are installed to the home cache. For anything else the generated conanbuildinfo-*.jam files contain absolute paths to the packages (source path in editable or output/full_deploy). The reason being that it's a simple path replacement:

```
f'<search>"{b2_path(d.replace(self._conanhome, "$(CONAN_HOME)"))}"' for d in cpp_info.libdirs+cpp_info.bindirs
f' <include>"{b2_path(d.replace(self._conanhome, "$(CONAN_HOME)"))}"' for d in cpp_info.includedirs]
```

One thing that I could do is to make the full_deploy paths be relative to the conanbuildinfo*.jam generated files instead of absolute paths. That way they are fully relocatable. But as you said, it's better to start "simple" :-) And that's something I can work out later since I suspect full_deploy is a less common feature.

TLDR; It works as is.

conan/tools/b2/b2deps.py
grafikrobot: @memsharded thanks for the explanations and example use case. And of course, thank you for finding that problem. The latest commit fixes this problem. I tested it with this example:

```
from conan import ConanFile


class Recipe(ConanFile):
    generators = "B2Deps"

    def requirements(self):
        if self.settings.build_type == "Release":
            self.requires("lyra/1.5.0")
            self.requires("libdeflate/1.17")
            self.requires("fmt/9.1.0")
        else:
            self.requires("lyra/1.6.1") # editable
            self.requires("libdeflate/1.17")
            self.requires("fmt/10.0.0", options={"header_only": True})
```

Which I think covers all the corners of this particular issue.

conan/tools/b2/b2deps.py
memsharded: > Thing is where b2 deps puts the generated files it's expected for them to be stable and likely included in source control.

I am afraid this is not the design of any Conan generators. They are not expected to be stable, but they are located in most cases inside the "build" folder (the ``generators`` folder inside ``build`` folder) and intended to be ``.gitignored`` and cleaned regularly. The reason for this is that these files can and will change for every single configuration change (architecture, shared, build_type), and not absolutely every binary variant can be captured in a file path of a generator.

I still think that this must be simplified and follow the rest of the generators design, allowing absolute paths in the generated files and managing them as non stable and not in version control.

grafikrobot: > I am afraid this is not the design of any Conan generators. They are not expected to be stable, but they are located in most cases inside the "build" folder (the `generators` folder inside `build` folder) and intended to be `.gitignored` and cleaned regularly. The reason for this is that these files can and will change for every single configuration change (architecture, shared, build_type), and not absolutely every binary variant can be captured in a file path of a generator.
> 
> I still think that this must be simplified and follow the rest of the generators design, allowing absolute paths in the generated files and managing them as non stable and not in version control.

I guess there are a couple of different things in that..

1. Stability of the generated files, in both names and content.
2. Absolute or "relative" paths in the generated files.
3. In version control.
4. Where they are generated.

Each of those is separable from the others.

(1) I'm perfectly fine not calling them stable. After all it's just a label. They might or might not be stable.

(2) I can certainly do that. And then they are almost certainly not going to be stable.

(3) Sure, I don't have to say they can go in version control. Users might still put them there regardless though.

(4) I can't change this. If they are generated some place other than as a sibling to the conanfile.txt/py they will not. Specifically the B2 package manager integration will not load them. The declaration of the sub-targets will not work. Users would be forced to instruct/define/setup some manual way to define that "build/generator" folder to b2 instead of the automatic integration. Which goes very much against the b2 goals of removing complexity.

grafikrobot: I guess I could abandon the idea of the B2Deps generator entirely and create something to read the conan packages directly, and entirely in b2. Which happens to be the way a pending b2 PR to support vcpkg integration works ATM.

PR #13597, opened 2023-04-02
fix lockfiles with overrides

### LINKED ISSUE #13519: [bug] openssl version conflict, even with override=True
### Environment details

* Compiler+version: N/A
* Conan version: 2.0.2
* Python version: 3.11

### Steps to reproduce

Given a project that requires openssl 1.1.1.t and also transitively 1.1.1s:

```
from conan import ConanFile
class Mwe(ConanFile):
  def requirements(self):
    self.requires("grpc/1.50.0#d2c60eb68fdeae68328741258897b5c8")
    self.requires("openssl/1.1.1t#7d9fc949345908e1b8609d9d9f528a08")
```

With it, I can reproduce the error from #13368. It looks like Conan can't order these version numbers.

```
$ conan lock create .
ERROR: Version conflict: grpc/1.50.0->openssl/1.1.1s, ->openssl/1.1.1t.
```

Changing the last line to

```
    self.requires("openssl/1.1.1t#7d9fc949345908e1b8609d9d9f528a08", override=True)
```

Causes `conan lock create .` to succeed, but then

```
$ conan graph info .
ERROR: Requirement 'openssl/1.1.1s' not in lockfile
```

Is this a bug? What is the proper way to depend on a different version of openssl? Thanks!

### Logs

_No response_

memsharded:
Changelog: Bugfix: Solve failing lockfiles when graph has requirements with ``override=True``
Docs: Omit

Test ``lock_install_test.py`` has been removed, the ``conan lock install`` has been completely removed in 2.0, with no replacement.

Close https://github.com/conan-io/conan/issues/13519


conans/model/graph_lock.py
```diff
@@ -146,6 +149,9 @@ def merge(self, other):
         self._requires.merge(other._requires)
         self._build_requires.merge(other._build_requires)
         self._python_requires.merge(other._python_requires)
+        # TODO: Check if this holds for merges
+        self._alias.update(other._alias)
```
AbrilRBS, line 152: I think this is a bugfix for `lock merge` too no? Maybe we should add a changelog line for it too if so. (After looking at the code, I haven't seen much `merge` tests which would have caught this so I'm now not sure)

AbrilRBS (approved):

conans/model/graph_lock.py
memsharded: Agree, doing a separate PR for it in https://github.com/conan-io/conan/pull/13763, with its own test

PR #13613, opened 2023-04-04
Add `-gc` to pass global confs in CLI

### LINKED ISSUE #13410: Add a way to define global confs from CLI
I think we will definitely need some command line ``-global-conf`` argument to be able to define core values like this one

_Originally posted by @memsharded in https://github.com/conan-io/conan/pull/13321#discussion_r1130707209_

AbrilRBS:
Changelog: Feature: Add `--global-conf` flag to pass global confs in CLI
Docs: 

In progress. I just realized that we don't only want `-gc` where passing profiles is allowed, but everywhere. Still need to do that, but for now passing `-gc` will allow you to override global.conf's contents

Closes #13410

### DIFF
--- a/conan/api/subapi/profiles.py
+++ b/conan/api/subapi/profiles.py
@@ -32,14 +32,14 @@ class ProfilesAPI:
         build = [self.get_default_build()] if not args.profile_build else args.profile_build
         host = [self.get_default_host()] if not args.profile_host else args.profile_host
         profile_build = self.get_profile(profiles=build, settings=args.settings_build,
-                                         options=args.options_build, globalconf=args.global_conf,
+                                         options=args.options_build, global_conf=args.global_conf,
                                          conf=args.conf_build)
         profile_host = self.get_profile(profiles=host, settings=args.settings_host,
-                                        options=args.options_host, globalconf=args.global_conf,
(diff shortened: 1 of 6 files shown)


SSE4:
just in case, can we stop abbreviating cli arguments like that? by [convention](https://man7.org/linux/man-pages/man3/getopt.3.html), `-gc` stands for `-g` + `-c` combined. otherwise, it's super confusing.

memsharded (changes requested):
I'd say with this approach, the "core" conf used internally by Conan will not be modified, please add some test to validate it.

The key is that it is not enough to inject core conf in both profiles, it should be injected in the core ``global_config`` object, guaranteeing it is modified before usage when accessed via ``cache.new_config``

AbrilRBS:
Ok, after looking into your point it turns out to be way more involved than I though it would be. There is no easy way for `ClientConfig` to get the args info to populate its `new_config` before it is used elsewhere, and it does not seem possible to inject the values from outside while ensuring that no usage has happened yet (The class even fetches one conf from its constructor so...)

So while I think this is something we want to, let's close this PR for 2.X once we have time to focus on these kinds of nice-to-have :)

PR #13662, opened 2023-04-10
fix conan list for packages with no settings or options

memsharded:
Changelog: Bugfix: ``conan list -p <package-query>`` failed when a package had no settings or options.
Docs: Omit


AbrilRBS (approved):

AbrilRBS:
I just hit this same issue, was about to file an issue, so pretty good timing there!

PR #13685, opened 2023-04-13
new build-order-prepare subcommand

memsharded:
Changelog: Feature: New ``conan graph build-order-prepare`` for CI flows simplification 

### DIFF
--- a/conan/cli/commands/graph.py
+++ b/conan/cli/commands/graph.py
@@ -4,13 +4,15 @@ import os
 from conan.api.output import ConanOutput, cli_out_write, Color
 from conan.cli import make_abs_path
 from conan.cli.args import common_graph_args, validate_common_graph_args
-from conan.cli.command import conan_command, conan_subcommand
+from conan.cli.command import conan_command, conan_subcommand, OnceArgument
 from conan.cli.formatters.graph import format_graph_html, format_graph_json, format_graph_dot
 from conan.cli.formatters.graph.graph_info_text import format_graph_info
 from conan.cli.printers.graph import print_graph_packages, print_graph_basic
 from conan.internal.deploy import do_deploys
 from conans.client.graph.install_graph import InstallGraph
 from conan.errors import ConanException
+from conans.model.graph_lock import Lockfile
+from conans.model.recipe_ref import RecipeReference
 
 
 @conan_command(group="Consumer")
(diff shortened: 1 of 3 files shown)


AbrilRBS (approved):

conan/cli/commands/graph.py
```diff
@@ -83,6 +85,37 @@ def graph_build_order(conan_api, parser, subparser, *args):
...
 
 
+@conan_subcommand(formatters={"text": cli_out_write})
+def graph_build_order_prepare(conan_api, parser, subparser, *args):
+    """
+    Compute the build order of a dependency graph.
+    """
+    subparser.add_argument("build_order_file", help="build-order json file")
+    subparser.add_argument("ref", help="recipe to build")
+    subparser.add_argument("package_id", help="package to build")
```
AbrilRBS, line 95: Couldn't we put this two arguments together so that you need to pass `pkg/1.0:abc123`? Your way is more explicit which might avoid confusion though..

memsharded: Yes, that is doable, whatever UI the team prefers.

conan/cli/commands/graph.py
```diff
@@ -83,6 +85,37 @@ def graph_build_order(conan_api, parser, subparser, *args):
     return install_order_serialized
 
 
+@conan_subcommand(formatters={"text": cli_out_write})
+def graph_build_order_prepare(conan_api, parser, subparser, *args):
+    """
+    Compute the build order of a dependency graph.
```
czoido, line 91: This should be updated

conan/cli/commands/graph.py
czoido: Yes, I think I agree with adding the package_id in the reference too instead of a separate argument

conan/cli/commands/graph.py
```diff
@@ -83,6 +85,37 @@ def graph_build_order(conan_api, parser, subparser, *args):
     return install_order_serialized
 
 
+@conan_subcommand(formatters={"text": cli_out_write})
+def graph_build_order_prepare(conan_api, parser, subparser, *args):
```
czoido, line 89: I think that maybe the name of the subcommand it's not so intuitive, what about something like:
- `conan graph install-prepare`
- `conan graph install-config`
Maybe more explicit like this?
- `conan graph generate-install-args`

memsharded:
Dropping this:
- The ``build-order`` will embed ``build_args`` processed string
- Adding ``--lockfile-overrides`` CLI arg that allows injecting the overrides from ``build-order``

Continue in https://github.com/conan-io/conan/pull/13597

PR #13688, opened 2023-04-13
Add info_invalid to graph node serialization

AbrilRBS:
Changelog: Feature: Add info_invalid to graph node serialization
Docs: Omit

This would have been useful while debugging the cppstd issue in c3i, so I'm adding it to help the next person who find themselves wanting to know why InvalidConfiguration was raised. This might also help in other ways for `validate_build` in c3i :)

### DIFF
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -219,6 +219,7 @@ class Node(object):
         result["binary"] = self.binary
         # TODO: This doesn't match the model, check it
         result["invalid_build"] = self.cant_build
+        result["info_invalid"] = self.conanfile.info.invalid if self.conanfile.info is not None and self.conanfile.info.invalid is not None else False
         # Adding the conanfile information: settings, options, etc
         result.update(self.conanfile.serialize())
         result["context"] = self.context
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -219,7 +219,7 @@ class Node(object):
         result["binary"] = self.binary
         # TODO: This doesn't match the model, check it
         result["invalid_build"] = self.cant_build
-        result["info_invalid"] = self.conanfile.info.invalid if self.conanfile.info is not None and self.conanfile.info.invalid is not None else False
+        result["info_invalid"] = getattr(getattr(self.conanfile, "info", None), "invalid", None)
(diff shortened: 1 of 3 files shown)


conans/client/graph/graph.py
```diff
@@ -219,6 +219,7 @@ def serialize(self):
         result["binary"] = self.binary
         # TODO: This doesn't match the model, check it
         result["invalid_build"] = self.cant_build
+        result["info_invalid"] = self.conanfile.info.invalid if self.conanfile.info is not None and self.conanfile.info.invalid is not None else False
```
memsharded: Why ``None`` isn't good here, instead of ``False``? If there is no information about it, I'd expect ``None`` as lack of existence, better than False?

conans/test/integration/command/info/info_test.py
```diff
@@ -356,6 +354,26 @@ def export(self):
         assert "ERROR: Package 'dep/0.1' not resolved: dep/0.1: Cannot load" in c.out
         assert exit_code == ERROR_GENERAL
 
+    def test_invalid_config_validate(self):
```
memsharded: Instead of a new test, I think a new check in the existing ``test_validate_create`` test would be easier (and faster)

conans/client/graph/graph.py
AbrilRBS: Because `invalid_build` seems to work that way, so I made it behave like it, not to have too many differing behaviours, but I'd also prefer having None here if it weren't for that

conans/test/integration/command/info/info_test.py
AbrilRBS: I knew there had to be a test for it already, but could not find it, thanks for the tip :)

conans/client/graph/graph.py
memsharded: ``invalid_build`` gets the value from ``cant_build``, and in this case it is find to be ``False``, you don't lack the information, you know if it is either valid or invalid (boolean), as opposed to the ``info_invalid`` which is string (or None if not defined)

AbrilRBS: You're right! I'll change it

memsharded (approved):

PR #13689, opened 2023-04-13
test for transitive build-scripts

### LINKED ISSUE #13683: [question] CMakeDeps not considering transitive build requirements?
### What is your question?

Hi everybody,

I'm facing a setup where a conan ```package A``` contains a bunch of build scripts ```package_type = "build_scripts"``` and there is another conan ```package B``` with  ```package_type = "build_scripts"``` as well, which builds upon and wraps A, but still only contains build scripts . 

In my conan ```package C``` I require these build scripts. Instead of adding direct dependencies to both it does make sense in my mind to only depend on ```B```, which itself depends on ```A``` via ```tool_requires(A, visible=True)```, since ```B```  cannot stand alone anyway and ```A``` will only be utilized by calls to ```B```.

Using conan 2.0.3 the dependency graph reflects the setup as I would expect, but coming to CMakeDeps generator only the direct build dependencies are considered. This obviously breaks a ```find_package(A)``` call within ```B```.

Am I missing some configuration to make this use case possible? If it is not by design, I would be interested in why CMakeDeps restricts this, when the user takes care of downstream visibility for build requirements with the ```visible``` parameter already?

memsharded:
Changelog: Omit
Docs: Omit

Close https://github.com/conan-io/conan/issues/13683

### DIFF
--- a/conans/test/functional/graph/test_transitive_build_scripts.py
+++ b/conans/test/functional/graph/test_transitive_build_scripts.py
@@ -1,0 +1,85 @@
+import textwrap
+
+import pytest
+
+from conans.test.utils.tools import TestClient
+
+
+@pytest.mark.tool("cmake")
+def test_transitive_build_scripts():
+    c = TestClient()
+    scriptsa = textwrap.dedent("""
+        from conan import ConanFile
+        from conan.tools.files import copy
+        class ScriptsA(ConanFile):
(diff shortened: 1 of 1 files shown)


AbrilRBS (approved):

memsharded:
This would deserve an ``example`` in the examples and docs repos.

PR #13700, opened 2023-04-15
add: hint that sha256 sum triggers download cache

elsamuko:
see https://github.com/conan-io/docs/issues/3505
and https://github.com/conan-io/conan/blob/release/2.0/conans/client/downloaders/caching_file_downloader.py#L30

Changelog: (Feature | Fix | Bugfix): Added doc line that sha256sum caches downloads

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -185,7 +185,7 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
     :param headers: A dictionary with additional headers
     :param md5: MD5 hash code to check the downloaded file
     :param sha1: SHA-1 hash code to check the downloaded file
-    :param sha256: SHA-256 hash code to check the downloaded file
+    :param sha256: SHA-256 hash code to check the downloaded file. If set, the download is cached.
     """
     config = conanfile.conf
 


elsamuko:
@memsharded I'm not sure, if this is the correct branch to make PRs too, the contributor doc says [`develop`](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md#dev-flow--pull-requests), `release/2.0` seems to be the active one, and the [user doc](https://docs.conan.io/2/reference/tools/files/downloads.html) links to `develop2`.

conan/tools/files/files.py
```diff
@@ -185,7 +185,7 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
     :param headers: A dictionary with additional headers
     :param md5: MD5 hash code to check the downloaded file
     :param sha1: SHA-1 hash code to check the downloaded file
-    :param sha256: SHA-256 hash code to check the downloaded file
+    :param sha256: SHA-256 hash code to check the downloaded file. If set, the download is cached.
```
memsharded, line 188: But this is not generally true. This only happens if the download cache is activated in the ``conf``, but that is not a default. On the other hand, if the cache is activated and the download does not define ``sha256``, a warning will be printed to the output, so it will be clear what is happening.

memsharded (commented):
Thanks for the contribution @elsamuko 

However, I'd say the comment would not be fully correct, as the caching does not happen in the general case, only when the cache is activated.

elsamuko:
@memsharded Thanks for the information. I just found out, that my sample download is cached even without providing sha256:  
https://github.com/elsamuko/linuxtag_2023_conan/blob/main/from_source/conanfile.py#L35  
And there is no warning, that no sha256 was provided.

But it seems, only, if the conanfile is unchanged. Any change in the conanfile.py triggers the download, even with sha256.

PR #13706, opened 2023-04-17
Feature/warnings quiet

### LINKED ISSUE #13517: [feature] Provide ability to suppress Conan 1.x legacy warnings
### What is your suggestion?

While I certainly see the value of issuing warnings about deprecated features when developing or porting Conan 2.0 recipes, these warnings are noise when running a CI pipeline. For example:

> zlib/1.2.13: Already installed! (4 of 4)
> WARN: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
> WARN:     'env_info' used in: cmake/3.25.1, gcc/10.3.0
> WARN:     'cpp_info.names' used in: zlib/1.2.13

It would be nice if there was an easy way to globally suppress this type of warning--that is warnings that will go away when the Conan 1.x support is stripped out of Conan 2.0 compatible recipes. Thanks.

memsharded:
Changelog: Feature: Introduce new ``core:skip_warns`` configuration to be able to silence some warnings in the output.
Docs: https://github.com/conan-io/docs/pull/3215

Close https://github.com/conan-io/conan/issues/13517

### DIFF
--- a/conan/api/output.py
+++ b/conan/api/output.py
@@ -1,13 +1,12 @@
-import json
 import sys
 
 from colorama import Fore, Style
 
 from conans.client.userio import color_enabled
+from conans.errors import ConanException
 from conans.util.env import get_env
 
 LEVEL_QUIET = 80  # -q
-
 LEVEL_ERROR = 70  # Errors
 LEVEL_WARNING = 60  # Warnings
 LEVEL_NOTICE = 50  # Important messages to attract user attention.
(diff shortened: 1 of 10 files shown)


AbrilRBS (commented):
I only have one UX question:
I'm wondering if we're going to reuse some "id"s in the future, seeing how general the proposed `deprecated` one is. If so, maybe it's not `warn_id` but `warn_category`/`warn_tag` or the likes?

memsharded:
> I'm wondering if we're going to reuse some "id"s in the future, seeing how general the proposed deprecated one is. If so, maybe it's not warn_id but warn_category/warn_tag or the likes?

I initially implemented some kind of abstract ID, like W01, but then I realized that it was only adding an unnecessary abstraction, and too much boilerplate in code, translating codes back an forth. So I'd prefer to have something with a bit of semantics. We can discuss if we want ``Wdeprecation``, ``Wdeprecated2.0``, ``c3i_removed``, or any other, but I'd suggest using plain text identifiers with a bit of meaning.

czoido (approved):

AbrilRBS (approved):

PR #13712, opened 2023-04-17
build_requires visible can be overriden

### LINKED ISSUE #13638: [question] version overrides for dependencies that are both host and build (via tool_requires)
### What is your question?

Adding this ticket to look into this - could be a potential feature or document a proper way of dealing with this.

Some packages include _both_ libraries and executables - and the way Conan 2.0 handles propagation of usage requirements, we can handle cross-compilation very nicely. For this, we recommend expressing the dependency like this:

```
def requirements(self):
    self.requires("foobar/x.y.z")

def build_requirements(self):
    self.tool_requires("foobar/x.y.z")
```

When cross-building, two different packages will be pulled: one for an os/architecture that we can _run_ on the build machine (the `tool_requires`), and another one for the library to link against for the _host_ context.

However, the library requirement (host context), is liable to overrides from downstream consumers.  Some of these tools (that contain library+executables) are very sensitive to a mismatch in versions and have a stronger coupling.

Such a case is protobuf: the protobuf compiler (`tool_requires`) needs to generate `.pb.cc` and `.pb.h` files that are compatible with the version of protobuf resolved for the library requirement. If this is not the case, we will eventually get a compiler error.

I have uploaded a full reproducible example here (requires Docker): https://github.com/jcar87/conan-tool-require-override 

The example involves cross-building, but note that this situation can and will also happen when _not_ cross compiling, however a lot of recipes in conan center are including logic to work around this when not cross building, in a way that forces the executable from the _host_ package to be used. This is non-trivial logic that is hard to follow for a lot of contributors, and it only hides the problem: it really doesn't solve it when cross-building. 

memsharded:
Changelog: Feature: Define new ``self.tool_requires("pkg/<host_version>")`` to allow some tool-requires to follow and use the same version as the "host" regular requires do.
Docs: https://github.com/conan-io/docs/pull/3223

Close https://github.com/conan-io/conan/issues/13638


conans/test/integration/build_requires/build_requires_test.py
```diff
@@ -747,3 +747,54 @@ def requirements(self):
...
+    c = TestClient()
+    pkg = textwrap.dedent("""
+        from conan import ConanFile
+        class ProtoBuf(ConanFile):
+            name = "pkg"
+            version = "0.1"
+            def requirements(self):
+                self.requires("protobuf/1.0")
+            def build_requirements(self):
+                self.tool_requires("protobuf/1.0", visible=True)
```
memsharded, line 762:
```suggestion
                version = self._conan_node.transitive_deps.get("protobuf").version
                self.tool_requires("protobuf/{version}))
```
Try something like this.

AbrilRBS (approved):

czoido (approved):

franramirez688 (approved):

PR #13716, opened 2023-04-18
Improve `conan inspect` output, it now understands `set_name`/`set_version`

### LINKED ISSUE #13628: conan v2: Expose `conan inspect` through the API to access attributes
https://github.com/conan-io/conan/blob/5bb2b8ddd7b23f460c1b81ee82c78aaf56c208ad/conan/cli/commands/inspect.py#L22

AbrilRBS:
Changelog: Feature: Improve `conan inspect` output, it now understands `set_name`/`set_version`.
Docs: https://github.com/conan-io/docs/pull/3204

`conan inspect` now parses the conanfile and returns its serialization

Close https://github.com/conan-io/conan/issues/13628


conan/cli/commands/inspect.py
```diff
@@ -24,20 +23,27 @@ def inspect(conan_api, parser, *args):
...
-
-    return ret
+    lockfile = conan_api.lockfile.get_lockfile(lockfile=args.lockfile,
+                                               conanfile_path=path,
+                                               cwd=os.getcwd(),
+                                               partial=None)
+    remotes = conan_api.remotes.list(args.remote) if args.remote else []
+    conanfile = conan_api.local.inspect(path, remotes=remotes, lockfile=lockfile)
+    result = conanfile.serialize()
+    possible_options = conanfile.options.possible_values
```
memsharded: I think the options possible values might be inside ``conanfile.serialize()`` (it will exist also in ``conan graph info``, but maybe it is ok there too?)

Also, it might be better to have this as ``possible_options`` or ``options_definition`` or something like that, so the ``options`` member always has the same meaning (actual values)?

conans/model/conan_file.py
```diff
@@ -121,15 +122,18 @@ def __init__(self, display_name=""):
...
-                  "upload_policy",
-                  "revision_mode", "provides", "deprecated", "win_bash", "win_bash_run"):
+        for a in ("name", "version", "user", "channel", "url", "license",
+                  "author", "description", "topics", "homepage", "build_policy", "upload_policy",
+                  "revision_mode", "provides", "deprecated", "win_bash", "win_bash_run",
+                  "default_options", "options_description"):
             v = getattr(self, a)
             if v is not None:
-                result[a] = v
+                result[a] = str(v)
```
memsharded: default_options and options description are plain dicts, maybe it is not necessary to serialize it to string, but it can be dictionaries?

conan/cli/commands/inspect.py
```diff
@@ -1,21 +1,20 @@
...
 
 from conan.api.output import cli_out_write
-from conan.cli.command import conan_command
+from conan.cli.command import conan_command, OnceArgument
 from conan.cli.formatters import default_json_formatter
 
 
 def inspect_text_formatter(data):
-    for name, value in data.items():
+    for name, value in sorted(data.items(), key=lambda x: x[0]):
```
memsharded: why this lambda? Not very intuitive

memsharded (commented):
Looks good, just a couple of open issues about serialization to discuss.

conans/model/conan_file.py
AbrilRBS: Yep. The issue this solves was `version` being a `Version`. The simpler choice would be to special-case it, but this seemed more robust against future changes

conan/cli/commands/inspect.py
AbrilRBS: While playing with the output I realized it would be insertion-order dependant and implemented a sort on the items to have them show up alphabetically. Maybe we want to have a custom order and just have to be carefurl about when we insert the result in the serialization?

conans/model/conan_file.py
```diff
@@ -121,16 +122,23 @@ def __init__(self, display_name=""):
...
+                  "revision_mode", "provides", "deprecated", "win_bash", "win_bash_run",
+                  "default_options", "options_description", "generators", "requires"):
+            v = getattr(self, a, None)
             if v is not None:
                 result[a] = v
-
+        if self.version is not None:
+            result["version"] = str(self.version)
+        if self.topics is not None:
+            result["topics"] = list(self.topics)
```
AbrilRBS, line 135: Not happy about this, but some tests were failing when calling `graph create` for json formatter (`test/integration/command/create_test.py:475`). I do wonder why now and not with the old changes the also has the topics returned

memsharded (approved):

conans/model/conan_file.py
```diff
@@ -121,16 +122,23 @@ def __init__(self, display_name=""):
...
         result = {}
 
-        for a in ("url", "license", "author", "description", "topics", "homepage", "build_policy",
-                  "upload_policy",
-                  "revision_mode", "provides", "deprecated", "win_bash", "win_bash_run"):
-            v = getattr(self, a)
+        for a in ("name", "user", "channel", "url", "license",
+                  "author", "description", "homepage", "build_policy", "upload_policy",
+                  "revision_mode", "provides", "deprecated", "win_bash", "win_bash_run",
+                  "default_options", "options_description", "generators", "requires"):
```
memsharded: Serialization of the ``requires`` field is a bit uncertain how is being serialized, as it is a Python ``Requirements`` object?

conans/model/requires.py
```diff
@@ -143,6 +143,13 @@ def __str__(self):
                                              self.visible)
         return "{}, Traits: {}".format(self.ref, traits)
 
+    def serialize(self):
+        return {attribute: getattr(self, attribute) for attribute in ("run", "libs", "skip",
```
memsharded: where is ``ref``?

AbrilRBS: In the tests but did not stash the changes for this file 🥴

conan/cli/commands/inspect.py
memsharded: Still not sure. ``sorted(mydict.items())`` already sorts by the first element (key), no need for the lambda

memsharded (approved):

Issue #9772, opened 2021-10-09
[feature] Accept verbose build for Meson


uilianries:
The new [Meson](https://docs.conan.io/en/latest/reference/conanfile/tools/meson.html#meson) build helper does not offer a way to execute a verbose build, when used with `MesonToolchain`. So far, we need to trick `target` argument with `--verbose`. My suggestion is adding a new parameter to be used when building:

```python
from conans import ConanFile
from conan.tools.meson import Meson, MesonToolchain

class App(ConanFile):
    settings = "os", "arch", "compiler", "build_type"
    requires = "hello/0.1"
    generators = "pkg_config"

    def generate(self):
        tc = MesonToolchain(self)
        tc.generate()

    def build(self):
        meson = Meson(self)
	    meson.configure(source_folder="src")
	    meson.build(verbose=True)
```

It will be equivalent to:

    meson compile -C "src/" --verbose

- [x] I've read the [CONTRIBUTING guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).

SSE4:
it should be aligned with `CMakeToolchain` as well.
e.g. we don't set `CMAKE_VERBOSE_MAKEFILE` right now in the new toolchain.
perhaps, it needs a common conf section?

AbrilRBS:
Closed by https://github.com/conan-io/conan/pull/13729 which implemented `tools.compilation:verbosity` support for the Meson integration in Conan 2.0.7

PR #13729, opened 2023-04-19
Split build & compilation verbosity control into two confs

### LINKED ISSUE #12239: [feature] Allow users to enable verbose Makefile in CMakeToolchain

### LINKED ISSUE #13703: [feature] conan 2.0 verbose build option for meson
### What is your suggestion?

Hi,
Conan 2.0 seems to be missing a verbose build option for meson.

I was doing the following with 1.x:
```python
meson = Meson(self)
meson.configure(<configure-opts>)
meson.build(args=['--verbose'])
```

I've updated my conanfile.py as follows for 2.0
```python
tc = MesonToolchain(self)
tc.project_options.update(<options>)
tc.properties.update(<props>)
tc.c_args = [<C defines>]
tc.generate()
meson = Meson(self)
meson.configure()
meson.build()
```
But now the `build` function doesn't take any arguments that it passes along to the backend.

AbrilRBS:
Changelog: Feature: Split build & compilation verbosity control to two confs.
Docs: https://github.com/conan-io/docs/pull/3277

Adds verbosity confs for:

| tool name | `build:verbosity` | `compilation:verbosity` |
|------|-------|-------------|
| cmake      | ✔️ | ✔️ |
| xcodebuild*| ✔️ | ✔️ |
| msbuild    | ✔️ | - |
| meson      | ✔️ | ✔️ |
---------------------

* xcodebuild is not clear to me nor to @czoido what verbosity controls, so both confs activate it! Probably want to change it

Closes #13703
Closes #12239

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -104,9 +104,7 @@ class CMake(object):
         arg_list.extend(['-D{}="{}"'.format(k, v) for k, v in self._cache_variables.items()])
         arg_list.append('"{}"'.format(cmakelist_folder))
 
-        verbosity = self._verbosity
-        if verbosity is not None:
-            arg_list.append(f'--loglevel={verbosity}')
+        arg_list.extend(self._cmake_log_levels_args)
 
         if cli_args:
             arg_list.extend(cli_args)
@@ -138,6 +136,8 @@ class CMake(object):
             cmd_line_args.extend(build_tool_args)
         if cmd_line_args:
             args += ['--'] + cmd_line_args
+
+        args.extend(self._compilation_verbosity_arg)
 
         arg_list = ['"{}"'.format(bf), build_config, cmd_args_to_string(args)]
         arg_list = " ".join(filter(None, arg_list))
@@ -185,6 +185,7 @@ class CMake(object):
         arg_list = ["--install", build_folder, build_config, "--prefix", pkg_folder]
         if component:
             arg_list.extend(["--component", component])
+        arg_list.extend(self._compilation_verbosity_arg)
         arg_list = " ".join(filter(None, arg_list))
         command = "%s %s" % (self._cmake_program, arg_list)
         self._conanfile.run(command)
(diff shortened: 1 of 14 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -209,3 +213,19 @@ def test(self, build_type=None, target=None, cli_args=None, build_tool_args=None
...
+
+    @property
+    def _verbosity(self):
+        verbosity = self._conanfile.conf.get("tools.build:verbosity")
+        if verbosity:
+            # ERROR, WARNING, NOTICE, STATUS (default), VERBOSE, DEBUG, or TRACE
+            verbosities = {
+
+            }
+            if verbosity not in ("quiet", "error", "warning", "notice", "status", "normal",
```
memsharded: lets add a ``conf.get("myconf", choices=iterable-of-choices)`` to do the validation inside ``conf.get``, in the same way we check the return type? Because it seems this pattern is common enough

AbrilRBS: 👍 good idea

franramirez688: I think that implementation should have a dedicated PR and keep this PR only fixing what it's supposed to fix.

From my point of view, I think that `choices` should have some kind of predefined fixed values to make them automatically checked (apart from this development which would be necessary as well). For instance, `verbosity`, `compiler_executables`, `microsoft.bash:subsystem`, etc.  could have those fixed values, won't they? Then, it could be better to have something like:


```python
BUILT_IN_CONFS = {
"conf_1": ("doc", DEFAULT, [CHOICE_1, CHOICE_2]),
...
}
```

I'm not saying that this implementation should be the only one but it could be a proposal.

AbrilRBS: I like your idea about having the type be part of the conf definition, I have some questions about how it would interact with non-built-in confs, but this is something we can discuss :)

franramirez688: Non-built-in confs would use your proposal I guess (`conf.get("xxxx", choices=())`), but sure, we can discuss it later.

franramirez688 (approved):

memsharded (approved):
Please check the broken test

AbrilRBS:
Can't repro locally, I ran those tests before commiting. Weird. 🤔

conan/tools/cmake/cmake.py
```diff
@@ -104,6 +104,10 @@ def configure(self, variables=None, build_script_folder=None, cli_args=None):
         arg_list.extend(['-D{}="{}"'.format(k, v) for k, v in self._cache_variables.items()])
         arg_list.append('"{}"'.format(cmakelist_folder))
 
+        verbosity = self._verbosity
+        if verbosity is not None:
+            arg_list.append(f"--log-level={verbosity}")
```
SpaceIm: @RubenRBS I believe it's unrelated to https://github.com/conan-io/conan/issues/12239 (you mention this issue as closed by this PR).

`--log-level` doesn't control what is displayed during build. https://github.com/conan-io/conan/issues/12239 is asking for a conf enabling https://cmake.org/cmake/help/latest/variable/CMAKE_VERBOSE_MAKEFILE.html in CMake, to display all build commands (usually compilation & linker commands).
https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-log-level controls the maximum level of `message()` type displayed by CMake during configuration.

So it seems that this PR is trying to map a common conf to arguments which have very different meaning in these build system.

For example in meson, `--verbose` controls verbosity level of make & ninja, so I guess this PR properly addresses https://github.com/conan-io/conan/issues/9772 & https://github.com/conan-io/conan/issues/13703. But in CMake, the same feature is `CMAKE_VERBOSE_MAKEFILE`.
In both case all these values (vv, debug, trace, error etc) don't make sense, it's enabled or disabled.

(regarding XCodeBuild & MSBuild, I don't know, it might also be a different meaning).

AbrilRBS: You're right @SpaceIm in that the conf is mapping to flags with different meanings in each build system, this was somewhat on purpose, but you're right that it's unexpected. Let me rethink this with the team :)

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -730,3 +730,13 @@ def build(self):
...
         assert "using FindComandante.cmake" in client.out
+
+
+@pytest.mark.tool("cmake")
+def test_cmake_verbosity():
+    tc = TestClient()
+    tc.run("new cmake_exe -d name=pkg -d version=1.0")
+    tc.run("create . -c tools.build:verbosity=verbose -c tools.compilation:verbosity=verbose")
+    assert "--loglevel=VERBOSE" in tc.out
+    assert "unrecognized option" not in tc.out
```
memsharded: Maybe a bit heavy test to just check the ``--verbose`` output?. Maybe it could be reduced, removing the actual compilation, and avoiding slow compiler checks instead of the ``cmake_exe`` template?

memsharded: Or integrate the checks in an existing test

memsharded (approved):
Looks very good

conans/test/functional/toolchains/cmake/test_cmake.py
AbrilRBS: Yes, that's what I'm going to do, this test was done before the others, didn't rememebr to do the same as in those :)

franramirez688 (approved):

PR #13731, opened 2023-04-20
Add test showing getting version of root in deployer

AbrilRBS:
Changelog: Omit

Some ideas as to how to access the version of the generated package. If this is something we want to document, a similar example might be useful in the docs

### DIFF
--- a/conans/test/integration/extensions/test_deployers.py
+++ b/conans/test/integration/extensions/test_deployers.py
@@ -1,0 +1,29 @@
+import textwrap
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+def test_deployer_custom_version_output():
+    deployer = textwrap.dedent("""
+    import os
+    def deploy(graph, output_folder, **kwargs):
+        # When running graph info with --requires, the root of the graph is a virtual node, skip it
+        if graph.root.recipe == "Cli":
+            ref_version = str(graph.nodes[1].ref)
+        else:
+            ref_version = str(graph.root.ref)
+        # ref_version is of the form name/version{@user/channel}
+        # More granular control is possible by accessing ref.name, ref.version etc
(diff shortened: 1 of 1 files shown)


conans/test/integration/extensions/test_deployers.py
```diff
@@ -0,0 +1,29 @@
...
+from conans.test.utils.tools import TestClient
+
+
+def test_deployer_custom_version_output():
+    deployer = textwrap.dedent("""
+    import os
+    def deploy(graph, output_folder, **kwargs):
+        # When running graph info with --requires, the root of the graph is a virtual node, skip it
+        if graph.root.recipe == "Cli":
+            ref_version = str(graph.nodes[1].ref)
```
memsharded, line 13: This seems a bit weird, and the access with ``[1]`` probably internal implementation detail that might break in the future.

memsharded (commented):
IIUC, the use case is a deployer that works the same for a local ``<path>`` installation than for a ``--requires=mypkg`` installation?
Because it is not only the ``ref`` of the package, but everything changes, like the layout, the folders that need to be managed to do copies or whatever... I am not sure if this wide scope of the deployer is the best approach, but I have no better ideas, lets talk about it.

AbrilRBS:
It working for both ways is just me trying to show that it _can_ be done, but the real use case is to be able to get the ref of the recipe the user is interested in, so that a deploy can be made relative to that name/version. I did it this "general" way so that the linked issue has as much of a staring point base as possible

conans/test/integration/extensions/test_deployers.py
AbrilRBS: Yes, you're right. We need to think of a way to make this more handy

AbrilRBS:
Let's maybe document a tiny bit better how the user could find this data if needed, let's move it to the docs

PR #13741, opened 2023-04-20
allow non-relocatable packages for upload_policy=skip

### LINKED ISSUE #13237: [question] Non-relocatable packages don't work in 2.0. Is this a supported use case?
### What is your question?

Is there a way to force packages to be built directly into their final cache location in conan 2.0? I have a number of non-relocatable packages, and they have `~/.conan2/p/t/<id>` rather than `~/.conan2/p/<id>` embedded into the files which makes everything non functional because the path with `t` in it doesn't exist when it comes time to consume the packages.

My assumption is that the `t` in the path here is shorthand for `temporary`. What would resolve my problem would be to be able to force all operations in the `package(self)` method to operate on the `~/.conan2/p/<id>` directory rather than `~/.conan2/p/t/<id>`.

memsharded:
Changelog: Feature: Allow non-relocatable packages for ``upload_policy=skip``

Close https://github.com/conan-io/conan/issues/13237

### DIFF
--- a/conan/api/subapi/export.py
+++ b/conan/api/subapi/export.py
@@ -30,11 +30,13 @@ class ExportAPI:
         out.info("Exporting binary from user folder to Conan cache")
         conanfile = pkg_node.conanfile
 
+        skip_upload = conanfile.upload_policy == "skip"
+
         package_id = pkg_node.package_id
         assert package_id is not None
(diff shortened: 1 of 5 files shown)


memsharded:
superseded by #13833

PR #13757, opened 2023-04-24
Allow specifying a custom output for the deployers

### LINKED ISSUE #13728: [question] Conan 2.0 custom deployer + output folder
### What is your question?

Hi, I am trying to deploy the binaries of a package with Conan 2.0. I created a simple custom deployer (my_deployer.py) that basically copies the binaries directly to the output folder:

```

def deploy(graph, output_folder, **kwargs):
    for name, dep in graph.root.conanfile.dependencies.items():
        source_folder = str(dep.folders.package_folder)
        copy(dep, "*.dll", source_folder, output_folder)

```
According to Conan 2.0 I should be specifiying the custom deployer together AND the output folder as arguments to the install command like:

`conan install --requires my_package/1.0.0@foo/testing  --deploy my_deployer --output-folder <path_to_output_folder> `

The problem is that by specifying --output-folder argument alone the install command will deploy generated files (like conanbuild.bat) in addition to the files my custom deployer copies to the very same output folder. I am trying to reproduce the same behavior deploy() method had in Conan 1.x, that´s it, deploying ONLY the files my custom deployer deploys.  However I did not figure out yet how to that.

AbrilRBS:
Changelog: Feature: Allow to specify a custom deployer output folder.
Docs: https://github.com/conan-io/docs/pull/3275

Closes #13728


conan/cli/commands/graph.py
```diff
@@ -118,6 +119,8 @@ def graph_info(conan_api, parser, subparser, *args):
                            help='Print information only for packages that match the patterns')
     subparser.add_argument("-d", "--deployer", action="append",
                            help='Deploy using the provided deployer to the output folder')
+    subparser.add_argument("--output-deployer",
```
memsharded: I guess not ``--deployer-output`` or ``--deployer-folder`` to avoid conflicts/overlap with ``--deployer`` in ``argparse``?
Maybe something that contains ``folder`` is a bit more descriptive/intuitive?

conan/cli/commands/graph.py
```diff
@@ -166,7 +169,8 @@ def graph_info(conan_api, parser, subparser, *args):
                                                       clean=args.lockfile_clean)
         conan_api.lockfile.save_lockfile(lockfile, args.lockfile_out, os.getcwd())
         if args.deployer:
-            base_folder = os.getcwd()
+            base_folder = args.output_deployer or os.getcwd()
+            mkdir(base_folder)
```
memsharded: There is already a ``mkdir()`` for this inside ``do_deploys``, this shouldn't be necessary?

memsharded (approved):
Looks good!
I'd say we can include this in 2.0.5

memsharded:
Regarding the "development re-locatable Conan-agnostic" deploys feature https://github.com/conan-io/examples2/pull/94, this could be a useful use case:

- it allows to do multiple ``conan install``, different configurations, like ``linux/windows``, or different compilers or architectures will require a different ``--output-folder`` or defined ``build_folder_vars`` to not overwrite each other
- But it is not intended that every one will contain its own ``full_deploy`` copy, because for example ``tool_requires`` for the build system for a given ``cmake.exe``, could contain multiple copies. An extra argument to redefine the full_deploy root folder could solve this.

AbrilRBS:
Picking this back up, the PR felth thru the cracks last month

conan/cli/commands/graph.py
AbrilRBS: Not that I can see?

conan/cli/commands/graph.py
AbrilRBS: It works good with `--deployer`, but some tests (and users!) might still be using the old `--deploy` syntax, which _would_ conflict. Do we force users to not use `--deploy` anymore and fix the tests?

memsharded (approved):

PR #13771, opened 2023-04-26
avoid hitting servers for binaries when upload_policy=skip

### LINKED ISSUE #13770: [feature] build_policy = "missing" and upload_policy = "skip" should work when later remotes are offline
### What is your suggestion?

Package download ignores `upload_policy = "skip"` and still looks for binaries that, according to that recipe, could never be uploaded. After failing to find binaries in the remote from which it retrieved the recipe, that means it will search subsequent remotes too, not only the one in which that recipe was found (and still not find anything, since the recipe would never allow binaries to be uploaded.

This is undesirable when there are additional remotes (later in the priority order) that are not always accessible, e.g intranet-only ones that require signing into a corporate VPN. `conan install` will error out upon failing to check them, even though it already found the requested recipe, and that recipe told it no binaries that could have been uploaded there (or anywhere else)

memsharded:
Changelog: Feature: Optimize and avoid hitting servers for binaries when ``upload_policy=skip``
Docs: Omit

Close https://github.com/conan-io/conan/issues/13770

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -1,3 +1,4 @@
+from conan.api.output import ConanOutput
 from conans.client.graph.build_mode import BuildMode
 from conans.client.graph.compatibility import BinaryCompatibility
 from conans.client.graph.compute_pid import compute_package_id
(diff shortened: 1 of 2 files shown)


conans/client/graph/graph_binaries.py
```diff
@@ -23,14 +24,9 @@ def __init__(self, conan_app):
...
-        # For cascade mode, we need to check also the "modified" status of the lockfile if exists
-        # modified nodes have already been built, so they shouldn't be built again
+        # check dependencies, if they are being built, "cascade" will try to build this one too
         if build_mode.cascade:
-            for dep in node.dependencies:
-                dep_node = dep.dst
-                if dep_node.binary == BINARY_BUILD:
-                    with_deps_to_build = True
-                    break
+            with_deps_to_build = any(dep.dst.binary == BINARY_BUILD for dep in node.dependencies)
```
memsharded, line 29: this is a pure refactor, no functional change

czoido (approved):

franramirez688 (approved):

PR #13788, opened 2023-04-28
fixed graph_lock.py - lock_node(): consider context

### LINKED ISSUE #13775: Multiple versions of same build_requires package in lockfile?
Hello,

we are facing the issue that we are generatng a lockfile and in this lockfile we have seen that the same build_requires package is listed twice (with different versions).

After analysing this we have found out that one version is provided by the profile and one by the recipe itself.

Is that supposed to be allowed? I would have expected this to be forbidden.
Also when installing using the lockfile which version is then installed?

BTW. We are using Conan 1.59.

SzBosch:
fixes #13775

Changelog: Bugfix: Fixed graph_lock.py - lock_node(): consider context of build_require from lockfile.
Docs: Omit

### DIFF
--- a/conans/model/graph_lock.py
+++ b/conans/model/graph_lock.py
@@ -3,7 +3,7 @@ import os
 from collections import OrderedDict
 
 from conans import DEFAULT_REVISION_V1
-from conans.client.graph.graph import RECIPE_VIRTUAL, RECIPE_CONSUMER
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Fantastic, thanks very much, happy to know that this bug seems possible to fix.
I will be contributing the tests later, and I have assigned it for next 1.60, that seems will be released likely end of next week, beginning of the following one.

PR #13794, opened 2023-04-30
trying to simplify components

memsharded:
Changelog: Omit
Docs: Omit

Internal simplifications of CppInfo:

- Dicts are now ordered, we can drop DefaultOrderedDict
- the CppInfo shallow copy seems doing nothing

### DIFF
--- a/conan/tools/premake/premakedeps.py
+++ b/conan/tools/premake/premakedeps.py
@@ -1,9 +1,8 @@
-import itertools 
+import itertools
 import glob
 import re
 
 from conan.internal import check_duplicated_generator
-from conans.model.build_info import CppInfo
 from conans.util.files import save
 
 # Filename format strings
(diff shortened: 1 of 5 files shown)


conan/tools/premake/premakedeps.py
```diff
@@ -84,34 +83,31 @@
...
-            '"%s"' % p for p in deps_cpp_info.cxxflags) if deps_cpp_info.cxxflags else ""
-        self.cflags = ", ".join(
-            '"%s"' % p for p in deps_cpp_info.cflags) if deps_cpp_info.cflags else ""
-        self.sharedlinkflags = ", ".join('"%s"' % p.replace('"', '\\"') for p in
-                                         deps_cpp_info.sharedlinkflags) \
-            if deps_cpp_info.sharedlinkflags else ""
-        self.exelinkflags = ", ".join('"%s"' % p.replace('"', '\\"') for p in
-                                      deps_cpp_info.exelinkflags) \
-            if deps_cpp_info.exelinkflags else ""
+    def __init__(self, dep_cpp_info, pkg_folder):
```
memsharded, line 86: This PremakeDeps refactor is using the ``package_folder`` directly in the generation, instead of modifying the dependencies ``cpp_info`` objects.

conans/client/installer.py
```diff
@@ -414,12 +414,12 @@ def _call_package_info(self, conanfile, package_folder, is_editable):
                     # cppinfos inside
 
                     # convert directory entries to be relative to the declared folders.build
-                    build_cppinfo = conanfile.cpp.build.copy()
+                    build_cppinfo = conanfile.cpp.build
```
memsharded, line 417: This copy() was shallow, so modifications (making paths absolute with ``set_relative_base_folder``) was still mutating the dependency ``cpp.build`` object.

czoido (approved):

conans/client/installer.py
franramirez688: Was it a bug? Was it simply working because we actually wanted to change the dependency `cpp.build` object?

franramirez688 (approved):

conans/client/installer.py
memsharded: It was unnecessary, not a bug: The copy is shallow, does not really protect against mutating the internal lists, which is what the ``set_relative_base_folder()`` is doing. Also, it doesn't matter that such state is mutated, such changes are already being applied to ``cpp_info`` some lines below ``conanfile.cpp_info.merge(full_editable_cppinfo, overwrite=True)``, so it makes sense that the ``cpp.source`` and ``cpp.build`` are also updated.

PR #13795, opened 2023-04-30
Feature/simplify cpp info2

memsharded:
Changelog: Omit
Docs: Omit

Continuation of https://github.com/conan-io/conan/pull/13794
Refactors of:
- CppInfo models the global conf as ``._package`` instead of ``.components[None]``. There was some ugly fallback of attribute indirection to it
- Extracted repeated code to ``_Component`` and calling it from ``CppInfo``
- Avoiding the check for ``componentname is None`` in several places (can probably be done more in generators)

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -513,11 +513,11 @@ class CppInfo:
         - Check that every internal component require actually exist
         It doesn't check that external components do exist
         """
-        if not self.has_components and not self.requires:
+        if not self.has_components and not self._package.requires:
             return
         # Accumulate all external requires
-        external = set()
-        internal = set()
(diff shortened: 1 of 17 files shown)


AbrilRBS (approved):

conans/model/build_info.py
```diff
@@ -334,124 +334,121 @@ def get_property(self, property_name):
             pass
 
     def get_init(self, attribute, default):
+        # Similar to dict.setdefault
```
franramirez688, line 337: Not for now, but we should change it to `setdefault` for instance. `get_init` sounds weird, wdyt?

franramirez688 (approved):

czoido (approved):

PR #13796, opened 2023-05-01
test settings.possible_values()

### LINKED ISSUE #13793: [feature] A way to query values in settings.yml from conanfile.py
### What is your suggestion?

Hi Team,

It would be really useful from a compatibility perspective to be able to query the possible values for a given setting in settings.yml from a conanfile.py. To illustrate this with a usecase, I've been running into issues with libcxx compatibility in tool_requires packages, specifically doxygen as documented in https://github.com/conan-io/conan-center-index/issues/17034 with a potential fix in https://github.com/conan-io/conan-center-index/pull/17051.

Ultimately, this has lead to a `compatibility()` definition as follows:

```python
    @property
    def _conan_home(self):
        conan_home_env = "CONAN_USER_HOME" if Version(conan_version).major < 2 else "CONAN_HOME"
        conan_home_dir = ".conan" if Version(conan_version).major < 2 else ".conan2"
        return os.environ.get(conan_home_env, pathlib.Path(pathlib.Path.home(), conan_home_dir))

    def compatibility(self):
        # is there a better way of reading all possible versions from settings.yml?
        settings_yml_path = pathlib.Path(self._conan_home, "settings.yml")
        with open(settings_yml_path, "r") as f:
            settings_yml = yaml.safe_load(f)

        compatible_versions = [{"settings": [("compiler.version", v), ("build_type", "Release")]}
            for v in settings_yml["compiler"][str(self.settings.compiler)]["version"] if v <= Version(self.settings.compiler.version)]
        return compatible_versions
```

This is necessary, because this compatibility is based on possible compiler versions - if the available binary was built with an older version of the compiler then it will be compatible, so my understanding is that I need to provide a list of the possible values an "older version of the compiler" could possibly have here. 

This raises additional questions about the way libcxx is being modelled (similar to the questions around libc), where the compiler is being used as a heuristic to model the libcxx dependency, but these are probably out of scope for this issue so lets leave that for another thread.

The above `compatibility()` definition works when creating/consuming the package locally for me, but alas it does not pass the sniff test when executing `conan info` on the CI servers because it fails to locate `settings.yml`. I'm not sure how robust my logic is above to identifying the conan home, but it strikes me that it would be useful to have a direct hook into the conan home being used by the conanfile at the time of interpretation, rather than relying on some imperfect heuristic as I am above. This would also be useful for custom user settings to model the libc compatibility described in https://github.com/conan-io/docs/issues/3960.

If there's a better way to approach this, any advice is appreciated.

memsharded:
Changelog: Feature: New ``settings.possible_values()`` method to query the range of possible values for a setting.
Docs: https://github.com/conan-io/docs/pull/3212
Close https://github.com/conan-io/conan/issues/13793

### DIFF
--- a/conans/client/conf/__init__.py
+++ b/conans/client/conf/__init__.py
@@ -7,7 +7,7 @@ os:
     WindowsStore:
         version: ["8.1", "10.0"]
     WindowsCE:
-        platform: ANY
+        platform: [ANY]
         version: ["5.0", "6.0", "7.0", "8.0"]
     Linux:
     iOS:
--- a/conans/model/settings.py
+++ b/conans/model/settings.py
@@ -152,12 +152,12 @@ class SettingsItem(object):
         if isinstance(self._definition, dict):
             self._definition[self._value].validate()
 
-    def get_definition(self):
+    def possible_values(self):
         if isinstance(self._definition, list):
             return self.values_range.copy()
         ret = {}
         for key, value in self._definition.items():
-            ret[key] = value.get_definition()
+            ret[key] = value.possible_values()
         return ret
 
     def rm_safe(self, name):
@@ -341,11 +341,10 @@ class Settings(object):
                 result.append("%s=%s" % (name, value))
         return '\n'.join(result)
 
-    def get_definition(self):
-        """Check the range of values of the definition of a setting. e.g:
-           get_definition_values("compiler.gcc.version") """
-
+    def possible_values(self):
+        """Check the range of values of the definition of a setting
+        """
         ret = {}
         for key, element in self._data.items():
-            ret[key] = element.get_definition()
+            ret[key] = element.possible_values()
         return ret
--- a/conans/model/settings.py
+++ b/conans/model/settings.py
@@ -154,7 +154,7 @@ class SettingsItem(object):
 
     def get_definition(self):
         if isinstance(self._definition, list):
-            return [e if e != 'None' else None for e in self.values_range]
+            return self.values_range.copy()
         ret = {}
         for key, value in self._definition.items():
             ret[key] = value.get_definition()
--- a/conans/test/integration/settings/test_settings_definition.py
+++ b/conans/test/integration/settings/test_settings_definition.py
@@ -41,10 +41,19 @@ def test_settings_definitions_compiler():
                 self.output.info("HOST: " + ", ".join(definition))
                 definition = self.settings_build.compiler.version.get_definition()
                 self.output.info("BUILD: " + ", ".join(definition))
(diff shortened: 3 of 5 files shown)


conans/test/integration/settings/test_settings_definition.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def test_settings_definitions():
+    c = TestClient()
+    conanfile = textwrap.dedent("""
+        from conan import ConanFile
+        class Pkg(ConanFile):
+            settings = "os"
+            def generate(self):
+                definition = self.settings.get_definition()
```
prince-chrismc: How could users disambiguate (too much ☕ I have big words) or tell a part a `None` from a "not present" value?

Like if a profile was missing cppstd from 1.x and not in the graph then `self.settings.compiler.cppstd.get_definition()` would be None?

conans/client/conf/__init__.py
```diff
@@ -7,7 +7,7 @@
     WindowsStore:
         version: ["8.1", "10.0"]
     WindowsCE:
-        platform: ANY
+        platform: [ANY]
```
memsharded, line 10: This was a bug, realized while printing things.

conans/model/settings.py
```diff
@@ -154,7 +154,7 @@ def validate(self):
 
     def get_definition(self):
         if isinstance(self._definition, list):
-            return [e if e != 'None' else None for e in self.values_range]
```
memsharded, line 157: We no longer use the ``"None"`` literal string for this

conans/test/integration/settings/test_settings_definition.py
memsharded: Not sure what you mean, that gets the range of values (added it to the test):
```python
cppstds = self.settings.compiler.cppstd.get_definition()
self.output.info("CPPSTDS: " + str(cppstds))
````
profile:
```
        [settings]
        compiler=msvc
        compiler.version=192
        compiler.runtime=dynamic
```
```
conan install . -pr=profile -s:b compiler=gcc
...
conanfile.py: CPPSTDS: [None, '14', '17', '20', '23']
```
Python ``None`` will be in the range of possible values

memsharded:
Naming possibilities:

- ``defined_values()``
- ``values_ranges()``
- ``possible_values()``
- ?

conans/test/integration/settings/test_settings_definition.py
prince-chrismc: Exactly this was the case I was thinking :)

prince-chrismc:
defined/possible _values makes the most sense to me... slight preference to defined :)

czoido (approved):

PR #13801, opened 2023-05-02
new test for CMake+editable+components

memsharded:
Changelog: Omit
Docs: https://github.com/conan-io/docs/pull/3220

https://github.com/conan-io/conan/issues/13639


franramirez688 (approved):
@memsharded this test is not using the components at all, is it?

memsharded:
> @memsharded this test is not using the components at all, is it?

It was failing in different places, not in the consumption with ``MyGreetings::MyGreetings``, but that makes sense, I will add some extra checks to validate it can link with a single component too.

franramirez688 (approved):

PR #13807, opened 2023-05-03
add option to sepcify the version for export_conandata_patches and ap…

boussaffawalid:
Changelog: Feature : add version param for apply_conandata_patches and export_conandata_patches

In our case conan.version is changing dynamically for every git commit. we use git sha as part of the version.
 
### DIFF
--- a/conan/tools/files/patches.py
+++ b/conan/tools/files/patches.py
@@ -63,7 +63,7 @@ def patch(conanfile, base_path=None, patch_file=None, patch_string=None, strip=0
         raise ConanException("Failed to apply patch: %s" % patch_file)
 
 
-def apply_conandata_patches(conanfile):
+def apply_conandata_patches(conanfile, version=None):
     """
     Applies patches stored in 'conanfile.conan_data' (read from 'conandata.yml' file). It will apply
     all the patches under 'patches' entry that matches the given 'conanfile.version'. If versions are
@@ -103,8 +103,9 @@ def apply_conandata_patches(conanfile):
         return
 
     if isinstance(patches, dict):
-        assert conanfile.version, "Can only be applied if conanfile.version is already defined"
-        entries = patches.get(conanfile.version, [])
+        assert version or conanfile.version, "Can only be applied if version or conanfile.version is already defined"
+        version = version if version else conanfile.version
+        entries = patches.get(version, [])
     elif isinstance(patches, list):
         entries = patches
     else:
@@ -122,7 +123,7 @@ def apply_conandata_patches(conanfile):
                                  " entry for every patch to be applied")
 
 
-def export_conandata_patches(conanfile):
+def export_conandata_patches(conanfile, version=None):
     """
     Exports patches stored in 'conanfile.conan_data' (read from 'conandata.yml' file). It will export
     all the patches under 'patches' entry that matches the given 'conanfile.version'. If versions are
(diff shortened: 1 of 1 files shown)


memsharded (commented):
Can you please elaborate a bit more about the use case?

The ``apply_conandata_patches`` is mostly for third-party, ConanCenter-like packages, which version would be defined by the upstream version like ``zlib/1.2.13``, so not very clear how the git sha would be used.

There would be interactions also with the trimming of versions different to the package one done by the hook dedicated to that purpose (active in ConanCenter by default, but should be used by anyone following the ConanCenter recipes structure with conandata.yml), which would trim the patches even if you try to use them, because the patches will be under a different version than the current one.

memsharded:
Hi @boussaffawalid 

Any update regarding this PR?
It broke the CI, but that is not very important. In order to better consider this it would be necessary:

- Explain a bit better the use case. It is not clear the problem solved.
- Add some unit-test that exemplify the functionality and covers it
- Fix other broken tests in CI

boussaffawalid:
@memsharded we want to reuse `apply_conandata_patches` and `export_conandata_patches` for our internal projects, However we use a dynamic version that include git sha as part of the version so that every commit generate a different version. 

Here is what we want to using zlib as a example

```
class ZlibConan(ConanFile):
    name = "zlib"
    upstream_version = "1.2.13"

    def set_version(self):
        self.version = self.upstream_version + "-" + git_sha_short

    def export_sources(self):
        export_conandata_patches(self, self.upstream_version)
```

My changes are supposed to be backward compatible and I am Working on fixing the tests, but it seems like I cannot run them locally. I tried to run a basic test using 
```
python -m pytest conans/test/functional/basic_build_test.py
```
but I got the error:
```
Failed: Required 'cmake' tool version 'Any' is not available
```
Screenshot 2023-05-30 at 2 00 11 PM

memsharded:
Thanks for clarifying.

I'd like to ask why aren't you using the ``recipe_revision``, if you want to use the ``git`` commit as the revision, just adding ``revision_mode = "scm"`` and you will get it, and it will have extra advantages as being able to automatically resolve to latest if you want.

boussaffawalid:
We used recipe_revision initially but we had two issues back then. 

* Builds were not deterministic before conan introduced lockfile, the same branch build could pass and then fail if there is a new revision for the same dependency version. 
* We had to use the `--update` flag to always pull latest revision which was extremely slow. We have a large dependency graph with around 150 conan packages.

memsharded:
What I am thinking is about explicit versioning.
If you already create a new ``zlib/1.2.13-commit1`` version, you need to require it as ``requires = "zlib/1.2.13-commit1"`` explicitly in order to use it, don't you? How do you require your dependencies if not?

If you are doing it, then using the recipe revision will be replacing ``-`` by ``#`` to ``requires="zlib/1.2.13#commit1"``. No need of lockfiles, no need to ``--update``, when using explicit revisions things resolve automatically.

And then, only if in the future you want to drop the revision and write ``requires = "zlib/1.2.13"`` only, then you will be able to do it (and then, yes, using lockfiles for reproducibility).

boussaffawalid:
@memsharded yes we do `requires = "zlib/1.2.13-commit1"`
I was not aware that you can conan support this syntax `requires="zlib/1.2.13#commit1"`. Thanks. I guess we will switch to it eventually.

I guess this feature is not needed then.

memsharded:
Great, thanks for the feedback.

> I was not aware that you can conan support this syntax requires="zlib/1.2.13#commit1". Thanks. I guess we will switch to it eventually.

Yes, absolutely. It also works in ``conanfile.txt`` and in all CLI commands, like ``conan install --requires=zlib/1.2.13#recipe_revision``.

> I guess this feature is not needed then.

I am closing this PR at the moment, but please let me know how it goes, and if needs to be re-considered, we can re-open and talk again.

In any case, until you move to use the revisions, you might easily rely in practice to something like:
```python
def exports_sources(self):
     for patch in conanfile.conan_data["patches"][self.upstream_version]:
          shutil.copy2(os.path.join(..., path["patch_file"]), ...)
```
As most of the methods are boilerplate to handle some unlikely error scenarios, I think it is possible to implement the equivalent in just 2-3 lines of code.

 Thanks!

PR #13810, opened 2023-05-03
do not move cache build folders to allow debugging in all cases

### LINKED ISSUE #13799: [bug] Debug package source path set incorrectly
### Environment details

* Operating System+version:  Win11 21H2
* Compiler+version: msvc v193
* Conan version: 2.0.4
* Python version: 3.11.0

I am trying to use a debugger to step into a conan package (ffmpeg on windows). To do this, I create a debug version of the package and use the debug configuration in the build. When I attempt to step into the package, its seems like it has the symbols, but the source code location is incorrectly set to the [temp file location](https://github.com/conan-io/conan/blob/deb13ad983c89386fb2cd8f54e5daf63ba3966b2/conans/client/cache/cache.py#L59), rather than the actual package source location. Maybe this is a bug?

Visual Studio lets me know that the source file it's expecting to find in the temp folder (`.conan2/p/t/ffmpe...`) cannot be found and prompts for the correct location. I can then use `conan cache path --folder source ffmpeg/$VERSION` to get the location to sources the debugger can use. At this point I can debug regularly

### Steps to reproduce

1. Use conan2
2. Build and install debug version of ffmpeg to conan cache
3. Add dependency to on ffmpeg to visual studio project using [MSBuildDeps ](https://docs.conan.io/2/reference/tools/microsoft/msbuilddeps.html#conan-tools-microsoft-msbuilddeps) generator (need to add the path to the generated `conan_ffmpeg.props` file in .vcxproj file for appropriate configurations)
4. Set visual studio solution configuration to Debug
5. Add break point somewhere and attempt to step into ffmpeg

### Logs

_No response_

memsharded:
Changelog: Feature: Do not move "build" folders in cache when ``package-revision`` is computed to allow locating sources for dependencies debuggability with step-into
Docs: omit

Fix https://github.com/conan-io/conan/issues/13799

NOTE: This is mostly an experiment, no guarantees this makes sense

### DIFF
--- a/conan/api/subapi/cache.py
+++ b/conan/api/subapi/cache.py
@@ -1,4 +1,4 @@
-import shutil
+import os
 
 from conan.internal.conan_app import ConanApp
 from conan.internal.integrity_check import IntegrityChecker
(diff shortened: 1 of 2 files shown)


memsharded:
Superseded by https://github.com/conan-io/conan/pull/13833

PR #13811, opened 2023-05-04
Add nicer error for None backup remote

AbrilRBS:
Changelog: Omit
Docs: Omit

CI scripts might fail when inserting their own download URL. Make sure we have a nice error for those cases.

The old error was an uncaught `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'` exception which did not make it clear what the issue was. This differed from other errors in this section of the code in that those do show what the current conf value is (Or at least the offending one), so this PR just makes it easier to debug a None in the conf that might have gotten there by some CI issue

### DIFF
--- a/conans/client/downloaders/caching_file_downloader.py
+++ b/conans/client/downloaders/caching_file_downloader.py
@@ -62,10 +62,11 @@ class SourcesCachingDownloader:
                 self._output.info(f"Source {urls} retrieved from local download cache")
             else:
                 with set_dirty_context_manager(cached_path):
+                    if None in backups_urls:
+                        raise ConanException("Trying to download sources from None backup remote."
+                                             f" Remotes were: {backups_urls}")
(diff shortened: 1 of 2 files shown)


conans/client/downloaders/caching_file_downloader.py
```diff
@@ -63,6 +63,9 @@ def _caching_download(self, urls, file_path,
             else:
                 with set_dirty_context_manager(cached_path):
                     for backup_url in backups_urls:
+                        if backup_url is None:
+                            raise ConanException("Trying to download sources from None backup remote"
+                                                 f". Remotes were: {backups_urls}")
```
franramirez688: If we're going to raise an exception, should it not be better to put this condition outside of the loop?

memsharded (approved):

franramirez688 (approved):

PR #13833, opened 2023-05-06
use neww uuids for new builds folders in cache

### LINKED ISSUE #13237: [question] Non-relocatable packages don't work in 2.0. Is this a supported use case?

### LINKED ISSUE #13799: [bug] Debug package source path set incorrectly

memsharded:
Changelog: Feature: Allow step-into dependencies debugging for packages built locally with ``--build``
Changelog: Feature: Allow non relocatable, locally built packages with ``upload_policy="skip"`` and ``build_policy="missing"``
Docs: https://github.com/conan-io/docs/pull/3210

Close https://github.com/conan-io/conan/issues/13799
Close https://github.com/conan-io/conan/issues/13237

This PRs does the following:
- For the package binaries built in the cache, use a random UUID to compute their folder. Such folder will be different each time it is built, but it is also expected that the final package-revision would be different so it shouldn't be a big deal
- The same can **not** be done for recipes, because creating a new folder with a random uuid each time a recipe is exported would completely invalidate the ``source`` folder, making the caching of the sources in the cache completely useless (we could use a mechanism similar to the ``build_id`` lookup in the cache, but seems overkill)
- The downloaded packages still get downloaded to the same folder, not using uuid, but hashing the full reference including the package-revision, so they are always located in the same place (could help with compilation caches as ccache)
- It was necessary to augment a but the ``conan cache clean``, because builds that not succeed will leave extra trailing folders that can't be removed with ``conan remove`` as they don't have a final reference


AbrilRBS (approved):

conans/test/integration/command_v2/test_cache_clean.py
```diff
@@ -50,4 +56,61 @@ def test_cache_clean_all():
...
+    c.run("create .")
+    create_out = c.out
+    c.run("cache path pkg/0.1:da39a3ee5e6b4b0d3255bfef95601890afd80709")
+    path2 = str(c.stdout)
+    assert path2 in create_out
+    assert path1 != path2
+
+    builds_folder = os.path.join(c.cache_folder, "p", "b")
+    assert len(os.listdir(builds_folder)) == 1  # only one build
+    c.run('cache clean')
```
franramirez688, line 83: Should we add a test using `conan cache clean -b`?

franramirez688 (approved):

czoido (approved):

PR #13835, opened 2023-05-07
Refactor/remotes

memsharded:
Changelog: Omit
Docs: Omit

- Small refactor of Remotes that would be necessary for potential merge operations
- Small changes in the RemotesAPI receiving names instead of Remote objects for rename and update operations


AbrilRBS (approved):

conan/api/subapi/remotes.py
```diff
@@ -48,42 +48,39 @@ def disable(self, pattern):
...
         for remote in remotes:
             app.cache.remotes_registry.remove(remote.name)
             users_clean(app.cache.localdb, remote.url)
 
-    def update(self, remote: Remote):
-        app = ConanApp(self.conan_api.cache_folder)
-        app.cache.remotes_registry.update(remote)
-
-    def move(self, remote: Remote, index: int):
+    def update(self, remote_name, url=None, secure=None, disabled=None, index=None):
```
memsharded, line 76: This API was changed, to not require passing a Remote object, but just the updates to do. The problem with passing the Remote object is that the update semantics are more implicit/hidden, and there can be more inconsistencies like coming with a new ``remote.name`` not existing in the current ones.

conan/api/subapi/remotes.py
```diff
@@ -48,42 +48,39 @@ def disable(self, pattern):
...
         app = ConanApp(self.conan_api.cache_folder)
-        app.cache.remotes_registry.move(remote, index)
+        app.cache.remotes_registry.update(remote_name, url, secure, disabled=disabled,
+                                          index=index)
 
-    def rename(self, remote: Remote, new_name: str):
+    def rename(self, remote_name: str, new_name: str):
```
memsharded, line 81: Same, this API was changed to only require the previous name and the new one, not a full ``Remote`` object to just change the name

czoido (approved):

PR #13839, opened 2023-05-08
Changed default -v level

franramirez688:
Changelog: Feature: `-v` argument defaults to the `VERBOSE` level.
Docs: omit

### DIFF
--- a/conan/api/output.py
+++ b/conan/api/output.py
@@ -51,17 +51,6 @@ class ConanOutput:
     # Singleton
     _conan_output_level = LEVEL_STATUS
     _silent_warn_tags = []
-    _output_levels = {"quiet": LEVEL_QUIET,  # -vquiet 80
-                      "error": LEVEL_ERROR,  # -verror 70
-                      "warning": LEVEL_WARNING,  # -vwaring 60
-                      "notice": LEVEL_NOTICE,  # -vnotice 50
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
What controls the default level now? `_conan_output_level`?

franramirez688:
> What controls the default level now? `_conan_output_level`?

The logic is exactly the same right now, what am I missing?

conan/api/output.py
```diff
@@ -51,6 +51,17 @@ class ConanOutput:
     # Singleton
     _conan_output_level = LEVEL_STATUS
     _silent_warn_tags = []
+    _output_levels = {"quiet": LEVEL_QUIET,  # -vquiet 80
```
memsharded: Making this global makes more difficult to see the actual change here.

Furthermore, this dict is used only once, inside the ``define_log_level``, so it could perfectly stayed there (giving it the  minimum necessary scope, no more)

franramirez688: I think it should be a class attribute because it should be something that it's purely defined by the class itself and not by a function. Anyway, I can revert that change for sure.

memsharded: IMO, not really. If it is used only inside one method/function of the class, it is an internal implementation detail of that method/function, no need to give it class scope.

memsharded (approved):

PR #13841, opened 2023-05-08
CMakeDeps: Fix imported library config suffix

### LINKED ISSUE #13504: [bug] In the CMakeDeps Generator IMPORTED_LOCATION is overridden by IMPORTED_LOCATION${config_suffix}
### Environment details

* Compiler+version: MSVC 19
* Conan version: 2.0.2
* Python version: 3.10

### Steps to reproduce

1. Download a dependency (e.g. thrift) with conan using the CMakeDeps generator
2. Have a look at the created cmakedeps_macros.cmake

In the function `conan_package_library_targets` line 48 and 49 for windows the correct target properties are set for IMPORTED_IMPLIB and IMPORTED_LOCATION. 

`set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION ${CONAN_SHARED_FOUND_LIBRARY})
 set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_IMPLIB ${CONAN_FOUND_LIBRARY})`

Unfortunately those values are superseded by the line 61 where a per config value for IMPORTED_LOCATION_${config_suffix} is set, making the previous setting useless.

`set_target_properties(${_LIB_NAME} PROPERTIES IMPORTED_LOCATION${config_suffix} ${CONAN_FOUND_LIBRARY})`

3. Create a cmake project using 

`  add_custom_command(TARGET mytarget POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy_if_different $<TARGET_RUNTIME_DLLS:mytarget> $<TARGET_FILE_DIR:mytarget>)`

4. Instead of copying the dlls files to the binary folder upon building, the .lib files are copied.

### Logs

_No response_

Untzelmann:
Currently the imported libraries are set without config suffix first. At the end there is one additional set using the config suffix, which overwrites the already set libraries.

Fixes #13504

Changelog: Fix: Fix imported library config suffix.
Docs: Omit

Note: I think no documentation changes are required. Should I still make a change in the documentation?


memsharded (commented):
Thanks for your contribution!

A couple of suggestions:
- If we want this to be part of the 2.0.X patches, it needs to be targeted to the ``release/2.0`` branch. The ``develop2`` branch is intended for 2.1
- Some tests in the test suite that proves that this is solving an issue would be necessary, also to guarantee that this doesn't break in the future. Doing tests might be a bit more difficult, we can try to help contributing them or guiding.
- Signing the CLA is necessary for accepting contributions, please have a look to it when possible.

Untzelmann:
Ok, 

- I changed this to release/2.0. Will it still be in 2.1, or do I need to open a second pull request to get this into both versions?
- I signed the cla.
- I am unsure how to write the test: The issue we wanted to fix is: We have a dependency to a shared library on windows. We want to copy all required DLL files of an exectuable by cmake code like this

`add_custom_command(TARGET exe POST_BUILD
  COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_RUNTIME_DLLS:exe> ${OUTPUT_DIRECTORY}
  COMMAND_EXPAND_LISTS
  )`


Before the fix, this did not work. After the fix it does. In general, the issue is only observed in Windows, when the if-clause at the top of the function sets IMPORTED_IMPLIB and IMPORTED_LOCATION.

How to write a test for this?

AbrilRBS:
> I changed this to release/2.0. Will it still be in 2.1, or do I need to open a second pull request to get this into both versions?

It will be available in 2.1 too yes :)

> How to write a test for this?

The functional tests look like the place to do this, with an decorator guard to only execute on Windows (There are some examples already there), but let us know if you need extra help with that!

jcar87:
Thanks @Untzelmann - indeed this fixes a bug that existed with two different properties conflicting when on Windows with shared libraries, and the `.lib` file was being copied instead. 

I'll add the pertinent tests ahead of the merge, thanks so much for your contribution!

Untzelmann:
@jcar87 Thanks for adding the tests for me. I actually intended to have a look at implementing the a test for this, but I will probably not find time for this in the next days. So I am very thankful, if you add the test, and I will still have a look how to implement tests, for my next contributions.

jcar87:
Hi @Untzelmann  - thanks! Just pushed the test which replicates the problem, and tests that the DLL from a dependency is copied next to the executable (also tests that the executable actually runs: if the DLLs isn't there, on the command line it will produce no visible output).

This was indeed a bug, where the `IMPORTED_LOCATION` property was being overwritten. Since these targets would now only have the `IMPORTED_LOCATION_<CONFIG>`  property, it does make me wonder if there would be any side effects, however the tests to cover a wide range of cases, and if they pass I'm confident that's not the case. If there were any issues in the future, I suspect we would need to set the `IMPORTED_CONFIGURATIONS` property - I'm hesitant to do so in this PR since it's not directly related to the issue being fixed.

Once again, thanks for contributing and for fixing this! And this is a very good example of using the `TARGET_RUNTIME_DLLS` generator expression :D

jcar87 (approved):

memsharded (approved):

AbrilRBS (approved):

franramirez688 (approved):

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps.py
```diff
@@ -661,6 +661,37 @@ def package_info(self):
     assert "talk: Release!" in client.out
     assert "main: Debug!" in client.out
 
+@pytest.mark.tool("cmake", "3.21")
```
franramirez688: I think `3.21` is not installed in CI

```suggestion
@pytest.mark.tool("cmake", "3.23")
```

Untzelmann:
I think it should not have any side effects. We usually only set the _RELEASE / _DEBUG versions in our CMake files and works fine. I think the only reason why the setting of the IMPORTED_LOCATION${config_suffix} was made is, to make sure that the other values are not overwritten, if release and debug is generated for a project. However the IMPORTED_IMPLIB was not considered. It would probably also have been fine to just set the IMPORTED_IMPLIB${config_suffix} at the end. However, then the value of the variables without config suffix would have the values of debug or release at random. So I preferred the solution were we do not set these variables at all.

Thank you for your assistance. By looking at your change for the tests, I think I already understood how that works. My next contributions will contain a unit test directly, then.

PR #13857, opened 2023-05-09
Cache the CMakeDeps Jinja2 templates

### LINKED ISSUE #13845: [bug] The CMakeDeps generation is very slow
### Environment details

* Conan version: I'm using 1.59, but Conan 2.0 has the same issue.
* Python version: 3.6.10

### Steps to reproduce

1. Create a conanfile with several dependencies
2. run `conan install`
3. Observe that the `generate` step of each package takes a long time.

For example in one of our smaller projects, the install command took 23.6 seconds. And profiling using the cProfile revealed that 21.4 seconds of that was spent compiling Jinja2 templates. The actual rendering took just a fraction of a second. This slowness really hurts our iteration speed, and for bigger projects it takes much longer.

The root cause is wrong usage of Jinja2, you are supposed to create an environment with a loader, and load all files through that environment. That way Jinja2 will automatically cache the results. 

I was considering making a pull request, but then I noticed that Jinja2 is used in the same wrong way all our the place in the codebase, so there need to be some bigger architectural consideration on how to pass the environment around, or perhaps use some kind of singleton. All templates also need to be given an unique name or stored in files, so that the Jinja2 environment uniquely can identify them.

### Logs

_No response_

fredizzimo:
Changelog:  Fix: Speed up the CMakeDeps generation
Docs: omit

The compilation of the Jinja2 templates is quite slow, and several templates needs to be compiled per dependency, so for larger projects with a lot of dependencies it can be quite slow. This is fixed by caching the templates per class instance. For a medium size project with around 80 packages and dependencies I saw an improvement from 1:10 seconds to 10 seconds with this fix, so it's quite a bit faster.

NOTE: I'm not particularly happy with the implementation itself, but any alternative implementation I can think of need all the template files to be modified. So I decided to go with this non-intrusive implementation. But I can change it if you prefer.

Fixes https://github.com/conan-io/conan/issues/13845

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/__init__.py
+++ b/conan/tools/cmake/cmakedeps/templates/__init__.py
@@ -44,8 +44,15 @@ class CMakeDepsFileTemplate(object):
             raise ConanException("error generating context for '{}': {}".format(self.conanfile, e))
         if context is None:
             return
-        return Template(self.template, trim_blocks=True, lstrip_blocks=True,
-                        undefined=jinja2.StrictUndefined).render(context)
+
+        # Cache the template instance as a class attribute to greatly speed up the rendering
+        # NOTE: this assumes that self.template always returns the same string
+        template_instance = getattr(type(self), "template_instance", None)
+        if template_instance is None:
+            template_instance = Template(self.template, trim_blocks=True, lstrip_blocks=True,
+                                         undefined=jinja2.StrictUndefined)
+            setattr(type(self), "template_instance", template_instance)
+        return template_instance.render(context)
 
     def context(self):
         raise NotImplementedError()


conan/tools/cmake/cmakedeps/templates/__init__.py
```diff
@@ -44,8 +44,15 @@ def render(self):
             raise ConanException("error generating context for '{}': {}".format(self.conanfile, e))
         if context is None:
             return
-        return Template(self.template, trim_blocks=True, lstrip_blocks=True,
-                        undefined=jinja2.StrictUndefined).render(context)
+
+        # Cache the template instance as a class attribute to greatly speed up the rendering
+        # NOTE: this assumes that self.template always returns the same string
```
memsharded, line 49: Doubled checked, all ``.template`` are indeed constant strings, all good.

memsharded (approved):
Looks good, thanks!

I have realized that there is a potentially unnecessary slow ``textwrap.dedent()`` for every large string, for every file, and this PR also optimizes it, not only jinja.

memsharded:
> NOTE: I'm not particularly happy with the implementation itself, but any alternative implementation I can think of need all the template files to be modified. So I decided to go with this non-intrusive implementation. But I can change it if you prefer.

I was also initially thinking something a bit more explicit, without ``getattr(type(self)...``, but the PR is short, only in one single place instead of multiple files, so I think it has merit, and is good enough to move forward. Thanks again!

PR #13905, opened 2023-05-16
Add `timeout` argument to ConanFile `run` method

AbrilRBS:
Changelog: Feature: Add `timeout` argument to ConanFile `run` method
Docs: In progress

Adds a way for recipes to specify the timeout of their `self.run` calls
Also adds the same functionality to `tools.utils.run_command` to keep symmetry.

### DIFF
--- a/conans/model/conan_file.py
+++ b/conans/model/conan_file.py
@@ -313,7 +313,7 @@ class ConanFile:
         return Path(self.generators_folder)
 
     def run(self, command, stdout=None, cwd=None, ignore_errors=False, env="", quiet=False,
-            shell=True, scope="build"):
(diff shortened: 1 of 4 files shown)


conans/model/conan_file.py
```diff
@@ -313,7 +313,7 @@ def generators_path(self) -> Path:
         return Path(self.generators_folder)
 
     def run(self, command, stdout=None, cwd=None, ignore_errors=False, env="", quiet=False,
-            shell=True, scope="build"):
+            shell=True, scope="build", timeout=None):
```
memsharded, line 316: Beware this breaks Conan 1.X usage, if explicitly used in recipes.

PR #13917, opened 2023-05-18
DuplicateOptionError Conan v1.60.0

ninamueller13:
Dear Conan community,

I am working with Conan for the first time as part of a software development project. The software developer who provides the software I use recommends a Conan version released before v2.0.0. I installed v1.60.0 on Linux with the following command:

`sudo pip install -U "conan<2.0"`

After that I have made the configuration. 
In the next step I wanted to compile one of the samples which was also provided by the software developer. I received the following error message

`ERROR: while executing system_requirements(): System requirements: "libasound2, libasound2-dev" are missing but can´t install because tools.system.package_manager:mode is "check". Please update packages manually or set "tools.system.package_manager:mode" to "install" in the [conf] section of the profile, or in the command line using "-c tools.system.package_manager:mode=install"`

libasound2 and libasound2-dev are definitely installed and up to date.
The software manufacturer recommends adding the following lines to global.conf in case of this error

`tools.system.package_manager:mode=install`
`tools.system.package_manager:sudo=True`

Unfortunately, I cannot find a global.conf in version 1.60.0 and have not yet found any useful information on how to create one or in which path it should be stored. Therefore, I have tried to modify the global.conf accordingly

`[log]`
`run_to_output = True`
`run_to_file = False`
`level = critical`
`print_run_commands = False`

`[general]`
`default_profile = default`
`compression_level = 9`
`sysrequires_sudo = True`
`request_timeout = 60`
`default_package_id_mode = semver_direct_mode`
`revision_enabled = 1`

`[storage]`
`path = ./data`

`[proxies]`

`[hooks]`
`attribute_checker`

`[tools]`
`system.package_manager:mode = install`
`system.package_manager:sudo = True`

Unfortunately, when trying to compile the sample with this conan.conf, I get this error message

`raise DuplicateOptionError(sectname, optname, configparser.DuplicateOptionError: While reading from "/home/miti/.conan/conan.conf" [line 25]: option "system.package_manager" in section "tools" already exists`

I also tried to solve the problem with `unset`, but that didn't work either.
Unfortunately I don't know how to fix this problem, so I would be very grateful for tips on how to create the global.conf or how to adjust the conan.conf so that the compilation of the sample works. 

Thanks a lot!

n1n14

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -543,11 +543,10 @@ def collect_libs(conanfile, folder=None):
     return result
 
 
-# TODO: Do NOT document this yet. It is unclear the interface, maybe should be split
-def move_folder_contents(src_folder, dst_folder):
-    """ replaces the current folder contents with the contents of one child folder. This
-    is used in the SCM monorepo flow, when it is necessary to use one subproject subfolder
-    to replace the whole cloned git repo
+def move_folder_contents(conanfile, src_folder, dst_folder):
+    """ replaces the dst_folder contents with the contents of the src_folder, which can be a
+    child folder of dst_folder. This is used in the SCM monorepo flow, when it is necessary
+    to use one subproject subfolder to replace the whole cloned git repo
     /base-folder                       /base-folder
         /pkg  (src folder)                 /other/<otherfiles>
           /other/<otherfiles>              /pkg/<pkgfiles>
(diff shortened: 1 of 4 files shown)


memsharded:
Hi @n1n14 

PRs among our own branches are not allowed.
Only contributions from your own repository fork are allowed.
Also it is better to open an issue first, to discuss ideas and propose changes.

memsharded:
Regarding the issue described, even if the ``global.conf`` file doesn't exist in Conan 1.60, it works by creating an empty ``global.conf`` in the root folder of the Conan cache, and adding there the configurations (without ``[conf]`` header, that is only used in profile files). Please try that, and if any further question, create an issue for it. Thanks!

PR #13918, opened 2023-05-18
Metadata files upload/download

### LINKED ISSUE #4610: Embed test_package in the recipe
Provide a way to avoid having a `test_package` folder. Maybe declaring the contents of a `main.c/cpp` in the recipe could be enough.  It would be nice because:

- You could contribute with only a conanfile.py to a library, without polluting the repository.
- The `conan install --build` could run the test_package too.

To brainstorm:

- New conanfile method `test_package`? calling some helpers that will generate on the flight the test package process? Or should it be only a standard variable name containing the c/c++ code?
- Should this mechanism will be compatible or support having the classic `test_package` folder?
- Should accept a path to a file in case the repository already have an example?

### LINKED ISSUE #13715: [question] Test reports
### What is your question?

Dear Conan Folks,

I am a bit divided about the topic where shall test reports for a package belong to.
In this case for every package I have number of tests: Unit tests, static code analysis, functional tests, simulation tests, on device tests. Each of them will generate a report. I think that this shall be shipped along with the binaries or libraries of the package so that the binary and the reports are all in the same unit and can be referenced by the consumer of the package. But what I am concerened about is :
- The size of the package will increase.
Other alternatives that I thought about is:
- Having a generic repo for reports but then that will come with a disadvantage of linking the reports belonging to a package.
-  Having another conan package for storing reports again there also we need to link the reports  belonging to a package.
Could you please help me what is the suggestion from the experts?

Thanks!
Br,
Maitrey

memsharded:
Changelog: Feature: New metadata files associated to recipes and packages that can be uploaded, downloaded and added after the package exists.
Docs: Omit

This proposal implements upload + download flows, but not ``conan install`` or ``conan create`` installation of metadata. That would add too much extra complexity to the code and uncertainty and fragile developer flows, like metadata not being pulled from server because the package is already in the cache, despite users trying to ``conan install --metadata=*``, thinking that it will pull new metadata from servers.

So probably the best flow to support full-graph download of metadata would be:
- ``conan install ... --format=json > graph.json``
- convert ``graph.json`` => ``pkg_list.json``
- ``conan download --list=pkg_list.json --metadata=logs*``

This flow has been already implemented in https://github.com/conan-io/conan/pull/13928

Close https://github.com/conan-io/conan/issues/13715
Close https://github.com/conan-io/conan/issues/4610


AbrilRBS (approved):
Seems like we'll need to implement the package list inputs sooner rather than later then :)

conan/internal/upload_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+            relpath = os.path.relpath(abs_path, folder)
+            if metadata:
+                if not any(fnmatch.fnmatch(relpath, m) for m in metadata):
+                    continue
+            path = os.path.join("metadata", relpath).replace("\\", "/")
+            result[path] = abs_path
+    return result
+
+
+def gather_metadata(upload_data, cache, metadata):
```
franramirez688: I think it's clearer to replace `upload_data` by `pkg_list`

conan/internal/upload_metadata.py
```diff
@@ -0,0 +1,38 @@
...
+                recipe_bundle["upload"] = True
+
+        for pref, pkg_bundle in upload_data.prefs(rref, recipe_bundle):
+            if metadata or pkg_bundle["upload"]:
+                metadata_folder = cache.pkg_layout(pref).metadata()
+                files = _metadata_files(metadata_folder, metadata)
+                if files:
+                    ConanOutput(scope=str(pref)).info(f"Package metadata: {len(files)} files")
+                    pkg_bundle.setdefault("files", {}).update(files)
+                    pkg_bundle["upload"] = True
```
franramirez688, line 38: This module is called `upload_metadata` but I think it's confusing. What about only `metadata`?

conans/test/integration/metadata/test_metadata_commands.py
```diff
@@ -0,0 +1,59 @@
...
+        # Add new files to the metadata
+        myfile = os.path.join(metadata_path, "logs", "mylogs2.txt")
+        save(myfile, "mylogs2!!!!")
+        myfile = os.path.join(pkg_metadata_path, "logs", "mybuildlogs2.txt")
+        save(myfile, "mybuildlogs2!!!!")
+        # Upload the metadata, even if the revisions exist in the server
+        # adding the new metadata logs files
+        c.run("upload * -c -r=default --metadata=*")
+        assert "pkg/0.1: Recipe metadata: 2 files" in c.out
+        assert "pkg/0.1:da39a3ee5e6b4b0d3255bfef95601890afd80709: Package metadata: 2 files" in c.out
```
franramirez688, line 41: Then, have we replaced the existing uploaded metadata files?

conan/internal/upload_metadata.py
memsharded: This is for the "upload" functionality exclusively at this point. When we structure the "internal" I think this should belong to some "upload" folder, not to a general "metadata" folder. But we will talk about it when we re-structure the project layout.

conan/internal/upload_metadata.py
memsharded: Done

conans/test/integration/metadata/test_metadata_commands.py
memsharded: It has added 1 file, the other didn't change, so it hasn't been replaced. I can add a replace test.

memsharded: Test added

franramirez688 (approved):

czoido (approved):

PR #13926, opened 2023-05-20
Bugfix/premakedeps - Double absolute paths

Ohjurot:
OK... as it seems the only thing that I have keeps as is from the old premake5 generator is now falling on my feet...

Changelog: Bugfix: Fix "double" absolute paths in premakedeps.
Docs: Omit

Conan 2.0.5 seems to already factor in the package path when calling `aggregated_components()`
The premake5 implementation manually did this. This resulted in a "double" absolute path.

### DIFF
--- a/conan/tools/premake/premakedeps.py
+++ b/conan/tools/premake/premakedeps.py
@@ -83,11 +83,11 @@ end
 
 # Helper class that expands cpp_info meta information in lua readable string sequences
 class _PremakeTemplate(object):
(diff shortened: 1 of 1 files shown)


memsharded (approved):

memsharded:
Thanks for your contribution.

It is clear that there are not enough tests covering this, ideally this should have been captured by tests. And unfortunately it seems there is not enough demand for Premake to push this higher into our tons of things to prioritize. So thanks for keep pushing it.

PR #13928, opened 2023-05-21
Feature/pkg list

### LINKED ISSUE #6647: [feature] Ability to specify a profile to match during `conan upload`
You can currently use `--query QUERY` during a `conan upload` to only upload packages matching the specified settings.  This is quite useful, but it would be nice to expand on this functionality to allow you to specify a profile to match, rather than having to write out a query.

We have run into a couple of cases where it would be helpful to be able to structure package creation and upload as
```
conan create <package ref> -pr <profile A>
conan upload <package ref> -pr <profile A>
```

While this is possible by instead specifying the package ID in the upload, this is not so straightforward in an automated build.  This is also possible with the existing `--query` flag but this is not so nice to use when needing to write out every setting in a given profile.

### LINKED ISSUE #6925: [feature] Option for uploading multiple specific references

I understand there is some work in progress (https://github.com/conan-io/conan/issues/5743) that might fulfill this use-case I'm gonna explain, but I just want to be sure it will really be possible when that issue is implemented, or maybe I can go other way to achieve what I want. Anyway, I'd be happy to have your input on this.

### Summary

I need the possibility to be more specific about the references I want to upload, i.e. I'd like to be able to do something like:
```bash
conan upload -r my_remote "boost/1.1.1@auth/ch" "zlib/1.2.3@auth/ch" "libressl/1.2.3@auth/ch" --all --parallel
```
Which will end up only uploading (compressing etc.) the packages of those references. Is there any way to achieve this atm?

### Motivation

In our company's ecosystem we applied an on-demand approach for building the conan packages for different environments (archs/flags/build_types etc.), instead of being required to build in advance packages for all our environments settings (which are many). We only upload to the remote conan server the recipe and a package with default settings, so the consumers will be able to build the desired package by their own.
Thus, when running the pipeline of a consumer app, we do the conan install with `--build missing`, building the dependent packages with the required settings. In order to avoid building always those packages, after the install command, we upload the packages we have in local cache.

Up until now, we used to always do `conan upload '*'` after `conan install --build missing` but we really want to get rid of this since we found out the upload command would take a lot of time (because it is compressing large deps like boost) even if no upload would need to be made (no package build in the `install` phase).
So we'd like to only upload the packages we built (we'll parse the json generated by the install command for this) and also upload them in parallel (so running multiple conan upload commands is not of so much help).

### LINKED ISSUE #10436: [question] How to upload a binary package that was just created in CI?
Hi,

On my CI server I am creating a package by calling ```conan create```. After that I want to upload the binary package that was just created. The poblem I face now is that I do not know how I can automate uploading exactly the package that was created with the previous create call. I think I could use the full reference like ```conan upload MyLib/1.0.0@jenkins/stable:<binary package hash> -r myremote``` but I don't know how I can get the `<binary package hash>` part via the command line. How is this usually done?

Thanks for your time!

### LINKED ISSUE #13274: [question] Conan 2.X and CI
### What is your question?

Hello everyone,

I'm in the process of upgrading my projects in order to use conan V2,

I was heavily using conan CPT in my CI, and I'm now updating my workflow as CPT is deprecated / not part of V2 (as far as I understand),

In CPT I was using the `CONAN_UPLOAD_DEPENDENCIES` environement variable, in order to push to my personal repository  (jfrog) all dependencies that were built during the build, extremly usefull to save time for later builds,

How could I reproduce this behavious with conan ?

My first idea was to analyse the log output in order to identify the dependencies that were build, however this sounds complicated (at least for me), and I'm not sure about how to proceed. Maybe using the `--format=json` of the `conan create` comamnd ? Unless there is an other / easier way ?

Thanks!

- [x] I've read the CONTRIBUTING guide

memsharded:
Changelog: Feature: Use ``PackagesList`` as input argument for ``conan upload --list=pkglist.json``.
Changelog: Feature: Use ``--graph`` input for ``conan list`` to create a ``PackagesList`` that can be used as input for ``conan upload``.
Docs: https://github.com/conan-io/docs/pull/3257

Close https://github.com/conan-io/conan/issues/13274
Close https://github.com/conan-io/conan/issues/10436
Close https://github.com/conan-io/conan/issues/6925
Close https://github.com/conan-io/conan/issues/6647


conans/test/integration/command_v2/test_combined_pkglist_flows.py
```diff
@@ -0,0 +1,139 @@
...
+        client.run(f"download {pattern} -r=default --format=json", redirect_stdout="pkglist.json")
+        # TODO: Discuss "origin"
+        assert "Local Cache" in client.load("pkglist.json")
+        # Download binary too! Just to make sure it is in the cache, but not uploaded
+        # because it is not in the orignal list of only recipes
+        client.run(f"download * -r=default")
+        client.run("remove * -r=default -c")
+        client.run("upload --list=pkglist.json -r=default")
+        assert f"Uploading recipe 'zlib/1.0.0" in client.out
+        assert f"Uploading recipe 'zli/" not in client.out
```
AbrilRBS, line 162: This neatly shows the usecase, I think it was a good call to have both download & upload available at first :)

AbrilRBS (approved):
Nice! Looks good to get merged now, and iterate on it further down the line

czoido (approved):

conan/cli/commands/download.py
```diff
@@ -17,27 +21,45 @@ def download(conan_api: ConanAPI, parser, *args):
...
+                             "e.g: zlib/1.2.13:* means all binaries for zlib/1.2.13. "
+                             "If revision is not specified, it is assumed latest one.")
     parser.add_argument("--only-recipe", action='store_true', default=False,
                         help='Download only the recipe/s, not the binary packages.')
     parser.add_argument('-p', '--package-query', default=None, action=OnceArgument,
                         help="Only download packages matching a specific query. e.g: os=Windows AND "
                              "(arch=x86 OR compiler=gcc)")
     parser.add_argument("-r", "--remote", action=OnceArgument, required=True,
                         help='Download from this specific remote')
+    parser.add_argument("-l", "--list", help="Package list file")
```
franramirez688, line 35: Perhaps the `--graph-list` name fits better? I thought of this name like the param from `conan list` is  --graph`. Even a better help message could be enough... Wdyt?

conan/cli/commands/download.py
```diff
@@ -17,27 +21,45 @@ def download(conan_api: ConanAPI, parser, *args):
...
+
     remote = conan_api.remotes.get(args.remote)
+
+    if args.list:
+        listfile = make_abs_path(args.list)
+        multi_package_list = MultiPackagesList.load(listfile)
+        try:
+            package_list = multi_package_list[remote.name]
+        except KeyError:
+            raise ConanException(f"The current package list do not contain remote '{remote.name}'")
```
franramirez688:
```suggestion
            raise ConanException(f"The current package list does not contain remote '{remote.name}'")
```

conan/api/model.py
```diff
@@ -33,6 +35,70 @@ def __repr__(self):
...
+        pkglist = MultiPackagesList()
+        cache_list = PackagesList()
+        if graph_recipes is None and graph_binaries is None:
+            recipes = ["*"]
+            binaries = ["*"]
+        else:
+            recipes = [r.lower() for r in graph_recipes or []]
+            binaries = [b.lower() for b in graph_binaries or []]
+
+        pkglist.lists["Local Cache"] = cache_list
```
franramirez688, line 86: Then, the `graphfile` param and the function itself are only supposed to fill the `Local Cache` field, aren't they? Should `load_graph` mention anything in a docstring to clarify it? If `load` are adding `remotes` and this is not, I think both docstrings could be useful.

franramirez688 (approved):

conan/cli/commands/download.py
memsharded: It is not a graph list, it is a package list, the result of the ``conan list`` command.

conan/api/model.py
memsharded: This is a good point.

Indeed, this is hardcoding the assumption that ``load_graph`` at the moment only happens for packages that have been installed in the cache, and then exists there for sure. 

We need to check if there is some scenario with ``graph info`` that do not retrieve the binaries to the local cache if this makes sense or not. I'll have a look.

conan/cli/commands/download.py
franramirez688: Sure! That was I meant. I think it could be a bit confusing to say only `
Package list file` instead of:

```suggestion
    parser.add_argument("-l", "--list", help="Package list file. See more information in [Docs_url]/commands/list#format-json")
```
Or something like that. At least, pointing out some kind of documentation about that file.

memsharded: Too much in my opinion. Why this and not others? Following this, we should be adding links to the docs to many of the arguments of many commands, and note that links can read badly in the CLI help output.

franramirez688 (approved):
Great! 👍

franramirez688 (approved):
Great! 👍

franramirez688 (approved):
Great! 👍

jcar87 (approved):

PR #13930, opened 2023-05-22
New ``local-recipes-index`` feature

memsharded:
Changelog: Omit
Docs: Omit

Not documenting the feature until the blog post that announces it
- Feature: Allow adding a folder with a clone of conan-center-index or same folder structure as a remote.

How to use it:

```bash
$ git clone git@github.com:conan-io/conan-center-index
$ conan remote add c3ifork file:///path/to/cloned/conan-center-index
// use it as normal
$ conan search opencv*
$ conan graph info --requires=opencv/4.5.5 -r=c3ifork --format=html > file.html
```


conans/client/remote_manager_git.py
```diff
@@ -0,0 +1,167 @@
...
+from conans.client.cmd.export import cmd_export
+from conans.errors import ConanException, PackageNotFoundException, RecipeNotFoundException
+from conans.model.recipe_ref import RecipeReference
+
+
+class GitRemoteManager:
+
+    def __init__(self, remote):
+        self._remote = remote
+        cache_folder = self._remote.url.replace("file:///", "")
```
prince-chrismc: This does not support the host name properly https://datatracker.ietf.org/doc/html/rfc8089#section-2 😱  it can also be `file:/absolute/path` or  `file://hostname/absolute/path` as well

This protocol is super buggy with different parsers using and not using the 1,2, or 3 slashes correctly in my experience 

If you drop the third `/` that would make it a relative path? unless you are requiring 4 slashs and that i've never seen before ;p

conans/client/remote_manager.py
```diff
@@ -18,14 +19,18 @@
...
     """ Will handle the remotes to get recipes, packages etc """
-
     def __init__(self, cache, auth_manager):
         self._cache = cache
         self._auth_manager = auth_manager
         self._signer = PkgSignaturesPlugin(cache)
 
+    @staticmethod
+    def _git_remote(remote):
+        if remote.url.startswith("file:///"):  # FIXME: remote type?
```
prince-chrismc: I like inventing these so that we do not need to follow any existing specifications 

```suggestion
        if remote.url.startswith("conandiskremote://"):
```

Plain and easy, even if it's not registered it's hard to see that conflicting nor are we supporting generic protocols like a browser

conans/client/remote_manager_git.py
memsharded: > This does not support the host name properly https://datatracker.ietf.org/doc/html/rfc8089#section-2 😱 it can also be file:/absolute/path or file://hostname/absolute/path as well

It is only intended to support local filesystem folders,  which means ``file:///`` is the way to specify it. It doesn't intend to support hostnames at all. In any case, this is a small implementation detail, that might be changed to any thing else, like a new Conan remote type.

conans/client/remote_manager.py
memsharded: Yes not even that, but ``conan remote add myremote <path> --is-conandisk-remote``

AbrilRBS (commented):
Right now running `conan list "*:*" -r=local` for the current head of cci fails because it finds Conan 1-only recipes and does not get past that. A try-catch in the `select()` method inner loop of `subapi/list.py` might be enough to skip those cases?

conans/client/remote_manager.py
prince-chrismc: I think that's a better approach 👍  the URL path encoding for windows and network paths are painful and since this is not doing any encoding. With my background the overloading of the protocol scheme is confusing 🙃 

Does it need a flag at that point? What about just letting python sort it out? `if path.exists() and path.is_dir(): else it's a URL` 🤔  I wouldn't be surprised if that was more feature complete too.

From UX these could both work? Is there a reason for disk to be optional?

```
conan remote add disk-remote /usr/home/prince/conan-center-index-look-a-like
conan remote add disk-remote C:\Users\Chris Mc\Documents\conan-center-index-look-a-like
conan remote add local-remote https://myserver.prince.co/api
```

memsharded:
> Right now running conan list "*:*" -r=local for the current head of cci fails because it finds Conan 1-only recipes and does not get past that. A try-catch in the select() method inner loop of subapi/list.py might be enough to skip those cases?

I have improved that by running a quick check of ``from conans`` in recipes, seems to be enough atm. I am also caching the result of the export to avoid re-exporting all the time.

danimtb (approved):

AbrilRBS:
To recap, some alternatives that might be nice to look into:
 * `conan remote add-local /home/barbarian/conan-center-index/`
 * `conan remote add /home/barbarian/conan-center-index/ --local`
 * `conan remote add /home/barbarian/conan-center-index/`, @czoido suggests that if no "real" remote can have the protocol missing, there's no overlap between `/absolute/paths` and `http(s)`

I think the UX feels better this way, and less confusion about protocols
(@czoido justo had the same issue about the number of `/` after the `file:` I experienced last week)

czoido:
This is supercool, my only comments are about changing the UX a bit, I would vote for adding an argument or not adding anything and distinguish between them with the `http:`

memsharded:
Lets add a explicit argument, just in case in the future we want to support the flow clone + use, and then we need a URL

memsharded:
Added ``conan remote add /path/to/repo --type=local`` to allow future flexibility and have full command irrespective of url input arg.

Name ``--type`` and ``"local"`` can be discussed, of course.

patmantru:
Request: rely on --type, don't decide based on file:// which would require a local clone of the repo, which isn't desirable in all use cases.

That would allow for something like:
https://github.com/mycompany/conan-recipes --type=git
file://localhost/conan-recipes --type=file

Please?

memsharded:
Thanks for the feedback @patmantru 
This is something that we want to do, but it is still a bit early. Note that one of the main advantages of using the conan-center-index structure is that users using a fork can perfectly reproduce things by checking out any desired commit. Specifying just the URL of the git remote implies using the latest HEAD main branch, but that opens many other questions and feature requests:
- Automatically updating the fork to latest. If this is a pull, it can create conflicts.
- Being able to specify branches, tags, commits in the remote definition
- Automate the push of the repo for ``conan upload`` to that repo
- Cloning to a different location, to use different clones (what would be the location in the first place?)

So we think it is better to start with the lower level (folder in disk) until the core functionality, which is using packages from there is more stable, and after that and with the learnings from many more users, we will also know better different use cases  about automating the clone and checkout from git remote.

czoido:
After testing this a bit, two comments:

- I would prefer that the recipes are not cached. For my use case, mostly for testing fixes for conflicts in Conan Center, the ideal is that every change in a recipe is taken when I do a new `conan install`

- Also, I noticed that it seems that using this branch Conan does not to export the boost recipe correctly, because that recipe exports a yaml with dependencies that does not seem to be exported to the cache:

```
$ conan install --requires=boost/1.83.0 


Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=15
os=Macos

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=15
os=Macos


boost/1.83.0: Not found in local cache, looking in remotes...
boost/1.83.0: Checking remote: c3ifork
  File "/Users/carlosz/.pyenv/versions/3.9.12/lib/python3.9/shutil.py", line 823, in move
    os.rename(src, real_dst)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.conan2/p/boost8fa1c3c0f8b8f/d/dependencies/dependencies-1.83.0.yml' -> '/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.conan2/p/boost8fa1c3c0f8b8f/e/dependencies/dependencies-1.83.0.yml'

During handling of the above exception, another exception occurred:

  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/cli/cli.py", line 272, in main
    cli.run(args)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/cli/cli.py", line 172, in run
    command.run(self._conan_api, self._commands[command_argument].parser, args[0][1:])
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/cli/command.py", line 134, in run
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/cli/commands/install.py", line 69, in install
    deps_graph = conan_api.graph.load_graph_requires(args.requires, args.tool_requires,
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/api/subapi/graph.py", line 126, in load_graph_requires
    deps_graph = self.load_graph(root_node, profile_host=profile_host,
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conan/api/subapi/graph.py", line 175, in load_graph
    deps_graph = builder.load_graph(root_node, profile_host, profile_build, lockfile)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/graph_builder.py", line 50, in load_graph
    new_node = self._expand_require(require, node, dep_graph, profile_host,
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/graph_builder.py", line 90, in _expand_require
    new_node = self._create_new_node(node, require, graph, profile_host, profile_build,
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/graph_builder.py", line 260, in _create_new_node
    resolved = self._resolve_recipe(require.ref, graph_lock)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/graph_builder.py", line 208, in _resolve_recipe
    result = self._proxy.get_recipe(ref, self._remotes, self._update, self._check_update)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/proxy.py", line 23, in get_recipe
    resolved = self._get_recipe(ref, remotes, update, check_update)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/proxy.py", line 41, in _get_recipe
    remote, new_ref = self._download_recipe(reference, remotes, output, update, check_update)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/proxy.py", line 133, in _download_recipe
    self._download(latest_rref, remote)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/graph/proxy.py", line 138, in _download
    self._remote_manager.get_recipe(ref, remote)
  File "/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.env/lib/python3.9/site-packages/conans/client/remote_manager.py", line 86, in get_recipe
    shutil.move(file_path, os.path.join(export_folder, file_name))
  File "/Users/carlosz/.pyenv/versions/3.9.12/lib/python3.9/shutil.py", line 843, in move
    copy_function(src, real_dst)
  File "/Users/carlosz/.pyenv/versions/3.9.12/lib/python3.9/shutil.py", line 444, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/Users/carlosz/.pyenv/versions/3.9.12/lib/python3.9/shutil.py", line 266, in copyfile
    with open(dst, 'wb') as fdst:
FileNotFoundError: [Errno 2] No such file or directory: '/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.conan2/p/boost8fa1c3c0f8b8f/e/dependencies/dependencies-1.83.0.yml'

ERROR: [Errno 2] No such file or directory: '/Users/carlosz/Documents/developer/conan/git-remotes-demo/conan-center-index/.conan-home/.conan2/p/boost8fa1c3c0f8b8f/e/dependencies/dependencies-1.83.0.yml'
```

conan/cli/commands/remote.py
```diff
@@ -72,9 +72,11 @@ def remote_add(conan_api, parser, subparser, *args):
                            help="Insert the remote at a specific position in the remote list")
     subparser.add_argument("-f", "--force", action='store_true',
                            help="Force the definition of the remote even if duplicated")
+    subparser.add_argument("-t", "--type", choices=["local"],
+                           help="Define the remote type")
```
AbrilRBS, line 82: My only gripe with this is that only having "local" as a choice makes it seem in the help output that only "local" is supported, not that it's an opt-in for local remotes. Do you think we could also add a "remote" (or something) default  value?

conans/client/rest_client_folder.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    Implements the RestAPI but instead of over HTTP for a remote server, using just
+    a local folder assuming the conan-center-index repo layout
+    """
+
+    def __init__(self, remote):
+        self._remote = remote
+        cache_folder = self._remote.url
+        self._remote_cache_dir = "{}/.cache".format(cache_folder)
+        hook_folder = HomePaths(self._remote_cache_dir).hooks_path
```
AbrilRBS: this is pretty clever, using the same folder as the place to place the extension in!

AbrilRBS (commented):
Minor comment about UX

conans/client/rest_client_folder.py
memsharded: Yes, it is a nested fully fledged cache inside the folder, so we can leverage all the existing "export" code, so putting the hook in the cache was a no-brainer.

conan/cli/commands/remote.py
memsharded: Agree, I haven't thought the UX or the naming at all. There are other cases we might want to consider too.

conan/internal/api/new/oss_recipe.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], destination=self.source_folder)
```
czoido:
```suggestion
        get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
```

conan/internal/api/new/oss_recipe.py
```diff
@@ -0,0 +1,100 @@
...
+    {%- else -%}
+    sha256: "Put here your tarball sha256"
+    {% endif -%}
+"""
+
+
+conanfile = """\
+from conan import ConanFile
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, load, replace_in_file, save
```
czoido:
```suggestion
from conan.tools.files import apply_conandata_patches, export_conandata_patches, get
```

conan/internal/api/new/oss_recipe.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class {{package_name}}Recipe(ConanFile):
+    name = "{{name}}"
+    version = "{{version}}"
```
czoido:
```suggestion
```

conan/api/model.py
```diff
@@ -9,14 +9,18 @@
 from conans.util.files import load
 from conans.model.version_range import VersionRange
 
+OSS_RECIPES = "oss-recipes"
+OSS_RECIPES_GIT = "oss-recipes-git"
```
czoido: One detail to check: for both the `oss-recipes` and the `oss-recipes-git` with an ssl url I get a warning about the format of the URL like `WARN: The URL 'git@github.com:conan-io/conan-center-index.git' is invalid. It must contain scheme and hostname`.

conan/internal/api/new/oss_recipe.py
```diff
@@ -0,0 +1,100 @@
...
+
+    {% if tool_requires is defined -%}
+    def build_requirements(self):
+        {% for require in tool_requires -%}
+        self.tool_requires("{{ require }}")
+        {% endfor %}
+    {%- endif %}
+"""
+
+oss_recipe_files = {"recipes/{{name}}/config.yml": config_yml,
```
czoido: What about adding a template for a _test_package_ as well, even if it's a "hello world"...? I think it would be useful when scaffolding a new recipe for Conan Center

conan/internal/api/new/oss_recipe.py
memsharded: I had to remove that, because it was failing in the test if there is no root folder in the zip

conan/api/model.py
memsharded: Yes, that's right, I need to get rid of it

conan/internal/api/new/oss_recipe.py
memsharded: The ``test_package`` is not involved at all when using the oss-recipes via ``remote add``, I wasn't sure about this

conan/internal/api/new/oss_recipe.py
czoido: Maybe we could fake it in the test with something like:

```
        save_files(repo_folder, {"somefolder/CMakeLists.txt": cmake,
                                      "somefolder/pkg.h": gen_function_h(name="pkg"),
                                      "somefolder/pkg.cpp": gen_function_cpp(name="pkg")})
```

Because I think it will fail otherwisr for most of the real world zip sources file?

memsharded: Done

conan/internal/api/new/oss_recipe.py
czoido: It's just a detail, but testing this new template I found it very easy to scaffold a new recipe for Conan Center and I missed the `test_package` there, maybe we can discuss if it makes sense.

conan/api/model.py
memsharded: Removed warning from ``conan remote add``. The warning will remain in ``conan remote update``, but that one can wait for a refactor of ``RemoteRegistry``

conan/internal/api/new/oss_recipe.py
memsharded: I can do it, it is very simple and it won't harm in most cases, while convenient in many cases

czoido (approved):

conan/api/subapi/new.py
```diff
@@ -38,7 +40,8 @@ def get_builtin_template(self, template_name):
                          "bazel_exe": bazel_exe_files,
                          "autotools_lib": autotools_lib_files,
                          "autotools_exe": autotools_exe_files,
-                         "alias": alias_file}
+                         "alias": alias_file,
+                         "local_recipes_index": local_recipes_index_files}
```
jcar87, line 44: mmm, running `conan new --help` from this branch doesn't list this new template:

```
% conan new --help
usage: conan new [-h] [-v [V]] [-cc CORE_CONF] [-d DEFINE] [-f] template

Create a new example recipe and source files from a template.

positional arguments:
  template              Template name, either a predefined built-in or a user-provided one. Available built-in templates: basic, cmake_lib, cmake_exe, meson_lib, meson_exe,
                        msbuild_lib, msbuild_exe, bazel_lib, bazel_exe, autotools_lib, autotools_exe. E.g. 'conan new cmake_lib -d name=hello -d version=0.1'. You can define
                        your own templates too by inputting an absolute path as your template, or a path relative to your conan home folder.
```

AbrilRBS: It's because it's not generated from this list https://github.com/conan-io/conan/blob/f7f99b1fb964620864f2b2e598049ef6a6488c26/conan/cli/commands/new.py#L19 :)

memsharded: I'll fix it.

tsondergaard:
Instead of requiring a local clone in the filesystem, could this allow me to specify just a git URL? For example:

```
conan remote add tsondergaard-cci https://github.com/tsondergaard/conan-center-index.git
```

With this setup can I have recipes in git and binaries in Artifactory?

memsharded:
> Instead of requiring a local clone in the filesystem, could this allow me to specify just a git URL? For example:

We have considered, even implemented it, and reverted. A couple of reasons:

- Fetching always the "head" and using it is not that great. Users want stability, being able to use a specific tag, branch or commit in many many cases. And that required to invent a whole DSL to define it: ``conan remote add repo git-url --tag=mytag`` and the like, really dirty thing
- Another very important use case, probably the most important is also being able to locally edit the recipes, and have the changes immediately available. This require a "user space" clone that users have simple access. If ``conan remote`` adds git urls, then those clones need to be in a private Conan folder, Conan can't be doing clones in user space.
- The control and semantics of updating/pulling/merging is also lost, or Conan would also need to invent yet another DSL around remotes like ``conan remote myremote --pull``

So it was just like creating a more limited, poorer layer over ``git`` repos and commands, while also not allowing users to modify sources, and it was dropped, not really worth it. Looks like convenient at first sight, but in practice it doesn't work great for majority of cases.


> With this setup can I have recipes in git and binaries in Artifactory?

Yes, Conan 2 binary management allows to use recipes from one repo (including this new local-recipes-index one) and binaries from other repos, because using revisions everywhere to guarantee consistency it allows to do so and guarantee that the binaries from one repo belongs to the same source recipe (recipe revision). Note that when binaries are uploaded to Artifactory, the recipe is also uploaded there together with the binary, so once this happens, it is no longer needed to use both and users can also get recipes and binaries from server repo too. This would be a very typical setup, and we already have many users that are doing this approach, even without the ``local-recipes-index``: they work with a fork of ``conan-center-index``, build their binaries from there and upload to their servers, and their developers and CI are just "consumers" of their own recipes+binaries. This is the approach documented in: https://docs.conan.io/2/devops/using_conancenter.html

tsondergaard:
Vcpkg suggests consuming projects add vcpkg as a git submodule. Would that pattern work with the mechanism you are building here? Let's say my company maintains a fork of conan-center-index at https://github.com/acme/conan-index.git that is extended also to contain our own libraries. Could consuming applications then add that as a git submodule and use that as the repository as the source registry and then artifactory would only be used for binary packages?

memsharded:
> Would that pattern work with the mechanism you are building here? 

Yes, this would work with this new ``local-recipes-index`` repo too, it uses a very similar approach. Nevertheless, it is possible that this approach is not the best in all scenarios, depending on the needs.

> that is extended also to contain our own libraries

The local-recipes-index is intended for third party libraries. For packaging your own libraries, the approach of putting the ``conanfile.py`` directly in the repo might work much better, specially if you need to maintain different versions of your packages, because in that way the recipe evolves in parallel to the code, and it enables simpler flows like ``git clone + conan install + cmake``, with just one single repo, no submodules. Git submodules management is not that straightforward and has issues too.

PR #13946, opened 2023-05-23
Improve graph conflict message when Conan can't show one of the conflicting recipes

AbrilRBS:
Changelog: Fix: Improve graph conflict message when Conan can't show one of the conflicting recipes.
Docs: Omit

The error now looks like:

```
ERROR: Version conflict: Conflict between math/1.0.1 and math/1.0 in the graph.
       Conflict originates from ai/1.0
       Run conan graph info ... --format=html to inspect the graph errors.

```
The graph output with conflict errors now looks like this:

image

Pinging @SSE4

### DIFF
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -3,7 +3,6 @@ import os
 
 from jinja2 import Template, select_autoescape
 
-
 from conan.api.output import cli_out_write
 from conan.cli.formatters.graph.graph_info_text import filter_graph
 from conan.cli.formatters.graph.info_graph_dot import graph_info_dot
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -111,7 +111,7 @@ def format_graph_html(result):
         error = {
             "label": graph.error.require.ref,
             "type": "conflict",
-            "from": graph.error.node.ref or graph.error.base_previous.ref,
+            "from": graph.error.node.id or graph.error.base_previous.id,
             "prev_require": graph.error.prev_require.ref
         }
     cli_out_write(_render_graph(graph, error, template, template_folder))
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -111,8 +111,7 @@ def format_graph_html(result):
             "type": "conflict",
             "from": graph.error.node.id or graph.error.base_previous.id,
             "prev_node": graph.error.prev_node,
-            "should_highlight_node": lambda node: node.id == graph.error.prev_node.id or
-                                                  node.id == graph.error.node.id
+            "should_highlight_node": lambda node: node.id == graph.error.node.id
         }
     else:
         error = {
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -106,13 +106,20 @@ def format_graph_html(result):
     template_folder = os.path.join(conan_api.cache_folder, "templates")
     user_template = os.path.join(template_folder, "graph.html")
     template = load(user_template) if os.path.isfile(user_template) else graph_info_html
-    error = None
     if isinstance(graph.error, GraphConflictError):
         error = {
             "label": graph.error.require.ref,
             "type": "conflict",
             "from": graph.error.node.id or graph.error.base_previous.id,
-            "prev_require": graph.error.prev_require.ref
+            "prev_node": graph.error.prev_node,
+            "should_highlight_node": lambda node: node.id == graph.error.prev_node.id or
+                                                  node.id == graph.error.node.id
+        }
+    else:
+        error = {
+            "type": "unknown",
+            "error:": graph.error,
+            "should_highlight_node": lambda node: False
         }
     cli_out_write(_render_graph(graph, error, template, template_folder))
     if graph.error:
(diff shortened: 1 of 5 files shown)


SSE4:
@RubenRBS can we output graph to the stdout (e.g. json format) instead of asking to run `--html`? if error happens on CI (e.g. in ConanCenterIndex) it's too hard to debug this way.

conans/client/graph/graph_error.py
```diff
@@ -15,8 +15,18 @@ def __init__(self, node, require, prev_node, prev_require, base_previous):
...
+                   f"{self.base_previous.ref}->{self.prev_require.ref}."
+        else:
+            conflicting_node = self.node.ref or self.base_previous.ref
+            conflicting_node_msg = ""
+            if conflicting_node is not None:
+                conflicting_node_msg = f"\nConflict originates from {conflicting_node}\n"
+            return f"Version conflict: " \
+                   f"Conflict between {self.require.ref} and {self.prev_require.ref} in the graph." \
+                   f"{conflicting_node_msg}" \
+                   f"\nRun conan graph info ... --format=html to inspect the graph errors."
```
memsharded: I would print this ``--format=html`` (with redirect ``> graph.html``) in all cases, even the "simple" part of the ``if``

conans/test/integration/graph/conflict_diamond_test.py
```diff
@@ -64,3 +64,7 @@ def _game_conanfile(version, reverse=False):
             c.assert_overrides({"math/1.0": [f"math/{v}"],
                                 "math/1.0.1": [f"math/{v}#8e1a7a5ce869d8c54ae3d33468fd657c"]})
             c.assert_listed_require({f"math/{v}": "Cache"})
+
+        c.run("install --requires=engine/1.0  --requires=ai/1.0", assert_error=True)
+        assert "Conflict between math/1.0.1 and math/1.0 in the graph"
+        assert "Conflict originates from ai/1.0"
```
memsharded, line 70: Check that ``--format=html`` is also suggested

memsharded (commented):
The ``--format=html`` suggestion for conflicts is not useful at the moment, because the html output is not that evident yet. I would add the html changes in this PR too.

SSE4:
to execute this suggestion on CI server (e.g. ConanCenterIndex) I would need:
- modify build script to run `--html` command (and for CCI, it's out of contributor's control)
- run one more build and wait (potentially, hours)
- somehow capture/copy produced .html file
IMO too much extra burden for users. I would just print JSON graph straight outta stderr.

memsharded:
> to execute this suggestion on CI server (e.g. ConanCenterIndex) I would need:

The idea is that not all debugging needs to happen in the CI server. Especially something that should have been detected by the developer in their machine while testing locally in the first place. It is not that a version conflict is something that only happens and can be debugged in the CI server.

So the recommended approach is not start modifying the CI, but execute a ``conan graph info`` locally, by the developer in the developer machine, not in CI. Even for other platforms and cross build different to the current machine it is completely possible to reproduce a version conflict.

conans/client/graph/graph_error.py
SSE4: can it print actual arguments here instead of `...`? so it could be copy-pasted easily.

conan/cli/formatters/graph/graph.py
```diff
@@ -104,7 +106,15 @@ def format_graph_html(result):
     template_folder = os.path.join(conan_api.cache_folder, "templates")
     user_template = os.path.join(template_folder, "graph.html")
     template = load(user_template) if os.path.isfile(user_template) else graph_info_html
-    cli_out_write(_render_graph(graph, template, template_folder))
+    error = None
+    if isinstance(graph.error, GraphConflictError):
+        error = {
+            "label": graph.error.require.ref,
+            "type": "conflict",
+            "from": graph.error.node.ref or graph.error.base_previous.ref,
```
memsharded: Not sure that providing the ``ref`` as ``from`` is the best approach. This can be problematic for graphs where there are multiple nodes with same ``ref`` which is totally possible. It seems that the ``node`` actual instance, or unique ``id`` should be used.

AbrilRBS: I did it this way becuase I was convinced that the error id would not be valid, but alas! Works nicely after testing!

conan/cli/formatters/graph/info_graph_html.py
```diff
@@ -69,12 +69,31 @@
...
-            var edges = new vis.DataSet([
+            ]
+            var edges = [
                 {%- for src, dst in graph.edges %}
                     { from: {{ src.id }}, to: {{ dst.id }} }{%- if not loop.last %},{% endif %}
                 {%- endfor %}
-            ]);
+            ]
+
+            {% if error is not none and error["type"] == "conflict" %}
```
memsharded, line 81: I would try to also highlight in big-red the other part of the conflict. In case it is a relatively large graph it will not be evident to see it. 2 big red things easy to identify would probably be the best.

AbrilRBS:
The graph now looks like this 
image

conan/cli/formatters/graph/info_graph_html.py
AbrilRBS: Done! see my comment for the new look

memsharded (approved):

AbrilRBS:
Remove red from UI, change lib link to red

memsharded (approved):
Fix broken test (checking for the wrong output message), and can be merged

czoido (approved):

PR #13964, opened 2023-05-25
Fix format for id property in graph output json

czoido:
Changelog: Fix: Output "id" property in graph json output as str instead of int.
Docs: https://github.com/conan-io/docs/pull/3236

### DIFF
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -210,7 +210,7 @@ class Node(object):
     def serialize(self):
         result = OrderedDict()
         result["ref"] = self.ref.repr_notime() if self.ref is not None else "conanfile"
-        result["id"] = getattr(self, "id")  # Must be assigned by graph.serialize()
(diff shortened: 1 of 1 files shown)


memsharded (approved):

AbrilRBS (approved):

conans/client/graph/graph.py
```diff
@@ -210,7 +210,7 @@ def __repr__(self):
     def serialize(self):
         result = OrderedDict()
         result["ref"] = self.ref.repr_notime() if self.ref is not None else "conanfile"
-        result["id"] = getattr(self, "id")  # Must be assigned by graph.serialize()
+        result["id"] = str(getattr(self, "id"))  # Must be assigned by graph.serialize()
```
franramirez688: `DepsGraph` class is injecting the `id` attribute, so I think it fits better there. Check `DepsGraph.serialize()`

franramirez688 (approved):

PR #13967, opened 2023-05-25
[graph][json] Same json output for conan graph info, create, install and export-pkg

franramirez688:
Changelog: Bugfix: Ensuring the same graph-info JSON output for  `graph info`, `create`, `export-pkg`, and `install`.
Docs: https://github.com/conan-io/docs/pull/3236

### DIFF
--- a/conan/cli/commands/create.py
+++ b/conan/cli/commands/create.py
@@ -1,24 +1,18 @@
-import json
 import os
 import shutil
 
-from conan.api.output import ConanOutput, cli_out_write
+from conan.api.output import ConanOutput
+from conan.cli.args import add_lockfile_args, add_common_install_arguments
 from conan.cli.command import conan_command, OnceArgument
 from conan.cli.commands.export import common_args_export
-from conan.cli.args import add_lockfile_args, add_common_install_arguments
+from conan.cli.formatters.graph import format_graph_json
 from conan.cli.printers import print_profiles
 from conan.cli.printers.graph import print_graph_packages, print_graph_basic
 from conan.errors import ConanException
 from conans.util.files import mkdir
 
 
-def json_create(deps_graph):
-    if deps_graph is None:
-        return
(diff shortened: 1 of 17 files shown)


conan/cli/commands/create.py
```diff
@@ -102,7 +96,8 @@ def create(conan_api, parser, *args):
                                                       clean=args.lockfile_clean)
 
     conan_api.lockfile.save_lockfile(lockfile, args.lockfile_out, cwd)
-    return deps_graph
+    return {"graph": deps_graph,
+            "conan_api": conan_api}
```
franramirez688, line 100: `conan_api` is not used for now. I kept it just for future implementations.

AbrilRBS: Just like `graph info` does right?

AbrilRBS (approved):

conan/cli/commands/create.py
memsharded: ``conan_api`` not used, it can be removed.
Actually, it is only the serialization of ``graph`` that changes, so this doesn't need to be changed.

UPDATE: Sorry, nevermind, I didn't see you are reusing the other formatter.

conan/cli/commands/install.py
```diff
@@ -1,20 +1,15 @@
...
 from conan.cli.command import conan_command
-from conan.cli import make_abs_path
+from conan.cli.formatters.graph import format_graph_json
 from conan.cli.printers import print_profiles
 from conan.cli.printers.graph import print_graph_packages, print_graph_basic
 
 
-def json_install(info):
-    deps_graph = info
-    cli_out_write(json.dumps({"graph": deps_graph.serialize()}, indent=4))
```
memsharded, line 14: It is better to leave the first level ``graph`` item, just in case?

memsharded (approved):
Perfect!

Issue #13962, opened 2023-05-25
[question] cross building packages that uses PkgConfig.fill_cpp_info


LadyzRoman:
### What is your question?

I've been trying to cross compile some packages from Windows to raspberry pi with conan 2.0.4
host profile:
```
[settings]
os=Linux
arch=armv7
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=8
[conf]
tools.build:compiler_executables={"c": "arm-linux-gnueabihf-gcc-8", "cpp": "arm-linux-gnueabihf-g++"}
tools.build:sysroot=C:\\SysGCC\\raspberry\\arm-linux-gnueabihf\\sysroot
tools.gnu:make_program=C:/SysGCC/raspberry/bin/make
[buildenv]
PKG_CONFIG_PATH=C:\SysGCC\raspberry\arm-linux-gnueabihf\sysroot\lib\arm-linux-gnueabihf\pkgconfig;C:\SysGCC\raspberry\arm-linux-gnueabihf\sysroot\usr\lib\arm-linux-gnueabihf\pkgconfig;C:\SysGCC\raspberry\arm-linux-gnueabihf\sysroot\usr\lib\pkgconfig;C:\SysGCC\raspberry\arm-linux-gnueabihf\sysroot\usr\share\pkgconfig
```
Build profile 
```
[settings]
os=Windows
arch=x86_64
build_type=Release
compiler=msvc
compiler.runtime=dynamic
compiler.cppstd=20
compiler.version=193
[conf]
tools.cmake.cmaketoolchain:generator=Ninja
```
And I having an issue with any system package, for example opengl/system, that uses PkgConfig.fill_cpp_info in package_info method. 
```
ERROR: opengl/system: Error in package_info() method, line 61
        pkg_config.fill_cpp_info(self.cpp_info, is_system=self.settings.os != "FreeBSD")
        ConanException: Command 'pkg-config --print-provides gl --print-errors' failed with errorcode '1'
b"Package gl was not found in the pkg-config search path.\r\nPerhaps you should add the directory containing `gl.pc'\r\nto the PKG_CONFIG_PATH environment variable\r\nNo package 'gl' found\r\n"
```
I can set PKG_CONFIG_PATH before calling `conan install` and everything works quite well then, but it seems not right for me. Is there a way to set this environment variable by conan?

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

[...]

PR #13985, opened 2023-05-27
PkgConfig uses ``conanfile.run()`` instead of independent runner to use Conan environment

memsharded:
Changelog: Fix: ``PkgConfig`` uses ``conanfile.run()`` instead of internal runner to get full Conan environment from profiles and dependencies.
Docs: Omit

For https://github.com/conan-io/conan/issues/13962


conan/tools/gnu/pkgconfig.py
```diff
@@ -23,11 +24,16 @@ def __init__(self, conanfile, library, pkg_config_path=None):
...
         command = cmd_args_to_string([executable, '--' + option, self._library, '--print-errors'])
+
         env = Environment()
         if self._pkg_config_path:
             env.prepend_path("PKG_CONFIG_PATH", self._pkg_config_path)
         with env.vars(self._conanfile).apply():
-            return check_output_runner(command).strip()
+            # This way we get the environment from ConanFile, from profile (default buildenv)
+            output = StringIO()
+            self._conanfile.run(command, output)
```
franramirez688: Let's use the kwarg to avoid confusion:

 ```suggestion
             self._conanfile.run(command, stdout=output)
```

File: conans/model/build_info.py [hunk 2]
```diff
@@ -112,6 +114,34 @@ def serialize(self):
...
+            "defines",
+            "cflags",
+            "cxxflags",
+            "sharedlinkflags",
+            "exelinkflags",
+            "objects",
+            "sysroot",
+            "requires",
+            "properties"
+        ]
```
franramirez688: If serialize function uses the same fields, could we put them as a class variable? wdyt?

conans/test/functional/toolchains/gnu/test_pkg_config.py
```diff
@@ -68,3 +70,67 @@ def generate(self):
         assert "conanfile.py: PROVIDES: libastral = 6.6.6" in c.out
         assert "conanfile.py: VERSION: 6.6.6" in c.out
         assert "conanfile.py: VARIABLES: /usr/local" in c.out
+
+
+def test_pkg_config_round_tripe_cpp_info():
```
franramirez688, line 75: Should we have dedicated tests for `CppInfo` new functions? or is it enough with this one?

franramirez688 (approved):

czoido (approved):

PR #13986, opened 2023-05-27
remove undocumented code, not planned atm

memsharded:
Changelog: Omit
Docs: Omit

This was never documented in 1.X, not planned to be released at the moment in 2.0, better remove it


AbrilRBS (approved):
Will ask tomorrow what the initial plan for this was :)

franramirez688 (approved):

memsharded:
The reason is connected to https://github.com/conan-io/conan/pull/13985

We have now much better serialization of CppInfo, that allows us to serialize and save it in ``package()``, load it in ``package_info()``. So the extra ``CppPackage`` would be boilerplate. Let's design the interface correctly ,and then the different ways to fill the interface with data, we can build on top of it.

PR #13989, opened 2023-05-29
print something during long downloads

memsharded:
Changelog: Feature: Small UX improvement to print some info while downloading large files.
Docs: Omit

### DIFF
--- a/conans/client/downloaders/file_downloader.py
+++ b/conans/client/downloaders/file_downloader.py
@@ -103,7 +103,7 @@ class FileDownloader:
 
         try:
             total_length = get_total_length()
-            is_large_file = total_length > 10000000  # 10 Mb
+            is_large_file = total_length > 10000000  # 10 MB
             t_start = time.time()
             base_name = os.path.basename(file_path)
             if is_large_file:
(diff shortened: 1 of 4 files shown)


conans/client/remote_manager.py
```diff
@@ -220,9 +219,9 @@ def _call_remote(self, remote, method, *args, **kwargs):
 def uncompress_file(src_path, dest_folder, scope=None):
     try:
         filesize = os.path.getsize(src_path)
-        big_file = filesize > 1000000  # 10 Mb
+        big_file = filesize > 10000000  # 10 Mb
```
AbrilRBS: Oops nice catch! This is why I prefer doing `10 * (1 << 20)` for things like this.

Also, got confused with the comment:
```suggestion
        big_file = filesize > 10000000  # 10 MB
```

conans/client/downloaders/file_downloader.py
```diff
@@ -102,12 +103,13 @@ def get_total_length():
 
         try:
             total_length = get_total_length()
-            if total_length > 1000000:  # 10 Mb
-                from conan.tools.files.files import _human_size
-                hs = _human_size(total_length)
+            is_large_file = total_length > 10000000  # 10 Mb
```
AbrilRBS: Seems like this was also missing a 0, oops!

Same thing, I know it's extreme nitpick but:

```suggestion
            is_large_file = total_length > 10000000  # 10 MB
```

AbrilRBS (approved):
Nice! Some extreme nitpicks but nothing that blocks merge if needed :)

czoido (approved):

PR #13994, opened 2023-05-29
global cpp_info can be used to initialize components

memsharded:
Changelog: Feature: Global ``cpp_info`` can be used to initialize components values.
Docs: omit

This is a long standing issue, where users that define ``self.cpp_info.includedirs = ["myinclude"]`` are surprised that individual components will not use the ``myinclude`` folder too.

There could be different approaches to this, I have implemented the following:

- Components are initialized at first instantiation from the values of the global ``cpp_info``
- That means that later changes to ``self.cpp_info`` are not taken into account.

### DIFF
--- a/conans/test/functional/layout/test_editable_cmake_components.py
+++ b/conans/test/functional/layout/test_editable_cmake_components.py
@@ -32,6 +32,7 @@ def test_editable_cmake_components():
 
             def layout(self):
                 cmake_layout(self, src_folder="src")
+                self.cpp.source.includedirs = []
                 self.cpp.source.components["hello"].includedirs = ["."]
                 self.cpp.source.components["bye"].includedirs = ["."]
                 bt = "." if self.settings.os != "Windows" else str(self.settings.build_type)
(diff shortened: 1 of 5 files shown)


conans/test/integration/command/create_test.py
```diff
@@ -572,6 +572,8 @@ def package_info(self):
                 self.cpp_info.set_property("pkg_config_name", "pkg_other_name")
                 self.cpp_info.set_property("pkg_config_aliases", ["pkg_alias1", "pkg_alias2"])
                 self.cpp_info.components["cmp1"].libs = ["libcmp1"]
+                self.cpp_info.components["cmp1"].libdirs = ["lib"]
```
memsharded, line 575: These tests were ill-formed in the first place, because they were defining things that doesn't work. Yes, they are serialized, but most generators just discard the global cpp_info definition.

conans/test/functional/layout/test_editable_cmake_components.py
```diff
@@ -32,6 +32,7 @@ def build(self):
 
             def layout(self):
                 cmake_layout(self, src_folder="src")
+                self.cpp.source.includedirs = []
```
memsharded, line 35: It is important to invalidate the non-existing ``includedirs`` in ``conanfile.cpp.source.includedirs = ["include"]`` inside ``cmake_layout()``

memsharded:
This is too risky, better wait until ``cpp_info`` is changed (maybe for CPS work)

Todiq:
Hello @memsharded,

Is there a chance to have this feature in a next conan release? It would be especially usefully to prevent looping through all the components to set the same `cpp_info` as the global one. Many thanks!

memsharded:
Hi @Todiq 

I am afraid this was discarded because of the risk, so it is not planned at this moment.
This might only be changed when the full interface is changed, that is, when fully relying on CPS information. But not possible to do it based on the global ``cpp_info`` information for every component, too high risk of break things.

PR #14002, opened 2023-05-30
upload --dry-run and tests fixes

memsharded:
Changelog: Feature: Add ``conan upload --dry-run`` equivalent to 1.X ``conan upload --skip-upload``.
Docs: https://github.com/conan-io/docs/pull/3274

Also, recover or remove some xfailed tests (the PR started there, then realized that the ``--dry-run`` feature was not there, and I recall it being used by some users in 1.X

### DIFF
--- a/conan/api/subapi/graph.py
+++ b/conan/api/subapi/graph.py
@@ -1,5 +1,3 @@
-import os
-
 from conan.api.output import ConanOutput
 from conan.internal.conan_app import ConanApp
 from conans.client.graph.graph import Node, RECIPE_CONSUMER, CONTEXT_HOST, RECIPE_VIRTUAL, \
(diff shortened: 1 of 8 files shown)


AbrilRBS (approved):
Is there any other command where this functionallity might be missing?

memsharded:
> Is there any other command where this functionallity might be missing?

This was the only one in Conan 1.X (btw, I am open to keep the ``--skip-upload`` 1.X syntax).
One use case was doing all the heavy compressions and integrity checks in one pass, then uploading all, so the actual upload was faster (compression can take a lot of time for large packages), improving a bit the atomicity of the upload. Also, the idea is that this will also support the upload via Artifactory API, but we can discuss this, if premature I can remove it, it was not the intent of this PR, but to fix tests.

I am not sure other commands do need it. For download/remove we already have the ``package-lists`` feature, that lists you the things that will be removed/downloaded.

AbrilRBS:
Thanks for the extra context, quite helpful :)

> btw, I am open to keep the --skip-upload 1.X syntax

I like the `--dry-run` syntax better

franramirez688 (approved):

czoido (approved):

PR #14051, opened 2023-06-08
Add prefix to pkgconfig root

### LINKED ISSUE #14049: [feature] add prefix var to pkg-config alias
### What is your suggestion?

When a package contains components, conan pkg-config generator generates one .pc file per component, plus a root .pc file that includes all the components .pc files.

However, the generated root .pc file does not contain the `prefix` variable. When I use the meson build system, I may need to access that variable, since this is the easiest way to get the location of the installed conan package. For instance, I may need this path to find a tool installed along with this pacakge. An other example is the `boost` package. meson expects the `boost.pc` package to have the `prefix` variable, and not having it causes issues (https://github.com/mesonbuild/meson/pull/11487 was an attempt to fix this on the meson side, but the real problem is on the conan generated file, imho).

Another problem with the generated root .pc file is that it doesn't honnor the `pkg_config_custom_content` property. Therefore, it is not possible to inject custom contents in the generated root .pc file.

bruchar1:
Changelog: Bugfix: Add prefix var and any custom content (through the `pkg_config_custom_content` property) to already generated pkg-config root .pc files by `PkgConfigDeps`.
Docs: omit

Fixes #14049

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -149,6 +149,10 @@ class _PCContentGenerator:
 
     shortened_template = textwrap.dedent("""\
         prefix={{ prefix_path }}
+        {% if pkg_config_custom_content %}
+        # Custom PC content
+        {{ pkg_config_custom_content }}
+        {% endif %}
 
         Name: {{ name }}
         Description: {{ description }}
@@ -201,8 +205,14 @@ class _PCContentGenerator:
             else self._dep.package_folder
         prefix_path = root_folder.replace("\\", "/")
 
+        if info.cpp_info:
+            custom_content = info.cpp_info.get_property("pkg_config_custom_content")
+        else:
+            custom_content = None
+
         context = {
             "prefix_path": prefix_path,
+            "pkg_config_custom_content": custom_content,
             "name": info.name,
             "description": info.description,
             "version": self._dep.ref.version,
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -346,8 +363,8 @@ def _update_pc_files(info):
         # Issue related: https://github.com/conan-io/conan/issues/10341
         pkg_name = _get_package_name(self._dep, self._build_context_suffix)
         if f"{pkg_name}.pc" not in pc_files:
-            package_info = _PCInfo(pkg_name, pkg_requires, f"Conan package: {pkg_name}", None,
-                                   _get_package_aliases(self._dep))
+            package_info = _PCInfo(pkg_name, pkg_requires, f"Conan package: {pkg_name}",
```
memsharded, line 360: This change not really necessary, the ``shortened_content`` still do not require the ``cpp_info``

memsharded (approved):
Looks good, thanks!
Assigning @franramirez688 for final review

conan/tools/gnu/pkgconfigdeps.py
bruchar1: it needs it for `custom_content`...

memsharded: I see, yes, you are right I got confused looking at my local copy of the code.

franramirez688:
Thanks for the contribution @bruchar1

I'm wondering why this is needed. Which is the use case? Was it actually a bug?

bruchar1:
> I'm wondering why this is needed. Which is the use case? Was it actually a bug?

This is detailed in the linked issue.

The fact the prefix variable is not in the .pc file is not really a bug, but having it solves issues with meson integration, especially because meson needs this variable to know where the package is installed.

The fact `pkg_custom_content` does not work with the root .pc file from a package with components could be considered as a bug.

franramirez688:
> This is detailed in the linked issue.

Oh, so sorry, my fault, I did not realize about the issue. Let me have a look at it. Perhaps, it makes sense to add a dedicated test for this change.

conans/test/integration/toolchains/gnu/test_pkgconfigdeps.py
```diff
@@ -431,11 +441,25 @@ def package_info(self):
...
 
     pc_content = client.load("pkg_other_name.pc")
-    assert "Description: Conan package: pkg_other_name" in pc_content
-    assert "Requires: compo1" in pc_content
+    content = textwrap.dedent(f"""\
+    {prefix}
+    # Custom PC content
+
+    datadir=${{prefix}}/share
+    schemasdir=${{datadir}}/mylib/schemas
```
franramirez688, line 449: Added this extra check to ensure that the custom content is being added only to the root .pc file.

franramirez688 (approved):

PR #14052, opened 2023-06-08
remove unused (only in testing) get_timestamps

memsharded:
Changelog: Omit
Docs: Omit

Removing dead code (only used in testing), because refs already contain timestamp, no need for a separate query


conan/internal/cache/db/recipes_table.py
```diff
@@ -33,16 +33,6 @@ def _where_clause(self, ref):
             [f'{k}="{v}" ' if v is not None else f'{k} IS NULL' for k, v in where_dict.items()])
         return where_expr
 
-    def _set_clause(self, ref: RecipeReference, path=None):
```
memsharded, line 36: dead code, not used or referenced anywhere

czoido (approved):

PR #14054, opened 2023-06-08
Cache LRU removal

### LINKED ISSUE #3587: Automatic cache cleanup with customizable retention period
Over time, the conan cache grows accumulating stale dependencies. There should be an automatic cleanup mechanism, which automatically removes packages in a customizable way based on some criteria:
* creation/installation time
* time of last use
* channel

For example, we are using testing and stable channels. Packages from testing channels are rebuilt at least daily, so packages older than a day will not normally be used again. Packages from stable channels might be used over months, but could also be removed when they have not been used for some days.

### LINKED ISSUE #5453: Implement something like git gc to clean out older versions of dependencies.

### LINKED ISSUE #7727: [feature] Track artifacts in Conan cache, expire least recently used
One of the challenges we see is maintaining free space on CI nodes, as well as development machines, with Conan cache on them. There is, of course, an atomic option to blow it all away, but that means the next few (or not so few) builds will be slow, retrieving the dependencies from the remote.
Something that would help is tracking access to the artifacts, and ability to remove least recently used ones. A few strategies I could see are:
- shrink the cache to a maximum size specified, deleting LRU items
- delete all items that hasn't been accessed in X days
- same as above, but with a search constraint

### LINKED ISSUE #10667: [question] Is there a way to control the size of Conan cache?
Hi,
we bumped into the situation on our CI system that Conan's cache practically grows endlessly over time.
Is there a plan / way to "garbage collect" - via Conan itself (e.g. a command line extension to conan, or a separate python script packages into the Conan distribution, I don't know) - the cache? I mean like erase the oldest N entries to make sure that the cache's size doesn't go above X Gb?
Thanks,

### LINKED ISSUE #11990: [question] Remove package revisions that have not been used for a certain period of time

My ~/.conan/data directory comtinues to grow, and some packages or some revisions of the packages are no longer required.
Is it possible for me to set a expire ttl for cached package revisions? When ttl is  reached, conan removes them automatically.

memsharded:
Changelog: Feature: Implement cache LRU control for cleaning of unused artifacts.
Docs: https://github.com/conan-io/docs/pull/3455

AFTER https://github.com/conan-io/conan/pull/14398

Close https://github.com/conan-io/conan/issues/3587
Close https://github.com/conan-io/conan/issues/7727
Close https://github.com/conan-io/conan/issues/10667
Close https://github.com/conan-io/conan/issues/5453
Close https://github.com/conan-io/conan/issues/11990


AbrilRBS (approved):
You mentioned table migrations at some point. Can I get a link to the code that does that for now columns? Otherwise looks good as far as I can see, which admitedly isnt much in this case :)

memsharded:
> You mentioned table migrations at some point. Can I get a link to the code that does that for now columns? Otherwise looks good as far as I can see, which admitedly isnt much in this case :)

I think this hasn't been done yet, so we don't have that code now, this has to be implemented as a new thing.

conan/cli/commands/list.py
```diff
@@ -108,15 +108,21 @@ def list(conan_api: ConanAPI, parser, *args):
     parser.add_argument("-g", "--graph", help="Graph json file")
     parser.add_argument("-gb", "--graph-binaries", help="Which binaries are listed", action="append")
     parser.add_argument("-gr", "--graph-recipes", help="Which recipes are listed", action="append")
+    parser.add_argument('--lru', default=None, action=OnceArgument,
+                        help="List only elements that have been recently used")
```
czoido: I think this description should be extended, it gives no clue about what type of argument you need to add to `--lru`

conans/util/dates.py
```diff
@@ -38,15 +37,25 @@ def timestamp_to_str(timestamp):
...
-        raise ConanException("Incorrect time interval definition: %s" % interval)
+        time_value = int(time_value)
+    except TypeError:
+        raise ConanException(f"Time value '{time_value}' must be an integer")
+    time_units = expression[-1]
+    units = {"w": 7*24*60*60,
+             "d": 24*60*60,
+             "h": 60*60,
+             "m": 60,
+             "s": 1}
```
czoido: Don't we want to accept years here as well?

conans/client/migrations.py
```diff
@@ -50,3 +52,24 @@ def _apply_migrations(self, old_version):
         # Update profile plugin
         from conans.client.profile_loader import migrate_profile_plugin
         migrate_profile_plugin(cache)
+
+        if old_version and old_version < "2.0.8":
+            _migrate_pkg_db_lru(cache)
+
+
+def _migrate_pkg_db_lru(cache):
```
czoido: Maybe we should consider the case of one user that install conan 2.0.8, the db is migrated but for some reason that user has to downgrade and then the Conan cache becomes unusable?

conans/util/dates.py
memsharded: Too much? I don't think users will be wanting to remove artifacts only those artifacts that are +1 year old, running with the same cache for more than 1 year without cleaning sounds highly improbable. They can always do ``--lru=365d`` too

conans/client/migrations.py
memsharded: Yeah, this has always been a problem with migrations. Migrations don't work backward, the only possibility would be to create a multi-release migration path, introducing some back-migration in 2.0.8, and then waiting for 2.0.9 to introduce the feature and the forward-migration. Still will fail if users downgrade to 2.0.7.

I am not sure if it is worth the effort, but we can discuss it.

conans/util/dates.py
czoido: Ok, I commented that because my first use of the feature was doing a `--lru=1y` and I was surprised it did not exist.

conans/client/migrations.py
czoido: Yes, I agree. Maybe it's not worth the effort, what worries me about this one is that it may make the whole cache unusable because of a feature you don't event use. Maybe we should think about this, if we want to make the databases retro compatible at least. In this case it would be that if we add a column to the database, previous versions can work with the database that has that extra column but wont use the feature because they don't know about that. Maybe that's something to consider in case we add another feature that modifies the db.

conans/util/dates.py
memsharded: Added month M and year 'y'

conans/client/migrations.py
memsharded: I have finally added the copy of the DB to a ``cache.sqlite3.backup`` file, with some visible warnings telling about this file and instructing users to restore it if downgrading to < 2.0.10. 

- The automatic down-migrations is a bit too much at this moment. Seems relatively simple on the surface, but not really
- The python script to down-migrate is possible, but not that straightforward, it is not one sqlite command, but a bit more complicated. I thought that a file copy could be a good compromise.

czoido (approved):

conans/client/migrations.py
memsharded: Changed the backup approach for a fully automatic back-migration removing the LRU column, please re-review and try the back-migration.

AbrilRBS (approved):

czoido (approved):

PR #14060, opened 2023-06-09
fix with caching of aggregated components

### LINKED ISSUE #14022: [bug] `full_deploy` generates paths to `res` relative to the cache instead of the deploy folder
### Environment details

* Operating System+version: Windows Server 2019
* Compiler+version: Clang 15.0.7
* Conan version: 2.0.5, 2.0.6
* Python version: 3.11.3

### Steps to reproduce

See [this comment](https://github.com/conan-io/conan/issues/14022#issuecomment-1574036327) for an example command that causes the issue.

1. Install `vulkan-loader` & `vulkan-headers` to some output directory using the `full_deploy` deployer
2. `find_package(Vulkan)`
2. Observe that the imported target for `Vulkan::Vulkan` has a path which is relative to the Conan cache and not the deploy folder
```cmake
  Imported target "Vulkan::Vulkan" includes non-existent path

    "[redacted]/full_deploy/host/vulkan-loader/1.3.239.0/Debug/x86_64/../../../vulka0e9c2dc266fb1/p/res/vulkan/registry"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.
  * An install or uninstall procedure did not complete successfully.
  * The installation package was faulty and references files it does not
  provide.
  ```

### Logs

_No response_

### PR DESCRIPTION
**Pull request by memsharded**
Changelog:  Fix: Avoid caching issues when some intermediate package in the graph calls ``aggregated_components()`` over some dependency and using ``--deployer``, generators still pointed to the Conan cache and not deployed copy. 
Docs: Omit

Close https://github.com/conan-io/conan/issues/14022

### DIFF
--- a/conans/model/build_info.py
+++ b/conans/model/build_info.py
@@ -403,7 +403,6 @@ class CppInfo:
     def __init__(self, set_defaults=False):
         self.components = defaultdict(lambda: _Component(set_defaults))
         self._package = _Component(set_defaults)
-        self._aggregated = None  # A _NewComponent object with all the components aggregated
 
     def __getattr__(self, attr):
         # all cpp_info.xxx of not defined things will go to the global package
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by AbrilRBS — APPROVED** [review]

File: conans/model/build_info.py [hunk 1]
```diff
@@ -490,22 +489,24 @@ def get_sorted_components(self):
 
     def aggregated_components(self):
         """Aggregates all the components as global values, returning a new CppInfo"""
-        if self._aggregated is None:
```
AbrilRBS, line 493: Asking myself if there are more instances of caching in the code that can also be an issue

memsharded: I had a quick look, there is caching in some other places related to the computation of the graph, but nothing evident at first sight that would affect the generators

czoido (approved):

PR #14063, opened 2023-06-09
Git.clone change folder to target one

### LINKED ISSUE #14058: [bug] Git.checkout() with a target folder doesn't work with the remaining Git functions
### Environment details

* Operating System+version: n/a
* Compiler+version: n/a
* Conan version: 2.0.6
* Python version: 3.11

### Steps to reproduce

The `conan.tools.scm.Git` interface seems lacking. If I specify a `target` folder in the `clone()` command, there seems to be no straight-forward way to correctly use the other `Git` functions like `checkout()`. They're all going to be executed in the wrong directory.

```python
    def source(self):
        git = Git(self)
        git.clone(url=self.scm["url"], target=os.path.join(self.source_folder, "src"))
        git.checkout(commit=self.scm["revision"])
```

### Logs

Giving this error:
```
        git.checkout(commit=self.scm["revision"])
        ConanException: Command 'git checkout 42353e2' failed with errorcode '128'
b'fatal: not a git repository (or any parent up to mount point /)\nStopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).\n'
```

memsharded:
Changelog: Fix: Allow spaces in path in ``Git`` helper.
Docs: https://github.com/conan-io/docs/pull/3271

Close https://github.com/conan-io/conan/issues/14058

### DIFF
--- a/conans/test/functional/tools/scm/test_git.py
+++ b/conans/test/functional/tools/scm/test_git.py
@@ -233,12 +233,12 @@ class TestGitBasicClone:
                     # git.checkout(commit="{commit}")
 
                     git = Git(self)
-                    git.clone(url="{url}", target="target") # git clone url target
-                    git.folder = "target"                   # cd target
+                    git.clone(url="{url}", target="tar get") # git clone url target
+                    git.folder = "tar get"                   # cd target
                     git.checkout(commit="{commit}")         # git checkout commit
 
-                    self.output.info("MYCMAKE: {{}}".format(load(self, "target/CMakeLists.txt")))
-                    self.output.info("MYFILE: {{}}".format(load(self, "target/src/myfile.h")))
+                    self.output.info("MYCMAKE: {{}}".format(load(self, "tar get/CMakeLists.txt")))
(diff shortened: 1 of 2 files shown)


conans/test/functional/tools/scm/test_git.py
```diff
@@ -210,6 +210,49 @@ def test_clone_checkout(self):
...
+                    self.folders.source = "source"
+
+                def source(self):
+                    # Alternative, first defining the folder
+                    # git = Git(self, "target")
+                    # git.clone(url="{url}", target=".")
+                    # git.checkout(commit="{commit}")
+
+                    git = Git(self)
+                    git.clone(url="{url}", target="target") # git clone url target
```
AbrilRBS:
```suggestion
                    git.clone(url="{url}", target="target with spaces") # git clone url target
```
So that the quotes behaviour is also checked for, maybe?

AbrilRBS (approved):
Nice! One minor comment about the test

conans/test/functional/tools/scm/test_git.py
memsharded: Great idea, changed it!

PR #14074, opened 2023-06-12
Redundant None checks

jputcu:
Changelog: Omit
Docs: Omit

Lookup returns None when not found, so no default None needed and if it didn't have a value it was already None.
I'm not a Python expert, just getting my feet wet.

### DIFF
--- a/conan/tools/build/flags.py
+++ b/conan/tools/build/flags.py
@@ -178,7 +178,7 @@ def cppstd_flag(settings):
             "apple-clang": _cppstd_apple_clang,
             "msvc": _cppstd_msvc,
             "intel-cc": _cppstd_intel_cc,
-            "mcst-lcc": _cppstd_mcst_lcc}.get(compiler, None)
+            "mcst-lcc": _cppstd_mcst_lcc}.get(compiler)
     flag = None
     if func:
         flag = func(Version(compiler_version), str(cppstd))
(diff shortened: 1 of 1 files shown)


conan/tools/build/flags.py
```diff
@@ -203,11 +203,11 @@ def cppstd_msvc_flag(visual_version, cppstd):
     if visual_version >= "193":
         v23 = "c++latest"
 
-    return {"14": v14, "17": v17, "20": v20, "23": v23}.get(str(cppstd), None)
+    return {"14": v14, "17": v17, "20": v20, "23": v23}.get(str(cppstd))
 
 def _cppstd_msvc(visual_version, cppstd):
     flag = cppstd_msvc_flag(visual_version, cppstd)
-    return "/std:%s" % flag if flag else None
+    return "/std:%s" % flag
```
memsharded: This one might might return a literal string ``/std:None`` which is undesired. Which is different than returning ``None`` (the priority seems to be first the string interpolation, then the if, so ``return None`` if flag is not defined

memsharded (commented):
It seems the ``.get()`` can be simplified, but the ``flags`` cannot. 

Thanks for the contribution!

AbrilRBS (commented):
Thanks a lot for your contribution, we really appreciate it :)
As @memsharded pointed out, the flag changes seem to have unintended consequences, but we love the spirit of improving code quality where possible :)

AbrilRBS (approved):

conan/tools/build/flags.py
jputcu: Thanks for the quick feedback, trying with small steps. The unittests did a fine job ;)

memsharded (approved):

memsharded:
Thanks!

PR #14078, opened 2023-06-13
Refactor/db latest recipes

memsharded:
Changelog: Omit
Docs: Omit

Only internal refactors around DB, including some optimizations avoiding double calls to DB

### DIFF
--- a/conans/client/graph/proxy.py
+++ b/conans/client/graph/proxy.py
@@ -33,13 +33,14 @@ class ConanProxy:
             return conanfile_path, RECIPE_EDITABLE, None, reference
 
         # check if it there's any revision of this recipe in the local cache
-        if reference.revision:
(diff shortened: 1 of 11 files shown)


conans/client/graph/proxy.py
```diff
@@ -33,10 +33,14 @@ def _get_recipe(self, reference, remotes, update, check_update):
             return conanfile_path, RECIPE_EDITABLE, None, reference
 
         # check if it there's any revision of this recipe in the local cache
-        ref = self._cache.get_latest_recipe_reference(reference)
+        try:
```
memsharded, line 36: Not necessary to do 2 DB calls, first get the reference, then the necessary layout, but we can get in one db call the whole layout.

czoido (approved):

PR #14082, opened 2023-06-13
`conan remove` pkg-lists support

memsharded:
Changelog: Feature: Allow ``conan remove --list=pkglist`` to remove package-lists.
Changelog: Feature: Output for ``conan remove --format`` both text (summary of deleted things) and json.
Docs: https://github.com/conan-io/docs/pull/3270

Need it as pre-condition for LRU functionality in https://github.com/conan-io/conan/pull/14054

### DIFF
--- a/conan/cli/commands/remove.py
+++ b/conan/cli/commands/remove.py
@@ -1,6 +1,6 @@
 from conan.api.conan_api import ConanAPI
 from conan.api.model import ListPattern, MultiPackagesList
-from conan.api.output import cli_out_write
+from conan.api.output import cli_out_write, ConanOutput
 from conan.cli import make_abs_path
 from conan.cli.command import conan_command, OnceArgument
 from conan.cli.commands.list import print_list_json, print_serial
@@ -18,12 +18,12 @@ def summary_remove_list(results):
     for remote, remote_info in info.items():
         new_info = result.setdefault(remote, {})
         for ref, content in remote_info.items():
-            for rev, rev_content in content["revisions"].items():
+            for rev, rev_content in content.get("revisions", {}).items():
                 pkgids = rev_content.get('packages')
                 if pkgids is None:
-                    new_info.setdefault(f"{ref}:{rev}", "Removed recipe and all binaries")
+                    new_info.setdefault(f"{ref}#{rev}", "Removed recipe and all binaries")
                 else:
-                    new_info.setdefault(f"{ref}:{rev}", f"Removed binaries: {list(pkgids)}")
+                    new_info.setdefault(f"{ref}#{rev}", f"Removed binaries: {list(pkgids)}")
     print_serial(result)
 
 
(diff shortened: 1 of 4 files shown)


conan/cli/commands/remove.py
```diff
@@ -27,26 +50,44 @@ def remove(conan_api: ConanAPI, parser, *args):
...
 
-    ref_pattern = ListPattern(args.pattern, rrev="*", prev="*")
-    select_bundle = conan_api.list.select(ref_pattern, args.package_query, remote)
-
-    if ref_pattern.package_id is None:
-        if args.package_query is not None:
+    if args.list:
+        listfile = make_abs_path(args.list)
+        multi_package_list = MultiPackagesList.load(listfile)
+        package_list = multi_package_list["Local Cache" if not remote else remote.name]
```
AbrilRBS, line 72: Maybe time to move "Local Cache" to some kind of constant? (Not on this PR) It's in the codebase a few times by now

AbrilRBS (approved):
Should we be testing the new argument in small tests too? I know there's one describing a use case for download + upload, but then this one would seem to be missing

memsharded:
> Should we be testing the new argument in small tests too? I know there's one describing a use case for download + upload, but then this one would seem to be missing

Yes, definitely. I will add tests. I was trying to make sure things didn't break and interface looked good, and I forgot to finally add them.

AbrilRBS (approved):

conans/test/integration/command_v2/test_combined_pkglist_flows.py
```diff
@@ -162,3 +162,51 @@ def test_download_upload_only_recipes(self, client, prev_list):
...
+            "zli.py": GenConanfile("zli", "1.0.0")
+        })
+        c.run("create zli.py")
+        c.run("create zlib.py --version=1.0.0 --user=user --channel=channel")
+        c.run("upload * -r=default -c")
+        return c
+
+    def test_remove_nothing_only_refs(self, client):
+        # It is necessary to do *#* for actually removing something
+        client.run(f"list * --format=json", redirect_stdout="pkglist.json")
```
memsharded, line 182: This is one of the slightly not great things: the difference in defaults between ``conan list`` and ``conan remove``:

- ``conan remove *`` is able to remove everything (defaults to #*:*#*)
- ``conan list *`` only lists the recipes, not even revisions. If we feed the output of this pattern to ``conan remove``, it won't remove anything (this is the reason I added a warning)

So it is necessary to do ``conan list *#* >pkglist`` + ``conan remove --list=pkglist`` if we want to actually remove something. I think it is consistent with the ``conan list`` behavior, but please review.

memsharded: We have discussed about this, and it seems ok. 
It will require explaining it in the docs, but this is expected and aligned with the ``conan remove`` command defaults, which are way more important.

czoido (approved):

PR #14084, opened 2023-06-13
Some more small code fixes

jputcu:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conan/tools/microsoft/subsystems.py
+++ b/conan/tools/microsoft/subsystems.py
@@ -7,8 +7,8 @@ def unix_path(conanfile, path, scope="build"):
 
 
 def unix_path_package_info_legacy(conanfile, path, path_flavor=None):
-    message = f"The use of 'unix_path_legacy_compat' is deprecated in Conan 2.0 and does not " \
-              f"perform path conversions. This is retained for compatibility with Conan 1.x " \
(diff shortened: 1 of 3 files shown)


memsharded (approved):
Fantastic, thanks!

AbrilRBS (approved):
TIL about NotImplementedError vs NotImplemented, great catch!
Need to double check that nothing os trying to catch the wrong one, but the changes look good

jputcu:
This time, I can't take any credit, I just unleashed `pylint` and `pyflakes` source directories.
There are lots of warnings/hints, so perhaps it is worth a CI/CD step?

memsharded:
> There are lots of warnings/hints, so perhaps it is worth a CI/CD step?

We might add it at some point, but we have been prioritizing testing over linting, typically because of the noise of false positives it can get a bit annoying (or nobody really looks at them as warnings in CI, only hard failures are paid attention...)

So I think at some point we want a quality build that includes linting, coverage, etc, but not sure how to introduce it in a way that it is not too annoying for the current status and dev flow.

jputcu:
One way would be to only activate them for certain files or directories, especially new ones.
I understand you have a lot of work to do, so I won't try to bother you with the smaller ones.

PR #14086, opened 2023-06-14
Add functionality to cmake_lib template

jcar87:
Changelog: Omit
Docs: Omit

Two small features added to the `cmake-lib` template:

* Print the msvc runtime variant used - use the macros described in the [documentation](https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170#remarks) to determine at compile time which runtime library is being used - this can be useful in tests to assert that the intended runtime is used

* Add function to the template library that passes `std::vector<std::string>` in its arguments, and call it from the `test_package` - this can be a good way of exercising the correct C++ runtime is used on both the calling site and the library. If there are mismatches, this could cause iterator debug level errors with Visual C++, or linker errors due to libstdc++ C++11 ABI.

### DIFF
--- a/conan/internal/api/new/cmake_lib.py
+++ b/conan/internal/api/new/cmake_lib.py
@@ -155,13 +155,13 @@ void {{package_name}}(){
 
     // MSVC runtime
     #if defined(_DEBUG)
-        #if defined(_MT) and defined(_DLL)
+        #if defined(_MT) && defined(_DLL)
         std::cout << "  {{name}}/{{version}}: MSVC runtime: MultiThreadedDebugDLL\n";
         #elif defined(_MT)
         std::cout << "  {{name}}/{{version}}: MSVC runtime: MultiThreadedDebug\n";
         #endif
     #else
-        #if defined(_MT) and defined(_DLL)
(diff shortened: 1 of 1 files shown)


conan/internal/api/new/cmake_lib.py
```diff
@@ -208,6 +227,12 @@ def build_requirements(self):
     std::cout << "  {{name}}/{{version}}: __CYGWIN__" << __CYGWIN__<< "\n";
     #endif
 }
+
+void {{package_name}}_print_vector(const std::vector<std::string> &strings) {
+    for(std::vector<std::string>::const_iterator it = strings.begin(); it != strings.end(); ++it) {
+        std::cout << "{{package_name}}/{{version}} " <<  *it << std::endl;
```
AbrilRBS:
```suggestion
        std::cout << "{{package_name}}/{{version}} " << *it << std::endl;
```

AbrilRBS (approved):
Nice! Some tiny minor styling fix, but otherwise looks great

memsharded (approved):

PR #14096, opened 2023-06-15
import conan tools improve

memsharded:
Changelog: Omit
Docs: Omit

I am concerned of excessive importing ``from conans`` in our own ``conan.tools`` that users copying/forking it, will also contain too many ``from conans`` imports that are private and will eventually break. So:

- Quick fix for exceptions that is an issue one
- Open the conversation about ``save, load`` helpers

### DIFF
--- a/conan/tools/cmake/cmakedeps/cmakedeps.py
+++ b/conan/tools/cmake/cmakedeps/cmakedeps.py
@@ -9,9 +9,9 @@ from conan.tools.cmake.cmakedeps.templates.macros import MacrosTemplate
 from conan.tools.cmake.cmakedeps.templates.target_configuration import TargetConfigurationTemplate
 from conan.tools.cmake.cmakedeps.templates.target_data import ConfigDataTemplate
 from conan.tools.cmake.cmakedeps.templates.targets import TargetsTemplate
+from conan.tools.files import save
 from conans.client.generators import relativize_generated_file
-from conans.errors import ConanException
-from conans.util.files import save
+from conan.errors import ConanException
 
 
 class CMakeDeps(object):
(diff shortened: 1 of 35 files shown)


conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -42,7 +42,7 @@ def generate(self):
         # Current directory is the generators_folder
         generator_files = self.content
         for generator_file, content in generator_files.items():
-            save(generator_file, content)
+            save(self._conanfile, generator_file, content)
```
memsharded, line 45: Should we change the ``save()`` and ``load()`` usages to use the ``from conan.tools.files`` ones?

franramirez688 (approved):

conan/tools/cmake/cmakedeps/cmakedeps.py
AbrilRBS: My question goes the other way around. Any reason why we would not want to?

memsharded: The current implementation of internal ones feature a ``save(..., only_if_modified, )`` which in theory is useful for generators, but in practice not used anymore, that the ``tools.files`` ones are missing.

But otherwise, yes, I think trying to make the ``conan.tools`` as self-contained as possible is good. There will be some exceptions for our own internal deep things that should go to ``from conan.internal`` to clearly expose and differentiate that detail.
The problem with the current status is that it is nearly unnoticeable that you are using things from a "private/internal" scope that you shouldn't be using

memsharded:
I am merging it as it is good, the ``load()``, ``save()`` requires a bit more of time, as they don't have access to the conanfile everywhere, it is good that we know that we want to move there, and it can happen in future PRs.

PR #14102, opened 2023-06-15
refactor some profile tests

memsharded:
Changelog: Omit
Docs: Omit

Xfailed tests were too old and broken, non-recoverable, removing them
Also moving test to integration, as it doesn't use any external tools


AbrilRBS (approved):

AbrilRBS:
The CI says there are a few xpassed tests in general, might be cool to check those out, they ought to be easy to update then

Issue #13889, opened 2023-05-13
[question] adding libraries to LDLIBS


igagis:
### What is your question?

Conan adds library path to all libraries from dependencies to `LDFLAGS` when using `AutotoolsDeps` generator.
But it does not add the actual libraries (from `cpp_info.libs`) to `LDLIBS`. Is it possible to make it do that?

Also, I have library C which depends on library B, which, in turn depends on library A, but B doesn't expose any API from A. I.e. B only needs A while building and uses it internally.

Now, if A, B and C are all packaged as static libraries. This means that when using library C in some program, one has to give `-lC` linker flag and also `-lB -lA`, otherwise linker will complain that there are undefined references to A in B and to B in C.

So, I want that when using C conan generator would add `-lA -lB -lC` to `LDLIBS`, is that possible to achieve?

### Have you read the CONTRIBUTING guide?

- [x] I've read the CONTRIBUTING guide

igagis:
Or maybe, when using C there is no need to add `-lC` automatically, but only `-lA -lB`, since those come transitively, and in case of dynamically linked `C` those don't have to be given to linker, but in case of static `C` we need those `-lA -lB`.

igagis:
The explanation is a bit cumbersome, but I hope you get the idea :)

igagis:
I've just found that in the docs https://docs.conan.io/2/reference/tools/gnu/autotoolsdeps.html
it says that `LIBS` variable will be set, and I checked that it is actually set to `-l<...>` values.
I'm wondering why is it `LIBS` and not `LDLIBS` which is the actual one used by `GNU make`, according to https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-LDLIBS ?

agagniere:
Yep, I always have to add `LDLIBS += $(LIBS)` in my Makefiles, which kind of defeats the point of AutotoolsDeps...
Maybe we can create a pull request to add `LDLIBS` in addition to `LIBS` : https://github.com/conan-io/conan/blob/release/2.0/conan/tools/gnu/autotoolsdeps.py#L72

agagniere:
So, while GNU make uses `LDLIBS` by default, it seems the rest of the Autotools project does not follow that convention, and in `configure.ac` of big projects like ImageMagick, I see :
```bash
for var in CC CFLAGS CPPFLAGS CXX CXXCPP LDFLAGS AR_FLAGS LIBS ; do
```
[...]

PR #14133, opened 2023-06-20
Add MakeDeps generator

### LINKED ISSUE #13394: [question] What's the intended way to work with makefiles without autotools in 2.0
### What is your question?

I'm new to Conan with 2.0 and my team is currently using custom makefiles without autotools to build our projects. Googling around I end up on 1.x docs of Conan where there does seem to be built in generators for makefiles. So how do I work with makefiles in 2.0?

Packaging at least seems easy. I can just not have a `build()` function and use a wrapper script to build with make first and then call `conan create`. Maybe I could have the `build()` function call the python `subprocess` library to handle the build itself though?

However consuming conan packages is not as straight forward. Should I just be using the `VirtualBuildEnv` and relying on env vars in my makefiles?

Are there any good example projects that build with makefiles and conan 2.0?

uilianries:
Changelog: Feature: Add support to Makefile by the new MakeDeps generator
Docs: https://github.com/conan-io/docs/pull/3348
closes #13394

This PR adds a new feature, the `MakeDeps` generator, that can be used with Conan install command: `conan install --requires=foobar/0.1.0 -g MakeDeps`. The idea is bringing back the `make` generator from Conan 1.x but with steroids, including support for components.
As real example, using Poco package, we will have:

```
% conan install --requires=poco/1.12.1 -g MakeDeps 
% ls *.mk
conandeps.mk
``` 

The are 3 levels of dependencies here:

* Each Poco component and its dependencies will generate separated variables, that contains values related to that component, plus, dependencies;
* Then, each package, include Poco, will have variable with all components listed, and all variables appending components;
* As top level in conandeps.mk, all global variables like CONAN_LIB_DIRS, will list all dependencies.


conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,602 @@
...
+            return _get_package_aliases(dep)
+        raise ConanException("Component '{name}::{cname}' not found in '{name}' "
+                             "package requirement".format(name=_get_package_reference_name(dep),
+                                                          cname=comp_name))
+    comp_aliases = dep.cpp_info.components[comp_name].get_property("make_aliases")
+    return comp_aliases or []
+
+
+def _get_package_name(dep, build_context_suffix=None):
+    pkg_name = dep.cpp_info.get_property("make_name") or _get_package_reference_name(dep)
```
memsharded: Why custom make names are necessary? As there is no standard in this regard, and Conan is inventing this integration, I think we can mandate the makefile names to always match package names and components, no need for properties.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,602 @@
...
+            "requires": info.requires,
+            "makeifyed_requires": [self._makefy(r) for r in info.requires],
+            "variables": variables,
+        }
+        template = Template(self.component_template, trim_blocks=True,
+                            lstrip_blocks=True, undefined=StrictUndefined)
+        return template.render(context)
+
+
+class _MakeGlobalContentGenerator(_MakeContentGenerator):
```
memsharded: I'd prefer to avoid inheritance, and use composition, or reuse of free functions.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,602 @@
...
+    Generate and store all the .mk files for all the dependencies
+    """
+
+    def __init__(self, conanfile):
+        """
+        :param conanfile: The current recipe object
+        """
+        self._conanfile = conanfile
+        self.build_context_activated = []
+        self.build_context_suffix = {}
```
memsharded: We can drop the build context at the moment, lets start small and simpler with host context first

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,213 @@
...
+import re
+import textwrap
+
+import pytest
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+@pytest.mark.skipif(platform.system() not in ["Linux", "Darwin"], reason="Make")
```
memsharded: The test doesn't really execute Make, so it should run in Windows too?

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,213 @@
...
+    from conan import ConanFile
+
+    class Dep(ConanFile):
+
+        settings = "os", "arch", "compiler", "build_type"
+
+        def package_info(self):
+            self.cpp_info.includedirs = []
+            self.cpp_info.includedirs.append("path/includes/{}".format(self.name))
+            self.cpp_info.includedirs.append("other\\include\\path\\{}".format(self.name))
```
memsharded: If test not Windows, why using \\ backslash in the path?

memsharded (commented):
It is looking good, please have a look at the comments.

Also, it would be better to focus on 1 good functional test that proves that this generator can actually be used to link with a regular library, and start from there to more advanced and complete cases (but that can happen later in other PRs)

conan/tools/gnu/makedeps.py
uilianries: Indeed not needed, I'll remove it, one less feature to be maintained/tested.

conan/tools/gnu/makedeps.py
uilianries: Okay, makes sense. Most of this generator I made based on pkg-config deps, that's a very complete generator, but I forgot a very important information, pkg-config is consumed by autotools, but make deps is something totally separated.

conans/test/integration/toolchains/gnu/test_makedeps.py
uilianries: sorry, those tests are not ready, I need to change many things here.

conans/test/integration/toolchains/gnu/test_makedeps.py
uilianries: this test is invalid, I'll remove it. I would prefer not supporting Windows with MakeDeps for now, we have NMake already for windows.

memsharded: Yes, make sense, you can start with support for Linux/OSX only, and we can bring Windows (via msys2) later

iskunk:
Trying this out on my side. A few comments:

* What's the motivation for breaking things out into various separate `*.mk` files? I haven't seen any other generators do that.

* In the `conandeps.mk` file, I see a bit like
  ```
  # Global dependencies
  include libxslt.mk libxml2.mk zlib.mk libiconv.mk
  ```
  This assumes that the files are in the current directory, so if your project is doing e.g. `include /path/to/conandeps.mk`, Make will fail to find them.

* Why the ubiquitous use of `+=`? Note that non-GNU Make programs may not support this operator. (I considered using this, perhtaps with a tool option to allow disabling them if needed, but then I didn't really see a point to it. Appending values is meaningfully different only if you have variable names colliding (which shouldn't happen) or if the consuming project sets a value on them first (I couldn't think of why they would want to do that).

uilianries:
@iskunk thank you for reviewing, your feedback is valuable.

> What's the motivation for breaking things out into various separate *.mk files? I haven't seen any other generators do that.

Most of generators split in several files, including pkg-config and cmake, please, take a look in https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#generated-files for instance.

> This assumes that the files are in the current directory, so if your project is doing e.g. include /path/to/conandeps.mk, Make will fail to find them.

Indeed it's a fragile point, need to be improved, thank you for spotting it.

> Why the ubiquitous use of +=? Note that non-GNU Make programs may not support this operator. (I considered using this, perhtaps with a tool option to allow disabling them if needed, but then I didn't really see a point to it. Appending values is meaningfully different only if you have variable names colliding (which shouldn't happen) or if the consuming project sets a value on them first (I couldn't think of why they would want to do that).

GNU Make is by far, the most popular and well documented Make implementation, that's why the choice. Append variables is not expected but can occur, as users demand all kind of customization. Anyway, using only `=` should simplify the implementation even more, so it should it adapted.

iskunk:
> Most of generators split in several files, including pkg-config and cmake, please, take a look in https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#generated-files for instance.

Ah, okay, now I remember those.

I'd point out that the CMake `find_package()` and pkg-config mechanisms both obtain information on a specific dependency by looking for a corresponding filename (e.g. `FindFoo.cmake` or `foo.pc`), so by necessity they have to break things out into separate files. That's not the case with plain Make, so I don't really see a benefit to taking that approach here.

> GNU Make is by far, the most popular and well documented Make implementation, that's why the choice. Append variables is not expected but can occur, as users demand all kind of customization. Anyway, using only = should simplify the implementation even more, so it should it adapted.

One of my potential use cases is running a build on AIX or Solaris, and GNU Make isn't always available, so the greater compatibility will be a plus. (I forgot to point out the `?=` and `$(sort ...)` bits as well. The `sort` was especially concerning since the order of at least `CONAN_LIBS` can be meaningful.)

uilianries:
@iskunk 

> I'd point out that the CMake find_package() and pkg-config mechanisms both obtain information on a specific dependency by looking for a corresponding filename (e.g. FindFoo.cmake or foo.pc), so by necessity they have to break things out into separate files. That's not the case with plain Make, so I don't really see a benefit to taking that approach here.

Yes, but keeping the flexibility of what you want to include is also a feature. For those cases when you want to use a separated folder to generate your makefiles, now you can use the definition `CONAN_MAKEFILES_DIR` when invoking make, so `include` directive will have a prefix with the folder name.

> One of my potential use cases is running a build on AIX or Solaris, and GNU Make isn't always available, so the greater compatibility will be a plus. (I forgot to point out the ?= and $(sort ...) bits as well. The sort was especially concerning since the order of at least CONAN_LIBS can be meaningful.)

Those are very rare cases, most used for production, not development (I know some people using AIX). About the $(sort) you are right, need to be improved and will. The `+=` and `?=` were replaced by the simple `=`.

iskunk:
> Yes, but keeping the flexibility of what you want to include is also a feature. For those cases when you want to use a separated folder to generate your makefiles, now you can use the definition `CONAN_MAKEFILES_DIR` when invoking make, so `include` directive will have a prefix with the folder name.

All these files do is assign variables. If there are some dependencies in there that I don't want to use, then I can just not use those variables in my makefile. What problem is solved by including `foobar.mk` instead of `conandeps.mk`?

I notice that the individual `*.mk` files are each including their own component dependencies, too, which means that many of these are being multiply included---by a large margin. I used your example of generating for `poco/1.12.1`,  and added this line to the top of `poco_foundation.mk`:
```
$(info hello from poco_foundation.mk)
```
Each time GNU Make sees that, it will print out that message. I wrote a trivial makefile that includes `conandeps.mk`. When I run it, here is what I see:
```
$ make
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
hello from poco_foundation.mk
true
```

Also, `CONAN_LIBS_POCO` is the concatenation of all the component `CONAN_LIBS_POCO_*` variables, which is incorrect behavior. Whatever the package sets for its top-level `cpp_info.libs` field is what should show up there; the generator should not make up its own value in this manner. If it makes sense for a package to aggregate all its component libraries into the top-level component, then it will do so, in its `package_info()` method. (Same issue applies to the other variables; it is just particularly egregious for `LIBS`)

> Those are very rare cases, most used for production, not development (I know some people using AIX).

My organization is using Conan to drive production builds of third-party dependencies. If all I cared about were development, we'd just use Conda.

The Conan buildenv mechanism is especially helpful on traditional Unix systems that can't run Python 3---I can run Conan on a Linux system, generate into a shared NFS workspace, have the Unix system load the buildenv and build in that NFS workspace, and then Conan on Linux can pick up the artifacts.

iskunk:
Much obliged @uilianries, this is looking pretty good.

I have a concern with how variables for components are named. In my implementation, I did a straight conversion from the `foo::bar` names, for example

* `boost::atomic` --> `CONAN_LIBS_BOOST_ATOMIC`
* `pcre2::pcre2-8` --> `CONAN_LIBS_PCRE2_PCRE2_8`

You mentioned that you didn't like the "PCRE2_PCRE2" name-doubling, and I see your code detects and drops the redundant part of the name, e.g.

* `boost::atomic` --> `CONAN_LIBS_BOOST_ATOMIC`
* `pcre2::pcre2-8` --> `CONAN_LIBS_PCRE2_8`

But say you have a package with component names like

* `foo::bar`
* `foo::foo-bar`

You'll get two `CONAN_LIBS_FOO_BAR` assignments. (This name combination may be unusual, but I don't believe it's prohibited)

The name-doubling from my implementation wasn't pretty, but it only reflected the redundant names in the original components. In the case of `pcre2`, the overall package is called `pcre2`, and the individual component libraries are named `pcre2-8`, `pcre2-16`, etc.---they are never referred to as `2-8`, nor just `8`, etc. There is no shorter, recognized name that can be given to them that would eliminate "pcre2" appearing twice in the package::component identifier.

(I have remarks on some other aspects of the current draft, but this is the main one for now)

uilianries:
@iskunk that's a nice feedback indeed, I didn't consider that scenario, because is rare, but still valid. I'll update the PR with the package name as prefix. Thank you!

iskunk:
All right, I gave the generator another test run. Here are my observations:

* It's good to have the `CONAN_HOST_*` and `CONAN_BUILD_*` settings variables, but didn't @memsharded say those needed to go into a toolchain file, rather than the dependencies? (I'm assuming this means something like a separate `conantoolchain.mk` file, but am not clear on the distinction)

* A couple things with a line like
  ```
  CONAN_SYSROOT_ZLIB = $(CONAN_ROOT_ZLIB)/.
  ```
  What's the `/.` at the end for? That is a redundant directory reference. (`/tmp/foo`, `/tmp/foo/.`, and `/tmp/foo/./././.` all refer to the same directory)

  Secondly, why is the SYSROOT variable being set for a package that doesn't define it? (If you look at zlib's conanfile, there is no `self.cpp_info.sysroot` assignment, so this value is coming out of thin air)

* I noticed the behavior of how (at least) `CONAN_{LIBS,DEFINES,REQUIRES}_*` are aggregated from the components if there is no value set at the top level. So for example, for `pcre2/10.42`, we get
  ```
  CONAN_LIBS_PCRE2 = \
        $(CONAN_LIBS_PCRE2_PCRE2_8) \
        $(CONAN_LIBS_PCRE2_PCRE2_POSIX) \
        $(CONAN_LIBS_PCRE2_PCRE2_16) \
        $(CONAN_LIBS_PCRE2_PCRE2_32)
  ```
  I don't see how this is helpful... this assumes that the consumer might want to use all the component libraries together, but that's not the case for a package like PCRE2. (Most applications, if they use PCRE2, use only one of those libraries.) Maybe it makes sense to do this for some packages, but then, those packages can always set things up this way explicitly in their conanfile. Setting an aggregated value like this in the generator has the potential to confuse the developer, as it suggests a way of using the package that is not correct.

  In general, I'd be very careful with having the generator "lie" about the value of package variables for the sake of convenience. Usually, it is better to fix up the conanfile to give those more convenient values.

* If a package sets `.defines` at both the top and component levels, then the top-level `CONAN_DEFINES_*` variable is missing the `$(CONAN_DEFINE_FLAG)` prefix. (For example, try adding `self.cpp_info.defines = "FOO"` to the PCRE2 conanfile)

uilianries:
@iskunk Thank you for checking the implementation again. The patch is now much simpler, without jinja2.

> What's the /. at the end for? That is a redundant directory reference. (/tmp/foo, /tmp/foo/., and /tmp/foo/./././. all refer to the same directory)
> Secondly, why is the SYSROOT variable being set for a package that doesn't define it? (If you look at zlib's conanfile, there is no self.cpp_info.sysroot assignment, so this value is coming out of thin air)

The `/.` means current directory, but indeed is not needed. Indeed it's a bug that should be fixed, as sysroot is not assigned by default in cpp_info. It should be removed in a next commit. Thank you for spotting it.

> It's good to have the CONAN_HOST_* and CONAN_BUILD_* settings variables, but didn't @memsharded say those needed to go into a toolchain file, rather than the dependencies? (I'm assuming this means something like a separate conantoolchain.mk file, but am not clear on the distinction)

Yes, splitting makes sense, as we have done for environment files in Conan 2.x

> I don't see how this is helpful... this assumes that the consumer might want to use all the component libraries together, but that's not the case for a package like PCRE2. (Most applications, if they use PCRE2, use only one of those libraries.) Maybe it makes sense to do this for some packages, but then, those packages can always set things up this way explicitly in their conanfile. Setting an aggregated value like this in the generator has the potential to confuse the developer, as it suggests a way of using the package that is not correct.

If you want to use only one component, you pick only those variables related to that specific component. I don't see the scenario that you describe, maybe could write a test to validate the case, but I still see it as pretty rare case that would not worth covering now.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,216 @@
...
+                "REQUIRES": None,
+            }
+
+            for var, prefix_var in common_variables.items():
+                cppinfo_value = getattr(cpp_info, var.lower())
+                # Use component cpp_info info when does not provide any value
+                if not cppinfo_value and hasattr(cpp_info, "components"):
+                    cppinfo_value = [self._var_ref(f"CONAN_{var}_{_makeify(var_id)}_{_makeify(name)}") for name, obj in cpp_info.components.items() if getattr(obj, var.lower())]
+                    # avoid repeating same prefix twice
+                    prefix_var = None
```
iskunk: To reiterate, I'm leery of this idea of synthesizing values for unset top-level package variables from the components. Do other generators do this?

While it poses no immediate problem for my usage (since I would ignore those variables anyway), I can see this causing headaches in maintaining packages with components. For example, a `foo` package maintainer starts out not setting `cpp_info.libs`, so we get the synthesized `CONAN_LIBS_FOO` aggregate value. People use the synthesized `CONAN_LIBS_FOO`, since it seems like the obvious thing to do. Later on, the maintainer decides to start setting `.libs` to some default library for the package (something that makes sense in context vis-a-vis the components). Now, people who were depending on the synthesized value get broken builds, as they may need to link in component libraries that are no longer listed in `CONAN_LIBS_FOO`.

And the maintainer is pressured to walk back the change. This discourages them from using a variable that they had never set before, because by not setting it, they were "opting in" to using this synthesized value.

IMHO, if it makes sense for `CONAN_LIBS_FOO` (or some other top-level package variable) to list the aggregated component values, then that should be set explicitly in the conanfile. Having the generator try to be "helpful" in this way is liable to cause problems down the road.

iskunk:
In the `conandeps.mk` output for `pcre2/10.42`, I see e.g.
```
CONAN_LIB_DIRS_PCRE2 = $(CONAN_LIB_DIRS_FLAG)$(CONAN_ROOT_PCRE2)/lib

...

CONAN_LIBS_PCRE2_PCRE2_8 = $(CONAN_LIB_FLAG)pcre2-8
```

When there is a `CONAN_xxxxx_FLAG` variable, I can see an argument for `xxxxx` being plural (e.g. `LIB_DIRS`) or singular (`LIB_DIR`). But either way, it should be consistent.

On a separate note: Don't lose sight of using the `_var_ref()` and `_var_assign_{single,multiple}()` methods to support a "bonus" shell generator. The logic is mostly the same as with Make; the only difference is that variables are expanded immediately, so you have to make sure they are set before referencing them.

conans/test/functional/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,119 @@
...
+    client.run("export . --name=hello --version=0.1.0")
+    client.run("install --requires=hello/0.1.0 --build=missing -g MakeDeps")
+    client.run_command(f"make --print-data-base -f {CONAN_MAKEFILE_FILENAME}", assert_error=True)
+    assert r'CONAN_CXXFLAGS_HELLO = flag2=\"my flag2\"' in client.out
+    assert r'CONAN_CFLAGS_HELLO = flag1=\"my flag1\"' in client.out
+    assert r'CONAN_DEFINES_HELLO = $(CONAN_DEFINE_FLAG)USER_CONFIG="user_config.h" $(CONAN_DEFINE_FLAG)OTHER="other.h"' in client.out
+
+
+@pytest.mark.tool("cmake")
+def test_makedeps_with_tool_requires():
```
memsharded, line 36: This is a very expensive test, with creation of 3 full packages, compiled libraries, for very light checking, that could be done with empty recipes that do not build anything. So this could be an integration test that doesn't depend on cmake at all

conans/test/functional/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,119 @@
...
+                            copy(self, "libhello.a", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"))
+                            copy(self, "hello.h", src=self.source_folder, dst=os.path.join(self.package_folder, "include"))
+
+                        def package_info(self):
+                            self.cpp_info.libs = ["hello"]
+                    ''')
+                     })
+        client.run('create . --name=hello --version=0.1.0 -c tools.microsoft.bash:subsystem=msys2 -c tools.microsoft.bash:path=bash')
+    with client.chdir("app"):
+        client.run("install --requires=hello/0.1.0 -pr:b=default -pr:h=default -g MakeDeps -of build")
```
memsharded, line 106: But if ``MakeDeps`` is used only in the consumer side, why the complexity of building the package with Autotools, that is slow and complicated? It should work with a standard template recipe, even cmake_lib

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,429 @@
...
+        import os
+        from conan import ConanFile
+
+        class TestMakeDirsConan(ConanFile):
+            name = "mylib"
+            version = "0.1"
+
+            def package_info(self):
+                self.cpp_info.frameworkdirs = []
+                self.cpp_info.filter_empty = False
```
memsharded: ``filter_empty`` no longer exist or has meaning in 2.0, can be removed

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,216 @@
...
+                comp_name_list.append(comp_name)
+                _output_cpp_info_vars(f"{dep_id}_{comp_id}", component, root, dep_id, False, dep)
+
+            comp_section = _divert()
+
+            line_buffer.extend(dep_section)
+            _output_var(f"CONAN_COMPONENTS_{dep_id}", comp_name_list)
+            line_buffer.extend(comp_section)
+
+            body.extend(_divert())
```
memsharded: This rendering approach reads a bit complicated. I'd prefer if there would be some visibility over the final result template.

conans/test/functional/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,119 @@
...
+        client.save({"Makefile": textwrap.dedent('''
```
memsharded, line 107: We need a test like this, but linking a package with a couple of components.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,216 @@
...
+            root_with_sep = root + os.path.sep
+            root_var_ref = self._var_ref(f"CONAN_ROOT_{root_id}")
+            return [root_var_ref + path[root_len:]  \
+                    if path.startswith(root_with_sep) else path for path in path_list]
+
+        # Output a variable assignment, appropriately formatted
+        def _output_var(name, item, root=None, root_id=None, prefix_var=None, blank_after=True):
+            if item is None:
+                return False
+            if type(item).__name__ in ('Component', 'SettingsItem', 'Version', 'str'):
```
memsharded: How are SettingsItem involved here?

I think this ``_output_var()`` function is quite complicated to understand what is doing.

memsharded (commented):
It would be good to try to make the generation more readable, I find this code challenging to understand. I'd try the following:

- Split the generation of code for 1 dependency into an independent class (divide and conquer)
- Use templates more heavily. That helps a lot to understand the final expected output.
- Make ``var_aggregates`` a explicit return of the generation of partial code.

conans/test/functional/toolchains/gnu/test_makedeps.py
uilianries: Done. No longer using build step, but only checking cpp_info through dependencies.

conans/test/functional/toolchains/gnu/test_makedeps.py
uilianries: Fixed. Updated to use only `make` command directly.

conans/test/functional/toolchains/gnu/test_makedeps.py
uilianries: Done. Included extra step that links to components.

conan/tools/gnu/makedeps.py
uilianries: I once found a comment in a code, like: "Only God can understand this code"

I'll simplify by using Jinja templates. I need to design something better, because we have 3 levels here, global (CONAN_LIBS), dependencies (hellolib, zlib, ...) and their components. I could write a template for each and then unity them in a single file.

uilianries:
@memsharded Ready for a second round of review. Some considerations:

- Toolchain tests now are simplified, don't need to build most of cases, as we only want to check cpp_info
- Using Jinja2 and templates to provide the output, but still, need to burn some phosphate to understand. 
- There are 3 levels in the Makefile:
  - Global stores only information about dependencies, but it also consider those cases where components add a folder. It also lists all dependencies as their original names (CONAN_DEPS)
  - Dependencies are listed individually, folders, flags, variables are listed
  - Components are listed together to its Main dependency as we need to rootify folders

iskunk:
I looked at the output from the latest code. Need a blank line before `CONAN_DEPS`, and there should be two blank lines before and after each package to aid in navigating the file.

Also, don't forget to make the `CONAN_xxxxx_FLAG` pluralization consistent.

The rest looks fine.

uilianries:
@iskunk Good catch, thank you!

conans/test/functional/toolchains/gnu/test_makedeps.py
memsharded: Great!

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,588 @@
...
+        self._info = MakeInfo(self._dep.ref.name, [], [])
+
+    def make_global_info(self) -> MakeInfo:
+        """
+        Remove duplicated folders and flags from current MakeInfo
+        :return: Dependency folder and flags
+        """
+        self._info.dirs = list(set(self._info.dirs))
+        self._info.flags = list(set(self._info.flags))
+        return self._info
```
memsharded: This way of removing duplicates can and probably will change the order of directories and flags, which is likely an undesired behavior. The defined order by users and topology of the graph must be respected.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,588 @@
...
+        return glob_content_gen.deps_content(dependencies)
+
+
+class DepComponentContentGenerator:
+    """
+    Generates Makefile content for each dependency component
+    """
+
+    template = textwrap.dedent("""\
+        {%- macro format_map_values(values) -%}
```
memsharded: When recommending to use more jinja templates, I was thinking of being able to visualize a bit better the actual contents, like in https://github.com/conan-io/conan/blob/c53f2879f5ca6d7324455c8fcf9ea4ab8e358ab9/conan/tools/cmake/cmakedeps/templates/target_data.py#L112
```cmake
set({{ pkg_name }}_INCLUDE_DIRS{{ config_suffix }} {{ global_cpp.include_paths }})
set({{ pkg_name }}_RES_DIRS{{ config_suffix }} {{ global_cpp.res_paths }})
set({{ pkg_name }}_DEFINITIONS{{ config_suffix }} {{ global_cpp.defines }})
set({{ pkg_name }}_SHARED_LINK_FLAGS{{ config_suffix }} {{ global_cpp.sharedlinkflags_list }})
```

I am not sure this template helps much in understanding the actual file contents, maybe it is because still too much abstraction over the dirs and flags.

conan/tools/gnu/makedeps.py
uilianries: It's not like that. When creating a global variables, like CONAN_LIBS, first we need to know if a dependency has libs declared.

However, there are cases when the `cpp_info.libs = []` but then we have a component with libs defined. For this case, we well have `CONAN_LIBS = $(CONAN_LIBS_FOO)` because what matters is the dependency, not its component (for global).

And there is a third case, when we have both `cpp_info.libs` defined and a component with another library. This will result that we will that same dependency duplicated in the list (because we push the variable/folder/flag for each component or dependency). 

The filter does not affect the linkage order or something fragile, but avoids duplicating the same dependency name in the variable list in `GlobalGenerator::_get_dependency_dirs` and `GlobalGenerator::_get_dependency_flags`

conan/tools/gnu/makedeps.py
uilianries: It's possible to be more explicit, but it would result in multiple ifs as we don't want to expose unused variables, otherwise the final conandeps.mk would be polluted with multiple empty variables. Let me update to a more explicit way, we still can revert in case we don't like.

uilianries: @memsharded Please, take a look again. Now variables are explicit in the template. I moved the method that formats lines, otherwise we would need to replicate 3 times. It's possible to avoid even more duplication, but I guess on this current state is more readable.

iskunk:
Thanks @uilianries. One small nit: There should be only one blank line after the package/version and "Aggregated" comments. The two blank lines serve as (visual) separators, so having two after the comment disassociates it from the information that follows.

SSE4:
> GNU Make is by far, the most popular and well documented Make implementation, that's why the choice. 

if generator is only compatible with GNU-specific Make flavour (e.g. using GNU-specific syntax or GNU extensions), then it should be named `GNUMakeDeps` or `GMakeDeps`, to avoid confusion. I suppose it will not work for other derivatives (e.g. BSDMake or NMake) without significant tweaks.

iskunk:
@SSE4, there is no need for this generator to use GNU Make-specific features; all it does is define a bunch of variables. My original implementation was portable across different Make implementations, and I've ensured it remains that way.

igagis:
I'll just repost my question from other [discussion](https://github.com/conan-io/conan/issues/13889#issuecomment-1676818985) , as it might be relevant here:

Why using custom `CONAN_` variables as it was in conan 1? Why not go with same approach as in `AutotoolsDeps`? I.e. setting well known (by `GNU make`) variables like `CXXFLAGS`, `LDFLAGS`, `LDLIBS` in the `MakeDeps` generator?

In my opinion, adding `MakeDeps` generator would be just to fork the `AutotoolsDeps` generator, removing autotools related stuff from there and add setting `LDLIBS` variable instead of `LIBS` varaible. Not sure if it that easy, but what I'd expect from `MakeDeps` generator is just to set `ASFLAGS`, `CFLAGS`, `CXXFLAGS`, `CPPFLAGS`, `LDFLAGS`, `LDLIBS` varaibles, as listed in `GNU make` docs: https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-CFLAGS. I'm not sure the generator has to set all the weird vars listed there, like `PFLAGS`...

Using these well known variables will result in that modifying `makefiles` of existing libraries will not be required at all to package them for conan

iskunk:
Hi @igagis,

There are several issues with the AutotoolsDeps approach using the `*FLAGS` variables, not least that (1) it assumes which variables are meaningful to the build system (what if you need to set `LOCAL_CXX_FLAGS` instead of `CXXFLAGS`? or pass them as a command-line arg instead of an env var?), and (2) there is no easy way for subsequent scripting to obtain flags for specific dependencies; it is all of them or nothing. The MakeDeps generator is intended to be fully general, in the way that Make itself is fully general. It sets `CONAN_*` variables to avoid collisions with ones already in use, and lets you make use of them however you like.

reedhedges:
It is useful to have separate flags variables for different Conan packages
and components. If you have a reasonably complex Makefile, you probably are
building up custom flags variables used by different targets or in
different conditions.

igagis:
Well, in addition to `CONAN_` variables it could also set `CFLAGS` etc. variables, should not harm

iskunk:
> Well, in addition to `CONAN_` variables it could also set `CFLAGS` etc. variables, should not harm

Someone else would complain that they don't want the generator assigning to such variables---either because they want to set a different value, or because they don't want these set at all. (For example, I'm using the profile `buildenv` section to set `CFLAGS`, so I *really* don't want that showing up in the generator output.)

There is a bright line between "makes my use case easy to support" and "supports my use case out of the box." This generator opts for the former, because the latter is impossible to attain without a gazillion parameters.

igagis:
As far as I understand, different conan package means different `conanfile.py`. And which flags to add to `CFLAGS` etc. is defined by the `conanfile.py`.
Packaging a library for conan means writing `conanfile.py` for it and not modifying the library's build system files.

So, anyway, I'm just suggesting. Looks like I'll continue using `AutotoolsDeps` generator for my purposes, as it works pretty well, just needs one `LDLIBS=LIBS` small workaround.

conans/test/functional/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,144 @@
...
+    client = TestClient(path_with_spaces=False)
+    with client.chdir("lib"):
+        client.save({"conanfile.py": conanfile.replace("libname", "hello")})
+        client.run("create . --name=hello --version=0.1.0 -tf=\"\"")
+    with client.chdir("test"):
+        client.save({"conanfile.py": conanfile.replace("libname", "test")})
+        client.run("create . --name=test --version=0.1.0 -tf=\"\"")
+    with client.chdir("tool"):
+        client.save({"conanfile.py": conanfile.replace("libname", "tool")})
+        client.run("create . --name=tool --version=0.1.0 -tf=\"\"")
```
memsharded:
```suggestion
                self.cpp_info.libs = [self.name]
        ''')
    client = TestClient(path_with_spaces=False)
    client.save({"conanfile.py": conanfile})
    client.run("create . --name=hello --version=0.1.0")
    client.run("create . --name=test --version=0.1.0")
    client.run("create . --name=tool --version=0.1.0")
```

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+                self.cpp_info.libdirs = [lib_dir, lib_dir2]
+        """)
+    client = TestClient()
+    client.save({"conanfile.py": conanfile})
+    client.run("create .")
+    client.run("install --requires=mylib/0.1@ -g MakeDeps")
+
+    makefile_path = os.path.join(client.current_folder, CONAN_MAKEFILE_FILENAME)
+    assert os.path.isfile(makefile_path)
+    makefile_content = load(makefile_path)
```
memsharded:
```suggestion
    makefile_content = client.load(CONAN_MAKEFILE_FILENAME)
```

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+        assert os.path.isabs(path) is True
+
+    lines = makefile_content.splitlines()
+    for line_no, line in enumerate(lines):
+        includedir_pattern = "CONAN_INCLUDE_DIRS_MYLIB = $(CONAN_INCLUDE_DIRS_FLAG)"
+        if line.startswith(includedir_pattern):
+            assert_is_abs(line[len(includedir_pattern):])
+            assert line.endswith("include")
+        elif line.startswith("\t$(CONAN_LIB_DIRS_FLAG)") and 'my_absoulte_path' in line:
+            assert_is_abs(line[len("\t$(CONAN_LIB_DIRS_FLAG)"):-2])
```
memsharded:
```suggestion
    lines = makefile_content.splitlines()
    for line_no, line in enumerate(lines):
        includedir_pattern = "CONAN_INCLUDE_DIRS_MYLIB = $(CONAN_INCLUDE_DIRS_FLAG)"
        if line.startswith(includedir_pattern):
            assert os.path.isabs(line[len(includedir_pattern):])
            assert line.endswith("include")
        elif line.startswith("\t$(CONAN_LIB_DIRS_FLAG)") and 'my_absoulte_path' in line:
            assert os.path.isabs(line[len("\t$(CONAN_LIB_DIRS_FLAG)"):-2])
```

Boiler plate

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+    makefile_path = os.path.join(client.current_folder, CONAN_MAKEFILE_FILENAME)
+    assert os.path.exists(makefile_path) is True
+    makefile_content = load(makefile_path)
+    assert 'CONAN_ROOT_MYLIB' in makefile_content
+    assert 'SYSROOT' not in makefile_content
+    assert 'CONAN_INCLUDE_DIRS' not in makefile_content
+    assert 'CONAN_LIB_DIRS' not in makefile_content
+    assert 'CONAN_BIN_DIRS' not in makefile_content
+    assert 'CONAN_LIBS' not in makefile_content
+    assert 'CONAN_FRAMEWORK_DIRS' not in makefile_content
```
memsharded, line 91: Does ``make`` and ``Makefiles`` fail if referencing a variable that doesn't exist, or just treat it as empty?

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+                self.cpp_info.frameworkdirs = []
+        """)
+    client = TestClient()
+    client.save({"conanfile.py": conanfile})
+    client.run("create .")
+    client.run("install --requires=mylib/0.1@ -g MakeDeps")
+
+    makefile_path = os.path.join(client.current_folder, CONAN_MAKEFILE_FILENAME)
+    assert os.path.exists(makefile_path) is True
+    makefile_content = load(makefile_path)
```
memsharded:
```suggestion
    makefile_content = client.load(CONAN_MAKEFILE_FILENAME)
```

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+        class TestMakeDeps(ConanFile):
+            def package_info(self):
+                self.cpp_info.libs = ["lib%s"]
+        """)
+    with client.chdir("app"):
+        client.save({"conanfile.py": conanfile % "app"})
+        client.run("create . --name=app --version=1.0")
+    with client.chdir("test"):
+        client.save({"conanfile.py": conanfile % "test"})
+        client.run("create . --name=test --version=1.0")
```
memsharded:
```suggestion
                self.cpp_info.libs = [f"lib{self.name}"]
        """)

    client.save({"conanfile.py": conanfile})
    client.run("create . --name=app --version=1.0")
    client.run("create . --name=test --version=1.0")
```

conans/test/integration/toolchains/gnu/test_makedeps.py
```diff
@@ -0,0 +1,428 @@
...
+    # Create library having build and test requires
```
memsharded, line 314: Comment doesn't match code

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+- CONAN_DEFINE_FLAG: Add a prefix to all defines variables, e.g. -D
+- CONAN_SYSTEM_LIB_FLAG: Add a prefix to all system_libs variables, e.g. -l
+- CONAN_INCLUDE_DIR_FLAG: Add a prefix to all include dirs variables, e.g. -I
+- CONAN_LIB_DIR_FLAG: Add a prefix to all lib dirs variables, e.g. -L
+- CONAN_BIN_DIR_FLAG: Add a prefix to all bin dirs variables, e.g. -L
+
+
+The conandeps.mk file layout is as follows:
+
+- CONAN_DEPS: list all transitive and intransitive dependencies names without version (e.g. zlib)
```
memsharded:
```suggestion
- CONAN_DEPS: list all transitive and direct dependencies names without version (e.g. zlib)
```

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+    """
+
+    template = textwrap.dedent("""\
+
+        # {{ dep.ref.name }}/{{ dep.ref.version }}
+
+
+        CONAN_NAME_{{ name }} = {{ dep.ref.name }}
+
+        CONAN_VERSION_{{ name }} = {{ dep.ref.version }}
```
memsharded, line 373: Why not user/channel too? Maybe a full ref ``CONAN_REF_``?

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+        """
+        context = {
+            "dep": self._dep,
+            "name": _makefy(self._dep.ref.name),
+            "root": self._root,
+            "sysroot": self._sysroot,
+            "include_dirs": self._dirs.get("include_dirs", []),
+            "lib_dirs": self._dirs.get("lib_dirs", []),
+            "bin_dirs": self._dirs.get("bin_dirs", []),
+            "src_dirs": self._dirs.get("src_dirs", []),
```
memsharded: Isn't it possible to use the ``None`` in the rendering as an empty indicator, instead of having to default to all empty lists?

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+
+    def __init__(self, conanfile: object):
+        """
+        :param conanfile: ``< ConanFile object >`` The current recipe object. Always use ``self``.
+        """
+        self._conanfile = conanfile
+
+    def generate(self) -> None:
+        """
+        Parse a ConanFile object, collecting all dependencies and components, then, generating a Makefile
```
memsharded:
```suggestion
        Collects all dependencies and components, then, generating a Makefile
```

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+
+        host_req = self._conanfile.dependencies.host
+        build_req = self._conanfile.dependencies.build  # tool_requires
+        test_req = self._conanfile.dependencies.test
+
+        content_buffer = f"{self._title}\n"
+        deps_buffer = ""
+
+        # Filter the build_requires not activated for any requirement
+        dependencies = [tup for tup in list(host_req.items()) + list(build_req.items()) + list(test_req.items()) if not tup[0].build]
```
memsharded, line 617: Just drop ``build-requires`` if they are not going to be used at all?

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+        self._info = MakeInfo(self._dep.ref.name, [], [])
+
+    def make_global_info(self) -> MakeInfo:
+        """
+        Remove duplicated folders and flags from current MakeInfo
+        :return: Dependency folder and flags
+        """
+        self._info.dirs = list(set(self._info.dirs))
+        self._info.flags = list(set(self._info.flags))
+        return self._info
```
memsharded: This is a weird pattern, with inputs and outputs all inside ``MakeInfo``, seems this should be encapsulated and automatic in ``MakeInfo``

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+            if "flags" in var:
+                cppinfo_value = [var.replace('"', '\\"') for var in cppinfo_value]
+            if cppinfo_value:
+                self._info.flags.append(var)
+                flags[var] = [_conan_prefix_flag(prefix_var) + it for it in cppinfo_value]
+        return flags
+
+    def _get_sysroot(self, root: str) -> list:
+        """
+        Get the sysroot of the dependency. Sysroot is a list of directories, or a single directory
```
memsharded, line 556: So far we haven't treated ``sysroot`` as multiple directories, maybe this would be better to align with others and treat it jointly with other generators behavior.

conan/tools/gnu/makedeps.py
```diff
@@ -0,0 +1,642 @@
...
+    def _get_dependency_flags(self, dependency: object) -> dict:
+        """
+        List common variables from cpp_info and format them to be used in the makefile
+        :param dep: Dependency object
+        """
+        flags = {}
+        for var, prefix_var in _common_cppinfo_variables().items():
+            cppinfo_value = getattr(dependency.cpp_info, var)
+            # Use component cpp_info info when does not provide any value
+            if not cppinfo_value and hasattr(dependency.cpp_info, "components"):
```
memsharded: The ``cpp_info`` always has ``components`` defined, it is in the constructor of ``CppInfo``, this ``hasattr`` check is useless?

conans/test/integration/toolchains/gnu/test_makedeps.py
uilianries: Just treats as empty, for instance:

    $(CC) $(TORTILLA) main.c -o main
    
In case `TORTILLA` is not declared, but still, `make` will ignore it and use empty as replacement.

conans/test/integration/toolchains/gnu/test_makedeps.py
uilianries: fixed, testing both test_requires and build_requires together in the commit 988a3192e5c22e8c556404a957f585553b8234fe

conan/tools/gnu/makedeps.py
uilianries: Added `CONAN_REFERENCE_{{ name }}` and keeping `CONAN_NAME_` and `CONAN_VERSION_`. Commit a1bcfa89c42188f24ce81cb77b3c0d57de140736

conan/tools/gnu/makedeps.py
uilianries: Done in the commit c5cd0a5bb

conan/tools/gnu/makedeps.py
uilianries: Yes, I just simplified it on the commit 2f0b565bc58ceb99d35bd0c5136193f2848aec73. Thank you for catching it.

conan/tools/gnu/makedeps.py
uilianries: Doing some tests, it seems be. The idea here is capturing cpp_info value from a component, when a dependency has no such cpp_info value. It's available on the commit 614c2731e

conan/tools/gnu/makedeps.py
uilianries: Exactly. We could add another feature, just like for `pkgconfigdeps`, the `build_context_activated`, a list with reference that should be part of the makefile.

conan/tools/gnu/makedeps.py
uilianries: Indeed is not usual having multiple sysroot, but only one when cross-building. TBH, it seems be a bug, because returning `['']` seems be something mixed, it should be empty string, but then was converted to a list to keep same pattern as other cpp_info dirs

iskunk:
Thanks @memsharded, @uilianries. This has been a long time in coming.

I've pushed a PR (#14605) with some minor tweaks on top of this, to polish it off.

PR #14161, opened 2023-06-23
removing some legacy and duplications

memsharded:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conans/client/conanfile/configure.py
+++ b/conans/client/conanfile/configure.py
@@ -1,44 +1,40 @@
 from conans.errors import conanfile_exception_formatter
 from conans.model.pkg_type import PackageType
 from conans.model.requires import BuildRequirements, TestRequirements, ToolRequirements
-from conans.util.env import no_op
 
 
 def run_configure_method(conanfile, down_options, profile_options, ref):
     """ Run all the config-related functions for the given conanfile object """
 
-    # Avoid extra time manipulating the sys.path for python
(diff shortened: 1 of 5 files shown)


conans/client/conanfile/configure.py
```diff
@@ -1,44 +1,40 @@
...
 from conans.model.pkg_type import PackageType
 from conans.model.requires import BuildRequirements, TestRequirements, ToolRequirements
-from conans.util.env import no_op
 
 
 def run_configure_method(conanfile, down_options, profile_options, ref):
     """ Run all the config-related functions for the given conanfile object """
 
-    # Avoid extra time manipulating the sys.path for python
-    with no_op():  # TODO: Remove this in a later refactor
```
memsharded, line 11: All of this change is just removing this ``no_op()`` context, that as it says, does nothing.

conans/model/requires.py
```diff
@@ -276,20 +276,7 @@ def transform_downstream(self, pkg_type, require, dep_pkg_type):
             return
 
         # Regular and test requires
-        if dep_pkg_type is PackageType.SHARED:
-            if pkg_type is PackageType.SHARED:
```
memsharded, line 280: The implementation was duplicated for both SHARED and STATIC, as the traits account for it already.

AbrilRBS (approved):

PR #14164, opened 2023-06-23
Don't use realpath() to determine cache location (#14136)

### LINKED ISSUE #14136: [bug] Conan mishandles cache path in symlinked homedir
### Environment details

* Operating System+version: Red Hat Enterprise Linux 7
* Conan version: 2.0.6
* Python version: 3.8.3

### Steps to reproduce

At my site, home directories are mounted on a typical Unix system using a combination of an automounter and symlinks. For example, given a user `robert`, the automounter will mount his homedir at a path like
```
/nfs/home14.internal.example.com/vol/vol374/r/ro/robert
```
but then for easier access, it will also create a symlink:
```
/user/robert -> /nfs/home14.internal.example.com/vol/vol374/r/ro/robert
```

At present, even though `HOME` is set to `/user/robert`, Conan will find the ugly long NFS path and use it for all cache operations. This is undesirable, especially when the long path gets baked into generated files and symlinks.

----

After some digging, I found that the culprit is an `os.path.realpath()` call at the top of `class DataCache`:

https://github.com/conan-io/conan/blob/81b195262350a13525a44a0b079c9a4eab9fb1cd/conan/internal/cache/cache.py#L20

Changing that to `os.path.abspath()` is enough to allow Conan to use the short path ubiquitously.

I believe `abspath` is in the spirit of what was originally intended with `realpath`, and if there is no objection, I'd like to submit a PR that makes this change.

### Logs

_No response_

iskunk:
Changelog: Fix: Respect symlinked path for cache location.
Docs: Omit

Close #14136

### DIFF
--- a/conan/internal/cache/cache.py
+++ b/conan/internal/cache/cache.py
@@ -17,7 +17,7 @@ from conans.util.files import rmdir, renamedir
 class DataCache:
 
     def __init__(self, base_folder, db_filename):
-        self._base_folder = os.path.realpath(base_folder)
+        self._base_folder = os.path.abspath(base_folder)
         self._db = CacheDatabase(filename=db_filename)
 
     def _create_path(self, relative_path, remove_contents=True):
--- a/conan/internal/cache/cache.py
+++ b/conan/internal/cache/cache.py
@@ -30,6 +30,8 @@ class DataCache:
         rmdir(self._full_path(relative_path))
 
     def _full_path(self, relative_path):
+        # This one is used only for rmdir and mkdir operations, not returned to user
+        # or stored in DB
         path = os.path.realpath(os.path.join(self._base_folder, relative_path))
         return path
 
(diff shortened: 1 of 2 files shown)


memsharded (commented):
This sounds good, I think this can be moved forward.

It would need a "integration" test that covers this case, to make sure this is working and won't be a problem. This test needs to define the ``symlink`` scenario that the issue described, and create and use a dummy package from there, generate some file like ``--format=json`` and check the paths.

Do you think that you will be able to contribute such test? Please let us know, we can help or we can even contribute the test ourselves. Thanks for your contribution!

iskunk:
Thanks @memsharded. I doubt I'd be able to write an acceptable test for this from scratch within a reasonable timeframe, but might there be an existing test that is already pretty close to what is needed? Others I could crib relevant bits from? I should be able to mix and match.

memsharded:
Don't worry about it, I have added a test in https://github.com/conan-io/conan/pull/14164/commits/c0104a2dc551bc3a772b0b7a9e02662b1b9ea168, that should be enough

iskunk:
Much obliged, good sir! It would have taken me quite a while to write a test that succinct :-]

PR #14167, opened 2023-06-25
Fixes #14166: Add strip option to cmake.install tool

### LINKED ISSUE #14166: [feature] Add '--strip' option to cmake install
### What is your suggestion?

Can you add a parameter to `cmake.install` function in order to be able to strip binaries before install ?

install and strip were added to CMake about 4 years ago, and if I got it correctly in version 3.15
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/3069

Strip is very useful to remove debug symbols (actually all DWARF extension) from binaries and especially release binaries.

I tried many options to make it work with conan, but without success, so I think adding this option as a parameter to `install` function in `cmake` tool will be the best solution instead of some workarounds.

sagi-ottopia:
Changelog: Feature: Add new ``tools.cmake:install_strip`` conf to add ``--strip`` option to cmake install.
Docs: https://github.com/conan-io/docs/pull/3281

Fix: #14166

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -133,7 +133,7 @@ class CMake(object):
     def build(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
         self._build(build_type, target, cli_args, build_tool_args)
 
-    def install(self, build_type=None, component=None, strip=False):
+    def install(self, build_type=None, component=None):
         mkdir(self._conanfile, self._conanfile.package_folder)
 
         bt = build_type or self._conanfile.settings.get_safe("build_type")
@@ -147,7 +147,9 @@ class CMake(object):
         arg_list = ["--install", build_folder, build_config, "--prefix", pkg_folder]
         if component:
             arg_list.extend(["--component", component])
-        if strip:
+
+        do_strip = self._conanfile.conf.get("tools.cmake.install:strip", check_type=bool)
+        if do_strip:
             arg_list.append("--strip")
         arg_list = " ".join(filter(None, arg_list))
         command = "%s %s" % (self._cmake_program, arg_list)
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -133,7 +133,7 @@ class CMake(object):
     def build(self, build_type=None, target=None, cli_args=None, build_tool_args=None):
         self._build(build_type, target, cli_args, build_tool_args)
 
-    def install(self, build_type=None, component=None):
+    def install(self, build_type=None, component=None, strip=False):
         mkdir(self._conanfile, self._conanfile.package_folder)
 
         bt = build_type or self._conanfile.settings.get_safe("build_type")
@@ -147,6 +147,8 @@ class CMake(object):
         arg_list = ["--install", build_folder, build_config, "--prefix", pkg_folder]
         if component:
             arg_list.extend(["--component", component])
+        if strip:
+            arg_list.append("--strip")
         arg_list = " ".join(filter(None, arg_list))
         command = "%s %s" % (self._cmake_program, arg_list)
         self._conanfile.output.info("CMake command: %s" % command)
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -148,7 +148,7 @@ class CMake(object):
         if component:
             arg_list.extend(["--component", component])
 
-        do_strip = self._conanfile.conf.get("tools.cmake.install:strip", check_type=bool)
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Thanks for your contribution.

as the ``CMake`` is a wrapper around the command line I see no strong reason against this feature and this PR. However, this approach requires modifying recipes, maybe it would make sense to have a ``conf`` like ``tools.cmake.install:strip=xxx`` that controls that behavior, so it is possible to do it at scale for the whole dependency graph?

For this PR, it would be great if you could add some unit test that covers the behavior. I think that a ``unittest`` could be enough for this one, please have a look, and if you need help, let us know, we can contribute the test. Thanks again!

sagi-ottopia:
Hi,
Regarding a `conf` value, I think it should be easy to add it to the PR as well. I will look for some example and will update the PR accordingly.

Regarding unittests, sure.

memsharded:
> Regarding a conf value, I think it should be easy to add it to the PR as well. I will look for some example and will update the PR accordingly.

Only if it makes sense as a use case, is it something that you will prefer over modifying the recipes? Is it something that makes sense to do for all packages in the graph or is it something specific to some?

I am adding @jcar87 for some feedback about this use case too.

sagi-ottopia:
> > Regarding a conf value, I think it should be easy to add it to the PR as well. I will look for some example and will update the PR accordingly.
> 
> Only if it makes sense as a use case, is it something that you will prefer over modifying the recipes? Is it something that makes sense to do for all packages in the graph or is it something specific to some?
> 
> I am adding @jcar87 for some feedback about this use case too.

I think it makes sense, 
because when using the a conf value I can add it just in my Release builds profile without the need to have something like the following in my `conanfile.py`: 
```
do_strip = False
if self.settings.build_type == 'Release':
   do_strip = True
...
cmake.install(strip=do_strip)
```

memsharded:
> if self.settings.build_type == 'Release':

Ok, makes sense, though nowadays, I'd expect that ``Release`` builds already don't contain any debug information to be stripped, I understand it might not be the case for all platforms and tools, so I am ok with making this a opt-in conf.

conan/tools/cmake/cmake.py
```diff
@@ -147,7 +147,9 @@ def install(self, build_type=None, component=None, strip=False):
         arg_list = ["--install", build_folder, build_config, "--prefix", pkg_folder]
         if component:
             arg_list.extend(["--component", component])
-        if strip:
+
+        do_strip = self._conanfile.conf.get("tools.cmake.install:strip", check_type=bool)
```
```suggestion
        do_strip = self._conanfile.conf.get("tools.cmake:install_strip", check_type=bool)
```
It would be a bit more aligned with existing configs ``tools.cmake:cmake_program``

conans/test/unittests/tools/cmake/test_cmake_install.py
```diff
@@ -31,9 +31,38 @@ def test_run_install_component():
...
+    settings.os = "Linux"
+    settings.arch = "x86_64"
+    settings.build_type = "Release"
+    settings.compiler = "gcc"
+    settings.compiler.version = "11"
+
+    conanfile = ConanFileMock()
+
+    conanfile.conf = Conf()
+    conanfile.conf["tools.cmake.install:strip"] = True
```
memsharded:
```suggestion
    conanfile.conf["tools.cmake:install_strip"] = True
```

memsharded (approved):
Looks great, many thanks!

sagi-ottopia:
Hi @memsharded,
saw that you added this into 2.0.8 timeline, 
is it possible to merge it also to 1.XX  latest version ? 
Thanks.

memsharded:
Yes, sorry, didn't check the target branch. Moved it to 1.61, and then I will merge it myself to 2.0 branch (2.0.8 will probably be out earlier than 1.61)

jcar87 (approved):
> > if self.settings.build_type == 'Release':
> 
> Ok, makes sense, though nowadays, I'd expect that `Release` builds already don't contain any debug information to be stripped, I understand it might not be the case for all platforms and tools, so I am ok with making this a opt-in conf.

From https://discourse.cmake.org/t/what-does-strip-do/4302 and if I'm not mistaken, it's not only debug information that gets stripped (if there is any), but also symbols that don't need to be externally visible, so this should have a benefit.

PR looks good to me, thanks @sagi-ottopia  for your contribution!!!

@memsharded  one thing I'd consider outside of this PR is adding a `cli_args` argument to `cmake.install()` - to mirror what we have for `cmake.configure()` and `cmake.build()` - while the current setup for install pretty much covers the bases, it may be useful for some advanced users to have the added flexibility

PR #14168, opened 2023-06-26
Fix/XcodeDeps enable custom configurations (#14149)

### LINKED ISSUE #14149: [question][bug] XcodeDeps with custom configuration does not behave as expected
### What is your question?

Hi,

I'm trying to understand the example from the [XcodeDeps tutorial with custom configurations](https://docs.conan.io/2/reference/tools/apple/xcodedeps.html#custom-configurations).

I've added the `conanfile.py` from the example and ran
```
> conan profile show
Host profile:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu20
compiler.libcxx=libc++
compiler.version=14
os=Macos

Build profile:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu20
compiler.libcxx=libc++
compiler.version=14
os=Macos

> conan install . --output-folder=build --build=missing -o 'shared=True'
```
 I was expecting that the config specifier in the resulting xcconfig files would now reflect the adapted configuration name, e.g. `ReleaseShared` - I thought that was the whole point.
But when I look inside the resulting `conan_zlib_zlib_release_arm64.xcconfig` then it's just `HEADER_SEARCH_PATHS_zlib_zlib[config=Release][arch=arm64][sdk=macosx*] ...`.

Is my understanding wrong?

We have several differently named Debug and Release configurations, e.g. "Release with asserts" and for that I need an xcconfig for all of them.

wAuner:
Changelog: Fix: Use ``configuration`` in ``XcodeDeps`` instead of always ``build_type``.
Docs: Omit

fixes #14149 

* adds support for Xcode custom configurations as it is already [documented](https://docs.conan.io/2/reference/tools/apple/xcodedeps.html#custom-configurations) but not actually implemented
* adds unit test

### DIFF
--- a/conans/test/integration/toolchains/apple/test_xcodedeps.py
+++ b/conans/test/integration/toolchains/apple/test_xcodedeps.py
@@ -40,11 +40,11 @@ def expected_files(current_folder, configuration, architecture, sdk_version):
     return files
 
 
-def check_contents(client, deps, configuration, architecture, sdk_version):
+def check_contents(client, deps, build_type, architecture, sdk_version, custom_config_name=None):
     for dep_name in deps:
         dep_xconfig = client.load("conan_{dep}_{dep}.xcconfig".format(dep=dep_name))
         conf_name = "conan_{}_{}{}.xcconfig".format(dep_name, dep_name,
-                                                 _get_filename(configuration, architecture, sdk_version))
+                                                 _get_filename(build_type, architecture, sdk_version))
 
         assert '#include "{}"'.format(conf_name) in dep_xconfig
         for var in _expected_dep_xconfig:
@@ -53,7 +53,8 @@ def check_contents(client, deps, configuration, architecture, sdk_version):
 
         conan_conf = client.load(conf_name)
         for var in _expected_conf_xconfig:
-            assert var.format(name=dep_name, configuration=configuration, architecture=architecture,
+            xcode_config_name = custom_config_name if custom_config_name else build_type
+            assert var.format(name=dep_name, configuration=xcode_config_name, architecture=architecture,
                               sdk="macosx", sdk_version=sdk_version) in conan_conf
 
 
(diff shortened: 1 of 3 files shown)


conans/test/integration/toolchains/apple/test_xcodedeps.py
```diff
@@ -40,11 +40,11 @@ def expected_files(current_folder, configuration, architecture, sdk_version):
     return files
 
 
-def check_contents(client, deps, configuration, architecture, sdk_version):
+def check_contents(client, deps, build_type, architecture, sdk_version, custom_config_name=None):
     for dep_name in deps:
         dep_xconfig = client.load("conan_{dep}_{dep}.xcconfig".format(dep=dep_name))
         conf_name = "conan_{}_{}{}.xcconfig".format(dep_name, dep_name,
-                                                 _get_filename(configuration, architecture, sdk_version))
+                                                 _get_filename(build_type, architecture, sdk_version))
```
memsharded: I have a question here: why the filename is not using the ``configuration`` instead of the build_type? because generated files should use different names for different configurations, to avoid collisitions, like a ``conan_releaseshared_...xconfig`` and ``conan_releasestatic_...xconfig``?

memsharded (commented):
Thanks for your contribution!
Can you please have a look to the comment?

conans/test/integration/toolchains/apple/test_xcodedeps.py
wAuner: That's a good question. My rationale for this was the following: the classic build type variants (Debug/Release) are defined by the dependency author (including preprocessor variables like `NDEBUG` etc) while the custom Xcode configuration name is something internal to the consuming Xcode project. So the dependency will not be compiled with the custom configuration.
Does that make sense, or should the name contain the custom configuration name? I don't know which would be better. Assuming my understanding isn't wrong (which is totally possible), I guess one could argue for both.

memsharded: Not really. This is the code of the CMakeDeps:
```python
 def filename(self):
        data_fname = "" if not self.generating_module else "module-"
        data_fname += "{}-{}".format(self.file_name, self.configuration.lower())
```
So it is using the configuration as the filename.

The reason is that this is the configuration for the consumer. Lets say that the consumer has:
```python
def requirements(self):
     if self.settings.build_type == "Release":
          self.requires("mydep/1.0")
    else:
          self.requires("mydep/1.2")
```
And ``mydep`` is a header-only library, not depending on the build_type at all.
The ``mydep_release.cmake`` file will be pointing to the ``mydep/1.0``, while the ``mydep_debug.cmake`` will be pointing to the ``mydep/1.2``. It is not the build_type of the dependency the one that is relevant here, but the ``configuration`` that is being built downstream.

wAuner: Sure, it makes sense. I'll make the necessary changes.

wAuner: I've implemented the configuration-based naming convention. Could you have another look? Let me know if there's anything else I should change. 

BTW is it normal that many tests are failing locally (not specific to my branch, happens also on develop/2)? I can't get a completely successful test run on my machine. I'm using a python 3.11 venv and I'm setting the python path as described.

memsharded: There is a lot of configuration in the ``conftest.py`` file, to be overriden by ``conftest_user.py`` to configure many external tools in order to be able to pass all tests.

Don't worry much about that, typically during development running the relevant tests only is enough, and CI will do the rest.

czoido (approved):

memsharded:
This was merged to ``develop2``, I will merge it to ``release/2.0`` so it is in next 2.0.8 too

PR #14169, opened 2023-06-26
trim_conandata()

memsharded:
Changelog: Feature: New ``trim_conandata()`` function to make things explicit, not depend on hook.
Docs: https://github.com/conan-io/docs/pull/3373

### DIFF
--- a/conan/tools/files/conandata.py
+++ b/conan/tools/files/conandata.py
@@ -31,3 +31,31 @@ def update_conandata(conanfile, data):
     recursive_dict_update(conandata, data)
     new_content = yaml.safe_dump(conandata)
     save(path, new_content)
+
+
+def trim_conandata(conanfile):
+    """
+    Tool to modify the ``conandata.yml`` once it is exported, to limit it to the current version
+    only
+    """
+    if not hasattr(conanfile, "export_folder") or conanfile.export_folder is None:
+        raise ConanException("The 'trim_conandata()' can only be used in the 'export()' method")
+    path = os.path.join(conanfile.export_folder, "conandata.yml")
+    if not os.path.exists(path):
+        raise ConanException("conandata.yml file doesn't exist")
+
+    conandata = load(path)
+    conandata = yaml.safe_load(conandata)
+
+    version = str(conanfile.version)
+    result = {}
+    for k, v in conandata.items():
+        if not isinstance(v, dict):
+            result[k] = v
+            continue  # to allow user extra conandata, common to all versions
+        version_data = v.get(version)
+        if version_data is not None:
+            result[k] = {version: version_data}
+
+    new_conandata_yml = yaml.safe_dump(result, default_flow_style=False)
+    save(path, new_conandata_yml)
--- a/conan/tools/files/__init__.py
+++ b/conan/tools/files/__init__.py
@@ -7,4 +7,4 @@ from conan.tools.files.cpp_package import CppPackage
 from conan.tools.files.packager import AutoPackager
 from conan.tools.files.symlinks import symlinks
 from conan.tools.files.copy_pattern import copy
-from conan.tools.files.conandata import update_conandata
+from conan.tools.files.conandata import update_conandata, trim_conandata
--- a/conans/test/integration/conanfile/conan_data_test.py
+++ b/conans/test/integration/conanfile/conan_data_test.py
@@ -331,3 +331,53 @@ class TestConanDataUpdate:
         c.save({"conanfile.py": conanfile})
         c.run("export .")  # It doesn't fail
         assert "pkg/0.1: Calling export()" in c.out
+
+
+def test_conandata_trim():
+    """ test the explict trim_conandata() helper
+    """
+    c = TestClient()
+    conanfile = textwrap.dedent("""
+        from conan import ConanFile
+        from conan.tools.files import trim_conandata
+
+        class Pkg(ConanFile):
+            name = "pkg"
+            def export(self):
+                trim_conandata(self)
+        """)
+    conandata_yml = textwrap.dedent("""\
+        sources:
+          "1.0":
+            url: "url1"
+            sha256: "sha1"
+        patches:
+          "1.0":
+            - patch_file: "patches/some_patch"
+              base_path: "source_subfolder"
+        something: else
+          """)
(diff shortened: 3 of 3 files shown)


conan/tools/files/conandata.py
```diff
@@ -31,3 +31,31 @@ def recursive_dict_update(d, u):
...
+    for k, v in conandata.items():
+        if not isinstance(v, dict):
+            result[k] = v
+            continue  # to allow user extra conandata, common to all versions
+        version_data = v.get(version)
+        if version_data is not None:
+            result[k] = {version: version_data}
+
+    new_conandata_yml = yaml.safe_dump(result, default_flow_style=False)
+    save(path, new_conandata_yml)
```
SSE4, line 61: I think it's easy to lose all the local modifications this way without any ability to restore.
at minimum, it should create a backup of an old conandata file.
but ultimately, I would prefer if it accepted `conandata_in` and `conandata_out` as arguments, so recipe immutability principle is not violated here.

memsharded: I don't know what you are talking about.
It is not possible to create any other file than the stripped/trimmed ``conandata.yml``, because that would break the immutability of recipe revisions, which is exactly the goal of this feature.
This is not new functionality, but moving what the hook is already doing to making it explicit in the recipes (it is biting too many users that are forking conan-center-index)

SSE4: > I don't know what you are talking about.

see https://github.com/conan-io/tribe/pull/16, it says `Recipes and packages will always be immutable, and will not be able to change` and `Packages and recipes, identified by their complete reference including the revision, all artifact bare file contents (conanfile.py, source files, compiled libraries and executables, etc) stored in the Conan cache and in servers are always immutable.`. in other words, once `conandata.yml` is stored in cache, it is immutable.

> It is not possible to create any other file than the stripped/trimmed conandata.yml

it is possible by simply doing something like `save(path + ".2", new_conandata_yml)`. conan recipe is turing-complete and can store files anywhere in filesystem if permissions allows (e.g. in temp directory).

> because that would break the immutability of recipe revisions

in this PR `conandata.yml` is modified in place. modification is mutation by definition. computed recipe revision is different in this case.
specifically, the exactly same recipe will result in different revisions per different version specified on command line.

> This is not new functionality, but moving what the hook is already doing

it might be done in hook or from the inside client itself, still any modification of `conandata.yml` leads to a new recipe revision. moving the same code to the different place doesn't change that.

AbrilRBS: Hi @SSE4 thanks for your input, I just wanted to clarify one point:

> I think it's easy to lose all the local modifications this way without any ability to restore.

and
> in this PR conandata.yml is modified in place

are not "right". See the PR linked test, where it shows how this is going to be used in the export method (Way before any manifest is created), and that it runs in the export folder, so no actual user file is overwritten, just the one that gets _exported_, so the user won't see any difference :)

memsharded: I am afraid this is not how it works, and https://github.com/conan-io/tribe/pull/16 is not related at all.
Whatever executes in ``export()`` happens **before** the recipe is fully created in the cache, **before** the ``conanmanifest.txt`` is computed and **before** the recipe revision is computed. It cannot break immutability because there is still nothing to be immutable, and the ``conandata.yml`` hasn't been stored yet.

SSE4: @RubenRBS 
> not right

if you take a look at changes:
```
    conandata = load(path)
    conandata = yaml.safe_load(conandata)
    ...
    save(path, new_conandata_yml)
```
it reads and write file from the same path. it means the file (`conandata.yml`) is modified in place.
I am not talking about manifest (it's irrelevant here), I don't see how manifest is connected at all.
what I mean is there is a file in conan cache (in `export_folder` specifically, which is obviously a part of a conan cache) named `conandata.yml` and it's modified inside the cache (as code does load/save on a file within cache).
so what I mean, is `conandata.yml`is stored in cache, and then `trim_conandata` modifies the file that is already stored in cache prior to the execution of the [export](https://docs.conan.io/2/reference/conanfile/methods/export.html?highlight=export_folder) method.
that violates tribe/16 proposal as it clearly says `Packages and recipes, identified by their complete reference including the revision, all artifact bare file contents (conanfile.py, source files, compiled libraries and executables, etc) stored in the Conan cache and in servers are always immutable.`.
it reads as once a file (e.g. `conandata.yml`) is stored within cache (e.g. `export_folder`) it's immutable. it cannot be mutated (modified) after it was already stored.
clearly, no file modification can happen on a file stored in `export_folder`.
`The immutability applies to the actual file contents: the bytes stored in the different raw files: conanfile.py, C++ sources and headers, compiled libraries (.lib, .a, .so), executables (.exe), etc.` - it means byte contents of `conandata.yml` cannot change.
so my suggestion is to keep immutability and do something like:
```
   conandata = load(path)
   conandata = yaml.safe_load(conandata)
   ...
   self.new_conandata = tempfile.NamedTemporaryFile().name
   save(self.new_conandata, new_conandata_yml)
```
and then just read from `self.new_conandata` (e.g. from [apply_conandata_patches](https://docs.conan.io/2/reference/tools/files/patches.html?highlight=apply%20conandata#conan.tools.files.patches.apply_conandata_patches) or whenever else needed).

> and the conandata.yml hasn't been stored yet.

it's stored in `self.export_folder` (which is in conan cache) at the beginning of `trim_conandata` call, and `load` reads stored file (otherwise, it wouldn't work here and raised for "file doesn't exist" - it uses `path = os.path.join(conanfile.export_folder, "conandata.yml")`).

memsharded: > that violates tribe/16 proposal as it clearly says 

No, it clearly doesnt. You just cited: ``Packages and recipes, identified by their complete reference including the revision, ``

The revision cannot be computed until the ``conanmanifest.txt`` is computed. And that is only computed after ``export()`` is executed. So the ``conanmanifest.txt`` is completely relevant here. 

It doesn't mean that every folder in the cache cannot be modified, otherwise it would be impossible to apply patches or build things, because the cache folders contents cannot be modified

AbrilRBS (approved):

czoido (approved):

PR #14183, opened 2023-06-28
trying abs path for config install

### LINKED ISSUE #14178: [bug] `conan config install` does not work on mounted network drives on Windows
### Environment details

* Compiler+version: any
* Conan version: 1.59.0
* Python version: 3.11.2

### Steps to reproduce

1. Run the following as Administrator in an elevated prompt to create a share:
```
mkdir C:\tmp-conan-share
net share tmp-conan-share=C:\tmp-conan-share /GRANT:EVERYONE,FULL
```
2. Run the following in your user's prompt to mount the share as a drive and try to run Conan on it. You can use arbitrary config.
```
net use U: \\127.0.0.1\tmp-conan-share
set CONAN_HOME=U:\.conan
set CONAN_USER_HOME=U:\.conan2
conan config install https://github.com/conan-io/conanclientcert.git
```
3. Observe that Conan fails very early on, in a way very similar to one described in #6173. However, here it involves real network drives, not just a simple `subst` call, so it still does not work.
4. Unmount the share:
```
 net use U: /DELETE 
```
5. Remove the share by running the following commands in an elevated command line:
```
net share tmp-conan-share /delete
del /F C:\tmp-conan-share
```

I believe the underlying reason is that Python's `subprocess.Popen` does not work on network shares with `shell=True` because `cmd.exe` itself does not support it. Moreover, mounting network share as a drive does not help because Conan decides to resolve the path for the temporary directory and ends up on the network share nonetheless: https://github.com/conan-io/conan/blob/b43eb83956f053a47cc3897cfdd57b9da13a16e6/conans/client/conf/config_installer.py#L50

If I comment out this line, at least `conan config install` completes successfully, but I wouldn't be surprised if there are similar troubles in other places e.g. in caches or build operations.

### Logs

```
C:\Users\egor.suvorov>conan config install https://github.com/conan-io/conanclientcert.git
Trying to clone repo: https://github.com/conan-io/conanclientcert.git
ERROR: Failed conan config install: Can't clone repo: '\\127.0.0.1\tmp-conan-share\.conan\tmp_config_install'
CMD.EXE was started with the above path as the current directory.
UNC paths are not supported.  Defaulting to Windows directory.
fatal: destination path '.' already exists and is not an empty directory.
```

memsharded:
Changelog: Fix: Use ``abspath`` for ``conan config install`` to avoid symlinks issues.
Docs: Omit

This will probably FAIL in OSX, it needs to be the ``realpath``.

Close https://github.com/conan-io/conan/issues/14178


AbrilRBS (approved):
This works good on my end when installing from `/var/...`, both locally and after modifying the test

PR #14190, opened 2023-06-29
Fix --format=html output for graphs with conflict errors

AbrilRBS:
Changelog: Bugfix: Improve display of graph conflicts in `conan graph info` in html format.
Docs: Omit

@czoido found a case (from which a simplified version is now in the tests) where the graph was not properly printed with conflicts. This PR aims to fix that, and to make it more robust against the rest of possible configurations

### DIFF
--- a/conans/test/integration/command/info/test_graph_info_graphical.py
+++ b/conans/test/integration/command/info/test_graph_info_graphical.py
@@ -190,4 +190,22 @@ def test_graph_info_html_output():
     tc.run("new cmake_lib -d name=math -d version=1.0 -d requires=lib/2.0 -f")
     tc.run("export .")
     tc.run("graph info --requires=math/1.0 --requires=ui/1.0 --format=html", assert_error=True)
-    assert "// Add some helpful visualizations for conflicts" in tc.out
+    assert "// Add error conflict node" in tc.out
+    assert "// Add edge from node that introduces the conflict to the new error node" in tc.out
+    assert "// Add edge from base node to the new error node" not in tc.out
+    assert "// Add edge from previous node that already had conflicting dependency" in tc.out
+
+    tc.run("new cmake_lib -d name=libiconv -d version=1.0 -d requires=lib/2.0 -f")
+    tc.run("export .")
+    tc.run("new cmake_lib -d name=boost -d version=1.0 -d requires=libiconv/1.0 -f")
+    tc.run("export .")
+    tc.run("new cmake_lib -d name=openimageio -d version=1.0 -d requires=boost/1.0 -d requires=lib/1.0 -f")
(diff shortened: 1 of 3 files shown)


conans/test/integration/command/info/test_graph_info_graphical.py
```diff
@@ -190,4 +190,22 @@ def test_graph_info_html_output():
...
+    assert "// Add error conflict node" in tc.out
+    assert "// Add edge from node that introduces the conflict to the new error node" in tc.out
+    assert "// Add edge from base node to the new error node" not in tc.out
+    assert "// Add edge from previous node that already had conflicting dependency" in tc.out
+
+    tc.run("new cmake_lib -d name=libiconv -d version=1.0 -d requires=lib/2.0 -f")
+    tc.run("export .")
+    tc.run("new cmake_lib -d name=boost -d version=1.0 -d requires=libiconv/1.0 -f")
+    tc.run("export .")
+    tc.run("new cmake_lib -d name=openimageio -d version=1.0 -d requires=boost/1.0 -d requires=lib/1.0 -f")
```
memsharded:
```suggestion
    tc.save({"conanfile.py": GenConanfile("openimageio", "1.0").with_requires("boost/1.0", "lib/1.0")
```

We have helpers to create conanfiles easily without needing the overhead of calling a full command (and can be used for ``create`` without firing a build.)

conans/test/integration/command/info/test_graph_info_graphical.py
```diff
@@ -190,4 +190,22 @@ def test_graph_info_html_output():
...
+    tc.run("export .")
+    tc.run("graph info . --format=html", assert_error=True)
+    assert "// Add error conflict node" in tc.out
+    assert "// Add edge from node that introduces the conflict to the new error node" in tc.out
+    assert "// Add edge from base node to the new error node" in tc.out
+    assert "// Add edge from previous node that already had conflicting dependency" not in tc.out
+    # There used to be a few bugs with weird graphs, check for regressions
+    assert "jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'context'" not in tc.out
```
memsharded: This check can be easily a false negative even if it happens. Reduce it to reduce false negatives.

```suggestion
    assert "jinja2.exceptions" not in tc.out
```

memsharded (approved):

conans/test/integration/command/info/test_graph_info_graphical.py
franramirez688: Yeah, I think it's worth changing all those `conan new` by this approach.

franramirez688 (approved):

czoido (approved):

PR #14194, opened 2023-06-29
[feature] Manage shared, fPIC and header_only options automatically

danimtb:
Changelog: Feature: Manage shared, fPIC, and header_only options automatically.
Changelog: Feature: Manage package ID for header-library package type automatically.
Docs: https://github.com/conan-io/docs/pull/3296


uilianries:
Is it possible to add a new recipe attribute later, only working with Conan 2.x, that could be a shortcut to those new methods?

conans/client/conanfile/configure.py
```diff
@@ -9,6 +9,8 @@ def run_configure_method(conanfile, down_options, profile_options, ref):
     if hasattr(conanfile, "config_options"):
         with conanfile_exception_formatter(conanfile, "config_options"):
             conanfile.config_options()
+    else:
+        conanfile._conan_config_options_default()
```
AbrilRBS: I'm not the biggest fan of these being private members of conanfiles that just call the tools, and would rather just call the tools in here and not have those private ones at all, but also not a blocker for me, happy to hear everyone's thoughts :)

AbrilRBS (commented):
I have a comment about the Conan private methods, to see what you all think, but otherwise it is looking great so far :)

conans/client/conanfile/configure.py
uilianries: The protected method has the advantage when using an IDE with autocomplete, that you probably will not use the method in your recipe. But don't see problem by using public method here, since it's just a tool and should not violate the design access. Anyway, python allows calling protected methods just like public, it will complain, but will not break.

conan/tools/options/helpers.py
```diff
@@ -0,0 +1,17 @@
...
+def handle_common_configure_options(conanfile):
+    if conanfile.options.get_safe("header_only"):
+        conanfile.options.rm_safe("fPIC")
+        conanfile.options.rm_safe("shared")
+    elif conanfile.options.get_safe("shared"):
+        conanfile.options.rm_safe("fPIC")
+
+
+def handle_common_package_id_options(conanfile):
+    if conanfile.options.get_safe("header_only"):
```
memsharded: This can be done to ``package_type == HEADER`` instead, so it covers that case too (I'd say it is expected)

conans/client/conanfile/configure.py
memsharded: Yes, I think it would be better to not make it a private method of ``ConanFile``, but directly call the ``handle`` method instead.

conans/model/conan_file.py
```diff
@@ -341,3 +341,15 @@ def set_deploy_folder(self, deploy_folder):
...
+        from conan.tools.options import handle_common_config_options
+        handle_common_config_options(self)
+
+    def _conan_configure_default(self):
+        from conan.tools.options import handle_common_configure_options
+        handle_common_configure_options(self)
+
+    def _conan_package_id_default(self):
+        from conan.tools.options import handle_common_package_id_options
+        handle_common_package_id_options(self)
```
memsharded: I think better to remove these, a bit of boilerplate

conans/model/options.py
```diff
@@ -139,6 +139,8 @@ def rm_safe(self, field):
             delattr(self, field)
         except ConanException:
             pass
+        except KeyError:
```
memsharded: This needs to be clearly explained, it might change existing behavior, would need a separate PR with tests.

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -396,7 +394,7 @@ def test_toolchain_linux(self, build_type, cppstd, arch, libcxx, shared):
                 "CMAKE_SHARED_LINKER_FLAGS": arch_str,
                 "CMAKE_EXE_LINKER_FLAGS": arch_str,
                 "COMPILE_DEFINITIONS": defines,
-                "CMAKE_POSITION_INDEPENDENT_CODE": "ON"
+                "CMAKE_POSITION_INDEPENDENT_CODE": "ON" if not shared else ""
```
memsharded, line 398: I'd add a comment why this is conditional now

conans/test/unittests/tools/options/test_options_tools.py
```diff
@@ -0,0 +1,60 @@
+from conan.tools.options import (
```
memsharded: Coverage is already excellent with ``integration`` test, I think this can be removed in favor of the integration tests.

memsharded (commented):
Looking good so far!

conans/test/unittests/tools/options/test_options_tools.py
danimtb: I prefer to keep the unit tests, I think they are easier to understand and modify them during development

conans/test/functional/toolchains/cmake/test_cmake.py
danimtb: done

conans/model/options.py
danimtb: Yes, I agree this is a major change. In this case, because in the `_ensure_exists()` method there is a check when the options are `_constrained`. I am not sure what this means so I left this as it is right now and wait for feedback

conans/client/conanfile/configure.py
danimtb: done

danimtb:
> Is it possible to add a new recipe attribute later, only working with Conan 2.x, that could be a shortcut to those new methods?

@uilianries I think we already discussed this before starting the implementation and preferred not to bother with attributes as it will be less intuitive. Maybe in the future, we change our minds, but at the moment it is very straightforward to just declare the options or package type

conans/test/unittests/tools/options/test_options_tools.py
memsharded: On the contrary, my experience is the opposite. They are a burden to maintain, changing something in internals, a small refactor, etc, results in broken unit tests that requires more time fixing than the refactor itself. While the integration tests are rock solid, and provide a perfect coverage already. Is there any functionality in these unittests that is not covered by the integration tests? If that is the case, then it would make sense to keep them, but if the functionality is already covered by the integration tests, I'd prefer to remove the unit tests. Mocks are not easier to understand when something breaks and need maintenance (and it happens quite often, and it is a pain)

conans/client/conanfile/configure.py
```diff
@@ -38,3 +43,11 @@ def run_configure_method(conanfile, down_options, profile_options, ref):
...
         with conanfile_exception_formatter(conanfile, "build_requirements"):
             conanfile.build_requirements()
+
+
+def _conanfile_config_options_default(conanfile):
+    handle_common_config_options(conanfile)
+
+
+def _conanfile_configure_default(conanfile):
+    handle_common_configure_options(conanfile)
```
memsharded: Still boilerplate? What is the purpose of these functions? I'd remove them and call the ``handle_xxxx`` directly in the above calls.

conans/test/integration/options/test_configure_options.py
```diff
@@ -0,0 +1,170 @@
...
+        ["Linux", True, True, False, [True, None, False]],
+        ["Linux", True, False, False, [True, None, False]],
+        ["Linux", True, True, True, [None, None, True]],
+        ["Linux", True, True, True, [None, None, True]],
+        ["Linux", False, True, False, [False, True, False]],
+        ["Linux", False, True, False, [False, True, False]],
+    ])
+    def test_methods_defined_explicit(self, settings_os, shared, fpic, header_only, result):
+        """
+        Test that options are managed automatically when methods config_otpions and configure are not
```
memsharded: Wrong docstring? In this case the methods are defined.

conans/model/options.py
memsharded: Lets make sure this PR is not merged until this is addressed, it is still not evident why it is necessary, if the functionality is basically what the recipes in ConanCenter are doing and ``xxx_safe()`` is used.

conans/client/conanfile/configure.py
danimtb: ok

conans/model/options.py
```diff
@@ -139,6 +139,8 @@ def rm_safe(self, field):
             delattr(self, field)
         except ConanException:
             pass
+        # except KeyError:
```
memsharded: It would make sense to open a separate PR with the simplest integration test that can reproduce the potential need for this.

danimtb: After checking the issue that led to this change, it seems that it is not working properly with the "virtual/consumert" conanfile. In that case, the attribute "_constrained" seems to be false and the `ConanException` is not raising at `_ensure_exists()`.

I don't know how this `_constrained` attribute works, so I don't know how to proceed with that.

-----

Details of the debugging:

1. "Virtual/Consumer" conanfile does not have `config_options()` method defined, so it executes the `handle_common_config_options() method` here: https://github.com/conan-io/conan/blob/07cfcc37c62abc04ab80a0e72f755402d3b9710d/conans/client/conanfile/configure.py#L11 (not this line but the ``else`` clause proposed in this PR).

2. Then, `self.options.rm_safe("fPIC")` is executed by the helper: https://github.com/conan-io/conan/blob/07cfcc37c62abc04ab80a0e72f755402d3b9710d/conans/model/options.py#L137

3. Then, the `ensure_exists()` method is called https://github.com/conan-io/conan/blob/07cfcc37c62abc04ab80a0e72f755402d3b9710d/conans/model/options.py#L175

4. As the `_constrained` attribute is False at this point https://github.com/conan-io/conan/blob/07cfcc37c62abc04ab80a0e72f755402d3b9710d/conans/model/options.py#L159, the ConanException is not raised and not captured by `rm_safe()`. Instead, a `KeyError` is thrown when actually deleting the non-existing option at https://github.com/conan-io/conan/blob/07cfcc37c62abc04ab80a0e72f755402d3b9710d/conans/model/options.py#L176

------

I have opened a small PR with the isolated change of the keyerror to see if it affects the current behavior https://github.com/conan-io/conan/pull/14226

Also, if this is not the right approach I will need some help to understand how `_constrained` attribute works

conan/tools/options/helpers.py
```diff
@@ -0,0 +1,19 @@
...
+def handle_common_configure_options(conanfile):
+    if conanfile.options.get_safe("header_only"):
+        conanfile.options.rm_safe("fPIC")
+        conanfile.options.rm_safe("shared")
+    elif conanfile.options.get_safe("shared"):
+        conanfile.options.rm_safe("fPIC")
+
+
+def handle_common_package_id_options(conanfile):
+    if conanfile.options.get_safe("header_only") or conanfile.package_type == PackageType.HEADER:
```
AbrilRBS: I'm thinking that if by the point this gets executed the package type has already been properly computed, why rely on the option? I don't actually know if having it is better or worse, so happy to hear your opinion :)

conans/client/graph/compute_pid.py
```diff
@@ -81,5 +82,10 @@ def run_validate_package_id(conanfile):
         with conanfile_exception_formatter(conanfile, "package_id"):
             with conanfile_remove_attr(conanfile, ['cpp_info', 'settings', 'options'], "package_id"):
                 conanfile.package_id()
-
+    else:
+        _conanfile_package_id_default(conanfile)
     conanfile.info.validate()
+
+
+def _conanfile_package_id_default(conanfile):
```
AbrilRBS: Leftover helper? The `configure` ones do not have it and it feels like this one shouldn't either

AbrilRBS (approved):
This is looking great, good job @danimtb! Thanks for taking the initiative on it and implementing it :)

I have two minor questions besides the reviews that everyone else has already done, the rest looks great, congrats!

conan/tools/options/helpers.py
memsharded:
```suggestion
    if conanfile.options.get_safe("header_only") or conanfile.package_type is PackageType.HEADER:
```
Enums in Python better compared with ``is`` (https://stackoverflow.com/questions/25858497/should-enum-instances-be-compared-by-identity-or-equality)

conans/client/graph/compute_pid.py
memsharded: Yeah, better align with the configure() approach

danimtb: indeed, I will remove it

conans/test/utils/mocks.py
```diff
@@ -62,6 +62,9 @@ def __getattr__(self, name):
         except KeyError:
             raise ConanException("'%s' value not defined" % name)
 
+    def rm_safe(self, name):
+        self.values.pop(name)
```
AbrilRBS: Maybe 
```suggestion
        self.values.pop(name, None)
```
so it matches the new behaviour and does not raise when the code wouldnt either

AbrilRBS (approved):
Looks good :)

memsharded (approved):

PR #14195, opened 2023-06-29
Locate otool and install_name_tool with xcrun

### LINKED ISSUE #14188: [feature] Allow custom otool and install_name_tool programs in fix_apple_shared_install_name
### What is your suggestion?

I'm cross compiling some Conan recipes from Linux to macOS using [osxcross](https://github.com/tpoechtrager/osxcross) to make the toolchain. `fix_apple_shared_install_name` currently hardcodes `otool` and `install_name_tool`, which doesn't work since osxcross provides it as e.g. `x86_64-apple-darwin22-install_name_tool`. It seems like this could be handled with a set of new config settings like what https://github.com/conan-io/conan/pull/13940 did for cmake.

```
ERROR: libvpx/1.13.0: Error in package() method, line 110
	ConanException: Command 'otool -hv /home/circleci/project/build/conan/p/b/libvpedd69c3a97bd1/p/lib/libvpx.8.dylib' failed with errorcode '127'
b'/bin/sh: 1: otool: not found\n'
```

If that sounds reasonable I can put up a PR.

sfackler:
Changelog: Feature: The `fix_apple_shared_install_name` tool now uses `xcrun` to resolve the `otool` and `install_name_tool` programs.
Docs: Omit

Closes #14188 

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -175,9 +175,19 @@ class XCRun(object):
         """path to libtool"""
         return self.find('libtool')
 
+    @property
+    def otool(self):
+        """path to otool"""
+        return self.find('otool')
 
-def _get_dylib_install_name(path_to_dylib):
-    command = "otool -D {}".format(path_to_dylib)
+    @property
+    def install_name_tool(self):
+        """path to install_name_tool"""
+        return self.find('install_name_tool')
+
+
+def _get_dylib_install_name(otool, path_to_dylib):
+    command = f"{otool} -D {path_to_dylib}"
     output =  iter(check_output_runner(command).splitlines())
     # Note: if otool return multiple entries for different architectures
     # assume they are the same and pick the first one.
@@ -194,26 +204,30 @@ def fix_apple_shared_install_name(conanfile):
     *install_name_tool* utility available in macOS to set ``@rpath``.
     """
 
+    xcrun = XCRun(conanfile)
+    otool = xcrun.otool
+    install_name_tool = xcrun.install_name_tool
+
     def _darwin_is_binary(file, binary_type):
         if binary_type not in ("DYLIB", "EXECUTE") or os.path.islink(file) or os.path.isdir(file):
             return False
-        check_file = f"otool -hv {file}"
+        check_file = f"{otool} -hv {file}"
         return binary_type in check_output_runner(check_file)
 
     def _darwin_collect_binaries(folder, binary_type):
         return [os.path.join(folder, f) for f in os.listdir(folder) if _darwin_is_binary(os.path.join(folder, f), binary_type)]
 
     def _fix_install_name(dylib_path, new_name):
-        command = f"install_name_tool {dylib_path} -id {new_name}"
+        command = f"{install_name_tool} {dylib_path} -id {new_name}"
         conanfile.run(command)
 
     def _fix_dep_name(dylib_path, old_name, new_name):
-        command = f"install_name_tool {dylib_path} -change {old_name} {new_name}"
+        command = f"{install_name_tool} {dylib_path} -change {old_name} {new_name}"
         conanfile.run(command)
 
     def _get_rpath_entries(binary_file):
         entries = []
-        command = "otool -l {}".format(binary_file)
+        command = f"{otool} -l {binary_file}"
         otool_output = check_output_runner(command).splitlines()
         for count, text in enumerate(otool_output):
             pass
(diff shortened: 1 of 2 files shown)


memsharded (commented):

I think this makes sense, and it is cleanly implemented.

It would be good to have some new unit test that validates a bit more the implementation (and ensures it is not broken in the future). Do you think you could add them? (no problem if you don't manage, we can contribute them). Thanks!

jcar87:
Hi @sfackler - thank you for opening this PR.

I'm not entirely familiar with `osxcross`, but is there a chance it contains the `xcrun` tool? I believe the most agnostic way of handling this would be to invoke `xcrun --find install_name_tool` to derive the path to the tool. This is what happens implicitly on macOS when one runs `/usr/bin/install_name_tools` - so I think it would be correct on both macOS and otherwise to use `xcrun` instead. 

Out of curiosity I wonder how `osxcross` handles this in general -  the utility is called `install_name_tool` and that's what most build systems _will_ assume, e.g. https://github.com/Kitware/CMake/blob/7bd24c1da5d093199dedab3fcfad6ec93d4c7aa5/Modules/Platform/Darwin.cmake#L71-L79 - it being rather tedious to have to change it everywhere (CMake, Conan, etc), due to the unconventional name.

sfackler:
Yep, it does contain an unprefixed xcrun:

```
root@409d72e9166d:/# xcrun -find install_name_tool
/cross-macos-0.14.0-2-gf41a2bc-linux-x86_64/bin/x86_64-apple-darwin20.4-install_name_tool
```

I can update this to remove the confs and resolve install_name_tool and otool from xcrun if you'd prefer.

osxcross does ship a cmake toolchain that handles the install_name_tool name for that build system at least: https://github.com/tpoechtrager/osxcross/blob/564e2b9aa8e7a40da663d890c0e853a1259ff8b1/tools/toolchain.cmake#L41

Happy to add tests as well!

memsharded:
> Yep, it does contain an unprefixed xcrun:

Ok, that is great, thanks for checking.

> I can update this to remove the confs and resolve install_name_tool and otool from xcrun if you'd prefer.

Yes, I think it would be better if the ``xcrun`` brings those tools automatically, and there is no need for extra confs. If you think you can do that, that would be amazing.

sfackler:
Updated to use xcrun and remove the new confs.

conan/tools/apple/apple.py
```diff
@@ -194,26 +204,33 @@ def fix_apple_shared_install_name(conanfile):
     *install_name_tool* utility available in macOS to set ``@rpath``.
     """
 
+    if not is_apple_os(conanfile):
+        return
+
+    xcrun = XCRun(conanfile)
+    otool = xcrun.otool
+    install_name_tool = xcrun.install_name_tool
```
sfackler, line 212: Alternatively, we could just prefix all of the commands with `xcrun`, e.g. `xcrun otool -hv {file}`

memsharded: I think it is good as is, thanks!

conan/tools/apple/apple.py
```diff
@@ -194,26 +204,33 @@ def fix_apple_shared_install_name(conanfile):
     *install_name_tool* utility available in macOS to set ``@rpath``.
     """
 
+    if not is_apple_os(conanfile):
+        return
+
+    xcrun = XCRun(conanfile)
```
memsharded, line 210: Just curious (and possibly ignorance about how it works). How this ``XCRun`` connects to the specific ``osxcross`` environment? Via ``os.sdk`` setting? I guess that you have tested against your ``osxcross``, and it works, but as it is not covered by full functional tests, I'd like to understand how it works.

memsharded (approved):
I think it looks good, but I'll defer it to @jcar87 that knows more than me in OSX.

conan/tools/apple/apple.py
sfackler: I think the xcrun has that relative SDK path hardcoded as the default search location when it was compiled, but I'm not totally sure.

On a semi-related note, I do need to explicitly set `tools.apple:sdk_path` for some CMake logic to work properly - would it be worth it to have the CMakeToolchain tool query XCRun for the SDK path if it's not explicitly configured?

jcar87 (approved):
Awesome, thanks so much for your contribution @sfackler!

memsharded:
Thanks for the contribution, this will be in next 2.0.8.

I am spinning a new issue from your comment above, I think it is necessary to follow up

PR #14223, opened 2023-07-04
Allow `tools.gnu:make_program` to work in every case

### LINKED ISSUE #14197: How to properly set CMAKE_MAKE_PROGRAM?
### What is your question?

I'm new to Conan and I'm trying to get a sample package (sqlite) building for Android on my Windows machine. The main problem I'm having is that I cannot figure out how to set the CMAKE_MAKE_PROGRAM environment variable such that Conan picks it up properly and propagates it to CMake.

After a fair bit of tinkering, I've figured out how to get it to work, but I'm 100% sure this is the wrong way to do it:

```py
    def generate(self):
        deps = CMakeDeps(self)
        deps.generate()
        tc = CMakeToolchain(self)
        # this feels very, very wrong
        tc.variables["CMAKE_MAKE_PROGRAM"]="c:/Users/andrew/AppData/Local/Android/Sdk/cmake/3.22.1/bin/ninja.exe"
        tc.generate()
```

I know enough about Conan at this point to know that embedding local paths into recipes is a terrible idea. Also while it does build successfully, it fails the test step, which is probably a good thing since it further dissuades anyone from doing this. But how do I set CMAKE_MAKE_PROGRAM properly? I've tried setting it in the profile like this:

```
include(default)

[settings]
os=Android
os.api_level=26
arch=armv8
compiler=clang
compiler.version=14
compiler.libcxx=c++_static
compiler.cppstd=17

[buildenv]
# this felt like it should work, but it doesn't
CMAKE_MAKE_PROGRAM=c:/Users/andrew/AppData/Local/Android/Sdk/cmake/3.22.1/bin/ninja.exe

[runenv]
# this felt like it could work, but it doesn't
CMAKE_MAKE_PROGRAM=c:/Users/andrew/AppData/Local/Android/Sdk/cmake/3.22.1/bin/ninja.exe

[tool_requires]
android-ndk/r25

[conf]
tools.cmake.cmaketoolchain:generator=Ninja
```

But that doesn't seem to have an effect: I just get:

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
```

There must be a way to do this, but I cannot find an example anywhere and I'm running out of things to read. What am I doing wrong? Any help would be greatly appreciated!

Thanks!

AbrilRBS:
Changelog: Fix: Allow `tools.gnu:make_program` to affect every CMake configuration.
Docs: Omit

The PR that implemented it was https://github.com/conan-io/conan/pull/10770 and already had a check for only mingw, but sadly no insight into _why_ it was gated off of the rest of configurations. After checking a bit and not seeing anything that might be _wrong_ with allowing it, I'm proposing to allow it in every configuration

Closes https://github.com/conan-io/conan/issues/14197

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -28,11 +28,11 @@ class _CMakePresets:
             if "CMAKE_SH" not in cache_variables:
                 cache_variables["CMAKE_SH"] = "CMAKE_SH-NOTFOUND"
 
-            cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
-                                                    default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
-            if cmake_make_program:
-                cmake_make_program = cmake_make_program.replace("\\", "/")
-                cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
+        cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
+                                                default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
+        if cmake_make_program:
+            cmake_make_program = cmake_make_program.replace("\\", "/")
+            cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
 
         if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
             cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
--- a/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
+++ b/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
@@ -552,6 +552,11 @@ def test_toolchain_cache_variables():
     assert "-DCMAKE_TOOLCHAIN_FILE=" in client.out
     assert f"-G {_format_val('MinGW Makefiles')}" in client.out
 
+    client.run("install . --name=mylib --version=1.0 -c tools.gnu:make_program='MyMake'")
+    presets = json.loads(client.load("CMakePresets.json"))
+    cache_variables = presets["configurePresets"][0]["cacheVariables"]
+    assert cache_variables["CMAKE_MAKE_PROGRAM"] == "MyMake"
+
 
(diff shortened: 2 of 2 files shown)


memsharded (commented):
Looks good, if it doesn't break anything, I think we can add an integration test to cover it and it could be merged.

AbrilRBS:
> I think we can add an integration test to cover it and it could be merged.

I didn't initially add one becase the test that is already there looked fine, but after double checking, it does not hurt to add an extra check! Good catch :)

memsharded:
> I didn't initially add one becase the test that is already there looked fine, but after double checking, it does not hurt to add an extra check! Good catch :)

By definition, the test is not there, because the necessary test is one that will not be using mingw. It should be a red-green test, failing without the fix, passing with the fix.

conan/tools/cmake/presets.py
```diff
@@ -28,11 +28,11 @@ def generate(conanfile, toolchain_file, generator, cache_variables, preset_prefi
...
-            cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
-                                                    default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
-            if cmake_make_program:
-                cmake_make_program = cmake_make_program.replace("\\", "/")
-                cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
+        cmake_make_program = conanfile.conf.get("tools.gnu:make_program",
+                                                default=cache_variables.get("CMAKE_MAKE_PROGRAM"))
+        if cmake_make_program:
+            cmake_make_program = cmake_make_program.replace("\\", "/")
+            cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
```
memsharded, line 35: I know this was already there, and not strictly related to this PR. But why is it a cache variable (presets) and not a toolchain one? Because it means that this is an extra things that users not using presets should remember to pass in the command line too. It would be worth checking if this works in the toolchain (probably will fail, and this is the reason it is a cache variable)

czoido (approved):

PR #14226, opened 2023-07-05
[wip] options keyerror rm_safe()

danimtb:
Related to https://github.com/conan-io/conan/pull/14194

### DIFF
--- a/conans/model/options.py
+++ b/conans/model/options.py
@@ -139,6 +139,8 @@ class _PackageOptions:
             delattr(self, field)
         except ConanException:
             pass
+        except KeyError:
+            pass
 
     def validate(self):
         for child in self._data.values():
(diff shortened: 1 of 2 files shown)


conans/model/options.py
```diff
@@ -139,6 +139,8 @@ def rm_safe(self, field):
             delattr(self, field)
         except ConanException:
             pass
+        except KeyError:
```
memsharded: Instead of adding this, the best would try to start with a test that breaks, if this is not here. Then we can take it from there, and see the best solution, but better first to illustrate the problem.

danimtb: I have pushed the minimal changes that make it fail and removed this

memsharded:
Superseded by https://github.com/conan-io/conan/pull/14238

PR #14227, opened 2023-07-05
Support multiple custom global generators

BobIsOnFire:
Changelog: Bugfix: Fix conan not finding generator by name when multiple custom global generators are detected.
Docs: omit

Hi!
Noticed this when migrating my generators from 1.X to 2.0. Looks like `result` is always populated only with the last class that it found on the way, not sure if this is intended.

Looking forward to the review!


memsharded (approved):

It was certainly a bug, and there was enough testing to cover it.
Good test too, excellent!

memsharded:
Merged, it will be released in 2.0.8, thanks again!

PR #14233, opened 2023-07-06
[PkgConfigDeps] `pkg_config_custom_content` property overwrites variables

franramirez688:
Changelog: Feature: Let `pkg_config_custom_content` overwrite default `*.pc` variables created by `PkgConfigDeps`.
Changelog: Feature: Let `pkg_config_custom_content` be a dict-like object too.
Docs: https://github.com/conan-io/docs/pull/3293

UPDATED:
* Simplified internal `PkgConfigDeps` logic: only one template and more straightforward, one context, Python is formatting the template, etc.
* `pkg_config_custom_content` overwrites Conan-defined freeform variables, not keyword metadata ones.
* `pkg_config_custom_content` could be defined as a dictionary instead of a string.

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -1,12 +1,13 @@
 import os
+import re
 import textwrap
 from collections import namedtuple
 
 from jinja2 import Template, StrictUndefined
 
+from conan.errors import ConanException
 from conan.internal import check_duplicated_generator
 from conan.tools.gnu.gnudeps_flags import GnuDepsFlags
-from conan.errors import ConanException
 from conans.model.dependencies import get_transitive_requires
 from conans.util.files import save
 
@@ -125,9 +126,9 @@ class _PCContentGenerator:
         {%- endfor -%}
         {%- endmacro -%}
 
-        {%- if prefix -%}
+        {% if prefix %}
         prefix={{ prefix }}
-        {%- endif -%}
+        {% endif %}
         {% for path in libdirs %}
         {{ "libdir{}={}".format((loop.index0 or ""), path) }}
         {% endfor %}
@@ -153,9 +154,9 @@ class _PCContentGenerator:
     """)
 
     shortened_template = textwrap.dedent("""\
-        {%- if prefix -%}
+        {% if prefix %}
         prefix={{ prefix }}
-        {%- endif -%}
+        {% endif %}
         {% if pkg_config_custom_content %}
         # Custom PC content
         {{ pkg_config_custom_content }}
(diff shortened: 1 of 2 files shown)


conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -209,8 +234,9 @@ def shortened_content(self, info):
         assert isinstance(info, _PCInfo)
         custom_content = info.cpp_info.get_property("pkg_config_custom_content") if info.cpp_info \
             else None
+        prefix_path = "" if self._is_var_match("prefix", custom_content) else self._get_prefix_path()
```
czoido: IMHO we should leave the prefix path out of this logic as it contains the package folder? If we want users to be able to define prefix in the custom content, another option would be to not use the prefix at all and add the absolute paths directly in our pc files?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -175,32 +180,52 @@ def _get_prefix_path(self):
...
+            """
+            custom_content = cpp_info.get_property("pkg_config_custom_content")
+            prefix_path = "" if self._is_var_match("prefix", custom_content) else \
+                self._get_prefix_path()
+            libdirs = [] if self._is_var_match("libdir", custom_content) else \
+                _get_formatted_dirs(cpp_info.libdirs, prefix_path)
+            includedirs = [] if self._is_var_match("includedir", custom_content) else \
+                _get_formatted_dirs(cpp_info.includedirs, prefix_path)
+            bindirs = [] if self._is_var_match("bindir", custom_content) else \
+                _get_formatted_dirs(cpp_info.bindirs, prefix_path)
```
memsharded: I think can have a cleaner implementation and cleaner output file
- Use a regex that capture all variable names in the file
- Filter the variable names out in the dict below (line 217)

franramirez688: What I'm thinking is to let users put that property as a dict-like instead of a string. That will simplify everything.

memsharded: But that could be breaking, and free format is still necessary (that would be a new feature? but we still need to make it work with current text input, right?)

franramirez688: Yes, absolutely! It's going to be only a second way to proceed.

conan/tools/gnu/pkgconfigdeps.py
franramirez688: I don't think so. `prefix` is part of the freeform variables added to all the *.pc files so if a user decides to overwrite it, let's allow it. I think it won't ever happen, but it's so, it would be a similar case if they decide to override the `package_folder` or any other variable defined by Conan, wouldn't it?

conan/tools/gnu/pkgconfigdeps.py
```diff
@@ -175,50 +126,74 @@ def _get_prefix_path(self):
...
+                    key, value = match.group(1).strip(), match.group(2).strip()
+                    pc_variables[key] = value
+        return pc_variables
+
+    def _get_lib_flags(self, libdirvars, cpp_info):
+        gnudeps_flags = GnuDepsFlags(self._conanfile, cpp_info)
+        libdirsflags = ['-L"${%s}"' % d for d in libdirvars]
+        system_libs = ["-l%s" % l for l in (cpp_info.libs + cpp_info.system_libs)]
+        shared_flags = [f for f in (cpp_info.sharedlinkflags + cpp_info.exelinkflags)]
+        framework_flags = [f for f in (gnudeps_flags.frameworks + gnudeps_flags.framework_paths)]
```
memsharded:
```suggestion
        shared_flags = cpp_info.sharedlinkflags + cpp_info.exelinkflags
        framework_flags = gnudeps_flags.frameworks + gnudeps_flags.framework_paths
```

memsharded (approved):

conan/tools/gnu/pkgconfigdeps.py
franramirez688: copy&paste from Jinja template 😓

czoido (approved):

PR #14253, opened 2023-07-08
fix leaking of CMakeDeps lib_suffix variable

### LINKED ISSUE #14225: [bug] Linking a package with two dlls in the lib directory fails
### Environment details

* Compiler+version: msvc 193
* Conan version: 2.0.6
* Python version: 3.10

### Steps to reproduce

1. Export a prebuilt package on windows with following characteristics:
 - The conanfile has either option shared=True or package_type = "shared-library"
 - The package has at least two dlls with import libs.
 - The dlls and import libs copied to the same lib/ directory inside the package.
2. Use the package in another project that uses CMakeDeps generator and Visual Studio.
3. The build fails because Visual Studio will try to link to the second dll instead of the second import lib.

For example with "prebuilt_local_project" conan tutorial modified to contain 2 dlls:
`C:\conan_examples2\tutorial\creating_packages\other_packages\prebuilt_local_project>conan new cmake_lib -d name=hello -d version=0.1`

Modify CMakeLists.txt:
```
project(hello2 CXX)
add_library(hello2 src/hello.cpp)
target_include_directories(hello2 PUBLIC include)
set_target_properties(hello2 PROPERTIES PUBLIC_HEADER "include/hello.h")
```
Modify conanfile.py:
```
import os

    def package(self):
        copy(self, pattern="*.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, pattern="*.dll", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, pattern="*.h", src=self.recipe_folder, dst=self.package_folder, keep_path=True)
    def package_info(self):
        self.cpp_info.libs = ["hello", "hello2"]
        self.cpp_info.bindirs = ["lib"]
```
`C:\conan_examples2\tutorial\creating_packages\other_packages\prebuilt_local_project>conan build . -s build_type=Debug -o shared=True`
`C:\conan_examples2\tutorial\creating_packages\other_packages\prebuilt_local_project>conan export-pkg . -s build_type=Debug -o shared=True`
```
...
hello/0.1 (test package): RUN: cmake --build "C:\conan_examples2\tutorial\creating_packages\other_packages\prebuilt_local_project\test_package\build\msvc-193-x86_64-14-debug" --config Debug
MSBuild version 17.5.1+f6fdcf537 for .NET Framework
  Checking Build System
  Building Custom Rule C:/conan_examples2/tutorial/creating_packages/other_packages/prebuilt_local_project/test_package/CMakeLists.txt
  example.cpp
C:\Users\tmusto\.conan2\p\b\hellocbc01ad6d9aa3\p\lib\hello2.dll : fatal error LNK1107: invalid or corrupt file: cannot read at 0x340 [C:\conan_examples2\tutorial\creating_packages\other_packages\prebuilt_local_project\test_package\build\msvc-193-x86_64-14-debug\example.vcxproj]
```

### Logs

_No response_

memsharded:
Changelog: Bugfix: Fix leaking of ``CMakeDeps`` ``CMAKE_FIND_LIBRARY_SUFFIXES`` variable.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14225

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/macros.py
+++ b/conan/tools/cmake/cmakedeps/templates/macros.py
@@ -58,9 +58,12 @@ class MacrosTemplate(CMakeDepsFileTemplate):
                    set(_LIB_NAME CONAN_LIB::${package_name}_${_LIBRARY_NAME}${config_suffix})
 
                    if(is_host_windows AND library_type STREQUAL "SHARED")
+                     # Store and reset the variable, so it doesn't leak
+                     set(_OLD_CMAKE_FIND_LIBRARY_SUFFIXES ${CMAKE_FIND_LIBRARY_SUFFIXES})
                      set(CMAKE_FIND_LIBRARY_SUFFIXES .dll ${CMAKE_FIND_LIBRARY_SUFFIXES})
                      find_library(CONAN_SHARED_FOUND_LIBRARY NAMES ${_LIBRARY_NAME} PATHS ${package_bindir}
                                   NO_DEFAULT_PATH NO_CMAKE_FIND_ROOT_PATH)
+                     set(CMAKE_FIND_LIBRARY_SUFFIXES ${_OLD_CMAKE_FIND_LIBRARY_SUFFIXES})
                      if(NOT CONAN_SHARED_FOUND_LIBRARY)
                        message(STATUS "Cannot locate shared library: ${_LIBRARY_NAME}")
                        message(DEBUG "DLL library not found, creating UNKNOWN IMPORTED target")


memsharded:
I haven't added a test for this, because it is a very niche bug (only Windows, only DLLs, only having more than 1 DLL and having them in the "lib" folder, not the "bin" folder, which is extremely unusual), the test would take me way too much time compared with the fix itself.

AbrilRBS (approved):

PR #14254, opened 2023-07-08
Allow `upload --list` to accept empty lists

### LINKED ISSUE #14203: [bug] conan upload --list errors on empty recipe list
### Environment details

* Operating System+version:
Manjaro Linux 23.0.0

* Compiler+version:
N/A

* Conan version:
Conan version 2.0.7

* Python version:
Python 3.11.3

### Steps to reproduce

1. Create a list file as follows (can be generated from something like `conan list --graph install_graph.json --graph-binaries build --format json`)
```json
{
  "Local Cache": {}
}
```
2. conan upload --remote foo --list list.json
3. Conditional check fails: https://github.com/conan-io/conan/blob/release/2.0/conan/cli/commands/upload.py#L84-L85

Note that I believe this should work since some automation workflows in CI may result in an empty list if a specific package was just cached and `--graph-binaries build` is set in the list command. I think the conditional check listed below should probably only error if the `--list` argument was not provided.

### Logs

```
$> conan upload --remote foo --list list.json
ERROR: No recipes found matching pattern 'None'
```

n-morales:
Changelog: Fix: Using `upload` with the `--list` argument now permits empty recipe lists.
Docs: Omit

My approach was to still error on recipe patterns that are empty when not using a list. Therefore, I added two tests -- one to ensure that passing a `--list` argument that had no recipes in it was valid, and the second to ensure passing a non-matching pattern still errored with the "No recipes found matching pattern" message.

Fixes #14203

### DIFF
--- a/conan/cli/commands/upload.py
+++ b/conan/cli/commands/upload.py
@@ -1,6 +1,6 @@
 from conan.api.conan_api import ConanAPI
 from conan.api.model import ListPattern, MultiPackagesList
-from conan.api.output import cli_out_write
+from conan.api.output import cli_out_write, ConanOutput
 from conan.cli import make_abs_path
 from conan.cli.command import conan_command, OnceArgument
 from conan.cli.commands.list import print_list_json, print_serial
@@ -96,8 +96,11 @@ def upload(conan_api: ConanAPI, parser, *args):
         if not args.dry_run:
             conan_api.upload.upload(package_list, remote)
             conan_api.upload.upload_backup_sources(package_list)
-    elif not args.list:
-        # Don't error on no recipes for automated workflows using list
+    elif args.list:
+        # Don't error on no recipes for automated workflows using list,
+        # but warn to tell the user that no packages were uploaded
+        ConanOutput().warning(f"No packages were uploaded because the package list is empty.")
+    else:
         raise ConanException("No recipes found matching pattern '{}'".format(args.pattern))
 
     pkglist = MultiPackagesList()
--- a/conan/cli/commands/upload.py
+++ b/conan/cli/commands/upload.py
@@ -81,23 +81,24 @@ def upload(conan_api: ConanAPI, parser, *args):
         ref_pattern = ListPattern(args.pattern, package_id="*", only_recipe=args.only_recipe)
         package_list = conan_api.list.select(ref_pattern, package_query=args.package_query)
 
-    if not package_list.recipes:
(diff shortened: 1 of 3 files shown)


conan/cli/commands/upload.py
```diff
@@ -81,24 +81,25 @@ def upload(conan_api: ConanAPI, parser, *args):
...
+            _ask_confirm_upload(conan_api, package_list)
+
+        conan_api.upload.prepare(package_list, enabled_remotes, args.metadata)
+
+        if not args.dry_run:
+            conan_api.upload.upload(package_list, remote)
+            conan_api.upload.upload_backup_sources(package_list)
+    elif not args.list:
+        # Don't error on no recipes for automated workflows using list
         raise ConanException("No recipes found matching pattern '{}'".format(args.pattern))
```
AbrilRBS, line 104: It might be good to output a warning in the case of empty lists, so that the user knows what happened :)

AbrilRBS (commented):
Thanks a lot! I just have a small suggestion for a warning in case the list is empty.

Other than that, it looks great! Thanks a lot for your contribution :)

conan/cli/commands/upload.py
n-morales: Thanks for the suggestion. I added a warning for this

memsharded (approved):

conans/test/integration/command_v2/test_combined_pkglist_flows.py
```diff
@@ -37,6 +37,18 @@ def test_list_upload_packages(self, client):
...
+        client.run(f"install --requires=zlib/1.0.0@user/channel -f json",
+                   redirect_stdout="install_graph.json")
+
+        # Generate an empty pkglist.json
+        client.run(f"list --format=json --graph=install_graph.json --graph-binaries=bogus",
+                   redirect_stdout="pkglist.json")
+
+        # No binaries should be uploaded since the pkglist is empty, but the command
+        # should not error
+        client.run("upload --list=pkglist.json -r=default")
```
AbrilRBS, line 50:
```suggestion
        client.run("upload --list=pkglist.json -r=default")
        assert "No packages were uploaded because the package list is empty." in client.out
```

AbrilRBS (approved):

AbrilRBS:
Thanks a lot for taking the time to contribute this feature, we really appreciate it :)

PR #14296, opened 2023-07-13
Fix duplicate presets

### LINKED ISSUE #13968: [question] Using Ninja (Multi-Config) with Extending Own CMake Presets Example
### What is your question?

Howdy, I am having troubles getting one of the CMake examples working with Ninja, and I assume I am doing something wrong, but can't for the life of me figure out what.

I have cloned (pulled changes this morning) the examples2 repo and am starting with the extend_own_cmake_presets example.  For my tests, I am just trying to specify the generator in conanfile.py on line 29.  When attempting to use `tc = CMakeToolchain(self, generator="Ninja Multi-Config")`, I am consistently getting bad CMakePresets which doesn't let me continue.

- On Both Windows and Linux, if I run `cmake --list-prests` before installing, I get the expected "can't read ConanPresets.json" since the referenced file isn't created yet.
- Next, I ran `conan install .` on both operating systems and running `cmake --list-presets` again, I got errors of "Duplicate Presets" on both Operating systems, however they did different things under the covers.  For both systems, ConanPresets.json contains a "duplicate conan-debug" buildPreset.  On Linux, it appeared to use a single config template as the CMakePresets.json file that was included is in the "build/Release/generators" directory as opposed to the "build/generators" directory that was created on Windows.
- Next I ran `conan install . -s build_type=Debug` on both systems.  Windows has appeared to correct the previous "Duplicate" presets warning as `cmake --list-presets` now lists all 6 presets.  However, on linux I'm still getting a "Duplicate Presets" error.  The duplicate buildPreset in ConanPresets.json has been cleared, up but now there are two includes from the build/Release and build/Debug directories that both define different versions of the conan-default configurePreset.

I blew away ConanPresets.json and the build directory to try the regular non Multi-Config version of Ninja using `tc = CMakeToolchain(self, generator="Ninja")`

- I started again by running `conan install .` to create ConanPresets.json.  It appears the "exact" same files were generated as before, with Windows now using the Multi-Config option even though just "Ninja" was requested.
- I now ran `conan install . -s build_type=Debug` to generate the other configuration and the results again were reversed.  On Linux, the ConanPresets.json duplicate was resolved and including the two files correctly resolved the rest of the conan- presets being inherited from the main file.  On Windows, ConanPresets was also resolved, but generated CMakePresets.json file in build/generators REPLACED the original Release configuration with Debug configuration causing `cmake --list-presets` to report an "Invalid preset" since none of the files have conan-release that we are trying to inherit from.

I'm not sure if I'm doing something wrong or if I'm running into a bug of some sort.  Even though the generator I pass into the CMakeToolchain call is showing up in the Preset files, I don't think it is actually being fully honored/evaluated when trying to generate the files.

Also, is it the expectation that I would need to install ALL configurations before I can configure/build one of them?  For example, once I add "RelWithDebInfo" to my CMakePresets, it seems like I would have to install it, otherwise I'll keep running into the issue of a "missing" preset since nothing generated conan-relwithdebinfo?

Thanks!

wayneph01:
Changelog: Bugfix: ConanPresets.json contains duplicate presets if multiple user presets inherit from the same conan presets.
Docs: Omit

fixes #13968

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -268,13 +268,14 @@ class _IncludingPresets:
             if os.path.exists(user_file):
                 user_json = json.loads(load(user_file))
                 for preset_type in types:
+                    conan_inherits = []
                     for preset in user_json.get(preset_type, []):
                         inherits = preset.get("inherits", [])
                         if isinstance(inherits, str):
                             inherits = [inherits]
-                        conan_inherits = [i for i in inherits if i.startswith(preset_prefix)]
-                        if conan_inherits:
-                            collected_targets.setdefault(preset_type, []).extend(conan_inherits)
+                        conan_inherits.extend([i for i in inherits if i.startswith(preset_prefix)])
+                    if len(conan_inherits):
+                        collected_targets.setdefault(preset_type, []).extend(list(set(conan_inherits)))
         return collected_targets
 
     @staticmethod
--- a/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
+++ b/conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
@@ -406,6 +406,67 @@ def test_cmake_presets_binary_dir_available():
     assert presets["configurePresets"][0]["binaryDir"] == build_dir
 
 
+def test_cmake_presets_shared_preset():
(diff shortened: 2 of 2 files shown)


memsharded (approved):
Many thanks! Merging this for next 2.0.9 release

PR #14302, opened 2023-07-14
Add apple-clang 15

AbrilRBS:
Changelog: Feature: Add `apple-clang` 15.
Docs: Omit

I recently downloaded XCode 15 (Beta 3) to try some things, found that after a `conan profile detect` I had 15 which was not supported by default.


franramirez688 (approved):

memsharded (approved):
Sounds good.
Did you check that apple-clang 15 doesn't change anything in term of default cppstd, like any change necessary in ``_apple_clang_supported_cppstd`` or similar?

AbrilRBS:
Yes! The team suggested that I should look into it and that C++ versions were usually the reason that updates were delayed, so I made sure that: The default C++ version hasn't changed, it's still 98 if not specified, and no new c++ flag is available (c++2c _might be_, but we don't add those so didn't even check)

PR #14317, opened 2023-07-18
fixed pyyaml installation due to cython update

piri-nyan:
Changelog: (Fix | Bugfix): PyYAML installation fix

- Problem occured at my company this morning, fixed it like this:
`pip install "cython<3.0.0" && pip install --no-build-isolation pyyaml==6.0`
but its a workaround, so here is an actual fix.

- I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- I've followed the PEP8 style guides for Python code.

Its a follow up on issue in PyYAML: https://github.com/yaml/pyyaml/issues/724

### DIFF
--- a/conans/requirements.txt
+++ b/conans/requirements.txt
@@ -1,7 +1,7 @@
 requests>=2.25, <3.0.0
 urllib3>=1.26.6, <1.27
 colorama>=0.3.3, <0.5.0
-PyYAML>=3.11, <=6.0
+PyYAML==6.0.1
 patch-ng>=1.17.4, <1.18
 fasteners>=0.14.1
 six>=1.10.0,<=1.16.0
--- a/conans/__init__.py
+++ b/conans/__init__.py
@@ -20,4 +20,4 @@ OAUTH_TOKEN = "oauth_token"
 SERVER_CAPABILITIES = [COMPLEX_SEARCH_CAPABILITY, REVISIONS]  # Server is always with revisions
 DEFAULT_REVISION_V1 = "0"
 
-__version__ = '1.59.0'
+__version__ = '1.59.0-dev'
--- a/conans/__init__.py
+++ b/conans/__init__.py
@@ -20,4 +20,4 @@ OAUTH_TOKEN = "oauth_token"
 SERVER_CAPABILITIES = [COMPLEX_SEARCH_CAPABILITY, REVISIONS]  # Server is always with revisions
 DEFAULT_REVISION_V1 = "0"
 
-__version__ = '1.59.0-dev'
+__version__ = '1.59.0'


memsharded (changes requested):
Hi @piri-nyan 

Thanks for your contribution.
However, as you can see, it is breaking the CI https://ci.conan.io/blue/organizations/jenkins/ConanTestSuite/detail/PR-14317/2/pipeline

It would be necessary to explain a bit better how the cython update is related. Conan doesn't use cython at all, so it shouldn't be affected. Thanks!

piri-nyan:
Hello! Conan doesn't use cython directly, but 1.59 requires PyYAML<6.0 which uses cython without strict version indication.
This 'not specified version of cython' in PyYAML package led to errors in versions of PyYAML<6.0 which conan 1.59 uses. PyYAML released a fix in 6.0.1 version, but conan installation prohibits its usage since it is strictly indicated to use PyYAML<6.0 in requirements.
Hope we can resolve this issue, thank you in advance!

piri-nyan:
Also, i dont think that it is my changes which caused errors in CI.

memsharded:
> Also, i dont think that it is my changes which caused errors in CI.

It could be something related to CI.

In any case, this fix should target first the latest released 1.X, not older ones. Then, we can discuss how long to backport it, but it is not possible to target 1.59 without having patched first 1.60 (with a 1.60.2 patch, the latest released in 1.60.1).

Please change your PR to target latest 1.60 first

conans/requirements.txt
```diff
@@ -1,7 +1,7 @@
 requests>=2.25, <3.0.0
 urllib3>=1.26.6, <1.27
 colorama>=0.3.3, <0.5.0
-PyYAML>=3.11, <=6.0
+PyYAML==6.0.1
```
memsharded, line 4: I am concerned of too restrictive constraint would break some special users that might workaround the cython issue, use older PyYAML version in some special distros or whatever. I have done https://github.com/conan-io/conan/pull/14323 to propose allowing future PyYAML patches too, I think it should be pretty safe and better in general than a strong pinning to 6.0.1

piri-nyan: Okiii, we can decline this PR then, i guess we'll have to wait for PyYAML to fix their builds :disappointed:

piri-nyan: Thank you!

memsharded: Thanks to you for raising the issue and contributing a fix!

>  i guess we'll have to wait for PyYAML to fix their builds

But https://github.com/conan-io/conan/pull/14323 will also fix immediately the issue in Conan 1.60.2, won't it?

piri-nyan: It will, i don't question that, i just hope that PyYAML will resolve issues with their builds. Because if they wont, this issue will come up again sometime :)

PR #14320, opened 2023-07-18
Add implements attribute in ConanFile to provide opt-in automatic management of some options and settings.

### LINKED ISSUE #14308: fPIC
### Environment details

* Compiler+version: gcc 12.1
* Conan version: 2.0.8
* Python version: 3.10

Referring to commit 2af05b5: Why is fPIC automatically disabled when setting shared in the conanfile options?  This was a sneaky change that took a while to discover and presented an instability between Conan versions.

I still need fPIC enabled in order for the cmake toolchain to properly set the fPIC option.

### Steps to reproduce

_No response_

### Logs

_No response_

czoido:
Changelog: Feature: Add `implements` attribute in ConanFile to provide automatic management of some options and settings.
Docs: https://github.com/conan-io/docs/pull/3303

Closes: https://github.com/conan-io/conan/issues/14308


memsharded:
Did some minor changes:
- Replaced ``getattr`` for normal access
- Changed ``GenConanFile`` specialized for generic ``with_class_attribute``

> Add options_auto_handle=True to the conan new template?

Probably not yet? The advantage of the current template is that it will work too as-is in latest Conan 1.X

conans/client/conanfile/configure.py
```diff
@@ -10,7 +10,7 @@ def run_configure_method(conanfile, down_options, profile_options, ref):
     if hasattr(conanfile, "config_options"):
         with conanfile_exception_formatter(conanfile, "config_options"):
             conanfile.config_options()
-    else:
+    elif conanfile.activate_default_behaviors:
```
memsharded: Don't love the name, but challenging to find a better one. Some ideas:
- ``auto_configuration = True`` 
- ``manage_options = True``
- ``auto_manage_library_type_configuration = True``
- ``auto_manage_library_configuration = True``
- ``auto_library_configuration = True``
- ``library_type_defaults = True``

Generic list of things?
- ``auto_manage = ["library_type_configuration", "app_type_configuration"]``
- ``auto_manage = ["library_configuration", "app_configuration"]``
- ``auto = ["library_configuration", "app_configuration"]``

- ``policies = ["auto_shared_fpic", "auto_header_only", "auto_cmake"]``
- ``auto_manage = ["shared_fpic", "header_only"]``
- ``rules, auto, defaults, implicits, auto_behaviors, mods, builtins, mixin, recipe_auto, recipe_defaults, config_templates``
- ``implements = ["auto_shared_fpic", "cmake_template"]``
- ``applies = [``
- ``sidecar = ["auto_shared_fpic", ]``
- ``recipe_presets = [``
- ``salt_and_pepper = ["auto_shared_fpic"``
- ``pre-cooked = ["auto_shared_``

``build_policy = "missing", upload_policy = "skip"``

The language case would be a different one most likely:
- ``languages = "C", "C++", "fortran"``

Maybe "configuration" and "options" should be in the var name?

FINAL DECISION:
- ``implements = ["auto_shared_fpic", "auto_header_only"]``

danimtb (approved):
I like the attribute name "activate_default_behavior" (without S at the end).

conans/client/conanfile/configure.py
franramirez688: I agree. We should think of another shorter one if possible.

conans/client/conanfile/configure.py
```diff
@@ -20,7 +20,7 @@ def run_configure_method(conanfile, down_options, profile_options, ref):
     if hasattr(conanfile, "configure"):
         with conanfile_exception_formatter(conanfile, "configure"):
             conanfile.configure()
-    else:
+    elif conanfile.activate_default_behaviors:
         default_configure(conanfile)
```
memsharded:
```suggestion
        if ... => auto_shared_fpic_configure(conanfile)
        if ... => auto_header_only_configure(conanfile)
```

memsharded (commented):
Drop the public methods

conans/client/conanfile/implementations.py
```diff
@@ -0,0 +1,43 @@
+import inspect
+
+from conans.model.pkg_type import PackageType
+
+
+def call(module, implementation_name, conanfile_method, *args, **kwargs):
+    # first check that we have at least one function in the module that starts
+    # with the implementation name, then, when calling to the function name composed
+    # with the conanfile_method name we pass on the exception
+    def _has_function_starting_with(module, prefix):
```
czoido: Maybe we want this to be checked earlier

conans/client/conanfile/implementations.py
```diff
@@ -0,0 +1,43 @@
...
+        pass
+
+
+def auto_shared_fpic_config_options(conanfile):
+    if conanfile.settings.get_safe("os") == "Windows":
+        conanfile.options.rm_safe("fPIC")
+
+
+def auto_shared_fpic_configure(conanfile):
+    if conanfile.options.get_safe("header_only"):
```
memsharded, line 10: Should this ``if`` be part of the ``header_only`` automation? Maybe not, I guess the meaning of ``auto-shared_fpic`` is "you can remove these options if they are not necessary"

conans/client/conanfile/configure.py
```diff
@@ -10,8 +10,9 @@ def run_configure_method(conanfile, down_options, profile_options, ref):
     if hasattr(conanfile, "config_options"):
         with conanfile_exception_formatter(conanfile, "config_options"):
             conanfile.config_options()
-    else:
-        default_config_options(conanfile)
+    elif conanfile.implements:
```
memsharded:
```suggestion
    elif "auto_shared_fpic" in conanfile.implements:
         auto_shared_fpic_config_options(conanfile)
```
I think I'd have preferred a minimal, explicit implementation to start with this feature, and only generalize a bit later when there is more evidence.

memsharded: (But I am fine with it, not a big issue)

czoido: Yes, I did it this way in case we wanted to respect the order of execution stated in the list, in fact, I prefer the minimal, explicit implementation

conans/client/conanfile/implementations.py
memsharded: Yes, seems a bit too much of extra checking, for every recipe loading in the graph (hundreds of them), this will be repeated over and over.

I am fine with no error checking for the initial implementation

conans/client/conanfile/implementations.py
danimtb: I think is fine as is but would require a clear explanation in the docs

danimtb (approved):
I like it!

conans/client/conanfile/configure.py
memsharded: I'd say lets hardcode the order, we know it works. If in the future order-defined application of ``implements`` is really a thing... maybe they should be explicit in their methods.

Yeah, I think we shouldn't care about the user order at all, it would be another point of variability and issues.

memsharded (approved):

uilianries:
Just like I asked before, a magic attribute 😝

PR #14326, opened 2023-07-18
using conanfile.run() so environment is injected

### LINKED ISSUE #14297: [question] Should XCRun use the conanfile's build env?
### What is your question?

I realized that #14195 doesn't fully solve the issues I was having when doing a Linux -> macOS cross build, since XCRun still assumes that `xcrun` is in the ambient `PATH`. I can work around it without too much trouble by just doing:

```python
with VirtualBuildEnv(self).vars().apply():
```

but ideally the tool would apply the build env internally.

memsharded:
Changelog: Fix: Invoke XCRun using conanfile.run() so that environment is injected.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14297

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -1,4 +1,5 @@
 import os
+from io import StringIO
 
 from conans.util.runners import check_output_runner
 from conan.tools.build import cmd_args_to_string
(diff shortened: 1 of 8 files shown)


franramirez688 (approved):

memsharded (commented):
This change means that more logs of ``RUN: git ....`` and ``RUN: xcrun ...`` will be displayed in the user output, and maybe it is too much. 

Moving it to 2.0.10 and marking it as ``look-into``

czoido (approved):

PR #14347, opened 2023-07-21
Implement test_package ``--build-test=missing`` for full control of ``test_package`` deps

### LINKED ISSUE #14314: conan create may fail if test_package requires the tested reference in a different configuration
### What is your question?

For discussion / look into.

When running `conan create` to build a specific package (given input profiles, options, etc), there are some circumstances in which the package may be created, but `test_package` can fail, some examples:

* when the `test_package` has a `self.requires()` **and** a `self.tool_requires()` on `self.tested_reference_str`. In some circumstances this can lead to _two_ different package IDs being required for a successful test package, for example:
   * `--profile:build` and `--profile:host` are different when cross compiling - the test package will fail if the package for the _build_ profile does not previously exist. 
   * `-o "*:shared=True"` - the option only applies to the _host_ context, but not the _build_ context (the package ID resolved for the tool require) - here we are _NOT_ cross building but it would still fail, unless the same options and settings were passed to both build and host contexts.
   
* when the package is always used as a `tool_requires`, and the test package has `self.tool_requires()` on the tested reference - in this case, the options and settings will apply to what is being created, but a different package ID will be required when testing. In these instances, `--build-require` should be passed to the command line when invoking `conan create` - but this requires external knowledge that the tool is used as a tool requirement. 

For the first scenario, knowledge of this situation is required in order to derive a correct build order, or alternatively one may want to silence the test package (`-tf=""`). For the second scenario, one must know that the package is (exclusively?) used as a tool requirement to pass `--build-require` when invoking `conan create`.

- [x] I've read the CONTRIBUTING guide

### LINKED ISSUE #14352: [bug] Assertion error self.binary == node.binary
### Environment details

* Operating System+version: Macos 13.4.1
* Compiler+version: apple-clang 14.0.3
* Conan version: 2.0.9
* Python version: 3.11.4

### Steps to reproduce

Not completely sure, but happens sometimes when I `conan test` package with `--build missing` where multiple binaries should be build by dependency.

The usual workaround is to first `conan install --requires --build missing` those packages and then `conan test`. I didn't find a pattern or minimal reproducer. Hopefully the stack trace will help discovering the issue.

### Logs

Stack trace:

```
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/cli/cli.py", line 272, in main
    cli.run(args)
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/cli/cli.py", line 172, in run
    command.run(self._conan_api, args[0][1:])
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/cli/command.py", line 125, in run
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/cli/commands/test.py", line 40, in test
    deps_graph = run_test(conan_api, path, ref, profile_host, profile_build, remotes, lockfile,
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/cli/commands/test.py", line 74, in run_test
    conan_api.install.install_binaries(deps_graph=deps_graph, remotes=remotes)
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conan/api/subapi/install.py", line 20, in install_binaries
    installer.install_system_requires(deps_graph)  # TODO: Optimize InstallGraph computation
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conans/client/installer.py", line 196, in install_system_requires
    install_graph = InstallGraph(graph)
                    ^^^^^^^^^^^^^^^^^^^
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conans/client/graph/install_graph.py", line 190, in __init__
    self._initialize_deps_graph(deps_graph)
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conans/client/graph/install_graph.py", line 230, in _initialize_deps_graph
    existing.add(node)
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conans/client/graph/install_graph.py", line 129, in add
    install_pkg_ref.add(node)
  File "/Users/dodo/Desktop/conan2-venv/lib/python3.11/site-packages/conans/client/graph/install_graph.py", line 51, in add
    assert self.binary == node.binary
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError

ERROR:
```

memsharded:
Changelog: Feature: ``conan create --build-test=missing`` new argument to control what is being built as dependencies of the ``test_package`` folder.
Docs: https://github.com/conan-io/docs/pull/3336

Close https://github.com/conan-io/conan/issues/14314
Close https://github.com/conan-io/conan/issues/14352


czoido (approved):

jcar87 (commented):
I have tested this flag with a `test_package` like this:
```
    def requirements(self):
        self.requires(self.tested_reference_str)

    def build_requirements(self):
        self.tool_requires(self.tested_reference_str)
```

while invoking `conan create` in a way that results in both of these being satisfied by two different package IDs of the same reference, and this flag works perfectly fine and builds the missing reference.

However I still find it a bit counter intuitive, or perhaps it may need to be better documented

`conan create . --build=missing` will fail for the code below if this results in two different package IDs being required:

```
    def requirements(self):
        self.requires(self.tested_reference_str)

    def build_requirements(self):
        self.tool_requires(self.tested_reference_str)
```



But the following scenario, where:

* `foobar/123` is exported
* it is *NOT* a dependency of `self.tested_reference_str`  and only appears as a requirement of the test package
* there is currently no package generated for it

```
    def requirements(self):
        self.requires(self.tested_reference_str)
        self.requires("foobar/123")
```

In this scenario,

```
conan create . --build=missing
```

will currently succeed (even before this PR), and will generate the `missing` package for `foobar/1.2.3`. So what is difficult to reason, on documentation alone, is why `--build=missing` works for _some_ of the references required by the test package, but specifically not for the one being created. I am aware why, but perhaps it needs a clarification in the documentation. 

This can create confusion as to why `--build-test=missing` does that `--build=missing` doesn't, when passed as an argument to Conan create. I only see two useful scenarios:
* when one knows beforehand that there are multiple requirements on `self.tested_reference_str` 
and perhaps the more unusual but also valid:
* when one wants `conan create` to force a build of the reference, rather than reuse an existing package, but still wants it to build any missing dependencies of the test_package

memsharded:
Thanks very much for the thorough review and testing @jcar87 

I think the ``self.requires("foobar/123")`` behavior might be an overlook, you are right, it doesn't look like the most intuitive behavior, I will add some tests and check it.

memsharded:
I think it is working fine, adding an extra test to the PR that validates it:
```python
    def test_build_test_package_dep(self):
        c = TestClient()
        c.save({"dep/conanfile.py": GenConanfile("dep", "0.1"),
                "pkg/conanfile.py": GenConanfile("pkg", "0.1"),
                "pkg/test_package/conanfile.py": GenConanfile().with_requires("dep/0.1")
                                                               .with_test("pass")})
        c.run("export dep")
        c.run("create pkg --build=missing", assert_error=True)
        c.assert_listed_binary({"pkg/0.1": ("da39a3ee5e6b4b0d3255bfef95601890afd80709", "Build")})
        c.assert_listed_binary({"dep/0.1": ("da39a3ee5e6b4b0d3255bfef95601890afd80709", "Missing"),
                                "pkg/0.1": ("da39a3ee5e6b4b0d3255bfef95601890afd80709", "Cache")},
                               test_package=True)
```
This works, and raise an error because the binary is ``missing``.
Can you please double check? Is it possible that you got the binary in the cache built in a previous command?

jcar87 (approved):
LGTM! 

I think the behaviour changes makes sense, that is `--build=missing` now applies only to the graph of the reference being _created_, and `--build-test` missing applies to the graph of the test_package, which may contain additional references (if the test package has them).  

I think this makes sense with the premise that `conan create` has a premise where for the tested reference it should only build what is requested

franramirez688 (approved):

PR #14364, opened 2023-07-25
detect msvc and version when CC/CXX point to cl.exe

jcar87:
Changelog: Feature: `conan profile detect` can now detect the version of msvc when invoked within a Visual Studio prompt where `CC` or `CXX` are defined and pointing to the `cl` compiler executable
Docs: Omit

This PR fixes issue where `conan profile detect` is invoked from an active Visual Studio prompt that _also_ happens to have `CC` or `CXX` defined. This may be unusual for a developer to actively configure a prompt like this, but Visual Studio invoke may invoke CMake in such an environment, and this may cause issues if `conan profile detect` needs to run when using `cmake-conan`.

This assumes that `cl.exe /?`always prints the copyright and the version in the first line. It doesn't assume a specific text since it can be localized to different languages, but it does assume that the string contains `Microsoft` and that the version contains 3 or more components. We also avoid running this logic if `clang` is part of the string - e.g. `clang-cl` would not be handled and would fall back to the current behaviour. 

Tests cover defining `CC` in multiple ways, e.g. `cl` and `cl.exe` (falls back to finding it in PATH), or the full path to `cl.exe`, which may be surrounded in double quotes as the path is very likely to contain spaces if it's installed in the default location.

### DIFF
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -45,7 +45,7 @@ def _msvc_cl_compiler(compiler_exe="cl"):
         if not "Microsoft" in first_line:
             return None
         compiler = "msvc"
-        version_regex = re.search(r"(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.([0-9]+)\.?([0-9]+)?", out)
+        version_regex = re.search(r"(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.([0-9]+)\.?([0-9]+)?", first_line)
         if not version_regex:
             return None
         # 19.36.32535 -> 193
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -34,6 +34,17 @@ def _gcc_compiler(compiler_exe="gcc"):
             return compiler, installed_version
     except Exception:
         return None
+    
+def _msvc_cl_compiler(compiler_exe="cl"):
+    try:
+        ret, out = detect_runner(f"{compiler_exe} /?")
+        if ret != 0 or not "Microsoft" in out:
+            return None
+        compiler = "msvc"
+        full_version = re.search(r"([0-9]+)\.([0-9]+)\.([0-9]+)\.?([0-9]+)?", out).group()
+        
+    except Exception:
+        return None
 
 
 def _clang_compiler(compiler_exe="clang"):
@@ -96,6 +107,9 @@ def _get_default_compiler():
             return gcc
         if platform.system() == "SunOS" and command.lower() == "cc":
             return _sun_cc_compiler(command)
+        if platform.system() == "Windows" and command.endswith(("cl", "cl.exe")) and not "clang" in command:
+            vs = _msvc_cl_compiler(command)
+
         # I am not able to find its version
         output.error("Not able to automatically detect '%s' version" % command)
         return None
--- a/conans/client/conf/detect.py
+++ b/conans/client/conf/detect.py
@@ -37,12 +37,20 @@ def _gcc_compiler(compiler_exe="gcc"):
     
 def _msvc_cl_compiler(compiler_exe="cl"):
     try:
-        ret, out = detect_runner(f"{compiler_exe} /?")
-        if ret != 0 or not "Microsoft" in out:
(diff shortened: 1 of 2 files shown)


conans/client/conf/detect.py
```diff
@@ -34,6 +34,25 @@ def _gcc_compiler(compiler_exe="gcc"):
...
+    try:
+        compiler_exe = compiler_exe.strip('"')
+        ret, out = detect_runner(f'"{compiler_exe}" /?')
+        if ret != 0:
+            return None
+        first_line = out.splitlines()[0]
+        if not "Microsoft" in first_line:
+            return None
+        compiler = "msvc"
+        version_regex = re.search(r"(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.([0-9]+)\.?([0-9]+)?", out)
```
jcar87:
```suggestion
        version_regex = re.search(r"(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.([0-9]+)\.?([0-9]+)?", first_line)
```

conans/client/conf/detect.py
```diff
@@ -34,6 +34,25 @@ def _gcc_compiler(compiler_exe="gcc"):
...
+    try:
+        compiler_exe = compiler_exe.strip('"')
+        ret, out = detect_runner(f'"{compiler_exe}" /?')
+        if ret != 0:
+            return None
+        first_line = out.splitlines()[0]
+        if not "Microsoft" in first_line:
+            return None
+        compiler = "msvc"
+        version_regex = re.search(r"(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.([0-9]+)\.?([0-9]+)?", first_line)
```
memsharded, line 48: Maybe the last optional fourth group can just be dropped, no need to even match it?

conans/test/functional/command/profile_test.py
```diff
@@ -128,3 +128,26 @@ def test_profile_new(self):
...
+        cl_executable = c.out.splitlines()[-1]
+        assert os.path.isfile(cl_executable)
+        cl_location = os.path.dirname(cl_executable)
+        
+        # Try different variations, including full path and full path with quotes 
+        for var in ["cl", "cl.exe", cl_executable, f'"{cl_executable}"']:
+            output = RedirectedTestOutput()
+            with redirect_output(output):
+                with environment_update({"CC": var, "PATH": cl_location}):
+                    result = detect_defaults_settings()
```
memsharded: It this is a functional test, maybe do we want to run ``client.run("profile detect")`` instead?

memsharded (approved):
Looking good

conans/test/functional/command/profile_test.py
jcar87: I considered that - but I need to be able to influence the environment the `TestClient`  is exposed to with at least two environment variables, do we have convenience functionality for this, or would influencing via `os.environ` be sufficient? This test is similar to the `test_detect_default_in_mac_os_using_gcc_as_default` test above, which also needs environment variables.

conans/client/conf/detect.py
jcar87: I considered that - I added the last option groups since so far the version seems to always have 3 or 4 components - figured it's safer to get a match of the longest string that looks like a version? In case something ever changes in that output :/ but happy to drop this to two

memsharded: Not necessary, it is fine, I guess it shouldn't be a problem

conans/test/functional/command/profile_test.py
memsharded: I think that running ``client.run()`` under that same ``environment_update()`` should have the same behavior?

jcar87: Done!

memsharded (approved):
great!

Can we then move out of draft and merge it? Anything missing?

jcar87:
> great!
> 
> Can we then move out of draft and merge it? Anything missing?

Oops, should have taken it out of draft already :D will merge once CI reports a pass (I think changing it to not-a-draft restarted a build).

PR #14392, opened 2023-07-28
Feature/credentials json

### LINKED ISSUE #14354: [feature] Add authentication option with .netrc file
### What is your suggestion?

Using `.netrc` file for authentication will complement current options:
* When using Git LFS with Artifactory, Git LFS already uses `.netrc`, so we can keep credentials in one place.
* Can avoid secret tokens in environment variables. Environment variables are more prone to exposure compared to the dot-files.

### LINKED ISSUE #14525: Where to put credentials for CI?
### What is your question?

Seeing a few recent Issues around this issue and a few things in work, figure I'd just start an issue to figure out the current state of things and concretely express our need since it seems there's still some [ambiguity ](https://github.com/conan-io/conan/pull/14392/files#r1278897465) of what users need. In this case I'm one of the maintainers of our internal build pipeline processes & private registries (Artifactory) and don't use Conan at all myself. Our users want to use Conan in their projects so in our CI pipelines we really need a specific, configurable, spot on the filesystem to put credentials for our private registries for build processes using Conan for each build. That's in line with most other package managers out there - pip, docker, helm, & npm all support that in some fashion. They all have a credentials file that defaults to a user scoped directory of some kind, and an environment variable to override that location. Generally what we're doing is we have common tooling across our organization that installs appropriate credentials to the appropriate environments for any specific build running on our CI platform, just need to know where to put them for a given package manager. We also may not have access to the package manager binary itself yet at this stage since providing that is the responsibility of build steps further down the chain and separate from our steps that furnish credentials, so running a conan CLI command to reset them every run would be less-than ideal too.

Seems the existing way for Conan is to pass credentials in is with environment variables. We strongly discourage putting any credentials in environment variables - we see those get picked and stored up by logging, debugging, and profiling tools way more often than files on disk. Same with the CLI '-p' option which is also seems Conan supports, command lines are ridiculously leaky and love to be captured by build log files.

I'm seeing [docs ](https://docs.conan.io/2/reference/config_files/source_credentials.html) about a source_credentials.json file. Looks like that's in the "Conan cache"(?) Need some docs or links on that page explaining what that is and if it's configurable. Seems like it's currently in the conan home directory? Then I'm seeing the recent ["Feature/credentials json"](https://github.com/conan-io/conan/pull/14392) PR. Sorta looks like that's revising source_credentials.json ?

Reading through #7793 is scaring me too. Conan should absolutely be using credentials specified by the user, loaded on every invocation. The user shouldn't need to be aware that it's caching things, or that apparently some Conan servers use JWT tokens. Tokens should be opaque strings for all practical purposes to appropriately support other private registry implementations, which practically has to be done to support users that need to run private registries. Worst case maybe an environment variable to ignore cache could be added?

memsharded:
Changelog: Feature: New ``credentials.json`` file to store credentials for Conan remotes.
Docs: https://github.com/conan-io/docs/pull/3350

Close https://github.com/conan-io/conan/issues/14354
Close https://github.com/conan-io/conan/issues/14525

### DIFF
--- a/conans/client/rest/remote_credentials.py
+++ b/conans/client/rest/remote_credentials.py
@@ -1,0 +1,65 @@
+import json
+import os
+import platform
+
+from jinja2 import Template
+
+from conan.api.output import ConanOutput
+from conans.client.userio import UserInput
+from conans.errors import ConanException
+from conans.util.files import load
+
+
+class RemoteCredentials:
+    def __init__(self, cache):
+        self._cache = cache
+        self._urls = {}
+        creds_path = os.path.join(cache.cache_folder, "credentials.json")
+        if not os.path.exists(creds_path):
+            return
+        template = Template(load(creds_path))
+        content = template.render({"platform": platform, "os": os})
+        content = json.loads(content)
+
+        self._urls = {credentials["remote"]: {"user": credentials["user"],
+                                              "password": credentials["password"]}
+                      for credentials in content["credentials"]}
+
+    def auth(self, remote, user=None, password=None):
+        if user is not None and password is not None:
+            return user, password
+
+        # First prioritize the cache "credentials.json" file
+        creds = self._urls.get(remote)
+        if creds is not None:
+            try:
+                return creds["user"], creds["password"]
+            except KeyError as e:
+                raise ConanException(f"Authentication error, wrong credentials.json: {e}")
+
+        # Then, check environment definition
+        env_user, env_passwd = self._get_env(remote, user)
+        if env_passwd is not None:
+            if env_user is None:
+                raise ConanException("Found password in env-var, but not defined user")
+            return env_user, env_passwd
+
+        # If not found, then interactive prompt
+        ui = UserInput(self._cache.new_config.get("core:non_interactive", check_type=bool))
+        input_user, input_password = ui.request_login(remote, user)
+        return input_user, input_password
+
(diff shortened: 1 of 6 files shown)


conans/client/rest/remote_credentials.py
```diff
@@ -0,0 +1,65 @@
...
+from conans.client.userio import UserInput
+from conans.errors import ConanException
+from conans.util.files import load
+
+
+class RemoteCredentials:
+    def __init__(self, cache):
+        self._cache = cache
+        self._urls = {}
+        creds_path = os.path.join(cache.cache_folder, "credentials.json")
```
franramirez688, line 17: Let's add a `conf` to let users define this file wherever they want. This value could be the default one for sure.

franramirez688 (commented):
@memsharded This is great! 👏
I'm missing some unit tests testing purely the new `RemoteCredentials` class too. Wdyt?

conans/client/rest/remote_credentials.py
memsharded: Better wait for users to request it? I think it is better to focus on the integration tests, and see if they could miss some important use cases to test, think if there are risks involved, etc.

franramirez688 (approved):

AbrilRBS (approved):

czoido (approved):

PR #14397, opened 2023-07-30
cmake: add support for cli_args in install()

### LINKED ISSUE #14235: Add cli_args argument to cmake.install()
> > if self.settings.build_type == 'Release':
> 
> Ok, makes sense, though nowadays, I'd expect that `Release` builds already don't contain any debug information to be stripped, I understand it might not be the case for all platforms and tools, so I am ok with making this a opt-in conf.

From https://discourse.cmake.org/t/what-does-strip-do/4302 and if I'm not mistaken, it's not only debug information that gets stripped (if there is any), but also symbols that don't need to be externally visible, so this should have a benefit.

PR looks good to me, thanks @sagi-ottopia  for your contribution!!!

@memsharded  one thing I'd consider outside of this PR is adding a `cli_args` argument to `cmake.install()` - to mirror what we have for `cmake.configure()` and `cmake.build()` - while the current setup for install pretty much covers the bases, it may be useful for some advanced users to have the added flexibility

_Originally posted by @jcar87 in https://github.com/conan-io/conan/pull/14167#pullrequestreview-1516018103_

mkmkme:
This commit fixes #14235

Changelog: Feature: Add ``cli_args`` argument for ``CMake.install()``.
Docs: https://github.com/conan-io/docs/pull/3314

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -163,7 +163,7 @@ class CMake(object):
         self._conanfile.output.info("Running CMake.build()")
         self._build(build_type, target, cli_args, build_tool_args)
 
-    def install(self, build_type=None, component=None):
+    def install(self, build_type=None, component=None, cli_args=None):
         """
         Equivalent to run ``cmake --build . --target=install``
 
@@ -172,6 +172,9 @@ class CMake(object):
                            It can fail if the build is single configuration (e.g. Unix Makefiles),
                            as in that case the build type must be specified at configure time,
                            not build type.
+        :param cli_args: A list of arguments ``[arg1, arg2, ...]`` for the underlying
+                         build system that will be passed to the command line:
+                         ``cmake --install ... arg1 arg2``
         """
         self._conanfile.output.info("Running CMake.install()")
         mkdir(self._conanfile, self._conanfile.package_folder)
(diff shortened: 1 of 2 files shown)


conan/tools/cmake/cmake.py
```diff
@@ -193,6 +196,14 @@ def install(self, build_type=None, component=None):
         if do_strip:
             arg_list.append("--strip")
 
+        if cli_args:
+            if "--install" in cli_args:
+                raise ConanException("Do not pass '--install' argument to 'install()'")
+            if "--strip" in cli_args:
```
memsharded: I am not sure about this warning, it should either be normally allowed or raise an error, but from my experience this kind of warning doesn't help much as it pass unnoticed, or it annoys the user that actually wants to do that.

memsharded (commented):
Looking good, just one minor comment about the warning

conan/tools/cmake/cmake.py
mkmkme: Sure :) The rationale for that warning was that I wanted the user to be aware of that feature. It wasn't even documented, and I got to know about it myself while reading the code.

I removed the warning. The configuration mentioning in the docs is added separately at https://github.com/conan-io/docs/pull/3314

memsharded (approved):

PR #14405, opened 2023-08-01
Add format 'cyclonedx' as output format for the graph command

### LINKED ISSUE #9787: [feature] Add support for CycloneDX output to the `info` command
Hi, full disclosure, I work on the OWASP CycloneDX software bill of materials project https://cyclonedx.org/. 

A software bill of materials, in a very basic sense, is a list of all the components that make up a piece of assembled software.

The current output of the `info` command provides this information.

But that output format is specific to the Conan ecosystem. CycloneDX is an ecosystem agnostic format.

Having this information in a standard format helps with a number of use cases. Identifying components with known vulnerabilities, OSS license compliance, and supply chain component analysis, as examples. Especially in organisations that are building software in multiple ecosystems.

I had been planning to implement a specific tool to do this. But after digging into it, I think it might be better to add it as an additional output format for the `info` command. Perhaps a `--cyclonedx-json` option?

If this is something you'd accept a PR for let me know and I'll start working on the implementation and tests.

hedtke:
Changelog: Feature: Add format 'cyclonedx' as output format for the graph command

Close https://github.com/conan-io/conan/issues/9787

The new output format `cyclonedx` for the graph command creates a SBOM in CycloneDX JSON 1.4 format to be parsed by dependency management / vulnerability tools like https://www.dependencytrack.org/

### DIFF
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -144,3 +144,44 @@ def format_graph_json(result):
     cli_out_write(json_result)
     if graph.error:
         raise graph.error
+
+
+def format_graph_cyclonedx(result):
+    """
+    # creates a CycloneDX JSON according to https://cyclonedx.org/docs/1.4/json/
+    """
+    def licenses(conanfilelic):
+        def entry(id):
+            return {"license": {
+                "id": id
+            }}
+        if conanfilelic is None:
+            return []
+        elif isinstance(conanfilelic, str):
+            return [entry(conanfilelic)]
+        else:
+            return [entry(i) for i in conanfilelic]
+
+    result["graph"].serialize()  # fills ids
+    deps = result["graph"].nodes[1:]  # first node is app itself
+    cyclonedx = {
+        "bomFormat": "CycloneDX",
+        "specVersion": "1.4",
+        "version": 1,
+        "dependencies": [n.id for n in deps],
+        "components": [
+            {
+                "type": "library",
+                "bom-ref": n.id,
+                "purl": n.package_url().to_string(),
+                "licenses": licenses(n.conanfile.license),
+                "name": n.name,
+                "version": n.conanfile.version,
+                "supplier": {
+                    "url": [n.conanfile.homepage]
+                }
+            } for n in deps
+        ]
+    }
+    json_result = json.dumps(cyclonedx, indent=4)
+    cli_out_write(json_result)
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
 
+from packageurl import PackageURL
+
 from conans.model.package_ref import PkgReference
 from conans.model.recipe_ref import RecipeReference
 
@@ -228,6 +230,29 @@ class Node(object):
         result["context"] = self.context
         result["test"] = self.test
         return result
+
+    def package_url(self):
+        """
+        Creates a PURL following https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst
+        """
+        qualifiers = {
+            "prev": self.prev
+        }
+        if self.ref.user:
+            qualifiers["user"] = self.ref.user
+        if self.ref.channel:
+            qualifiers["channel"] = self.ref.channel
+        if self.ref.revision:
+            qualifiers["rref"] = self.ref.revision
+        if self.remote:
+            qualifiers["repository_url"] = self.remote
+        else:
+            qualifiers["repository_url"] = "https://center.conan.io"
+        return PackageURL(
+            type="conan",
+            name=self.name,
+            version=str(self.ref.version),
+            qualifiers=qualifiers)
 
     def overrides(self):
 
--- a/conans/requirements.txt
+++ b/conans/requirements.txt
@@ -7,3 +7,4 @@ fasteners>=0.15
 distro>=1.4.0, <=1.8.0; sys_platform == 'linux' or sys_platform == 'linux2'
 Jinja2>=3.0, <4.0.0
 python-dateutil>=2.8.0, <3
+packageurl-python>=0.10.0
--- a/conan/cli/commands/graph.py
+++ b/conan/cli/commands/graph.py
@@ -4,7 +4,7 @@ from conan.api.output import ConanOutput, cli_out_write, Color
 from conan.cli import make_abs_path
 from conan.cli.args import common_graph_args, validate_common_graph_args
 from conan.cli.command import conan_command, conan_subcommand
-from conan.cli.formatters.graph import format_graph_html, format_graph_json, format_graph_dot
+from conan.cli.formatters.graph import format_graph_html, format_graph_json, format_graph_dot, format_graph_cyclonedx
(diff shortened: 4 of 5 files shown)


hedtke:
Hi, this is my first PR. I am not sure whether I work on the correct branches

hedtke:
The resulting JSON can be parsed by dependencytrack. Here is an example: 
image

conans/requirements.txt
```diff
@@ -7,3 +7,4 @@ fasteners>=0.15
 distro>=1.4.0, <=1.8.0; sys_platform == 'linux' or sys_platform == 'linux2'
 Jinja2>=3.0, <4.0.0
 python-dateutil>=2.8.0, <3
+packageurl-python>=0.10.0
```
memsharded, line 10: We are avoiding adding new dependencies like this one, unless it is extremely necessary. Every additional Python dependency has proven to be a liability (the state of dependency management in Python is not great), so we prefer to avoid this.

memsharded (commented):
Thanks for your contribution @hedtke 

At this moment we are avoiding this kind of specialized formats as built-in, and we are implementing them as extensions in https://github.com/conan-io/conan-extensions (thanks to ``conan config install`` it is very easy to use them).

No need to do any changes, we will check this, investigate a bit and discuss the approaches and let you know. Thanks!

tsondergaard:
I hope this PR is approved. Every organization have to deal with SBOM generation and having it built in just makes it easier for all of us.

memsharded (changes requested):
We have had a look at this, and the main problem is that the full cycloneDX specification is way more complete than the currently proposed one: https://cyclonedx.org/specification/overview/

This means that this is very likely to evolve, and we would like Conan to create a more complete one, and this cannot be implemented in a "formatter", it is too much responsibility for it. Having it in the core as built-in is also problematic for faster evolution.

So my recommendation is to move it to the ``conan-extensions`` repository, it fits much better there. There is already a very similar functionality which is the creation of "BuildInfo" (another type of SBOM).

We can discuss there what is the best approach for it, if it is a custom command or a deployer. I'd say try with a custom command first. We also need to think about the "Location" where a potential new command like this belongs, the command naming too, etc.

conan/cli/formatters/graph/graph.py
```diff
@@ -144,3 +144,44 @@ def format_graph_json(result):
...
+            return []
+        elif isinstance(conanfilelic, str):
+            return [entry(conanfilelic)]
+        else:
+            return [entry(i) for i in conanfilelic]
+
+    result["graph"].serialize()  # fills ids
+    deps = result["graph"].nodes[1:]  # first node is app itself
+    cyclonedx = {
+        "bomFormat": "CycloneDX",
```
jkowalleck, line 168: would you add the schema to the json, so its more clear how to validate the result? 
```py
  "$schema": "http://cyclonedx.org/schema/bom-1.4.schema.json",
```

conan/cli/formatters/graph/graph.py
```diff
@@ -144,3 +144,44 @@ def format_graph_json(result):
...
+        "components": [
+            {
+                "type": "library",
+                "bom-ref": n.id,
+                "purl": n.package_url().to_string(),
+                "licenses": licenses(n.conanfile.license),
+                "name": n.name,
+                "version": n.conanfile.version,
+                "supplier": {
+                    "url": [n.conanfile.homepage]
```
jkowalleck, line 181: homepage - why not `externalReferences` of type `website`? 
see https://cyclonedx.org/use-cases/#external-references

conans/client/graph/graph.py
```diff
@@ -229,6 +231,29 @@ def serialize(self):
         result["test"] = self.test
         return result
 
+    def package_url(self):
+        """
+        Creates a PURL following https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst
```
jkowalleck, line 236: read the spec: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#conan

conan/cli/formatters/graph/graph.py
```diff
@@ -144,3 +144,44 @@ def format_graph_json(result):
...
+        else:
+            return [entry(i) for i in conanfilelic]
+
+    result["graph"].serialize()  # fills ids
+    deps = result["graph"].nodes[1:]  # first node is app itself
+    cyclonedx = {
+        "bomFormat": "CycloneDX",
+        "specVersion": "1.4",
+        "version": 1,
+        "dependencies": [n.id for n in deps],
```
jkowalleck, line 171: This code looks wrong, according to CycloneDX spec.
Could you provide an example SBOM generated with your code, so we can check if the JSON results is actually valid according to <http://cyclonedx.org/schema/bom-1.4.schema.json>

memsharded:
@hedtke it seems the review from @jkowalleck is also supporting our conclusions above: this would be better as an "extension" where all these matters are way easier to fix and evolve.

conans/client/graph/graph.py
hedtke: Do you want me to change the comment, or is there an error in how I build the JSON?

conan/cli/formatters/graph/graph.py
hedtke: What is the advantage of using that over supplier? external references are not part of the BOM: https://cyclonedx.org/docs/1.4/json/#components_items_supplier VS https://cyclonedx.org/docs/1.4/json/#components_items_externalReferences

conan/cli/formatters/graph/graph.py
hedtke: Thanks, that is not correct. I will fix it in the conan extension

conans/client/graph/graph.py
jkowalleck: wanted to put a more precise link here, so nobody needs to scroll trough the whole document.

conan/cli/formatters/graph/graph.py
jkowalleck: > What is the advantage of using that over supplier

It's a matter of correctness.
One is the URL to the homepage of a supplier(or contractor, legal proxy, etc),
the other is the URL to the homepage of the component/project.

From what I've seen, `n.conanfile.homepage` contains the project's homepage, not the homepage of an author or supplier.
see https://docs.conan.io/2/reference/conanfile/attributes.html?highlight=homepage#homepage

I found that `n.conanfile.url` could be the `externalReference` of type `vcs`
see https://docs.conan.io/2/reference/conanfile/attributes.html?highlight=url#url


Linking a component's metadata and URLs via `externalReferences` is the intended way.
see the many examples here: https://cyclonedx.org/use-cases/

hedtke: `conanfile.url` points to the GitHub repo of conan-center. Not sure that is useful in this case.

I will use the homepage attribute and move it to `externalReferences`

hedtke:
Here is the "extension" version: https://github.com/conan-io/conan-extensions/pull/66

memsharded:
Ok, thanks very much, lets follow up everyone in the extensions repo: https://github.com/conan-io/conan-extensions/pull/66

PR #14407, opened 2023-08-01
Unified mocks definition (boilerplate)

franramirez688:
Changelog: omit
Docs: omit

Fixed apple_test.py and unified both classes (actually the same)

### DIFF
--- a/conans/test/functional/toolchains/meson/test_cross_compilation.py
+++ b/conans/test/functional/toolchains/meson/test_cross_compilation.py
@@ -7,7 +7,7 @@ import pytest
 
 from conan.tools.apple.apple import _to_apple_arch, XCRun
 from conans.test.assets.sources import gen_function_cpp, gen_function_h
-from conans.test.utils.mocks import MockConanfile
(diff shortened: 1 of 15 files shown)


memsharded (commented):
As you can see, it is not that simple to unify the Mocks, and there are 45 tests broken: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuitev2/detail/PR-14407/1/pipeline/35

I like the effort, and I'd really like to unify these mocks, this is a great initiative, I love it, but I'd suggest to fix first the broken ``release/2.0`` job, then try to unify these mocks.

franramirez688:
> As you can see, it is not that simple to unify the Mocks, and there are 45 tests broken: https://ci.conan.io/blue/organizations/jenkins/ConanTestSuitev2/detail/PR-14407/1/pipeline/35

Yes, but it was only one module. IMO, I think it can be done in one shot because that test broken was due to bad usage of those mocks (mixing both classes). Let's see if it passes now.

memsharded (approved):
Fantastic job!!

PR #14423, opened 2023-08-03
cmaketoolchain: add user_presets_path config

mkmkme:
CMake documentation [1] states:

	If `CMakePresets.json` and `CMakeUserPresets.json` are both present,
	`CMakeUserPresets.json` implicitly includes `CMakePresets.json`,
	even with no include field, in all versions of the format.

	If a preset file contains presets that inherit from presets in another
	file, the file must include the other file either directly or
	indirectly. Include cycles are not allowed among files. If `a.json`
	includes `b.json`, `b.json` cannot include `a.json`.

Since conan generates `CMakeUserPresets.json` by default, it's not possible to inherit from the presets specified by Conan by default. This can be changed by settings `CMakeToolchain.user_presets_path` to a custom value. However, for some project this single line might be the only reason to switch from `conanfile.txt` to `conanfile.py`.

This commit adds new configuration
`tools.cmake.cmaketoolchain:user_presets_path` so that `CMakeToolchain.user_presets_path` can be changed via command-line argument.

[1]: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#includes

Changelog: (Feature): Added config `tools.cmake.cmaketoolchain:user_presets_path` to change the user presets path via command-line argument.

### DIFF
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -184,6 +184,8 @@ class CMakeToolchain(object):
         if toolchain_file is None:  # The main toolchain file generated only if user dont define
             save(os.path.join(self._conanfile.generators_folder, self.filename), self.content)
             ConanOutput(str(self._conanfile)).info(f"CMakeToolchain generated: {self.filename}")
+        user_presets_path = self._conanfile.conf.get("tools.cmake.cmaketoolchain:user_presets_path",
+                                                     self.user_presets_path)
         # If we're using Intel oneAPI, we need to generate the environment file and run it
         if self._conanfile.settings.get_safe("compiler") == "intel-cc":
             IntelCC(self._conanfile).generate()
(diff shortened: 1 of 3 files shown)


memsharded (commented):
Hi @mkmkme 

I think this is not really a configuration item, but a workaround to avoid writing a ``conanfile.py``.

Our general design guideline is that when something really doesn't work unless you define a ``conf``, then it is not a ``conf`` and it should be written in the recipe. In this case if the user presets are going to inherit the Conan generated presets, it is absolutely mandatory that Conan generates the filename the user presets is expecting. It is not something configurable, it is hardcoded in the project, and it breaks if the user doesn't provide the configuration.

The strong recommendation here is to switch to ``conanfile.py``. As explained in https://docs.conan.io/2/tutorial/consuming_packages/the_flexibility_of_conanfile_py.html, the idea is that ``conanfile.txt`` is a format mainly for beginners and very simple use cases, but it is expected that in lots of cases users will do a ``conanfile.py`` for many different reasons. So forcing a new ``conf`` to avoid this doesn't sound to fit the current design.

Thanks very much for your contribution anyway!

mkmkme:
Alright, thanks for your feedback!

Issue #10341, opened 2022-01-14
[feature] Make the generation order of .pc files deterministic for PkgConfigDeps and pkg_config


czoido:
After the discussion [here](https://github.com/conan-io/conan-center-index/pull/8716#discussion_r782406117) and [here](https://github.com/conan-io/conan-center-index/pull/8623#discussion_r778436720) we have decided to provide as a feature the possibility to force one component to overwrite the `.pc` file of the main root cpp_info and leave the user to take responsibility for this behavior. So, if for example a recipe defines this cpp_info:

```python
    def package_info(self):
        self.cpp_info.set_property("pkg_config_name", "samplerate")
        self.cpp_info.components["onecomponent"].set_property("pkg_config_name", "samplerate")
```

When we install this package with `pkg_config` or `PkgConfigDeps` generator, the `.pc` file that will be generated is the one corresponding to the component.

PR #14428, opened 2023-08-04
Fix pkgconfig name collision on Windows

domin144:
Changelog: (Bugfix): Fix pkgconfig name collision on Windows

This fix is related to https://github.com/conan-io/conan/issues/10341. The previous fix did not take into account cases insensitive filesystems. As a result the issue was still present on Windows. OpenEXR 2.x is an example of affected recipe. The "OpenEXR.pc" file for the "openexr_ilmimf" component was overwritten by "openexr.pc" created based on package name.

### DIFF
--- a/conan/tools/gnu/pkgconfigdeps.py
+++ b/conan/tools/gnu/pkgconfigdeps.py
@@ -336,7 +336,7 @@ class _PCGenerator:
         # if it does not already exist in components one
         # Issue related: https://github.com/conan-io/conan/issues/10341
         pkg_name = _get_package_name(self._dep, self._build_context_suffix)
-        if f"{pkg_name}.pc" not in pc_files:
+        if not f"{pkg_name}.pc".lower() in [x.lower() for x in pc_files]:
             package_info = _PCInfo(pkg_name, pkg_requires, f"Conan package: {pkg_name}",
                                    self._dep.cpp_info, _get_package_aliases(self._dep))
             # It'll be enough creating a shortened PC file. This file will be like an alias
--- a/conans/test/functional/toolchains/gnu/test_pkgconfigdeps.py
+++ b/conans/test/functional/toolchains/gnu/test_pkgconfigdeps.py
@@ -113,3 +113,68 @@ def test_pkgconfigdeps_bindir_and_meson():
     # Executing directly "meson test" fails if the bindir field does not exist
     client.run_command("meson test -C test_package/build-release")
     assert "1/1 ./src/example OK"
+
+
+@pytest.mark.tool("meson")
+@pytest.mark.tool("pkg_config")
+def test_pkgconfigdeps_pc_name_conflict():
+    """
+    This test checks that a pc file provided by component is not overwritten by
+    conan with pc file named after the package if the names collide.
+    Cases insensitive filesystems have to be taken into account.
+    """
+    client = TestClient()
+    client.run("new meson_lib -d name=hello -d version=1.0")
+    client.run("create . -tf=\"\"")
+    conanfile = textwrap.dedent("""
+    import os
+    from conan import ConanFile
+    from conan.tools.meson import MesonToolchain, Meson
+    from conan.tools.layout import basic_layout
+    from conan.tools.files import copy
+
+    class helloConan(ConanFile):
+        name = "hello"
+        version = "1.0"
+        package_type = "library"
+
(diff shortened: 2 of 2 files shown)


memsharded (commented):
Thanks very much for the contribution @domin144 

It would be great to start the PR with a failing "red" test failing for this bug, and then see it fixed with this contributed fix. Tests are necessary anyway, so having them from the start help a lot understanding the fixes.

domin144:
I am having some trouble running the tests. For now I wrote some test, which tries to reproduce an example I tried manually. I did not run the test though. The important lines are:
```
        self.cpp_info.components["component"].set_property("pkg_config_name", "Hello")
        self.cpp_info.components["component"].libs = ["hello"]
```
This is where conflicting `.pc` file comes from. Notice the upper case `H` i `hello`. `Hello.pc` created for this component will be overwritten on Windows with the `hello.pc` created based on the package name.

When I run the example, I had to add `tool_requires = "meson/[>=1.1.0]", "pkgconf/[>=1.9]"` to `test_package\conanfile.py`, but I can see other tests don't to this, so I assume that in the testing environment it is not required. I didn't put this in the test. I attach the example.

BTW: where do you normally get `pkg-config` on Windows to use with meson + visual studio?

Please bear with me. This is all new to me. I did not use conan before and I don't have much experience with python. I usually don't develop on windows also and hoped conan will make a temporary transition easier for me.

I assume you have a working setup, where you are able to run test. I would appreciate, if you tried to run this test and let me know if it passes with the fix and fails without it.

memsharded:
> tool_requires = "meson/[>=1.1.0]", "pkgconf/[>=1.9]"

yes, it is not possible to download things from ConanCenter in testing. They need to be installed in the machine, and injected via ``@pytest.mark.tool("...")`` annotations, that work with the definition of ``conftest.py`` and ``conftest_user.py``

> BTW: where do you normally get pkg-config on Windows to use with meson + visual studio?

there is no standard location, this is why we are using the ``conftest.py`` and ``conftest_user.py`` files to define it.

> Please bear with me. This is all new to me. I did not use conan before and I don't have much experience with python. I usually don't develop on windows also and hoped conan will make a temporary transition easier for me.

Yes, thanks very much for your contribution and putting the effort! 🙂 I think @franramirez688 can help the following days with the information you provided in your comment and draft some test that covers it.

franramirez688:
Hi @domin144 ,

Thanks for raising this problem.

I'm not sure if this is the best way to proceed with this error. Let me talk to the team because I'd like a more generic solution. This one sounds like a breaking change for systems with case-sensitive file systems, so let's see if we could get another approach or even if this is the best one.

domin144:
Hi @franramirez688,

I think the OpenEXR recipe tried to reproduce the pc file name, provided by OpenEXR itself (https://github.com/AcademySoftwareFoundation/openexr/blob/d688d05d99dc54f6d928c4e6d6ceee53b3368267/cmake/OpenEXR.pc.in).

Any application prepared to use the original `OpenEXR.pc` file, will not be broken by absence of the additional `openexr.pc` file present only when package is provided by conan. It's hard for me to imagine a case, where creating two pc files with names differing only by letter cases is intentional.

With the current, case sensitive match, we have a discrepancy between platforms: on case sensitive file systems two files are created, on case insensitive filesystems the first one is overwritten and only second is available. Case ignoring match makes the procedure consistent.

Anyway, I checked the `_get_package_name` function and I can see the default pc file name can be overridden, by changing `pkg_config_name` property of the `cpp_info` object. This means the issue with OpenEXR could be worked around in the recipe. I will try to make such workaround, but I still think conan should somehow prevent the overwrite of the file or at least provide a warning.

franramirez688:
Hi @domin144 ,

Thanks for your feedback!

We were checking this issue, and we think that the best approach would have these two points:

* First, `PkgConfigDeps` should have a mechanism to define `set_property("pkg_config_name", "none")` as a placeholder to avoid creating those `*.pc` files.
* Then, modify the recipe and add this `"none"` one to the main OpenEXR's `cpp_info` to make that possible and avoid overwriting the component one.

We still think that this change would bring breaking changes for several recipes so we prefer to implement those ones instead, and keep that check as case-sensitive.

domin144:
Great!
I am happy you have a concept on how to resolve the issue. I am looking forward to the implementation.

PR #14466, opened 2023-08-11
Implement conf that avoids binary skipping in the graph

AbrilRBS:
Changelog: Feature: Add `tools.graph:skip_binaries` to control binary skipping in the graph.
Docs: https://github.com/conan-io/docs/pull/3342
Docs: https://github.com/conan-io/docs/pull/3341

Tests missing, will asign someone once I write them up :)

Motivation comes from https://github.com/conan-io/conan-extensions/issues/65 (Pinging @fschoenm) even though this was a known issue before that. Another alternative for that specific issue is listed in its comments, but this change stands on its own either way.

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -23,8 +23,6 @@ BUILT_IN_CONFS = {
     "core.download:retry_wait": "Seconds to wait between download attempts from Conan server",
     "core.download:download_cache": "Define path to a file download cache",
     "core.cache:storage_path": "Absolute path where the packages and database are stored",
-    # TODO. Find better name
-    "core.graph:skip_binaries": "Lorem Ipsum",
     # Sources backup
     "core.sources:download_cache": "Folder to store the sources backup",
     "core.sources:download_urls": "List of URLs to download backup sources from",
@@ -71,6 +69,8 @@ BUILT_IN_CONFS = {
     "tools.deployer:symlinks": "Set to False to disable deployers copying symlinks",
     "tools.files.download:retry": "Number of retries in case of failure when downloading",
     "tools.files.download:retry_wait": "Seconds to wait between download attempts",
+    # TODO. Find better name
+    "tools.graph:skip_binaries": "Lorem Ipsum",
     "tools.gnu:make_program": "Indicate path to make program",
     "tools.gnu:define_libcxx11_abi": "Force definition of GLIBCXX_USE_CXX11_ABI=1 for libstdc++11",
     "tools.gnu:pkg_config": "Path to pkg-config executable used by PkgConfig build helper",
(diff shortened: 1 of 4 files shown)


memsharded (commented):
Looks good.
I think this would deserve some tests or some extra checks, like adding this to some of the already existing functional tests, to validate things are not breaking. The idea is validate that even with not skipping binaries, linkage and usage of packages is exactly the same with generators like CMakeDeps, CMakeToolchain and VirtualBuild/RunEnv.

fschoenm:
@RubenRBS Thanks for implementing something so quickly, that seems very helpful :)

What effect does not skipping packages have on the build process? Does it just add e.g. CMake configs for packages that would usually be skipped or are there any other side effects?

What happens with requirements for a `tool_require`, that is already built? At least those would usually not be required for my use case (OSS license handling). Is it worth it to handle different scenarios like this or is not skipping packages an all-or-nothing thing?

memsharded:
Thanks for the comments @fschoenm 

I think those make some sense, and I am thinking @RubenRBS that maybe ``tools.graph:skip_binaries`` could be beneficial:
- Easy, clear and known separation between host and build context. For licenses the difference is very clear
- It can be applied per-package if necessary like ``*@:tools.graph:skip_binaries`` to apply to all packages from ConanCenter (without user/channel)

conans/model/conf.py
```diff
@@ -69,6 +69,8 @@
     "tools.deployer:symlinks": "Set to False to disable deployers copying symlinks",
     "tools.files.download:retry": "Number of retries in case of failure when downloading",
     "tools.files.download:retry_wait": "Seconds to wait between download attempts",
+    # TODO. Find better name
+    "tools.graph:skip_binaries": "Lorem Ipsum",
```
memsharded: Yeah, probably we want to negate it and be ``force_install=True``, ``dont_skip=True``?
I don't oppose to that name, it is not that bad ``tools.graph:skip_binaries`` doesn't sound bad.

memsharded (approved):

conans/model/conf.py
AbrilRBS: Let's go with the already present one then :)

conans/model/conf.py
```diff
@@ -69,6 +69,7 @@
     "tools.deployer:symlinks": "Set to False to disable deployers copying symlinks",
     "tools.files.download:retry": "Number of retries in case of failure when downloading",
     "tools.files.download:retry_wait": "Seconds to wait between download attempts",
+    "tools.graph:skip_binaries": "Allow binary skipping in the graph",
```
czoido: I think it would be nice to complete the description mentioning what the default value for this configuration is.

AbrilRBS:
@fschoenm 
> What happens with requirements for a tool_require, that is already built? At least those would usually not be required for my use case (OSS license handling). Is it worth it to handle different scenarios like this or is not skipping packages an all-or-nothing thing?

The conf can be specified just for build-context packages. The idea would be to do something like:
`conan install --requires fast-dds/2.10.1 --deployer=licenses --conf:host "*:tools.graph:skip_binaries=False" --conf:build "*:tools.graph:skip_binaries=True"`

This would force host-context packages not to be skipped, and build-context ones to be allowed to skip (But won't get skipped if they are needed in the install, you are the one that has to ensure that first!)

(Or write the conf with the correct value to each of the corresponding profiles)

czoido (approved):

PR #14467, opened 2023-08-11
Refactor ugly class attributes into parameters

AbrilRBS:
Changelog: Omit
Docs: Omit

Leaving the place a bit prettier than when I came across it when working on #14466


memsharded (approved):
Fantastic, many thanks!

PR #14478, opened 2023-08-14
Fix 'conan remove': output only confirmed entries

Tobulus:
Changelog: Bugfix: 'conan remove' was outputting all entries in the cache matching the filter not just the once which where confirmed by the user.
Docs: Omit


memsharded (commented):
Thanks for the fix @Tobulus, it is looking good.
It would be necessary to add a unit test, to guarantee that this is not broken in the future. Do you think you could add it, or need some help with that?

conan/cli/commands/remove.py
```diff
@@ -79,18 +80,23 @@ def confirmation(message):
             raise ConanException('--package-query supplied but the pattern does not match packages')
         package_list = conan_api.list.select(ref_pattern, args.package_query, remote)
         multi_package_list = MultiPackagesList()
-        multi_package_list.add("Local Cache" if not remote else remote.name, package_list)
+        multi_package_list.add(cache_name, package_list)
 
+    results_pkg_list = PackagesList() 
+    results = MultiPackagesList()
```
memsharded: You cannot just recreate the lists here, unconditionally, because this discard the information in the above ``multi_package_list``? A lot of tests in https://ci.conan.io/blue/organizations/jenkins/ConanTestSuitev2/detail/PR-14478/2/pipeline/36 are broken, please have a look.

memsharded (changes requested):

AbrilRBS:
Hi! Thanks a lot for your PR. Quick question,  as I'm afk right now and cant check by myself:

conan upload also has confirmation logic. Does the same issue also arise there?

Tobulus:
> Thanks for the fix @Tobulus, it is looking good. It would be necessary to add a unit test, to guarantee that this is not broken in the future. Do you think you could add it, or need some help with that?

Let me take a look, I try to add testing. If I need help, I will ask. Thanks!

conan/cli/commands/remove.py
Tobulus: the idea was to use `multi_package_list` only for iterating the stuff possible to be removed by the pattern and collect what was confirmed by the user in a separate list.

I'll take a look at the failing tests.

Tobulus:
> Thanks for the fix @Tobulus, it is looking good. It would be necessary to add a unit test, to guarantee that this is not broken in the future. Do you think you could add it, or need some help with that?

Should this be a unittest? Or can I add a test case [here](https://github.com/conan-io/conan/blob/release/2.0/conans/test/integration/command/remove_test.py)?

Tobulus:
> Hi! Thanks a lot for your PR. Quick question, as I'm afk right now and cant check by myself:
> 
> conan upload also has confirmation logic. Does the same issue also arise there?

Looks like the same problem there. Maybe we should fix that in a separate PR?

memsharded:
> Looks like the same problem there. Maybe we should fix that in a separate PR?

Yes, I think it is worth doing the upload in a separate PR, as there is some stuff there we need to discuss regarding the UploadAPI and how the ``package_lists`` are managed in the API. I will create a ticket dedicated from it from this comment.

Tobulus:
@memsharded The last failing test case actually revealed another problem, if I understand correctly.
This test case was passing before but was not actually doing what it is outputting.

[This](https://github.com/conan-io/conan/blob/release/2.0/conans/test/integration/command_v2/test_combined_pkglist_flows.py#L219) outputs that some binaries are deleted, but if you step through it with the debugger, this is not actually done (which now pops up as the output is corrected).

The problem seems [this part](https://github.com/conan-io/conan/blob/release/2.0/conan/cli/commands/remove.py#L89) which is not providing elements for iteration.

I don't understand the datastructure completly to tell why the iterator is empty ... can you help out?

If its helpful I can still provide a test case. As asked above: is it ok to not provide a unit test but a integration test?

memsharded:
Yes, it is possible that some tests are incorrect/buggy too, and those should be fixed. But there are too many tests broken, that typically means that there is something that is actually breaking valid working behavior.

> If its helpful I can still provide a test case. As asked above: is it ok to not provide a unit test but a integration test?

yes, integration tests are usually our preferred choice for most tests.

Tobulus:
> Yes, it is possible that some tests are incorrect/buggy too, and those should be fixed. But there are too many tests broken, that typically means that there is something that is actually breaking valid working behavior.
> 
> > If its helpful I can still provide a test case. As asked above: is it ok to not provide a unit test but a integration test?
> 
> yes, integration tests are usually our preferred choice for most tests.

2 failed, 1821 passed

I think you looked at the old test run? I force pushed one commit which I forgot, so there is no newer commit visible.

conan/cli/commands/remove.py
memsharded: Ok, that makes sense, now I see the ``add_xxx`` calls and less failing tests.
I am facing a similar problem in my "upload" branch (I am preparing the same fix for the  upload command), and I have done something similar.
The issue with this is that it is missing the ``package_id`` configuration description. Don't worry about it, it is fine, I will revisit this myself when finishing the upload PR, I might change the API a bit for ``PackageLists`` to help with this.

memsharded:
Yes, my bad, I hadn't realize about the new commits and new test run, looking better.

Sure, you can fix the tests and include them in the PR.

memsharded:
The PR https://github.com/conan-io/conan/pull/14512 for fixing the ``conan upload`` output has been merged.

Finally no API has changed, no new methods in PackagesList.

Please let me know if you would like a hand with this PR or you can manage to fix the tests.

Tobulus:
> The PR #14512 for fixing the `conan upload` output has been merged.
> 
> Finally no API has changed, no new methods in PackagesList.
> 
> Please let me know if you would like a hand with this PR or you can manage to fix the tests.

I'm out for the next two weeks. So I would be happy for help.

PR #14493, opened 2023-08-16
Warn of unknown version range options

AbrilRBS:
Changelog: Feature: Warn of unknown version range options.
Docs: Omit

As part of the fixes that came from https://github.com/conan-io/conan-center-index/pull/18513

As discussed with @czoido, I don't think this should break to allow for new options being used in older Conan versions, so a warning is the next best thing

### DIFF
--- a/conans/model/version_range.py
+++ b/conans/model/version_range.py
@@ -94,6 +94,9 @@ class VersionRange:
             if "include_prerelease" in t:
                 prereleases = True
                 break
+            else:
+                from conan.api.output import ConanOutput
+                ConanOutput().warning(f'Unrecognized version range option "{t}" in "{expression}"')
         version_expr = tokens[0]
         self.condition_sets = []
         for alternative in version_expr.split("||"):
--- a/conans/model/version_range.py
+++ b/conans/model/version_range.py
@@ -95,8 +95,12 @@ class VersionRange:
                 prereleases = True
                 break
             else:
+                t = t.strip()
                 from conan.api.output import ConanOutput
-                ConanOutput().warning(f'Unrecognized version range option "{t}" in "{expression}"')
+                if len(t) == 0 or t[0].isalpha():
+                    ConanOutput().warning(f'Unrecognized version range option "{t}" in "{expression}"')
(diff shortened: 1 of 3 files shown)


conans/model/version_range.py
```diff
@@ -95,8 +95,12 @@ def __init__(self, expression):
                 prereleases = True
                 break
             else:
+                t = t.strip()
                 from conan.api.output import ConanOutput
-                ConanOutput().warning(f'Unrecognized version range option "{t}" in "{expression}"')
+                if len(t) == 0 or t[0].isalpha():
+                    ConanOutput().warning(f'Unrecognized version range option "{t}" in "{expression}"')
+                else:
+                    ConanOutput().error(f'"{t}" in version range "{expression}" is not a valid option')
```
memsharded: Not raise exception?

conans/test/integration/graph/core/test_version_ranges.py
```diff
@@ -418,6 +418,9 @@ def test_different_user_channel_resolved_correctly():
 
 def test_unknown_options():
     c = TestClient()
-    c.save({"conanfile.py": GenConanfile("pkg", "1.0").with_requirement("lib/[>1.2, <1.4]")})
-    c.run("install .", assert_error=True)
-    assert 'Unrecognized version range option " <1.4" in ">1.2, <1.4"' in c.out
+
+    c.run("graph info --requires=lib/[>1.2,<1.4]", assert_error=True)
```
memsharded, line 424: This ``assert_error=True`` is failing for another reason, not the bad syntax.

conans/model/version_range.py
AbrilRBS: Woops, yeah that makes more sense

conans/test/integration/graph/core/test_version_ranges.py
AbrilRBS: Yep! Good catch. Either way, the following assert was still valid, but fixed it not to raise for unrelated version :)

memsharded (approved):

PR #14508, opened 2023-08-17
Add `tools.files.download:verify`

AbrilRBS:
Changelog: Feature: Add `tools.files.download:verify`.
Docs: https://github.com/conan-io/docs/pull/3341
As a look into to see if we event want this.

The motivation from this comes from trying to locally develop packages whose sources do not play nice with our JFrog VPN. This is usually `pkgconf` (For me and @jcar87 both at least), but I've seen it in a few others.

As editing the recipe is a pain in the neck, my usual go-to is changing Conan not to verify any download, which is not ideal either.

This would allow to do things such as
`conan create . --version=1.0.0 --build=missing -c:b "pkgconf/*:tools.files.download:verify=False"` and have finegrained control over the verification.

Another possible solution if this does not go thru is to use the backup of CCI sources. I'm not 100% sure that an exception on origin will trigger a check of the backups, it might be worth considering if not.

Also, tests are missing because I'm not quite sure how to properly test this. Mocking the downloader seems like the way to go, but I still need to check on that

### DIFF
--- a/conan/tools/files/files.py
+++ b/conan/tools/files/files.py
@@ -193,6 +193,7 @@ def download(conanfile, url, filename, verify=True, retry=None, retry_wait=None,
     retry = config.get("tools.files.download:retry", check_type=int, default=retry)
     retry_wait = retry_wait if retry_wait is not None else 5
     retry_wait = config.get("tools.files.download:retry_wait", check_type=int, default=retry_wait)
+    verify = config.get("tools.files.download:verify", check_type=bool, default=verify)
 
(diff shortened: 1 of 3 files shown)


memsharded (commented):
I understand the use case and this might be useful for other users, most likely not best practice (document it explicitly this shouldn't be done in production), but still useful under some dev circumstances, I don't oppose to adding it.

memsharded (approved):

jcar87 (approved):

PR #14510, opened 2023-08-17
resolve version ranges for editables too

### LINKED ISSUE #14507: [question] Editables not considered for version ranges
### What is your question?

Say you have two packages pa and pb where pb uses pa via a version range. (E. g. `self.requires("pa/[>1.0]")`)
You can now create the packages pa (version 1.0.0) and pb (version 2.0.0).
Now I want to make changes to pa and test those changes with pb. So I increase the version of pa to e. g. 1.0.1 and put it in "editable" mode. Then I have a `pa/1.0.0` in the local cache and a `pa/1.0.1` in the list of editables.
When I now recreate package pb, it still uses pa/1.0.0 from the cache instead of the newer editable package. If I delete pa/1.0.0 from the cache and try to recreate pb, I get an error message: `ERROR: Package 'pa/[>1.0]' not resolved: Version range '>1.0' from requirement 'pa/[>1.0], Traits: ... required by 'pb/2.0.0' could not be resolved.` even though there is a `pa/1.0.1` in the editables list.
When I change the recipe of pb to  `self.requires("pa/1.0.1")` then the editable pa package is used as expected.

Question is: Am I missing something or do version ranges ignore the editable packages alltogether?

Depending on the answer, feel free to change the type of this issue to "bug report" or "featue request". 😃 

(I am using Conan version 2.0.9, btw.)

memsharded:
Changelog: Feature: Allow ``version-ranges`` to resolve to editable packages too.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14507

### DIFF
--- a/conans/client/graph/range_resolver.py
+++ b/conans/client/graph/range_resolver.py
@@ -54,7 +54,9 @@ class RangeResolver:
             # TODO: This is still necessary to filter user/channel, until search_recipes is fixed
             local_found = [ref for ref in local_found if ref.user == search_ref.user
                            and ref.channel == search_ref.channel]
-            local_found.extend(self._cache.editable_packages.edited_refs)
+            local_found.extend(r for r in self._cache.editable_packages.edited_refs
+                               if r.name == search_ref.name and r.user == search_ref.user
+                               and r.channel == search_ref.channel)
             self._cached_cache[pattern] = local_found
         if local_found:
             return self._resolve_version(version_range, local_found, self._resolve_prereleases)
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):

conans/client/graph/range_resolver.py
```diff
@@ -54,6 +54,7 @@ def _resolve_local(self, search_ref, version_range):
             # TODO: This is still necessary to filter user/channel, until search_recipes is fixed
             local_found = [ref for ref in local_found if ref.user == search_ref.user
                            and ref.channel == search_ref.channel]
+            local_found.extend(self._cache.editable_packages.edited_refs)
```
AbrilRBS: Does edited_refs not need the filtering done in the line above this one?

AbrilRBS (commented):
One question after further looking into it

conans/client/graph/range_resolver.py
memsharded: Yes, it does, I will extend the tests and fix this, very good catch!

memsharded: Fixed, please re-review

AbrilRBS (approved):

czoido (approved):

PR #14512, opened 2023-08-18
Print only packages confirmed interactively in ``conan upload``

### LINKED ISSUE #14490: Redesign UploadAPI management of package_lists and fix user confirmation
The UploadAPI is labeling as ``upload: False`` things that we don't want to upload, and that is being used to do nothing with them. Maybe they should just be removed from the ``PackageList`` (or create and return a new package lists?)

> Looks like the same problem there. Maybe we should fix that in a separate PR?

Yes, I think it is worth doing the upload in a separate PR, as there is some stuff there we need to discuss regarding the UploadAPI and how the ``package_lists`` are managed in the API. I will create a ticket dedicated from it from this comment.

_Originally posted by @memsharded in https://github.com/conan-io/conan/issues/14478#issuecomment-1679702453_

memsharded:
Changelog: Bugfix: Print only packages confirmed interactively in ``conan upload``.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14490

### DIFF
--- a/conans/test/integration/command/upload/upload_complete_test.py
+++ b/conans/test/integration/command/upload/upload_complete_test.py
@@ -9,6 +9,7 @@ from requests import ConnectionError
 from conans.paths import CONAN_MANIFEST
 from conans.test.utils.tools import (NO_SETTINGS_PACKAGE_ID, TestClient, TestRequester, TestServer,
                                      GenConanfile)
+from conans.util.env import environment_update
 from conans.util.files import load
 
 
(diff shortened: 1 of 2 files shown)


conans/test/integration/command/upload/upload_complete_test.py
```diff
@@ -71,20 +72,38 @@ def test_upload_with_pattern():
...
     assert "Uploading recipe 'hello2/1.2.1@frodo/stable" not in client.out
 
 
+def test_check_upload_confirm_question_yes():
+    server = TestServer()
+    client = TestClient(servers={"default": server},
+                        inputs=["yes", "yes", "yes", "yes", "yes", "admin", "password"])
+    client.save({"conanfile.py": GenConanfile("hello1", "1.2.1")})
+    client.run("create . ")
+    client.save({"conanfile.py": GenConanfile("hello1", "1.2.1").with_package_file("file.txt",
```
AbrilRBS, line 96: TIL, super nice function to know!

AbrilRBS (approved):

czoido (approved):

PR #14522, opened 2023-08-18
Print exception stacktrace from command run into stderr

BobIsOnFire:
Changelog: Bugfix: Print exception stacktrace (when `-vtrace` is set) into stderr instead of stdout
Docs: Omit

Hi! A `ConanException` from some command like `conan <cmd> -v trace -f json > output.json` redirected my error into JSON file and didn't write anything into the console. I believe such output should be always written into stderr.


memsharded (approved):
Yes, you are right, this better go to stderr. Thanks for this fix @BobIsOnFire!

PR #14561, opened 2023-08-23
Fix/version prerelease hyphens

### LINKED ISSUE #14557: [bug] Pre-release versions are not always considered "later" than a release version (semver issue?)
### Environment details

* Compiler+version: N/A
* Conan version: 2.0.9
* Python version: 3.10.8

### Steps to reproduce

Use a version which has more than one hyphen:
```
>>> from conan.tools.scm import Version
>>> Version("0.0.1") > Version("0.0.1-3-1")
False
```

Quote:
```1.1-alpha.1``` is older than ```1.1```, not newer.

From:
https://docs.conan.io/2/tutorial/versioning/version_ranges.html#semantic-versioning

My understanding is that a pre-release should always be lower priority than a release, and the presence of a hyphen makes it a prerelease. According to [the semver spec (item 9)](https://semver.org/#spec-item-9) a prerelease may itself contain a hyphen, so this should be legal.

(a workaround which works for me, by the way, is to instead use + inside the prerelease string: 0.0.1-3+1)

### Logs

_No response_

memsharded:
Changelog: Fix: Allow extra hyphens in versions pre-releases.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14557


AbrilRBS (approved):
TIL. Happy to talk abou the impact this has, but Im not against adding it

memsharded:
See comments in https://github.com/conan-io/conan/issues/14557#issuecomment-1690785983

memsharded:
I have managed to make the PR a bit more compliant, without nested pre-released, but handling "-" as identifiers too, please re-review.

AbrilRBS (approved):

czoido (approved):

PR #14594, opened 2023-08-28
Allow `--format=json` in `conan create` for `python-requires`

### LINKED ISSUE #14577: [bug] conan create --format=json throws when creating a python_requires package
### Environment details

* Operating System+version: Ubuntu Linux 22.04
* Conan version: 2.0.9 / 2.0.10-dev
* Python version: 3.10.12

### Steps to reproduce

1. Create basic conanfile.py which shall be used as a python_requires. E. g. with the contents of the upper one from [this example page](https://docs.conan.io/2/reference/extensions/python_requires.html?highlight=python_requires):
```
from conan import ConanFile

myvar = 123

def myfunct():
    return 234

class Pkg(ConanFile):
    name = "pyreq"
    version = "0.1"
    package_type = "python-require"
```
2. Do a 
```
$ conan create . --format=json
```
This results in a:
```
ERROR: 'NoneType' object has no attribute 'serialize'
```
While a pure `conan create .` works as expected.

When searching the issue database, I stumbled across [this merge request](https://github.com/conan-io/conan/pull/14529) which looked like it would fix my exact problem. So I tried the source-version (commit 0e168973a58d284814c517d03baf841767c84d9f) but I still get the same error.

Background is: I want to use the generated json file with a `conan list --graph=conancreategraph.json --graph-binaries=build --format=json ...` and then a `conan upload --list=conanlist.json ...` command to upload the generated package to our conan remote. This should work with python_requires packages, as well, shoudn't it?

### Logs

_No response_

AbrilRBS:
Changelog: Fix: Allow `--format=json` in `conan create` for `python-requires`
Docs: Omit

Couldn't find anything concrete when testing locally, but this now makes it so calling `conan create .` with `--update` flag for python-requires does _something_ (Updates the recipe times). 

BUT, it's not capable of using a newer recipe _version_

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -258,9 +258,7 @@ class ConanFileLoader:
                 conanfile.requires(repr(reference))
 
         if python_requires:
-            conanfile.python_requires = []
-            for python_require in python_requires:
-                conanfile.python_requires.append(python_require.repr_notime())
+            conanfile.python_requires = [pr.repr_notime() for pr in python_requires]
 
         if self._pyreq_loader:
             self._pyreq_loader.load_py_requires(conanfile, self, graph_lock, remotes,
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -244,7 +244,8 @@ class ConanFileLoader:
 
         return conanfile
 
-    def load_virtual(self, requires=None, tool_requires=None, python_requires=None):
+    def load_virtual(self, requires=None, tool_requires=None, python_requires=None, graph_lock=None,
+                     remotes=None, update=None, check_updates=None):
         # If user don't specify namespace in options, assume that it is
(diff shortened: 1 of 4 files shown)


conans/client/loader.py
```diff
@@ -256,6 +257,15 @@ def load_virtual(self, requires=None, tool_requires=None):
             for reference in requires:
                 conanfile.requires(repr(reference))
 
+        if python_requires:
+            conanfile.python_requires = []
+            for python_require in python_requires:
+                conanfile.python_requires.append(python_require.repr_notime())
```
memsharded:
```suggestion
        if python_requires:
            conanfile.python_requires = [pr.repr_notime() for pr in python_requires]
```

memsharded (approved):

czoido (approved):

PR #14597, opened 2023-08-28
extended tests for `find_package_prefer_config`

thorsten-klein:
Changelog: Omit
Docs: Omit

By default `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is empty, which behaves exactly the same as setting it to `OFF`. 
Ref: #14550 

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -722,9 +722,14 @@ class TestCMakeFindPackagePreferConfig:
                      "FindComandante.cmake": find,
                      "ComandanteConfig.cmake": config,
                      "profile_true": profile.format(True),
-                     "profile_false": profile.format(False)})
+                     "profile_false": profile.format(False),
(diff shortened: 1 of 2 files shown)


AbrilRBS:
Hi @thorsten-klein thanks a lot for the contribution.

While I'm the first one to like extra tests when the situation calls for it, this PR introduces a total of 293 new test runs, making the functional test running times jump by quite a lot:


|Os|Old Time|New time|
|----|------|--------|
|Macos|5min|7min 22s|
|Windows|9min 38s|14min 51s (Wow!)|
|Linux|7min 44s|9min|

Is there a chance to simplify the amount of testing needed while keeping the same overall coverage? Thanks!

thorsten-klein:
Yes, this is true. 
I can mark them as slow running tests, so that they are skipped in most CI runs.
But better: I will have a look how I can optimize them. I guess it will be quite hard, since cmake has to run for real at this kind of test.

thorsten-klein:
I will close this PR since it is more a test of cmake than of conan.

PR #14599, opened 2023-08-28
Fix `editable add` remote args

AbrilRBS:
Changelog: Omit
Docs: https://github.com/conan-io/docs/pull/3345

Fix `conan editable` remote args to now properly work - the group was being done in the wrong parser

Found that the new editable remote args were not properly passed. I've fixed it there and checked that the rest of the mutual exclusive groups in the code base is done properly


memsharded (approved):
Curious, why it didn't fail before?

PR #14602, opened 2023-08-29
list python_requires in create graph.json

memsharded:
Changelog: Omit
Docs: Omit

Fix for https://github.com/conan-io/conan/issues/14577#issuecomment-1697078628


AbrilRBS (approved):
Makes sense that the if needs to be after adding the pyrequires, good find :)

PR #14605, opened 2023-08-29
MakeDeps: Provide "require" information, and more styling tweaks

iskunk:
Changelog: Feature: MakeDeps: Provide "require" information, and more styling tweaks.
Docs: Omit

Minor follow-on to #14133 

This includes the following:

* Whitespace adjustments for better visual arrangement;
* Include `(indirect dependency)` comment on applicable components (just an example of what can be done with the `require` info);
* Pass the `require` info to `DepContentGenerator` and `DepGenerator`, as it provides useful context for a dependency;
* Add a footer comment, to clearly indicate the end of the file.

### DIFF
--- a/conan/tools/gnu/makedeps.py
+++ b/conan/tools/gnu/makedeps.py
@@ -230,7 +230,6 @@ class GlobalContentGenerator:
             """)
 
     template_deps = textwrap.dedent("""\
-
             {{ define_variable_value("CONAN_DEPS", deps) }}
             """)
 
@@ -371,10 +370,12 @@ class DepContentGenerator:
 
     template = textwrap.dedent("""\
 
-        # {{ dep.ref }}
+        # {{ dep.ref }}{% if not req.direct %} (indirect dependency){% endif +%}
+
         CONAN_NAME_{{ name }} = {{ dep.ref.name }}
         CONAN_VERSION_{{ name }} = {{ dep.ref.version }}
         CONAN_REFERENCE_{{ name }} = {{ dep.ref }}
+
         CONAN_ROOT_{{ name }} = {{ root }}
 
         {{  define_variable_value("CONAN_SYSROOT_{}".format(name), sysroot) -}}
@@ -398,8 +399,9 @@ class DepContentGenerator:
         {{- define_variable_value("CONAN_COMPONENTS_{}".format(name), components) -}}
         """)
 
-    def __init__(self, dependency, root: str, sysroot, dirs: dict, flags: dict):
+    def __init__(self, dependency, require, root: str, sysroot, dirs: dict, flags: dict):
         self._dep = dependency
+        self._req = require
         self._root = root
         self._sysroot = sysroot
         self._dirs = dirs or {}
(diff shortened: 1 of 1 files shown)


conan/tools/gnu/makedeps.py
```diff
@@ -507,8 +512,9 @@ class DepGenerator:
     Process a dependency cpp_info variables and generate its Makefile content
     """
 
-    def __init__(self, dependency):
+    def __init__(self, dependency, require=None):
```
memsharded: If the ``require`` is always passed, better remove the ``None`` default

conan/tools/gnu/makedeps.py
```diff
@@ -597,6 +603,7 @@ class MakeDeps:
     """
 
     _title = "# This Makefile has been generated by Conan. DO NOT EDIT!\n"
+    _footer = f"# end {CONAN_MAKEFILE_FILENAME}\n"
```
memsharded: Not sure about the value of this footer. No other Conan generated file adds this end-of-file footer, so better remove it to align.

conan/tools/gnu/makedeps.py
```diff
@@ -371,10 +370,14 @@ class DepContentGenerator:
 
     template = textwrap.dedent("""\
 
-        # {{ dep.ref }}
+        # {{ dep.ref }}{% if not req.direct %} (indirect dependency){% endif +%}
+
         CONAN_NAME_{{ name }} = {{ dep.ref.name }}
+
```
memsharded: I removed these lines, because I like visually more compact files, and group all similar things in blocks. For me, all the name, version, ref, are conceptually part of the same "block" of information, and I like to have some visual hints of these information groups.

memsharded (commented):
Thanks for your contribution!

conan/tools/gnu/makedeps.py
iskunk: Whoops, minor oversight there. Was thinking of making it optional at first.

conan/tools/gnu/makedeps.py
iskunk: I'll remove it, but I would advise adopting this as a general convention. A clear footer indicates that you have the entire file, that it was not truncated for one reason or another.

conan/tools/gnu/makedeps.py
iskunk: Okay, I'm glad to see there is thought behind this. I'll leave the blank right before `_ROOT`, as that relates more to the other variables.

conan/tools/gnu/makedeps.py
memsharded: We haven't been reported a truncated file ever in many years. Creation of generated files happen in memory, and written all at once in single operation, extremely unlikely to happen.

iskunk:
Let me know if you have any further concerns with this (updated) patch.

memsharded (approved):

uilianries (approved):

iskunk:
Thanks @memsharded and @uilianries. I'm happy to see this generator finished, merged, and polished off.

Looking forward to seeing what you have in mind for MakeToolchain.

uilianries:
Thank you for your words @iskunk !! About the MakeToolchain, your feedback will be very important too! Regards!

Issue #14620, opened 2023-08-30
[feature] allow regex pattern for `tools.info.package_id:confs`


thorsten-klein:
### What is your suggestion?

It would be nice if users can specify patterns for conf `tools.info.package_id:confs` instead of listing each single conf name.

e.g.
```
[conf]
tools.info.package_id:confs = [ "tools.cmake.*" ]
```

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

memsharded:
Implemented in https://github.com/conan-io/conan/pull/14621 for 2.0.11

PR #14621, opened 2023-08-30
allow regex pattern for conf tools.info.package_id:confs

thorsten-klein:
Changelog: Feature: Supporting regex pattern for conf `tools.info.package_id:confs`
Docs: https://github.com/conan-io/docs/pull/3382

Ref: #14620 

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -473,10 +473,12 @@ class Conf:
         # Reading the list of all the configurations selected by the user to use for the package_id
         package_id_confs = self.get("tools.info.package_id:confs", default=[], check_type=list)
         for conf_name in package_id_confs:
-            value = self.get(conf_name)
-            # Pruning any empty values, those should not affect package ID
-            if value:
-                result.define(conf_name, value)
+            matching_confs = [c for c in self._values if re.match(conf_name, c)] or [conf_name]
+            for name in matching_confs:
+                value = self.get(name)
+                # Pruning any empty values, those should not affect package ID
+                if value:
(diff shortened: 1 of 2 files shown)


conans/model/conf.py
```diff
@@ -470,10 +470,12 @@ def copy_conaninfo_conf(self):
         # Reading the list of all the configurations selected by the user to use for the package_id
         package_id_confs = self.get("tools.info.package_id:confs", default=[], check_type=list)
         for conf_name in package_id_confs:
-            value = self.get(conf_name)
-            # Pruning any empty values, those should not affect package ID
-            if value:
-                result.define(conf_name, value)
+            matching_confs = [c for c in BUILT_IN_CONFS if re.match(conf_name, c)] or [conf_name]
```
memsharded: A quick question: If we also want to match the ``user.mycomp.*`` confs, then we would need to iterate better the defined values in the current conf, rather than matching against the ``BUILT_IN_CONFS``? I think this would be a more general approach.

memsharded (commented):
Hi @thorsten-klein 

Thanks for your contribution!
I think this is looking good, but I'd try to generalize it to cover also ``user.*`` confs, I think it is doable.
Also, it is better if you target the ``release/2.0`` branch, as that means it will go into the next ``2.0.11``, not need to wait until the ``2.1``.

memsharded (approved):

AbrilRBS (approved):

czoido (approved):

PR #14622, opened 2023-08-30
clear info.conf for header-only libraries

memsharded:
Changelog: Bugfix: ``tools.info.package_id:confs`` shouldn't affect header-only libraries.
Docs: Omit


AbrilRBS (approved):
Good catch!

franramirez688 (approved):

PR #14633, opened 2023-09-01
[MesonToolchain] Raises error if consumer uses raw options

franramirez688:
Changelog: omit
Docs: omit

UPDATED: Finally reverted by https://github.com/conan-io/conan/pull/14692

### DIFF
--- a/conan/tools/meson/helpers.py
+++ b/conan/tools/meson/helpers.py
@@ -99,8 +99,6 @@ def to_meson_value(value):
         return 'true' if value else 'false'
     elif isinstance(value, list):
         return '[{}]'.format(', '.join([str(to_meson_value(val)) for val in value]))
-    elif isinstance(value, dict):
-        return '{{{}}}'.format(', '.join([f'{k}: {str(to_meson_value(v))}' for k, v in value.items()]))
     elif isinstance(value, _PackageOption):
         raise ConanException("Conan options cannot be specified directly as a valid Meson data type."
                              " Please, use another Python data type like int, str, list or boolean.")
--- a/conan/tools/meson/helpers.py
+++ b/conan/tools/meson/helpers.py
@@ -1,4 +1,6 @@
 from conan.tools.build.flags import cppstd_msvc_flag
+from conans.errors import ConanException
+from conans.model.options import _PackageOption
 
 __all__ = ["to_meson_machine", "to_meson_value", "to_cppstd_flag"]
 
@@ -90,12 +92,18 @@ def to_meson_value(value):
     :return: formatted value as a ``str``.
     """
     # https://mesonbuild.com/Machine-files.html#data-types
+    # we don't need to transform the integer values
     if isinstance(value, str):
-        return "'%s'" % value
+        return f"'{value}'"
     elif isinstance(value, bool):
-        return 'true' if value else "false"
+        return 'true' if value else 'false'
     elif isinstance(value, list):
-        return '[%s]' % ', '.join([str(to_meson_value(val)) for val in value])
+        return '[{}]'.format(', '.join([str(to_meson_value(val)) for val in value]))
+    elif isinstance(value, dict):
+        return '{{{}}}'.format(', '.join([f'{k}: {str(to_meson_value(v))}' for k, v in value.items()]))
+    elif isinstance(value, _PackageOption):
+        raise ConanException("Conan options cannot be specified directly as a valid Meson data type."
+                             " Please, use another Python data type like int, str, list or boolean.")
     return value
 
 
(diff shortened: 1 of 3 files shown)


conan/tools/meson/helpers.py
```diff
@@ -90,12 +92,16 @@ def to_meson_value(value):
...
+        return f"'{value}'"
     elif isinstance(value, bool):
-        return 'true' if value else "false"
+        return 'true' if value else 'false'
     elif isinstance(value, list):
-        return '[%s]' % ', '.join([str(to_meson_value(val)) for val in value])
+        return '[{}]'.format(', '.join([str(to_meson_value(val)) for val in value]))
+    elif isinstance(value, _PackageOption):
+        raise ConanException("Conan options cannot be specified directly as a valid Meson data type."
+                             " Please, use another Python data type like int, str, list or boolean.")
```
franramirez688, line 104: @memsharded we were already transforming all the values into valid Meson data types

conans/test/functional/toolchains/meson/test_meson.py
```diff
@@ -296,3 +296,60 @@ def build(self):
     client.run("build .")
     assert "unrecognized character escape sequence" not in str(client.out)  # if Visual
     assert "unknown escape sequence" not in str(client.out)  # if mingw
+
+
+@pytest.mark.tool("meson")
```
memsharded: This doesn't need to be a full functional test that needs meson. The ``MesonToolchain`` can be safely operate without Meson installed at all. I think downgrading this to an integration test, or in this case, even better, a unittest, would be better.

memsharded (commented):
Looks good! 👍 
Just have a look to the test, I think a functional one is too much, as Meson is not really involved, errors happen before even calling meson.

conans/test/functional/toolchains/meson/test_meson.py
franramirez688: You're right, but I thought that it was safer if Meson minimally reads the native configuration file just to be sure that we're not making some syntax mistakes. This test is not compiling anything; it's only calling Meson to read that file. Anyway, I'm not opposed to converting it to a unit test.

memsharded: The test with Meson reading different options values is already there, in ``test_meson.py``:
```python
tc.project_options["STRING_DEFINITION"] = "Text"
tc.project_options["TRUE_DEFINITION"] = True
tc.project_options["FALSE_DEFINITION"] = False
tc.project_options["INT_DEFINITION"] = 42
tc.project_options["ARRAY_DEFINITION"] = ["Text1", "Text2"]
```

This test was about launching an error message if options is being passed, because it was erroring anyway, so it is enough to cover the error case, the other was already covered.

franramirez688: You're right. I did not look hard enough to see that test was already there. Thanks!

conans/test/functional/toolchains/meson/test_meson.py
```diff
@@ -13,94 +13,99 @@
...
+                tc.generate()
 
-    _meson_build = textwrap.dedent("""
-    project('tutorial', 'cpp')
-    add_global_arguments('-DSTRING_DEFINITION="' + get_option('STRING_DEFINITION') + '"',
-                         language : 'cpp')
-    add_global_arguments('-DHELLO_MSG="' + get_option('HELLO_MSG') + '"', language : 'cpp')
-    hello = library('hello', 'hello.cpp')
-    executable('demo', 'main.cpp', link_with: hello)
-    """)
```
franramirez688, line 70: Moved class attributes into the test. No changes here.

conans/test/functional/toolchains/meson/test_meson.py
```diff
@@ -13,94 +13,99 @@
...
-
-        self.t.run("install .")
-
+        self.t.run("build .", assert_error=True)
+        assert "Conan options cannot be specified directly as a valid Meson data type. " \
+               "Please, use another Python data type like int, str, list or boolean." in self.t.out
+        # Let's transform the Conan option into other allowed data type to solve the issue
+        self.t.save({"conanfile.py": conanfile_py.format(dynamic="True if self.options.shared "
+                                                                 "else False")})
+        self.t.run("build .")
```
franramirez688, line 94: This is the most meaningful part.

memsharded (approved):

AbrilRBS (approved):
My only concern comes from the differing behaviuours that CMakeToolchain now presents, we should thing about how to handle that, and user expectations. For now just having this properly documented is a-ok for me :)

PR #14642, opened 2023-09-01
cli_out_write: don't decorate data if colors are not set

### LINKED ISSUE #14614: [feature] Don't add escape sequences to stdout of (some) query commands, regardless of CLICOLOR_FORCE
### What is your suggestion?

Hi!

My use case: I am running various conan commands in a wrapper script. The wrapper is using its own buffering when running subprocesses, so when I run the script in the terminal, conan doesn't know that it actually runs in terminal, and disables colors in logs. I set CLICOLOR_FORCE env variable when I want to see colored logs, which works nicely for me.

Sadly, when I want to capture stdout of a command and do something with it, CLICOLOR_FORCE adds unwanted escape sequences and breaks my script:

```bash
$ conan list 'zlib/*' --format=json | cat -A
{$
    "Local Cache": {$
        "zlib/1.2.13": {}$
    }$
}$
$ CLICOLOR_FORCE=1 conan list 'zlib/*' --format=json | cat -A
{$
    "Local Cache": {$
        "zlib/1.2.13": {}$
    }$
}^[[0m$
```

Same can be observed if I run simple `conan config home`. This is kinda expected since I add a variable to force colors, but JSON output format is meant to be processed automatically, so I think it shouldn't be touched with color decorations at all.

BobIsOnFire:
Changelog: Bugfix: Fix unnecessarily decorating command stdout with escape sequences.
Docs: omit

Fixes #14614.

### DIFF
--- a/conan/api/output.py
+++ b/conan/api/output.py
@@ -227,7 +227,7 @@ def cli_out_write(data, fg=None, bg=None, endline="\n", indentation=0):
 
     fg_ = fg or ''
     bg_ = bg or ''
-    if color_enabled(sys.stdout):
+    if (fg or bg) and color_enabled(sys.stdout):
         data = f"{' ' * indentation}{fg_}{bg_}{data}{Style.RESET_ALL}{endline}"
     else:
         data = f"{' ' * indentation}{data}{endline}"


memsharded (approved):

memsharded:

Issue #14560, opened 2023-08-23
[bug] CMakeDeps: Target name '<name>' already exists


BobIsOnFire:
### Environment details

* Operating System+version: Oracle Linux 8
* Compiler+version: gcc 8
* Conan version: 2.0.9
* Python version: 3.6

#### Minimal project example

```
.
├── CMakeLists.txt
└── conanfile.txt
```

`conanfile.txt`:

```ini
[requires]
libarchive/3.6.2

[options]
libarchive/*:with_lz4 = True

[generators]
CMakeDeps
```

`CMakeLists.txt`:

```cmake
cmake_minimum_required(VERSION 3.15)

project(test)

find_package(LibArchive REQUIRED)
find_package(lz4 REQUIRED)
```

#### Commands

* `conan install . --output-folder ./conan/ --build missing`
* `cmake -B ./build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=$(pwd)/conan -DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON`

### Logs

Logs from `cmake` command:

```
-- The C compiler identification is GNU 8.5.0
-- The CXX compiler identification is GNU 8.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Conan: Target declared 'LibArchive::LibArchive'
-- Conan: Target declared 'ZLIB::ZLIB'
-- Conan: Target declared 'Iconv::Iconv'
-- Conan: Target declared 'LZ4::lz4_static'
CMake Warning at conan/lz4Targets.cmake:25 (message):
  Target name 'lz4::lz4' already exists.
Call Stack (most recent call first):
  conan/lz4-config.cmake:16 (include)
  CMakeLists.txt:8 (find_package)


-- Configuring done (0.3s)
-- Generating done (0.0s)
-- Build files have been written to: /home/user/repro/build
```

[...]

Issue #12958, opened 2023-01-23
[bug] Warning when using target aliases with CMakeDeps


phamelin:
### Environment details

* Conan version: 1.54

### Steps to reproduce

If a package has defined  `cmake_target_aliases` property, e.g.:

```
def package_info(self):
    # ...
    self.cpp_info.set_property("cmake_target_name", "mypackage::mypackage")
    self.cpp_info.set_property("cmake_target_aliases", ["mypackage"])
```
then, there's a warning when the find_package is called twice:
```
CMake Warning at build/Release/conan/mypackageTargets.cmake:25 (message):
   Target name 'mypackage' already exists.
 Call Stack (most recent call first):
   build/Release/conan/mypackageTargets-config.cmake:10 (include)
   CMakeLists.txt:54 (find_package)
```
This is explained by this part of the `mypackageTargets.cmake`:
```
if(NOT TARGET mypackage::mypackage)
    add_library(mypackage::mypackageINTERFACE IMPORTED)
    message(STATUS "Conan: Target declared 'mypackage::mypackage'")
endif()
if(NOT TARGET mypackage)
    add_library(mypackageINTERFACE IMPORTED)
    set_property(TARGET mypackagePROPERTY INTERFACE_LINK_LIBRARIES mypackage::mypackage)
else()
    message(WARNING "Target name 'mypackage' already exists.")
endif()
```
We see that the `else()` condition is present only for the target alias. We can avoid to call `find_package(mypackage)` twice in the consumer package. However, if another package also depends on `mypackage`, `find_dependency` macro will be called by the other package config file and the warning will still be present.

If there's no warning for the main target, I suggest that there should have no warning for the target alias as well.

Hopobcn:
Any feedback on this? It's really a bit annoying to have these warnings TBH. They dirty the cmake configuration process which lowers the confidence on the conan generated code.

> New in version 3.15: Added the NOTICE, VERBOSE, DEBUG, and TRACE levels.
  
Perhaps now that cmake supports 3.15 these warnings could be moved to another level?

memsharded:
Thanks for the feedback.
https://github.com/conan-io/conan/pull/14644 from @BobIsOnFire will remove this warning, it was more annoying that helpful. It will be in next 2.0.11

PR #14644, opened 2023-09-01
 CMakeDeps: remove warning about existing alias

BobIsOnFire:
Changelog: Fix: CMakeDeps: Remove "Target name ... already exists" warning about duplicating aliases.
Docs: omit

This PR benefits #14560 (and #12958)

### DIFF
--- a/conan/tools/cmake/cmakedeps/templates/targets.py
+++ b/conan/tools/cmake/cmakedeps/templates/targets.py
@@ -80,8 +80,6 @@ class TargetsTemplate(CMakeDepsFileTemplate):
         if(NOT TARGET {{alias}})
             add_library({{alias}} INTERFACE IMPORTED)
             set_property(TARGET {{ alias }} PROPERTY INTERFACE_LINK_LIBRARIES {{target}})
-        else()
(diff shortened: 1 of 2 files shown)


memsharded:
Thanks for your contribution @BobIsOnFire 

I think I'd advocate for removing the warning completely, I think it is there to protect against colliding aliases from different packages, but that check should better be done globally, I'll check with the team.

memsharded:
It seems this warning is more problematic than helpful for most users. Lets do the following:

- Please remove the warnings completely from CMake code (do you want to do it in this same PR @BobIsOnFire ?). Note: there is a test ``test_collide_component_alias`` that explicitly checks for this warning, this test can remove the check, and add a comment that this is possible (alias collision), and it will not warn at this moment.
- As a future thing, we could consider checking in the ``CMakeDeps`` python code possible aliases colisions and warn, or even error from there.

BobIsOnFire:
Hi @memsharded 

Thanks! I have updated branch to remove warning / adjust tests.

jcar87 (approved):

memsharded (approved):

PR #14645, opened 2023-09-02
Remove last uneeded `StoppableThreadBottle` server in tests

### LINKED ISSUE #13448: Improve test that use StoppableThreadBottle
There are some tests that use StoppableThreadBottle that could use with some common abstraction such as a TestFileServer

AbrilRBS:
Changelog: Omit
Docs: Omit

Found the linked issue while searching for cool 2.X goals and saw this. Went looking into the code and to my suprise, only 1 uneeded usage remained! 

With this, only 3 usages of `StoppableThreadBottle` remain, which are in the backup sources tests as part of http errors, so it's harder to substitute them.

Closes https://github.com/conan-io/conan/issues/13448

### DIFF
--- a/conans/test/functional/command/config_install_test.py
+++ b/conans/test/functional/command/config_install_test.py
@@ -12,10 +12,10 @@ from conans.client.conf.config_installer import _hide_password
 from conans.client.downloaders.file_downloader import FileDownloader
 from conans.paths import DEFAULT_CONAN_HOME
 from conans.test.assets.genconanfile import GenConanfile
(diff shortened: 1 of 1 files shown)


memsharded (commented):
This test is not easy to remove, because the ``conan config install`` doesn't inject the server in the same way, if I recall correctly.

memsharded:
> With this, only 3 usages of StoppableThreadBottle remain, which are in the backup sources tests as part of http errors, so it's harder to substitute them.

We might try to extend the server with more endpoints with errors if that help to replace these tests

PR #14659, opened 2023-09-04
Add `conan config install` merge strategy

### LINKED ISSUE #13821: [feature] Allow "conan config install" to merge remotes
### What is your suggestion?

Currently `conan config install` overrides all existing remotes. I'd like to be able to tell conan that the remotes should be merged instead.
Our use case is the following:
- We have "base" repo where we store company wide conan configuration including shared remote. We use `conan config install <repo>` to setup basic configuration.
- Every team within a company can tune the configuration and add team-specific remote. If a team has additional configuration, it's configured by second invocation of `conan config install` pointing to a team-specific configuration repo. Having only "merge" strategy for remotes forces teams to repeat shared remote in their configurations which contradicts with DRY principle. The reason to have multiple remotes is permissioning - some teams don't want their packages to be used outside of their teams or they want to restrict what packages can be shared (i.e. published to shared remote).

AbrilRBS:
Changelog: Feature: Add `conan config install` merge strategy
Docs: TODO

Rethink how we want this to work. Also, the tests for config install need to be reviewed, I plan on doing it in this PR to make them prettier if possible.

Closes #13821

### DIFF
--- a/conan/api/subapi/config.py
+++ b/conan/api/subapi/config.py
@@ -10,7 +10,7 @@ class ConfigAPI:
         return self.conan_api.cache_folder
 
     def install(self, path_or_url, verify_ssl, config_type=None, args=None,
-                source_folder=None, target_folder=None, strategy="append"):
+                source_folder=None, target_folder=None, strategy="replace"):
         # TODO: We probably want to split this into git-folder-http cases?
         from conans.client.conf.config_installer import configuration_install
         app = ConanApp(self.conan_api.cache_folder)
--- a/conan/api/subapi/config.py
+++ b/conan/api/subapi/config.py
@@ -10,13 +10,14 @@ class ConfigAPI:
         return self.conan_api.cache_folder
 
     def install(self, path_or_url, verify_ssl, config_type=None, args=None,
-                source_folder=None, target_folder=None):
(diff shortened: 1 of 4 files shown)


conan/api/subapi/config.py
```diff
@@ -10,13 +10,14 @@ def home(self):
         return self.conan_api.cache_folder
 
     def install(self, path_or_url, verify_ssl, config_type=None, args=None,
-                source_folder=None, target_folder=None):
+                source_folder=None, target_folder=None, strategy="replace"):
```
memsharded, line 13: I think the approach is not that much as cli args, but as information encoded in files.

memsharded (commented):
I think the approach is information in files. Some possibilities could be:
- An extra field in the files like ``strategy: merge`` in the ``remotes.json`` file
- An extra file that encodes which files will follow a "merge" strategy instead of a "replace" strategy
- Some python plugin-like code that can do arbitrary operations over the files.

Lets discuss it :)

AbrilRBS:
We won't go ahead with this. After discussing the feature with the team, we've decided that there's too much variability for a feature like this to be clear to use. See the linked issue for a more detailed explanation

PR #14660, opened 2023-09-05
Feature/collect metadata

memsharded:
Changelog: Omit
Docs: Omit

Complements the metadata functionality with a custom command "collect" test, plus some fixes and interface necessary for implementing it


AbrilRBS (approved):

conans/client/graph/install_graph.py
```diff
@@ -229,7 +244,7 @@ def _initialize_deps_graph(self, deps_graph):
             else:
                 existing.add(node)
 
-    def install_order(self):
+    def install_order(self, by_levels=True):
```
franramirez688: I think that `by_levels` is a little bit confusing. I'm thinking of `flat=True` instead, where `flat = True` means `by_levels = False`. Wdyt?

czoido (approved):

conans/test/integration/metadata/test_metadata_collect.py
```diff
@@ -0,0 +1,201 @@
...
+        if os.path.exists(folder):
+            conanfile.output.warning(f"Folder for {conanfile} already exist, removing it: {folder}")
+            shutil.rmtree(folder)
+        conanfile.output.info(f"Copying recipe metadata from {conanfile.recipe_metadata_folder}")
+        _copy_metadata(conanfile.recipe_metadata_folder, os.path.join(folder, "recipe"))
+        for package in install_reference.packages.values():
+            pkg_metadata_folder = package.conanfile.package_metadata_folder
+            conanfile.output.info(f"Copying package metadata from {pkg_metadata_folder}")
+            _copy_metadata(pkg_metadata_folder, os.path.join(folder, "package"))
+'''
```
franramirez688, line 118: Perhaps it's unnecessary now, but it could be great to save this piece of code as a Python module in `assets` and copy and paste that file when the tests start. It'd make our lives easier when reading the diff.

franramirez688 (approved):

conans/test/integration/metadata/test_metadata_collect.py
memsharded: This is very specific for this test, not like a general asset, isn't it?

Issue #13424, opened 2023-03-13
[feature] can't specify c++20 with meson generator


lukester1975:
### What is your suggestion?

`/std:c++20` is supported by `cl` since [VS 2019 16.11](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-160) :

> The /std:c++20 option enables C++20 standard-specific features and behavior. Available starting in Visual Studio 2019 version 16.11

and clang-cl since [13](https://github.com/llvm/llvm-project/commit/a8f75d49).

`c++latest` is far from ideal as it turns on 23 features (and generates a ton of deprecation warnings). An additional consideration is that only meson [1.0.1](https://github.com/mesonbuild/meson/pull/11221) supports `c++20` for `clang-cl`. So ultimately 

```
_vs_cppstd_map = {
    '14': "vc++14",
    '17': "vc++17",
    '20': "vc++latest"
}
```
is somewhat simplistic, as ideally it's a function of compiler, compiler version and meson version... But maybe a trivial change to `vc++20` is pragmatic?

Thanks

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

memsharded:
Hi @lukester1975 

Thanks for your report

I think that mapping do not affect ``clang`` compiler and only ``msvc`` compiler:
```python
    if compiler == "msvc":
        return _vs_cppstd_map.get(cppstd)
```

But ``'20': "c++20", 'gnu20': "gnu++20"`` would be valid for ``clang`` compiler.

lukester1975:
Well ... I'd say yes and no :) I really meant the `clang-cl` driver, rather than `clang`, but I was going to open an issue about that separately - getting conan to actually use the `clang-cl` driver (i.e. via the `ClangCL` toolset) seems difficult, but I've not read *all* the relevant issues on that subject yet (there are quite a few!)... That said I have a 3 line mod (2 in profile, 1 in `settings.yml`) that makes it "just work". I really don't want the `clang` driver...

But as for the subject of this issue, yeah proper `cl` + `/std:c++20` support is the concern, I just mention meson 1.0.1 and `clang-cl` here for completeness. Ultimately the mapping should be a function of compiler version and meson version, I suppose. But maybe not worth the effort?

Thanks

memsharded:
You might want to read https://blog.conan.io/2022/10/13/Different-flavors-Clang-compiler-Windows.html.

[...]

PR #14664, opened 2023-09-05
Feature/set meson b_vscrt with clang-cl

lukester1975:
Changelog: Feature: Ensure meson toolchain sets `b_vscrt` with clang-cl.
Docs: Omit

This is a follow up to #https://github.com/conan-io/conan/issues/13424#issuecomment-1471974865, specifically item 3.

* The `if compiler in ("msvc", "clang")` is a bit ugly but as I mentioned in the issue, I couldn't see a way to see if the compiler is really `clang-cl` at that point. Safe enough given the `msvc_runtime_flag` implementation (note I changed it to return `None` for all non-`compiler.runtime` cases such that _b_vscrt is consistently `None`; all other usages seem OK with that).
* Given that, arguably the `if compiler in ("msvc", "clang")` test could be removed. Intel's cl-like frontend?
* In the tests I couldn't get the VS17/clang16 `pytest.mark.tool` working, so they are commented out. I'm sure I'm just missing some magic incantation somewhere despite VS2022 being enable in my env.
* VS 17.7 has bumped clang from 15 -> 16. So I guess if the test runners aren't updated, these tests may fail. I use a jinja profile for clang-cl with this hack: `compiler.version={{ os.popen("clang-cl --version").read().removeprefix("clang version ").split(".")[0] }}`. I guess this could do something similar if necessary?

### DIFF
--- a/conans/test/integration/toolchains/meson/test_mesontoolchain.py
+++ b/conans/test/integration/toolchains/meson/test_mesontoolchain.py
@@ -1,4 +1,7 @@
+import platform
 import textwrap
+
+import pytest
 
 from conan.tools.meson import MesonToolchain
 from conans.test.utils.tools import TestClient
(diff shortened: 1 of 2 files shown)


conans/test/integration/toolchains/meson/test_mesontoolchain.py
```diff
@@ -165,3 +168,127 @@ def generate(self):
     content = t.load(MesonToolchain.native_filename)
     assert "wrap_mode " not in content
     assert "nofallback" not in content
+
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="requires Win")
+# @pytest.mark.tool("visual_studio", "17")
+# @pytest.mark.tool("clang", "16")
+def test_clang_cl_vscrt_mdd():
```
memsharded: It seems that these tests are very similar, it would be great to factorize (using ``pytest.mark.parametrize``) to reduce code. If not sure how to do it, don't worry, just tell, we can help with that.

conan/tools/microsoft/visual.py
```diff
@@ -177,7 +177,7 @@ def msvc_runtime_flag(conanfile):
     Gets the MSVC runtime flag given the ``compiler.runtime`` value from the settings.
 
     :param conanfile: ``< ConanFile object >`` The current recipe object. Always use ``self``.
-    :return: ``str`` runtime flag.
+    :return: ``str`` runtime flag or ``None``.
```
memsharded: I am concerned that this is breaking existing usages of this function, as this is exposed as a tool. I'd prefer to keep the current behavior returning ``""`` empty string.

memsharded (commented):
Thanks for your contribution! Please have a look to the comments.

conans/test/integration/toolchains/meson/test_mesontoolchain.py
lukester1975: Great thanks for the pointer, I assumed there'd be a nice way... hopefully I got it right.

What do you think about the VS17 / clang 15 or 16 issue?

conan/tools/microsoft/visual.py
lukester1975: Ah OK; done.

conans/test/integration/toolchains/meson/test_mesontoolchain.py
memsharded: Yes, looks good, thanks!

I don't think that using the full tool vs/clang or whatever is necessary, the testing you did sounds like the right balance.

memsharded: > In the tests I couldn't get the VS17/clang16 pytest.mark.tool working, so they are commented out. I'm sure I'm just missing some magic incantation somewhere despite VS2022 being enable in my env.

If you tried making it work in CI, only ``functional`` tests have the tools installed, using ``pytest.mark.tool`` will fail.

The tests as you did them don't need the tools installed at all, just Conan, so things are looking good.

memsharded (approved):

PR #14672, opened 2023-09-05
error message for usages of package_folder in generate()

memsharded:
Changelog: Fix: Show error message if using ``self.package_folder`` in ``generate()`` method.
Docs: TODO

NOTE: The implementation is not great, but python magic to patch the conanfile was also not great, too complicated. We are already doing it in ``conanfile_remove_attr``, but a passthrough error msg was a bit more complicated. Open to suggestions and ideas.

### DIFF
--- a/conans/client/generators/__init__.py
+++ b/conans/client/generators/__init__.py
@@ -4,7 +4,7 @@ import traceback
 import importlib
 
 from conans.client.subsystems import deduce_subsystem, subsystem_path
-from conans.errors import ConanException, conanfile_exception_formatter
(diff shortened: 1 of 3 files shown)


memsharded:
As pointed by @czoido CMakeToolchain is internally using ``self.package_folder``, which is bad. We need to wait and discuss about this.

memsharded:
Dropping this, package_folder might be convenient to use in some scenarios like allowing ``cmake.install()`` in the cache, with the toolchain defining the output folders. It should just be documented that it can be ``None`` in the local flow and cannot be leveraged nor written at at ``generate()`` time.

PR #14676, opened 2023-09-06
[MesonToolchain] Binaries `c`, `cpp` and `ld` (added recently) can be lists

### LINKED ISSUE #13824: [question] subtle difference between [buildenv] and [env]?
### What is your question?

Hi,

see this post for failing cross-build of libsystemd: https://github.com/conan-io/conan-center-index/issues/17270#issuecomment-1534901887

Is there any difference in the implementation of `[buildenv]` compared to `[env]` which explains this behaviour?

### LINKED ISSUE #14028: [bug] MesonToolchain does not handle CXX, CC with arguments included, needs separate LD handling
### Environment details

* Operating System+version: Debian 11 bullseye
* Compiler+version: GCC 10
* Conan version: 1.60.0
* Python version: 3.9.2

### Steps to reproduce

I have toolchain called `fsl-imx-wayland-glibc-x86_64-core-image-base-cortexa53-crypto-imx8mmevk-toolchain-5.10-hardknott.sh` which has encironment setup script `environment-setup-cortexa53-crypto-poky-linux` defining CXX an C like this:
```bash
export CC="aarch64-poky-linux-gcc  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=$SDKTARGETSYSROOT"
export CXX="aarch64-poky-linux-g++  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=$SDKTARGETSYSROOT"
```

In other words, CC and CXX has arguments embedded together with executable name. And Meson does not like that:

```
../src/meson.build:1:0: ERROR: Unknown compiler(s): [['aarch64-poky-linux-gcc  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/opt/fsl-imx-wayland/5.10-hardknott/sysroots/cortexa53-crypto-poky-linux']]
```

Because geenrated `conan_meson_cross.ini` has:

```ini
[binaries]
c = 'aarch64-poky-linux-gcc  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/opt/fsl-imx-wayland/5.10-hardknott/sysroots/cortexa53-crypto-poky-linux'
cpp = 'aarch64-poky-linux-g++  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/opt/fsl-imx-wayland/5.10-hardknott/sysroots/cortexa53-crypto-poky-linux'
```
It Works For Me™ after implementing workaround/hack/fix in `tools/meson/toolchain.py` by generating `c` and `cpp` as lists:
```diff
42,43c42,43
<     {% if c %}c = '{{c}}'{% endif %}
<     {% if cpp %}cpp = '{{cpp}}'{% endif %}
---
>     {% if c %}c = {{c}}{% endif %}
>     {% if cpp %}cpp = {{cpp}}{% endif %}
49a50
>     {% if ld %}ld = {{ld}}{% endif %}
173,174c174,175
<         self.c = compilers_by_conf.get("c") or build_env.get("CC") or default_comp
<         self.cpp = compilers_by_conf.get("cpp") or build_env.get("CXX") or default_comp_cpp
---
>         self.c = self._quote_or_none(compilers_by_conf.get("c") or self._get_env_list(build_env.get("CC")) or default_comp)
>         self.cpp = self._quote_or_none(compilers_by_conf.get("cpp") or self._get_env_list(build_env.get("CXX")) or default_comp_cpp)
176,177c177,179
<         self.c_ld = build_env.get("CC_LD") or build_env.get("LD")
<         self.cpp_ld = build_env.get("CXX_LD") or build_env.get("LD")
---
>         self.c_ld = build_env.get("CC_LD")
>         self.cpp_ld = build_env.get("CXX_LD")
>         self.ld = self._get_env_list(build_env.get("LD", []))
201a204,210
>     def _quote_or_none(self, value):
>         # do not quote lists, ignore None
>         if value is None or isinstance(value, list):
>             return value        
>         
>         return f"'{value}'"
> 
303a313,315
>         if v is None:
>             return v
>
```
Basically, I had to remove quotes around `cc` and `cxx` in the template, as lists should not have those, but quote later when needed.

Also, I had to separate `LD` env. value into `ld` for meson because of:
```
../src/meson.build:1:0: ERROR: Unsupported linker, only bfd, gold, and lld are supported, not aarch64-poky-linux-ld  --sysroot=/opt/fsl-imx-wayland/5.10-hardknott/sysroots/cortexa53-crypto-poky-linux.
```
From IRC:

> [Friday, June 2, 2023] [2:23:00 PM EEST] <elibrokeit> Talkless: the cpp_ld value needs to be something that can be passed as -fuse-ld= 

> [Friday, June 2, 2023] [2:41:56 PM EEST] <elibrokeit> LD is not the same as CXX_LD so I guess conan should not do this
[Friday, June 2, 2023] [2:42:37 PM EEST] <elibrokeit> LD is the kind of thing I'd expect to see being used when manually linking objects with ld, not when using the compiler as the linker driver

### Logs

_No response_

franramirez688:
Changelog: Feature: Added mechanism to transform `c`, `cpp`, and/or `ld` binaries variables from Meson into lists if declared blank-separated strings.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/14028

Maybe closes: https://github.com/conan-io/conan/issues/13824 https://github.com/conan-io/conan/issues/14006


conans/test/integration/toolchains/meson/test_mesontoolchain.py
```diff
@@ -205,3 +206,80 @@ def test_clang_cl_vscrt(build_type, runtime, vscrt):
...
+    assert "c = 'CC_VALUE'" in content
+    assert "cpp = 'CXX_VALUE'" in content
+    assert "ld = 'LD_VALUE'" in content
+    assert "c_ld = 'CC_LD_VALUE'" in content
+    assert "cpp_ld = 'CXX_LD_VALUE'" in content
+    assert "ar = 'AR_VALUE'" in content
+    assert "strip = 'STRIP_VALUE'" in content
+    assert "as = 'AS_VALUE'" in content
+    assert "windres = 'WINDRES_VALUE'" in content
+    assert "pkgconfig = 'PKG_CONFIG_VALUE'" in content
```
franramirez688, line 259: This test is coming from the `test_meson_build_require`. Previously, it was functional, but it shouldn't.

conans/test/integration/toolchains/meson/test_mesontoolchain.py
```diff
@@ -205,3 +206,80 @@ def test_clang_cl_vscrt(build_type, runtime, vscrt):
...
+       [settings]
+       os=Windows
+       arch=x86_64
+       compiler=gcc
+       compiler.version=9
+       compiler.cppstd=17
+       compiler.libcxx=libstdc++11
+       build_type=Release
+       [buildenv]
+       CC=aarch64-poky-linux-gcc  -mcpu=cortex-a53 -march=armv8-a+crc+crypto
```
memsharded, line 274: So Meson needs this defined as a list for the ``c``, ``cpp``, etc? This reads a bit ugly imho, I'd expect to have the compiler as one thing and the list of flags as cxxflags or similar

franramirez688: I agree with that, but the thing is that you can pass all those arguments to the CC in one shot as a list and it seems to be working as it's said in the issue https://github.com/conan-io/conan/issues/14028.

Overall, we're not changing the behavior at all. We agree that this should not be the way to go, but we're only allowing the users to have those compiler and compiler flags together without changing those env variables (up to them). You can check the Meson documentation too (https://mesonbuild.com/Machine-files.html#binaries).

memsharded (approved):

AbrilRBS (approved):

czoido (approved):

conans/test/integration/toolchains/meson/test_mesontoolchain.py
jwillikers: @memsharded This is unforunately how Yocto configures the `CC` and similar environment variables out-of-the-box.

conan/tools/meson/toolchain.py
```diff
@@ -380,11 +391,11 @@ def _context(self):
...
             "project_options": {k: to_meson_value(v) for k, v in self.project_options.items()},
             # https://mesonbuild.com/Builtin-options.html#directories
-            # TODO : we don't manage paths like libdir here (yet?)
             # https://mesonbuild.com/Machine-files.html#binaries
             # https://mesonbuild.com/Reference-tables.html#compiler-and-linker-selection-variables
-            "c": self.c,
-            "cpp": self.cpp,
+            "c": _sanitize_format(self.c),
+            "cpp": _sanitize_format(self.cpp),
+            "ld": _sanitize_format(self.ld),
```
jwillikers, line 398: @franramirez688 According to the Meson documentation [here](https://mesonbuild.com/howtox.html#set-linker), I didn't think using a multivalue argument for the linker options made sense sense it would have to be one of `gold`, `lld`, `bfd` for Clang / GCC and only for MSVC / Clang-Cl should it be an executable. I think this is pretty different behavior than how autotools make use of the `LD` environment variable.

Fyi, I really appreciate you improving the interface for this stuff in this PR.

franramirez688: Yeah, indeed @jwillikers 
I added that possibility only to be sure that I was not missing any corner case where you could want to pass several arguments to LD. Anyway, I hope consumers will use `<>_ld` instead, as Meson recommends in its documentation.

Issue #14678, opened 2023-09-06
[bug] conan does not consider RREV for --build-require


thorsten-klein:
### Environment details

* Operating System+version: Ubuntu20
* Compiler+version: -
* Conan version: 1.60.1
* Python version: 3.8.10

### Steps to reproduce

Install a specific revision of a package, e.g. `conan install --build-require cmake/3.22.3@#8d3544c0d9b1e909d14f45e4eba896f3`

### Logs

```
$ conan install --build-require cmake/3.22.3@#8d3544c0d9b1e909d14f45e4eba896f3
Configuration:
[settings]
arch=x86_64
build_type=Release
compiler=system
os=Linux
[options]
[build_requires]
[env]
[conf]
tools.env.virtualenv:auto_use=True
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True
tools.cmake.cmaketoolchain:find_package_prefer_config=False

cmake/3.22.3: Not found in local cache, looking in remotes...
cmake/3.22.3: Trying with 'conancenter'...
Decompressing conan_export.tgz completed [0.00k]                                         
cmake/3.22.3: Downloaded recipe revision a6a56a74df4d58b36d981508706815f0
```

Issue: Revision a6a56a74df4d58b36d981508706815f0 is downloaded.

memsharded:
Closed by https://github.com/conan-io/conan/pull/14679

PR #14679, opened 2023-09-06
bugfix conan 1.x: consider revision for build_requires

thorsten-klein:
Changelog: Bugfix: Consider revision for build_requires in cli arguments.
Docs: Omit

This solves #14678 

### DIFF
--- a/conans/client/loader.py
+++ b/conans/client/loader.py
@@ -349,7 +349,7 @@ class ConanFileLoader(object):
         conanfile.settings = profile_host.processed_settings.copy_values()
 
         if is_build_require:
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Thanks for fixing this @thorsten-klein 

This seems indeed an issue, and the fix looks correct.
Would be great to have a unit test that covers the fix, if you can add it, please do, otherwise don't worry, we will add it.

Assigning it for next 1.61 release

thorsten-klein:
I have added some test for it now 😊

memsharded (approved):
Great, thanks!

PR #14682, opened 2023-09-06
Support clang-cl in default profile plugin

lukester1975:
Changelog: Feature: Support clang-cl in default profile plugin.
Docs: https://github.com/conan-io/docs/pull/3387

This is a follow up to #https://github.com/conan-io/conan/issues/13424#issuecomment-1471974865, item 2 this time.

It would be nice if `cl` and `clang-cl` were consistent wrt `runtime_type`, though since the profile can set the value explicitly, it's not the end of the world if you think this is a bad idea (as far as I've looked `runtime` would only be set with a `cl`-like compiler, so should be good?).

If it's viable:

* I suppose https://docs.conan.io/2/reference/extensions/profile_plugin.html could do with an update but I'm not sure where that lives (a quick check of the docs and web repos was unsuccessful).
* I've removed `runtime_type` from #14664 tests; the now fail without the plugin change and succeed with. Or maybe there should be a test both with and without?

Thanks

### DIFF
--- a/conans/client/profile_loader.py
+++ b/conans/client/profile_loader.py
@@ -30,7 +30,7 @@ _default_profile_plugin = """\
 
 def profile_plugin(profile):
     settings = profile.settings
-    if settings.get("compiler") == "msvc" and settings.get("compiler.runtime"):
+    if settings.get("compiler") in ("msvc", "clang") and settings.get("compiler.runtime"):
         if settings.get("compiler.runtime_type") is None:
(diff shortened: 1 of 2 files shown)


memsharded (approved):
Fantastic, thanks!

Yes, this is viable, the change makes sense, and it has no risk, as ``runtime`` is an user input, they explicitly decide if it is clang with msvc runtime (both clang and clang-cl) or clang with gnu/msys2 backend

Issue #14590, opened 2023-08-28
[question] Why does CMakeToolchain's `cache_variables` convert recipe options to CMake bool but `variables` converts them to strings?


tgurriet:
### What is your question?

Hello,

While exploring the files generated by the CMakeToolchain generator, I realized that when assigning an option to `cache_variables`, it resulted in a CMake `boolean` variable in the `CMakePresets.json`, wherease passing the same option to `variables` generated a CMake `string` variable in the `conan_toolchain.cmake`.

Is there a reason for that difference? I couldn't find anything in the doc about this.

The recipe:
```
from conan import ConanFile
from conan.tools.cmake import CMakeDeps, CMakeToolchain, cmake_layout

class helloRecipe(ConanFile):
    name = "my_lib"
    version = "1.0"

    settings = "os", "compiler", "build_type", "arch"
    options = {"option": [True, False]}
    default_options = {"option": True}
    exports_sources = "CMakeLists.txt", "src/*", "include/*"

    def layout(self):
        cmake_layout(self)

    def generate(self):
        dep = CMakeDeps(self)
        dep.generate()

        tc = CMakeToolchain(self)
        tc.variables["option"] = self.options.get_safe("option")
        tc.cache_variables["option"] = self.options.get_safe("option")
        tc.variables["option2"] = self.options.option
        tc.cache_variables["option2"] = self.options.option
        tc.variables["option3"] = True
        tc.cache_variables["option3"] = True
        tc.generate()
```

The generated CMakePresets.json:
```
...
"cacheVariables": {
  "option": "ON",
  "option2": "ON",
  "option3": "ON",
  "CMAKE_POLICY_DEFAULT_CMP0091": "NEW",
  "CMAKE_BUILD_TYPE": "Debug"
},
...
```
The generated conan_toolchain.cmake:
```
...
# Variables
set(option "True" CACHE STRING "Variable option conan-toolchain defined")
set(option2 "True" CACHE STRING "Variable option2 conan-toolchain defined")
set(option3 ON CACHE BOOL "Variable option3 conan-toolchain defined")
...
```

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

tgurriet:
@memsharded Should I generate a bug report with more info maybe?

franramirez688:
Hi, @tgurriet - Thanks a lot for the report.

[...]

PR #14692, opened 2023-09-07
[MesonToolchain] Showing a warning message if raw options are used

### LINKED ISSUE #14453: [bug] MesonToolchain: project options of type string are not quoted (ERROR: Malformed value in machine file variable)
### Environment details

* Operating System+version: macOS Ventura
* Compiler+version: apple clang 14
* Conan version: 2.0.9
* Python version: 3.11.4

### Steps to reproduce

Create a conan recipe based on Meson, and pass a recipe option of type `["ANY"]` (with a string as default_options) to `project_options`.

For example:

```python
options = {
    "foo": ["ANY"],
    "bar": ["ANY"],
}
default_options = {
    "foo": "/var/run/foo",
    "bar": "BAR",
}

(...)

tc = MesonToolchain(self)
tc.project_options["foo"] = self.options.foo
tc.project_options["bar"] = self.options.bar
tc.generate()
```

The problem is that values are not quoted in generated machine file. Surprisingly, they are quoted if value is passed directly `tc.project_options["foo"] = "/var/run/foo"`, I guess there is something wrong when self.options.foo is resolved.

(I'm not sure, but for the fix you likely want to avoid to quote boolean options)

### Logs

First option gives this error:

```
ERROR: Malformed value in machine file variable 'foo': Unsupported node type.
```

Second option gives this error:

```
ERROR: Undefined constant 'BAR' in machine file variable 'bar'.
```

franramirez688:
Changelog: Feature: MesonToolchain shows a warning message if any options are used directly.
Docs: https://github.com/conan-io/docs/pull/3381
Closes: https://github.com/conan-io/conan/issues/14453

### DIFF
--- a/conan/tools/meson/helpers.py
+++ b/conan/tools/meson/helpers.py
@@ -1,3 +1,4 @@
+from conan.api.output import ConanOutput
 from conan.tools.build.flags import cppstd_msvc_flag
 from conans.errors import ConanException
 from conans.model.options import _PackageOption
@@ -99,6 +100,10 @@ def to_meson_value(value):
         return 'true' if value else 'false'
     elif isinstance(value, list):
         return '[{}]'.format(', '.join([str(to_meson_value(val)) for val in value]))
+    elif isinstance(value, _PackageOption):
+        ConanOutput().warning(f"Please, do not use a Conan option value directly. "
+                              f"Convert 'options.{value.name}' into a valid Python"
+                              f"data type, e.g, bool(self.options.shared)", warn_tag="deprecated")
     return value
 
 
--- a/conan/tools/meson/helpers.py
+++ b/conan/tools/meson/helpers.py
@@ -99,9 +99,6 @@ def to_meson_value(value):
         return 'true' if value else 'false'
     elif isinstance(value, list):
         return '[{}]'.format(', '.join([str(to_meson_value(val)) for val in value]))
-    elif isinstance(value, _PackageOption):
-        raise ConanException("Conan options cannot be specified directly as a valid Meson data type."
-                             " Please, use another Python data type like int, str, list or boolean.")
     return value
 
 
--- a/conans/model/options.py
+++ b/conans/model/options.py
@@ -73,6 +73,10 @@ class _PackageOption:
         return other == self.__str__()
 
     @property
+    def name(self):
+        return self._name
+
+    @property
     def value(self):
         return self._value
 
--- a/conans/test/functional/toolchains/meson/test_meson.py
+++ b/conans/test/functional/toolchains/meson/test_meson.py
@@ -91,6 +91,10 @@ class MesonToolchainTest(TestMesonBase):
                      "main.cpp": app})
         self.t.run("build .", assert_error=True)
         # Using directly the options, the result is -> HELLO_MSG = Hi everyone, which is incorrect
+        assert "WARN: deprecated: Please, do not use a Conan option value directly. " \
+               "Convert 'options.shared' into" in self.t.out
+        assert "WARN: deprecated: Please, do not use a Conan option value directly. " \
+               "Convert 'options.msg' into" in self.t.out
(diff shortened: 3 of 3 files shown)


memsharded (approved):
If this hasn't been released yet, just remove the ``Changelog`` in both PRs

franramirez688:
> If this hasn't been released yet, just remove the `Changelog` in both PRs

Keeping only the feature about showing the warning message

czoido (approved):

SpaceIm:
Is it expected that boolean options display this warning?

In dav1d recipe, there is this warning now:

```
WARN: deprecated: Please, do not use a Conan option value directly. Convert 'options.assembly' into a valid Pythondata type, e.g, bool(self.options.shared)
WARN: deprecated: Please, do not use a Conan option value directly. Convert 'options.with_tools' into a valid Pythondata type, e.g, bool(self.options.shared)
```

Though, in recipe, there is:

```python
    options = {
        "shared": [True, False],
        "fPIC": [True, False],
        "bit_depth": ["all", 8, 16],
        "with_tools": [True, False],
        "assembly": [True, False],
        "with_avx512": [True, False],
    }
    default_options = {
        "shared": False,
        "fPIC": True,
        "bit_depth": "all",
        "with_tools": True,
        "assembly": True,
        "with_avx512": False,
    }

    (...)

    def generate(self):
        env = VirtualBuildEnv(self)
        env.generate()

        tc = MesonToolchain(self)
        tc.project_options["enable_tests"] = False
        tc.project_options["enable_asm"] = self.options.assembly
        if Version(self.version) < "1.0.0":
            tc.project_options["enable_avx512"] = self.options.get_safe("with_avx512", False)
        tc.project_options["enable_tools"] = self.options.with_tools
        if self.options.bit_depth == "all":
            tc.project_options["bitdepths"] = "8,16"
        else:
            tc.project_options["bitdepths"] = str(self.options.bit_depth)
        tc.generate()
```

So why this warning? I though that boolean values were properly converted, unlike string. I don't understand why there can't be an automated conversion of PackageOption like in CMakeToolchain which has:

```python
            elif isinstance(value, _PackageOption):
                if str(value).lower() in ["true", "false", "none"]:
                    cache_variables[name] = "ON" if bool(value) else "OFF"
                elif str(value).isdigit():
                    cache_variables[name] = int(value)
                else:
                    cache_variables[name] = str(value)
```

I would have expected such fix for https://github.com/conan-io/conan/issues/14453, to get consistent behavior of toolchains.

memsharded:
Yes, it is expected. The options have to be explicitly converted to ``bool(self.options.xxx)`` if desired that Meson treats them as booleans. Right now it is working by luck, because being mapped as strings and meson accepts the casing.

SpaceIm:
But why this inconsistency between CMakeToolchain and MesonToolchain? CMakeToolchain converts PackageOption evaluated as boolean or digits or string to their CMake format, but not MesonToolchain. It's weird.

SpaceIm:
I see that @RubenRBS had the same concern https://github.com/conan-io/conan/pull/14633#pullrequestreview-1607751451

franramirez688:
> But why this inconsistency between CMakeToolchain and MesonToolchain? CMakeToolchain converts PackageOption evaluated as boolean or digits or string to their CMake format, but not MesonToolchain. It's weird.

Indeed, I agree with both of you @SpaceIm and @RubenRBS. The thing is that I don't see the point of converting anything like CMakeToolchain is doing under the hood. Actually, it's doing that only for cache_variables and not for the rest of the variables so that's the weirdest thing. IMHO, we should not convert anything at all, and we should manage those conversions through the internal `_PackageOption` class instead somehow. Also, I cannot prune that conversion from CMakeToolchain because of likely breaking changes. It's clear that it's a limitation of this class for now, and we should solve it in the future.

franramirez688:
Related issue: https://github.com/conan-io/conan/issues/14590

PR #14694, opened 2023-09-07
New conan graph explain command to search, compare and explain missing binaries

### LINKED ISSUE #4089: [feature request] Explain why a match was not found (`conan why`)
I have a relatively frequent use-case which could be automated in conan (maybe as a `conan why` command). First, let me show an example how I do this manually and then I'll propose `conan why`.

## The example and manual solution

  * Trying to install a package for a certain profile
  * There is no matching package in the cache or remotes
  * I need to figure out which setting or option is wrong and what's the best match

I'm served with a message that looks like this:

```
PlexMediaServer/1.15.0@plex/stable: WARN: Can't find a 'PlexMediaServer/1.15.0@plex/stable' package for the specified options and settings:
- Settings: arch=x86_64, build_type=Release, compiler=clang, compiler.libcxx=libc++, compiler.version=5.0, os=Macos
- Options: devel=False, sanitize=none, variation=nano, boost:shared=False, bzip2:fPIC=True, bzip2:shared=False, cppnetlib:shared=False, curl:shared=False, fmt:shared=False, freei
mage:shared=False, libidn:shared=False, libxml2:shared=False, minizip:shared=False, opencv:shared=False, plex-openssl:shared=False, soci:shared=False, sqlite:shared=False, taglib
:shared=False, zlib:shared=False
- Package ID: b9cf955f067ee02ab9107d531142201be6422019
```

Now, *in this case* I know I have exactly one package in my cache, so I proceed to list its settings and options:

```
$ conan info PlexMediaServer/1.15.0@plex/stable

[I] ➜ ./conan-bin search PlexMediaServer/1.15.0@plex/stable
Existing packages for recipe PlexMediaServer/1.15.0@plex/stable:

    Package_ID: f5f11a16f57dea41ac9cbfd09c094673f35b4e4f
        [options]
            devel: True
            sanitize: none
            variation: nano
        [settings]
            arch: x86_64
            build_type: Release
            compiler: clang
            compiler.libcxx: libc++
            compiler.version: 5.0
            os: Macos
        [requires]
            boost/1.59.0-37@plex/stable:f3008bb2d07dac7305d36831eff30641d375e5c3
            bzip2/1.0.6-10@plex/stable:0c446de161870e43ba015ddcae0051201b8523e8
            ca-bundle/2018-10-17-1@plex/stable:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
            cmakecache/1-15@plex/stable:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
            cmaketoolchain/1-14@plex/stable:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
            cotire/1.8.0-391bf6b-10@plex/stable:5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9
            cppnetlib/0.10.1-35@plex/stable:78ff5bd100860b10449c8f9203c37913498e118f
            curl/7.56.1-44@plex/stable:82e67269b6c125788ab8cdb27ae6f546537f47be
            fmt/4.1.0-135ab5c-9@plex/stable:27de2ffeada2aa97bda8c9d4a57fea3c49ff6655
            freeimage/3.17.0-13@plex/stable:27de2ffeada2aa97bda8c9d4a57fea3c49ff6655
            libidn/2.0.5-14@plex/stable:2ecea8b4c00f534446d65d264899e590afe91c27
            libxml2/2.9.8-7@plex/stable:f3881625999479f891a7fcd01279b92e7993f60d
            minizip/1.2.8-12@plex/stable:f3881625999479f891a7fcd01279b92e7993f60d
            opencv/2.4.13-07711e4-16@plex/stable:64a735b0d0e2664d702117e3988fb9e3c6b21525
            plex-openssl/1.0.2p-17@plex/stable:f3881625999479f891a7fcd01279b92e7993f60d
            rapidjson/1.1.0-9@plex/stable:27de2ffeada2aa97bda8c9d4a57fea3c49ff6655
            soci/3.0.0-9a9baa1-10@plex/stable:6f085b771649cc265b6b0f42cc878ab40e474a91
            sqlite/3.23.1-13@plex/stable:2ecea8b4c00f534446d65d264899e590afe91c27
            taglib/1.11.1-35@plex/stable:8bedfc069c29e5a10c786ab5501478c56fdc37e9
            universalchardet/1.0.0-9e41626-9@plex/stable:27de2ffeada2aa97bda8c9d4a57fea3c49ff6655
            zlib/1.2.8-12@plex/stable:2ecea8b4c00f534446d65d264899e590afe91c27
        Outdated from recipe: False
```

It's not easy to spot the wrong setting or option there, especially that the error message above listed the full_options, but `search` only lists the package options. I don't know where the mismatch is at this moment.

So I do the following: I dump the sorted options from the conaninfo.txt from the package I'd expect to be a match, and I dump the options from the error output (breaking up the lines & removing commas manually and sorting both files). So now I have files with contents like this (showing only one for brevity):

```
boost:shared=False
bzip2:fPIC=True
bzip2:shared=False
cppnetlib:shared=False
curl:shared=False
devel=True
fmt:shared=False
freeimage:shared=False
libidn:shared=False
libxml2:shared=False
minizip:shared=False
opencv:shared=False
plex-openssl:shared=False
sanitize=none
soci:shared=False
sqlite:shared=False
taglib:shared=False
variation=nano
zlib:shared=False
```

Then I diff them:

```
➜ diff ../../pms/opts1.txt ../../pms/opts2.txt
6c6
< devel=False
---
> devel=True
```

And there I have my answer. 

Needless to say, I was lucky I had only one package. It's much harder to figure out when there are multiple packages. In that case, I usually search with `-q` to reduce the number of hits, but in many cases it's still a lot of legwork to spot what's wrong. I don't always do this diffing process, many times I just scroll up and down and try keep each option and setting in my head.

## Proposed solution: `conan why`

When an install command fails with the error above, I'd like to be able to ask conan what went wrong and what's the closest match. Matches should be ranked by the number of options and settings that matched between the consuming installation context ("wanted") and the package that's being considered for installation. Maybe a percentage value could be used to unify the matching metric. A perfect match is 1.0. My propsed workflow would be (`#` denoting my comments):

```
$ conan install xxxx 
# get the error from above

$ conan why
Failed matches while resolving PlexMediaServer/1.15.0@plex/stable:
  * Considered f5f11a16f57dea41ac9cbfd09c094673f35b4e4f (match: 0.92):
    - Mismatching settings: None
    - Mismatching options:
      - `devel`
        Wanted:  devel=False
        Package: devel=True
  # Listing any further package IDs after this

# Listing any further missing packages the same way after this
```

Ideally, one should be able to limit the number of listed packages, because I'd imagine usually only the first few best matches will be interesting.

I think all the information is present at runtime when `conan install` is executed.
 What are your thoughts?

---

To help us debug your issue please explain:

- [x] I've read the [CONTRIBUTING guide](https://raw.githubusercontent.com/conan-io/conan/develop/.github/CONTRIBUTING.md).
- [n/a] I've specified the Conan version, operating system version and any tool that can be relevant.

### LINKED ISSUE #6770: [feature] Add "Best Matches" Binary Search Mode
This feature request was already somewhat outlined by @fulara in this other ticket, as a possible concept: 

https://github.com/conan-io/conan/issues/6364#issuecomment-600916751

This is a very old problem.  From a user experience perspective, the common scenario is that we know that there are a bunch of binaries are built for some dependency, and we're pretty sure this includes the one we want.  But, when we run Conan we get an error that there's no binary built with the specific package ID hash requested.  The existing search has improved over the years to print out the information about the requested package ID hash so that users could deduce the mismatch.  

The "problem" is that this is very onerous, tedious, and time-consuming a lot of the time. It relies on the user to perform a visual comparison setting-by-setting, option-by-option, between the error message and EACH search result in the list of search results to HOPEFULLY see the problem.  Alternatively, users can manually construct a query for the `conan search -q` which contains all the settings they're trying to use.  This reduces the comparison to just looking at all the different options and values.   Furthermore, if the difference in `package_id` is related to the list of dependencies, it's very hard to see it, especially for newcomers. 

Matching binary package ID's is central to Conan's functionality, and this situation comes up frequently enough that it feels like a big hole in the UX overall.

The suggestion in the aforementioned ticket is a sound strategy for providing a new search feature that improves the UX around this.  From a search perspective, each binary is effectively describable by it's matrix of options, settings, and dependencies (pre-hash).  The suggested conan search feature is effectively to do a matrix comparison against the requested but unhashed values and the unhashed values of the binaries in the available repositories. Conan would then need to use some sort of scoring/distinace/best match algorithm and then present to the user a list of results in order of "best score/best match" first.  For each result, the output should display the "DIFF" with the requested binary so users can easily see what's different.  

Logically, this is reasonable.  Based on how the Conan dependency tree graph is calculated, dynamic options and settings are reconciled, and then everything is factored into the `package_id` .... the implementation of this might be very difficult.  In any case, I think it's fair to say that "best match" search is reasonable, and would be very useful here.  I will try to do some research into established "Matrix comparison" methods with best match and diff functionality. 

### LINKED ISSUE #14464: [question] filtering packages with "conan list" based on an option not part of the package id
### What is your question?

Hi,

consider `gtest/1.13.0` which has an option [`no_main`](https://github.com/conan-io/conan-center-index/blob/master/recipes/gtest/all/conanfile.py#L26) which is not part of the [package_id](https://github.com/conan-io/conan-center-index/blob/master/recipes/gtest/all/conanfile.py#L83).

No mather which value you try to filter for such an option, this will not find a package:
```
$ conan list "gtest/1.13.0:*" -p "options.no_main=False"
Local Cache
  gtest
    gtest/1.13.0
      revisions
        53120f2a905d958ebe9cb0dca34a2432 (2023-07-26 07:55:02 UTC)
          packages
$ conan list "gtest/1.13.0:*" -p "options.no_main=True"
Local Cache
  gtest
    gtest/1.13.0
      revisions
        53120f2a905d958ebe9cb0dca34a2432 (2023-07-26 07:55:02 UTC)
          packages
```
This is not obvious to someone who doesn't know the recipe:
```
$ conan inspect $(conan cache path gtest/1.13.0)               
default_options:
    shared: False
    fPIC: True
    build_gmock: True
    no_main: False
    hide_symbols: False
    debug_postfix: d
description: Google's C++ test framework
generators: []
homepage: https://github.com/google/googletest
label: 
license: BSD-3-Clause
name: gtest
options:
    build_gmock: True
    debug_postfix: d
    fPIC: True
    hide_symbols: False
    no_main: False
    shared: False
options_definitions:
    shared: ['True', 'False']
    fPIC: ['True', 'False']
    build_gmock: ['True', 'False']
    no_main: ['True', 'False']
    hide_symbols: ['True', 'False']
    debug_postfix: ['ANY']
package_type: None
requires: []
revision_mode: hash
settings: ['os', 'arch', 'compiler', 'build_type']
topics: ['testing', 'google-testing', 'unit-test']
url: https://github.com/conan-io/conan-center-index
```

Now the problem is, that this option `no_main` is outputted in the json output from `conan create` and I create filters based on this report e.g. to find a "Debug" package of a build for a debugging tool.

I have no way of removing this option from the filter creation as I can't get the info that the option is not part of the package_id.
How to deal with this situation?

### LINKED ISSUE #14810: [feature] A friendlier output when pre-built packages are missing (precisely listing the config diffs).
### What is your suggestion?

This one might be difficult, but it's also quite important for us.

If at all possible (I know already the answer is no), back-porting this feature to conan 1 would really help. We've been refactoring like crazy but we aren't even close to switching to conan 2 :)

Conan would do much better having a beginner friendly (first time user understandable) message when it cannot find pre-built packages for a given config / profile. It's hard to tell, even with configs you control yourself, what is clashing.

An example ideal output :
```
Hey buddy, we couldn't find any pre-built packages for your current configuration.
You may pass --build missing to build the package on your system.

Asked for config :
[settings]
bla
bla
bla
*flaflee* # in bold, closest matching configs dont have this option
[options]
bla
de
bla

Closest matching configurations found :
[settings]
bla
*notquitebla* # in bold, showing the difference
bla
[options]
bla
de
bla

# more profiles that are close, see below
```

Figuring out what configs are closest might be tricky, but also not that much.
You could go the simple route of comparing available configs, listing configs that have the least differences (multiple if more than one have the least amount of differences).

Once in a while it'll give weird results, for example a different OS maybe. But when that happens it should immediately raise flags and help us figure out the problem.

I know this is a big one, but we really *really* need something like this to stop relying on `--build missing`.
Thanks!

memsharded:
Changelog: Feature: New ``conan graph explain`` command to search, compare and explain missing binaries.
Docs: https://github.com/conan-io/docs/pull/3486
Close https://github.com/conan-io/conan/issues/14464
Close https://github.com/conan-io/conan/issues/14810
Close https://github.com/conan-io/conan/issues/4089
Close https://github.com/conan-io/conan/issues/6770


conan/api/subapi/list.py
```diff
@@ -84,9 +84,23 @@ def filter_packages_configurations(pkg_configurations, query):
...
+
+        def _matches(d):  # Only implements settings matching at this moment
+            settings = d.get("settings", {})
+            for k, v in settings.items():
+                profile_value = profile_settings.get(k)
+                if profile_value is not None and profile_value != v:
+                    return False
+            return True
+
+        return {pref: data for pref, data in pkg_configurations.items() if _matches(data)}
```
franramirez688: Would it be interesting to have implemented the `__eq__` operator on the Settings side? I see that `SettingsItem` already has it so I think it would make sense.

conan/cli/commands/list.py
```diff
@@ -102,6 +102,8 @@ def list(conan_api: ConanAPI, parser, *args):
     parser.add_argument('-p', '--package-query', default=None, action=OnceArgument,
                         help="List only the packages matching a specific query, e.g, os=Windows AND "
                              "(arch=x86 OR compiler=gcc)")
+    parser.add_argument('-pr', '--profile', default=None, action="append",
+                        help="Profiles to filter the package binaries")
```
franramirez688: I'm not sure if this param name is clear enough to say "filtering packages by this profile". We already have the `-p` param so I think it could be a bit confusing.

conans/test/integration/command_v2/list_test.py
```diff
@@ -663,6 +663,34 @@ def test_list_empty_settings():
     assert pkgs == {NO_SETTINGS_PACKAGE_ID: {"info": {}}}
 
 
+def test_list_package_filter_profile():
+    c = TestClient(default_server_user=True)
+    c.save({"pkg/conanfile.py": GenConanfile("pkg", "1.0"),
+            "other/conanfile.py": GenConanfile("other", "1.0").with_settings("os"),
+            "windows": "[settings]\nos=Windows",
+            "linux": "[settings]\nos=Linux"})
```
franramirez688: I prefer to see a more complex test using several settings lines.

franramirez688 (commented):
It's looking great! Only a few observations/doubts about the changes.

AbrilRBS:
Pinging @danimtb which just spoke about this a few days ago

danimtb (commented):
I think it is fine for an initial implementation, although my main request would be to make it work with options as well. Also, as a nice feedback for the user, it would be awesome to output (maybe via stderr?) the closest match for the profile provided (but I guess that could be done in the future)

memsharded:
> I think it is fine for an initial implementation, although my main request would be to make it work with options as well. Also, as a nice feedback for the user, it would be awesome to output (maybe via stderr?) the closest match for the profile provided (but I guess that could be done in the future)

Sure, the ``options`` will be added, it was just the POC idea.

I'll check about the ``closest`` binary thing.

franramirez688:
@memsharded is this supposed to happen?
```bash
$ conan listx find-binaries "hell/1.0.0" -pr proof
Finding binaries in the cache
Finding binaries in remote conancenter
Local Cache
  hell/1.0.0
    hell/1.0.0#3f7c271fed307c6bee398d19e00f8b28 (2023-11-27 15:32:04 UTC)
      hell/1.0.0#3f7c271fed307c6bee398d19e00f8b28:40a4bd5c82fcddf7dcfbe92d65687290441859b6
        settings: Macos, x86_64, Release, apple-clang, 17, libc++, 15
        options: fPIC=True, shared=False
        diff
          settings: Debug
          explanation: This binary was built with different settings (compiler, build_type).
conancenter
  WARN: There are no matching recipe references
```

Where `proof` profile contains:
```
[options]
*:shared=False
[settings]
build_type=Debug
```

There are no binaries built with Debug mode, but it's showing me one binary with those `diff` points.

franramirez688:
@memsharded yes, I just saw some tests showing the closest binary found.

memsharded:
Yes, the report is as follow:
- The ``conan listx find-binaries`` report for each binary:
```
settings: Macos, x86_64, Release, apple-clang, 17, libc++, 15
options: fPIC=True, shared=False
```
This means this binary is built with ``Release``.

Then the ``diff`` says:
```
diff
          settings: Debug
          explanation: This binary was built with different settings (compiler, build_type).
```
It means that you are asking for a ``Debug`` binary, but the found one is different in this ``setting``.

AbrilRBS (approved):

AbrilRBS (approved):

conan/cli/commands/graph.py
```diff
@@ -179,3 +197,75 @@ def graph_info(conan_api, parser, subparser, *args):
...
+                conaninfo = node.conanfile.info
+                break
+    else:
+        raise ConanException("There is no missing binary")
+
+    pkglist = conan_api.list.explain_missing_binaries(ref, conaninfo, remotes)
+
+    ConanOutput().title("Closest binaries")
+    return {
+        "Closest binaries": pkglist.serialize(),
```
franramirez688: I think that this should be `closest_binaries` instead. It only represents the namespace for the closest binaries section, so I believe that it'd be better to use a variable following the Python syntax.

franramirez688 (approved):
This is great! 👏 
I've only suggested one minor thing.

czoido (approved):

franramirez688 (approved):

PR #14705, opened 2023-09-08
Handle the case where `remotes` is `None`.

cqc-alec:
Changelog: (Bugfix): Handle the case where `remotes` is `None` in `evaluate_graph()`

This fixes a regression in conan 2.0.10, introduced in #14467 .

The `remotes` argument to `evaluate_graph()` may be `None`, as is shown by the following stack trace from a `conan export-pkg` command:

```
   File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conan\cli\commands\export_pkg.py", line 93, in export_pkg
    deps_graph = run_test(conan_api, test_conanfile_path, ref, profile_host, profile_build,
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conan\cli\commands\test.py", line 70, in run_test
    conan_api.graph.analyze_binaries(deps_graph, build_modes, remotes=remotes, update=update,
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conan\api\subapi\graph.py", line 201, in analyze_binaries
    binaries_analyzer.evaluate_graph(graph, build_mode, lockfile, remotes, update,
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conans\client\graph\graph_binaries.py", line 344, in evaluate_graph
    self._evaluate_node(node, build_mode, remotes, update)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conans\client\graph\graph_binaries.py", line 137, in _evaluate_node
    self._process_node(node, build_mode, remotes, update)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conans\client\graph\graph_binaries.py", line 196, in _process_node
    self._evaluate_download(node, remotes, update)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conans\client\graph\graph_binaries.py", line 261, in _evaluate_download
    self._get_package_from_remotes(node, remotes, update)
  File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\conans\client\graph\graph_binaries.py", line 54, in _get_package_from_remotes
    for r in remotes:
TypeError: 'NoneType' object is not iterable
```

The above-mentioned PR removed handling of that case.

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -341,7 +341,7 @@ class GraphBinariesAnalyzer(object):
                     if locked_prev:
                         self._process_locked_node(node, build_mode, locked_prev)
                         continue
-                self._evaluate_node(node, build_mode, remotes, update)
+                self._evaluate_node(node, build_mode, remotes or [], update)
 
         self._skip_binaries(deps_graph)
 


memsharded (commented):
Hi @cqc-alec 

I'd prefer to start it with a unit test (integration folder, most likely) that can reproduce the bug, in this way we make sure that we cover that scenario for the future, because clearly there is a gap there. 
Then, it sounds that the fix might happen earlier in the chain, but that we can discuss when we have the repro test. Do you think you could do it? If not, please detail a bit more the steps to reproduce, so we can contribute it. Thanks!

cqc-alec:
Thank you @memsharded , I agree a unit test would be good! I am having a bit of difficulty making a minimal test case (it shows up with a rather complex project). Something like this should do it:

```shell
conan new cmake_lib -d name=basic -d version=1.0
conan install .
conan build .
conan export-pkg .
```

Except that when the last command reaches `GraphBinariesAnalyzer._process_node()` (in `conans/client/graph/graph_binaries.py`), it finds something in cache (`cache_latest_prev` is not `None`), so it does _not_ call `_evaluate_download()`. If it did, it would call `_get_package_from_remotes()` and hit the error because
`remotes = None`.

So I'm now starting to wonder if the real bug is somewhere else.

This is where the problem actually arose: https://github.com/CQCL/tket/blob/1efc10c391c5ce5d68cccd4bcec6038ba6ef4080/.github/workflows/build_and_test.yml#L197

(It's not Windows-specific; I can reproduce it locally on Linux with this project.)

memsharded:
I have managed to reproduce, but with kind of a very edge case, this is the test:
```python
def test_remote_none():
    # https://github.com/conan-io/conan/pull/14705
    c = TestClient(default_server_user=True)
    c.save({"dep/conanfile.py": GenConanfile("dep", "0.1"),
            "pkg/conanfile.py": GenConanfile("pkg", "0.1"),
            "pkg/test_package/conanfile.py": GenConanfile().with_test("pass").with_requires("dep/0.1")})
    c.run("create dep")
    c.run("upload dep* -r=default -c")
    c.run("install pkg")
    c.run("build pkg")
    c.run("remove dep*:* -c")
    c.run("export-pkg pkg")  # This crashes
    print(c.out)
```
This means that:
- the export-pkg package must have a ``test_package``
- the ``test_package/conanfile.py`` must have some extra dependencies, not just the tested dependency
- The binary for the extra dependency must not be in the local cache, but only the recipe

Could you please verify this is the issue in your project? Thanks!

cqc-alec:
@memsharded Good that you've reproduced it! My case does seem slightly different, however, in that the `test_package/conanfile.py` does not have any extra dependencies. It is very simple: https://github.com/CQCL/tket/blob/develop/tket/test_package/conanfile.py .

I think you should be able to reproduce the issue if you do the following:

```shell
git checkout git@github.com:CQCL/tket.git
cd tket
conan profile detect
conan remote add tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs --index 0
conan install tket -o boost/*:header_only=True
conan build tket -o boost/*:header_only=True
conan export-pkg tket -o boost/*:header_only=True
```

(It does take a while to build; sorry about that.)

cqc-alec:
(But in the end, if you have a test that shows the issue, and the fix fixes my issue, maybe that is enough. ;-) )

memsharded:
Quick feedback: In Conan 2.0 you don't need ``conan install`` + ``conan build``, as ``conan build`` is already internally doing the ``conan install`` step.

Thanks very much for the feedback!
I have managed to reproduce with your instructions, and created a new unit test to ensure that such case is also covered by the fix. It is in https://github.com/conan-io/conan/pull/14712, it will be in 2.0.11, we will try to release it soon.

cqc-alec:
Brilliant, thanks very much!

franramirez688:
Closed by https://github.com/conan-io/conan/pull/14712

This change will be released in the next Conan v2.0.11 version.

Issue #648, opened 2016-11-09
Use LZMA compression for transmission of binary packages


annulen:
Gzip is not really efficient for compressing large amounts of binary data. Also, LZMA has special filters improving compression ratio for executable code. You could use 7z or tar.xz for transmission of packages

memsharded:
Thanks for telling. I think this might be a duplicate to: https://github.com/conan-io/conan/issues/52

We never worked on this feature because a preliminary attempt showed that this will be very complicated from the operations point of view: creating the conan application for PyPi, creating the installers that bundle it for different systems (Win installer, deb, archlinux, OSX homebrew...), as the mainstream python version is still Python2.7, that do not include such functionality.

However, this would be a very nice to have functionality, and very decoupled, not necessary to know about the very internals of conan, if someone knows how to do it and is able to do the work, it will be very welcome.

annulen:
This is not a duplicate of #52 as that issue is talking about use of 7z in conanfile code, though it seems to be a necessary prerequisite.

memsharded:
Yes, you are right. Basically the underlying problem is the same, once provided support for LZMA in conan, such issue #52 would be trivial.

tru:
@memsharded would you be open to having some kind of opt-in to use lzma for packages instead of gzip? lzma makes a LOT of sense for the package storage and transfer, and in our case we can guarantee that all clients are using python3. Would a environment or conan.conf option for uploading .xz packages be something you would accept if we created such a patch?

lasote:
The problem is maintaining the installers and pypi package, the distinction between py2 and py3 to manage will be a pain. Also managing different compressed files on the server side could be very hard to control if the users start to mix them in the public repositories. It's possible but it looks like a nightmare for us.

tru:
I see that - I am not really that worried about the public repos (selfish I know), but in our closed ecosystem this would still be a huge win for us.

memsharded:
I wouldn't oppose to have a branch with it, see how it looks like, and think if there is some way to guarantee 100% that it won't be possible to use it in public repos.

lasote:
I think probably we should probably wait for the definitive dropping of python 2 support. But maybe we should start implementing in the next Conan release a kindly error when a `conan_package.7z` is found to be ready to the future when (probably Conan 2.x) starts to upload 7z packages to public repositories. Then past versions from 1.6 could raise nicely saying: "The package xxx uses 7z compression and your conan version doesn't support it, please upgrade to the latest".
[...]

PR #14706, opened 2023-09-08
Add support for zstd compression

grossag:
Changelog: (Feature): Add support for zstd compression
Docs: Will create one if this PR is acceptable

As discussed in issue #648, this change adds zstd support to conan in the following ways:

1. The person or build running `conan upload` can set a config value `core.upload:compression_format = zstd` to upload binaries using zstd instead of gzip.
2. The zstd compression is done entirely in Python using a combination of `tarfile` and `python-zstandard`. Then the file is uploaded as normal.
3. When downloading packages, if a .tar.zst file is encountered, the extraction code uses `tarfile` and `python-zstandard` to extract.
4. Adds a test to cover zstd compression and decompression.

I chose python-zstandard as the library because that is what urllib3 uses. The package has not yet hit 1.0 but urllib3 is a mature package and it says a lot to me that they chose python-zstandard.

I apologize in advance if I'm missing important parts of the developer workflow. If this approach is acceptable, I'll create a docs PR as requested.

Developer docs on all branches say to open pull requests against `develop` but AFAICT that is Conan 1.x. I'm opening this against `develop2` instead because that appears to be Conan 2.x; I hope that's the right thing to do.


grossag:
I am working through my company's CLA approval process and hope to sign it by end of day today. In the meantime, I wrote a script to test compression and decompression of a test folder using various gzip and zstd compression levels and ran it overnight on a 7.1GB folder with 16000 files. I put the script here in case you all find it useful: https://gist.github.com/grossag/525f3cdaf7d985b625a38df55a7c9087

Run on a VM using shared NAS storage:
```
gzip level 7:
	- Compression time: 554.04 seconds
	- Compression size: 1.988 GB
	- Decompression times in seconds: 89.50 mean, 88.05 median, 2.04 stdev
gzip level 8:
	- Compression time: 1079.89 seconds
	- Compression size: 1.978 GB
	- Decompression times in seconds: 87.20 mean, 87.05 median, 1.84 stdev
gzip level 9 (default compression level):
	- Compression time: 2080.31 seconds
	- Compression size: 1.976 GB
	- Decompression times in seconds: 85.34 mean, 86.40 median, 3.10 stdev

zstd level 3 (default compression level):
	- Compression time: 136.69 seconds
	- Compression size: 1.498 GB
	- Decompression times in seconds: 54.33 mean, 54.14 median, 4.04 stdev
zstd level 4:
	- Compression time: 125.80 seconds
	- Compression size: 1.478 GB
	- Decompression times in seconds: 52.21 mean, 52.98 median, 2.19 stdev
zstd level 5:
	- Compression time: 115.56 seconds
	- Compression size: 1.398 GB
	- Decompression times in seconds: 50.87 mean, 51.97 median, 4.49 stdev
```

My work laptop has widely varying performance right now, where zstd decompression of the same files jumps between 47 and 60 seconds. So here are the first results which I need to rerun:
```
gzip level 9 (default compression level):
	- Compression time: 1144.69 seconds
	- Compression size: 1.975 GB
	- Decompression times in seconds: 94.60 mean, 97.29 median, 8.86 stdev
zstd level 5:
	- Compression time: 69.69 seconds
	- Compression size: 1.394 GB
	- Decompression times in seconds: 55.63 mean, 59.06 median, 8.00 stdev
```

zstd is interesting because my results are showing that decompression time doesn't change as you increase the compression level, maybe with the exception of the really high levels 20-22. But overall my results are summarized as: on both machines, zstd level 5 shows a size reduction of 30% and a decompression time reduction of 35-40% as compared to gzip level 9.

grossag:
Looks like my virus scanner was causing the high variance in hash performance testing. Here are some results, comparing zstd level 9 with gzip level 9 on my laptop:

Boost (1.1GB and 15000 files):
```
gzip level 9 (default compression level):
	- Compression time: 178.97 seconds
	- Compression size: 192.741 MB
	- Decompression times in seconds: 9.89 mean, 9.95 median, 0.09 stdev
zstd level 9:
	- Compression time: 12.34 seconds
	- Compression size: 130.144 MB
	- Decompression times in seconds: 6.99 mean, 7.01 median, 0.10 stdev
```

Compiler toolset (7.1 GB and 16000 files):
```
gzip level 9 (default compression level):
	- Compression time: 1473.80 seconds
	- Compression size: 1.975 GB
	- Decompression times in seconds: 34.64 mean, 34.54 median, 0.23 stdev
zstd level 9:
	- Compression time: 91.72 seconds
	- Compression size: 1.258 GB
	- Decompression times in seconds: 17.33 mean, 16.95 median, 1.01 stdev
```

So my tests are still showing 20-50% improvements in decompression time.

13steinj:
one way or the other I'll have to implement this + more for my org eventually--

can this be changed to be done in an expandable manner? Something _like_:

```
core.packager.binaries.compressor.native = false # one of true, false; true uses a command
core.packager.binaries.compressor = gzip  # one of pigz, gzip, bzip2, xz, lzip, lzma, lzop, gzip, zstd; if native, also arbitrary
core.packager.binaries.compressor.suffix = auto  # one of auto, or if compressor.native AND unknown compressor, custom defaulting to first word of compressor (program)
core.packager.binaries.archiver = python, native # one of python, tar. Defaults to decompressor, native defaults to tar. 
core.packager.binaries.decompressor = python # one of python, tar (auto detect default), or native (based off of suffix
core.packager.binaries.dearchiver = python  # one of python, native (tar). Defaults to decompressor (tar and native defaults to tar). 
```

maybe some other variations... bit of a hard problem to make this workable for everyone.

grossag:
> one way or the other I'll have to implement this + more for my org eventually--
> 
> 
> 
> can this be changed to be done in an expandable manner? Something _like_:
> 
> 
> 
> ```
> 
> core.packager.binaries.compressor.native = false # one of true, false; true uses a command
> 
> core.packager.binaries.compressor = gzip  # one of pigz, gzip, bzip2, xz, lzip, lzma, lzop, gzip, zstd; if native, also arbitrary
> 
> core.packager.binaries.compressor.suffix = auto  # one of auto, or if compressor.native AND unknown compressor, custom defaulting to first word of compressor (program)
> 
> core.packager.binaries.archiver = python, native # one of python, tar. Defaults to decompressor, native defaults to tar. 
> 
> core.packager.binaries.decompressor = python # one of python, tar (auto detect default), or native (based off of suffix
> 
> core.packager.binaries.dearchiver = python  # one of python, native (tar). Defaults to decompressor (tar and native defaults to tar). 
> 
> ```
> 
> 
> 
> maybe some other variations... bit of a hard problem to make this workable for everyone.

Hey, thanks for the review! What is your use case that requires this additional customization? I found that using a separate tar process made things difficult to manage and was actually a tiny bit slower than in-proc python-zstandard.

13steinj:
> Hey, thanks for the review! What is your use case that requires this additional customization? I found that using a separate tar process made things difficult to manage and was actually a tiny bit slower than in-proc python-zstandard.

With respect to native vs non-native (subprocess or python-level), my experience has unfortunately been that the parallel downloads feature is "not actually parallel" because tar extraction is not parallel (partially due to the GIL, partially due to how the code is structured). This was tested on conan 1.5X. On large binary packages and no other core/job restrictions, `parallel_downloads` was fastest set to _2_ rather than 16, with a large amount of time spent wasted in `tarfile` :cry: .

While at a previous org, monkeypatch-experimenting (because everything is python, yay!) to replace with a native call to `tar` + `pigz` (so called "fake" parallelism for decompression) was faster, and I expect [`pugz`](https://github.com/Piezoid/pugz) to be even faster.

This isn't to say I don't want this feature you've written, I do! But with conan's committal to backwards compatibility in 2.0, I would expect that config options need to either have a lot of granularity in order to suffice for future use cases (for example, some binary packaged data that I've had played with over the past year suggests that `bz2 ` is optimal instead).

I'm less so asking for additional customization right now and more so for the config to be structured so that additional customization can be added later. Ex `core.upload` may be a poor choice, and there is already, unfortunately, `core.gzip:compresslevel` that I assume would be better off under some sub-namespace but now has to work  for the foreseeable future.

conans/client/cmd/uploader.py
```diff
@@ -165,11 +167,22 @@ def _prepare_package(self, pref, prev_bundle):
...
+        if compression_format == "gzip":
+            compress_level_config = "core.gzip:compresslevel"
+            package_file_name = PACKAGE_TGZ_NAME
+            package_file = os.path.join(download_pkg_folder, PACKAGE_TGZ_NAME)
+        elif compression_format == "zstd":
+            compress_level_config = "core.zstd:compresslevel"
+            package_file_name = PACKAGE_TZSTD_NAME
+            package_file = os.path.join(download_pkg_folder, PACKAGE_TZSTD_NAME)
+        else:
+            raise ConanException(f"Unsupported compression level '{compression_format}'")
```
13steinj: I'll review while here...
```suggestion
            raise ConanException(f"Unsupported compression format '{compression_format}'")
```

conans/client/cmd/uploader.py
```diff
@@ -165,11 +167,22 @@ def _prepare_package(self, pref, prev_bundle):
 
     def _compress_package_files(self, layout, pref):
         download_pkg_folder = layout.download_package()
-        package_tgz = os.path.join(download_pkg_folder, PACKAGE_TGZ_NAME)
-        if is_dirty(package_tgz):
-            self._output.warning("%s: Removing %s, marked as dirty" % (str(pref), PACKAGE_TGZ_NAME))
-            os.remove(package_tgz)
-            clean_dirty(package_tgz)
+        compression_format = self._app.cache.new_config.get("core.upload:compression_format") or "gzip"
```
13steinj: is `...new_config` not a dictionary? aka 
```suggestion
        compression_format = self._app.cache.new_config.get("core.upload:compression_format", "gzip")
```

conans/client/cmd/uploader.py
```diff
@@ -165,11 +167,22 @@ def _prepare_package(self, pref, prev_bundle):
...
+            package_file = os.path.join(download_pkg_folder, PACKAGE_TGZ_NAME)
+        elif compression_format == "zstd":
+            compress_level_config = "core.zstd:compresslevel"
+            package_file_name = PACKAGE_TZSTD_NAME
+            package_file = os.path.join(download_pkg_folder, PACKAGE_TZSTD_NAME)
+        else:
+            raise ConanException(f"Unsupported compression level '{compression_format}'")
+
+        if is_dirty(package_file):
+            self._output.warning("%s: Removing %s, marked as dirty" % (str(pref), package_file_name))
```
13steinj: conan requires python >= 3.6, should probably be using f-strings
```suggestion
            self._output.warning(f"{pref!s}: Removing {package_file_name!s}, marked as dirty")
```

grossag:
> > Hey, thanks for the review! What is your use case that requires this additional customization? I found that using a separate tar process made things difficult to manage and was actually a tiny bit slower than in-proc python-zstandard.
> 
> With respect to native vs non-native (subprocess or python-level), my experience has unfortunately been that the parallel downloads feature is "not actually parallel" because tar extraction is not parallel (partially due to the GIL, partially due to how the code is structured). This was tested on conan 1.5X. On large binary packages and no other core/job restrictions, `parallel_downloads` was fastest set to _2_ rather than 16, with a large amount of time spent wasted in `tarfile` 😢 .
> 
> While at a previous org, monkeypatch-experimenting (because everything is python, yay!) to replace with a native call to `tar` + `pigz` (so called "fake" parallelism for decompression) was faster, and I expect [`pugz`](https://github.com/Piezoid/pugz) to be even faster.
> 
> This isn't to say I don't want this feature you've written, I do! But with conan's committal to backwards compatibility in 2.0, I would expect that config options need to either have a lot of granularity in order to suffice for future use cases (for example, some binary packaged data that I've had played with over the past year suggests that `bz2 ` is optimal instead).
> 
> I'm less so asking for additional customization right now and more so for the config to be structured so that additional customization can be added later. Ex `core.upload` may be a poor choice, and there is already, unfortunately, `core.gzip:compresslevel` that I assume would be better off under some sub-namespace but now has to work for the foreseeable future.

This is something I would want direction from the maintainers on if I was to do. In the discussion with @memsharded in https://github.com/conan-io/conan/issues/648 , the idea of deferring compression and decompression to native tools was not ideal because of testing and compatibility concerns. That is why I tried to do this in all Python code. I am very happy with the Python zstd decompression performance so far across Windows, Mac, and Linux. The python-zstandard library releases the GiL before calling into the Zstd C library, so you aren’t losing performance there.

This change represents the most minimal one I could do while still accomplishing my goals. On the GH issue I referenced, accepting this PR was not guaranteed so I wanted to limit the complexity to show that it is supportable long-term.

13steinj:
Fair enough; I'm mainly concerned about other package types and how this interacts with compatibility concerns.

To be clear I'm not suggesting you implement all of these methods right now, just for the config key chosen to be extensible for the future.

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
+            if compresslevel is not None:
+                zstd_kwargs["level"] = compresslevel
+
+            dctx = zstandard.ZstdCompressor(**zstd_kwargs)
+
+            # Create a zstd stream writer so tarfile writes uncompressed data to
+            # the zstd stream writer, which in turn writes compressed data to the
+            # output tar.zst file.
+            with dctx.stream_writer(tarfile_obj) as stream_writer:
+                with tarfile.open(mode="w|", fileobj=stream_writer,
```
Ext3h: You may want to increase the `bufsize` argument to around 512kB for both reading and writing. The default of 10kB is oddly chosen to best match exactly 5 TAR records, but it's hitting way too low for modern storage types / is rendering hot loops inside `tarfile.py` CPU-bound.

ZSTD also works even better with larger block sizes, and since you don't have latency constraints, you can go wild. (Up to the point where buffer reallocation is a bottleneck...)

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
-    with set_dirty_context_manager(tgz_path), open(tgz_path, "wb") as tgz_handle:
-        tgz = gzopen_without_timestamps(name, mode="w", fileobj=tgz_handle,
-                                        compresslevel=compresslevel)
-        for filename, abs_path in sorted(files.items()):
-            # recursive is False in case it is a symlink to a folder
-            tgz.add(abs_path, filename, recursive=False)
-        tgz.close()
+
+    if compressformat == "zstd":
+        with open(tar_path, "wb") as tarfile_obj:
```
Ext3h: This code should have gone next to https://github.com/grossag/conan/blob/a33394daa263e19fa0929ed73d95eca90bdc8e87/conans/util/files.py#L247 - not inline?

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
+            def reset_tarinfo(tarinfo):
```
Ext3h: This is actually not doing the same, at all. You are now stripping the `mtime` from each individual file *inside* the tar archive.

This is not the same as `gzopen_without_timestamps()` which merely omitted an extra timestamp *outside* the tar archive, added to each compressed chunk in `GzipFile`.

conans/client/cmd/uploader.py
Ext3h: Btw.: I suspect this is also part of the reason why the Gzip path is so slow / GIL heavy. ~~And probably also why it compresses so badly - as this means restarting compression every 32kB while the format is designed to use true stream compression of *huge* continuous memory segments with up to 64kB dictionaries and deduplication with 32kB sliding windows.~~

EDIT: GZip has jet another BufferedWriter allocated internally, and *is* doing stream compression. But 10kB is still missing any sane default size for file handling.

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
+            # Create a zstd stream writer so tarfile writes uncompressed data to
+            # the zstd stream writer, which in turn writes compressed data to the
+            # output tar.zst file.
+            with dctx.stream_writer(tarfile_obj) as stream_writer:
+                with tarfile.open(mode="w|", fileobj=stream_writer,
+                                  format=tarfile.PAX_FORMAT) as tar:
+                    for filename, abs_path in sorted(files.items()):
+                        tar.add(abs_path, filename, recursive=False, filter=reset_tarinfo)
+    else:
+        # FIXME, better write to disk sequentially and not keep tgz contents in memory
```
Ext3h: Unrelated , but this comment appears to be outdated. There is nothing kept in memory apart from the buffer inside the tarfile  instance and some `BufferedWriter` instances.

conans/client/remote_manager.py
```diff
@@ -255,8 +265,15 @@ def uncompress_file(src_path, dest_folder, scope=None):
         if big_file:
             hs = human_size(filesize)
             ConanOutput(scope=scope).info(f"Decompressing {hs} {os.path.basename(src_path)}")
+
         with open(src_path, mode='rb') as file_handler:
-            tar_extract(file_handler, dest_folder)
+            if src_path.endswith(".tar.zst"):
+                dctx = zstandard.ZstdDecompressor()
```
Ext3h: This should also go with `tar_extract` instead of being spread out inline.

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
+                """
+                tarinfo.mtime = 0
+                return tarinfo
+
+            # Only provide level if it was overridden by config.
+            zstd_kwargs = {}
+            if compresslevel is not None:
+                zstd_kwargs["level"] = compresslevel
+
+            dctx = zstandard.ZstdCompressor(**zstd_kwargs)
```
Ext3h: You should be applying `write_checksum=True`, otherwise this will run by default in an unsafe mode where corrupted data is potentially still decompressed.

conans/client/cmd/uploader.py
```diff
@@ -246,21 +263,48 @@ def upload_package(self, pref, prev_bundle, remote):
...
+
+            # Create a zstd stream writer so tarfile writes uncompressed data to
+            # the zstd stream writer, which in turn writes compressed data to the
+            # output tar.zst file.
+            with dctx.stream_writer(tarfile_obj) as stream_writer:
+                with tarfile.open(mode="w|", fileobj=stream_writer,
+                                  format=tarfile.PAX_FORMAT) as tar:
+                    for filename, abs_path in sorted(files.items()):
+                        tar.add(abs_path, filename, recursive=False, filter=reset_tarinfo)
```
Ext3h: You will most likely want to issue an `flush(zstandard.COMPRESSOBJ_FLUSH_BLOCK)` whenever the written file size exceeds a couple of megabytes.

This has no relevance for Conan as a client, as it's only accessing the compressed file as a stream, but it is very much relevant for Artifactory and alike as they can build on that to properly ***index*** into for random access into the .tar.zstd, as de-compression can be restarted after every frame.

It's still entirely compatible with the tar / zstd command line tools.

Ext3h:
> - Decompression times in seconds: 17.33 mean, 16.95 median, 1.01 stdev

This still ain't looking right. This is under performing by a factor 2-3x compared to what it should look like.

There is likely some major overhead as tarfile is serializing an embarrassingly parallel problem by hopping between blocking file system accesses and decompression in a single thread...

conans/client/cmd/uploader.py
Ext3h: You should be exposing/using the `threads` attribute for compression:  https://python-zstandard.readthedocs.io/en/latest/multithreaded.html

You trade a minimal loss in compression ratio for a pretty good scaling in compression speed.

This also decouples `tarfile.TarFile`'s processing from the actual compression pass, so even `threads = 1` already provides a measurable speedup!

Ext3h:
https://github.com/python/cpython/issues/106980 The issue with `extractall` was already known upstream.

conans/client/cmd/uploader.py
grossag: I'd like to leave this one separate, as that function is specifically focused on GZip-related operations.

conans/client/cmd/uploader.py
grossag: I tested decompression with various `bufsize` values and am finding that the sweet spot is right around 32KB. I'll run more tests to find the best value but my first test for a 7GB package with 16000 files showed this:
```
Decompress+extract for bufsize 10KB took 107.5195472240448 seconds
Decompress+extract for bufsize 16KB took 109.06598091125488 seconds
Decompress+extract for bufsize 32KB took 104.0440125465393 seconds
Decompress+extract for bufsize 64KB took 111.28909516334534 seconds
Decompress+extract for bufsize 128KB took 103.44610142707825 seconds
Decompress+extract for bufsize 256KB took 126.36465835571289 seconds
Decompress+extract for bufsize 512KB took 128.68103170394897 seconds
Decompress+extract for bufsize 1024KB took 141.1732087135315 seconds
Decompress+extract for bufsize 2048KB took 290.1585919857025 seconds
```
and second test showed this:
```
Decompress+extract for bufsize 10KB took 109.60621428489685 seconds
Decompress+extract for bufsize 16KB took 105.29482531547546 seconds
Decompress+extract for bufsize 32KB took 103.24112486839294 seconds
Decompress+extract for bufsize 64KB took 107.41227602958679 seconds
Decompress+extract for bufsize 128KB took 118.18165183067322 seconds
Decompress+extract for bufsize 256KB took 129.25403714179993 seconds
Decompress+extract for bufsize 512KB took 136.5764811038971 seconds
```

grossag:
@memsharded Are you able to review this PR?

conans/client/cmd/uploader.py
```diff
@@ -253,22 +272,50 @@ def upload_package(self, pref, prev_bundle, remote):
...
+                # The variance in compression times at bufsize<=64KB is small. It is only
+                # when bufsize>=128KB that compression times start increasing.
+                with tarfile.open(mode="w|", fileobj=stream_writer, bufsize=32768,
+                                  format=tarfile.PAX_FORMAT) as tar:
+                    unflushed_bytes = 0
+                    for filename, abs_path in sorted(files.items()):
+                        tar.add(abs_path, filename, recursive=False)
+
+                        unflushed_bytes += os.path.getsize(abs_path)
+                        if unflushed_bytes >= 2097152:
```
memsharded: Why this magic number here? Seems unexpected there is not some constant or other better way to check this

conans/client/cmd/uploader.py
```diff
@@ -253,22 +272,50 @@ def upload_package(self, pref, prev_bundle, remote):
...
-        tgz.close()
+
+    if compressformat == "zstd":
+        with open(tar_path, "wb") as tarfile_obj:
+            # Only provide level if it was overridden by config.
+            zstd_kwargs = {}
+            if compresslevel is not None:
+                zstd_kwargs["level"] = compresslevel
+
+            dctx = zstandard.ZstdCompressor(write_checksum=True, threads=-1, **zstd_kwargs)
```
memsharded: I'd probably make an independent function that implements the compression

conans/client/remote_manager.py
```diff
@@ -164,14 +166,21 @@ def _get_package(self, layout, pref, remote, scoped_output, metadata):
                                              metadata, only_metadata=False)
             zipped_files = {k: v for k, v in zipped_files.items() if not k.startswith(METADATA)}
             # quick server package integrity check:
-            for f in ("conaninfo.txt", "conanmanifest.txt", "conan_package.tgz"):
+            for f in ("conaninfo.txt", "conanmanifest.txt"):
                 if f not in zipped_files:
                     raise ConanException(f"Corrupted {pref} in '{remote.name}' remote: no {f}")
+            accepted_package_files = [PACKAGE_TZSTD_NAME, PACKAGE_TGZ_NAME]
+            package_file = next((f for f in zipped_files if f in accepted_package_files), None)
```
memsharded, line 173: Basically, a package could contain both compressed artifacts, but it will prioritize and only download the zstd one if existing?

Wouldn't it be a bit less confusing to not allow to have both compressed formats artifacts in the same package?

conans/client/cmd/uploader.py
```diff
@@ -197,15 +212,19 @@ def _compress_package_files(self, layout, pref):
...
 
-        if not os.path.isfile(package_tgz):
-            tgz_files = {f: path for f, path in files.items()}
-            compresslevel = self._global_conf.get("core.gzip:compresslevel", check_type=int)
-            tgz_path = compress_files(tgz_files, PACKAGE_TGZ_NAME, download_pkg_folder,
-                                      compresslevel=compresslevel, ref=pref)
-            assert tgz_path == package_tgz
-            assert os.path.exists(package_tgz)
+        if os.path.isfile(package_file):
+            output.info(f"Not writing '{package_file_name}' because it already exists.")
```
memsharded: A recommendation for "risky" PRs: try to keep the diff to a minimum. While this message might be useful, it makes the diff larger and more difficult to read.

It is probably fine to keep the ``tgz_files`` variable name, no need to generalize to ``source_files``, and that way the diff could also keep as short as possible.

conans/client/rest/rest_client_v2.py
```diff
@@ -81,8 +81,12 @@ def get_package(self, pref, dest_folder, metadata, only_metadata):
...
         # Download only known files, but not metadata (except sign)
         if not only_metadata:  # Retrieve package first, then metadata
-            accepted_files = ["conaninfo.txt", "conan_package.tgz", "conanmanifest.txt",
-                              "metadata/sign"]
+            accepted_package_files = [PACKAGE_TZSTD_NAME, PACKAGE_TGZ_NAME]
+            accepted_files = ["conaninfo.txt", "conanmanifest.txt", "metadata/sign"]
+            for f in accepted_package_files:
+                if f in server_files:
+                    accepted_files = [f] + accepted_files
+                    break
```
memsharded, line 89: If we assumed there can only be 1 compressed artifact in one of the formats, this would be simplified?

conans/model/conf.py
```diff
@@ -48,8 +48,11 @@
     "core.net.http:cacert_path": "Path containing a custom Cacert file",
     "core.net.http:client_cert": "Path or tuple of files containing a client cert (and key)",
     "core.net.http:clean_system_proxy": "If defined, the proxies system env-vars will be discarded",
-    # Gzip compression
+    # Compression for `conan upload`
+    "core.upload:compression_format": "The compression format used when uploading Conan packages. "
+                                      "Possible values: 'zstd', 'gzip' (default=gzip)",
     "core.gzip:compresslevel": "The Gzip compression level for Conan artifacts (default=9)",
+    "core.zstd:compresslevel": "The zstd compression level for Conan artifacts",
```
memsharded: Indicate the default if possible

conan/internal/paths.py
```diff
@@ -82,6 +82,7 @@ def _user_home_from_conanrc_file():
 CONAN_MANIFEST = "conanmanifest.txt"
 CONANINFO = "conaninfo.txt"
 PACKAGE_TGZ_NAME = "conan_package.tgz"
+PACKAGE_TZSTD_NAME = "conan_package.tar.zst"
```
memsharded: Maybe a shorter extension would be nice, like ``txz``? Does this exist?

conans/requirements.txt
```diff
@@ -7,3 +7,4 @@ fasteners>=0.15
 distro>=1.4.0, <=1.8.0; platform_system == 'Linux' or platform_system == 'FreeBSD'
 Jinja2>=3.0, <4.0.0
 python-dateutil>=2.8.0, <3
+zstandard>=0.20, <= 0.23
```
memsharded: This is one of the major issues, the ``ztandard`` library requires ``Python>=3.8``.
Conan is still supporting Python 3.6, even if it is EOL, but users request. It will be dropped probably soon, but still supported. So adding dependencies that requires Python 3.8 is not possible at the moment.

Furthermore, the requirements of this ``zstandard`` are probably a risk, see https://python-zstandard.readthedocs.io/en/latest/installing.html#requirements. This means that this library might have issues being installed in some non-mainstream platform, and there are Conan users running in very different and non-mainstream platforms (for example Windows non-x86 archs, see for example https://github.com/indygreg/python-zstandard/issues/237)

memsharded (commented):
Thanks for your contribution, sorry that we haven't been able to have time to review this.

This PR as is, is looking a bit risky, one of the main reasons the addition of the new ``zstandard`` library dependency. It is likely that it might be better added as conditional requirement (and protect the import of it with a ``try-except`` with a clear message). 

But I'd say that it is not impossible to move it forward, based on the diff, I think the code changes risk might be controlled. Please check the comments.

Thanks again for your contribution.

conan/internal/paths.py
grossag: Yeah actually, I did some searching and found that `.tzst` appears to be pretty standard. I hadn't known about that previously.

conans/client/cmd/uploader.py
grossag: This was suggested by @Ext3h in https://github.com/conan-io/conan/pull/14706#discussion_r1377565167 . Although, now that I am looking into it, I don't think flushing blocks matter.

The maximum zstd blocksize is `1<<17` (131072) and is in zstd as `ZSTD_BLOCKSIZE_MAX` and exposed in python-zstandard as `BLOCKSIZE_MAX`. By the time that we get to 2MB of data, zstd would already have had to break it into at least 16 blocks. I can remove this logic, unless @Ext3h has any additional thoughts on why this would help.

My only guess is that @Ext3h wanted me to create a new frame every 2MB or so and the reference to `flush(zstandard.COMPRESSOBJ_FLUSH_BLOCK)` was a typo and should have been `flush(zstandard.FLUSH_FRAME)` but I'm not sure. Splitting the object into multiple frames would be something that we would need to benchmark and investigate to see if there are any performance tradeoffs.

conans/client/remote_manager.py
grossag: A package is only supposed to contain one. Let's say an organization switches to zstd compression on Jan 1 2025. The expectation would be that packages produced before then would have `.tgz` extension and packages produced after then would have `.tzst` extension. I would like to avoid producing both because it would result in unnecessary storage usage in Artifactory.

conans/client/rest/rest_client_v2.py
grossag: Sorry, I think I'm missing what you are saying here. I don't have the context about if/how these accepted files changed over time. But Artifactory would only have .tgz or .tzst, not both. If that means we can simplify this a bit, that's fine with me.

conans/model/conf.py
grossag: The default is controlled by python-zstandard and is currently level 3. I'll put `(default=3)` but let me know if you instead want me to put `(default is set by python-zstandard)`

conans/requirements.txt
grossag: I have made this optional in the files module, so I have remove this line. The code will only raise an exception if someone enables zstd compression but doesn't have python-zstandard installed.

conans/client/cmd/uploader.py
Ext3h: You are correct, it should have been frames, not blocks. Frames are where the decoder can arbitrarily restart, while blocks within a frame have a data dependency in the uncompressed stream. (Therefor compression can be trivially threaded, but decompression can't.)

Just enforcing frames does not provide any parallel decompression capabilities with the pure zstd library out of the box either, unfortunately, so still only compression can be threaded. (I would have referenced `pzstd` as an example of how to multithread decompression as well - but it's on a deprecation path and did actually require skippable non-standard frames in between the compressed frames in order to actually achieve multi-threaded decompression from multiple offsets of the input file. Without those extra index frames, you still can't find the end of a frame without having to scan the file first.)

Where enforced framing is useful though, is in 3rd party tools (e.g. Artifactory!) that need to provide pseudo random access which doesn't work at all with the fully streamed default format. I know conan center doesn't care, but packages with debug symbols for some of the larger recipes tend to go into the multi GB range, and having to decode the full file to serve something from in the middle (e.g. because you have a symbol server that's pointing straight into the (indexed) archive contents in an Artifactory!) is simply madness.

2MB is way too little though. ZSTD is realistically designed to process 50MB+ in a single streaming pass, with the ***highest*** compression levels reaching up to an 128MB window size. Realistically, you would only want to enforce a frame break at around 64-128MB, as this ensures that the loss in compression ratio is still minimal, but "random access" is still reasonably (as in: up to a 50-100ms of additional CPU time, rather than 10s or more) costly.

memsharded:
Hi all.

I am resuming work on this, now that Python 3.14 has built-in zstd compression.

The current PR, derived from this one is in: https://github.com/conan-io/conan/pull/19337

We are planning to try to include it in Conan 2.25, in January 2026. It would be very useful to have testing and feedback, so if you want to experiment running from source from my branch in https://github.com/conan-io/conan/pull/19337, that would be very useful. Many thanks!

13steinj:
>I am resuming work on this, now that Python 3.14 has built-in zstd compression.

> The current PR, derived from this one is in: https://github.com/conan-io/conan/pull/19337

> We are planning to try to include it in Conan 2.25, in January 2026. It would be very useful to have testing and feedback,

At this time I have no good way to test this; however I would like to express the desire to make the implementation configurable-- that is, it's fine if built-in is used by default, but being able to offload to a command on the system with arguments will _probably_ lead to better performance, which is a bit of a necessity for things like CI times.

memsharded:
> At this time I have no good way to test this; however I would like to express the desire to make the implementation configurable-- that is, it's fine if built-in is used by default, but being able to offload to a command on the system with arguments will probably lead to better performance, which is a bit of a necessity for things like CI times.

Understood, but I think in this regard we need to go step by step. First ensuring that the different compressed formats work fine in the ecosystem, they don't break in unexpected scenarios, etc. This is the reason why we are putting on hold the "compression plugin" approach we also implemented in https://github.com/conan-io/conan/pull/18314, because we think that it can create a mess/chaos in the ecosystem, unless wide support for different compression formats, as built-in is there.

Once the xz and zstd formats are cleanly supported, as built-in, I understand the use case to want to optimize further the zip/unzip operations, and that might make more sense as a plugin without creating issues, as that would be way more optional and wouldn't break users that don't install the plugin.

memsharded:
https://github.com/conan-io/conan/pull/19337 has been merged, implementing both zstd (for Python>=3.14) and xz compressions built-in. It will be released as experimental in next Conan 2.25. Feedback welcome!

PR #14727, opened 2023-09-13
Add shorthand syntax in cli to specify host and build in 1 argument

### LINKED ISSUE #14696: [feature] Same host and build profile option?
### What is your suggestion?

In "developer mode sitting at my desk" mode, i.e. when building and running code on the same machine and manually typing `conan install ....` often, I'd find it useful to save some typing by adding a command line option to set both profiles the same easily.

So far I've not had a problem with only using `--profile`, but that merging of profiles makes me a bit nervous depending on what might be in `default` vs. what I really want (not specifying any profiles and just relying on `default` for both is not what I'm after, as I have multiple differing profiles).

Not sure if I'm missing something obvious or it's a really bad idea for some reason I'm not aware of, but in case it's a reasonable idea I hacked something up at: https://github.com/lukester1975/conan/commit/d554a6ad84015007e0005d8aa027d7877b64bef2 

Thanks!

AbrilRBS:
Changelog: Feature: Add shorthand syntax in cli to specify host and build in 1 argument
Docs: https://github.com/conan-io/docs/pull/3439

There's some split preferences between possible names. This pr implements (for `-pr`, `-s`, `-o`, `-c`) `-s:a` (`--settings:all`), but other possible names we discussed are:

* `-s:hb` (`settings:hostbuild`): Makes it more clear that this affects the host and build profiles
* `-s:h,b` (`settings:host,build`): Separates both names with a comma (Which is not a special char in any terminal as far as I can see), maybe easier to read?

I'm open to change it to whatever everyone feels like a better one

The new profile args read like this:

```
$ conan build -h
usage: conan build [-h] [-v [V]] [--name NAME] [--version VERSION] [--user USER]
                   [--channel CHANNEL] [-g GENERATOR] [-of OUTPUT_FOLDER] [-d DEPLOYER]
                   [--deployer-folder DEPLOYER_FOLDER] [--build-require] [-b BUILD]
                   [-r REMOTE | -nr] [-u] [-pr PROFILE] [-pr:b PROFILE_BUILD]
                   [-pr:h PROFILE_HOST] [-pr:a PROFILE_ALL] [-o OPTIONS] [-o:b OPTIONS_BUILD]
                   [-o:h OPTIONS_HOST] [-o:a OPTIONS_ALL] [-s SETTINGS] [-s:b SETTINGS_BUILD]
                   [-s:h SETTINGS_HOST] [-s:a SETTINGS_ALL] [-c CONF] [-c:b CONF_BUILD]
                   [-c:h CONF_HOST] [-c:a CONF_ALL] [-l LOCKFILE] [--lockfile-partial]
                   [--lockfile-out LOCKFILE_OUT] [--lockfile-packages] [--lockfile-clean]
                   [--lockfile-overrides LOCKFILE_OVERRIDES]
                   [path]

Install dependencies and call the build() method.

positional arguments:
  path                  Path to a python-based recipe file or a folder containing a
                        conanfile.py recipe. conanfile.txt cannot be used with conan build.

options:
  -v [V]                Level of detail of the output. Valid options from less verbose to more
                        verbose: -vquiet, -verror, -vwarning, -vnotice, -vstatus, -v or
                        -vverbose, -vv or -vdebug, -vvv or -vtrace
  --name NAME           Provide a package name if not specified in conanfile
  --version VERSION     Provide a package version if not specified in conanfile
  --user USER           Provide a user if not specified in conanfile
  --channel CHANNEL     Provide a channel if not specified in conanfile
  -g GENERATOR, --generator GENERATOR
                        Generators to use
  -of OUTPUT_FOLDER, --output-folder OUTPUT_FOLDER
                        The root output folder for generated and build files
  -d DEPLOYER, --deployer DEPLOYER
                        Deploy using the provided deployer to the output folder
  --deployer-folder DEPLOYER_FOLDER
                        Deployer output folder, base build folder by default if not set
  --build-require       Whether the provided path is a build-require
  -b BUILD, --build BUILD
                        Optional, specify which packages to build from source. Combining
                        multiple '--build' options on one command line is allowed. Possible
                        values: --build="*" Force build from source for all packages.
                        --build=never Disallow build for all packages, use binary packages or
                        fail if a binary package is not found, it cannot be combined with other
                        '--build' options. --build=missing Build packages from source whose
                        binary package is not found. --build=cascade Build packages from source
                        that have at least one dependency being built from source.
                        --build=[pattern] Build packages from source whose package reference
                        matches the pattern. The pattern uses 'fnmatch' style wildcards.
                        --build=~[pattern] Excluded packages, which will not be built from the
                        source, whose package reference matches the pattern. The pattern uses
                        'fnmatch' style wildcards. --build=missing:[pattern] Build from source
                        if a compatible binary does not exist, only for packages matching
                        pattern.
  -r REMOTE, --remote REMOTE
                        Look in the specified remote or remotes server
  -nr, --no-remote      Do not use remote, resolve exclusively in the cache
  -u, --update          Will check the remote and in case a newer version and/or revision of
                        the dependencies exists there, it will install those in the local
                        cache. When using version ranges, it will install the latest version
                        that satisfies the range. Also, if using revisions, it will update to
                        the latest revision for the resolved version range.
  -pr PROFILE, --profile PROFILE
                        Apply the specified profile. By default, or if specifying -pr:h
                        (--profile:host), it applies to the host context. Use -pr:b
                        (--profile:build) to specify the build context, or -pr:a
                        (--profile:all) to specify both contexts at once
  -pr:b PROFILE_BUILD, --pr:build PROFILE_BUILD
  -pr:h PROFILE_HOST, --pr:host PROFILE_HOST
  -pr:a PROFILE_ALL, --profile:all PROFILE_ALL
  -o OPTIONS, --options OPTIONS
                        Apply the specified profile. By default, or if specifying -o:h
                        (--options:host), it applies to the host context. Use -o:b
                        (--options:build) to specify the build context, or -o:a (--options:all)
                        to specify both contexts at once. Example: -o pkg:with_qt=true
  -o:b OPTIONS_BUILD, --o:build OPTIONS_BUILD
  -o:h OPTIONS_HOST, --o:host OPTIONS_HOST
  -o:a OPTIONS_ALL, --options:all OPTIONS_ALL
  -s SETTINGS, --settings SETTINGS
                        Apply the specified profile. By default, or if specifying -s:h
                        (--settings:host), it applies to the host context. Use -s:b
                        (--settings:build) to specify the build context, or -s:a
                        (--settings:all) to specify both contexts at once. Example: -s
                        compiler=gcc
  -s:b SETTINGS_BUILD, --s:build SETTINGS_BUILD
  -s:h SETTINGS_HOST, --s:host SETTINGS_HOST
  -s:a SETTINGS_ALL, --settings:all SETTINGS_ALL
  -c CONF, --conf CONF  Apply the specified profile. By default, or if specifying -c:h
                        (--conf:host), it applies to the host context. Use -c:b (--conf:build)
                        to specify the build context, or -c:a (--conf:all) to specify both
                        contexts at once. Example: -c
                        tools.cmake.cmaketoolchain:generator=Xcode
  -c:b CONF_BUILD, --c:build CONF_BUILD
  -c:h CONF_HOST, --c:host CONF_HOST
  -c:a CONF_ALL, --conf:all CONF_ALL
  -l LOCKFILE, --lockfile LOCKFILE
                        Path to a lockfile. Use --lockfile="" to avoid automatic use of
                        existing 'conan.lock' file
  --lockfile-partial    Do not raise an error if some dependency is not found in lockfile
  --lockfile-out LOCKFILE_OUT
                        Filename of the updated lockfile
  --lockfile-packages   Lock package-id and package-revision information
  --lockfile-clean      Remove unused entries from the lockfile
  --lockfile-overrides LOCKFILE_OVERRIDES
                        Overwrite lockfile overrides
```

Closes #14696


AbrilRBS:
Can't find a way to still show the `-pr:b` etc variations in the usage but not in the help

conan/api/subapi/profiles.py
```diff
@@ -29,12 +29,26 @@ def get_default_build(self):
         return loader.get_default_build()
 
     def get_profiles_from_args(self, args):
-        build = [self.get_default_build()] if not args.profile_build else args.profile_build
-        host = [self.get_default_host()] if not args.profile_host else args.profile_host
-        profile_build = self.get_profile(profiles=build, settings=args.settings_build,
-                                         options=args.options_build, conf=args.conf_build)
-        profile_host = self.get_profile(profiles=host, settings=args.settings_host,
-                                        options=args.options_host, conf=args.conf_host)
+        all_profiles = [] if not args.profile_all else args.profile_all
```
memsharded:
```suggestion
        all_profiles = args.profile_all or []
```
If the comparison is ``not profile_all``, short circuit ``or`` might be easier to read

AbrilRBS: nice python-fu, you're right :)

conan/api/subapi/profiles.py
```diff
@@ -29,12 +29,26 @@ def get_default_build(self):
...
-        profile_build = self.get_profile(profiles=build, settings=args.settings_build,
-                                         options=args.options_build, conf=args.conf_build)
-        profile_host = self.get_profile(profiles=host, settings=args.settings_host,
-                                        options=args.options_host, conf=args.conf_host)
+        all_profiles = args.profile_all or []
+        all_settings = args.settings_all or []
+        all_options = args.options_all or []
+        all_conf = args.conf_all or []
+
+        build_profiles = all_profiles + (args.profile_build or [self.get_default_build()])
```
memsharded: Still unsure about this. I am thinking that if ``-pr:a=xx`` is defined, that is equal to ``-pr:h=xx -pr:b=xx``, but not to ``-pr:h=xx -pr:h=default -pr:b=xx -pr:b=default``

conans/test/integration/build_requires/build_requires_test.py
```diff
@@ -859,6 +859,33 @@ def build_requirements(self):
         c.run("install app --lockfile=app/conan.lock")
         c.assert_listed_require({"protobuf/1.1": "Cache"}, build=True)
 
+    def test_overriden_host_version_transitive_deps(self):
```
memsharded: This test seems unrelated? Need rebase?

conan/cli/args.py
```diff
@@ -56,45 +58,76 @@ def add_common_install_arguments(parser):
...
-                            action="append",
-                            dest="options_{}".format(machine),
-                            help='Define options values ({} machine), e.g.:'
-                                 ' -o{} Pkg:with_qt=true'.format(machine, short_suffix))
-
-    def conf_args(machine, short_suffix="", long_suffix=""):
-        parser.add_argument("-c{}".format(short_suffix),
-                            "--conf{}".format(long_suffix),
+    class ContextAllAction(argparse.Action):
+        def __init__(self,
```
memsharded: Probably args/kwargs can be used to simplify code, both here and in ``super()`` code

conan/cli/args.py
```diff
@@ -56,45 +58,76 @@ def add_common_install_arguments(parser):
...
-        item_fn("build", ":b", ":build")
-        item_fn("host", ":h", ":host")
+                            dest=f"{long}_host",
+                            metavar=long.upper(),
+                            help='Apply the specified profile. '
+                                 f'By default, or if specifying -{short}:h (--{long}:host), it applies to the host context. '
+                                 f'Use -{short}:b (--{long}:build) to specify the build context, '
+                                 f'or -{short}:a (--{long}:all) to specify both contexts at once'
+                                  + ('' if not example else f". Example: {example}"))
+        contexts = ["build", "host"]
```
memsharded: If we hardcode this as a const, instead of an arg, probably the Action can be further simplified?

conan/cli/args.py
AbrilRBS: I wish the append Action was public in argparse, so this could be 
```py
    class ContextAllAction(argparse._AppendAction):
        def __call__(self, action_parser, namespace, values, option_string=None):
            for context in contexts:
                items = getattr(namespace, self.dest + "_" + context, None)
                items = items[:] if items else []
                items.append(values)
                setattr(namespace, self.dest + "_" + context, items)
```
but alas

memsharded (approved):

franramirez688 (approved):

czoido (approved):

PR #14731, opened 2023-09-13
better CMakeToolchain blocks filter

### LINKED ISSUE #14640: [feature] CMakeToolchain: improve `tc.blocks` interface
### What is your suggestion?

The contents of `conan_toolchain.cmake` consist of a set of logical blocks that are controlled via `https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#customizing-the-content-blocks`

currently, it is possible to remove blocks, or add/replace with new templates. But if we want to "choose" which blocks to keep, it isn't as easy, as removing is only possible if the names are known beforehand.

use case:
* user would like to keep the find paths (`CMAKE_PREFIX_PATH` and so on), but nothing else, as the user already has pre-existing CMake toolchain files that they must use

The only way of doing it _today_ is by knowing the names of all existing blocks:
```
    def generate(self):
        tc = CMakeToolchain(self)
        self.output.warning(f"Blocks: {tc.blocks}")
        for block in ["user_toolchain", "generic_system", "compilers", "android_system",
                    "apple_system", "fpic", "arch_flags", "linker_scripts", "libcxx",
                    "vs_runtime", "cppstd", "parallel", "extra_flags",
                    "cmake_flags_init", "try_compile", "pkg_config", "rpath", "shared", "output_dirs"]:
            tc.blocks.remove(block)
        tc.generate()
```

It would be good if this could be simplified, to something like:
```
for block in tc.block.names():
    if name != "find_paths":
        tc.block.remove(block)
```

which would cause the `conan_toolchain.cmake` file to be generated with only the selected contents.

We should also be aware that this would make the names of the blocks part of the API - renaming blocks could be a breaking change, so may want to settle on an interface that is robust to changes in a way that future Conan updates don't break recipes

### LINKED ISSUE #14641: [feature] conan_toolchain.cmake: consider segregating blocks in separate files
### What is your suggestion?

The contents of `conan_toolchain.cmake` consist of several logic "blocks". For some users in tightly controlled environments, they may only be interested in some subsets of the file, wanting to avoid the rest.

If the logic blocks are in separate files, users could selectively `include()` them for CMakeLists.txt. This can be useful in situations where the users have pre-existing CMake toolchain files that must be used, or where the variables/behaviours set by `conan_toolchain.cmake` conflict with the project in ways where the user wants greater control.

Consider segregating the different logic blocks into separate files - `conan_toolchain.cmake` can still `include()` them all - and users can optionally include only what they need, where they have enough control to do the right thing ™️ . 

From team discussion:
- this behaviour could be enabled under an opt-in conf, retaining the current behaviour otherwise.

memsharded:
Changelog: Feature: Better ``CMakeToolchain`` blocks interface. Added new ``.blocks.select()``, ``.blocks.keys()``.
Docs: https://github.com/conan-io/docs/pull/3384

Close https://github.com/conan-io/conan/issues/14640
Close https://github.com/conan-io/conan/issues/14641

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -840,8 +840,20 @@ class ToolchainBlocks:
             for name, block in items:
                 self._blocks[name] = block(conanfile, toolchain)
 
-    def remove(self, name):
+    def keys(self):
+        return self._blocks.keys()
+
+    def items(self):
+        return self._blocks.items()
+
+    def remove(self, name, *args):
         del self._blocks[name]
+        for arg in args:
+            del self._blocks[arg]
+
+    def filter(self, name, *args):
+        to_keep = [name] + list(args)
+        self._blocks = OrderedDict((k, v) for k, v in self._blocks.items() if k in to_keep)
 
     def __setitem__(self, name, block_type):
         # Create a new class inheriting Block with the elements of the provided one
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -851,7 +851,10 @@ class ToolchainBlocks:
         for arg in args:
             del self._blocks[arg]
 
-    def filter(self, name, *args):
+    def select(self, name, *args):
+        """
+        keep the blocks provided as arguments, remove the others
+        """
         to_keep = [name] + list(args)
         self._blocks = OrderedDict((k, v) for k, v in self._blocks.items() if k in to_keep)
 
--- a/conans/test/unittests/tools/cmake/test_cmaketoolchain.py
+++ b/conans/test/unittests/tools/cmake/test_cmaketoolchain.py
@@ -63,7 +63,7 @@ def test_remove(conanfile):
 
 def test_filter(conanfile):
     toolchain = CMakeToolchain(conanfile)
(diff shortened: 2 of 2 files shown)


AbrilRBS (approved):

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -840,8 +840,20 @@ def __init__(self, conanfile, toolchain, items=None):
...
+        return self._blocks.items()
+
+    def remove(self, name, *args):
         del self._blocks[name]
+        for arg in args:
+            del self._blocks[arg]
+
+    def filter(self, name, *args):
+        to_keep = [name] + list(args)
+        self._blocks = OrderedDict((k, v) for k, v in self._blocks.items() if k in to_keep)
```
franramirez688, line 859: IMO, the name `filter` sounds like this function is going to return something. As far as I see, this filter is going to perform a kind of `include` or `keep` those names that you're passing, isn't it?

```python
def include(self, name, *args):  # Or directly def include(self, *args):
# Or
def keep(self, name, *args):  # Or directly def keep(self, *args):
```

memsharded: Let's do ``select()``

franramirez688 (approved):

PR #14734, opened 2023-09-14
Sketch parallel compatibility checks

AbrilRBS:
Had this idea while eating breakfast, wanted to test if it would work.

For a local installation of flatbuffers on my end seems to work nicely (unexpected, I know)

As a PR to see what everyone thinks about it :)

For #14799

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -1,3 +1,5 @@
+from multiprocessing.pool import ThreadPool
+
 from conan.api.output import ConanOutput
 from conans.client.graph.build_mode import BuildMode
 from conans.client.graph.compatibility import BinaryCompatibility
@@ -97,7 +99,6 @@ class GraphBinariesAnalyzer(object):
 
     def _process_compatible_packages(self, node, remotes, update):
         conanfile = node.conanfile
-        original_binary = node.binary
         original_package_id = node.package_id
 
         compatibles = self._compatibility.compatibles(conanfile)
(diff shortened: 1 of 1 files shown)


conans/client/graph/graph_binaries.py
```diff
@@ -106,28 +107,38 @@ def _process_compatible_packages(self, node, remotes, update):
...
-                # Trick to allow mutating the options (they were freeze=True)
-                # TODO: Improve this interface
-                conanfile.options = conanfile.options.copy_conaninfo_options()
-                conanfile.options.update_options(compatible_package.options)
-                break
-        else:  # If no compatible is found, restore original state
-            node.binary = original_binary
-            node._package_id = original_package_id
+
+            thread_pool = ThreadPool(4)
```
AbrilRBS, line 119: 4 or whatever the conf says, again, this is just a sketch

conans/client/graph/graph_binaries.py
```diff
@@ -106,28 +107,38 @@ def _process_compatible_packages(self, node, remotes, update):
...
+            import copy
+            result = thread_pool.starmap(check_compatible, [(copy.copy(node), pkg_id, compatible_package)
```
AbrilRBS, line 121: The copy is needed because the `_process_compatible_node` modifies internal node attributes. We could maybe create a new one copying only the fields that will be used, but this seemed more robus for future changes

AbrilRBS, line 121: Maybe `starmap_async` too?

AbrilRBS (commented):
Some comments

AbrilRBS:
A better approach would be to fetch all the package info needed in 1 single API call to the server, that's where the proper solution is at - Let's fall back to this only if that path is a no-go after pursuing it first!

PR #14740, opened 2023-09-14
Add missing minor macos versions

### LINKED ISSUE #14656: [bug] CMAKE_OSX_DEPLOYMENT_TARGET for macOS does not support point releases
### Environment details

* Operating System+version: macOS
* Compiler+version: Xcode 15 (SDK 14.0) running on macOS 13.
* Conan version: 2.0.10
* Python version:

### Steps to reproduce

Discovered this using `cmake-conan`'s logic when passing `os.version`.

* `os.version` is only defined in `cmake-conan` if `CMAKE_OSX_DEPLOMENT_TARGET`
* `CMAKE_OSX_DEPLOYMENT_TARGET` is usually "empty" when targetting macOS, unless:
   - it is set manually
   - the SDK version detected is _newer_ than then OS we are currently running, in that case, it is set such that the binaries are compatibles with the OS we are usin

The SDKs (and thus. `os.version`) support point releases as valid targets, but for macOS, `settings.yml` only models the `.0` release for each major.  For the other Apple OSs (iOS, tvOS, watchOS), we do properly track the point releases.

Suggestion: accept point releases as valid `os.version` setting for macOS 11+ 

### Logs

```
ERROR: Invalid setting '13.5' is not a valid 'settings.os.version' value.
Possible values are ['10.6', '10.7', '10.8', '10.9', '10.10', '10.11', '10.12', '10.13', '10.14', '10.15', '11.0', '12.0', '13.0', '14.0']
```

AbrilRBS:
Changelog: Fix: Add missing minor macos versions
Docs: https://github.com/conan-io/docs/pull/3389

Closes: https://github.com/conan-io/conan/issues/14656

The list comes from https://support.apple.com/es-es/HT201260 (Thanks to @czoido for finding it, my googling was giving many results)


czoido (approved):

memsharded (approved):

barracuda156:
@RubenRBS Does something prevent this from working on 10.5?

AbrilRBS:
Hi @barracuda156  - I'm not aware of any _explicit_ reason for it not to work.

If you want to try, you can add support for it by adding the version into your own `settings_user.yml` file ([Example here](https://docs.conan.io/2/examples/config_files/settings/settings_user.html#examples-config-files-settings-user)) - but in such an old version, things might just not work and we can't commit to support them, but happy to hear from you if you test it out :)

Issue #14334, opened 2023-07-20
[bug] Conan 1.x: `CMAKE_PREFIX_PATH` cannot be set to `self.package_folder`


thorsten-klein:
### Environment details

* Operating System+version: Ubuntu 20
* Compiler+version: irrelevant
* Conan version: 1.60.1
* Python version: 3.8.10

### Steps to reproduce

I am using `CMakeToolchain` generator.

`conanfile.py` (mylib)
```
def package_info(self):
        self.cpp_info.builddirs.append(self.package_folder)
```

`CMakeLists.txt` (test_package)
```
message(STATUS "CMAKE_PREFIX_PATH '${CMAKE_PREFIX_PATH}'"
```

I can see that the explicitly specified `self.cpp_info.builddirs` is not added to `CMAKE_PREFIX_PATH`.
According to the comments in `blocks.py` it `# The explicitly defined "builddirs" of "host" context dependencies must be in PREFIX_PATH`

thorsten-klein:
My hello project:

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.15)
project(hello)
add_library(hello INTERFACE)
install(TARGETS hello EXPORT helloConfig)
install(EXPORT helloConfig
    DESTINATION "${CMAKE_INSTALL_PREFIX}/share/cmake/hello"
    NAMESPACE hello::
)
```

conanfile.py
```
from conans import ConanFile
from conan.tools.cmake import CMake
import os
class TestConan(ConanFile):
    name = "hello"
    version = "1.0.0"
    settings = "os", "compiler", "build_type", "arch"
    exports_sources = "*"
    generators = "CMakeToolchain"

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.install()
```

My consumer:

CMakeLists.txt
```
cmake_minimum_required(VERSION 3.15)
project(consumer)
find_package(hello REQUIRED)
```

conanfile.py
```
from conans import ConanFile
from conan.tools.cmake import CMake
import os
class TestConan(ConanFile):
    name = "consumer"
    version = "1.0.0"
    settings = "os", "compiler", "build_type", "arch"
    exports_sources = "*"
    generators = "CMakeToolchain"
    requires = "hello/1.0.0@"

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.install()
```

memsharded:
Hi @thorsten-klein 

Thanks very much for your code, it helps.
[...]

PR #14746, opened 2023-09-15
CMakeToolchain generator sets `CMAKE_FRAMEWORK_PATH`

thorsten-klein:
Changelog: (Feature): CMakeToolchain generator sets CMAKE_FRAMEWORK_PATH to host package roots

Let cmake `find_package()` do its default search procedure instead of maintaining a list of directories as `cpp_info.builddirs`.

With this change users need to maintain only cmake module paths in `cpp_info.builddirs` (containing `FindXXX.cmake`).
cmake config paths (containing `XXX-Config.cmake`) do not need to be maintained anymore.
conan will add each host package root path automatically to `CMAKE_FRAMEWORK_PATH`, which allows cmake to find them automatically when doing its default search procedure 
(Ref: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure)

Statement from Kitware:
```
The CMAKE_FRAMEWORK_PATH variable will be used by find_* commands, on all platforms. 
The main difference is that by default, on non-Mac platforms, find_* will first look for 
CMAKE_PREFIX_PATH variable entries and then  CMAKE_FRAMEWORK_PATH. 
Order is inverted on Mac by default. 
```

Ref: #14334 

- [x ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one.

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -461,7 +461,11 @@ class FindFiles(Block):
         list(PREPEND CMAKE_MODULE_PATH {{ generators_folder }})
         {% endif %}
 
-        # Definition of CMAKE_PREFIX_PATH, CMAKE_XXXXX_PATH
+        # Definition of CMAKE_PREFIX_PATH, CMAKE_FRAMEWORK_PATH, CMAKE_XXXXX_PATH
+        {% if host_build_paths_root %}
+        # The root dirs of all "host" context dependencies should be considered during find_package()
+        list(PREPEND CMAKE_FRAMEWORK_PATH {{ host_build_paths_root }})
+        {% endif %}
         {% if host_build_paths_noroot %}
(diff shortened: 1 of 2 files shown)


thorsten-klein:
Ref: #14748

thorsten-klein:
We are now setting environment variable `CMAKE_FRAMEWORK_PATH` which is also considered by cmake automatically. This solves our issue well. But nevertheless it would be benefitial for other users if the `CMakeToolchain` generator sets this cmake variable and works out of the box.

The big advantage what I see is:
if the `CMAKE_FRAMEWORK_PATH` is used, conan does not need to store any information about the `builddirs` anymore. cmake will automatically search within some default directories when doing `find_package()`, `find_library()` or `find_path()`. These are exactly the things within host context.
`find_program()` (which should only search in build context) is not affected by this variable

Ref: #14748 

I guess that with this change also some cmake variables (which are currently set in `conan_toolchain.cmake`) could become obsolete. But this can be checked in a subsequent PR then.

PR #14752, opened 2023-09-15
Feature/Add an option to install() method in package_manager

### LINKED ISSUE #14567: [bug] Install packages compatible with incorrect architecture
### Environment details

* Operating System+version: all
* Compiler+version: all
* Conan version: 2.0.9 (latest)
* Python version: all

### Steps to reproduce

1. Make sure settings.arch in your profile differs your host architecture, for example:
 If your host architecture is x86 and in your profile is defined:
```
[settings]
arch=armv7
....
[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True
```

2. In the conan file you should use `_SystemPackageManagerTool` in some way, for example:
```
        apt = Apt(self)
        apt.install(["tree"], update=True, check=True)
```
For installing requirements for building the conan package.

3. Then execute `conan build -pr <your_profile> ....`

4. If on your system the command `tree` is not found (in linux you can uninstall it with the command `apt-get remove tree` to try this), conan will try to install the package for arm and not for your host architecture (x86), means it will try to install `tree:arm`  instead of just `tree`.

### Logs

conan executes something like that `apt-get install tree:arm` on x86 architecture and we get an error that the architecture is suitable for our device.

- [x]  I've read the CONTRIBUTING guide

LielHarel:
Changelog: Feature: Add `host_tool` to `install()` method in `package_manager` to indicate whether the package is a host tool or a library.
Docs: https://github.com/conan-io/docs/pull/3401

According to that the we will install a package for the needed arch (host arch or destination arch which can be different in case of cross building)

This option enables the user controls whether the installed package is for the host or needed for the destination arch, this is useful specifically when cross building and we want to avoid misunderstanding about what we want to install. 

Closes https://github.com/conan-io/conan/issues/14567

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -62,9 +62,12 @@ class _SystemPackageManagerTool(object):
                 if d in os_name:
                     return tool
 
-    def get_package_name(self, package):
-        # TODO: should we only add the arch if cross-building?
-        if self._arch in self._arch_names and cross_building(self._conanfile):
+    def get_package_name(self, package, host_package=True):        
+        # Only if the package is for building, for example a library,
+        # we should add the host arch when cross building.
+        # If the package is a tool that should be installed on the current build 
+        # machine we should not add the arch.
+        if self._arch in self._arch_names and cross_building(self._conanfile) and host_package:
             return "{}{}{}".format(package, self._arch_separator,
                                    self._arch_names.get(self._arch))
         return package
@@ -149,7 +152,7 @@ class _SystemPackageManagerTool(object):
             self._conanfile.output.warning(str(error))
         raise ConanException("None of the installs for the package substitutes succeeded.")
 
-    def _install(self, packages, update=False, check=True, **kwargs):
+    def _install(self, packages, update=False, check=True, host_package=True, **kwargs):
         pkgs = self._conanfile.system_requires.setdefault(self._active_tool, {})
         install_pkgs = pkgs.setdefault("install", [])
         install_pkgs.extend(p for p in packages if p not in install_pkgs)
@@ -157,7 +160,7 @@ class _SystemPackageManagerTool(object):
             return
 
         if check or self._mode in (self.mode_check, self.mode_report_installed):
-            packages = self.check(packages)
+            packages = self.check(packages, host_package=host_package)
             missing_pkgs = pkgs.setdefault("missing", [])
             missing_pkgs.extend(p for p in packages if p not in missing_pkgs)
 
@@ -178,7 +181,7 @@ class _SystemPackageManagerTool(object):
             if update:
                 self.update()
 
-            packages_arch = [self.get_package_name(package) for package in packages]
+            packages_arch = [self.get_package_name(package, host_package=host_package) for package in packages]
             if packages_arch:
                 command = self.install_command.format(sudo=self.sudo_str,
                                                       tool=self.tool_name,
@@ -196,8 +199,8 @@ class _SystemPackageManagerTool(object):
             command = self.update_command.format(sudo=self.sudo_str, tool=self.tool_name)
             return self._conanfile_run(command, self.accepted_update_codes)
 
-    def _check(self, packages):
-        missing = [pkg for pkg in packages if self.check_package(self.get_package_name(pkg)) != 0]
+    def _check(self, packages, host_package=True):
+        missing = [pkg for pkg in packages if self.check_package(self.get_package_name(pkg, host_package=host_package)) != 0]
         return missing
 
     def check_package(self, package):
@@ -234,7 +237,7 @@ class Apt(_SystemPackageManagerTool):
 
         self._arch_separator = ":"
 
-    def install(self, packages, update=False, check=True, recommends=False):
+    def install(self, packages, update=False, check=True, host_package=True, recommends=False):
         """
         Will try to install the list of packages passed as a parameter. Its
         behaviour is affected by the value of ``tools.system.package_manager:mode``
@@ -243,13 +246,15 @@ class Apt(_SystemPackageManagerTool):
         :param packages: try to install the list of packages passed as a parameter.
         :param update: try to update the package manager database before checking and installing.
         :param check: check if the packages are already installed before installing them.
+        :param host_package: install the packages for the host machine architecture (the machine
+               that will run the software), it has an effect when cross building.
         :param recommends: if the parameter ``recommends`` is ``False`` it will add the
                ``'--no-install-recommends'`` argument to the *apt-get* command call.
         :return: the return code of the executed apt command.
         """
         recommends_str = '' if recommends else '--no-install-recommends '
         return super(Apt, self).install(packages, update=update, check=check,
-                                        recommends=recommends_str)
+                                        host_package=host_package, recommends=recommends_str)
 
 
 class Yum(_SystemPackageManagerTool):
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -237,7 +237,7 @@ class Apt(_SystemPackageManagerTool):
 
         self._arch_separator = ":"
 
-    def install(self, packages, update=False, check=True, host_package=True, recommends=False):
+    def install(self, packages, update=False, check=True, recommends=False, host_package=True):
         """
         Will try to install the list of packages passed as a parameter. Its
         behaviour is affected by the value of ``tools.system.package_manager:mode``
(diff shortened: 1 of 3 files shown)


conan/tools/system/package_manager.py
```diff
@@ -56,9 +55,11 @@ def get_default_tool(self):
...
                     return tool
 
-    def get_package_name(self, package):
-        # TODO: should we only add the arch if cross-building?
-        if self._arch in self._arch_names and cross_building(self._conanfile):
+    def get_package_name(self, package, host_tool=True):        
+        # Only if the package is for building, for example a library,
+        # we should add the destination arch (important especially in cross building).
+        # If the package is a tool that should be installed on the host we should not add the arch.
+        if self._arch in self._arch_names and not host_tool:
```
memsharded: Maybe keep the ``cross_building()`` condition? Even for host packages, if we are not cross-compiling, maybe it is better not to add architecture specifiers, and go with the native one?

memsharded (commented):
Thanks for contributing to Conan!

memsharded (commented):
Some kind of testing to cover this would be necessary. Likely to be unit-test with some patching/mocking, not easy for a full test with this. Please have a look, don't hesitate to ask for help, we might even be able to contribute the tests ourselves.

conan/tools/system/package_manager.py
LielHarel: I though that if the package should be linked to the built package maybe it a good practice to add the arch anyway since it is not used as a tool for the host. But is seems that it will not make any dramatic different so I can keep this condition :) What do you think?

LielHarel:
> Some kind of testing to cover this would be necessary. Likely to be unit-test with some patching/mocking, not easy for a full test with this. Please have a look, don't hesitate to ask for help, we might even be able to contribute the tests ourselves.

I have tried to run the current test with the command `python -m pytest .` and get this error:
```
ImportError while loading conftest '/home/liel/conan/conan/conans/test/conftest.py'.
../__init__.py:1: in <module>
    from conans.client.build.autotools_environment import AutoToolsBuildEnvironment
E     File "/home/liel/conan/conan/conans/client/build/autotools_environment.py", line 37
E       conanfile.output.warning(f"**** The 'AutoToolsBuildEnvironment' helper is deprecated. "
E                                                                                             ^
E   SyntaxError: invalid syntax
``` 
Do you know that error?

conan/tools/system/package_manager.py
memsharded: We usually tend to favor commands as close to the developer experience as possible. So if a user uses ``apt install xxxx`` without architecture as default (and they do), it is slightly better that the conan command does the same.

conan/tools/system/package_manager.py
```diff
@@ -56,9 +56,11 @@ def get_default_tool(self):
...
                     return tool
 
-    def get_package_name(self, package):
-        # TODO: should we only add the arch if cross-building?
-        if self._arch in self._arch_names and cross_building(self._conanfile):
+    def get_package_name(self, package, host_tool=True):        
+        # Only if the package is for building, for example a library,
+        # we should add the destination arch when cross building.
+        # If the package is a tool that should be installed on the host we should not add the arch.
+        if self._arch in self._arch_names and cross_building(self._conanfile) and not host_tool:
```
memsharded: I think that the naming might be the opposite. We are using GNU, which means:
- "build" is the current build machine
- "host" is the machine that will run the software

If we are cross-building in a X86_64 box for an armv8 target, our "build" is x86_64 and the "host" is "armv8".
Then, I understand that the "mypkg:arm" qualifier needs to be added when it is a "host_package" (I'd probably avoid the ``_tool`` name here, as there are both libraries and tools in system packages), so maybe the semantics are reversed?

conan/tools/system/package_manager.py
LielHarel: Sounds logical :+1: I keep the condition

conan/tools/system/package_manager.py
LielHarel: Ops, I refer to the host machine as the current build machine, I will change it.

conan/tools/system/package_manager.py
```diff
@@ -56,11 +56,12 @@ def get_default_tool(self):
                 if d in os_name:
                     return tool
 
-    def get_package_name(self, package, host_tool=True):        
+    def get_package_name(self, package, host_package=False):
```
memsharded: I think this would be breaking wrt to existing usage:
- The current default is assuming that all packages are "host", because the architecture will be applied if cross-building, always
- Users using this might be broken by this change
- The fact that the tests have to be changed to pass, is also an indicator of this

I think we need to use the opposite default.

memsharded (commented):
I think we are almost there, thanks for your work! 🙂 

I am concerned that this change might be breaking existing usage, please have a look to the comments.
When the existing tests doesn't need to be changed, that is a very good sign, then adding a new test, that checks the other non-default new behavior. Please let me know if you need with that new test.

conan/tools/system/package_manager.py
LielHarel: I have changed also in the docs repo

conan/tools/system/package_manager.py
LielHarel: But the opposite default will cause the adding of the arch atomically and it is not the native way the users use the install command. Are we fine with the fact that the default value can add the arch and do something that can feel like a magic ? This what happens to me when I used this feature at the first place and it makes me very confused :)

LielHarel: I feel that the default behavior should be as simple as possible.
About the tests - I just think I should add tests for both cases

memsharded: > But the opposite default will cause the adding of the arch atomically and it is not the native way the users use the install command. Are we fine with the fact that the default value can add the arch and do something that can feel like a magic ?

Yes, the default behavior is intended for "host" libraries, not "build" tools. If you check all the ``system_requirements()`` in ConanCenter, you will see those are libraries intended for the "host" context. With your change, all these recipes in ConanCenter will be broken, and would require to add the ``host_package=True``. In the same way there can be tons of users with private recipes that rely on this default.

The capability that this PR is adding is the one of specifying: "I want to install a "build" tool requirement, please do not try to install it in the host architecture", which is fine to be an opt-in behavior, it doesn't seem the most popular use case so far, so the current default seems good.

LielHarel: I understand, I have changed this and return the test like they were before. I will add more test for the second case :+1:

AbrilRBS (approved):
Thanks for taking the opportunity to contribute the fix by yourself, we really appreciate it :)

AbrilRBS:
Thanks! I've rebased this against the `release/2.0` branch - I'll fix the conflicts tomorrow :)

LielHarel:
> Thanks! I've rebased this against the `release/2.0` branch - I'll fix the conflicts tomorrow :)

I have tried to solve the conflicts of my PR - I hope it helps
Pay attention that there is a commit from`develop` I think that causes problems and I do not know if it really should be here.

LielHarel:
> Thanks! I've rebased this against the `release/2.0` branch - I'll fix the conflicts tomorrow

memsharded:
There was something breaking due to some concurrent changes (scons), not fault of this PR. I have updated the branch, hopefully now it will pass

conans/test/functional/toolchains/scons/test_sconsdeps.py
```diff
@@ -8,7 +8,7 @@
 
 @pytest.mark.skipif(platform.system() != "Linux", reason="SCons functional tests"
                                                          "only for Linux")
-@pytest.mark.tool("scons")
+@pytest.mark.tool_scons
```
memsharded:
```suggestion
@pytest.mark.tool("scons")
```

conan/tools/system/package_manager.py
```diff
@@ -234,7 +237,7 @@ def __init__(self, conanfile, arch_names=None):
 
         self._arch_separator = ":"
 
-    def install(self, packages, update=False, check=True, recommends=False):
+    def install(self, packages, update=False, check=True, host_package=True, recommends=False):
```
memsharded:
```suggestion
    def install(self, packages, update=False, check=True, recommends=False, host_package=True):
```
The likelihood to break existing code is lower if new arguments are added at the end.

conans/test/integration/tools/system/package_manager_test.py
```diff
@@ -229,11 +229,51 @@ def fake_check(*args, **kwargs):
             return ["package1", "package2"]
         from conan.tools.system.package_manager import _SystemPackageManagerTool
         with patch.object(_SystemPackageManagerTool, 'check', MagicMock(side_effect=fake_check)):
-            tool.install(["package1", "package2"])
+            tool.install(["package1", "package2"], host_package=True)
```
memsharded:
```suggestion
            tool.install(["package1", "package2"])
```
Default not necessary

memsharded (approved):

AbrilRBS (approved):
Thanks a lot! This will be available in Conan 2.0.12 :)

PR #14760, opened 2023-09-17
Add `--dry-run` for `conan remove`

### LINKED ISSUE #14245: [feature] conan remove should have a --dry-run option
### What is your suggestion?

`conan remove` asks for every package if it really should be deleted. This is really bothersome when removing many packages (e.g. when removing all but the last revision with `conan remove '*#!latest'`). 

However, it would be nice to see what is deleted before actually executing `conan remove -c`. It would be nice to have a `--dry-run` option for that use cases.

AbrilRBS:
Changelog: Feature: Add `--dry-run` for `conan remove`.
Docs: https://github.com/conan-io/docs/pull/3404

After discussing with the team, the deciding factor was the difference in defaults between the semantics of `conan list` selection pattern and that of `conan remove`, which might make the usage of the pkglist feature a bit suprising (Even if that's still the recommended method!)

Closes #14245

### DIFF
--- a/conan/cli/commands/remove.py
+++ b/conan/cli/commands/remove.py
@@ -67,7 +67,7 @@ def remove(conan_api: ConanAPI, parser, *args):
     cache_name = "Local Cache" if not remote else remote.name
 
     def confirmation(message):
-        return (args.confirm or ui.request_boolean(message)) and not args.dry_run
+        return args.confirm or ui.request_boolean(message)
 
     if args.list:
         listfile = make_abs_path(args.list)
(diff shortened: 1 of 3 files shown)


conan/cli/commands/remove.py
```diff
@@ -65,7 +67,7 @@ def remove(conan_api: ConanAPI, parser, *args):
     cache_name = "Local Cache" if not remote else remote.name
 
     def confirmation(message):
-        return args.confirm or ui.request_boolean(message)
+        return not args.dry_run and (args.confirm or ui.request_boolean(message))
```
memsharded: I am not sure the ``dry_run`` should skip the confirmation, maybe it should be done after the ``ui.request``?

conan/cli/commands/remove.py
```diff
@@ -91,7 +91,7 @@ def confirmation(message):
         if packages is None:
             if confirmation(f"Remove the recipe and all the packages of '{ref.repr_notime()}'?"):
                 conan_api.remove.recipe(ref, remote=remote)
-            else:
+            elif not args.dry_run:
```
memsharded: I think this reads a bit confusing. I'd remove the dry-run from the confirmation function and just:
```python
if confirmation():
    if not args.dry_run:
           conan_api.remove...
else:
```

memsharded (approved):

PR #14771, opened 2023-09-18
Conan 1.x: Do not output error in case of empty build_type

thorsten-klein:
Changelog: (Feature | Fix | Bugfix): Remove error output in case of non existent build_type setting

If there is no `build_type` setting considered in conanfile.py, `CMake` will throw an exception during`build()` (Which is correct). This means we must explicitly specify the `build_type` argument when calling `build()`.

Use case: A build tool which should always be build with build type "Release".

Unfortunately we get some ugly error output when doing this:
```
    settings = "os", "compiler", "arch"
    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build(build_type="Release")
```
output (Note: The build passes succesful):
```
ERROR: Don't specify 'build_type' at build time for single-config build systems
```

### DIFF
--- a/conan/tools/cmake/cmake.py
+++ b/conan/tools/cmake/cmake.py
@@ -104,11 +104,10 @@ class CMake(object):
         bf = self._conanfile.build_folder
         is_multi = is_multi_configuration(self._generator)
         if build_type and not is_multi:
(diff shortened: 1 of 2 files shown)


thorsten-klein:
Ref: according PR for conan 2.x: #14349

PR #14775, opened 2023-09-19
Fix min version flag

NoWiseMan:
Changelog: (Feature | Fix | Bugfix): 
Apple changes the way how the the min target os version flag is create
Docs: https://developer.apple.com/forums/thread/737595

This fix the flag, but the early returns are not really fitting into the current code style of the patched function.


NoWiseMan:
oh, it looks like I need to rebase to develop2 ... sorry

memsharded (commented):
For changes to the next 2.0.X, better branch from and do the PR to ``release/2.0`` branch, please. Thanks for contributing!

NoWiseMan:
OK, I close this PR and create a new one

PR #14504, opened 2023-08-17
Add support for Apple visionOS to Conan 2.0

stevn:
Changelog: Feature: Add support for Apple visionOS.
Docs:Omit

Closes: https://github.com/conan-io/conan/issues/14496

…f32db2322677d55d3d13 and resolve merge conflicts).

Issue: https://github.com/conan-io/conan/issues/14496
Original pull request for Conan 1.x: https://github.com/conan-io/conan/pull/14497

- [x] Refer to the issue that supports this Pull Request.
- [x] If the issue has missing info, explain the purpose/use case/pain/need that covers this Pull Request.
- [x] I've read the [Contributing guide](https://github.com/conan-io/conan/blob/develop/.github/CONTRIBUTING.md).
- [x] I've followed the PEP8 style guides for Python code.
- [ ] I've opened another PR in the Conan docs repo to the ``develop`` branch, documenting this one.

PR #14776, opened 2023-09-19
Fix min target flag for xros and xros-simulator

NoWiseMan:
Changelog: Fix: Fix min target flag for xros and xros-simulator.
Docs: omit

Apple changes the way the "min target os version" flag is created
see: https://developer.apple.com/forums/thread/737595

This fixes the generation of the min target version flag for visionsOS and visionsOS simulator

### DIFF
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -79,9 +79,11 @@ def apple_min_version_flag(os_version, os_sdk, subsystem):
     elif 'appletvsimulator' in os_sdk:
         flag = '-mtvos-simulator-version-min'
     elif 'xros' in os_sdk:
-        flag = '-mxros-version-min'
+        # need early return as string did not fit into the "normal" schema
+        return f'--target arm64-apple-xros{os_version}'
     elif 'xrsimulator' in os_sdk:
-        flag = '-mxros-simulator-version-min'
+        # need early return as string did not fit into the "normal" schema
+        return f'-target arm64-apple-xros{os_version}-simulator'
 
     if subsystem == 'catalyst':
         # especial case, despite Catalyst is macOS, it requires an iOS version argument
--- a/conan/tools/apple/apple.py
+++ b/conan/tools/apple/apple.py
@@ -80,7 +80,7 @@ def apple_min_version_flag(os_version, os_sdk, subsystem):
         flag = '-mtvos-simulator-version-min'
     elif 'xros' in os_sdk:
         # need early return as string did not fit into the "normal" schema
-        return f'--target arm64-apple-xros{os_version}'
+        return f'-target arm64-apple-xros{os_version}'
     elif 'xrsimulator' in os_sdk:
         # need early return as string did not fit into the "normal" schema
         return f'-target arm64-apple-xros{os_version}-simulator'


memsharded:
These lines were added by @stevn in https://github.com/conan-io/conan/pull/14504, it would be great to get their feedback.

Also, this functionality is clearly not covered by tests if we can change it and it doesn't break anything, it would be good to add some tests that validate this, or at least do some checking.

NoWiseMan:
FYI:
The current CMake nightly build has a similar "bug"

memsharded:
Any feedback wrt this change @stevn?
Do you think that you could try to add some tests to cover this change @NoWiseMan? If not sure, please ask for help, or we can try to contribute a test ourselves.

Thanks to both!

NoWiseMan:
Besides conan I did not any python development.
I can try to implement a test, but I first need to understand how to run them locally

memsharded:
Don't worry @NoWiseMan if you don't have much experience with Python, we can contribute the tests, no need to do a very large effort, thanks!

NoWiseMan:
I'm not sure if there is any support for visionOS in the Intel-based simulator.
If yes this PR need more work.

conan/tools/apple/apple.py
```diff
@@ -79,9 +79,11 @@ def apple_min_version_flag(os_version, os_sdk, subsystem):
     elif 'appletvsimulator' in os_sdk:
         flag = '-mtvos-simulator-version-min'
     elif 'xros' in os_sdk:
-        flag = '-mxros-version-min'
+        # need early return as string did not fit into the "normal" schema
+        return f'--target arm64-apple-xros{os_version}'
```
franramirez688:
```suggestion
        return f'-target arm64-apple-xros{os_version}'
```

NoWiseMan:
There is an intel-based simulator.

So the flag `-target arm64-apple-xros{os_version}` is only correct for arm
the intel version needs `-target x86_64-apple-xros{os_version}`
d
for that `def apple_min_version_flag(os_version, os_sdk, subsystem):` needs an additional `arch`parameter

memsharded:
Hi @NoWiseMan 

@stevn hasn't given any feedback regarding this change, and we are also overwhelmed, so difficult to test this too on our side.
If this change is working for you @NoWiseMan, I am happy to merge it as-is. If you could please explicitly comment on the different versions (cmake, OSX, sdk, etc) that are needed for this to work, that could be enough, and I could merge this for next 2.0.14. Thanks again!

stevn:
Hi,
yeah sorry for going quiet - I meant to look into this but currently can't find the time. Thanks

memsharded:
> yeah sorry for going quiet - I meant to look into this but currently can't find the time. Thanks

No prob! Thanks for following up 🙂

NoWiseMan:
Yes for me this is working.
Luckily with Xcode 15.1 Beta Apple has **removed** the x86_64 target 
-target **_x86_64_**-apple-xros1.0 
and we no longer need the arch parameter I mentioned earlier and the PR is  perfectly fine yet.

But there is still a unfixed issue in cmake 3.28rc2.
It is not possible to build cmake based libs with conan and cmake for visionOS simulator without patching 

Apple-Clang.cmake
```
....
  elseif(_CMAKE_OSX_SYSROOT_PATH MATCHES "/XRSimulator")
    set(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG "")
....
```

and replacing the flag with an empty string

memsharded:
Ok, so the PR seems good, and the patch that is still missing seems something on the CMake side, not something that Conan should fix, but if you think it would be easy to add something to the ``CMakeToolchain`` that automatically adds ``set(CMAKE_${lang}_OSX_DEPLOYMENT_TARGET_FLAG "")`` conditioned on the ``VisionOS`` and that could work, that could be added, feel free to open another PR if you'd like. 

I am merging this, thanks very much!

PR #14780, opened 2023-09-19
Fix cmake build --config XXXX for multi-config with hardcoded configurations

memsharded:
Changelog: Fix: Allow ``cmake.build(build_type="Release")`` for recipes built with multi-config systems but without ``build_type`` setting. 
Docs: Omit

Investigation on https://github.com/conan-io/conan/pull/14349


thorsten-klein:
👍🏻 

**Note 1:**

It seems that in `presets.py` the `CMAKE_BUILD_TYPE` of the toolchain is always overwritten by the value of the profile setting.
Is this behavior intended?

**Note 2:**
There are currently 2 possible places within `conanfile.py` where the `build_type` / `CMAKE_BUILD_TYPE` can defined.
1. tc.cache_variables["CMAKE_BUILD_TYPE"] = "MinSizeRel"
2. cmake.build(build_type="Release") 
The second one only has some effect if `is_multi_configuration`. This might be confusing for users.

**My expected behavior:**
priorities for value of CMAKE_BUILD_TYPE:
1. tc.cache_variables["CMAKE_BUILD_TYPE"]
2. 'build_type' if build_type is considered under settings
5. None (empty)

**Proposed changes:*
I have extended your test in order to reproduce & align to my expected behavior:
https://github.com/memsharded/conan/compare/fix/cmake_build_configs...thorsten-klein:conan:fix/cmake_build_configs

memsharded:
I have had a look to your code in your branch extending the test and your comment:

> My expected behavior:
priorities for value of CMAKE_BUILD_TYPE:

> tc.cache_variables["CMAKE_BUILD_TYPE"]
'build_type' if build_type is considered under settings

I am afraid this is against Conan design. The behavior for Conan is that downstream/consumer definitions should always have precedence over recipes definition. In this case, the profile ``build_type``, if specified, should be applied to recipes, as long as recipes depend on it ``settings = ..., "build_type"``. What is not correct is that the profile defines ``build_type=Release``, this is also stored in the ``conaninfo.txt`` and implies a ``package_id``, but being changed under the hood so the package binary is actually using a different build-type like ``MinSizeRel`` without Conan being aware of it in the binary representation. This cannot happen.

memsharded:
I see there are still some unknowns, I have done a review in https://github.com/memsharded/conan/pull/51#pullrequestreview-1785549184

czoido (approved):

Issue #14566, opened 2023-08-24
[feature] System Provided Packages


nicosmd:
### What is your suggestion?

Hey, I'm thinking about this kind of issue for a long time now and finally decided to create a feature request for this since I think there are also others struggling with that.

When dealing with `sysroots` it is a very common scenario that users consuming recipes from conan-center but want to have fine-grained control over which dependency is provided by the sysroot and therefore should not be consumed using conan. This could be the case when any vendor is providing closed-source libraries which are already linked against specific versions of Open Source dependencies in the sysroot ( https://github.com/conan-io/conan/issues/12597 ) or there are dependencies which needs to be consumed as a system package for other reasons ( https://github.com/conan-io/conan-center-index/issues/17101 ). In my project we are currently using ugly workarounds (heavily mis)using `override` and custom system package to deal with this and I would be really glad to get rid of them in future, since they are not scaling. 

There are several unresolved issues which goes in a similar direction:
- https://github.com/conan-io/conan/issues/14343
- https://github.com/conan-io/conan/issues/12478
- https://github.com/conan-io/conan/issues/12597
- https://github.com/conan-io/conan/issues/12459

and I would guess there are even more since this is a quite common use case.

Conan currently does not provide any convenient feature to archive this goal easily, which is really a pain point for us at the moment. That's why I want to already try to come up with a rough idea how this might be modeled into conan. Let's call this feature idea `System Provided Packages` for the moment.

System Provided Packages is inspired by the recently introduced system_tools feature, which allows users to define which tools are already provided by the system. Therefore I hope that the feasibility in general is given. 

Similar to `system_tools`, a new section needs to be introduced in the profile called e.g. [system_provided] where users can define, which version of any dependency is provided by the system or the sysroot and should neither be build by conan nor consumed from the conan cache. The version defined under [system_provided] should either override the version for a particular dependency in the whole graph, so that it is guaranteed that the system provided packages are being used, or show show up conflicts, which needs to be resolved upstream to have a more explicit solution.

[...]

Issue #12726, opened 2022-12-17
[question] Is it possible to customize lib version in CMake config file generated by `CMakeDeps`?


SpaceIm:
### What is your question?

In `vulkan-headers` recipe of conan-center, version in CMake config file generated by CMakeDeps must be "sanitized" to really follow upsteam CMake config file content (always `x.y.z` while we may have "sdk version" like `1.3.236.0` in conan-center). It's important because build system of a downstream recipe (`vulkan-validationlayers`) relies on this version in CMake config file to inject this value into a manifest file. If this value doesn't exactly follow this `x.y.z` convention, it leads to an illformed manifest file which will break logic of another library at runtime (`Vulkan-Loader`) when it tries to find validation layers.
see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/4973

So is it possible to change this version in `package_info()`? Something like this in `vulkan-headers` recipe for example:

```python
def package_info():
    vk_version = Version(self.version)
    self.cpp_info.set_property("cmakedeps_version", f"{vk_version.major}.{vk_version.minor}.{vk_version.patch}")
```

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

memsharded:
Hi @SpaceIm 

There is nothing in the ``cpp_info`` for this very specific need. Also, it doesn't look right to force this from one package to all its consumers, as some might want or need to use the full package version specification. If the consumers like ``vulkan-validationlayers`` don't want to use the full package version they can use a different one. I guess this might be possible to do a ``replace`` in ``generate()`` after the ``CMakeDeps.generate()``

SpaceIm:
> Also, it doesn't look right to force this from one package to all its consumers, as some might want or need to use the full package version specification

I though that the goal of `CMakeDeps` was to properly emulate upstream CMakeLists, and if it fails in this task it looks wrong to me. Recipe always know better than consumers what should be the content of CMake config or pkgconfig files.

> I guess this might be possible to do a replace in generate() after the CMakeDeps.generate()

It's fragile and it means that we have to patch forever. Actually it took me quite a lot of time to understand this bug (because interaction between 3 libraries are involved, and the final consequence is seen at runtime), and find that it was coming from emulation of vulkan-headers config file by CMakeDeps.

Why PkgConfigDeps has this feature and CMakeDeps should not have it?

memsharded:
This was solved some time ago with property ``system_package_version`` or ``package_version``.

PR #14808, opened 2023-09-22
add cmakedeps listening to system_package_version property

### LINKED ISSUE #14789: [feature] Add possibility to change the version written to XXXConfigVersion.cmake generated by CMakeDeps generator
### What is your suggestion?

Hey, 

I would like to propose a feature for the CMakeDeps generator. It is basically a feature which is already available for other generators like PkgConfigDeps. Using PkgConfigDeps generator it is possible to influence the version written to the `*.pc` file by setting `self.cpp_info.set_property("component_version", "1.14.12")` in the package_info stage.
This is actually a very useful feature when creating some kind of wrapper packages like system packages, where the package version might be something like `system` but pointing to an actual version of a library. Therefore it would be great to have something like this for the CMakeDeps generator as well to make the actual version of a library transparent to CMake, I think.

nicosmd:
Changelog: Feature: Add CMakeDeps and PkgConfigDeps generators listening to system_package_version property.
Docs: https://github.com/conan-io/docs/pull/3399

Close: https://github.com/conan-io/conan/issues/14789


memsharded (commented):
Thanks @nicosmd for your contribution.

It seems there is a gap in the ``CMakeDeps`` that is being fixed by https://github.com/conan-io/conan/pull/14813. This should follow the same pattern ``self.cmakedeps.get_property(...)``, to make sure that the overriding from consumer side also works.

If it is not clear, don't worry, I can try to fix that later. In any case, it would be necessary to add a test that covers it, if you think you can contribute it, great, otherwise, please let me know and we'll contribute it.

nicosmd:
Hey @memsharded, Ok, I will give it a try to fix it and to provide a test for this feature. Using the test from your PR as a template should make it easier for me. I will let you know if I need help. Thanks!

nicosmd:
@memsharded fixed the findings. Let me know if you have further feedback.

memsharded (commented):
Sorry for the delay, it has been some busy weeks.
I am reviewing this, and realized that ``component_version`` is used in the PkgConfigDeps context specifically for components versions, not for the package version, so this might not be the best choice of variable. I'll review the goal and alternatives and update, thanks for your patience.

nicosmd:
Thank you very much @memsharded, let me know when further work is necessary for this PR

memsharded:
Ok, yes, it seems that ``component_version`` would be a bit misleading name, and we would need a different property for this one.
The test is doing a check using the consumer ``CMakeDeps.set_property(...)``, but if I understood it correctly, the main use case would be for ``pkg/system`` recipes, and then it would be the responsibility of the "system" recipe to detect the system version automatically and define it in its ``package_info()`` method? How does this look like? How is it possible to check the version of some system libraries, for example? Do you have some real examples of system library + version + find_package() with specific version number that illustrate the case? What system dependency triggered this in your case? That would help to come up with a property name, something like ``package_version`` seems a bit confusing to me, like why would you define this if you already have the ``self.version`` which is the version of the package? (yes, for system libraries, but sounds confusing). Thanks very much!

nicosmd:
Hey @memsharded, 

> Ok, yes, it seems that component_version would be a bit misleading name, and we would need a different property for this one.

Yeah, I also had the feeling that this might be not the right property. I think I did not made the use case clear enough in the beginning so let me try to answer your questions so we can find a appropriate name for the property. 

> The test is doing a check using the consumer CMakeDeps.set_property(...), but if I understood it correctly, the main use case would be for pkg/system recipes, and then it would be the responsibility of the "system" recipe to detect the system version automatically and define it in its package_info() method?

This is exactly how I'm currently using it. The system package itself is responsible for setting the property correctly.

> How does this look like?

I have created a system package for libalsa, which looks like: 

```python
class LibalsaConan(ConanFile):
    name = "libalsa"
    version = "system"
    license = "LGPL-2.1-or-later"
    url = "https://github.com/conan-io/conan-center-index"
    homepage = "https://github.com/alsa-project/alsa-lib"
    topics = ("alsa", "sound", "audio", "midi")
    description = (
        "Library of ALSA: The Advanced Linux Sound Architecture, that provides audio "
        "and MIDI functionality to the Linux operating system"
    )
    package_type = "shared-library"
    settings = "os", "arch"
    build_policy = "missing"
    upload_policy = "skip"

    revision_mode = "scm"

    def build_requirements(self):
        self.tool_requires("pkgconf/1.9.3")

    def export(self):
        git = Git(self, self.recipe_folder)
        scm_url, scm_commit = git.get_url_and_commit()
        update_conandata(self, {"sources": {"commit": scm_commit, "url": scm_url}})

    def package(self):
        pkg_config = PkgConfig(self, "alsa")
        cpp_info = CppInfo(self)
        cpp_info.set_property("component_version", pkg_config.version)
        pkg_config.fill_cpp_info(cpp_info, is_system=False, system_libs=["m"])
        cpp_info.save(os.path.join(self.package_folder, "cpp_info.json"))

    def package_info(self):
        self.cpp_info = CppInfo(self).load("cpp_info.json")
        self.cpp_info.set_property("pkg_config_name", "alsa")
        self.cpp_info.set_property("cmake_file_name", "ALSA")
        self.cpp_info.set_property("cmake_target_name", "ALSA::ALSA")

```

To be honest, I have the feeling that I am using a bug here as I think `component_version` should not be applicable to `self.cpp_info` directly but only to `self.cpp_info.components[]` but it works for PkgConfigDeps generator. This was necessary since gst-plugins-base for gstreamer which is checking the actual version of libalsa for compatibility and it failed if it was just `system`. 

> Do you have some real examples of system library + version + find_package() with specific version number that illustrate the case? What system dependency triggered this in your case?

As I said, I had the error mainly when using the libalsa system package together with the official gstreamer plugin gst-plugins-base package as here it actually failed. But this package is build using Meson + PkgConfig and I mainly raised this issue for consistency reasons. When for meson builds it is possible to make the actual version used from the system transparent to the build it should also be possible for CMake, I think. I'm quite sure that there are cases where the actual used version of a system dependency is required also during CMake builds. This might be relevant for `openssl` for example. When cross compiling for embedded platforms, I might want to use the openssl library provided by the sysroot, so I create a `openssl/system` package for it like the alsa example above. Now there are projects which actually validating the openssl version like [the Fast-DDS project](https://github.com/eProsima/Fast-DDS/blob/f631faad7539d5aa98dd54a47bccecd8b8d30c9c/src/cpp/CMakeLists.txt#L525) so it will fail when they are receiving `system` instead of the actual version of the system. And I expect there are many other use cases for this. 

I'm not sure if this feature is only relevant in cross compilation context where it is necessary to work with sysroots. At least this is the case for me so far. Maybe the property name could go in this direction? Like `sysroot_package_version`? 

To be honest, it is also not completely clear to me if this feature is maybe only required for me since I'm workaround the missing feature described in #14566 by creating those system packages and it is not a good idea to add this feature since it wont be useful once a big solution is available. This would depend a bit on when a big solution could become available, I think. But maybe you have an opinion here.

Sorry for the long answer but I wanted to make the use case as clear as possible.

nicosmd:
Hey again @memsharded,

Some addition: I came across this question by chance #12726, which is exactly what this PR and my question is about. So it seems that there are more use cases for this feature rather then working with sysroots.

memsharded:
I have finally renamed it to ``system_package_version``, I think the intent and scope is way more clear, and also avoids misuse when the package version is already defined in the recipe, wdyt?

nicosmd:
I think, that should fit, thanks a lot! One question: As I've mentioned I use 'component_version' property for PkgConfigDeps generator to archive the same goal but my feeling is, that I'm misusing the property for that purpose. Do you think PkgConfigDeps should also get a dedicated `system_package_version` property?

memsharded:
> I think, that should fit, thanks a lot! One question: As I've mentioned I use 'component_version' property for PkgConfigDeps generator to archive the same goal but my feeling is, that I'm misusing the property for that purpose. Do you think PkgConfigDeps should also get a dedicated system_package_version property?

You are right. I have just added processing of ``system_package_version`` in ``PkgConfigDeps`` too

nicosmd:
Thanks, is there something needed from my side to finalize the PR?

memsharded:
Nop, all good, just merged, it will be in next 2.0.14. Thanks again!!

PR #14814, opened 2023-09-24
Fix/version range pres

### LINKED ISSUE #14803: How to include pre-releases in the lower bound of a version range
In Conan version expression, how do I define the range to include all versions compatible with "2"? Including: 2.0.0-pre, 2.0.0, 2.1.0. 

The 2.0.0-pre is important! It seems the expressions "~2-" and ">=2-" all exclude the 2.0.0-pre.

This could be an overlook in the lower bound of version ranges, we need to check it.

memsharded:
Changelog: BugFix: Version ranges ending with ``-`` do not automatically activate pre-releases resolution in the full range.
Changelog: BugFix: Fix version ranges so pre-releases are correctly included in the lower bound and excluded in the upper bound.
Docs: https://github.com/conan-io/docs/pull/3454

Close https://github.com/conan-io/conan/issues/14803


AbrilRBS (approved):

czoido (approved):

AbrilRBS (approved):
We just need to make it clear in the docs that the range to allow for 2.0-prereleases is now [... <=2] where before it was only [... <2] when activating pres

Just for confirmation: [>1 <=2.0-] does not actually accept the 2.0-pres anymore

aander80:
Very happy to see these changes, thanks! 😄

PR #14822, opened 2023-09-26
Add warning for suprising version range option behaviour

AbrilRBS:
Changelog: Omit
Docs: https://github.com/conan-io/docs/pull/3437

This warns on version ranegs like `[>=1 <2,include_prerelease=False]`, which does exactly the opposite of what the user wants.

We might want this as an error, or to listen to the value if assigned

### DIFF
--- a/conans/model/version_range.py
+++ b/conans/model/version_range.py
@@ -94,6 +94,11 @@ class VersionRange:
         prereleases = False
         for t in tokens[1:]:
             if "include_prerelease" in t:
+                if "include_prerelease=" in t:
+                    from conan.api.output import ConanOutput
+                    ConanOutput().warning(
+                        f'include_prerelease version range option in "{expression}" does not take an attribute, '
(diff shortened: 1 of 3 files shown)


conans/model/version_range.py
```diff
@@ -94,6 +94,11 @@ def __init__(self, expression):
         prereleases = False
         for t in tokens[1:]:
             if "include_prerelease" in t:
+                if "include_prerelease=" in t:
+                    from conan.api.output import ConanOutput
+                    ConanOutput().warning(
+                        f'include_prerelease version range option in "{expression}" does not take an attribute, '
+                        'its presence unconditionally enables prereleases')
```
franramirez688, line 101: Having a simple unittest checking that this message is appearing could be great: `conans/test/unittests/model/version/test_version_range.py`

memsharded (approved):

franramirez688 (approved):

Issue #14519, opened 2023-08-18
[bug] Possible Bazel toolchain deficiencies


valgur:
Copying over some observations regarding the Bazel generators / build system from one of the few CCI recipes targeting a Bazel-based project: https://github.com/conan-io/conan-center-index/pull/17575#discussion_r1205539939

* The `tools.build:compiler_executables` config values were not being applied.
* The `cppstd` value was not being translated into a `--std=c++14` build flag, as is done with `CMakeToolchain`. I might simply be using the `BazelToolchain` incorrectly, though.
* There's no obvious way (to me at least) how the build results should be installed for `package()` using Bazel. I added some explicit `copy()` commands in the conanfile for this. 

Conan version: 2.0.4

### Steps to reproduce

1. Clone https://github.com/embarktrucks/embag/compare/master...valgur:embag:conan
2. Run `conan create` on the recipe.

memsharded:
Thanks @valgur for the report.
It is very likely that the Bazel integration is far less mature than other integrations. @franramirez688 has pending some work to revisit the integration and some of the open tickets around it.

franramirez688:
Hi @valgur 

Thanks a lot for reporting this!

All your points should be solved by https://github.com/conan-io/conan/pull/14958, but the latest one:

> There's no obvious way (to me at least) how the build results should be installed for package() using Bazel. I added some explicit copy() commands in the conanfile for this.

Bazel has no explicit mechanism to do that, so it's not still obvious to make it automatically. I hope that future versions of Conan can automatize it in a better way.

valgur:
> Hi @valgur 
> 
> Thanks a lot for reporting this!
> 
> All your points should be solved by https://github.com/conan-io/conan/pull/14958, but the latest one:
> 
> > There's no obvious way (to me at least) how the build results should be installed for package() using Bazel. I added some explicit copy() commands in the conanfile for this.
> 
> Bazel has no explicit mechanism to do that, so it's not still obvious to make it automatically. I hope that future versions of Conan can automatize it in a better way.

Thank you for the hard work on this, @franramirez688! Looks like everything I previously mentioned has been addressed and the issue can be closed.

PR #14845, opened 2023-09-28
[Bazel][Part 1]: Bazel/BazelToolchain refactor

### LINKED ISSUE #14282: [feature] (Possibility to) change output path for generated Bazel BUILD files by BazelDeps
### What is your suggestion?

Hello!

I have tried using Bazel using conan 2.0.7. While I can get it running, I have noticed that the dependency BUILD files are in subfolders directly under the build_folder. If bazel_layout is used, this causes `<dep>/BUILD` files to be generated. In git directories, this requires adding <dep> to .gitignore for each dependency we may have, which is tedious to maintain. A suggestion for solving this is to either put all of the generated folders per requirement into a subfolder under generators folder (something like `generators_folder/conandeps/<dep>/BUILD`), or to provide an option when calling BazelDeps to put the generated files into a single subfolder; this would help sorting generated files and avoid missing to add generated files to a git repo.

I guess this is related to the changes with regards to #10676.

An example for reproducing this:

``` shell
$ git clone https://github.com/aander80/bazel-test && cd bazel-test
$ conan --version
Conan version 2.0.7
$ cd conan
$ conan create say
$ cd hello
$ conan install .
# say/BUILD is now generated
```

Another suggestion for properly maintaining Bazel repos using conan is very much appreciated; I am new to Bazel and am happy to learn!

franramirez688:
Changelog: Feature: BazelToolchain creates a `conan_bzl.rc` file which defines the `conan-config` configuration. If it exists, Bazel helper will use it automatically.
Changelog: Global refactor in BazelToolchain and Bazel conan.tools.google classes.
Closes: https://github.com/conan-io/conan/issues/14282 https://github.com/conan-io/conan/issues/14519

Summary:

* bazel_layout is now creating all the Conan-generated files inside another folder, `bazel-conan-tools`.
     * WORKSPACE file (consumer side) before my change:

      load("@//:dependencies.bzl", "load_conan_dependencies")
      load_conan_dependencies()

     * Now:

      load("@//bazel-conan-tools:dependencies.bzl", "load_conan_dependencies")
      load_conan_dependencies()
  
* BazelToolchain is generating a `conan_bzl.rc` file with a `build:conan-config` configuration depending on your settings/options.
* BazelToolchain is not generating a real Bazel toolchain (https://bazel.build/extending/toolchains) because they require much more knowledge and expertise in Bazel to make a dynamic Bazel toolchain per user configuration.
* Bazel is moving forward to platforms (https://bazel.build/extending/platforms) for future releases instead of toolchains but it does not imply that BazelToolchain could be incompatible but some changes to inject those platforms constraints in our conan-config configuration.


conan/internal/api/new/bazel_lib.py
```diff
@@ -9,12 +9,12 @@
 class {{package_name}}Recipe(ConanFile):
     name = "{{name}}"
     version = "{{version}}"
-    package_type = "library"
+    package_type = "static-library"
```
franramirez688, line 12: static one as Bazel, AFAIK, has no dynamic way to build a shared/static library. It has options as `--dynamic_mode=fully` but it's not working as expected. The only way that I got it was to add `linkshared=True` to the BUILD file in the test_package but it made useless the use of `shared=True` dynamically

conan/tools/google/toolchain.py
```diff
@@ -1,24 +1,154 @@
...
+        # FIXME: Keeping this option but it's not working as expected. It's not creating the shared
+        #        libraries at all.
+        self.dynamic_mode = "fully" if shared else "off"
+        self.cppstd = cppstd_flag(self._conanfile.settings)
+        self.copt = []
+        self.conlyopt = []
+        self.cxxopt = []
+        self.linkopt = []
+        self.strip = 'sometimes'  # 'always', 'sometimes', 'never'
+        self.compilation_mode = "fastbuild"  # 'fastbuild', 'dbg', 'opt'
```
czoido, line 86: Checking the docs I think we don't need to define `strip` because it has a default of `sometimes` in Bazel, that will only strip the debug information for `fastbuild` mode and `fastbuild` mode I think that should only be defined by the user because it does not directly correspond to neither Debug or Release of our settings. So I would propose setting something like:

```
self.strip = None
self.compilation_mode = {'Release': 'opt', 
                         'Debug': 'dbg'}.get(self.settings.build_type)
```

The user will be able to change this in case they have another custom build type defined.

conan/tools/google/toolchain.py
```diff
@@ -1,24 +1,154 @@
...
+        self.extra_ldflags = []
+        self.extra_defines = []
+
+        # Bazel build parameters
+        shared = self._conanfile.options.get_safe("shared")
+        fpic = self._conanfile.options.get_safe("fPIC")
+        self.force_pic = fpic if (not shared and fpic is not None) else None
+        # FIXME: Keeping this option but it's not working as expected. It's not creating the shared
+        #        libraries at all.
+        self.dynamic_mode = "fully" if shared else "off"
```
franramirez688, line 79: As I commented above, it's not working as expected because it depends on the user's BUILD file. I've asked in the official Bazel Slack channel to know more about this.

conan/tools/google/layout.py
```diff
@@ -1,11 +1,13 @@
...
-    # used in test package for example, to know where the binaries are (editables not supported yet)
-    conanfile.cpp.build.bindirs = ["bazel-bin"]
-    conanfile.cpp.build.libdirs = ["bazel-bin"]
+def bazel_layout(conanfile, src_folder="."):
+    """Bazel layout is so limited. It does not allow to create its special symlinks in other
+    folder. See more information in https://bazel.build/remote/output-directories"""
+    subproject = conanfile.folders.subproject
+    conanfile.folders.source = src_folder if not subproject else os.path.join(subproject, src_folder)
+    conanfile.folders.build = "."  # Bazel always build the whole project in the root folder
+    conanfile.folders.generators = "bazel-conan-tools"  # one
```
czoido, line 10: Minor thing, but what about changing the name of the folder to `conan` or `generators` like in other generators layouts?

```suggestion
    conanfile.folders.generators = "conan"  # one
```

franramirez688:
Superseded by https://github.com/conan-io/conan/pull/14958

PR #14846, opened 2023-09-28
Properly lock release candidates of python requires

AbrilRBS:
Changelog: Bugfix: Properly lock release candidates of python requires
Docs: Omit

The `core.version_ranges:resolve_prereleases` conf works fine for resolving python require prereleases, but a missing argument was not letting lockfiles to also work, this fixes it

### DIFF
--- a/conans/test/integration/lockfile/test_lock_pyrequires.py
+++ b/conans/test/integration/lockfile/test_lock_pyrequires.py
@@ -95,6 +95,8 @@ def test_lock_pyrequires_prereleases():
     tc.run("export dep")
     dep_rrev = tc.exported_recipe_revision()
     tc.run("lock create app --lockfile-out=app.lock", assert_error=True)
+    # Makes sense, prereleases are not active
+    assert "Version range '>=0' from requirement 'dep/[>=0]' required by 'python_requires' could not be resolved"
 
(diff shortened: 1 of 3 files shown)


conans/test/integration/lockfile/test_lock_pyrequires.py
```diff
@@ -87,6 +88,21 @@ def configure(self):
     assert "pkgb/0.2: dep: dep/0.2!!" in client.out
 
 
+def test_lock_pyrequires_prereleases():
+    tc = TestClient()
+    tc.save({"dep/conanfile.py": GenConanfile("dep", "1.0-0.1"),
+             "app/conanfile.py": GenConanfile("app", "1.0").with_python_requires("dep/[>=0]")})
+    tc.run("export dep")
+    dep_rrev = tc.exported_recipe_revision()
+    tc.run("lock create app --lockfile-out=app.lock", assert_error=True)
```
memsharded, line 97: Good be nice to check what error is being raised, just in case

memsharded (approved):
Very good catch!
Indeed ``resolve_prereleases=None`` was there as default, but no one passing ever a value into it 🤦‍♂️

franramirez688 (approved):
👏

PR #14852, opened 2023-09-29
fix conan source with python_requires in remotes

### LINKED ISSUE #14848: [bug] conan source failing with python_requires
### Environment details

* Conan version: 2.0.9 & 2.0.13
* Python version: 3.11.0

### Steps to reproduce

1. Use a "python_requires" in the recipe
2. Make sure the python requires recipe is **not** in the local cache, only on a remote
3. Call `conan source .`

By looking at the call stack, it seems no remote is ever defined from the `conan source` command, which explains why at the time the python requires is searching in remotes, it fails.

### Logs

```
 > conan source .
test_package/0.1.0: Not found in local cache, looking in remotes...
ERROR: Error loading conanfile at 'D:\workdir\test_package\conanfile.py': Cannot resolve python_requires 'recipe_utils/0.1.0': No remote defined
```

Call stack:
image
Looking at each steps from the very beginning (source method) the `remotes` variable is always empty.

memsharded:
Changelog: Fix: Define remotes in ``conan source`` command in case recipe has ``python_requires`` that need to be downloaded from remotes.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14848

### DIFF
--- a/conans/test/integration/command/source_test.py
+++ b/conans/test/integration/command/source_test.py
@@ -289,3 +289,4 @@ def test_source_python_requires():
     c.save({"conanfile.py": GenConanfile().with_python_requires("pytool/0.1")}, clean_first=True)
     c.run("source . ")
     assert "pytool/0.1: Not found in local cache, looking in remotes" in c.out
+    assert "pytool/0.1: Downloaded recipe" in c.out
--- a/conans/test/integration/command/source_test.py
+++ b/conans/test/integration/command/source_test.py
@@ -277,3 +277,15 @@ class TestSourceWithoutDefaultProfile:
         client.run("source .")
         assert "conanfile.py: Calling source()" in client.out
         assert "CACHE:MYCACHE!!" in client.out
+
+
+def test_source_python_requires():
+    c = TestClient(default_server_user=True)
+    c.save({"conanfile.py": GenConanfile("pytool", "0.1")})
+    c.run("export . ")
+    c.run("upload * -r=default -c")
+    c.run("remove * -c")
+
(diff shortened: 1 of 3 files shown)


AbrilRBS (approved):

AbrilRBS (approved):
Between this and what we found a few weeks ago, wondering if we should check the api's usage in the code, in case we can cath some extra issues

conans/test/integration/command/source_test.py
```diff
@@ -277,3 +277,15 @@ def source(self):
...
+def test_source_python_requires():
+    c = TestClient(default_server_user=True)
+    c.save({"conanfile.py": GenConanfile("pytool", "0.1")})
+    c.run("export . ")
+    c.run("upload * -r=default -c")
+    c.run("remove * -c")
+
+    c.save({"conanfile.py": GenConanfile().with_python_requires("pytool/0.1")}, clean_first=True)
+    c.run("source . ")
+    assert "pytool/0.1: Not found in local cache, looking in remotes" in c.out
```
docbrown1955, line 291: Depending on what this test is supposed to check, there might be an assert missing. The "looking in remotes" message is always displayed, including in the pathological case fixed in this PR.

The issue #14848 shows this in the log, where the "looking in remotes" is displayed just before the failure.

AbrilRBS: The test would fail if Conan returned a non 0 exit code so the test is covering the case, but you're right, it might be nice to assert more output :)

czoido (approved):

PR #14871, opened 2023-10-04
New [platform_requires]  similar to system-tools, but for all requires

### LINKED ISSUE #1330: Add ability to force using system-installed dependency instead of Conan package
Hello, I am writing a Qt-based library and want to create a Conan package.
But, I want to give the end-user a choice:
- install Qt with Conan, as a usual Conan dependency

OR

- make Conan assume, that Qt is already installed by the platform-dependent package manager (apt, rpm, homebrew, ... whatever), and rebuild the package from source (to mitigate the possible version incompatibility).

But, I want this mechanism to be more generic and standardized, so the end-user could mark any dependency to be "system-installed" (using Conan profile, for example) and all packages, dependent on this will be rebuilt, assuming, that this dependency is present on the system.

### LINKED ISSUE #14566: [feature] System Provided Packages

### LINKED ISSUE #15101: Extend [system_tools] to [system_packages]
### What is your suggestion?

As of now, we can use profiles to describe our different build environments. Conan also has a possible to integrate with system package managers to resolve dependencies, which may be present in a base system. 
This is nice, but a lot of times, especially in embedded usecases the whole system setup is a bit more subtle. So why don't take the [system_tools] approach a bit further, and extend it to [system_packages].

Often, when dealing with embedded devices, they come with an SDK, which I will refer to as "the platform".
A typical usecase would be that I want to build my (existing) software stack for such a new platform.
This means that for this concept of a platform (which could, but musn't be a setting - maybe an OS subsetting) - I need to be able to proxy Conan recipes, or tell Conan that there is a set of dependencies, which it shouldn't resolve. Ideally, I don't want to touch existing recipes in the process, but opt-in into the whole thing.
What would be even better, if I could provide a fixed version for a given library, for a given platform, and Conan would at least tell me if version checks are satisfied, or the platform versions are conflicting in any places.

```
--------------------------
  Platform X
 ``` 

```
--------------------------
  Platform Y
  - libA == 1.0
```  
  
```  
--------------------------
  Platform Z
  - libA == 1.1
  - libB == 1.0
```  
  
So let's see the recipes:
`libA` & `libB` don't have any dependencies, they are available in Version 1.0 and 1.1
App would be:
```
def requirements():
  self.requires("libA/[>=1.0]@")
  self.requires("libB/[>=1.0]@")
```
  
So now I would like to build App for Platform X, Y and Z.
For X, conan would resolve `libA=1.1`, `libB=1.1`, build those and also build App.
For Y, conan would see libA is in the platform. So it would resolve `libB=1.1`, build it, and then build App, using libB from Conan and libA from the platform. 
For Z, both libA and libB are available in the platform, so Conan can go ahead, App would resolve as if it had no dependencies, and App could build entirely against the platform.

Even better if I had
```
---------------------------
  Platform U
  - libA == 0.9
 ```
 
Conan would tell me, that App cannot be build, because requirement libA/[>=1.0] is not fulfilled by the platform.

Like with `[system_tools]`, it could be modeled in the profile.

```
[settings]
os=Linux
os.distribution=platformZ
arch=x86_64
compiler=gcc
compiler.version=9
compiler.libcxx=libstdc++11
compiler.cppstd=gnu11

[system_packages]
libA=1.1
libB=1.0
```  
   
If you are targeting only Windows & Ubuntu, something like this might not be a big of a deal. (Although of today our recipes are not very nice looking, trying just to model that for the different Ubuntu platfoms). But if you have a software stack, which you want to build on top of a bunch of different embedded systems, it is a real usecase - and a real pain, something that Conan does not handle well yet.
So curious to hear what you think!

memsharded:
Changelog: Feature: New ``[platform_requires]`` profile definition to be able to replace Conan dependencies by platform-provided dependencies.
Changelog: Fix: Deprecate ``[system_tools]`` in favor of ``[platform_tool_requires]`` to align with ``[platform_requires]`` for regular dependencies. Changed output from "System tool" to "Platform".
Docs: https://github.com/conan-io/docs/pull/3495

Close https://github.com/conan-io/conan/issues/14566
Address https://github.com/conan-io/conan/issues/15001
Close https://github.com/conan-io/conan/issues/15101
Close https://github.com/conan-io/conan/issues/1330

TO LOOK INTO:
- Renamed to ``[system_requires]``, I didn't like ``[system_deps]``, but still not convinced because of overlap with ``system_requirements()``
- The cases of replacement of a requires with another Conan recipe, more like override, needs to be treated in a different way
- Some generalization of "System tool" => "System requirement" is needed, but lets agree on the naming first


memsharded:
Possible naming:
```
PROFILE
[alternative_requires]
protobuf/*: protobuf/*@system

[alternative_tool_requires]



[system_tools]
cmake/3.25

[system_bypass]
[platform_requires]
[deferred]
[bypass_requires]
[system_dependencies]
boost/1.81.0#rev1:package_id_system

```

czoido:
I think I would try to make the name consistent with the _requires_ and _tool_requires_... and change `system_tools` to `<whatever_we_choose>_tool_requires`. 

From the names in the issue and others I could come up with, the ones I would prefer are:

- `platform_tool_requires` and `platform_requires`
- `preinstalled_tool_requires` and `preinstalled_requires`
- `defaulted_tool_requires` and `defaulted_requires`

KerstinKeller:
Actually, I think I like best the
`defaulted_tool_requires` and `defaulted_requirest` - because it describes best what the feature actually does: it provides default versions to use for a specific profile.

That they are given / defined by the underlying platform, could be one reason.
But it could alswo be used to provide fix "sdks", against which users shall develop for a given project.

memsharded:
After some discussion we are settling down with the ``platform_xxx``. Because this feature defers to things that need to be provided by the platform (os, sysroot, toolchain). It is different than the ``replace/alternatives`` proposal in the other PR.

czoido (approved):

conans/model/dependencies.py
```diff
@@ -133,7 +133,7 @@ def direct_host(self):
 
     @property
     def direct_build(self):
-        return self.filter({"build": True, "direct": True}, remove_system_tools=True)
```
franramirez688, line 136: Could this change be breaking?

franramirez688 (approved):
Looks promising! Only one doubt 😁

conans/model/dependencies.py
memsharded: I changed the default above to ``remove_system=True`` so the net default should still be the same. the ``filter()`` method itself is not documented yet, so it shouldn't be a problem

conans/client/graph/graph_builder.py
```diff
@@ -246,13 +245,12 @@ def _create_new_node(self, node, require, graph, profile_host, profile_build, gr
                                      "host dependency")
             require.ref.version = transitive.require.ref.version
 
+        resolved = self._resolved_system(node, require, profile_build, profile_host,
```
AbrilRBS, line 248: I like the change for `platform_requires` to now be lockable (if Im reading the change right!), maybe we should add a changelog entry for that too as a bugfix?

At the very least, I think that at least one test should now check that the lock properly works, just in case :)


Edit: Thinking about it, maybe _not_ locking on the subsitution was ok? Let me sleep on it, see what I think tomorrow about it

AbrilRBS (commented):
I love how clean this is! I only have one comment about the lockability of the platform_requires, otherwise looks good!

conans/model/dependencies.py
franramirez688: Oh, sorry, I did not realize about it. Thanks!

conans/client/graph/graph_builder.py
memsharded: I thought the behavior before was buggy, the relevant change was in ``def test_system_tool_require_range(self):`` (this is the reason I didn't refactor the names in this test @franramirez688, to not hide this change in the diff)

Update, sorry, my bad, this was actually one of my "search and replace" to fix broken tests. Sorry about it, and it is bad it is hiding the test changed, I should have highlighted it.

But it is true that this is questionable, lets have a second thought about it and discuss it.

AbrilRBS (approved):

PR #14882, opened 2023-10-06
Refactor of ClientCache/Home

memsharded:
Changelog: Omit
Docs: Omit

Refactor to remove some state from ``ClientCache``, to prepare for having a better access to ``global.conf`` for cli-args definition of ``core.xxxx`` confs.

Note that every ``ClientCache=>HomePaths`` replacement is avoiding a full parsing+jinja template of the ``global.conf`` configuration, for example.


conan/internal/cache/home_paths.py
```diff
@@ -0,0 +1,73 @@
...
+    def custom_generators_path(self):
+        return os.path.join(self._home, EXTENSIONS_FOLDER, "generators")
+
+    @property
+    def hooks_path(self):
+        return os.path.join(self._home, EXTENSIONS_FOLDER, HOOKS_EXTENSION_FOLDER)
+
+    @property
+    def wrapper_path(self):
+        return os.path.join(self._home, "extensions", "plugins", "cmd_wrapper.py")
```
AbrilRBS: Bikesheding: I guess we can use the `EXTENSIONS_FOLDER` const in these other functions too, even if it makes no difference at all

AbrilRBS (approved):
N i c e, this reads so much cleaner now!

conan/internal/cache/home_paths.py
memsharded: Yeah, that was some copy&paste, lets do that.

czoido (approved):

PR #14912, opened 2023-10-11
Feature/better version ranges

### LINKED ISSUE #14905: [bug] unexpected version resolving conflict with range
### Environment details

* Operating System+version: Windows 11 21H2
* Compiler+version: no matter
* Conan version: 2.0.10
* Python version: 3.10.8

### Steps to reproduce

1. write a simple conanfile.txt:

```
[requires]
paho-mqtt-cpp/[>=1.2.0 <2]
openssl/[>=1.1.1 <2]
```

2. run `conan install conanfile.txt` or `conan graph info conanfile.txt`
3. version conflict

However, this conanfile.txt (using qualified version) works:

```
[requires]
paho-mqtt-cpp/[>=1.2.0 <2]
openssl/1.1.1w
```

and this (just change the order) works too:

```
[requires]
openssl/[>=1.1.1 <2]
paho-mqtt-cpp/[>=1.2.0 <2]
```

The `paho-mqtt-cpp` actually requires `openssl/[>=1.1 <4]`

As i understand, version range in downstream will (and should) take precedence (this is [documented in Conan 1](https://docs.conan.io/1/versioning/version_ranges.html) but not mentioned in Conan 2 Doc). Those three conanfile.txt show some inconsistency.

We may need a bugfix or clarifying in document.

### Logs

Here is the log showing conflict:

```
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmake_layout:build_folder_vars=['settings.arch']

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmake_layout:build_folder_vars=['settings.arch']


Graph root
    conanfile.txt: D:\src\test_conan\fuck_range\conanfile.txt
Requirements
    openssl/3.1.2#8879e931d726a8aad7f372e28470faa1 - Cache
    paho-mqtt-c/1.3.9#3bec15070efa4907c9750d43afce9040 - Cache
    paho-mqtt-cpp/1.2.0#8fb67b3e02d623651be24b62cbae0d8b - Cache
    zlib/1.3#06023034579559bb64357db3a53f88a4 - Cache
Build requirements
    nasm/2.15.05#ca1c14c2060af0ec151048b54ced9262 - Cache
    strawberryperl/5.32.1.1#8f83d05a60363a422f9033e52d106b47 - Cache
Resolved version ranges
    openssl/[>=1.1 <4]: openssl/3.1.2
    paho-mqtt-cpp/[>=1.2.0 <2]: paho-mqtt-cpp/1.2.0
    zlib/[>=1.2.11 <2]: zlib/1.3
ERROR: Version conflict: Conflict between openssl/[>=1.1.1 <2] and openssl/3.1.2 in the graph.
Run conan graph info with your recipe and add --format=html to inspect the graph errors in an easier to visualize way.
```

memsharded:
Changelog: Feature: Compute version-ranges intersection to avoid graph version conflicts for compatible ranges
Docs: Omit

Close https://github.com/conan-io/conan/issues/14905


AbrilRBS:
I've pushed support for or'ed expressions and check that prereleases work as expected - But more testing needs to be done, I'm not confident in either of those things to 100%

conans/test/unittests/model/version/test_version_range_intersection.py
```diff
@@ -7,34 +7,40 @@
     ['>1.0', ">1.0", ">1.0"],
     ['>=1.0', ">1.0", ">1.0"],
     ['>1.0', ">1.1", ">1.1"],
-    ['>1.0', ">=1.1", ">=1.1"],
-    ['>=1.0', ">=1.1", ">=1.1"],
+    ['>1.0', ">=1.1", ">=1.1-"],
```
memsharded: I am not sure why the result of the intersection has changed.

conans/model/version_range.py
```diff
@@ -167,29 +167,26 @@ def contains(self, version: Version, resolve_prerelease: Optional[bool]):
...
+            if limits:
+                return sorted(limits, reverse=operator == ">")[0]
+
+        for lhs_conditions in self.condition_sets:
+            for rhs_conditions in other.condition_sets:
+                internal_conditions = []
+                lower_limit = _calculate_limits(">", lhs_conditions, rhs_conditions)
+                upper_limit = _calculate_limits("<", lhs_conditions, rhs_conditions)
+                internal_conditions.append(lower_limit)
+                internal_conditions.append(upper_limit)
```
memsharded: just ``internal_conditions = [lower_limit, upper_limit]``?

conans/model/version_range.py
```diff
@@ -167,29 +167,26 @@ def contains(self, version: Version, resolve_prerelease: Optional[bool]):
...
+                if internal_conditions and (not lower_limit or not upper_limit or lower_limit <= upper_limit):
```
memsharded, line 205: ``if internal_conditions`` is always True

AbrilRBS: Oops, nice catch, leftover from previous iteration

conans/test/unittests/model/version/test_version_range_intersection.py
AbrilRBS: Because after the changes in #14814, the canonical representation of >=1.1 is >=1.1-, which is what the conditions now store and so what gets used in the calculations of the intersection

AbrilRBS: (Note that I updated the branch be in sync with release/2.0)

memsharded: Ok, on one hand it makes sense, it is expected, on the other hand it made me a bit uncomfortable for some reason, doesn't read very intuitive that the intersection apparently changes the limits, it should be kind of orthogonal. I have had a look, it is possible to change it, by storing the original version in the ``Condition`` object, but probably it is not worth it, so lets leave it as-is.

conans/model/version_range.py
AbrilRBS: I modified the logic to better reflect the actual behaviour, making the check now puroseful

conans/test/unittests/model/version/test_version_range_intersection.py
AbrilRBS: I think it might be worth pursuing the change? I share the feeling that it might be counterintuitive for users to see that the ranges are changing. Storing the original version expression seems doable to me, and tests seem to be passing locally. Will discuss later

RubenRBS:
Tests are failing on pourpose, we found a bug which has no current clear solution

czoido (approved):

AbrilRBS (approved):

PR #14923, opened 2023-10-13
Cache save and restore new functionality

### LINKED ISSUE #4316: Recommended mechanism to copy/archive a repository
I've been enjoying the repeatability that conan brings to building C/C++ libraries, and have run into a gap in my understanding of package/repository management. In YUM/RPM world, a YUM repository is a filesystem tree which can be tar/zip bundled and copied or archived, which is very convenient behavior.

In conan, when recipes/packages are cached locally they are distributed among user home directory ".conan" and short-path directory "C:\.conan" which makes it difficult to make a full copy of a local cache. Is there a canonical or recommended way to zip up a repository or local cache in a redistributable way?

I use Artifactory to manage local conan repositories and have considered to use the Artifactory export form for conan repository redistribution, but it seems like this ties the repository tightly to Artifactory which I would like to avoid if possible.

### LINKED ISSUE #10740: [question] Copy a package from machine to machine without internet
Hello,

I have two systems one which creates the conan packages (conan export-pkg) and one which needs to install them.
The second system is not connected to the internet at all, so I have to copy the package from one system to the other via a usb stick.
So if the offline system wants to install the package, it can do so without needing a internet connection, since the packages are already on the system.

My question is how do i copy a package from one device to the other, for example with a USB Stick.
Where do i find the package from the first machine to copy to my usb stick and how do i add this package to the offline machine so conan knows its there?

Thanks in advance!

### LINKED ISSUE #11687: [question] create a shippable deployment
Hello,
we store our conan recipes in our own JFrog artifactory, so that they can be consumed by our customers. 
Now we need to ship these packages to some colleages, which do not have access to our artifactory. 
Is there any deployable file format for this purpose?

I would expect something similar to `docker save`, which stores an container image as file. This file can then be shipped and imported on another PC via `docker load`.

Would it make sense to introduce some mechanism for this in conan 2.0?
```
conan save mylib/1.0.0@user/channel -o myfile.tar.gz # export the recipe only (export + exports_sources)
conan save mylib/1.0.0@user/channel -p <packageid1> -p <packageid2> -o myfile.tar.gz # + precompiled packageids
conan load myfile.tar.gz # load the conan package into local cache
conan load myfile.tar.gz user1/channel1 # combines "load" + "copy"
```

Background:
Currently we do not know, which packages or old RREVs of a package we can remove. A customer has maybe used it (with an explicit #RREV). If we remove the package now, we will break his (old) builds. Therefore we want to have a possibility to provide packages without "releasing" them into an artifactory. Even having a "development remote", this is not a solution for us. people may use the explicit package with #RREV.

This feature would also be helpful for CI system / pipelines: 

The CI does not need to upload any conan package to artifactory anymore, but instead it can create tar.gz artifacts.
Only in a last step (if all previous steps succeeded), the CI may upload the conan package to artifactory (in case of a release).
During development the package is not consumed via artifactory but via CI artifact (which the developer can download and load locally). As a result there is no need to upload all packages to artifactory anymore and waste a lot of space. 

Further Refs: #4316

### LINKED ISSUE #14548: [question] how to install from conan_package.tgz (not covered in docs)
### What is your question?

I have locally on my machine:

- conanmanifest.txt
- conaninfo.txt
- conan_package.tgz

(downloaded from package registry)

How I can install the package from these files locally into my conan environment?

memsharded:
Changelog: Feature: Implement package load/restore from the cache, for CI workflows and moving packages over air-gaps.
Docs: https://github.com/conan-io/docs/pull/3453

At the moment this is not intended for long-term storage of Conan packages, as the archive format and serialization might change in future Conan versions, but looking good.

Close https://github.com/conan-io/conan/issues/10740
Close https://github.com/conan-io/conan/issues/4316
Close https://github.com/conan-io/conan/issues/11687
Close https://github.com/conan-io/conan/issues/14548


conan/api/subapi/cache.py
```diff
@@ -101,6 +109,72 @@ def clean(self, package_list, source=True, build=True, download=True, temp=True)
...
+        cache = ClientCache(self.conan_api.cache_folder)
+        for ref, ref_bundle in package_list.refs().items():
+            ref.timestamp = revision_timestamp_now()
+            recipe_layout = cache.get_or_create_ref_layout(ref)
+            recipe_folder = ref_bundle["recipe_folder"]
+            rel_path = os.path.relpath(recipe_layout.base_folder, cache.cache_folder)
+            assert rel_path == recipe_folder, f"{rel_path}!={recipe_folder}"
+            out.info(f"Put: {ref} in {recipe_folder}")
+            for pref, pref_bundle in package_list.prefs(ref, ref_bundle).items():
+                pref.timestamp = revision_timestamp_now()
```
AbrilRBS, line 165: If I do `conan cache save zlib/1.3:*#*`, does this mean that it forgets which one is the "latest" prev?

AbrilRBS (commented):
Quick question by glancing at the code

conan/api/subapi/cache.py
memsharded: That is a very good point. I think we need to make sure  we are keeping the order, to make sure they are restored in the same order. I'll do it.

memsharded: Being based in package-lists, it keeps revisions in order, restoring them in order. I have added a test that validates the order for recipe-revision, which would be the most common use case, package-revisions should be way less common.

AbrilRBS (approved):

czoido (approved):

PR #14938, opened 2023-10-15
fix recreation of same package revision

### LINKED ISSUE #14933: `conan install` corrupts the cache and prevents other projects from compiling
### Environment details

* Operating System+version: Windows 11 22H2
* Conan version: 2.0.13
* Python version: 3.11.2

### Steps to reproduce

We are migrating from Conan v1 to v2 and encountered a bug. Here is a simple demonstration: [ConanCache.zip](https://github.com/conan-io/conan/files/12904230/ConanCache.zip).

conanfile.txt:
```
[requires]
fmt/10.1.1

[generators]
MSBuildDeps

[options]
fmt/*:header_only=True
```
First, we compile the dependencies for Project1, and Project1 works fine. But after we compile the dependencies for Project2, Project1 fails to compile because it cannot find <fmt/format.h>.

This error only occurs under specific configurations. If we remove the `header_only` option or use `--build=missing` in `conan install`, everything works fine.

### Logs

_No response_

memsharded:
Changelog: Fix: Solve re-build of an existing package revision (like forcing rebuild of a an existing header-only package), while previous folder was still used by other projects.
Docs: Omit

Close https://github.com/conan-io/conan/issues/14933

### DIFF
--- a/conan/internal/cache/cache.py
+++ b/conan/internal/cache/cache.py
@@ -1,5 +1,6 @@
 import hashlib
 import os
+import shutil
 import uuid
 
 from conan.internal.cache.conan_reference_layout import RecipeLayout, PackageLayout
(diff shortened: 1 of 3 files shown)


conan/internal/cache/cache.py
```diff
@@ -187,9 +188,16 @@ def assign_prev(self, layout: PackageLayout):
...
+            # There was a previous package folder for this same package reference (and prev)
             pkg_layout = self.get_package_layout(pref)
+            # We remove the old one and move the new one to the path of the previous one
+            # this can be necessary in case of new metadata or build-folder because of "build_id()"
             pkg_layout.remove()
+            shutil.move(layout.base_folder, pkg_layout.base_folder)  # clean unused temporary build
+            layout._base_folder = pkg_layout.base_folder  # reuse existing one
+            # TODO: The relpath would be the same as the previous one, it shouldn't be ncessary to
+            #  update it, the update_package_timestamp() can be simplified and path dropped
+            relpath = os.path.relpath(layout.base_folder, self._base_folder)
```
AbrilRBS, line 200: I have little insight into why this is needed again given the conext of the comment above, I'll trust you on this one but maybe I just need to look deeper into this bit of code functionality

AbrilRBS (approved):
I have 1 unknown but approving nontheless

czoido (approved):

franramirez688 (approved):

PR #14958, opened 2023-10-18
[Bazel][1.x][Part 1]: Bazel + BazelToolchain refactor

### LINKED ISSUE #14282: [feature] (Possibility to) change output path for generated Bazel BUILD files by BazelDeps

### LINKED ISSUE #14519: [bug] Possible Bazel toolchain deficiencies

franramirez688:
Changelog: Feature: `BazelToolchain` creates a `conan_bzl.rc` file which defines the `conan-config` configuration. If it exists, Bazel helper will use it automatically.
Docs: https://github.com/conan-io/docs/pull/3457
Closes: https://github.com/conan-io/conan/issues/14282 https://github.com/conan-io/conan/issues/14519

Next PR (BazelDeps): https://github.com/conan-io/conan/pull/14959

Overall, this PR is introducing no compatible changes, but deprecating some existing variables.

Summary:

* **BazelToolchain** is generating a `conan_bzl.rc` file with a `build:conan-config` configuration depending on your settings/options.
* **BazelToolchain** is not generating a real Bazel toolchain (https://bazel.build/extending/toolchains) because they require much more knowledge and expertise in Bazel to make a dynamic Bazel toolchain per user configuration.
* Bazel is moving forward to platforms (https://bazel.build/extending/platforms) for future releases instead of toolchains but it does not imply that BazelToolchain could be incompatible but some changes to inject those platforms constraints in our conan-config configuration.


conan/tools/google/bazel.py
```diff
@@ -1,44 +1,50 @@
...
-
-    def _get_bazel_project_configuration(self):
-        toolchain_file_content = load_toolchain_args(self._conanfile.generators_folder,
-                                                     namespace=self._namespace)
-        configs = toolchain_file_content.get("bazel_configs")
-        self._bazel_config = configs.split(",") if configs else []
-        self._bazelrc_path = toolchain_file_content.get("bazelrc_path")
+        # This is very important for Windows, as otherwise the bazel server locks files
+        if platform.system() == "Windows":
+            self._conanfile.run("bazel shutdown")
```
franramirez688: @memsharded these 3 lines are coming from our `release/2.0` branch. I just backported them. Do you remember why this was needed? Or is there any issue associated with that?

franramirez688: https://github.com/conan-io/conan/commit/0973aa2800c19f65cd6c26435fbe97140d29d2d2

franramirez688:
@czoido seems that bazel is not installed in macOS machines 😢 

The same test was already passing correctly for Conan 2.x (https://github.com/conan-io/conan/pull/14845) 

UPDATED:
@czoido Tests passing, solved! 👏

conans/assets/templates/new_v2_bazel.py
```diff
@@ -28,8 +28,7 @@ def layout(self):
 
     def build(self):
         bazel = Bazel(self)
-        bazel.configure()
-        bazel.build(label="//main:{name}")
+        bazel.build()
```
czoido, line 40: If this new template is not going to work with conan < 1.62, then I think we should add a `conan_min_required` to the template

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,50 @@
...
-        # )
-        command = 'bazel {} build {} {}'.format(
-            bazelrc_path,
-            bazel_config,
-            label
-        )
+    def __init__(self, conanfile):
+        self._conanfile = conanfile
 
+    def build(self, target="//...", cli_args=None):
```
czoido: To try to break as less as we can to people that already used the previous implementation maybe we should consider:
- Not changing `label` to `target` as `label` seems to  be aligned with Bazel terminology (https://bazel.build/concepts/labels).
- Not removing the `configure `method in the new implementation, just outputting a warning saying it's deprecated and that it will be removed in the future but at least it won't crash for users that update

I think it would be good that a package created with the `conan new` template from previous versions would work with the new implementation or at least need minimal changes. What do you think?

franramirez688: I know what you mean, but, IMO, the correct term for that is "target" as it shows the help message:
```
$ bazel help build
                                                  [bazel release 6.3.2-homebrew]
Usage: bazel build <options> <targets>

Builds the specified targets, using the options.

See 'bazel help target-syntax' for details and examples on how to
specify targets to build.
```
The most correct term would be `target_label` because it fully represents it and appears on the getting started most of the time.

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,50 @@
...
-
-    def _get_bazel_project_configuration(self):
-        toolchain_file_content = load_toolchain_args(self._conanfile.generators_folder,
-                                                     namespace=self._namespace)
-        configs = toolchain_file_content.get("bazel_configs")
-        self._bazel_config = configs.split(",") if configs else []
-        self._bazelrc_path = toolchain_file_content.get("bazelrc_path")
+        # This is very important for Windows, as otherwise the bazel server locks files
+        if platform.system() == "Windows":
+            self._conanfile.run("bazel shutdown")
```
danimtb: maybe add this in a try/finally clause to ensure it is always executed in case the buld command fails (it proved to be problematic in Windows if not executed)

franramirez688: absolutely! 😓

conan/tools/google/toolchain.py
```diff
@@ -1,24 +1,158 @@
...
+        self.dynamic_mode = "fully" if shared else "off"
+        self.cppstd = cppstd_flag(self._conanfile.settings)
+        self.copt = []
+        self.conlyopt = []
+        self.cxxopt = []
+        self.linkopt = []
+        self.strip = 'sometimes'  # 'always', 'sometimes', 'never'
+        self.compilation_mode = {'Release': 'opt', 'Debug': 'dbg'}.get(
+            self._conanfile.settings.get_safe("build_type"), "fastbuild"  # Bazel default
+        )
```
czoido: I think that we should not define `fastbuild` as the default, because it's already the Bazel default and it's not related to our settings, I would do: 

```
self.compilation_mode = {'Release': 'opt', 
                         'Debug': 'dbg'}.get(self.settings.build_type)
```

Then, if some some reason, Bazel deprecates `fastbuild` or changes the syntax we won't be affected.

conan/tools/google/toolchain.py
```diff
@@ -1,24 +1,158 @@
...
+        self.force_pic = fpic if (not shared and fpic is not None) else None
+        # FIXME: Keeping this option but it's not working as expected. It's not creating the shared
+        #        libraries at all.
+        self.dynamic_mode = "fully" if shared else "off"
+        self.cppstd = cppstd_flag(self._conanfile.settings)
+        self.copt = []
+        self.conlyopt = []
+        self.cxxopt = []
+        self.linkopt = []
+        self.strip = 'sometimes'  # 'always', 'sometimes', 'never'
```
czoido: Do we need to define this? (it's the default for Bazel). Maybe we want to set this to `None` or remove it and let users define it via a cli argument in `build()`?

danimtb (approved):

czoido:
I think there's a problem that we should resolve on how Bazel caches the artifacts and Conan packages it after the build.
The problem is that if for example, you create a bazel package and the library is packaged ok, because the artifacts are taken from the Bazel cache to the Conan cache, but then you make some modifications to the package, and change the name of the library for another one, then the old and new library are taken from the Bazel cache and two libs are packaged. To reproduce:

```
conan new mypackage/1.0 -m bazel_lib
conan create . -> libmypackage.a is stored in Conan cache
# replace name = "mypackage", in the BUILD file for name = "mypackagewithothername",
conan create . -> libmypackage.a and libmypackagewithothername.a are both stored in Conan cache
```

conans/assets/templates/new_v2_bazel.py
```diff
@@ -23,13 +24,18 @@ def config_options(self):
...
+                                 "know more about it: https://bazel.build/reference/be/c-cpp")
+
     def layout(self):
         bazel_layout(self)
 
     def build(self):
         bazel = Bazel(self)
-        bazel.configure()
-        bazel.build(label="//main:{name}")
+        bazel.build()
```
czoido, line 40: Would it be possible in the template to create different targets and do something like:

```
        bazel = Bazel(self)
        target_suffix = "_shared" if self.options.shared else ""
        bazel.build(target=f"//main:mypackage{target_suffix}")
```

to enable creating both shared an static libraries?

conans/assets/templates/new_v2_bazel.py
```diff
@@ -23,13 +24,18 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
+    def validate(self):
+        if self.settings.os in ("Windows", "Macos") and self.options.shared:
+            raise ConanException("Windows and Macos needs extra BUILD configuration to be able "
+                                 "to create a shared library. Please, check this reference to "
+                                 "know more about it: https://bazel.build/reference/be/c-cpp")
+
```
franramirez688, line 32: The next PR solves this issue. That one introduces changes in the template to let users create/consume shared libraries in all the platforms.

conans/assets/templates/new_v2_bazel.py
franramirez688: See https://github.com/conan-io/conan/pull/14958/files#r1374405258

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,70 @@
...
+
+from conan.tools.google import BazelToolchain
 
 
 class Bazel(object):
-    def __init__(self, conanfile, namespace=None):
+
+    def __init__(self, conanfile):
         self._conanfile = conanfile
-        self._namespace = namespace
```
memsharded, line 7: Removing ``namespace`` can break, but lets wait to see what happens, hopefully was not used.

conan/tools/google/layout.py
```diff
@@ -1,11 +1,13 @@
...
-    conanfile.cpp.build.libdirs = ["bazel-bin"]
+def bazel_layout(conanfile, src_folder="."):
+    """Bazel layout is so limited. It does not allow to create its special symlinks in other
+    folder. See more information in https://bazel.build/remote/output-directories"""
+    subproject = conanfile.folders.subproject
+    conanfile.folders.source = src_folder if not subproject else os.path.join(subproject, src_folder)
+    conanfile.folders.build = "."  # Bazel always build the whole project in the root folder
+    conanfile.folders.generators = "conan"  # one
+    # FIXME: used in test package for example, to know where the binaries are (editables not supported yet)?
+    conanfile.cpp.build.bindirs = [os.path.join(conanfile.folders.build, "bazel-bin")]
```
memsharded, line 15: Not sure how this changes something if ``build="."``, then this should be the same?

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,70 @@
...
+        conan_bazelrc = os.path.join(self._conanfile.generators_folder, BazelToolchain.bazelrc_name)
+        use_conan_config = os.path.exists(conan_bazelrc)
+        bazelrc_paths = []
+        bazelrc_configs = []
+        if use_conan_config:
+            bazelrc_paths.append(conan_bazelrc)
+            bazelrc_configs.append(BazelToolchain.bazelrc_config)
+        # User bazelrc paths have more prio than Conan one
+        # See more info in https://bazel.build/run/bazelrc
+        bazelrc_paths.extend(self._conanfile.conf.get("tools.google.bazel:bazelrc_path",
```
memsharded: Moving these configs to a list (mandating to be a list), is also breaking. I'd allow them being single values too, and converting to list internally, to avoid breaking.

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,70 @@
...
+                                                      default=[], check_type=list))
+        command = "bazel"
+        for rc in bazelrc_paths:
+            command += f" --bazelrc={rc}"
+        command += " build"
+        bazelrc_configs.extend(self._conanfile.conf.get("tools.google.bazel:configs",
+                                                        default=[], check_type=list))
```
memsharded: Same about conf being a list.

conan/tools/google/layout.py
```diff
@@ -1,11 +1,13 @@
...
-    # used in test package for example, to know where the binaries are (editables not supported yet)
-    conanfile.cpp.build.bindirs = ["bazel-bin"]
-    conanfile.cpp.build.libdirs = ["bazel-bin"]
+def bazel_layout(conanfile, src_folder="."):
+    """Bazel layout is so limited. It does not allow to create its special symlinks in other
+    folder. See more information in https://bazel.build/remote/output-directories"""
+    subproject = conanfile.folders.subproject
+    conanfile.folders.source = src_folder if not subproject else os.path.join(subproject, src_folder)
+    conanfile.folders.build = "."  # Bazel always build the whole project in the root folder
+    conanfile.folders.generators = "conan"  # one
```
memsharded: What is the downside of making this opt-in to not break? I am fine if this is the only thing breaking, but if there are other things, then I'd reconsider.

conans/test/functional/toolchains/google/test_bazel.py
```diff
@@ -26,15 +26,16 @@ def base_profile():
         build_type={build_type}
 
         [conf]
-        tools.google.bazel:bazelrc_path={curdir}/mybazelrc
+        tools.google.bazel:bazelrc_path=["{curdir}/mybazelrc"]
```
memsharded: Breaking change to try to avoid.

memsharded (commented):
Overall looking good, but my concerns about breaking too much, and I'd try to minimize it.

conan/tools/google/layout.py
franramirez688: I changed that in my next PR because of the editable mode, so let me keep it here to avoid more conflicts

memsharded (approved):

conan/tools/google/toolchain.py
```diff
@@ -1,24 +1,160 @@
...
 
     def __init__(self, conanfile, namespace=None):
         self._conanfile = conanfile
-        self._namespace = namespace
+        # TODO: Remove namespace and check_using_build_profile in Conan 2.x
+        if namespace:
+            self._conanfile.output.warn("In BazelToolchain() call, namespace param has been "
+                                        "deprecated as it's not used anymore. Use the bazel_layout()"
+                                        " and the 'tools.google.bazel_layout:generators_folder=xxxx' configuration"
+                                        " to specify another folder for the Conan-generated-files.")
```
czoido:
```suggestion
        if namespace:
            self._conanfile.output.warn("In BazelToolchain() call, namespace param has been "
                                        "deprecated as it's not used anymore.")
```
I would removed that because the conf does not replace the namespace...

conan/tools/google/bazel.py
```diff
@@ -1,44 +1,86 @@
...
     def __init__(self, conanfile, namespace=None):
         self._conanfile = conanfile
-        self._namespace = namespace
-        self._get_bazel_project_configuration()
+        # TODO: Remove namespace in Conan 2.x
+        if namespace:
+            self._conanfile.output.warn("In Bazel() call, namespace param has been "
+                                        "deprecated as it's not used anymore. Use the bazel_layout()"
+                                        " and the 'tools.google.bazel_layout:generators_folder=xxxx' configuration"
+                                        " to specify another folder for the Conan-generated-files.")
```
czoido: Same as the comment for the toolchain

conan/tools/google/layout.py
```diff
@@ -1,11 +1,16 @@
...
-    conanfile.cpp.build.libdirs = ["bazel-bin"]
+def bazel_layout(conanfile, src_folder="."):
+    """Bazel layout is so limited. It does not allow to create its special symlinks in other
+    folder. See more information in https://bazel.build/remote/output-directories"""
+    subproject = conanfile.folders.subproject
+    conanfile.folders.source = src_folder if not subproject else os.path.join(subproject, src_folder)
+    conanfile.folders.build = "."  # Bazel always build the whole project in the root folder
+    # FIXME: Keeping backward-compatibility. Defaulting to "conan" in Conan 2.x.
+    conanfile.output.warning("In bazel_layout() call, generators folder changes its default value "
+                             "from './' to './conan/' in Conan 2.x")
```
czoido, line 12: I think we can make here a explicit recomendation of setting `self.folders.generators_folder = "conan"` so that users can have a clue on what to do, but we can change that in the next PR

czoido (approved):

PR #14966, opened 2023-10-18
Add extra flags to CMakeToolchain

czoido:
Changelog: Feature: Add extra flags to CMakeToolchain.
Docs: https://github.com/conan-io/docs/pull/3452

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -567,8 +567,8 @@ class ExtraFlagsBlock(Block):
         # Now, it's time to get all the flags defined by the user
         cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list) + self._toolchain.extra_cxxflags
         cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list) + self._toolchain.extra_cflags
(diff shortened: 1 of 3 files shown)


AbrilRBS (approved):

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -565,10 +565,10 @@ class ExtraFlagsBlock(Block):
 
     def context(self):
         # Now, it's time to get all the flags defined by the user
-        cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list)
-        cflags = self._conanfile.conf.get("tools.build:cflags", default=[], check_type=list)
-        sharedlinkflags = self._conanfile.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)
-        exelinkflags = self._conanfile.conf.get("tools.build:exelinkflags", default=[], check_type=list)
+        cxxflags = self._conanfile.conf.get("tools.build:cxxflags", default=[], check_type=list) + self._toolchain.extra_cxxflags
```
memsharded: cxxflags have higher priority if they are appended. I think that the profile flags should have priority over the recipe flags if there is conflict, so we might want to reverse this order, what do you think?

czoido: Yes, totally makes sense, I'll change the preference order

memsharded (approved):

PR #14980, opened 2023-10-19
smart cache path

### LINKED ISSUE #14828: [feature] Conan cache could be more helpful about resolving packages from short references
### What is your suggestion?

Very often, I'm interested in finding information about a package, to which I only have a "short" reference. I.e:
```
package_name/version@company
```

In these cases, Conan will either just give me information about the recipe, or say that it needs to know the package ID.

However, there is only one single package in existence which would match this particular reference, and Conan probably knows it. Why not have Conan then resolve this for me, instead of me having to jump through hoops to find the hash of the package_id?

I understand that if there is more than one package, Conan can't resolve it, but even in that case it would be _much more_ helpful if it listed which packages it found.

For example:
```
conan cache path package_name/version@company --folder=build
ERROR: '--folder build' requires a valid package reference
```

Could instead be:
```
conan cache path package_name/version@company --folder=build
/home/user/.conan2/p/packdf916cc80f201/b
```

Or, if more than one package_id matches this reference, in my cache:
```
conan cache path package_name/version@company --folder=build
ERROR: More than one package matches this reference. Need a package_id:
package_name/version@company#recipe_hash:package_id1
package_name/version@company#recipe_hash:package_id2
package_name/version@company#recipe_hash:package_id3
```

If this were implemented, I would also love to be able to specify ```--folder=package```. Then Conan would understand that I want the package path, and not the recipe path:
```
conan cache path package_name/version@company --folder=package
/home/user/.conan2/p/packdf916cc80f201/p
```

memsharded:
Changelog: Feature: Smart ``conan cache path`` command that is able to deduce the ``package_id`` if it is a single one, and to retrieve recipe, export and source folders when providing a full package-reference
Docs: omit

Just a proof of concept, to discuss, not guaranteed that it will be moved forward.

Close https://github.com/conan-io/conan/issues/14828

### DIFF
--- a/conan/api/subapi/cache.py
+++ b/conan/api/subapi/cache.py
@@ -37,6 +37,16 @@ class CacheAPI:
         ref_layout = app.cache.recipe_layout(ref)
         return ref_layout.source()
 
+    def unique_binary(self, ref:RecipeReference):
+        app = ConanApp(self.conan_api.cache_folder)
+        ref = _resolve_latest_ref(app, ref)
+        prefs = app.cache.get_package_references(ref, only_latest_prev=True)
+        if len(prefs) != 1:
+            raise ConanException(f"There are {len(prefs)} binaries for {ref}. "
+                                 f"This only works when there is exactly 1 binary. "
+                                 "Please specify the package_id too.")
+        return prefs[0]
+
     def build_path(self, pref: PkgReference):
         app = ConanApp(self.conan_api.cache_folder)
         pref = _resolve_latest_pref(app, pref)
--- a/conan/cli/commands/cache.py
+++ b/conan/cli/commands/cache.py
@@ -21,38 +21,44 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
     Show the path to the Conan cache for a given reference.
     """
     subparser.add_argument("reference", help="Recipe reference or Package reference")
-    subparser.add_argument("--folder", choices=['export_source', 'source', 'build', 'metadata'],
+    subparser.add_argument("--folder", choices=['recipe', 'export_source', 'source', 'build',
+                                                'metadata', 'package', 'recipe_metadata',
+                                                'package_metadata'],
                            help="Path to show. The 'build' requires a package reference. "
                                 "If the argument is not passed, it shows 'exports' path for recipe references "
                                 "and 'package' folder for package references.")
 
     args = parser.parse_args(*args)
+    pref = ref = None
     try:
         pref = PkgReference.loads(args.reference)
     except ConanException:
-        pref = None
-
-    if not pref:  # Not a package reference
         ref = RecipeReference.loads(args.reference)
-        if args.folder is None:
-            path = conan_api.cache.export_path(ref)
-        elif args.folder == "export_source":
-            path = conan_api.cache.export_source_path(ref)
-        elif args.folder == "source":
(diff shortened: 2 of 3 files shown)


AbrilRBS (approved):
I like the overall idea and know that some users have requested it - I was a bit worried that now users would expect similar behaviours in other commands, but thinking about it, there's not any other command that could work similarly right?

memsharded:
> I like the overall idea and know that some users have requested it - I was a bit worried that now users would expect similar behaviours in other commands, but thinking about it, there's not any other command that could work similarly right?

There are many commands that can receive different inputs like refs, refs-with-recipe-revision, prefs, etc, with different patterns, all the ``list, upload, download, remove``. I think they already define their own logic and semantics to define what they want to list, upload, download or remove, so I think yes, this could be restricted to the ``conan cache path`` command only.

memsharded:
To be re-considered after https://github.com/conan-io/conan/pull/15011

smoofra:
I use this plugin command for a similar purpose.  It takes the `-pr:h` argument, so you can specify what profile you want it for instead of searching for the package_id first.    I not sure if it's implemented in an entirely sound fashion, but I do find it pretty useful.


```
from conan.cli.args import add_profiles_args
from conan.cli.command import conan_command
from conan.cli.printers import print_profiles
from conan.cli.printers.graph import print_graph_packages, print_graph_basic
import conan.api.output

verbose = False


@conan_command(group="Consumer")
def find(conan_api, parser, *raw_args):
    """
    Find the path to a package in the local cache
    """
    add_profiles_args(parser)
    parser.add_argument("requires", nargs="+")
    parser.add_argument("--export", action="store_true")
    parser.add_argument("--export-source", action="store_true")
    parser.add_argument("--source", action="store_true")
    parser.add_argument(
        "-r",
        "--remote",
        action="append",
        default=None,
        help="Look in the specified remote or remotes server",
    )
    parser.add_argument(
        "-nr",
        "--no-remote",
        action="store_true",
        help="Do not use remote, resolve exclusively in the cache",
    )
    args = parser.parse_args(*raw_args)

    conan.api.output.ConanOutput.define_log_level("warning")

    remotes = conan_api.remotes.list(args.remote) if not args.no_remote else []

    profile_host, profile_build = conan_api.profiles.get_profiles_from_args(args)
    if verbose:
        print_profiles(profile_host, profile_build)

    deps_graph = conan_api.graph.load_graph_requires(
        args.requires, [], profile_host, profile_build, None, remotes, False
    )

    if verbose:
        print_graph_basic(deps_graph)
    deps_graph.report_graph_error()

    conan_api.graph.analyze_binaries(deps_graph, remotes=remotes)

    if verbose:

    for p in deps_graph.root.neighbors():
        if args.source:
            print(conan_api.cache.source_path(p.ref))
        elif args.export:
            print(conan_api.cache.export_path(p.ref))
        elif args.export_source:
            print(conan_api.cache.export_source_path(p.ref))
        else:
            print(conan_api.cache.package_path(p.pref))

```

memsharded:
Let's go at the moment with the new ``conan list *:* --format=compact`` it seems convenient enough. This proposed PR is still a bit too much of an edge case, that it happens to have just 1 binary, not worth the complexity. 
We can always reconsider again in the future, and there are always user commands, but at this moment we think the current tools should be enough. Thanks for all the feedback!

PR #14986, opened 2023-10-20
Add config passing to conan profile detect

AbrilRBS:
Changelog: Feature: Add ability to pass configs to override results of `conan profile detect`
Docs: TODO

Besides the `multiple_contexts` change for the profile arg helper, my main concern is that this would be the first command to accept `-pr` etc, but not context specifiers, as they don't make any sense in this context

### DIFF
--- a/conan/api/subapi/profiles.py
+++ b/conan/api/subapi/profiles.py
@@ -91,8 +91,7 @@ class ProfilesAPI:
 
     def _get_profile(self, profiles, settings, options, conf, cwd, cache_settings,
                      profile_plugin, global_conf):
-        loader = ProfileLoader(self._conan_api.cache_folder)
-        profile = loader.from_cli_args(profiles, settings, options, conf, cwd)
+        profile = self.load_profile(profiles, settings, options, conf, cwd)
         if profile_plugin is not None:
             try:
                 profile_plugin(profile)
@@ -105,6 +104,10 @@ class ProfilesAPI:
         # Apply the new_config to the profiles the global one, so recipes get it too
         profile.conf.rebase_conf_definition(global_conf)
         return profile
+
+    def load_profile(self, profiles, settings, options, conf, cwd):
+        loader = ProfileLoader(self._conan_api.cache_folder)
+        return loader.from_cli_args(profiles, settings, options, conf, cwd)
 
     def get_path(self, profile, cwd=None, exists=True):
         """
--- a/conans/test/functional/command/profile_test.py
+++ b/conans/test/functional/command/profile_test.py
@@ -3,8 +3,6 @@ import os
 import platform
 import subprocess
 import textwrap
-import unittest
-
 import pytest
 
 from conans.client.conf.detect import detect_defaults_settings
--- a/conans/test/functional/command/profile_test.py
+++ b/conans/test/functional/command/profile_test.py
@@ -1,3 +1,4 @@
+import json
 import os
 import platform
 import subprocess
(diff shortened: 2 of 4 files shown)


conans/test/functional/command/profile_test.py
```diff
@@ -14,140 +13,152 @@
...
+
+    client.save({"profile1": profile1})
+    client.run("profile show -pr=profile1")
+    assert "[settings]\nos=Windows" in client.out
+    assert "MyOption=32" in client.out
+    assert "tools.build:jobs=20" in client.out
+
+    client.run("profile show -pr=profile1 -pr:b=profile1 -f=json", redirect_stdout="profile1.json")
+    profile = json.loads(client.load("profile1.json"))
+    assert profile == {'host': {'settings': {'os': 'Windows'}, 'package_settings': {},
```
AbrilRBS, line 73: Addition to show that profile show properly works with its -f formatter, just in case

conans/test/functional/command/profile_test.py
```diff
@@ -14,140 +13,152 @@
...
+
+    c.run("profile detect --name=./MyProfile2", assert_error=True)
+    assert "MyProfile2' already exists" in c.out
+
+    c.run("profile detect --name=./MyProfile2 --force")  # will not raise error
+
+    c.run("profile detect --name=./MyProfile3 -o *:my_custom_option=42 -s compiler.cppstd=gnu98")
+    assert "*:my_custom_option=42" in c.out
+    assert "compiler.cppstd=gnu98" in c.out  # Override from default detected one
+
```
AbrilRBS, line 141: These is the only new test in the file, the rest is moving away from unittest

conans/test/functional/command/profile_test.py
```diff
@@ -14,140 +13,152 @@
...
-        # See: https://github.com/conan-io/conan/issues/2231
-        output = check_output_runner("gcc --version", stderr=subprocess.STDOUT)
-
-        if "clang" not in output:
-            # Not test scenario gcc should display clang in output
-            # see: https://stackoverflow.com/questions/19535422/os-x-10-9-gcc-links-to-clang
-            raise Exception("Apple gcc doesn't point to clang with gcc frontend anymore!")
-
-        output = RedirectedTestOutput()  # Initialize each command
+
```
AbrilRBS, line 16: Moved the file away from unittest - Maybe I should have left the classes be?

AbrilRBS (commented):
Some comments to clarify the test changes

conan/api/subapi/profiles.py
```diff
@@ -106,6 +105,10 @@ def _get_profile(self, profiles, settings, options, conf, cwd, cache_settings,
         profile.conf.rebase_conf_definition(global_conf)
         return profile
 
+    def load_profile(self, profiles, settings, options, conf, cwd):
```
memsharded, line 108: This could be a bit confusing, we already have the ``get_profile()`` with same arguments. It seems the difference is the application of profile_plugin, isn't it? I think we want to discuss this (future) public API

conans/test/functional/command/profile_test.py
memsharded: A bit difficult to see the relevant changes that could be caused by the new feature, but not a big problem

memsharded (commented):
``profile detect`` having profiles as arguments read too much for me. settings, options and conf, maybe, but maybe still too much. 

What is the use case? I can understand fine-tuning some of the detection with some ``setting``, in some circumstances, but being able to fully hack the ``conan profile detect`` is kind of bad? Why would you like to hardcode so much information just in your CI job, instead of having it in a profile file?

AbrilRBS:
Let's get back t the drawing board with this approach, as discussed, `conan profile show` can work as a replacement with less work

PR #15004, opened 2023-10-23
editable shouldn't download sources

### LINKED ISSUE #15003: [bug] conan graph info with tools.build:download_source fails for editable packages
### Environment details

* Operating System+version: Ubuntu 20.04.6 LTS
* Compiler+version: not applicable
* Conan version: 2.0.13
* Python version: 3.8.10

### Steps to reproduce

1. Set up a conan project A
2. Set up a conan project B, with a dependency to A
3. Add A as an editable package using `conan editable add <path/to/A>`
4. Make sure `conan install ...` works as expected in project B, using A as an editable package
5. `conan graph info <path/to/B> --conf tools.build:download_source=True` --> Should succeed but fails

### Logs

`ERROR: Recipe '<recipe-of-A>' not found`

memsharded:
Changelog: Bugfix: ``tools.build:download_source=True`` will not fail when there are editable packages.
Docs: https://github.com/conan-io/docs/pull/3448

Close https://github.com/conan-io/conan/issues/15003

Note I have opted for not trying to do the download. The download can be a destructive action, removing local changes, which defeats the purpose of editables. ``conan source`` must be manually called by users on the editable explicitly, but it will not be called by the consumers ``install`` even if ``tools.build:download_source=True``.

### DIFF
--- a/conans/client/installer.py
+++ b/conans/client/installer.py
@@ -182,6 +182,9 @@ class BinaryInstaller:
             return
 
         conanfile = node.conanfile
+        if node.binary == BINARY_EDITABLE:
+            return
+
         recipe_layout = self._cache.recipe_layout(node.ref)
         export_source_folder = recipe_layout.export_sources()
         source_folder = recipe_layout.source()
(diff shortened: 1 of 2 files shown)


AbrilRBS:
I don't think we should omit docs on this one, maybe a note to clarify this behavaiour would be nice? It might be suprising behaviour for people that run this if they expected all sources to be fetched.

AbrilRBS (approved):

franramirez688 (approved):

PR #15013, opened 2023-10-25
Add `conan cache backup-upload`

### LINKED ISSUE #15000: [question] backup third-party source not done if the package already exists
### What is your question?

Hi,

I saw the docs for backing up third party sources: https://docs.conan.io/2/devops/backup_sources/sources_backup.html
Now the problem I have is that if the package already exists on artifactory but the sources are not backed up yet und you add the stuff mentioned in the docs, this is not detected and the backup is skipped.

Is this intended somehow? How can I backup the sources now for already created packages?

AbrilRBS:
Changelog: Feature: Add `conan cache backup-upload` to upload all the backup sources in the cache, regardless of which references they are from
Docs: https://github.com/conan-io/docs/pull/3438

Took me a bit to find the better way to cleanly separate the backup file recolection api, but I like how simple it looks in the end.

Not sold on the api names nor command names, but those are minor details that can be changed before merging, the PR is more to see if we like this approach

Note that I tried to add an option akin to the `--metdata` flag in `conan upload`, but quickly found that it didn't feel a nice fit, so turned away from it

### DIFF
--- a/conan/api/subapi/upload.py
+++ b/conan/api/subapi/upload.py
@@ -57,22 +57,38 @@ class UploadAPI:
         executor = UploadExecutor(app)
         executor.upload(package_list, remote)
 
-    def upload_backup_sources(self, package_list):
+    def get_all_backup_sources(self):
         app = ConanApp(self.conan_api.cache_folder)
         config = app.cache.new_config
-        url = config.get("core.sources:upload_url")
+
+        download_cache_path = config.get("core.sources:download_cache", check_type=str)
+        download_cache_path = download_cache_path or HomePaths(
+            self.conan_api.cache_folder).default_sources_backup_folder
+        excluded_urls = config.get("core.sources:exclude_urls", check_type=list, default=[])
+        files = DownloadCache(download_cache_path).get_all_backup_sources_files(excluded_urls)
+        return files
+
+    def get_associated_backup_sources(self, package_list):
+        app = ConanApp(self.conan_api.cache_folder)
+        config = app.cache.new_config
+        download_cache_path = config.get("core.sources:download_cache")
+        download_cache_path = download_cache_path or HomePaths(
+            self.conan_api.cache_folder).default_sources_backup_folder
+        excluded_urls = config.get("core.sources:exclude_urls", check_type=list, default=[])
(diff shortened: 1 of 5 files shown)


conan/api/subapi/upload.py
```diff
@@ -57,22 +57,38 @@ def upload(self, package_list, remote):
...
-        url = config.get("core.sources:upload_url")
+
+        download_cache_path = config.get("core.sources:download_cache", check_type=str)
+        download_cache_path = download_cache_path or HomePaths(
+            self.conan_api.cache_folder).default_sources_backup_folder
+        excluded_urls = config.get("core.sources:exclude_urls", check_type=list, default=[])
+        files = DownloadCache(download_cache_path).get_all_backup_sources_files(excluded_urls)
+        return files
+
+    def get_associated_backup_sources(self, package_list):
```
memsharded: The implementation of this method that is almost identical to the above one, together with the "qualifier" ``all`` or ``associated`` in the name, makes me think that it could be one method that receives an optional ``package_list`` filtering argument. If the package-list is None, then picks everything.

conans/client/downloaders/download_cache.py
```diff
@@ -44,42 +44,66 @@ def lock(self, lock_id):
...
+                       for url in urls)
+
+        result = []
+        for path in os.listdir(path_backups):
+            path = os.path.join(path_backups, path)
+            if path.endswith(".json"):
+                f = os.path.join(path)
+                content = json.loads(load(f))
+                refs = content["references"]
+                # unknown entries are not uploaded at this moment, the flow is not expected.
```
memsharded: This comment is no longer true for this case?

memsharded (commented):
I like it! it is indeed clean and nice.

Just a minor comment about the public API, if we wanted to avoid splitting it into 2, but not really opposed to current proposal.

conan/api/subapi/upload.py
AbrilRBS: Ahg, I knew I was missing something obvious! It looks way better now, thanks for the tip

memsharded (approved):
Amazing!

czoido (approved):

PR #15015, opened 2023-10-26
Add new Apple OS versions to settings.yml

valgur:
Changelog: Fix: Add new Apple OS versions to settings.yml
Docs: Omit

Adds new Apple OS and SDK versions that were released in the past few weeks.

`macos-latest` currently fails in Github due to a missing `settings.os.version` value in the default settings.yml because of this: https://github.com/valgur/novatel_edie/actions/runs/6650201881/job/18069930032#step:7:65

I updated the lists based on:
* https://en.wikipedia.org/wiki/MacOS_version_history#Releases
* https://en.wikipedia.org/wiki/TvOS#tvOS_17
* https://en.wikipedia.org/wiki/WatchOS#watchOS_10
* https://en.wikipedia.org/wiki/IOS_version_history#iOS_17_/_iPadOS_17
* https://en.wikipedia.org/wiki/Xcode#Toolchain_versions (I did not add 15.1, since the list seems to only include x.0 versions)


memsharded (approved):

memsharded (approved):
Thanks very much for your contribution, and also for providing the links with the info about the versions and releases, that helps!

PR #15023, opened 2023-10-28
Check commits even with shallow clone of tags

### LINKED ISSUE #14928: [bug] Find Commit in Remote on Shallow Clone
### Environment details

* Compiler+version: gcc11
* Conan version: 2.0.13
* Python version: 3.10.6

### Steps to reproduce

The git helper is able to add the repo url and commit to our conandata.yml file.  It uses `git branch -r --contains {commit_hash}`.  My CI system does a shallow clone of the repo in certain cases and the `git branch` command can't find the commit.  It is easier to recreate outside of conan rather than inside a recipe.

1. git clone --depth=1 --branch=2.0.13 https://github.com/conan-io/conan.git
2. git log # Shows that the commit hash is 8cb007e15865d06d999658d72f396a69400aefcb
3. git branch -r --contains 8cb007e15865d06d999658d72f396a69400aefcb

Note that the `git branch` doesn't find the commit.  If you repeat the steps without the `depth=1` then git is able to find the commit.

### Logs

Inside of the `conan create` when I call `Git(self).get_url_and_commit()` I get the following warning.
WARN: Current commit f033eaf9380ec2177257d73ea9b4b8625f1b81d1 doesn't exist in remote origin
This revision will not be buildable in other computer

That commit does exist in the remote url.

### Solution
The current code in `conan/tools/git/` uses `git branch -r --contains {hash}`
I found that instead running: `git fetch origin --dry-run --depth=1 {hash}` works even if the commit is shallow.    https://stackoverflow.com/questions/65611995/how-to-check-whether-a-git-commit-exists-on-some-remote

I'm happy to submit a pull request for this.

jsallay:
Changelog: Fix: Make ``Git()`` check commits in remote server even for shallow clones.
Docs: omit

Fixes https://github.com/conan-io/conan/issues/14928

When performing a shallow clone of a tag (e.g. `git clone https://github.com/conan-io/conan --depth=1 -b 2.0.13`) it does not copy and of the branches or history.  The `commit_in_remote` method of the `Git` class only checks local information.  This PR will additionally check the remote if needed.

I tested this by cloning a repo twice - one as a normal clone and the other with a shallow clone of a tag.  I created an additional commit in each repo.  I ran of total of 8 tests - checking all possibilities of:
1. Shallow vs Full clone
2. Commit in repo
3. Network connection available.

The `git fetch` performed requires a connection to the remote repo.  I wanted to make sure that it failed gracefully in an offline scenario.  The function performed as expected in all cases.

Without an internet connection, the function waits until the command times out and then throws an error (which is communicated to the user).

### DIFF
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -91,15 +91,11 @@ class Git:
 
         try:
             # This will raise if commit not present.
-            self.run("fetch origin --dry-run --depth=1 {}".format(commit))
+            self.run("fetch {} --dry-run --depth=1 {}".format(remote, commit))
             return True
         except Exception as e:
-            # The error message contains the command and the commit if it can't find it.
-            if str(e).count(commit) == 2:
-                return False
-            else:
-                # Perhaps a connection error.
-                raise ConanException("Unable to check remote commit in '%s': %s" % (self.folder, str(e)))
+            # Don't raise an error because the fetch could fail for many more reasons than the branch.
+            return False
 
     def is_dirty(self):
         """
--- a/conan/tools/scm/git.py
+++ b/conan/tools/scm/git.py
@@ -80,11 +80,26 @@ class Git:
         """
         if not remote:
             return False
+        # Potentially do two checks here.  If the clone is a shallow clone, then we won't be
+        # able to find the commit.
         try:
             branches = self.run("branch -r --contains {}".format(commit))
-            return "{}/".format(remote) in branches
+            if "{}/".format(remote) in branches:
+                return True
         except Exception as e:
             raise ConanException("Unable to check remote commit in '%s': %s" % (self.folder, str(e)))
+
+        try:
+            # This will raise if commit not present.
+            self.run("fetch origin --dry-run --depth=1 {}".format(commit))
+            return True
+        except Exception as e:
+            # The error message contains the command and the commit if it can't find it.
+            if str(e).count(commit) == 2:
+                return False
+            else:
+                # Perhaps a connection error.
+                raise ConanException("Unable to check remote commit in '%s': %s" % (self.folder, str(e)))
 
     def is_dirty(self):
         """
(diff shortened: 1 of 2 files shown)


conan/tools/scm/git.py
```diff
@@ -80,12 +80,27 @@ def commit_in_remote(self, commit, remote="origin"):
...
         except Exception as e:
             raise ConanException("Unable to check remote commit in '%s': %s" % (self.folder, str(e)))
 
+        try:
+            # This will raise if commit not present.
+            self.run("fetch origin --dry-run --depth=1 {}".format(commit))
+            return True
+        except Exception as e:
+            # The error message contains the command and the commit if it can't find it.
+            if str(e).count(commit) == 2:
```
memsharded: It seems this check is not capturing all scenarios.

memsharded (commented):
Thanks for your contribution @jsallay !

Some CI tests are failing, the new check is producing unexpected errors, it probably needs to be more conservative and raise only on very evident cases.

memsharded (commented):
This looks good, and tests are passing, that is great.

It would be great to add a test that validates the new code is working fine. Do you think you could try to add it? Let us know if you need help or guidance.

conans/test/functional/tools/scm/test_git.py
```diff
@@ -889,3 +889,76 @@ def set_version(self):
...
+                in_remote = git.commit_in_remote(commit)
+                self.output.info("COMMIT IN REMOTE: {}".format(in_remote))
+                self.output.info("DIRTY: {}".format(git.is_dirty()))
+        """)
+
+    def test_find_tag_in_remote(self):
+        """
+        a shallow cloned repo won't have the new commit locally, but can fetch it.
+        """
+        return
```
memsharded: This return is skipping the test
```suggestion
```

jsallay:
I've added in tests for a case where it should find the commit and one where it should not.  Let me know if you have any other suggestions.

memsharded:
> I've added in tests for a case where it should find the commit and one where it should not. Let me know if you have any other suggestions.

It is great. I'll merge when the PR is green

PR #15026, opened 2023-10-29
package_manager: handle linux mint

RMZeroFour:
Changelog: Fix: Default to apt-get package manager in Linux Mint
Docs: https://github.com/conan-io/docs/pull/3441

Linux Mint, being based on Ubuntu, should default to using the apt-get package manager. Being a decently popular OS for Linux beginners, ideally Conan should work on it with as little configuration from the user as possible, and the tool name is one such configuration that can be preset.

Right now it fails to install system packages even with sudo=True and mode=install in the global.conf file, as it doesn't know which package manager to use.

### DIFF
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -61,6 +61,11 @@ class _SystemPackageManagerTool(object):
             for d in distros:
                 if d in os_name:
                     return tool
+                
+        # No default package manager was found for the system,
+        # so notify the user
+        self._conanfile.output.info("A default system package manager couldn't be found for {}, "
+                                    "system packages will not be installed.".format(os_name))
 
     def get_package_name(self, package, host_package=True):
         # Only if the package is for building, for example a library,
--- a/conan/tools/system/package_manager.py
+++ b/conan/tools/system/package_manager.py
@@ -39,7 +39,7 @@ class _SystemPackageManagerTool(object):
             os_name = distro.id() or os_name
         elif os_name == "Windows" and self._conanfile.conf.get("tools.microsoft.bash:subsystem") == "msys2":
             os_name = "msys2"
-        manager_mapping = {"apt-get": ["Linux", "ubuntu", "debian", "raspbian"],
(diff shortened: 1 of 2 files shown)


memsharded (commented):
Thanks for your contribution @RMZeroFour 
Could you please add this to the tests? In ``def test_package_manager_distro(distro, tool):`` test, it should be just a new entry.

RMZeroFour:
There should probably be a helpful error or at least a warning on systems where the default package manager cannot be found and the user hasn't set one, right now Conan only prints out the name of the package not found, with no indication for why (no package manager). I'd be willing to contribute another commit for raising said error, if someone would be kind enough to point me to where might be the best place for this. 

Will the run() method below be a suitable place to raise an error, if self._active_tool is not set?

memsharded:
> There should probably be a helpful error or at least a warning on systems where the default package manager cannot be found and the user hasn't set one

Sounds that the ``_SystemPackageManagerTool.get_default_tool`` would be a good place, ``run()`` might be too noisy. 
Also, it shouldn't be a warning, because it is very common that recipes define some system packages for Linux that are not necessary neither in Windows or in OSX, but recipes do not explicitly check for this. So it shouldn't be a warning in Windows or OSX, for example. But some informational message, like "A default system package manager couldn't be found for XXX, system packages will not be installed" could be good.

RMZeroFour:
> > There should probably be a helpful error or at least a warning on systems where the default package manager cannot be found and the user hasn't set one
> 
> Sounds that the `_SystemPackageManagerTool.get_default_tool` would be a good place, `run()` might be too noisy. Also, it shouldn't be a warning, because it is very common that recipes define some system packages for Linux that are not necessary neither in Windows or in OSX, but recipes do not explicitly check for this. So it shouldn't be a warning in Windows or OSX, for example. But some informational message, like "A default system package manager couldn't be found for XXX, system packages will not be installed" could be good.

Gotcha, that makes sense. A self._conanfile.output.info at the end of the get_default_tool function should work right? Please have a look.

memsharded (approved):

memsharded:
Sounds good, approved, it will be merged for next 2.0.14 release, thanks again!

PR #15047, opened 2023-11-02
Correct encoding of make_program path for AutoTools + msys2

### LINKED ISSUE #15046: [bug] AutoTools.make() fails under MSYS2 with custom make program
### Environment details

* Conan version: 1.61.0
* Python version: 3.11.0

### Steps to reproduce

1. Choose an existing `autoconf` project as target. I used `libexif`
2. Modify `build_requirements()` to use current `make` package. Code snippet:
~~~
    def build_requirements(self):
        if self._settings_build.os == "Windows":
            self.win_bash = True
            if not self.conf.get("tools.microsoft.bash:path", check_type=str):
                self.tool_requires("msys2/cci.latest")
            else:
                self.tool_requires("make/4.4")

    def build(self):
        autotools = Autotools(self)
        autotools.configure()
        autotools.make()
~~~
3. Execute `conan create conanfile.py nmassey/devel`

### Logs

Sections of output
~~~
$ conan create conanfile.py nmassey/devel
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Debug
compiler=Visual Studio
compiler.runtime=MDd
compiler.version=17
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
[conf]
tools.microsoft.bash:path=C:\msys64\usr\bin\bash.exe
tools.microsoft.bash:subsystem=msys2
libexif/0.6.24@nmassey/devel: Forced build from source
Build requirements
    make/4.4 from local cache - Cache
Build requirements packages
    make/4.4:01edd76db8e16db9b38c3cca44ec466a9444c388 - Cache
make/4.4: Already installed!
libexif/0.6.24@nmassey/devel: Applying build-requirement: make/4.4
libexif/0.6.24@nmassey/devel: WARN: Build folder is dirty, removing it: C:\.build\37392e\1
libexif/0.6.24@nmassey/devel: Copying sources to build folder
libexif/0.6.24@nmassey/devel: Building your package in C:\.build\37392e\1
libexif/0.6.24@nmassey/devel: Generator txt created conanbuildinfo.txt
libexif/0.6.24@nmassey/devel: Calling generate()
libexif/0.6.24@nmassey/devel: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
libexif/0.6.24@nmassey/devel: Aggregating env generators
libexif/0.6.24@nmassey/devel: Calling build()
libexif/0.6.24@nmassey/devel: Calling:
 > "/c/.build/37392e/1/src/configure" --disable-shared --enable-static --prefix=/ --bindir=${prefix}/bin --sbindir=${prefix}/bin --libdir=${prefix}/lib --includedir=${prefix}/include --oldincludedir=${prefix}/include --disable-docs --disable-nls --disable-rpath --disable-dependency-tracking 
----Running------
> "C:\.build\37392e\1\build-debug\conan\conanbuild.bat" && "C:\.build\37392e\1\build-debug\conan\msys2_mode.bat" && C:\msys64\usr\bin\bash.exe --login -c ^". \^"C:\.build\37392e\1\build-debug\conan\conanbuild.sh\^" ^&^& \^"/c/.build/37392e/1/src/configure\^" --disable-shared --enable-static --prefix=/ --bindir=${prefix}/bin --sbindir=${prefix}/bin --libdir=${prefix}/lib --includedir=${prefix}/include --oldincludedir=${prefix}/include --disable-docs --disable-nls --disable-rpath --disable-dependency-tracking ^"
-----------------
~~~
Lots of config output
~~~
Configuration (libexif 0.6.24):
  Build
    Source code location:          /c/.build/37392e/1/src
    Compiler:                      /c/.build/37392e/1/src/compile cl -nologo
    Ship binaries in tarball:      false
    Use translations:              no
You may run "make" and "make install" now.
----Running------
> "C:\.build\37392e\1\build-debug\conan\conanbuild.bat" && "C:\.build\37392e\1\build-debug\conan\msys2_mode.bat" && C:\msys64\usr\bin\bash.exe --login -c ^". \^"C:\.build\37392e\1\build-debug\conan\conanbuild.sh\^" ^&^& C:\.build\c9af77\1\bin\gnumake.exe -j2^"
-----------------
~~~
...
~~~
[vcvarsall.bat] Environment initialized for: 'x64'
/usr/bin/bash: line 1: C:.buildc9af771bingnumake.exe: command not found
libexif/0.6.24@nmassey/devel: 
libexif/0.6.24@nmassey/devel: ERROR: Package '164640aad040835ac8988[239](https://dev2.sec.kmbs.us/dispatcher-phoenix/conan/libexif/-/jobs/254246#L239)3a96d89200694f04' build failed
libexif/0.6.24@nmassey/devel: WARN: Build folder C:\.build\37392e\1\build-debug
  File "C:\Python311\Lib\site-packages\conans\errors.py", line 34, in conanfile_exception_formatter
    yield
  File "C:\Python311\Lib\site-packages\conans\client\conanfile\build.py", line 16, in run_build_method
    conanfile.build()
  File "C:\.conan\data\libexif\0.6.24\nmassey\devel\export\conanfile.py", line 104, in build
    autotools.make()
  File "C:\Python311\Lib\site-packages\conan\tools\gnu\autotools.py", line 58, in make
    self._conanfile.run(command)
  File "C:\Python311\Lib\site-packages\conans\model\conan_file.py", line 464, in run
    raise ConanException("Error %d while executing %s" % (retcode, command))
conans.errors.ConanException: Error 127 while executing C:\.build\c9af77\1\bin\gnumake.exe -j2
~~~

nmasseyKM:
Changelog: Bugfix: Solve path issues with ``tools.gnu:make_program`` in Windows subsystems.
Docs: Omit

fixes #15046

- [x] Refer to the issue that supports this Pull Request. (fixes #15046)
  I'd be happy to, but I'm not sure what docs would need to be updated


memsharded:
Thanks very much for your PR!
I will try to reduce a bit the test scope, it is taking 1:30 in my Windows machine to run, I'll try to cover the case but reducing the time to run.

memsharded:
In [5b33028](https://github.com/conan-io/conan/pull/15047/commits/5b33028889dc3aa322bd976204e3cfd71e810c84) I have reduced the test so it runs in a few seconds, and still covers the use case.

However, I am a bit concerned about the change:
- At the moment the users can define the ``make_program`` via ``conf``. They can specify there any value they want, with the syntax and "unix_path" they want. 
- It can be failing if users provide a path with \\ backslash, but the solution is easy, provide a path with forward slash
- If we force now to convert to ``unix_path``, then in all scenarios it will be tried to be converted. If for some reason some users have a Make outside of the subsystem, it might fail. I know it is not that likely, but still possible that users are using a subsystem and still their toolchain, including make is not installed in that subsystem. For example, this is more likely for some use cases of ``Autotools`` with bare Makefiles, not really configure scripts. I know this was not the original intention of ``Autotools``, but it happens.

Please let me know what you think.

nmasseyKM:
> In [5b33028](https://github.com/conan-io/conan/pull/15047/commits/5b33028889dc3aa322bd976204e3cfd71e810c84) I have reduced the test so it runs in a few seconds, and still covers the use case.
> 
> However, I am a bit concerned about the change:
> 
> * At the moment the users can define the `make_program` via `conf`. They can specify there any value they want, with the syntax and "unix_path" they want.

They can also specify "bash:path" with the syntax they want (in system or "unix_path" syntax), and autotools works correctly. That's the expected behavior I am trying to replicate.

> * It can be failing if users provide a path with \ backslash, but the solution is easy, provide a path with forward slash

Agreed. If the user is providing the path manually, they can conform to a required syntax.

 However, this path is being supplied programmatically in by `make`'s conanfile.py as
```
        make = os.path.join(self.package_folder, "bin", "gnumake.exe" if self.settings.os == "Windows" else "make")
        self.conf_info.define("tools.gnu:make_program", make)
```
and it seems like that syntax should result in a valid path.

Of secondary consideration, trying to script this in windows tools (such as powershells `Get-Command` cmdlit) will also provide the paths in Windows native form.

> * If we force now to convert to `unix_path`, then in all scenarios it will be tried to be converted. If for some reason some users have a Make outside of the subsystem, it might fail.

You're right, I think  922cb85ba is a cleaner approach.

> I know it is not that likely, but still possible that users are using a subsystem and still their toolchain, including make is not installed in that subsystem. For example, this is more likely for some use cases of `Autotools` with bare Makefiles, not really configure scripts. I know this was not the original intention of `Autotools`, but it happens.
> 

It's definitely possible to have a `msys2` subsystem without make installed. That's how I discovered this issue. :-)
If that's the case, and "tools.gnu:make_program" is not set, then `Autotools.make()` is going to fail regardless of what we do to the default value: default behavior is to expand to a make which is alongside bash.

The user has the option of using their profile to configure the package they would like to provide "tools.gnu:make_program", which seems a sensible path since that configuration would be alongside the configuration of "bash:path" and "bash:subsystem".
Recipes _could_ address this if they want, along the lines of
```
    def build_requirements(self):
        if self._settings_build.os == "Windows":
            self.win_bash = True
            if not self.conf.get("tools.microsoft.bash:path", check_type=str):
                self.tool_requires("msys2/cci.latest")
            else if not self.conf.get("tools.gnu:make_program", check_type=str):
                self.tool_requires("make/[>=4.0]")
```

memsharded:
Sorry that it took so long to follow up on this.

As there were apparently trivial workarounds, like putting the ``/`` in the profile conf, this was never a priority.

Also, the recipe for ``make`` in ConanCenter could have easily done:
```python
make = os.path.join(self.package_folder, "bin", "gnumake.exe" if self.settings.os == "Windows" else "make")
make = make.replace("\\", "/")
self.conf_info.define("tools.gnu:make_program", make)
```
And problem solved.

But I have just checked, and the ``make`` recipe hadn't changed.
It is also true that this has never been an issue in ConanCenter because there is no overlap between ``msys2`` and ``make`` pacakges. Basically, if some package is using ``msys2``, they do not require ``make``, as they will use the make from inside ``msys2``.


I have merged from develop2, modernized and fixed the test to work for Conan2. I will take it to the ConanCenter team to look into this.

If you are still there, and would like to clarify what is the use case of using a different ``make`` than the one provided by ``msys2``, that could help to push this.

Thanks very much again for your contribution and your patience.

memsharded:
Merged, it will be in 2.27, many thanks!

PR #15082, opened 2023-11-10
test for build-missing of transitive tool-requires

memsharded:
Changelog: Bugfix: ``--build=missing`` was doing unnecessary builds of packages that were not needed and could be skipped, in the case of ``tool_requires`` having transitive dependencies.
Docs: Omit

### DIFF
--- a/conans/client/graph/graph.py
+++ b/conans/client/graph/graph.py
@@ -65,6 +65,7 @@ class Node(object):
         self.error = None
         self.cant_build = False  # It will set to a str with a reason if the validate_build() fails
         self.should_build = False  # If the --build or policy wants to build this binary
+        self.build_allowed = False
 
     def __lt__(self, other):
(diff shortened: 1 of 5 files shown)


AbrilRBS (approved):
🤯 
After playing around with this seems like pure bad luck that we didn't catch it earlier

memsharded:
> After playing around with this seems like pure bad luck that we didn't catch it earlier

The main reason could be that it doesn't really fail, but it builds from source, so unless it is something slow and you are paying attention, it can go unnoticed. And of course, also that the scenario is not that common:
- tool_requires with transitive dependencies
- Doing a ``--build=missing``
- Actually missing those transitive dependencies binaries, which is very unlikely (I realized because I copied to a private repo and remove conancenter remote)

czoido (approved):

PR #15102, opened 2023-11-15
CMP0091 should be only set on Windows

ShuangLiu1992:
CMP0091 should be only set on Windows, otherwise it will generate a warning on newer version of CMake (tested on 3.27.0)

CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_POLICY_DEFAULT_CMP0091

### DIFF
--- a/conan/tools/cmake/presets.py
+++ b/conan/tools/cmake/presets.py
@@ -170,8 +170,8 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables,
             cmake_make_program = cmake_make_program.replace("\\", "/")
             cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
 
-    if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
-        cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
+        if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
+            cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
 
     if "BUILD_TESTING" not in cache_variables:
         if conanfile.conf.get("tools.build:skip_test", check_type=bool):


conan/tools/cmake/presets.py
```diff
@@ -170,8 +170,8 @@ def write_cmake_presets(conanfile, toolchain_file, generator, cache_variables,
             cmake_make_program = cmake_make_program.replace("\\", "/")
             cache_variables["CMAKE_MAKE_PROGRAM"] = cmake_make_program
 
-    if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
-        cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
+        if "CMAKE_POLICY_DEFAULT_CMP0091" not in cache_variables:
+            cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
```
memsharded, line 174: I think there might be some unusual but possible scenarios of cross-building from Linux to Windows that still need this definition. So it cannot be conditioned to ``platform.system() == Windows`` but it should be conditioned to ``settings.os == "Windows"``.

memsharded: Also, there are scenarios using MinGW makefiles that still use ``msvc`` as compiler and will need the Visual runtime, so this shouldn't be conditioned to the generator.

memsharded (commented):
Thanks for your contribution!
I have a couple of concerns, please have a look.

conan/tools/cmake/presets.py
ShuangLiu1992: Hi, thank you for your reply.
Do you have any other suggestion to skip setting "CMAKE_POLICY_DEFAULT_CMP0091" from elsewhere? I don't know where it should be and it's quite annoying to see the warning on newer version of CMake.

SpaceIm: It would be a good idea to use the same condition than the one which makes `VSRuntimeBlock` returning something not None.

`VSRuntimeBlock` is a little bit convoluted, but basically its consolidated condition seems to be:

`settings.get_safe("os") == "Windows" and settings.get_safe("compiler") in ("msvc", "clang", "intel-cc") and settings.get_safe("compiler.runtime") and settings.get_safe("build_type")`

So it could be factorized and used in `VSRuntimeBlock` and `_CMakePresets` classes.

memsharded:
I think we should be pragmatic here.
This warning is also inconvenient for many Windows-MSVC cases. Every project that contains a ``cmake_minimum_required(3.15)`` or above, will still warn because the policy is no longer read. Doing the following:
```diff
diff --git a/conan/tools/cmake/toolchain/toolchain.py b/conan/tools/cmake/toolchain/toolchain.py  
index 7e47045c8..37a2a194e 100644                                                                 
--- a/conan/tools/cmake/toolchain/toolchain.py                                                    
+++ b/conan/tools/cmake/toolchain/toolchain.py                                                    
@@ -115,6 +115,8 @@ class CMakeToolchain(object):                                                 
         {% endfor %}                                                                             
         # Preprocessor definitions per configuration                                             
         {{ iterate_configs(preprocessor_definitions_config, action='add_compile_definitions') }} 
+                                                                                                 
+        set(_unused ${CMAKE_POLICY_DEFAULT_CMP0091})                                             
         """)                                                                                     
                                                                                                  
     def __init__(self, conanfile, generator=None):                                               
```
Removes the warning for absolutely all cases, with 0 risks of breaking something. While limiting the cases in which we define the variable might break some legacy projects that depends on this and do not satisfy some of the conditions for some particular reasons, and will still be warning in Windows-msvc, with that cmake_minimum_require

What do you think? Want to submit that fix yourself? If not, we could close this one and I can do a new PR myself.

ShuangLiu1992:
> I think we should be pragmatic here. This warning is also inconvenient for many Windows-MSVC cases. Every project that contains a `cmake_minimum_required(3.15)` or above, will still warn because the policy is no longer read. Doing the following:
> 
> ```diff
> diff --git a/conan/tools/cmake/toolchain/toolchain.py b/conan/tools/cmake/toolchain/toolchain.py  
> index 7e47045c8..37a2a194e 100644                                                                 
> --- a/conan/tools/cmake/toolchain/toolchain.py                                                    
> +++ b/conan/tools/cmake/toolchain/toolchain.py                                                    
> @@ -115,6 +115,8 @@ class CMakeToolchain(object):                                                 
>          {% endfor %}                                                                             
>          # Preprocessor definitions per configuration                                             
>          {{ iterate_configs(preprocessor_definitions_config, action='add_compile_definitions') }} 
> +                                                                                                 
> +        set(_unused ${CMAKE_POLICY_DEFAULT_CMP0091})                                             
>          """)                                                                                     
>                                                                                                   
>      def __init__(self, conanfile, generator=None):                                               
> ```
> 
> Removes the warning for absolutely all cases, with 0 risks of breaking something. While limiting the cases in which we define the variable might break some legacy projects that depends on this and do not satisfy some of the conditions for some particular reasons, and will still be warning in Windows-msvc, with that cmake_minimum_require
> 
> What do you think? Want to submit that fix yourself? If not, we could close this one and I can do a new PR myself.

Yes I'd close this one and leave it you. Thank you again for your quick response :)

PR #15116, opened 2023-11-16
Removed useless distutils import

### LINKED ISSUE #15114: [bug] Building Conan fails with Python 3.12
### Environment details

* Operating System+version: any
* Compiler+version: any
* Conan version: 2.0.14
* Python version: 3.12

### Steps to reproduce

1. Make sure Python 3.12 is installed
2. Run `pip install -e .`
3. Run `python pyinstaller.py`

With Python 3.12 the deprecated `distutils` package was removed ([PEP 632](https://peps.python.org/pep-0632/#migration-advice)) which now makes the Conan build fail. See: https://github.com/conan-io/conan/blob/889ffbc1243a2542ac87d4761e005e6e0fab8466/pyinstaller.py#L22

### Logs

```
[2023-11-15T21:52:13.937Z] Traceback (most recent call last):
[2023-11-15T21:52:13.937Z]   File "c:\c\ws\6989kc6findlvv8\conan-2.0.14\pyinstaller.py", line 22, in <module>
[2023-11-15T21:52:13.937Z]     from distutils import dir_util
[2023-11-15T21:52:13.937Z] ModuleNotFoundError: No module named 'distutils'
```

franramirez688:
Changelog: Bugfix: `pyinstaller.py` was broken for Python 3.12 due to a useless `distutils` import.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/15114

### DIFF
--- a/pyinstaller.py
+++ b/pyinstaller.py
@@ -103,7 +103,6 @@ def pyinstall(source_folder):
     conan_path = os.path.join(source_folder, 'conans', 'conan.py')
     hidden = ("--hidden-import=glob "  # core stdlib
               "--hidden-import=pathlib "
-              "--hidden-import=distutils.dir_util "
               # Modules that can be imported in ConanFile conan.tools and errors
               "--collect-submodules=conan.cli.commands "
               "--hidden-import=conan.errors "
--- a/pyinstaller.py
+++ b/pyinstaller.py
@@ -19,7 +19,6 @@ import os
 import platform
 import shutil
 import subprocess
-from distutils import dir_util
 
 from conans import __version__
 from conans.util.files import save


pyinstaller.py
```diff
@@ -19,7 +19,6 @@
 import platform
 import shutil
 import subprocess
-from distutils import dir_util
```
gjasny, line 22: Further down there's the following line:
```
"--hidden-import=distutils.dir_util "
```
Could that go as well?

franramirez688: It was fine locally even keeping that line, but let's wait for @czoido @memsharded feedback.

memsharded: What Python version are we using to generate the installers? Do we have that Python version documented somewhere? It would be necessary to test the ``pyinstaller`` script in that minimum Python, to check that it doesn't break.

czoido: I think those can be removed, they are legacy from Conan 1.X. I have checked with the ci job that generates the installers and everything works removing both the imports and the "--hidden-import=distutils.dir_util" line. There we use 3.6 in Windows and 3.8 and 3.9 in Linux and Macos respectively. So I think these changes should be fine.

franramirez688: I tried it over Python 3.11 and 3.12. Both are working correctly.

czoido (approved):

AbrilRBS (approved):
Works fine for me locally as well in the few versions I've tried

PR #15130, opened 2023-11-17
bug fix: migrations do not respect core.cache:storage_path

### LINKED ISSUE #15129: 2.0.14's cache migrations do not respect core.cache:storage_path
### Environment details

* Operating System+version: Mac OS 13.4.1
* Compiler+version: n/a
* Conan version: 2.0.14
* Python version: 3.10.8

### Steps to reproduce

* install conan 2.0.13

* set `core.cache:storage_path` to something  other than `~/.conan2/p`.  remove `~/.conan2/p`

* upgrade to conan 2.0.14

* try to use conan, it will fail with "Could not complete the 2.0.14 DB migration"

### Logs

```
➜  conan  list \*    
WARN: Upgrade cache from Conan version '2.0.13'
WARN: Running 2.0.14 Cache DB migration to add LRU column
ERROR: unable to open database file
```

smoofra:
migrations do not respect core.cache:storage_path.  If the user has this option set, and tries to migrate from conan 2.0.13 to 2.0.14, the migration will fail.

To fix, client migrations are split into two migrators, one for $CONAN_HOME, and one for the local cache.

The local cache now has its own version.txt and migrations/ directory,
for future migrations.   But the undo script from 2.0.14 is still
placed in $CONAN_HOME, because existing versions of conan will
not know to look for it anywhere else.

The undo script from 2.0.14 is also updated to use the config to find the correct path to the local cache.

fixes https://github.com/conan-io/conan/issues/15129


memsharded (commented):
Hi @smoofra 

Thanks very much for this contribution!
I think there are 2 different things here:
- The fix of the bug that is ignoring the ``global.conf`` ``storage`` definition
- A new system for migrations that is able to migrate the "storage", assigning a version to the storage too.

I think we should decouple both, first, fix quickly the bug, so migration happens even if ``storage`` conf is defined, assuming the current 1:1 assumption between the Conan home and the package storage.

Then, we should consider more carefully the change of migration approach with versions in the storage folder, because this can have some challenges, better handle them separately.

smoofra:
> I think we should decouple

OK, how's it look now?

conan/api/conan_api.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, cache_folder=None):
         self.graph = GraphAPI(self)
         self.export = ExportAPI(self)
         self.remove = RemoveAPI(self)
-        self.config = ConfigAPI(self)
+        self.config = ConfigAPI(self.home_folder)
```
memsharded, line 52: Better not change the interface of ConfigAPI constructor. All subapis are expected to receive the ConanAPI as argument.

conan/api/subapi/config.py
```diff
@@ -14,18 +14,18 @@
 
 class ConfigAPI:
 
-    def __init__(self, conan_api):
-        self.conan_api = conan_api
+    def __init__(self, home_folder):
+        self.home_folder = home_folder
```
memsharded, line 18: All changes in this file seems unnecessary, just a side effect of changing the ``ConfigAPI``  constructor. Better do not change this file.

memsharded (commented):
I have proposed https://github.com/conan-io/conan/pull/15135 as an alternative for this PR.

Thanks very much for your contribution, it is technically good, but there some architecture aspects that we would like to do slightly different:
- We prefer all the subapis to have the same interface for construction
- The LRU DB migration is a temporary one, once the 2.X train starts (2.1, 2.2...) can be removed later, so it is better that it doesn't affect other parts of the code
- So a little bit of repetition like the one that happens in the migration can be preferred

I have also included a test in my PR to check that the problem is really addressed.

smoofra:
> I have proposed https://github.com/conan-io/conan/pull/15135 as an alternative for this PR.
> 

ok 😃 but I think you should still include a fixed version of the undo script so that migration backwards to 2.13, and then forwards to 2.15 still works

PR #15135, opened 2023-11-19
fix migration lru

### LINKED ISSUE #15129: 2.0.14's cache migrations do not respect core.cache:storage_path

memsharded:
Changelog: Bugfix: Fix the 2.0.14 migration to add LRU data to the cache when ``storage_path`` conf is defined.
Docs: Omit

Close https://github.com/conan-io/conan/pull/15130
Close https://github.com/conan-io/conan/issues/15129


smoofra (changes requested):
What about the undo script for backwards migration?

You should be able to migrate backwards and forward between 2.0.13 and 2.0.15 without a problem right? So we still need to include a fixed version of the undo script I think.

memsharded:
Fixed the back migration too, good catch.

conans/client/migrations.py
```diff
@@ -75,8 +78,21 @@ def _migrate_pkg_db_lru(cache_folder, old_version):
...
             def migrate(cache_folder):
-                db = os.path.join(cache_folder, 'p', 'cache.sqlite3')
+                config = os.path.join(cache_folder, "global.conf")
+                global_conf = open(config, "r").read() if os.path.isfile(config) else ""
+                storage = os.path.join(cache_folder, "p")
+                for line in global_conf.splitlines():
+                    tokens = line.split("=")
+                    if len(tokens) != 2:
+                        continue
+                    conf, value = tokens
```
smoofra: shoudln't this use ConfDefinition to parse quotes and whatnot?

memsharded: The truth is that I was trying to do a shortcut, to avoid having to check every change in the code in the previous versions, because the code has to be compatible with the old version, not that evident with all the work and changes since 2.0.0. But yeah, probably leveraging some of the previous code is more robust than this, I'll try to find some common base.

memsharded: Done

AbrilRBS (approved):

PR #15136, opened 2023-11-19
require-replaces proposal

memsharded:
Changelog: Feature: New ``[replace_requires]`` and ``[replace_tool_requires]`` in profile for redefining requires, useful for package replacements like ``zlibng/zlib``, to solve conflicts, and to replace some dependencies by system alternatives wrapped in another Conan package recipe.
Docs: https://github.com/conan-io/docs/pull/3495

Complement https://github.com/conan-io/conan/pull/14871


czoido:
I have been testing the feature and I think it could be quite useful. Throwing here some other names that could fit, to discuss in the team (as well as in [the other](https://github.com/conan-io/conan/pull/14871) PR, I personally prefer to keep the names as: _something_requires_ and _something_tool_requires_):

- `fallback_requires` and `fallback_tool_requires`
- `substitute_requires` and `substitute_tool_requires`
- `alternate_requires` and `alternate_tool_requires`
- `replacement_requires` and `replacement_tool_requires`

memsharded:
I think my preferred one would be ``alternate_requires and alternate_tool_requires``, lets see what other say.

conans/client/profile_loader.py
```diff
@@ -238,6 +239,20 @@ def get_profile(profile_text, base_profile=None):
...
         platform_requires = [RecipeReference.loads(r) for r in doc_platform_requires.splitlines()]
 
+        def load_replace(doc_replace_requires):
+            result = {}
+            for r in doc_replace_requires.splitlines():
+                r = r.strip()
+                if not r or r.startswith("#"):
+                    continue
+                src, target = r.split(":", 1)
+                target = RecipeReference.loads(target.strip())
```
AbrilRBS: A `RecipeReference` performs no validation of its own, so there's nothing really stopping users from specifying a package id in the target (I think we break if specified as part of the src due to the split above). Do they get an error some other way?

Do we want to guard against those cases? I wouldn't expect users to try to add those bits if we tell them it's not meant for things like these, but I never know anymore

AbrilRBS (commented):
Looks good. I have a concern regarding the `RecipeReference` usage, but the rest works really nicely!

conans/client/profile_loader.py
memsharded: Yeah, better to add some checking and good error messages. I'll add it.

memsharded: Added, please re-check

AbrilRBS (approved):

czoido (approved):
Just one question. Thinking about resolving conflicts. 

Would it make sense that we support multiple replacements resolved in order like this:

```
include(default)
[replace_requires]
xz_utils/5.4.*: xz_utils/5.4.4
xz_utils/4.5.*: xz_utils/4.5.0
xz_utils/3.2.*: xz_utils/3.2.3
...
```

I'm thinking in using the same profile for different projects. Maybe I don't want to do `xz_utils/*: xz_utils/5.4.4` in my profile and always resolve to that same version. 

Maybe we could consider in the future if that makes sense, anyway I think it's great as it is right now.

AbrilRBS (approved):

memsharded:
> Would it make sense that we support multiple replacements resolved in order like this:

I think it makes sense to check this now, add a test or something and understand the use case before it could be too late. I'll check it @czoido

memsharded:
Ok, I recall now why this is challenging:

- The pattern ``pkg/1.2.*`` is effectively a version range, or equivalent to ``pkg/[>=1.2.0 <1.3]`` mostly
- If we allow version ranges like this, what happens when we find a requires like ``requires = "pkg/[>=1.0 <1.2.5]``?
- Note now we have ranges intersection capabilities, but still it sounds a bit weird, if we compute the intersection, and find a non-empty match of the intersection, then we should proceed and do the replacement?

This was the reason why so far I have only consider the pattern ``*``, all-or-nothing matching the version field, but not partial matching versions. Let's discuss it.

franramirez688 (commented):
This is great @memsharded 👏 

A couple of doubts:

* Should we show anything in the "Computing the dependency graph" output? For instance:

```
Graph root
    cli
Requirements
    hell/1.0.0#655853d1fb17b93c66482a7f86f7c0ff - Cache
    zlib/1.2.11#dc0e384f0551386cd76dc29cc964c95e - Cache
Replacements
    zlib/1.2.xx: zlib/1.2.11#dc0e384f0551386cd76dc29cc964c95e
```

* I tried something like:

```
[replace_requires]
zlib/1.2.123: zlib/*
```
The output:
```
zlib/1.2.123: Not found in local cache, looking in remotes...
zlib/1.2.123: Checking remote: conancenter
Graph root
    cli
Requirements
    hell/1.0.0#655853d1fb17b93c66482a7f86f7c0ff - Cache
ERROR: Package 'zlib/1.2.123' not resolved: Unable to find 'zlib/1.2.123' in remotes.
```
Now, I changed it:
```
[replace_requires]
zlib/1.2.123: zlib/1.2.*
```
Output:
```
zlib/1.2.*: Not found in local cache, looking in remotes...
zlib/1.2.*: Checking remote: conancenter
Graph root
    cli
Requirements
    hell/1.0.0#655853d1fb17b93c66482a7f86f7c0ff - Cache
ERROR: Package 'zlib/1.2.*' not resolved: Unable to find 'zlib/1.2.*' in remotes.
```
I'm not saying that this is a valid flow, but I think that we should cover those cases with a more meaningful output. Wdyt?

memsharded:
> Should we show anything in the "Computing the dependency graph" output?

I think adding some explicit information about the replacements, at least a summary, would be good, yes, I'll try to add it.

> zlib/1.2.123: zlib/1.2.*

I'll try to add some validation and test for this case too

franramirez688 (approved):

czoido (approved):

PR #15140, opened 2023-11-20
Try to address NMakeDeps quoting issues

### LINKED ISSUE #13603: [bug] NMakeDeps: doesn't properly handle cpp_info.defines with special characters
### Environment details

* Conan version: 2.0.3
* Python version: 3.10.6

### Steps to reproduce

Call `conan install --requires="gdal/3.4.3" -o "sqlite3/*:shared=True" -b missing` (you may have to write a conanfile.py in order to override versions of some dependencies since there are some conflicts currently I think).

My analysis of this issue:
When compiler is Visual Studio, gdal recipe (if version < 3.5) relies on `NMakeToolchain` & `NMakeDeps`. sqlite3 is an optional dependency of gdal, but enabled by default. When sqlite3 is shared on Windows, its recipe appends `SQLITE_API=__declspec(dllimport)` into `cpp_info.defines` (https://github.com/conan-io/conan-center-index/blob/eb52bc7ae38ab3cd857d38b0d32be39c45500fa6/recipes/sqlite3/all/conanfile.py#L207-L209).

I guess that my PR https://github.com/conan-io/conan/pull/12944 is not robust to some special characters in these definitions, they may have to be protected more than what is currently implemented here: https://github.com/conan-io/conan/blob/2.0.3/conan/tools/microsoft/nmakedeps.py#L54

### Logs

I'll attach the log soon, but basically all compilation units of gdal including sqlite3.h fail. There is one error per line where `SQLITE_API` appears, something like 'string' unknown.

AbrilRBS:
Changelog: Bugfix: Fix NMakeDeps quoting issues.
Docs: Omit

This is a first try, need to double check with @uilianries (Thanks to @danimtb for checking this locally for me!)

cc @SpaceIm 

https://learn.microsoft.com/en-us/cpp/build/reference/d-preprocessor-definitions?view=msvc-170#remarks

Closes #13603

### DIFF
--- a/conan/tools/microsoft/nmakedeps.py
+++ b/conan/tools/microsoft/nmakedeps.py
@@ -50,8 +50,12 @@ class NMakeDeps(object):
                     # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
+                    if value and not value.isnumeric():
+                        value = f'\"{value}\"'
                     define = f"{macro}#{value}"
-                return f'/D"{define}"'
+                    if " " in define:
+                        define = f"\"{define}\""
+                return f"/D{define}"
 
             cl_flags = [f'-I"{p}"' for p in cpp_info.includedirs or []]
             cl_flags.extend(cpp_info.cflags or [])
--- a/conan/tools/microsoft/nmakedeps.py
+++ b/conan/tools/microsoft/nmakedeps.py
@@ -53,9 +53,7 @@ class NMakeDeps(object):
                     if value and not value.isnumeric():
                         value = f'\"{value}\"'
                     define = f"{macro}#{value}"
-                    if " " in define:
-                        define = f"\"{define}\""
-                return f"/D{define}"
+                return f"/D\"{define}\""
 
             cl_flags = [f'-I"{p}"' for p in cpp_info.includedirs or []]
             cl_flags.extend(cpp_info.cflags or [])
--- a/conan/tools/microsoft/nmakedeps.py
+++ b/conan/tools/microsoft/nmakedeps.py
@@ -50,10 +50,11 @@ class NMakeDeps(object):
                     # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
-                    if value and not value.isnumeric():
-                        value = f'\\"{value}\\"'
+                    value = value.replace("(", "\\(").replace(")", "\\)")
                     define = f"{macro}#{value}"
-                return f"/D{define}"
+                if " " in define:
+                    return f'"/D{define}"'
+                return f'/D{define}'
 
             cl_flags = [f'-I"{p}"' for p in cpp_info.includedirs or []]
             cl_flags.extend(cpp_info.cflags or [])
--- a/conan/tools/microsoft/nmakedeps.py
+++ b/conan/tools/microsoft/nmakedeps.py
@@ -50,7 +50,6 @@ class NMakeDeps(object):
                     # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
-                    value = value.replace("(", "\\(").replace(")", "\\)")
                     define = f"{macro}#{value}"
                 if " " in define:
                     return f'"/D{define}"'
--- a/conan/tools/microsoft/nmakedeps.py
+++ b/conan/tools/microsoft/nmakedeps.py
@@ -51,9 +51,7 @@ class NMakeDeps(object):
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
                     define = f"{macro}#{value}"
-                if " " in define:
-                    return f'"/D{define}"'
-                return f'/D{define}'
+                return f'/D"{define}"'
 
             cl_flags = [f'-I"{p}"' for p in cpp_info.includedirs or []]
             cl_flags.extend(cpp_info.cflags or [])
--- a/conan/tools/microsoft/nmaketoolchain.py
+++ b/conan/tools/microsoft/nmaketoolchain.py
@@ -38,7 +38,9 @@ class NMakeToolchain(object):
                 if value and not value.isnumeric():
                     value = f'\\"{value}\\"'
                 define = f"{macro}#{value}"
-            formated_defines.append(f"/D{define}")
+                if " " in define:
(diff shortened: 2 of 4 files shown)


conan/tools/microsoft/nmakedeps.py
```diff
@@ -50,10 +50,10 @@ def format_define(define):
                     # CL env-var can't accept '=' sign in /D option, it can be replaced by '#' sign:
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
-                    if value and not value.isnumeric():
```
memsharded, line 53: Aren't we missing something here? Maybe it was not covered by tests? If value is a number, process is different?

memsharded (approved):
LGTM

SpaceIm:
`conans/test/integration/toolchains/microsoft/test_nmakedeps.py` doesn't prove that it fixes https://github.com/conan-io/conan/issues/13603, just that you write in toolchain file what you want to write.
Extending `conans/test/functional/toolchains/test_nmake_toolchain.py` with additional definitions like `FOO=__delcspec(dllimport)` would prove that it works.

valgur:
The `mslex` Python package might be useful here to solve any .bat quoting issues once and for all?
The core of the library is small enough to be vendored with Conan:
https://github.com/smoofra/mslex/blob/master/mslex/__init__.py
and here is its test suite:
https://github.com/smoofra/mslex/blob/master/tests/test_mslex.py

danimtb (approved):

conans/test/functional/toolchains/test_nmake_toolchain.py
```diff
@@ -13,7 +13,8 @@
     [
         ("msvc", "190", "dynamic", "14", "Release", [], [], [], [], []),
         ("msvc", "190", "dynamic", "14", "Release",
-         ["TEST_DEFINITION1", "TEST_DEFINITION2=0", "TEST_DEFINITION3=", "TEST_DEFINITION4=TestPpdValue4"],
+         ["TEST_DEFINITION1", "TEST_DEFINITION2=0", "TEST_DEFINITION3=", "TEST_DEFINITION4=TestPpdValue4",
```
memsharded, line 16: Missing assert/checks in test

memsharded (approved):

conans/test/functional/toolchains/test_nmake_toolchain.py
AbrilRBS: TIL https://github.com/conan-io/conan/blob/694b7d1c7e72c4cf39ac9bc8c6f5b0697f8aa82a/conans/test/functional/utils.py#L140 this is already checked! No need to manually do that

Called in line 87 with `check_exe_run` below

conan/tools/microsoft/nmaketoolchain.py
```diff
@@ -38,7 +38,9 @@ def _format_defines(self, defines):
                 if value and not value.isnumeric():
                     value = f'\\"{value}\\"'
                 define = f"{macro}#{value}"
-            formated_defines.append(f"/D{define}")
+                if " " in define:
+                    define = f"\"{define}\""
+            formated_defines.append(f'/D{define}')
```
memsharded: Why this different than ``NMakeDeps``? can't we use the same approach (always quoting)?

conan/tools/microsoft/nmakedeps.py
```diff
@@ -51,8 +51,10 @@ def format_define(define):
                     # https://learn.microsoft.com/en-us/cpp/build/reference/cl-environment-variables
                     macro, value = define.split("=", 1)
                     if value and not value.isnumeric():
-                        value = f'\\"{value}\\"'
+                        value = f'\"{value}\"'
                     define = f"{macro}#{value}"
+                    if " " in define:
+                        define = f"\"{define}\""
```
memsharded: But this will put an extra quotes for strings with spaces, because above has already put quotes for not-numbers?

conan/tools/microsoft/nmaketoolchain.py
danimtb: I have removed the conditional to the whitespace, as it works quoting always. However, it seems that the string values need an extra quote pair. I have tested it at the nmake toolchain side and is the only combination that works. Removing the non-numeric conditional throws the following errors:

```
simple.cpp(101): error C2065: 'TestPpdValue4': undeclared identifier
E           simple.cpp(101): error C2059: syntax error: '__declspec(dllexport)'
E           simple.cpp(101): error C2065: 'foo': undeclared identifier
E           simple.cpp(101): error C2146: syntax error: missing ';' before identifier 'bar'
E           simple.cpp(101): error C2065: 'bar': undeclared identifier
```

conans/test/integration/toolchains/microsoft/test_nmakedeps.py
```diff
@@ -38,8 +41,11 @@ def package_info(self):
     bat_file = client.load("conannmakedeps.bat")
     # Checking that defines are added to CL
     for flag in (
-        r"/DTEST_DEFINITION1", r"/DTEST_DEFINITION2#0",
-        r"/DTEST_DEFINITION3#", r'/DTEST_DEFINITION4#\\"foo\\"',
+        r'/D\"TEST_DEFINITION1\"', r'/D\"TEST_DEFINITION2#0\"',
+        r'/D\"TEST_DEFINITION3#\"', r'/D\"TEST_DEFINITION4#\"foo\"\"',
+        r'/D\"TEST_DEFINITION5#\"__declspec\(dllexport\)\"\"',
+        r'/D\"TEST_DEFINITION6#\"foo bar\"\"',
+        r'/D\"TEST_DEFINITION7#7\"'
```
memsharded: Can't we remove the ``r`` and the escaping backslashes? They seem unnecessary in most cases except ``/D\"TEST_DEFINITION6#\"foo bar\"\"`` (which by the way, it is the surprising thing, I am very surprised that nesting regular quotes works well.

danimtb: done!

AbrilRBS:
@danimtb should we check about what happens when using `tools.env.virtualenv:powershell=True`?

danimtb:
> @danimtb should we check about what happens when using `tools.env.virtualenv:powershell=True`?

@RubenRBS it works with test_nmake_toolchain and produces de following line in `nmakedeps.ps1` file:
```
$env:CL="$env:CL -I`"C:\Users\danielm\AppData\Local\Temp\tmpty9hbstsconans\path with spaces\.conan2\p\b\test-486504e48c907\p\include`" /D`"TEST_DEFINITION4#`"foo`"`" /D`"TEST_DEFINITION5#`"__declspec(dllexport)`"`" /D`"TEST_DEFINITION6#`"foo bar`"`" /D`"TEST_DEFINITION7#7`" /D`"TEST_DEFINITION3#`" /D`"TEST_DEFINITION2#0`" /D`"TEST_DEFINITION1`""
```

As both nmaketoolchain and nmakedeps produces env files, the format seems to be correct and it works

memsharded (approved):
Looks good to me.
If this is needed in ConanCenter, then a backport for 1.X might be necessary

PR #15148, opened 2023-11-21
add option copy if_different to conan.tools.files.copy()

boussaffawalid:
Changelog: (Feature): add option copy if_different to conan.tools.files.copy()

- [ ] Reimplemented https://github.com/conan-io/conan/pull/9389 for onan.tools.files.copy

### DIFF
--- a/conans/test/unittests/tools/files/test_tool_copy.py
+++ b/conans/test/unittests/tools/files/test_tool_copy.py
@@ -256,3 +256,29 @@ class ToolCopyTest(unittest.TestCase):
                          sorted(os.listdir(os.path.join(dst_folder, "include"))))
         self.assertEqual(sorted(["AttributeStorage.h", "file.h"]),
                          sorted(os.listdir(os.path.join(dst_folder, "include", "sub"))))
+
+    def test_if_different(self):
+        src_folder = temp_folder()
+        save(os.path.join(src_folder, "file.h"), "")
+
+        dst_folder = temp_folder()
+
+        copy(None, "*.h", src=src_folder, dst=dst_folder)
+
+        # creation time
+        dst_file = os.path.join(dst_folder, "file.h")
+        dst_file_ctime = os.stat(dst_file).st_ctime_ns
+
+        # test that file is not copied again
+        copy(None, "*.h", if_different=True, src=src_folder, dst=dst_folder)
+        self.assertEqual(dst_file_ctime, os.stat(dst_file).st_ctime_ns)
+
+        # test that file is copied again
+        copy(None, "*.h", if_different=False, src=src_folder, dst=dst_folder)
+        self.assertNotEqual(dst_file_ctime, os.stat(dst_file).st_ctime_ns)
+
+        # test that file is copied again when source changes
+        dst_file_ctime = os.stat(dst_file).st_ctime_ns
+        save(os.path.join(src_folder, "file.h"), "dummy")
+        copy(None, "*.h", if_different=True, src=src_folder, dst=dst_folder)
+        self.assertNotEqual(dst_file_ctime, os.stat(dst_file).st_ctime_ns)
--- a/conan/tools/files/copy_pattern.py
+++ b/conan/tools/files/copy_pattern.py
@@ -1,4 +1,5 @@
 import fnmatch
+import filecmp
 import os
 import shutil
 from collections import defaultdict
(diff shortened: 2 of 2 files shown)


boussaffawalid:
@memsharded should I open another PR for develop2 branch ?

conans/test/unittests/tools/files/test_tool_copy.py
```diff
@@ -256,3 +256,29 @@ def test_ignore_case_excludes(self):
...
         self.assertEqual(sorted(["AttributeStorage.h", "file.h"]),
                          sorted(os.listdir(os.path.join(dst_folder, "include", "sub"))))
+
+    def test_if_different(self):
+        src_folder = temp_folder()
+        save(os.path.join(src_folder, "file.h"), "")
+
+        dst_folder = temp_folder()
+
+        copy(None, "*.h", src=src_folder, dst=dst_folder)
```
memsharded, line 266: The usage of ``copy()`` is to copy files to a target empty folder. It would be good to have tests in ``integration`` that illustrate and cover the main problem or pain it is trying to address.

memsharded (commented):
Thanks for your contribution @boussaffawalid 

It is not very clear what problem this is trying to solve. The ``copy()`` is mostly used for copying to an empty folder, so checking against a possible existing file never really happens.


This should no longer target Conan 1.X, new functionalities like this one (an optimization) only go to Conan 2.0.

conans/test/unittests/tools/files/test_tool_copy.py
boussaffawalid: this is mostly useful in the `imports` method, (`generate` method for conan 2) for local dev workflow.

For local dev workflow we reuse the same build directory to speedup development. 

This feature is resolving two problems:
* Speed up configure time by not copying files over and over even if there are no changes (we have large projects)
*  Avoid permissions issue when some processes are running from build folder.

memsharded: Thanks for the feedback

> For local dev workflow we reuse the same build directory to speedup development.

In Conan 2.0 it is not that ``imports`` has been removed, it is that copying files at ``generate()`` method shouldn't be that common. Using generators and virtualenvs, or deployers would be recommended. Are you still copying heavy files?

It is important to note that ``deployers`` can also implement and use any copy, the built-in modern ``copy()`` can be easily replaced by ``shutil.copy`` if desired, and any extra checks can be done in the deployer code for efficiency.


In any case, this shouldn't target 1.X anymore, but 2.0. And for Conan 2.0, it probably deserves first a ticket explaining a bit:
- The use case, what files are being copied and for what purpose. Discuss what other alternatives approaches could be for that case that would avoid the copy
- The cost: how long does it take the copy, how many files and MBs?. How much time is saved with this feature? Measure in a couple of different OSs if possible including Windows too (it is not that evident that having to read 2 large files will be faster than overwriting a existing file without checking, in all OSs)

boussaffawalid: We do copy heavy files, one of our projects build folder is about 10 Gb :( 

Files being copied:
- Resource files from dependencies/manifest files
- Dlls for windows: We use conan VirtualRunEnv but some developers dont like using scripts and they prefer a solution that works with all IDE out of the box and therefore we need to copy all runtime DLLs.

filecmp does not read files to be able to compare them, it uses [os.stat()](https://docs.python.org/3/library/os.html#os.stat) signatures (file type, size, and modification time) of both files to compare them.

https://docs.python.org/3/library/filecmp.html#filecmp.cmp

We are still far from upgrading to conan 2, but I will try the deployers approach once we are there.

memsharded: > https://docs.python.org/3/library/filecmp.html#filecmp.cmp

That is interesting, I thought that the default was ``shallow=False``, not ``True``, but I was wrong, thanks for clarifying.

I think this can be considered for Conan 2.0. Unfortunately for Conan 1.X really not possible, we don't have enough resources to keep adding, documenting, supporting these kinds of new features into Conan 1.X, so unless those are bugs or strictly necessary for the migration, still better to target this to Conan 2.0. 

For those very large 10 Gbs projects in Conan 1.X you might try a workaround, to put the import/deployer functionality in the ``generate()`` method explicitly, or in a ``python_requires`` and use ``shutil.copy`` and ``filecmp.cmp`` explicitly, it will be a few lines of code, but relatively easy to do.

memsharded:
Hi @boussaffawalid 

This kind of new feature is no longer being added to 1.X releases, so I am closing this PR.

However, I think it is valid as a 2.0 feature, so if you want to open a PR for ``develop2`` for 2.X releases, I think it can be added to 2.X. Thanks!

PR #15149, opened 2023-11-21
Add `core:warnings_as_errors` configuration option

### LINKED ISSUE #14909: [feature] Add "warnings as errors" mode
### What is your suggestion?

For example, sometimes when i'm developing a package I get: "Warning: No files in package"...I would like that to show up as an error. Of course, I guess that also implies the need to suppress warnings - so perhaps it's not so simple.

AbrilRBS:
Changelog: Feature: Add `core:warnings_as_errors` configuration option to make Conan raise on warnings and errors.
Docs: https://github.com/conan-io/docs/pull/3484

First try for `core:warnings_as_errors`. It worries me that people might want to have Conan stop processing on errors in this case (We mostly do that now when we print proper errors as they are next to raised exceptions), but I'm not quite sure if I should just raise here too. Any opinions are welcomed :)

Closes #14909


conan/internal/conan_app.py
```diff
@@ -44,8 +44,12 @@ def __init__(self, cache_folder, global_conf):
 
         home_paths = HomePaths(self.cache_folder)
         self.hook_manager = HookManager(home_paths.hooks_path)
-        # Wraps an http_requester to inject proxies, certs, etc
+
+        ConanOutput.set_warnings_as_errors(global_conf.get("core:warnings_as_errors",
+                                                           default=False, check_type=bool))
```
memsharded: Maybe unify ``ConanOutput.configure(global_conf)``?

conan/api/output.py
```diff
@@ -204,11 +209,16 @@ def success(self, msg):
...
+            return (warn_tag is not None and warn_tag in self._silent_warn_tags
+                    or warn_tag is None and "unknown" in self._silent_warn_tags)
+        if self._conan_output_level <= LEVEL_WARNING or (self._warnings_as_errors and self._conan_output_level <= LEVEL_ERROR):
+            if _skip_warning():
                 return self
             warn_tag_msg = "" if warn_tag is None else f"{warn_tag}: "
-            self._write_message(f"WARN: {warn_tag_msg}{msg}", Color.YELLOW)
+            message_type = "ERROR" if self._warnings_as_errors else "WARN"
+            message_color = Color.RED if self._warnings_as_errors else Color.YELLOW
+            self._write_message(f"{message_type}: {warn_tag_msg}{msg}", message_color)
```
memsharded: Not raising?

memsharded (commented):
I was kind of expecting to raise an exception to produce a hard error, otherwise, it is just move error messages from yellow to read and WARN=>ERROR; but still can be skipped? And if we need to check for output, we can directly check the warnings.

conan/api/output.py
AbrilRBS: We need to check this in more depth. If we raise here, it begs the question as to why normal `error`s do not raise in some cases. That's the main reason I didn't do that in the first place.

Maybe we need to come up with the idea of `core:raise_on_all_errors`, or knowing which errors should raise on print or something? let's look into it

memsharded: ``error()`` messages should be there mostly for 2 reasons:
- They will end up raising an exception, but we want to print a bunch of error messages before raising the exception instead of creating a huge exception message, sometimes a bit more difficult to read
- They are legacy, something was bad, but we couldn't really raise because of risk of breaking users, but we still wanted to print a red error message.

So maybe we can assume that module the legacy, error messages are only printed before exceptions?

conan/internal/api/detect_api.py
```diff
@@ -264,18 +266,17 @@ def detect_compiler():
         if "gcc" in command or "g++" in command or "c++" in command:
             gcc, gcc_version = _gcc_compiler(command)
             if platform.system() == "Darwin" and gcc is None:
-                output.error("%s detected as a frontend using apple-clang. "
-                             "Compiler not supported" % command)
+                raise ConanException("%s detected as a frontend using apple-clang. "
```
memsharded: I think better not change this to exception always, only when warning-as-errors

conan/internal/api/detect_api.py
```diff
@@ -264,18 +266,17 @@ def detect_compiler():
...
-        if platform.system() == "Windows" and command.rstrip('"').endswith(("cl", "cl.exe")) \
-                and "clang" not in command:
+        if (platform.system() == "Windows" and command.rstrip('"').endswith(("cl", "cl.exe"))
+            and "clang" not in command):
             return _msvc_cl_compiler(command)
 
         # I am not able to find its version
-        output.error("Not able to automatically detect '%s' version" % command)
-        return None, None
+        raise ConanException("Not able to automatically detect '%s' version" % command)
```
memsharded: I think better not change this to exception always, only when warning-as-errors

conan/internal/cache/db/cache_database.py
```diff
@@ -13,8 +14,8 @@ class CacheDatabase:
 
     def __init__(self, filename):
         version = sqlite3.sqlite_version
-        if Version(version) < "3.7.11":  # Not an exception, in case some false positives
-            ConanOutput().error(f"Your sqlite3 '{version} < 3.7.11' version is not supported")
+        if Version(version) < "3.7.11":
+            raise ConanException(f"Your sqlite3 '{version} < 3.7.11' version is not supported")
```
memsharded: I think better not change this to exception always, only when warning-as-errors

conan/internal/deploy.py
```diff
@@ -77,7 +77,7 @@ def _deploy_single(dep, conanfile, output_folder, folder_name):
     except Exception as e:
         if "WinError 1314" in str(e):
             ConanOutput().error("full_deploy: Symlinks in Windows require admin privileges "
-                                "or 'Developer mode = ON'")
+                                "or 'Developer mode = ON'", error_type="context")
```
memsharded: What is a "context" error? The meaning is not evident to me

conans/client/userio.py
```diff
@@ -129,5 +129,5 @@ def request_boolean(self, msg, default_option=None):
             elif s.lower() in ['no', 'n']:
                 ret = False
             else:
-                self._out.error("%s is not a valid answer" % s)
```
memsharded, line 132: This can remain an error

memsharded (approved):

memsharded (approved):

franramirez688 (approved):

PR #15150, opened 2023-11-21
Add `redirect_stdout` to CMake integration methods

### LINKED ISSUE #15108: [question] Build log file generation in recipe
### What is your question?

Hello everyone,

I searched for that topic, but I couldn't find any result here.
Is it somehow feasible to generate a log file from the build in the build function of the Conan recipe? My personal attempts here weren't successfull, but I am also not the most skilled Python developer. The idea would be to generate a log file during the `conan create .` so that the log can be archived as part of the package itself.
Does anyone have any ideas here?

BR,
Karsten

AbrilRBS:
Changelog: Feature: Add `redirect_stdout` to CMake integration methods.
Docs: Omit

This PR focuses on adding the functionality to the CMake integrations only. We're happy to make them available in other integrations if there're requests for it from users

Closes #15108


conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1353,3 +1353,45 @@ def build(self):
...
+
+        def layout(self):
+            cmake_layout(self)
+
+        def generate(self):
+            CMakeToolchain(self).generate()
+
+        def build(self):
+            self.output.info(f"Metadata: {self.package_metadata_folder}")
+
```
memsharded: I'd probably try to simplify this test:
- generators = "CMakeToolchain"
- remove output.info()
- no ``layout()`` necessary
- Probably better decouple the test from the metadata feature. Just dump in the build folder. If build goes wrong no metadata is generated

conan/tools/cmake/cmake.py
```diff
@@ -254,3 +255,14 @@ def _cmake_log_levels_args(self):
         if log_level == "quiet":
             log_level = "error"
         return ["--loglevel=" + log_level.upper()] if log_level is not None else []
+
+    def _run(self, command, redirect_stdout=None, **kwargs):
+        if redirect_stdout is not None:
+            with open(redirect_stdout, "w") as f:
```
memsharded: this is overwriting, not appending log, isn't it? Is this the most expected behavior, or users passing the same file to 3 configure(), build() and install() will expect the logs to append?

conan/tools/cmake/cmake.py
```diff
@@ -112,9 +113,9 @@ def configure(self, variables=None, build_script_folder=None, cli_args=None):
 
         command = " ".join(arg_list)
         with chdir(self, build_folder):
-            self._conanfile.run(command)
+            self._run(command, redirect_stdout=redirect_stdout)
```
memsharded: Maybe we want the ``conanfile.run(..., stdout_file=<filepath>)`` interface that avoids the context ``with open() as f`` to pass it to ``conanfile.run()``?

memsharded (commented):
Just some quick ideas and feedback

conan/tools/cmake/cmake.py
AbrilRBS: When brainstorming it I realized that my intuition for that would be a total overwrite, yes. That does not mean that I'm right, so let's ask the team about what they think

czoido:
Maybe we could we consider an alternative syntax that's more aligned with the `conanfile.run()` method? Like:

```python
def build(self):
    cmake = CMake(self)
    cmake.configure()
    my_buf = StringIO()
    cmake.build(stdout=my_buf)
    # then you do whatever you want with the output
```

I know this won't output the results while it's running, but the we don't have to take decisions like how we handle the files, etc.

What do you think @memsharded, @RubenRBS ?

memsharded:
> I know this won't output the results while it's running, but the we don't have to take decisions like how we handle the files, etc.

I like the idea of being uniform with ``self.run()``, and I can probably agree that maybe the users should do what they want and not Conan. Maybe they even want no build output in CI to have their CI logs cleaner and check the stored build.logs if they want to check that.

memsharded (approved):

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1353,3 +1353,54 @@ def build(self):
...
+        def build(self):
+            cmake = CMake(self)
+
+            config_stdout, config_stderr = StringIO(), StringIO()
+            cmake.configure(stdout=config_stdout, stderr=config_stderr)
+            self.output.info(f"Configure stdout: {config_stdout.getvalue()}")
+            self.output.info(f"Configure stderr: {config_stderr.getvalue()}")
+
+            build_stdout, build_stderr = StringIO(), StringIO()
+            cmake.configure(stdout=build_stdout, stderr=build_stderr)
```
czoido:
```suggestion
            cmake.build(stdout=build_stdout, stderr=build_stderr)
```

conans/test/utils/tools.py
```diff
@@ -774,6 +774,10 @@ def assert_listed_binary(self, requires, build=False, test=False, test_package=F
             else:
                 raise AssertionError(f"Cant find {r}-{kind} in {reqs}")
 
+    def created_build_folder(self, ref):
```
memsharded: Not really used in the PR?

AbrilRBS: Oops leftover from previous iteration, good catch

conans/test/unittests/tools/cmake/test_cmake_presets_definitions.py
```diff
@@ -42,7 +42,7 @@ def test_cmake_cmake_program(conanfile):
 
 
 def test_cmake_make_program(conanfile):
-    def run(command):
+    def run(command, **kwargs):
```
AbrilRBS, line 45: Needed to make tests pass

memsharded (approved):

czoido (approved):

PR #15151, opened 2023-11-21
Allow querying the contents of settings.yml (and settings_user!) from ConfigAPI

### LINKED ISSUE #15125: [question][api] Access to `possible_values` settings after version 2.0.14
### What is your question?

We have a custom command for setting up profiles. Up to version 2.0.13 I had the following code to access to possible value for settings (settings.yml and settings_user.yml):
```python
conan_cache = ClientCache(conan_api.cache_folder)
... = conan_cache.settings.possible_values()
```

I think it broke with PR #14969.

Do you have any hint what's the new way to access this value? I had a look for the code that was in `cache.py` previously and it seems to have moved into `profiles.py` but the whole `_settings()` function there seems to be private.

AbrilRBS:
Changelog: Feature: Allow accessing the contents of `settings.yml` (and `settings_user`!) from `ConfigAPI`.
Changelog: Feature: Add builtin conf access from `ConfigAPI`.
Docs: Omit

Closes #15125

### DIFF
--- a/conan/api/subapi/config.py
+++ b/conan/api/subapi/config.py
@@ -72,8 +72,8 @@ class ConfigAPI:
                 save(global_conf_path, default_global_conf)
         return self._new_config
 
-    @staticmethod
-    def builtin_confs():
+    @property
+    def builtin_confs(self):
         return BUILT_IN_CONFS
 
     @property
--- a/conan/api/subapi/config.py
+++ b/conan/api/subapi/config.py
@@ -1,14 +1,17 @@
 import os
 import platform
 import textwrap
-
+import yaml
 from jinja2 import Environment, FileSystemLoader
 
 from conan import conan_version
+from conans.client.conf import default_settings_yml
 from conan.internal.api import detect_api
 from conan.internal.cache.home_paths import HomePaths
 from conan.internal.conan_app import ConanApp
-from conans.model.conf import ConfDefinition
+from conans.errors import ConanException
+from conans.model.conf import ConfDefinition, BUILT_IN_CONFS
+from conans.model.settings import Settings
 from conans.util.files import load, save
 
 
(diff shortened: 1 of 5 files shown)


conan/api/subapi/config.py
```diff
@@ -68,3 +71,47 @@ def global_conf(self):
                     """)
                 save(global_conf_path, default_global_conf)
         return self._new_config
+
+    @staticmethod
```
memsharded: Maybe a property0 Why settings_yaml is property and this is method? would be a bit non homogeneous UI.

conan/cli/commands/config.py
```diff
@@ -77,3 +76,12 @@ def config_show(conan_api, parser, subparser, *args):
...
     return conan_api.config.show(args.pattern)
+
+
+@conan_subcommand(formatters={"text": list_text_formatter, "json": default_json_formatter})
+def config_settings_yml(conan_api, parser, subparser, *args):
+    """
+    Get the content of the settings.yml file
+    """
+    parser.parse_args(*args)
+    return conan_api.config.settings_yml.possible_values()
```
memsharded: It might be useful, whenever I am checking for valid values, I don't want the full output. If not very readable on terminal, we might want to explore some syntax to select subsets.

conan/api/subapi/config.py
memsharded:
```suggestion
    @property
```
Seems this is breaking the tests, this might fix it

memsharded (approved):
Looks good!

conan/cli/commands/config.py
AbrilRBS: Yeah, I looked into that usecase when implementing it, but came to the conclusion as elsewhere that if `conan config settings-yml --format=json | jq .os` can do that for you, do we need to implement something for the text case too? I can totally pretty-print the text output, that's a good idea

memsharded: Sounds good, better than inventing a new argument or syntax

memsharded (commented):
Looks good to me

memsharded (approved):

AbrilRBS:
Will move the sertial_print outsude its current location as a separate PR

czoido:
I was thinking if instead of adding such an specific subcommand as `conan config settings-yml` we could do a more generic command such as `conan config view` that can accept other configuration files. As this is also adding the user settings and it's not the contents of the `settings.yml` directly maybe it would be something like this?

```
$ conan config view settings
# if we don't care that's not exactly the contents of the file we could do
$ conan config view settings.yml
```

Should we add another command that locates the config files? Like:

```
$ conan config locate settings.yml
```

AbrilRBS:
I've now simplified this PR to only the requested API changes, as we're not sure this is useful outside super specific cases - We'll be happy to reevaluate this if needs arise for this feature, but until then let's not complicate ourselves

memsharded (approved):

czoido (approved):

PR #15153, opened 2023-11-21
proposal for source() buildenv discussion

### LINKED ISSUE #15152: [question] Using tool_requires dependencies in source method
### What is your question?

I maintain a number of Conan packages for internal use by our team, and have recently been working to migrate them to Conan 2.x. Some of these packages have actions in their `source` method that require an external tool — for example:
- Using CVS to check out source files from legacy source control.
- Using CMake to run a time-consuming code generation step that's independent of the host context (and would therefore be redundant to repeat for each build configuration).
- Using NuGet to fetch prebuilt binaries. (I think that this step does belong in the `source` method, because each NuGet package contains a collection of binaries for every supported build configurations: the desired NuGet package is therefore the same no matter the settings of the Conan package being built, and it's redundant to fetch it again for each configuration.)

I've packaged these external tools (CVS, CMake, Nuget) as `tool_requires` dependencies; this worked well under Conan 1.x.

However, it seems that the changes to how `tool_requires` dependencies work in Conan 2.x mean that this is no longer possible. Specifically, now that the `tool_requires` dependencies are made available via the `VirtualBuildEnv` generator (which can't be called from the `source` method), I can't see a supported way to use those tools in the `source` method.

The recipes do remain correct if I move these steps out of the `source` method and into `build`, but doing so means that a lot of redundant work is repeated for each build configuration. In some cases, this increases the time taken to build a set of packages from a recipe by several hours.

Is there any other way to use this `tool_requires` functionality from the `source` method in Conan 2.x?

memsharded:
Changelog: Feature: Allow ``tool-requires`` to be used in ``source()`` method injecting environment.
Docs: https://github.com/conan-io/docs/pull/3493

Close https://github.com/conan-io/conan/issues/15152


conans/test/integration/build_requires/test_build_requires_source_method.py
```diff
@@ -0,0 +1,91 @@
...
+
+        class Pkg(ConanFile):
+            name = "pkg"
+            version = "0.1"
+            tool_requires = "tool/0.1"
+
+            def source(self):
+                cmd = "mytool.bat" if platform.system() == "Windows" else "mytool.sh"
+                with buildenv(self):
+                    self.run(cmd)
```
memsharded: I have tried this approach, as we discussed, but this doesn't seem possible.
The test fail at the local ``conan source`` command, while trying to execute this, because it doesn't have enough information to construct the ``buildenv``, the ``conan source`` command do not construct the graph or have dependencies at all. Inhibiting this so it doesn't do anything and it is no-op when it is a local ``conan source`` command is not straightforward at all.

memsharded (commented):
Pleas team, check the comment

conans/test/integration/build_requires/test_build_requires_source_method.py
AbrilRBS: This was my fear when first proposed. Do you have any preferred fallback for this case?

conan/tools/env/virtualbuildenv.py
```diff
@@ -12,21 +12,27 @@ def __init__(self, conanfile):
...
-            self.configuration = self.configuration.lower()
-        self.arch = conanfile.settings.get_safe("arch")
-        if self.arch:
-            self.arch = self.arch.lower()
+        self.configuration = None
+        self.arch = None
 
     @property
     def _filename(self):
+        if not self.configuration:
```
AbrilRBS, line 21: Let's keep this change even if no longer necessary - feel more correct to me this way

AbrilRBS (commented):
I'm happy about the current approach, but I know @franramirez688 will try to convince us tomorrow about using a opt-out instead, let's see what we decide upon before approving :)

czoido (approved):

franramirez688 (approved):

conan/tools/env/virtualbuildenv.py
```diff
@@ -8,9 +8,10 @@ class VirtualBuildEnv:
         .bat or .sh script
     """
 
-    def __init__(self, conanfile):
+    def __init__(self, conanfile, skip_auto_generate=True):
```
czoido: I was thinking in other alternatives for the name of the argument, maybe one of these or any variation, makes sense:

- `freeze`
- `hold`
- `no_refresh`
- `keep`
- `inmutable`

Anyway, I think that `skip_auto_generate` or just `auto_generate` are also good enough

memsharded: Lets do ``auto_generate``, I think it would be the most (relatively speaking) intuitive.

czoido (approved):

memsharded:
This opt-out is being changed in https://github.com/conan-io/conan/pull/15319 for 2.0.16, making it a opt-in

PR #15172, opened 2023-11-24
proposal for deploy() feature

### LINKED ISSUE #14811: [feature] Add a `deploy()` method to conanfile for application specific deployment logic
### What is your suggestion?

Hi Team,

# Bottom Line Up Front
### The proposal

Add a `deploy()` method to the conanfile, that would allow the author to specify very specific deployment instructions for applications where a more general deployment script is not appropriate. This might be invoked as follows:

```python
$ conan install --requires="myapp/1.0.0" --deploy /path/to/deployment/location
```
This does not necessarily need to conflict with `--deployer` (both may be useful for different applications, i.e. a custom deployer to aggregate licenses or generate a SBOM is not mutually exclusive with deployment for production usage).

### The use case
It would be good to have a method of writing a deployment mechanism for a single application. My specific use case here is for a GUI application that is unlikely to be invoked via the command line or with a virtual environment activated. Without knowing a heap about these applications, it might be useful to think of how an application such as Adobe Photoshop or Blender might be managed and deployed using conan.

For this case, these applications have the following properties:
* Highly specific project structure that needs to aggregate dependencies and store/read configuration from the deployed location
* Not useful as part of any developer process - this is an end-point deployment
* Unlikely to be executed from the command line
* Unlikely to use virtual environments (even though these are useful, some legacy applications won't play nicely)

The current method of doing this might be to use option (3) below, however it seems to me that this is at the wrong level of abstraction. If I have a number of applications like this, each would require their own deployer to live in the cache, which would necessitate distribution of every applications deployment recipe with global configuration. There is a UX issue here that could lead to one of these highly specific custom deployers being used to deploy packages they're not intended for, and would fail and throw a bunch of errors. It seems to me that for highly application specific deployment requirements, this deploy script/recipe is best placed with the conanfile of the applications being deployed itself.

# Context - A summary of alternatives

Currently, there are a few of methods that we have available for application deployment:

### 1. Don't deploy

For many use cases, especially when using build tools or other command line utilities, this may be sufficient:

```bash
conan install --tool_requires="cmake/[>3.25]"
conanbuild.bat
cmake.exe --version
# when you are done
deactivate_conanbuild.bat
``` 

And this works great when the application can be engaged from the command line. This does not satisfy the use case of system level deployment of a GUI application that expects to be bundled with its dependencies somewhere.

### 2. Built in deployers
Otherwise, for deployment, the build in deployers are great for simple applications or applications that adhere more to the unix philosophy in terms of application structure

```bash
$ conan install --requires="cmake/[>3.25]" --deployer=direct_deploy
$ tree direct_deploy -L 3
direct_deploy
└── cmake
    ├── bin
    │   ├── ccmake
    │   ├── cmake
    │   ├── cmake-gui
    │   ├── cpack
    │   └── ctest
    ├── conanbuildenv-x86_64.sh
    ├── conanbuild.sh
    ├── conaninfo.txt
    ├── conanmanifest.txt
    ├── conanrunenv-x86_64.sh
    ├── conanrun.sh
    ├── deactivate_conanbuild.sh
    ├── deactivate_conanrun.sh
    ├── licenses
    │   └── Copyright.txt
    └── share
        ├── aclocal
        ├── applications
        ├── bash-completion
        ├── cmake-3.25
        ├── emacs
        ├── icons
        ├── mime
        └── vim
```
You can invoke the generated environment files and use them temporarily, or you can add them to `PATH` in a more permanent way if desired. This may work excellently on unix like systems where these structures can be injected into the system easily.

However, it's not flexible enough to satisfy the use case of system level deployment of a GUI application that can't be easily invoked by command line, or where the deployment is targeting users that are not comfortable with the command line.

### 3. Custom global deployer

These are also great when built in deployers don't do what we need them to, and shine especially in situations where the deployer can perform a deployment that applies to a large number of packages. I.e., if this deploys `cmake` in a way that would be equally as valid as deploying `meson` or `protobuf`.

```bash
$ conan install --requires="cmake/[>3.25]" --deployer=mycustomdeployer
$ conan install --requires="meson/[>1.0.0]" --deployer=mycustomdeployer # Great! adheres to DRY principle and we get re-use of common deployment logic
```

This _does_ satisfy the use case - I _can_ write a highly specific custom deployer for my application here. The objection I have to this method is that it lives within the `$CONAN_HOME/extensions/deploy` directory and can be applied to other applications where it would not make sense. To illustrate, I could:

```bash
$ conan install --requires="myapp/[>3.25]" --deployer=myapp # This would work great
```
Which would work, but I could also:

```bash
$ conan install --requires="myotherapp/[>3.25]" --deployer=myapp # Error! This makes no sense
```

If I need to write a deployer that is highly specific to a single application, I think that there should be a way of coupling this with the package recipe rather than having to distribute multiple components (the deployer and the recipe) separately to make this possible.

memsharded:
Changelog: Feature: New ``deploy()`` method in recipes for explicit per-recipe deployment.
Docs: https://github.com/conan-io/docs/pull/3494

Close https://github.com/conan-io/conan/issues/14811

### DIFF
--- a/conan/cli/commands/install.py
+++ b/conan/cli/commands/install.py
@@ -33,7 +33,7 @@ def install(conan_api, parser, *args):
                         help='Deploy using the provided deployer to the output folder')
     parser.add_argument("--deployer-folder",
                         help="Deployer output folder, base build folder by default if not set")
-    parser.add_argument("--deployer-pkg", action="append",
+    parser.add_argument("--deploy-package", action="append",
                         help="Execute the deployment of the specific packages")
     parser.add_argument("--build-require", action='store_true', default=False,
                         help='Whether the provided path is a build-require')
@@ -72,7 +72,7 @@ def install(conan_api, parser, *args):
     conan_api.install.install_binaries(deps_graph=deps_graph, remotes=remotes)
     ConanOutput().title("Finalizing install (deploy, generators)")
     conan_api.install.install_consumer(deps_graph, args.generator, source_folder, output_folder,
-                                       deploy=args.deployer, deploy_pkg=args.deployer_pkg,
+                                       deploy=args.deployer, deploy_package=args.deploy_package,
                                        deploy_folder=args.deployer_folder)
     ConanOutput().success("Install finished successfully")
 
--- a/conan/cli/commands/install.py
+++ b/conan/cli/commands/install.py
@@ -33,7 +33,7 @@ def install(conan_api, parser, *args):
                         help='Deploy using the provided deployer to the output folder')
     parser.add_argument("--deployer-folder",
                         help="Deployer output folder, base build folder by default if not set")
-    parser.add_argument("--deploy-package", action="append",
+    parser.add_argument("--deployer-package", action="append",
                         help="Execute the deployment of the specific packages")
     parser.add_argument("--build-require", action='store_true', default=False,
                         help='Whether the provided path is a build-require')
@@ -72,7 +72,7 @@ def install(conan_api, parser, *args):
     conan_api.install.install_binaries(deps_graph=deps_graph, remotes=remotes)
     ConanOutput().title("Finalizing install (deploy, generators)")
     conan_api.install.install_consumer(deps_graph, args.generator, source_folder, output_folder,
-                                       deploy=args.deployer, deploy_package=args.deploy_package,
+                                       deploy=args.deployer, deploy_package=args.deployer_package,
                                        deploy_folder=args.deployer_folder)
     ConanOutput().success("Install finished successfully")
 
--- a/conan/cli/commands/install.py
+++ b/conan/cli/commands/install.py
@@ -33,6 +33,8 @@ def install(conan_api, parser, *args):
                         help='Deploy using the provided deployer to the output folder')
     parser.add_argument("--deployer-folder",
                         help="Deployer output folder, base build folder by default if not set")
+    parser.add_argument("--deployer-pkg", action="append",
+                        help="Execute the deployment of the specific packages")
     parser.add_argument("--build-require", action='store_true', default=False,
                         help='Whether the provided path is a build-require')
     args = parser.parse_args(*args)
@@ -70,7 +72,8 @@ def install(conan_api, parser, *args):
     conan_api.install.install_binaries(deps_graph=deps_graph, remotes=remotes)
     ConanOutput().title("Finalizing install (deploy, generators)")
     conan_api.install.install_consumer(deps_graph, args.generator, source_folder, output_folder,
-                                       deploy=args.deployer, deploy_folder=args.deployer_folder)
+                                       deploy=args.deployer, deploy_pkg=args.deployer_pkg,
+                                       deploy_folder=args.deployer_folder)
     ConanOutput().success("Install finished successfully")
 
     # Update lockfile if necessary
--- a/conans/test/integration/conanfile/test_deploy_method.py
+++ b/conans/test/integration/conanfile/test_deploy_method.py
@@ -1,0 +1,47 @@
+import textwrap
+
+from conans.test.utils.tools import TestClient
+
+
+def test_deploy_method():
+    c = TestClient()
+    conanfile = textwrap.dedent("""
+        import os
+        from conan import ConanFile
+        from conan.tools.files import copy, save
+        class Pkg(ConanFile):
+            name = "{name}"
+            version = "0.1"
+            {requires}
+            def package(self):
+                save(self, os.path.join(self.package_folder, f"my{name}file.txt"), "HELLO!!!!")
+            def deploy(self):
+                copy(self, "*", src=self.package_folder, dst=self.deploy_folder)
+            """)
+    c.save({"dep/conanfile.py": conanfile.format(name="dep", requires=""),
+            "pkg/conanfile.py": conanfile.format(name="pkg", requires="requires='dep/0.1'")})
+    c.run("create dep")
+    assert "Executing deploy()" not in c.out
+    c.run("create pkg")
+    assert "Executing deploy()" not in c.out
+
+    # Doesn't install by default
+    c.run("install --requires=pkg/0.1")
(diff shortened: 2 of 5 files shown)


czoido:
The only thing I'm not sure about is the naming of the argument. Perhaps something like ``--deploy-requires=...`` would be more aligned with the naming of other arguments?

AbrilRBS:
> Perhaps something like --deploy-requires=...

I don't agree with this one, I think requires would give the wrong impression here? Not sure though, happy to look-into it tomorrow :)

memsharded:
> The only thing I'm not sure about is the naming of the argument. Perhaps something like --deploy-requires=... would be more aligned with the naming of other arguments?

There might be an issue, this introduces some ambiguity with tool-requires, will there be a ``--deploy-tool-requires`` too?

czoido:
> > The only thing I'm not sure about is the naming of the argument. Perhaps something like --deploy-requires=... would be more aligned with the naming of other arguments?
> 
> There might be an issue, this introduces some ambiguity with tool-requires, will there be a `--deploy-tool-requires` too?

Then maybe `--deploy-package/s` instead `--deployer-pkg` ?

conan/cli/commands/install.py
```diff
@@ -33,6 +33,8 @@ def install(conan_api, parser, *args):
                         help='Deploy using the provided deployer to the output folder')
     parser.add_argument("--deployer-folder",
                         help="Deployer output folder, base build folder by default if not set")
+    parser.add_argument("--deploy-package", action="append",
```
AbrilRBS: While I understand that `--deploy-package` gives a better meaning of performing an action, I think that having symmetric names with the two options above feels cleaner? My suggestion would thus be `--deployer-package`

czoido: Yes, I think that's a valid point, not having to remember if it was _deploy_ or _deployer_ everytime so for me it's fine `--deployer-package`

memsharded: Sure, lets change it

memsharded: Done

czoido (approved):

franramirez688 (approved):

AbrilRBS (approved):

conan/cli/commands/install.py
```diff
@@ -33,6 +33,8 @@ def install(conan_api, parser, *args):
                         help='Deploy using the provided deployer to the output folder')
     parser.add_argument("--deployer-folder",
                         help="Deployer output folder, base build folder by default if not set")
+    parser.add_argument("--deployer-package", action="append",
+                        help="Execute the deployment of the specific packages")
```
samuel-emrys, line 37: Not sure if this is too late to amend, but I think that this is an ambiguous description
```suggestion
                        help="Execute the deploy() method of the named package references")
```

conans/test/integration/conanfile/test_deploy_method.py
```diff
@@ -0,0 +1,46 @@
...
+    # Doesn't install by default
+    c.run("install --requires=pkg/0.1")
+    assert "Executing deploy()" not in c.out
+
+    # Doesn't install with other patterns
+    c.run("install --requires=pkg/0.1 --deployer-package=other")
+    assert "Executing deploy()" not in c.out
+
+    # install can deploy all
+    c.run("install --requires=pkg/0.1 --deployer-package=* --deployer-folder=mydeploy")
```
samuel-emrys, line 37: Yep, I think that this is a great interface. You're right, we might want to deploy multiple packages at once. I'm wondering about overloading the `--deployer-folder` though - is the design intention here that each deployer is responsible for namespacing itself within the `deployer-folder`? For instance, if I run

```
$ conan install --requires=foo/0.1.0 --deployer-package=foo/0.1.0 --deployer=licenses --deployer=full_deploy --deployer-folder=/tmp/deploy-test
```
then each of the deployers (kind of) conflict:
```
$tree /tmp/deploy-test -L 2
/tmp/deploy-test
├── conaninfo.txt
├── conanmanifest.txt
├── full_deploy
│   └── host
├── include
│   └── foo.h
├── lib
│   └── libfoo.a
└── licenses.zip
```

`full_deploy` namespaces itself properly, licenses.py (from the conan-extensions repo) places licenses.zip in the root of the deployer folder, and in this case I've emulated your `deploy()` method, which copies the package folder to the root. If this is the most desirable design, then a best practice would need to emerge to avoid contaminating other deployments (i.e., by outputing to a subfolder of `deploy-folder`). Or is there another design that might remove the need for a best practice?

I guess removing the need for a best practice also has issues - this is the most flexible, and gives the user the most room to construct things the way they want, at the cost of a (potential) maintenance overhead in ensuring deployers deploy consistently if there's a possibility of a clash.

conan/cli/commands/install.py
memsharded: Agree, lets change it (in a new PR)

conans/test/integration/conanfile/test_deploy_method.py
memsharded: Yes, I think it should be the responsibility of the users to namespace their deploys if that is what they want/need. If we try to namespace something, then it would be certainly blocking for users that want to deploy all .dlls in a single common folder to all dependencies for example, and they want to do it in their ``deploy()`` method.

Once deployers are running from different origins and different packages ``deploy()`` now, the space for conflict is always there (they could be deploying directly to ``/usr/bin`` etc, and conflicting anyway). The ``full_deploy`` implement the namespace in itself, it is not something external, with the intention that users can take it as inspiration and easily customize the namespace/layout of the deployment to fit their needs.

The rule of thumb is that deployers copy to user space, and as such it is the full responsibiliy of the user (via the different deployers implementations) to control that user space. And then, yes, some common practices might happen with some common namespaces patterns, but that should be it, just some common practices, not something mandated or implemented by the tool.

conan/cli/commands/install.py
memsharded: Fixed

PR #15174, opened 2023-11-24
Removed conan migration compatibility warning

### LINKED ISSUE #15161: [feature] Supress Conan 2 compatibility message
### What is your suggestion?

When a recipe fails to build Conan 2 will print a migration note regarding v2 compatibility:

```
*********************************************************
Recipe 'conanfile.py (...)' cannot build its binary
It is possible that this recipe is not Conan 2.0 ready
If the recipe comes from ConanCenter check: https://conan.io/cci-v2.html
If it is your recipe, check if it is updated to 2.0
*********************************************************
```

This is confusing our developers though because they assume a Conan issue when in fact its a good old compiler failure right above that flashy message. Looking at the code I found now way to disable this - is there any plans to offer an option for that?

juansblanco:
Changelog: Feature: Removed a warning about a potential issue with conan migration that would print every time a build failed.
Docs: Omit

Removed the corresponding test that was checking that the error would print.

Closes #15161

### DIFF
--- a/conan/cli/cli.py
+++ b/conan/cli/cli.py
@@ -191,16 +191,6 @@ class Cli:
                     "If it is your recipe, check if it is updated to 2.0\n" \
                     "*********************************************************\n"
             ConanOutput().writeln(error, fg=Color.BRIGHT_MAGENTA)
-        result = re.search(r"(.*): Error in build\(\) method, line", message)
-        if result:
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
😁

memsharded (approved):

PR #15186, opened 2023-11-28
Add pyproject.toml

jcar87:
Changelog: Omit
Docs: Omit

Move to pyproject.toml from legacy setup.py

DRAFT PR - not ready for merging

Testing:
* running `pip install .` from the top of the repository should result in the installation of the same files into the virtual environment, from both `setup.py` and from `pyproject.toml`
* running `pip install -e .` should still work
* generating a wheel with `python -m pip wheel .` at the root level of the repository, should generate an equivalent wheel (same files) with both `setup.py` and `pyproject.toml`


conans/__init__.py
```diff
@@ -1,5 +1,7 @@
+import importlib.metadata
+
 CHECKSUM_DEPLOY = "checksum_deploy"  # Only when v2
 REVISIONS = "revisions"  # Only when enabled in config, not by default look at server_launcher.py
 OAUTH_TOKEN = "oauth_token"
 
-__version__ = '2.0.15-dev'
+__version__ = importlib.metadata.version("conan")
```
memsharded: I think this is only Python 3.8, we might need a bit of duplication until we move to 3.8

jcar87: tricky but doable - I think the version needs to be in a separate file altogether, that both `__init__.py` and `pyproject.toml` reference

jcar87:
My initial attempt was to replace `setup.py` and move entirely to `pyproject.toml`, as the structure is relative simple.
However, it appears the most recent version of `setuptools` that is available in Python 3.6 (which we still support), is unable to parse the metadata and ends up generating a package of name "UNKNOWN" and version `0.0.0`.

It appears that both can files can coexist, in the sense that `setuptools` runs either way. But it is unclear what information can be where (apparently some can even be in setup.cfg) - and what happens in case of duplication.
When the `pyproject.toml` was self-sufficient, but `setup.py` had additional metadata, I got some warnings pointing me in the direction of adding them to pyproject.toml - so clearly both files are being considered. 

In the end I've opted for a minimal `pyproject.toml` that only contains the `[build-system]` section - this more or less emulates `pip` fallback behaviour of what happens when a `pyproject.toml` is not present: https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour 
Also consistent with the advice here: https://scikit-hep.org/developer/guidelines#packaging-required

Tools have continued to evolve and it is now possible for a setuptools-based project to rely only on pyproject.toml - but this is only supported in newer versions of setuptools (beyond what's supported on Python 3.6). Once we don't have to support Python 3.6 we can probably do away with setup.py entirely and rely on metadata defined in pyproject.toml.

PR #15192, opened 2023-11-30
Add environment information to CMakePresets

### LINKED ISSUE #14271: [feature] CMakeToolChain should add environment section to CMakePresets.json
### What is your suggestion?

Hi! I would like for Conan to be able to insert the environment variables defined in env_info and the env section of a profile into the CMakePresets.json so that CMake uses these environment variables when building.

According to https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html an environment section can be added to the config preset.

In my case, I need an environment variable setting to access the license for the compiler, which I am adding as a Conan package with a toolchain file. I am setting the required env var in the profile to include the required compiler, but the license is not found. When I added manually to the generated CMakePresets.json, then the license was found indicating that CMake had used it.

I would like this to happen automatically. I would rather not use the virtualenv generator, as it adds an extra step for my developers to remember - whereas this approach is working well with the modern approach you have defined for working with Conan and Cmake (and using VSCode's CMake plugin).

Note that I am currently using Conan 1.59.0 (and attempting to define recipes that are conan 2.x ready)

Attached:
[profile.txt](https://github.com/conan-io/conan/files/12012574/profile.txt) - The profile with a env_var defined
[CMakePresets.json.txt](https://github.com/conan-io/conan/files/12012582/CMakePresets.json.txt) - Modified CMakePresets.json file with the additional environment section I would like to see added.

### LINKED ISSUE #15012: [feature] Consider passing the buildenv via CMake Presets
### What is your suggestion?

CMake presets allow setting the environment that is "activated" during the CMake run: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html#id4

For example:
```
      "environment": {
        "MY_ENVIRONMENT_VARIABLE": "Test",
        "PATH": "$env{HOME}/ninja/bin:$penv{PATH}"
      },
```

Currently, in Conan it is advised to 'activate' the run environment:
```
source ./Release/generators/conanbuild.sh
generators\conanbuild.bat
# (and the one for powershell)
```

if we are able to gather the information and express it in the generated CMake presets by Conan, the workflow could be the same for all platforms:
```
conan install .
cmake --preset conan-release
cmake --build --preset conan-release
```

and it would do TheRightThing :tm: 

this would narrow the gap of having to run different commands on Windows / macOS / Linux depending on whether the generator is multi-generator, without multigenerator, or whether we are using windows cmd prompt, powershell, or bash.  Just some quality of life imprvement

czoido:
Changelog: Feature: Add build environment information to CMake configure preset and run environment information to CMake test presets.
Docs: https://github.com/conan-io/docs/pull/3488

Closes: https://github.com/conan-io/conan/issues/15012
Closes: https://github.com/conan-io/conan/issues/14271


conans/test/integration/toolchains/cmake/test_cmaketoolchain.py
```diff
@@ -717,6 +717,27 @@ def _format_val(val):
     assert cache_variables["CMAKE_MAKE_PROGRAM"] == "MyMake"
 
 
+def test_variables_types():
```
czoido, line 720: moved from unit to integration, but testing the same

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -214,8 +215,23 @@ def generate(self):
             else:
                 cache_variables[name] = value
 
+        # Get environment information for the presets
+        # FIXME: check if we have an alternative to avoid this way of getting this
+
+        prev_status = self._conanfile.virtualbuildenv
+        build_env = VirtualBuildEnv(self._conanfile).vars()
+        self._conanfile.virtualbuildenv = prev_status
```
memsharded: This pattern is becoming annoying, yes, we should check what other possibilities there could be to avoid this.

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -214,8 +215,23 @@ def generate(self):
...
+        buildenv = {name: value for name, value in
+                    build_env.items(variable_reference="$penv{{{name}}}")}
```
memsharded: Is this something to fix? This looks necessary, the placeholder really differs for different usage (cmd, powershell), and this is not known by VirtualBuildEnv until variable generation time.

czoido: Sorry, no, just for the way of getting the environment, this is ok, I'll change the comment to clarify

memsharded (commented):
This is amazingly simple and powerful, good job.

I think we need to consider a bit the UX, user expectations, defaults, etc., just in case, but lets move this forward.

memsharded (approved):

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -214,8 +215,27 @@ def generate(self):
             else:
                 cache_variables[name] = value
 
+        buildenv, runenv = None, None
+
+        if self._conanfile.conf.get("tools.cmake.cmaketoolchain:presets_environment", default="",
+                                    check_type=str, choices=("disabled", "")) != "disabled":
```
franramirez688, line 221: Could it be clearer to use a boolean value instead?

```suggestion
        if self._conanfile.conf.get("tools.cmake.cmaketoolchain:disabled_presets_environment", default=False,
                                    check_type=bool):
```

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -1424,3 +1424,117 @@ def package(self):
...
+    if platform.system() != "Windows":
+        c.run_command("cmake --preset conan-debug")
+        assert "MY_BUILD_VAR:MY_BUILDVAR_VALUE" in c.out
+        assert "MY_RUNVAR NOT FOUND" in c.out
+
+    c.run_command("cmake --build --preset conan-debug --target run_mytool --target test_env")
+    assert "running: mytool/0.1" in c.out
+
+    c.run_command("ctest --preset conan-debug")
+    assert "tests passed" in c.out
```
franramirez688, line 1540: It reads fine, but IMO it'd be great if we could split this test into three different ones:
* one for conf
* one for buildenv/configure
* one for runenv/tests

franramirez688 (approved):

conan/tools/cmake/toolchain/toolchain.py
czoido: Yes, that was intentional, in case in the future we let users generate the environments in the presets with different modes (like for example adding env to configure preset but not to test or inhibiting generating environment in the scripts...)

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
czoido: Maybe we can keep this for the moment, as the PR is already green, and probably refactor when we add more tests in the future if we extend this feature?

conan/tools/env/virtualrunenv.py
```diff
@@ -29,13 +29,14 @@ class VirtualRunEnv:
         .bat or .sh script
     """
 
-    def __init__(self, conanfile):
+    def __init__(self, conanfile, auto_generate=False):
```
czoido, line 32: adding the same argument that was added here: https://github.com/conan-io/conan/pull/15153

jcar87 (approved):
This is great!!!

In practice - given that the CMakeToolchain adds all the `bindirs` of the dependencies in the `build` context to `CMAKE_PROGRAM_PATH` - that already covers the most used case for environment variables that would otherwise make activating `conanbuild.sh` a requirement.

But I've found a useful practical case - e.g.


`conanfile.txt`:
```
[requires]
fmt/10.1.1

[tool_requires]
mold/2.0.0

[generators]
CMakeDeps
CMakeToolchain

[layout]
cmake_layout
```

`CMakeLists.txt`:
```
cmake_minimum_required(VERSION 3.26)

string(APPEND CMAKE_CXX_FLAGS_INIT " -fuse-ld=mold")
project(hello LANGUAGES CXX)

message("CMAKE_CXX_FLAGS=${CMAKE_CXX_FLAGS}")

find_package(fmt REQUIRED)

add_executable(hello hello.cpp)

target_link_libraries(hello PRIVATE fmt::fmt)
```

note how CMake doesn't really let you control which linker to use, we do this via the compiler with `-fuse-ld=mold` - in this case, mold needs to be in the `PATH` (as seen by the compiler), in order to work.

With the changes proposed in this PR:
```
cmake --preset conan-release
cmake --build --preset conan-release
```

This works without activating any environment :D Without the changes in this PR, configuring CMake fails unless the `conanbuild.sh` is activated.

Massive win !!!

db4:
Can this feature be backported to Conan v1?

memsharded:
Hi @db4 
I am afraid it is not possible, Conan 2.0 is now 1 year old (2.5 years since beta), there are not enough resources to backport this kind of new features anymore, and the solution is straightforward, keep calling the environment scripts to activate the environment.

PR #15196, opened 2023-11-30
Fix `conan graph info --format=html` reporting misleading conflicting nodes

AbrilRBS:
Changelog: Bugfix: Fix `conan graph info --format=html` reporting misleading conflicting nodes.
Docs: Omit

The tests didn't catch this because I only tested with one level, which left the ids the same!

Don't merge yet, will add a test to ensure this keeps working in the future

### DIFF
--- a/conan/cli/formatters/graph/graph.py
+++ b/conan/cli/formatters/graph/graph.py
@@ -54,7 +54,8 @@ class _PrinterGraphItem(object):
 class _Grapher(object):
     def __init__(self, deps_graph):
         self._deps_graph = deps_graph
-        self.nodes, self.edges, self.node_map = self._build_graph()
+        self.node_map, self.edges = self._build_graph()
+        self.nodes = self.node_map.values()
 
     def _build_graph(self):
         graph_nodes = self._deps_graph.by_levels()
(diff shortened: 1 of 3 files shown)


conan/cli/formatters/graph/info_graph_html.py
```diff
@@ -56,7 +56,7 @@
                         font: { color: "{% if highlight_node %}white{% else %}black{% endif %}" },
                         fulllabel: '<h3>{{ node.label }}</h3>' +
                                    '<ul>' +
-                                   '    <li><b>id</b>: {{ node.package_id }}</li>' +
+                                   '    <li><b>package id</b>: {{ node.package_id }}</li>' +
```
AbrilRBS, line 59: Unrelated change, but felt that id was too general in this case

conan/cli/formatters/graph/graph.py
```diff
@@ -73,7 +73,7 @@ def _build_graph(self):
                 dst = _node_map[node_to]
                 edges.append((src, dst))
 
-        return _node_map.values(), edges
+        return _node_map.values(), edges, _node_map
```
AbrilRBS: Could probably simplify this to only return the node_map and the edges, will do so later today

AbrilRBS:
New look: 
image

memsharded (approved):
Looks good

franramirez688 (approved):

PR #15204, opened 2023-12-01
reusing fixture, will reduce CI time

memsharded:
Changelog: Omit
Docs: Omit

There is much more that we can do, and specially if we want to keep expanding functional test without incurring in very slow test suite.


AbrilRBS (approved):
Oh, I like the idea 👍

czoido (approved):

PR #15207, opened 2023-12-03
Feature/cmakedeps legacy adapter

### LINKED ISSUE #14623: [feature] Bring back CMake targets generator for Conan 2
### What is your suggestion?

The new way of injecting dependency information into CMake is a lot more verbose, complex and error prone on the CMake side than conan 1's `conan_basic_setup(TARGETS)`.

1. Need to use find_package for each dependency instead of a single include. (more verbose)
2. The name in find_package is often slightly different, e.g. ZLIB (CMake) instead of zlib (conan). (error prone, complex)
3. The find scripts often do things in a non package-manager friendly way via extra options, e.g. boosts autolinking option. Conan 1's way neatly sidestepped that unnecessary complexity. (error prone, complex, verbose)
4. find_package falls back to looking for system libs silently if not configured correctly (error prone, complex)

I'd very much like to have an option to import targets the conan 1 way with conan 2, be it as an option or a separate generator. I can see how the new way works better for 'unintrusive' conan support in libraries. But I maintain several applications that are specifically built on conan with 10-20 dependencies that are a huge pain to port to conan 2 right now.

memsharded:
Changelog: Feature: Make ``CMakeDeps`` generator create a ``conandeps.cmake`` that aggregates all direct dependencies in a ``cmake``-like generator style.
Docs: https://github.com/conan-io/docs/pull/3492

Close https://github.com/conan-io/conan/issues/14623


AbrilRBS (approved):

franramirez688 (approved):

czoido (approved):

conans/test/functional/toolchains/cmake/cmakedeps/test_cmakedeps_aggregator.py
```diff
@@ -0,0 +1,44 @@
...
+                cmake.build()
+                self.run(os.path.join(self.cpp.build.bindir, "app"))
+        """)
+
+    cmakelists = textwrap.dedent("""
+        cmake_minimum_required(VERSION 3.15)
+        set(CMAKE_CXX_COMPILER_WORKS 1)
+        set(CMAKE_CXX_ABI_COMPILED 1)
+        project(app CXX)
+        include(${CMAKE_BINARY_DIR}/generators/conandeps.cmake)
```
jcar87: I would probably suggest renaming this to `conandeps_legacy.cmake` or something similar - to reinforce the message that this is compatible with a legacy behaviour.

conan/tools/cmake/cmakedeps/cmakedeps.py
```diff
@@ -162,3 +186,38 @@ def get_find_mode(self, dep):
...
+
+        template = textwrap.dedent("""\
+            message(STATUS "Conan: Using CMakeDeps conandeps.cmake aggregator via include()")
+            message(STATUS "Conan: It is recommended to use explicit find_package() per dependency instead")
+
+            {% for config in configs %}
+            find_package({{config.file_name}})
+            {% endfor %}
+
+            add_library(conandeps INTERFACE)
```
jcar87: Consider using a variable like the original generator, where the value of the variable is a list of targets
https://docs.conan.io/1/integrations/build_system/cmake/cmake_generator.html#global-variables-approach

(I'd also consider using a name like `CONAN_DEPS_LEGACY` or something like that)

PR #15211, opened 2023-12-04
allow hard set_name() set_version()

### LINKED ISSUE #15208: [question] usage of options in set_version
### What is your question?

Hi

I would like to create packages with and with our a timestamp in the version number.
So I try to use an option called with_timestamp and read that option in the set_version function.
But this does not work as expected.

this is my recipe:

```
import os
import pdb

from conan import ConanFile
from conan.tools.files import load
from datetime import datetime, timezone

class ComponentConanFile(ConanFile):
    name = "demo"
    version = "1.2.3"    
    options = {
        "with_timestamp" : [True, False]  
        }

    def config_options(self):   
        if self.options.with_timestamp == None:
            self.options.with_timestamp = True
    
    def set_version(self):
        #breakpoint()
        orig_version = self.version
        self.output.info("ORIG: " + orig_version)
        try:
            option_dict = vars(self.options.with_timestamp)
            print("Content of self.options.with_timestamp:")
            print(option_dict)
            if self.options.with_timestamp:
                current_datetime = datetime.utcnow().replace(tzinfo=timezone.utc)
                formatted_datetime = current_datetime.strftime("%Y%m%d-%H%M")
                print(formatted_datetime)
                self.version = orig_version + "-" + formatted_datetime
                print(self.version)
        except Exception as e:
            self.output.error(e)

```

Is it the intended behaviour that options are not available within set_version? 

memsharded:
Changelog: Fix: Allow hard ``set_name()`` and ``set_version()`` to mutate name and version provided in command line.
Docs: https://github.com/conan-io/docs/pull/3491

Close https://github.com/conan-io/conan/issues/15208


hubertba:
Would be a good enhancement 👍

AbrilRBS (approved):

memsharded:
Discussed with the team, we will be enabling this as a new feature for 2.0.15

czoido (approved):

PR #15212, opened 2023-12-04
Fix `conan graph explain` binary distance calculation

AbrilRBS:
Changelog: Omit
Docs: Omit

We were calculating the number of _keys_, which was either 2 or 0, not by how much it was different.
There might be a tad more elegant solution but I like this one being straightforward

Can add a test if the team thinks it's necessary

### DIFF
--- a/conan/api/subapi/list.py
+++ b/conan/api/subapi/list.py
@@ -267,8 +267,10 @@ class _BinaryDistance:
 
     @property
     def distance(self):
-        return len(self.platform_diff), len(self.settings_diff), \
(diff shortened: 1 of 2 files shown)


memsharded (commented):
This was definitely a bug. It worked fine before moving to the ``expected``, ``existing`` keys, because it measured the size of the diff. 

I think yes, we need to add one test to make sure this is covered and doesn't happen again. It can be just one, in one dimension, no need to do a super thorough and costly test of all combinations.

AbrilRBS:
Done, please re-check

memsharded (approved):

PR #15222, opened 2023-12-05
Fix range escaping in conflict reports involving ranges

### LINKED ISSUE #14793: [bug] Version ranges in `graph info -f=html` are doubly escaped
### Environment details

* Conan version: 2.0.12-dev

### Steps to reproduce

1. Run `graph info ... -f=html` with a graph that contains conflicting version ranges
image

### Logs

_No response_

AbrilRBS:
Changelog: Fix: Fix range escaping in conflict reports involving ranges.
Docs: Omit

Closes https://github.com/conan-io/conan/issues/14793

This makes it so that the string is parsed to json which escapes the ranges to unicode codes before having a chance to be automatically HTML escaped by Jinja2.

Captura de pantalla 2023-12-06 a las 0 05 15

### DIFF
--- a/conan/cli/formatters/graph/info_graph_html.py
+++ b/conan/cli/formatters/graph/info_graph_html.py
@@ -82,11 +82,13 @@ graph_info_html = """
                 // Add error conflict node
                 nodes.push({
                     id: "{{ error["type"] }}",
-                    label: "{{ error["context"].require.ref }}",
+                    label: {{ error["context"].require.ref|string|tojson }},
                     shape: "circle",
                     font: { color: "white" },
                     color: "red",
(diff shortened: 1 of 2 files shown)


conan/cli/formatters/graph/info_graph_html.py
```diff
@@ -82,11 +82,13 @@
                 // Add error conflict node
                 nodes.push({
                     id: "{{ error["type"] }}",
-                    label: "{{ error["context"].require.ref }}",
+                    label: {{ error["context"].require.ref|string|tojson }},
```
AbrilRBS, line 85: Note that the `tojson` function adds proper quoting to the label, so that they are not needed in the js code itself

memsharded: I dont know enough of this lib, but it seems it makes sense 👍

memsharded (approved):
Looks good to me

PR #15223, opened 2023-12-06
Sketch to show that populating download cache on conan upload is possible

AbrilRBS:
Just to show that it is technically possible, even though right now is hideous - just wanted to have the PR up to quickly show it around.

This PR won't get merged, and it doesn't either mean that proper work on the feature will for sure happen, I just wrote this because I got bored :)


conans/client/cmd/uploader.py
```diff
@@ -244,6 +245,18 @@ def upload_package(self, pref, prev_bundle, remote):
...
+        for cache_file in cache_files:
+            self._local_cache_file(pref, cache_file, remote)
+
+    def _local_cache_file(self, pref, cache_file, remote):
+        approved_files = ["conaninfo.txt", "conan_package.tgz", "conanmanifest.txt"]
+        if cache_file not in approved_files:
+            return
+        _download_cache = self._app.global_conf.get("core.download:download_cache")
+        if _download_cache and os.path.isabs(_download_cache):
+            url = self._app.remote_manager._auth_manager._get_rest_client(remote)._get_api().router.package_file(pref, cache_file)
```
AbrilRBS: 🫣

conans/client/cmd/uploader.py
```diff
@@ -244,6 +245,18 @@ def upload_package(self, pref, prev_bundle, remote):
         self._app.remote_manager.upload_package(pref, cache_files, remote)
         duration = time.time() - t1
         self._output.debug(f"Upload {pref} in {duration} time")
+        for cache_file in cache_files:
+            self._local_cache_file(pref, cache_file, remote)
+
+    def _local_cache_file(self, pref, cache_file, remote):
+        approved_files = ["conaninfo.txt", "conan_package.tgz", "conanmanifest.txt"]
```
memsharded: Recipe files have to be cached too, ``conanfile.py``, ``conan_export.tgz``, etc

conans/client/cmd/uploader.py
memsharded: 😱 😂

conans/client/cmd/uploader.py
```diff
@@ -244,6 +245,18 @@ def upload_package(self, pref, prev_bundle, remote):
...
+        if cache_file not in approved_files:
+            return
+        _download_cache = self._app.global_conf.get("core.download:download_cache")
+        if _download_cache and os.path.isabs(_download_cache):
+            url = self._app.remote_manager._auth_manager._get_rest_client(remote)._get_api().router.package_file(pref, cache_file)
+            cached_path = DownloadCache(_download_cache).cached_path(url)
```
memsharded: Is someone really doing a copy to the cache? I am missing it. This would be the critical part, my biggest concern about this is the locking, the download cache is concurrent-safe

AbrilRBS: No, only printint for now, ran out of time (Got too sleepy) before being able to add that - will iterate further next week in between other tasks

memsharded (commented):
Needs:
- More testing
- Understand what the filters checking the possible files were doing, maybe we need to filter metadata files out?

This should be a bit cleaner

memsharded (commented):
I am re-thinking this.
Even if from the performance point of view, it makes sense, on the other hand if something goes wrong in the server, there is some misconfiguration or whatever, that makes the upload finish OK, but not possible to download that file again from the server, maybe forcing at least the first download is not a bad thing to guarantee everything works? Or maybe the likelihood of that happening is so low, the performance criteria wins?

AbrilRBS:
Closed because https://github.com/conan-io/conan/pull/20163 implemented this

Issue #15241, opened 2023-12-08
[bug] Cannot specify compiler.toolset=v142 when it is not the default


popowskm:
### Environment details

* Operating System+version: Windows 11 Pro
* Compiler+version: msvc 193
* Conan version: 2.0.9
* Python version: 3.11.3

### Steps to reproduce

1. Set compiler.toolset=v142 in your conan profile
2. Attempt to run a conan command, i.e. `conan install .` 
* Expectation: conan attempts to build the .props files and other outputs
* Result: conan throws an error saying that 'v142' is not a valid toolset value

Note: This can be resolved by editing settings.yml to include 'v142' in the enumerated list of options for msvc.toolset, but it requires freezing settings.yml and must be done manually on new installations.

#### Original
```
$ conan install .
ERROR: Invalid setting 'v142' is not a valid 'settings.compiler.toolset' value.
Possible values are ['v110_xp', 'v120_xp', 'v140_xp', 'v141_xp']
Read "http://docs.conan.io/2/knowledge/faq.html#error-invalid-setting"
```
#### After editing settings.yml
```
$ conan install .

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.toolset=v142
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.toolset=v142
compiler.version=193
os=Windows


======== Computing dependency graph ========
Graph root
...
```

memsharded:
Hi  @popowskm 

Thanks for your feedback.

It seems that the profile is not correct. The way to define VS 2022 + toolset v142 is by:

- ``compiler.version=192``. The ``compiler.version`` is the main input here, it defines the compiler version irrespective of the VS IDE version. The v142 toolset of VS 2022 is the same as the default toolset of VS 2019. The binary compatibility only depends on the ``compiler.version`` for msvc compiler, not on the IDE version.
- If you want to use another VS IDE version different to the compiler default, then define the ``conf`` ``tools.microsoft.msbuild:vs_version=17`` (17-2022, 16-2019, etc). This conf is not part of the ``package_id`` by default.

Please use that and let us know.

popowskm:
[...]

PR #15250, opened 2023-12-11
better ux for VS not found error

memsharded:
Changelog: Omit
Docs: Omit

For https://github.com/conan-io/conan/issues/15241

I haven't added a test for this, not that easy to reproduce without mocking, too much for the value of just an error msg.

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -234,7 +234,7 @@ def _vcvars_path(version, vs_install_path):
     if not vs_path or not os.path.isdir(vs_path):
         raise ConanException(f"VS non-existing installation: Visual Studio {version}. "
                              "If using a non-default toolset from a VS IDE version consider "
-                             "'tools.microsoft.msbuild:vs_version' conf to specify it")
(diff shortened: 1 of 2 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -232,7 +232,9 @@ def _vcvars_path(version, vs_install_path):
     # TODO: This comes from conans/client/tools/win.py vcvars_command()
     vs_path = vs_install_path or vs_installation_path(version)
     if not vs_path or not os.path.isdir(vs_path):
-        raise ConanException("VS non-existing installation: Visual Studio %s" % version)
+        raise ConanException(f"VS non-existing installation: Visual Studio {version}. "
+                             "If using a non-default toolset from a VS IDE version consider "
+                             "'tools.microsoft.msbuild:vs_version' conf to specify it")
```
AbrilRBS:
```suggestion
                             "If using a non-default toolset from a VS IDE version consider "
                             "specifying it with the 'tools.microsoft.msbuild:vs_version' conf")
```

AbrilRBS (approved):
👍

PR #15257, opened 2023-12-12
[command][cache path] Raising error if folder path does not exist

franramirez688:
Changelog: Feature: `conan cache path xxx --folder xxxx` raises an error if the folder requested does not exist.
Docs: omit

### DIFF
--- a/conan/api/subapi/cache.py
+++ b/conan/api/subapi/cache.py
@@ -25,43 +25,43 @@ class CacheAPI:
         app = ConanApp(self.conan_api.cache_folder, self.conan_api.config.global_conf)
         ref.revision = None if ref.revision == "latest" else ref.revision
         ref_layout = app.cache.recipe_layout(ref)
-        return ref_layout.export()
+        return _check_folder_existence(ref, "export", ref_layout.export())
 
     def recipe_metadata_path(self, ref: RecipeReference):
         app = ConanApp(self.conan_api.cache_folder, self.conan_api.config.global_conf)
         ref = _resolve_latest_ref(app, ref)
         ref_layout = app.cache.recipe_layout(ref)
-        return ref_layout.metadata()
+        return _check_folder_existence(ref, "metadata", ref_layout.metadata())
 
     def export_source_path(self, ref: RecipeReference):
         app = ConanApp(self.conan_api.cache_folder, self.conan_api.config.global_conf)
         ref.revision = None if ref.revision == "latest" else ref.revision
         ref_layout = app.cache.recipe_layout(ref)
-        return ref_layout.export_sources()
+        return _check_folder_existence(ref, "export_sources", ref_layout.export_sources())
 
     def source_path(self, ref: RecipeReference):
         app = ConanApp(self.conan_api.cache_folder, self.conan_api.config.global_conf)
         ref.revision = None if ref.revision == "latest" else ref.revision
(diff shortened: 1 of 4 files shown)


conan/api/subapi/cache.py
```diff
@@ -16,6 +16,27 @@
...
+
+def _resolve_latest_pref(app, pref):
+    pref.ref = _resolve_latest_ref(app, pref.ref)
+    if pref.revision is None or pref.revision == "latest":
+        pref.revision = None
+        result = app.cache.get_latest_package_reference(pref)
+        if result is None:
+            raise ConanException(f"'{pref.repr_notime()}' not found in cache")
+        pref = result
+    return pref
```
memsharded: Has something changed here, or is this only just a move for these functions? Why moving them? private implementations better at the end of the file not the top, specially for APIs

conan/cli/commands/cache.py
```diff
@@ -31,30 +33,34 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
...
+    elif args.folder == "source":
+        path = conan_api.cache.source_path(ref)
+    elif args.folder == "metadata":
+        path = conan_api.cache.recipe_metadata_path(ref) if ref else \
+               conan_api.cache.package_metadata_path(pref)
+    elif args.folder == "build":
+        path = conan_api.cache.build_path(pref)
+    # Assuming that pref or ref would have raised an exception if args.folder is None
+    if args.folder and not os.path.exists(path):
+        reference = f'recipe reference {ref}' if ref else f'package reference {pref}'
```
memsharded: Better not change this, this looks like a potential refactor, but it hides the actual behavior change of raising the errors when folders do not exist.

memsharded (changes requested):

conan/cli/commands/cache.py
```diff
@@ -46,6 +48,8 @@ def cache_path(conan_api: ConanAPI, parser, subparser, *args):
             path = conan_api.cache.recipe_metadata_path(ref)
         else:
             raise ConanException(f"'--folder {args.folder}' requires a valid package reference")
+        if args.folder and not os.path.exists(path):
+            raise ConanException(f"'{args.folder}' folder does not exist for the reference {ref}")
```
memsharded: I think behavior like this should belong to the API, not the command layer?

(side note, it seems message in the above line 50 is wrong)

franramirez688: This check seems to be something particular about this command, should the API ensure that the folder exists? Not sure either.

> (side note, it seems message in the above line 50 is wrong)

No, it's not wrong but confusing. That's why I made that refactor.

memsharded: > This check seems to be something particular about this command, should the API ensure that the folder exists? Not sure either.

The problem could be the API returning the ``source`` folder path if it doesn't exist. What could be the use case? If the user writes something to that folder, that is undefined behavior, and Conan will not execute the ``source()`` method when necessary, because it thinks the sources are already (I know users shouldn't write in the cache in any case, but it is more likely to happen if the API gives them a folder path that they can write)

> No, it's not wrong but confusing. That's why I made that refactor.

Yes, make sense to try to improve it later in another PR (not urgent)

memsharded (approved):

czoido (approved):

PR #15260, opened 2023-12-12
Make paths format in layout consistent on Windows

juansblanco:
Changelog: Omit
Docs: Omit

The paths in `source_folder`, `build_folder` and `generators_folder` can have different path formats depending on how they are set (layout, output-folder, etc). When accessing those values from the output json the resulting path would sometimes apply the `normpath` function. This would yield different results in Windows producing inconsistencies in the resulting paths. (This only happens in Windows because `normpath` trasforms all `/` to `\`)

For example, using `cmake_layout()` produced a path with `\`, while using `output-folder` or no setting produced a path with `/`.

This changes makes those paths to always use the `normpath` so the result is consistent (it should always be `\` on Windows). Note that the paths in the output json will still be either `\\` or `/ ` but when extracting the value it will always be `\`.

### DIFF
--- a/conans/model/layout.py
+++ b/conans/model/layout.py
@@ -90,7 +90,7 @@ class Folders(object):
         if self._base_source is None:
             return None
         if not self.source:
-            return self._base_source
+            return os.path.normpath(self._base_source)
 
         return os.path.normpath(os.path.join(self._base_source, self.source))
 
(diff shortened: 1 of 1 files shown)


conans/model/layout.py
```diff
@@ -90,7 +90,7 @@ def source_folder(self):
         if self._base_source is None:
             return None
         if not self.source:
-            return self._base_source
+            return os.path.normpath(self._base_source)
```
AbrilRBS, line 93: Is this pattern repeated elsewhere?

juansblanco: It's fixed for all the cases found here.

AbrilRBS (approved):

jcar87 (approved):
Good catch!

PR #15262, opened 2023-12-13
allow to copy & paste from compact format into conan-lock-add

memsharded:
Changelog: Feature: Allow copy&paste of recipe revisions with timestamps from ``--format=compact`` into ``conan lock add``.
Docs: https://github.com/conan-io/docs/pull/3533

### DIFF
--- a/conan/cli/commands/list.py
+++ b/conan/cli/commands/list.py
@@ -79,7 +79,7 @@ def print_list_text(results):
             for k, v in item.items():
                 if isinstance(v, dict) and v.get("timestamp"):
                     timestamp = v.pop("timestamp")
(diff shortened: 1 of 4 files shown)


AbrilRBS (approved):

AbrilRBS (changes requested):
After speaking with the team, let's try to have both the %float formatter and the human readable timestamp in the output, see how that looks

AbrilRBS:
This now looks something like this:
image

I've also added it to the normal --format=text output, I think this is ok

memsharded:
Looks good, I think we can move this forward with this (tests broken). How does the normal --format=text looks like on screen?

AbrilRBS:
Captura de pantalla 2024-01-04 a las 12 24 55

PR #15272, opened 2023-12-13
adding tools.microsoft:winsdk_version conf

### LINKED ISSUE #14383: [bug] It's impossible to set winsdk_version argument for vcvars script
### Environment details

* Windows 11
* MSVC 2022
* Conan version: 2.0.9
* Python version: 3.10

### Steps to reproduce

The generated `conanvcvars.bat` contains the following line

```
set "VSCMD_START_DIR=%CD%" && call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC/Auxiliary/Build/vcvarsall.bat"  amd64 -vcvars_ver=14.3
```

`winsdk_version` is not set at all.
For some legacy code I need to set older SDK version for MSVC among several installed.
Please, correct me if I wrong, but I looked through the documentation and python scripts and it's looks like it's impossible to set it.

Thank you.

### Logs

_No response_

memsharded:
Changelog: Feature: Adding ``tools.microsoft:winsdk_version`` conf to make ``VCVars`` generator to use the given ``winsdk_version``.
Docs: https://github.com/conan-io/docs/pull/3487

Close https://github.com/conan-io/conan/issues/14383

### DIFF
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -788,15 +788,11 @@ class GenericSystemBlock(Block):
 
         return system_name, system_version, system_processor
 
-    def context(self):
-        generator = self._toolchain.generator
-        generator_platform = self._get_generator_platform(generator)
-        toolset = self._get_toolset(generator)
-        system_name, system_version, system_processor = self._get_cross_build()
-
-        # This is handled by the tools.apple:sdk_path and CMAKE_OSX_SYSROOT in Apple
-        cmake_sysroot = self._conanfile.conf.get("tools.build:sysroot")
-        cmake_sysroot = cmake_sysroot.replace("\\", "/") if cmake_sysroot is not None else None
+    def _get_winsdk_version(self, system_version, generator_platform):
+        compiler = self._conanfile.settings.get_safe("compiler")
+        if compiler not in ("msvc", "clang") or "Visual" not in str(self._toolchain.generator):
+            # Ninja will get it from VCVars, not from toolchain
+            return system_version, None, None
 
         winsdk_version = self._conanfile.conf.get("tools.microsoft:winsdk_version", check_type=str)
         if winsdk_version:
@@ -815,6 +811,21 @@ class GenericSystemBlock(Block):
         gen_platform_sdk_version = [generator_platform,
                                     f"version={winsdk_version}" if winsdk_version else None]
         gen_platform_sdk_version = ",".join(d for d in gen_platform_sdk_version if d)
+
+        return system_version, winsdk_version, gen_platform_sdk_version
+
+    def context(self):
+        generator = self._toolchain.generator
+        generator_platform = self._get_generator_platform(generator)
+        toolset = self._get_toolset(generator)
+        system_name, system_version, system_processor = self._get_cross_build()
+
+        # This is handled by the tools.apple:sdk_path and CMAKE_OSX_SYSROOT in Apple
+        cmake_sysroot = self._conanfile.conf.get("tools.build:sysroot")
+        cmake_sysroot = cmake_sysroot.replace("\\", "/") if cmake_sysroot is not None else None
+
+        result = self._get_winsdk_version(system_version, generator)
+        system_version, winsdk_version, gen_platform_sdk_version = result
 
         return {"toolset": toolset,
                 "generator_platform": generator_platform,
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -804,7 +804,7 @@ class GenericSystemBlock(Block):
                 self._conanfile.output.warning("Both cmake_system_version and winsdk_version confs"
                                                " defined, prioritizing winsdk_version")
             system_version = winsdk_version
-        elif "Windows" in self._conanfile.get_safe("os", ""):
+        elif "Windows" in self._conanfile.settings.get_safe("os", ""):
             winsdk_version = self._conanfile.settings.get_safe("os.version")
             if system_version:
                 if winsdk_version:
--- a/conan/tools/cmake/toolchain/blocks.py
+++ b/conan/tools/cmake/toolchain/blocks.py
@@ -577,7 +577,7 @@ class ExtraFlagsBlock(Block):
                                                             check_type=bool)
         if android_ndk_path and (cxxflags or cflags) and android_legacy_toolchain is not False:
             self._conanfile.output.warning("tools.build:cxxflags or cflags are defined, but Android NDK toolchain may be overriding "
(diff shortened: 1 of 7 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -132,12 +132,14 @@ def generate(self, scope="build"):
             vcvars_ver = _vcvars_vers(conanfile, compiler, vs_version)
         vcvarsarch = _vcvars_arch(conanfile)
 
+        winsdk_version = conanfile.conf.get("tools.microsoft:winsdk_version", check_type=str)
+        winsdk_version = winsdk_version or conanfile.setttings.get_safe("os.version")
```
jcar87: ~I think this might cause issues if an SDK that matches the os version is not installed - but also, I would advise against making this the default.
Last time I checked, the default behaviour of vcvars was to pick up "the most recent" version, unless told otherwise - I'd try and keep that behaviour. We did have a a lot of issues in Conan Center when CMake was picking up a version that matches the os.version and not the latest. Recently CMake has added a policy to NOT behave like that: https://cmake.org/cmake/help/latest/policy/CMP0149.html#policy:CMP0149~

Edit: nevermind, thought we were asking the current OS version, but this is the Conan profile!!

memsharded: This will only be used if:
- Users adds ``version`` to Windows subsettings in ``settings.yml``
- Defines ``os.version`` in their profiles

I have kept ``os.version`` instead of ``sdk_version`` because other Windows oss, like CE in ``settings.yml`` already have a ``version`` subsetting that represents the sdk

jcar87 (approved):
LGTM. Although makes me wonder whether we want the same setting to apply to CMake when using the Visual Studio generators?

memsharded:
> LGTM. Although makes me wonder whether we want the same setting to apply to CMake when using the Visual Studio generators?

yes, I think so, if you have some insights how to do it, that would be appreciated.

jcar87:
> > LGTM. Although makes me wonder whether we want the same setting to apply to CMake when using the Visual Studio generators?
> 
> yes, I think so, if you have some insights how to do it, that would be appreciated.

* if cmake policy CMP0149 is with new behaviour, and the CMake generator is visual studio, we need to pass `version=` in the CMake generator platform (see https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_PLATFORM.html#variable:CMAKE_GENERATOR_PLATFORM)
* else (prior to cmake 3.27 or with the policy set to old behaviour)
we need to set `CMAKE_SYSTEM_VERSION` to the version of the SDK

conan/tools/microsoft/visual.py
memsharded: Added integrations for VCVars, MSBuildToolchain (so opening directly from IDE works) and CMakeToolchain

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -656,10 +659,24 @@ class GenericSystemBlock(Block):
...
 
-        {% if generator_platform %}
+        {% if generator_platform and not winsdk_version %}
         set(CMAKE_GENERATOR_PLATFORM "{{ generator_platform }}" CACHE STRING "" FORCE)
+        {% elif winsdk_version %}
+        if(POLICY CMP0149)
+            cmake_policy(GET CMP0149 _POLICY_WINSDK_VERSION)
+        endif()
+        if(_POLICY_WINSDK_VERSION STREQUAL "NEW")
+            message(STATUS "Conan toolchain: CMAKE_GENERATOR_PLATFORM={{gen_platform_sdk_version}}")
```
jcar87, line 669: I think we need to take into account that this is only useful for the "Visual Studio" CMake generators - I'm assuming this block doesn't appear for things like Ninja? (where the SDK version is selected when activating vcvars)

conans/test/functional/toolchains/cmake/test_cmake_toolchain.py
```diff
@@ -555,6 +555,20 @@ def test_cmake_toolchain_runtime_types_cmake_older_than_3_15():
...
+def test_cmake_toolchain_winsdk_version():
+    client = TestClient(path_with_spaces=False)
+    client.run("new cmake_lib -d name=hello -d version=0.1")
+    # cmake = client.load("CMakeLists.txt")
+    # cmake = cmake.replace("cmake_minimum_required(VERSION 3.15)",
+    #                       "cmake_minimum_required(VERSION 3.27)")
+    # client.save({"CMakeLists.txt": cmake})
+    client.run("create . -s arch=x86_64 -s compiler.version=193 "
+               "-c tools.microsoft:winsdk_version=8.1")
+    assert "Conan toolchain: CMAKE_GENERATOR_PLATFORM=x64" in client.out
```
jcar87, line 573: In the CMake lists we can print the value of `CMAKE_VS_WINDOWS_TARGET_PLATFORM_VERSION` - which is a read-only variable of the version of the SDK that CMake selected - although I'm not 100% what happens if you choose a version that you don't have installed

memsharded: If you pick a version that you don't have installed, with the OLD policy, it will fallback to the latest SDK

memsharded: I'll add that check, good idea.

conan/tools/cmake/toolchain/blocks.py
memsharded: Good catch, adding a check that avoids winsdk_verison in toolchain if generator!=Visual

jcar87 (approved):
LGTM!!

conan/tools/cmake/toolchain/blocks.py
```diff
@@ -781,12 +824,17 @@ def context(self):
         cmake_sysroot = self._conanfile.conf.get("tools.build:sysroot")
         cmake_sysroot = cmake_sysroot.replace("\\", "/") if cmake_sysroot is not None else None
 
+        result = self._get_winsdk_version(system_version, generator)
```
mocabe, line 827: @memsharded 
Forgive me if I'm wrong: should this be `self._get_winsdk_version(system_version, generator_platform)` instead?
I have following conf in profile: 
```
tools.cmake.cmaketoolchain:generator=Visual Studio 17 2022
tools.cmake.cmaketoolchain:system_version=10.0.22000
```
and now I get `set(CMAKE_GENERATOR_PLATFORM "Visual Studio 17 2022,version=10.0.22000" CACHE STRING "" FORCE)` in `conan_toolchain.cmake`, which seems incorrect.

memsharded: Hi @mocabe 

Thanks very much for reporting this. Lets have a look.
As a general hint, it is better to create a new ticket (issue) that links to this, that comments in closed PRs, because these might pass unnoticed. (I'll create a ticket for this)

PR #15281, opened 2023-12-14
Fix for tests affected by path normalization change

juansblanco:
Changelog: Omit
Docs: Omit

Fix for tests broken by path changes in  https://github.com/conan-io/conan/pull/15260

### DIFF
--- a/conans/test/unittests/tools/files/test_patches.py
+++ b/conans/test/unittests/tools/files/test_patches.py
@@ -43,9 +43,9 @@ def test_single_patch_file(mock_patch_ng):
     conanfile.folders.set_base_export_sources("/my_source")
     conanfile.display_name = 'mocked/ref'
     patch(conanfile, patch_file='patch-file')
-    assert Path(mock_patch_ng.filename) == Path('/my_source/patch-file')
+    assert mock_patch_ng.filename.replace('\\', '/') == '/my_source/patch-file'
     assert mock_patch_ng.string is None
-    assert Path(mock_patch_ng.apply_args[0]) == Path('/my_source')
+    assert mock_patch_ng.apply_args[0].replace('\\', '/') == '/my_source'
     assert mock_patch_ng.apply_args[1] == 0
     assert mock_patch_ng.apply_args[2] is False
 
@@ -57,7 +57,7 @@ def test_single_patch_file_from_forced_build(mock_patch_ng):
     patch(conanfile, patch_file='/my_build/patch-file')
     assert mock_patch_ng.filename == '/my_build/patch-file'
     assert mock_patch_ng.string is None
-    assert Path(mock_patch_ng.apply_args[0]) == Path('/my_source')
+    assert mock_patch_ng.apply_args[0].replace('\\', '/') == '/my_source'
     assert mock_patch_ng.apply_args[1] == 0
     assert mock_patch_ng.apply_args[2] is False
 
@@ -69,7 +69,7 @@ def test_base_path(mock_patch_ng):
     conanfile.folders.source = "src"  # This not applies to find the patch file but for applying it
     conanfile.display_name = 'mocked/ref'
     patch(conanfile, patch_file='patch-file', base_path="subfolder")
-    assert Path(mock_patch_ng.filename) == Path('my_source/patch-file')
+    assert mock_patch_ng.filename.replace('\\', '/') == 'my_source/patch-file'
     assert mock_patch_ng.string is None
     assert mock_patch_ng.apply_args == (os.path.join("my_source", "src", "subfolder"), 0, False)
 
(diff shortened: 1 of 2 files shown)


conans/test/unittests/tools/files/test_patches.py
```diff
@@ -42,10 +43,11 @@ def test_single_patch_file(mock_patch_ng):
     conanfile.folders.set_base_export_sources("/my_source")
     conanfile.display_name = 'mocked/ref'
     patch(conanfile, patch_file='patch-file')
-    assert mock_patch_ng.filename.replace("\\", "/") == '/my_source/patch-file'
+    assert Path(mock_patch_ng.filename) == Path('/my_source/patch-file')
```
memsharded: The check with ``Path`` is problematic, as it doesn't compare the exact path, so ``Path("/home/lib") == Path("\home\lib")`` is ``True``. If we want to check that the paths contains exactly one slash direction, then this might not be the best check?

conans/test/unittests/tools/gnu/autotools_test.py
```diff
@@ -28,8 +28,8 @@ def test_source_folder_works(chdir_mock):
     assert conanfile.command.replace("\\", "/") == '"/path/to/sources/configure" -foo bar '
 
     autotools.autoreconf(build_script_folder="subfolder")
-    chdir_mock.assert_called_with(autotools, os.path.join("/path/to/sources", "subfolder"))
+    chdir_mock.assert_called_with(autotools, os.path.normpath(os.path.join("/path/to/sources", "subfolder")))
 
     autotools.autoreconf()
-    chdir_mock.assert_called_with(autotools, "/path/to/sources")
-    assert conanfile.command.replace("\\", "/") == 'autoreconf -bar foo'
+    chdir_mock.assert_called_with(autotools, os.path.normpath("/path/to/sources"))
```
memsharded, line 34: Something similar here. I think that some of this checks want to have a certain slash orientation, it is likely that autotools, even if in Windows, it might expect to receive forward slashes? If that is the case, this would be hiding this fact.

memsharded (commented):
Some doubts about checks being agnostic of the ``\\`` vs ``/``

conans/test/unittests/tools/files/test_patches.py
jcar87: Judging by the previous state of this file, I think this is telling:

```
assert mock_patch_ng.filename.replace("\\", "/") == '/path/to/sources/patch-file'
```
☝️  this assert clearly expects that paths may have double `\\` and replaces it before the comparison
but

```
assert mock_patch_ng.apply_args == ("/path/to/sources", 23, True)
```
☝️  this one doesn't do anything with paths 

From intuition this tells me that the test isn't really intending or concerned with exact string comparison, and it doesn't care about the direction of the slashes as long as it is composed in a correct way.

The fact that some asserts were like that, but others were like:
```
    assert mock_patch_ng.apply_args[0] == os.path.join("/my_build", "subfolder").replace("\\", "/")
    assert mock_patch_ng.apply_args[1] == 0
    assert mock_patch_ng.apply_args[2] is False
```

This tells me that prior authors when writing these tests simply adapted the test to whatever output Conan was giving at that particular point in time, and that the discrepancy in slash direction ended up unnecessarily expressed in the tests, when it is not the concern of the test.

I think using `Path` is not incorrect - my perception from this test is that it is not concerned with exact slash directions so long as it is correct and can be understood by the underlying tools (in particular the patch functionality). `Path` thus serves as an agnostic way (since it's more "immune" to slash direction) - and we can rely on Path doing the right thing internally. If in the future we wanted to change the slashes to forward slashes (while still being valid), then the same tests would still work - which is what I would have expected to begin with! 


On the other hand, and following the principle of doing what's already there - @juansblanco  - I would suggest replicating what was already being done here: https://github.com/conan-io/conan/blob/9d916b2ed021a0b5ef4512fd3ced1ed2a1f43194/conans/test/unittests/tools/files/test_patches.py#L81-L83 for the exact same case.

juansblanco: The problem we have is that now Windows and Linux/MacOS output different paths in some tests. @jcar87 suggested using `Path` as a better way to solve having fixed paths in asserts. It's true what you said maybe we don't want to use it, I thin I can still fix it with the `normpath` and later discuss if we want to use `Path`.

juansblanco: Yeah I needed to assert the values separately anyways. Changing the `replace` for `Path` is the extra step. So for now I shouldn't use `Path` and stick to the `replace` as it is being used, right?

conans/test/unittests/tools/gnu/autotools_test.py
juansblanco: In this case, I think it's just replicating what the current new functionality is like. In Windows the source path will always come with `\\`, so maybe we should check if the path changes aren't interfering with Autotools, if not the test should be correct.

conans/test/unittests/tools/files/test_patches.py
jcar87: I'd say so - I think we missed that earlier, that there were already asserts in the same file breaking down the tuple to replace the slashes in the first element.- I'd have suggested doing what the file was already doing elsewhere

juansblanco:
I reverted the Path changes so there should be just the minimum changes needed for the fix. I'll check tomorrow if we need to do more work in the Autotools part.

conans/test/unittests/tools/gnu/autotools_test.py
jcar87: yeah I agree that in this case it's important to check there are no regressions - since autotools on Windows might be sensitive to path formats

jcar87 (approved):

jcar87:
Merging this now to unblock tests. Have opened this: https://github.com/conan-io/conan/issues/15291 to follow up

PR #15282, opened 2023-12-15
new CMake.ctest() method

### LINKED ISSUE #14798: [feature] Add better CTest support to CMake helper
### What is your suggestion?

Right now, the `CMake` helper has a `test()` method that uses the `test`/`RUN_TESTS` target for running tests. However, CTest has a lot of additional options that cannot be configured with just environment variables (see https://cmake.org/cmake/help/latest/manual/ctest.1.html).

For example, it does not seem possible to set the `--output-junit` parameter that way. But there are a lot more parameters that cannot be used, e.g. excluding and include tests, repeated runs, etc.

I think it would be nice to have an additional `ctest()` method that uses CTest directly instead of the limited method above. It should be possible to use presets anyway, so calling e.g. `ctest --preset conan-debug` on the command already starts CTest with the correct preset.

memsharded:
Changelog: Feature: New ``CMake.ctest()`` helper method to launch directly ``ctest`` instead of via ``cmake --target=RUN_TEST``.
Docs: Omit

This is documented in code docstrings

Close https://github.com/conan-io/conan/issues/14798

### DIFF
--- a/conans/test/functional/toolchains/cmake/test_cmake.py
+++ b/conans/test/functional/toolchains/cmake/test_cmake.py
@@ -7,7 +7,6 @@ import unittest
 import pytest
 from parameterized.parameterized import parameterized
 
-from conans.model.recipe_ref import RecipeReference
 from conans.test.assets.cmake import gen_cmakelists
 from conans.test.assets.sources import gen_function_cpp, gen_function_h
 from conans.test.functional.utils import check_vs_runtime, check_exe_run
@@ -624,6 +623,7 @@ class TestCmakeTestMethod:
                     cmake.configure()
                     cmake.build()
                     cmake.test()
+                    cmake.ctest()
             """)
 
         cmakelist = textwrap.dedent("""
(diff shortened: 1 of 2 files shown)


jcar87 (approved):
LGTM!!

conans/test/functional/toolchains/cmake/test_cmake.py
```diff
@@ -648,7 +648,8 @@ def build(self):
 
         # The create flow must work
         c.run("create . --name=pkg --version=0.1 -pr:b=default -o test*:shared=True")
-        assert "1/1 Test #1: example ..........................   Passed" in c.out
+        assert str(c.out).count("1/1 Test #1: example ..........................   Passed") == 2
+        assert "pkg/0.1: RUN: ctest --build-config Release --parallel"
```
PeteAudinate, line 652: Looks like this assertion can never fail, as it should end with `in c.out`.

Thanks for adding this feature though :+1:

PR #15284, opened 2023-12-16
new lock remove command

### LINKED ISSUE #14523: [feature] ability to update a reference to an earlier version in a conan lockfile
### What is your suggestion?

From: https://cpplang.slack.com/archives/C41CWV9HA/p1692362444235349

At the moment, it is possible to update a lockfile to a _new_ version (and also update the revision), but with `conan lock add` it's not possible to add an earlier version in a meaningful way (it would not resolve to that one).

Some options:
* `conan lock remove`  + `conan lock add` + `conan lock create` 
* `conan lock update` (some challenges: lockfile may have different reversions/revisions of the same recipe, which one gets updated, consider a pattern)

### LINKED ISSUE #14524: [feature] ability to add reference to lockfile for a specific version and revision
### What is your suggestion?

At the moment it is super easy to recalculate a conanlock to update everything to the latest, and it's also easy to update a specific reference via:

Update everything to latest:
```
conan install . --lockfile="" --lockfile-out=conan.lock --update
```

Update a specific new version (where `conan.lock` was previously pinned to an earlier version)
```
conan lock add --requires=openssl/3.1.2
conan lock create . 
```

But if we want to update/change a specific reference to a specific version/revision that may _not_ be the latest, it is not as easy and a user may have to get the list from `conan list` (or the website if this information is available), and manually edit the `conan.lock` file.

memsharded:
Changelog: Feature: New ``conan lock remove`` command to remove requires from lockfiles.
Docs: https://github.com/conan-io/docs/pull/3496

I am proposing this PR to close:

Close https://github.com/conan-io/conan/issues/14523
Close https://github.com/conan-io/conan/issues/14524

The idea is that updates are equivalent to something like:
```bash
$ conan remove --requires=zlib/* && conan add --requires=zlib/1.2.11
```
But it will allow us to first iterate and learn about the logic of removal of versions from lockfiles, because ``conan remove`` makes sense in isolation, even without ``lock update`` or ``lock add``, like for example using it before resolving it to latest with ``--lockfile-partial``. When we have clear what is the UX and UI for the removal, we can do a single command ``conan update`` or ``conan add`` with extra arguments

### DIFF
--- a/conans/test/integration/lockfile/test_user_overrides.py
+++ b/conans/test/integration/lockfile/test_user_overrides.py
@@ -1,4 +1,7 @@
 import json
+import textwrap
+
+import pytest
 
 from conans.test.assets.genconanfile import GenConanfile
 from conans.test.utils.tools import TestClient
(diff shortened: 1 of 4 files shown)


conans/test/integration/lockfile/test_user_overrides.py
```diff
@@ -211,3 +214,105 @@ def test_lock_add_error():
...
+        ("--requires=math/*", ["math"]),
+        ("--requires=math/2.0", []),
+        ("--build-requires=cmake/1.0", ["cmake"]),
+        # Not valid ("--build-requires=*", ["cmake", "ninja"]),
+        ("--build-requires=*/*", ["cmake", "ninja"]), # But this is valid
+        ("--python-requires=mytool/*", ["mytool"]),
+        ("--python-requires=*tool/*", ["mytool", "othertool"]),
+        # With version ranges
+        ('--requires="math/[>=1.0 <2]"', ["math"]),
+        ('--requires="math/[>1.0]"', []),
```
AbrilRBS, line 230: I dont think it works, but what about `--requires="*/[>=1.0]"`?

AbrilRBS (approved):
Nice! Only have 1 doubt about a pattern case, otherwise looks good to me

conans/test/integration/lockfile/test_user_overrides.py
memsharded: No, that doesn't work yet, the name has to match for version ranges, I thought it didn't make sense to remove certain range for all versions, but the truth is that the patterns without version ranges can do it, so I'll make it the same and accept patterns too for the name.

memsharded: Done, please check

AbrilRBS (approved):

czoido (approved):

jcar87 (approved):

PR #15285, opened 2023-12-16
fix self.info.clear() for python_requires and tool_requires

memsharded:
Changelog: Bugfix: Make ``self.info.clear()`` and header-only packages to remove ``python_requires`` and ``tool_requires``.
Docs: https://github.com/conan-io/docs/pull/3485

### DIFF
--- a/conans/model/info.py
+++ b/conans/model/info.py
@@ -394,6 +394,8 @@ class ConanInfo:
         self.options.clear()
         self.requires.clear()
         self.conf.clear()
+        self.build_requires.clear()
+        self.python_requires.clear()
 
     def validate(self):
         # If the options are not fully defined, this is also an invalid case
--- a/conans/test/integration/package_id/package_id_test.py
+++ b/conans/test/integration/package_id/package_id_test.py
@@ -2,7 +2,9 @@ import textwrap
 
 import pytest
 
+from conans.test.assets.genconanfile import GenConanfile
 from conans.test.utils.tools import NO_SETTINGS_PACKAGE_ID, TestClient, TestServer
+from conans.util.files import save
 
 
 def test_double_package_id_call():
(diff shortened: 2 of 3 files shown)


AbrilRBS (approved):
😶

SSE4:
might it be a breaking change? seems like it can change package ID for existing recipes and packages

memsharded:
> might it be a breaking change? seems like it can change package ID for existing recipes and packages

Might be changing some package-ids, but it is a bug, and can be fixed, even if that is the case.

However, it is unlikely to hit, and this is the reason it hasn't been detected so far:

- It requires a header-only to use python-requires, which by the simple nature of header-only libs, is very uncommon
- It requires a header-only to use tool-requires (more likely but still unusual, just maybe when it comes injected from the profile), but also changing the package ``build_mode``, which is even more unlikely

SSE4:
I read a [stability commitment](https://docs.conan.io/2/introduction.html#stability) documentation, and it's kinda unclear:
```
Moving forward to following minor versions 2.1, 2.2, …, 2.X should never break existing recipes, packages or command line flows
If something is breaking, it will be considered a regression and reverted.
Bug fixes will not be considered breaking, recipes and packages relying on the incorrect behavior of such bugs will be considered already broken.
```
so I see, if you consider some behavior buggy (retrospectively or retroactively), you are always free to change it, and it's not breaking (that gives pretty much freedom to break things, actually).
about `self.info.clear()` in particular, right now it's not fully clear for reader, because docs never actually explain an effect on `self.python_requires`.
relevant paragraphs I found only say:
```
We have a package_id() method calling self.info.clear(). This is internally removing the settings from the package ID calculation so we generate only one configuration for our header-only library.
```
```
 For that case, there’s a simplified way of declaring that the generated package ID should not take into account settings, options or any information from the requirement which is using the self.info.clear() method inside another recipe method called package_id()
```
so doc mainly talks about settings and options.
at very least, documentation should be much more explicit on what does `self.info.settings.clear` actually do.

memsharded:
> Bug fixes will not be considered breaking.

This is a clear bug. ``self.info.clear()`` should completely clear the ``info``, this was always the intent, the expected behavior, and even the automatic ``implements`` relies on it. In the docs it says "or any information from the requirement", that might need some clarification, but that is the intended meaning of that sentence.

czoido (approved):

PR #15287, opened 2023-12-17
Backport `<host_version:track ref>` to Conan v1

AbrilRBS:
Changelog: Feature: Add tracking of different ref to `<host_version>`.
Changelog: Bugfix: Make `<host_version>` only track for same user/channel.
Docs: https://github.com/conan-io/docs/pull/3575

Need to check with the team if we want the bugfix. Conan v2 behaves similarly

### DIFF
--- a/conans/test/integration/build_requires/build_requires_test.py
+++ b/conans/test/integration/build_requires/build_requires_test.py
@@ -716,7 +716,6 @@ class TestBuildTrackHost:
         c.save({"protobuf/conanfile.py": GenConanfile("protobuf"),
                 "pkg/conanfile.py": pkg})
         c.run(f"create protobuf 1.0@{requires_tag}")
-        c.run(f"create protobuf 1.0@{tool_requires_tag}")
 
         c.run("create pkg", assert_error=fails)
         if fails:
--- a/conans/test/integration/build_requires/build_requires_test.py
+++ b/conans/test/integration/build_requires/build_requires_test.py
@@ -616,7 +616,7 @@ class TestBuildTrackHost:
         c.save({"conanfile.py":
                 GenConanfile("pkg").with_build_requirement("protobuf/<host_version>")})
         c.run(f"install . {build_profile}", assert_error=True)
-        assert "'protobuf/<host_version>': didn't find a matching host dependency" in c.out
+        assert "pkg/None require 'protobuf': didn't find a matching host dependency" in c.out
 
     @pytest.mark.parametrize("build_profile", [False, True])
     def test_track_host_error_wrong_context(self, build_profile):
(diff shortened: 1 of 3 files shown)


franramirez688 (approved):

conans/test/integration/build_requires/build_requires_test.py
```diff
@@ -616,7 +616,7 @@ def test_track_host_error_nothost(self, build_profile):
         c.save({"conanfile.py":
                 GenConanfile("pkg").with_build_requirement("protobuf/<host_version>")})
         c.run(f"install . {build_profile}", assert_error=True)
-        assert "'protobuf/<host_version>': didn't find a matching host dependency" in c.out
+        assert "pkg/None require 'protobuf': didn't find a matching host dependency" in c.out
```
memsharded: Isn't this output more confusing that it was before?

memsharded (approved):

conans/test/integration/build_requires/build_requires_test.py
AbrilRBS: Yeah, let me fix that, I think we can make it a bit clearer 👍

AbrilRBS: Done, please re-check

SSE4:
what is `track ref` in the first place? searching in conan docs provides zero results for that term. and why is it needed to be added to 1.x?

AbrilRBS:
> what is track ref in the first place?

You're right that the PR title is not the most obvious thing (Note that the proper changelog goes into the `Changelog` entry in the PR body which I think does make sense) - It's meant to represent the ability to track a different ref other than the one being required in the `<host_version>` instance.

> and why is it needed to be added to 1.x?

We need to add this to V1 because CCI needs this fix for libgettext/gettext - and as we're still building v1 & v2 there, backporting this made sense :)

SSE4:
> We need to add this to V1 because CCI needs this fix for libgettext/gettext - and as we're still building v1 & v2 there, backporting this made sense :)

it's still not clear from the PR body Changelog entry, which says `Bugfix: Make <host_version> only track for same user/channel`. AFAIK CCI doesn't use username and channel at all, so it's unclear what exactly does it fix in the context of CCI.

IMO it needs more explanation of the use-case to make it crystal clear (what's the problem are we trying to solve, what's the command line, and some example how to use `host_version:track ref` in the recipe).

e.g. from PR diff it looks like tests are using syntax `self.tool_requires("protobuf/<host_version>@tool_requires_tag")`, while PR title states it should be `self.tool_requires("protobuf/<host_version>:tool_requires_tag")`, which is confusing right now.

to add more to confusion, PR title uses term `ref` (I assume short for `reference`), while code uses term `tag` for some reason, while in reality it's just `username/channel` in tests (could it be a full package or recipe reference here? unclear).

PR #15296, opened 2023-12-18
Make user confs have at leats one `:` separator

AbrilRBS:
Changelog: Bugfix: Ensure `user` confs have at least 1 `:` separator
Changelog: Fix: Fix `user` confs package scoping when no separator was given
Docs: Omit

Fix comes from trying to fix error messages when they disdnt have a `:` but were scoped


conans/model/conf.py
```diff
@@ -493,13 +490,28 @@ def set_relative_base_folder(self, folder):
 
     @staticmethod
     def _check_conf_name(conf):
-        if USER_CONF_PATTERN.match(conf) is None and conf not in BUILT_IN_CONFS:
-            raise ConanException(f"[conf] '{conf}' does not exist in configuration list. "
-                                 f" Run 'conan config list' to see all the available confs.")
+        if (CORE_CONF_PATTERN.match(conf) is not None or TOOLS_CONF_PATTERN.match(conf) is not None):
```
AbrilRBS: It's a bit longer, but I think it's a lot clearer how it's checking it. Maybe duplication can be reduced if someone feels like it's needed, happy to do that

conans/test/integration/package_id/compatible_test.py
```diff
@@ -279,7 +279,7 @@ def package_info(self):
 
         client.save({"conanfile.py": GenConanfile().with_require("pkg/0.1")})
         client.run("lock create . -s os=Windows --lockfile-out=deps.lock")
-        client.run("install . --lockfile=deps.lock")
+        client.run("install . -s os=Windows --lockfile=deps.lock")
```
AbrilRBS: Unrelated to the PR, but this was failing - Change needed to pass on my local Macos machine

conans/test/unittests/model/test_conf.py
```diff
@@ -277,3 +277,21 @@ def test_conf_choices_default():
...
     assert c.get("user.category:option2", choices=[1, 2], default=7) == 7
+
+
+@pytest.mark.parametrize("conf, assert_message", [
+    ("pkg/1.0:user.foo:bar=1", None),
+    ("pkg/1.0:user.baz=1", "'pkg/1.0:user.baz' must have at least one ':' separator"),
+    ("user.foo:bar=1", None),
+    ("user.baz=1", "'user.baz' must have at least one ':' separator"),
+])
+def test_conf_scope_patterns_battery(conf, assert_message):
```
AbrilRBS: Failing test, fixing it now

conans/test/integration/package_id/compatible_test.py
memsharded: This is slightly concerning why CI is not detecting this on the Mac build, the ``release/2.0`` branch builds runs ``integration`` in Mac too. Might be worth investigating, if too much on your plate let me know and I'll check it.

conans/test/unittests/model/test_conf.py
```diff
@@ -277,3 +277,21 @@ def test_conf_choices_default():
...
+    final_conf = scope + conf
+    c = ConfDefinition()
+    if assert_message is not None:
+        with pytest.raises(ConanException) as exc_info:
+            c.loads(final_conf)
+            c.validate()
+        assert assert_message in str(exc_info.value)
+    else:
+        c.loads(final_conf)
+        c.validate()
```
memsharded: Reads very different, sounds more like 2 separate tests

conans/model/conf.py
```diff
@@ -493,13 +490,24 @@ def set_relative_base_folder(self, folder):
 
     @staticmethod
     def _check_conf_name(conf):
-        if USER_CONF_PATTERN.match(conf) is None and conf not in BUILT_IN_CONFS:
+        if CORE_CONF_PATTERN.match(conf) is not None or TOOLS_CONF_PATTERN.match(conf) is not None:
+            if conf in BUILT_IN_CONFS:
+                return  # It's a built-in conf
+            else:
+                raise ConanException(f"[conf] '{conf}' does not exist in configuration list. "
+                                     f" Run 'conan config list' to see all the available confs.")
```
memsharded: Why this necessary? The built-in confs are already fully validated, so this seems redundant?

memsharded (commented):
Looking good, just a couple of details

conans/model/conf.py
AbrilRBS: At this point the only validation done is the one a in the outter if - unknown confs can get here

memsharded: Ok, maybe just 
```python
if USER_CONF_PATTERN.match(conf):
     ...
elif conf not in BUILTIN_CONFS:
     raise
```
Would be enough?

conans/test/integration/package_id/compatible_test.py
memsharded: There is something off here. In my branch, there is already `` client.run("install . -s os=Windows --lockfile=deps.lock")`` in this line 😕

memsharded: @franramirez688 also confirms this is passing for in in ``release/2.0``, maybe need to rebase.

franramirez688:
@RubenRBS From my point of view, the changes to be made should be something like this:

```diff
diff --git a/conans/model/conf.py b/conans/model/conf.py
index 4a46d91ec..45f1d4000 100644
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -116,9 +116,9 @@ BUILT_IN_CONFS = {
 BUILT_IN_CONFS = {key: value for key, value in sorted(BUILT_IN_CONFS.items())}


-CORE_CONF_PATTERN = re.compile(r"^core[.:]")
-TOOLS_CONF_PATTERN = re.compile(r"^tools[.:]")
-USER_CONF_PATTERN = re.compile(r"^user[.:]")
+CORE_CONF_PATTERN = re.compile(r"^(core\..+|core):.*")
+TOOLS_CONF_PATTERN = re.compile(r"^(tools\..+|tools):.*")
+USER_CONF_PATTERN = re.compile(r"^(user\..+|user):.*")


 def _is_profile_module(module_name):
@@ -496,8 +496,9 @@ class Conf:
     @staticmethod
     def _check_conf_name(conf):
         if USER_CONF_PATTERN.match(conf) is None and conf not in BUILT_IN_CONFS:
-            raise ConanException(f"[conf] '{conf}' does not exist in configuration list. "
-                                 f" Run 'conan config list' to see all the available confs.")
+            raise ConanException(f"[conf] Either '{conf}' does not exist in configuration list or "
+                                 f"the conf format introduced is not valid. Run 'conan config list' "
+                                 f"to see all the available confs.")
```

I think that this should cover everything 😁 

UPDATED: I just updated the pattern 😅

memsharded (approved):

franramirez688 (approved):

PR #15297, opened 2023-12-19
Make `core:warnings_as_errors` accept a list

AbrilRBS:
Changelog: Fix: Make `core:warnings_as_errors` accept a list
Docs: Omit

WIP, uploading to get it out as soon as it's done :)


memsharded (commented):
Looks good, please check:
```python
def error(self, msg, error_type=None):
        if self._conan_output_level <= LEVEL_ERROR:
            if self._warnings_as_errors and error_type != "exception":
                raise ConanException(msg)
```
I wouldn't condition the exception to the log-level LEVEL_ERROR. Even if it is not logged, we want it to raise and report it.

memsharded (approved):

franramirez688 (approved):

PR #15300, opened 2023-12-19
ensure --build=editable and --build=cascade works together

### LINKED ISSUE #15292: [bug] `<host_version>` ignores `@user/channel`
### Environment details

* Operating System+version: MacOS 14.2
* Compiler+version: apple-clang 15.0.0
* Conan version: 2.0.14
* Python version: 3.11.5

### Steps to reproduce

Let's just use the example from the [official documentation](https://docs.conan.io/2/reference/conanfile/methods/build_requirements.html#host-version).

Simply, change the override in the app's conanfile to the internal company version of protobuf, e.g.

```python
from conan import ConanFile

class myappRecipe(ConanFile):
    name = "myapp"
    version = "0.1"
    def requirements(self):
        self.requires("mylib/0.1")
        self.requires("protobuf/3.21.9.1@company/stable", override=True)
```

The override and conan installation will fail because _mylib_ will tool-require `protobuf/3.21.9.1` (which does not exist) instead of `protobuf/3.21.9.1@company/stable`.

I guess the `<host_version>` placeholder should contain both the version and `@user/channel` in that case.

However, I'll be also OK with the solution with multiple placeholders, e.g. `protobuf/<host_version>@<host_user>/<host_channel>`, but I think the former should be simpler to use and implement.

### Logs

_No response_

memsharded:
Changelog: Feature: Ensure ``--build=editable`` and ``--build=cascade`` works together.
Docs: https://github.com/conan-io/docs/pull/3550

Close https://github.com/conan-io/conan/issues/15292

### DIFF
--- a/conans/client/graph/build_mode.py
+++ b/conans/client/graph/build_mode.py
@@ -89,7 +89,3 @@ class BuildMode:
         for pattern in self.build_missing_patterns:
             if ref_matches(conanfile.ref, pattern, is_consumer=False):
                 return True
-
-    def report_matches(self):
-        for pattern in self._unused_patterns:
-            ConanOutput().error(f"No package matching '{pattern}' pattern found.", error_type="context")
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -29,7 +29,8 @@ class GraphBinariesAnalyzer(object):
         with_deps_to_build = False
         # check dependencies, if they are being built, "cascade" will try to build this one too
         if build_mode.cascade:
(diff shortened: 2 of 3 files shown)


conans/client/graph/build_mode.py
```diff
@@ -89,7 +89,3 @@ def should_build_missing(self, conanfile):
         for pattern in self.build_missing_patterns:
             if ref_matches(conanfile.ref, pattern, is_consumer=False):
                 return True
-
-    def report_matches(self):
```
memsharded, line 93: This was dead code

AbrilRBS (approved):

PR #15304, opened 2023-12-19
Feature/get commit repo

### LINKED ISSUE #15259: [question] Inconvenient way to retrieve the scm commit with scm.Git. Why not taking the latest commit of the root folder?
We followed the [Conan migration guide](https://docs.conan.io/1/migrating_to_2.0/recipes.html#the-scm-attribute) to use the new `conan.tools.scm.Git`.

In particular, this line seems to have an unfortunate consequence: `git = Git(self, self.recipe_folder)`

From [the doc of `get_commit()`](https://docs.conan.io/2.0/reference/tools/scm/git.html?highlight=get_commit#conan.tools.scm.git.Git.get_commit), this will lead to the command `git rev-list HEAD -n 1 -- <self.recipe_folder>`. By providing a folder to the command (instead of simply ignoring this last argument), this returns the latest commit where any file in the folder was modified. This means it is almost certainly wrong as soon as the recipe does not live at the root of the repository.

Note that we tried to use `self.folders.root` instead for the `Git(folder)`, since we define it explicitly in the `layout()`, but this value is not available in `export()`.

Why not returning the latest commit of the whole repository by default?
Otherwise, how to get the root folder here without duplication the logic from `layout()`?

memsharded:
Changelog: Feature: Add ``Git.get_commit(..., repository=True)`` to obtain the repository commit, not the folder commit.
Docs: omit

Close https://github.com/conan-io/conan/issues/15259

### DIFF
--- a/conans/test/functional/tools/scm/test_git.py
+++ b/conans/test/functional/tools/scm/test_git.py
@@ -28,9 +28,11 @@ class TestGitBasicCapture:
             def export(self):
                 git = Git(self, self.recipe_folder)
                 commit = git.get_commit()
+                repo_commit = git.get_commit(repo=True)
                 url = git.get_remote_url()
                 self.output.info("URL: {}".format(url))
                 self.output.info("COMMIT: {}".format(commit))
+                self.output.info("REPO_COMMIT: {}".format(repo_commit))
                 in_remote = git.commit_in_remote(commit)
                 self.output.info("COMMIT IN REMOTE: {}".format(in_remote))
                 self.output.info("DIRTY: {}".format(git.is_dirty()))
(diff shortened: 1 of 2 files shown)


conans/test/functional/tools/scm/test_git.py
```diff
@@ -28,9 +28,11 @@ class Pkg(ConanFile):
             def export(self):
                 git = Git(self, self.recipe_folder)
                 commit = git.get_commit()
+                repo_commit = git.get_commit(repo=True)
```
memsharded:
```suggestion
                repo_commit = git.get_commit(repository=True)
```

Issue #15311, opened 2023-12-20
[bug] PkgConfigDeps fails in generate() (Conan 1)


db4:
### Environment details

* Conan version: 1.62.0

### Steps to reproduce

Unpack the attached archive and run `build.bat`. BTW, Conan 2 doesn't suffer from that. 
[test_case.zip](https://github.com/conan-io/conan/files/13727614/test_case.zip)

### Logs

```

U:\Work\conan_test\override>conan create pkg2.py pkg2/1.0@  --build missing 
Conan 1 is on a deprecation path, please consider migrating to Conan 2
Exporting package recipe
pkg2/1.0: The stored package has not changed
pkg2/1.0: Using the exported files summary hash as the recipe revision: e303c7731dd0a637598b9e4f19459095 
pkg2/1.0: Exported revision: e303c7731dd0a637598b9e4f19459095
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=15
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

Installing package: pkg2/1.0
Requirements
    pkg2/1.0 from local cache - Cache
    zlib/1.2.13 from 'conancenter' - Cache
Packages
    pkg2/1.0:606fdb601e335c2001bdf31d478826b644747077 - Cache
    zlib/1.2.13:6cc50b139b9c3d27b3e9042d5f5372d327b3a9f7 - Cache

Installing (downloading, building) binaries...
zlib/1.2.13: Already installed!
pkg2/1.0: Already installed!

U:\Work\conan_test\override>conan create pkg2.py pkg2/1.0+private@ 
Conan 1 is on a deprecation path, please consider migrating to Conan 2
Exporting package recipe
pkg2/1.0+private: The stored package has not changed
pkg2/1.0+private: Using the exported files summary hash as the recipe revision: e303c7731dd0a637598b9e4f19459095 
pkg2/1.0+private: Exported revision: e303c7731dd0a637598b9e4f19459095
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=15
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

pkg2/1.0+private: Forced build from source
Installing package: pkg2/1.0+private
Requirements
    pkg2/1.0+private from local cache - Cache
[...]

PR #15320, opened 2023-12-21
[1.x][bugfix] Skipped binaries do not appear on requires

franramirez688:
Changelog: Bugfix: Non-visible transitive dependencies are not included in the requires (PkgConfigDeps).
Changelog: Bugfix: `from_old_cppinfo` helper checks if the old cpp_info is not `None` before merging.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/15311

### DIFF
--- a/conans/test/integration/toolchains/gnu/test_pkgconfigdeps.py
+++ b/conans/test/integration/toolchains/gnu/test_pkgconfigdeps.py
@@ -795,9 +795,6 @@ def test_pkg_config_deps_and_private_deps():
             self.requires("pkg/0.1")
     """)
     client.save({"conanfile.py": conanfile}, clean_first=True)
-    # Now, it passes and creates the pc files correctly
+    # Now, it passes and creates the pc files correctly (the skipped one is not created)
     client.run("install .")
-    # bug???? private dependency is skipped so PC file is not created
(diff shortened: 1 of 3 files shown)


conans/test/integration/toolchains/gnu/test_pkgconfigdeps.py
```diff
@@ -770,3 +770,34 @@ def build(self):
...
+
+        def requirements(self):
+            self.requires("pkg/0.1")
+    """)
+    client.save({"conanfile.py": conanfile}, clean_first=True)
+    # Now, it passes and creates the pc files correctly
+    client.run("install .")
+    # bug???? private dependency is skipped so PC file is not created
+    # assert client.load("private.pc)
+    # But pkg still depends on it!!
```
franramirez688: @memsharded I found a side effect related to skipped binaries. Let's discuss it.

memsharded (approved):
I am a bit concerned about this fix, but it looks correct, so lets try it.

franramirez688:
> I am a bit concerned about this fix, but it looks correct, so let's try it.

Yes 😞 Let's remember that we don't need any of these changes in the Conan 2.x branches, only the test hopefully.

PR #15336, opened 2023-12-22
fix upload_policy=skip --build=missing issues

### LINKED ISSUE #15330: [question] upload_policy = "skip" breaks skipping binaries?
### What is your question?

Looks like `upload_policy = "skip"` attribute in a non-visible dependency doesn't work as one would expect: Conan builds a binary package that should be skipped. The test case is attached. When pkg3 is built pkg1 binary should be skipped, but it's built from sources due to `upload_policy = "skip"` in pkg1.
[test_case.zip](https://github.com/conan-io/conan/files/13752720/test_case.zip)

memsharded:
Changelog: Fix: Fix ``upload_policy=skip`` with ``--build=missing`` issues.
Docs: Omit

Close https://github.com/conan-io/conan/issues/15330

### DIFF
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -206,12 +206,10 @@ class GraphBinariesAnalyzer(object):
                 break
 
         if node.conanfile.upload_policy == "skip":
+            # Download/update shouldn't be checked in the servers if this is "skip-upload"
             if cache_latest_prev:
                 node.binary = BINARY_CACHE
                 node.prev = cache_latest_prev.revision
-            elif build_mode.allowed(node.conanfile):
-                node.should_build = True
-                node.binary = BINARY_BUILD
             else:
                 node.binary = BINARY_MISSING
         elif cache_latest_prev is None:  # This binary does NOT exist in the cache
--- a/conans/client/graph/graph_binaries.py
+++ b/conans/client/graph/graph_binaries.py
@@ -207,6 +207,7 @@ class GraphBinariesAnalyzer(object):
 
         if node.conanfile.upload_policy == "skip":
             # Download/update shouldn't be checked in the servers if this is "skip-upload"
+            # The binary can only be in cache or missing.
             if cache_latest_prev:
                 node.binary = BINARY_CACHE
                 node.prev = cache_latest_prev.revision
--- a/conans/test/integration/command/install/install_test.py
+++ b/conans/test/integration/command/install/install_test.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import textwrap
 from collections import OrderedDict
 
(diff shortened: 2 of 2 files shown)


conans/client/graph/graph_binaries.py
```diff
@@ -206,12 +206,10 @@ def _process_node(self, node, build_mode, remotes, update):
                 break
 
         if node.conanfile.upload_policy == "skip":
+            # Download/update shouldn't be checked in the servers if this is "skip-upload"
```
db4, line 209: I don't understand why **upload**_policy should affect **download**/update in any way. Why `if node.conanfile.upload_policy == "skip":` is needed here and cannot be removed completely?

memsharded: Because if a binary cannot be uploaded, it doesn't make sense that Conan would be looking for binaries in remotes, so Conan should avoid doing that. Some users were requesting that Conan should be able to work offline, without trying to reach the servers at all for upload-policy=skip, and it makes sense. So download of packages must check it.

db4: > Because if a binary cannot be uploaded, it doesn't make sense that Conan would be looking for binaries in remotes, so Conan should avoid doing that.

Just to save some time? With `upload_policy=skip`, the binaries simply can't end up on the server. I don't think this kludge is worth it, but of course, I may not understand something.

> Some users were requesting that Conan should be able to work offline, without trying to reach the servers at all for upload-policy=skip, and it makes sense. So download of packages must check it.

The recipe itself still has to be downloadable/updatable,  `upload_policy` just affects binaries. So it seems to help the offline work not much, Conan servers still need to be accessible. And if `upload_policy=skip` were triggering the fully offline mode, one would need **all** his/her recipes to have this attribute? I don't think it's a realistic scenario.

memsharded: > Just to save some time? With upload_policy=skip, the binaries simply can't end up on the server. I don't think this kludge is worth it, but of course, I may not understand something.

Not for performance, but for "offline" mode. Conan should be able to work offline if it has everything it needs in the cache, and this is the case if ``upload_policy=skip``.

> The recipe itself still has to be downloadable/updatable, upload_policy just affects binaries. So it seems to help the offline work not much, Conan servers still need to be accessible. And if upload_policy=skip were triggering the fully offline mode, one would need all his/her recipes to have this attribute? I don't think it's a realistic scenario.

But the recipe can already be in the cache without the binary. This can happen in several scenarios, for example with a ``conan export``. The problem is that if every ``conan install`` doesn't find binaries in the cache it will go and check in the server, and then fail if there is no network or something like that. The current implementation makes this case work smoothly as users expect, not checking in the servers if not necessary.

db4: > The current implementation makes this case work smoothly as users expect, not checking in the servers if not necessary.

This use of `upload_policy=skip` is not documented in any way. Should it? I wouldn't say this behavior is quite obvious.  And you write `# Download/update shouldn't be checked in the servers if this is "skip-upload"`. Why "update"? It only applies to recipes, not binaries, and is not prevented by `upload_policy=skip`.

memsharded: It is assumed to be the default and expected behavior, not hitting the servers when not necessary, this is why it is not documented that explicitly. But we can definitely improve the docs in to clarify this https://docs.conan.io/2/reference/conanfile/attributes.html#upload-policy

AbrilRBS: Tracking in https://github.com/conan-io/docs/issues/3507

AbrilRBS (approved):

PR #15337, opened 2023-12-22
Disable creating editables without name and version

AbrilRBS:
Changelog: Bugfix: Disable creating editables without name and version.
Docs: Omit

After looking into it with the team, we have decided that creating editables without specifying a name and version is an anti-pattern

### DIFF
--- a/conans/test/integration/editable/test_editables_layout.py
+++ b/conans/test/integration/editable/test_editables_layout.py
@@ -317,3 +317,14 @@ def test_editable_components_absolute_paths():
     c.run("create .")
     # This used to crash due to "include" is not absolute path, now it works
     assert "Testing the package" in c.out
+
+
+def test_editable_no_version_test_package():
+    tc = TestClient()
+    tc.save({"conanfile.py": GenConanfile("app"),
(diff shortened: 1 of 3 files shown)


conans/test/integration/editable/test_editables_layout.py
```diff
@@ -317,3 +317,14 @@ def test(self):
...
+
+
+def test_editable_no_version_test_package():
+    tc = TestClient()
+    tc.save({"conanfile.py": GenConanfile("app"),
+             "test_package/conanfile.py": GenConanfile("test_package")
+             .with_class_attribute("test_type = 'explicit'")
+            .with_test("self.output.info('Testing the package')")})
+    tc.run("editable add .")
+    tc.run("test test_package app/None")
```
memsharded: This explicit ``None`` reads a bit weird, are we support we want to support this as-is, or is it possible to improve this?

AbrilRBS: Good catch, I also think we could have better syntaxt here, but not too sure on the how - let's discuss this next thursday!

memsharded (approved):

PR #15348, opened 2023-12-26
test-poc for cppstd compatibility exact

### LINKED ISSUE #15347: [feature] Provide a strict compatibility checker for cppstd
### What is your suggestion?

Related to https://github.com/conan-io/conan-center-index/issues/19008

The Catch2 project supports both C++14 and C++20, and uses separated features in case building with C++20, making it incompatible to link to projects that were built with C++14.

So far, we don't have a helper to specify it directly. For now, we can use [compatibility.py](https://docs.conan.io/2/reference/extensions/binary_compatibility.html), or some workaround in `validate()`.

memsharded:
Changelog: Feature: Add `extension_properties` attribute to pass information to extensions from recipes.
Changelog: Feature: Implement `compatibility_cppstd` in `extension_properties` for the ``compatibility.py`` plugin to disable fallback to other cppstd for the recipe.
Docs: https://github.com/conan-io/docs/pull/3549

Close https://github.com/conan-io/conan/issues/15347


conans/client/graph/compatibility.py
```diff
@@ -29,7 +29,8 @@ def compatibility(conanfile):
 
 def cppstd_compat(conanfile):
     # It will try to find packages with all the cppstd versions
-
+    if not getattr(conanfile, "compatibility_cppstd", True):
```
memsharded: Lets explore the ``conf`` mechanism instead of attributes:
- With intermediate priorities, soft and hard overwritten by user profiles

conans/client/graph/compatibility.py
```diff
@@ -29,7 +29,9 @@ def compatibility(conanfile):
 
 def cppstd_compat(conanfile):
     # It will try to find packages with all the cppstd versions
-
+    if not conanfile.conf_info.get("user.conan.plugins:compatibility_cppstd", check_type=bool,
```
memsharded: Note it reads directly ``conf_info``, not ``conf``

The specifics of ``user.conan.plugins:compatibility_cppstd`` are to be discussed:
- Do we want a builtin ``tools.conan ....`` (tools are more for recipe tools doesn't seem the best match)
- Do we want a ``core.plugins:compatibility_cppstd``? Maybe sounds better

The ``conf_info`` is only propagated to consumers ``conf`` in the case of direct ``tool_requires``:
- Is it an issue that consumers of this recipe gets the ``compatibility_cppstd`` information from the dependency? Doesn't seem harmful, consumers will mostly ignore it, not an issue if you don't read it

conans/test/integration/package_id/test_validate.py
```diff
@@ -661,6 +660,43 @@ def package_id(self):
...
+            from conan import ConanFile
+            from conan.errors import ConanInvalidConfiguration
+            class Pkg(ConanFile):
+                name = "pkg"
+                version = "0.1"
+                settings = "compiler"
+
+                def compatibility(self):
+                    # It also works as class attribute, but more explicit here
+                    self.conf_info.define("user.conan.plugins:compatibility_cppstd", False)
```
memsharded:
```
class Pkg(Conanfile):
     extensions_properties = {"compatibility_cppstd": False}
```

conans/client/graph/compatibility.py
memsharded: Changed to ``extension_properties`` instead of conf

AbrilRBS (approved):

czoido (approved):

jcar87 (approved):

PR #15349, opened 2023-12-26
Don't take `.` folder into consideration for `tools.files.copy()` `excludes` patterns

### LINKED ISSUE #9208: [question] Strange behavior of `self.copy` excludes patterns
Hi, 
I have a few questions/concerns about how the `copy` method excludes files.

1. If I want to exclude all hidden files (i.e., using the `.*` pattern) no files are packaged at all (i.e., all files are excluded). Is this a bug or by design?
2. If I want to exclude a particular file (e.g., `test.txt`) from the `src` copy directory and also from all subdirectories of `src`, it is required to set 2 patterns for each file: `['test.txt', '*/'test.txt']`. Is this how it should work? Shouldn't the pattern not depend on the file path?

Thanks

AbrilRBS:
Changelog: Bugfix: Don't take `.` folder into consideration for `tools.files.copy()` `excludes` patterns.
Docs: Omit

Closes https://github.com/conan-io/conan/issues/9208

As per https://docs.python.org/3/library/os.path.html#os.path.relpath and my own testing, `relpath` folders do not start with a `./` from the source

### DIFF
--- a/conans/test/unittests/tools/files/test_tool_copy.py
+++ b/conans/test/unittests/tools/files/test_tool_copy.py
@@ -161,6 +161,7 @@ class ToolCopyTest(unittest.TestCase):
         save(os.path.join(folder1, "MyLib.txt"), "")
         save(os.path.join(folder1, "MyLibImpl.txt"), "")
         save(os.path.join(folder1, "MyLibTests.txt"), "")
+
         folder2 = temp_folder()
         copy(None, "*.txt", folder1, folder2, excludes="*Test*.txt")
         self.assertEqual({'MyLib.txt', 'MyLibImpl.txt'}, set(os.listdir(folder2)))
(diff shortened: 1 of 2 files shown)


conans/test/unittests/tools/files/test_tool_copy.py
```diff
@@ -169,6 +170,29 @@ def test_excludes(self):
...
+        mkdir(folder_foo)
+        mkdir(folder_foo_bar)
+        mkdir(folder_hidden)
+
+        save(os.path.join(folder1, "file1.txt"), "")
+        save(os.path.join(folder1, ".hiddenfile"), "")
+        save(os.path.join(folder_foo, "file2.txt"), "")
+        save(os.path.join(folder_foo, ".hiddenfile2"), "")
+        save(os.path.join(folder_hidden, "file3.txt"), "")
+        save(os.path.join(folder_foo_bar, "file4.txt"), "")
```
memsharded: You might do:
```python
save_files(folder1, {"foo/bar/.hiddenfolder/filer.txt": "",
                      "foo/bar/...
```
That might make a bit more evident to the test reader all the files and folders involved.

memsharded (approved):
Looks good, have a look at the possible test improvement (you can discard it too, no prob), and you can merge it.

conans/test/unittests/tools/files/test_tool_copy.py
AbrilRBS: Ohh, didn't know that!

memsharded (approved):

PR #15353, opened 2023-12-27
Fix `tools.info.package_id:confs` when pattern did not match any defined conf

AbrilRBS:
Changelog: Fix: Fix `tools.info.package_id:confs` when pattern did not match any defined conf.
Docs: Omit

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -481,7 +481,7 @@ class Conf:
         # Reading the list of all the configurations selected by the user to use for the package_id
         package_id_confs = self.get("tools.info.package_id:confs", default=[], check_type=list)
         for conf_name in package_id_confs:
-            matching_confs = [c for c in self._values if re.match(conf_name, c)] or [conf_name]
+            matching_confs = [c for c in self._values if re.match(conf_name, c)]
             for name in matching_confs:
                 value = self.get(name)
                 # Pruning any empty values, those should not affect package ID
--- a/conans/test/integration/package_id/package_id_and_confs_test.py
+++ b/conans/test/integration/package_id/package_id_and_confs_test.py
@@ -111,5 +111,6 @@ def test_conf_pkg_id_user_pattern_not_defined():
     })
     tc.save_home({"global.conf": "tools.info.package_id:confs=['user.*']"})
 
+    # This used to break the build because `user.*` conf was not valid
     tc.run("create lib")
     assert "lib/1.0: Package 'da39a3ee5e6b4b0d3255bfef95601890afd80709' created" in tc.out
--- a/conans/test/integration/package_id/package_id_and_confs_test.py
+++ b/conans/test/integration/package_id/package_id_and_confs_test.py
@@ -2,8 +2,7 @@ import textwrap
 
 import pytest
 
-from assets.genconanfile import GenConanfile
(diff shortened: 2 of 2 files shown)


conans/model/conf.py
```diff
@@ -481,7 +481,7 @@ def copy_conaninfo_conf(self):
         # Reading the list of all the configurations selected by the user to use for the package_id
         package_id_confs = self.get("tools.info.package_id:confs", default=[], check_type=list)
         for conf_name in package_id_confs:
-            matching_confs = [c for c in self._values if re.match(conf_name, c)] or [conf_name]
+            matching_confs = [c for c in self._values if re.match(conf_name, c)]
```
AbrilRBS, line 484: The original idea from @thorsten-klein seems to have been that if no match was found, then the passed conf would be a specific conf id - This is not the case for patterns that just don't match anything, either because they are malformed if they are meant to add built in confs, or because they target user. confs that are not currently defined.

I don't see a need to keep the or branch as proper confs will only match themselves, so removing that bit should not change behaviour in any case, so this not breaking, only exclusively a bugfix

This bit comes from the original pr at https://github.com/conan-io/conan/pull/14621

conans/test/integration/package_id/package_id_and_confs_test.py
```diff
@@ -16,14 +20,12 @@
     ('["tools.build:cxxflags", "tools.build:cflags"]', PKG_ID_1),
     ('["tools.build:defines"]', PKG_ID_2),
     ('["tools.build:cxxflags", "tools.build:sharedlinkflags"]', PKG_ID_3),
+    ('["user.foo:value"]', PKG_ID_USER_1),
+    ('["user.foo:value", "user.bar:value"]', PKG_ID_USER_2),
+    ('["user.*"]', PKG_ID_USER_3),
```
AbrilRBS, line 24: @memsharded this is unrelated to the PR - I initially added this to check that user confs were being handled properly. But note how the generated pkgids are different in these 2 cases, we need to look into that too

conans/test/integration/package_id/package_id_and_confs_test.py
```diff
@@ -98,3 +104,13 @@ class Pkg(ConanFile):
...
+def test_conf_pkg_id_user_pattern_not_defined():
+    tc = TestClient(light=True)
+    tc.save({
+        "lib/conanfile.py": GenConanfile("lib", "1.0"),
+    })
+    tc.save_home({"global.conf": "tools.info.package_id:confs=['user.*']"})
+
+    # This used to break the build because `user.*` conf was not valid
+    tc.run("create lib")
+    assert "lib/1.0: Package 'da39a3ee5e6b4b0d3255bfef95601890afd80709' created" in tc.out
```
AbrilRBS, line 115: This is the proper test that was failing before

AbrilRBS (commented):
Some explanatory comments

conans/test/integration/package_id/package_id_and_confs_test.py
AbrilRBS: This is solved by https://github.com/conan-io/conan/pull/15356

memsharded (approved):
I am also fine if this is released in 2.0.17 as a fix too (it can be in 2.1 too)

AbrilRBS:
Noted, will cherry-pick the squashed commit once this lands in `develop2`

PR #15356, opened 2023-12-27
Order `conf` items when dumping them

AbrilRBS:
Changelog: Bugfix: Fix `package_id` calculation when including conf values thru `tools.info.package_id:confs`.
Changelog: Bugfix: Order `conf` items when dumping them to allow reproducible `package_id` independent of the order the confs were declared.
Docs: Omit

This bugfix will generate new package ids for the corner cases where two different orders of the same set of confs were used in `tools.info.package_id:confs`, making the lex sort order only possible one

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -347,7 +347,7 @@ class Conf:
         """
         Returns a string with the format ``name=conf-value``
         """
-        return "\n".join([v.dumps() for v in reversed(self._values.values())])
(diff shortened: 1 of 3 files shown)


memsharded (approved):

memsharded:
> Changelog: Bugfix: Fix package_id calculation when including conf values thru tools.info.package_id:confs
Changelog: Bugfix: Order conf items when dumping them to allow reproducible package_id independent of the order the confs were declared

It looks like the same changelog?

Issue #15346, opened 2023-12-26
[question] visible=False doesn't work for build requirements?


db4:
### What is your question?

Consider the following test recipes:
pkg1/1.0
```python
from conan import ConanFile


class Pkg1Conan(ConanFile):
    settings = "os", "compiler", "arch", "build_type"
```
pkg2/1.0
```python
from conan import ConanFile


class Pkg2Conan(ConanFile):
    settings = "os", "arch", "build_type", "compiler"

    def requirements(self):
        self.requires("pkg1/1.0", build=True)
```
pkg3/1.0
```python
from conan import ConanFile


class Pkg3Conan(ConanFile):
    settings = "os", "arch", "build_type", "compiler"

    def requirements(self):
        self.requires("pkg2/1.0", visible=False)
```
pkg4/1.0
```python
from conan import ConanFile


class Pkg4Conan(ConanFile):
    settings = "os", "arch", "build_type", "compiler"

    def requirements(self):
        self.requires("pkg3/1.0")
```
If I create them in sequence, pkg4 requires pkg1 as a build requirement:
```
pkg4/1.0: Forced build from source
Requirements
    pkg3/1.0#7603eec4a39e35fa51c302be5dc39edf:4f95ed7b4a57821394071c702e54ae2871aa8e91#197e6f60d1dc2e4bc34e15cfeb21e6d4 - Cache
    pkg4/1.0#6b1269936c6af238dac6f28435d83888:c4237a8a579c71a763944d01b215ad8a801871f0 - Build
Build requirements
    pkg1/1.0#9bfaf61a619093fa366d262be46a6b77:2068baaab67a8d1ae925bbe12c37df5b885a6d56#bd38b1b7983e749d27f789e51c8ec4b6 - Cache
Skipped binaries
    pkg2/1.0
```
Why? Shouldn't `visible=True` in `pkg3/1.0` prevent that?

### Have you read the CONTRIBUTING guide?

- [ ] I've read the CONTRIBUTING guide

memsharded:
Hi @db4 

Thanks for the report.
It seems it could be a bug in the propagation of traits down the graph, I am having a look.

db4:
@memsharded Great, thank you! But if you decide to release 2.0.17 with that fix, maybe you also include https://github.com/conan-io/conan/pull/15336 there?

memsharded:
I am submitting https://github.com/conan-io/conan/pull/15357 to fix it.

[...]

PR #15357, opened 2023-12-27
fix trait visible for build requires

memsharded:
Changelog: Bugfix: Fix visible trait propagation with ``build=True`` trait.
Docs: Omit

Close https://github.com/conan-io/conan/issues/15346


AbrilRBS:
Quick question: When did this break? Was it introduced in 2.0.15?

AbrilRBS (approved):

memsharded:
> Quick question: When did this break? Was it introduced in 2.0.15?

It never worked. The use case is subtle:
- Using a ``build=True`` trait, but with a normal require, making it ``visible=True``, which is a very corner case, should probably be avoided for most cases
- Mixing ``visible=False`` with the above ``build=True`` + ``visible=True``

So this case was not covered by tests.

PR #15358, opened 2023-12-28
Intel oneapi detection

### LINKED ISSUE #15344: [feature] Improve detect API to support Intel oneAPI compiler
### What is your suggestion?

Hi there,
Is there anything that we can do in order to better support the Intel oneAPI compiler in the [detect API](https://github.com/conan-io/conan/blob/37573eeb176b9d382d999d5b3c45d220727f8702/conan/internal/api/detect_api.py#L205)?
Any suggestion is welcome, I will be happy to contribute.
Thanks.

jakeheke75:
Changelog: Feature: Intel oneAPI compiler detection improvement.
Docs: Omit
Fixes #15344 

- [x] I've opened another PR in the Conan docs repo to the ``develop2`` branch, documenting this one.

### DIFF
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -447,4 +447,6 @@ def default_compiler_version(compiler, version):
         return major
     elif compiler == "msvc":
         return major
+    elif compiler == "intel-cc":
+        return major
     return version
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -185,6 +185,8 @@ def detect_libcxx(compiler, version):
         return "libCstd"
     elif compiler == "mcst-lcc":
         return "libstdc++"
+    elif compiler == "intel-cc":
+        return "libstdc++"    
 
 
 def default_msvc_runtime(compiler):
@@ -228,6 +230,7 @@ def default_cppstd(compiler, compiler_version):
     default = {"gcc": _gcc_cppstd_default(compiler_version),
                "clang": _clang_cppstd_default(compiler_version),
                "apple-clang": "gnu98",
+               "intel-cc": "17",
                "msvc": _visual_cppstd_default(compiler_version),
                "mcst-lcc": _mcst_lcc_cppstd_default(compiler_version)}.get(str(compiler), None)
     return default
@@ -268,6 +271,10 @@ def detect_compiler():
                 output.error("%s detected as a frontend using apple-clang. "
                              "Compiler not supported" % command)
             return gcc, gcc_version
+        # FIXME add the Windows identifiers
+        if "icpx" in command or "icx" in command:
+            intel, intel_version = _intel_compiler(command)
+            return intel, intel_version        
         if platform.system() == "SunOS" and command.lower() == "cc":
             return _sun_cc_compiler(command)
         if (platform.system() == "Windows" and command.rstrip('"').endswith(("cl", "cl.exe"))
@@ -342,6 +349,18 @@ def _gcc_compiler(compiler_exe="gcc"):
     except (Exception,):  # to disable broad-except
         return None, None
 
+def _intel_compiler(compiler_exe="icx"):
+    try:
+        ret, out = detect_runner("%s --version" % compiler_exe)
+        if ret != 0:
+            return None, None
+        compiler = "intel-cc"
+        installed_version = re.search(r"([0-9]+(\.[0-9])?)", out).group()
+        if installed_version:
+            ConanOutput(scope="detect_api").info("Found %s %s" % (compiler, installed_version))
+            return compiler, Version(installed_version)
+    except (Exception,):  # to disable broad-except
+        return None, None
 
 def _sun_cc_compiler(compiler_exe="cc"):
     try:
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -186,7 +186,7 @@ def detect_libcxx(compiler, version):
     elif compiler == "mcst-lcc":
         return "libstdc++"
     elif compiler == "intel-cc":
-        return "libstdc++"    
+        return "libstdc++11"    
 
 
 def default_msvc_runtime(compiler):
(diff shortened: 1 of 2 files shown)


conan/internal/api/detect_api.py
```diff
@@ -228,6 +230,7 @@ def _mcst_lcc_cppstd_default(version):
     default = {"gcc": _gcc_cppstd_default(compiler_version),
                "clang": _clang_cppstd_default(compiler_version),
                "apple-clang": "gnu98",
+               "intel-cc": "17",
```
memsharded: Any version of intel OneAPI defaults to C++17?
If there is some public official link that defines that, feel free to paste it in code as a comment, that helps for maintenance when new oneapi versions are released.

conans/client/conf/__init__.py
```diff
@@ -121,7 +121,8 @@
         libcxx: [libstdc++, libc++]
         cppstd: [null, 98, gnu98, 11, gnu11, 14, gnu14, 17, gnu17, 20, gnu20, 23, gnu23]
     intel-cc:
-        version: ["2021.1", "2021.2", "2021.3"]
+        version: ["2021.1", "2021.2", "2021.3", "2021.4", "2022.1", "2022.2",
```
memsharded, line 124: Do you know what is the default binary compatibility of compiler versions? 
In ``gcc`` we define both ``6`` (the detected default) and then ``6.1``, ``6.2``, etc, including the minor, for users that want to opt-in to use the minor too, but in general the major only is good enough for general binary compatibility.

memsharded (approved):
This is looking good, just a couple of questions, but I think it can otherwise be merged for next 2.1

conan/internal/api/detect_api.py
jakeheke75: You are perfectly right. Intel oneAPI is switching from C++14 to C++17 [starting from release 2023.0](https://www.intel.com/content/www/us/en/developer/articles/troubleshooting/icx-changes-default-cpp-std-to-cpp17-with-2023.html). Thank you for pointing this out, I will add code to manage this.

conan/internal/api/detect_api.py
```diff
@@ -225,9 +227,15 @@ def _visual_cppstd_default(version):
     def _mcst_lcc_cppstd_default(version):
         return "gnu14" if version >= "1.24" else "gnu98"
 
+    def _intel_cppstd_default(version):
+        tokens = version.main
+        major = tokens[0]
+        return "17" if major >= "2023" else "14"
```
memsharded, line 234:
```suggestion
        # https://www.intel.com/content/www/us/en/developer/articles/troubleshooting/icx-changes-default-cpp-std-to-cpp17-with-2023.html
        return "17" if major >= "2023" else "14"
```

conan/internal/api/detect_api.py
```diff
@@ -342,6 +353,18 @@ def _gcc_compiler(compiler_exe="gcc"):
     except (Exception,):  # to disable broad-except
         return None, None
 
+def _intel_compiler(compiler_exe="icx"):
+    try:
+        ret, out = detect_runner("%s --version" % compiler_exe)
+        if ret != 0:
+            return None, None
+        compiler = "intel-cc"
+        installed_version = re.search(r"([0-9][0-9][0-9][0-9]+(\.[0-9])?)", out).group()
```
memsharded:
```suggestion
        installed_version = re.search(r"(202[0-9]+(\.[0-9])?)", out).group()
```
I am happy simplifying the regex and hardcode the first digits ,like ``(202[0-9]+(\.[0-9])?)``. It might be a bit simpler and more robust, detecting only 202X years versions.

memsharded (approved):

memsharded:
Looks good, I think you can take it out of draft :)

conans/client/conf/__init__.py
jakeheke75: I am newbie and therefore probably missing this point. Do you say that probably in the file it is enough to define the major version of the compiler? Meanwhile I am trying to sort out the matter of binary compatibility.

memsharded: The way the default settings.yml + auto-detection model binary compatibility for ``gcc`` is:

- The profile detection will return ``compiler=gcc`` and ``compiler.version=6`` (Just the major, excluding the minor), for all minor versions of gcc, 6.0, 6.1, 6.2, etc
- That means that the same binary compiled with any of the 6.X minors will be considered compatible for the other versions. A binary built with 6.1, will still have ``compiler.version=6`` in the profile, and result in an identical ``package_id`` than if building with 6.2, because that will still be modeled by the same ``compiler.version=6``. That is the minor is dropped from the binary model
- The minor versions in the ``settings.yml`` after the ``6``, like ``6, 6.0, 6.1, 6.2...`` are for users that want to explicitly model binary incompatibility among different gcc6 minors. They just need to provide the version explicitly in their profiles with ``compiler.version=6.2`` and that will guarantee a different binary than if writing ``compiler.version=6.1``. The tradeoff is more strict binary compatibility guarantees, at the expense of having to build from source more often, for every minor release of the compiler.

The Conan proposed "default" binary compatibility is just a balanced one, achieving binary compatibility/incompatibility in the vast majority of cases, without needed to build too much from source. 

Please let me know if this clarifies a bit.

jakeheke75: It's clear, thank you for taking the time to explain everything in detail. I totally agree with the "default" binary compatibility proposed in Conan. I found in the Intel website the logic behind the [versioning schema](https://www.intel.com/content/www/us/en/developer/articles/tool/oneapi-toolkit-and-component-versioning-schema.html). Intel is following the Semantic Versioning (SemVer).
A SemVer number looks like this: MAJOR.MINOR.PATCH. 

- The MAJOR release has many new features and changes when ABI/API compatibility is broken.
- The MINOR release is an update to the MAJOR release. Features can include an expanded API or new features that are backwards-compatible. 
- Intel oneAPI components may issue a critical PATCH for specific bug and security fixes.

So we could say that also in this case it makes sense to return the major release. I will update the code of method `default_compiler_version` accordingly. Thanks.

memsharded (approved):

memsharded:
Merged, good job, thanks very much!
This will be in next Conan 2.1

jakeheke75:
> Merged, good job, thanks very much! This will be in next Conan 2.1

It would not have been possible without your help. Thank you for all the support.

PR #15360, opened 2023-12-29
Feature: Upload parallel

### LINKED ISSUE #13224: [feature] conan 2.x parallel upload
### What is your suggestion?

It seems that `parallel` upload is missing in Conan 2.x. This feature was present in Conan 1.x and valuable for optimizing uploads.

memsharded:
Changelog: Feature: New ``conan upload`` ``core.upload:parallel`` for faster parallel uploads.
Docs: https://github.com/conan-io/docs/pull/3540

Close https://github.com/conan-io/conan/issues/13224


conan/api/subapi/upload.py
```diff
@@ -57,6 +59,40 @@ def upload(self, package_list, remote):
         executor = UploadExecutor(app)
         executor.upload(package_list, remote)
 
+    def full_upload(self, package_list, remote, enabled_remotes, check_integrity=False, force=False,
```
czoido: It would be nice to add a docstring here, because this is used for the docs to add documentation for the public API methods.

czoido (approved):

conan/api/subapi/upload.py
memsharded: Suggestions for a better name? I didn't love it, but also not better ideas. I thought repurposing the old ``upload()`` to mean this full one, and rename the old one to ``execute_upload`` or ``upload_transfer`` or something like that, but I am afraid that this could break users even if this is not public/stable

czoido: I could not think of a better name, but I think I would use `upload_full` instead of `full_upload` so it's easier to find.

memsharded: Changed to ``upload_full``

czoido (approved):

PR #15366, opened 2023-12-31
Improve the error message when trying to use a `build-scripts` package as a regular requirement

valgur:
It currently fails on an assert with no further context provided in the stack trace. There might be a better place to do this check, though. 

Also, might want to add a check for `PackageType.PYTHON` as well.

Changelog: Fix: Improve the error message when trying to use a `build-scripts` package as a regular requirement
Docs: Omit

### DIFF
--- a/conans/model/requires.py
+++ b/conans/model/requires.py
@@ -289,6 +289,8 @@ class Requirement:
                 downstream_require = Requirement(require.ref, headers=False, libs=False, run=require.run)
             elif pkg_type is PackageType.HEADER:
                 downstream_require = Requirement(require.ref, headers=require.headers, libs=require.libs, run=require.run)
+            elif pkg_type is PackageType.BUILD_SCRIPTS:
+                raise ConanError("build-scripts package must be used as a tool_requires(), not a regular requirement")
             else:
                 assert pkg_type == PackageType.UNKNOWN
                 # TODO: This is undertested, changing it did not break tests


memsharded (commented):
Looks good, just a couple of quick things:
- Is there any scenario that this could break to users already leveraging this, even if it could be considered a bug? (I'd also like to check this with the team, labeling as look-into to discuss, it seems it makes sense, but just in case there could be some angle that I'd be missing)
- It would be good to cover this with a test, if you need help or guidance, please ask, we can even contribute the test ourselves

conans/model/requires.py
```diff
@@ -289,6 +289,8 @@ def transform_downstream(self, pkg_type, require, dep_pkg_type):
                 downstream_require = Requirement(require.ref, headers=False, libs=False, run=require.run)
             elif pkg_type is PackageType.HEADER:
                 downstream_require = Requirement(require.ref, headers=require.headers, libs=require.libs, run=require.run)
+            elif pkg_type is PackageType.BUILD_SCRIPTS:
+                raise ConanError("build-scripts package must be used as a tool_requires(), not a regular requirement")
```
memsharded, line 293: This check is about a build-script depending on a library, not really about it being used as regular ``requires``. This might be an extra check to do, but it seems we want to do a general check over build-scripts being used only as ``tool-requires``

tbsuht:
I just want to mention one case where we are currently using a `build-scripts` package via `requires()`.
https://github.com/conan-io/conan/issues/15925

memsharded (changes requested):
Seeing the feedback, this is too risky, I think it needs some changes:
- Make it a warning instead, not an error
- Most likely better suited in ``DepsGraphBuilder._create_new_node``

PR #15382, opened 2024-01-03
Add built in runtime deployer

### LINKED ISSUE #15381: [feature] add a runtime deployer
### What is your suggestion?

The purposes are:
- ease the conan 2 migration by giving an alternative to the conan import command that was removed
- be able to create an installer, by copying all the .dll into a single folder
- be able to launch the executable on windows, without using the conanrun.sh script (so that you can use the Visual Studio IDE with cmake support more easily) if you have a cmake version that do not allow you to use the cmake presets

The original request come from this PR: https://github.com/conan-io/conan-extensions/pull/94, from @memsharded

> I think, and I have been wanting to have a built-in deployer in conan for runtime (dll, exes). So I think it would be good to move this to the conan repo, and try to contribute this as a built-in:
>- Lets call it runtime_deployer
>- It should target all platform, collecting dlls, dylibs, so and executables (basically the bin folder plus some lib artifacts)
>- It should collect them "flat", all in the same folder
>- I think it should also collect licenses, in different folders so they don't overwrite

fdgStilla:
Changelog: Feature: Add built in `runtime_deploy` deployer.
Docs: https://github.com/conan-io/docs/pull/3789

Fix https://github.com/conan-io/conan/issues/15381

This PR is currently a draft, only tested manually on windows with:
```
conan install --requires=zlib/1.3 --requires=cmake/3.28.1 -of C:\Users\florian.degaulejac\tmp\zlib-cmake --deployer runtime_deploy.py -vtrace --deployer-folder C:\Users\florian.degaulejac\tmp\zlib-cmake\deploy -o "zlib/*:shared=True"
```

Remaining tasks that I won't have the time to complete:

- [ ]  write autotest instead of manual test
- [ ] test other env (linux, apple...)
- [ ] symbolic links
- [ ] test with deep hierarchy and check it is flatten

And a weird behavior I observed: dep.cpp_info.bindirs and dep.cpp_info.libdirs are populated with absolute paths instead of relative path like described in the documentation.


conan/internal/deploy.py
```diff
@@ -80,6 +81,72 @@ def full_deploy(graph, output_folder):
...
+                continue
+            file_count = _flatten_directory(dep, conanfile, bindir_path, output_folder)
+            conanfile.output.info(f"Copied {file_count} files from {dep.ref}, directory named {bindir_name}")
+
+        for libdir_name in dep.cpp_info.libdirs:
+            libdir_path = os.path.join(dep.package_folder, libdir_name)
+            if not os.path.isdir(libdir_path):
+                conanfile.output.verbose(f"{libdir_path} does not exist")
+                continue
+            file_count = _flatten_directory(dep, conanfile, libdir_path, output_folder, [".dll", ".dylib",".so"])
```
memsharded: Maybe we just want to do .dylib and .so, but not .dll? because .dll in Windows go to ``bindirs``, not ``libdirs`` (that is the convention)

memsharded (commented):
Thanks for contributing this, I think it is a good start.

Do you think you could add some basic, simple test? Or do you want me to contribute it? No prob, I can help with that.

fdgStilla:
> Do you think you could add some basic, simple test? Or do you want me to contribute it? No prob, I can help with that.

I am currently struggling to migrate all our projects to conan 2 and I won't have the time to do work more to this PR so any contribution is more than welcome!

conan/internal/deploy.py
fdgStilla: yes maybe

memsharded:
Started to add some tests and minor code changes

conan/internal/deploy.py
fdgStilla: I'am working on the onnxruntime recipe and I noticed that they create dll in the lib directory.
From what I understand they create some "modules" that are supposed to loaded at runtime, and they are considered as "LIBRARY" and not "RUNTIME" by cmake during the installation: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html#installing-targets

Note that according to their documentation the main dll shall be next the to the modules: https://onnxruntime.ai/docs/build/eps.html#loading-the-shared-providers so for this specific recipe we shall move the dll in the bin dir.
What is the correct solution here?
- Patch the onnxruntime cmake file so that they are copied to the bin folder?
- In the conanfile:package method, after the cmake install move the dll from the lib to the bin folder?

I am asking in this thread because it shows that a normal cmake install with modules create dll in the lib folder so maybe we have to manage this case in the runtime deployer.

Todiq:
Hello @memsharded,

Do you have time to keep working on this? I think it could come in very handy for many people. Thanks a lot

memsharded:
Sorry for the delay, things have been a bit busy, and this was stuck in the backlog.

I have added some missing things:
- Components support
- Output scope
- Avoid unnecessary copies to speed up repeated deploys

Ready to move this for next 2.5

AbrilRBS (approved):

czoido (approved):

czoido (approved):
I was thinking that maybe as we have now three built-in deployers, it would be nice to list them in the `--help`, something like this?

```
  -d DEPLOYER, --deployer DEPLOYER
                        Deploy using the provided deployer to the output folder.
                        Built-in deployers: 'full_deploy', 'direct_deploy', 'runtime_deploy'

```

memsharded:
Done, added docstrings.

valgur:
I'm kind of surprised this was included in the 2.5.0 release with the two major bugs still present (`.so` files and symlinks are currently broken, #16527).

PR #15396, opened 2024-01-05
Load extensions from different folders using a conf

AbrilRBS:
Just a sketch for now to discuss with the team


conan/internal/conan_app.py
```diff
@@ -45,7 +45,10 @@ def __init__(self, conan_api):
         self.cache = ClientCache(self.cache_folder, global_conf)
 
         home_paths = HomePaths(self.cache_folder)
-        self.hook_manager = HookManager(home_paths.hooks_path)
+        hooks_paths = global_conf.get("core:extensions_paths", default=["CONAN_HOME"], check_type=list)
+        hooks_paths = [HomePaths(self.cache_folder if path == "CONAN_HOME" else path).deployers_path
+                       for path in hooks_paths]
+        self.hook_manager = HookManager(hooks_paths)
```
memsharded, line 51: This ``hooks_paths = .... deployers_path`` seems wrong, or at least very confusing

memsharded (commented):
I'd like to understand better the user stories of having different extensions in different folders, because it will add extra complexity. I can see future issues for example, precedence of origins when there are name hiding/overlap, while the user stories or problems to solve is not that clear.

So far we have one use case which is testing of custom commands, it is way easier to test them directly in the source origin, without requiring a copy to the Conan cache. But this is for testing, not really for production, so maybe this only requires testing access, not general user configuration.

PR #15413, opened 2024-01-08
test to cover test_requires potential loop

### LINKED ISSUE #15412: [bug] test_requires causing cycle/loop in the dependency graph
### Environment details

* Operating System+version: macos 14.2.1
* Compiler+version: apple-clang 15.0.0
* Conan version: 2.0.16
* Python version: 3.11.7

### Steps to reproduce

It appears that `test_requires` dependencies somehow get leaked and thus causing the cycles/loops when they shouldn't be.

Here is my setup:

`abseil/20230802.1.0@company/stable` **test_requires** `gtest/[~1]@company/stable`
`re2/20231101.0.0@company/stable` **requires** `abseil/[^20230802]@company/stable` and **test_requires** `gtest/[~1]@company/stable`

`gtest/1.11.0@company/stable` has no dependencies
`gtest/1.14.0@company/stable` **requires** both `abseil/[^20230802]@company/stable` and `re2/[^20231101]@company/stable`.

The build order is the following:

- first, we built `gtest/1.11.0@company/stable`
- then, we built `abseil/202300802.1.0@company/stable` which used `gtest/1.11.0@company/stable` for its testing purposes
- then, we built `re2/20231101.0.0@company/stable` which used `gtest/1.11.0@company/stable` for its testing purposes and has a real dependency on `abseil/20230802.1.0@company/stable`
- then, we updated `gtest` to v1.14 and added dependencies to `abseil` and `re2`, and built the new `gtest` package.
- then, an attempt to use `gtest/1.14.0@company/stable` as **test_requires** from some other package (e.g. `opencv`), fails with the following error:

```
|| ERROR: There is a cycle/loop in the graph:
||     Initial ancestor: gtest/1.14.0@company/stable
||     Require: gtest/1.14.0@company/stable
||     Dependency: gtest/1.14.0@company/stable
```

I don't see how this would be a cycle because:

- `opencv` test_requries `gtest`, which requires `abseil` and `re2` (which also requires `abseil`)
- the fact that both `abseil` and `re2` **test_require** `gtest` should not be leaked to `opencv`, as it's not relevant (it may appear in the lock file, but should not enter dependency resolution)

Currently, I can still build `gtest/1.14` without `abseil` and `re2` dependencies as they are still optional, and this is my workaround.

However, Google has [announced](https://github.com/google/googletest?tab=readme-ov-file#coming-soon) that soon gtest will depend on abseil (which itself test-depends on gtest), so this issue needs to be resolved.

### Logs

_No response_

memsharded:
Changelog: Omit
Docs: Omit

Close https://github.com/conan-io/conan/issues/15412


conans/test/integration/graph/core/test_build_requires.py
```diff
@@ -405,6 +405,24 @@ def test_trait_aggregated(self):
         _check_transitive(lib, [(gtest, True, True, False, False),
                                 (zlib, True, True, False, False, False)])
 
+    def test_test_require_loop(self):
+        # https://github.com/conan-io/conan/issues/15412
+        self._cache_recipe("gtest/1.11", GenConanfile())
+        self._cache_recipe("abseil/1.0", GenConanfile().with_test_requires("gtest/[>=1 <1.14]"))
+        self._cache_recipe("gtest/1.14", GenConanfile().with_requires("abseil/1.0"))
+        deps_graph = self.build_graph(GenConanfile("opencv", "1.0").with_require("gtest/1.14"))
```
DoDoENT: maybe you wanted `with_test_require("gtest/1.14")` here?

memsharded: Sure, I was testing that the issues and fixes were the same for requires and test_requires and I left this without restoring its original ``test_requires``. Test fixed, thanks for noticing!

czoido (approved):

PR #15418, opened 2024-01-09
Bugfix libcxx detection when using CC/CXX

### LINKED ISSUE #15399: [bug] profile detect does not try to use the compiler provided
### Environment details

* Operating System+version: Ubuntu 22.04.3 LTS
* Compiler+version: Ubuntu clang version 17.0.6
* Conan version: Conan version 2.0.16
* Python version: Python 3.10.12

### Steps to reproduce

`CC=clang-17 CXX=clang++-17 conan profile detect`
I would expect it to use clang++-17 when detecting libcxx

### Logs

```
detect_api: CC and CXX: clang-17, clang++-17 
detect_api: Found clang 17.0
detect_api: clang>=8, using the major as version
detect_api: WARN: compiler.libcxx check error: /bin/sh: 1: clang++: not found
...
```

AbrilRBS:
Changelog: Bugfix: Fix libcxx detection when using `CC/CXX` env vars.
Docs: https://github.com/conan-io/docs/pull/3509

Closes: https://github.com/conan-io/conan/issues/15399

### DIFF
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -263,11 +263,11 @@ def detect_compiler():
         if "clang" in command.lower():
             return _clang_compiler(command)
         if "gcc" in command or "g++" in command or "c++" in command:
-            gcc, gcc_version = _gcc_compiler(command)
+            gcc, gcc_version, compiler_exe = _gcc_compiler(command)
             if platform.system() == "Darwin" and gcc is None:
                 output.error("%s detected as a frontend using apple-clang. "
                              "Compiler not supported" % command)
-            return gcc, gcc_version, command
+            return gcc, gcc_version, compiler_exe
         if platform.system() == "SunOS" and command.lower() == "cc":
             return _sun_cc_compiler(command)
         if (platform.system() == "Windows" and command.rstrip('"').endswith(("cl", "cl.exe"))
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -344,7 +344,7 @@ def _gcc_compiler(compiler_exe="gcc"):
 
 
 def detect_compiler():
-    ConanOutput(scope="detect_api").warning("detect_compiler() is deprecated, use detect_default_compiler()")
+    ConanOutput(scope="detect_api").warning("detect_compiler() is deprecated, use detect_default_compiler()", warn_tag="deprecated")
     compiler, version, _ = detect_default_compiler()
     return compiler, version
 
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -122,7 +122,7 @@ def _get_e2k_architecture():
     }.get(platform.processor())
 
 
-def detect_libcxx(compiler, version, compiler_exe):
+def detect_libcxx(compiler, version, compiler_exe=None):
     assert isinstance(version, Version)
 
     def _detect_gcc_libcxx(version_, executable):
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -243,17 +243,17 @@ def detect_cppstd(compiler, compiler_version):
     return cppstd
 
 
-def detect_compiler():
+def detect_default_compiler():
     """
-    find the default compiler on the build machine
-    search order and priority:
-    1. CC and CXX environment variables are always top priority
-    2. Visual Studio detection (Windows only) via vswhere or registry or environment variables
-    3. Apple Clang (Mac only)
-    4. cc executable
-    5. gcc executable
-    6. clang executable
-    """
+        find the default compiler on the build machine
+        search order and priority:
+        1. CC and CXX environment variables are always top priority
+        2. Visual Studio detection (Windows only) via vswhere or registry or environment variables
+        3. Apple Clang (Mac only)
+        4. cc executable
+        5. gcc executable
+        6. clang executable
+        """
     output = ConanOutput(scope="detect_api")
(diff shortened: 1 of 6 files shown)


conan/internal/api/detect_api.py
```diff
@@ -122,7 +122,7 @@ def _get_e2k_architecture():
     }.get(platform.processor())
 
 
-def detect_libcxx(compiler, version):
+def detect_libcxx(compiler, version, compiler_exe):
     assert isinstance(version, Version)
 
     def _detect_gcc_libcxx(version_, executable):
```
AbrilRBS, line 128: This function checks for `if executable == "g++": ` to do some special handling, but now the executable can be whatever. Should we keep this around?

conan/internal/api/detect_api.py
```diff
@@ -122,7 +122,7 @@ def _get_e2k_architecture():
     }.get(platform.processor())
 
 
-def detect_libcxx(compiler, version):
+def detect_libcxx(compiler, version, compiler_exe):
```
memsharded: It can be optional defaulted to None, and use ``compiler`` if not provided

conan/internal/api/detect_api.py
```diff
@@ -326,21 +326,21 @@ def _gcc_compiler(compiler_exe="gcc"):
             _, out = detect_runner("%s --version" % compiler_exe)
             out = out.lower()
             if "clang" in out:
-                return None, None
```
memsharded, line 329: Maybe we can simplify the private ``_xxx_compiler`` and they can still return 2 things, it will be just the root ``detect_compiler`` that will handle it.

memsharded (commented):
Please also check broken tests

conans/client/conf/detect.py
```diff
@@ -14,7 +14,7 @@ def detect_defaults_settings():
     arch = detect_arch()
     if arch:
         result.append(("arch", arch))
-    compiler, version = detect_compiler()
+    compiler, version, compiler_exe = detect_compiler()
```
memsharded: @czoido note this is a breaking change for those that started to use the ``detect_api`` feature, but we thought that being it declared as unstable and recently released, we are on time to change it. 

The other alteranative was to opt-in via ``detect_compiler(return_compiler_exe=True)`` or the like

memsharded (approved):

conans/client/conf/detect.py
czoido: I also think that the chances to break are not so high, but my concern is that if anyone is using this, the error is going to be difficult to track, because it will look like this:

```
File "/Users/test/conan/conan/cli/cli.py", line 270, in main
    cli.run(args)
  File "/Users/test/conan/conan/cli/cli.py", line 180, in run
    command.run(self._conan_api, args[0][1:])
  File "/Users/test/conan/conan/cli/command.py", line 126, in run
  File "/Users/test/conan/conan/cli/commands/install.py", line 55, in install
    profile_host, profile_build = conan_api.profiles.get_profiles_from_args(args)
  File "/Users/test/conan/conan/api/subapi/profiles.py", line 66, in get_profiles_from_args
    profile_host = self._get_profile(host_profiles, args.settings_host, args.options_host, args.conf_host,
  File "/Users/test/conan/conan/api/subapi/profiles.py", line 88, in _get_profile
    profile = loader.from_cli_args(profiles, settings, options, conf, cwd)
  File "/Users/test/conan/conans/client/profile_loader.py", line 95, in from_cli_args
    tmp = self.load_profile(p, cwd)
  File "/Users/test/conan/conans/client/profile_loader.py", line 105, in load_profile
    profile = self._load_profile(profile_name, cwd)
  File "/Users/test/conan/conans/client/profile_loader.py", line 130, in _load_profile
    text = rtemplate.render(context)
  File "/Users/test/conan-virtual-env/lib/python3.9/site-packages/jinja2/environment.py", line 1301, in render
    self.environment.handle_exception()
  File "/Users/test/conan-virtual-env/lib/python3.9/site-packages/jinja2/environment.py", line 936, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 2, in top-level template code
ValueError: too many values to unpack (expected 2)

ERROR: too many values to unpack (expected 2)
```

If we don't have a way to raise an error that guides users on how to fix the problem, I personally would prefer a non-breaking change, like adding a new function and deprecate the current one adding a deprecation warning.

memsharded: I think I agree, we can do it non-breaking and deprecating old behavior.

memsharded (approved):

conan/internal/api/detect_api.py
```diff
@@ -326,21 +326,27 @@ def _gcc_compiler(compiler_exe="gcc"):
...
             ConanOutput(scope="detect_api").info("Found %s %s" % (compiler, installed_version))
-            return compiler, Version(installed_version)
+            return compiler, Version(installed_version), compiler_exe
     except (Exception,):  # to disable broad-except
-        return None, None
+        return None, None, None
+
+
+def detect_compiler():
+    ConanOutput(scope="detect_api").warning("detect_compiler() is deprecated, use detect_default_compiler()")
```
czoido: I think the `warn_tag` is missing
```suggestion
    ConanOutput(scope="detect_api").warning("detect_compiler() is deprecated, use detect_default_compiler()", warn_tag="deprecated")
```

AbrilRBS: Damn, really nice catch

memsharded (approved):

franramirez688 (approved):

PR #15436, opened 2024-01-10
[dev] Added env variable to append other custom command location

franramirez688:
Changelog: omit
Docs: omit

### DIFF
--- a/conan/cli/cli.py
+++ b/conan/cli/cli.py
@@ -50,6 +50,8 @@ class Cli:
         conan_custom_commands_path = HomePaths(self._conan_api.cache_folder).custom_commands_path
         # Important! This variable should be only used for testing/debugging purpose
         developer_custom_commands_path = os.getenv("_CONAN_INTERNAL_CUSTOM_COMMANDS_PATH")
+        # Notice that in case of having same custom commands file names, the developer one has
+        # preference over the Conan default location because of the sys.path.append(xxxx)
         custom_commands_folders = [developer_custom_commands_path, conan_custom_commands_path] \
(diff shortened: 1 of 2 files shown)


conan/cli/cli.py
```diff
@@ -47,35 +47,41 @@ def _add_commands(self):
             for k, v in self._commands.items():  # Fill groups data too
                 self._groups[v.group].append(k)
 
-        custom_commands_path = HomePaths(self._conan_api.cache_folder).custom_commands_path
-        if not os.path.isdir(custom_commands_path):
-            return
+        conan_custom_commands_path = HomePaths(self._conan_api.cache_folder).custom_commands_path
+        # Important! This variable should be only used for testing/debugging purpose
+        developer_custom_commands_path = os.getenv("_CONAN_INTERNAL_CUSTOM_COMMANDS_PATH")
+        custom_commands_folders = [developer_custom_commands_path, conan_custom_commands_path] \
```
AbrilRBS, line 55: This and the test below confuse me a bit, let's check

AbrilRBS (commented):
🤔

conan/cli/cli.py
memsharded, line 55: I think the order should be reversed, to give priority to the developer ones

franramirez688 (commented):
@memsharded @RubenRBS @czoido 

The order is confusing but it's fine as it is, and the test is already checking that. I just added a comment to explain better what's happening under the hood. Basically, we're using `sys.path.append`, so the developer ones will have preference over any other custom commands defined.

memsharded (approved):

czoido (approved):

PR #15441, opened 2024-01-10
Add new ``--core-conf`` command line argument to allow passing `core.` confs via CLI

### LINKED ISSUE #13410: Add a way to define global confs from CLI

memsharded:
Changelog: Feature: Add new ``--core-conf`` command line argument to allow passing `core.` confs via CLI.
Docs: https://github.com/conan-io/docs/pull/3515

Close https://github.com/conan-io/conan/issues/13410


conan/cli/command.py
```diff
@@ -100,9 +102,14 @@ class ConanArgumentParser(argparse.ArgumentParser):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
```
AbrilRBS, line 105: What if we add the -gc arg in this file so that every command gets it by default? Just like for format

AbrilRBS (commented):
Looking good! A suggestion I have, but yes let's talk about it!

conan/cli/command.py
memsharded: It is being added to all commands already, isn't it?

memsharded: If you mean the parsing and handling, we need the ``conan_api``, which doesn't seem easily available here.

AbrilRBS: Oh missread which file was being modified, we actually had the same idea then, 👍

AbrilRBS:
> I still have some concerns, most of the confs are really not for command line, but some it makes sense.

Yeah, I think we should warn for things like `core.cache:storage_path` which are not set early enough, right?

memsharded:
> Yeah, I think we should warn for things like core.cache:storage_path which are not set early enough, right?

Actually, it is possible, I have extended the test to prove it

AbrilRBS:
Ohhhhh, it actually is! This is great!

memsharded:
I have changed this PR, no need to modify every ``parse_args`` call to inject the api, with this change, every single Conan command gets a ``-gc/--global-conf`` argument that injects all the conf the very first thing in the ``conan_api.config.global_conf`` that everybody else reads

conan/cli/command.py
```diff
@@ -97,12 +99,18 @@ def _format(self, parser, info, *args):
...
 
     def parse_args(self, args=None, namespace=None):
         args = super().parse_args(args)
         ConanOutput.define_log_level(os.getenv("CONAN_LOG_LEVEL", args.v))
+        if args.global_conf:
+            from conans.model.conf import ConfDefinition
+            confs = ConfDefinition()
+            confs.loads("\n".join(args.global_conf))
+            self._conan_api.config.global_conf.update_conf_definition(confs)
```
AbrilRBS, line 118: This is great ⭐

AbrilRBS (approved):
😍

franramirez688 (approved):
This is great 👏

czoido (approved):

franramirez688 (approved):

czoido (approved):

AbrilRBS (approved):

PR #15447, opened 2024-01-12
Consider adding cmakeExecutable to presets

### LINKED ISSUE #15427: [feature] configurePresets.cmakeExecutable in CMakePresets.json should be filled in
### What is your suggestion?

In `CMakePresets.json`, you can set `configurePresets.cmakeExecutable` to set a specific CMake binary. 
While this doesn't work at alll with Visual Studio (stupid IDE insists on using built-in, outdated CMake version), it does work for IDEs like Visual Studio Code and similar ones.

This avoids a nasty issue when working a "hybrid" model between IDE with CMake integration mixed with conan commands in the same workspace.

Why you would want to do that?... E.g. because the *initial* configuration of the CMake cache is something which still behaves somewhat different between doing it within the IDE vs doing it in Conan. Stuff like environment variables required for compiler detection not being correctly set in the `CMakePresets.json`, and loads of other similar (environment related) issues.

When you have such a workflow, and you are working in a setup where CMake is distributed as a Conan dependency (and not installed system wide like in Conan Center), then you end up with Conan and the IDE trying to use different CMake versions on the same cache with guaranteed conflicts.

Specifying the CMake version `conan` would use in the presets ensures that configuration of the workspace with Conan, but subsequent work *without* Conan is possible in the first place.

While at it, also have a look at `configurePreset.environment` please, that's yet another thing Conan isn't filling in, leaving you in an odd situation where IDEs invoking CMake will do something ***entirely*** different from what  Conan had prepared as an environment.

Worst case they pick up an entirely different compiler toolchain. E.g. trying to use Ninja together with MSVC and VSCode is something which should work out of the box, but absolutely doesn't when Conan is in the mix so you have to start re-wire commands just to make sure at least VCVars is eventually called before CMake.

czoido:
Changelog: Feature: Add cmakeExecutable to configure preset.
Docs: https://github.com/conan-io/docs/pull/3548

Closes: https://github.com/conan-io/conan/issues/15427

### DIFF
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -184,6 +184,18 @@ class CMakeToolchain(object):
         content = relativize_generated_file(content, self._conanfile, "${CMAKE_CURRENT_LIST_DIR}")
         return content
 
+    def _find_cmake_exe(self):
+        for req in self._conanfile.dependencies.build.values():
+            if req.ref.name == "cmake":
+                for bindir in req.cpp_info.bindirs:
+                    cmake_path = os.path.join(bindir, "cmake")
+                    cmake_exe_path = os.path.join(bindir, "cmake.exe")
+
+                    if os.path.exists(cmake_path):
+                        return cmake_path
+                    elif os.path.exists(cmake_exe_path):
+                        return cmake_exe_path
+
     def generate(self):
         """
           This method will save the generated files to the conanfile.generators_folder
@@ -215,7 +227,7 @@ class CMakeToolchain(object):
             else:
                 cache_variables[name] = value
 
-        buildenv, runenv = None, None
+        buildenv, runenv, cmake_executable = None, None, None
 
         if self._conanfile.conf.get("tools.cmake.cmaketoolchain:presets_environment", default="",
                                     check_type=str, choices=("disabled", "")) != "disabled":
@@ -228,8 +240,11 @@ class CMakeToolchain(object):
             runenv = {name: value for name, value in
                       run_env.items(variable_reference="$penv{{{name}}}")}
 
+            cmake_executable = self._find_cmake_exe()
+
         write_cmake_presets(self._conanfile, toolchain, self.generator, cache_variables,
-                            self.user_presets_path, self.presets_prefix, buildenv, runenv)
+                            self.user_presets_path, self.presets_prefix, buildenv, runenv,
+                            cmake_executable)
 
     def _get_generator(self, recipe_generator):
         # Returns the name of the generator to be used by CMake
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -185,7 +185,7 @@ class CMakeToolchain(object):
         return content
 
     def _find_cmake_exe(self):
-        for req in self._conanfile.dependencies.build.values():
+        for req in self._conanfile.dependencies.direct_build.values():
             if req.ref.name == "cmake":
                 for bindir in req.cpp_info.bindirs:
                     cmake_path = os.path.join(bindir, "cmake")
--- a/conan/tools/cmake/toolchain/toolchain.py
+++ b/conan/tools/cmake/toolchain/toolchain.py
@@ -240,7 +240,7 @@ class CMakeToolchain(object):
             runenv = {name: value for name, value in
                       run_env.items(variable_reference="$penv{{{name}}}")}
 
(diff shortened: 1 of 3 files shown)


conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -184,6 +184,18 @@ def content(self):
         content = relativize_generated_file(content, self._conanfile, "${CMAKE_CURRENT_LIST_DIR}")
         return content
 
+    def _find_cmake_exe(self):
+        for req in self._conanfile.dependencies.build.values():
```
memsharded: Better ``direct_build`` only?
```suggestion
        for req in self._conanfile.dependencies.direct_build.values():
```

memsharded (commented):
looks good

conan/tools/cmake/toolchain/toolchain.py
```diff
@@ -184,6 +184,18 @@ def content(self):
         content = relativize_generated_file(content, self._conanfile, "${CMAKE_CURRENT_LIST_DIR}")
         return content
 
+    def _find_cmake_exe(self):
+        for req in self._conanfile.dependencies.direct_build.values():
+            if req.ref.name == "cmake":
```
jcar87, line 189: This assumes a package called `cmake` - which indeed is likely what it's going to be called, but I'm aways wary of blurring the separation between the client and the packages (which may or may not exist). However this is likely to be the package name and it is harmless otherwise :D

Question: what happens when the profile has a `platform_tool_requires` on cmake?

jcar87 (approved):
Very good! just one comment, but it shouldn't be a problem.

I wonder if it would be useful to have a an attribute in CMakeToolchain that the users could set in the `generate()` method, if they so wish? , in case they're in a universe where cmake is not in the system path, and also not a tool_requirement (think a CI machine that has multiple CMake versions installed in different places)

conan/tools/cmake/toolchain/toolchain.py
czoido: > This assumes a package called `cmake` - which indeed is likely what it's going to be called, but I'm aways wary of blurring the separation between the client and the packages (which may or may not exist). However this is likely to be the package name and it is harmless otherwise :D

Yes, I agree, that was the same thing I was not 100% convinced about this.

> 
> Question: what happens when the profile has a `platform_tool_requires` on cmake?

Good point, I'll check if something weird happens in that case.

memsharded: The ``platform_tool_requires`` in theory shouldn't get to this point in the CMakeToolchain code, it is filtered before, so it will not be in ``direct_build`` dependencies?

memsharded:
> I wonder if it would be useful to have a an attribute in CMakeToolchain that the users could set in the generate() method, if they so wish? , in case they're in a universe where cmake is not in the system path, and also not a tool_requirement (think a CI machine that has multiple CMake versions installed in different places)

We already have
```python
self._cmake_program = conanfile.conf.get("tools.cmake:cmake_program", default="cmake")
```

Inside the ``CMake`` helper, to override which CMake is being used. Maybe that is what could go directly to the presets as well?

conan/tools/cmake/toolchain/toolchain.py
czoido: Yes, that's exactly what happens, I expanded the test with that to verify anyway

czoido:
> > I wonder if it would be useful to have a an attribute in CMakeToolchain that the users could set in the generate() method, if they so wish? , in case they're in a universe where cmake is not in the system path, and also not a tool_requirement (think a CI machine that has multiple CMake versions installed in different places)
> 
> We already have
> 
> ```python
> self._cmake_program = conanfile.conf.get("tools.cmake:cmake_program", default="cmake")
> ```
> 
> Inside the `CMake` helper, to override which CMake is being used. Maybe that is what could go directly to the presets as well?

I made changes to read that conf with preference

czoido:
> > I wonder if it would be useful to have a an attribute in CMakeToolchain that the users could set in the generate() method, if they so wish? , in case they're in a universe where cmake is not in the system path, and also not a tool_requirement (think a CI machine that has multiple CMake versions installed in different places)
> 
> We already have
> 
> ```python
> self._cmake_program = conanfile.conf.get("tools.cmake:cmake_program", default="cmake")
> ```
> 
> Inside the `CMake` helper, to override which CMake is being used. Maybe that is what could go directly to the presets as well?

I made changes to read that conf with preference

memsharded (approved):

PR #15461, opened 2024-01-16
Add support for vcvars when using Powershell

### LINKED ISSUE #15267: [bug] Build env scripts generated with virtualenv:powershell are incomplete
### Environment details

* Operating System+version:  Windows 10 22H2
* Compiler+version:  msvc 193
* Conan version: 2.0.14
* Python version: 3.11.6

### Steps to reproduce

1. Create a simple `conanfile.py` with a dependency. For example:
```python
from conan import ConanFile
from conan.tools.cmake import CMake, cmake_layout, CMakeToolchain, CMakeDeps

class Conan(ConanFile):

   requires = [("gtest/1.12.1")]

   generators = 'CMakeDeps', 'CMakeToolchain'

   def layout(self):
      cmake_layout(self)
```
2. Run `conan install . -c tools.env.virtualenv:powershell=True`

Expected Result:
- The generated `conanbuild.bat` and `conanbuild.ps1` basically does the same.

Result:
- `conanbuild.bat` activate the Visual Studio 17 environment while   `conanbuild.ps1` does not.

-----------------------------------------------------
My understanding of the `conanbuild` scripts was that the visual studio environment would be activated when running them. So I was expecting the `conanbuild.ps1` script to do so.

Maybe my understanding is incorrect but I feel like both version should do the same.

### Logs

```powershell
PS C:\Local> conan profile show
Host profile:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja

Build profile:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja

PS C:\Local> conan --version
Conan version 2.0.14
PS C:\Local> python --version
Python 3.11.6
PS C:\Local> conan install . -c tools.env.virtualenv:powershell=True

Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja
tools.env.virtualenv:powershell=True

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja


Graph root
    conanfile.py: C:\Local\conanfile.py
Requirements
    gtest/1.12.1#3ed1009f8004c2ac29ab5b1d9fbfbf53 - Cache

Requirements
    gtest/1.12.1#3ed1009f8004c2ac29ab5b1d9fbfbf53:1c9c0f79f33093a353ed3a111698ab362c007c19#96cb726491a605d4ad6dc216969ccb06 - Cache

gtest/1.12.1: Already installed! (1 of 1)
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'cpp_info.names' used in: gtest/1.12.1

======== Finalizing install (deploy, generators) ========
conanfile.py: Writing generators to C:\Local\build\Release\generators
conanfile.py: Generator 'CMakeToolchain' calling 'generate()'
conanfile.py: CMakeToolchain generated: conan_toolchain.cmake
conanfile.py: Preset 'conan-release' added to CMakePresets.json. Invoke it manually using 'cmake --preset conan-release' if using CMake>=3.23
conanfile.py: If your CMake version is not compatible with CMakePresets (<3.23) call cmake like: 'cmake <path> -G Ninja -DCMAKE_TOOLCHAIN_FILE=C:\Local\build\Release\generators\conan_toolchain.cmake -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Release'
conanfile.py: CMakeToolchain generated: CMakePresets.json
conanfile.py: Generator 'CMakeDeps' calling 'generate()'
conanfile.py: Generating aggregated env files
conanfile.py: Generated aggregated env files: ['conanbuild.bat', 'conanbuild.ps1', 'conanrun.ps1']
Install finished successfully
PS C:\Local> cat .\build\Release\generators\conanbuild.bat
@echo off
call "%~dp0/conanvcvars.bat"
PS C:\Local> cat .\build\Release\generators\conanvcvars.bat
@echo off
set __VSCMD_ARG_NO_LOGO=1
set VSCMD_SKIP_SENDTELEMETRY=1
echo conanvcvars.bat: Activating environment Visual Studio 17 - amd64 - vcvars_ver=14.3
set "VSCMD_START_DIR=%CD%" && call "C:\Program Files\Microsoft Visual Studio\2022\Professional\VC/Auxiliary/Build/vcvarsall.bat"  amd64 -vcvars_ver=14.3
PS C:\Local> cat .\build\Release\generators\conanbuild.ps1
& "$PSScriptRoot/conanbuildenv-release-x86_64.ps1"
PS C:\Local> cat .\build\Release\generators\conanbuildenv-release-x86_64.ps1
Push-Location $PSScriptRoot
"echo `"Restoring environment`"" | Out-File -FilePath "deactivate_conanbuildenv-release-x86_64.ps1"
$vars = (Get-ChildItem env:*).name
$updated_vars = @()

foreach ($var in $updated_vars)
{
    if ($var -in $vars)
    {
        $var_value = (Get-ChildItem env:$var).value
        Add-Content "deactivate_conanbuildenv-release-x86_64.ps1" "`n`$env:$var = `"$var_value`""
    }
    else
    {
        Add-Content "deactivate_conanbuildenv-release-x86_64.ps1" "`nif (Test-Path env:$var) { Remove-Item env:$var }"
    }
}
Pop-Location
```

juansblanco:
Changelog: Feature: Add support for use of vcvars env variables when calling from powershell.
Docs: https://github.com/conan-io/docs/pull/3541

Trying to activate the generated `conanbuild.ps1` file to set the environment from a Powershell console was not working.
To fix this there's now a new `conanvcvars.ps1` file that saves the environment from calling `conanvcvars.bat` to be used in the Powershell.

Fixes: https://github.com/conan-io/conan/issues/15267

### DIFF
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -154,7 +154,7 @@ class VCVars:
         is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
         if is_ps1:
             content_ps1 = textwrap.dedent(f"""\
-            if (-not $env:VSCMD_ARG_HOST_ARCH){{
+            if (-not $env:VSCMD_ARG_VCVARS_VER){{
                 cmd /c "$PSScriptRoot/conanvcvars.bat&set" |
                 foreach {{
                   if ($_ -match "=") {{
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -144,23 +144,20 @@ class VCVars:
                                 winsdk_version=winsdk_version, vcvars_ver=vcvars_ver,
                                 vs_install_path=vs_install_path)
 
-        content = f"""\
+        is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
+        ps_command = 'powershell.exe -noexit -command "%1"' if is_ps1 else ""
+        content = textwrap.dedent(f"""\
             @echo off
             set __VSCMD_ARG_NO_LOGO=1
             set VSCMD_SKIP_SENDTELEMETRY=1
             echo conanvcvars.bat: Activating environment Visual Studio {vs_version} - {vcvarsarch} - winsdk_version={winsdk_version} - vcvars_ver={vcvars_ver}
-            {vcvars}"""
+            {vcvars}
+            echo hola%1
+            {ps_command}
+            """)
 
-        is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
         from conan.tools.env.environment import create_env_script
-        if is_ps1:
-
-            bat_content = textwrap.dedent(content + f'\n echo cmd env \n powershell.exe -noexit -command "exit"')
-            ps_content = textwrap.dedent(content + f'\n echo powershell env \n powershell.exe -noexit -command ""')
-            create_env_script(conanfile, bat_content, CONAN_VCVARS_BAT, scope)
-            create_env_script(conanfile, ps_content, CONAN_VCVARS_PS1, scope)
-        else:
-            create_env_script(conanfile, content, CONAN_VCVARS_BAT, scope)
+        create_env_script(conanfile, content, CONAN_VCVARS_BAT, scope)
 
 
 def vs_ide_version(conanfile):
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -151,6 +151,8 @@ class VCVars:
             {vcvars}
             """)
         from conan.tools.env.environment import create_env_script
+        create_env_script(conanfile, content, CONAN_VCVARS_BAT, scope)
+
         is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
         if is_ps1:
             content_ps1 = textwrap.dedent(f"""\
@@ -167,7 +169,6 @@ class VCVars:
             """)
             create_env_script(conanfile, content_ps1, CONAN_VCVARS_PS1, scope)
 
-        create_env_script(conanfile, content, CONAN_VCVARS_BAT, scope)
 
 
 def vs_ide_version(conanfile):
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -155,12 +155,14 @@ class VCVars:
         if is_ps1:
             content_ps1 = textwrap.dedent(f"""\
             if (-not $env:VSCMD_ARG_VCVARS_VER){{
-                cmd /c "$PSScriptRoot/conanvcvars.bat&set" |
+                Push-Location "$PSScriptRoot"
+                cmd /c "conanvcvars.bat&set" |
                 foreach {{
                   if ($_ -match "=") {{
                     $v = $_.split("=", 2); set-item -force -path "ENV:\$($v[0])"  -value "$($v[1])"
                   }}
                 }}
+                Pop-Location
                 write-host conanvcvars.ps1: Activated environment}}
             """)
             create_env_script(conanfile, content_ps1, CONAN_VCVARS_PS1, scope)
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -138,17 +138,23 @@ class VCVars:
         # C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
         # C:\Program Files (x86)\Microsoft Visual Studio\2017\Community
         # C:\Program Files (x86)\Microsoft Visual Studio 14.0
-        vcvars = vcvars_command(vs_version, architecture=vcvarsarch, platform_type=None,
-                                winsdk_version=winsdk_version, vcvars_ver=vcvars_ver,
-                                vs_install_path=vs_install_path)
-
+        is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool)
+        is_ps1 = False
+        if is_ps1:
+            vcvars = '& "$PSScriptRoot/conanvcvars.bat"'
+        else:
+            vcvars = vcvars_command(vs_version, architecture=vcvarsarch, platform_type=None,
+                                    winsdk_version=winsdk_version, vcvars_ver=vcvars_ver,
+                                    vs_install_path=vs_install_path)
+        new_line = 'powershell.exe -noexit -command "& { %* }"'
         content = textwrap.dedent("""\
             @echo off
             set __VSCMD_ARG_NO_LOGO=1
             set VSCMD_SKIP_SENDTELEMETRY=1
             echo conanvcvars.bat: Activating environment Visual Studio {} - {} - winsdk_version={} - vcvars_ver={}
             {}
-            """.format(vs_version, vcvarsarch, winsdk_version, vcvars_ver, vcvars))
+            {}
+            """.format(vs_version, vcvarsarch, winsdk_version, vcvars_ver, vcvars, new_line))
         from conan.tools.env.environment import create_env_script
         create_env_script(conanfile, content, CONAN_VCVARS_FILE, scope)
 
--- a/conan/tools/microsoft/visual.py
+++ b/conan/tools/microsoft/visual.py
@@ -7,7 +7,8 @@ from conan.errors import ConanException, ConanInvalidConfiguration
 from conan.tools.scm import Version
 from conan.tools.intel.intel_cc import IntelCC
 
-CONAN_VCVARS_FILE = "conanvcvars.bat"
+CONAN_VCVARS_BAT = "conanvcvars.bat"
+CONAN_VCVARS_PS1 = "conanvcvarsps.bat"
 
 
 def check_min_vs(conanfile, version, raise_invalid=True):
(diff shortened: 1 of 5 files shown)


conan/tools/microsoft/visual.py
```diff
@@ -142,15 +143,31 @@ def generate(self, scope="build"):
...
+            {vcvars}
+            """)
         from conan.tools.env.environment import create_env_script
-        create_env_script(conanfile, content, CONAN_VCVARS_FILE, scope)
+        is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
+        if is_ps1:
+            generator_path = conanfile.generators_path
+            content_ps1 = textwrap.dedent(f"""\
+            pushd "{generator_path}"
+            cmd /c "conanvcvars.bat&set" |
```
jcar87: we could use `$PSScriptRoot/conanvcvars.bat` instead of directly by name? this would mean we don't have to do `pushd` and `popd` - and use the same structure as `conanbuild.ps1` already does - unless I'm missing something?

(see [docs](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7.4#psscriptroot))

conans/client/generators/__init__.py
```diff
@@ -186,7 +186,13 @@ def bat_content(files):
                  bat_content(deactivates(bats)))
         if ps1s:
             def ps1_content(files):
-                return "\r\n".join(['& "{}"'.format(b) for b in files])
+                content = "\r\n"
+                for b in files:
+                    if "vcvars" in b:
+                        content += 'if (-not $env:VSCMD_ARG_HOST_ARCH){{ & "{}" }}\n'.format(b)
```
jcar87: I would add this conditional in the contents of `conanvcvars.ps1` and keep the logic here as it already was? it means the file would "always" be called when activating `conanbuild.ps1`, but it would immediately skip the logic if the variable is defined.

As a second question: is this `VSCMD_ARG_HOST_ARCH` guaranteed to exist for all versions of visual studio supported by Conan 2.0?

juansblanco: To the first question: this is something I thought about, I thought it would be better to not even call the other script, but it's true it is a good idea to remove the logic from here and leave it completely at the file generation.

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
```diff
@@ -114,3 +114,53 @@ def test(self):
     assert "MYPS1!!!!" in client.out
     assert "MYVC_CUSTOMVAR1=PATATA1" in client.out
     assert "MYVC_CUSTOMVAR2=PATATA2" in client.out
+
+@pytest.mark.skipif(platform.system() != "Windows", reason="Requires Windows powershell")
+def test_vcvars():
```
memsharded, line 121: This test is NOT running in CI because it is Window only and "integration" tests in Windows do not pass for regular PR, only when merged.

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
```diff
@@ -114,3 +114,53 @@ def test(self):
...
+    client = TestClient()
+    conanfile = textwrap.dedent(r"""
+        from conan import ConanFile
+        from conan.tools.cmake import CMake, cmake_layout, CMakeToolchain, CMakeDeps
+
+        class Conan(ConanFile):
+           settings = "os", "compiler", "build_type", "arch"
+
+           test_requires = [("gtest/1.12.1")]
```
memsharded: You cannot use the real ``gtest`` in tests, you need to create a fake "hello" or "pkg" dependency for testing. 
Most likely the dependency is not critical for the scope of the test, better drop it.

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
```diff
@@ -114,3 +114,53 @@ def test(self):
...
+        find_package(GTest REQUIRED)
+
+        add_executable(hello hello.cpp)
+        target_link_libraries(hello PRIVATE GTest::gtest GTest::gtest_main)
+    """)
+    client.save({"CMakeLists.txt": cmakelists})
+    client.run("install . -c tools.env.virtualenv:powershell=True -c tools.cmake.cmaketoolchain:generator=Ninja -s compiler.version=193")
+    conanbuild = client.load(r".\build\Release\generators\conanbuild.ps1")
+    vcvars_ps1 = client.load(r".\build\Release\generators\conanvcvars.ps1")
+    assert "conanvcvars.ps1" in conanbuild
```
memsharded, line 154: I think a full building example that shows everything fully works would be necessary.

conans/client/generators/__init__.py
memsharded: I agree with @jcar87 better keep this logic as-is, and move it to the other file, calling it is cheap, and it will be cleaner

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
juansblanco: Is this okay? The other tests that are exclusive to Powershell do the same. Do we need additional checks?

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
juansblanco: I added the build command but this only tests something that was already working before. 
The original issue appeared when activating the environment manually, which I'm not sure if it can be done in the test environment.

juansblanco:
The deactivate environments needs fixing, I'm looking into it.

Edit: It was already broken when trying to deactivate conanvcvars envs, it will be fixed in a different PR.

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
memsharded: Have a look to how the powershell is launched, and try to do in the test something like ``powershell ....ps1 ; do-something-to-check``

conans/test/integration/toolchains/env/test_virtualenv_powershell.py
memsharded: Good point, probably those deserve to be moved to "functional", try to do it, hopefully github diff will be good and not show a huge remove/add diff

conan/tools/microsoft/visual.py
```diff
@@ -142,15 +143,31 @@ def generate(self, scope="build"):
...
+            echo conanvcvars.bat: Activating environment Visual Studio {vs_version} - {vcvarsarch} - winsdk_version={winsdk_version} - vcvars_ver={vcvars_ver}
+            {vcvars}
+            """)
         from conan.tools.env.environment import create_env_script
-        create_env_script(conanfile, content, CONAN_VCVARS_FILE, scope)
+        is_ps1 = conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool, default=False)
+        if is_ps1:
+            content_ps1 = textwrap.dedent(f"""\
+            if (-not $env:VSCMD_ARG_VCVARS_VER){{
+                Push-Location "$PSScriptRoot"
```
juansblanco, line 160: I reverted this back to using "push/pop" as calling the path directly in the cmd command was causing trouble when path had spaces.

conans/test/functional/toolchains/env/test_virtualenv_powershell.py
```diff
@@ -114,3 +114,53 @@ def test(self):
...
+    """)
+    client.save({"hello.cpp": hello_cpp})
+    cmakelists = textwrap.dedent(r"""
+        cmake_minimum_required(VERSION 3.15)
+
+        project(hello-world LANGUAGES CXX)
+        message("C++ compiler: ${CMAKE_CXX_COMPILER}")
+
+        add_executable(hello hello.cpp)
+    """)
```
memsharded: These files can most likely be generated in 2 lines with ``gen_function_cpp`` and ``gen_cmakelists`` helpers.

conan/tools/microsoft/visual.py
```diff
@@ -142,15 +143,31 @@ def generate(self, scope="build"):
...
+                cmd /c "conanvcvars.bat&set" |
```
memsharded, line 161: If this is using ``conanvcvars.bat`` I'd probably generate the ``conanvcvars.ps1`` after generating the ``conanvcvars.bat``. A very minor detail, but I was confused at first with the diff because it seemed that the .bat was removed.

memsharded (approved):
Looks good!

memsharded (approved):

jcar87 (approved):

conan/tools/microsoft/visual.py
```diff
@@ -142,15 +143,32 @@ def generate(self, scope="build"):
...
+                foreach {{
+                  if ($_ -match "=") {{
+                    $v = $_.split("=", 2); set-item -force -path "ENV:\$($v[0])"  -value "$($v[1])"
+                  }}
+                }}
+                Pop-Location
+                write-host conanvcvars.ps1: Activated environment}}
+            """)
```
jmarrec, line 169: I know no-one asked for my opinion ( :) ) but this makes me uneasy.

here's how you could do it similarly than the Launch-VsDevShell.ps1 (`$vspath\Common7\Tools\Launch-VsDevShell.ps1`)

```powershell
$vspath=$(vswhere -products '*' -requires Microsoft.Component.MSBuild -property installationPath -latest)
$devShellModule = "$vspath\Common7\Tools\Microsoft.VisualStudio.DevShell.dll"
Import-Module $devShellModule
Enter-VsDevShell -VsInstallPath $vspath -SkipAutomaticLocation -Arch amd64
```

jmarrec:
```powershell
Get-Help Enter-VsDevShell

NAME
    Enter-VsDevShell

SYNTAX
    Enter-VsDevShell -VsInstallPath <string> [-SkipExistingEnvironmentVariables] [-StartInPath <string>] [-Arch {Default | x86 | amd64 | arm | arm64}] [-HostArch {Default | x86 | amd64}]
    [-DevCmdArguments <string>] [-DevCmdDebugLevel {None | Basic | Detailed | Trace}] [-SkipAutomaticLocation] [-SetDefaultWindowTitle] [-ReportNewInstanceType {PowerShell | Cmd |
    LaunchScript}] [<CommonParameters>]

    Enter-VsDevShell [-VsInstanceId] <string> [-SkipExistingEnvironmentVariables] [-StartInPath <string>] [-Arch {Default | x86 | amd64 | arm | arm64}] [-HostArch {Default | x86 |
    amd64}] [-DevCmdArguments <string>] [-DevCmdDebugLevel {None | Basic | Detailed | Trace}] [-SkipAutomaticLocation] [-SetDefaultWindowTitle] [-ReportNewInstanceType {PowerShell | Cmd
    | LaunchScript}] [<CommonParameters>]

    Enter-VsDevShell [-Arch {Default | x86 | amd64 | arm | arm64}] [-HostArch {Default | x86 | amd64}] [-Test] [-DevCmdDebugLevel {None | Basic | Detailed | Trace}] [<CommonParameters>]


ALIASES
    None


REMARKS
    None
```

jcar87: Hi @jmarrec - thanks for your feedback.

This is the approach we considered initially, but discarded it for a couple of reasons:
- `Launch-VsDevShell.ps1` and the underlying implementation do not expose the same interface that vcvars provides, in particular the ability to specify the toolset with the `-vcvars_ver`, or using the Windows Store SDK, or specifying the WIndows SDK version. We do rely on mapping Conan settings/ conf to some of these - and the interface for the powershell ones do not expose these. 
- The documented way would be to invoke `Launch-VsDevShell.ps1` once located - rather than importing `DevShell.dll` - this is evidenced:
  - in the [documentation](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022), where it says:

> Launch-VsDevShell.ps1 is the recommended way to initialize Developer PowerShell interactively or for scripting build automation.
- This response by a Microsoft employee, where it says that the interface to `Enter-VsDevShell` is considered undocumented and can change without notice, and once again reiterates that `Launch-VsDevShell.ps1` is to be used - https://developercommunity.visualstudio.com/t/MicrosoftVisualStudioDevShell-document/1348197

we wouldn't want to use undocumented/subject to breakages features, and the `Launch-VsDevShell.ps1` does not provide feature parity, so we relied on this fallback.

jmarrec: Gotcha, thanks a lot for the detailed explanation!

PR #15462, opened 2024-01-16
Refactor for RemotesAPI

memsharded:
Changelog: Feature: Initial documentation to make ``RemotesAPI`` publicly available (experimental).
Docs: Omit

I plan to docoument and publish in docs the ``RemotesAPI`` as the first "experimental" but released and documented sub API

Should probably be merged after https://github.com/conan-io/conan/pull/15464


conans/test/integration/command/remote_test.py
```diff
@@ -335,6 +335,7 @@ def test_add_wrong_conancenter():
     c.run("remote add whatever https://conan.io/center", assert_error=True)
     assert "Wrong ConanCenter remote URL. You are adding the web https://conan.io/center" in c.out
     assert "the correct remote API is https://center.conan.io" in c.out
+    c.run("remote add conancenter htpps://center.conan.io")
```
AbrilRBS: If the typo is intentional, maybe a comment specifying such?

AbrilRBS (approved):
Looks reaally nice. We still need to discuss how much we want to/if we want to specify types for the methods

conans/test/integration/command/remote_test.py
memsharded: Not intentional, I just needed `conancenter` remote name to exist otherwise the `conan update` fails because not existing

czoido (approved):

PR #15464, opened 2024-01-17
Add `--allowed-packges` to remotes to limit what references a remote can supply

AbrilRBS:
Changelog: Feature: Add `--allowed-packges` to remotes to limit what references a remote can supply.
Docs: https://github.com/conan-io/docs/pull/3534

### DIFF
--- a/conans/client/cache/remote_registry.py
+++ b/conans/client/cache/remote_registry.py
@@ -33,9 +33,9 @@ class _Remotes:
         data = json.loads(text)
         for r in data.get("remotes", []):
             disabled = r.get("disabled", False)
-            filters = r.get("filters", None)
+            allowed_packages = r.get("allowed_packages", None)
             # TODO: Remote.serialize/deserialize
-            remote = Remote(r["name"], r["url"], r["verify_ssl"], disabled, filters)
+            remote = Remote(r["name"], r["url"], r["verify_ssl"], disabled, allowed_packages)
             result._remotes[r["name"]] = remote
         return result
 
@@ -45,8 +45,8 @@ class _Remotes:
             remote = {"name": r.name, "url": r.url, "verify_ssl": r.verify_ssl}
             if r.disabled:
                 remote["disabled"] = True
-            if r.filters:
-                remote["filters"] = r.filters
+            if r.allowed_packages:
+                remote["allowed_packages"] = r.allowed_packages
             remote_list.append(remote)
         ret = {"remotes": remote_list}
         return json.dumps(ret, indent=True)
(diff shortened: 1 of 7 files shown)


conans/client/cache/remote_registry.py
```diff
@@ -110,6 +113,10 @@ def update(self, remote_name, url=None, secure=None, disabled=None, index=None,
             remotes.insert(index, remote)
             self._remotes = {r.name: r for r in remotes}
 
+        if filters is not None:
+            remote.filters = remote.filters or []
+            remote.filters.extend(filters)
```
memsharded: The update shouldn't be an ``extend`` but a replacement. Otherwise it is not possible to remove things

conans/client/graph/proxy.py
```diff
@@ -97,6 +97,9 @@ def _find_newest_recipe_in_remotes(self, reference, remotes, update, check_updat
 
         results = []
         for remote in remotes:
+            if remote.filters and not any(reference.matches(f, is_consumer=False) for f in remote.filters):
+                output.info(f"Excluding remote {remote.name} because recipe is filtered out")
```
memsharded: This will be very verbose, I suggest `debug` level

memsharded (approved):
Looks good!

franramirez688 (commented):
IMO, I think that `--filter` name is a bit confusing. It's being used in `conan remote` command, and it's filtering recipe references. I'd prefer to see another param name.

AbrilRBS:
Let's talk with the team about what the best naming is, yeah :)

czoido:
> Let's talk with the team about what the best naming is, yeah :)

I think the name suggest that the packages added with `--filter` will be filtered out, but those are the ones that you can fetch from the remote, so I thin it may be better to use a name that suggests those packages are the ones allowed in the server. Something like `--allowed`, `--allowed-packages`, `--allowed-patterns` ? or any derivatives

conans/test/integration/command/remote_test.py
```diff
@@ -346,50 +345,56 @@ def test_add_wrong_conancenter():
...
+    tc.run("upload * -r=default -c")
+    tc.run("remove * -c")
+
+    tc.run("install --requires=app/1.0 -r=default")
+    assert "app/1.0: Downloaded recipe revision 04ab3bc4b945a2ee44285962c277906d" in tc.out
+    assert "liba/1.0: Downloaded recipe revision 4d670581ccb765839f2239cc8dff8fbd" in tc.out
+    tc.run("remove * -c")
+
+    # lib/2.* pattern does not match the one being required by app, should not be found
+    tc.run('remote update default --allowed-packages="app/*" --allowed-packages="liba/2.*"')
```
memsharded, line 366: If I have a relatively long list, maybe it makes sense to also have a short `-a=` too?

memsharded (approved):

PR #15465, opened 2024-01-17
[MesonToolchain] Using `get_apple_sdk_fullname` to avoid empty apple minimum version flag

### LINKED ISSUE #15459: [bug] MesonToolchain cannot set `-mmacosx-version-min` flag automatically when build for macos target
### Environment details

* Operating System+version: macOS Sonoma 14.2.1
* Compiler+version: Xcode 15.1
* Conan version: 2.0.13
* Python version: 3.10.5

### Steps to reproduce

1. Build lib dav1d from conan-center-index
2. check conan_meson_cross.ini - No flag found
3. otool -l ${path_to_libdav1d.dylib} - minos is 14.0

### Logs

In file conan/tools/meson/toolchain.py 

line 304:
```
        os_sdk = self._conanfile.settings.get_safe('os.sdk')
        if not os_sdk and self._os != "Macos":
            raise ConanException("Please, specify a suitable value for os.sdk.")
```
When built for macos, there is no os.sdk in settings.yml, so this os_sdk is None

line 314
```
        self.apple_min_version_flag = [apple_min_version_flag(os_version, os_sdk, subsystem)]
```
In function `apple_min_version_flag`
```
def apple_min_version_flag(os_version, os_sdk, subsystem):
    """compiler flag name which controls deployment target"""
    if not os_version or not os_sdk:
        return ''

    # FIXME: This guess seems wrong, nothing has to be guessed, but explicit
    flag = ''
    if 'macosx' in os_sdk:
        flag = '-mmacosx-version-min'
    elif 'iphoneos' in os_sdk:
        flag = '-mios-version-min'
    elif 'iphonesimulator' in os_sdk:
        flag = '-mios-simulator-version-min'
    elif 'watchos' in os_sdk:
        flag = '-mwatchos-version-min'
    elif 'watchsimulator' in os_sdk:
        flag = '-mwatchos-simulator-version-min'
    elif 'appletvos' in os_sdk:
        flag = '-mtvos-version-min'
    elif 'appletvsimulator' in os_sdk:
        flag = '-mtvos-simulator-version-min'
    elif 'xros' in os_sdk:
        # need early return as string did not fit into the "normal" schema
        return f'-target arm64-apple-xros{os_version}'
    elif 'xrsimulator' in os_sdk:
        # need early return as string did not fit into the "normal" schema
        return f'-target arm64-apple-xros{os_version}-simulator'

    if subsystem == 'catalyst':
        # especial case, despite Catalyst is macOS, it requires an iOS version argument
        flag = '-mios-version-min'

    return f"{flag}={os_version}" if flag else ''
```
If os_sdk is none, it returns ''
while expected `-mmacosx-version-min={os_version}`

So, MesonToolchain never write -mmacosx-version-min={os_version}  to the meson_build.ini, which will be compiler default (eg. 14.0)

franramirez688:
Changelog: Bugfix: MesonToolchain calculates a valid `apple_min_version_flag`.
Docs: omit
Closes: https://github.com/conan-io/conan/issues/15459

### DIFF
--- a/conan/tools/meson/toolchain.py
+++ b/conan/tools/meson/toolchain.py
@@ -6,7 +6,7 @@ from jinja2 import Template
 from conan.errors import ConanException
 from conan.internal import check_duplicated_generator
 from conan.tools.apple.apple import to_apple_arch, is_apple_os, apple_min_version_flag, \
-    apple_sdk_path
+    apple_sdk_path, get_apple_sdk_fullname
 from conan.tools.build.cross_building import cross_building
 from conan.tools.build.flags import libcxx_flags
 from conan.tools.env import VirtualBuildEnv
@@ -296,13 +296,8 @@ class MesonToolchain(object):
                 "Apple SDK path not found. For cross-compilation, you must "
                 "provide a valid SDK path in 'tools.apple:sdk_path' config."
             )
-
-        # TODO: Delete this os_sdk check whenever the _guess_apple_sdk_name() function disappears
-        os_sdk = self._conanfile.settings.get_safe('os.sdk')
-        if not os_sdk and self._os != "Macos":
-            raise ConanException("Please, specify a suitable value for os.sdk.")
-
         # Calculating the main Apple flags
+        os_sdk = get_apple_sdk_fullname(self._conanfile)
         arch = to_apple_arch(self._conanfile)
         self.apple_arch_flag = ["-arch", arch] if arch else []
         self.apple_isysroot_flag = ["-isysroot", sdk_path] if sdk_path else []
--- a/conans/test/functional/toolchains/meson/test_cross_compilation.py
+++ b/conans/test/functional/toolchains/meson/test_cross_compilation.py
@@ -57,17 +57,15 @@ option('STRING_DEFINITION', type : 'string', description : 'a string option')
 
 @pytest.mark.tool("meson")
 @pytest.mark.skipif(platform.system() != "Darwin", reason="requires Xcode")
-@pytest.mark.parametrize("arch, os_, os_version, os_sdk, platform_assertion", [
-    # platform_assertion -> https://stackoverflow.com/questions/44188023/how-to-check-a-lib-static-or-dynamic-is-built-for-ios-simulator-or-mac-osx
-    ('armv8', 'watchOS', '10.1', 'watchos', "platform 4"),
-    ('armv8', 'iOS', '17.1', 'iphoneos', "platform 2"),
-    ('armv7', 'iOS', '10.0', 'iphoneos', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
-    ('x86', 'iOS', '10.0', 'iphonesimulator', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
-    ('x86_64', 'iOS', '10.0', 'iphonesimulator', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
-    ('armv8', 'Macos', None, None, "platform 1"),  # MacOS M1
-    ('armv8', 'Macos', '10.11', None, "platform 1"),  # MacOS M1
+@pytest.mark.parametrize("arch, os_, os_version, os_sdk", [
+    ('armv8', 'iOS', '17.1', 'iphoneos'),
+    ('armv7', 'iOS', '10.0', 'iphoneos'),
+    ('x86', 'iOS', '10.0', 'iphonesimulator'),
+    ('x86_64', 'iOS', '10.0', 'iphonesimulator'),
+    ('armv8' if platform.machine() == "x86_64" else "x86_64", 'Macos', None, None),
+    ('armv8' if platform.machine() == "x86_64" else "x86_64", 'Macos', '10.11', None),
 ])
-def test_apple_meson_toolchain_cross_compiling(arch, os_, os_version, os_sdk, platform_assertion):
+def test_apple_meson_toolchain_cross_compiling(arch, os_, os_version, os_sdk):
     profile = textwrap.dedent("""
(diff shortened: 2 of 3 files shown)


conans/test/integration/toolchains/meson/test_mesontoolchain.py
```diff
@@ -281,3 +281,38 @@ def test_check_c_cpp_ld_list_formats():
     assert "c = ['aarch64-poky-linux-gcc', '-mcpu=cortex-a53', '-march=armv8-a+crc+crypto']" in content
     assert "cpp = ['aarch64-poky-linux-g++', '-mcpu=cortex-a53', '-march=armv8-a+crc+crypto']" in content
     assert "ld = ['aarch64-poky-linux-ld', '--sysroot=/opt/sysroots/cortexa53-crypto-poky-linux']" in content
+
+
+@pytest.mark.skipif(platform.system() != "Darwin", reason="Requires OSX")
+def test_apple_meson_add_min_version_flag():
```
memsharded: I think there is a problem, and there are insufficient tests of ``MesonToolchain`` with different ``os.sdk`` and ``os.sdk_version`` values, for cross-building. I cannot find any coverage of these cases in the test suite, and I am concerned, following the code paths, that something might not be ok.

conan/tools/meson/toolchain.py
```diff
@@ -296,13 +296,8 @@ def _resolve_apple_flags_and_variables(self, build_env, compilers_by_conf):
...
                 "provide a valid SDK path in 'tools.apple:sdk_path' config."
             )
-
-        # TODO: Delete this os_sdk check whenever the _guess_apple_sdk_name() function disappears
-        os_sdk = self._conanfile.settings.get_safe('os.sdk')
-        if not os_sdk and self._os != "Macos":
-            raise ConanException("Please, specify a suitable value for os.sdk.")
-
         # Calculating the main Apple flags
+        os_sdk = get_apple_sdk_fullname(self._conanfile)
```
memsharded, line 300: Using this reads weird:
- The ``get_apple_sdk_fullname`` returns a value with the sdk+sdk_version.
- But then, later, this value is passed to ``apple_min_version_flag()`` which requires the sdk_version separately, ignores the value in the ``os_sdk`` and use the ``os.version`` instead.

The whole flow is confusing (it is not because of this PR, it was already confusing)

conans/test/integration/toolchains/meson/test_mesontoolchain.py
franramirez688: I don't think so. You can have a look at the `conans/test/functional/toolchains/meson/test_cross_compilation.py` module.

conan/tools/meson/toolchain.py
franramirez688: Yes, I agree. Do you want me to open an issue to solve this?

franramirez688: I just saw that this flow also appears in the AutotoolsToolchain generator.

memsharded: I see, good point. Still readability and understanding what and why it is doing it is complicated. Lets try to do a later refactor of this then.

conans/test/integration/toolchains/meson/test_mesontoolchain.py
memsharded: The test that is covering it has:
```
t.run("install . --profile:build=default --profile:host=profile_host")
t.run("build .")
```
That is legacy 1.X syntax and behavior, that means that the test is not really cross-compiling, and just using the ``--profile:host=default`` to build the application! And this is the test that covers the cases of ``os_sdk`` and ``os_version``

memsharded: Maybe fixing this in the test make it fail without this PR fix?

franramirez688: You're completely right, but the test was cross-building because of the info gathered by the MesonToolchain` in the `conan_meson_cross.ini`. Despite that, I'm changing this test to verify if it was failing as you said.

conans/test/functional/toolchains/meson/test_cross_compilation.py
```diff
@@ -56,19 +56,18 @@ def build(self):
...
-    ('x86_64', 'iOS', '10.0', 'iphonesimulator'),
-    ('armv8', 'Macos', None, None)  # MacOS M1
+@pytest.mark.parametrize("arch, os_, os_version, os_sdk, platform_assertion", [
+    ('armv8', 'watchOS', '10.1', 'watchos', "platform 4"),
+    ('armv8', 'iOS', '17.1', 'iphoneos', "platform 2"),
+    ('armv7', 'iOS', '10.0', 'iphoneos', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
+    ('x86', 'iOS', '10.0', 'iphonesimulator', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
+    ('x86_64', 'iOS', '10.0', 'iphonesimulator', "cmd LC_VERSION_MIN_IPHONEOS"),  # "platform" is missing!
+    ('armv8', 'Macos', None, None, "platform 1"),  # MacOS M1
+    ('armv8', 'Macos', '10.11', None, "platform 1"),  # MacOS M1
```
franramirez688: Those platform checks: https://stackoverflow.com/questions/44188023/how-to-check-a-lib-static-or-dynamic-is-built-for-ios-simulator-or-mac-osx

memsharded: Don't be afraid to put links like that in the test itself, so it is easier for developers to read it without having to blame + track back to Github PR + find this specific comment in the review that might be hidden if we resolve the conversation.

memsharded (approved):

PR #15467, opened 2024-01-17
[TestClient] Additions

franramirez688:
Changelog: omit
Docs: omit

### DIFF
--- a/conans/test/utils/tools.py
+++ b/conans/test/utils/tools.py
@@ -30,7 +30,7 @@ from conan.internal.cache.home_paths import HomePaths
 from conans import REVISIONS
 from conan.api.conan_api import ConanAPI
 from conan.api.model import Remote
-from conan.cli.cli import Cli
+from conan.cli.cli import Cli, _CONAN_INTERNAL_CUSTOM_COMMANDS_PATH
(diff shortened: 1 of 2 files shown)


conans/test/utils/tools.py
```diff
@@ -429,6 +429,9 @@ def __init__(self, cache_folder=None, current_folder=None, servers=None, inputs=
         else:
             text = default_profiles[platform.system()]
         save(self.cache.default_profile_path, text)
+        # Using internal env variable to add another custom commands folder
+        if custom_commands_folder:
+            os.environ.update({_CONAN_INTERNAL_CUSTOM_COMMANDS_PATH: custom_commands_folder})
```
memsharded: Adding this way to ``os.environ`` for tests is quite problematic, as ``os.environ`` is not necessarily cleaned, so it might remain from test to test, etc. I'd suggest trying to find a way that is temporary, like with a context like ``with environment_update()`` or something like that.

franramirez688: You're right. Better do this temporarily instead of the whole test instance.

memsharded (approved):

PR #15491, opened 2024-01-22
[DRAFT] Add environment to presets inheriting from other presets

czoido:
Changelog: Feature: Add environment to presets inheriting from other presets.
Changelog: Feature: Ability to use modified environment in generate() method in CMakePresets.

It would also allow to do the same proposed here: https://github.com/conan-io/conan/pull/15470 
This would be an alternative to the other PR.


czoido:
Closed in favor of https://github.com/conan-io/conan/pull/15470
Inheriting environment through presets feels a bit forced and prone to generate problems with presets in the future.

PR #15501, opened 2024-01-23
Avoid missing file stacktrace when no metadata exists for a source backup

AbrilRBS:
Changelog: Fix: Avoid missing file stacktrace when no metadata exists for a source backup.
Docs: Omit

### DIFF
--- a/conans/client/downloaders/download_cache.py
+++ b/conans/client/downloaders/download_cache.py
@@ -3,6 +3,7 @@ import os
 from contextlib import contextmanager
 from threading import Lock
 
+from conan.api.output import ConanOutput
 from conans.util.dates import timestamp_now
 from conans.util.files import load, save
 from conans.util.locks import SimpleLock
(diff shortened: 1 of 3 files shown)


franramirez688 (approved):

conans/test/integration/cache/backup_sources_test.py
```diff
@@ -154,6 +156,7 @@ def source(self):
             path=self.client.cache.cache_folder)
         self.client.run("create .")
         self.client.run("upload * -c -r=default")
+        assert "FileNotFoundError: [Errno 2] No such file or directory" not in self.client.out
```
memsharded: This might be tricky. It is possible that with other Pythons, other OSs, the error message is not the same. I'd try to ``assert "error" not in str(self.client.out).lower()`` or something like that.

conans/client/downloaders/download_cache.py
```diff
@@ -78,6 +79,9 @@ def should_upload_sources(package):
             if not path.endswith(".json"):
                 blob_path = os.path.join(path_backups, path)
                 metadata_path = os.path.join(blob_path + ".json")
+                if not os.path.exists(metadata_path):
+                    ConanOutput().warning(f"Missing metadata file for backup source {blob_path}")
```
memsharded: I am not sure about just a warning. This seems like a violation of the consistency / broken storage of backup sources, shouldn't this be a hard error?

AbrilRBS: Makes sense, changed to raise a hard error

franramirez688 (approved):

memsharded (approved):

PR #15503, opened 2024-01-23
Add Fish support for environments

AbrilRBS:
Changelog: Feature: Add Fish shell command support for environments
Docs: https://github.com/conan-io/docs/pull/3690

Still need to check if the `-g` scope for the variable erasure is ok, and how activating both `:powershell` and `:fish` confs work


memsharded (commented):
Looks good, just I agree it would need a bit more testing of functionality and possible interactions with other shells.

uilianries:
Well, I'm officially converted to the cult of Fish now, so I'm interested as well.

I'll take a look with Ruben to push this PR again.

conan/tools/env/environment.py
```diff
@@ -520,12 +520,52 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+              {deactivate}
+              """).format(deactivate=deactivate if generate_deactivate else "")
+        result = [capture]
+        for varname, varvalues in self._values.items():
+            value = varvalues.get_str("${name}", self._subsystem, pathsep=self._pathsep)
+            value = value.replace('"', '\\"')
+            if value:
+                result.append('set -gx {} "{}"'.format(varname, value))
+            else:
+                result.append('set -ge {}'.format(varname))
```
uilianries: I guess it could be simplified by using `set -p` (prepend), instead of expanding all values.

https://fishshell.com/docs/current/cmds/set.html

conans/test/functional/toolchains/env/test_virtualenv_fish.py
```diff
@@ -0,0 +1,43 @@
...
+
+    with open(os.path.join(client.current_folder, "conanbuildenv.fish"), "r") as f:
+        buildenv = f.read()
+        assert 'set -gx MYPATH1 "/path/to/ar"' in buildenv
+
+    client.run_command("fish -c 'source conanbuildenv.fish && set'")
+    assert 'MYPATH1 /path/to/ar' in client.out
+
+    client.run_command("fish -c 'source conanbuildenv.fish && set && source deactivate_conanbuildenv.fish && set'")
+    assert str(client.out).count('MYPATH1 /path/to/ar') == 1
```
uilianries: Need to add a test appending variables too.

conan/tools/env/environment.py
```diff
@@ -520,12 +520,52 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+
+        content = "\n".join(result)
+        content = relativize_generated_file(content, self._conanfile, "$script_folder")
```
memsharded: Isn't it possible to not use this and relativize paths as the other environment do?

conan/tools/env/environment.py
```diff
@@ -534,18 +574,27 @@ def save_script(self, filename):
         if ext:
             is_bat = ext == ".bat"
             is_ps1 = ext == ".ps1"
+            is_fish = ext == ".fish"
         else:  # Need to deduce it automatically
             is_bat = self._subsystem == WINDOWS
             is_ps1 = self._conanfile.conf.get("tools.env.virtualenv:powershell", check_type=bool)
-            if is_ps1:
+            is_fish = self._conanfile.conf.get("tools.env.virtualenv:fish", check_type=bool)
```
memsharded, line 619: There are no wrappers for ``.fish`` files, are they? If there are no wrappers, this will break when set and used to build any dependency. 
This should most likely apply only to end-consumer conanfile, not cache dependencies, and only if not used for ``conan build`` or anything like that.

conan/tools/env/environment.py
```diff
@@ -520,12 +520,52 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+    def save_fish(self, file_location, generate_deactivate=True):
+        filepath, filename = os.path.split(file_location)
+        deactivate_file = os.path.join(filepath, "deactivate_{}".format(filename))
+        values = self._values.keys()
+        if len(values) == 0:
+            # Empty environment, nothing to restore (Easier to handle in Fish)
+            deactivate = ("""echo "echo Nothing to restore" > {deactivate_file}"""
+                          .format(deactivate_file=deactivate_file))
+        else:
+            deactivate = textwrap.dedent("""\
```
memsharded: For new one, I'd try to do directly an in-memory or in-tmp-folder deactivate script rather than a local generated one.

memsharded (commented):
This is a bit risky in its current form, because it breaks when applied to dependencies or ``conan build``

uilianries:
To understand the current error when trying to use `conanrun.sh` with fish:

```
⋊> source conanrun.sh 
/private/tmp/foo/conanrunenv-release-armv8.sh (line 1): Unsupported use of '='. In fish, please use 'set script_folder "/private/tmp/foo"'.
script_folder="/private/tmp/foo"
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
from sourcing file /private/tmp/foo/conanrunenv-release-armv8.sh
	called on line 1 of file conanrun.sh
from sourcing file conanrun.sh
.: Error while reading file '/private/tmp/foo/conanrunenv-release-armv8.sh'
```

So, it's incompatible with Fish right now. Just to have an idea, `pyenv` generates both `activate` and `activate.fish`.

I'll check if we can do something compatible with both, because it's a really small script that we have.

uilianries:
Add new test to validate consuming a tool requirement using Fish shell. It should be able to load the tool without errors.

conan/tools/env/environment.py
```diff
@@ -19,16 +22,20 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
...
     # TODO: This implemantation is dirty, improve it
     for f in filenames:
         f = f if os.path.isabs(f) else os.path.join(env_folder, f)
         if f.lower().endswith(".sh"):
             if os.path.isfile(f) and "sh" in accept:
                 f = subsystem_path(subsystem, f)
                 shs.append(f)
+        elif f.lower().endswith(".fish"):
+            if os.path.isfile(f) and "fish" in accept:
+                f = subsystem_path(subsystem, f)
```
memsharded: I don't think we should be supporting fish at the moment in Windows subsystems, mostly used externally with ``win_bash``

conan/tools/env/environment.py
```diff
@@ -62,6 +73,10 @@ def environment_wrap_command(env_filenames, env_folder, cmd, subsystem=None,
     elif shs:
         launchers = " && ".join('. "{}"'.format(f) for f in shs)
         return '{} && {}'.format(launchers, cmd)
+    elif fishs:
+        launchers = " && ".join('. "{}"'.format(f) for f in fishs)
+        print(f"LAUNCHERS: {launchers}")
```
memsharded:
```suggestion
```

conan/tools/env/environment.py
```diff
@@ -520,12 +535,64 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+        script_content = Template(textwrap.dedent("""\
+            function {{ function_name }}
+                echo "echo Restoring environment"
+                {% for item in vars_unset %}
+                set -e {{ item }}
+                {% endfor %}
+                {% for item, value in vars_restore.items() %}
+                set -gx {{ item }} {{ value }}
+                {% endfor %}
+            end
```
memsharded: The restore should work restoring the variables that were defined at activation time, not the variables that were defined at the time of generating the script. It might be a bit more complicated, the activation needs to store the variables in memory or something too, so the restore can use them

conans/client/generators/__init__.py
```diff
@@ -167,6 +168,8 @@ def deactivates(filenames):
                 bats.append("%~dp0/"+path)
             elif env_script.endswith(".sh"):
                 shs.append(subsystem_path(subsystem, path))
+            elif env_script.endswith(".fish"):
+                fishs.append(subsystem_path(subsystem, path))
```
memsharded: Same as above, no subsystem for fish initially, IMHO

uilianries:
@memsharded Thank you for your early review. I just talked to @RubenRBS few minutes ago and found other cases to be testes using spaces and some alternatives for deactivation via memory function.

Most of these changes were made to track the creation of `conanbuild.sh` and how `conan_run` invokes with wrapper. I have to confess I spent few hours to understand the flow.

conan/tools/env/environment.py
```diff
@@ -520,12 +534,66 @@ def save_sh(self, file_location, generate_deactivate=True):
...
 
+    def save_fish(self, file_location):
+        """Save a fish script file with the environment variables defined in the Environment object.
+
+        It generates a function to deactivate the environment variables configured in the Environment object.
+
+        :param file_location: The path to the file to save the fish script.
+        """
+        filepath, filename = os.path.split(file_location)
+        function_name = f"deactivate_{Path(filename).stem}"
```
memsharded, line 550: Why the function_name is this one? Why not just the ``group`` with the random chars? Is there risk of collision of ``deactivate_xxxx`` function name?

conan/tools/env/environment.py
```diff
@@ -520,12 +534,66 @@ def save_sh(self, file_location, generate_deactivate=True):
...
+                {% for item in vars_define.keys() %}
+                if set -q conan_{{ group }}_{{ item }}
+                    set -gx {{ item }} "$conan_{{ group }}_{{ item }}"
+                end
+                {% endfor %}
+            end
+            {% for item, value in vars_define.items() %}
+            {% if item == "PATH" %}
+            set -g conan_{{ group }}_{{ item }} "${{ item }}"
+            fish_add_path -pg "{{ value }}"
```
memsharded: There are other variables that are paths too, like PYTHONPATH, and every other var marked as "path", shouldn't they do the same?

memsharded (commented):
Looks good! I think it is almost ready.

conan/tools/env/environment.py
uilianries: Unfortunately `fish_add_path` is a feature only to PATH: https://fishshell.com/docs/current/cmds/fish_add_path.html

I didn't find a similar feature for custom paths, but I'll check with Rubén tomorrow for insights.

conan/tools/env/environment.py
uilianries: It follow the common name used by other virtualenv like: `conanbuild.sh` -> `deactivate_conanbuild.sh`. So you can actually can call from your terminal directly, like:

```
source conanbuild.fish
...
deactivate_conanbuild
```

In Fish, functions are exported as a command, so you can run directly.

About the random to avoid collision: In case we have multiple dependencies add bin folder to PATH, we will need to undo the configuration after deactivating it. Usually it would be store in a deactivate script, but here we are using a variable. Actually, the random is just an idea, but could be the build folder name too, because is kind unique name.

uilianries:
I'm not sure about the logic that I added: In case a variable already exist, it will be prepend to that variable.

Pros: Preserves system variables and variables already configured by another Conan dependency.

Cons: If an user runs `conanbuild.fish` twice, all variables will be broken as they will be value duplicated.


I still need to fix some important points when manipulating paths, it's not working right now when needing to prepend.

conan/tools/env/environment.py
uilianries: Done, please, take a look in the test `test_transitive_tool_requires`. It generates a tool requirement, then a second recipe consumes it and runs an executable that prints an environment variable. All using Fish.

conans/test/functional/toolchains/env/test_virtualenv_fish.py
uilianries: Done. `test_prepend_path`

uilianries:
We still need to activate Fish in the CI, otherwise, we will not be able to validate it. 

It works in my machine™:

```
pytest -vvv conans/test/functional/toolchains/env/test_virtualenv_fish.py                                                                                                                                                 18:34:59
============================================================================================================================= test session starts ==============================================================================================================================
platform darwin -- Python 3.10.4, pytest-7.4.4, pluggy-1.4.0 -- /Users/uilian/Development/conan/conan-io/venv/bin/python
cachedir: .pytest_cache
rootdir: /Users/uilian/Development/conan/conan-io
configfile: pytest.ini
plugins: xdist-3.5.0
collected 11 items                                                                                                                                                                                                                                                             

conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_define_new_vars PASSED                                                                                                                                                                               [  9%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_prepend_path[/path/to/fake/folder] PASSED                                                                                                                                                            [ 18%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_prepend_path[/path/with space/to/folder] PASSED                                                                                                                                                      [ 27%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea del Toboso-True-True] PASSED                                                                                                                                       [ 36%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea del Toboso-True-False] PASSED                                                                                                                                      [ 45%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea del Toboso-False-True] PASSED                                                                                                                                      [ 54%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea del Toboso-False-False] PASSED                                                                                                                                     [ 63%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea-Del-Toboso-True-True] PASSED                                                                                                                                       [ 72%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea-Del-Toboso-True-False] PASSED                                                                                                                                      [ 81%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea-Del-Toboso-False-True] PASSED                                                                                                                                      [ 90%]
conans/test/functional/toolchains/env/test_virtualenv_fish.py::test_transitive_tool_requires[Dulcinea-Del-Toboso-False-False] PASSED                                                                                                                                     [100%]
```

uilianries:
Fish is now installed and configured with tests. Time for review!

conans/test/functional/toolchains/env/test_virtualenv_fish.py
```diff
@@ -0,0 +1,208 @@
...
+
+import pytest
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.test_files import temp_folder
+from conans.test.utils.tools import TestClient, load
+from conans.util.files import save
+
+# INFO: Fish only supports Cygwin and WSL https://github.com/fish-shell/fish-shell?tab=readme-ov-file#windows
+pytestmark = pytest.mark.skipif(platform.system() not in ("Darwin", "Linux"), reason="Fish is well supported only in Linux and MacOS")
```
memsharded: Why this syntax and not a normal @pytest.mark.skip?

conans/client/generators/__init__.py
```diff
@@ -179,6 +183,14 @@ def sh_content(files):
             save(os.path.join(conanfile.generators_folder, filename), sh_content(shs))
             save(os.path.join(conanfile.generators_folder, "deactivate_{}".format(filename)),
                  sh_content(deactivates(shs)))
+        if fishs:
+            def fish_content(files):
+                return ". " + " && . ".join('"{}"'.format(s) for s in files)
+            fishs = [it for it in fishs if os.path.exists(it)]
+            if fishs:
+                filename = "conan{}.fish".format(group)
```
memsharded, line 198: Here it seems it is missing the deactivation function (not script)?

conan/tools/env/environment.py
memsharded: Yes, having something deterministic would make more sense, I probably still didn't understand the issue.

conans/test/functional/toolchains/env/test_virtualenv_fish.py
uilianries: Because the entire file only work with Linux and MacOS. Otherwise, would need to apply for each test. It's a documented feature in pytest: https://docs.pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules

conans/client/generators/__init__.py
uilianries: Actually the deactivation is not an script, but a function exported by Fish in the environment. So we can call the deactivate_conanbuild, but there is no file associated. The function is created on-the-fly: https://github.com/conan-io/conan/pull/15503/files#diff-56b36c23589a66b0bd803c545703684c74400cbb7bd8538b8949ed84613e3fc7R550

AbrilRBS:
We're closing this until there's push from the community to support Fish natively, as the current risk/reward ratio for this feature is not something we're comfortable with currently.

We don't expect Fish users to be unaware of the pitfalls of using it, it's expected that we need to switch back to another terminal to source files like the `conanbuild.sh` etc.

If this is a feature you'd like to have, please do let us know so we can properly prioritize this again, but for now I'm closing this as not planned for the foreseeable future

jasal82:
I'd really like to have this feature, I'm using fish exclusively.

kambala-decapitator:
I'm also a fish adept :) But tbh so far I've sourced build env script in terminal only once.

aagor:
I would like to propose another solution, that makes that topic easier and is more maintainable and future-proof, as the current approach requires support for every shell specifically:

You could add a command like `conan env --env=build fish` (the default), `conan env --env=run fish`.
These commands need to take similar arguments like `conan build` to be able to pass settings/options via CLI (in order to get the correct environment variables needed).

The command can set the environment variables from Python via `os.environ`/`os.putenv`. Then execute the given command via one of the exec family methods.

This has a few benefits:
* Every shell is supported without special code for each.
* It's possible to call commands directly with the correct environment without intermediary shell.
  Which is sometimes very nice, because shell means ugly escaping of different stuff.
  Even more important, different shells, especially on different OS, have different escaping rules.
* You don't need custom deactivate logic. Just exit the shell and you're back in **exactly** the same state as before.

Should I create a seperate issue for that suggestion?

memsharded:
> You could add a command like conan env --env=build fish (the default), conan env --env=run fish

It is not possible to permanently modify the current shell environment via a Python subprocess execution, so it is impossible to do it with a Conan command. This is not allowed by the OSs. Environment of the current process can only be modified executing in the current shell process, not a subprocess one. This happens in Windows cmd naturally while executing ``.bat`` files, and this is the reason why in Linux shells it is necessary to ``source``/``.`` the environment file instead of executing it, because executing it is also a subprocess and that cannot modify the parent process environment.

aagor:
> It is not possible to permanently modify the current shell environment via a Python subprocess execution...

Yeah, I know this and your statement is 100% correct.

However, why do I need to modify the environment of the **current** shell?
Is there any reason I shouldn't use a separate shell process?
I can just use another instance (process) of the same shell with appropriate environment, which, as I tried to describe above, is also nice, because you get easily rid of such a modified by exiting this shell and returning to the main shell (which, as you correctly noted, still has the unmodified environment).

I mean, this is exactly what `conan build` internally does. It creates a subprocess, with appropriate environment, and uses that to execute the build commands.
If I want to execute commands manually with `conan install`, it seems reasonable to be able to do the same.

Basically, this is exactly how `sudo` works. You don't give privileges (= env variables) to the current shell, a new shell is created with that privileges (= env variables). If you're done you exit that privileged (= build/runenv) shell.
This has proven since many years.

And I'm not proposing to drop the current thing, just to add that as a generic approach for every currently not supported shell.

memsharded:
Uhmm, I see, interesting.

I think it might be a bit too much, but it would be interesting to explore the idea, this has been suggested in the past but never moved forward. Yes, please, open a ticket for it for discussion.

> I mean, this is exactly what conan build internally does. It creates a subprocess, with appropriate environment, and uses that to execute the build commands.

Conan commands and internal subprocess start and finish, and the Conan process is also terminated.
Implementing something like the proposed approach requires a different flow, the first ``conan xxxx`` command subprocess will be kept alive, launch a new shell with the new environment as a subprocess. There might be challenges with this, as the current running Conan process might hold some resources, but as long as the subprocess is used exclusively for building (not for launching Conan again, that would be re-entrancy forbidden in the core guidelines), maybe it can be done, lets explore that.

aagor:
> Yes, please, open a ticket for it for discussion.

I will create one (but not next week).

> ..., the first conan xxxx command subprocess will be kept alive, ...

You can use exec and terminate the conan process.
However I have a similar idea, how this could be solved, which I will explain when creating the feature request.

Thanks for your comments already!

PR #15514, opened 2024-01-24
Log Git tool commands when running in verbose mode

### LINKED ISSUE #7029: [feature] Print git commands when print_run_commands = True
It would be nice to have `git` commands from the `tools.Git` helper be printed out if the `print_run_commands` variable is set to `True`. Could even be a new configuration option if desired

AbrilRBS:
Changelog: Feature: Log Git tool commands when running in verbose mode.
Docs: Omit

Closes #7029


franramirez688 (approved):

conan/tools/scm/git.py
```diff
@@ -24,6 +24,7 @@ def run(self, cmd):
 
         :return: The console output of the command.
         """
+        self._conanfile.output.verbose(f"Running git {cmd}")
```
memsharded: The ``Conanfile.run()`` already has some output, with a specific color, etc, why not the same format and color?

TODO: Is there any risk that the `cmd` will contain the URL that includes credentials?

memsharded (approved):
I like the idea of hiding the full URL, no risks this way.

conan/tools/scm/git.py
```diff
@@ -31,12 +32,14 @@ def __init__(self, conanfile, folder=".", excluded=None):
...
 
-    def run(self, cmd):
+    def run(self, cmd, hidden_output=None):
         """
         Executes ``git <cmd>``
 
         :return: The console output of the command.
         """
+        print_cmd = cmd if hidden_output is None else cmd.replace(hidden_output, "<hidden>")
+        self._conanfile.output.verbose(f"RUN: git {print_cmd}", fg=Color.BRIGHT_BLUE)
```
memsharded: Why not ``info``? Other tools like ``cmake.build()``, using ``self.run()`` will display the CMD in info

PR #15535, opened 2024-01-25
Add pattern-to-pattern replace test case

AbrilRBS:
Changelog: Omit
Docs: https://github.com/conan-io/docs/pull/3530

As response to Slack message, we probably want to document something like this?


memsharded (approved):
Looks good any special risk I am not seeing?

AbrilRBS:
@memsharded I don't think so - merging now as-is, will add a docu example as a follow-up

PR #15552, opened 2024-01-29
Fix msvc profile detect test

juansblanco:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conans/test/functional/test_profile_detect_api.py
+++ b/conans/test/functional/test_profile_detect_api.py
@@ -18,7 +18,6 @@ class TestProfileDetectAPI:
             compiler={{compiler}}
             compiler.version={{detect_api.default_compiler_version(compiler, version)}}
             compiler.runtime={{runtime}}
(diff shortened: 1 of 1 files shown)


AbrilRBS:
Found while asking @juansblanco to run some windows test for me - I added this back then without locally testing

conans/test/functional/test_profile_detect_api.py
```diff
@@ -18,7 +18,6 @@ def test_profile_detect_compiler(self):
             compiler={{compiler}}
             compiler.version={{detect_api.default_compiler_version(compiler, version)}}
             compiler.runtime={{runtime}}
-            compiler.libcxx={{detect_api.detect_libcxx(compiler, version, compiler_exe)}}
```
memsharded, line 21: Why this was not failing before in CI? I thought CI was running VS 17 already

AbrilRBS: Investigating, curious nontheless

PR #15600, opened 2024-02-05
Refactor `DownloadApi`

AbrilRBS:
Changelog: Omit
Docs: Omit

### DIFF
--- a/conan/api/subapi/download.py
+++ b/conan/api/subapi/download.py
@@ -49,6 +49,7 @@ class DownloadAPI:
 
     def package(self, pref: PkgReference, remote: Remote, metadata: Optional[List[str]] = None):
         """Download the package specified in the pref from the remote.
+        The recipe for this package binary must already exist in the cache.
         If the package is already in the cache it will be skipped,
         but the specified metadata will be downloaded."""
         output = ConanOutput()
--- a/conan/api/subapi/download.py
+++ b/conan/api/subapi/download.py
@@ -1,3 +1,6 @@
+import time
+from multiprocessing.pool import ThreadPool
+
 from conan.api.model import Remote
 from conan.api.output import ConanOutput
 from conan.internal.conan_app import ConanApp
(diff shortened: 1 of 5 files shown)


conans/test/integration/command/download/download_parallel_test.py
```diff
@@ -11,15 +11,14 @@ def test_basic_parallel_download():
 
     package_ids = []
     for i in range(counter):
-        client.run("create . --name=pkg --version=0.1 --user=user --channel=testing -o pkg/*:myoption=%s" % i)
-        package_id = client.created_package_id("pkg/0.1@user/testing")
-        package_ids.append(package_id)
+        client.run(f"create . --name=pkg{i} --version=0.1 --user=user --channel=testing -o pkg{i}/*:myoption={i}")
+        package_id = client.created_package_id(f"pkg{i}/0.1@user/testing")
+        package_ids.append((i, package_id))
```
memsharded: This test was covering multiple binaries for the same pkg/version. With this change, it will test multiple pkgi/version with one single binary for each one. Maybe we want to cover both cases, adding some binaries for the same package name too.

conan/api/subapi/download.py
```diff
@@ -40,7 +47,10 @@ def recipe(self, ref: RecipeReference, remote: Remote, metadata=None):
         app.remote_manager.get_recipe_sources(ref, recipe_layout, remote)
         return True
 
-    def package(self, pref: PkgReference, remote: Remote, metadata=None):
+    def package(self, pref: PkgReference, remote: Remote, metadata: Optional[List[str]] = None):
+        """Download the package specified in the pref from the remote.
```
memsharded, line 51:
```suggestion
        """Download the package specified in the pref from the remote.
        The recipe for this package binary must already exist in the cache.
```

memsharded (approved):

czoido (approved):

PR #15601, opened 2024-02-05
Check sha256 of local back source cache files and improve `.dirty` handling on backup upload 

AbrilRBS:
Changelog: Fix: Improve handling of `.dirty` download files when uploading backup sources.
Docs: Omit

### DIFF
--- a/conans/client/downloaders/caching_file_downloader.py
+++ b/conans/client/downloaders/caching_file_downloader.py
@@ -61,6 +61,7 @@ class SourcesCachingDownloader:
             remove_if_dirty(cached_path)
 
             if os.path.exists(cached_path):
+                self._file_downloader.check_checksum(cached_path, md5, sha1, sha256)
                 self._output.info(f"Source {urls} retrieved from local download cache")
             else:
                 with set_dirty_context_manager(cached_path):
--- a/conans/client/downloaders/caching_file_downloader.py
+++ b/conans/client/downloaders/caching_file_downloader.py
@@ -61,7 +61,6 @@ class SourcesCachingDownloader:
             remove_if_dirty(cached_path)
 
             if os.path.exists(cached_path):
-                self._file_downloader.check_checksum(cached_path, md5, sha1, sha256)
                 self._output.info(f"Source {urls} retrieved from local download cache")
             else:
                 with set_dirty_context_manager(cached_path):
--- a/conans/client/downloaders/download_cache.py
+++ b/conans/client/downloaders/download_cache.py
@@ -5,7 +5,7 @@ from threading import Lock
 
 from conans.errors import ConanException
 from conans.util.dates import timestamp_now
-from conans.util.files import load, save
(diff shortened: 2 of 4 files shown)


conans/client/downloaders/caching_file_downloader.py
```diff
@@ -61,6 +61,7 @@ def _caching_download(self, urls, file_path,
             remove_if_dirty(cached_path)
 
             if os.path.exists(cached_path):
+                self._file_downloader.check_checksum(cached_path, md5, sha1, sha256)
```
memsharded: This won't cover all cases of ``.dirty``, only controlled ones. There can be some cases where things will crash and still the ``.dirty`` file will be there.

AbrilRBS: Yup! That's why I was not confident in pushing this without being a draft, I still need to check those cases :)

Thanks for the heads up!

conans/client/downloaders/download_cache.py
```diff
@@ -73,22 +73,33 @@ def should_upload_sources(package):
...
 
-        files_to_upload = []
+        path_backups_contents = []
 
+        dirty_ext = ".dirty"
         for path in os.listdir(path_backups):
+            if remove_if_dirty(os.path.join(path_backups, path)):
+                continue
+            if path.endswith(dirty_ext):
+                # TODO: Clear the dirty file marker if it does not have a matching downloaded file
```
AbrilRBS, line 83: When we rethink how the `.dirty` files are handled, we won't need to check here nor do anything special

conans/test/integration/cache/backup_sources_test.py
```diff
@@ -677,3 +678,111 @@ def source(self):
...
+            # This check should go away once we refactor dirty handling
+            assert os.path.exists(os.path.join(self.download_cache_folder, "s", f"{sha256}.dirty"))
+
+        if upload:
+            self.client.run("cache backup-upload")
+        else:
+            self.client.run("source .")
+        assert f"{sha256} is dirty, removing it" in self.client.out
+
+    @pytest.mark.skip("Recover when .dirty files are properly handled")
```
AbrilRBS, line 752: To be recovered after https://github.com/conan-io/conan/issues/15612

conans/util/files.py
```diff
@@ -44,6 +44,8 @@ def remove_if_dirty(item):
             else:
                 rmdir(item)
         clean_dirty(item)
+        return True
+    return False
```
AbrilRBS, line 48: Let's the calling method know if a removal happened, useful when skipping dirty files in the backup upload file gathering

AbrilRBS (commented):
Some comments

conans/util/files.py
memsharded: Sure, but this is not returning removal, just that the thing was dirty.

memsharded (approved):

PR #15613, opened 2024-02-06
Add format json option to conan cache path

### LINKED ISSUE #15599: [feature] add `--format=json` to conan cache path
### What is your suggestion?

I would like for conan to be able to give me a machine readable file, rather than having to pipe the output of the `conan cache path` command.

There's nothing wrong with piping, but this makes life slightly easier on a CI system that needs to support both bash syntax and bat/cmd syntax on Windows. 

juansblanco:
Changelog: Feature: Add `--format=json` option to `conan cache path`.
Docs: Omit

Added `--format=json` option to the `conan cache path` command that returns a json as: 
`{"cache_path": "C:\\Users\\user\\.conan2\\p\\zlib2cad50ee1214e\\e"}`

Closes: https://github.com/conan-io/conan/issues/15599

### DIFF
--- a/conans/test/integration/command_v2/test_cache_path.py
+++ b/conans/test/integration/command_v2/test_cache_path.py
@@ -1,3 +1,6 @@
+import json
+import os
+
 import pytest
 
 from conans.test.assets.genconanfile import GenConanfile
@@ -111,3 +114,14 @@ def test_cache_path_does_not_exist_folder():
     client.run(f"install --requires mypkg/0.1")
     client.run(f"cache path {pref} --folder build", assert_error=True)
     assert f"ERROR: 'build' folder does not exist for the reference {pref}" in client.out
+
+def test_cache_path_output_json():
+    client = TestClient()
+    conanfile = GenConanfile("mypkg", "0.1")
+    client.save({"conanfile.py": conanfile})
+    client.run("export .")
+    layout = client.exported_layout()
+    client.run("cache path mypkg/0.1 --format=json", redirect_stdout="out.json")
+    output = json.loads(client.load("out.json"))
+    cache_path = os.path.join(layout.base_folder, "e")
+    assert output == {"cache_path": f"{cache_path}"}
--- a/conans/test/integration/command_v2/test_cache_path.py
+++ b/conans/test/integration/command_v2/test_cache_path.py
@@ -123,5 +123,4 @@ def test_cache_path_output_json():
     layout = client.exported_layout()
     client.run("cache path mypkg/0.1 --format=json", redirect_stdout="out.json")
     output = json.loads(client.load("out.json"))
-    cache_path = os.path.join(layout.base_folder, "e")
-    assert output == {"cache_path": f"{cache_path}"}
+    assert output == {"cache_path": os.path.join(layout.base_folder, "e")}
--- a/conans/test/integration/command_v2/test_cache_path.py
+++ b/conans/test/integration/command_v2/test_cache_path.py
@@ -121,6 +121,6 @@ def test_cache_path_output_json():
     client.save({"conanfile.py": conanfile})
     client.run("export .")
     layout = client.exported_layout()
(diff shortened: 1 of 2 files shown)


conans/test/integration/command_v2/test_cache_path.py
```diff
@@ -111,3 +114,14 @@ def test_cache_path_does_not_exist_folder():
...
+def test_cache_path_output_json():
+    client = TestClient()
+    conanfile = GenConanfile("mypkg", "0.1")
+    client.save({"conanfile.py": conanfile})
+    client.run("export .")
+    layout = client.exported_layout()
+    client.run("cache path mypkg/0.1 --format=json", redirect_stdout="out.json")
+    output = json.loads(client.load("out.json"))
+    cache_path = os.path.join(layout.base_folder, "e")
+    assert output == {"cache_path": f"{cache_path}"}
```
AbrilRBS:
```suggestion
    assert output == {"cache_path": os.path.join(layout.base_folder, "e")}
```

conans/test/integration/command_v2/test_cache_path.py
```diff
@@ -111,3 +114,14 @@ def test_cache_path_does_not_exist_folder():
...
     assert f"ERROR: 'build' folder does not exist for the reference {pref}" in client.out
+
+def test_cache_path_output_json():
+    client = TestClient()
+    conanfile = GenConanfile("mypkg", "0.1")
+    client.save({"conanfile.py": conanfile})
+    client.run("export .")
+    layout = client.exported_layout()
+    client.run("cache path mypkg/0.1 --format=json", redirect_stdout="out.json")
+    output = json.loads(client.load("out.json"))
```
memsharded:
```suggestion
    client.run("cache path mypkg/0.1 --format=json")
    output = json.loads(client.stdout)
```

memsharded (approved):

jcar87:
LGTM as it is, but I wonder if we might want something slightly more elaborate?

At the moment, the following two invocationS:
```
conan cache path foo/1.0.0 --format=json
```
and
```
conan cache path foo/1.0.0 --folder=metadata --format=json
```

would both return a json have the key `cache_path` in the returned json, but pointing to a different folder. This mirrors the behaviour of the regular CLI output, but it makes me wonder if we might want something more "rich"?

e.g.
```
{
    "cache_path": "C:\\Users\\user\\.conan2\\p\\zlib2cad50ee1214e\\e", 
    "folder": "exports", 
    "reference": "foo/1.0.0"
}
```

and
```
   "cache_path": "xxx",
   "folder: "metadata", 
   "reference: "foo/1.0.0"
```

or even a richer
```
{
 "ref": "foo/1.0.0#rrev1", 
 "folders": 
     { 
         "exports": "path1",
         "metadata": "path2"
         "export_source": "path3",
         // etc...
      }
}
```

Just an idea, but perhaps if we decide to do something richer in the future, leave the json today in a way that can be extended without breaking?

memsharded:
I think it should be good enough with the command to return 1 path only because all the use cases I have found so far it is to do operations on 1 folder, like the metadata one, but it is unlikely to need several folders, and it that is the case they can do several ``conan cache path`` calls, not a big issue. So the 
```
{
    "cache_path": "C:\\Users\\user\\.conan2\\p\\zlib2cad50ee1214e\\e", 
    "folder": "exports", 
    "reference": "foo/1.0.0"
}
```
Should be future-proof enough, and maybe if we want a multiple-paths command and accept multiple ``--folder`` arguments, the output format can change anyway to accommodate for that, as this would be an opt-in.

Because for the basic case I am concerned that you'd need to use both the argumetn ``--folder=xxxx`` and then yet again the same in the ``jq`` query or to parse the json with ``data["folders"][xxxxx]`` to get the value you requested as an argument?

So I think might be good as-is

jcar87 (approved):

jcar87:
Makes sense! Merging :) Thanks @juansblanco !!

PR #15619, opened 2024-02-07
Docstrings

franramirez688:
Changelog: omit
Docs: omit

### DIFF
--- a/conans/model/conf.py
+++ b/conans/model/conf.py
@@ -57,8 +57,8 @@ BUILT_IN_CONFS = {
     "tools.build:download_source": "Force download of sources for every package",
     "tools.build:jobs": "Default compile jobs number -jX Ninja, Make, /MP VS (default: max CPUs)",
     "tools.build:sysroot": "Pass the --sysroot=<tools.build:sysroot> flag if available. (None by default)",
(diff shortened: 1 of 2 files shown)


memsharded (approved):
But the ``conf`` was not just a docstring, how test could pass without it? I think because it was just a unit test that is not running validation of the ``conf`` inputs

franramirez688:
> But the `conf` was not just a docstring, how test could pass without it? I think because it was just a unit test that is not running validation of the `conf` inputs

It's only an order change. It was completely right.

conans/model/conf.py
```diff
@@ -108,7 +108,6 @@
     "tools.apple:enable_visibility": "(boolean) Enable/Disable Visibility Apple Clang flags",
     "tools.env.virtualenv:powershell": "If it is set to True it will generate powershell launchers if os=Windows",
     # Compilers/Flags configurations
-    "tools.build.cross_building:cross_build": "Decides whether cross-building or not regardless of arch/OS settings. Used by 'conan.tools.build.cross_building' function.",
```
franramirez688, line 111: @memsharded It was here. Now, it's above. Otherwise, the test would fail.

memsharded: Oh, you are right! Sorry I didn't see it.

Issue #15579, opened 2024-02-01
[question/bug] conan install takes more than 90 seconds


Ariox41:
### What is your question?

We've been using conan 2 for a while and it worked quite well. Our modules depend on boost and qt, `conan install` was executed in about 20 seconds. However, today I added a module that depends on `libsndfile` and `openal-soft`, and after that the `conan install` for the final application increased to 90 seconds. This clearly exceeds the time that developers are willing to put up with.

This does not apply to most of our modules, so we are still ready to put up with it. But I did a little research anyway:
- In total, the application has about 220 dependencies, 130 of them from conancenter.
- Profiling shows that the DepsGraph.add_edge() method takes a significant amount of time.
- After deleting the `assert src in self.nodes and dst in self.nodes` line, the `conan install` time is reduced to 60 seconds, which is significant.
- When I try to run the `python3 -O ~/.local/bin/conan install .` command instead, I get an error (ubuntu 22.04, python3.10, conan 2.0.17):
  ```shell
  ...
  File "~/.local/lib/python3.10/site-packages/conan/internal/cache/db/recipes_table.py", line 91, in all_references
      with self.db_connection() as conn:
    File "/usr/lib/python3.10/contextlib.py", line 142, in __exit__
      next(self.gen)
    File "~/.local/lib/python3.10/site-packages/conan/internal/cache/db/table.py", line 32, in db_connection
      self._lock.release()
  RuntimeError: release unlocked lock
  ```
  It looks like a serious bug in conan.
- When trying to use `pypy3`, I get a login request, and 'canon remote login` does not solve the problem. Conan doesn't seem to work with pypy. This may also be a consequence of errors in conan.
- I also noticed that conan recipe methods are called hundreds, if not thousands of times. There doesn't seem to be any caching. Could this be a problem in the recipes themselves? Can `self.requires("...", force=True)` in some recipes lead to a significant slowdown?

Thus, at least part of the problem is on the conan side.

Has anyone faced the same problem? It seems like 200 dependencies shouldn't be uncommon, considering that most of them are external dependencies.

### Have you read the CONTRIBUTING guide?

- [X] I've read the CONTRIBUTING guide

memsharded:
Hi @Ariox41 

Thanks for your report.

It seems there are some different things here in this issue, but overall it would be related to performance.
[...]

PR #15623, opened 2024-02-07
some low hanging fruits for optimizing speed for large graphs

memsharded:
Changelog: Omit
Docs: Omit

For https://github.com/conan-io/conan/issues/15579

Pure refactors for speed.
Achieved a 15% speed up for the test graph of 40*40 = 1600 nodes dependency graph, reduced from 7.7s to 6.1 approx on my machine

### DIFF
--- a/conans/test/performance/test_large_graph.py
+++ b/conans/test/performance/test_large_graph.py
@@ -1,0 +1,43 @@
+import cProfile
+import json
+import pstats
+import time
+from pstats import SortKey
+
+from conans.test.assets.genconanfile import GenConanfile
+from conans.test.utils.tools import TestClient
+
+
+def test_large_graph():
+    c = TestClient(cache_folder="T:/mycache")
+    num_test = 40
+    num_pkgs = 40
+
+    """for i in range(num_test):
+        conanfile = GenConanfile(f"test{i}", "0.1")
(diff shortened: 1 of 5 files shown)


AbrilRBS (approved):

conans/test/performance/test_large_graph.py
```diff
@@ -0,0 +1,43 @@
...
+    t = time.time()
+    pr = cProfile.Profile()
+    pr.enable()
+    c.run(f"install --requires=pkg{num_pkgs - 1}/0.1")
+    pr.disable()
+    print(time.time()-t)
+
+    sortby = SortKey.CUMULATIVE
+    ps = pstats.Stats(pr).sort_stats(sortby)
+    ps.print_stats()
```
AbrilRBS, line 40: I like this. I used to do it manually by hand, but this seems so much easier having a test already written!

AbrilRBS (approved):

conans/test/performance/test_large_graph.py
memsharded: It is still very dirty, because the generation of the test packages takes way longer than the test itself, so I cache it. Definitely not for CI, but having it there to play with it is useful

PR #15630, opened 2024-02-08
proposal for cli-commands api

memsharded:
Changelog: Feature: New ``CommandAPI`` subapi in the ``ConanAPI`` that allows calling other commands.
Docs: https://github.com/conan-io/docs/pull/3635

### DIFF
--- a/conans/test/integration/command_v2/custom_commands_test.py
+++ b/conans/test/integration/command_v2/custom_commands_test.py
@@ -2,6 +2,8 @@ import json
 import os
 import textwrap
 
+import pytest
+
 from conans.test.assets.genconanfile import GenConanfile
 from conans.test.utils.test_files import temp_folder
 from conans.test.utils.tools import TestClient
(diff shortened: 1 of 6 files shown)


franramirez688 (commented):
I think it looks really simple 👏

jcar87:
I already have a few custom commands in mind if this is possible!! :D

czoido (approved):

conans/test/integration/command_v2/custom_commands_test.py
```diff
@@ -286,3 +288,84 @@ def hello(conan_api, parser, *args, **kwargs):
...
+            import json
+            from conan.cli.command import conan_command
+            from conan.cli.formatters.graph import format_graph_json
+
+            @conan_command(group="custom commands", formatters={"json": format_graph_json})
+            def mycommand(conan_api, parser, *args, **kwargs):
+                \"""
+                mycommand help
+                \"""
+                result = conan_api.command.run("create", ".", "--version=1.0.0")
```
jcar87: this fails if expressed this way, but passes if we call it with the arguments as a list
```
result = conan_api.command.run("create", [".", "--version=1.0.0"])
```

which seems to be a different syntax than the test above that passes them as separate arguments

jcar87 (approved):
LGTM!

AbrilRBS (approved):
Can't check right now because the question came to my mind just after I closed the laptop for the day, but I assume this works just fine for custom commands, right?

Otherwise looks good and feels nice to use! 👍

memsharded:
> Can't check right now because the question came to my mind just after I closed the laptop for the day, but I assume this works just fine for custom commands, right?

@RubenRBS you mean that one custom command can also call other custom command? Yes, it should, I can add a test to cover this.

AbrilRBS (approved):

czoido (approved):

PR #15635, opened 2024-02-09
Add test showing that recipes can override their pyreq methods just fine

AbrilRBS:
Changelog: Omit
Docs: Omit

As answer to a CppLang slack question. I'm not a super fan of the approach, but fits their usage :)

### DIFF
--- a/conans/test/integration/py_requires/python_requires_test.py
+++ b/conans/test/integration/py_requires/python_requires_test.py
@@ -1246,3 +1246,37 @@ def test_export_pkg():
     c.run("list *")
     assert "WARN: There are no matching recipe references" in c.out
     assert "pytool/0.1" not in c.out
+
+
+def test_py_requires_override_method():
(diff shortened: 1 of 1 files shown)


memsharded (approved):

memsharded (commented):
While I am fine with this, I think we already have tests that show this? like
```python
    def test_reuse_super(self):
        client = TestClient(default_server_user=True)
        self._define_base(client)
        reuse = textwrap.dedent("""
               from conan import ConanFile
               class PkgTest(ConanFile):
                   python_requires = "base/1.1@user/testing"
                   python_requires_extend = "base.MyConanfileBase"

                   def source(self):
                       super().source()
                       self.output.info("MY OWN SOURCE")
               """)
```?

AbrilRBS:
Yes, I just wanted to show the user that the specific of overriding methods still worked fine with Conan's syntax :)

PR #15664, opened 2024-02-13
Switched to cc pattern in compiler detection

### LINKED ISSUE #14575: Conan fails to detect the compiler although `CC` and `CXX` are set
### Environment details

* Operating System+version: ubuntu-22.04
* Compiler+version: 13.1.0
* Conan version: 2.0.9
* Python version: 3.11.4

### Steps to reproduce

I am creating a Conda package of my software, as part of a Github C/I workflow. My software depends on other software (e.g.: Boost). For dependencies for which there isn't a Conda package available I use Conan to install them.

Here is the idea, scattered over the various files (`.github/workflows/linux-conda.yml`, `environment/conda/build.sh`):
- Create a Conda environment
- Install the boa Conda package builder and execute: `conda mambabuild environment/conda --override-channels --channel conda-forge`

Then `build.sh` is executed which includes:

```bash
conan profile detect                                                                                             
conan install . \                                                           
    --profile=default \                                                                                          
    --settings:host compiler.cppstd=17 \                                                                         
    --settings:build compiler.cppstd=17 \                                                                        
    --settings:build build_type=Release \                                                                        
    --build=missing \                                                                                            
    --output-folder=build
```

This results in an error (see log below). Conan is not able to correctly detect the compiler. The `CC` and `CXX` variables are set, so I would expect that those are used to detect the compiler.

I tried passing in the compiler explicitly, when calling `conan install`, using `--setting compiler=$CXX`, but that failed with `ERROR: 'settings.compiler' value not defined`. Maybe because the whole compiler section is missing from the profile?

I think this is a bug, given that the compiler is installed and pointed to by the standard variables. If not, then my question is how to get Conan to detect the compiler correctly in this situation.

I use the same conan commands to install Conan packages outside of the Conda build environment (on Linux, macOS, Windows), and those work fine. It seems that the Conda build environment is confusing Conan. Maybe because of the atypical setting of `CC` and `CXX`?

### Logs

```
2023-08-24T16:10:09.7983374Z CC and CXX: /usr/share/miniconda/envs/test/conda-bld/lue_1692893106258/_build_env/bin/x86_64-conda-linux-gnu-cc, /usr/share/miniconda/envs/test/conda-bld/lue_1692893106258/_build_env/bin/x86_64-conda-linux-gnu-c++ 
2023-08-24T16:10:09.7984827Z ERROR: Not able to automatically detect '/usr/share/miniconda/envs/test/conda-bld/lue_1692893106258/_build_env/bin/x86_64-conda-linux-gnu-cc' version
2023-08-24T16:10:09.7985898Z No compiler was detected (one may not be needed)
2023-08-24T16:10:09.7986712Z WARN: This profile is a guess of your environment, please check it.
2023-08-24T16:10:09.7987819Z WARN: The output of this command is not guaranteed to be stable and can change in future Conan versions.
2023-08-24T16:10:09.7988301Z WARN: Use your own profile files for stability.
2023-08-24T16:10:09.7988673Z Saving detected profile to /home/runner/.conan2/profiles/default
2023-08-24T16:10:09.8382941Z + conan install . --profile=default --settings:host compiler.cppstd=17 --settings:build compiler.cppstd=17 --settings:build build_type=Release --build=missing --output-folder=build
2023-08-24T16:10:10.1836158Z ERROR: 'settings.compiler' value not defined
```

Somewhat later on in the log:

```
2023-08-24T16:10:10.3898835Z Detected profile:
2023-08-24T16:10:10.3899230Z [settings]
2023-08-24T16:10:10.3899457Z arch=x86_64
2023-08-24T16:10:10.3899668Z build_type=Release
2023-08-24T16:10:10.3900101Z os=Linux
```

hansenms:
Changelog: Fix: Fix gcc detection in conda environments.
Docs: Omit

In a conda environment, the `CC` and `CXX` variables point to /opt/conda/envs/<env name>/bin/x86_64-conda-linux-gnu-cc and /opt/conda/envs/<env name>/bin/x86_64-conda-linux-gnu-c++ and the `command` searched for pattern ends up being /opt/conda/envs/<env name>/bin/x86_64-conda-linux-gnu-cc and thus `_gcc_compiler(command)` is never called.

Fixes #14575 

### DIFF
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -276,7 +276,7 @@ def detect_default_compiler():
         command = cc or cxx
         if "clang" in command.lower():
             return _clang_compiler(command)
-        if "gcc" in command or "g++" in command or "c++" in command:
+        if "cc" in command or "g++" in command or "c++" in command:
             gcc, gcc_version, compiler_exe = _gcc_compiler(command)
             if platform.system() == "Darwin" and gcc is None:
                 output.error("%s detected as a frontend using apple-clang. "
--- a/conan/internal/api/detect_api.py
+++ b/conan/internal/api/detect_api.py
@@ -276,7 +276,7 @@ def detect_default_compiler():
         command = cc or cxx
         if "clang" in command.lower():
             return _clang_compiler(command)
-        if "cc" in command or "g++" in command or "c++" in command:
+        if "gnu-cc" in command or "gcc" in command or "g++" in command or "c++" in command:
             gcc, gcc_version, compiler_exe = _gcc_compiler(command)
             if platform.system() == "Darwin" and gcc is None:
                 output.error("%s detected as a frontend using apple-clang. "


conan/internal/api/detect_api.py
```diff
@@ -276,7 +276,7 @@ def detect_default_compiler():
         command = cc or cxx
         if "clang" in command.lower():
             return _clang_compiler(command)
-        if "gcc" in command or "g++" in command or "c++" in command:
+        if "cc" in command or "g++" in command or "c++" in command:
```
memsharded: I think this can break some detections of compilers like ``icc``, or ``qcc`` that would suddenly start to be detected as ``gcc`` and that could be breaking.

To solve the issue for Conda, I think it would be better to add ``gnu-c++`` and ``gnu-cc`` to the checks, that would be good enough to detect the Conda one without the risk of affecting other compilers.

memsharded (commented):
Thanks very much for the contribution @hansenms 

Please have a look to the comment. 2.1 is about to be branched, if ready it could be in, otherwise might need to wait until next release.

conan/internal/api/detect_api.py
hansenms: Makes sense. I think adding `gnu-cc` should do the trick since `c++` would already hit `gnu-c++`. I have made that change.

memsharded (approved):
Thanks for the contribution and the review!

